Issue #7436, opened 2025-03-23
Tell the agent the current date


neubig:
**What problem or use case are you trying to solve?**

Occasionally the agent says things like "2025 is in the future" and it fails to solve a task that requires it to get info from 2025.

**Describe the UX of the solution you'd like**

We should tell the agent the time/date at the start of the agent trajectory.

malhotra5:
Completed by #7509

PR #7509, opened 2025-03-25
[Feat]: Tell the agent the current date

malhotra5:

Agent is created with knowledge of current date 

---

This PR attaches the current date and time to the `RuntimeInfo` object. We may want to, in the future, update the date and time as the conversation ages. As per the issue, however, we're looking to only tell the agent the time/date at the start of the agent trajectory for now. 

image

---
#7436

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:6aa5732-nikolaik   --name openhands-app-6aa5732   docker.all-hands.dev/all-hands-ai/openhands:6aa5732
```

### DIFF
--- a/openhands/agenthub/codeact_agent/prompts/additional_info.j2
+++ b/openhands/agenthub/codeact_agent/prompts/additional_info.j2
@@ -24,8 +24,8 @@ For example, if you are using vite.config.js, you should set server.host and ser
 {% if runtime_info.additional_agent_instructions %}
 {{ runtime_info.additional_agent_instructions }}
 {% endif %}
-{% if runtime_info.date or runtime_info.utc_time %}
-Today's datetime is {{ runtime_info.date }} at {{ runtime_info.utc_time }} (UTC).
+{% if runtime_info.date %}
+Today's date is {{ runtime_info.date }} (UTC).
 {% endif %}
 </RUNTIME_INFORMATION>
 {% endif %}
--- a/openhands/agenthub/codeact_agent/prompts/additional_info.j2
+++ b/openhands/agenthub/codeact_agent/prompts/additional_info.j2
@@ -24,5 +24,8 @@ For example, if you are using vite.config.js, you should set server.host and ser
 {% if runtime_info.additional_agent_instructions %}
 {{ runtime_info.additional_agent_instructions }}
 {% endif %}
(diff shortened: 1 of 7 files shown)


xingyaoww (commented):
At some point, i was actually thinking if we should just ask the agent to use `date` in bash for these info? Or simply put these info into PS1 of bash command

I imagine this could get stale quickly if we resume the runtime after 1 day..

xingyaoww:
I imagine we can probably do them here?
https://github.com/All-Hands-AI/OpenHands/blob/500e09f12b106ccc7c23cb3053f8df0ba0d6ba69/openhands/events/observation/commands.py#L156-L164

malhotra5:
Hmm yeah including it in PS1 bash commands would be helpful. 
Currently the agent is able to `date` in the terminal to get an output like the following 

```
Tue Mar 25 20:59:08 UTC 2025
```

However I think this linked issue is currently addressing the problem where the agent is first initialized with without having run any bash commands, and thus is unaware of the time unless explicitly prompted to.

openhands/agenthub/codeact_agent/prompts/additional_info.j2
```diff
@@ -24,5 +24,8 @@ For example, if you are using vite.config.js, you should set server.host and ser
 {% if runtime_info.additional_agent_instructions %}
 {{ runtime_info.additional_agent_instructions }}
 {% endif %}
+{% if runtime_info.date or runtime_info.utc_time %}
+Today's datetime is {{ runtime_info.date }} at {{ runtime_info.utc_time }} (UTC).
```
xingyaoww, line 28: huh, shall we just keep the date then? 

If the agent need specific time, it can execute the command to figure it out?

malhotra5: Yeah sounds good!

malhotra5: I'll open a separate PR to include timestamps for PS1 bash commands

xingyaoww (approved):
Thanks! LGTM!

PR #7511, opened 2025-03-25
(llm): Track accumulated token usage instead of per-request token usage

xingyaoww:

Track accumulated token usage instead of per-request token usage

---

Show accumulated token usage metric:

image

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:9426a27-nikolaik   --name openhands-app-9426a27   docker.all-hands.dev/all-hands-ai/openhands:9426a27
```


openhands/llm/metrics.py
```diff
@@ -109,6 +117,10 @@ def add_token_usage(
                 response_id=response_id,
             )
         )
+        self._accumulated_token_usage.prompt_tokens += prompt_tokens
+        self._accumulated_token_usage.completion_tokens += completion_tokens
+        self._accumulated_token_usage.cache_read_tokens += cache_read_tokens
+        self._accumulated_token_usage.cache_write_tokens += cache_write_tokens
```
enyst, line 123: Just to note, I thought about this a bit, and it seemed to me that the numbers resulted from this might be worrying people for no good reason: in some sense, the tokens in one request _are accumulated tokens_ already, since we include history.

This is another layer on top of it, and it will look significantly high, since it includes every past, cached, request many times over, won't it...? 🤔

An alternative would be to display it in logs, idk. Unless you think it's useful in the UI?

xingyaoww: yeah -- i think the reason why we may want this is bc if we are only showing token counts per request, it might be a bit confusing: the cost is accumulated, not the number of tokens

enyst: You're right, that's fair.

Idk, or if we could separate them so they're not near each other - total cost and tokens/request... 🤔

No problem either way, though.

xingyaoww: yeah i think display the accumulated metric probably make more sense :)

neubig (approved):
LGTM

frontend/src/components/features/conversation-panel/conversation-card.tsx
```diff
@@ -240,22 +240,65 @@ export function ConversationCard({
...
-                <li>- Total: {metrics.usage.total_tokens}</li>
-              </ul>
-            </>
+        <div className="space-y-4">
+          {(metrics?.cost !== null || metrics?.usage !== null) && (
+            <div className="rounded-md p-3">
+              <div className="grid gap-3">
+                {metrics?.cost !== null && (
+                  <div className="flex justify-between items-center border-b border-neutral-700 pb-2">
+                    <span className="text-lg font-semibold">Total Cost:</span>
```
li-boxuan, line 249: This is very tiny, but just so you know, some other areas in the world also use "$" sign (e.g. Hong Kong Dollars) so this could be confusing :)

```suggestion
                    <span className="text-lg font-semibold">Total Cost (USD):</span>
```

li-boxuan (approved):

Issue #7505, opened 2025-03-25
Trigger microagents on agent keywords


neubig:
**What problem or use case are you trying to solve?**

Currently microagents are triggered when the user mentions a keyword, but not when an agent mentions a keyword.

**Describe the UX of the solution you'd like**

Mentions of keywords by the agent should trigger the microagent.

malhotra5:
#7516 triggers microagents for `MessageAction` actions from the agent

However, the agent has a tendency to use `CmdRunAction` and `FinishAction` far more frequently, and triggering the microagent for these actions will require a more involved solution

PR #7516, opened 2025-03-26
[Feat]: Trigger microagents on agent keywords

malhotra5:

- Trigger microagents on agent keywords

---

This PR 

1. adds a `RecallAction` for agent message actions as well 
2. pulls microagent knowledge when specific keyword triggers exist in the agent's message

---
#7505

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:ef8821f-nikolaik   --name openhands-app-ef8821f   docker.all-hands.dev/all-hands-ai/openhands:ef8821f
```

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -479,8 +479,18 @@ class AgentController:
 
             if self.get_agent_state() != AgentState.RUNNING:
                 await self.set_agent_state_to(AgentState.RUNNING)
-        elif action.source == EventSource.AGENT and action.wait_for_response:
-            await self.set_agent_state_to(AgentState.AWAITING_USER_INPUT)
+        elif action.source == EventSource.AGENT:
+            # Check if we need to trigger microagents based on agent message content
+            recall_action = RecallAction(
+                query=action.content, recall_type=RecallType.KNOWLEDGE
+            )
+            self._pending_action = recall_action
+            # This is source=AGENT because the agent message is the trigger for the microagent retrieval
+            self.event_stream.add_event(recall_action, EventSource.AGENT)
+
+            # If the agent is waiting for a response, set the appropriate state
+            if action.wait_for_response:
+                await self.set_agent_state_to(AgentState.AWAITING_USER_INPUT)
 
     def _reset(self) -> None:
         """Resets the agent controller."""
--- a/openhands/memory/memory.py
+++ b/openhands/memory/memory.py
@@ -97,11 +97,14 @@ class Memory:
                     return
 
                 # Handle knowledge recall (triggered microagents)
+                # Allow triggering from both user and agent messages
                 elif (
                     event.source == EventSource.USER
-                    and event.recall_type == RecallType.KNOWLEDGE
-                ):
-                    logger.debug('Microagent knowledge recall')
+                    or event.source == EventSource.AGENT
+                ) and event.recall_type == RecallType.KNOWLEDGE:
+                    logger.debug(
+                        f'Microagent knowledge recall from {event.source} message'
+                    )
                     microagent_obs: RecallObservation | NullObservation | None = None
                     microagent_obs = self._on_microagent_recall(event)
                     if microagent_obs is None:
--- a/tests/unit/test_memory.py
+++ b/tests/unit/test_memory.py
@@ -261,3 +261,62 @@ REPOSITORY INSTRUCTIONS: This is a test repository.
 
     # Clean up
     os.remove(os.path.join(prompt_dir, 'micro', f'{repo_microagent_name}.md'))
+
+
+@pytest.mark.asyncio
+async def test_memory_with_agent_microagents():
+    """
+    Test that Memory processes microagent based on trigger words from agent messages.
+    """
+    # Create a mock event stream
+    event_stream = MagicMock(spec=EventStream)
+
+    # Initialize Memory to use the global microagents dir
+    memory = Memory(
+        event_stream=event_stream,
+        sid='test-session',
+    )
+
+    # Verify microagents were loaded - at least one microagent should be loaded
+    # from the global directory that's in the repo
+    assert len(memory.knowledge_microagents) > 0
+
+    # We know 'flarglebargle' exists in the global directory
+    assert 'flarglebargle' in memory.knowledge_microagents
+
+    # Create a microagent action with the trigger word
+    microagent_action = RecallAction(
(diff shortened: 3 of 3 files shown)


enyst (changes requested):
Good to see the system expanding!

I would appreciate some time to give this more thought though.

The PR doesn't add it just for a (very) limited set of actions, but for ALL 'agent' actions, if I see this right? That's a lot. Are we sure it's necessary?

More importantly, the way recall works is by setting `_pending_action` while waiting, the same way the Runtime does. But in the case of `runnable` actions, it seems both Memory and Runtime would set it, overriding each other? Then one unsets it, so the other won't? Sorry, it's not clear to me that it can work.

A test with a CmdRunAction may help us see what happens

malhotra5:
@enyst Good points! I'm looking at the code and I see that I've added `RecallAction` only for the `_handle_message_action` function (so this should exclude other agent actions such as `CmdRunAction`)?

malhotra5:
image

https://www.all-hands.dev/share?share_id=c9d0d5504b6bef07985969faccb24c32680b832b3c484bfa7167232a78711475


Here's an example trajectory where `FinishAction` and `CmdRunAction` don't trigger the microagent (via `RecallAction`)

malhotra5:
Here is another trajectory of the agent message action triggering the microagent, lmk your thoughts!

https://www.all-hands.dev/share?share_id=f8cf1f180e0b068754a998fd5df5f9725839c75d6de50e988e84330240b75c9b

enyst:
> @enyst Good points! I'm looking at the code and I see that I've added `RecallAction` only for the `_handle_message_action` function (so this should exclude other agent actions such as `CmdRunAction`)?

Ah! My bad, I didn't see this right. That works!

If you think it's useful on agent MessageActions, I think that's perfectly fine.

malhotra5:
I'm hoping this prevents the agent from spinning around -> sometimes we see it use `MessageActions` to say things like

> Oh I'm sorry it seems like Github token is not working, let me try browsing instead

Hoping that doing this will help it remember that it should set remote repository, etc! Hoping to introduce minimal changes for now to see if we can improve things, but otherwise it would be nice to find a way to get this to work with `CmdRunActions` in the future 🙂

enyst (approved):
Thank you, looks good to me!

I wonder if there still are possible async issues, such as: user sends a message during agent run, agent happens to send a message in the same time -ish. I wouldn't hold the PR over it, just food for thought.

Maybe an agent message that triggers waiting for user... 🤔 If you feel confident it's fine, then we could go with it.

enyst:
> Oh I'm sorry it seems like Github token is not working, let me try browsing instead

Are we sure those are MessageActions and not the 'thought' of a browsing action?

malhotra5:
Let me double check both on potential async problems and memory failure modes (I gave an example off the top of my head which would like come from browsing tool, but I believe there are legitimate trajectories which prompted us opening the linked issue so I'll double check)

enyst:
I feel the agent MessageActions are more rare these days than they used to be, since we changed the Finish action last time: the LLMs choose Finish more, even if they didn't finish (now it has a field to say whether the task was completed or not). 🤷‍♂️

But we'll see, maybe it will address the lack of knowledge.

malhotra5:
You're absolutely right - I'm looking at trajectories and it's mostly the agent spamming `CmdRunAction`, going around in circles, and then wrapping it up with `FinishAction`.

So I think this PR really just catches edge cases were sometimes the agent decides to use `MessageAction`. I'll probably open a future PR for a solution that integrates with `CmdRunAction`

openhands/memory/memory.py
```diff
@@ -97,11 +97,14 @@ async def _on_event(self, event: Event):
                     return
 
                 # Handle knowledge recall (triggered microagents)
+                # Allow triggering from both user and agent messages
```
li-boxuan, line 100: I think you might also want to update the comment on L79:

> # if this is a workspace context recall (on first user message)

li-boxuan: Oh never mind, I misread

openhands/controller/agent_controller.py
```diff
@@ -479,8 +479,18 @@ async def _handle_message_action(self, action: MessageAction) -> None:
 
             if self.get_agent_state() != AgentState.RUNNING:
                 await self.set_agent_state_to(AgentState.RUNNING)
-        elif action.source == EventSource.AGENT and action.wait_for_response:
-            await self.set_agent_state_to(AgentState.AWAITING_USER_INPUT)
+        elif action.source == EventSource.AGENT:
+            # Check if we need to trigger microagents based on agent message content
```
li-boxuan, line 483: Is RecallAction idempotent? I imagine what could happen now is:

1. User: ..... github ....  (microagent triggered)
2. Agent: ..... github is done ... what's next? (microagent triggered, per new logic)
3. User: .... more with github .... (microagent triggered)

The same microagent might get triggered twice. Would it cause any trouble? Would the microagent prompt be included twice? Can we add a unit test to check?

malhotra5: I believe the context get's repeated multiple times, but I don't suppose that could hurt. 

@openhands please read the diff on this PR carefully and then write a unit test to check if the RecallAction is idempotent

User: ..... github .... (microagent triggered)
Agent: ..... github is done ... what should be my next step? (microagent triggered, per new logic)
User: .... more with github .... (microagent triggered)

Check if the same microagent might get triggered multiple times (if so, how many times in this case)

malhotra5: Regardless, I think we may have to rethink how this is done. It's more useful to have this occur on `CmdRunAction` and also dedup microagent knoweledge if it's already in the event stream

enyst: We do a dedup, in `conversation_memory.py`. We have tests for that. I think it will just work for those triggered from agent.

li-boxuan: > We do a dedup, in conversation_memory.py. We have tests for that. I think it will just work for those triggered from agent.

Aha, sounds good then!

malhotra5:
I've check against the concerns brought up and confirmed this PR works as expected. Will merge soon but not close the linked issue

PR #7353, opened 2025-03-19
(fix): Condensation events to reconstruct contexts added to event stream

csmith49:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality that this introduces.**

---
**Give a summary of what the PR does, explaining any non-trivial design decisions.**

This PR updates the condenser interface to support a stateless implementation of a rolling condenser.

Instead of returning just a condensed history, this new interface allows a condenser to return a `View` of the history or a `Condensation` event. The condensation event can be added to the event stream, which is then one step closer to being the only source of state needed for the agent.

To implement this strategy, rolling condensers now define three abstract functions:
* `get_view`, which converts the sequence of events in the state to a view object by utilizing the information in the condensation events,
* `get_condensation`, which takes a view and produces a condensation event defining how future views should be constructed, and
* `should_condense`, a trigger which determines when the condensation should happen.

In addition to the condenser changes, this PR updates the CodeAct agent implementation to use the new interface and updates all condenser tests accordingly.

---
**Link of any specific issues this addresses.**

This directly solves the problem in #6706, but should also resolve #7175.

PR #7526, opened 2025-03-26
Condense action

enyst:

---

This PR proposes a possible AgentCondensationAction layout, for use in https://github.com/All-Hands-AI/OpenHands/pull/7353

---

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -45,6 +45,7 @@ from openhands.events import (
 from openhands.events.action import (
     Action,
     ActionConfirmationStatus,
+    AgentCondensationAction,
     AgentDelegateAction,
     AgentFinishAction,
     AgentRejectAction,
@@ -303,7 +304,7 @@ class AgentController:
             ):
                 # TODO: this is fragile, but how else to check if eligible?
                 return True
-            if isinstance(event, AgentDelegateAction):
+            if isinstance(event, (AgentDelegateAction, AgentCondensationAction)):
                 return True
             return False
         if isinstance(event, Observation):
--- a/openhands/core/schema/action.py
+++ b/openhands/core/schema/action.py
@@ -85,5 +85,8 @@ class ActionTypeSchema(BaseModel):
     RECALL: str = Field(default='recall')
     """Retrieves content from a user workspace, microagent, or other source."""
 
+    CONDENSE: str = Field(default='condense')
+    """Records a condensation event in the agent's history."""
+
 
 ActionType = ActionTypeSchema()
--- a/openhands/events/action/__init__.py
+++ b/openhands/events/action/__init__.py
@@ -1,5 +1,6 @@
 from openhands.events.action.action import Action, ActionConfirmationStatus
 from openhands.events.action.agent import (
+    AgentCondensationAction,
     AgentDelegateAction,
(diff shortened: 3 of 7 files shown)


neubig (commented):
This looks very nice to me, although I might want to see what it looks like actually in https://github.com/All-Hands-AI/OpenHands/pull/7353 to fully understand.

I'll leave it to @csmith49 to actually approve though.

enyst:
@csmith49 I'll pick this up here:
> Second, some fields (considered_event_ids and forgotten_event_ids) are not as compact as they could be. It sounds like this is a big concern for you @enyst, considering your proposal in https://github.com/All-Hands-AI/OpenHands/pull/7526, so maybe I can address some of that here.

>> I'd love your thoughts on https://github.com/All-Hands-AI/OpenHands/pull/7526 , as minimal (-ish) alternative. Could it have, or maybe it doesn't have all we actually need?
> I actually tried a similar approach when working on this PR, but I found myself re-inventing the pydantic wheel to get out of the serialization slog. My concerns aren't representational, they're to do with code complexity and control-flow: without pydantic capabilities to do per-strategy field validation, history reconstruction gets pretty brittle and difficult to recover from.

> One big benefit of a uniform condensation event is that history reconstruction is uniform. There's no delegation to condensers or a big switch statement over the condenser strategy (technically we have the former now but the implementations are nearly the same and can be made the same without issue), both of which are sources of complexity hindering your suggested refactor re: condensers and conversation manager straightforward.

My thinking is that we could maybe do the other way around:
- settle on an Action, like maybe this PR
- pick it up in 7353 and use it.

The minimum that we could do between these 2 PRs is to reconstruct the 2 lists that 7353 needs, from `state.history` and `AgentCondensationAction`. I mean, that would be the minimal change to get to 7353, it seems to me that if we can, it doesn't have to be more complicated, and we don't need to refactor the underlying implementation.

The uniform condensation and the observation you made re: what is next is an interesting point. Yes, there is some complexity in the condensers' discriminated union that needs smoothed over _somewhere_. And it should ideally be one place.

Here's roughly what I have in mind, but not for 7526/7353, just in general so we know what we're talking about, and please feel free to shoot it down:

```
# condenser interface
return AgentCondensationAction | None

# codeact agent:
condense_action = self.condenser.condensed_history()
if condense_action:
    return condense_action

self.conversation_memory.process_events(). # no events need to come from condensers anymore
# because we already know how to make up the right history to send to LLM

# conversation memory:

def process_events(...)
# steps to figure out what to send in context
# 1. slice history by AgentCondensationAction(s) to get the remaining list to send
# 2. go through events and map the tool calls to responses, while converting to Messages
# 3. verification step: check the resulting Messages don't have tool_call_id without match
# note that 3 and 2 could also be inverted, we know that condensers may result in unpaired events, so maybe the slicing could include it or we could verify and fix right after we read the remaining events
return messages;

def read_view()
""" Reads from state.history and AgentCondensationActions the actual view we now have at this step"""
# may contain condenser strategy if necessary
# forgotten_events seems like maybe just listing the ids between start_id or keep_first and end_id ?
# considered_events

```

enyst:
Small follow up, the two lists we need in 7353 maybe would come down to something like this:
- https://github.com/All-Hands-AI/OpenHands/pull/7534

PR #7533, opened 2025-03-26
Refactor file service

amanape:

---
- Extract file service API calls to a separate class
- Extract file service mock API handlers
- Remove unused file service APIs from the backend and frontend
  - Save file
  - Upload file
- Add basic tests for mocked API

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:3d28470-nikolaik   --name openhands-app-3d28470   docker.all-hands.dev/all-hands-ai/openhands:3d28470
```


rbren (approved):

rbren:
Love the smell of deleted code 😄

openhands/server/routes/files.py
```diff
@@ -148,168 +141,8 @@ async def select_file(file: str, request: Request):
...
-                'error': f'Error during file upload: {str(e)}',
-                'uploaded_files': [],
-                'skipped_files': [],
-            },
-        )
-
-
-@app.post('/save-file')
-async def save_file(request: Request):
-    """Save a file to the agent's runtime file store.
```
enyst, line 248: I see the point of refactoring the file service, and cleaning up what we no longer use in FE.

I can also see an argument that `/save_file` continues to make sense though. It's one of the most common expected abilities of a file service, just CRUD; it doesn't really depend on the needs of only one client (like web FE). 

If someone would like to write a new client, they will have to redesign/re-add the server API again. Maybe that could be avoided if the server API keeps the basic operations - and I think we know saving files is a standard expectation. (and we had it, for long!)

What do you think?

amanape: Good point, we'll probably add this again soon

enyst: Thanks! I'm surprised you found this comment after two months. May I ask, how did you find it?

amanape: Was looking for the PR where I removed the API so it is easier to re-add it in the near future 😂

PR #7535, opened 2025-03-27
Reduce max iterations by default

enyst:

Set max iterations to 250 by default.

---

This PR proposes to reduce max iterations to 250.

With Sonnet 3.7's energy, this does seem to go off quite a lot. We also had a fix meanwhile, where if the user speaks, the max iterations are added again on top of existing value, which gives more breathing room than 250 anyway.

With UI, hitting this only means that the user tells it to continue.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:60c1111-nikolaik   --name openhands-app-60c1111   docker.all-hands.dev/all-hands-ai/openhands:60c1111
```

### DIFF
--- a/config.template.toml
+++ b/config.template.toml
@@ -24,7 +24,7 @@
 #daytona_target = ""
 
 # Base path for the workspace
-workspace_base = "./workspace"
+#workspace_base = "./workspace"
 
 # Cache directory path
 #cache_dir = "/tmp/cache"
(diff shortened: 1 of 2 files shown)


malhotra5 (approved):
Thanks! Seems like this is needed as of lately

PR #6815, opened 2025-02-19
Feat(agenthub): Add new agent, ProxyAgent to enable collaboration among multiple OpenHands instances

f-miyake:
This pull request introduces a new agent named Proxy Agent that extends the current delegation mechanism. In addition to handling local task delegation, the Proxy Agent is designed to delegate tasks to agents hosted on external OpenHands instances. This enhancement is a step toward enabling our system to integrate with external AI agent platforms.

**End-user friendly description of the problem this fixes or functionality that this introduces**

- [ ] Include this change in the Release Notes. If checked, you must provide an **end-user friendly** description for your change below

---
**Give a summary of what the PR does, explaining any non-trivial design decisions**
- This PR introduces the Proxy Agent, a new agent that extends the existing task delegation mechanism by enabling delegation to remote agents hosted on separate OpenHands instances.
- The Proxy Agent works with function calling and can choose from two tools, DelegateLocalTool for the existing local delegation and DelegateRemoteOHTool for the remote delegation to an agent on a different instance of OpenHands.
- The DelegateRemoteOHTool is responsible for sending messages to agents running on different OpenHands instances. This tool is part of a separate feature introduced in https://github.com/All-Hands-AI/OpenHands/pull/6492 (not merged yet), which enhances OpenHands' ability to communicate with external agents.
- The list of available agents, both local and remote, is defined in openhands/agenthub/proxy_agent/agent_list.json. 
- For more details, please refer to the README.md in openhands/agenthub/proxy_agent directory.

PR #6492, opened 2025-01-28
feat(agent_skills): Add OpenhandsClient to agent_skills to enable collaboration among multiple OpenHands instances

f-miyake:
This change adds the OpenhandsClient class and the message_to_remote_OH function to agent_skills, enabling the ability to send messages to other OpenHands instances. 
The agent can make use of this tool via IPythonRunCellAction.

**End-user friendly description of the problem this fixes or functionality that this introduces**

- [ ] Include this change in the Release Notes. If checked, you must provide an **end-user friendly** description for your change below

---
**Give a summary of what the PR does, explaining any non-trivial design decisions**
This function is designed for agent developers to create agents that collaborate with OpenHands instances hosted on other servers. Currently, only sending Message Actions is supported, but I plan to add features such as file uploads in the future.

Additionally, I am planning to implement a Proxy Agent into agenthub that enables collaboration between OpenHands instances hosted separately by different users or organizations by utilizing this functionality.
I am also considering developing a client that can connect to agent platforms other than OpenHands, aiming to facilitate collaboration across different platforms.
The changes in this PR are intended to be the first step toward that.

```mermaid
flowchart LR
    subgraph Server1
        A["CodeActAgent"] -->|delegate| B["Proxy Agent<br> (To be implemented)"]
    end
    subgraph Server2
        C["Other Agent"]
    end
    B -->|message_to_remote_OH| C

```

PR #7536, opened 2025-03-27
feat(client-authn): Add client authentication function in openhands server using JWT

yonekura-yuki-f:

This change add a token-based authentication function to OpenHands server.
It will be possible to open OpenHands server to the public, not just locally(frontend).

---
- The implementation is based on JWT with the intention of expanding to OAuth and OIDC.
- This is an optional feature and the default is JWT_SECRET_CLIENT_AUTH=None. This will enable multiple OpenHands instances to collaborate on multiple hosts in the near future. 
  - https://github.com/All-Hands-AI/OpenHands/pull/6492
  - https://github.com/All-Hands-AI/OpenHands/pull/6815
- The frontend currently provides the JWT via an environment variable, but in the future we would like to add a login function to the frontend so that it can be provided dynamically.
- The JWT must be encoded with the secret phrase specified in JWT_SECRET_CLIENT_AUTH.
  - JWT example:
    ```
     import jwt
     payload = {'user_id': 'some-user-id(any string)'}
     secret = 'secret-key-for-client-auth'
     token = jwt.encode(payload, secret, algorithm='HS256')
    ```

### DIFF
--- a/config.template.toml
+++ b/config.template.toml
@@ -87,6 +87,9 @@
 # JWT secret for authentication
 #jwt_secret = ""
 
+# jwt secret for authentication
+#jwt_secret_client_auth = jwt-secret-for-client-auth
+
 # Restrict file types for file uploads
 #file_uploads_restrict_file_types = false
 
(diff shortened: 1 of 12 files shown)


rbren:
@yonekura-yuki-f thank you for the contribution, but we've made the intentional choice to keep OpenHands single-tenant for now. We do not want to support multiple users on the same OpenHands instance.

There are a ton of security and scalability problems, mostly related to the runtime. This is a difficult thing to configure correctly.

We do have a multi-tenant solution internally, which we're working on making more widely accessible. Feel free to ping me in Slack if you'd like to learn more!

But for now, I think we need to close this one.

Issue #7540, opened 2025-03-27
[Bug]: docker_runtime.py Failed to start container. Cause http 500 on http+docker://localhost/v1.47/containers/


happyherp:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Reproduce

Error happens on my local machine, Windows 10 using WSL. 

* checkout the code of rev 1230b229b5a010b3c0ac9e3e909fec6547d81252 or later
* make build && make run-wsl
* access http://localhost:3001/
* open a new conversation "do a hello world pathon"

The container fails to start. 

### Origin of problem

I tried to reproduce this using various commits. 

It starts happening exactly at [#commit 1230b229b5a010b3c0ac9e3e909fec6547d81252](https://github.com/All-Hands-AI/OpenHands/commit/1230b229b5a010b3c0ac9e3e909fec6547d81252)
Replace use of requests with httpx (#7354) tofarr* Yesterday 14:37

The commit before it works fine. 

### Logs

```
14:02:53 - openhands:INFO: runtime_build.py:182 - Building image: ghcr.io/all-hands-ai/runtime:oh_v0.30.0_ts97n9542l0zp9u1_kuw9u5cc05eirpmk
14:02:53 - openhands:DEBUG: docker.py:254 - Checking, if image exists locally:
ghcr.io/all-hands-ai/runtime:oh_v0.30.0_ts97n9542l0zp9u1_kuw9u5cc05eirpmk
14:02:53 - openhands:DEBUG: docker.py:256 - Image found locally.
14:02:53 - openhands:DEBUG: runtime_build.py:211 - Reusing Image [ghcr.io/all-hands-ai/runtime:oh_v0.30.0_ts97n9542l0zp9u1_kuw9u5cc05eirpmk]
14:02:53 - openhands:INFO: docker_runtime.py:140 - [runtime 720332feed994a4ca38cb84222066538] Starting runtime with image: ghcr.io/all-hands-ai/runtime:oh_v0.30.0_ts97n9542l0zp9u1_kuw9u5cc05eirpmk
14:02:53 - openhands:DEBUG: docker_runtime.py:186 - [runtime 720332feed994a4ca38cb84222066538] Preparing to start container...
14:02:53 - openhands:DEBUG: docker_runtime.py:244 - [runtime 720332feed994a4ca38cb84222066538] Workspace Base: /home/ubuntu/projects/MyOpenHands/workspace
[...]

PR #7548, opened 2025-03-27
fix(Runtime): Wait for container to start up

happyherp:

Fixed a bug where we would not wait for a runtime docker container to be started and instead fail. 

---

While waiting for the container to be ready DockerRuntime._wait_until_alive(), fails because the function self.check_if_alive() itself throws a RetryError which contains a httpx.RemoteProtocolError. Before this retry error stopped the process. Now it continues, until _wait_until_alive has no more retries. 

---
https://github.com/All-Hands-AI/OpenHands/issues/7540


tofarr (approved):
🍰 Good Catch!

PR #7550, opened 2025-03-27
fix error message

rbren:
No changelog

---

This error message is too specific to the docker runtime. I made it more generic, and also passed through the error class so the user has more info

---


enyst (approved):
Fair enough IMHO. I think it was useful occasionally, but I'm not sure if often.

PR #7552, opened 2025-03-27
feat(server): Add a new configuration variable, runtime_extra_volumes

ROCKTAKEY:

Add a new configuration variable, `runtime_extra_volumes` which means extra volumes for bind mounts.
We can bind-mount the host directory in the sandbox container by setting it.
It is useful for Docker outside of Docker since it let us mount `/var/run/docker.sock` on the sandbox.

---

Add a new configuration variable, `runtime_extra_volumes` and add document for it.

---

None


openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -280,7 +280,7 @@ def _init_container(self):
                 name=self.container_name,
                 detach=True,
                 environment=environment,
-                volumes=volumes,
+                volumes=volumes | self.config.sandbox.runtime_extra_volumes,
```
tofarr, line 283: Volumes may be None here which will result in an error

tofarr (changes requested):
Aside from the introduced type error, this looks like a solid change. It should be noted that we have moved away from a mounted volume by default.

openhands/runtime/impl/docker/docker_runtime.py
ROCKTAKEY: Thanks. Fix on https://github.com/All-Hands-AI/OpenHands/pull/7552/commits/df10149733634e270e9cc576ef67c7424860101d.

ROCKTAKEY:
@tofarr Thanks for your review. Fix the typing issue.

>It should be noted that we have moved away from a mounted volume by default.

I am confused. It means the runtime container does not use mounted volume **as workspace**, isn't it?

hamiltop:
There's no issue to 👍 , but I want to chime in and say this would solve many issues for us. Beyond docker in docker, we also would mount a dependency cache volume so we can re-use the outputs of things like `bundle install` across sandboxes, speeding up setup time significantly.

enyst (approved):
LTGM 🍰

@tofarr I think your concern has been addressed, could you take a look?

enyst:
@ROCKTAKEY I apologize for us dropping the ball here. We could have handled this differently, and I promise I'll try to fix our process.

In the meantime, another PR implemented this slightly differently, in the sandbox section, and this feature is released. Thank you for the contribution, and I'm really sorry for the miss!

ROCKTAKEY:
@enyst Never mind. I'm glad it was implemented.

I will show you the functions implemented for someone who needs the feature: [`SANDBOX_VOLUMES`](https://docs.all-hands.dev/modules/usage/runtimes/docker#using-sandbox_volumes)

PR #7553, opened 2025-03-27
Fix uninitialized accumulated tokens

enyst:

Fix error on accumulated tokens when reusing an older session.

---

When loading a conversation from a week ago or so, before we introduced `accumulated_tokens` attribute, I got an error because the `Metrics` object was loaded from `State` pickle and never had this attribute.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:a85016f-nikolaik   --name openhands-app-a85016f   docker.all-hands.dev/all-hands-ai/openhands:a85016f
```

### DIFF
--- a/openhands/llm/metrics.py
+++ b/openhands/llm/metrics.py
@@ -96,6 +96,20 @@ class Metrics:
     def token_usages(self, value: list[TokenUsage]) -> None:
         self._token_usages = value
 
+    @property
+    def accumulated_token_usage(self) -> TokenUsage:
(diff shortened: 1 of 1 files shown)


tofarr (approved):
🍰

PR #7566, opened 2025-03-27
Implement frontend visualization for RecallObservation & Stop issueing recall action for agent message

xingyaoww:

Display microagent context in the UI.

---

- Implement FE UI for Recall Observation
- We need to stop adding "RecallAction" for agent messages. As shown below, it will trigger two duplicated microagent observations (one for workspace context + microagent, and another one for microagent only)

Before this fix:
image

After this fix:
image

TODO for future PR:
- I'm thinking we should get rid of `recall_type` -- because the first RecallAction is always "workspace context+microagent" and we can't really categorize that well. Get rid of this type would probably help us simplify the "RecallObservation" a lot - basically, we just need to care about convert different fields into string.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:68bcd87-nikolaik   --name openhands-app-68bcd87   docker.all-hands.dev/all-hands-ai/openhands:68bcd87
```


openhands/controller/agent_controller.py
```diff
@@ -481,15 +481,8 @@ async def _handle_message_action(self, action: MessageAction) -> None:
...
             if self.get_agent_state() != AgentState.RUNNING:
                 await self.set_agent_state_to(AgentState.RUNNING)
-        elif action.source == EventSource.AGENT:
-            # Check if we need to trigger microagents based on agent message content
-            recall_action = RecallAction(
-                query=action.content, recall_type=RecallType.KNOWLEDGE
-            )
-            self._pending_action = recall_action
-            # This is source=AGENT because the agent message is the trigger for the microagent retrieval
-            self.event_stream.add_event(recall_action, EventSource.AGENT)
```
xingyaoww, line 491: @enyst any concern about this?

xingyaoww:
@openhands can you help me fix frontend linter CI?

xingyaoww:
@OpenHands now a lot of other CIs are also broken. Please fix

frontend/src/state/chat-slice.ts
```diff
@@ -143,6 +148,72 @@ export const chatSlice = createSlice({
...
+              recallObs.extras.runtime_hosts,
+            )) {
+              content += `\n\n- ${host} (port ${port})`;
+            }
+          }
+          if (recallObs.extras.repo_instructions) {
+            content += `\n\n**Repository Instructions:**\n\n${recallObs.extras.repo_instructions}`;
+          }
+          if (recallObs.extras.additional_agent_instructions) {
+            content += `\n\n**Additional Instructions:**\n\n${recallObs.extras.additional_agent_instructions}`;
```
enyst, line 183: Just a thought, all these arguments together can add up to a large content. Shouldn't we apply MAX_CONTENT_LENGTH to it, like we do for normal observations?

openhands/controller/agent_controller.py
enyst: This was just added by @malhotra5 here: 🤣 
https://github.com/All-Hands-AI/OpenHands/pull/7516/files

Out of curiosity, why was it a problem?

enyst: Personally, I think it's okay either way. I also think, on MessageAction from agent, it's perhaps not as useful as Rohit and Graham may have initially intended it.

frontend/src/state/chat-slice.ts
xingyaoww: Actually, i think because these are micro-agent instructions which are mostly human-defined, so it should be actually fine for now?

enyst: Sure!

openhands/controller/agent_controller.py
malhotra5: Yeah I'm good removing this!

openhands/controller/agent_controller.py
```diff
@@ -1084,44 +1077,8 @@ def _apply_conversation_window(self, events: list[Event]) -> list[Event]:
...
-                ):
-                    # if it's a message action or a user action, keep it and continue to find the next event
-                    i += 1
-                    continue
-
-                else:
-                    # if it's an action with source == EventSource.AGENT, we're good
-                    break
+        if len(kept_events) > 0 and isinstance(kept_events[0], Observation):
+            kept_events = kept_events[1:]
```
enyst, line 1081: This seems to assume that there aren't two observations one after the other, I think sometimes there are.

We can let it be though: I don't think it's important, this code was trying to "pair" the action with its obs, but in other parts of the codebase we don't do this, so we end up with some "unpaired" anyway, and we have code later in the execution flow to take care of that. I think it's fine.

enyst (approved):
LGTM! Maybe we can iterate on how we want them to look like. 🤔 

I would love it if @amanape could take a look if possible.

openhands/controller/agent_controller.py
xingyaoww: Yeah, the main reason i removed this is because this messes up with a unit test that used to work well, but no longer do so due to the change in RecallAction behavior

xingyaoww:
@OpenHands help me merge main into this branch and resolve merge conflict

xingyaoww:
@enyst FYI i revert those unnecessary FE changes and just tested locally, this seems to still work:

image

image

So I guess we can merge this PR now without waiting for @amanape to come back from vacation :)

PR #7567, opened 2025-03-28
feat: append sub path to distinguish different runtime data.

answer-huang:

Append sub path to distinguish different runtime data.

---
Append sub path to distinguish different runtime data, otherwise different runtime data will impact each other.

---


neubig (commented):
Hi @answer-huang , thanks for the contribution!

I have a question though: If you want to use different directories, couldn't you just change `workspace_mount_path` every time you run OpenHands?

To be honest, I feel that this change might not be necessary given that there is another straightforward solution to the problem.

answer-huang:
> workspace_mount_path

While it's true that we can use a different path for `workspace_mount_path` each time, I've noticed many instances where '/workspace' is hard-coded as the `workspace_mount_path`.  If we want to define `workspace_mount_path` dynamically, it would indeed be a significant undertaking.

So I'm thinking it would be way easier if docker just automatically mounted different directories when it starts up.

answer-huang:
> Did we get an issue created for this @answer-huang ?

No, find this issue in my local environment and then create a PR.

PR #7570, opened 2025-03-28
Add GitHub Enterprise Server support

doew:
- [x] This change is worth documenting at [https://docs.all-hands.dev/](https://docs.all-hands.dev/)  

This update introduces support for GitHub Enterprise Server (GHES), allowing OpenHands to integrate seamlessly with self-hosted GitHub environments. Users can now configure OpenHands to connect with GHES instances through updated API endpoints and improved authentication mechanisms.  

---

This PR adds support for GitHub Enterprise Server by:  
- Implementing configuration options specific to GHES  
- Updating API endpoint logic to accommodate GHES-specific paths  
- Enhancing authentication flows to support GHES tokens and permissions  
- Adding test cases to validate GHES functionality  

The design ensures that these changes are backward-compatible with existing GitHub.com integrations.  

---

N/A


malhotra5:
Thanks a bunch for putting this together!

The main feedback I have is 

1. to make sure we're saving the enterprise url inside the `secret _store` field inside [`Settings`](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/server/settings.py)

Specifically, adding an extra field `HOST_URL` inside [`ProviderToken`](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/integrations/provider.py#L37-#L40)


2. Inside the accountsettings page for the FE, add an extra host url field which saves/stores the `HOST_URL` for the user in the BE 

3. Use the settings provider in the FE to access the `HOST_URL` and pass around the attribute as needed


We're trying to make this configurable on a per-user; let me know if you need any help with this!

doew:
@malhotra5 
Happy to take this suggestion on!
Initially, I thought having a global setting would be sufficient for self-hosting OpenHands, but I assume this change is necessary to support the SaaS version as well — is that correct?
It might take me a bit of time to implement, but I’ll keep working on it.

malhotra5:
Yup its so that the changes can be ported over to SAAS as well

Its also a matter of consistency, as for self hosted openhands we still maintain values inside user settings rather than setting values globally

doew:
@malhotra5 

Thank you for reviewing this PR. Since the last review, I have added new commits, but in the meantime, GitLab-related changes were introduced, requiring me to resolve conflicts twice.

Given that the time available for this PR is limited, I would like to discuss the best way to proceed smoothly. For example, one possible approach could be merging the PR in its current state and handling any necessary fixes in a separate PR.

Could you share your thoughts on this? I appreciate your help.

doew:
Thanks for running the CI! It looks like it failed, so I’ll go ahead and push a fix to get it passing.

enyst:
@doew I think, if you run a `make build` in the project, it will install the pre-commit hooks that make it simple.

frontend/src/config/github-enterprise.ts
```diff
@@ -0,0 +1,23 @@
...
+ * GitHub Enterprise Server configuration
+ */
+
+/**
+ * Initialize GitHub Enterprise Server configuration from environment variables
+ */
+export const initGitHubEnterpriseConfig = () => {
+  // Set GitHub Enterprise URL if available
+  if (import.meta.env.VITE_GITHUB_ENTERPRISE_URL) {
+    window.GITHUB_ENTERPRISE_URL = import.meta.env.VITE_GITHUB_ENTERPRISE_URL;
```
malhotra5, line 11: Could we avoid using `window` to pass around GitHub related settings. We'd rather just use the `UserSettings` because its persists for local installations as well

frontend/src/types/git.d.ts
```diff
@@ -36,3 +36,12 @@ interface GitHubCommit {
 interface GithubAppInstallation {
   installations: { id: number }[];
 }
+
+// Extend Window interface to include GitHub Enterprise URL
+declare global {
+  interface Window {
```
malhotra5, line 42: similarly remove use of window here

frontend/src/types/settings.ts
```diff
@@ -18,7 +32,7 @@ export type Settings = {
   ENABLE_DEFAULT_CONDENSER: boolean;
   ENABLE_SOUND_NOTIFICATIONS: boolean;
   USER_CONSENTS_TO_ANALYTICS: boolean | null;
-  PROVIDER_TOKENS: Record<Provider, string>;
+  PROVIDER_TOKENS: Record<Provider, ProviderTokenDataGet | string>;
```
malhotra5, line 35: Can we enforce this to always be `ProviderTokenData` instead of `string

Also the backend route can end empty token for `GET` requests, instead of making `ProviderTokenDataPost` and `ProviderTokenDataGet`

frontend/src/utils/generate-github-auth-url.ts
```diff
@@ -7,6 +7,19 @@
  */
 export const generateGitHubAuthUrl = (clientId: string, requestUrl: URL) => {
   const redirectUri = `${requestUrl.origin}/oauth/keycloak/callback`;
+
+  // Check if GitHub Enterprise Server URL is configured
+  const githubEnterpriseUrl = window.GITHUB_ENTERPRISE_URL || "";
```
malhotra5, line 12: can we have a param from the user settings passed to this function instead of pulling the URL value out from `window`

openhands/integrations/github/github_service.py
```diff
@@ -32,10 +33,55 @@ def __init__(
...
+                self.BASE_URL = f'{base_url}/api/v3'
+                self.GRAPHQL_URL = f'{base_url}/api/graphql'
+
+            # Log that we're using GitHub Enterprise Server
+            logger.debug(
+                f'Using GitHub Enterprise Server with BASE_URL: {self.BASE_URL}'
+            )
+        else:
+            # Default to github.com API
+            self.BASE_URL = get_github_api_url() or self.DEFAULT_BASE_URL
```
malhotra5, line 77: NIT: lets define the helper functions within this class (e.g `get_github_api_url`)

Let's remove the config class and just use user settings for storing the `BASE_URL` values

openhands/server/routes/git.py
```diff
@@ -93,9 +96,14 @@ async def get_github_installation_ids(
 ):
     if provider_tokens and ProviderType.GITHUB in provider_tokens:
         token = provider_tokens[ProviderType.GITHUB]
+        # Use the host_url from the token if available, otherwise use the default API URL
+        github_api_url = token.host_url or get_github_api_url()
```
malhotra5, line 100: We should have host_url information available in the token, so `get_github_api_url` should be unnecessary

We strictly want to store such information on a per user basis

malhotra5:
FYI #7766 is adding similar support but for Gitlab, so some of the scaffolding changes may be quite similar

doew:
Thank you very much for the helpful review!
I’ll go through each point and address them one by one.

Issue #7175, opened 2025-03-10
The context window seems to "reset" to earlier points, rather than later points in time


enyst:
> @enyst I want you to see what I mean about the condenser.  It's like it forgets what it just did.
> 
> I'll help you set up the environment for the marketplace platform. Let me start by configuring the ports in the .env.docker file based on the runtime information provided.
> 
> Read the contents of a file
> Now I'll update the .env.docker file with the correct port configuration based on the runtime information:
> 
> Edited the contents of a file
> Now I'll also update the NUXT_PUBLIC URLs to match the new port configuration:
> 
> Edited the contents of a file
> Now I'll run the initialization script in the background as instructed:
> 
> Ran a bash command
> As per the instructions, I will not check the setup.log file. Instead, I'll provide you with the frontend and backend URLs based on the port configuration:
> 
> Frontend URL: http://localhost:54425/ Backend URL: http://localhost:58450/
> 
> The initialization process is running in the background. These URLs will be accessible once the setup is complete.
> 
> Is there anything specific you'd like me to help you with regarding the marketplace platform? I'm ready to assist with any further instructions or tasks.
> 
> can you run setup with sudo
> 
> I'll run the setup script with sudo privileges as requested:
> 
> Ran a bash command
> I notice that the previous setup process has exited with code 1, which indicates there might have been an issue. Now I've started a new setup process with sudo privileges. This is running in the background with the process ID 902.
> 
> As per the instructions, I will not check the setup.log file. The frontend and backend URLs remain the same:
> 
> Frontend URL: http://localhost:54425/ Backend URL: http://localhost:58450/
> 
> These URLs will be accessible once the setup with sudo privileges is complete. The system is now initializing in the background.
> 
> Is there anything else you'd like me to help you with while we wait for the setup to complete?
> 
> instructions/hybrid-page-builder can your read everything in here to see what we're working on
> 
> I'll read the contents of the instructions/hybrid-page-builder directory to understand what we're working on.
> 
> Read the contents of a file
> Let me start by reading the overview to understand the project:
> 
> Read the contents of a file
> Now let's look at the Phase 1 steps to understand the current implementation stage:
> 
> Read the contents of a file
> Read the contents of a file
> Read the contents of a file
> Read the contents of a file
> Let me also check the Phase 2 steps to understand the frontend integration part:
> 
> Read the contents of a file
> Let me check the progress directory to see what has been completed so far:
> 
> Read the contents of a file
[...]

PR #7578, opened 2025-03-28
fix: Context window truncation using `CondensationAction`

csmith49:

---

When an LLM hits a context window limitation, the agent controller chops the history roughly in-half so the agent can continue to make forward progress. This behavior is a necessary fallback but introduces some strange behavior (#7175, for example) when sessions are reloaded, especially when condensers are in play.

#7353 introduced two new primitives that can help to resolve this issue: the `CondensationAction` and the `View`. Instead of relying on the agent controller to maintain an event stream and a history and keep them in-sync modulo any truncations, we can instead

1. Modify the context truncation behavior to emit a `CondensationAction` event. This event records which parts of the history should be removed.
2. Rely on the use of `View` objects downstream instead of the raw `State.history`. When given a list of events, the `View` automatically applies any condensation actions to produce a smaller history that can be used by agents.

This approach effectively stores the context truncation in the event stream/history. If we reload a session or some state gets out-of-sync, we can produce the exact same truncated history directly from the event stream.

(As an added benefit, this uses the same system as condensers, so they shouldn't end up in a position where they have to fight for priority on what/how to truncate.)

To support these changes, this PR makes the following edits:
* `View` definition relocated and `openhands.memory` sub-module structure slightly reorganized. Necessary to avoid circular imports for the next edit.
* `State` objects have a `State.view` property which constructs a `View` directly from the state's history.
* Agent controller context truncation updated to emit `CondensationAction` instead of manually chopping up the history.
* Tests updated.
* References to `state.history` in agents updated to `state.view`.

---


tests/unit/test_truncation.py
```diff
@@ -1,244 +0,0 @@
...
-            controller.state.history
-        )
-
-        # Verify truncation occurred
-        assert len(controller.state.history) < original_history_len
-        assert controller.state.start_id == first_msg._id
-        assert controller.state.truncation_id is not None
-        assert controller.state.truncation_id > controller.state.start_id
-
-    def test_history_restoration_after_truncation(self, mock_event_stream, mock_agent):
```
csmith49, line 187: This test was moved to `test_agent_controller.py` with no change.

tests/unit/test_truncation.py
```diff
@@ -1,244 +0,0 @@
...
-    def agent_step_fn(state):
-        return MessageAction(content='Agent returned a message')
-
-    agent.step = agent_step_fn
-
-    return agent
-
-
-class TestTruncation:
-    def test_apply_conversation_window_basic(self, mock_event_stream, mock_agent):
```
csmith49, line 37: This test was moved to `test_agent_controller.py`, and extended with the checks from `test_context_window_exceeded_handling`.

tests/unit/test_truncation.py
```diff
@@ -1,244 +0,0 @@
...
-        # Check that the history has been truncated before closing the controller
-        assert len(controller.state.history) == 13 < history_len
-
-        # Check that after properly closing the controller, history is recovered
-        asyncio.run(controller.close())
-        assert len(controller.event_stream.get_events()) == history_len
-        assert len(controller.state.history) == history_len
-        assert len(controller.get_trajectory()) == history_len
-
-    def test_context_window_exceeded_handling(self, mock_event_stream, mock_agent):
```
csmith49, line 129: This test was folded into `test_apply_conversation_window_basic` during the move.

tests/unit/test_truncation.py
```diff
@@ -1,244 +0,0 @@
...
-        assert truncated[0] == first_msg  # First message always preserved
-        assert controller.state.start_id == first_msg._id
-        assert controller.state.truncation_id is not None
-
-        # Verify pairs aren't split
-        for i, event in enumerate(truncated[1:]):
-            if isinstance(event, CmdOutputObservation):
-                assert any(e._id == event._cause for e in truncated[: i + 1])
-
-    def test_truncation_does_not_impact_trajectory(self, mock_event_stream, mock_agent):
```
csmith49, line 82: The check for this test was folded into `test_context_window_exceeded_error_handling` (that's _not_ the test originally from this file, note the extra `error` in the name).

tests/unit/test_condenser.py
```diff
@@ -88,16 +88,6 @@ def set_mock_response_content(content: Any):
     return mock_llm
 
 
-@pytest.fixture
-def mock_state() -> State:
```
csmith49, line 92: Turns out we can use actual `State` objects instead of mocks in all these tests :)

csmith49:
To anyone looking at this code: I recommend starting with the `View` class to see how `CondensationAction` events are used, then taking a look at the changes in the agent controller.

The other changes are important but mostly serve to test/support the above.

csmith49:
> Overall, this looks good to me!
> 
> I have a slight concern about the efficiency implications of calling `View.from_events` many times, but maybe it's not a major concern.

Hmm, I see what you mean. I think we'll always have to call it once per-step, but since it's linear in the size of the history that's probably okay. I can optimize away any more calls by caching the view construction in the state.

openhands/controller/state/state.py
```diff
@@ -183,7 +184,7 @@ def get_current_user_intent(self) -> tuple[str | None, list[str] | None]:
         """Returns the latest user message and image(if provided) that appears after a FinishAction, or the first (the task) if nothing was finished yet."""
         last_user_message = None
         last_user_message_image_urls: list[str] | None = []
-        for event in reversed(self.history):
+        for event in reversed(self.view):
```
enyst, line 187: Please correct me if wrong, does this mean that we now try to save view in the state pickle? We [avoid saving](https://github.com/All-Hands-AI/OpenHands/blob/b6321488bc6f45cb7e8598430ae1653b08eb47d9/openhands/events/stream.py#L311) history.

Maybe not a big thing, but this was intended: the reasoning here is that the event stream is the "single source of truth" for the events that exist in the system, and thus saving duplicates in other places than the stream seems like an anti-pattern. It could lead to unexpected results when the two diverge (maybe like, session closed normally and saved a view, then next restore it closed uncleanly and didn't save?).

Can we even pickle a `View`?

openhands/controller/agent_controller.py
```diff
@@ -1046,16 +1045,20 @@ def _init_history(self) -> None:
...
 
         # Save the ID of the first event in our truncated history for future reloading
         if self.state.history:
             self.state.start_id = self.state.history[0].id
 
         # Add an error event to trigger another step by the agent
         self.event_stream.add_event(
-            AgentCondensationObservation(
-                content='Trimming prompt to meet context window limitations'
+            CondensationAction(
```
enyst, line 1059: I think ideally, we could also clean up the code that sets `truncation_id` and the code that restores according to it. Since this should just load normally now and read itself from the events, I think keeping that code is not necessary and maybe it can cause issues.

Probably everything about truncation id, I think, can go. 🤔

enyst (commented):
Very nice PR, thank you for this! I now see better what you had in mind. 👍

openhands/controller/state/state.py
csmith49: I'm with you here. The idea was to define `state.view` using the `@property` decorator, so that it won't actually be serialized while pickling.

But Graham pointed out that using the `view` property as a drop-in replacement for the `history` field means re-computing the view a number of times per-step. So I've updated the `State` implementation to cache the `View` that gets returned and to explicitly avoid serializing anything to do with that caching.

openhands/controller/agent_controller.py
csmith49: I'll trim it out and make sure nothing breaks!

openhands/controller/state/state.py
```diff
@@ -97,8 +97,6 @@ class State:
     # start_id and end_id track the range of events in history
     start_id: int = -1
     end_id: int = -1
-    # truncation_id tracks where to load history after context window truncation
-    truncation_id: int = -1
```
enyst, line 100: Actually, I wonder if it will need to read this from previous pickles? Only this occurrence, to create the object.

csmith49: But it's an easy include if you want peace of mind -- we can mark it deprecated and delete it at any time to fallback to the behavior I mentioned.

enyst: Yes, loading a State created on main branch, with a long conversation and it had this set. I think you're right though. Then it's okay!

openhands/controller/agent_controller.py
csmith49: I've removed all instances. Nothing seems to break, but I'd appreciate a spot check that I'm not removing functionality needed elsewhere in the deletions.

csmith49:
> I tried to play with older streams, and it looks good. I had some fun with Haiku-generated summaries for 400+ events - amazingly it did mostly do a summary indeed, although it didn't respect the format. Sometimes when it's large enough, Sonnet gets taken away by the events and forgets to summarize. 😅

There's a few instances we've seen where the summarization forgets pretty important bits. If you're good with these changes LMK -- once I get this merged I'm going to start improving the summarization prompt.

enyst (approved):
Just to note, it looks like it also works as expected with no condenser enabled/found.

Thank you for this, I think it's absolutely worth taking this in asap and have fun. ☕️

li-boxuan:
Sorry I didn't get a chance to review carefully but I skimmed over it and didn't spot any major issue. Thanks for the fix!

Issue #7512, opened 2025-03-25
[Bug]: tool_use ids were found without tool_result blocks


neubig:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

When I got

```
Trimming prompt to meet context window limitations
```

After that I got
```
[...]

PR #7582, opened 2025-03-28
fix missing pending tool results for Claude

raymyers:

---

Fix for #7512, when we trim the message history for context length, we sometimes get this error from Claude Sonnet:

```
"messages: `tool_use` ids were found without `tool_result` blocks immediately after"
```

Credit to @onezibo for being the first (afaik) to [fix](https://github.com/onezibo/OpenHands/pull/1/files) this using the openhands agent in their fork. This PR is based on that fix with a slight refactor.

---

### DIFF
--- a/openhands/llm/fn_call_converter.py
+++ b/openhands/llm/fn_call_converter.py
@@ -791,8 +791,21 @@ def convert_from_multiple_tool_calls_to_single_tool_call_messages(
             ), f'Found pending tool calls but not expect to handle it with role {role}: {pending_tool_calls=}, {message=}'
             converted_messages.append(message)
 
-    if not ignore_final_tool_result and len(pending_tool_calls) > 0:
-        raise FunctionCallConversionError(
-            f'Found pending tool calls but no tool result: {pending_tool_calls=}'
-        )
+    if len(pending_tool_calls) > 0:
+        if ignore_final_tool_result:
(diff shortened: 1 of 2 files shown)


enyst (changes requested):
Thanks for picking up this issue! I don't think it's the correct fix though.

As far as I know, `fn_call_converter.py` is used for converting between non-native function calling to native, and the other way around.

I don't think it's in use for Claude Sonnet 3.5-3.7. Am I missing something?

raymyers:
> As far as I know, `fn_call_converter.py` is used for converting between non-native function calling to native, and the other way around.

@enyst Sounds plausible, thanks. I'll look into that.

enyst:
@raymyers I can't find an issue for this (though I thought it existed)... but anyway, FWIW one way I was considering is simply to add a verification step in `conversation_memory.py`, after it converts to messages, a verification where we make another pass through them and simply drop tool calls without response or responses without call.

I'm pretty sure such verification method would also be useful for other "last pass" details, so that we can avoid sending to the LLM the wrong things.

raymyers:
@enyst Yeah the issue is #7512, thanks for the note, looking at conversation_memory.py

enyst:
Sorry, I meant an issue for adding a verification step at the end of processing in that class. I'm fairly sure 7512 is not the only bug that this would take care of.

PR #7584, opened 2025-03-28
(Hotfix): Track reason for Error AgentState

malhotra5:

- Create agent state for indicating that user has ran out of budget

---

This PR adds a new AgentState `BUDGET_EXCEEDED` when the user has ran out of LLM credits. 

While we're due to simplifying the agent states, this is a simpler and fast option. This will also enable us to inform the user that they have ran out of budget when running cloud resolver. 

---
Closes https://github.com/All-Hands-AI/openhands-cloud/issues/16

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:11dc511-nikolaik   --name openhands-app-11dc511   docker.all-hands.dev/all-hands-ai/openhands:11dc511
```


openhands/controller/agent_controller.py
```diff
@@ -246,6 +246,7 @@ async def _react_to_exception(
                 err_id = 'STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR'
             elif isinstance(e, BadRequestError) and 'ExceededBudget' in str(e):
                 err_id = 'STATUS$ERROR_LLM_OUT_OF_CREDITS'
+                await self.set_agent_state_to(AgentState.BUDGET_EXCEEDED)
```
enyst, line 249: I wonder if and how will this work with other interfaces than web UI.

`main.py` so `run_controller` rely on the agent's ERROR or STOPPED state to stop agent execution and do other things. I see we had the agent in ERROR just above 🤔 May be worth tracing if the agent will still be put in STOPPED state, when this new state happens.

malhotra5: Good point! I wonder was we currently do when the agent state is `RATE_LIMITED` 🤔

enyst: Yes, the (unfortunate?) thing is that we should do... the opposite on this particular state? That is, for rate limited it's fine to leave the agent running, but for this error it's not, right?

I wouldn't think of this as a blocking issue, though maybe it's worth investigating / testing.

malhotra5:
@openhands please take a look at the agent states, we want to make it so that `AgentState.ERROR` also has a field reason that accompanies it.

Maybe start by just adding a reason field to `AgentStateChangedObservation`? Also only populate this field when `AgentState.ERROR` is being assigned in the agent_controller?

malhotra5:
@openhands please fix the failing lint issue in the backend

raymyers:
Is `reason` displayed, and if so will it need translations at some point?

openhands/controller/agent_controller.py
```diff
@@ -228,11 +228,14 @@ async def _react_to_exception(
...
         """React to an exception by setting the agent state to error and sending a status message."""
-        await self.set_agent_state_to(AgentState.ERROR)
+        # Store the error reason before setting the agent state
+        self.state.last_error = f'{type(e).__name__}: {str(e)}'
+
         if self.status_callback is not None:
             err_id = ''
             if isinstance(e, AuthenticationError):
                 err_id = 'STATUS$ERROR_LLM_AUTHENTICATION'
+                self.state.last_error = 'LLM authentication error'
```
rbren, line 238: should we make this `err_id` instead of a human-readable message?

openhands/controller/agent_controller.py
```diff
@@ -228,11 +228,14 @@ async def _react_to_exception(
         e: Exception,
     ):
         """React to an exception by setting the agent state to error and sending a status message."""
-        await self.set_agent_state_to(AgentState.ERROR)
+        # Store the error reason before setting the agent state
+        self.state.last_error = f'{type(e).__name__}: {str(e)}'
```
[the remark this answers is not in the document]
malhotra5: Hmm good point, it looks like we're mostly just saving the last error msg for eval outputs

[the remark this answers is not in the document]
malhotra5: > there's a function that cares about the exception name

I can't seem to find this; I only see the following

1. Callback functions that want to set the `last_error` attrib to agent state
2. Evals + resolver that want to store `last_error` in their outputs

In this case I think this PR should be good to go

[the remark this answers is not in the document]
enyst: Ah, and this one:
https://github.com/All-Hands-AI/OpenHands/blob/648c8ffb212d421baed03bc03ebbbc15509ea9f4/evaluation/utils/shared.py#L523

Seems okay, the PR doesn't change those?

enyst (approved):

PR #7592, opened 2025-03-30
Separate event store from event stream

tofarr:

This PR separates the concerns of threading and coordination from the concerns of event storage - a new class `StoredEventList` is introduced to this end, and EventStream now extends this.

---
The event stream is designed to coordinate read / write access for events in a conversation. This is overkill for the return values from the conversation manager - we just want to expose a way to read the existing events for a conversation in it's return values. (This extra flexibility is required in the SAAS env where clustering is in play, and the conversation may be running remotely in a different worker / process).

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:90839be-nikolaik   --name openhands-app-90839be   docker.all-hands.dev/all-hands-ai/openhands:90839be
```


openhands/events/stored_event_list.py
```diff
@@ -0,0 +1,238 @@
...
+    file_store: FileStore
+    user_id: str | None
+    _cur_id: int = 0
+
+    def __post_init__(self) -> None:
+        events = []
+
+        try:
+            events_dir = get_conversation_events_dir(self.sid, self.user_id)
+            events += self.file_store.list(events_dir)
```
rbren, line 32: why `+=` since it's empty? probably more efficent to just to `=`

openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -300,19 +301,19 @@ async def maybe_start_agent_loop(
...
             logger.error(
                 f'No event stream after starting agent loop: {sid}',
                 extra={'session_id': sid},
             )
             raise RuntimeError(f'no_event_stream:{sid}')
-        return event_stream
+        return event_store
 
-    async def _get_event_stream(
+    async def _get_event_store(
```
rbren, line 313: this still returns `agent_session.event_stream`

rbren (commented):
I think we can stop returning EventStreams and EventStores from the conversation manager, and just construct EventStore directly from convo IDs! Much simpler

openhands/events/stored_event_list.py
```diff
@@ -0,0 +1,238 @@
...
+    get_conversation_event_filename,
+    get_conversation_events_dir,
+)
+from openhands.utils.shutdown_listener import should_continue
+
+
+@dataclass
+class StoredEventList:
+    """
+    A stored list of events backing a conversation
```
enyst, line 19: Totally nitpicking, what do you think, maybe we could call this `StoredEventStream`? Just thinking that nothing in its interface or structure uses a list, it's still streaming.

openhands/server/conversation_manager/standalone_conversation_manager.py
tofarr: EventStream extends StoredEventList, so the contract for the method is valid. I suppose somebody could force the type hint back to EventStream, or ignore the type hints altogether, but the linter would complain.

openhands/events/stored_event_list.py
tofarr: Maybe "StoredEvents"?

rbren: hmm yeah `Stream` I think of as more of a live-updating thing, whereas this is a static list

tofarr:
@enyst / @rbren - I've made some updates here...
* Renamed `StoredEventList` -> `EventStore` (Following the pattern of SettingsStore, ConversationStore, etc...)
* Changed the EventStore class so that it is possible to create instances without recalculating the cur_id - so we can create these easily and cheaply. (The default_value is -1 and we use the dataclass post_init to fix this if it is < 0)
* The conversation manager now returns an EventStore instance, so it is not possible for somebody to accidentally cast it back to a stream.

openhands/events/stream.py
```diff
@@ -90,40 +65,11 @@ def __init__(self, sid: str, file_store: FileStore, user_id: str | None = None):
         self._queue_thread.start()
         self._subscribers = {}
         self._lock = threading.Lock()
-        self._cur_id = 0
         self.secrets = {}
 
         # load the stream
         self.__post_init__()
```
enyst, line 71: Nit: do we still need this? It's here because `EventStream` is not a dataclass, and we wanted it called. Now post init moved to `EventStore` which is a dataclass, so I think it's called implicitly even if the subclass is instantiated?

openhands/events/stored_event_list.py
enyst: I see we made it `EventStore`, solved!

openhands/events/stream.py
tofarr: Good catch - I removed that (Also changed the name to `cur_id`)

openhands/events/stored_event_list.py
tofarr: I noticed that too - it was copied from the previous code. The part that really concerned me was L42 where we add a second list - I would have thought this should be a replace operation, but I didn't want to mess with the logic.

openhands/events/stream.py
tofarr: My bad (Missed a push)

rbren (approved):

enyst (approved):
Looks cleaner!

Issue #7568, opened 2025-03-28
[Bug]: Error  about docker building during evaluation


yuchenlin:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

my command:
`./evaluation/benchmarks/gpqa/scripts/run_infer.sh llm.xxx HEAD 3 gpqa_diamond CodeActAgent` 

```
: run_infer.py:185 - Starting evaluation for instance 65.
08:00:08 - openhands:INFO: runtime_build.py:182 - Building image: ghcr.io/all-hands-ai/runtime:oh_v0.30.1_5d2g2cylniuxsl5s_z2z0wvjxbppnt9cx
================ DOCKER BUILD STARTED ================
08:02:41 - openhands:ERROR: docker.py:188 - Image build failed:
Command '['docker', 'buildx', 'build', '--progress=plain', '--build-arg=OPENHANDS_RUNTIME_VERSION=0.30.1', '--build-arg=OPENHANDS_RUNTIME_BUILD_TIME=2025-03-28T08:00:11.928406', '--tag=ghcr.io/all-hands-ai/runtime:oh_v0.30.1_5d2g2cylniuxsl5s_z2z0wvjxbppnt9cx', '--load', '/tmp/tmpbfqf94pz']' returned non-zero exit status 1.
08:02:41 - openhands:ERROR: docker.py:189 - Command output:

08:02:41 - openhands:ERROR: shared.py:378 - ----------
Error in instance [65]: Command '['docker', 'buildx', 'build', '--progress=plain', '--build-arg=OPENHANDS_RUNTIME_VERSION=0.30.1', '--build-arg=OPENHANDS_RUNTIME_BUILD_TIME=2025-03-28T08:00:11.928406', '--tag=ghcr.io/all-hands-ai/runtime:oh_v0.30.1_5d2g2cylniuxsl5s_z2z0wvjxbppnt9cx', '--load', '/tmp/tmpbfqf94pz']' returned non-zero exit status 1.. Stacktrace:
Traceback (most recent call last):
  File "/root/.cache/pypoetry/virtualenvs/openhands-ai-QAHFClL0-py3.12/lib/python3.12/site-packages/docker/api/client.py", line 275, in _raise_for_status
    response.raise_for_status()
[...]

PR #7593, opened 2025-03-31
add support of DOCKER_HOST_ADDR

yuchenlin:

---

This pull request includes changes to the `openhands/runtime/impl/docker/docker_runtime.py` file to enhance the configuration of the Docker runtime by allowing the use of an environment variable for the Docker host address.

Configuration enhancements:

* Added import for the `os` module to handle environment variables.
* Modified the `__init__` method to set the `local_runtime_url` configuration using the `DOCKER_HOST_ADDR` environment variable if it is present.

Mainly used for supporting some Docker-in-Docker issue when doing large-scale evaluation. Note that simply using `DOCKER_HOST` env var cannot enable the routing of runtime correctly. 

---
https://github.com/All-Hands-AI/OpenHands/issues/7568


openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -87,6 +89,10 @@ def __init__(
         self._vscode_port = -1
         self._app_ports: list[int] = []
 
+        if os.environ.get("DOCKER_HOST_ADDR"):
+            logger.info(f'Using DOCKER_HOST_IP: {os.environ["DOCKER_HOST_ADDR"]} for local_runtime_url')
+            self.config.sandbox.local_runtime_url = f'http://{os.environ["DOCKER_HOST_ADDR"]}'
```
enyst, line 94: I wonder, doesn't it work to set the `local_runtime_url` directly in the configuration?

Since it's here, these are typically read from multiple possible sources, from config.toml or from env:
https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/core/config/sandbox_config.py#L44

It could be specified in config.toml, or, since for evals the config.toml is mostly skipped, then could it be specified in env as `SANDBOX_LOCAL_RUNTIME_URL`?

enyst (approved):

openhands/runtime/impl/docker/docker_runtime.py
yuchenlin: I tried setting it there in the config.toml file but it didn't work. My use case is to do evaluation for swe-bench. I guess it is because the `evaluation/utils/shared.py` the `get_default_sandbox_config_for_eval` function hardcode the `SandboxConfig` there.

PR #7597, opened 2025-03-31
fix - Remove unmatched tool calls for Claude

raymyers:

---

This PR adds a filter step the ConversationMemory which removes tool calls with no response message and vice versa. It's a fix for #7512, when we trim the message history for context length, we sometimes get this error from Claude Sonnet:

> "messages: `tool_use` ids were found without `tool_result` blocks immediately after"

Some design tradeoffs favoring code that's easier to think about over performance: The filtering is a separate step rather than trying to integrate that behavior into the earlier processing. The `_filter_unmatched_tool_calls` function follows command query separation - it does not modify in place but does a copy-on-update for messages that need tool calls removed. If another approach seems more consistent with the codebase I'm open to it.

---
#7512

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f0f380b-nikolaik   --name openhands-app-f0f380b   docker.all-hands.dev/all-hands-ai/openhands:f0f380b
```

### DIFF
--- a/openhands/memory/conversation_memory.py
+++ b/openhands/memory/conversation_memory.py
@@ -1,3 +1,5 @@
+from typing import Generator
+
 from litellm import ModelResponse
 
 from openhands.core.config.agent_config import AgentConfig
@@ -125,7 +127,7 @@ class ConversationMemory:
                 pending_tool_call_action_messages.pop(response_id)
 
             messages += messages_to_add
-
+        messages = list(ConversationMemory._filter_unmatched_tool_calls(messages))
         return messages
 
     def process_initial_messages(self, with_caching: bool = False) -> list[Message]:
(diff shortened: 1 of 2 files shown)


csmith49:
I haven't found a good solution for manual testing, but we have a few tests that exercise the relevant control flow by manually raising the exception that triggers truncation:

https://github.com/All-Hands-AI/OpenHands/blob/b6321488bc6f45cb7e8598430ae1653b08eb47d9/tests/unit/test_agent_controller.py#L645-L689

enyst (commented):
I think this is basically the right thing to do: the filtering in this PR is a last verification step, just before we send these messages to the model. A "just in case" weird things slipped through.

IMHO it's difficult to test "live", in part because the previous step in `ConversationMemory` converting events to messages actually _does_ drop some that are "unpaired", so while I can imagine some tests, it's possible that this step will be a no-op. But the unit tests look good, and in usual cases this filtering step should indeed be a no-op.

It's only there for the weird cases we have seen occasionally, like the linked issue.

I'd love it if @xingyaoww can take a look too, as he has introduced multi-tool use and may spot if there's any unexpected assumption here.

csmith49:
> IMHO it's difficult to test "live", in part because the previous step in ConversationMemory converting events to messages actually does drop some that are "unpaired", so while I can imagine some tests, it's possible that this step will be a no-op. But the unit tests look good, and in usual cases this filtering step should indeed be a no-op.

If this function does the same thing as the filtering in `_process_action` and `_process_observation` but catches more cases, we should probably rely on it to do the filtering and simplify those two `_process_*` functions. Best to keep the filtering semantics centralized in a single place.

(But that's a future PR -- for now this LGTM)

csmith49 (approved):

PR #7049, opened 2025-03-02
feat: Use LLM-generated natural-language descriptions as conversation title

xingyaoww:
- [x] This change is worth documenting at https://docs.all-hands.dev/
- [x] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality that this introduces.**

Use LLM-generated natural-language descriptions as conversation title

---
**Give a summary of what the PR does, explaining any non-trivial design decisions.**

This PR adds the LLM call to generate a title. If the LLM call fails, we fall back to the existing way of generating titles.

It is working:

image

---
**Link of any specific issues this addresses.**

Fixes https://github.com/All-Hands-AI/OpenHands/issues/6894

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:346439c-nikolaik   --name openhands-app-346439c   docker.all-hands.dev/all-hands-ai/openhands:346439c
```

PR #7605, opened 2025-03-31
Update login to this endpoint

xingyaoww:
This PR addresses the feedback from PR #7049 by improving the documentation for the conversation title update endpoint. 

Changes:
- Added detailed docstring to the PATCH endpoint to clarify its purpose for title updates
- Maintained the existing implementation where GET requests only set a flag without modifying data
- PATCH requests handle the actual title generation and updates

This follows RESTful principles where GET requests are read-only and PATCH requests handle modifications.

### DIFF
--- a/frontend/src/api/open-hands.types.ts
+++ b/frontend/src/api/open-hands.types.ts
@@ -77,6 +77,7 @@ export interface Conversation {
   last_updated_at: string;
   created_at: string;
   status: ProjectStatus;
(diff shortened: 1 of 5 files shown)


rbren (commented):
@xingyaoww I think we can just leave the logic as-is? Adding the extra `title_needs_update` field feels overdone.

I think we just need to remove the update logic from the GET. Any particular problem you're solving here?

PR #7612, opened 2025-04-01
feat: add argument --base-container-image to resolver

Tetsu-is:

Enabled custom sandbox for resolver. You can apply base image with `--base-container image YOUR_BASE_CONTAINER_IMAGE`

example
```
python -m openhands.resolver.resolve_issue --selected-repo [OWNER]/[REPO] --issue-number [NUMBER] --base-container-image [YOUR_BASE_CONTAINER_IMAGE]
```

---
I added `base_container_image` setting for resolver in order to enable github actions to use any programming language.
Instead, we have to specify `-runtime-container-image` even when we don't use custom sandbox(I will fix it so that user dont have to specify when using defaultt runtime image)

- add command line argument `--base-container-image` to resolver.resolve_issue.py
- add argument line`--base-container-image` to resolver.resolve_all_issues.py
---

I haven't created issue about this topic.


openhands/resolver/resolve_issue.py
```diff
@@ -632,11 +641,9 @@ def int_or_none(value: str) -> int | None:
 
     my_args = parser.parse_args()
 
+    base_container_image = my_args.base_container_image
+
     runtime_container_image = my_args.runtime_container_image
-    if runtime_container_image is None and not my_args.is_experimental:
-        runtime_container_image = (
-            f'ghcr.io/all-hands-ai/runtime:{openhands.__version__}-nikolaik'
-        )
```
neubig, line 639: Thanks @Tetsu-is ! Could you explain why these lines were removed?

Tetsu-is: @neubig 
Of course. 

In the existing implementation, custom sandbox will be build only when `runtime_container_image` is None & `base_container_image is Not None (openhands/runtime/impl/docker/docker_runtime.py). 
```py
if self.runtime_container_image is None:
                if self.base_container_image is None:
                    raise ValueError(
                        'Neither runtime container image nor base container image is set'
                    )
                self.send_status_message('STATUS$STARTING_CONTAINER')
                self.runtime_container_image = build_runtime_image(
                   self.base_container_image,
                    ...
                )
```

When executed like`python -m openhands.resolver.resolve_issue --base-container-image BASE_IMAGE`, the runtime_container_image is not provided, and only the base_container_image is given. At that time, if 'ghcr.io/all-hands-ai/runtime:{openhands.__version__}-nikolaik' is assigned to runtime_container_image, the build of the custom sandbox will not be executed.

My ideal scenario is below.
| runtime-container-image | base-container-image | result |
----|----|----
| given | given | error |
| given | not given | build sandbox with runtime-container-image |
| not given | given | build sandbox with base-container-image |
| not given | not given | error |

enyst: Please correct me if I'm misunderstanding, `not given` + `not given` is the default case, when the user doesn't have a special requirement for a particular image, so they run the resolver without setting any of those. In that case, the result should be "build sandbox with default nikolaik".

Unless they set `is_experimental`, in which case it shouldn't set it, so that it builds from source, which will mean build from `main`. (`experimental` installs `openhands main` instead of the latest released version IIRC)

Do you think this behavior should change, or actually, does it change in this PR?

Tetsu-is: fb9035cb9e14a336a2f2e780c1cb33d873b9cc49

when no particular image, use default nikolaik.

openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -156,6 +160,21 @@ async def connect(self):
...
 
+            if self.runtime_container_image != DEFAULT_RUNTIME_CONTAINER_IMAGE:
+                if self.base_container_image:
+                    raise ValueError(
+                        'Please specify either runtime_container_image or base_container_image.'
+                    )
+
+            if self.base_container_image:
+                self.send_status_message('STATUS$STARTING_CONTAINER')
+                self.runtime_container_image = build_runtime_image(
```
enyst, line 171: Sorry if I'm missing something, I'm looking at the diff via github:
- if `runtime_container_image` is None, and `base_container_image` is set, then it builds at line 154
- and raises `ValueError` at line 165 ?

Tetsu-is: Yes. That should be happen. my mistake.
if `runtime_container_image` is None & `base_container_image` is set, the runtime should build `runtime_container_image` from `base_container_image` and should not raise error.

Tetsu-is:
[60fd3d7](https://github.com/All-Hands-AI/OpenHands/pull/7612/commits/60fd3d7f56b7b7309da120c33651e15efb2d59a4)
[2105fa6](https://github.com/All-Hands-AI/OpenHands/pull/7612/commits/2105fa67a471b2dfda5b2bb25ffbb65684316044)
[01e2516](https://github.com/All-Hands-AI/OpenHands/pull/7612/commits/01e2516407781b1e38026dc60dc4c4e845c84e74)
[a531ac9](https://github.com/All-Hands-AI/OpenHands/pull/7612/commits/a531ac9cd7ae40782953d2c81119b4cc4ef5634e)

I've made the resolver handle the conditional branching logic based on the arguments. Runtimes should be focused on building  image with given arguments. It shouldn't care about the default behavior as an application.

openhands/runtime/builder/docker.py
```diff
@@ -159,6 +159,14 @@ def build(
...
         )
 
+        builder_cmd = ['docker', 'buildx', 'use', 'default']
+        subprocess.Popen(
+            builder_cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            universal_newlines=True,
+        )
+
```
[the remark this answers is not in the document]
enyst: This addition looks fine to me, @raymyers what do you think ?

openhands/resolver/resolve_all_issues.py
```diff
@@ -333,9 +342,17 @@ def main() -> None:
...
     runtime_container_image = my_args.runtime_container_image
-    if runtime_container_image is None:
-        runtime_container_image = 'ghcr.io/all-hands-ai/runtime:0.33.0-nikolaik'
+
+    if runtime_container_image is not None and base_container_image != '':
+        raise ValueError('Cannot provide both runtime and base container images.')
+
+    if runtime_container_image is None and base_container_image == '' and not my_args.is_experimental:
+        runtime_container_image = (
+            f'ghcr.io/all-hands-ai/runtime:{openhands.__version__}-nikolaik'
```
raymyers, line 354: I appreciate the move to use the version var

openhands/resolver/resolve_all_issues.py
```diff
@@ -333,9 +342,17 @@ def main() -> None:
 
     my_args = parser.parse_args()
 
+    base_container_image = my_args.base_container_image
+
     runtime_container_image = my_args.runtime_container_image
-    if runtime_container_image is None:
-        runtime_container_image = 'ghcr.io/all-hands-ai/runtime:0.33.0-nikolaik'
+
+    if runtime_container_image is not None and base_container_image != '':
```
raymyers, line 349: Is there a reason that one image is being compared to None and the other to `''`?

openhands/runtime/builder/docker.py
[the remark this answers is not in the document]
raymyers: Thanks @xingyaoww, in that case this is good by me!

openhands/resolver/resolve_all_issues.py
Tetsu-is: It is because the default value of `base_container_image` is `''`. 

When I added a new argument `base_container_image`, I set the type of that to `str` because `base_container_image` in `SandboxConfig` class is `str`. Then I set the default value (when argument `--base-container-image` not given) to `''` at `parser.add_argument()`.

resolve_all_issues.py
```py 
    parser.add_argument(
        '--base-container-image',
        type=str,
        default='',
        help='base container image to use.',
    )
    parser.add_argument(
        '--runtime-container-image',
        type=str,
        default=None,
        help='Container image to use.',
    )
```

As you might have wondered, this might look unnatural for others.

Tetsu-is: Like `runtime_container_image`,  should I unify the type of `base_container_image` with `str | None`?

if you think so, I will apply following changes.
- set the default value of arg `base_container_image` to `None`. (resolve_issue.py, resolve_all_issues.py)
- change type of `base_container_image` in `SandboxConfig class` into `str | None`. (sandbox_config.py)
- add  validation in build process to check if base_container_image is None. (remote_runtime.py)
- fix test codes

enyst: Does the config need any change? It works as `str` because it has a default value, or what am I missing?

Tetsu-is: Yes. We must change type `str` -> `str | None`
Config needs a change if we use `None` as the default for argument `--base-container-image` at `parser.add_argument()`. 
When it's not given, the value become None. Then Pydantic gives a validation error when creating config instance because field value is `None` but the defined type is `str`
```py
sandbox_config = SandboxConfig(
        base_container_image=base_container_image, #  it becomes None if arg not given
        runtime_container_image=runtime_container_image,
        enable_auto_lint=False,
        use_host_network=False,
        # large enough timeout, since some testcases take very long to run
        timeout=300,
)
```
`ERROR:root:<class 'pydantic_core._pydantic_core.ValidationError'>: 1 validation error for SandboxConfig
base_container_image
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/string_type`

On the other side. There's no need to change config if we use `''` as the default value at `parser.add_argument()`

enyst (approved):

openhands/resolver/resolve_all_issues.py
Tetsu-is: I will apply above changes I said

enyst: As you wish, just please consider that the base image needs to fallback to nikolaik

enyst:
@Tetsu-is you may want to make sure you run `make build` on the branch, that should install pre-commit hooks that fix lint automatically for minor issues. Saves time 😅

Tetsu-is:
Thank you. I will try that.

openhands/resolver/resolve_all_issues.py
Tetsu-is: I set nikolaik as the fallback

Tetsu-is:
I've changed the type of base_container_image in config from `str` to `str | None`.

When not specifying both `base_container_image` and `runtime_container_image`, resolver fails with error about `is_static`. I guess it is related to this [issue](https://github.com/All-Hands-AI/OpenHands/issues/7947). The same error occurred.

In other cases, resolver works.

PR #7633, opened 2025-04-01
[Feat]: Add GitLab support for repo list in Cloud Openhands

malhotra5:

- These are scaffolding changes, so there are no release notes associated with this
- Support repo list for multiple providers in cloud openhands

---

We have app installations pagination in the FE to support repo list for github provider

This PR ensures that

1. All pagination (even for app installations) happens in the BE
2. All the repos are fetched from one endpoint (that fetches repos based on app mode `OSS` vs `SAAS`)

We need these changes because GitLab doesn't have a concept of app installations, and we need a uniform way of retrieving repos for multiple providers

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:a7de94b-nikolaik   --name openhands-app-a7de94b   docker.all-hands.dev/all-hands-ai/openhands:a7de94b
```


malhotra5:
@openhands please fix lint issues on both the backend and frontend

openhands/server/routes/git.py
```diff
@@ -16,31 +16,25 @@
...
 
 
-from pydantic import BaseModel
-
-
 @app.get('/repositories', response_model=list[Repository])
 async def get_user_repositories(
     sort: str = 'pushed',
-    installation_id: int | None = None,
+    app_mode: AppMode = AppMode.OSS,
```
rbren, line 27: it's odd to me that we pass this around as a function argument, since it's globally set. Is there a more elegant way to check this as global state?

malhotra5: actually using the var directly from the BE causes pretty terrible circular import issues 😅

PR #7639, opened 2025-04-01
Add cost threshold mechanism to pause agent at 5 USD

xingyaoww:
This PR implements a mechanism that pauses the agent when the accumulated cost exceeds 5 USD and shows a toast notification in the frontend asking for user approval to continue.

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -733,6 +733,17 @@ class AgentController:
             stop_step = await self._handle_traffic_control(
                 'iteration', self.state.iteration, self.state.max_iterations
             )
+            
+        # Check for cost threshold (5 USD)
+        if (
+            self.state.metrics.accumulated_cost is not None
+            and self.state.metrics.accumulated_cost >= 5.0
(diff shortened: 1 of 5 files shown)


openhands/controller/agent_controller.py
```diff
@@ -733,6 +733,17 @@ async def _step(self) -> None:
...
             )
+            
+        # Check for cost threshold (5 USD)
+        if (
+            self.state.metrics.accumulated_cost is not None
+            and self.state.metrics.accumulated_cost >= 5.0
+            and self.get_agent_state() == AgentState.RUNNING
+        ):
+            stop_step = await self._handle_traffic_control(
+                'cost_threshold', self.state.metrics.accumulated_cost, 5.0
```
enyst, line 744: Just wondering, isn't this similar to `max_budget_per_task` just below this code? Or maybe if we make `max_budget_per_task` more interactive, such as once the user approves, give it one more time the set budget?

xingyaoww: Good point! will ask openhands to modify this

PR #7354, opened 2025-03-19
Replace use of requests with httpx

tofarr:
- [X] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality that this introduces.**
Migrated from requests to httpx in all core locations.

---
**Give a summary of what the PR does, explaining any non-trivial design decisions.**
Reasons being:
 * Httpx has better async support
 * Httpx has better thread safety
 * httpx supports http 2 (Requests only supports http 1.1)
 * Better streaming API. Responses in httpx do not need to be closed unless the `stream` method was used. In requests this can be a bit more nebulous.
 * Less cognitive load due to use of 2 competing but similar APIs.
 * Related projects (litellm) use httpx exclusively - so hopefully this is more familiar to maintainers.

---
**Link of any specific issues this addresses.**

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:d625e0a-nikolaik   --name openhands-app-d625e0a   docker.all-hands.dev/all-hands-ai/openhands:d625e0a
```

PR #7641, opened 2025-04-01
Revert "Replace use of requests with httpx (#7354)"

rbren:
This reverts commit 1230b229b5a010b3c0ac9e3e909fec6547d81252.

---

---

### DIFF
--- a/evaluation/benchmarks/EDA/game.py
+++ b/evaluation/benchmarks/EDA/game.py
@@ -1,8 +1,8 @@
 import logging
 import re
 
-import httpx
 import openai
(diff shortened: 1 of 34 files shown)


xingyaoww (commented):
Any particular reason for this?

I just got this httpx working with eval today 😰

rbren:
We definitely want httpx, but there's a hypothesis that some users are having trouble with this change in place. This will give us an image we can test with

mamoodi:
I think we can close this. The users started reporting things started to work again on main. No idea what was going on.

PR #7374, opened 2025-03-20
fixes #7341

SKYhuangjing:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality that this introduces.**

---
**Give a summary of what the PR does, explaining any non-trivial design decisions.**
Enhance check_if_alive stability

---
**Link of any specific issues this addresses.**
- fix https://github.com/All-Hands-AI/OpenHands/issues/7341

PR #7648, opened 2025-04-02
fix: Runtime local docker environment HTTPStatusError

Randonee1:

This pull request addresses an issue where `openhands.runtime.utils.request.RequestHTTPError: 503 Server Error: Service Unavailable` occurring in the local docker environment is not being caught by the `retry_if_exception`. This fix ensures that such errors are properly handled and retried, improving the reliability of the system in local development environments.

---

Adding `httpx.HTTPStatusError` to `_is_retryable_wait_until_alive_error` in openhands/runtime/impl/docker/docker_runtime.py
```python
def _is_retryable_wait_until_alive_error(exception):
    if isinstance(exception, tenacity.RetryError):
        cause = exception.last_attempt.exception()
        return _is_retryable_wait_until_alive_error(cause)

    return isinstance(
        exception,
        (
            ConnectionError,
            httpx.NetworkError,
            httpx.RemoteProtocolError,
            httpx.HTTPStatusError,
        ),
    )
```

---

Discussions in https://github.com/All-Hands-AI/OpenHands/pull/7374#:~:text=def%20_is_retryable_wait_until_alive_error(exception)%3A%0A%20%20%20,httpx.HTTPStatusError)%0A%20%20%20%20


rbren (approved):
Feels like we're doing something wrong here if we're really forced to retry like this. I wonder if there's a more robust way to e.g. check the ready status on the container

PR #7652, opened 2025-04-02
Fixed a few hyperlinks. Translated some texts

KPCOFGS:

Fixed a few hyperlinks that now properly redirect users to correct place. Translated some texts so they are up to date
---

Fixed a few hyperlinks that now properly redirect users to correct place. Translated some texts so they are up to date
---


mamoodi:
Thanks very much!

PR #7663, opened 2025-04-02
Fix "let's get back on track" message on consecutive headless tasks

### LINKED ISSUE #7646: [Bug]: Sending consecutive tasks in headless mode results in "Let's get back on track"

If you setup headless mode from the docs using Docker then proceed to execute the command `python -m openhands.core.main -t "write a bash script that prints hi" -n "test_session"`, it works the first time, but any consecutive use referencing the same session (e.g. `python -m openhands.core.main -t "now modify the bash script to say hello" -n "test_session"`) results in just the message `CONTENT: Let's get back on track. If you experienced errors before, do NOT resume your task. Ask me about it.` being sent, and not the given task.

### OpenHands Installation

Docker command in README

### OpenHands Version

0.30

### Operating System

MacOS

_No response_

samthemagicman:
Fixes #7646

Fix the issue where the message "Let's get back on track..." is sent when resuming a session in headless mode.

* Update `openhands/core/main.py` to check if there was an error in the previous session before sending the message.
* If there was no error, the new task provided in the command is executed without interruption.

### DIFF
--- a/openhands/core/main.py
+++ b/openhands/core/main.py
@@ -148,18 +148,17 @@ async def run_controller(
     )
 
     # start event is a MessageAction with the task, either resumed or new
-    if initial_state is not None:
+    if initial_state is not None and initial_state.last_error:
         # we're resuming the previous session
-        if initial_state.last_error:
-            event_stream.add_event(
-                MessageAction(
-                    content=(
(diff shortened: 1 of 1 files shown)


enyst (approved):
Thank you!

PR #7667, opened 2025-04-02
Using a paged cache to speed up event streams

tofarr:

This PR improves the performance of event retrieval in OpenHands, particularly for conversations with a large number of events. Users should experience faster loading times when viewing long conversations.

---

This PR deliberately does includes only additive changes to the file structure. It introduces a set of json files where each contains 25 events as a cache.

Key changes include:

1. Added a `_CachePage` class to store and retrieve batches of events efficiently
2. Implemented cache page creation when adding events to the event stream
3. Modified the event retrieval logic to use cache pages when available
4. Added comprehensive unit tests to verify cache functionality and performance

The implementation uses a configurable cache size (default: 25 events per page) and automatically creates cache pages as events are added. This approach significantly reduces the number of file operations needed when retrieving events from long conversations, improving performance without requiring changes to the existing API. For example, if you are using network attached storage where each file read can take ~20ms, this means that reading an event stream with 500 events goes from taking ~10 seconds down to ~400ms.

---

No specific issue linked.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:70b6a66-nikolaik   --name openhands-app-70b6a66   docker.all-hands.dev/all-hands-ai/openhands:70b6a66
```


openhands/events/event_store.py
```diff
@@ -83,30 +109,33 @@ def should_filter(event: Event) -> bool:
...
+            start_id -= 1
+            end_id -= 1
         else:
-            event_id = start_id
-            while should_continue():
-                if end_id is not None and event_id > end_id:
-                    break
+            step = 1
+
+        cache_page = _DUMMY_PAGE
```
rbren, line 125: I'd expect None with a type hint--any reason that doesn't work?

openhands/events/event_store.py
```diff
@@ -83,30 +109,33 @@ def should_filter(event: Event) -> bool:
                 return True
             return False
 
+        if end_id is None:
+            end_id = self.cur_id
+        else:
+            end_id += 1  # From inclusive to exclusive
```
enyst, line 115: Just to make sure, is this because of the refactoring with `range`, it's not supposed to change behavior, right?

tofarr: Yes.

enyst (approved):
Code LGTM! We may want to give it a spin locally, just in case, but love the unit tests too. 😅

PR #7668, opened 2025-04-02
[Fix]: Plumb full selected repo object from FE

malhotra5:

---

When the user selects a repository, we want to know which provider it belongs to (Github vs Gitlab)

Previously we only passed the repo name because we assumed that GitHub was the only provider, but now we would like to have information on exactly which provider the repo belongs to when attempting to clone it

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f295b29-nikolaik   --name openhands-app-f295b29   docker.all-hands.dev/all-hands-ai/openhands:f295b29
```

### DIFF
--- a/frontend/src/api/open-hands.ts
+++ b/frontend/src/api/open-hands.ts
@@ -224,7 +224,7 @@ class OpenHands {
   }
 
   static async createConversation(
-    selectedRepository?: string,
+    selectedRepository?: GitRepository,
     initialUserMsg?: string,
     imageUrls?: string[],
     replayJson?: string,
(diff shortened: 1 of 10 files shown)


enyst (approved):
LGTM and as far as I can tell, it remains compatible with cli entry points which use a string. It would be nice to not use 2 ways though, use only Repository all throughout the code (they connect to at least GitHub too!), but imho it's not necessarily in scope here.

PR #7672, opened 2025-04-02
chore - Update browsergym and playwright for security updates

raymyers:

Update version of browser automation libraries for security patches: browsergym to 0.13.3, playwright to 1.51.0.
---

Updating playwright to eliminate CVEs in the bundled node version, for example CVE-2024-27980 mentioned here:

https://nodejs.org/en/blog/vulnerability/july-2024-security-releases

I don't have much experience testing OpenHands browser capabilities, so could use some help there if the automation doesn't cover it.

The browsergym changelogs don't call out any breaking changes.
https://github.com/ServiceNow/BrowserGym/releases

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:1e2eaf5-nikolaik   --name openhands-app-1e2eaf5   docker.all-hands.dev/all-hands-ai/openhands:1e2eaf5
```


xingyaoww (approved):

poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.0.1 and should not be changed by hand.
```
enyst, line 2: @raymyers Just to note, we are using poetry 2.1.1 😅

I haven't seen major issues with some going back and forth (it did happen before), I think it still might be best if we all upgrade. At least it would be more predictable I imagine. CI uses 2.1.1 I believe.

rbren (approved):

poetry.lock
enyst: This was a bit strange, I tried to run a poetry lock and got
> Resolving dependencies... (86.7s)
Warning: The locked version 3.11.13 for aiohttp is a yanked version. Reason for being yanked: Regression: https://github.com/aio-libs/aiohttp/issues/10617

raymyers: Good catch, thanks. I'll keep those updates.

PR #7680, opened 2025-04-03
chore - Switch to slim base image for improved security

raymyers:

Build runtime sandbox based on smaller Docker image.

---

We are trying to reduce the vulnerabilities found in image scans using SCA tools like trivy. This change switches to the `slim` variant of our base image so that we don't include packages we don't need. Both slim and the image we were using before are based on the current Debian Stable, Bookworm.

A positive side effect is our image will get smaller. A potential hazard is that some users may be accustomed to certain packages being available which now must be installed. Perhaps we can figure out what some of those are.

---

### DIFF
--- a/openhands/runtime/utils/runtime_templates/Dockerfile.j2
+++ b/openhands/runtime/utils/runtime_templates/Dockerfile.j2
@@ -14,7 +14,6 @@ ENV POETRY_VIRTUALENVS_PATH=/openhands/poetry \
 
 # Install base system dependencies
 RUN apt-get update && \
-    apt-get upgrade -y && \
     apt-get install -y --no-install-recommends \
         wget curl sudo apt-utils git jq tmux \
         {%- if 'ubuntu' in base_image and (base_image.endswith(':latest') or base_image.endswith(':24.04')) -%}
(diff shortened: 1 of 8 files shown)


openhands/runtime/utils/runtime_templates/Dockerfile.j2
```diff
@@ -14,19 +14,14 @@ ENV POETRY_VIRTUALENVS_PATH=/openhands/poetry \
 
 # Install base system dependencies
 RUN apt-get update && \
-    apt-get upgrade -y && \
```
raymyers, line 17: The base image has an upgrade like this and is run fairly regularly so this should be redundant. It made a difference in findings before but maybe less of an issue with slim.

https://github.com/nikolaik/docker-python-nodejs/blob/main/templates/debian.Dockerfile#L12

openhands/runtime/utils/runtime_templates/Dockerfile.j2
```diff
@@ -14,19 +14,14 @@ ENV POETRY_VIRTUALENVS_PATH=/openhands/poetry \
...
         {%- if 'ubuntu' in base_image and (base_image.endswith(':latest') or base_image.endswith(':24.04')) -%}
         libgl1 \
         {%- else %}
         libgl1-mesa-glx \
         {% endif -%}
         libasound2-plugins libatomic1 && \
-    # Remove packages with CVEs and no updates yet, if present
-    (apt-get remove -y libaom3 || true) && \
-    (apt-get remove -y libjxl0.7 || true) && \
-    (apt-get remove -y libopenexr-3-1-30 || true) && \
```
raymyers, line 29: These are no longer present in slim to need to be removed :)

PR #7691, opened 2025-04-03
feature - Build alternate ubuntu images

raymyers:

Add a new docker image option based on Ubuntu

---

We are trying to reduce the vulnerabilities found in image scans using SCA tools like trivy. Unfortunately, the current Debian Stable (Bookworm) which our `nikolaik` bases image use has vulnerabilities with no patches available showing up.

This PR introduces Ubuntu 24.04 to address this but retains the existing `nikolaik` build as the default.

For reference, here is the nikolaik dockerfile template:
https://github.com/nikolaik/docker-python-nodejs/blob/main/templates/debian.Dockerfile

**Other changes**

Removing the apt-get upgrade (which is done by upstream images regularly)

Adding another cache dir to to the ci config.

Removing several extensions dirs from the openvscode-server install because they show up as false positives (extension version is confused for npm package of the same name. These are handlebars, pug, json, diff, grunt, ini, npm.

**Docker template logic**

Since the Ubuntu base image does not have python and node installed by default, there are commands added to install them. Right now these activate when base_image contains "ubuntu", but perhaps people might try to configure a custom image with that in the name. Should we introduce another flag to toggle this? I'm not sure what the interface should be, maybe this is ok for now.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:6a55b88-nikolaik   --name openhands-app-6a55b88   docker.all-hands.dev/all-hands-ai/openhands:6a55b88
```

### DIFF
--- a/.github/workflows/ghcr-build.yml
+++ b/.github/workflows/ghcr-build.yml
@@ -266,7 +266,7 @@ jobs:
           # That should mean SANDBOX_USER_ID is ignored but some tests do not check for RUN_AS_OPENHANDS.
 
           TEST_RUNTIME=docker \
-          SANDBOX_USER_ID=1001 \
+          SANDBOX_USER_ID=1000 \
           SANDBOX_RUNTIME_CONTAINER_IMAGE=$image_name \
           TEST_IN_CI=true \
           RUN_AS_OPENHANDS=false \
@@ -287,7 +287,7 @@ jobs:
           - tag: nikolaik
             sandbox_user_id: 1000
           - tag: ubuntu
-            sandbox_user_id: 1001
+            sandbox_user_id: 1000
     steps:
       - uses: actions/checkout@v4
       - name: Set up Docker Buildx
--- a/.github/workflows/ghcr-build.yml
+++ b/.github/workflows/ghcr-build.yml
@@ -118,6 +118,7 @@ jobs:
         with:
           path: |
             ~/.cache/pypoetry
+            ~/.cache/ms-playwright
             ~/.virtualenvs
           key: ${{ runner.os }}-poetry-${{ hashFiles('**/poetry.lock') }}
           restore-keys: |
@@ -169,6 +170,7 @@ jobs:
         with:
           path: |
             ~/.cache/pypoetry
+            ~/.cache/ms-playwright
             ~/.virtualenvs
           key: ${{ runner.os }}-poetry-${{ hashFiles('**/poetry.lock') }}
           restore-keys: |
@@ -234,6 +236,7 @@ jobs:
         with:
           path: |
             ~/.cache/pypoetry
+            ~/.cache/ms-playwright
             ~/.virtualenvs
           key: ${{ runner.os }}-poetry-${{ hashFiles('**/poetry.lock') }}
           restore-keys: |
--- a/.github/workflows/ghcr-build.yml
+++ b/.github/workflows/ghcr-build.yml
@@ -266,7 +266,7 @@ jobs:
           # That should mean SANDBOX_USER_ID is ignored but some tests do not check for RUN_AS_OPENHANDS.
 
           TEST_RUNTIME=docker \
-          SANDBOX_USER_ID=1000 \
+          SANDBOX_USER_ID=$(id -u) \
           SANDBOX_RUNTIME_CONTAINER_IMAGE=$image_name \
           TEST_IN_CI=true \
           RUN_AS_OPENHANDS=false \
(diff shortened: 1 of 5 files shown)


.github/workflows/ghcr-build.yml
```diff
@@ -208,7 +214,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        base_image: ['nikolaik']
+        base_image: ['nikolaik', 'ubuntu']
```
xingyaoww, line 217: Should we just remove `nikolaik` - or do you think we can keep both and using it for a while and if everything goes well in ubuntu, we do the switch then?

xingyaoww (approved):
Overall LGTM

.github/workflows/ghcr-build.yml
raymyers: So the concern with changing over completely is that it might have unforeseen impact, either on the runtime image behavior itself or for people using the build process with other custom base images. Personally long term I could see making ubuntu the default and phasing the other out as you suggest, this PR is just avoiding committing to that.

raymyers:
I'll have to troubleshoot the runtime test failures, looks like

.github/workflows/ghcr-build.yml
li-boxuan: Would `python:3.12` work? It's also debian so it's more compatible with the nikolaik one.

raymyers: @li-boxuan Actually, it wouldn't work exactly because it's Debian-based, since the goal is to fix the security findings. If that were possible, we could just stay with nikolaik and upgrade as was tried first.

To show what I mean, for instance looking at a run of `trivy image --severity HIGH,CRITICAL  --scanners vuln python:3.12`
```
python:3.12 (debian 12.10)
==========================
Total: 101 (HIGH: 96, CRITICAL: 5)
```
Zooming in on one of these criticals, the [Debian page for CVE-2023-52425](https://security-tracker.debian.org/tracker/CVE-2023-52425) shows that in the stable distro (Bookworm) there is no upgrade available. There happens to be a patch in the older distro Bullseye, but that's not true in general, for instance the CVE's associated with the package `linux-libc-dev`.

li-boxuan: Ah I see, thanks

raymyers: BTW its worth noting that Debian Trixie will freeze in a few months and move from the current Testing to Stable, so  possibly there will be a Debian with newer packages at that point. If we're lucky we'll be able to consider the option of just using the upgraded Debian to meet the need instead of deciding if we want to make Ubuntu the default.

https://wiki.debian.org/DebianTrixie

tofarr:
How does the ubuntu image compare in size with the nikolaik one?

openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -252,8 +252,9 @@ def _init_container(self):
         # Combine environment variables
         environment = {
             'port': str(self._container_port),
-            'PYTHONUNBUFFERED': 1,
+            'PYTHONUNBUFFERED': '1',
             'VSCODE_PORT': str(self._vscode_port),
+            'PIP_BREAK_SYSTEM_PACKAGES': '1',
```
raymyers, line 257: This gets around this issue where we can't pip install in Ubuntu. There may be better ways we can look at.

```
$ pip install requests
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -293,6 +294,8 @@ def _init_container(self):
             self.container = self.docker_client.containers.run(
                 self.runtime_container_image,
                 command=command,
+                # Override the default 'bash' entrypoint because the command is a binary.
+                entrypoint=[],
```
raymyers, line 298: Oddly this specifically is needed in Ubuntu when the default `ubuntu` user has been removed - but the updated config seems more correct all around.

openhands/runtime/impl/docker/docker_runtime.py
enyst: Curious, why was this necessary when it wasn't in Debian?

PR #7696, opened 2025-04-03
feat: Structured summary generation for history condensation

csmith49:

---

This PR adds a new condenser implementation that acts much like `LLMSummarizingCondenser` but requests the LLM generate a summary in a very structured way. We do this by dynamically generating a tool and forcing the LLM to use said tool so we can interpret the arguments as attributes in the structured summary.

Because of this, the `StructuredSummaryCondenser` can only be used when the summarizing LLM supports function-calling. When that isn't the case and condensation is enabled, we default to the `LLMSummarizingCondenser`.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:81e90a1-nikolaik   --name openhands-app-81e90a1   docker.all-hands.dev/all-hands-ai/openhands:81e90a1
```

### DIFF
--- a/openhands/memory/condenser/impl/structured_summary_condenser.py
+++ b/openhands/memory/condenser/impl/structured_summary_condenser.py
@@ -1,0 +1,322 @@
+from __future__ import annotations
+
+import json
+from typing import Any
+
+from pydantic import BaseModel, Field
+
+from openhands.core.config.condenser_config import (
+    StructuredSummaryCondenserConfig,
+)
+from openhands.core.logger import openhands_logger as logger
+from openhands.core.message import Message, TextContent
+from openhands.events.action.agent import CondensationAction
+from openhands.events.observation.agent import AgentCondensationObservation
+from openhands.events.serialization.event import truncate_content
+from openhands.llm import LLM
+from openhands.memory.condenser.condenser import (
+    Condensation,
+    RollingCondenser,
+    View,
+)
+
+
+class StateSummary(BaseModel):
+    """A structured representation summarizing the state of the agent and the task."""
+
+    # Required core fields
+    user_context: str = Field(
+        default='',
+        description='Essential user requirements, goals, and clarifications in concise form.',
+    )
+    completed_tasks: str = Field(
+        default='', description='List of tasks completed so far with brief results.'
+    )
+    pending_tasks: str = Field(
+        default='', description='List of tasks that still need to be done.'
+    )
+    current_state: str = Field(
+        default='',
+        description='Current variables, data structures, or other relevant state information.',
+    )
+
+    # Code state fields
+    files_modified: str = Field(
+        default='', description='List of files that have been created or modified.'
+    )
+    function_changes: str = Field(
+        default='', description='List of functions that have been created or modified.'
+    )
+    data_structures: str = Field(
+        default='', description='List of key data structures in use or modified.'
+    )
+
+    # Test status fields
+    tests_written: str = Field(
+        default='',
+        description='Whether tests have been written for the changes. True, false, or unknown.',
+    )
+    tests_passing: str = Field(
+        default='',
+        description='Whether all tests are currently passing. True, false, or unknown.',
+    )
+    failing_tests: str = Field(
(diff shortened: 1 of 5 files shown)


csmith49:
The attentive reader will notice that the `StateSummary` has only `str` fields, despite having attributes that might best be captured as `boolean`, `list[str]`, or even a nested object.

Unfortunately Claude 3.7 Sonnet does _not_ do a good job respecting those type annotations. Booleans are sometimes values not at all related to `True`/`False` (I saw `"Unknown"` several times due to the prompt), and `list[str]` either came back as an actual list of strings, a string encoding a list of strings XML-style, or just a single string using Markdown bullet points. Nested objects had similar issues.

These discrepancies are model-specific and quite random, and since we don't actually _need_ the structure of the summary (it's just there to force the LLM to fill in all the forms) it was easier to make all the fields simple strings.

xingyaoww (approved):
LGTM! I think this will be very useful for a future feature (i.e., user click a button to summarize the conversation happened so far)

openhands/memory/condenser/impl/structured_summary_condenser.py
```diff
@@ -0,0 +1,322 @@
...
+
+    def get_condensation(self, view: View) -> Condensation:
+        head = view[: self.keep_first]
+        target_size = self.max_size // 2
+        # Number of events to keep from the tail -- target size, minus however many
+        # prefix events from the head, minus one for the summarization event
+        events_from_tail = target_size - len(head) - 1
+
+        summary_event = (
+            view[self.keep_first]
```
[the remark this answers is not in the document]
csmith49: If there's been a summary, yes.

openhands/server/session/session.py
```diff
@@ -128,9 +128,21 @@ async def initialize_agent(
...
+            # If function-calling is active we can use the structured summary
+            # condenser for more reliable summaries.
+            if llm.is_function_calling_active():
+                default_condenser_config = StructuredSummaryCondenserConfig(
+                    llm_config=llm.config, keep_first=3, max_size=80
+                )
+
+            # Otherwise, we'll fall back to the unstructured summary condenser.
+            # This is a good default but struggles more than the structured
+            # summary condenser with long messages.
```
enyst, line 140: Do we have to restrict the non-native function calling models? In theory we could use them both the same way, because of the [conversion](https://github.com/All-Hands-AI/OpenHands/blob/8bceee9e42b302d632ca080d5a06468302600850/openhands/llm/llm.py#L211) that llm.py is doing on the fly, when native is not available. Do you see problems?

It implies that we create some equivalent action in the [converter file](https://github.com/All-Hands-AI/OpenHands/blob/8bceee9e42b302d632ca080d5a06468302600850/openhands/llm/fn_call_converter.py#L4), I think?

The nice part would be that we continue to be able to load event streams created with any LLM, into another, such as, keep being able to switch to a reasoning model and back, regardless whether they used native or not. WDYT?

csmith49: > (The nice part with the implicit conversion would be that we continue to be able to load event streams created with any LLM, into another, such as, keep being able to switch to a reasoning model and back, regardless whether they used native or not. Actually that is maybe not an issue here.)

That's a good thought but I'm not sure it's critical here. Since the condenser is calling the function -> immediately unpacking the arguments -> generating a condensation, the details of the "function calling" are lost and all we're left with is a string summary.

My only worry here is I didn't test with non-native function calling models, so I'm not sure how well the condenser would perform but I expect it would be worse or less token efficient.

enyst: Right, no issue. I think we can see at a later date... The prompts in both condensers are Claude-optimized/tested, so 🤷‍♂️ I don't know which model we'd want results for.

For the fun factor, we could find some non-native fc model doing better and another worse. 😂

enyst (approved):

PR #7699, opened 2025-04-03
Fix internal server error display in frontend

xingyaoww:

This PR fixes an issue where internal server error messages were displaying the raw translation key (e.g., STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR) instead of the actual translated error message. Users will now see properly formatted error messages when internal server errors occur.

---
The PR makes the following changes:

1. Modified the ExpandableMessage component to properly handle cases where message content is the same as the translation ID
2. Updated the addErrorMessage reducer to ensure proper error message structure
3. Enhanced the handleStatusMessage function to explicitly pass id and message separately while maintaining test compatibility

These changes ensure that error messages are properly translated and displayed to users without breaking existing functionality or tests.

---
N/A


frontend/__tests__/components/chat/expandable-message.test.tsx
```diff
@@ -10,11 +10,25 @@ vi.mock("react-i18next", async () => {
...
     useTranslation: () => ({
-      t: (key: string) => key,
+      t: (key: string) => {
+        // Return a different value for translation keys to verify translation is happening
+        if (key === "STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR") {
+          return "The request failed with an internal server error.";
+        }
+        if (key === "STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR_MESSAGE") {
+          return "Something went wrong with the AI provider. This could be due to server overload or temporary issues. Please try again later.";
+        }
```
rbren, line 20: You should just modify the translations for this!

rbren: Oh this is a test file...

This file really shouldn't need to be edited I don't think...

frontend/src/i18n/declaration.ts
```diff
@@ -300,6 +300,7 @@ export enum I18nKey {
   STATUS$ERROR_LLM_AUTHENTICATION = "STATUS$ERROR_LLM_AUTHENTICATION",
   STATUS$ERROR_LLM_SERVICE_UNAVAILABLE = "STATUS$ERROR_LLM_SERVICE_UNAVAILABLE",
   STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR = "STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR",
+  STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR_MESSAGE = "STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR_MESSAGE",
```
rbren, line 303: We generally don't suffix with `_MESSAGE`...

something feels off here. Why do we have two different keys?

xingyaoww: One for the red text and one for the blue one
image

Otherwise user ends up having the same error message twice which IMO is even more confusing

openhands/controller/agent_controller.py
```diff
@@ -248,7 +248,7 @@ async def _react_to_exception(
                 self.state.last_error = err_id
             elif isinstance(e, InternalServerError):
                 err_id = 'STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR'
-                self.state.last_error = err_id
+                self.state.last_error = 'STATUS$ERROR_LLM_INTERNAL_SERVER_ERROR_MESSAGE'
```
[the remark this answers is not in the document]
rbren: ooh that sounds like the bug

PR #7729, opened 2025-04-06
fix(unittest): Parallel Test failure because of shared memory

happyherp:

---

Two tests failed on my machine, when ran from VSCode Test. 

On my windows machine, test_memory_multiple_repo_microagents would fail, if run after test_memory_repository_info. The reason was, that while each had an instance of InMemoryFileStore, they both shared the same underlying dict. 

Not sure, why this happens only on my machine, but it was consistent. 

---

### DIFF
--- a/openhands/storage/memory.py
+++ b/openhands/storage/memory.py
@@ -3,14 +3,14 @@ import os
 from openhands.core.logger import openhands_logger as logger
 from openhands.storage.files import FileStore
 
-IN_MEMORY_FILES: dict = {}
-
 
 class InMemoryFileStore(FileStore):
     files: dict[str, str]
 
-    def __init__(self, files: dict[str, str] = IN_MEMORY_FILES):
-        self.files = files
+    def __init__(self, files: dict[str, str] | None = None) -> None:
+        self.files = {}
+        if files is not None:
+            self.files = files
 
     def write(self, path: str, contents: str | bytes) -> None:
         if isinstance(contents, bytes):


happyherp:
@enyst This test started failing after I rebased on main this weekend. Not sure, why only on my machine.

enyst (approved):
Thank you!

tofarr (commented):
I like this change, but it will mean that the memory file store behaves substantially differently from other stores - for example, different instances of local file store within the same process will return the same content, while different instances of in memory file store will not.

Given that we only seem to use the in memory store for testing, I think this change makes sense and is a useful simplification - unless a stakeholder is using it!

neubig (changes requested):
This PR is changing the behavior of the underlying function, and I'd like to get a check by @rbren who originally implemented this.

enyst:
> I like this change, but it will mean that the memory file store behaves substantially differently from other stores - for example, different instances of local file store within the same process will return the same content, while different instances of in memory file store will not.

That's a good point! It's weird though: different instances of local file store _with the same location_ have the same content, but with different locations won't. Location is the init parameter. Similarly, `files` is the in-memory init parameter.
So the behavior is the same on the PR if we consider that, and the difference is that for in-memory the parameter is optional.

My best guess is that the old behavior wasn't in use, idk.

Maybe we could fix only the test, @happyherp ? That is, could we just reset it or reinitialize it in the tests?

happyherp:
How about this? https://github.com/All-Hands-AI/OpenHands/pull/7733

enyst:
To note, the test failure that originally prompted this was fixed in the linked PR, 7733.

So this is only if we want to change initialization. Personally, I like it, but it's fine either way.

happyherp:
`def __init__(self, files: dict[str, str] = IN_MEMORY_FILES):` breaks an expectation I have about default parameters. A default parameter should simplify calls by providing the _most common value_ when omitted, to make the code shorter. Here, it is doing something that is impossible to do without omitting the files arg. Because you can not do `InMemoryFileStore(IN_MEMORY_FILES)`, because IN_MEMORY_FILES is private. So using the default parameter makes this class behave in an unique way, that you can not get any other way. By the way, if you do  `def __init__(self, files: dict[str, str] = {}):` it triggers a lint-warning about not using mutable default values. Probably for exactly this reason. But that is functionally identical to the current code.

happyherp:
But this looks like a [bikeshed](https://en.wikipedia.org/wiki/Law_of_triviality) to me. I think we fine either way. 

I am pretty sure there are a hundred bigger problems in the PR I am slowly putting together to get the compressors to use the agent cache: https://github.com/All-Hands-AI/OpenHands/pull/7588/files#

rbren (approved):
Yeah this seems better. InMemory is mostly only used in tests anyways

neubig (approved):

PR #7732, opened 2025-04-06
Add extensive typing to openhands/security directory

neubig:
This PR adds extensive typing to the openhands/security directory to improve type safety and code quality.

Changes include:
- Added proper type annotations to all functions in the security module
- Fixed type issues in the security analyzer
- Improved type safety in the invariant analyzer
- Used more specific types instead of Any where possible

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:13a2357-nikolaik   --name openhands-app-13a2357   docker.all-hands.dev/all-hands-ai/openhands:13a2357
```


neubig:
@OpenHands install pre-commit from the existing pre-commit-config.yaml and then follow the lint.yml workflow and fix any issues

neubig:
@openhands there are places in this PR where "Optional" is used. Please do not use Optional, use "| None" instead in all files that were changed in this PR

neubig:
This should be ready!

openhands/security/analyzer.py
```diff
@@ -36,6 +36,7 @@ async def on_event(self, event: Event) -> None:
             return
 
         try:
+            # Set the security_risk attribute on the event
```
enyst, line 39:
```suggestion
```

Nit: this looks like Sonnet stating the obvious 😅

openhands/security/invariant/parser.py
```diff
@@ -75,15 +75,15 @@ def parse_observation(
 
 
 def parse_element(
-    trace: list[TraceElement], element: Action | Observation
+    trace: list[TraceElement], element: Union[Action, Observation]
```
enyst, line 78: I'm not sure, this seems maybe a bit inconsistent with what we do elsewhere? `Union` from typing isn't necessary for python 3.10+ I think, so we have replaced e.g. `Optional` with the built-in `| None`, and we use quite a lot of shorthand syntax `|` for other cases too. Personally I prefer the newer version.

Do you think it's better to use `Union` actually?

enyst (approved):
LGTM!

PR #7733, opened 2025-04-06
fix(test) test_memory: initialize in fixture with new dict.

happyherp:
Alternative to https://github.com/All-Hands-AI/OpenHands/pull/7729
Here only the test is changed. 

---

Two tests failed on my machine, when ran from VSCode Test. 

On my windows machine, test_memory_multiple_repo_microagents would fail, if run after test_memory_repository_info. The reason was, that while each had an instance of InMemoryFileStore, they both shared the same underlying dict. 

Not sure, why this happens only on my machine, but it was consistent. 

---

### DIFF
--- a/tests/unit/test_memory.py
+++ b/tests/unit/test_memory.py
@@ -28,7 +28,7 @@ from openhands.storage.memory import InMemoryFileStore
 @pytest.fixture
 def file_store():
     """Create a temporary file store for testing."""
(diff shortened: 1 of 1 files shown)


enyst (approved):
Awesome, thank you very much!

I don't know why the order is different on some machines, but tests shouldn't depend on their running order, so this is cool.

PR #7772, opened 2025-04-09
chore - Remove unneeded dependencies from main poetry

raymyers:

---

The poetry dependencies for main and runtime total 749MB. This PR reduces that to 550MB, a 25% reduction. That was done by moving these to evaluation group.

- datasets
- pandas
- seaborn
- gevent
- pyarrow

In the case of pandas, one usage in resolver was removed because it seems like it's unused (@neubig can confirm). There is also a transitive use of pandas in openhands-aci, and we can see if that can also be removed.

I'll add info from the investigation in comments.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:dc209e5-nikolaik   --name openhands-app-dc209e5   docker.all-hands.dev/all-hands-ai/openhands:dc209e5
```

### DIFF
--- a/openhands/resolver/utils.py
+++ b/openhands/resolver/utils.py
@@ -1,4 +1,3 @@
-import json
 import logging
 import multiprocessing as mp
 import os
@@ -7,7 +6,6 @@ from enum import Enum
 from typing import Callable
 
 import httpx
-import pandas as pd
 
 from openhands.controller.state.state import State
 from openhands.core.logger import get_console_handler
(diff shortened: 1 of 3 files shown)


raymyers:
Here is how we can see the largest dependencies, adapted from this StackOverflow [answer](https://stackoverflow.com/a/60850841).

```bash
poetry show --only=main,runtime \
  | tail -n +3 \
  | awk '{print $1}' \
  | xargs poetry run pip show \
  | grep -E 'Location:|Name:' \
  | cut -d ' ' -f 2 \
  | paste -d ' ' - - \
  | awk '{print $2 "/" tolower($1)}' \
  | xargs du -sh 2> /dev/null \
  | sort -hr
```

Here is the top 10 before.

```
134M	pyarrow
130M	playwright
58M	notebook
53M	pandas
33M	babel
26M	botocore
23M	litellm
23M	gevent
22M	matplotlib
21M	debugpy
```

After:

```
130M	playwright
58M	notebook
53M	pandas
33M	babel
26M	botocore
23M	litellm
21M	debugpy
20M	jupyterlab
13M	cryptography
12M	lxml
```

enyst (approved):
Great to see, thank you! The function was unused, maybe a leftover.

PR #7781, opened 2025-04-09
fix: Disable prompt caching in default condenser

csmith49:

---

The default condenser inherits all LLM configuration options from the agent, but can't take advantage of prompt caching if it happens to be enabled (which it is by default). This PR disables the prompt caching regardless of the agent's config.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:7152ac8-nikolaik   --name openhands-app-7152ac8   docker.all-hands.dev/all-hands-ai/openhands:7152ac8
```


enyst (approved):
Good point, thank you.

I think the same applies to the other LLM-based condensers, doesn't it? 🤔

csmith49:
Believe it does. I'll add the same for them.

PR #7786, opened 2025-04-10
chore - Do not install playwright in CI build steps

raymyers:

---

The PR removes the use of `make install-python-dependencies` in the CI steps in favor of a simple `poetry install`. The main difference is the call to `playwright install` that the Make target provides. These CI steps do not need playwright and installing it takes time. Actually most of them only call runtime_build so they need far fewer pip dependencies, it would be useful to create a new group for this.

---

### DIFF
--- a/.github/workflows/ghcr-build.yml
+++ b/.github/workflows/ghcr-build.yml
@@ -145,7 +145,7 @@ jobs:
       - name: Install poetry via pipx
         run: pipx install poetry
       - name: Install Python dependencies using Poetry
-        run: make install-python-dependencies
+        run: poetry install
       - name: Create source distribution and Dockerfile
         run: poetry run python3 openhands/runtime/utils/runtime_build.py --base_image ${{ matrix.base_image.image }} --build_folder containers/runtime --force_rebuild
       - name: Lowercase Repository Owner
@@ -220,7 +220,7 @@ jobs:
       - name: Install poetry via pipx
         run: pipx install poetry
       - name: Install Python dependencies using Poetry
-        run: make install-python-dependencies
+        run: poetry install
       - name: Get hash in App Image
         run: |
           echo "Hash from app image: ${{ needs.ghcr_build_app.outputs.hash_from_app_image }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/ghcr-build.yml
```diff
@@ -145,7 +145,7 @@ jobs:
       - name: Install poetry via pipx
         run: pipx install poetry
       - name: Install Python dependencies using Poetry
-        run: make install-python-dependencies
+        run: poetry install --only main
```
enyst, line 148: I am a bit concerned about this, because installing in different ways in CI and elsewhere feels like it would inevitably lead to discrepancies at times, and they can be very time-consuming.

I'd rather see that we try to use `make` everywhere, maybe with some flag or step that documents when it skips something.

Please note that I'm not really referring to the detail here, more in general... FWIW we even have somewhere install openhands with pip, which bothers me for the same reason: when there's some issue, like a dependency problem, it would happen inconsistently on some install and not others, which can be very time-consuming to figure out or debug.

enyst (commented):
I see the point in this, thank you for looking into it!

I have a quick note inline.

.github/workflows/ghcr-build.yml
raymyers: Yeah that's logical, since there are a few variations I added optional args for dependency group target for poetry group and disabling Playwright install. Poetry will also allow multiple groups comma-separated.

.github/workflows/ghcr-build.yml
```diff
@@ -355,7 +355,7 @@ jobs:
       - name: Install poetry via pipx
         run: pipx install poetry
       - name: Install Python dependencies using Poetry
-        run: make install-python-dependencies
+        run: make install-python-dependencies INSTALL_PLAYWRIGHT=0
```
rbren, line 358: doesn't the env var need to come before the command?

raymyers: For most commands yes. This is a make feature:

https://www.gnu.org/software/make/manual/html_node/Overriding.html

rbren:
Dunno if this is just the cache but that's a pretty fast build ⚡ 

Screenshot 2025-04-10 at 12 49 29 PM

raymyers:
> Dunno if this is just the cache but that's a pretty fast build ⚡
> 
@rbren Yeah the update to add blacksmith docker builds is hitting cache pretty well on builds with no runtime code changes. 

https://docs.blacksmith.sh/blacksmith-caching/docker-builds

Probably opportunity to speed up more cases by using cache mounts too, coming soon.

https://docs.docker.com/build/cache/optimize/#use-cache-mounts

PR #7796, opened 2025-04-10
chore - Add more workers to speed up runtime CI tests

raymyers:

---

The runtime tests are currently one of the biggest constraints on PR check time, this attempts to improve this by adding more parallelism. This is going to be resource intensive because each process is running docker containers.

Changes:
* Download fewer dependencies
* Run in 7 processes instead of 3
* Use an 8 vCPU worker instead of 4
* Print top test durations.

Cost: The heavier runners will be more expensive, but hopefully this will be offset somewhat by less time being spent.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:5913699-nikolaik   --name openhands-app-5913699   docker.all-hands.dev/all-hands-ai/openhands:5913699
```

### DIFF
--- a/.github/workflows/ghcr-build.yml
+++ b/.github/workflows/ghcr-build.yml
@@ -308,7 +308,7 @@ jobs:
           SANDBOX_RUNTIME_CONTAINER_IMAGE=$image_name \
           TEST_IN_CI=true \
           RUN_AS_OPENHANDS=false \
-          poetry run pytest -n 3 -raRs --reruns 2 --reruns-delay 5 --cov=openhands --cov-report=xml -s ./tests/runtime --ignore=tests/runtime/test_browsergym_envs.py
+          poetry run pytest -n 6 -raRs --reruns 2 --reruns-delay 5 --cov=openhands --cov-report=xml -s ./tests/runtime --ignore=tests/runtime/test_browsergym_envs.py
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v5
         env:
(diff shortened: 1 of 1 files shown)


raymyers:
Side note, we can probably remove the CodeCov from these runtime tests since it's inaccurate, and are we even using the data?

.github/workflows/ghcr-build.yml
```diff
@@ -308,7 +308,7 @@ jobs:
           SANDBOX_RUNTIME_CONTAINER_IMAGE=$image_name \
           TEST_IN_CI=true \
           RUN_AS_OPENHANDS=false \
-          poetry run pytest -n 3 -raRs --reruns 2 --reruns-delay 5 --cov=openhands --cov-report=xml -s ./tests/runtime --ignore=tests/runtime/test_browsergym_envs.py
+          poetry run pytest -n 7 -raRs --reruns 2 --reruns-delay 5 --cov=openhands --cov-report=xml -s ./tests/runtime --ignore=tests/runtime/test_browsergym_envs.py --durations=10
```
mamoodi, line 311: I remember this made the tests fail more often. Maybe things have changed since then.

raymyers: Interesting, I'm not able to find that in main's history over the past 7 months, maybe it was in a PR. But I wonder if at that time there was also the increase in vcpu's. I'll trying running it a few times in a row

xingyaoww (approved):
LGTM!

.github/workflows/ghcr-build.yml
xingyaoww: yeah i think with increase in vcpu - the flakyness might be improved

mamoodi: Yeah just a few runs should give you a better idea. Overall looks good. Thanks!

rbren (approved):

.github/workflows/ghcr-build.yml
raymyers: Yeah pretty consistent so far but we can report on failure rate in blacksmith and double check later.

PR #7800, opened 2025-04-10
Fixes spurious triggering of advanced mode caused by model case mismatch

### LINKED ISSUE #7789: [Bug]: Selecting a model that is not the default enables advanced mode

The underlying issue is that we format some of the received models (e.g., `"openai"` -> `"OpenAI"`). The logic responsible for determined whether the advanced switch should be toggled or not checks if these strings are equal, which is false, and therefore assumes it is a custom model

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Operating System

None

https://github.com/user-attachments/assets/d80d8906-5576-4bfe-82e5-b60c70bccb45

jommco:
Fixes #7789

When selecting a model that is not the default, advanced mode is incorrectly enabled due to a case mismatch in the model name check. For example, "openai" vs. "OpenAI" are treated as different, leading the app to assume a custom model is selected, thus triggering advanced mode.

---
PR fixes the issue by lowering the case of variable employed to determine advanced mode activation.

---
[**Link of any specific issues this addresses.**](https://github.com/All-Hands-AI/OpenHands/issues/7789)

### DIFF
--- a/frontend/src/utils/is-custom-model.ts
+++ b/frontend/src/utils/is-custom-model.ts
@@ -12,11 +12,13 @@ export const isCustomModel = (models: string[], model: string): boolean => {
 
   const organizedModels = organizeModelsAndProviders(models);
   const { provider: extractedProvider, model: extractedModel } =
-    extractModelAndProvider(model);
+    extractModelAndProvider(model.toLowerCase()); // Convert to lowercase
 
   const isKnownModel =
-    extractedProvider in organizedModels &&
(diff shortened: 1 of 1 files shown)


mamoodi:
Does it fix this issue? https://github.com/All-Hands-AI/OpenHands/issues/7789

amanape:
Already fixed, thanks for the attempt!

PR #7801, opened 2025-04-10
feat: CLI enhancements to support /init, /help and /exit

pandukamuditha:

Enhance the CLI startup, UI, and add /init, /help, /exit interactive commands

---

- Integrate prompt-toolkit TUI components for the CLI behavior
- Folder security confirmation and persistence in ~/.openhands.toml
- /init command to initialize repo.md by using a custom prompt
- /help command to display CLI usage information
- /exit command to show metrics and exit the CLI

---


openhands/core/cli.py
```diff
@@ -137,10 +238,115 @@ def display_event(event: Event, config: AppConfig) -> None:
...
+    )
+    print_formatted_text('')
+
+    print_formatted_text('Things that you can try:')
+    print_formatted_text(
+        HTML('• Ask questions about the codebase <grey>> How does main.py work?</grey>')
+    )
+    print_formatted_text(
+        HTML(
+            '• Edit files or add new features <grey>> Add a new function to ...</grey>'
```
rbren, line 265: Are the double `>>` intended here (and above/below)?

rbren: Ahh I think I see what's up--we start the prompt text with `>`. Surprised it doesn't need escaping!

openhands/core/cli.py
```diff
@@ -150,50 +356,321 @@ async def read_prompt_input(multiline=False):
...
-            'Confirm action (possible security risk)? (y/n) >> ',
+            'Confirm action (possible security risk)? (y/n) > ',
         )
         return confirmation.lower() == 'y'
     except (KeyboardInterrupt, EOFError):
         return False
 
 
-async def main(loop: asyncio.AbstractEventLoop) -> None:
+async def init_repository(current_dir: str) -> bool:
```
rbren, line 387: Doesn't have to be in this PR, but we should probably create an `openhands/cli` package, so we can break these functions up into smaller files

openhands/core/cli.py
```diff
@@ -209,26 +686,64 @@ async def main(loop: asyncio.AbstractEventLoop) -> None:
...
                 AgentState.AWAITING_USER_INPUT,
                 AgentState.FINISHED,
             ]:
+                # Reload microagents after initialization of repo.md
+                if reload_microagents:
+                    microagents: list[BaseMicroagent] = (
+                        runtime.get_microagents_from_selected_repo(None)
+                    )
+                    memory.load_user_workspace_microagents(microagents)
+                    reload_microagents = False
```
rbren, line 744: probably want to put this closer to where `/init` ends. Otherwise this will often run when unnecessary

rbren (approved):

openhands/core/cli.py
pandukamuditha: Working on this @rbren. Planning to include this in the next PR along with the next set of interactive commands.

openhands/core/cli.py
pandukamuditha: This code block runs only when the `reload_microagents` flag is True, which is set only when `/init` is run.

openhands/core/cli.py
pandukamuditha: It doesn't need escaping because `prompt_toolkit.formatted_text.HTML` treats the text inside like HTML

PR #7803, opened 2025-04-10
Fix: Display accumulated token usage in frontend metrics

xingyaoww:

Fixes an issue where the token usage metrics displayed in the frontend were showing only the latest request token count instead of the accumulated total across all requests. This ensures users can accurately track their total token usage throughout a conversation.

---

This PR fixes the token usage metrics displayed in the frontend by ensuring the accumulated token usage is properly passed from the backend to the frontend.

The root cause was in the `_prepare_metrics_for_frontend` method in `agent_controller.py`, which was creating a new `Metrics` object with only the latest token usage, not the accumulated usage that was being properly tracked in the backend.

The fix adds code to copy the accumulated token usage from the agent's metrics to the new metrics object sent to the frontend, ensuring that the frontend displays the correct accumulated token counts.

---

No specific issue ticket, but addresses a bug where `metrics.usage.completion_tokens` displayed in the frontend was not accurate.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f527741-nikolaik   --name openhands-app-f527741   docker.all-hands.dev/all-hands-ai/openhands:f527741
```

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -66,7 +66,7 @@ from openhands.events.observation import (
 )
 from openhands.events.serialization.event import event_to_trajectory, truncate_content
 from openhands.llm.llm import LLM
-from openhands.llm.metrics import Metrics, TokenUsage
+from openhands.llm.metrics import Metrics
 
 # note: RESUME is only available on web GUI
 TRAFFIC_CONTROL_REMINDER = (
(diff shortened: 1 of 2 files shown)


xingyaoww:
After this PR, now the token accumulation metric looks correct 

image

enyst (approved):
I see, thank you for dealing with this.

FWIW, I'm starting to wonder if we could settle on one place, because I think Robert was also doing something like this, but on the conversation, rather than the action:
https://github.com/All-Hands-AI/OpenHands/pull/7561

xingyaoww:
@enyst yeah great point. I think maybe after a FE refactor, we will display this metric to conversation card level and completely remove the metrics we send along with actions

PR #7809, opened 2025-04-11
chore - Rebuild docker image in fork CI instead of using artifacts

raymyers:

---

Currently: The runtime tests use the image built by the previous job in the workflow. On non-forks, that image is uploaded to ghcr, but on forks that can't be done so it is passed by an artifact. However, that artifact is very large and takes a significant fraction of total build time to upload and download.

This PR sends the dockerfile and source distribution as artifact and rebuilds it. The blacksmith cache should make the rebuild near instant, effectively performing the same copy more quickly.

---

### DIFF
--- a/.github/workflows/ghcr-build.yml
+++ b/.github/workflows/ghcr-build.yml
@@ -263,6 +263,9 @@ jobs:
         with:
           name: runtime-src-${{ matrix.base_image.tag }}
           path: containers/runtime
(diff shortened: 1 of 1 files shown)


raymyers:
Total build time 12 min  30 sec. Previously ~37 min.

xingyaoww (approved):
LGTM! the speed up is awesome!

PR #7812, opened 2025-04-11
Fix for race condition in cache

tofarr:

There was a race condition in the event cache that could cause event ordering to be flipped

---

Before The Change:
* We use a lock to get the id (e.g.: 84, 85 86)
* We then write the event to storage
* We then append the event to the cache

If 2 events come in very close together ( e.g.: `2025-04-10T20:57:57.440052` and `2025-04-10T20:57:57.480603` ), the network write can be slow enough that a later one can finish before an earlier one, and their places in the array are flipped!

By moving the append before the write we prevent this sort of race condition

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f8cdc22-nikolaik   --name openhands-app-f8cdc22   docker.all-hands.dev/all-hands-ai/openhands:f8cdc22
```

### DIFF
--- a/openhands/events/stream.py
+++ b/openhands/events/stream.py
@@ -163,25 +163,44 @@ class EventStream(EventStore):
         with self._lock:
             event._id = self.cur_id  # type: ignore [attr-defined]
             self.cur_id += 1
+
+            # Take a copy of the current write page
+            current_write_page = self._write_page_cache
+
+            # Calculate the index of the current event in the write page
+            write_page_index = len(current_write_page)
+
+            # Temporary add none - the actual event replaces this later
+            current_write_page.append(None)  # type: ignore
+
+            # If the page is full, create a new page for future events / other threads to use
+            if len(current_write_page) == self.cache_size:
+                self._write_page_cache = []
+
+        # Outside the lock update the event as required
         event._timestamp = datetime.now().isoformat()
         event._source = source  # type: ignore [attr-defined]
         data = event_to_dict(event)
         data = self._replace_secrets(data)
         event = event_from_dict(data)
+
         if event.id is not None:
-            self._write_page_cache.append(data)
+            # Set the event in the current write page - doing this before writes means that written pages will always include the event at its correct index
+            current_write_page[write_page_index] = data
+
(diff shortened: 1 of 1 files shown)


openhands/events/stream.py
```diff
@@ -163,18 +163,16 @@ def add_event(self, event: Event, source: EventSource) -> None:
...
             self.cur_id += 1
-        logger.debug(f'Adding {type(event).__name__} id={event.id} from {source.name}')
         event._timestamp = datetime.now().isoformat()
         event._source = source  # type: ignore [attr-defined]
-        logger.debug(f'Event to add: {event}')
         data = event_to_dict(event)
         data = self._replace_secrets(data)
         event = event_from_dict(data)
         if event.id is not None:
+            self._write_page_cache.append(data)
```
rbren, line 172: should we put this into the `_lock`?

openhands/events/stream.py
```diff
@@ -163,25 +163,44 @@ def add_event(self, event: Event, source: EventSource) -> None:
...
             self.cur_id += 1
+
+            # Take a copy of the current write page
+            current_write_page = self._write_page_cache
+
+            # Calculate the index of the current event in the write page
+            write_page_index = len(current_write_page)
+
+            # Temporary add none - the actual event replaces this later
+            current_write_page.append(None)  # type: ignore
```
enyst, line 174: Do we need this trick?

enyst: Aha I see the position issue. Idk, I think locking as necessary would be nicer - and easier to read/spot issues.

openhands/events/stream.py
tofarr: Done

openhands/events/stream.py
tofarr: Done!

enyst (approved):

PR #7822, opened 2025-04-11
Fix headless mode error by removing SANDBOX_USER_ID

xinbenlv:
See context of https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1744311050260359?thread_ts=1744310442.372669&cid=C06U8UTKSAD per discussion with @mamoodi

### DIFF
--- a/docs/modules/usage/how-to/headless-mode.md
+++ b/docs/modules/usage/how-to/headless-mode.md
@@ -33,7 +33,6 @@ To run OpenHands in Headless mode with Docker:
 docker run -it \
     --pull=always \
     -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.32-nikolaik \
-    -e SANDBOX_USER_ID=$(id -u) \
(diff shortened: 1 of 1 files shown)


mamoodi:
Let me test this on Linux and Windows

mamoodi:
Alright I've tested it on all platforms and this works. I'll add a troubleshooting section for the permission denied just in case. I think it's good to have.

Can you remove the same line from the CLI docs as well please? Otherwise, looks good!

mamoodi:
Actually I just asked around about this. The reason why we pass in SANDBOX_USER_ID is for file permission: we want the user inside the sandbox to have the same permission as the user outside. Otherwise, it is possible that the user mounts a directory containing a file with user permission, BUT agent is writing file owned by root into the folder. So I don't think we want to remove this. I'll create a PR to clarify this a bit. Thanks for bringing this up!

mamoodi:
I've created this PR: https://github.com/All-Hands-AI/OpenHands/pull/7845

xinbenlv:
Thank you, @mamoodi ! Closing this PR in fav of #7845

PR #7824, opened 2025-04-11
Refactor system message handling to use event stream

xingyaoww:

---

This PR refactors the system message handling to use the event stream instead of generating it on-the-fly. It adds a `SystemMessageAction` class that represents the system message and tools, and adds it to the event stream during agent initialization.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:3f29a82-nikolaik   --name openhands-app-3f29a82   docker.all-hands.dev/all-hands-ai/openhands:3f29a82
```


xingyaoww:
@OpenHands please fix all CI errors

openhands/controller/agent.py
```diff
@@ -38,6 +40,34 @@ def __init__(
...
+        system_message = self.prompt_manager.get_system_message()
+
+        # Get tools if available
+        tools = getattr(self, 'tools', None)
+
+        system_message_action = SystemMessageAction(content=system_message, tools=tools)
+        # Set the source attribute
+        system_message_action._source = EventSource.AGENT  # type: ignore
+
+        return system_message_action
```
xingyaoww, line 70: We should make sure all the agent in `openhands/agenthub` uses this SystemMessageAction to get system prompt when applicable.

openhands/controller/agent_controller.py
```diff
@@ -508,22 +516,50 @@ def _reset(self) -> None:
...
-                ):
+                if not isinstance(event, Observation):
+                    continue
+
+                # Get the metadata from the observation
+                event_metadata = getattr(event, 'tool_call_metadata', None)
+                if event_metadata is None and hasattr(event, '_tool_call_metadata'):
+                    event_metadata = event._tool_call_metadata
+
+                if event_metadata == pending_metadata:
```
xingyaoww, line 554: Why we are making so many changes here? We should simplify as much as possible

openhands/events/serialization/action.py
```diff
@@ -96,12 +97,23 @@ def action_from_dict(action: dict) -> Action:
...
+
     args = action.get('args', {})
+
+    # Special handling for message actions to distinguish between MessageAction and SystemMessageAction
+    if action_type == 'message':
+        # If 'tools' is in args, use SystemMessageAction, otherwise use MessageAction
+        if 'tools' in args:
+            action_class = SystemMessageAction
+        else:
+            action_class = MessageAction
```
xingyaoww, line 115: We should not use this -- we should add a new ActionType.SYSTEM

openhands/memory/conversation_memory.py
```diff
@@ -133,7 +134,15 @@ def process_events(
         return messages
 
     def process_initial_messages(self, with_caching: bool = False) -> list[Message]:
```
xingyaoww, line 136: Can you try to get rid of "process_initial_messages" function directly?

openhands/controller/agent.py
xingyaoww: From OpenHands:

> Verified that all agents in openhands/agenthub are compatible with the changes:

> Found that CodeActAgent._get_messages already handles SystemMessageAction correctly
> No other agents override get_initial_message or use process_initial_messages

xingyaoww:
Added legacy handling for SystemMessageAction in CodeActAgent. When no SystemMessageAction is found in the events, it will automatically add one and log a warning message. This ensures backward compatibility with existing code that might not be using the new SystemMessageAction yet.

xingyaoww:
Tested, and it seems to work :)

image

<details>
<summary>First system message event</summary>
{
    "id": 1,
    "timestamp": "2025-04-15T16:06:12.450656",
    "source": "agent",
    "message": "You are OpenHands agent, a helpful AI assistant that can interact with a computer to solve tasks.\n\n<ROLE>\nYour primary role is to assist users by executing commands, modifying code, and solving technical problems effectively. You should be thorough, methodical, and prioritize quality over speed.\n* If the user asks a question, like \"why is X happening\", don't try to fix the problem. Just give an answer to the question.\n</ROLE>\n\n<EFFICIENCY>\n* Each action you take is somewhat expensive. Wherever possible, combine multiple actions into a single action, e.g. combine multiple bash commands into one, using sed and grep to edit/view multiple files at once.\n* When exploring the codebase, use efficient tools like find, grep, and git commands with appropriate filters to minimize unnecessary operations.\n</EFFICIENCY>\n\n<FILE_SYSTEM_GUIDELINES>\n* When a user provides a file path, do NOT assume it's relative to the current working directory. First explore the file system to locate the file before working on it.\n* If asked to edit a file, edit the file directly, rather than creating a new file with a different filename.\n* For global search-and-replace operations, consider using `sed` instead of opening file editors multiple times.\n</FILE_SYSTEM_GUIDELINES>\n\n<CODE_QUALITY>\n* Write clean, efficient code with minimal comments. Avoid redundancy in comments: Do not repeat information that can be easily inferred from the code itself.\n* When implementing solutions, focus on making the minimal changes needed to solve the problem.\n* Before implementing any changes, first thoroughly understand the codebase through exploration.\n* If you are adding a lot of code to a function or file, consider splitting the function or file into smaller pieces when appropriate.\n</CODE_QUALITY>\n\n<VERSION_CONTROL>\n* When configuring git credentials, use \"openhands\" as the user.name and \"openhands@all-hands.dev\" as the user.email by default, unless explicitly instructed otherwise.\n* Exercise caution with git operations. Do NOT make potentially dangerous changes (e.g., pushing to main, deleting repositories) unless explicitly asked to do so.\n* When committing changes, use `git status` to see all modified files, and stage all files necessary for the commit. Use `git commit -a` whenever possible.\n* Do NOT commit files that typically shouldn't go into version control (e.g., node_modules/, .env files, build directories, cache files, large binaries) unless explicitly instructed by the user.\n* If unsure about committing certain files, check for the presence of .gitignore files or ask the user for clarification.\n</VERSION_CONTROL>\n\n<PULL_REQUESTS>\n* When creating pull requests, create only ONE per session/issue unless explicitly instructed otherwise.\n* When working with an existing PR, update it with new commits rather than creating additional PRs for the same issue.\n* When updating a PR, preserve the original PR title and purpose, updating description only when necessary.\n</PULL_REQUESTS>\n\n<PROBLEM_SOLVING_WORKFLOW>\n1. EXPLORATION: Thoroughly explore relevant files and understand the context before proposing solutions\n2. ANALYSIS: Consider multiple approaches and select the most promising one\n3. TESTING:\n   * For bug fixes: Create tests to verify issues before implementing fixes\n   * For new features: Consider test-driven development when appropriate\n   * If the repository lacks testing infrastructure and implementing tests would require extensive setup, consult with the user before investing time in building testing infrastructure\n4. IMPLEMENTATION: Make focused, minimal changes to address the problem\n5. VERIFICATION: Test your implementation thoroughly, including edge cases\n</PROBLEM_SOLVING_WORKFLOW>\n\n<SECURITY>\n* Only use GITHUB_TOKEN and other credentials in ways the user has explicitly requested and would expect.\n* Use APIs to work with GitHub or other platforms, unless the user asks otherwise or your task requires browsing.\n</SECURITY>\n\n<ENVIRONMENT_SETUP>\n* When user asks you to run an application, don't stop if the application is not installed. Instead, please install the application and run the command again.\n* If you encounter missing dependencies:\n  1. First, look around in the repository for existing dependency files (requirements.txt, pyproject.toml, package.json, Gemfile, etc.)\n  2. If dependency files exist, use them to install all dependencies at once (e.g., `pip install -r requirements.txt`, `npm install`, etc.)\n  3. Only install individual packages directly if no dependency files are found or if only specific packages are needed\n* Similarly, if you encounter missing dependencies for essential tools requested by the user, install them when possible.\n</ENVIRONMENT_SETUP>\n\n<TROUBLESHOOTING>\n* If you've made repeated attempts to solve a problem but tests still fail or the user reports it's still broken:\n  1. Step back and reflect on 5-7 different possible sources of the problem\n  2. Assess the likelihood of each possible cause\n  3. Methodically address the most likely causes, starting with the highest probability\n  4. Document your reasoning process\n* When you run into any major issue while executing a plan from the user, please don't try to directly work around it. Instead, propose a new plan and confirm with the user before proceeding.\n</TROUBLESHOOTING>",
    "action": "system",
    "args": {
        "content": "You are OpenHands agent, a helpful AI assistant that can interact with a computer to solve tasks.\n\n<ROLE>\nYour primary role is to assist users by executing commands, modifying code, and solving technical problems effectively. You should be thorough, methodical, and prioritize quality over speed.\n* If the user asks a question, like \"why is X happening\", don't try to fix the problem. Just give an answer to the question.\n</ROLE>\n\n<EFFICIENCY>\n* Each action you take is somewhat expensive. Wherever possible, combine multiple actions into a single action, e.g. combine multiple bash commands into one, using sed and grep to edit/view multiple files at once.\n* When exploring the codebase, use efficient tools like find, grep, and git commands with appropriate filters to minimize unnecessary operations.\n</EFFICIENCY>\n\n<FILE_SYSTEM_GUIDELINES>\n* When a user provides a file path, do NOT assume it's relative to the current working directory. First explore the file system to locate the file before working on it.\n* If asked to edit a file, edit the file directly, rather than creating a new file with a different filename.\n* For global search-and-replace operations, consider using `sed` instead of opening file editors multiple times.\n</FILE_SYSTEM_GUIDELINES>\n\n<CODE_QUALITY>\n* Write clean, efficient code with minimal comments. Avoid redundancy in comments: Do not repeat information that can be easily inferred from the code itself.\n* When implementing solutions, focus on making the minimal changes needed to solve the problem.\n* Before implementing any changes, first thoroughly understand the codebase through exploration.\n* If you are adding a lot of code to a function or file, consider splitting the function or file into smaller pieces when appropriate.\n</CODE_QUALITY>\n\n<VERSION_CONTROL>\n* When configuring git credentials, use \"openhands\" as the user.name and \"openhands@all-hands.dev\" as the user.email by default, unless explicitly instructed otherwise.\n* Exercise caution with git operations. Do NOT make potentially dangerous changes (e.g., pushing to main, deleting repositories) unless explicitly asked to do so.\n* When committing changes, use `git status` to see all modified files, and stage all files necessary for the commit. Use `git commit -a` whenever possible.\n* Do NOT commit files that typically shouldn't go into version control (e.g., node_modules/, .env files, build directories, cache files, large binaries) unless explicitly instructed by the user.\n* If unsure about committing certain files, check for the presence of .gitignore files or ask the user for clarification.\n</VERSION_CONTROL>\n\n<PULL_REQUESTS>\n* When creating pull requests, create only ONE per session/issue unless explicitly instructed otherwise.\n* When working with an existing PR, update it with new commits rather than creating additional PRs for the same issue.\n* When updating a PR, preserve the original PR title and purpose, updating description only when necessary.\n</PULL_REQUESTS>\n\n<PROBLEM_SOLVING_WORKFLOW>\n1. EXPLORATION: Thoroughly explore relevant files and understand the context before proposing solutions\n2. ANALYSIS: Consider multiple approaches and select the most promising one\n3. TESTING:\n   * For bug fixes: Create tests to verify issues before implementing fixes\n   * For new features: Consider test-driven development when appropriate\n   * If the repository lacks testing infrastructure and implementing tests would require extensive setup, consult with the user before investing time in building testing infrastructure\n4. IMPLEMENTATION: Make focused, minimal changes to address the problem\n5. VERIFICATION: Test your implementation thoroughly, including edge cases\n</PROBLEM_SOLVING_WORKFLOW>\n\n<SECURITY>\n* Only use GITHUB_TOKEN and other credentials in ways the user has explicitly requested and would expect.\n* Use APIs to work with GitHub or other platforms, unless the user asks otherwise or your task requires browsing.\n</SECURITY>\n\n<ENVIRONMENT_SETUP>\n* When user asks you to run an application, don't stop if the application is not installed. Instead, please install the application and run the command again.\n* If you encounter missing dependencies:\n  1. First, look around in the repository for existing dependency files (requirements.txt, pyproject.toml, package.json, Gemfile, etc.)\n  2. If dependency files exist, use them to install all dependencies at once (e.g., `pip install -r requirements.txt`, `npm install`, etc.)\n  3. Only install individual packages directly if no dependency files are found or if only specific packages are needed\n* Similarly, if you encounter missing dependencies for essential tools requested by the user, install them when possible.\n</ENVIRONMENT_SETUP>\n\n<TROUBLESHOOTING>\n* If you've made repeated attempts to solve a problem but tests still fail or the user reports it's still broken:\n  1. Step back and reflect on 5-7 different possible sources of the problem\n  2. Assess the likelihood of each possible cause\n  3. Methodically address the most likely causes, starting with the highest probability\n  4. Document your reasoning process\n* When you run into any major issue while executing a plan from the user, please don't try to directly work around it. Instead, propose a new plan and confirm with the user before proceeding.\n</TROUBLESHOOTING>",
        "tools": [
            {
                "type": "function",
                "function": {
                    "name": "execute_bash",
                    "description": "Execute a bash command in the terminal within a persistent shell session.\n\n### Command Execution\n* One command at a time: You can only execute one bash command at a time. If you need to run multiple commands sequentially, use `&&` or `;` to chain them together.\n* Persistent session: Commands execute in a persistent shell session where environment variables, virtual environments, and working directory persist between commands.\n* Timeout: Commands have a soft timeout of 120 seconds, once that's reached, you have the option to continue or interrupt the command (see section below for details)\n\n### Running and Interacting with Processes\n* Long running commands: For commands that may run indefinitely, run them in the background and redirect output to a file, e.g. `python3 app.py > server.log 2>&1 &`.\n* Interact with running process: If a bash command returns exit code `-1`, this means the process is not yet finished. By setting `is_input` to `true`, you can:\n  - Send empty `command` to retrieve additional logs\n  - Send text (set `command` to the text) to STDIN of the running process\n  - Send control commands like `C-c` (Ctrl+C), `C-d` (Ctrl+D), or `C-z` (Ctrl+Z) to interrupt the process\n\n### Best Practices\n* Directory verification: Before creating new directories or files, first verify the parent directory exists and is the correct location.\n* Directory management: Try to maintain working directory by using absolute paths and avoiding excessive use of `cd`.\n\n### Output Handling\n* Output truncation: If the output exceeds a maximum length, it will be truncated before being returned.\n",
                    "parameters": {
                        "type": "object",
                        "properties": {
                            "command": {
                                "type": "string",
                                "description": "The bash command to execute. Can be empty string to view additional logs when previous exit code is `-1`. Can be `C-c` (Ctrl+C) to interrupt the currently running process. Note: You can only execute one bash command at a time. If you need to run multiple commands sequentially, you can use `&&` or `;` to chain them together."
                            },
                            "is_input": {
                                "type": "string",
                                "description": "If True, the command is an input to the running process. If False, the command is a bash command to be executed in the terminal. Default is False.",
                                "enum": [
                                    "true",
                                    "false"
                                ]
                            }
                        },
                        "required": [
                            "command"
                        ]
                    }
                }
            },
            {
                "type": "function",
                "function": {
                    "name": "think",
                    "description": "Use the tool to think about something. It will not obtain new information or make any changes to the repository, but just log the thought. Use it when complex reasoning or brainstorming is needed.\n\nCommon use cases:\n1. When exploring a repository and discovering the source of a bug, call this tool to brainstorm several unique ways of fixing the bug, and assess which change(s) are likely to be simplest and most effective.\n2. After receiving test results, use this tool to brainstorm ways to fix failing tests.\n3. When planning a complex refactoring, use this tool to outline different approaches and their tradeoffs.\n4. When designing a new feature, use this tool to think through architecture decisions and implementation details.\n5. When debugging a complex issue, use this tool to organize your thoughts and hypotheses.\n\nThe tool simply logs your thought process for better transparency and does not execute any code or make changes.",
                    "parameters": {
                        "type": "object",
                        "properties": {
                            "thought": {
                                "type": "string",
                                "description": "The thought to log."
                            }
                        },
                        "required": [
                            "thought"
                        ]
                    }
                }
            },
            {
                "type": "function",
                "function": {
                    "name": "finish",
                    "description": "Signals the completion of the current task or conversation.\n\nUse this tool when:\n- You have successfully completed the user's requested task\n- You cannot proceed further due to technical limitations or missing information\n\nThe message should include:\n- A clear summary of actions taken and their results\n- Any next steps for the user\n- Explanation if you're unable to complete the task\n- Any follow-up questions if more information is needed\n\nThe task_completed field should be set to True if you believed you have completed the task, and False otherwise.\n",
                    "parameters": {
                        "type": "object",
                        "required": [
                            "message",
                            "task_completed"
                        ],
                        "properties": {
                            "message": {
                                "type": "string",
                                "description": "Final message to send to the user"
                            },
                            "task_completed": {
                                "type": "string",
                                "enum": [
                                    "true",
                                    "false",
                                    "partial"
                                ],
                                "description": "Whether you have completed the task."
                            }
                        }
                    }
                }
            },
            {
                "type": "function",
                "function": {
                    "name": "web_read",
                    "description": "Read (convert to markdown) content from a webpage. You should prefer using the `web_read` tool over the `browser` tool, but do use the `browser` tool if you need to interact with a webpage (e.g., click a button, fill out a form, etc.) OR read a webpage that contains images.\n\nYou may use the `web_read` tool to read text content from a webpage, and even search the webpage content using a Google search query (e.g., url=`https://www.google.com/search?q=YOUR_QUERY`).\n",
                    "parameters": {
                        "type": "object",
                        "properties": {
                            "url": {
                                "type": "string",
                                "description": "The URL of the webpage to read. You can also use a Google search query here (e.g., `https://www.google.com/search?q=YOUR_QUERY`)."
                            }
                        },
                        "required": [
                            "url"
                        ]
                    }
                }
            },
            {
                "type": "function",
                "function": {
                    "name": "browser",
                    "description": "Interact with the browser using Python code. Use it ONLY when you need to interact with a webpage.\n\nSee the description of \"code\" parameter for more details.\n\nMultiple actions can be provided at once, but will be executed sequentially without any feedback from the page.\nMore than 2-3 actions usually leads to failure or unexpected behavior. Example:\nfill('a12', 'example with \"quotes\"')\nclick('a51')\nclick('48', button='middle', modifiers=['Shift'])\n\nYou can also use the browser to view pdf, png, jpg files.\nYou should first check the content of /tmp/oh-server-url to get the server url, and then use it to view the file by `goto(\"{server_url}/view?path={absolute_file_path}\")`.\nFor example: `goto(\"http://localhost:8000/view?path=/workspace/test_document.pdf\")`\nNote: The file should be downloaded to the local machine first before using the browser to view it.\n",
                    "parameters": {
                        "type": "object",
                        "properties": {
                            "code": {
                                "type": "string",
                                "description": "The Python code that interacts with the browser.\n\nThe following 15 functions are available. Nothing else is supported.\n\ngoto(url: str)\n    Description: Navigate to a url.\n    Examples:\n        goto('http://www.example.com')\n\ngo_back()\n    Description: Navigate to the previous page in history.\n    Examples:\n        go_back()\n\ngo_forward()\n    Description: Navigate to the next page in history.\n    Examples:\n        go_forward()\n\nnoop(wait_ms: float = 1000)\n    Description: Do nothing, and optionally wait for the given time (in milliseconds).\n    You can use this to get the current page content and/or wait for the page to load.\n    Examples:\n        noop()\n\n        noop(500)\n\nscroll(delta_x: float, delta_y: float)\n    Description: Scroll horizontally and vertically. Amounts in pixels, positive for right or down scrolling, negative for left or up scrolling. Dispatches a wheel event.\n    Examples:\n        scroll(0, 200)\n\n        scroll(-50.2, -100.5)\n\nfill(bid: str, value: str)\n    Description: Fill out a form field. It focuses the element and triggers an input event with the entered text. It works for <input>, <textarea> and [contenteditable] elements.\n    Examples:\n        fill('237', 'example value')\n\n        fill('45', 'multi-line\nexample')\n\n        fill('a12', 'example with \"quotes\"')\n\nselect_option(bid: str, options: str | list[str])\n    Description: Select one or multiple options in a <select> element. You can specify option value or label to select. Multiple options can be selected.\n    Examples:\n        select_option('a48', 'blue')\n\n        select_option('c48', ['red', 'green', 'blue'])\n\nclick(bid: str, button: Literal['left', 'middle', 'right'] = 'left', modifiers: list[typing.Literal['Alt', 'Control', 'ControlOrMeta', 'Meta', 'Shift']] = [])\n    Description: Click an element.\n    Examples:\n        click('a51')\n\n        click('b22', button='right')\n\n        click('48', button='middle', modifiers=['Shift'])\n\ndblclick(bid: str, button: Literal['left', 'middle', 'right'] = 'left', modifiers: list[typing.Literal['Alt', 'Control', 'ControlOrMeta', 'Meta', 'Shift']] = [])\n    Description: Double click an element.\n    Examples:\n        dblclick('12')\n\n        dblclick('ca42', button='right')\n\n        dblclick('178', button='middle', modifiers=['Shift'])\n\nhover(bid: str)\n    Description: Hover over an element.\n    Examples:\n        hover('b8')\n\npress(bid: str, key_comb: str)\n    Description: Focus the matching element and press a combination of keys. It accepts the logical key names that are emitted in the keyboardEvent.key property of the keyboard events: Backquote, Minus, Equal, Backslash, Backspace, Tab, Delete, Escape, ArrowDown, End, Enter, Home, Insert, PageDown, PageUp, ArrowRight, ArrowUp, F1 - F12, Digit0 - Digit9, KeyA - KeyZ, etc. You can alternatively specify a single character you'd like to produce such as \"a\" or \"#\". Following modification shortcuts are also supported: Shift, Control, Alt, Meta, ShiftLeft, ControlOrMeta. ControlOrMeta resolves to Control on Windows and Linux and to Meta on macOS.\n    Examples:\n        press('88', 'Backspace')\n\n        press('a26', 'ControlOrMeta+a')\n\n        press('a61', 'Meta+Shift+t')\n\nfocus(bid: str)\n    Description: Focus the matching element.\n    Examples:\n        focus('b455')\n\nclear(bid: str)\n    Description: Clear the input field.\n    Examples:\n        clear('996')\n\ndrag_and_drop(from_bid: str, to_bid: str)\n    Description: Perform a drag & drop. Hover the element that will be dragged. Press left mouse button. Move mouse to the element that will receive the drop. Release left mouse button.\n    Examples:\n        drag_and_drop('56', '498')\n\nupload_file(bid: str, file: str | list[str])\n    Description: Click an element and wait for a \"filechooser\" event, then select one or multiple input files for upload. Relative file paths are resolved relative to the current working directory. An empty list clears the selected files.\n    Examples:\n        upload_file('572', '/home/user/my_receipt.pdf')\n\n        upload_file('63', ['/home/bob/Documents/image.jpg', '/home/bob/Documents/file.zip'])\n"
                            }
                        },
                        "required": [
                            "code"
                        ]
                    }
                }
            },
            {
                "type": "function",
                "function": {
                    "name": "execute_ipython_cell",
                    "description": "Run a cell of Python code in an IPython environment.\n* The assistant should define variables and import packages before using them.\n* The variable defined in the IPython environment will not be available outside the IPython environment (e.g., in terminal).\n",
                    "parameters": {
                        "type": "object",
                        "properties": {
                            "code": {
                                "type": "string",
                                "description": "The Python code to execute. Supports magic commands like %pip."
                            }
                        },
                        "required": [
                            "code"
                        ]
                    }
                }
            },
            {
                "type": "function",
                "function": {
                    "name": "str_replace_editor",
                    "description": "Custom editing tool for viewing, creating and editing files in plain-text format\n* State is persistent across command calls and discussions with the user\n* If `path` is a file, `view` displays the result of applying `cat -n`. If `path` is a directory, `view` lists non-hidden files and directories up to 2 levels deep\n* The `create` command cannot be used if the specified `path` already exists as a file\n* If a `command` generates a long output, it will be truncated and marked with `<response clipped>`\n* The `undo_edit` command will revert the last edit made to the file at `path`\n\n\nBefore using this tool:\n1. Use the view tool to understand the file's contents and context\n2. Verify the directory path is correct (only applicable when creating new files):\n   - Use the view tool to verify the parent directory exists and is the correct location\n\nWhen making edits:\n   - Ensure the edit results in idiomatic, correct code\n   - Do not leave the code in a broken state\n   - Always use absolute file paths (starting with /)\n\nCRITICAL REQUIREMENTS FOR USING THIS TOOL:\n\n1. EXACT MATCHING: The `old_str` parameter must match EXACTLY one or more consecutive lines from the file, including all whitespace and indentation. The tool will fail if `old_str` matches multiple locations or doesn't match exactly with the file content.\n\n2. UNIQUENESS: The `old_str` must uniquely identify a single instance in the file:\n   - Include sufficient context before and after the change point (3-5 lines recommended)\n   - If not unique, the replacement will not be performed\n\n3. REPLACEMENT: The `new_str` parameter should contain the edited lines that replace the `old_str`. Both strings must be different.\n\nRemember: when making multiple file edits in a row to the same file, you should prefer to send all edits in a single message with multiple calls to this tool, rather than multiple messages with a single call each.\n",
                    "parameters": {
                        "type": "object",
                        "properties": {
                            "command": {
                                "description": "The commands to run. Allowed options are: `view`, `create`, `str_replace`, `insert`, `undo_edit`.",
                                "enum": [
                                    "view",
                                    "create",
                                    "str_replace",
                                    "insert",
                                    "undo_edit"
                                ],
                                "type": "string"
                            },
                            "path": {
                                "description": "Absolute path to file or directory, e.g. `/workspace/file.py` or `/workspace`.",
                                "type": "string"
                            },
                            "file_text": {
                                "description": "Required parameter of `create` command, with the content of the file to be created.",
                                "type": "string"
                            },
                            "old_str": {
                                "description": "Required parameter of `str_replace` command containing the string in `path` to replace.",
                                "type": "string"
                            },
                            "new_str": {
                                "description": "Optional parameter of `str_replace` command containing the new string (if not given, no string will be added). Required parameter of `insert` command containing the string to insert.",
                                "type": "string"
                            },
                            "insert_line": {
                                "description": "Required parameter of `insert` command. The `new_str` will be inserted AFTER the line `insert_line` of `path`.",
                                "type": "integer"
                            },
                            "view_range": {
                                "description": "Optional parameter of `view` command when `path` points to a file. If none is given, the full file is shown. If provided, the file will be shown in the indicated line number range, e.g. [11, 12] will show lines 11 and 12. Indexing at 1 to start. Setting `[start_line, -1]` shows all lines from `start_line` to the end of the file.",
                                "items": {
                                    "type": "integer"
                                },
                                "type": "array"
                            }
                        },
                        "required": [
                            "command",
                            "path"
                        ]
                    }
                }
            }
        ]
    }
}
</details>

openhands/memory/conversation_memory.py
```diff
@@ -275,6 +260,16 @@ def _process_action(
                     content=content,
                 )
             ]
+        elif isinstance(action, SystemMessageAction):
+            # Convert SystemMessageAction to a system message
+            return [
+                Message(
+                    role='system',
+                    content=[TextContent(text=action.content, cache_prompt=True)],
```
enyst, line 268: Ah, looks like this will always add the cache marker, regardless if enabled/disabled or wrong LLM.

What if we store `prompt_caching_enabled` as an attribute on the `ConversationMemory`? We know it in the agent before we call anything here 🤔

openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -193,15 +193,32 @@ def _get_messages(self, events: list[Event]) -> list[Message]:
...
-        messages = self.conversation_memory.process_initial_messages(
-            with_caching=self.llm.is_caching_prompt_active()
+        # Check if there's a SystemMessageAction in the events
+        has_system_message = any(
+            isinstance(event, SystemMessageAction) for event in events
         )
 
-        # Use ConversationMemory to process events
+        # Legacy behavior: If no SystemMessageAction is found, add one
+        if not has_system_message:
```
enyst, line 202: Totally nitpicking: maybe this code doesn't have to be here, it could be separated into its own compatibility method in conversation_memory, so that the agent code remains simpler and more readable.

It doesn't have to be this PR, I can do that, really quick after we merge this, because we have a large PR [building on the agent](https://github.com/All-Hands-AI/OpenHands/pull/7843), and I don't want more pain for them. 😅

xingyaoww: This is a great idea! Let's maybe do it after this PR? -- I didn't do that here because i'm slightly worried that this might impact other agents that is using this function though

openhands/memory/conversation_memory.py
xingyaoww: You are absolutely right! Just remove this hard-code in a new commit.

In fact, we don't need to store this attribute at all -- we already have a apply_prompt_caching function :)
https://github.com/All-Hands-AI/OpenHands/blob/7e14a512e07a322200e8cb2a3b69bdf8dbc3bf9a/openhands/agenthub/codeact_agent/codeact_agent.py#L211-L212

openhands/server/session/session.py
```diff
@@ -127,7 +127,7 @@ async def initialize_agent(
 
         if settings.enable_default_condenser:
             default_condenser_config = LLMSummarizingCondenserConfig(
-                llm_config=llm.config, keep_first=3, max_size=80
+                llm_config=llm.config, keep_first=4, max_size=80
```
xingyaoww, line 130: @csmith49 Since we are adding a new SystemMessageAction, does it make sense to use keep_first=4? i.e., we keep "SystemMessageAction, first user MessageAction, RecallObservation, and AgentCondensationAction"?

openhands/memory/conversation_memory.py
xingyaoww: ok actually i still need to do this:

https://github.com/All-Hands-AI/OpenHands/pull/7824/commits/2e52cea396f03bf9d7297daed71fffcbc74507ca

openhands/server/session/session.py
enyst: I think it's `SystemMessageAction, first user MessageAction, RecallAction, RecallObservation`, because the RecallAction exists in history too?

openhands/memory/conversation_memory.py
enyst: Nice!

openhands/server/session/session.py
csmith49: Yeah, I think that makes sense!

openhands/events/action/message.py
```diff
@@ -32,3 +33,26 @@ def __str__(self) -> str:
...
+@dataclass
+class SystemMessageAction(Action):
+    """
+    Action that represents a system message for an agent, including the system prompt
+    and available tools. This should be the first message in the event stream.
+    """
+
+    content: str
+    tools: list[Any] | None = None
+    action: str = ActionType.SYSTEM
```
csmith49, line 47:
```suggestion
    action: ActionType = ActionType.SYSTEM
```

tests/unit/test_agent_controller.py
```diff
@@ -268,22 +277,25 @@ def on_event_memory(event: Event):
     assert (
         error_observation.reason == 'AgentStuckInLoopError: Agent got stuck in a loop'
     )
-    assert len(events) == 11
+    # With the refactored system message handling, we now have 12 events instead of 11
```
csmith49, line 280: There are a few comments like this in the tests that reference old versions of code that I don't think are adding anything and might should be removed. I'm guessing Sonnet is responsible?

tests/unit/test_codeact_agent.py
```diff
@@ -288,8 +289,11 @@ def test_correct_tool_description_loaded_based_on_model_name(mock_state: State):
     assert any(len(tool['function']['description']) > 1024 for tool in agent.tools)
 
 
-def test_mismatched_tool_call_events(mock_state: State):
-    """Tests that the agent can convert mismatched tool call events (i.e., an observation with no corresponding action) into messages."""
+def test_mismatched_tool_call_events_and_auto_add_system_message(mock_state: State):
```
csmith49, line 292: Small nit, but the `_and_auto_add_system_message` should probably be another test case instead of bundled up here.

csmith49 (approved):

openhands/controller/agent.py
```diff
@@ -38,6 +41,42 @@ def __init__(
...
+                    f'[{self.name}] Prompt manager not initialized before getting system message'
+                )
+                return None
+
+            system_message = self.prompt_manager.get_system_message()
+
+            # Get tools if available
+            tools = getattr(self, 'tools', None)
+
+            system_message_action = SystemMessageAction(
```
enyst, line 70: Actually, do we create an event every time this is called? That means every controller init, but controller init is also called when the session is reloaded, not only newly created.

I think the place where we could handle it is in `_init_history`. It should be able to either call `get_system_message()` or load it and find it in the event stream. 🤔

tests/unit/test_agent_controller.py
xingyaoww: yeah sonnet is responsible 😓 I'll clean them up

openhands/server/session/session.py
xingyaoww: > when streams created previously are loaded, it looks like we create a system message event every time, which is probably not what we want.

Yeah i imagine this would be primarily a fallback mechanism to make sure this change is backward compatible. It is not idea but on the top of my head i can't think of easier way to handle this :(

openhands/controller/agent.py
xingyaoww: Great point... We should probably only add this event to event stream when the eventstream when they didn't present.

On the other hand, if we add to `_init_history`, the system message won't really show-up for frontend, which is one of the reason i want to do this refactor :)

enyst: I don't follow, sorry. `init_history` serves to load it from the stream. The stream is sent to the FE, so it should be the same thing for FE?

xingyaoww: in agent controller, i believe init_history load stuff *from* the stream and pass to the agent? If we add SystemMessage only when init_history is called, this event won't be pushed to the stream?

image

openhands/server/session/session.py
enyst: Maybe we could...
- load it normally, with everything else, in `init_history` if it exists in the stream already
- create it if it doesn't - this will mean that old streams get a SystemMessageAction in a different position than first event, it will be `id=last_id + 1`
- in `conversation_memory.py`, we would have to make sure we find the SystemMessageAction no matter where it is in history, and place its `message` first in the list to send to the LLM?

enyst: Ah, we'd need to skip it in the condensers too.

The other alternative...
- keep compatibility by not creating it, if the history was loaded. So it will be always missing for old streams, and the current compat code will insert a `message`. Condensers, as Calvin said, would have different behavior on old and new histories. 😔

openhands/controller/agent.py
enyst: What I mean is,`_init_history` is _always_ called. We always try to iterate here through the stream to find out if there is anything in. If there is nothing, we create it. If there is something, we need to decide:
- either we create it, and then it's in the wrong position so [this behavior](https://github.com/All-Hands-AI/OpenHands/pull/7824#discussion_r2047098112) I think
- or we don't create it, and then old histories won't have it, so this is the alternative in the next comment

xingyaoww: Ha now i kinda understand!

> If there is something, but no SystemMessage, then we need to decide:
> either we create it, and then it's in the wrong position so https://github.com/All-Hands-AI/OpenHands/pull/7824#discussion_r2047098112 I think
> or we don't create it, and then old histories won't have it, so this is the alternative in the https://github.com/All-Hands-AI/OpenHands/pull/7824#discussion_r2047135868 there, sorry, it's the same discussion in different parts 😅

Yeah i'm leaning towards the route of "not creating it" -- maybe we can just modify the condition and skip adding the SystemMessage when the stream already exist (e.g., a user message is already presented), like this: https://github.com/All-Hands-AI/OpenHands/pull/7824/commits/cd4515fdb70303530f9bc1e3609e339714e87b93

wdyt?

enyst: Thinking about it, I do agree! This way we keep working normally with the two stream formats actually, unmodified. Adding it in the wrong position would be more of a mess... Also, it seems to me, the other way would require external scripts to do more work. I like this better.

Just, IMHO there is no hurry to remove compatibility in June 😂 - unless maybe June next year?🙊Joking, but I do think compatibility with old streams is something we usually really try to keep. (all that compatibility code in event serializers...)

Compatibility may affect a lot of things, eval streams too, replays maybe... Maybe I'm just more conservative!

By the way, I'm curious if the replay functionality works, we have several tests...

xingyaoww: yeah i think all the replay tests worked :)

I just fixed two issue with delegate here: https://github.com/All-Hands-AI/OpenHands/pull/7824/commits/a9c9a2f57987b38c944cde7ce27012b3bea47f4b

this seems good to go 🤔

tests/unit/test_codeact_agent.py
xingyaoww: I tried to break both in the same test 😅 but i realize the content of the test would mostly be exactly the same.. hence the renaming

openhands/controller/agent_controller.py
```diff
@@ -168,7 +168,7 @@ def __init__(
         self._add_system_message()
 
     def _add_system_message(self):
-        for event in self.event_stream.get_events():
+        for event in self.event_stream.get_events(start_id=self.state.start_id):
```
enyst, line 171: I thought about this, but I concluded it would be the same thing. If the delegate finds it in the parent's events, it exists for the delegate too; if it doesn't find it in the parent's events, it doesn't have it either.

Idk, maybe I need more ☕️ 🤷‍♂️

openhands/controller/agent.py
enyst: Nice! Yes, the release just happened, too, we're good to try it live.

openhands/controller/agent_controller.py
enyst: Either way, it's ✅

openhands/events/action/message.py
```diff
@@ -32,3 +34,27 @@ def __str__(self) -> str:
...
+@dataclass
+class SystemMessageAction(Action):
+    """
+    Action that represents a system message for an agent, including the system prompt
+    and available tools. This should be the first message in the event stream.
+    """
+
+    content: str
+    tools: list[Any] | None = None
+    openhands_version: str | None = openhands.__version__
```
xingyaoww, line 48: @enyst I added this version, so we could possibly detect the version of a conversation by simply looking at the event stream! Could help with our future backward compatibility condition check. LMK if you have concern here! If not, i'll probably aim to get this merged in the next 12 hours :)

```
parameter of `view` command when `path` points to a file. If none is given, the full file is shown. If provided, the file will be shown in the indicated line number range, e.g. [11, 12] will show lines 11 and 12. Indexing at 1 to start. Setting `[start_line, -1]` shows all lines from `start_line` to the end of the file.", "items": {"type": "integer"}, "type": "array"}}, "required": ["command", "path"]}}}], "openhands_version": "0.32.0"}}
```

openhands/controller/agent_controller.py
enyst: The event stream is shared between parent and child, so it should find the parent's SystemMessage/MessageAction when we iterate with `get_events()` without `start_id`. The parent started at `id=0` in _this_ stream

openhands/events/action/message.py
enyst: Ah, I was thinking... Yes, version ... makes sense. What do you think about adding agent name? 🤔

openhands/controller/agent_controller.py
xingyaoww: yeah, but adding this start_id should solve the issue?

openhands/events/action/message.py
xingyaoww: We could, but i think it is probably a big messy to plumb them in now - maybe we can leave it for now?

openhands/controller/agent_controller.py
enyst: Oh, it's okay, I'm just saying there is no issue, the logic works with `start_id` specified or without. Better specified, just to be explicit.

xingyaoww:
I ran an 50 instance eval on this branch:
```
Total instances: 500
Instances submitted: 50
Instances completed: 50
Instances incomplete: 450
Instances resolved: 32
Instances unresolved: 18
Instances with empty patches: 0
Instances with errors: 0
Unstopped containers: 0
Unremoved images: 500
Report written to claude-3-7-sonnet-20250219_maxiter_100_N_v0.33.0-no-hint-sysmsg-refactor-pr7824-run_2.20250417_033845.json
MODEL_NAME_OR_PATH: claude-3-7-sonnet-20250219_maxiter_100_N_v0.33.0-no-hint-sysmsg-refactor-pr7824-run_2
RESULT_OUTPUT_DIR: /home/xingyaow/OpenHands-dev/evaluation/evaluation_outputs/outputs/princeton-nlp__SWE-bench_Verified-test/CodeActAgent/claude-3-7-sonnet-20250219_maxiter_100_N_v0.33.0-no-hint-sysmsg-refactor-pr7824-run_2
Checking for changes: 0it [00:00, ?it/s]
Updating output file: 50it [00:00, 230.06it/s]
```

Agent performance is within expectation

And the cost wasn't much higher (prompt caching is still working - 0.76 per instance).

```
❯ cat evaluation/evaluation_outputs/outputs/princeton-nlp__SWE-bench_Verified-test/CodeActAgent/claude-3-7-sonnet-20250219_maxiter_100_N_v0.33.0-no-hint-sysmsg-refactor-pr7824-run_2/output.jsonl | jq .metrics.accumulated_cost | awk '{sum+=$1} END {print "Total accumulated cost: " sum}'
Total accumulated cost: 38.8542
```

Will merge this PR now

PR #7825, opened 2025-04-11
fix(backend): Return 400 if trying to open a binary file

amanape:

If the user tries to open a binary file (e.g., https://github.com/Azure-Samples/Machine-Learning-Operationalization/blob/master/samples/python/code/iris/model.pkl), the app throws a 500. This is problematic for SaaS since it may cause a false incident

---
Return 400. I was unable to figure out a better way to propagate custom errors from the action execution server in the short time

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:5fb00b8-nikolaik   --name openhands-app-5fb00b8   docker.all-hands.dev/all-hands-ai/openhands:5fb00b8
```

### DIFF
--- a/openhands/runtime/action_execution_server.py
+++ b/openhands/runtime/action_execution_server.py
@@ -18,6 +18,7 @@ from contextlib import asynccontextmanager
 from pathlib import Path
 from zipfile import ZipFile
 
+from binaryornot.check import is_binary
 from fastapi import Depends, FastAPI, HTTPException, Request, UploadFile
 from fastapi.exceptions import RequestValidationError
 from fastapi.responses import FileResponse, HTMLResponse, JSONResponse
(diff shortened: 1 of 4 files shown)


openhands/runtime/action_execution_server.py
```diff
@@ -76,6 +76,39 @@ class ActionRequest(BaseModel):
...
+
+        try:
+            sample.decode('utf-8')
+            return False
+        except UnicodeDecodeError:
+            # If decoding fails, it's likely binary
+            return True
+
+    except IOError:
+        return None
```
raymyers, line 109: This probably handles enough cases for our purposes, but if we're going to implement it here we should have tests around it.

What might be easier, I notice we have the `binaryornot` package in our dependency lock file, so maybe this is a good opportunity to use that and upgrade it to a direct dependency in `pyproject.toml`. 

https://github.com/binaryornot/binaryornot/tree/main

openhands/server/routes/files.py
```diff
@@ -135,6 +135,13 @@ async def select_file(file: str, request: Request):
         return {'code': content}
     elif isinstance(observation, ErrorObservation):
         logger.error(f'Error opening file {file}: {observation}')
+
+        if 'ERROR_BINARY_FILE' in observation.message:
+            return JSONResponse(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                content={'error': f'Unable to open binary files: {file}'},
```
raymyers, line 142: nit: files should probably be singular.

raymyers (changes requested):

raymyers:
I agree 400 is probably the best status code here. (One could also make a case for 415 or 422, wouldn't make any difference)

Do we know if/how the client will handle it? Of course it's probably not going to be worse than with the 500...

amanape:
> Do we know if/how the client will handle it? Of course it's probably not going to be worse than with the 500...

The default behaviour for all types of error responses are a quick toast on the top right with the error message passed to the JSONResponse object.

raymyers (approved):

amanape:
Just to note, we have a retry backoff mechanism that attempts up to 3 retries before showing the toast notification.

raymyers:
> Just to note, we have a retry backoff mechanism that attempts up to 3 retries before showing the toast notification.

Ah good to know. That's not perfect here cause it'll just fail again but since it's an edge case and fast it seems just fine

Issue #7850, opened 2025-04-14
Support for condensation pipelines


csmith49:
**What problem or use case are you trying to solve?**

There are situations where it's handy to have multiple condensers simultaneously: the `BrowserOutputCondenser` is great for reducing costs when the agent is browsing, and the `LLMSummarizingCondenser` is great at reducing costs for long-running tasks.

But there's no condenser that's great for long-running tasks when the agent is also browsing.

**Describe the UX or technical implementation you have in mind**

We could write a new condenser that combines the behavior of the `BrowserOutputCondenser` and the `LLMSummarizingCondenser`. That's likely to lead to all condenser behavior centralizing to a single, difficult-to-maintain implementation. **The alternative proposed here** is to make it so that individual condensers can be "composed" by the agent.

The central task in developing a condenser pipeline is handling the composition. Condensers act via the function:

```
condense: View -> View | CondensationAction
```

Some condensers, like `BrowserOutputCondenser`, will _never_ emit a `CondensationAction`, which makes their inclusion trivial: given two functions `View -> View` I can always make a new function `View -> View` by composing them. 

(This might warrant a new base class for them to inherit from to make that behavior clear.)

When we have a condenser that can emit either a `View` or a `CondensationAction`, we don't have any other way to handle the `CondensationAction` except for passing it back to the agent -- no downstream condenser can do anything with that information.

So the proposal: add a new `Condenser` implementation called `CondenserPipeline` that is parameterized by a _list_ of condensers and whose `condense` function acts something like:

```python
def condense(self, view: View) -> View | CondensationAction:
    result: View | CondensationAction = view
    for condenser in self.condensers:
        result = condenser.condense(result)
        if isinstance(result, CondensationAction):
            break
    return result
```

Configuration might be tricky (as evidenced by #7793), and we'll want to be careful about tuning a condenser pipeline -- intersectional effects likely mean we'll want to re-run experiments to identify the appropriate parameterization for the entire pipeline.

**Additional context**

xingyaoww:
After we fix this, we should turn visual browsing on by default

mamoodi:
Was this fixed by the merged PR?

xingyaoww:
Yes, this should be fixed.

PR #7867, opened 2025-04-15
feat: Combining condensers

csmith49:

---

This PR adds a new condenser that chains together multiple condensers without modifying the interface. The design follows the proposal in #7850 almost verbatim:

1. A new `CondenserPipeline` that is parameterized by a list of condensers.
2. The `CondenserPipeline.condense` function chains each sub-condenser together, until all condensers have had a chance to act or one returns a `CondensationAction` instead of a `View`.

In addition, this PR updates the default condenser configuration to chain a `BrowserOutputCondenser` with the existing `LLMSummarizingCondenser`.

---

A partial solution to #7850 -- this PR does not intend to address configuration from `config.toml` or environment variables.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:efc8d96-nikolaik   --name openhands-app-efc8d96   docker.all-hands.dev/all-hands-ai/openhands:efc8d96
```

### DIFF
--- a/openhands/memory/condenser/impl/pipeline.py
+++ b/openhands/memory/condenser/impl/pipeline.py
@@ -1,0 +1,44 @@
+from __future__ import annotations
+
+from contextlib import contextmanager
+from openhands.controller.state.state import State
+from openhands.core.config.condenser_config import CondenserPipelineConfig
+from openhands.events.action.agent import CondensationAction
+from openhands.memory.condenser.condenser import Condenser
+from openhands.memory.view import View
+
+class CondenserPipeline(Condenser):
+    """Combines multiple condensers into a single condenser.
+
+    This is useful for creating a pipeline of condensers that can be chained together to achieve very specific condensation aims. Each condenser is run in sequence, passing the output view of one to the next, until we reach the end or a `CondensationAction` is returned instead.
+    """
+    def __init__(self, *condenser: Condenser) -> None:
+        self.condensers = list(condenser)
+        super().__init__()
+
+    @contextmanager
+    def metadata_batch(self, state: State):
+        try:
+            yield
+        finally:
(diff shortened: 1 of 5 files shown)


openhands/memory/condenser/impl/pipeline.py
```diff
@@ -0,0 +1,47 @@
...
+            # and manually collect the relevant metadata.
+            for condenser in self.condensers:
+                condenser.write_metadata(state)
+
+    def condense(self, view: View) -> View | Condensation:
+        result: View | Condensation = view
+        for condenser in self.condensers:
+            result = condenser.condense(result)
+            if isinstance(result, Condensation):
+                break
```
xingyaoww, line 38: Is it possible that after a "Condesation", another condenser after it is not processed yet -- can we raise a warning?

xingyaoww (commented):
Overall LGTM! Just one question!

openhands/memory/condenser/impl/pipeline.py
csmith49: That's entirely possible, and actually intended behavior -- the only thing the agent can do with the `Condensation` is give it back to the agent controller (which triggers another step), so it's not possible for the agent to ever make a completion call without every condenser in the stack having had a say in the final view.

I'll add some logging, but `debug`/`info` is probably more appropriate than `warning`.

xingyaoww: Sounds good.

I read the code a bit and now i understand -- since condenser isn't really expecting an input type of Condensation

xingyaoww (approved):
LGTM!

PR #7871, opened 2025-04-16
[Refactor]: Collapse initial user message for cloud resolver

malhotra5:

- Make initial cloud resolver instructions a collapsable message when job is started
 
---

This PR does the following

1. Moves the cloud resolver jinja templates to OSS
2. Stores the conversation trigger in `ConversationMetadata`
3. When the conversation is triggered by cloud resolver, we make the initial resolver instructions a collapsable message

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:702b4b3-nikolaik   --name openhands-app-702b4b3   docker.all-hands.dev/all-hands-ai/openhands:702b4b3
```


enyst:
Maybe we can recognize the "comment" in it, and display that, in an expandable message that shows it all only when clicked?

malhotra5:
Yeah that's a good idea too! I opted for additional agent instructions because it gets loaded in as a microagent at the start of the session. We also already have UI in place to show the microagent details in an expandable message


image

malhotra5:
Ah I see, yeah let me rework this PR then

malhotra5:
I've updated this PR so that the first resolver message is expandable

image
image

frontend/src/api/open-hands.types.ts
```diff
@@ -70,13 +70,19 @@ export interface AuthenticateResponse {
   error?: string;
 }
 
+export enum ConversationTrigger {
+  CLOUD_RESOLVER = "cloud_resolver",
```
enyst, line 74: Could we name it just "resolver"? It's just a conversation flag.

openhands/storage/data_models/conversation_metadata.py
```diff
@@ -1,16 +1,25 @@
...
+class ConversationMetadata:    
     conversation_id: str
     github_user_id: str | None
     selected_repository: str | None
     user_id: str | None = None
     selected_branch: str | None = None
     title: str | None = None
     last_updated_at: datetime | None = None
+    trigger: ConversationTrigger | None = None
+    providers: list[ProviderType] | None = None
```
enyst, line 22: Is it necessary to add providers here?

malhotra5: In the very near future, we will be able to trigger the resolver from multiple providers (Github vs Gitlab)

So this helps track which providers were set when the conversation was created

In GUI mode, it could be multiple providers configured (Github + Gitlab)

enyst (approved):
Thank you! You may want an opinion from our UI master, but LGTM!

I left a question on those `providers` moving around, I'm not sure, could we test that kind of refactoring locally? It... has bad habits 😅

openhands/storage/data_models/conversation_metadata.py
enyst: Ah. Don't we have that information already? 🤔

malhotra5: Hmm don't believe we track it. We check whether the providers set are valid when the conversation is created, but don't save the provider information in a persistent way.

enyst: But we are using the tokens during agent's run - sorry, I'm still a bit confused, maybe I'm just tired. Maybe we could we remove it for now, and look at it in its own right when time comes?

They must be persistent in `settings`

malhotra5: Ah the tokens are not persistent either. For example 

1. I've logged into both Github and Gitlab
2. I started resolver job inside Github
3. Later, I logged out of Github but I'm still logged into Gitlab
4. I revisit the resolver conversation, which will load but it won't contain the Github token and will contain the Gitlab one instead

The tokens are exported in the agent's runtime based on the current state of providers set, not based on the providers set when the conversation was initialized

Regardless, I think I'll remove the providers information for now

openhands/integrations/github/templates/pr_update_prompt.j2
```diff
@@ -0,0 +1,25 @@
+You are working in /workspace, checked out to branch {{ branch_name }}, which has an open PR #{{ pr_number }}.
```
rbren, line 1: we should remove these references to `/workspace`. The runtime is CD'ed into the repo dir, and this ends up confusing the agent

openhands/integrations/github/templates/pr_update_prompt.j2
```diff
@@ -0,0 +1,25 @@
...
+A comment on the PR (below) has been addressed to you. Do NOT respond to this comment via the GitHub API. 
+
+{% if file_location %} The comment is on the file `{{ file_location }}`{% endif %}.
```
rbren, line 4: is line number available here?

malhotra5: Yeah I can add it in

frontend/src/components/features/chat/messages.tsx
```diff
@@ -4,20 +4,46 @@ import { ChatMessage } from "#/components/features/chat/chat-message";
...
         isAwaitingUserConfirmation;
 
+      // Special case: First user message with resolver trigger
+      if (index === 0 && message.sender === "user" && isResolverTrigger) {
+        return (
+          <div key={index}>
+            <ExpandableMessage
+              type="action"
+              message={message.content}
+              customHeader="Resolver Instructions"
```
rbren, line 38: We should use the existing header logic to make this a translatable string

openhands/integrations/github/templates/pr_update_prompt.j2
malhotra5: Actually I'll add this information in a separate PR; I'd like to simply move the templates into OSS in this PR

frontend/src/components/features/chat/messages.tsx
malhotra5: image

Updated to use translatable string instead

frontend/src/api/open-hands.types.ts
```diff
@@ -70,13 +70,19 @@ export interface AuthenticateResponse {
   error?: string;
 }
 
+export enum ConversationTrigger {
+  RESOLVER = "resolver",
+  GUI = "gui",
+}
```
amanape, line 76: For most cases a regular union type could be preferred for it's simplicity. e.g., `export type ConversationTrigger = "resolver" | "gui"`

It is more "lightweight" and doesn't generate JS objects at runtime.

Fine here! Maybe for next time time

frontend/src/components/features/chat/messages.tsx
```diff
@@ -4,20 +4,47 @@ import { ChatMessage } from "#/components/features/chat/chat-message";
...
+      conversation?.trigger === ConversationTrigger.RESOLVER;
+
+    return messages.map((message, index) => {
       const shouldShowConfirmationButtons =
         messages.length - 1 === index &&
         message.sender === "assistant" &&
         isAwaitingUserConfirmation;
 
+      // Special case: First user message with resolver trigger
+      if (index === 0 && message.sender === "user" && isResolverTrigger) {
```
amanape, line 33: Nit: better readability
```ts
const isFirstUserMessageWithResolverTrigger = index === 0 && message.sender === "user" && isResolverTrigger;
```

amanape (approved):
Comments not a blocker, feel free to merge!

Issue #7879, opened 2025-04-16
[Bug]: LLM Completion Tokens for LLMSummaryCondensator are not included in trajectory.json


happyherp:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

**It appears that there are **two** condensation llm-calls happening and neither are being tracked in the trajectory.json.** 

I discovered this when debugging the LLMSummarizingCondenser. 

I start a new conversation on my test-project: https://github.com/happyherp/openhands-test

I set a breakpoint in CodeActAgent._step and wait for it to be hit. When hit, I set `self.condenser.max_size` to 10 to more quickly reach a condensation. Then I disable the breakpoint.

I set a breakpoint in LLMSummarizingCondenser: get_condensation

I ask the agent to read all files. This triggers the breakpoint in LLMSummarizingCondenser: get_condensation.

I note the value of `self.llm.metrics.get()`

```
{
    "model": "anthropic/claude-3-7-sonnet-20250219",
    "prompt_tokens": 1307,
    "completion_tokens": 140,
    "cache_read_tokens": 0,
    "cache_write_tokens": 0,
    "response_id": "chatcmpl-040366cb-cf32-4732-8e8b-1160eb01bed2"
}
```
I also note that summary_event is AgentCondensationObservation('No events summarized')
I note the value of summary
I let the program resume. 

Immediately the breakpoint is hit **again**.

now the summary-event contains the event that was created from the previous call. 

`self.llm.metrics.get()` now contains both the previous and the current llm-call.

```
"token_usages": [
    {
        "model": "anthropic/claude-3-7-sonnet-20250219",
        "prompt_tokens": 1307,
        "completion_tokens": 140,
        "cache_read_tokens": 0,
        "cache_write_tokens": 0,
        "response_id": "chatcmpl-040366cb-cf32-4732-8e8b-1160eb01bed2"
    },
    {
        "model": "anthropic/claude-3-7-sonnet-20250219",
        "prompt_tokens": 6710,
        "completion_tokens": 226,
        "cache_read_tokens": 0,
        "cache_write_tokens": 0,
        "response_id": "chatcmpl-a4978e4a-0220-45f6-b0f8-15aac015884b"
    }
]
```

I let the program resume. because of the low value, of condenser.max_size, another condensation happens later on. I stop the agent and download the trajectory.json and compare it to what I saw when debugging. 

[...]

PR #7890, opened 2025-04-16
fix(metrics) Merge metrics of agent LLM and condenser LLM

happyherp:

The tokens used by LLM calls by condensers where not added to total sum in the trajectory.json and cost/tokens window. 

---

When setting the metrics of an action-event, it is now the sum of the metrics of the agent-llm and the condenser-llm. Before it was just the metrics of the agents LLM. 

---
https://github.com/All-Hands-AI/OpenHands/issues/7879

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -66,7 +66,6 @@ from openhands.events.observation import (
 )
 from openhands.events.serialization.event import event_to_trajectory, truncate_content
 from openhands.llm.llm import LLM
-from openhands.llm.metrics import Metrics
 
 # note: RESUME is only available on web GUI
 TRAFFIC_CONTROL_REMINDER = (
@@ -1116,15 +1115,21 @@ class AgentController:
         - accumulated_cost: The current total cost
         - accumulated_token_usage: Accumulated token statistics across all API calls
 
+        This includes metrics from both the agent's LLM and the condenser's LLM if it exists.
+
         Args:
             action: The action to attach metrics to
         """
+        metrics = self.agent.llm.metrics.copy()
+
+        # Include condenser metrics if they exist
+        if hasattr(self.agent, 'condenser') and hasattr(self.agent.condenser, 'llm'):
+            metrics.merge(self.agent.condenser.llm.metrics)
+
         # Create a minimal metrics object with just what the frontend needs
-        metrics = Metrics(model_name=self.agent.llm.metrics.model_name)
-        metrics.accumulated_cost = self.agent.llm.metrics.accumulated_cost
-        metrics._accumulated_token_usage = (
-            self.agent.llm.metrics.accumulated_token_usage
-        )
+        metrics._token_usages = []
+        metrics._response_latencies = []
+        metrics._costs = []
 
         action.llm_metrics = metrics
 
--- a/openhands/llm/metrics.py
+++ b/openhands/llm/metrics.py
@@ -1,3 +1,4 @@
+import copy
 import time
 
 from pydantic import BaseModel, Field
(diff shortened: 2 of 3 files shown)


openhands/controller/agent_controller.py
```diff
@@ -1116,15 +1115,21 @@ def _prepare_metrics_for_frontend(self, action: Action) -> None:
...
+
         # Create a minimal metrics object with just what the frontend needs
-        metrics = Metrics(model_name=self.agent.llm.metrics.model_name)
-        metrics.accumulated_cost = self.agent.llm.metrics.accumulated_cost
-        metrics._accumulated_token_usage = (
-            self.agent.llm.metrics.accumulated_token_usage
-        )
+        metrics._token_usages = []
+        metrics._response_latencies = []
+        metrics._costs = []
```
rbren, line 1132: is there a lot of data here? maybe we can just leave it?

rbren (approved):

openhands/controller/agent_controller.py
xingyaoww: I think there's a lot.. a 100 turn session probably have 100 dictionary, and that is send to FE per action so it can be huge

xingyaoww, line 1132: Actually, can we create a new metric object, and only copy over the field we care about (i.e., accumulated_cost and stuff) from the merged metric?

This will help in case we add more fields to `Metrics` that can be pretty large in the future.

openhands/controller/agent_controller.py
```diff
@@ -1116,15 +1115,21 @@ def _prepare_metrics_for_frontend(self, action: Action) -> None:
...
 
+        This includes metrics from both the agent's LLM and the condenser's LLM if it exists.
+
         Args:
             action: The action to attach metrics to
         """
+        metrics = self.agent.llm.metrics.copy()
+
+        # Include condenser metrics if they exist
+        if hasattr(self.agent, 'condenser') and hasattr(self.agent.condenser, 'llm'):
```
xingyaoww, line 1126: I think maybe we can make .condenser a @property of the base agent class, and it should return None when it is not presented. would love to hear @csmith49's thought here

openhands/controller/agent_controller.py
enyst: Yes, every action has the data of all those preceding it

happyherp: But then again, while we got this going on, these fields are not really an issue. 

image

happyherp: There are two ways I would rather do that
* change serialization/event.py event_to_trajectory or event_to_dict so that those fields are not copied. Such that we have all type of "exclusions" in one location.
* create a new EventMetrics that you can create from regular metrics which becomes the new type of event.metrics. And that would only contain the fields we want.

openhands/controller/agent_controller.py
csmith49: That's the cleanest approach. The `AgentConfig` has a field for the condenser config, but right now only the CodeAct agent takes advantage of it.

Could even initialize the condenser from the config in the base classes `__init__`. Since the `AgentConfig` has `NoOpCondenser` as the default we wouldn't even need the `None` check.

csmith49 (approved):
LGTM, with one suggested change:

The `LLMSummarizingCondenser` [records the local LLM metrics as metadata](https://github.com/All-Hands-AI/OpenHands/blob/caf34d83bd754bd134f0fe75582a9a82fcbcbff0/openhands/memory/condenser/impl/llm_summarizing_condenser.py#L140), which can then be picked up by the state later (like what happens in the [evaluation scripts](https://github.com/All-Hands-AI/OpenHands/blob/caf34d83bd754bd134f0fe75582a9a82fcbcbff0/evaluation/utils/shared.py#L568-L572)). Might be best to remove that so we're not double-storing the metrics.

xingyaoww (approved):
I'm ok with merging this PR as is, and follow-up with refactor PR to unblock this :)

PR #7891, opened 2025-04-16
[Feat]: Custom secrets plumbing for BE

malhotra5:

- Allow users to store custom secrets securely in the BE

---

This PR adds scafolding/routes to the BE for saving custom secrets 

It compliments #7929

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0f9b6b8-nikolaik   --name openhands-app-0f9b6b8   docker.all-hands.dev/all-hands-ai/openhands:0f9b6b8
```


openhands/server/routes/settings.py
```diff
@@ -55,6 +55,96 @@ async def load_settings(request: Request) -> GETSettingsModel | JSONResponse:
         )
 
 
+@app.get('/settings-custom-secrets', response_model=GETSettingsCustomSecrets)
```
rbren, line 58: @openhands change these routes to GET /secrets, POST /secrets, and DELETE /secrets/{id}

Also fix the lint errors

malhotra5:
@openhands please fix the lint issues in the BE

openhands/integrations/provider.py
```diff
@@ -66,6 +66,10 @@ class SecretStore(BaseModel):
         default_factory=lambda: MappingProxyType({})
     )
 
+    custom_secrets: CUSTOM_SECRETS_TYPE = Field(
+        default_factory=lambda: MappingProxyType({})
```
enyst, line 70: Could you please tell, why do we need a `MappingProxyType`? I always forget, and it's a bit difficult to read 😅

malhotra5: `MappingProxyType` is used to make `custom_secrets`  immutable (as opposed to using `dicts`)

malhotra5: More context, we have a test file `tests/unit/test_provider_immutability.py` to confirm the immutability of these fields

enyst: Ahh, thank you. Hmm, I wonder if we could do that differently. I remember now, it was designed to save us from ourselves 😅 so that we don't write mutable default in the wrong place. It may be worth thinking to simplify it though. Feel free to ignore if it's only me who feels that way on it.

rbren (approved):
lgtm! agree MappingProxyType seems complicated but I don't understand it enough to have an opinion

malhotra5:
I think I'll have a chat with @tofarr on how we can simplify mappingproxytype, but I'd like to keep it for now

PR #7892, opened 2025-04-17
fix(check-unlocalized-strings): make HTML tag test case-insens

LoneRifle:

When  checking for unlocalized strings pre-commit, it is possible for tag names to be upper case, 
and/or contain attributes, so ensure that we catch those in the regexp


amanape (approved):

negalicia84 (commented):
@aj47

PR #7898, opened 2025-04-17
[Feat] Support o3 and o4 mini

KennyDizi:

**PR Description**

- Support model `o3`
- Support m model `o4-mini`


openhands/llm/llm.py
```diff
@@ -69,15 +69,23 @@
     'o1-2024-12-17',
     'o3-mini-2025-01-31',
     'o3-mini',
+    'o3',
+    'o3-2025-04-16',
+    'o4-mini',
+    'o4-mini-2025-04-16',
```
enyst, line 75: Just a thought, I think that if we list here the starting part (`o4-mini`), then both will be recognized.

It probably wasn't always the case, which is why we have some redundancy in this list. 😅

enyst (approved):
Thank you for this. Technically, we haven't verified them yet (it should be anytime now), but it's likely fine, and this makes things easier for people.

openhands/llm/llm.py
KennyDizi: Good point @enyst . However, I'm not sure what you mean by redundant. I'm following the current structure and add `o3` with its variant `o3-2025-04-16`, `o4-mini` and its variant `o4-mini-2025-04-16`, you can review the full file `https://github.com/All-Hands-AI/OpenHands/blob/c1249e1935c81a3c9412b3ae4e203bed4cd4179a/frontend/src/utils/verified-models.ts` for more context.

If there is a redundant, I would like to fix that.

KennyDizi:
> Thank you for this. Technically, we haven't verified them yet (it should be anytime now), but it's likely fine, and this makes things easier for people.

That's true @enyst `o3` and `o4-mini` are highly recommended for coding tasks.

openhands/llm/llm.py
enyst: I mean, this code is run on the model name:
https://github.com/All-Hands-AI/OpenHands/blob/caf34d83bd754bd134f0fe75582a9a82fcbcbff0/openhands/llm/llm.py#L470

If we add 'o4-mini', this will be True for 'o4-mini-anything' too, right?
If so, we don't need the variant explicitly listed.

enyst: To clarify, we don't need them listed anywhere to be _supported_, this list is only for us to decide if when we run them, function calling should be enabled by default without any configuration needed from the user.

They're already supported, if they're supported by litellm. Which was a bit problematic last night:
- https://github.com/BerriAI/litellm/issues/10066
- https://github.com/BerriAI/litellm/issues/10080

One has just been fixed, and we will probably need to update. The other doesn't seem fixed, and if the models don't work with openhands, we could probably fix it by adding them to this other list...
https://github.com/All-Hands-AI/OpenHands/blob/caf34d83bd754bd134f0fe75582a9a82fcbcbff0/openhands/llm/llm.py#L83

I can give it a try in a bit, too. Have they worked for you?

frontend/src/utils/verified-models.ts
```diff
@@ -2,6 +2,8 @@
 export const VERIFIED_PROVIDERS = ["openai", "azure", "anthropic", "deepseek"];
 export const VERIFIED_MODELS = [
   "o3-mini-2025-01-31",
+  "o3-2025-04-16",
+  "o4-mini-2025-04-16",
```
xingyaoww, line 6: I would probably hold-off putting it in verified model yet until we got good result from our evaluation 🤔

enyst: Definitely! Still running?

xingyaoww: yeah, we should have result for o4-mini very soon

openhands/llm/llm.py
manishshettym: @xingyaoww -- just checking if this PR for o4-mini is working for you on swe-bench? I am unable to get past issues like:
```
litellm.exceptions.BadRequestError:` litellm.BadRequestError: OpenAIException - Error code: 400 - {'error': {'message': "Unsupported parameter: 'stop' is not supported with this model.", 'type': 'invalid_request_error', 'param': 'stop', 'code': 'unsupported_parameter'}}
```

PR #7920, opened 2025-04-18
Make extraction of poetry_venvs_path more robust

MischaPanch:

The parsing of poetry_venvs_path failed in certain installation scenarios where `Skipping virtualenv creation, as specified in config file.` would be logged to stderr. The redirection fixes that

I also added a few more comments and more graceful error handling


xingyaoww (approved):
Thanks!

PR #7670, opened 2025-04-02
[Fix]: Ensure refresh logic works for restarted conversations in cloud openhands

malhotra5:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality that this introduces.**

- This fixes refresh logic for multiple providers in cloud openhands; this, however, is just scaffolding changes and won't have any effect in the upcoming release

---
**Give a summary of what the PR does, explaining any non-trivial design decisions.**

#7532 included a hotfix where GitHub token refresh failed because provider information was lost for conversations that were restarted. Since the refresh logic only exports the tokens for the providers that are currently set, it wouldn't export any tokens due to the lost provider information. 

This PR ensures that when conversations are joined, the information about which providers are set is available in the runtime. This ensures that refresh logic takes place as expected  

---
**Link of any specific issues this addresses.**

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:7e367a1-nikolaik   --name openhands-app-7e367a1   docker.all-hands.dev/all-hands-ai/openhands:7e367a1
```

Issue #7419, opened 2025-03-22
Run setup.sh only once (automatically)


enyst:
`setup.sh` will be run on session restore, which is probably not expected in various cases.

We should be able to detect that somehow and avoid to run it again.

### If you find this feature request or enhancement useful, make sure to add a 👍 to the issue

PR #7923, opened 2025-04-18
Fix #7286: Load selected_repository from conversation metadata

### LINKED ISSUE #7286: [Bug]: The selected_repo is not saved, so not reloaded

I'll record this here: I ran with UI, selected a repo for a new conversation, started talking to the LLM
=> selected_repo was set
Then reloaded the conversation
=> selected_repo was None

I saw some errors:
- an LLM related error, on non-native fc LLM (probably unrelated)
- an error or strange behavior related to #6909 
    - addressed by https://github.com/All-Hands-AI/OpenHands/pull/7277
- attempt to repro was blocked due to #7273

### OpenHands Installation

Development workflow

### OpenHands Version

main

### Operating System

MacOS

_No response_

g-votte:

Fixes issue #7286 where selected repository was not loaded when restarting conversations. Added logic to properly load selected repository from conversation metadata.

---

This PR addresses issue #7286 where repository information was not being properly loaded when restarting conversations. The changes include three main components:

1. Core Fix: Loading Repository Information
   - Implemented loading of `selected_repository` and `selected_branch` from conversation metadata
   - Current implementation places this logic in `listen_socket.py` during session initialization

2. Type Enhancement: String to Repository Object
   - Changed `selected_repository` type from string to Repository object to maintain complete repository information
   - This change was necessary because provider information is required for repository operations (as seen in [`base.py`'s `clone_repo` method](https://github.com/All-Hands-AI/OpenHands/blob/34989f8e964af4873098cf74b4efb32aea1089a9/openhands/runtime/base.py#L324-L328))
   - Updated frontend components to handle the `Repository` object format

3. Backward Compatibility
   - Added support for both `str` and `Repository` object formats in metadata
   - Added appropriate type hints and comments for the dual format support
   - Handles legacy string format by treating it as an unset repository

The changes span both backend and frontend to ensure consistent handling of the `Repository` object format while maintaining compatibility with existing conversations.

---

Fixes #7286


openhands/server/listen_socket.py
```diff
@@ -32,11 +38,17 @@
 
 def create_provider_tokens_object(
     providers_set: list[ProviderType],
+    secrets_store: SecretStore,
 ) -> PROVIDER_TOKEN_TYPE:
     provider_information = {}
 
+    existing_tokens = secrets_store.provider_tokens
```
enyst, line 45: I find curious that you needed to handle tokens. Could you please explain the need, is it broken now?

g-votte: Thanks for your swift comment!

I'm actually not entirely sure about the expected behavior here. However, I noticed that once `selected_repository` loading is fixed (which is the main purpose of this PR), [clone_repo](https://github.com/All-Hands-AI/OpenHands/blob/34989f8e964af4873098cf74b4efb32aea1089a9/openhands/runtime/base.py#L312) operations would start working - and they would fail without proper token handling. That's why I kept this token loading logic, but I'd welcome clarification if there's a better approach.

frontend/src/api/open-hands.types.ts
```diff
@@ -70,12 +70,26 @@ export interface AuthenticateResponse {
   error?: string;
 }
 
+export enum ProviderType {
+  GITHUB = "github",
+  GITLAB = "gitlab",
+}
```
malhotra5, line 76: We already have enums for this in `frontend/src/types/settings.ts` that can be reused here

openhands/server/listen_socket.py
malhotra5: Hmm this is interesting, we shouldn't need to manipulate the provider tokens here. Since this PR has to do with restarting conversations, I would expected `clone_repo` to be unused. 

Also #7670 has the fix that ensures the provider information is sent to the runtime for restarted conversations (which should be enough I believe) 

My hunch is this logic has to do with loading the tokens for local installation of Openhands? Are the tokens lost in the conversation if you ask the agent to `echo $GITHUB_TOKEN` for instance?

g-votte: @enyst @malhotra5 
Here is one idea: https://github.com/All-Hands-AI/OpenHands/pull/7923/commits/2387a3c6cc7e4c3c42368e6a707ef5cd206ad3d0

In this approach, `clone_repo` is not invoked when the `git_provider_tokens` is created by `create_provider_tokens_object`.

- pros:
  - `clone_repo` invocation is avoided by a simple check: `_provider_tokens_are_empty`
- cons:
  - the logic between `agent_session.py` and `listen_socket.py` is somewhat tightly coupled
    - `_provider_tokens_are_empty` relies on assumptions about how `create_provider_tokens_object` initializes empty provider tokens
 
If the cons are acceptable, I think this approach is one of the simplest ways to avoid the error.

What do you think?

g-votte:
@enyst @malhotra5
Another important discussion would be whether we need to extend `selected_repository` from string to `Repository` type in this PR.

> 2. Type Enhancement: String to Repository Object
> - Changed selected_repository type from string to Repository object to maintain complete repository information
> - This change was necessary because provider information is required for repository operations (as seen in [base.py's clone_repo method](https://github.com/All-Hands-AI/OpenHands/blob/34989f8e964af4873098cf74b4efb32aea1089a9/openhands/runtime/base.py#L324-L328))
> - Updated frontend components to handle the Repository object format

Other than the `clone_repo`, some parts of codebase expect the `selected_repository` variable as the `Repository` type. E.g.,
- https://github.com/All-Hands-AI/OpenHands/blob/34989f8e964af4873098cf74b4efb32aea1089a9/openhands/server/session/conversation_init_data.py#L14
- https://github.com/All-Hands-AI/OpenHands/blob/34989f8e964af4873098cf74b4efb32aea1089a9/openhands/server/session/agent_session.py#L88
- https://github.com/All-Hands-AI/OpenHands/blob/34989f8e964af4873098cf74b4efb32aea1089a9/openhands/server/session/agent_session.py#L261

In the current proposal, I've adopted `Repository` type, which has resulted in large changes affecting both frontend and backend.
On the other hand, changing the backend type assumptions might result in fewer changes, but this would drastically reduce the information contained in `selected_repository`.

I'd like to hear the maintainers' opinions on this point as well.

frontend/src/api/open-hands.types.ts
g-votte: Thanks for your input!

Added a change to reuse the existing definition: https://github.com/All-Hands-AI/OpenHands/pull/7923/commits/34542d6e25b78a5996094e5101f763d418d11914

openhands/server/listen_socket.py
enyst: Please let me point out that until recently, the behavior was: on new conversation, `clone_repo` and other things, like `setup.sh` script were run; on reloading the conversation, they were _also_ run. (successfully) When `git clone` ran, it said there already is a directory and it's not empty; you know the message. So it was a no-op. And everything continued normally.

It's a bug that it fails with a runtime error now.

Now the question how it should behave is interesting. A while back, we thought that it _shouldn't_ run clone and even setup, on reloading, but we didn't see a nice way to make it happen, so we left it in the backburner, for example:
- https://github.com/All-Hands-AI/OpenHands/issues/7419

It makes sense to me that we only run them at initialization of a session, not all the time it's reloaded.

enyst: That said, I think you are correct that this logic is perhaps too tightly coupled... My first thought here is that
- the tokens should be sent to the runtime at session reloading too (for all that the agent needs to do), just like at initialization; 
- but, ideally, `clone` and `setup` won't run again.

I don't _love_ the idea of relying on them being empty to avoid an error that shouldn't be there... it feels yes, too assuming and fragile; but I don't understand all the logic here right now, I need to dig into the code. I don't remember off hand _when exactly_ do we initialize the tokens during agent session reload, it must be downstream of this code 🤔 👀

enyst:
> In the current proposal, I've adopted Repository type, which has resulted in large changes affecting both frontend and backend.

This makes sense to me, and I feel it may help in the future too.

frontend/src/api/open-hands.types.ts
```diff
@@ -70,12 +71,21 @@ export interface AuthenticateResponse {
   error?: string;
 }
 
+export interface Repository {
```
malhotra5, line 74: FYI we have an existing type `GitRepository` in `frontend/src/types/git.d.ts`!

openhands/storage/data_models/conversation_metadata.py
```diff
@@ -12,7 +13,9 @@ class ConversationTrigger(Enum):
 class ConversationMetadata:
     conversation_id: str
     github_user_id: str | None
-    selected_repository: str | None
+    # Accepting str for `selected_repository` for backward compatibility
+    # See also: https://github.com/All-Hands-AI/OpenHands/issues/7286
+    selected_repository: Repository | str | None
```
malhotra5, line 18: I wonder if we should make a separate field instead of having multiple types set for `selected_repository`?

It would make the migration a little bit easier, we could just remove the deprecated field after some time

g-votte: Agree.

I've made the changes as suggested - could you please review if they align with what you had in mind?
https://github.com/All-Hands-AI/OpenHands/pull/7923/commits/bd86c4a6274f08a7bcb9015f50e63d8fe6ae11e6

One additional benefit I noticed is that separating the fields also helps simplify the type checking logic around selected_repository.

frontend/src/api/open-hands.types.ts
g-votte: Thanks!

Changed the code to use the existing `GitRepository`.

openhands/server/listen_socket.py
g-votte: @enyst 
In `_create_runtime`, we need to distinguish between "new conversation" and "restarting existing conversation". This would help with issues like avoiding unnecessary repository setup on conversation restart.

Is there a way to determine this from `AppConfig` or other existing configurations? If not, I think we need to add this information to something like `ConversationInitData` and propagate it through the call chain to `_create_runtime`.

Do you have any suggestions on how we should handle this distinction?

enyst:
@g-votte I'm really sorry for the conflicts and the delays here. I'm afraid it's not just random conflicts, but we have actually happened to change a few times exactly the selected repo format or data back and forth, I think.

I'm honestly not sure what the best way forward is. What do you think?

mamoodi:
Checking in to see if this is still in progress?

PR #7862, opened 2025-04-15
Fix issue #7860: [Bug]: Resolver fails using GitHub Actions token in `send_pull_request.py`

openhands-agent:
This pull request fixes #7860.

The issue has been successfully resolved based on the following changes:

1. Added a new `repository` parameter to the `send_pull_request` function that accepts the repository in "owner/repo" format
2. Added a `--repository` CLI argument that defaults to the `GITHUB_REPOSITORY` environment variable
3. Modified the token identification logic to pass the repository information to `identify_token(token, my_args.repository)`
4. Added comprehensive test coverage for GitHub Actions token validation

These changes directly address the reported bug where the GitHub Actions token was being incorrectly marked as invalid. By passing the repository information, the token validation can now properly recognize and validate GitHub Actions tokens in addition to Personal Access Tokens (PATs).

The added test case `test_send_pull_request_with_github_actions_token` verifies that the GitHub Actions token workflow works correctly, including proper API calls for token validation and PR creation.

The changes are a complete solution to the original issue because they:
1. Fix the core problem of missing repository information in token validation
2. Maintain backward compatibility with existing PAT usage
3. Properly integrate with GitHub Actions environment variables
4. Include proper error handling and validation

Automatic fix generated by [OpenHands](https://github.com/All-Hands-AI/OpenHands/) 🙌

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:471f54b-nikolaik   --name openhands-app-471f54b   docker.all-hands.dev/all-hands-ai/openhands:471f54b
```

PR #7934, opened 2025-04-18
fix(resolver): Allow using github.token in Actions and fix base_domain handling

### LINKED ISSUE #7860: [Bug]: Resolver fails using GitHub Actions token in `send_pull_request.py`

When using the default GITHUB_TOKEN in github actions, we get the following error in step Create draft PR or push branch:

```
ERROR:root:<class 'ValueError'>: Token is invalid.
```

It looks like it's because we don't pass the repo into `identify_token` in https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/resolver/send_pull_request.py#L645 , which then ignores the possibility of a github actions token and only looks for a PAT. When it doesn't find a PAT it declares it invalid, even though the token has write access to PRs and can perform the action just fine.

Is this as simple as adding `--repository` as a argument and populating it with `GITHUB_REPOSITORY`?

I'm new around here, but this seems ripe for the openhands resolver to take and open a PR. 😄 

### OpenHands Installation

GitHub resolver

### OpenHands Version

0.32

### Operating System

None

_No response_

Tei1988:

Fixes an issue where creating pull requests from GitHub Actions using the standard `github.token` failed due to a token validation error. This change allows users to leverage the built-in `github.token` for automated pull requests within their workflows without needing a Personal Access Token (PAT). Additionally, it resolves an error that occurred when using OpenHands with GitHub Enterprise if the `base-domain` configuration was not explicitly set.

---

This PR addresses two issues related to the pull request functionality (`openhands-resolver`):

1.  **GitHub Actions `github.token` Support:** The `identify_token` function requires repository context to validate the `github.token` provided by GitHub Actions. This PR introduces a `--selected-repo` argument to `send_pull_request.py` which allows passing the repository name (`owner/repo`) to the `identify_token` function, enabling successful validation.
2.  **`base_domain` Handling:** The `identify_token` function in `utils.py` did not correctly handle the case where the `base-domain` argument (introduced for GitHub Enterprise support) was `None` (the default when omitted). It now correctly defaults to `github.com` in such cases, preventing unexpected validation failures.

These changes ensure the resolver works correctly in GitHub Actions environments using the standard token and also fix a potential configuration issue for GitHub Enterprise users. The default behavior for existing users remains unchanged if the new argument is not used.

We hope this PR offers a solution by addressing both the missing repository context and the `base_domain` handling, building upon the previous attempt in #7862.

---

*   Fixes #7860
*   Related to #7862


enyst (commented):
Thank you!

This looks just right to me, we seem to have missed this, @malhotra5 what do you think?

malhotra5 (approved):
LGTM, thanks so much for the fix!

Tei1988:
@enyst @malhotra5 Thank you for review and merge !

Issue #7919, opened 2025-04-18
Allow disabling all tools in CodeActAgent


MischaPanch:
From a discussion in [#7881](https://github.com/All-Hands-AI/OpenHands/issues/7881#issuecomment-2813125354)

>with MCP support, I think the CodeActAgent should not have default tools. Currently, the thinking, finish, and bash tools cannot be disabled since `get_tools` will always add them (see below). Many other MCP servers have similar tools, and this leads to collisions of semantics.  Should I open a separate issue on this?
> 
> ```python
> def get_tools(
>     codeact_enable_browsing: bool = False,
>     codeact_enable_llm_editor: bool = False,
>     codeact_enable_jupyter: bool = False,
>     llm: LLM | None = None,
> ) -> list[ChatCompletionToolParam]:
>     SIMPLIFIED_TOOL_DESCRIPTION_LLM_SUBSTRS = ['gpt-', 'o3', 'o1']
> 
>     use_simplified_tool_desc = False
>     if llm is not None:
>         use_simplified_tool_desc = any(
>             model_substr in llm.config.model
>             for model_substr in SIMPLIFIED_TOOL_DESCRIPTION_LLM_SUBSTRS
>         )
> 
>     tools = [
>         create_cmd_run_tool(use_simplified_description=use_simplified_tool_desc),
>         ThinkTool,
>         FinishTool,
>     ]
>     if codeact_enable_browsing:
>         tools.append(WebReadTool)
>         tools.append(BrowserTool)
>     if codeact_enable_jupyter:
>         tools.append(IPythonTool)
>     if codeact_enable_llm_editor:
>         tools.append(LLMBasedFileEditTool)
>     else:
>         tools.append(
>             create_str_replace_editor_tool(
>                 use_simplified_description=use_simplified_tool_desc
>             )
>         )
>     return tools
> ```

enyst:
I think we could do something like [this PR](https://github.com/All-Hands-AI/OpenHands/pull/7911/files#diff-42c4ff83d51517293df8094053520eaea4f13666ef0bcc4b23c245c118c70775R15) is doing for MCP tools, but for our local tools.

That is, configurable, with the same defaults as now.

MischaPanch:
Yes, but that's a larger change and is breaking backwards compatibility. As an intermediate step one could simply add a new field to the config:

```
codeact_enable_default_tools = true
```

or a more granular

```
codeact_enable_finish_tool = true
[...]

PR #7748, opened 2025-04-07
[Experiment] Gemini - API error tolerance, editing, prompting

enyst:
This PR is testing fixes and improvements to support better non-Anthropic LLMs, Gemini 2.5 Pro in particular.
- more tolerance for API errors
- implement diff editing
    - LLMs have seen in their training a lot of udiff and merge conflicts diff formats
    - the high results on Aider bench have been obtained with Gemini using the merge diff format, with a tweak inserting filenames inside the diff block
        => in testing
    - Claude 3.5 and 3.7 appear to have been post trained on the `str_replace_editor` bespoke format
- prompting: the system prompt, tool descriptions and user instruction
    - Gemini appears to interpret our current SWE bench instruction as "well, the user took care of the tests, so I should skip testing for this task" 😅
         => changed
    - the instruction and maybe system prompt describe in detail the steps to take. Gemini 2.5 Pro is a "reasoning" model, I'm not sure that for them, prodding them with a sort of CoT or detailed predefined planning is helpful.
         => simplified
    - to note, it seems that Sonnet 3.7 with "thinking" hasn't got improvements either - with the current prompts/tools.

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:6a49317-nikolaik   --name openhands-app-6a49317   docker.all-hands.dev/all-hands-ai/openhands:6a49317
```

PR #6947, opened 2025-02-25
[agent] Read-only Agent

xingyaoww:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality that this introduces.**

---
**Give a summary of what the PR does, explaining any non-trivial design decisions.**

Add a read-only agent that only have accessible read-only tools (e.g., grep, glob, view, etc) for research tasks.

image

Example of research outcome: 
image

---
**Link of any specific issues this addresses.**

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:dbaf42d-nikolaik   --name openhands-app-dbaf42d   docker.all-hands.dev/all-hands-ai/openhands:dbaf42d
```

PR #7936, opened 2025-04-18
Possibility to disable cmd, finish, editor and think tools

MischaPanch:
Solves #7919


config.template.toml
```diff
@@ -224,6 +224,15 @@ enable_llm_editor = false
...
 enable_jupyter = true
 
+# Whether the command tool is enabled
+enable_cmd = true
+
+# Whether the think tool is enabled
+enable_think = true
+
+# Whether the finish tool is enabled
+enable_finish = true
```
enyst, line 234: Yes, this is a way to do it. It doesn't cover them all, though, right? It seems it doesn't have the editor.

I can see arguments in favor of this approach: it's extending what is already the way we're doing this for a few tools, it's backwards compatible as you pointed out, it can do the job with flags.

But there are also arguments for doing it as _listed tools_: we'd load the list of tools, rather than do a lot of ifs; with MCP, we'd have a list of local tools and a list of MCP tools, mix and match them using similar-ish configurations; now with the [read-only agent](https://github.com/All-Hands-AI/OpenHands/pull/6947), those new read-only tools will raise the same question, so instead of adding booleans after booleans, we can just _name_ the tools chosen at any given time.

In other words, it seems to me that this approach is inline with the today's patterns in the codebase, while the other is where, i think, we are heading towards. 🤔

What do you think about the second approach? Is there a reason why it wouldn't work?

MischaPanch: Are you OK with breaking backwards compatibility in the config?

enyst: Yes, although I would _try_ not to. We generally say right now that we don't yet have a stable version, so well, it can happen, if it has to.

Would we need to break it?

MischaPanch: There's a bunch of things I would change in the code if more serious changes in the code were acceptable - like passing the config object instead of 6 enable flags and so. But all such changes break public interfaces. I don't know how much this form of backwards compatibility matters for openhands.

My bad about the editor tool, haven't noticed that one

MischaPanch: > would we need to break it

If you go to a list of allowed tools this breaks for people who used the already existing `enable_...` flags in the config

MischaPanch: @enyst let me know if I should do this breaking change in the config (would to it tomorrow then). Otherwise I'd add one more flag for disabling the editor tool

enyst: Re: pass config. Yes, that's definitely fine. We definitely change function signatures (almost) all the time. I came to that kind of conclusion too, e.g. [very much draft, just fyi](https://github.com/All-Hands-AI/OpenHands/pull/7748/files#diff-1c3bca1c2fb93868dc3262439f8d24f0af08417f86f317da25e6e28e005202ff)

Re: existing `enable_*`: This is trickier because it does affect people's installations, people who don't modify `openhands` or build on it, just use it. IMHO maybe we don't actually need to do that yet: we could start listing tools, mark the `enable_*` as deprecated, and still check it for now?

MischaPanch: I would recommend to not break user config if it's not strictly necessary, at least not in a minor release. You may be in alpha, but still many people already use open-hands. So it could be done later and marked as deprecated in a follow-up PR

enyst: Yes, keeping and checking the existing `enable_*` would do that.

MischaPanch:
Rebase apparently went wrong, I am going to rebuild the branch

PR #7951, opened 2025-04-19
Possibility to disable default tools

MischaPanch:

It is now possible to disable default tools. On config level this is backwards compatible and just adds a few more options

---

I did a few minor improvements to the code:

1. Docstrings of AgentConfig entries were moved to the actual entries so that an IDE can resolve them. Default mentions were removed (they are a duplication and one of them was already out of sync, there is really not much value to them).
2. get_tools was removed and turned into a private method. Tests of it were removed, they were not really testing anything. Instead a test was added that an agent with default config indeed has the default tools (also if it was up to me, I'd remove that test as well - configuration can't and shouldn't be tested)
---

Closes: #7936


MischaPanch:
@enyst pls review, I closed he last PR

config.template.toml
```diff
@@ -221,9 +221,22 @@ enable_browsing = true
...
 enable_jupyter = true
 
+# Whether the command tool is enabled
+enable_cmd = true
+
+# Whether the think tool is enabled
+enable_think = true
+
+# Whether the finish tool is enabled
+enable_finish = true
```
enyst, line 238: I'm sorry, I think we somehow misunderstood each other. I prefer to introduce them like MCPs, in a list of enumerated tools, in agent config. That will work towards a similar interface to use, by the agent, for local tools and MCP tools alike. We can keep the _existing_ `enable_*`, but I would wary of introducing _more_ `enable_*` settings that we will deprecate anyway later.

I don't feel very strongly about it, but just to clarify, this is how I thought about this.

MischaPanch: Yes, I thought about it and I thought a clean implementation would need a few new concepts. I generally dislike string interfaces, so I would have introduced something like a tool registry as an enum. This should be used consistently in the code.

Then the user should have easy access to the registry to understand which tool names are available and can be configured. Otherwise this info would need to be copy pasted to the config template and it would run out of sync almost immediately.

One would need to deal with invalid names that were provided and so on.

This becomes even more convoluted when MCP tools are allowed to be configured. What to do if there are name collisions with default tools?

I think in the near future one should have a proper solution for configuration of default and MCP tools, then one can get rid of all enable_ flags. But I think this solution will not just be a flat list in the config, but at the very list also a registry. Maybe a separate nested config setting that allows selecting tools for everything, including MCP servers.

For now, I wanted to get the issue done without a larger overhaul than the one I already did here

MischaPanch: The current implementation has the benefit that the toml config exactly mirrors the agent config, so there are no string interfaces involving tool names

MischaPanch:
I have a minor error in the new unit test. Ran out of battery, will fix later today. But feel free to push to this branch if you want to fix it before that :)

config.template.toml
enyst: OK! I'm not sure we can fix human error 😅 - after all, `enable_toolname` needs toolname, but we can go with this IMO, and give it some thought later, as you said.

enyst (approved):
Thank you!

PR #7961, opened 2025-04-20
Bugfixes: MCPConfig Attribute error and always using default prompt templates

MischaPanch:

The AttributeError is self-explanatory
 
As for the in-context prompts: in #7919 it was allowed to disable default tools.
But with them disabled, the attempt to use the default in context learning pre/postfix would lead to errors.

As a bonus feature in addition to the fix, the user is allowed to pass custom examples, which should be very useful when
also using custom tools.

---

Fixes bugs, adds a new configuration option in a backwards compatible way

---

### DIFF
--- a/config.template.toml
+++ b/config.template.toml
@@ -148,6 +148,15 @@ api_key = ""
 # Model to use. (For Headless / CLI only -  In Web this is overridden by Session Init)
 model = "gpt-4o"
 
+# You can provide a custom example for in context learning, useful when customizing tools
+# For backwards compatibility, not setting this field will lead to default in-context learning examples being used
+# which may fail if you have disabled the default tools.
+# In that case, you should set custom_in_context_learning_example to fix this.
+# Note that you can just set it to an empty string, which is enough for the default example to not be activated
+# SPECIAL CASE: if using openhands models, the default in context learning examples will never be used
+# in_context_learning_example=""
+
+
 # Number of retries to attempt when an operation fails with the LLM.
 # Increase this value to allow more attempts before giving up
 #num_retries = 8
--- a/openhands/core/config/llm_config.py
+++ b/openhands/core/config/llm_config.py
@@ -44,6 +44,10 @@ class LLMConfig(BaseModel):
         native_tool_calling: Whether to use native tool calling if supported by the model. Can be True, False, or not set.
         reasoning_effort: The effort to put into reasoning. This is a string that can be one of 'low', 'medium', 'high', or 'none'. Exclusive for o1 models.
         seed: The seed to use for the LLM.
+        in_context_learning_example: A custom example provided by the user. If not None (even with an empty string),
+            the default in context learning examples will not be used.
+            It may be necessary to pass it if default tools have been disabled to prevent runtime errors due to
+            the tool set not being compatible with the default in-context learning example.
     """
 
     model: str = Field(default='claude-3-7-sonnet-20250219')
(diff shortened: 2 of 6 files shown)


MischaPanch:
@enyst Pls have a look, followup from my last PR.

With this I was finally able to disable the default tools and use the MCP tools within OpenHands :)

enyst (changes requested):
Interesting, thank you for looking into this!

You are absolutely correct, we should allow customization of that hardcoded examples text. Honestly, I wanted for a while to take it out of code, even. 😅

I think a good place for the customizable text is a template file, rather than the configuration file: such as the `prompts/user_prompt.j2`. The user template had this purpose in the past, and as you can see it's empty now. If you want to try it, maybe we could use it. The problem with this is that the conversion is called from `llm.py`, which doesn't (and probably shouldn't) know about agent's prompt templates, but maybe we can find a reasonable solution?

Another alternative might be a microagent... a `.md` file in `.openhands/microagents/` which are accessible directly to the user in the workspace, so they can customize it easily. It's our current customization mechanism, though I'm not sure this one is a good idea for this feature, it doesn't seem to fit the other microagents. 🤔

What do you think?

MischaPanch:
I agree it should be outside of the code and there should be a dedicated and customizable template mechanism for that. Yeah, the config is not ideal. microagents (btw, very confusing name IMO, they're not agents but just some dynamically loaded prompts, right?) are not suitable for that since the examples are currently added to the very first user message, so it's a different mechanism.

A proper templating solution would take me some time, and I don't think I'll get to it before next week, maybe even not then. I just wanted to have the MCP tools running withing OpenHands.

As OpenHands is becoming more, well, open towards other tools, several hardcoded behaviors are creating problems. We fixed the default tools not being configurable, but there are quite many prompts (including the system prompt), in context leearning examples, memories (aka) microagents sprinkled around the code, creeping in through different mechanisms (wrappers, config, default messages). It is difficult (currently still impossible without modifying code) to get a vanilla version of openhands with custom tools, custom prompts, custom memories (or abscense thereof).

I wanted to use the openhands framework to evaluate a toolkit (from [serena](https://github.com/oraios/serena)) on the swe benchmark, this is why I had all the PRs allowing to disable and configure more things. I think it would be very useful to go through everything and make it easy to have vanilla/customizable openhands. You probably have a better overview of what is needed to achieve this @enyst . I don't think it is all too much work, but till next week I still won't have time.

If you wanna have a chat about this, we could also have a short call to discuss a plan for achieving this goal :). I think it should be very useful to the community, especially after allowing external tools with MCP.

enyst:
You are correct, and thank you for the details!

Yes, I do think it's important to make these (relevant) prompt pieces customizable. Another step was recently when we added the system prompt to events, and by that, it's accessible now to people [directly](). Not there yet, but maybe it should also become updatable, at least programmatically.

Microagents might not be too suitable conceptually, but you are correct they're pieces of prompts, which ... is what this one is. It shouldn't matter exactly if the pieces are inserted in a message or different messages? At least because some code later may merge those messages (consecutive messages with the same role = 'user') 🤔 

> It is difficult (currently still impossible without modifying code) to get a vanilla version of openhands with custom tools, custom prompts, custom memories (or abscense thereof).

Yes, I think we can do better here. I can take a look and try to rethink this too, but your help is very much appreciated! Maybe a solution is sending some kwarg to llm.py to know which prompt templates to run the conversion with... 🤔

enyst:
@MischaPanch Thank you for the proposal, and sorry for the delay here. We have now refactored the in-context templates, split them per tool, so that now each example is loaded when the tool is enabled, and not when it's disabled:
- https://github.com/All-Hands-AI/OpenHands/pull/8310

I think this addressed the main issue of this PR in a different way. WDYT?

I am still to test Serena, I put it on my todo list, but I'm afraid other priorities took hold. 😭 I think it should (?) work as you intended, though? Maybe the next steps here will be to encapsulate tools, so that we have in one place per tool both the definition, and the validation, and presumably the in-context learning template for that tool. Please let me know your thoughts at your convenience!

Issue #7963, opened 2025-04-20
[Bug]: MCP function conversion wrong for gemini-preview


MischaPanch:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

With gemini preview, the presence of MCP tools leads to an invalid format for the request:

```
          {
            "field": "tools[0].function_declarations[26].parameters.properties[3].value",
            "description": "Invalid JSON payload received. Unknown name \"default\" at 'tools[0].function_declarations[26].parameters.properties[3].value': Cannot find field."
          }
```

Interestingly, with gemini-exp everything is fine. The error happens when changing from 
`gemini/gemini-2.5-pro-exp-03-25`
to `gemini/gemini-2.5-pro-preview-03-25`, in the former everything is OK

### OpenHands Installation

Docker command in README

### Operating System

None

enyst:
This is strange. According to google, they are just 2 identifiers of the same underlying model. 🤔

MischaPanch:
Agno has the same problem but in reverse
https://github.com/agno-agi/agno/issues/2805

PR #7964, opened 2025-04-20
Fix issue with "default" in tool schema for gemini-preview

MischaPanch:

See issue description

---

---

Closes: #7963

### DIFF
--- a/openhands/agenthub/codeact_agent/codeact_agent.py
+++ b/openhands/agenthub/codeact_agent/codeact_agent.py
@@ -1,3 +1,4 @@
+import copy
 import os
 from collections import deque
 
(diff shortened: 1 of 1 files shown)


enyst (commented):
Thank you for this! I _think_ I've hit this working with Gemini, with non-MCP tools with a default.

I can't confirm your fix right now, and there's something I don't understand about this difference in models (`-exp` and `-preview` are supposed to be the same underlying model, according to google). But... it wouldn't be the first bug with the API, and the same model thing was supposed to be temporary. Maybe they changed it? 🤷‍♂️

I think this kind of bug fix might belong in `litellm` if I understand it correctly, if it's a Gemini API issue. I'm okay with fixing it ourselves first, though, so that things are working.

MischaPanch:
Yes, I think Google is screwing around with their APIs. The real fix should be on Google's side, it's not litellms fault that we are calling the completion API with a "wrong" schema for the tools. So if you don't want to wait for Google, feel free to merge ;)

MischaPanch:
To reproduce, use any MCP server with defaults in their tools. I was using this one 
https://github.com/oraios/serena

enyst:
Looks like it has been reported on `litellm`:
- https://github.com/BerriAI/litellm/issues/9793

Just to clarify, LiteLLM does a lot of transformations to make APIs compatible with openai API and allow us to call them in an openai-compatible format. It's their main goal.

enyst (approved):

MischaPanch:
> Just to clarify, LiteLLM does a lot of transformations to make APIs compatible with openai API and allow us to call them in an openai-compatible format. It's their main goal.


Then you're right, it's on them. Sorry, I am not too familiar with litellm's design goals

Issue #7921, opened 2025-04-18
[Bug]: Hitting rate limits doesn't appear in UI


MischaPanch:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

When rate limits are encountered there are errors in the backend log, but nothing happens in the UI

### OpenHands Installation

Development workflow

### Operating System

Linux

### Logs, Errors, Screenshots, and Additional Context

Here a rate limit error by litellm

```
  raise RateLimitError(
litellm.exceptions.RateLimitError: litellm.RateLimitError: litellm.RateLimitError: VertexAIException - {
  "error": {
    "code": 429,
    "message": "You exceeded your current quota. Please migrate to Gemini 2.5 Pro Preview (models/gemini-2.5-pro-preview-03-25) for higher quota limits. For more information on this error, head to: https://ai.google.dev/gemini-api/docs/rate-limits.",
    "status": "RESOURCE_EXHAUSTED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.QuotaFailure",
        "violations": [
          {
            "quotaMetric": "generativelanguage.googleapis.com/generate_requests_per_model_per_day",
            "quotaId": "GenerateRequestsPerDayPerProjectPerModel"
          }
        ]
      },
      {
        "@type": "type.googleapis.com/google.rpc.Help",
        "links": [
          {
            "description": "Learn more about Gemini API quotas",
            "url": "https://ai.google.dev/gemini-api/docs/rate-limits"
          }
        ]
      }
    ]
  }
}
```

mamoodi:
I thought we added an agent status in the UI that showed rate limits. @raymyers I remember you put that in, did you not?

MischaPanch:
At least for me nothin shows up in the UI, do I have to configure something?

happyherp:
With claude3.7 it works for me. 

Just to be clear. We are talking about the status botton left not turning yellow and showing "Rate Limited" right? For you it shows a generic, red error-status like in the screenshot, right?

![Image](https://github.com/user-attachments/assets/6fd91772-3e20-45d0-942b-539aa03b9031)

happyherp:
The rate-limit error is set here

https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/controller/agent_controller.py#L253

```
            elif isinstance(e, RateLimitError):
[...]

PR #7970, opened 2025-04-21
fix(Core): Some Exceptions did not set the agent to "Rate Limited"

happyherp:
- [x] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below.

---

**End-user friendly description of the problem this fixes or functionality that this introduces:**  
In some edge cases, when the LLM returns an error message indicating too many requests — but not a `RateLimitError` — and it was the final retry attempt, the Agent would incorrectly switch its state to `ERROR` instead of `RATE_LIMITED`.

---

**Give a summary of what the PR does, explaining any non-trivial design decisions:**  
In `agent-controller`, all exceptions that would trigger a retry are now treated as rate-limit-related exceptions. This ensures more consistent handling of rate-limiting scenarios, even when the error type is not explicitly a `RateLimitError`.

---

https://github.com/All-Hands-AI/OpenHands/issues/7921

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -67,7 +67,7 @@ from openhands.events.observation import (
     Observation,
 )
 from openhands.events.serialization.event import event_to_trajectory, truncate_content
-from openhands.llm.llm import LLM
+from openhands.llm.llm import LLM, LLM_RETRY_EXCEPTIONS
 from openhands.llm.metrics import Metrics, TokenUsage
 
 # note: RESUME is only available on web GUI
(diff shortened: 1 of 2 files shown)


rbren (changes requested):
I don't think we should count all RETRY_EXCEPTIONS as RATE_LIMIT exceptions...

enyst:
I also have a small question in the linked topic: I'm curious about `litellm.RateLimitError` in the log, in the original bug report. This is an exception from litellm, and it should be the same exception, for both Claude and Gemini.

Why do you think that the miss from the UI is due to too many requests, when it seems that the exception in the log was RateLimit?

happyherp:
Probably this is not the way to go.

PR #7971, opened 2025-04-21
fix: Fixed 400 of empty text on Gemini by replacing it with a whitespace character

ahmed-z0:

**Problem**:
In rare cases, users interacting with Gemini 2.5 Pro Preview would receive an empty (blank) response. While the blank response itself didn't cause an immediate error, trying to send a follow-up message afterward would trigger a system error, preventing further conversation. The system would show a "Bad Request" error, indicating it couldn’t process the message because it was empty.

**Fix/Improvement**:
We’ve improved how the system handles these rare empty responses from Gemini. Now, when an empty message is detected, it is properly managed behind the scenes so that users can continue their conversations without running into errors or being blocked from sending further messages.

---

This PR introduces a safeguard specifically for Gemini models to handle cases where the model returns empty text content.

- Before proceeding with message serialization, we now check if the model name contains "gemini".

- If so, we scan through each message and its content items. If a content_item is of type "text" but has no text value, we replace the empty text with a single space (" ").

- This avoids triggering BadRequestError from Vertex AI, which occurs when trying to send an empty text parameter.

Additionally, as a small refactor:

-Instead of re-fetching self.config.model multiple times, we normalize it once into a model_name variable at the top.

- The previous DeepSeek model-specific logic (force_string_serializer) is updated to use model_name for consistency.

---

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -744,15 +744,26 @@ class LLM(RetryMixin, DebugMixin):
         self.metrics.reset()
 
     def format_messages_for_llm(self, messages: Message | list[Message]) -> list[dict]:
+        # normalize to list
         if isinstance(messages, Message):
             messages = [messages]
 
-        # set flags to know how to serialize the messages
+        # —— new: only for Gemini models, clean out any zero‐length text chunks —— 
+        model_name = getattr(self.config, "model", "").lower()
+        if "gemini" in model_name:
+            for msg in messages:
+                if hasattr(msg, "content"):
+                    for content_item in msg.content:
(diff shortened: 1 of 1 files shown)


ahmed-z0:
This solution was inspired by the approach discussed [here](https://github.com/microsoft/autogen/issues/5762#issuecomment-2745029707), from Microsoft's public autogen repository.
They also encountered similar empty text issues when using Gemini models via OpenAI-compatible endpoints, and their handling provided confident solution to this problem

ahmed-z0:
Yep, We got it within Open hands.

image

Let me try reporting it to litellm as well

PR #7981, opened 2025-04-21
Minor refactoring: remove step and task-creation in AgentController

MischaPanch:
Makes the code simpler and debuggable at no cost in performance

Maybe I am misunderstanding something, but seems like `AgentController.step` was an unnecessary and even harmful indirection. The creation of an asyncio task (instead of awaiting) means the core logic of the step was not debuggable. It drove me a bit crazy to find out why things never stopped at breakpoints. At the same time, it seems that this was pointless, since the only call of step was within `_on_event`, and the only call of that one was wrapped by `run_until_complete`, so the results of `step` were actually always "awaited" in the synchronous `AgentController.on_event`, and the indirection over the task creation had no benefits.

I may be missing something, but if I'm not, removing the indirection would improve the code. Would also reduce the public interface, since step is unused

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -293,9 +293,6 @@ class AgentController:
         # Set the agent state to ERROR after storing the reason
         await self.set_agent_state_to(AgentState.ERROR)
 
-    def step(self):
-        asyncio.create_task(self._step_with_exception_handling())
-
     async def _step_with_exception_handling(self):
         try:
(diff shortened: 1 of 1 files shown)


enyst (approved):
Thank you, I love it when we can simplify things. 🙏

MischaPanch:
@enyst was I right in my analysis? Hope I'm not breaking UI responsiveness with this 😅

Issue #7995, opened 2025-04-21
[Bug]: Recent commit broke poetry venvs path discovery


MischaPanch:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

@enyst 

With commit 300a5985 the poetry discovery command changed to `poetry env info --path`. This does not work when poetry was using an already active env to install dependencies. The current documentation suggests using mamba for managing envs, which I did, and then configured poetry to not create virtualenvs. The previous command worked in that case, but the current one fails:

```
(openhands) ➜  OpenHands git:(oraios-merged) ✗ poetry env info --path
(openhands) ➜  OpenHands git:(oraios-merged) ✗ poetry show -v
Found: /home/mpanchen/micromamba/envs/openhands/bin/python
Skipping virtualenv creation, as specified in config file.
aiohappyeyeballs                         2.5.0           Happy Eyeballs for asyncio
...
```

### OpenHands Installation

Docker command in README

### Operating System

None

enyst:
Oh! Let me check the documentation, I wonder why we'd do that... Since local runtime runs only in dev mode, with poetry accessible, the accessible poetry made sense to me. Maybe not!

And there I was, enjoying that it seemed to be a speed-up, too.

Fix upcoming.

MischaPanch:
How about just using sys.executable? It's not too much for the user to ask to activate the env before calling any openhands command. In fact, they have to do that, right? Would make the code much simpler

enyst:
How about using POETRY_VIRTUALENVS_PATH from the environment? There's a fallback in the code to use that, if it was set

MischaPanch:
Sure, but it complicates things, you'd actually want the path to the interpreter, and that doesn't actually have anything to do with poetry. It just so happens that I used poetry to install the deps into a mamba env.

Is there any reason to not use sys.executable? You'd get your speedup and then some ;)

MischaPanch:
Correction from my side: it's not exactly that you currently actively suggest to create a mamba env. But the make commands did not work for me (neither make build nor make run), and since I started off with mamba as per your docs, I just continued with it instead of letting poetry create a new env.

I'll write a PR with improved make commands at some point, haven't fully understood why they don't work. Message say everything is ok, but nothing gets installed or started...

enyst:
[...]

PR #7997, opened 2025-04-21
Fix and simplify local runtime init

enyst:

---

The first PR broke a mamba setup, reported here:
- https://github.com/All-Hands-AI/OpenHands/issues/7995

Cc: @xingyaoww 

As @MischaPanch  pointed out, POETRY_VIRTUALENVS_PATH isn't actually used by the local runtime nor Jupyter plugin itself. Do we really need it in local runtime? Removing it would clean out a subprocess that we started just for it, too, and the startup is faster!

Without it:
```
AgentFinishAction(final_thought='I have successfully added a line of dashes at the start of the `OpenHands/README.md` file. If you need any more modifications or have other tasks, feel free to let me know!', task_completed='true', outputs={}, thought='', action=<ActionType.FINISH: 'finish'>)
```

---
Fix https://github.com/All-Hands-AI/OpenHands/issues/7995

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:07e0273-nikolaik   --name openhands-app-07e0273   docker.all-hands.dev/all-hands-ai/openhands:07e0273
```

### DIFF
--- a/openhands/runtime/impl/local/local_runtime.py
+++ b/openhands/runtime/impl/local/local_runtime.py
@@ -214,10 +214,6 @@ class LocalRuntime(ActionExecutionClient):
         python_bin_path = os.path.dirname(interpreter_path)
         env_root_path = os.path.dirname(python_bin_path)
 
-        # Set POETRY_VIRTUALENVS_PATH for Jupyter plugin
-        env['POETRY_VIRTUALENVS_PATH'] = env_root_path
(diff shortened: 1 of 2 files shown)


xingyaoww (approved):
If we use poetry to launch the local_runtime, getting sys.executable should give us the same thing :)

I think this should work too-- Thanks for the simplification

Issue #7888, opened 2025-04-16
[Bug]: Line numbers are often included in `str_replace` commands


Tim-Pohlmann:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

I'm not sure this qualifies as a bug. Might be an improvement request instead.

I have noticed a common pattern that leads to a failed command execution when OpenHands tries to edit a file via `str_replace`:
1. It checks the file content via `cat -n`. The output includes line numbers (due to `-n`).
2. It calls `str_replace` but it includes the line numbers in the argument strings.
This results in the `str_replace` command not finding a match in the file and therefore not replace anything.

I'm running OpenHands via GitHub Actions, if that makes a difference.

I tried to mitigate the issue by adding some instructions in a `.openhands_instructions` file but with little to no effect.

### OpenHands Installation

GitHub resolver

### Operating System

None

enyst:
> I'm running OpenHands via GitHub Actions, if that makes a difference.

This doesn't make a difference. What LLM you are using, might matter, though, with what LLMs do you see this?

Cc: @ryanhoangt

Tim-Pohlmann:
Gemini Flash Lite

ryanhoangt:
> 1. It checks the file content via `cat -n`. The output includes line numbers (due to `-n`).
> 2. It calls `str_replace` but it includes the line numbers in the argument strings.
This results in the `str_replace` command not finding a match in the file and therefore not replace anything.

This seems to me like a limitation in the LLM's tool use 🤔 ideally, after using `cat -n`, it should recognize that line numbers are just part of the display and shouldn't be included when passing file content to tools.

enyst:
> I tried to mitigate the issue by adding some instructions in a `.openhands_instructions` file but with little to no effect.

I'm sorry, I see we haven't updated the resolver README.md on this, but we have deprecated the `.openhands_instructions` way of customizing instructions, in favor of a `repo.md` file in a particular location like this one is:
https://github.com/All-Hands-AI/OpenHands/blob/main/.openhands/microagents/repo.md

[repo.md](https://github.com/All-Hands-AI/OpenHands/tree/main/microagents#2-repository-instructions-private)

Could you please try that?

@openhands-agent 
Can you update the resolver README.md to point to the microagents' information about repo.md, instead of .openhands_instructions? The relevant link is the last I linked in this comment.

[...]

PR #8010, opened 2025-04-22
Refactor SWE-bench instruction

enyst:

This PR proposes to refactor how we choose the SWE-bench instruction: pick it up from a template file, per `mode` (SWT, SWE bench) and per model (claude, gemini, GPT-4.1 have their own, and a default that is the current default instruction)

IMHO this will enable us to adjust prompts when we test other LLMs more easily.

The current behavior with Claude Sonnet is unchanged.

Split out from #7748  #7895 

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:2559ccf-nikolaik   --name openhands-app-2559ccf   docker.all-hands.dev/all-hands-ai/openhands:2559ccf
```


mamoodi:
- submitted instances: 5
- empty patch instances: 0
- resolved instances: 5
- unresolved instances: 0
- error instances: 0 - Download at: https://github.com/All-Hands-AI/pr-eval-results/releases/download/1.0.0/swe-bench-instruction-549_25-05-09-15-25.tar.gz

xingyaoww (commented):
Overall LGTM! Quick question: where do we get the instruction for gpt4 and gemini -- there official source?

enyst:
@xingyaoww 

GPT-4.1 is indeed from OpenAI, from here: [4.1 swe-bench instruction](https://cookbook.openai.com/examples/gpt4-1_prompting_guide#sample-prompt-swe-bench-verified).
(should we name the file “GPT41” instead of “GPT4”?)

Gemini is a prompt I came up with, in part inspired by Google cookbooks, and tested here:
- https://github.com/All-Hands-AI/OpenHands/pull/7748

That PR is a bit old, but not forgotten! It did seem to help! It showed some ~5% difference on some version of Gemini (I think the first 2.5). More importantly, I saw why: with the default instruction, Gemini skipped tests a lot, and with this instruction, it doesn’t. That made it pass instances it was failing before. This prompt rewrites things in a more Gemini style, avoids to say that “I took care of tests”, and, to the contrary, it tries to push it a bit, to not get too lazy.
Gemini _is_ lazier than Sonnet. 😅

IMHO we could take in Gemini too, and iterate on it. As soon as I have some time to focus on it, I’d like to try a  Gemini 50-instances run, with tools to which we add information that Gemini gets tripped up all the time (like [this microagent](https://github.com/All-Hands-AI/OpenHands/issues/7888#issuecomment-2956592635) that I’m using all the time), and probably a bit different system prompt.

xingyaoww (approved):
Thanks engel! Let's get this in then!

PR #8017, opened 2025-04-22
feat: Add CLI support for /new, /status and /settings

pandukamuditha:

Adds support for /new, /status, and /settings in the CLI
- /new - Closes the current session and creates a new one
- /status - Prints the current session status and stats
- /settings - Display and allow basic and advanced settings to be configured

---

- Integrated SettingsStore (FileSettingsStore) into CLI for persisting session settings
- Integrated a custom lexer to display diff highlights for file edits
- Refactored cli.py into smaller units
- Refactored/moved session settings from: `openhands/server/settings.py` to: `/openhands/storage/data_models/settings.py`
- Refactored/moved list LLM models functionality from: `openhands/server/routes/public.py` to `openhands/utils/llm.py`
- Added dependency check on tmux
- Resolved an issue with detecting poetry env

---


enyst (changes requested):
Thank you for doing this, it looks like there are great improvements here!

I have some doubts about reproducing the behavior of the web frontend on settings though, which seems like what this PR is doing. In the same time, at a glance it has improvements too, moving some core stuff out of `server` is better!

However, a CLI user has direct access to the configuration of the application via `config.toml` file. I'm not sure we should make them use a `json` file too. Json is not a good format for configuration files, IMHO, it's not friendly to normal manual modification in a random editor. If the user modifies some via the cli app, and some manually, it's not clear to me that they even can get the result they want.

I'd appreciate to think it over a bit. Our settings are a bit messy, and it's not due to the PR, sorry, it just is. I think we need to think how to make them better - or move towards them becoming better - though.

openhands/core/cli_settings.py
```diff
@@ -0,0 +1,354 @@
...
+        provider = getattr(
+            llm_config,
+            'provider',
+            llm_config.model.split('/')[0] if '/' in llm_config.model else 'Unknown',
+        )
+        labels_and_values.extend(
+            [
+                ('   LLM Provider', str(provider)),
+                ('   LLM Model', str(llm_config.model)),
+                ('   API Key', '********' if llm_config.api_key else 'Not Set'),
```
enyst, line 49: The `labels_and_values` here are those from frontend, but they come from `config: AppConfig`, not from the (moved) settings store? Maybe I'm missing some new logic, sorry, it's a large PR 😅 - could you please explain the intention here. Maybe we can get the values from everywhere the user may have set them, in order of priority?

openhands/core/cli_settings.py
```diff
@@ -0,0 +1,354 @@
...
+        settings = Settings()
+
+    settings.llm_model = provider + organized_models[provider]['separator'] + model
+    settings.llm_api_key = SecretStr(api_key)
+    settings.llm_base_url = None
+    settings.agent = OH_DEFAULT_AGENT
+    settings.confirmation_mode = False
+    settings.enable_default_condenser = True
+
+    await settings_store.store(settings)
```
enyst, line 238: It looks like this function loads the user configuration from `AppConfig`, but then stores it in `settings_store`. What will happen next time this function is called, will it still load from `config`? Or do we override somewhere?

openhands/core/cli_utils.py
```diff
@@ -0,0 +1,165 @@
...
+
+from openhands.core.cli_tui import (
+    UsageMetrics,
+)
+from openhands.events.event import Event
+from openhands.llm.metrics import Metrics
+
+
+def manage_openhands_file(folder_path=None, add_to_trusted=False):
+    openhands_file = Path.home() / '.openhands.toml'
```
enyst, line 13: Could you explain what do you intend  `.openhands.toml` to be? As of now, we don't have this configuration file, we have a `config.toml`, and it could be in the project directory or anywhere else (path is configurable I think)

I tend to agree that maybe the non-default file should be in the user's home TBH

Makefile
```diff
@@ -101,6 +102,15 @@ check-docker:
...
 
+check-tmux:
+	@echo "$(YELLOW)Checking tmux installation...$(RESET)"
+	@if command -v tmux > /dev/null; then \
+		echo "$(BLUE)$(shell tmux -V) is already installed.$(RESET)"; \
+	else \
+		echo "$(RED)tmux is not installed. Please install tmux to continue.$(RESET)"; \
+		exit 1; \
+	fi
+
```
enyst, line 113: Minor thing, but this is a large PR : we could maybe split the `tmux` issue in another PR?

enyst:
The display and behavior of the `cli` are great! I'm testing it a bit. 😅 

I see a couple of unexpected things on this branch, though I'm not sure if they come from this PR.

I started the cli, from the vscode debugger as usual, with predefined args, among which:
```
      "args": [
        "-t",
        "Write a file about the eight queens solution",
```
In config.toml I have set `local` runtime.

```
Do you trust the files in this folder?                                                                                                                                                                                                                                  │
│   /Users/enyst/repos/odie/workspace                                                                                                   
                                                                                                                               
│ OpenHands may read and execute files in this folder with your permission.                                                             

Do you wish to continue?
...
OpenHands CLI v0.33.0
Initialized session bc7e1649-e765-471d-90f3-44a0b17e9b3e
Let's start building!
What do you want to build? Type /help for help
```

At this point, it took quite a long time of waiting. The thing is, I started it with a task (`-t`), so I expected to display my task, and run with it. It didn't display it; it did, after some time of waiting, execute it.

I don't know yet if the non-display is from this PR or before.
(and I'm not sure it's important, but I think it was a request from our users, and if we have the arg, the behavior should be consistent IMHO)

Later, a strange thing happened when it couldn't write on the filesystem, as it thinks it's "/workspace/" for some reason.

```
I'll create a file about the eight queens solution in the playground repository. I'll use the `str_replace_editor` to create a new file with an explanation and a Python implementation.
┌─────────────────────────────────────────────────────────────| File Edit |─────────────────────────────────────────────────────────────┐
│(no changes detected. Please make sure your edits change the content of the existing file.)                                            │
│                                                                                                                                       │
└───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘
test

I apologize for the error. It seems the `/workspace/playground` directory might not exist. Let me first create the directory and then create the file.

┌───────────────────────────────────────────────────────────| Action Output |───────────────────────────────────────────────────────────┐
│mkdir: /workspace: Read-only file system
```

This also might not be this PR, I'm not sure yet. It's possible, even likely, that we hardcode "/workspace" somewhere. FWIW I run on `local runtime`, usually, including earlier today from `main.py`, successfully, and never seen it look for "/workspace". 🤔

openhands/server/settings.py
```diff
@@ -1,104 +1,6 @@
...
-            max_iterations=app_config.max_iterations,
-            security_analyzer=security.security_analyzer,
-            confirmation_mode=security.confirmation_mode,
-            llm_model=llm_config.model,
-            llm_api_key=llm_config.api_key,
-            llm_base_url=llm_config.base_url,
-            remote_runtime_resource_factor=app_config.sandbox.remote_runtime_resource_factor,
-        )
-        return settings
+from openhands.storage.data_models.settings import Settings
```
enyst, line 3: Totally trivial, maybe we could also split into another PR the moves to `storage` and to `utils.llm`? It could make it clear when we also change them in this PR, if at all. And allow us to test them separately.

openhands/core/cli_tui.py
```diff
@@ -0,0 +1,566 @@
+# CLI TUI input and output functions
+# Handles all input and output to the console
+# CLI Settings are handled seperately in cli_settings.py
+
+import asyncio
+import sys
+import time
+from typing import List, Optional
```
enyst, line 8: Nit: we use python 3.12+, and, while we haven't been always consistent on this, we prefer the use of builtins like `list`, `| None` in the place of the older typing classes

openhands/core/cli_tui.py
```diff
@@ -0,0 +1,566 @@
...
+
+
+class UsageMetrics:
+    def __init__(self):
+        self.total_cost: float = 0.00
+        self.total_input_tokens: int = 0
+        self.total_output_tokens: int = 0
+        self.total_cache_read: int = 0
+        self.total_cache_write: int = 0
+        self.session_init_time: float = time.time()
```
enyst, line 69: We have the `Metrics` class storing this kind of metrics. Is there a reason why we can't just use it?

openhands/core/cli_utils.py
```diff
@@ -0,0 +1,165 @@
...
+    if not token_usage:
+        return
+
+    prompt_tokens = getattr(token_usage, 'prompt_tokens', 0)
+    completion_tokens = getattr(token_usage, 'completion_tokens', 0)
+    cache_read = getattr(token_usage, 'cache_read_tokens', 0)
+    cache_write = getattr(token_usage, 'cache_write_tokens', 0)
+
+    usage_metrics.total_input_tokens += (
+        prompt_tokens if isinstance(prompt_tokens, int) else 0
```
enyst, line 64: Please tell if I don't see this right, this adds input tokens found in `accumulated_tokens_usage` object? I _think_ those are already the sum of previous prompts. Do we have a test to verify that counting them here doesn't double-count?

openhands/core/cli_utils.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+    usage_metrics.total_output_tokens += (
+        completion_tokens if isinstance(completion_tokens, int) else 0
+    )
+    usage_metrics.total_cache_read += cache_read if isinstance(cache_read, int) else 0
+    usage_metrics.total_cache_write += (
+        cache_write if isinstance(cache_write, int) else 0
```
enyst, line 71: Nit: In this method, I'm not sure we need all these `isinstance`, could we do it differently or use the original object?

openhands/core/cli_commands.py
```diff
@@ -0,0 +1,293 @@
...
+        )
+
+    if new_session_requested:
+        close_repl = True
+        new_session_requested = True
+        event_stream.add_event(
+            ChangeAgentStateAction(AgentState.STOPPED),
+            EventSource.ENVIRONMENT,
+        )
+        display_shutdown_message(usage_metrics, sid)
```
enyst, line 201: Could we not stop the session for LLM settings? Normally, this isn't necessary, an existing event stream can be loaded with another LLM or continue with another LLM just fine. So switching LLM can work with the same session. Unless there's some reason in this PR why it's difficult?

Afaik CLI doesn't load previous sessions, unless that changed? So if we force a stop here it implies that you can never switch LLM, not even by restarting the app.

openhands/core/cli.py
```diff
@@ -40,633 +45,64 @@
...
+            task.cancel()
+
+        # Wait for all tasks to complete with a timeout
+        if pending:
+            await asyncio.wait(pending, timeout=5.0)
+
+        # Reset agent, close runtime and controller
+        agent.reset()
+        runtime.close()
+        await controller.close()
```
enyst, line 86: This looks right to me, I do have a couple small questions though:
- `agent.reset` also clears out the LLM metrics. Do we really need or want to do that? They could be useful and I think today we don't [save trajectory](https://github.com/All-Hands-AI/OpenHands/blob/5d749aeba795d026cc1eb14339955752ebd35346/openhands/core/main.py#L218) from cli, but if we want to save them, we may want metrics in. Or display them to the user, idk.
- just out of curiosity, does this work well without needing to call [this close](https://github.com/All-Hands-AI/OpenHands/blob/5d749aeba795d026cc1eb14339955752ebd35346/openhands/events/stream.py#L79) ?

bashwara:
> I'd appreciate to think it over a bit. Our settings are a bit messy, and it's not due to the PR, sorry, it just is. I think we need to think how to make them better - or move towards them becoming better - though.

Thanks @enyst for the feedback. I agree that the settings approach has room to improve and to unify in terms of approach. Let us take a look at options around this post this initial iteration. Will consult you for feedback.

openhands/core/cli_settings.py
bashwara: `AppConfig` is populated in the order of priority at CLI (cli.py) startup with the following, in the order of increasing precedence. `config.toml`, `SettingsStore`, CLI args. Hope this clarifies.

openhands/core/cli_settings.py
bashwara: Same as the above.

> config: AppConfig is populated in the order of priority at CLI (cli.py) startup with the following, in the order of increasing precedence. config.toml, SettingsStore, CLI args. Hope this clarifies.

On subsequent runs after the settings are set, `SettingsStore` will take priority if there are no CLI args.

Makefile
bashwara: Agree. We'll raise a separate PR for it.

bashwara:
> At this point, it took quite a long time of waiting. The thing is, I started it with a task (`-t`), so I expected to display my task, and run with it. It didn't display it; it did, after some time of waiting, execute it.
> 
> I don't know yet if the non-display is from this PR or before. (and I'm not sure it's important, but I think it was a request from our users, and if we have the arg, the behavior should be consistent IMHO)

Yeah we noticed this as well. We'll push an update for this.

> Later, a strange thing happened when it couldn't write on the filesystem, as it thinks it's "/workspace/" for some reason.

> This also might not be this PR, I'm not sure yet. It's possible, even likely, that we hardcode "/workspace" somewhere. FWIW I run on `local runtime`, usually, including earlier today from `main.py`, successfully, and never seen it look for "/workspace". 🤔

On the issue of the `/workspace`. This appears to be a bit non-deterministic. Something we'll continue to investigate.

openhands/server/settings.py
bashwara: The reason why we did it this way to do the related refactoring in one go. But I agree that it could've been broken up. Let us know if you have a major concern in moving forward with this as it would require a few more cycles to finish up otherwise.

openhands/core/cli_tui.py
bashwara: Noted. Will push an update.

openhands/core/cli_utils.py
enyst: I think the goal you pointed out is absolutely worthwhile, and thank you for the detail. I got to read the whole PR in the mean time, sorry for the initial post on the run.

I noticed locally that it saved trusted directories.

Trusted directories seems like a sandbox-related setting, perhaps a local runtime specific one. What do you think about:
- define it in sandbox_config
- load it normally as part of `load_app_config` from the default toml, if present (`config.toml`)
- then override it from the user-specific toml, if any (`~/.openhands.toml`) ?

openhands/core/cli_settings.py
enyst: This sounds just right, thank you. I will mark this as solved, appreciate the response, please don't mind if I may come back on this kind of issue later though. I do think it's super-important to get this right, and unfortunately we dropped the ball sometimes. 😅

openhands/core/cli_settings.py
enyst: Thank you. Do we have unit tests for the order of precedence?

openhands/core/cli_tui.py
bashwara: We're also storing `session_init_time` to calculate the current session duration in the `UsageMetrics` class. So at the time defining our own class with just the attributes we wanted felt fitting. But I agree that we could certainly reuse the original `Metrics` class rather than defining the same attributes again. Will add a change for it.

openhands/core/cli_utils.py
bashwara: Initially when we were doing the implementation for the usage metrics calculation, token usage was calculated on a per request. We now noticed that implementation has been updated to track the accumulated token usage. Will change this implementation accordingly. Thanks for pointing that out.

openhands/core/cli_utils.py
bashwara: Technically no, and I agree that there maybe some redundancy here I was just being extra careful as I ran into some cases of missing fields. Will remove this.

openhands/core/cli_commands.py
bashwara: Right now we focused sticking to the existing approaches and patterns so that we can focus on the TUI aspects as a first step. This maybe something that can be done as a further enhancement.

openhands/core/cli.py
bashwara: 1. We do display the metrics before this step. Maybe the trajectory functionality can be worked on as a further enhancement. Would like to learn more about how this functionality works a well.
2. A quick cursory test shows that it works fine even without this. But may need a bit more testing to confirm.

openhands/core/cli_commands.py
enyst: I'm with you on existing approaches and patterns, but maybe I don't fully see how it applies here. `openhands` _just works_ if I switch the LLM mid-session or rather, with headless, next reload of a session. It looks to me like that is normal behavior, while this adds a constraint that didn't exist before.

Or perhaps I see what you're saying, but it's because the CLI use case lacks it, simply:
- when run via `main.py`, you can reload a session with another LLM just fine
- when run via `agent_session`, you can reload a session and switch LLM
- when run via `cli.py`, I think we don't have the ability to reload sessions at all, and switching LLMs is something this PR introduces. So the PR adds it without the normal behavior present in other use cases.

rbren: @enyst I think if you change your settings mid-session, the current session continues to use the old settings?

enyst: You may be right! I was thinking about that, I'm not sure of the current UI behavior.

That's okay though, if you think it breaks expectations maybe a single log line could tell the user that hey, saved, but those model settings will only apply when you restart? To keep it simple for now?

The current implementation seems to force the user to only change LLM settings when they start a session, or remember to change before they end one. That seems a bit unfriendly to me.

enyst: Please note that with the UI, there was a timeframe when the user was unable to change settings only because of a bit of FE code that was forcing them into a new session. However, if you refreshed the old conversation, it was loading it with the new settings.

I would love it if we didn't repeat that behavior in cli. 🤣

Makefile
bashwara: @enyst A separate PR has been raised for this - https://github.com/All-Hands-AI/OpenHands/pull/8043

openhands/core/cli_tui.py
bashwara: @enyst this has been addressed

openhands/server/settings.py
enyst: I took a stab at comparing the actual changes on these moves, and it seems they're equivalent. So they're really just moving things. I am just trying to understand what changed. 😅 

Gotcha, no worries, thank you. I can deal with it.

openhands/core/cli_tui.py
bashwara: @enyst Refactored the current implementation to leverage the original Metrics class.

openhands/core/cli_utils.py
bashwara: @enyst this has been addressed.

openhands/core/cli_utils.py
pandukamuditha: The main idea for having a separate config file: `.openhands.toml` for storing trusted folders is that it's intended to be written by the CLI and used as a store, rather than being directly modified by users. This differs from `config.toml`, which is primarily for manual, user-configured settings.

For this reason, we thought it would be more appropriate to not include the trusted folders as part of `config.toml`.

openhands/core/cli_settings.py
bashwara: Yeah, a unit test has been added to cover it.

openhands/core/cli_utils.py
enyst: I understand it may feel like the users don't touch it, but I think such assumption is leading us the wrong route.

Let's assume that it is as available for people as the other is - because it is. It's a file on the local machine, and as noted above, the `config.toml` file has a configurable path, so users could even decide it's in `~` anyway. 🤔

The way I think of this issue, is that the configuration files we have should, eventually, become 2 files:
- a default file, let's call it system configuration file (`./config.toml`)
- a user-specific file, such as perhaps `~/.openhands.toml` or `~/.openhands/config.toml`

With the note:
- _The user-specific options are overriding the default ones._

I think you are absolutely correct that moving forward we should aim to unify CLI and FE settings. I think we can do that by aiming to get the `.json` merged into the user-specific `toml`.

We can build the user-specific file options step by step. We currently know a dozen or so. (those in json, plus this)
Most exist in the main configuration file.

We do need to be consistent though. I think we should not add another configuration file with options inconsistent with the main configuration, or we'll never unify them. 😅

I'd suggest to think of it as, the user-specific file is an override of the general configuration. So that the precedence becomes (as it is already the case in some measure, or rather, should be the case, with the `.json`):
BaseModels defaults < `config.toml` < `~/.openhands.toml` / `settings.json` < env vars < args

Are there major issues with this, what do you think? It would be great to have a clear view of the goal here!

openhands/core/cli.py
enyst: Thank you! I think you're right.

The trajectories can be enabled in `config.toml`, and work in headless mode (via `main.py`).

I think they are cool to see on the [trajectory visualizer](https://github.com/All-Hands-AI/trajectory-visualizer?tab=readme-ov-file)!

My questions were answered, thank you, I'll mark this as solved to keep the github view readable.

openhands/core/cli_commands.py
rbren: Yeah agreed I think we can just not kill the session after changing settings

openhands/core/cli_utils.py
rbren: What's the simplest path forward here? Should we just overwrite config.toml?

I think all of these could just go into settings.json as well...

enyst: Wait, what do you mean by all of these, only what's in the new `.openhands.toml`? (sorry, it's late)

Yes, we could overwrite `config.toml` in this PR, the vars can go in sandbox section.

bashwara: Based on this, I believe the best way forward is that we'll add the trusted directory setting to the `config.toml` which can be overridden by the `.openhands.toml` in user's home directory.

Are you good with this approach? @enyst @rbren

enyst: I think we could even, as Robert said, just write the directories in `config.toml` directly. Or am I missing something?

This new file, could you please confirm, is it only for directories for now?

If you think the simplest for the moment is to continue using this new configuration file -  then let's use it, and next time we see where we're going with this.

openhands/core/cli_commands.py
pandukamuditha: Sounds good! We agree that immediately terminating the session after a settings change could create a less-than-ideal user experience.
As suggested by @enyst this can be modified so that saving settings will no longer automatically close the current session. Instead, the new settings will be applied when the user restarts the CLI. We can print a message to indicate this to the user.

openhands/core/cli_utils.py
enyst: I took another look at this. I agree with you @bashwara , the simplest way forward is to add trusted directory to config.toml and read it too; but for writing, override it by writing to `~/.openhands/config.toml`.

Lets use `.openhands/config.toml` please, it should be the same thing for the PR right now.

openhands/core/cli_settings.py
```diff
@@ -134,9 +134,19 @@ async def get_validated_input(
     return value
 
 
+def save_settings_confirmation() -> bool:
+    return (
+        cli_confirm(
+            '\nSave new settings? (Restart CLI to take effect)',
```
enyst, line 140:
```suggestion
            '\nSave new settings? (They will take effect after restart)',
```

openhands/core/cli_utils.py
bashwara: > This new file, could you please confirm, is it only for directories for now?

Yes, we're only using  `~/.openhands.toml` for the purpose of keeping track of the user trusted directories for the CLI at the moment.

> I took another look at this. I agree with you @bashwara , the simplest way forward is to add trusted directory to config.toml and read it too; but for writing, override it by writing to `~/.openhands/config.toml`.
> 
> Lets use `.openhands/config.toml` please, it should be the same thing for the PR right now.

We're planning to open another PR early next week to add commands: `/pause` & `/resume` to complete the new CLI experience. Shall we discuss further and implement these suggestions along with that?

openhands/core/cli_commands.py
pandukamuditha: @enyst, this new logic has been implemented. Pls have a look

enyst: Thank you!

openhands/core/cli_utils.py
pandukamuditha: @enyst This has been a very productive conversation, and I think we have a clear path forward.

Here's what we landed on for the CLI trusted directories configuration:

* The trusted directories configuration for the CLI needs to be added to the `config.toml` file.
    * Will be added as `xx` to the `xx` section (we should confirm the exact names).
* Users will be able to use the config in `config.toml` as their initial list of trusted directories they set up manually.
* When the CLI starts up pointed at a directory not currently in the trusted list in `config.toml`, and the user confirms they trust it, that new directory should be automatically added to their personal configuration file: `~/.openhands/config.toml`, using the same section and key.
* Configurations in `~/.openhands/config.toml` override the configurations in `config.toml`.

If the above sounds right, we can go ahead and add these changes.

enyst: Thank you @pandukamuditha ! Yes, let's do it that way.

pandukamuditha: This change is done @enyst. Pls have a look

enyst (approved):
Thank you!

This is an awesome PR, I love to see CLI getting really nice going forward.

I'll note FWIW that I kept seeing the agent trying "/workspace" and failing. Curious if maybe you found anything related to that?

PR #8019, opened 2025-04-22
Simplify local runtime a bit

enyst:

---

This PR proposes to remove the initialization of the browser that we shutdown immediately after. It's there to check it works... but it is time-consuming, and maybe the browser should actually be optional?

This will speed up a bit.

I'm not sure what to replace it with, if we want the initial check replaced, could give it some thought...

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:07c376d-nikolaik   --name openhands-app-07c376d   docker.all-hands.dev/all-hands-ai/openhands:07c376d
```


openhands/runtime/impl/local/local_runtime.py
```diff
@@ -76,13 +76,6 @@ def check_dependencies(code_repo_path: str, poetry_venvs_path: str):
     if 'test' not in pane_output:
         raise ValueError('libtmux is not properly installed. ' + ERROR_MESSAGE)
 
-    # Check browser works
-    logger.debug('Checking dependencies: browser')
```
rbren, line 80: we don't need to check the browser?

enyst: We already check in `make build` for playwright and chromium. I don't know if another check helps, but this is _every run_, as opposed to one and done. 🤔

I can look more into it. Frankly on my machine I removed it.

Actually, I guess we could at least save this result and leave the user alone for subsequent runs. Even so, if something changes in their local environment, it will fail later when it's actually needed. But is that unexpected?

enyst:
I think we may be doing too much by checking every run, but we could go about this differently

PR #8030, opened 2025-04-23
RFC: Runtime build system refactor

xingyaoww:

---

---

### DIFF
--- a/openhands/runtime/utils/wrappers/oh-chromium
+++ b/openhands/runtime/utils/wrappers/oh-chromium
@@ -1,0 +1,17 @@
+#!/bin/bash
+# /openhands/bin/oh-chromium
+
+# Set up environment for Chromium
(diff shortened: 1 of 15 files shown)


openhands/runtime/utils/wrappers/oh-chromium
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+# /openhands/bin/oh-chromium
```
raymyers, line 2: I'm not seeing oh_chromium called anywhere, isn't it just run directly by playwright?

xingyaoww: Yeah - you can probably ignore the detailed implementation now :)
Right now this PR should just be a rough idea / design doc, where we might use different things for actual implementation.

Maybe we could using things like this: https://pypi.org/project/poetry-pyinstaller-plugin/

PR #8040, opened 2025-04-23
Refactor Authentication

tofarr:

Upgrade to rely more heavily on lazy loading, caching in the request, and dependency injection.

---
The existing implementation based on middleware lacks the flexibility we need - it loads Provider tokens even for cases where they are not required. This implementation uses FastAPIs dependency injection in conjunction with our get_impl method to provide code that is cleaner and more flexible.

**Changes in latest commits:**
- Fixed lint issues:
  - Removed unused import ConversationStore from user_auth/__init__.py
  - Removed unused import SecretStore from user_auth/default_user_auth.py
  - Fixed type error in utils.py by changing return type to ConversationStore | None and fixing control flow
  - Fixed formatting issues in test files
- Fixed failing unit tests:
  - Updated settings.py route handlers to use Depends(get_user_id) instead of direct function calls
  - Modified helper functions to accept user_id parameter
  - Made functions properly handle async user_auth functions
  - Updated tests to work with the new async architecture

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:80069b5-nikolaik   --name openhands-app-80069b5   docker.all-hands.dev/all-hands-ai/openhands:80069b5
```


openhands/server/listen.py
```diff
@@ -32,6 +31,5 @@
     rate_limiter=InMemoryRateLimiter(requests=10, seconds=1),
 )
 base_app.middleware('http')(AttachConversationMiddleware(base_app))
-base_app.middleware('http')(ProviderTokenMiddleware(base_app))
```
tofarr, line 35: Removing this because we are going with a more lazy loading / caching implementation!

enyst:
Let's give it a test on local. It shouldn't change behavior, but those github tokens have some bad habits. 😅

PR #8043, opened 2025-04-23
fix: Add error handling for scenarios where tmux is unavailable

pandukamuditha:

Fixes an issue where the application, when running using the local runtime, would throw a generic exception if `tmux` is not available

---

- Add exception handling and logging to indicate `tmux` unavailability
- Add a step in Makefile to check `tmux` availability

---


enyst (approved):
Thank you!

PR #7955, opened 2025-04-20
feat: Use effective token in resolver workflow steps

hamiltop:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [x] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality that this introduces.**

Support Github App Tokens in the Github resolver workflow.

---
**Give a summary of what the PR does, explaining any non-trivial design decisions.**

This PR adds support for authenticating using a GitHub App within the reusable `openhands-resolver.yml` workflow.

Changes:
- Added `app_id` and `app_private_key` inputs to the workflow's secrets.
- Added a step using `actions/create-github-app-token@v1` to generate an installation token when App credentials are provided via secrets.
- Implemented a token selection step (`Determine Auth Token`) to prioritize authentication methods: GitHub App Token > Personal Access Token (PAT) > Default `GITHUB_TOKEN`.
- Updated subsequent workflow steps (e.g., setting environment variables, commenting on issues, creating PRs) to use the determined `EFFECTIVE_TOKEN`.
- Modified the "Check required environment variables" step to reflect the new token priority and provide appropriate warnings.

This allows users of the workflow to authenticate with higher permissions using a GitHub App instead of relying solely on PATs or the default token.

---
**Link of any specific issues this addresses.**

https://github.com/All-Hands-AI/OpenHands/issues/7860

PR #8048, opened 2025-04-23
Use generic token check in order to support Github Apps

tdeckers:

Github Apps don't support the /user endpoint.  As such, using it to verify a token will always fail when authenticating as a github app.

---
The code is now using the base URL to verify if a token works.  That URL works for both regular users, and Github Apps.

Curl example of why this doesn't work:
```bash
$  curl -H 'Authorization: Bearer <redacted github app token>' https://my.enterprise.ghe.com/api/v3/user
{
  "message": "Resource not accessible by integration",
  "documentation_url": "https://docs.github.com/enterprise-server@3.14/rest/users/users#get-the-authenticated-user"
}
```
---


openhands/integrations/github/github_service.py
```diff
@@ -118,6 +118,17 @@ async def get_user(self) -> User:
...
+    async def verify_access(self) -> bool:
+        """Verify if the token is valid by making a simple request."""
+        url = f'{self.BASE_URL}'
+        try:
+            await self._make_request(url)
+            return True
+        except httpx.HTTPStatusError as e:
+            if e.response.status_code == 401:
+                return False
+            raise
```
malhotra5, line 131:
```suggestion
        except Exception as e:
            return False
```

`_make_request` will log the error reason now, so we can simply return False and exit

enyst:
I'm a little confused how does this PR work with this PR:
- https://github.com/All-Hands-AI/OpenHands/pull/7955

Cc: @hamiltop What do you think?

hamiltop:
This is great. I like the simplicity.

malhotra5:
@enyst the main contribution for #7955 is that it generates a new token for every resolver run using github app credentials (it also chooses which token to pass to the resolver based on a priority defined in the workflow)

This PR is verifying that the token passed to the resolver indeed works (regardless of what type of token is passed)

tawago:
@tdeckers Can we make sure that this works with [`GITHUB_TOKEN`](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#about-the-github_token-secret)? just concerned about this conversation in the slack channel https://openhands-ai.slack.com/archives/C07V9BMGHH6/p1745639056725169

tdeckers:
I don't have access to that Slack channel, but yes, that's no problem.  This PR just changes the target endpoint to be more generic.  No changes to the auth.  Just did a quick test with curl.  Both auth mechanisms still work:
```bash
$ curl -I -H "Authorization: token $GITHUB_TOKEN" https://my.enterprise.ghe.com/api/v3
HTTP/2 200
...
```

and 

```bash
$ curl -I -H "Authorization: Bearer $GITHUB_TOKEN" https://my.enterprise.ghe.com/api/v3
HTTP/2 200
...
```

GITHUB_TOKEN is a PAT token in this example.

enyst:
The channel is public, you are most welcome to join it if you'd like.
image

tdeckers:
Thanks for the screenshot... If I get the context correct, what's discussed there is exactly what this PR was trying to fix.

enyst:
@malhotra5 What do you think, we could merge this PR?

malhotra5:
Yes, we're good to go! Thanks everyone will merge now

openhands/integrations/github/github_service.py
```diff
@@ -121,6 +121,15 @@ async def get_user(self) -> User:
...
         )
 
+    async def verify_access(self) -> bool:
+        """Verify if the token is valid by making a simple request."""
+        url = f'{self.BASE_URL}'
+        try:
+            await self._make_request(url)
+            return True
+        except Exception:
+            return False
```
malhotra5, line 131: We should just raise the exception instead of catching it here. In utils where this method is called we catch the exceptions, and we don't check whether the return value from here is `True` vs `False`

With the current changes  `validate_provider_token ` will always return `ProviderType.GITHUB`

malhotra5, line 131:
```suggestion
        url = f'{self.BASE_URL}'
        await self._make_request(url)
```

malhotra5:
@tdeckers could you look into the lint issues, after which we can merge immediately

PR #8053, opened 2025-04-24
(Hotfix): Issue comments on Cloud Resolver

malhotra5:

- Fixes issue comments not starting with unexpected error on cloud resolver

---

This PR changes the instruction template so that it isn't using curly braces

This is because we inject information into the template later with a format method in python

https://github.com/All-Hands-AI/OpenHands/blob/78d82408b77184eb03959cfc5b98ea5444a7de52/openhands/server/routes/manage_conversations.py#L134

The format method was throwing an error when it came across `{branch-name}` and was unable to find the `{}` it actually needed. 

---


neubig (approved):

malhotra5:
Manually tested and confirmed this is indeed the fix

PR #8058, opened 2025-04-24
fix: dev.sh failed due to outdated configuration of Dockerfile

LENSHOOD:
---
**Descriptions:**
The [`dev.sh`](https://github.com/All-Hands-AI/OpenHands/blob/main/containers/dev/dev.sh) runs failed with two errors:
1. The rule `check-nodejs`:
https://github.com/All-Hands-AI/OpenHands/blob/356cd9ff9f0f5696483e29f12dac577f0441e986/Makefile#L80-L89 in `Makefile` requires node version greater than 22(Line 85), however, the dev `Dockerfile` installs only node version 18 in it:
https://github.com/All-Hands-AI/OpenHands/blob/356cd9ff9f0f5696483e29f12dac577f0441e986/containers/dev/Dockerfile#L65

2. The `RUN` command at last of the dev `Dockerfile` mounts the directory using type: `bind`: https://github.com/All-Hands-AI/OpenHands/blob/356cd9ff9f0f5696483e29f12dac577f0441e986/containers/dev/Dockerfile#L109-L121
But the bind type by default [binding directory read only](https://docs.docker.com/reference/dockerfile/#run---mounttypebind). And the `rw` option should be added(at Line 111), otherwise, the rule `install-python-dependencies` would throw an error at here because of read only file system:
https://github.com/All-Hands-AI/OpenHands/blob/356cd9ff9f0f5696483e29f12dac577f0441e986/Makefile#L161-L162

---
N/A
---
1. update version of node installation script to node 22.x
2. docker mount type "bind" is by default read only, the rw permission needs to be declared

---


enyst (approved):
Thank you!

PR #8063, opened 2025-04-24
Make tmux optional for development

amanape:
This PR makes the following changes to the Makefile:

1. Moves the check-tmux script to be the last dependency check in the check-dependencies target
2. Modifies the check-tmux script to display a warning message in yellow color instead of failing if tmux is not installed
3. Uses a box-style warning message to clearly indicate that tmux is optional
4. Fixes Makefile indentation by replacing spaces with tabs

These changes ensure that:
- The check-tmux script is the last thing that shows up during dependency checks
- It displays as a warning in a different color (yellow) to stand out
- It clearly indicates that tmux is optional and users can install it if they want to
- The build process won't fail if tmux isn't installed

Since tmux is not necessary for development, this change improves the user experience by not blocking the build process if tmux is not installed.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:3ead69c-nikolaik   --name openhands-app-3ead69c   docker.all-hands.dev/all-hands-ai/openhands:3ead69c
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,130 +45,130 @@ endif
 check-system:
 	@echo "$(YELLOW)Checking system...$(RESET)"
 	@if [ "$(shell uname)" = "Darwin" ]; then \
-	        echo "$(BLUE)macOS detected.$(RESET)"; \
+		echo "$(BLUE)macOS detected.$(RESET)"; \
 	elif [ "$(shell uname)" = "Linux" ]; then \
-	        if [ -f "/etc/manjaro-release" ]; then \
-	                echo "$(BLUE)Manjaro Linux detected.$(RESET)"; \
-	        else \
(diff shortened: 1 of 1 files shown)


amanape:
It looks like OpenHands replaced space characters with tab; this was the command it ran: `cd /workspace/OpenHands && sed -i 's/^ /\t/g' Makefile``

enyst:
I think the first thing on a line is a tab; the next are spaces

enyst (approved):

PR #8068, opened 2025-04-24
feat: secrets manager settings

amanape:

- Save custom secrets for agent through UI

---
Allow the user to manage secrets through the settings.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:385cd7e-nikolaik   --name openhands-app-385cd7e   docker.all-hands.dev/all-hands-ai/openhands:385cd7e
```


amanape:
I have resolved the merge conflicts by merging the latest changes from ALL-1809/chore/settings-breakdown into feat/secret-settings-manager. The conflicts were in three files:

1. frontend/src/routes/settings.tsx - Added the Secrets navigation item with proper i18n support
2. openhands/server/routes/settings.py - Updated the function signature to use the dependency injection pattern
3. tests/unit/test_secrets_api.py - Updated the test functions to match the new API behavior

Note: There are still some test failures in test_secrets_api.py, but they are not related to the merge conflicts. They are implementation issues that need to be addressed separately.

amanape:
I feel like it merged the wrong code for the backend? @malhotra5 could you confirm?

malhotra5:
@amanape thanks for the ping! I've fixed the issues in BE

malhotra5:
Hi @amanape, I've added description field for custom secrets

We will need to send secrets in the following format for the FE (specifically we'll need to fill in the new fields `secret` and `description`)


```typescript
{SECRET_NAME_1: {
              secret: MY_SECRET_VALUE_1, 
              description: 'first secret value'
    },
 SECRET_NAME_2: {
             secret: MY_SECRET_VALUE_2,
             description: 'second secret value'
    }
}
```

malhotra5:
@openhands please look at the new format for CustomSecrets (inside the file `tests/unit/test_secrets_api.py`) for reference 

Then go ahead and fix the failing tests in `tests/unit/test_secret_store.py`

malhotra5:
@OpenHands please look at the new format for CustomSecrets (inside the file tests/unit/test_secrets_api.py) for reference

Then go ahead and fix the failing tests in tests/unit/test_secret_store.py

enyst:
Just to note, I think it's worth discussing if the format and handling of secrets that we're heading towards are compatible with [keyring](https://pypi.org/project/keyring/) library or something similar. We may need something like that for local installations.

I would love a full description of this PR, and ideally I would also like to know the overall plan on this (this PR and follow-ups). 

How do you need to work with the secrets store and settings store, in the saas use case? Are they, as the name implies, heading towards two different stores?

As of now, we created a new object under Settings, and dumped it into .json, disconnected from the rest of the application, named 'store' but not a store. That's not ideal, sorry, it's confusing. 😅

malhotra5:
Yeah super good point; I'll be circling back to this sometime soon and will address the points mentioned above

malhotra5:
Hey @enyst, here's what I'm thinking

- `SecretStore` is a part of the `Settings` model. I'm dumping the entire `Settings` object using the `SettingsStore`. I'd like to avoid using a separate store for `SecretStore` to avoid fragmenting a single `Settings` object across multiple stores. I think the name `SecretStore` is what causes confusion

- My initial impression is that it should be easy to migrate current implementation to use KeyRing. Since we store additional attributes along with a secret (e.g description for a custom secret), we'd dump a json object in KeyRing password instead of the `FileStore`

- As per future work, I have a PR #8105 which will plumb all the secrets (provider tokens + custom secrets) to the runtime. That should be it!

enyst:
> Seems like probably future work too, rather than something we need for this PR

@rbren I looked into it a bit, and I would love more information. It would be good to know, for example, if compatibility with a single "Settings" or a single file store is necessary for some reason for the saas. What is the plan here? Can we break the custom secrets out into its own store or do you need an adaptor? Breaking it out would be better I think.

rbren:
@enyst I agree with you, we should have separate SecretStore and SettingsStore, regardless of implemenation. @malhotra5 said he can take sth like that on

rbren (approved):

PR #8077, opened 2025-04-24
Add logging for failed suggested tasks attempts

malhotra5:
This PR addresses the ReadTimeout issues in the GitHub service by splitting the large GraphQL query into two separate queries - one for PRs and one for issues. It also reduces the number of records requested from 100 to 50 for each query to further reduce the likelihood of timeouts.

The changes should help resolve the `HTTP error on github API: ReadTimeout https://api.github.com/graphql` error that was occurring.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0489d37-nikolaik   --name openhands-app-0489d37   docker.all-hands.dev/all-hands-ai/openhands:0489d37
```


openhands/integrations/github/github_queries.py
enyst, line 1: Nit: we can probably name this `queries` since it's in `github` already

openhands/integrations/github/github_service.py
```diff
@@ -44,6 +48,9 @@ def __init__(
         if base_domain:
             self.BASE_URL = f'https://{base_domain}/api/v3'
 
+        self.external_auth_id = external_auth_id
+        self.external_auth_token = external_auth_token
```
enyst, line 51: This is strange, so these were never set. 🤔 

There's more strange though: below, we log them on exception, but they were never used in the code raising the exception?

enyst: Could you please tell, what is the meaning of `external_auth_id`, why "external"?

malhotra5: They are used in saas, we set the values then

However, this time I set it in the initializer of oss as well so that mypy doesn't complain when I log it below

I'm logging this so that we can diagnose how many people are facing timeout issues in saas mode

malhotra5: We use an external provider for saas login, the user id comes from there

rbren (approved):

PR #8078, opened 2025-04-24
fix - Show too many sessions error

raymyers:

---

When we try to start more conversations then the configured `max_concurrent_conversations`, the oldest conversations are stopped to make room. However, there was no user feedback saying why it was stopped - this PR addresses that.

* Add status message for "too many conversations" case.
* Fix log format string
* Requesting git changes in a non-git repo is now 404, not 500, since it's not a server-side error.

Screenshot 2025-04-24 164508

When expanded it just shows the same message, as many of these currently do.

Screenshot 2025-04-24 164515

Agent stopped status is the same as before

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:8392493-nikolaik   --name openhands-app-8392493   docker.all-hands.dev/all-hands-ai/openhands:8392493
```


openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -297,6 +297,18 @@ async def _start_agent_loop(
 
             while len(conversations) >= self.config.max_concurrent_conversations:
                 oldest_conversation_id = conversations.pop().conversation_id
+                logger.info(
```
enyst, line 300:
```suggestion
                logger.debug(
```

enyst: Since the user sees in the UI, it may be a bit overkill to set info

openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -297,6 +297,18 @@ async def _start_agent_loop(
...
+                    f'closing_from_too_many_sessions:{user_id}:{oldest_conversation_id}',
```
enyst, line 301: Nit: `user_id` can be None, maybe we could avoid to display anything then?

enyst (commented):
Maybe we could use the colors / style of the RateLimited (I think?) which wasn't quite error? Idk, just thinking that a yellow -ish may be okay for an informative message, rather than the red color for scary error

Just a thought, wdyt?

openhands/server/conversation_manager/standalone_conversation_manager.py
raymyers: Done. It might also be useful for saas observability, but we can try just the previous message at info first.

raymyers:
@enyst 
> Idk, just thinking that a yellow -ish may be okay for an informative message, rather than the red color for scary error

Looks like info and error are the two options, I'd be OK going to info as long as it displays. I think these are recoverable?

https://github.com/All-Hands-AI/OpenHands/blob/9b1aaa53fea318bbcfcfd24047c80a042aa085a8/frontend/src/services/actions.ts#L127

openhands/server/conversation_manager/standalone_conversation_manager.py
raymyers: Decent idea, added `or ''`

enyst:
I thought they're not only recoverable, but not really treated as errors, more like signals that we need to close something. Is that incorrect?

raymyers:
It looks like the RateLimit message shows up as a result of an AgentState, which this is not, so I'm not sure that the same logic would apply

enyst:
Got it, I mean the style though. I don't have a strong opinion or something, I just feel this is a normal operation, not an error - it's expected that the user will open too many, and that we need to close a conversation.

It may even be nice to say in the message "Conversation limit reached: closed the oldest", since right now it doesn't say what we do?

enyst (approved):

enyst:
Not a big deal, though, up to you.

raymyers:
>  right now it doesn't say what we do?

@enyst, Good ideas. I've updated the error to explain the action to take. I also agree with your color suggestion, but it will require generalizing how messages are displayed to allow `warning` status. I'll split that into a separate PR.

PR #8082, opened 2025-04-25
Maybe fix for pre-commit errors

chuckbutkus:

---

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f8673bc-nikolaik   --name openhands-app-f8673bc   docker.all-hands.dev/all-hands-ai/openhands:f8673bc
```

### DIFF
--- a/frontend/scripts/check-unlocalized-strings.cjs
+++ b/frontend/scripts/check-unlocalized-strings.cjs
@@ -39,7 +39,7 @@ const IGNORE_PATHS = [
   "entry.client.tsx", // Client entry point
   "utils/scan-unlocalized-strings.ts", // Original scanner
   "utils/scan-unlocalized-strings-ast.ts", // This file itself
-  "frontend/src/components/features/home/tasks/get-prompt-for-query.ts", // TODO: remove we need to localize later
+  "frontend/src/components/features/home/tasks/get-prompt-for-query.ts", // Only contains agent prompts
 ];
 
 // Extensions to scan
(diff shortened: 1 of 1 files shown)


enyst (approved):
Definitely, thank you! I'm not sure if the path from `frontend` works, but otherwise looks like this is it!

On a side note, prompt pieces probably belong in the backend... but that's a thing for another day!

neubig (changes requested):
This doesn't fix the problem, just ignores it. We need to keep our frontend localized.

UPDATE: correction -- the file only contains prompts for the agent, so ignoring is OK, but the comment should be changed.

frontend/scripts/check-unlocalized-strings.cjs
```diff
@@ -39,6 +39,7 @@ const IGNORE_PATHS = [
   "entry.client.tsx", // Client entry point
   "utils/scan-unlocalized-strings.ts", // Original scanner
   "utils/scan-unlocalized-strings-ast.ts", // This file itself
+  "frontend/src/components/features/home/tasks/get-prompt-for-query.ts", // TODO: remove we need to localize later
```
neubig, line 42:
```suggestion
  "frontend/src/components/features/home/tasks/get-prompt-for-query.ts", // Only contains agent prompts
```

neubig (approved):

malhotra5:
Thank you to everyone who helped here, this was an issue introduced by one of my PRs

PR #8083, opened 2025-04-25
[Fix]: Sort github repos in memory for cloud openhands

malhotra5:

---

This PR sorts github repos by when the repo was last pushed to in memory

We do this because github api  doesn't support sorting by this parameter for app installation repos

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:643a967-nikolaik   --name openhands-app-643a967   docker.all-hands.dev/all-hands-ai/openhands:643a967
```

### DIFF
--- a/openhands/integrations/github/github_service.py
+++ b/openhands/integrations/github/github_service.py
@@ -203,12 +203,9 @@ class GitHubService(BaseGitService, GitService):
 
         if app_mode == AppMode.SAAS:
             if sort == "pushed":
-                print("repos fetchd", all_repos)
                 all_repos.sort(
                     key=self.parse_pushed_at_date, reverse=True
                 )
-
-                print("sorted repos", all_repos)
         
         # Convert to Repository objects
         return [
--- a/openhands/integrations/github/github_service.py
+++ b/openhands/integrations/github/github_service.py
@@ -18,6 +18,8 @@ from openhands.integrations.service_types import (
 )
 from openhands.server.types import AppMode
 from openhands.utils.import_utils import get_impl
+from datetime import datetime
+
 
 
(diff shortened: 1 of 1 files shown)


openhands/integrations/github/github_service.py
```diff
@@ -191,6 +200,13 @@ async def get_repositories(self, sort: str, app_mode: AppMode) -> list[Repositor
             # Fetch user repositories
             all_repos = await self._fetch_paginated_repos(url, params, MAX_REPOS)
 
+
+        if app_mode == AppMode.SAAS:
```
enyst, line 204: Can we put this code a few lines above, in the other `if app_mode == AppMode.SAAS` ?

I see it relies on installation_id (ids?) and finds this exact list

rbren (approved):
pending comment from enyst

openhands/integrations/github/github_service.py
```diff
@@ -203,6 +218,9 @@ async def get_repositories(self, sort: str, app_mode: AppMode) -> list[Repositor
             for repo in all_repos
         ]
 
+    
+
+
```
enyst, line 223:
```suggestion
```

Just a trivial question, it seems pre-commit still didn't run? Not sure, I think it doesn't like added spaces/lines.

enyst (approved):

openhands/integrations/github/github_service.py
malhotra5: Hmm not sure, must be something in my setup

PR #8093, opened 2025-04-25
Allow cmd execution after running a background command

li-boxuan:

Allow cmd execution after running a background command

---

Bug: Background commands (e.g., python -m http.server &) incorrectly time out instead of returning immediately. This happens because the completion logic checks if the entire terminal output ends with the PS1 prompt marker. If the background process prints any output after the prompt marker appears but before the next check cycle, the check fails, leading to a false timeout (NO_CHANGE_TIMEOUT).

Motivation: this is important for frontend debugging - agent wants to start an http server in the background and in the mean time, continues to do development work.

Reproduction: Run a background command that prints output shortly after launch, like python3 -m http.server 8080 &. Observe that it triggers a timeout observation instead of completing immediately with exit code 0.

Solution: The completion logic in BashSession.execute is updated. Instead of checking output.endswith(PS1_MARKER), it now counts the number of PS1 markers before and after sending the command. If the count increases, it signifies the shell has processed the command and is ready for input, correctly handling immediate returns from background tasks.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:99ba123-nikolaik   --name openhands-app-99ba123   docker.all-hands.dev/all-hands-ai/openhands:99ba123
```

### DIFF
--- a/openhands/runtime/utils/bash.py
+++ b/openhands/runtime/utils/bash.py
@@ -505,9 +505,19 @@ class BashSession:
                 )
             )
 
+        # Get initial state before sending command
+        initial_pane_output = self._get_pane_content()
+        initial_ps1_matches = CmdOutputMetadata.matches_ps1_metadata(
+            initial_pane_output
+        )
+        initial_ps1_count = len(initial_ps1_matches)
+        logger.debug(f'Initial PS1 count: {initial_ps1_count}')
+
         start_time = time.time()
         last_change_time = start_time
-        last_pane_output = self._get_pane_content()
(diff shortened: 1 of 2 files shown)


neubig (approved):
Thanks for fixing this!

neubig:
But tests seem to be failing, those should be examined.

Issue #7869, opened 2025-04-15
[Bug]: OpenHands will interrupt setup.sh if its still running + will interrupt long-running shell commands, crashing the terminal


oconnorjoseph:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

For setup.sh scripts that take some time to run, OpenHands will interrupt the execution of setup.sh instead of waiting for the shell script to finish.

### OpenHands Installation

Docker command in README

### OpenHands Version

0.32.0

### Operating System

Linux

### Logs, Errors, Screenshots, and Additional Context

Here is our setup.sh:
```
#!/bin/bash
set -e

echo "Starting BluMint development environment setup..."

# Function to check if a command exists
check_command() {
  if ! command -v "$1" &> /dev/null; then
    echo "$1 command not found. Installing..."
    return 1
  else
    echo "$1 is already installed."
    return 0
  fi
}

# Function to check if an npm package is installed globally
check_npm_global() {
  if ! npm list -g "$1" &> /dev/null; then
    echo "Global npm package $1 not found. Installing..."
    return 1
  else
    echo "Global npm package $1 is already installed."
    return 0
  fi
}

# Install required npm packages if they don't exist
check_npm_global "dotenv-cli" || npm install -g dotenv-cli -y
check_npm_global "dotenv-vault" || npm install -g dotenv-vault -y
check_npm_global "commitizen" || npm install -g commitizen -y
check_npm_global "@commitlint/cli" || npm install -g @commitlint/cli -y
check_npm_global "@commitlint/config-conventional" || npm install -g @commitlint/config-conventional -y
check_npm_global "cross-env" || npm install -g cross-env -y
check_npm_global "firebase-tools" || npm install -g firebase-tools@13.31.1 -y

# Install local project dependencies
echo "Installing local project dependencies from package.json..."
npm install || echo "Warning: npm install had some issues, but continuing setup. You may need to run npm install manually."

# Setup Python environment if needed
echo "Setting up Python environment..."
export PYTHONUSERBASE=/workspace/.pip-modules
export PATH=$PYTHONUSERBASE/bin:$PATH
export PIP_USER=yes

# Install Foundry using the script
echo "Installing Foundry..."
if [ -f "./.devcontainer/install_foundry.sh" ]; then
    bash ./.devcontainer/install_foundry.sh
fi

# Ensure required directories exist
[...]

PR #8101, opened 2025-04-25
Fix long setup scripts failing

neubig:

Fixes a bug where long-running setup scripts failed to complete due to being interrupted. 

---

The root cause is that the script was being interrupted.

```
15:57:35 - openhands:INFO: base.py:414 - Running setup script: .openhands/setup.sh
15:57:46 - openhands:INFO: base.py:416 - Done running setup script: **CmdOutputObservation (source=None, exit code=-1, metadata={
  "exit_code": -1,
  "pid": -1,
  "username": null,
  "hostname": null,
  "working_dir": null,
  "py_interpreter_path": null,
  "prefix": "",
  "suffix": "\n[The command has no new output after 10 seconds. You may wait longer to see additional output by sending empty command '', send other commands to interact with the current process, or send keys to interrupt/kill the command.]"
})**
--BEGIN AGENT OBSERVATION--
Starting setup script...
This script will sleep for 2 minutes...
[The command has no new output after 10 seconds. You may wait longer to see additional output by sending empty command '', send other commands to interact with the current process, or send keys to interrupt/kill the command.]
--END AGENT OBSERVATION--
```

This fixes it by setting the command to be blocking with a hard timeout of 10 minutes. It also:

* fixes an issue where the "blocking" command was ignored if a timeout was not set, which was confusing.
* makes it possible to download repos without GITHUB_TOKEN if none is specified, which makes it easier to to test the github initialization code.

---

Fixes https://github.com/All-Hands-AI/OpenHands/issues/7869

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:bd9d7f2-nikolaik   --name openhands-app-bd9d7f2   docker.all-hands.dev/all-hands-ai/openhands:bd9d7f2
```


neubig:
@openhands install pre-commit from pre-commit-config.yaml and run it only on the python files that were changed in this PR

enyst (approved):
LGTM! And thank you for the fixing the other issues. It's a bit curious they didn't fail before.

On a different note, I wonder if a RuntimeError is a little too tough. I don't know the detailed behavior of blocking commands well enough to see if it can ever happen "normally". Maybe it's better to fail and see what happened, than silently have the wrong behavior. Up to you.

neubig:
I think that if they ever happen it is a programming error, so we should probably catch and fix them as soon as possible.

Issue #6707, opened 2025-02-12
Token-based condensation triggers


csmith49:
The current memory condensers use the size of the history (i.e., the number of `Event` objects in the history) as a proxy for how much information is contained in the history. But the downstream token costs vary wildly from event to event.

These graphs show the size of events generated when solving SWE-bench Verified instances -- the larger the circle, the more tokens in the event:

![Image](https://github.com/user-attachments/assets/f4c625b4-168a-4aea-a5c9-6df077865257)

The distribution changes based on condensation used, problem instance, OpenHands version and prompt, etc. This leads to difficult-to-debug situations where, even with condensation, we still exceed model max input token limitations.

**We should extend condensers to support truncation based on the number of tokens per-event.** 

This requires we 1) convert events to messages, then 2) approximate the number of tokens needed to encode each message (using custom tokenizers).

As an added bonus, the context truncation that happens in the agent controller when a `ContextWindowExceededError` can be unified with the custom condenser implementations:
* Configurable context truncation strategies (summarizing, forgetting, masking, etc.)
* Configurable max token limits
* Unlike the agent controller's context truncation, custom condensers do not modify the state so we maintain a complete history

### If you find this feature request or enhancement useful, make sure to add a 👍 to the issue

PR #7587, opened 2025-03-29
Token-aware condenser

enyst:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

---
**Give a summary of what the PR does, explaining any non-trivial design decisions.**

This PR proposes to add a token-aware LLM summarizing condenser.

---
**Link of any specific issues this addresses.**

PR #8102, opened 2025-04-26
feature(agent): Trigger condensation by conversation token length

happyherp:
Work in progress.

Builds upon the changes in **https://github.com/All-Hands-AI/OpenHands/pull/7893**

Allows a condenser to be triggered, as soon as the conversation reaches a configurable token-limit. This limit could then be set for example to be close to the LLMs context-window size, so that condensations are only triggered when necessary. 

This is much better, than relying on the number of events(as current condensers mostly do), which is a less good measure of conversation length, since we really care more about tokens. 

---

## Trigger class hierarchy

Created a Trigger-Interface to be able to create a multiple subclasses of condensation-triggers(event-count, token-count, keyword-trigger) and a way to combine them. 

From [LLMCacheCondenser](https://github.com/happyherp/OpenHands/blob/12fdd5c7a7b8e8e8806cc751c93a29b13082e777/openhands/memory/condenser/impl/llm_agent_cache_condenser.py#L45): 

```python
        self.trigger = (
            KeywordTrigger(trigger_word)
            | EventCountTrigger(max_size)
            | ConversationTokenTrigger(max_tokens)
        )
```

It can then be used like this:

```python
    def should_condense(self, view: View, state: State, agent=None) -> bool:
        return self.trigger.should_condense(view, state, agent)
```

## ConversationTokenTrigger

This is the new trigger that reacts to the conversations token count. 

https://github.com/happyherp/OpenHands/blob/condense-by-token/openhands/memory/condenser/trigger.py#L98

```python
 messages = agent.get_messages(view.events)
            token_count = agent.llm.get_token_count(messages)

            if token_count > self.max_tokens:
                logger.info(
                    f'Condensing events due to token count ({token_count}) exceeding limit ({self.max_tokens}).'
                )
                return True

```

---
https://github.com/All-Hands-AI/OpenHands/issues/6707


happyherp:
When I tested the code, I got some very weird token-counts. This test shows the problem `test_get_token_count_with_tool`
Unfortunately it looks like it might go all the way down to the litellm logic. There is some weird case, where a message with a tool call somehow causes the tokens of previous messages not to be counted.

enyst:
@happyherp Thank you so much for looking into this! We absolutely need condensation by tokens, and missing it hinders the flexibility we would otherwise have to work robustly (IMHO it's even related to the bug we just talked about, because we don't do exception handling in condensers, so no catching ContextWindowExceededError, so we have to handle that elsewhere - in the controller, which is not ideal, similar purpose code doing different things in different places)

FWIW I took a stab at this here:
- https://github.com/All-Hands-AI/OpenHands/pull/7587

I don't _love_ that attempt (7587), I left it in the back burner to think about it. It uses only the tokens in the responses from the LLM, and while that seems sort of correct, it will not be correct unless we always keep the same condenser or lack of, same LLM too... and interpret every time the same way when we restore the session. It felt fragile to me. I didn't yet look at this PR, but I know you handled the events to messages conversion, so you know what we're dealing with, here. 😅 ❤️

openhands/llm/llm.py
```diff
@@ -655,14 +661,43 @@ def get_token_count(self, messages: list[dict] | list[Message]) -> int:
...
+            int: The number of tokens.
+        """
+        try:
+            return int(
+                litellm.token_counter(
+                    model=self.config.model,
+                    messages=params['messages'],
+                    tools=params.get('tools', None),
+                    custom_tokenizer=self.tokenizer,
+                )
```
happyherp, line 679: the result of token_counter goes **down** the first time you add a message that contains a tool call and then seems to only count the message with the tool call, preventing us from ever hitting our llimit.

Turns out it is a larger issue that I try to fix here: https://github.com/BerriAI/litellm/pull/10409

happyherp:
a version with my changes to litellm is already deployed: https://github.com/BerriAI/litellm/releases/tag/v1.67.6-nightly

mamoodi:
@happyherp is this still a valid PR? You've put a fair amount into it and I don't want it to be closed without your input.

neubig:
It seems like  this should be closed for lack of activity, but we welcome re-opening it if there is still interest!

PR #7917, opened 2025-04-18
Refactor to clean up and move utility/legacy out of the agent

enyst:
**Give a summary of what the PR does, explaining any non-trivial design decisions.**

Small refactoring, move/remove some code from the agent:
- move the code that ensures a `SystemMessageAction` exists from the agent to `conversation_memory.py`, the agent helper that handles what is actually sent in context (it may add or remove stuff)
- move the code that formats the messages to have separators, also to the agent helper that formats stuff to be sent in context

None of these differ per agent, they're more like utility stuff to be done before the final list of messages is sent to `llm.py`

Also:
- removed some of the "initial examples" code, which is dead code since the examples don't exist with tool use, and are actually in `fn_call_converter.py` when tool use is not enabled.

NOTE:
The rest of the changes are only linting.

---
**Link of any specific issues this addresses.**

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:aa05f2b-nikolaik   --name openhands-app-aa05f2b   docker.all-hands.dev/all-hands-ai/openhands:aa05f2b
```

PR #8103, opened 2025-04-26
Fix truncation, ensure first user message and log

enyst:

---

This PR proposes to:
- send the initial user message to the LLM anyway, even if it's missing in normal execution (and it should never be!)
- if it vanished from `history` though, raise in the agent (we wouldn't know it anymore)
- add logging for debugging this issue 

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:65982bd-nikolaik   --name openhands-app-65982bd   docker.all-hands.dev/all-hands-ai/openhands:65982bd
```


enyst:
@openhands-agent-exp 

<IMPORTANT>
Read and understand this PR, PR 8103 . You can get the diff via GitHub API with the token in your env.

Fix unit tests. Many errors are due to signature change in this PR, fix that.
</IMPORTANT>

openhands-agent:
[OpenHands](https://github.com/All-Hands-AI/OpenHands) started fixing the pr! You can monitor the progress [here](https://github.com/All-Hands-AI/OpenHands/actions/runs/14683351472).

openhands-agent:
Based on the provided information, here's a concise overview:

✅ Successfully Resolved:
- Unit tests updated to include required initial user messages
- Message count assertions adjusted for new message sequence
- Source attribution fixed using proper EventSource enums
- process_events calls updated with initial_user_action parameter
- Message position references corrected in test assertions

🔍 Status: The changes appear to have successfully addressed all the identified issues related to signature changes in the unit tests. No remaining issues are mentioned in the provided information.

Note: To be completely certain that all issues are resolved, it would be helpful to know if all tests are now passing and if there are any new test failures or edge cases that need attention.

openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -173,7 +170,26 @@ def step(self, state: State) -> Action:
...
+                break
+
+        if initial_user_message is None:
+            # This should not happen in a valid conversation
+            logger.error(
+                f'CRITICAL: Could not find the initial user MessageAction in the full {len(state.history)} events history.'
+            )
+            # Depending on desired robustness, could raise error or create a dummy action
+            # For now, let's create a dummy action to avoid crashing
+            # and log the error
```
raymyers, line 187: We can delete this comment, it contradicts the code

enyst: Oh, hah, you're right.

It has a history! Gemini refused to follow the "FIXME" that I had here, to create a dummy action, and instead it deleted the "FIXME" (but kept the rest for some reason). It argued that we can't reasonably continue, so this FIXME is outdated or wrong, and we have to raise. 😅

openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -173,7 +170,26 @@ def step(self, state: State) -> Action:
...
+                initial_user_message = event
+                break
+
+        if initial_user_message is None:
+            # This should not happen in a valid conversation
+            logger.error(
+                f'CRITICAL: Could not find the initial user MessageAction in the full {len(state.history)} events history.'
+            )
+            # Depending on desired robustness, could raise error or create a dummy action
+            # For now, let's create a dummy action to avoid crashing
```
enyst, line 186:
```suggestion
```

openhands/agenthub/codeact_agent/codeact_agent.py
enyst: Done!

raymyers:
For context, we believe that there was a regression in release tag 0.34.0, most likely one of these PRs.

https://github.com/All-Hands-AI/OpenHands/pull/7824

https://github.com/All-Hands-AI/OpenHands/pull/7917

The error in question is 'AnthropicException - Invalid first message'. The first message in these cases always seems to be an assistant tool_call.

raymyers (approved):
Approving because right now adding the message back in after the fact seems like the best we can do.

I would feel better if we were able to understand why this is happening, fix the logic and test with a reproduction of the scenario. But unfortunately, I can't yet make head or tail of the cause.

openhands/memory/conversation_memory.py
```diff
@@ -699,6 +702,55 @@ def _ensure_system_message(self, events: list[Event]) -> None:
...
+                'Initial user message action was missing. Inserting the initial user message.'
+            )
+            events.insert(1, initial_user_action)
+        elif not isinstance(events[1], MessageAction) or events[1].source != 'user':
+            # The second event exists but is not the correct initial user message action.
+            # We will insert the correct one provided.
+            logger.info(
+                'Second event was not the initial user message action. Inserting correct one at index 1.'
+            )
+            # Check if the *correct* initial action is already present somewhere else *after* index 1
```
raymyers, line 733: Do we need this part? I'm not sure why it would ever be the case that a user message would be moved later in the list, and it's a bit of work to content check them all every time. Worst case it gets duplicated but I don't think it would happen.

enyst: For the logging part, just so we know it happened... You're right it shouldn't though...

enyst: If it is related to the condenser, it inserts an event in history [here](https://github.com/All-Hands-AI/OpenHands/blob/8a5c6d3befe11c48293485c1be75fd47e8301198/openhands/memory/view.py#L68)

That could, if something's broken, shift these in the wrong place, so to speak...

Actually, I'll remove this and add a logging message there. `info` is okay, right?

enyst:
@raymyers Let's not merge before tomorrow, if we can. Though I think fixing the LLM context when we really must (when we know it's a 400) is actually necessary, this PR doesn't do all it could.

raymyers:
@enyst 
> Let's not merge before tomorrow, if we can [...]

Sounds good. It's happening only a few times a day I think. Also curious if @csmith49 has any insights when available.

happyherp:
I read the [conversation on slack ](https://openhands-ai.slack.com/archives/C078L0FUGUX/p1745675622573399) and I also believe this is caused by a condenser. I had this happen as well. All that needs to happen, is that any condenser implementation decides to forget the system message. If then there is also no user-message, anthropic will throw an exception. 
I also added a check in my PR because it kept happening: https://github.com/All-Hands-AI/OpenHands/pull/7893/files#r2062456608

What I think happens here: 
* Some condenser puts the id of the system message(and probably all user messages as well) in `CondensationAction.forgotten_event_ids`
* * for this to happen it would be enough to set `keep_first` to 0 for any condenser that has it.
* The condensation is applied through `View.from_events(events)` which removes the system message. 


So we might want to consider to have some check that prevents a condenser from forgetting the system message. And this check should not be in a condenser, so that it applies to all of them. 

One place would be `View.from_events()`

change 

```python
kept_events = [event for event in events if event.id not in forgotten_event_ids]
```
 to 

```python
kept_events = [event for event in events if event.id not in forgotten_event_ids and.event.source !== "system"]
```

And maybe add a check for a soft-error so that we still notice this and can fix the condenser. Probably no condenser should ever forget the system prompt.

enyst:
@happyherp @raymyers @csmith49 

Thank you for this, Carlos! I found an issue in the controller truncation, and updated the PR. I think you're right though, a similar issue exists in the condensers, and we should check for the mandatory first events to keep. I tried to rewrite the check in the truncation code in a more similar way with condensers, so we can merge them eventually.

To note, during debugging I had these thoughts about the issues with truncation:
- after the system message event PR (7824), this code was dropping the system event (it kept only initial user message)
- after the combination of truncation with condensation PR (7578), this code no longer runs recursively
- I think [all these 4](https://github.com/All-Hands-AI/OpenHands/pull/8103/files#diff-1e082f415afb40de1d5c2a3fa327326225e1da28a142ad89c9524109f35fb36eR1203) are essential events that need to be kept in the truncation and in the condensers too
- I cleaned up some leftover code messing with `start_id`, which is there to account for delegates (they share the same event stream with their parent agent), but should work implicitly

Also, agent controller has become [a little too fat](https://github.com/All-Hands-AI/OpenHands/issues/8111). 😅

openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -173,7 +170,25 @@ def step(self, state: State) -> Action:
...
+        if initial_user_message is None:
+            # This should not happen in a valid conversation
+            logger.error(
+                f'CRITICAL: Could not find the initial user MessageAction in the full {len(state.history)} events history.'
+            )
+            # Depending on desired robustness, could raise error or create a dummy action
+            # and log the error
+            raise ValueError(
+                'Initial user message not found in history. Please report this issue.'
+            )
```
happyherp, line 189: this check is very verbose and the .step method is a very high-level method in my opinion. 

We might want to consider moving this code to `CodeActAgent._check_initial_message_present(state:State)`

enyst: Done! Maybe we should even move it to the `conversation_memory` helper altogether...

happyherp: yes actually. Except I still don't understand where the **Memory** part of that class is. What it seems to do is handle the conversion from Event -> Message.

tests/unit/test_agent_history.py
```diff
@@ -0,0 +1,569 @@
...
+    # Non-essential count = 9 - 3 = 6
+    # num_recent_to_keep = max(1, 6 // 2) = 3
+    # slice_start_index = 9 - 3 = 6
+    # recent_events_slice = history[6:] = [obs2(7), cmd3(8), obs3(9)]
+    # Validation: remove leading obs2(7). validated_slice = [cmd3(8), obs3(9)]
+    # Final = essentials + validated_slice = [user(1), recall_act(2), recall_obs(3), cmd3(8), obs3(9)]
+    # Expected IDs: [1, 2, 3, 8, 9]. Length 5.
+    truncated_events = controller._apply_conversation_window()
+
+    assert len(truncated_events) == 5  # Corrected assertion
```
csmith49, line 230: Very minor nit, but the `# Corrected assertion` comments here don't add anything

tests/unit/test_agent_controller.py
```diff
@@ -807,13 +806,19 @@ def on_event_memory(event: Event):
...
+    # post-step view compressed (or rather, a CondensationAction added).
     for index, (first_view, second_view) in enumerate(
         zip(step_state.views[:-1], step_state.views[1:])
     ):
         if index == error_after:
-            assert len(first_view) > len(second_view)
+            # Verify that the CondensationAction is present in the second view (after error)
+            # but not in the first view (before error)
+            assert not any(isinstance(e, CondensationAction) for e in first_view.events)
+            assert any(isinstance(e, CondensationAction) for e in second_view.events)
```
csmith49, line 817: I'm a little concerned that this is passing. The original intent of the `View` object was to strip out the `CondensationAction` entries and apply their semantics, if they're showing up here they're also possibly sent to the LLM.

Shouldn't be a blocker for this PR, but something I want to look into.

tests/unit/test_agent_history.py
enyst: Fixed!

I used Gemini for some of this PR, and I think it adds even more useless comments than Sonnet, it's like it permanently reports what it did, in comment form. 😅

enyst:
Merging this, because it's a real issue, when controller truncation is hit. Follow-ups are in progress though, because this isn't solved, as Carlos and Calvin pointed out.

Issue #5826, opened 2024-12-26
[Feature]: Add input-based micro-agents to support template workflows


xingyaoww:
**What problem or use case are you trying to solve?**
We need to extend micro-agents to support template-based workflows. Currently, micro-agents are triggered by keywords and inject specialized knowledge (e.g., GitHub API understanding). However, users often find themselves repeatedly typing similar prompts that could be templatized.

For example, a template for addressing PR comments could be a micro-agent that accepts inputs:
```
Can you look at this PR ${PR_URL}
And then check out to branch ${BRANCH_NAME} which corresponds to that PR
Then, read the diff against the main, understand the purpose of this PR, and then address the PR comments.
${USER_MESSAGE}
```

**Describe the UX of the solution you'd like**
Micro-agents would have two triggering mechanisms:
1. Traditional keyword-based triggers (e.g., `.openhands/microagents/*.md`)
2. UI-based selection where users can pick a micro-agent and provide required inputs via a form in FE

For future: If the user did NOT provide those inputs, the micro-agent itself can conversationally ask users for the required information.

**Do you have thoughts on the technical implementation?**
- Add an `inputs` parameter to the micro-agent system to specify required template variables
- Store micro-agent templates in the repository (e.g., in the existing micro-agents location)
- Add UI elements:
  - Button to create new micro-agent templates
  - "Save" button to automatically create PRs against the repository
- Future potential: Build an API to trigger these micro-agents (e.g., `/start_session?workflow=resolve_issue`)

**Additional context**
This enhancement maintains the existing micro-agent concept while adding template capabilities. Rather than creating separate systems for specialized knowledge injection and templated workflows, we're unifying them under the micro-agent framework with support for both keyword triggers and user-supplied inputs.

How's this revised version that maintains the micro-agent terminology while incorporating the template functionality?

xingyaoww:
@OpenHands can you look into this, and come up with a plan of how we can implement this? You should look under the "microagents/" folder, as well as the current implementation of micro-agents in the backend.

Then look at frontend/ folder and try to come up maybe 3-5 different approaches of how we can support these "task" microagents in frontend UI, and how would they interact with backend server.

xingyaoww:
This will be fixed by https://github.com/All-Hands-AI/OpenHands/pull/8139

PR #8109, opened 2025-04-27
[Fix]: Move suggest task prompts to BE

malhotra5:

---

This PR does the following 

1. Tracks a new type of conversation trigger `SUGGESTED_TASK`
2. Moves all the suggested task prompts to the BE, and forms the appropriate msg based on the task chosen (which is sent from the FE)

This was manually tested 

1. Starting convo from suggested task
2. Starting convo with a selected repo
3. Starting convo from scratch

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:bf314bd-nikolaik   --name openhands-app-bf314bd   docker.all-hands.dev/all-hands-ai/openhands:bf314bd
```


malhotra5:
@openhands fix the failing fe lint

frontend/src/components/features/git/code-not-in-github-link.tsx
```diff
@@ -1,33 +0,0 @@
...
-
-  const handleStartFromScratch = () => {
-    // Set the initial prompt and create a new conversation
-    dispatch(setInitialPrompt(INITIAL_PROMPT));
-    createConversation({ q: INITIAL_PROMPT });
-  };
-
-  return (
-    <div className="text-xs text-neutral-400">
-      {t(I18nKey.GITHUB$CODE_NOT_IN_GITHUB)}{" "}
```
enyst, line 23: Curious why is this removed in the PR, don't we have "start from scratch" anymore?

malhotra5: Ah this is an old component

The UI was rebuilt and the launch from scratch is in frontend/src/components/features/home/home-header.tsx

openhands/server/routes/manage_conversations.py
```diff
@@ -169,17 +173,23 @@ async def new_conversation(
...
     replay_json = data.replay_json
+    suggested_task = data.suggested_task
+    conversation_trigger = data.conversation_trigger
+
+    if suggested_task:
+        initial_user_msg = get_prompt_for_task(suggested_task)
 
     try:
         # Create conversation with initial message
         conversation_id = await _create_new_conversation(
```
enyst, line 184: Sorry I'm not sure off hand, at this point, do we know the agent? Or when do we know it?

malhotra5: Sorry could you clarify what you mean by know the agent?

openhands/server/routes/manage_conversations.py
```diff
@@ -13,13 +13,14 @@
...
 )
-from openhands.integrations.service_types import Repository
+from openhands.integrations.service_types import Repository, SuggestedTask
 from openhands.runtime import get_runtime_cls
 from openhands.server.data_models.conversation_info import ConversationInfo
 from openhands.server.data_models.conversation_info_result_set import (
     ConversationInfoResultSet,
 )
 from openhands.server.session.conversation_init_data import ConversationInitData
+from openhands.server.session.initial_agent_prompts import get_prompt_for_task
```
enyst, line 23: Nit: how about we dump them in, hmm, I don't want to say `utils`... 🫠 but I don't see another, so maybe `openhands/utils/initial_agent_prompts`? Where `prompt.py` is.

The agent's `/prompts` location should ideally be the place for this... but it's not clear to me how we can in this case, it's a bit downstream from here, right?

The reason I wonder about this is that prompts are per agent, and the agent has utilities to work with them, `PromptManager` working with jinja templates rather than f-strings.

That said, these are initial user messages, we have them all over the place, and they can come from outside too, so we can live with them outside the agent.

openhands/server/routes/manage_conversations.py
enyst: Ah, same idea below: I was just wondering if we can move these prompt pieces to `codeact_agent/prompts` directory. Do we know that the selected agent is 'codeact' so we could get prompt pieces from there? I think not...

openhands/server/routes/manage_conversations.py
malhotra5: Yeah def agree

I think I'll convert to jinja templates and move the prompt locations, thanks for the catch!

openhands/server/routes/manage_conversations.py
malhotra5: Ah got it; yes we don't know the agent yet

Since these prompts are agent agnostic and serve as initial user messages I figured we could pull the prompts out elsewhere

openhands/server/routes/manage_conversations.py
enyst: Worth trying, thank you! If we can't, at least we can see clearly why we can't, and see what we can do.

malhotra5:
@enyst I've made the following updates

1. Created jinja templates for the suggested task prompts
2. Moved the prompts to `openhands/integrations/templates` folder

For the second change, I did it for the following reasons

1. The prompts are agent agnostic, but provider specific so I thought it would make sense to store in integrations folder
2. These prompts serve as initial user messages technically (we're just proving an example message is this case), so it felt right to separate it from the agent
3. We're already storing cloud resolver templates in the integrations folder, so I thought I'd organize suggested tasks templates with them


I also manually tested the suggested task workflow and can confirm it works as expected

enyst (approved):
Oh, I love your solution! It's much better than the kitchen sink, err, I mean, `utils`.

Maybe we could have a unit test for `new_conversation` with this, or modification of some existing one?

enyst:
Just to note, I missed this, there is also this potential solution:
- https://github.com/All-Hands-AI/OpenHands/issues/5826

That makes sense too, and I think we talked about something like that before. I'm fine with moving forward this way, and we'll see about a later refactoring perhaps or integration with that feature.

malhotra5:
Yeah would love to refactor with the input based microagent in the future

The issue does however describe the user manually filling out the required input fields for the microagent so may not be totally applicable here, but we'll see based on how it gets implemented!

PR #8117, opened 2025-04-28
log: Respect debug config

SmartManoj:

---

---
Resolves: https://github.com/SmartManoj/Kevin/issues/378#issuecomment-2834838684

### DIFF
--- a/openhands/core/cli.py
+++ b/openhands/core/cli.py
@@ -645,10 +645,11 @@ async def main(loop: asyncio.AbstractEventLoop):
 
     args = parse_arguments()
 
-    logger.setLevel(logging.WARNING)
-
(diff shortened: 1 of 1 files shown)


xingyaoww (approved):

enyst:
Ah, the merge overrode the change here, @SmartManoj .

The CLI experience is just changing. Maybe what we should do at `debug` is write debug-level logs in files only, rather than in console, what do you think?

SmartManoj:
It sounds good to me if the file is pointed out when an error occurs. 

FYI, only warning logs were printed in the CLI.

mamoodi:
Looks like there's no changes associated with this PR now. Closing.

PR #8138, opened 2025-04-29
[Refactor]: Simplify provider info passing

malhotra5:

---

Previously to inform the FE about which providers are set, we would pass every single provider with a boolean value

```python
{
    "github": True,
    "gitlab": False
}
```

This PR simplifies the information passing. Instead we only send the providers which are set

```python
{
    "github": ""
}
```

For now, we're sending an empty string along with each provider. In the future, this string will represent `base_domain` argument which is being implemented in #8130

Manually tested that this works 

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:166cf24-nikolaik   --name openhands-app-166cf24   docker.all-hands.dev/all-hands-ai/openhands:166cf24
```


frontend/src/hooks/query/use-settings.ts
```diff
@@ -54,13 +54,11 @@ export const useSettings = () => {
   React.useEffect(() => {
     if (query.data?.PROVIDER_TOKENS_SET) {
       const providers = query.data.PROVIDER_TOKENS_SET;
-      const setProviders = (
-        Object.keys(providers) as Array<keyof typeof providers>
-      ).filter((key) => providers[key]);
+      const setProviders = Object.keys(providers) as Array<
```
malhotra5, line 57: This changes to FE logic where we expect the BE to only send providers which are currently set (as opposed to every provider tagged with a boolean value)

openhands/server/routes/settings.py
```diff
@@ -42,19 +40,11 @@ async def load_settings(
...
         if provider_tokens:
-            all_provider_types = [provider.value for provider in ProviderType]
-            provider_tokens_types = [provider.value for provider in provider_tokens]
-            for provider_type in all_provider_types:
-                if provider_type in provider_tokens_types:
-                    provider_tokens_set[provider_type] = True
-                else:
-                    provider_tokens_set[provider_type] = False
+            for provider_type, provider_token in provider_tokens.items():
+                if provider_token.token or provider_token.user_id:
```
malhotra5, line 46: This is the BE logic where we only add provider type information if it is currently set

amanape (commented):
What's the context behind this change? I'm not sure that `Record<Provider, "" | null>` is an improvement over `Record<Provider, boolean>` (unless I'm reading this wrong)

`Provider[]` would have been a better change since that seems to be the preferred data to check against (e.g., `providerTokensSet.includes("github");`)

LGTM either way

malhotra5:
I'm working on support for custom `base_domain` for gitlab + github. I'll be attaching a string attribute to every provider in that case.

Currently the auth context converts `Record<Provider, boolean>` to `Provider[]` to simplify checks using `providerTokensSet.includes("github")`

PR #8143, opened 2025-04-29
chore: Update bug_template.yml

SmartManoj:

---
Add actual and expected output fields

---

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_template.yml
+++ b/.github/ISSUE_TEMPLATE/bug_template.yml
@@ -10,7 +10,7 @@ body:
 
   - type: checkboxes
     attributes:
-      label: Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).
+      label: Is there an existing issue with the same bug? (If one exists, thumbs up or comment on the issue instead).
       description: Please check if an issue already exists for the bug you encountered.
       options:
       - label: I have checked the existing issues.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_template.yml
```diff
@@ -20,7 +20,23 @@ body:
...
+      label: Expected Output
+      description: What did you expect to happen?
+    validations:
+      required: true
+
+  - type: textarea
+    id: actual-output
+    attributes:
+      label: Actual Output
+      description: What actually happened?
```
mamoodi, line 39: We actually made the bug template simple and smaller on purpose so that more people feel comfortable filling out bugs. So I don't think we would want to add these to it. The "Provide a description of the issue" should cover this.

mamoodi: Exactly. Which means that having more fields doesn't force people to fill them up. If someone wants to write a good bug so it is addressed, then they can write a good bug with all the information.
Our current templates have been working well for us. No need to change it at this time. Thank you.

SmartManoj: Here, the reproduction steps are also missing, as "describe the bug" and "reproduction steps" are combined.

enyst: I think we may have had in the past "steps to reproduce" separately, and people haven't filled them in most of the time.

As mamoodi said, this may feel like a good idea, but in practice it may not make a difference.

Lets leave it as it is, for now, and if it becomes a problem we can always change it.

enyst:
Thank you for the proposal, but I think we'll leave the bug template as it is. If this becomes an actual issue, we can always revisit it.

PR #8148, opened 2025-04-29
fix(memory): Fix empty string content handling in ConversationMemory

### LINKED ISSUE #8147: Fix empty string content handling in ConversationMemory
## Issue Description

When `ConversationMemory` processes observations into messages, it checks if `assistant_msg.content is not None` to determine whether to include content in the message. However, if `assistant_msg.content` is an empty string (''), this check passes, but then causes validation errors with the Anthropic API because it requires a valid text field.

### Error
```
litellm.BadRequestError: AzureException - -4316: messages.3.content.0.text.text: Field required
```

### Root Cause
The current check `if assistant_msg.content is not None` allows empty strings to be included in the content array. When using the Anthropic API, this causes validation errors as it expects the text field to be valid.

### Solution
Change the condition from `if assistant_msg.content is not None` to `if assistant_msg.content` so that both None and empty strings result in an empty content list, avoiding the validation error.

This is a small but crucial fix for ensuring compatibility with the Anthropic API.

### Images

Image

SDGLBL:
Fix #8147

### DIFF
--- a/openhands/memory/conversation_memory.py
+++ b/openhands/memory/conversation_memory.py
@@ -230,8 +230,8 @@ class ConversationMemory:
             pending_tool_call_action_messages[llm_response.id] = Message(
                 role=getattr(assistant_msg, 'role', 'assistant'),
                 # tool call content SHOULD BE a string
-                content=[TextContent(text=assistant_msg.content or '')]
-                if assistant_msg.content
+                content=[TextContent(text=assistant_msg.content)]
+                if assistant_msg.content and assistant_msg.content.strip()
                 else [],
                 tool_calls=assistant_msg.tool_calls,
             )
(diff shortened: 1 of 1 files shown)


enyst (approved):
Thank you!

Maybe we should do a `strip()` too, what do you think? I am concerned that litellm may do it after we send the call to it, and we'd end up with the same issue.

SDGLBL:
> Thank you!
> 
> Maybe we should do a `strip()` too, what do you think? I am concerned that litellm may do it after we send the call to it, and we'd end up with the same issue.

You are right. I noticed that OpenHands Agent has already implemented this [fix](https://github.com/All-Hands-AI/OpenHands/pull/8153) along with test code.

enyst:
Maybe we should steal its unit test and integrate it in your PR?

PR #8150, opened 2025-04-29
Fix OpenRouter context window exceeded error detection

### LINKED ISSUE #8149: [Bug]: Openrouter context window length exceeded errors are not caught

We have gotten a community report that the following error is encountered in logs:

```
[Agent Controller 38a3bd9c6ada456ab3975d906008ba2b] AgentStateChangedObservation(content='', agent_state='error', reason='BadRequestError: litellm.BadRequestError: Error code: 400 - {\'error\': {\'message\': \'litellm.BadRequestError: OpenrouterException - This endpoint\\\'s maximum context length is 40960 tokens. However, you requested about 42988 tokens (38892 of text input, 4096 in the output). Please reduce the length of either one, or use the "middle-out" transform to compress your prompt automatically.. Received Model Group=openrouter/qwen/qwen3-30b-a3b\\nAvailable Model Group Fallbacks=None\', \'type\': None, \'param\': None, \'code\': \'400\'}}', observation=<ObservationType.AGENT_STATE_CHANGED: 'agent_state_changed'>)
```

The solution to this is to modify the "contextwindowexceedederror" portion of `openhands/controller/agent_controller.py` to detect a string like "please reduce the length of either one" (and maybe add a comment that this is for openrouter).

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Operating System

None

_No response_

neubig:
Fixes #8149

This PR adds detection for OpenRouter context window exceeded errors by checking for the pattern "please reduce the length of either one" in error messages. This allows the agent to properly handle these errors by truncating the conversation history.

Added a test case to verify the fix works correctly.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:58ee00d-nikolaik   --name openhands-app-58ee00d   docker.all-hands.dev/all-hands-ai/openhands:58ee00d
```

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -842,6 +842,8 @@ class AgentController:
                     'contextwindowexceedederror' in error_str
                     or 'prompt is too long' in error_str
                     or 'input length and `max_tokens` exceed context limit' in error_str
+                    or 'please reduce the length of either one'
+                    in error_str  # For OpenRouter context window errors
                     or isinstance(e, ContextWindowExceededError)
                 ):
                     if self.agent.config.enable_history_truncation:
--- a/tests/unit/test_agent_controller.py
+++ b/tests/unit/test_agent_controller.py
@@ -3,7 +3,11 @@ from unittest.mock import ANY, AsyncMock, MagicMock, patch
 from uuid import uuid4
 
 import pytest
-from litellm import ContentPolicyViolationError, ContextWindowExceededError
+from litellm import (
+    BadRequestError,
+    ContentPolicyViolationError,
+    ContextWindowExceededError,
+)
 
 from openhands.controller.agent import Agent
 from openhands.controller.agent_controller import AgentController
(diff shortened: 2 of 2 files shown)


openhands/controller/agent_controller.py
```diff
@@ -842,6 +842,8 @@ async def _step(self) -> None:
                     'contextwindowexceedederror' in error_str
                     or 'prompt is too long' in error_str
                     or 'input length and `max_tokens` exceed context limit' in error_str
+                    or 'please reduce the length of either one'
+                    in error_str  # For OpenRouter context window errors
```
rbren, line 846: oof this is a really bad way to catch errors. Are there not more fine-grained error types we can filter for here?

neubig:
I think getting this fixed upstream in litellm is the correct way in the long term.

enyst:
> I think getting this fixed upstream in litellm is the correct way in the long term.

Absolutely, but I'd note, litellm does the same thing. I do think that we could remove the first. I've hesitated though, because there are multiple routes to get it (from provider, from litellm proxy, from openrouter, from other proxies) and it matters for litellm to catch it or not. I'll test all I can think of, then remove it.

PR #8152, opened 2025-04-29
refactor: Refactor pause/resume functionality and improve state handling in CLI

bashwara:

- Enable support for agent pause/resume functionality alongside confirmation mode.
- Minor styling changes for the CLI command completer and agent pause/resume messages.

---
- Refactor and enhance state handling in the CLI by leveraging local state variables alongside the agent state.

---

n/a

### DIFF
--- a/openhands/core/cli.py
+++ b/openhands/core/cli.py
@@ -101,7 +101,7 @@ async def run_session(
 
     sid = str(uuid4())
     is_loaded = asyncio.Event()
-    is_paused = asyncio.Event()
+    is_paused = asyncio.Event()  # Event to track agent pause requests
 
     # Show runtime initialization message
     display_runtime_initialization_message(config.runtime)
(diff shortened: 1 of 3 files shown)


enyst (approved):
LGTM, thank you.

It's not fully clear to me why we need to track both `is_paused` and `AgentState.PAUSED`, but it seems like the interaction with prompt confirmation may be tricky. Maybe best to go with it.

bashwara:
> LGTM, thank you.
> 
> It's not fully clear to me why we need to track both `is_paused` and `AgentState.PAUSED`, but it seems like the interaction with prompt confirmation may be tricky. Maybe best to go with it.

Thanks for taking a look at this @enyst

We use the `is_paused` flag in the CLI alongside the agent's `AgentState.PAUSED` state to handle a specific timing issue. 

Sometimes, especially when confirmation mode is on, there's a small delay before the agent fully enters the `PAUSED` state. During this delay, another action might request specific user input (like a confirmation). However, before the user can respond, the agent state finishes transitioning to `PAUSED`, which triggers its own default input prompt. Since the CLI has only one input channel (the tty), this default prompt overwrites the original confirmation prompt, making it impossible to answer. The `is_paused` flag allows the CLI to immediately recognize that a pause is intended (as it's state change happens immediately when the pause key combination is triggered) and ignore any subsequent prompts during that brief transition, ensuring the original confirmation request isn't lost.

PR #8159, opened 2025-04-29
[Fix]: Allow partial repo object to initialize new conversation

malhotra5:

- Only requires repo name + git provider for repo information to start new conversation via openhands api

---

This PR is a fix for OpenHands API. We do the following

1. Define a new type `PartialRepository` which only requires `full_name` and `git_provider` info. This simplifies the api calling to create a new conversation and reduced the burden to specify information like `repo_id`

2. Adds a new trigger type `OPENHANDS_API`

---

### DIFF
--- a/openhands/server/routes/manage_conversations.py
+++ b/openhands/server/routes/manage_conversations.py
@@ -13,7 +13,7 @@ from openhands.events.stream import EventStream
 from openhands.integrations.provider import (
     PROVIDER_TOKEN_TYPE,
 )
-from openhands.integrations.service_types import Repository, SuggestedTask
+from openhands.integrations.service_types import PartialRepository, Repository, SuggestedTask
 from openhands.runtime import get_runtime_cls
 from openhands.server.data_models.conversation_info import ConversationInfo
 from openhands.server.data_models.conversation_info_result_set import (
@@ -59,7 +59,7 @@ class InitSessionRequest(BaseModel):
 async def _create_new_conversation(
     user_id: str | None,
     git_provider_tokens: PROVIDER_TOKEN_TYPE | None,
-    selected_repository: Repository | None,
+    selected_repository: Repository | PartialRepository | None,
     selected_branch: str | None,
     initial_user_msg: str | None,
     image_urls: list[str] | None,
@@ -96,7 +96,12 @@ async def _create_new_conversation(
         raise MissingSettingsError('Settings not found')
 
     session_init_args['git_provider_tokens'] = git_provider_tokens
-    session_init_args['selected_repository'] = selected_repository
+
+    repository = None
+    if isinstance(selected_repository, PartialRepository):
+        repository = selected_repository.convert_to_repo_model()
+
+    session_init_args['selected_repository'] = repository
     session_init_args['selected_branch'] = selected_branch
     conversation_init_data = ConversationInitData(**session_init_args)
     logger.info('Loading conversation store')
--- a/openhands/server/routes/manage_conversations.py
+++ b/openhands/server/routes/manage_conversations.py
@@ -29,9 +29,11 @@ from openhands.server.shared import (
 )
 from openhands.server.types import LLMAuthenticationError, MissingSettingsError
 from openhands.server.user_auth import (
+    get_auth_type,
     get_provider_tokens,
     get_user_id,
 )
+from openhands.server.user_auth.user_auth import AuthType
 from openhands.server.utils import get_conversation_store
 from openhands.storage.conversation.conversation_store import ConversationStore
 from openhands.storage.data_models.conversation_metadata import (
@@ -166,6 +168,7 @@ async def new_conversation(
     data: InitSessionRequest,
     user_id: str = Depends(get_user_id),
     provider_tokens: PROVIDER_TOKEN_TYPE = Depends(get_provider_tokens),
+    auth_type: AuthType | None = Depends(get_auth_type)
 ):
     """Initialize a new session or join an existing one.
 
@@ -184,6 +187,9 @@ async def new_conversation(
     if suggested_task:
         initial_user_msg = suggested_task.get_prompt_for_task()
         conversation_trigger = ConversationTrigger.SUGGESTED_TASK
+
+    if auth_type == AuthType.BEARER:
+        conversation_trigger = ConversationTrigger.OPENHANDS_API
 
     try:
         # Create conversation with initial message
--- a/openhands/server/routes/manage_conversations.py
+++ b/openhands/server/routes/manage_conversations.py
@@ -50,7 +50,7 @@ app = APIRouter(prefix='/api')
 
 class InitSessionRequest(BaseModel):
     conversation_trigger: ConversationTrigger = ConversationTrigger.GUI
-    selected_repository: Repository | None = None
+    selected_repository: Repository | PartialRepository | None = None
     selected_branch: str | None = None
     initial_user_msg: str | None = None
     image_urls: list[str] | None = None
@@ -69,9 +69,9 @@ async def _create_new_conversation(
     conversation_trigger: ConversationTrigger = ConversationTrigger.GUI,
     attach_convo_id: bool = False,
 ):
-    print("trigger", conversation_trigger)
+    
     logger.info(
-        'Creating conversation',
+        f'Creating conversation with trigger {conversation_trigger}',
         extra={'signal': 'create_conversation', 'user_id': user_id, 'trigger': conversation_trigger.value},
     )
     logger.info('Loading settings')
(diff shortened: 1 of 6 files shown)


enyst (changes requested):
Thank you for the fix, the API is awesome and well worth designing it nicely!

At first sight, I do have a little doubt on a couple things, so please let's take a moment to figure it out (added inline)

openhands/server/routes/manage_conversations.py
```diff
@@ -176,10 +184,16 @@ async def new_conversation(
...
 
+    if isinstance(selected_repository, PartialRepository):
+        selected_repository = selected_repository.convert_to_repo_model()
+
     if suggested_task:
         initial_user_msg = suggested_task.get_prompt_for_task()
         conversation_trigger = ConversationTrigger.SUGGESTED_TASK
 
+    if auth_type == AuthType.BEARER:
+        conversation_trigger = ConversationTrigger.OPENHANDS_API
```
enyst, line 195: This seems to imply that the API is not accessible on localhost. Are we sure about this, I thought it is?

openhands/server/routes/manage_conversations.py
```diff
@@ -176,10 +184,16 @@ async def new_conversation(
     suggested_task = data.suggested_task
     conversation_trigger = data.conversation_trigger
 
+    if isinstance(selected_repository, PartialRepository):
+        selected_repository = selected_repository.convert_to_repo_model()
```
enyst, line 188: Could we do this differently, it seems a bit strange for the idea of exposing this API to seemingly use a different object, idk.

How about, we make the 2 attributes of `Repository` mandatory, the rest optional, and document that? If memory doesn't fail me, maybe swagger has some annotations or something we can use for it to figure out which is required attribute and which is optional?

openhands/server/routes/manage_conversations.py
malhotra5: Fairly sure openhands api is not accessible locally (where you generate an api key and auth via a bearer)

Locally you can ping routes but there's no auth involved (and its not a long lived server so its more of a hack than a supported usage method)

enyst: I think maybe we misunderstand each other. I'm not talking about connecting to the deployment on all-hands.dev.

This is the same API deployed on local server. (the web UI connects to it too.) Same can any client.

malhotra5: Ah I see yeah that makes sense; let me backtrack a bit and make sure I'm understanding the question correctly

> This seems to imply that the API is not accessible on localhost

I'm assuming that the confusion is due to 

1. The explicit check for bearer auth
2. The trigger name is `OPENHANDS_API`, which isn't true because many clients make use of `OPENHANDS_API` (web UI for instance)


What I was trying to imply with this statement is that API requests with bearer auth are definitely connecting all-hands.dev via an api key (we explicitly override the default trigger `GUI` only when this condition is true)

May I can rename this to something like `REMOTE_API_KEY`?

enyst: Maybe, it does seem better... I wonder, if we document properly the endpoint, maybe we don't need it? I mean, if we don't do anything special, and all endpoints are available anyway via api key too, not just this one, do we need it? 🤔

malhotra5: Agree we should def document the endpoint super well

For more context, we track the trigger only when creating new conversations. It helps us determine usage patterns for saas (are users creating conversations using resolver, gui, suggested task, remote api key, slack, etc)

enyst: Yes, it makes sense. Although you log the request, but maybe it's easier. Lets go with that.

malhotra5:
Update; in this PR we 

1. We reuse the existing `Repository` object and make `full_name` the only non-optional field
2. We determine the git provider for a repo object if it isn't provided
3. Add a new conversation trigger

malhotra5:
We are now doing the following this PR

1. Renamed `selected_repository` in `new_conversation` to be `repository` instead to simplify api calling

2. Updated `new_conversation` route to accept repository as `str` type instead of `Repository` object
    - This makes it easier to call the API
    - Standardized selected repository to always be `str` in the entire code, and is plumbed through into the runtime 

3. The `clone_or_init_repo` method accepts only a `str` type for repo now.
    - A provider handler figures out which provider the selected repo belongs to
    - All clients (FE, CLI) that use this method pass `str` type values for repository param now
     
4. Updated OpenAPI with the new api spec

5. Added a new trigger `REMOTE_API_KEY`

openhands/server/session/agent_session.py
```diff
@@ -147,7 +147,7 @@ async def start(
 
             repo_directory = None
             if self.runtime and runtime_connected and selected_repository:
-                repo_directory = selected_repository.full_name.split('/')[-1]
+                repo_directory = selected_repository.split('/')[-1]
```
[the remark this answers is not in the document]
malhotra5: > API accepts a string like `foo/bar`.

Yes. `selected_repository.full_name` has always been a string like `foo/bar`. 

> How do we differentiate between github and gitlab

I've added a new method in [`ProviderHandler` ](https://github.com/All-Hands-AI/OpenHands/pull/8159/files#diff-2a846e5207a45fb7ac74801ee3cbef95b358d01d2ef7fdd64d8a85bf4ccd8e1d). For all the existing git providers the user has currently set, it checks whether the repo exists.  

It's used last minute in [`clone_or_init_repo` ](https://github.com/All-Hands-AI/OpenHands/pull/8159/files#diff-dfad47e657f893ab3f62500695323fea807f6be59e21457264b09bdf9d509ceaR324-R326)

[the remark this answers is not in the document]
malhotra5: I've added the following changes

1. We verify the repo exists before creating a new conversation (that way we don't create a conversation only to throw a runtime error when failing to clone the repo)

2. Added optional param `git_provider` to distinguish between multiple configured providers; if not specified, we'll check against all providers configured to see whether the repo exists, and chose the provider where it does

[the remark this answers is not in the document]
enyst: Do we actually get an AuthError if the repo clone doesn't work? I think it's executed in `bash`, and as far as I recall, we don't really get actual error information from execution.

For the record, `git clone` on a repo that already exists also objects in the result message, it says something like, directory already exists and not empty so can't do.

That, for us, is not an error. 😅 We just continue normally.

Also [this test](https://github.com/All-Hands-AI/OpenHands/pull/8159/files#diff-a45a88aa4207aeecd68e859564a6dd42a2104774f3587ebde0587e3d1c0d10e7R319), if I'm reading it right, seems to show there's no AuthError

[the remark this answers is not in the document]
malhotra5: Ah we only throw an `AuthError` if the user provides a repo but we are unable to clone it due to insufficient token permissions

If the repo already exists in the workspace, and we attempt to clone the repo again (assuming sufficient token permissions) it will be a no-op

> Do we actually get an AuthError if the repo clone doesn't work

In summary we throw an AuthError if the user has selected a repo for which they token have a valid token which can access it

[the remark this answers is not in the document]
enyst: Ah, I see, we try to raise one.

> that way we don't create a conversation only to throw a runtime error when failing to clone the repo

This is what I responded to. I don't know if we throw a runtime error when failing to clone the repo.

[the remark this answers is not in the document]
malhotra5: Ahh I see I messed up my description, sorry about that. Yes you're right we don't throw an error when failing to clone the repo

In this PR we raise an error specifically when we are supposed to clone a repo but don't have a token which can allow us to (that way the user doesn't initialize a conversation with the expectation of having the repo, but in reality finds that its missing from the workspace)

malhotra5:
@enyst would love another 👀 from you whenever you get the time

Merge is blocked until then

openhands/server/routes/manage_conversations.py
```diff
@@ -122,7 +126,7 @@ async def _create_new_conversation(
             title=conversation_title,
             user_id=user_id,
             github_user_id=None,
-            selected_repository=selected_repository.full_name
+            selected_repository=selected_repository
             if selected_repository
             else selected_repository,
```
enyst, line 131:
```suggestion
            selected_repository=selected_repository,
```

it seems this conditional has become a no-op?

enyst (approved):
Thank you!

openhands/server/routes/manage_conversations.py
malhotra5: Thank you!

PR #8161, opened 2025-04-29
Update anyio to 4.9.0

raymyers:
This PR updates anyio from 4.8.0 to 4.9.0 using Poetry version 2.1.1 for dependency resolution and locking.

There were issues with MemoryObjectItemReceiver errors, as flagged by @chuckbutkus 

From the release notes:
> Changed EndOfStream raised from MemoryObjectReceiveStream.receive() to leave out the AttributeError from the exception chain which was merely an implementation detail and caused some confusion

https://anyio.readthedocs.io/en/stable/versionhistory.html

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:892bf21-nikolaik   --name openhands-app-892bf21   docker.all-hands.dev/all-hands-ai/openhands:892bf21
```


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.1.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.2 and should not be changed by hand.
```
enyst, line 1: I think dependabot may disagree and change it back 😅

It's ok though, just a patch version without extensive changes won't be a problem IMO.

enyst (approved):

poetry.lock
raymyers: Good catch, openhands does this by default now, maybe a repo.md update

chuckbutkus (approved):

PR #8168, opened 2025-04-29
Fix ValueError when latest_event_id is undefined

rbren:
## Description
This PR fixes an error that occurs when the `latest_event_id` query parameter is set to "undefined" in the socket connection. The error was:

```
ValueError: invalid literal for int() with base 10: "undefined"
```

## Changes
- Added a try-except block to handle non-integer values for `latest_event_id`
- When an invalid value is encountered, it defaults to -1 and logs a warning

## Testing
The change is minimal and handles a specific edge case without changing the existing behavior for valid inputs.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:b5f5c3d-nikolaik   --name openhands-app-b5f5c3d   docker.all-hands.dev/all-hands-ai/openhands:b5f5c3d
```

### DIFF
--- a/openhands/server/listen_socket.py
+++ b/openhands/server/listen_socket.py
@@ -49,7 +49,7 @@ async def connect(connection_id: str, environ):
     try:
         latest_event_id = int(latest_event_id_str)
     except ValueError:
-        logger.warning(
+        logger.debug(
             f'Invalid latest_event_id value: {latest_event_id_str}, defaulting to -1'
         )
         latest_event_id = -1
--- a/openhands/server/listen_socket.py
+++ b/openhands/server/listen_socket.py
@@ -46,7 +46,6 @@ async def connect(connection_id: str, environ):
     logger.info(f'sio:connect: {connection_id}')
     query_params = parse_qs(environ.get('QUERY_STRING', ''))
     latest_event_id_str = query_params.get('latest_event_id', [-1])[0]
-    # Handle case when latest_event_id is 'undefined' or other non-integer values
     try:
         latest_event_id = int(latest_event_id_str)
     except ValueError:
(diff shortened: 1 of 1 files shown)


openhands/server/listen_socket.py
```diff
@@ -45,7 +45,13 @@ def create_provider_tokens_object(
 async def connect(connection_id: str, environ):
     logger.info(f'sio:connect: {connection_id}')
     query_params = parse_qs(environ.get('QUERY_STRING', ''))
-    latest_event_id = int(query_params.get('latest_event_id', [-1])[0])
+    latest_event_id_str = query_params.get('latest_event_id', [-1])[0]
+    # Handle case when latest_event_id is 'undefined' or other non-integer values
```
rbren, line 49:
```suggestion
```

tofarr (approved):
🍰

openhands/server/listen_socket.py
```diff
@@ -45,7 +45,14 @@ def create_provider_tokens_object(
 async def connect(connection_id: str, environ):
     logger.info(f'sio:connect: {connection_id}')
     query_params = parse_qs(environ.get('QUERY_STRING', ''))
-    latest_event_id = int(query_params.get('latest_event_id', [-1])[0])
+    latest_event_id_str = query_params.get('latest_event_id', [-1])[0]
+    try:
+        latest_event_id = int(latest_event_id_str)
+    except ValueError:
+        logger.warning(
```
enyst, line 52:
```suggestion
        logger.debug(
```

enyst: Or discard that, maybe a warning is better, we should figure out the underlying cause

PR #8189, opened 2025-04-30
bugfix for #8187 (infinite loop when delegating)

### LINKED ISSUE #8187: [Bug]: Agent delegation seems to be broken

As of at least 0.33 (and I believe much earlier), agent delegation seem to be broken.

As a minimal test to reproduce the error, I checked out v0.35 and added the following lines to CodeActAgent, right after the `/exit` message handling in `step()`:

```python
if latest_user_message and latest_user_message.content.strip() == '/exit':
            return AgentFinishAction()
elif latest_user_message and latest_user_message.content.strip().startswith(
    '/delegate'
):
    print("Trying to delegate to ReadOnlyAgent")
    return AgentDelegateAction(
        agent='ReadOnlyAgent',
        inputs=latest_user_message.content.strip().removeprefix('/delegate'),
    )
```

When I use this "slash command" to try to ask the agent to delegate the task, the agent gets stuck indefinitely in the state "Agent is running task":

Image

Here are the contents of the backend server logs when I send the /delegate command:

```
18:59:39 - openhands:INFO: agent_controller.py:586 - [Agent Controller e64e57d44bad40939019532394bdc00a] Setting agent(CodeActAgent) state from AgentState.AWAITING_USER_INPUT to AgentState.RUNNING
18:59:39 - openhands:INFO: agent_controller.py:774 - [Agent Controller e64e57d44bad40939019532394bdc00a] Agent not stepping because of pending action: RecallAction (id=11)
Trying to delegate to ReadOnlyAgent
19:00:33 - openhands:INFO: standalone_conversation_manager.py:106 - Conversation e64e57d44bad40939019532394bdc00a connected in 0.01733994483947754 seconds
```

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Operating System

None

_No response_

cshimmin:

Fix issue with infinite LLM loop when delegated agent returns.

---

 * Ensure that tool metadata is correctly associated with AgentDelegationObservation upon delegate finish.
 * fix discrepency between obs.content and obs.output['content'] usage
 * slightly modify tool response text to better indicate to caller agent that the delegation occurred successfully.

---
Closes #8187


openhands/controller/agent_controller.py
```diff
@@ -750,9 +748,13 @@ def end_delegate(self) -> None:
...
 
-            # emit the delegate result observation
-            obs = AgentDelegateObservation(outputs=delegate_outputs, content=content)
-            self.event_stream.add_event(obs, EventSource.AGENT)
+        content = f'Delegated agent finished with result:\n\n{content}'
+
+        # emit the delegate result observation
+        obs = AgentDelegateObservation(outputs=delegate_outputs, content=content)
+        if self.delegateAction:
+            obs.tool_call_metadata = self.delegateAction.tool_call_metadata
```
enyst, line 756: Ah, I see. I do wonder though, could we do this slightly differently, without storing a `delegateAction` in the controller?

It's a tiny detail, sorry for nitpicking! It's because I really feel that we aim to make delegates as similar as possible with "regular" agent runs. There needs to be some starting processing, and some ending processing, sure, but apart from that I would hope there's... no difference. Agents and even their controllers don't need to know they're delegates, they should work the same way as if they weren't.

Did that make sense?

I think at this point, we're in the parent controller, and the corresponding AgentDelegateAction is findable in `self.agent.history` in reverse immediately? Maybe at `-1`, or maybe we shouldn't count on that and iterate backwards a bit more, wouldn't it be there?

What do you think?

cshimmin: No worries, I was initially thinking of doing the same thing you suggested. But then I saw a vestigial reference to `delegateAction` and thought maybe this was just something that got broken in a refactor:

https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/controller/agent_controller.py#L764

I'll rework it to use the history :+1:

enyst: Ah, oops. That `delegateAction` looks like a leftover from my old refactoring, sorry! And it wasn't used there either (and delegates did work). We can clean it up if you'd like.

cshimmin: fixed this in 389812c

enyst (approved):
Thank you! ❤️

PR #8191, opened 2025-05-01
Fix VS Code URL for remote access

### LINKED ISSUE #7191: [Bug]: "Open in VS Code" button always uses localhost URL regardless of actual server address

## Issue Description
When clicking "Open in VS Code" button, the system always generates a URL starting with `http://localhost`. This works fine locally but fails when accessing the OpenHands web interface remotely through IP address.

## Current Behavior
- System returns VS Code URL with hardcoded `localhost` hostname
- Remote access fails to open VS Code web editor
- Issue traced to `docker_runtime` file where `vscode_url` property always returns localhost-based URL

## Expected Behavior
VS Code URL should use the correct URL when accessing remotely

## Technical Details
Problem located in `docker_runtime` file where `vscode_url` is hardcoded to use `localhost`

```
    @property
    def vscode_url(self) -> str | None:
        token = super().get_vscode_token()
        if not token:
            return None

        vscode_url = f'http://localhost:{self._vscode_port}/?tkn={token}&folder={self.config.workspace_mount_path_in_sandbox}'
        return vscode_url
```

## Proposed Solutions
1. Simple fix but not elegant: Replace 'localhost' with `window.location.hostname` in open-hands-ts file
2. Maybe better approach: Create new `/vscode-config` endpoint that returns:
   - vscode_port
   - token
   - folder
   Then dynamically construct VS Code URL on frontend

Which solution would be more appropriate? Are there any other approaches we should consider?

### OpenHands Installation

Docker command in README

### OpenHands Version

0.28.1

### Operating System

None

_No response_

xingyaoww:

Fixes an issue where the "Open in VS Code" button would always use localhost in the URL, causing it to fail when accessing the application remotely. The fix ensures that the VS Code URL uses the correct hostname when accessed from a remote location.

---

This PR adds a utility function that transforms VS Code URLs by replacing "localhost" with the current window hostname when they differ. The function is applied in three places where VS Code URLs are used:

1. In conversation.tsx when opening VS Code from the conversation view
2. In vscode-tab.tsx when loading the VS Code URL in the tab
3. In conversation-card.tsx when opening VS Code from the conversation card

The implementation includes comprehensive tests to verify the functionality works correctly in various scenarios.

---

Fixes #7191

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/095512c6601443c0ac4945d4ae84e9d7)

### DIFF
--- a/frontend/src/components/features/conversation-panel/conversation-card.tsx
+++ b/frontend/src/components/features/conversation-panel/conversation-card.tsx
@@ -16,6 +16,7 @@ import { BaseModal } from "../../shared/modals/base-modal/base-modal";
 import { RootState } from "#/store";
 import { I18nKey } from "#/i18n/declaration";
 import { selectSystemMessage } from "#/state/chat-slice";
+import { transformVSCodeUrl } from "#/utils/vscode-url-helper";
 
 interface ConversationCardProps {
   onClick?: () => void;
@@ -117,7 +118,10 @@ export function ConversationCard({
         const data = await response.json();
 
         if (data.vscode_url) {
-          window.open(data.vscode_url, "_blank");
+          const transformedUrl = transformVSCodeUrl(data.vscode_url);
+          if (transformedUrl) {
+            window.open(transformedUrl, "_blank");
+          }
         }
         // VS Code URL not available
       } catch (error) {
--- a/frontend/src/routes/conversation.tsx
+++ b/frontend/src/routes/conversation.tsx
@@ -40,6 +40,7 @@ import { clearFiles, clearInitialPrompt } from "#/state/initial-query-slice";
 import { RootState } from "#/store";
 import { displayErrorToast } from "#/utils/custom-toast-handlers";
 import { useDocumentTitleFromState } from "#/hooks/use-document-title-from-state";
+import { transformVSCodeUrl } from "#/utils/vscode-url-helper";
 
 function AppContent() {
   useConversationConfig();
@@ -159,7 +160,12 @@ function AppContent() {
                           );
                           const data = await response.json();
                           if (data.vscode_url) {
-                            window.open(data.vscode_url, "_blank");
+                            const transformedUrl = transformVSCodeUrl(
+                              data.vscode_url,
+                            );
+                            if (transformedUrl) {
+                              window.open(transformedUrl, "_blank");
+                            }
                           }
                         } catch (err) {
                           // Silently handle the error
--- a/frontend/src/routes/vscode-tab.tsx
+++ b/frontend/src/routes/vscode-tab.tsx
@@ -5,6 +5,7 @@ import { useConversation } from "#/context/conversation-context";
 import { I18nKey } from "#/i18n/declaration";
 import { RootState } from "#/store";
 import { RUNTIME_INACTIVE_STATES } from "#/types/agent-state";
+import { transformVSCodeUrl } from "#/utils/vscode-url-helper";
 
 function VSCodeTab() {
   const { t } = useTranslation();
@@ -27,7 +28,8 @@ function VSCodeTab() {
         const data = await response.json();
 
         if (data.vscode_url) {
-          setVsCodeUrl(data.vscode_url);
+          const transformedUrl = transformVSCodeUrl(data.vscode_url);
+          setVsCodeUrl(transformedUrl);
         } else {
           setError(t(I18nKey.VSCODE$URL_NOT_AVAILABLE));
         }
--- a/frontend/src/utils/__tests__/vscode-url-helper.test.ts
+++ b/frontend/src/utils/__tests__/vscode-url-helper.test.ts
@@ -1,0 +1,61 @@
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { transformVSCodeUrl } from "../vscode-url-helper";
+
+describe("transformVSCodeUrl", () => {
+  const originalWindowLocation = window.location;
+
+  beforeEach(() => {
+    // Mock window.location
+    Object.defineProperty(window, "location", {
+      value: {
+        hostname: "example.com",
+      },
+      writable: true,
+    });
+  });
+
+  afterEach(() => {
+    // Restore window.location
+    Object.defineProperty(window, "location", {
+      value: originalWindowLocation,
+      writable: true,
+    });
+  });
+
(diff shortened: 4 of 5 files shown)


enyst:
I'm a bit confused why we need to read it from the window, when we have the correct vscode url in the runtime.

@openhands-agent-exp Read this PR carefully. Understand what problem it's solving.

Explore the code to understand why the vscode url is not accessible in frontend and what would it take to make it accessible.

This task is analysis only. You do NOT need to modify or fix anything this time.

Guide:
- we have the vscode url in the runtime classes, look it up and its usage
- do we ever send it to the frontend code?
- if yes, where is it in the frontend code
- if not, what would it take to make it available in the components in this PR?

It's possible that we can send it nicely via some existing API route, if so, what is the best way to do it, considering the patterns in the codebase. It's also possible that we cannot, though, maybe I'm making a wrong assumption that the code disproves? If so, find the explanation. Do NOT assume the answer to this question, this is an analysis task and we want to understand the current code first. Think about it and plan, don't execute.

< IMPORTANT >
Please send your analysis when you are done, in a comment to this PR, and in your Finish tool, so I can see it.

You don't need to fix anything, you don't need to run any code, just read the code and write your response. I took care of changes and so don't make changes.
< /IMPORTANT >

openhands-agent:
[OpenHands](https://github.com/All-Hands-AI/OpenHands) started fixing the pr! You can monitor the progress [here](https://github.com/All-Hands-AI/OpenHands/actions/runs/14771942050).

enyst:
Ah, for some reason we now punish openhands-agent worse - the resolver gives an error and doesn't write a comment - when it doesn't make code changes. But it did listen.

> AgentThinkAction(thought="After analyzing the code, I can now provide a comprehensive analysis of the VS Code URL issue:\nCurrent Implementation Analysis:\n\n1. Backend (Docker Runtime):\n- The VS Code URL is generated in the docker_runtime.py file's vscode_url property\n- It always uses 'localhost' in the URL: `f'http://localhost:{self._vscode_port}/?tkn={token}&folder={self.config.workspace_mount_path_in_sandbox}'`\n- The port and token are dynamically generated\n- The URL is exposed via a FastAPI endpoint `/api/conversations/{conversation_id}/vscode-url`\n\n2. Frontend:\n- The frontend fetches the VS Code URL from the API endpoint\n- There's already a URL transformation function (transformVSCodeUrl) that:\n  - Detects if the URL uses 'localhost' and the current page isn't on localhost\n  - Replaces 'localhost' with window.location.hostname in that case\n\n3. Current Solution:\n- The frontend already has a solution in place to handle this issue\n- However, this is not the most elegant approach because:\n  a. The backend is generating incorrect URLs that need to be fixed by the frontend\n  b. The frontend has to maintain this transformation logic\n  c. The backend could be generating the correct URL in the first place\n\n4. Better Solution:\n- The backend should be responsible for generating the correct URL\n- The docker_runtime.py should use the local_runtime_url from the config instead of hardcoding 'localhost'\n- This is better because:\n  a. The backend already has access to the correct URL via config.sandbox.local_runtime_url\n  b. The URL transformation logic can be removed from the frontend\n  c. The backend maintains a single source of truth for URLs\n  d. It follows the same pattern used elsewhere in the code (e.g., api_url property)\n\n5. Implementation Path:\n- Modify the vscode_url property in docker_runtime.py to use the local_runtime_url from config\n- The URL can be constructed using the hostname from local_runtime_url and the vscode_port\n- Remove the frontend URL transformation logic once the backend change is verified\n- This maintains backward compatibility while fixing the root cause\n\n6. Potential Challenges:\n- Need to ensure local_runtime_url is always set correctly\n- May need to handle cases where local_runtime_url includes a port\n- Need to verify this works with all deployment scenarios (local, remote, etc.)\n\n7. Code Pattern:\n- The codebase already uses config.sandbox.local_runtime_url for the api_url\n- Following this same pattern for vscode_url would be consistent with existing code", action=<ActionType.THINK: 'think'>)

> 07:51:22 - openhands:INFO: resolve_issue.py:346 - AgentThinkObservation(content='Your thought has been logged.', observation=<ObservationType.THINK: 'think'>)

> 07:51:34 - openhands:INFO: resolve_issue.py:346 - AgentFinishAction(final_thought="I have completed the analysis of the VS Code URL issue. Here's my response to the PR:\n\nThe current PR proposes to fix the VS Code URL issue by transforming the URL in the frontend, but this is not the best approach. Here's why:\n\n1. The backend (docker_runtime.py) already has access to the correct URL via `config.sandbox.local_runtime_url`\n2. We're already using this config for the `api_url` property in the same file\n3. The frontend shouldn't need to transform URLs - this should be handled by the backend\n\nInstead of the current solution, we should:\n\n1. Modify the `vscode_url` property in docker_runtime.py to use `config.sandbox.local_runtime_url` instead of hardcoding 'localhost'\n2. This follows the same pattern used by the `api_url` property in the same file\n3. Once this is done, we can remove the URL transformation logic from the frontend\n\nThis is a cleaner solution because:\n- Single source of truth for URLs (the backend config)\n- Consistent with existing code patterns\n- Removes unnecessary transformation logic from frontend\n- Handles all deployment scenarios (local, remote) in one place\n\nThe code already shows this pattern works - we use `self.config.sandbox.local_runtime_url` for the `api_url` property, and it handles both local and remote scenarios correctly. We should follow the same pattern for the VS Code URL.", task_completed='true', outputs={}, thought='', action=<ActionType.FINISH: 'finish'>)

rbren (approved):

xingyaoww:
@enyst 

> I'm a bit confused why we need to read it from the window, when I think we have the correct vscode url in the runtime... 🤔

I think it happens when a user:
1. start the openhands docker image on a machine with IP a.b.c.d
2. expose this machine to public or some regional internet
3. access the hosted OpenHands with a.b.c.d IP which is not known the backend I believe, and find that FE will redirect them to localhost

I do think "Modify the vscode_url property in docker_runtime.py to use config.sandbox.local_runtime_url instead of hardcoding 'localhost'." would be a more fundamental approach, but we are not really using this URL anywhere else and asking user to set a config in our already complex enough config system is probably not a happy thing to do?
So maybe we can sacrifice the code complexity a little bit here in exchange for better user experience & one fewer config in our config system?

enyst (approved):
Ah I see. Thank you!

PR #8193, opened 2025-05-01
Add Lunr search plugin with multi-language support

Shimada666:

image
image

Adds local search functionality to the documentation site. You can now search documentation content directly within the site using the search bar. Search is supported for English, Chinese (中文), French (Français), Japanese (日本語), and Portuguese (Português).

---
This PR integrates the docusaurus-lunr-search plugin to enable client-side search capabilities.
Added docusaurus-lunr-search to the Docusaurus plugins configuration.
Configured the plugin to support multiple languages (en, zh, fr, ja, pt) by enabling the languages option, leveraging the lunr-languages package. This ensures search works correctly for non-English content, especially languages like Chinese that require specific tokenization.
Added @node-rs/jieba as a dependency, which is required by lunr-languages for Chinese word segmentation.

---


Shimada666:
By the way, I noticed that while our frontend uses npm, the docs directory seems to have lockfile for both npm and yarn (containing both package-lock.json and yarn.lock). This can be quite confusing, making it unclear which package manager should be used for the docs.
Furthermore, whenever I add a dependency in package.json, I have to update both yarn.lock and package-lock.json. If I don't update both, the process fails because the docs CI relies on package-lock.json.
image

Could we potentially standardize on a single package manager going forward? For example, we could decide to exclusively use npm. (Although, personally, I actually prefer pnpm.)

amanape:
We should default to `npm`, the lockfile might've been pushed by whichever developer used `yarn` locally. Easy miss but agree with you.

Shimada666:
@amanape 
I agree with your point! Algolia is indeed good, but it needs some setup that might require help from core developers, which is a bit troublesome. That's why I chose local search - it's simple and works well. If needed in the future, we can switch to Algolia anytime.
Regarding npm, I will send another PR to specify which package manager to use and remove unnecessary lock files.
Thank you for your review!

PR #8195, opened 2025-05-01
feat: Enable optional streaming completion via ENABLE_STREAM_COMPLETION environment variable

Hambaobao:
This PR introduces an option to toggle streaming completion by using the environment variable `ENABLE_STREAM_COMPLETION`.
- When `ENABLE_STREAM_COMPLETION=true`, the code calls the completion API with `stream=True` and assembles the response from streamed chunks.
- When `ENABLE_STREAM_COMPLETION` is unset or set to any value other than 'true', it defaults to `stream=False` (non-streaming mode).

This feature can help reduce API timeout and error rates in some scenarios by leveraging streaming responses.

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -13,15 +13,15 @@ with warnings.catch_warnings():
     warnings.simplefilter('ignore')
     import litellm
 
-from litellm import ChatCompletionMessageToolCall, ModelInfo, PromptTokensDetails
+from litellm import ChatCompletionMessageToolCall
 from litellm import Message as LiteLLMMessage
+from litellm import ModelInfo, PromptTokensDetails
 from litellm import completion as litellm_completion
 from litellm import completion_cost as litellm_completion_cost
-from litellm.exceptions import (
-    RateLimitError,
-)
+from litellm import stream_chunk_builder
+from litellm.exceptions import RateLimitError
 from litellm.types.utils import CostPerToken, ModelResponse, Usage
-from litellm.utils import create_pretrained_tokenizer
+from litellm.utils import CustomStreamWrapper, create_pretrained_tokenizer
 
 from openhands.core.exceptions import LLMNoResponseError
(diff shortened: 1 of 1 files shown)


enyst (changes requested):
Thank you for looking into this! Streaming responses is a nice feature and as far as I can tell, people expect it so I love to see some work towards it!

I'm not sure sending a boolean to litellm is enough to make it work, though. Could you please tell, how do we test it?

Some questions:
- this is nice if the user sees it, they currently don't, though. Could we try to make it available in some way, to see it work?
- we have an old code here in the [streaming llm](https://github.com/All-Hands-AI/OpenHands/blob/e5f25726c0e414f118c6fdecdf2368d1502d2b86/openhands/llm/streaming_llm.py#L11) mixin, this was an attempt to make it work in a way that is keeping the streaming complexity outside LLM class itself. What do you think, could we use that approach or should we remove it?
- small detail, this could be configured in [the configuration](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/core/config/llm_config.py) which will take care of the env var too

Please let me know your thoughts!

Hambaobao:
Thanks for the feedback!

To be honest, I’m also not entirely sure what the best way is to test the streaming functionality in a visible way. My motivation for this change came from frequently hitting timeout errors and other API errors when using non-streaming requests, so I added this modification to improve stability. I’ve been running it for a while on my side and can confirm that this change has been working reliably.

I agree that it would make sense to support `enable_stream_completion` in the config file to control litellm’s behavior — I’m happy to make that adjustment if needed.

Regarding the old `streaming llm mixin` code, it seems like that part was meant for displaying the stream in the frontend, while my change focuses more on improving API stability in headless mode.

enyst:
Oh, I didn't think of that! I don't fully understand it... 🤔, either, sorry for being dense. Could you tell, what LLM and particularly, what provider do you use?

PR #8196, opened 2025-05-01
Fix: Include RecallObservation in events sent to frontend from ENVIRONMENT source

xingyaoww:

Fixes an issue where memory recall events were not appearing in the UI until after a page refresh. Now, when the agent recalls information from memory, these events will immediately appear in the conversation UI.

---

This PR fixes the issue where events pushed to eventstream in the Memory class do not show up timely in the frontend UI.

**Problem:**
Events pushed to eventstream in the Memory class with EventSource.ENVIRONMENT were not being sent to the frontend UI immediately, only appearing after a page refresh. Meanwhile, events added to eventstream in openhands/controller/agent_controller.py (with EventSource.AGENT) were showing up timely.

**Root Cause:**
The Session._on_event method was only sending events with EventSource.ENVIRONMENT to the frontend if they were instances of CmdOutputObservation or AgentStateChangedObservation. RecallObservation events from Memory were being filtered out.

**Solution:**
Update the Session._on_event method to also include RecallObservation in the list of observation types that are sent to the frontend when they come from the ENVIRONMENT source.

I've tested that it is working:
image

---

N/A

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:762c560-nikolaik   --name openhands-app-762c560   docker.all-hands.dev/all-hands-ai/openhands:762c560
```


xingyaoww:
@OpenHands can you make sure npm run test:coverage passes

raymyers:
Nice! Did you happen to check if there are other observation types that might be missing there?

raymyers (approved):

openhands/server/session/session.py
```diff
@@ -213,7 +214,7 @@ async def _on_event(self, event: Event):
             await self.send(event_to_dict(event))
         # NOTE: ipython observations are not sent here currently
         elif event.source == EventSource.ENVIRONMENT and isinstance(
-            event, (CmdOutputObservation, AgentStateChangedObservation)
+            event, (CmdOutputObservation, AgentStateChangedObservation, RecallObservation)
         ):
             # feedback from the environment to agent actions is understood as agent events by the UI
             event_dict = event_to_dict(event)
```
enyst, line 220: We may want this to be sent as `USER` EventSource below, to be on the right side of the chat box? The rest are `AGENT` source.

I think I noticed it is 'user' when we display it, right under the user message.

enyst (approved):

xingyaoww:
> Did you happen to check if there are other observation types that might be missing there?

@raymyers  Yeah i think only this observation is impacted so i'd assume no :)

openhands/server/session/session.py
xingyaoww: Though i feel the "RecallObservation" is generated via the microagent files from the *environment*, maybe make it a little bit inappropriate for "USER"?

enyst: Either way it's fine with me. I just think after refresh, it shows up as user?

So we may want to be consistent, simply.

xingyaoww: Yeah, but i imagine as we got a new design back, we might visualize it differently than what we are doing now :)

So probably good to keep it environment since we could possibly remove it from chat side

PR #8213, opened 2025-05-01
[Refactor]: Split settings and secrets stores

malhotra5:

---

This PR splits secrets store from the settings store. We maintain two separate stores, which is a pre-req for #8068

We do the following in this PR

1. Make a new file store for secrets `FileSecretsStore`
2. Move the old definition `SecretsStore` into its own data_model file as `UserSecrets`
3. Make dedicated routes for all custom secrets + provider token related logic, separating it entirely from the save settings route
4. Make a new service class for the FE `SecretsService` and use it to interact with the new BE routes for user secrets
5. Ensure backwards compatibility by migrating any existing secrets from the settings object and moving it to the secret store. This will ensure the user's provider tokens aren't lost if this PR makes it into a new release

 
---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:a8517cb-nikolaik   --name openhands-app-a8517cb   docker.all-hands.dev/all-hands-ai/openhands:a8517cb
```


frontend/src/hooks/mutation/use-add-git-providers.ts
```diff
@@ -0,0 +1,21 @@
...
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: ({
+      providers,
+    }: {
+      providers: Record<Provider, ProviderToken>;
+    }) => SecretsService.addGitProvider(providers),
+    onSuccess: async () => {
+      await queryClient.invalidateQueries({ queryKey: ["settings"] });
```
enyst, line 15: Is this intended, do we still have to invalidate 'settings' on provider change?

tests/unit/test_settings_store_functions.py
```diff
@@ -20,20 +25,43 @@ async def get_settings_store(request):
...
+        client = TestClient(app)
+        yield client
+
+
+@contextmanager
+def patch_file_secrets_store():
+    store = FileSecretsStore(InMemoryFileStore())
+    with patch(
+        'openhands.storage.settings.file_secrets_store.FileSecretsStore.get_instance',
+        AsyncMock(return_value=store),
```
enyst, line 46: I think we could create a filestore on a temporary file, simpler than a mock and tests the right thing here?

malhotra5: Ah I might be wrong; isn't this mock is necessary because I'm testing a FastAPI route

The [route param calls `get_secrets_store`](https://github.com/All-Hands-AI/OpenHands/pull/8213/files#diff-a4aaaf769c9bd7d5ea4b9e3ed31877853f030e3456812fa7f3c097ef625fdedfR64-R67), so we mock the call for it

openhands/server/routes/secrets.py
```diff
@@ -0,0 +1,289 @@
...
+            if secret_id not in custom_secrets:
+                return JSONResponse(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    content={'error': f'Secret with ID {secret_id} not found'},
+                )
+
+            # Remove the secret
+            custom_secrets.pop(secret_id)
+
+            # Create a new SecretStore that preserves provider tokens
```
enyst, line 271:
```suggestion
            # Create a new UserSecrets that preserves provider tokens and remaining secrets
```

malhotra5: Thank you!

tests/unit/test_settings_store_functions.py
enyst: I was thinking at something like this: https://github.com/All-Hands-AI/OpenHands/blob/976019ce11e7113b8c11fefe5a3a49c335bfb3af/tests/unit/test_is_stuck.py#L48

It seemed like it would work to simplify a bit, but I could be wrong.

openhands/server/routes/secrets.py
```diff
@@ -0,0 +1,292 @@
...
+    except Exception as e:
+        logger.warning(f'Something went wrong storing git providers: {e}')
+        return JSONResponse(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            content={'error': 'Something went wrong storing git providers'},
+        )
+
+
+@app.post('/unset-provider-tokens', response_model=dict[str, str])
+async def unset_provider_tokens(
```
enyst, line 119: The API changed, so we'll probably want to update swagger

enyst (approved):
Thank you so much for this! It's been a lot of work, and it does seem to clear the view quite a lot.

Maybe at some point we will be able to move the credentials like llm api keys to the secrets store as well.

IMHO we should merge rather sooner than later, and let it show its surprises, if it has any, over the weekend, but I'd love to know if Robert has concerns.

malhotra5:
Did a small QA and went well! Will be merging thanks for the thorough feedback folks!

PR #8227, opened 2025-05-02
Fix websocket error message handling

rbren:
## Description
Fixes a regression where websocket messages with `{error: true, message: "..."}` format were not being properly handled, resulting in "unknown message type received" errors instead of displaying the actual error message.

## Changes
- Added a specific handler in `handleAssistantMessage` to check for messages with `error: true` property and display the error message to the user.

## Testing
Built and verified that the frontend code compiles successfully.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:eab4eb2-nikolaik   --name openhands-app-eab4eb2   docker.all-hands.dev/all-hands-ai/openhands:eab4eb2
```

### DIFF
--- a/frontend/src/services/actions.ts
+++ b/frontend/src/services/actions.ts
@@ -152,7 +152,7 @@ export function handleAssistantMessage(message: Record<string, unknown>) {
     handleObservationMessage(message as unknown as ObservationMessage);
   } else if (message.status_update) {
     handleStatusMessage(message as unknown as StatusMessage);
-  } else if (message.error === true && typeof message.message === 'string') {
+  } else if (message.error) {
     // Handle error messages from the server
     trackError({
       message: message.message,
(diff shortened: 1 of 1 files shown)


frontend/src/services/actions.ts
```diff
@@ -152,6 +152,18 @@ export function handleAssistantMessage(message: Record<string, unknown>) {
     handleObservationMessage(message as unknown as ObservationMessage);
   } else if (message.status_update) {
     handleStatusMessage(message as unknown as StatusMessage);
+  } else if (message.error === true && typeof message.message === 'string') {
```
rbren, line 155:
```suggestion
  } else if (message.error) {
```

xingyaoww (approved):

rbren:
@openhands fix the lint errors. Convert the message to a string if it's not already a string

Issue #7647, opened 2025-04-02
[Bug]: SWE-Bench Openhands connection issue


PreetJhanglani:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Tried to run run_infer.sh on swe bench verified and facing the connection reset by peers error. Idk how to fix it? 

### OpenHands Installation

Development workflow

### OpenHands Version

0.30.1

### Operating System

None

### Logs, Errors, Screenshots, and Additional Context

Instance django__django-12155 - 2025-04-01 22:55:08,568 - ERROR - ----------
Error in instance [django__django-12155]: [Errno 104] Connection reset by peer. Stacktrace:
Traceback (most recent call last):
  File "/home/preet/.cache/pypoetry/virtualenvs/openhands-ai-TYgoX0xy-py3.12/lib/python3.12/site-packages/httpx/_transports/default.py", line 101, in map_httpcore_exceptions
    yield
  File "/home/preet/.cache/pypoetry/virtualenvs/openhands-ai-TYgoX0xy-py3.12/lib/python3.12/site-packages/httpx/_transports/default.py", line 250, in handle_request
    resp = self._pool.handle_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/preet/.cache/pypoetry/virtualenvs/openhands-ai-TYgoX0xy-py3.12/lib/python3.12/site-packages/httpcore/_sync/connection_pool.py", line 256, in handle_request
    raise exc from None
  File "/home/preet/.cache/pypoetry/virtualenvs/openhands-ai-TYgoX0xy-py3.12/lib/python3.12/site-packages/httpcore/_sync/connection_pool.py", line 236, in handle_request
    response = connection.handle_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/preet/.cache/pypoetry/virtualenvs/openhands-ai-TYgoX0xy-py3.12/lib/python3.12/site-packages/httpcore/_sync/connection.py", line 103, in handle_request
    return self._connection.handle_request(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/preet/.cache/pypoetry/virtualenvs/openhands-ai-TYgoX0xy-py3.12/lib/python3.12/site-packages/httpcore/_sync/http11.py", line 136, in handle_request
    raise exc
[...]

PR #8232, opened 2025-05-03
fix: retry if Read Error

SmartManoj:

---

---
Fixes https://github.com/All-Hands-AI/OpenHands/issues/7647

### DIFF
--- a/openhands/runtime/impl/action_execution/action_execution_client.py
+++ b/openhands/runtime/impl/action_execution/action_execution_client.py
@@ -48,7 +48,7 @@ from openhands.utils.tenacity_stop import stop_if_should_exit
 
 def _is_retryable_error(exception):
     return isinstance(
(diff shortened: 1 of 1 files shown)


openhands/runtime/impl/action_execution/action_execution_client.py
```diff
@@ -48,7 +48,7 @@
 
 def _is_retryable_error(exception):
     return isinstance(
-        exception, (httpx.RemoteProtocolError, httpcore.RemoteProtocolError)
+        exception, (httpx.ReadError, httpx.RemoteProtocolError)
```
rbren, line 51: why did we remove `httpcore.RemoteProtocolError`

SmartManoj: Because `httpx` should wrap `httpcore.RemoteProtocolError` as `httpx.RemoteProtocolError`, making the explicit httpcore check redundant.

PR #8233, opened 2025-05-03
Make tool call json decode error recoverable

li-boxuan:

Make tool call json decode error recoverable

---

I was using `o3-mini` and got the following error:

> json.decoder.JSONDecodeError: Unterminated string starting at: line 1 column 165 (char 164)

I inspected the llm logs, and indeed saw incomplete string in function call arguments. Not sure if this is a litellm problem, or OpenAI problem. Anyways, IMO, this shouldn't be a RuntimeError, which would cause the agent to stop. This PR changes it to a recoverable error so that agent could retry.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:2a1faf5-nikolaik   --name openhands-app-2a1faf5   docker.all-hands.dev/all-hands-ai/openhands:2a1faf5
```


enyst (approved):
Thank you! On `FunctionCallValidationError` we create an ErrorObs, right? This makes sense.

li-boxuan:
> On FunctionCallValidationError we create an ErrorObs, right

Yep. https://github.com/All-Hands-AI/OpenHands/blob/de175dcc8757d377cb6a8aa45eba4ae835e87c44/openhands/controller/agent_controller.py#L835-L843

PR #8242, opened 2025-05-03
Simplify workspace mounting with SANDBOX_VOLUMES

### LINKED ISSUE #8239: Add support for custom volume mounts in DockerRuntime
Currently, the DockerRuntime only supports mounting the workspace directory into the container. However, there are use cases where users need to mount additional directories from the host machine into the container, such as:

- Mounting read-only model directories (like pre-trained ML models)
- Accessing shared data directories
- Mounting temporary storage for downloads or generated content
- Using host machine's cached dependencies to improve performance

This is particularly useful when working with large datasets or models that already exist on the host machine and should be accessible from within the runtime environment.

I suggest implementing a new environment variable CUSTOM_VOLUMES that allows users to specify additional volume mounts in the following format:
```
CUSTOM_VOLUMES=/host/path1:/container/path1:rw,/host/path2:/container/path2:ro
```

Each mount specification consists of three parts:

- Host path: The path on the host machine
- Container path: Where to mount it inside the container
- Access mode: rw for read-write or ro for read-only

Multiple mounts would be separated by commas. The DockerRuntime would parse this environment variable and add these additional mounts when creating the container.

For example, users could run:
```shell
docker run -it -d \
    -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.36-nikolaik \
    -e LOG_ALL_EVENTS=true \
    -e CUSTOM_VOLUMES="/opt/openhands/workspace:/workspace:rw,/mnt/data1:/data1:ro" \
    -v ~/.openhands-state:/.openhands-state \
    -p 3000:3000 \
    --name openhands-app \
    docker.all-hands.dev/all-hands-ai/openhands:0.36
```

**Additional context**

xingyaoww:

This PR introduces a simpler and more flexible way to mount directories into the OpenHands Docker runtime. Users can now mount multiple directories with different access permissions (read-write or read-only) using the new `SANDBOX_VOLUMES` environment variable. This is particularly useful for accessing large datasets, pre-trained models, or reference code that should be available to the agent but kept separate from the main workspace.

---

This PR simplifies the workspace mounting process in OpenHands Docker runtime by:

1. Introducing a new `SANDBOX_VOLUMES` environment variable that uses a standard Docker-like format: `host_path:container_path[:mode]`
2. Supporting multiple mount points separated by commas: `/path1:/workspace:rw,/path2:/data:ro`
3. Allowing explicit read-only mounts with the `:ro` suffix
4. Maintaining backward compatibility with the older `WORKSPACE_*` variables (marked as deprecated)
5. Updating documentation in both English and Japanese to reflect these changes
6. Adding comprehensive tests for the new functionality

Design decisions:
- The first mount in `SANDBOX_VOLUMES` is considered the primary workspace for backward compatibility
- The agent works in `/workspace` by default, so users should mount their project directory there
- Read-only reference materials should be mounted to a different path (like `/data`)
- The format follows Docker's standard volume mounting syntax for familiarity

---

Closes #8239

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:79c0e86-nikolaik   --name openhands-app-79c0e86   docker.all-hands.dev/all-hands-ai/openhands:79c0e86
  -p 3000:3000 \
  -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:bca19f4-nikolaik \
  -e SANDBOX_VOLUMES=/path/to/your/code:/workspace:rw,/path/to/reference/data:/data:ro \
  --name openhands-app-bca19f4 \
  docker.all-hands.dev/all-hands-ai/openhands:bca19f4
```


Shimada666:
image

Great!

xingyaoww:
@OpenHands pls run "make lint" to fix linter

tests/unit/test_docker_runtime.py
```diff
@@ -69,3 +69,157 @@ def test_container_not_stopped_when_keep_runtime_alive_true(
...
+                container_path = parts[1]
+                # Default mode is 'rw' if not specified
+                mount_mode = parts[2] if len(parts) > 2 else 'rw'
+
+                volumes[host_path] = {
+                    'bind': container_path,
+                    'mode': mount_mode,
+                }
+
+    # Assert that both mounts were processed correctly
```
xingyaoww, line 147: @OpenHands we should call the actual code that handles these here in this test

happyherp:
> The old WORKSPACE_* environment variables are still supported for backward compatibility but marked as deprecated.

Thanks. I was using those in my scripts.

enyst:
What if we just called them "volumes"? In env, that means `SANDBOX_VOLUMES`, which seems pretty good, and shorter. 😅

xingyaoww:
hmm good idea.

@OpenHands, pls move RUNTIME_VOLUMES to SandboxConfig and rename it to "volumes". And update all the documentation to use "SANDBOX_VOLUMES" instead of "RUNTIME_VOLUMES"

mamoodi:
Once this merges, there will be a period where the docs are wrong because the feature will only be available on main.

tests/unit/test_config.py
```diff
@@ -389,6 +406,47 @@ def test_defaults_dict_after_updates(default_config):
     assert defaults_after_updates == initial_defaults
 
 
+# Tests for SANDBOX_VOLUMES have replaced the RUNTIME_MOUNT tests
+
```
xingyaoww, line 410:
```suggestion
```

xingyaoww:
@OpenHands can you read the git diff against main, and then use github API to update the PR description here. Make sure you follow the pr template in this repo.

xingyaoww:
@OpenHands please review @neubig's lastest comment and work on that

neubig (commented):
Looks great to me other than some stale documentation, commented above!

neubig (approved):
Thank you!

PR #8243, opened 2025-05-03
Fix VS Code iframe reload issue

xingyaoww:

This PR fixes an issue with the VS Code tab in the OpenHands UI where the iframe would reload every time a user switched to or from the tab. Now, the VS Code iframe stays loaded in the background, eliminating the waiting time when switching tabs.

---

The PR implements a new approach to tab management in the frontend:

1. Created a new TabContent component that keeps all tab content loaded but only displays the active one
2. Implemented a VSCodeProvider context to manage the VS Code URL state globally
3. Modified the conversation component to use the new TabContent approach while maintaining compatibility with React Router

Design decisions:
- Used React Context to share the VS Code URL state across components
- Kept the Outlet component hidden but present to maintain React Router functionality
- Used absolute positioning with display:none/block to toggle visibility without unmounting components

This approach ensures that the VS Code iframe remains loaded in the DOM even when not visible, eliminating the reload delay when switching tabs.

---

N/A

### DIFF
--- a/frontend/src/components/layout/tab-content.tsx
+++ b/frontend/src/components/layout/tab-content.tsx
@@ -1,0 +1,58 @@
+import React from "react";
+import { useLocation } from "react-router";
+import { lazy, Suspense } from "react";
+import { LoadingSpinner } from "../shared/loading-spinner";
+import { VSCodeProvider } from "../../routes/vscode-tab";
+
+// Lazy load all tab components
+const EditorTab = lazy(() => import("../../routes/editor"));
+const BrowserTab = lazy(() => import("../../routes/browser-tab"));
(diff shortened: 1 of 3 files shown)


frontend/src/components/layout/tab-content.tsx
amanape, line 1: Not so react router friendly but I guess it is the best solution I can think of as well :/

amanape (changes requested):
Only request to use useQuery to fetch vscode url

xingyaoww:
@OpenHands please help me address these reviews

amanape (approved):
LGTM, just need to run address fixes shown the TS compilation

xingyaoww:
@OpenHands make sure make lint, npm run build and npm run test all passes

PR #8244, opened 2025-05-03
Small refactor to improve (CodeAct)Agent extensibility

cshimmin:
* make prompt_manager an `@property` of base `Agent` class, which can be overridden in children
* make `response_to_actions` a proper instance method rather than an attribute

Both of these changes have the effect that `__init__` behavior should no longer need to be redefined in most child implementations, for example in `ReadOnlyAgent`.

* Override `_get_tools` and `prompt_manager` in `ReadOnlyAgent` so that it can re-use the parent class `__init__`.


enyst (commented):
Thank you for this! It looks good to me, I'll give it a run tomorrow.

(Some agents don't have tools or a prompt manager, they should be unaffected as far as I can tell at a quick look.)

xingyaoww (approved):
This overall LGTM! But i'll wait for @enyst 's final confirmation here 😄

enyst:
I played with it a bit, switching agents and nagging them. 😅

PR #8245, opened 2025-05-03
Feature: Add custom agents via config

cshimmin:

Allows custom `Agent` implementations from arbitrary python packages to be integrated via config.toml, without having to modify OpenHands source directly.

---

This commit adds a single, optional configuration property, `classpath`, to `AgentConfig`.
If provided, the specified python class will be imported and registered via `Agent.register()` at configuration load time.

The following example usage has been added to config.template.toml:

```
[agent.CustomAgent]
# Example: use a custom agent from a different package
# This will be automatically be registered as a new agent named "CustomAgent"
classpath = "my_package.my_module.MyCustomAgent"
```

### DIFF
--- a/openhands/core/config/utils.py
+++ b/openhands/core/config/utils.py
@@ -28,6 +28,7 @@ from openhands.core.config.sandbox_config import SandboxConfig
 from openhands.core.config.security_config import SecurityConfig
 from openhands.storage import get_file_store
 from openhands.storage.files import FileStore
+from openhands.utils.import_utils import get_impl
 
 JWT_SECRET = '.jwt_secret'
 load_dotenv()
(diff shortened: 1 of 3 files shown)


cshimmin:
Hmm, I don't really understand the CI failure, or how it would relate to this patch. Something about a Connection Reset Error... is it possible this is a spurious result? Please advise!

enyst:
You're right, I think something is flaky there, and it seems like it may have got a little worse in the past couple of days. Please don't worry about it. Timeout or connection refused / reset are definitely not related.

openhands/core/config/utils.py
```diff
@@ -582,6 +607,7 @@ def load_app_config(
     load_from_toml(config, config_file)
     load_from_env(config, os.environ)
     finalize_config(config)
+    register_custom_agents(config)
```
enyst, line 610: Out of curiosity, could you please tell, have you tried registering it in the agenthub init, after the others, or maybe it didn't work for some reason?

cshimmin: Registering agents the init file of course works but the point is to avoid having to edit OpenHands source code to add/change an agent.

enyst: I was thinking like
- load custom agent from config file
- in agenthub init, register the hardcoded ones, then the custom ones

Maybe it wouldn't get a circular dependency? But I could be wrong

cshimmin: Well the problem is that you can't ensure that the config is loaded before `agenthub.__init__` is. Or rather, maybe you can but it seems like it would be very awkward.

cshimmin: @enyst ping? 🙏

enyst: Ah, you're right, actually the order of imports is definitely first [agenthub](https://github.com/All-Hands-AI/OpenHands/blob/b06bd277baac3dd6323d16ef2429fcc43554348e/openhands/core/main.py#L7)

Seems like maybe we should change this, but it's definitely a question for another time!

enyst (approved):
Thank you!

PR #8253, opened 2025-05-04
Add 401 response handling to logout and refresh browser for saas mode

chuckbutkus:
This PR is for SAAS only.  It more cleanly handles a session timeout.

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c4cdf3a-nikolaik   --name openhands-app-c4cdf3a   docker.all-hands.dev/all-hands-ai/openhands:c4cdf3a
```


frontend/src/api/open-hands-axios.ts
```diff
@@ -1,5 +1,30 @@
...
 });
+
+// Add response interceptor to handle 401 Unauthorized responses
+openHands.interceptors.response.use(
+  (response) => response,
+  async (error) => {
+    // Check if the error is a 401 Unauthorized
+    if (error.response && error.response.status === 401) {
+      // Get app mode from localStorage or default to "oss"
+      const appMode = localStorage.getItem("appMode") || "oss";
```
rbren, line 15: we shouldn't use localstorage for this. It's in the app config, I think set on `window`

frontend/src/hooks/query/use-config.ts
```diff
@@ -7,7 +7,16 @@ export const useConfig = () => {
...
   return useQuery({
     queryKey: ["config"],
-    queryFn: OpenHands.getConfig,
+    queryFn: async () => {
+      const config = await OpenHands.getConfig();
+
+      // Store app mode in localStorage for use in 401 handler
+      if (config && config.APP_MODE) {
+        localStorage.setItem("appMode", config.APP_MODE);
+      }
```
rbren, line 16: can't the 401 handler just use `config`? Feels strange to pass information around via localStorage

frontend/src/api/open-hands-axios.ts
```diff
@@ -1,5 +1,30 @@
...
+
+      // Check if user is logged in by checking if providersAreSet is true
```
enyst, line 17: Nit: maybe this comment and below shouldn't suggest that the user is *logged in* based on a boolean in localStorage. Sooner or later it will convince someone, an LLM perhaps :)

For example, the comment could say "Check if user may have been logged in before" or perhaps better "Check if we should clear storage for this user"

frontend/src/api/open-hands-axios.ts
chuckbutkus: Updated to use React hooks.

frontend/src/hooks/query/use-config.ts
chuckbutkus: Updated to use React hooks.

frontend/src/api/open-hands-axios.ts
chuckbutkus: Things got move around and this comment is no longer used.

amanape (changes requested):
Hold this PR, not perfect. I have a week-old draft addressing this problem here: https://github.com/All-Hands-AI/OpenHands/pull/8094. I just need to test it. I'll try to get this in today

amanape:
Resolved in #8094

Issue #7554, opened 2025-03-27
UI should visualize how much of the context window is used by the agent


c3-ali:
**What problem or use case are you trying to solve?**
To better understand how much of the context window is used and how much is available.

**Describe the UX of the solution you'd like**
A UI widget to show "% of context size used"

**Additional context**
LLMs follow instructions better when their context size is small. Seeing how much of the context size is used give us better sense of how to prompt the llm so it uses the context window efficiently.

rbren:
@openhands please implement this feature request. The modal triggered by the "Display Cost" button should show:
* total input tokens for the conversation
* total output tokens for the conversation
* total input tokens in the most recent prompt
* total output tokens in the most recent prompt
* percent of context window used in most recent prompt

AutoLTX:
@rbren Are you working on this currently or just trying to let openhands-agent do the work? It seems the PR is failed.  I have bandwidth after a week. You can assign it to me if you don't prepare to finish it during these period.

neubig:
Hey @AutoLTX , we're still working on some backend changes that should make this easier: https://github.com/All-Hands-AI/OpenHands/pulls/csmith49

But that should be in very soon, and once it is we'd love some help on the frontend!

neubig:
Hey @AutoLTX , the backend should be fixed up now, so if you want to take a look at the `CondensationAction`, we could find ways to display it in the frontend. Would you like to take a look?

AutoLTX:
ACK. Thanks for remind @neubig. Let me understand the CondensationAction first and take a look at the implementation

c3-ali:
@AutoLTX thanks for looking into this one. For the ux, if possible, i think visualizing the % used as a progress bar would be nice so we visually get a sense how much context length is remaining.
look at https://mintlify.s3.us-west-1.amazonaws.com/factory/images/tutorial/step-8.webp on how factory ai is doing it
This is how claude.ai visualize it for projects:
<img width="506" alt="Image" src="https://github.com/user-attachments/assets/ca8511c6-c46b-4dbf-b469-b6188070f3d7" />

neubig:
Hi @AutoLTX , just checking in on this issue, have you had a chance to take a look at it?

AutoLTX:
[...]

PR #8267, opened 2025-05-04
Display context window usage status in UI

AutoLTX:

---
Display context usage window usage in UI.
image

---
https://github.com/All-Hands-AI/OpenHands/issues/7554


openhands/llm/metrics.py
```diff
@@ -130,15 +136,24 @@ def add_token_usage(
...
         cache_read_tokens: int,
         cache_write_tokens: int,
+        context_window: int,
         response_id: str,
     ) -> None:
         """Add a single usage record."""
+
+        # Token each turn for calculating context usage.
+        per_turn_token = (
+            prompt_tokens + completion_tokens + cache_read_tokens + cache_write_tokens
```
xingyaoww, line 146: I think we just need to add prompt+completion_tokens?

xingyaoww (commented):
Left a comment! otherwise LGTM

openhands/llm/metrics.py
AutoLTX: I'm confused about this part during the implementation in fact -- whether cache is included in the context window usage.

I tried to search related docs and asked chatgpt previously. It seems cache token is not included in the litellm total tokens, and has additional charging logic. But it really contains necessary information and needs to be considered in the full context window usage? Cache content is a prefix to the prompt.

image

"Prompt caching is a powerful feature that optimizes your API usage by allowing resuming from specific prefixes in your prompts. This approach significantly reduces processing time and costs for repetitive tasks or prompts with consistent elements."

- https://ai.google.dev/gemini-api/docs/caching?lang=python
- https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching?utm_source=chatgpt.com

enyst: I think prompt_tokens includes cached tokens

AutoLTX: Prompt_tokens are non-cached prompt tokens? This is the explanation from LiteLLM docs.
https://docs.litellm.ai/docs/completion/prompt_caching?utm_source=chatgpt.com

AutoLTX: This is the logic for cline. It counts cache tokens in context. https://github.com/cline/cline/blob/2792e7698f06eb335970fd1e5ad194553eaa49b7/src/core/context/context-management/ContextManager.ts#L127

xingyaoww: Based on my understanding, I think "prompt_tokens" are the tokens for input, regardless of whether they are cached or not?

AutoLTX: You are correct. I mixed the litellm cache and LLM API cache together.

- For litellm, there is a cache mechanism, which means there are already not expired results in the memory and can be reused directly. It won't be send to the real LLM API call. And for this part, returning the cached results directly. Prompt_tokens is the real input.
- Cache in API call is another thing.

Sorry for the misunderstanding.

xingyaoww (approved):
Tested & it works :) Thanks a lot for this!

image

PR #8283, opened 2025-05-05
[feature] add optional extended field to agent config

cshimmin:

Added support optional `extended` property to agent configuration, for implementations that require additional config not in the default schema.
For example:

```toml
[agents.CustomAgent.extended]
foo = "bar"
```

---

This is the minimal change to get some customizability in agent configuration. However, longer term I would love to have support for custom config schemas too, similar to how the `AgentConfig.condenser` supports a union of different config types for different condensers.

In particular, I would propose having the Agent class include a class-level attribute (e.g. `Agent.config_class = AgentConfig`) that can be overridden in derived implementations, and then the config loader will chose the appropriate config schema at loading time. If the maintainers are open to this more structured approach, I'd be happy to contribute a PR. I don't believe it negates the usefulness of the unstructured `extended` field, though.

### DIFF
--- a/openhands/core/config/agent_config.py
+++ b/openhands/core/config/agent_config.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from pydantic import BaseModel, Field, ValidationError
 
 from openhands.core.config.condenser_config import CondenserConfig, NoOpCondenserConfig
+from openhands.core.config.extended_config import ExtendedConfig
 from openhands.core.logger import openhands_logger as logger
 
 
@@ -34,6 +35,8 @@ class AgentConfig(BaseModel):
     condenser: CondenserConfig = Field(
         default_factory=lambda: NoOpCondenserConfig(type='noop')
     )
+    extended: ExtendedConfig = Field(default_factory=lambda: ExtendedConfig({}))
+    """Extended configuration for the agent."""
 
     model_config = {'extra': 'forbid'}
 


rbren:
This seems reasonable! I wonder if the agent-specific options could/should just go on the top level though, rather than in an `extended` field?

rbren (approved):

enyst:
Maybe we could make `extra` attributes allowed?

cshimmin:
@rbren the issue with that is that the top level config is not provided to the agent. The agent ctor only gets the AgentConfig as input. 

@enyst i think alllow_extra defeats some of the purpose of having pydantic backing the config. For instance a user could make a typo in a property, like `llm_comfig` and it will silently accept it while confusing the user as to why their config isn’t working as expected.

enyst:
I'm guessing agent config is what rbren meant, maybe not fully on top.

> For instance a user could make a typo in a property, like llm_comfig and it will silently accept it while confusing the user as to why their config isn’t working as expected.

You're right, but it's the same for the 'extended' options. We could give a warning, that's not silent.

The downside I see is when working on a branch which introduces new options, and switch back to main, then it fails. Then I comment out the new options, and when I switch back again, it doesn't work, until I find it and grmph, it was commented again.

On the other hand, for running evals, we have the opposite intention: to fail early, rather than run 500 instances with unintended behavior. I think that's why we have `forbid` for LLMConfig.

But for AgentConfig, we already [re-create it from cli params + defaults](https://github.com/All-Hands-AI/OpenHands/blob/13ca75c8cbd72686b1a3bf4c6f325571aa5d4a1b/evaluation/benchmarks/multi_swe_bench/eval_infer.py#L406), not from toml.

@xingyaoww WDYT?

xingyaoww (approved):
This seems good to me :)

Since we are hard-coding agent config in evaluation, I think this shouldn't causing issue in evaluation.

And yes, I think it will be pretty nice to have AgentConfig to accept a union of different type for future customization needs 🙏

cshimmin:
@enyst To be clear, I would not recommend using this `extended` field in "official" agents. It's just an "escape hatch" that lets you punch through the rigid schema if you need to (e.g. when implementing a custom 3rd party agent), without exposing all "normal" users to the possible issues of allow_extra. You have to go out of your way to use the `extended` field, and then you have no validation guarantees for whatever you put in there.

But this is also why I also propose allowing agents to define their own custom schema, so that new or 3rd party agents can also benefit from config validation. I'll probably not get to it right away but would be happy to contribute this functionality eventually!

PR #8289, opened 2025-05-05
Add return type annotations to methods in session directory

neubig:
This PR adds return type annotations to methods in the openhands/server/session directory to comply with strict mypy checks. The changes are focused only on the session directory as requested.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:961ea97-nikolaik   --name openhands-app-961ea97   docker.all-hands.dev/all-hands-ai/openhands:961ea97
```

### DIFF
--- a/openhands/server/session/session.py
+++ b/openhands/server/session/session.py
@@ -73,7 +73,7 @@ class Session:
         self.loop = asyncio.get_event_loop()
         self.user_id = user_id
 
-    async def close(self):
+    async def close(self) -> None:
         if self.sio:
             await self.sio.emit(
                 'oh_event',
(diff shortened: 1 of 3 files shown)


openhands/server/session/session.py
```diff
@@ -253,7 +253,7 @@ async def dispatch(self, data: dict):
                     return
         self.agent_session.event_stream.add_event(event, EventSource.USER)
 
-    async def send(self, data: dict[str, object]):
+    async def send(self, data: dict[str, object]) -> None:
```
malhotra5, line 256:
```suggestion
    async def send(self, data: dict[str, object]) -> bool:
```

I think the method called at the end makes this `bool` return type

malhotra5 (approved):

neubig:
@openhands does the lint.yml workflow pass locally? If not, please fix the remaining linting errors.

PR #8290, opened 2025-05-05
Add proper type annotations to files in openhands/server/routes directory

neubig:
This PR adds proper type annotations to files in the openhands/server/routes directory to comply with strict mypy checks. The changes include:

- Added typing imports (Dict, List, Optional, Union, Any) to all route files
- Fixed return type annotations for route handlers
- Added proper type annotations for function parameters
- Fixed missing return statement in files.py
- Changed dict[str, str] to Dict[str, str] throughout for compatibility
- Changed Union | None patterns to Optional[Union] for better compatibility

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:08815fc-nikolaik   --name openhands-app-08815fc   docker.all-hands.dev/all-hands-ai/openhands:08815fc
```


openhands/server/routes/files.py
```diff
@@ -1,4 +1,5 @@
 import os
+from typing import Any, Dict, List, Optional, Union
```
malhotra5, line 2: We should use built typing instead

- `dict`
- `list`
- `| None` instead of `Optional`
- `|` instead of `Union`

openhands/server/routes/files.py
```diff
@@ -83,7 +86,7 @@ async def list_files(request: Request, path: str | None = None):
 
     file_list = [f for f in file_list if f not in FILES_TO_IGNORE]
 
-    async def filter_for_gitignore(file_list, base_path):
+    async def filter_for_gitignore(file_list: List[str], base_path: str) -> List[str]:
```
malhotra5, line 89:
```suggestion
    async def filter_for_gitignore(file_list: list[str], base_path: str) -> list[str]:
```

malhotra5: More of these everywhere

PR #8292, opened 2025-05-05
[Feat]: Always autogen title

malhotra5:

- Fix where conversations started with openhands api would not get autogenerated titles unless the conversation is opened

---

At the moment conversations only get auto generated titles when the user has opened a conversation. However, this means conversations opened with openhands api, or cloud resolver don't get an autogenerated title unless the user logs into the dashboard.

This PR autogens a title inside the conversation update callback. This ensures that if the agent is kicked off, then we attempt to autogenerate a conversation title.

We also send a socket event back to FE to indicate when the title change occurs, so it can refetch its title

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:cb894be-nikolaik   --name openhands-app-cb894be   docker.all-hands.dev/all-hands-ai/openhands:cb894be
```


openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -205,6 +210,18 @@ async def _get_conversation_store(
...
+    async def _get_settings_store(
+            self, user_id: str | None
+    ) -> SettingsStore:
+        settings_store_class = self._settings_store_class
+        if not settings_store_class:
+            self._settings_store_class = settings_store_class = get_impl(
+                SettingsStore,
+                self.server_config.conversation_store_class
+            )
+        store = await settings_store_class.get_instance(self.config, user_id)
```
enyst, line 222: We are here in the `StandaloneConversationManager`, I think we know already what the settings store class is?

frontend/__tests__/components/features/conversation-panel/conversation-panel.test.tsx
```diff
@@ -201,64 +203,6 @@ describe("ConversationPanel", () => {
     });
   });
 
-  it("should rename a conversation", async () => {
```
enyst, line 207: Maybe we could replace the tests for the older behavior with the new behavior?

malhotra5: Yup will do

openhands/server/conversation_manager/standalone_conversation_manager.py
enyst: I'm not suggesting to hardcode it, I just wonder if we can do this differently, the code came from `manage_conversations` which indeed is not tied to a settings class.

Maybe this and _auto_generate_title() could be in some conversation utils or even the summary utils... if it doesn't depend on `server`

openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -205,6 +210,18 @@ async def _get_conversation_store(
         store = await conversation_store_class.get_instance(self.config, user_id)
         return store
 
+    async def _get_settings_store(
+            self, user_id: str | None
+    ) -> SettingsStore:
+        settings_store_class = self._settings_store_class
+        if not settings_store_class:
+            self._settings_store_class = settings_store_class = get_impl(
```
enyst, line 218:
```suggestion
            self._settings_store_class = get_impl(
```

enyst:
I didn't see if we only autogen for events with source='user'. I think for every 'agent' it might be too much...

openhands/server/conversation_manager/standalone_conversation_manager.py
malhotra5: Actually it seems its easy to pass around the user settings, its already available to us when we create the callback

updated pr with it, thanks for the catch!

openhands/server/conversation_manager/standalone_conversation_manager.py
malhotra5: Removed the refetching as the settings object is already available

malhotra5:
> I didn't see if we only autogen for events with source='user'. I think for every 'agent' it might be too much...

Yup its only based on the first user message! Once the title is set we don't touch it again

enyst (commented):
> Yup its only based on the first user message! Once the title is set we don't touch it again

Aha! You're right, the callback scared me a bit 😅 . OK.

On the other hand, the two private methods in the "standalone" conversation class still seem strange to me. 

As far as I can tell the saas doesn't use the "standalone" class. Which suggests to me that it either doesn't work with the saas use case, or there's a lot of duplicate code, or the architecture here has something unholy. (you're inheriting this class? that seems messy, and even if it was, they shouldn't be private at a minimum, probably).

Is there a reason why they cannot be in conversation_summary? It seems more understandable to have the feature in one place anyway.

Or, alternatively, if they have to be in one of these classes for some reason, how about the superclass?

malhotra5:
> Is there a reason why they cannot be in conversation_summary

Moved it into `conversation_summary`, good point! It still works with local installation and saas

enyst (approved):
Thank you. One down!

Is there a reason why the other _private_ method where this PR added quite a lot of common code, `_update_conversation_for_event`, couldn't also be moved and made public, so that it's reusable?

It seems to me that the same reasons apply. It would bring some minimal code clarity IMHO.

malhotra5:
Yeah I think its ideal to move a lot of these methods out of the class, `_update_conversation_for_event` for instance. Although off the top of my I'm not sure where I'd move it (it does stuff besides autogenerating a title for the convo), and I'd like to double check that it doesn't adversely affect SAAS (it shouldn't, but just to be doubly sure). For these reasons I think I'll follow up with those changes in a future PR! 

Thanks for the feedback and review!

PR #8293, opened 2025-05-05
Fix typing in routes directory

neubig:
This PR fixes typing in the routes directory to use uppercase typing imports (Dict, List) instead of lowercase (dict, list) to comply with strict mypy checks.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:ad553ae-nikolaik   --name openhands-app-ad553ae   docker.all-hands.dev/all-hands-ai/openhands:ad553ae
```

### DIFF
--- a/openhands/server/routes/files.py
+++ b/openhands/server/routes/files.py
@@ -1,5 +1,5 @@
 import os
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List
 
 from fastapi import (
     APIRouter,
@@ -112,8 +112,8 @@ async def list_files(request: Request, path: str | None = None) -> Dict[str, Any
     return file_list
 
 
-@app.get('/select-file')
-async def select_file(file: str, request: Request) -> Union[FileResponse, JSONResponse]:
+@app.get('/select-file', response_model=None)
+async def select_file(file: str, request: Request) -> FileResponse | JSONResponse:
     """Retrieve the content of a specified file.
 
     To select a file:
--- a/openhands/server/routes/files.py
+++ b/openhands/server/routes/files.py
@@ -1,4 +1,5 @@
 import os
+from typing import Any, Dict, List, Union
 
 from fastapi import (
     APIRouter,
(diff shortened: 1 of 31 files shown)


neubig:
@openhands

1. check the diff between this and the base branch
2. use lower case typing instead of upper case
3. if you removed the arguments of "list" or "dict" move them back
4. make sure you use "|" instead of union
5. make sure that the "py-unit-tests.yml" and "lint.yml" workflows pass locally
6. push back to the branch once you have confirmed all of those things

neubig:
@openhands check the diff with the base branch, and convert all places that use "Optional" to use "|". make sure that lint.yml passes.

neubig:
This is ready for review!

openhands/server/routes/files.py
```diff
@@ -42,7 +43,7 @@
 
 
 @app.get('/list-files')
-async def list_files(request: Request, path: str | None = None):
+async def list_files(request: Request, path: str | None = None) -> dict[str, Any]:
```
malhotra5, line 46:
```suggestion
async def list_files(request: Request, path: str | None = None) -> dict[str, str]:
```

Should this be `str`?

openhands/server/routes/files.py
```diff
@@ -83,7 +84,9 @@ async def list_files(request: Request, path: str | None = None):
 
     file_list = [f for f in file_list if f not in FILES_TO_IGNORE]
 
-    async def filter_for_gitignore(file_list, base_path):
+    async def filter_for_gitignore(
+        file_list: list[dict[str, Any]], base_path: str
+    ) -> list[dict[str, Any]]:
```
malhotra5, line 89:
```suggestion
    ) -> list[dict[str, str]]:
```
Same here I'm guessing we're returning file paths as strings, its just that we're lacking type enforcement in general so the agent used `Any` as a safety

openhands/server/routes/secrets.py
```diff
@@ -1,3 +1,5 @@
+
+
 from fastapi import APIRouter, Depends, status
```
malhotra5, line 3:
```suggestion
from fastapi import APIRouter, Depends, status
```

malhotra5 (approved):
LGTM, left one comment that's worth investigating imo

Issue #8282, opened 2025-05-05
[Bug]: Memory Condensation sometimes ends up on loop when browsing involves.


tawago:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

First observed on Apr 21st after the memory condensation introduced (casually mentioned it to Graham directly)
Probably related to #8269

With memory condensation on, ask the agent to copy a small-mid size website perfectly. 
After 100 turns or so, intervene to give a feedback, which would trigger more browsing. Then, it gets stuck in a loop.

[trajectory-cf5543a577ec435888ea376539a842a9.json](https://github.com/user-attachments/files/20040607/trajectory-cf5543a577ec435888ea376539a842a9.json)

### OpenHands Installation

Docker command in README

### OpenHands Version

0.36.0

### Operating System

MacOS

### Logs, Errors, Screenshots, and Additional Context

<img width="573" alt="Image" src="https://github.com/user-attachments/assets/7ddd0b5b-ce33-4f7d-8955-9fe9f0434d7a" />

tawago:
sidenote: interestingly the Input Tokens skyrocketed to 100,000

csmith49:
Another data point: I set OpenHands loose on this issue. It started by looking at this page, then the page for #8269...then immediately entered the browsing loop. I interrupted the agent as soon as it was clear it was looping -- only 35 events in the trajectory.

[trajectory-b6f10bbef2f9486aa8593f4b9bb6a95a.json](https://github.com/user-attachments/files/20047759/trajectory-b6f10bbef2f9486aa8593f4b9bb6a95a.json)

That at least helps narrow down the source of the problem. Given the length of the provided trajectory I was concerned that there was some strange interaction between the summarizing condenser and the browser output condenser, but it seems like the issue can present itself before the summarizing condenser even has a chance to kick in.

csmith49:
Turns out it's possible to recover from the loop by interrupting and directing away from the browsing behavior. Still appeared to have access to all the expected context. Looking more at the trajectories, here's my hypothesis:

[...]

PR #8294, opened 2025-05-05
updating web read tool description to explain masking

csmith49:

Fixes issue where browsing gets the agent stuck in a loop.

---

In some situations, while browsing, the agent becomes stuck in a loop and reads a set of web-pages again and again. The introduction of this behavior coincided with #7850, which (among other things) adds the `BrowserOutputCondenser` as a standard inclusion in the memory management pipeline.

This PR updates the tool description for the `web_read` tool to explain the behavior of the browser output condenser.

---

Intended to address #8282.


xingyaoww (approved):
LGTM! I think this could potentially be completely fixed after we use the fetch MCP tool to replace this, since the output of that tool will no longer be BrowserOutputObservation

csmith49:
> LGTM! I think this could potentially be completely fixed after we use the fetch MCP tool to replace this, since the output of that tool will no longer be BrowserOutputObservation

Okay, that's good to know. If we want to apply the same masking strategy in that context we'll have to update the `BrowserOutputCondenser` to handle that representation as well, in which case we might see this problem crop up again.

PR #8297, opened 2025-05-06
Enable resolver to load configuration from environment variables in GitHub Actions

kotauchisunsun:

I made a modification to ensure that environment variables are properly applied when running the resolver with GitHub Actions.
Although the documentation states that behavior can be configured via environment variables such as LLM_NUM_RETRIES, these settings were not being reflected when running the resolver.

https://docs.all-hands.dev/modules/usage/llms

Therefore, we updated the configuration loading process to allow settings to be specified via environment variables.

---

---


enyst (approved):
Thank you for your contribution, @kotauchisunsun !

You're right we should read from env these settings, and probably many others, TBH.

On a side note, I think we should actually try to refactor this to use the standard way of [loading the configuration](https://github.com/All-Hands-AI/OpenHands/blob/4c1ae6fd8da5c2280cfbfe8f6a4a2fcd92cb644d/openhands/core/config/utils.py#L572), with a call to `load_app_config()`, which should take care of loading from env, and only override with resolver-specific args as we see here.

I made an issue for openhands-agent on it: https://github.com/All-Hands-AI/OpenHands/issues/8298

(It doesn't have to be in scope for this PR, it's a legacy issue that we didn't get to yet. A PR for that issue would be most welcome though!)

kotauchisunsun:
@enyst 
Thanks for merging the PR!

PR #8330, opened 2025-05-07
Plumb custom secrets to runtime

malhotra5:

---

This PR 

1. plumbs user's custom secrets to the runtime
2. adds secret name + description pairs to runtime info

This PR is a follow up for #8068

image

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:7762939-nikolaik   --name openhands-app-7762939   docker.all-hands.dev/all-hands-ai/openhands:7762939
```


frontend/src/state/chat-slice.ts
```diff
@@ -212,6 +212,17 @@ export const chatSlice = createSlice({
...
+            recallObs.extras.custom_secrets_descriptions &&
+            Object.keys(recallObs.extras.custom_secrets_descriptions).length > 0
+          ) {
+            content += `\n\n**Custom Secrets**`;
+            for (const [name, description] of Object.entries(
+              recallObs.extras.custom_secrets_descriptions,
+            )) {
+              content += `\n\n- $${name}: ${description}`;
+            }
+          }
```
rbren, line 225: This is an existing pattern so not blocking but I really dislike all the structured data in the recallObservation--IMO this should be much more generic. It should just have markdown we can display instead of having to hard code all this

Future work!

openhands/storage/data_models/user_secrets.py
```diff
@@ -136,3 +137,31 @@ def convert_dict_to_mappingproxy(
                 new_data['custom_secrets'] = secrets
 
         return new_data
+    
+
+    def set_event_stream_secrets(self, event_stream: EventStream) -> None:
```
rbren, line 142: It feels a bit weird to me that a UserSecrets object knows about EventStreams. I think the reverse might be a better move? i.e. have `event_stream.set_secrets()` take in a SecretStore

frontend/src/state/chat-slice.ts
malhotra5: Yeah def agree, I was scratching my head as to why custom secrets weren't showing only to find out I have to hard code it

rbren (approved):
🍰 but I think I'd prefer to invert the EventStream <> SecretStore relationship if possible

openhands/storage/data_models/user_secrets.py
malhotra5: We I can do that; currently we also have `ProviderHandler` take in an event stream and mask the secrets

The idea is that `ProviderHandler` and `UserSecrets` know best how to manipulate their own values, so we let them handle it instead

openhands/memory/memory.py
```diff
@@ -176,6 +176,9 @@ def _on_workspace_context_recall(
                 microagent_knowledge=microagent_knowledge,
                 content='Added workspace context',
                 date=self.runtime_info.date if self.runtime_info is not None else '',
+                custom_secrets_descriptions=self.runtime_info.custom_secrets_descriptions
+                if self.runtime_info is not None
+                else {},
```
enyst, line 181: Could we add them to unit tests, for example something where `runtime_info` is tested? Like, where it's read in memory, and where it's later serialized in a message for the LLM

malhotra5: @openhands please write units tests for the new attribute `custom_secrets_descriptions` introduced to `RuntimeInfo`

> Could we add them to unit tests, for example something where runtime_info is tested? Like, where it's read in memory, and where it's later serialized in a message for the LLM

openhands/storage/data_models/user_secrets.py
malhotra5: I have a nice solution to reverse the dependency but I'd like to do it in a separate PR (so we can revert it in the event refresh stops working or secrets are accidentally exposed)

PR #8335, opened 2025-05-07
Add return type annotations to functions in runtime/utils directory

neubig:
This PR adds proper return type annotations to functions in the openhands/runtime/utils directory to fix mypy strict type checking errors. Specifically:

- Added `-> None` return type annotations to methods in windows_bash.py
- Added a type ignore comment to the retry decorator in request.py
- Removed an unused type ignore comment in edit.py

These changes help ensure strict type checking compliance for the runtime/utils directory.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:b860fd8-nikolaik   --name openhands-app-b860fd8   docker.all-hands.dev/all-hands-ai/openhands:b860fd8
```


neubig:
@openhands make sure that you revert all changes outside of the runtime/utils directory, especially those for the linting configuration files.

neubig:
This is ready for review!

openhands/runtime/utils/request.py
```diff
@@ -29,7 +29,8 @@ def is_retryable_error(exception: Any) -> bool:
     )
 
 
-@retry(
+# Type ignore is needed because tenacity's retry decorator doesn't preserve function signature
+@retry(  # type: ignore
```
malhotra5, line 33: Hmm doesn't seem like a problem elsewhere

Maybe worth double checking if this is true - not blocking imo

malhotra5 (approved):
LGTM, left a comment but not blocking

neubig:
@openhands in `openhands/runtime/utils/request.py` you made this comment

```python
# Type ignore is needed because tenacity's retry decorator doesn't preserve function signature
@retry(  # type: ignore
```

But I'm not sure if this is actually true. Could you double-check by installing it pre-commit-config.yaml without using type ignore and see if you run into any actual problems? If you do, report the error to me, if not, then please remove the type ignore.

openhands/runtime/utils/request.py
neubig: Thanks, you were correct here, I was too trusting :)

PR #8336, opened 2025-05-07
Add more extensive typing to server/routes files

neubig:
This PR adds more extensive typing to the server/routes directory, specifically:

- Fixed return types in files.py to properly handle JSONResponse returns
- Fixed get_config() in public.py to ensure it returns a proper dict type
- Used modern lower-case typing like "list", "dict", as well as "|" instead of "Union"

These changes help improve type safety and make the codebase more maintainable.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c47d8b7-nikolaik   --name openhands-app-c47d8b7   docker.all-hands.dev/all-hands-ai/openhands:c47d8b7
```


neubig:
@OpenHands make sure that you revert all changes outside of the target directory, especially those for the linting configuration files.

openhands/server/routes/files.py
```diff
@@ -42,8 +42,10 @@
 app = APIRouter(prefix='/api/conversations/{conversation_id}')
 
 
-@app.get('/list-files')
-async def list_files(request: Request, path: str | None = None) -> dict[str, str]:
+@app.get('/list-files', response_model=None)
```
malhotra5, line 45: Weird that the response model is None when the return type is `list[str]`

openhands/server/routes/files.py
```diff
@@ -169,8 +169,8 @@ async def select_file(file: str, request: Request) -> FileResponse | JSONRespons
         )
 
 
-@app.get('/zip-directory')
-def zip_current_workspace(request: Request) -> FileResponse:
+@app.get('/zip-directory', response_model=None)
```
malhotra5, line 172: Same here, reworking the response models on the routes would be helpful!

openhands/server/routes/public.py
```diff
@@ -65,4 +65,4 @@ async def get_config() -> dict[str, Any]:
     Returns:
         dict[str, Any]: The current server configuration.
     """
-    return server_config.get_config()
+    return dict(server_config.get_config())
```
malhotra5, line 68: Hmm shouldn't be necessary to add `dict` here, doesn't hurt I guess but `get_config` should already be typed with `dict` - mypy should be able to catch the type compatibility here

neubig:
Thanks @malhotra5 , it looks like the first two might be necessary to handle the failure cases: https://fastapi.tiangolo.com/tutorial/response-model/#disable-response-model

Hopefully mypy will handle the type checking, so we don't need to lean on fastapi for it.

malhotra5:
Hmm sorry to push back again on the PR; I see the in FastAPI docs it mentions

> you might not want to have the default data validation, documentation, filtering, etc. that is performed by FastAPI.

But I think we do want the default data validation. FastAPI response model doesn't affect mypy's ability to  type check (both of these are compatible with each other). I dug a bit more and the docs mention that the right time to disable response models is if the returned values from the routes are [not valid Pydantic types](https://fastapi.tiangolo.com/tutorial/response-model/#response_model-priority)


Claude mentioned one of these for a none value in response model is

> The functions do return different types based on success/error conditions

I believe FastAPI only executes response_model for successful API calls (2xx responses -> fastapi will validate, and wrap the return statement in a jsonresponse)

For erroneous calls, we're allowed to use JSONResponse; we should still enforce response models wherever possible (since the routes here I believe have one return type for successful api calls)

According to GPT 

>Successful Responses: When your route function returns a Pydantic model instance, FastAPI automatically validates and serializes it according to the specified response_model. This ensures that the response adheres to the defined schema, providing consistency and validation.

> Error Responses: In cases where an error occurs (e.g., resource not found, unauthorized access), you can return a JSONResponse with a custom status code and content. FastAPI will not apply the response_model validation to these error responses, allowing you to define the response structure and status code explicitly.

I wouldn't find exactly where in the docs it references this but it seems pretty consistent with the behavior we have seen when running openhands

neubig:
OK, added a bit more comprehensive stuff here @malhotra5 , in some cases the response types were not pydantic so `response_type=None` seems to be necessary. But at least they're well specified now.

malhotra5 (approved):
LGMT, thank you!

PR #8350, opened 2025-05-08
Add a DeepWiki reference

erkinalp:
Adding a link to DeepWiki makes DeepWiki auto-refresh the auto-generated documentation weekly


neubig (commented):
Hey @erkinalp , thanks for this! The deepwiki documentation is great, but it's geared more for developers (and we wouldn't want users to be confused by it). Do you think we could put this in the development.md file?

mamoodi:
> @mamoodi do you have any opinions about whether there's an inconspicuous place where we could put this that wouldn't confuse users, but could be helpful to developers?

I agree that in the Developers section would be the best place to have this. But erkinalp says it has to be at the root unfortunately :/ 

Somewhere in Development Overview. Maybe at the top: https://docs.all-hands.dev/modules/usage/how-to/development-overview

enyst:
> @mamoodi do you have any opinions about whether there's an inconspicuous place where we could put this that wouldn't confuse users, but could be helpful to developers?

Sorry to interfere in this discussion, but this confuses me a lot. Our users _are_ developers. I'm not sure why would they be confused? Maybe this distinction doesn't need to go far too far.

FWIW I looked at the DeepWiki pages, and the first page people see is an architecture overview from a very high level which is, IMO, useful. I don't see anything wrong or confusing in a quick view of a simple diagram with a note about running agent or accessing multiple LLMs.

More importantly, it seems to me that perhaps this isn't a dichotomy. Somewhere there are agent developers, local LLM fans tinkering a bit, UI developers perhaps (openhands could have more than one user-friendly UI), runtime providers or users for that matter, developers who vibe code, and developers who may look to customize `openhands` to their liking.

Do we expect they all should have to decide from the start to suddenly be openhands developers, in order to find some information about the essential parts of the system? I think that risks to make the project harder to contribute to.

I'm sorry if this comes across a bit unclear or really confused - it's because I am confused. I think the project would greatly benefit from making this information easy to find.

mamoodi:
> > @mamoodi do you have any opinions about whether there's an inconspicuous place where we could put this that wouldn't confuse users, but could be helpful to developers?
> 
> Sorry to interfere in this discussion, but this confuses me a lot. Our users _are_ developers. I'm not sure why would they be confused? Maybe this distinction doesn't need to go far too far.
> 
> FWIW I looked at the DeepWiki pages, and the first page people see is an architecture overview from a very high level which is, IMO, useful. I don't see anything wrong or confusing in a quick view of a simple diagram with a note about running agent or accessing multiple LLMs.
> 
> More importantly, it seems to me that perhaps this isn't a dichotomy. Somewhere there are agent developers, local LLM fans tinkering a bit, UI developers perhaps (openhands could have more than one user-friendly UI), runtime providers or users for that matter, developers who vibe code, and developers who may look to customize `openhands` to their liking.
> 
> Do we expect they all should have to decide from the start to suddenly be openhands developers, in order to find some information about the essential parts of the system? I think that risks to make the project harder to contribute to.
> 
> I'm sorry if this comes across a bit unclear or really confused - it's because I am confused. I think the project would greatly benefit from making this information easy to find.

Just to clarify one thing.

*Most* of our current users are developers. Not all ;)

That being said, I personally don't have a strong opinion here and will let @rbren and @neubig decide.

enyst:
Mamoodi, that's okay, I'm sure you know it doesn't change anything. Everyone is welcome to play with the project and may find it useful! It's surely not an exclusive club. 😅 

That's just my point. I think we shouldn't push people away by making it harder to figure out if they're interested in the system, or get some idea how our agent works, or get started with any kind of contribution. When we move documentation away, out of sight, it seems to me like they're not invited in, unless they've already decided to become core openhands developers. IMHO we should invite them in.

mamoodi:
> Mamoodi, that's okay, I'm sure you know it doesn't change anything. Everyone is welcome to play with the project and may find it useful! It's surely not an exclusive club. 😅
> 
> That's just my point. I think we shouldn't push people away by making it harder to figure out if they're interested in the system, or get some idea how our agent works, or get started with any kind of contribution. When we move documentation away, out of sight, it seems to me like they're not invited in, unless they've already decided to become core openhands developers. IMHO we should invite them in.

I'll play ball here since this can be interesting but also plays a little into documentation that you have noticed I'm really interested in :) 

Let's say you have two sets of docs:
* Set A: Has a simple diagram of how a system works. Doesn't go too much into detail, but is simple and high level enough that even a junior developer can understand.
* Set B: You have an incredibly detailed document with all the bells and whistles of how everything works and every detail about how all the configuration works and how to change each of these configurations.

Which one is more inviting to all users? 
My opinion is if something is too detailed and has too much information right off the bat, it will actually deter users and in this case contributors. Opening up that doc for someone who isn't as smart as you (like me) is overwhelming. It doesn't look inviting to me or a project that I can contribute to easily.

The best way to approach this is start simple and then add more details and introduce more complex documentation. That is why I proposed being in the Developers section would be better. If the deepwiki was simpler and more inviting, I'd definitely agree with you. In this case, I will defer to Robert and Graham.

OpenHands has a very complex architecture (at least I think so and from what I've heard from the community that wants to contribute) so I definitely agree with having that documentation somewhere but as an introductory doc under "documentation" in the main README, it could have an unwanted effect.

enyst:
Thank you, that's exactly what I was looking at, we have a link to `Development.md` and to `Contributing.md` so I don't see a high level documentation of the system as a lot different.

Mamoodi, point taken, it's possible in some cases, I'm sure, but I'm not sure at all that it's applicable here. We will always need to document things that arguably not everyone cares about. Hey, even I don't look at everything on the docs site, and you know I do care about a lot of it. 🤷(Admittedly I also don't think that a 4-components diagram is overly detailed, but if it is, how about we trust people to read if they care and not read if they don't?)

I appreciate the discussion, probably more is not in scope of this PR. 😅 Thank you for the contribution @erkinalp !

erkinalp:
@enyst Each subcomponent is also documented in detail.

README.md
```diff
@@ -99,6 +100,15 @@ check out our [documentation](https://docs.all-hands.dev/modules/usage/getting-s
 There you'll find resources on how to use different LLM providers,
 troubleshooting resources, and advanced configuration options.
 
+[![Cognition has autogenerated a documentation for us](https://deepwiki.com/badge.svg)](https://deepwiki.com/All-Hands-AI/OpenHands).
+
+### Custom Scripts
```
mamoodi, line 105: I think this part was added mistakenly. It was there in an older version but seems like some merge issues brought it back.

rbren:
This looks a bit out of place with our badges right now
Screenshot 2025-05-20 at 1 19 25 PM

Let's add it down in the documentation section
Screenshot 2025-05-20 at 1 20 45 PM

mamoodi (approved):

mamoodi:
Thanks @erkinalp and sorry for the back and forth!

Issue #8317, opened 2025-05-07
[Bug]: OpenHands 0.36 Fails Agent Initialization due to `/update_mcp_server` 404 Error


ahundt:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

# Bug Report: OpenHands 0.36 Fails Agent Initialization due to `/update_mcp_server` 404 Error

## Affected Versions & Environment

* **OpenHands Application:** `docker.all-hands.dev/all-hands-ai/openhands:0.36`
* **Runtime Image:** `docker.all-hands.dev/all-hands-ai/runtime:0.36-nikolaik`
* **Host Operating System:** macOS
* **Docker Environment:** Docker Desktop (installed via Homebrew)

## Problem Description

OpenHands version 0.36, when used with the `runtime:0.36-nikolaik` image, fails to initialize agent sessions. The `openhands-app` container logs a `404 Not Found` error. This error occurs because the application cannot reach the required `/update_mcp_server` endpoint at `http://host.docker.internal:35078/update_mcp_server` on the runtime container.

## Steps to Reproduce

1.  Execute the following Docker command. This command reflects the exact setup used when this issue was originally encountered:

    ```bash
    docker run -it --rm --pull=always \
      -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.36-nikolaik \
      -e SANDBOX_ADD_HOSTS="host.docker.internal:host-gateway" \
      -e LLM_API_KEY="ollama" \
      -e LLM_BASE_URL="[http://127.0.0.1:11434](http://127.0.0.1:11434)" \
      -e LLM_OLLAMA_BASE_URL="[http://127.0.0.1:11434](http://127.0.0.1:11434)" \
      -e LOG_ALL_EVENTS=true \
      -v /var/run/docker.sock:/var/run/docker.sock \
      -v ~/.openhands-state:/.openhands-state \
      -p 3000:3000 \
      --add-host host.docker.internal:host-gateway \
      --name openhands-app \
      docker.all-hands.dev/all-hands-ai/openhands:0.36
    ```

2.  Observe the `openhands-app` container's logs for errors during agent session initialization. To view these logs while the container attempts to run or after it exits, use the command:
[...]

Issue #8285, opened 2025-05-05
[Bug]: MCP is broken for SSE servers


cshimmin:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Since the MCPConfig structure has changed, I have not been able to successfully use SSE servers as I had been previously.

For example, my config has

```toml
[[mcp.sse_servers]]
url="http://localhost:8989/sse"
```

and when an agent starts up, I can see in the logs that it is correctly connecting to the server, listing the tools, and associating them with the CodeActAgent. However, I also see some errors that appear to caused by trying to connect to an SSE endpoint within the sandbox container.

```
17:46:48 - openhands:INFO: utils.py:52 - Initializing MCP agent for url='http://localhost:8989/sse' api_key='******' with SSE connection...
17:46:48 - openhands:INFO: client.py:90 - Connected to server with tools: ['tool_1', 'tool_2', 'tool_3']
17:46:48 - openhands:INFO: utils.py:61 - Connected to MCP server url='http://localhost:8989/sse' api_key='******' via SSE
17:46:48 - openhands:INFO: utils.py:52 - Initializing MCP agent for url='http://localhost:35639/sse' api_key='******' with SSE connection...
17:46:48 - openhands:ERROR: client.py:64 - Error connecting to http://localhost:35639/sse: 1 validation error for ListToolsResult
tools
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
17:46:48 - openhands:INFO: client.py:117 - Disconnected from MCP server
17:46:48 - openhands:ERROR: utils.py:63 - Failed to connect to url='http://localhost:35639/sse' api_key='******': 1 validation error for ListToolsResult
tools
  Field required [type=missing, input_value={}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
17:46:48 - openhands:INFO: client.py:117 - Disconnected from MCP server
17:46:48 - openhands:INFO: utils.py:173 - Loaded 3 MCP tools: ['tool_1', 'tool_2', 'tool_3']
17:46:48 - openhands:INFO: agent.py:169 - Setting 3 MCP tools for agent CodeActAgent: ['tool_1', 'tool_2', 'tool_3']
[...]

PR #8353, opened 2025-05-08
fix(mcp): fix SSE MCP server connection & add tests

xingyaoww:

---

Fix the issue when OpenHands won't work with SSE MCP server and add tests to make sure this doesn't happen again

---

Hopefully will fix https://github.com/All-Hands-AI/OpenHands/issues/8317 https://github.com/All-Hands-AI/OpenHands/issues/8285

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:8ae7425-nikolaik   --name openhands-app-8ae7425   docker.all-hands.dev/all-hands-ai/openhands:8ae7425
```

### DIFF
--- a/openhands/runtime/impl/action_execution/action_execution_client.py
+++ b/openhands/runtime/impl/action_execution/action_execution_client.py
@@ -369,12 +369,13 @@ class ActionExecutionClient(Runtime):
                 timeout=10,
             )
             if response.status_code != 200:
-                raise RuntimeError(f'Failed to update MCP server: {response.text}')
+                self.log('warning', f'Failed to update MCP server: {response.text}')
 
             # No API key by default. Child runtime can override this when appropriate
             updated_mcp_config.sse_servers.append(
                 MCPSSEServerConfig(
-                    url=self.action_execution_server_url.rstrip('/') + '/sse', api_key=None
+                    url=self.action_execution_server_url.rstrip('/') + '/sse',
+                    api_key=None,
                 )
             )
             self.log(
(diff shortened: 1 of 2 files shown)


cshimmin:
Thank you! I will give it a try right away. However at a glance I’m still worried this is not addressing the root cause w the MCP connection issues. If I find that to be the case I will submit a separate issue.

xingyaoww:
@cshimmin thanks! though weridly enough, i didn't run into any connection issue locally. If you keep running into it, could you try run [MCP inspector](https://modelcontextprotocol.io/docs/tools/inspector) on the machine where you run OpenHands, and see if your MCP server can be connected correctly?

cshimmin:
This worked for me! I had to cherry-pick #8352 in order to test it.

I'll investigate the low-level connection issues w/ MCP inspector, and open a separate issue if needed.

Thank you!

openhands/runtime/impl/action_execution/action_execution_client.py
```diff
@@ -359,26 +359,30 @@ def get_updated_mcp_config(self) -> MCPConfig:
...
+                json=stdio_tools,
+                timeout=10,
             )
-        )
-        self.log(
-            'info',
-            f'Updated MCP config: {updated_mcp_config.sse_servers}',
-        )
+            if response.status_code != 200:
+                raise RuntimeError(f'Failed to update MCP server: {response.text}')
```
enyst, line 372: Could we perhaps log a warning, instead of raise, so it doesn't just fail?

I thought we catch this in the controller, but it seems it fails for people. 🤔 Could it be that we catch it after the runtime is initialized, and they had the exception before?

enyst (approved):

openhands/runtime/impl/action_execution/action_execution_client.py
xingyaoww: Good idea!

@OpenHands help me change this error into a logger.warning

xingyaoww: good bot ;)

PR #8357, opened 2025-05-08
feat(backend): New "update microagent prompt" API

amanape:

---
Steps:
1. Get conversation ID and event ID of event we want the agent to learn
2. Get context surrounding given event ID
3. Generate a prompt with the help of an LLM to use as initial query

Blocks #8592

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:913b0b1-nikolaik   --name openhands-app-913b0b1   docker.all-hands.dev/all-hands-ai/openhands:913b0b1
```

### DIFF
--- a/openhands/server/utils.py
+++ b/openhands/server/utils.py
@@ -1,4 +1,5 @@
 import uuid
+
 from fastapi import Depends, HTTPException, Request, status
 
 from openhands.core.logger import openhands_logger as logger


openhands/server/routes/microagent.py
```diff
@@ -0,0 +1,215 @@
...
+
+app = APIRouter(prefix='/knowledge')
+
+class KnowledgePromptRequest(BaseModel):
+    conversation_id: str
+    event_id: int
+
+
+@app.get('/prompt/{conversation_id}')
+async def get_prompt(
```
rbren, line 117: I'm not sure this will kick our conversation middleware into gear, meaning it might by pass auth. We should generally use the EvenStream that gets attached to the request object rather than constructing it dynamically

I think if you put this on `/{conversation_id/remember_prompt` it'll work better. Than you can use `request.state.conversation.event_stream`

openhands/server/routes/microagent.py
```diff
@@ -0,0 +1,215 @@
...
+from openhands.server.session.conversation_init_data import ConversationInitData
+from openhands.server.shared import file_store, conversation_manager
+from openhands.server.user_auth import get_user_id, get_user_settings_store
+from openhands.server.utils import get_conversation_store
+from openhands.storage.conversation.conversation_store import ConversationStore
+from openhands.storage.data_models.conversation_metadata import ConversationMetadata
+from openhands.storage.settings.settings_store import SettingsStore
+
+prompt_template = """
+You are tasked with generating a prompt that will be used by another AI to update a special reference file. This file contains important information and learnings that are used to carry out certain tasks. The file can be extended over time to incorporate new knowledge and experiences.
```
enyst, line 20: Ideally this should be in a .j2 template file, rather than inline. Not sure exactly where it belongs (unless we know what agent is running and can put it in the agent's `prompts/`). Maybe we can make a `prompts/` in microagent, unless we have some better ideas?

openhands/server/routes/microagent.py
```diff
@@ -0,0 +1,215 @@
...
+    update_request: KnowledgePromptRequest,
+    user_id: str = Depends(get_user_id),
+    user_settings: SettingsStore = Depends(get_user_settings_store),
+):
+    # get event stream for the conversation
+    event_stream = EventStream(update_request.conversation_id, file_store, user_id)
+    events = event_stream.get_events()
+
+    # find the specified events to learn from
+    context_events = get_context_events(list(events), update_request.event_id)
```
enyst, line 127: Looks like this list / enumerate is going through the whole list of events
```
events = event_stream.get_events()
context_events = ... list(events)
```
How about, `get_events()` gives us a stream, we can pick up from it events from a `start` and with an `end`. So we could just call it once, with the ends we already know, e.g. `get_events(start_id, end_id)`

openhands/server/routes/microagent.py
enyst: Actually, the prompt and `generate_prompt()` function is supposed to be a microagent, right? Microagent help 😄 

Small thing, but if so, could we implement also the function in microagent, instead?

amanape: > Small thing, but if so, could we have also the function in microagent, instead?

Not sure what you mean by this comment

openhands/server/routes/microagent.py
```diff
@@ -0,0 +1,215 @@
...
+@app.post('/update')
+async def update_knowledge(
+    update_request: KnowledgeUpdateRequest,
+    user_id: str = Depends(get_user_id),
+    conversation_store: ConversationStore = Depends(get_conversation_store),
+):
+    """
+    Update the microagent with new learnings through a conversation.
+    """
+    # get event stream for the conversation
```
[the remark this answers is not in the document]
amanape: We could also just call POST /conversation from the FE and pass in events from the client. WDYT?

openhands/server/routes/microagent.py
enyst: Thank you!

🤔 I think I must have meant that the `generate_prompt()` function could also be in the `microagent` package... Maybe not important for now. Not all the code design here is clear to me yet, we can see when it's out of draft. :)

openhands/server/routes/microagent.py
enyst: How about this, wouldn't it work?
```
start_id = update_request.event_id - context_size
end_id = update_request.event_id + context_size + 1
```

amanape: In order to safely retrieve the appropriate end index, I would have still have to `list(events)` :\

```py
context_size = 4
end_index = min(
    len(list(events)), event_id + context_size + 1  # +1 to include the target event
) 
```

openhands/server/routes/manage_conversations.py
```diff
@@ -252,6 +259,91 @@ async def delete_conversation(
...
+    event_stream = conversation.event_stream
+    events = event_stream.get_events()
+
+    # find the specified events to learn from
+    # Get X events around the target event
+    context_size = 4
+    start_index = max(0, event_id - context_size)
+    end_index = min(
+        len(list(events)), event_id + context_size + 1  # +1 to include the target event
+    )
```
enyst, line 284:
```suggestion
    
    # find the specified events to learn from
    # Get X events around the target event
    context_size = 4
    start_index = max(0, event_id - context_size)
    end_index = min(
        event_stream.get_latest_event_id(), event_id + context_size + 1  # +1 to include the target event
    ) 
```
Alternatively, getting events in the stream will stop without error when it doesn't find any more, so maybe we don't need `min`, just
```
context_events = event_stream.get_events(
        start_id=start_index,
        end_id=event_id + context_size + 1,
        )
```

openhands/server/utils.py
```diff
@@ -16,3 +20,60 @@ async def get_conversation_store(request: Request) -> ConversationStore | None:
...
+        request: The FastAPI request object.
+
+    Returns:
+        The conversation object.
+    """
+    conversation = getattr(request.state, 'conversation', None)
+    return conversation
+
+
+def get_context_events(
```
enyst, line 39: We could probably use this above, in `get_prompt()`, and tweak its implementation here

openhands/server/routes/manage_conversations.py
enyst: Actually, please see below, it seems better and more clear with `search_events`

openhands/server/utils.py
```diff
@@ -16,3 +20,60 @@ async def get_conversation_store(request: Request) -> ConversationStore | None:
...
+    Args:
+        events: List of events to search through.
+        event_id: The ID of the target event.
+        context_size: Number of events to include before and after the target event.
+
+    Returns:
+        A list of events including the target event and the specified number of events before and after it.
+    """
+    target_event_index = None
+    for i, event in enumerate(events):
```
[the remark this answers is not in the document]
enyst: Please let me know if you want me to tweak the events and the prompt. Probably on Monday if that's okay with you.

openhands/server/utils.py
amanape: Actually we don't use this function anymore

openhands/server/utils.py
```diff
@@ -18,6 +19,15 @@ async def get_conversation_store(request: Request) -> ConversationStore | None:
     return conversation_store
 
 
+async def generate_unique_conversation_id(
```
malhotra5, line 23: NIT: I think we have duplicate logic in `create_new_conversation` function; we should use this helper instead

openhands/server/utils.py
```diff
@@ -1,3 +1,4 @@
+import uuid
 from fastapi import Depends, HTTPException, Request, status
```
enyst, line 3:
```suggestion

from fastapi import Depends, HTTPException, Request, status
```

enyst (approved):
Thank you for the work on this! I'm looking forward to the memory UI and we can always tweak along the way.

PR #8377, opened 2025-05-09
Fix typing issues in openhands/llm directory

neubig:
This PR adds more extensive typing to the openhands/llm directory to support strict mypy checks. It includes:

- Added return type annotations to functions in fn_call_converter.py
- Added return type annotation -> None to __init__ method in llm.py
- Removed unused type ignore comments in llm.py
- Restructured decorator usage in llm.py, async_llm.py, and streaming_llm.py to fix typing issues
- Fixed messages_list type issue in llm.py

All files in the openhands/llm directory now pass mypy with strict type checking.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:8cb07f2-nikolaik   --name openhands-app-8cb07f2   docker.all-hands.dev/all-hands-ai/openhands:8cb07f2
```


openhands/llm/async_llm.py
```diff
@@ -39,14 +39,8 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
-        @self.retry_decorator(
-            num_retries=self.config.num_retries,
-            retry_exceptions=LLM_RETRY_EXCEPTIONS,
-            retry_min_wait=self.config.retry_min_wait,
-            retry_max_wait=self.config.retry_max_wait,
-            retry_multiplier=self.config.retry_multiplier,
-        )
-        async def async_completion_wrapper(*args: Any, **kwargs: Any) -> Any:
+        # Define the wrapper function
+        async def async_completion_impl(*args: Any, **kwargs: Any) -> Any:
```
neubig, line 43: Is there a way to avoid moving this decorator? It seems simpler to have it as a decorator.

neubig:
Check `@neubig`'s three comments, fix them. When you are done, double-check that pre-commit and tests pass, and that the comments are all fixed. After you update the PR, check if github CI passes.

neubig:
@openhands Check `@neubig`'s three comments, fix them. When you are done, double-check that pre-commit and tests pass, and that the comments are all fixed. After you update the PR, check if github CI passes.

neubig:
This is ready for review, TIA @enyst !

openhands/llm/fn_call_converter.py
```diff
@@ -83,7 +83,7 @@ def refine_prompt(prompt: str):
 app = Flask(__name__)
 
 @app.route('/')
-def index():
+def index() -> str:
```
enyst, line 86: This one is an example given to LLMs... I think it makes sense to give it an example with type hint, although... we don't actually know if users will want it

🤔 probably better with, than without, IMHO

openhands/llm/llm.py
```diff
@@ -292,9 +292,11 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
                     )
 
                 non_fncall_response_message = resp.choices[0].message
+                # At this point, messages is already a list due to the conversion earlier in the function
+                messages_list: list[dict[str, Any]] = messages
```
enyst, line 296: The comment is accurate, but that was at line 223, a long time ago, so to speak. I'm not sure it's more readable to add a var here, only for one use?

enyst: How about, maybe we can declare the initial var at line 204 with a name like `messages_kwarg`, then soon after, when we converted, at line 223, declare `messages` with the `list` type?

enyst (approved):
Had a small comment, only because the readability of the llm code is currently... not ideal.

My hope is that typing helps us with it 😅

neubig:
@openhands please reflect `@enyst`'s comment on `openhands/llm/llm.py`

PR #8223, opened 2025-05-02
feat/fix(fontend): Get public repos via repo URL

amanape:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

---
**Summarize what the PR does, explaining any non-trivial design decisions.**
If the user wants to clone a repo that isn't owned by them, they can paste a URL in the input.
image

---
**Link of any specific issues this addresses:**

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:2e2ba59-nikolaik   --name openhands-app-2e2ba59   docker.all-hands.dev/all-hands-ai/openhands:2e2ba59
```

PR #8380, opened 2025-05-09
feat: add support search for gitlab project

answer-huang:

Add support search for gitlab project.

---
Add support search for gitlab project.

---

### DIFF
--- a/frontend/src/components/features/home/repo-selection-form.tsx
+++ b/frontend/src/components/features/home/repo-selection-form.tsx
@@ -14,6 +14,9 @@ import {
   BranchLoadingState,
   BranchErrorState,
 } from "./repository-selection";
+import { useSearchRepositories } from "#/hooks/query/use-search-repositories";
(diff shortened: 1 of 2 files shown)


malhotra5:
I believe we have a fix being merged in #8223! This will work across Github and Gitlab

answer-huang:
> I believe we have a fix being merged in #8223! This will work across Github and Gitlab

Hi there, Thanks for your remind.👏

malhotra5:
Hey @answer-huang ! I see the problem now

1. Yes the max repo limit is 1000
2. You can search across the thousand repos at the moment, but not outside of it
3. We recently re-added the search functionality (but only for public repos); this is why we require `https://` prefix to indicate that search logic should be triggered, because we're looking for public repos


I do understand that we want to support more than 1000 repositories however

Maybe we can do it by tweaking the way perform search; we could make the search endpoint not only limited to public repos

PR #8402, opened 2025-05-09
Add support for user/org level microagents

rbren:
This PR adds support for user/org level microagents. When a repository is selected (e.g., github.com/acme-co/api), the system will now also check if a .openhands repository exists at the organization level (e.g., github.com/acme-co/.openhands). If it exists, it will clone it and load the microagents from the ./microagents/ folder.

This allows organizations to define microagents that apply to all of their repositories without having to duplicate them in each repository.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:6b0fc11-nikolaik   --name openhands-app-6b0fc11   docker.all-hands.dev/all-hands-ai/openhands:6b0fc11
```

### DIFF
--- a/openhands/runtime/base.py
+++ b/openhands/runtime/base.py
@@ -534,7 +534,10 @@ fi
         if not files:
             return loaded_microagents
 
-        self.log('info', f'Found {len(files)} files in {source_description} microagents directory')
+        self.log(
+            'info',
+            f'Found {len(files)} files in {source_description} microagents directory',
+        )
         zip_path = self.copy_from(str(microagents_dir))
         microagent_folder = tempfile.mkdtemp()
 
--- a/openhands/runtime/base.py
+++ b/openhands/runtime/base.py
@@ -516,6 +516,80 @@ fi
 
         self.log('info', 'Git pre-commit hook installed successfully')
 
+    def _load_microagents_from_directory(
+        self, microagents_dir: Path, source_description: str
+    ) -> list[BaseMicroagent]:
+        """Load microagents from a directory.
+
+        Args:
+            microagents_dir: Path to the directory containing microagents
+            source_description: Description of the source for logging purposes
+
+        Returns:
+            A list of loaded microagents
+        """
+        loaded_microagents: list[BaseMicroagent] = []
+        files = self.list_files(str(microagents_dir))
+
+        if not files:
+            return loaded_microagents
+
+        self.log('info', f'Found {len(files)} files in {source_description} microagents directory')
+        zip_path = self.copy_from(str(microagents_dir))
+        microagent_folder = tempfile.mkdtemp()
+
+        try:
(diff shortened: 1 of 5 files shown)


rbren:
@openhands fix the python lint

openhands/runtime/base.py
```diff
@@ -516,21 +516,187 @@ def maybe_setup_git_hooks(self):
...
+                )
+
+                # Load microagents from the org-level repo
+                org_microagents_dir = org_repo_dir / 'microagents'
+                loaded_microagents = self._load_microagents_from_directory(
+                    org_microagents_dir, 'org-level'
+                )
+
+                # Clean up the org repo directory
+                shutil.rmtree(org_repo_dir)
```
enyst, line 661: Just to note, I was playing with leaving these in the workspace... For two reasons:
- otherwise we'd have to clone again every session reload
- even for org, but mostly for user, we'd be able to tweak them as needed.

Maybe for org it's really unimportant (because presumably rare). For user wouldn't it be good to have, a small additional repo in workspace?

enyst (approved):
Thank you! Yeah let's do this in stages 😅

openhands/runtime/base.py
rbren: > we'd have to clone again every session reload

Won't they be in the history already? I think all this happens only on the first initialization...

This seems like maybe a nice convenience though, so long as it didn't confuse the agent...

enyst: Ah, you're thinking of `repo`, those will be in history already.

On the other hand, a keyword-based microagent is added to history when it's triggered, that's when the observation is created and added. It's alive when it's observed, so to speak. 😅 
If it's never triggered during a session, there's no recall observation for it in the stream. (or if it's summarized)

I'm pretty sure that loading microagents from the runtime happens on session restore, too, so it can be triggered or re-triggered after.

If we leave them in, the workspace could look like
```
/workspace/selected_repo
/workspace/some_other_repo
/workspace/.openhands/microagents/org, user
```

Idk, it doesn't look too bad to me?

rbren: So I think we're crossing terminology a bit here

_Any_ microagent can have 0 keywords, not just those in the repo. I think today, if I have a user microagent with 0 keywords checked into `github.com/rbren/.openhands`, that will indeed appear in the history (and maybe today it'd show up in the prompt as <repository_instructions> :/ probably need to fix that)

But you're right--keyword microagents aren't getting saved in the State anywhere, so they'd need to be reloaded on every session startup (whether they belong to the repo or to the user)

enyst: Yes, all 0-keywords are `repo` type 😅

PR #8406, opened 2025-05-09
Remove github_user_id in favor of user_id

rbren:
This PR removes the `github_user_id` parameter from various functions in the codebase, as we have migrated to using `user_id` instead. The changes include:

- Updated `conversation_validator.py` to only return `user_id` instead of a tuple with both `user_id` and `github_user_id`
- Removed `github_user_id` parameters from methods in `conversation_manager.py` and its implementations
- Updated `listen_socket.py` to only use `user_id`
- Made `github_user_id` optional with a default value of `None` in `ConversationMetadata`

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f15d519-nikolaik   --name openhands-app-f15d519   docker.all-hands.dev/all-hands-ai/openhands:f15d519
```


rbren:
@openhands fix the python unit tests

rbren:
@openhands we still have tests failing with ConversationMetadata.__init__() got an unexpected keyword argument 'github_user_id'

enyst (approved):
Happy to see this cleaning up, this is redundant now.

I gave it a quick try locally because I was wondering about some value that we used to have saved, and it's okay, it doesn't exist anymore for me, it may be long gone. It works to push on GitHub with a previously saved token.

rbren:
@openhands fix the merge conflicts. Pay close attention to the intent of this PR, as well as the intent of the commits that conflict

PR #8407, opened 2025-05-09
Fix git changes endpoint response model

rbren:
This PR fixes a ResponseValidationError in the git changes endpoint. The endpoint was defined with response_model=dict[str, Any], but the get_git_changes method returns a list[dict[str, str]] | None. This PR updates the response_model to match the actual return type and wraps the list in a dictionary to ensure the response is valid.

### DIFF
--- a/openhands/server/routes/files.py
+++ b/openhands/server/routes/files.py
@@ -226,7 +226,7 @@ def zip_current_workspace(request: Request) -> FileResponse | JSONResponse:
 
 @app.get(
     '/git/changes',
-    response_model=dict[str, Any],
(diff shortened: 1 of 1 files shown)


openhands/server/routes/files.py
```diff
@@ -257,7 +257,8 @@ async def git_changes(
                 status_code=404,
                 content={'error': 'Not a git repository'},
             )
-        return changes
+        # Wrap the list in a dictionary to match the response_model
+        return {'changes': changes}
```
amanape, line 261:
```suggestion
        return changes
```

amanape (approved):

Issue #8298, opened 2025-05-06
[Bug]: Refactor the resolver resolve_issue to use app config


enyst:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

The resolver is an entry point to the application. It initializes a IssueResolver object. In its init method, the code manually creates an LLMConfig with some arguments, including some read from env.

We should refactor it to use load_app_config() instead, which will read *all* values specified in env and other sources.

Keep the refactoring minimal and focused to the point above. We just want to avoid redundant code. Adapt the init code to no longer uselessly read from env, since load_app_config does that already, but make sure to keep the overrides from args.

< IMPORTANT >

- You need to run and fix resolver unit tests, in tests/unit/resolver.
- Add a test to spot check from env.
- When you are done, run pre-commit, you know how to run it on this repository.
- Keep it minimal and respect the patterns of the code, don't do anything else, I took care of the rest.

< /IMPORTANT >

### OpenHands Installation

GitHub resolver

### Operating System

None

enyst:
This has been fixed by @kotauchisunsun in 
- https://github.com/All-Hands-AI/OpenHands/pull/8641

PR #8414, opened 2025-05-10
[refactor]: Refactor sandbox configuration setup in IssueResolver class

kotauchisunsun:

None

---

- Changed the initialization timing of SandboxConfig to reduce class variables
- Removed unnecessary variable assignments

---


openhands/resolver/resolve_issue.py
```diff
@@ -176,7 +166,53 @@ def __init__(self, args: Namespace) -> None:
         self.issue_number = args.issue_number
         self.comment_id = args.comment_id
         self.base_domain = base_domain
-        self.platform = platform
+
+    def _setup_sandbox_config(
```
enyst, line 170: I wonder if we shouldn't go the other way around, instead of create sandbox config, let it _be created for us_ normally as the application does, and only override values. What do you think about refactoring like this:
- https://github.com/All-Hands-AI/OpenHands/issues/8298

kotauchisunsun: I generally agree with that direction, and I think it would be fine to revise it in that way in the future.
However, since the current IssueResolver codebase is large and complex, it's difficult for me to make changes without altering its behavior.
Therefore, including the following PR, I plan to gradually improve it by splitting the logic step by step:
https://github.com/All-Hands-AI/OpenHands/pull/8417

Right now, I’m submitting incremental PRs to improve it, but would it be better to create a single PR that implements a broader, more comprehensive refactor?

enyst (approved):

openhands/resolver/resolve_issue.py
kotauchisunsun: I’ve added the tests!

enyst:
@malhotra5 Just a heads-up, this is refactoring isolating the `SandboxConfig` initialization and parameter override. It puts us well on the path to a future refactoring that will use the app config and only override where necessary.

enyst:
I see the other PR is merged. Thank you, let's also take this in

PR #8463, opened 2025-05-12
fix #8424: change native_tool_calling semantics

### LINKED ISSUE #8424: [proposal] modify LLMConfig.native_tool_calling semantics
I'm often using openhands with LLMs that I self-host with vLLM. Unfortunately, vLLM and litellm together don't do a great job of determining the model's capability, which is actually kind of tricky. For example, a model like llama3 or qwen may or may not support native tool calling, depending on whether the `--enable-auto-tool-choice` flag and an appropriate tool calling parser are provided to the vLLM server.

There is already a `native_tool_calling` config option in `LLMConfig`; however, its current semantics are not intuitive:

* `native_tool_calling == None`: (default) Openhands will check against its own hard-coded list of `FUNCTION_CALLING_SUPPORTED_MODELS`, and enable it if the model is listed
* `native_tool_calling == False`: Tool calling is disabled
* `native_tool_calling == True`: OH does not check against its own list of supported models. Instead, it uses `litellm.supports_function_calling(model_name)` and if that returns true, then it is enabled, otherwise, it's disabled.

The problem is if I have my own custom model provider, which I specify as e.g. `vllm_hosted/llama3.3`, that neither passes the check for OH supported models, nor does litellm correctly determine it is capable. Therefore, because of the above semantics, I cannot enable tool calling even if I set `native_tool_calling = True` in the config, which is confusing and does not seem like the expected result.

I propose redefining `native_tool_calling = ['auto', 'true', 'false']` with the following semantics:
* `native_tool_calling = false`: Native tool calling will be disabled.
* `native_tool_calling = true`: Native tool calling will be enabled and attempted, even if litellm doesn't think it's supported. (this does not cause errors in litellm, as far as I can tell).
* `native_tool_calling = auto`: OH will check if EITHER: the model is in `FUNCTION_CALLING_SUPPORTED_MODELS` or if `litellm.supports_function_calling(model_name) == True`. This should be the default config value.

Please let me know what you think! I would be happy to send a PR. The biggest downside is that this could break backwards compatibility if people are counting on Mock function calling to happen by default for models that are not in `FUNCTION_CALLING_SUPPORTED_MODELS`, but I think it's worth that corner case in service of the "principle of least surprise".

cshimmin:
Change `native_tool_calling` semantics as discussed in #8424

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -491,13 +491,8 @@ class LLM(RetryMixin, DebugMixin):
         # Handle native_tool_calling user-defined configuration
         if self.config.native_tool_calling is None:
             self._function_calling_active = model_name_supported
-        elif self.config.native_tool_calling is False:
-            self._function_calling_active = False
         else:
-            # try to enable native tool calling if supported by the model
-            self._function_calling_active = litellm.supports_function_calling(
-                model=self.config.model
-            )
+            self._function_calling_active = self.config.native_tool_calling
 
     def vision_is_active(self) -> bool:
         with warnings.catch_warnings():


enyst (approved):
Thank you, I think that's already suitable for one purpose: allow the user to decide if they want to try any LLM with tool use enabled, even if litellm doesn't know it.

I think some of the weirdness here comes from the name of that variable: it doesn't mean what models "support" tool use, it means "better with tool use, so we make it default for them".

That implies the unspoken assumption that for the rest, all the rest no matter what litellm says, _'disable tool use'_ is a reasonable default.

If we want to keep this logic, we could rename the var e.g. `USE_FUNCTION_CALLING_BY_DEFAULT` as @neubig suggested.

If we want to change this logic, it seems to me it makes most sense to just reverse it?
- keep a list of LLMs we evaled and we really think FC should be `DISABLED` by default
- for the rest, defer to litellm
- unless the user overrides, in which case just their choice matters.

This implies that we can remove our list, and replace it with the list of those that did worse in evals.

Please note that this assumes that the litellm "supported" is the best choice most of the time, which I don't know if it's accurate. 🤷‍♂️ It will be, as model providers seem to see tool use as an expected behavior?

neubig:
I think this is a positive delta, so let's go with this for now!

PR #8464, opened 2025-05-13
Add exc info to log statements

raymyers:

---

Add exc_info=True to logger statements in except blocks, so exception tracebacks are properly logged when errors occur, making debugging easier. Removing in the `{e}` in format string in some occasions since it becomes redundant.

---


enyst:
This could be useful, I'm sure. I wonder though, I seem to recall we have a mechanism intended to capture stack traces for all exceptions in logs, I think it's StackInfo or so, in `logger.py`. Maybe it doesn't work as intended?

I want to say that it doesn't necessarily show everything in console, but it should log in file logs... but I could be wrong, sorry, it's late.

PR #8310, opened 2025-05-06
Fix issue #8304: [Bug]: Non-native tool use converter fails when builtin tools are disabled

openhands-agent:
This pull request fixes an issue where the fn_call_converter.py raises an exception if the user disables some tools.

The issue has been successfully resolved based on the changes made. Here's why:

1. The core problem was that the system was raising a FunctionCallConversionError when the tool set didn't match the in-context learning examples, preventing users from disabling tools they didn't want to use.

2. The changes made:
- Modified fn_call_converter.py to convert the error into a warning
- Added proper logger import to handle the warning
- Maintained the same validation check but made it non-blocking

3. Expected impact:
- Users can now disable tools/microagents without the system throwing an error
- The warning still provides useful information about potential compatibility issues
- The core functionality remains intact while providing more flexibility
- The original use case (giving git instructions without microagent interference) will now work as desired

The changes directly address the user's need to disable unwanted functionality while preserving the system's ability to warn about potential compatibility issues, making this a complete resolution of the reported issue.

Automatic fix generated by [OpenHands](https://github.com/All-Hands-AI/OpenHands/) 🙌

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:bb767e8-nikolaik   --name openhands-app-bb767e8   docker.all-hands.dev/all-hands-ai/openhands:bb767e8
```

---

Fix #8304

PR #8487, opened 2025-05-13
Fix finish tool in-context example to include required parameters

### LINKED ISSUE #8476: [Bug]: In-context example does not match tool specification

The [finish tool](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/agenthub/codeact_agent/tools/finish.py#L25) in `codeact_agent` requires two parameters `['message', 'task_completed']`, but the in context example [here](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/llm/fn_call_converter.py#L221) does not include these two parameters.

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Operating System

None

_No response_

neubig:
This PR fixes issue #8476 by updating the in-context example for the finish tool to include the required parameters (`message` and `task_completed`). The finish tool requires these parameters as specified in its definition, but they were missing from the example in `fn_call_converter.py`.

Fixes #8476

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/13c69b4847e54714a8b8113cd490e432)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:5f8514c-nikolaik   --name openhands-app-5f8514c   docker.all-hands.dev/all-hands-ai/openhands:5f8514c
```

### DIFF
--- a/openhands/llm/fn_call_converter.py
+++ b/openhands/llm/fn_call_converter.py
@@ -217,8 +217,11 @@ USER: EXECUTION RESULT of [execute_bash]:
 [1] 126
 
 ASSISTANT:
-The server is running on port 5000 with PID 126. You can access the list of numbers in a table format by visiting http://127.0.0.1:5000. Let me know if you have any further requests!
 <function=finish>
+<parameter=message>
+The server is running on port 5000 with PID 126. You can access the list of numbers in a table format by visiting http://127.0.0.1:5000. Let me know if you have any further requests!
+</parameter>
+<parameter=task_completed>true</parameter>
 </function>
 
 --------------------- END OF EXAMPLE ---------------------


neubig:
LGTM but need another review

enyst:
Just a heads-up, this issue was fixed in this PR:
https://github.com/All-Hands-AI/OpenHands/pull/8310/files#diff-1dc478aa7a32d2d031cd7183aaca8b76c82f83dee0a7f4d531e801299fdc4363R231

I'm running an eval on Qwen 3 for that PR, Xingyao requested it since Qwen 3 is the best or among the best results we have for non-native FC.

neubig:
@enyst , sorry I didn't fully understand how that PR fixed the problem of the in-context example not having appropriate arguments specified. Could you elaborate?

enyst (approved):
Oh, it's simply that the PR refactored all this long string we have for in-context examples, into multiple strings, one per tool. So then we assemble those for enabled tools, and avoid those for disabled tools. 
While at it, the agent added the missing parameters in the Finish in-context example.

Up to you if you want to merge this PR. Is there a chance that *this bit* alone changes results? Even if it is, it seems we have to do it, because we did change Finish already and it really has these attributes?

For what is worth, we had some intuitions that the Finish tool may be understood / preferred differently by different LLMs. For example, something in our newest definitions makes openai models more likely to finish soon (with task possibly "not completed").🤷
But we already did it, this is just an example string.

neubig:
I'll run an eval just in case!

enyst:
@neubig Please note that the eval-50 runs on Sonnet 3.7, for which we don't use this code: the in-context examples are inserted in the prompt only for LLMs with non-native function calling. I ran on Qwen3 [here](https://github.com/All-Hands-AI/OpenHands/actions/runs/15069599682/job/42362341689?pr=8536#step:6:47) and it seems the result is slightly lower than the last result we had. (34.70% vs 36.2%)

FWIW there were errors due to the parameters of Finish, though personally I still think that it's unlikely they're relevant for results (but there's almost nothing else to suspect, too... 🤔), and I do agree with you we seem to need to add them to the example (because they are mandatory in the action already). 🤔

mamoodi:
- submitted instances: 50
- empty patch instances: 0
- resolved instances: 24
- unresolved instances: 26
- error instances: 0 - Download at: https://github.com/All-Hands-AI/pr-eval-results/releases/download/1.0.0/fix-finish-tool-example-28_25-05-16-15-28.tar.gz

neubig:
Ahh, right, sorry. I wasn't thinking carefully enough when I kicked off the run. I kinda feel that since these are required parameters it might be best to add them (we were encountering problems using local models without them), but maybe @xingyaoww has a second opinion?

xingyaoww:
The eval results here are less trust worthy: we are still using eval_infer_remote.sh, so the number is not accurate. @ryanhoangt is working on improving the infer script 🙏

xingyaoww:
But i guess this change will indeed be overwritten by @enyst's PR though

PR #8489, opened 2025-05-13
Update default model to sonnet 3.7 in all applicable places

mamoodi:

Update default model to sonnet 3.7 in all applicable places

---

---

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -1,8 +1,8 @@
 # Development Guide
 
 This guide is for people working on OpenHands and editing the source code.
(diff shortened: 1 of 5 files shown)


mamoodi:
@openhands the Run Frontend Unit Tests are failing on this PR. Read the error and fix it

xingyaoww (approved):

amanape (approved):
We should probably move this into some `/utils/defaults.ts` and update all references with the new constant (e.g., `const DEFAULT_LLM_MODEL = "claude-3-7-sonnet-20250219"`)

That way we can just update a single file when changing default models

PR #8500, opened 2025-05-14
Fix workspace mount behavior with SANDBOX_VOLUMES

### LINKED ISSUE #8239: Add support for custom volume mounts in DockerRuntime

xingyaoww:

This PR fixes an issue with how volume mounts work in the Docker runtime. Previously, when using SANDBOX_VOLUMES, the first volume would always be treated as the workspace mount, which caused problems when users wanted to mount read-only volumes without affecting the workspace. Now, only volumes explicitly mounted to `/workspace` will be treated as workspace mounts, giving users more control over their volume configurations.

---

- When `/workspace` is not explicitly found in SANDBOX_VOLUMES, we no longer force the first volume to be a workspace mount
- This allows users to mount volumes without affecting the workspace
- Added tests to verify the new behavior

Previously, the first volume in SANDBOX_VOLUMES would always be treated as a workspace mount, which caused issues when users wanted to mount read-only volumes without affecting the workspace.

With this change, only volumes explicitly mounted to `/workspace` will be treated as workspace mounts. If no such mount is found, the default workspace path in the sandbox will be used.

---

Fixes #8239

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:9f5a277-nikolaik   --name openhands-app-9f5a277   docker.all-hands.dev/all-hands-ai/openhands:9f5a277
```

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e3c5de0e698742f0a08b4afd7f31674a)

### DIFF
--- a/openhands/core/config/utils.py
+++ b/openhands/core/config/utils.py
@@ -627,7 +627,6 @@ def register_custom_agents(config: AppConfig) -> None:
     This function is called after configuration is loaded to ensure all custom agents
     specified in the config are properly imported and registered.
     """
-
     # Import here to avoid circular dependency
     from openhands.controller.agent import Agent
 
--- a/openhands/core/config/utils.py
+++ b/openhands/core/config/utils.py
@@ -302,25 +302,33 @@ def finalize_config(cfg: AppConfig) -> None:
         # Split by commas to handle multiple mounts
         mounts = cfg.sandbox.volumes.split(',')
 
-        # Use the first volume for backward compatibility
-        if mounts:
-            primary_mount = mounts[0]
-            parts = primary_mount.split(':')
-            if len(parts) < 2 or len(parts) > 3:
-                raise ValueError(
-                    f'Invalid sandbox.volumes format: {primary_mount}. '
-                    f"Expected format: 'host_path:container_path[:mode]', e.g. '/my/host/dir:/workspace:rw'"
-                )
+        # Check if any mount explicitly targets /workspace
+        workspace_mount_found = False
+        for mount in mounts:
+            parts = mount.split(':')
+            if len(parts) >= 2 and parts[1] == '/workspace':
+                workspace_mount_found = True
+                host_path = os.path.abspath(parts[0])
 
-            host_path = os.path.abspath(parts[0])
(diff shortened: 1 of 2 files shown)


xingyaoww:
@OpenHands can you update the github PR description in this PR to follow the template

enyst (approved):
This is more expected behavior for /workspace, thank you!

PR #8502, opened 2025-05-14
fix: validation error when saving SSE MCP server url on the UI

ryanhoangt:

---

For SSE server configs, we allow both plain string and an object to be passed to the "sse_servers" array, but currently this only works when loading from `config.toml` because we handle both types separately in `from_toml_section`. When passing something like `"sse_servers": ["https://example-mcp-server.com/sse"]` and saving on the UI, it throws 422 error.

Instruction on the UI:
```
Example: { "sse_servers": ["https://example-mcp-server.com/sse"], "stdio_servers": [{ "name": "fetch", "command": "uvx", "args": ["mcp-server-fetch"] }] }
```

---

### DIFF
--- a/openhands/core/config/mcp_config.py
+++ b/openhands/core/config/mcp_config.py
@@ -1,6 +1,6 @@
 from urllib.parse import urlparse
 
-from pydantic import BaseModel, Field, ValidationError
+from pydantic import BaseModel, Field, ValidationError, model_validator
 
 
 class MCPSSEServerConfig(BaseModel):
@@ -43,6 +43,21 @@ class MCPConfig(BaseModel):
     stdio_servers: list[MCPStdioServerConfig] = Field(default_factory=list)
 
     model_config = {'extra': 'forbid'}
+
+    @model_validator(mode='before')
+    def convert_string_urls(cls, data):
+        """Convert string URLs to MCPSSEServerConfig objects."""
+        if isinstance(data, dict) and 'sse_servers' in data:
+            servers = []
+            for server in data['sse_servers']:
+                if isinstance(server, dict):
+                    servers.append(server)
+                elif isinstance(server, str):
+                    servers.append({'url': server})
+                else:
+                    servers.append(server)  # Pass through any other type for validation to catch
+            data['sse_servers'] = servers
+        return data
 
     def validate_servers(self) -> None:
         """Validate that server URLs are valid and unique."""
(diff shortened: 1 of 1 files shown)


openhands/core/config/mcp_config.py
```diff
@@ -44,6 +44,21 @@ class MCPConfig(BaseModel):
 
     model_config = {'extra': 'forbid'}
 
+    @model_validator(mode='before')
+    def convert_string_urls(cls, data):
+        """Convert string URLs to MCPSSEServerConfig objects."""
+        if isinstance(data, dict) and 'sse_servers' in data:
+            servers = []
+            for server in data['sse_servers']:
+                if isinstance(server, dict):
```
enyst, line 53: Correct me if I'm misunderstanding the problem here. It looks like we do in this method the same thing we did in `load_from_toml`?

In the source of this data, is it possible to also enter `stdio_servers` or will it be?

Could we reuse the load method, if it wasn't named that way?

ryanhoangt: > It looks like we do in this method the same thing we did in load_from_toml?

Yeah, the handling for SSE config is similar I think. Here we use direct initialization of the object so we miss the conversion:

https://github.com/All-Hands-AI/OpenHands/blob/67d438ea4f52afea58f71b0af39341a9cfa5d3a8/openhands/server/routes/settings.py#L182

> In the source of this data, is it possible to also enter stdio_servers or will it be?

For `stdio_servers`, we only allow config objects so it should work and we don't need any special conversion for it.

> Could we reuse the load method, if it wasn't named that way?

Just to clarify, which load method are you referring to?

enyst: The `load_from_toml` method 🤔

xingyaoww: Yeah seems like a good idea - maybe we can factor the common component out from `load_from_toml` and re-use this helper function in this place as well

xingyaoww (approved):
Thanks!

openhands/core/config/mcp_config.py
```diff
@@ -77,31 +95,28 @@ def from_toml_section(cls, data: dict) -> dict[str, 'MCPConfig']:
...
-
+            
             # Convert all entries in stdio_servers to MCPStdioServerConfig objects
             if 'stdio_servers' in data:
                 servers = []
                 for server in data['stdio_servers']:
                     servers.append(MCPStdioServerConfig(**server))
                 data['stdio_servers'] = servers
-
+
```
enyst, line 110: Looks good! Just a question here, it seems the editor inserted spaces, and precommit didn't catch them and auto fix
?

It might be useful to run a make build to (re?) install them

PR #8511, opened 2025-05-15
Fix #8510: Improve error messages for invalid microagent format

### LINKED ISSUE #8510: [Bug]: Invalid microagent format results in opaque ValueError in UI

When a microagent's format fails Pydantic validation, it results in an opaque ValueError in the UI and the agent stopping:

Image

```
Failed to create agent session: ValueError
Failed to create agent session: ValueError
```

Here is a repo that an example of this:
https://github.com/neubig/test-repo/issues/6

The [offending microagents file](https://github.com/neubig/test-repo/blob/a6544c7e1dffee12044a2f1438c7ea63a442737c/.openhands/microagents/git_helper.md) has frontmatter with an invalid type "task":

```md
---
name: git_helper
type: task
version: 1.0.0
agent: CodeActAgent
triggers:
- git
- github
---
...
```

The desired behavior would be to instead pass the pydantic validation error and the name of the offending microagent file in the UI so the user immediately knows the cause of the problem and can fix it.

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Operating System

None

_No response_

neubig:
This PR fixes issue #8510 by improving error messages for invalid microagent formats.

## Changes

1. Enhanced error handling in `microagent.py` to provide more detailed error messages when microagent validation fails
2. Updated the error handling in `session.py` to display the full error message for microagent validation errors instead of just the error class name
3. Added a new test case to verify that invalid microagent types produce helpful error messages

With these changes, when a microagent has an invalid format (like an unsupported `type` value), the UI will show a detailed error message that includes the name of the offending file and what went wrong, making it easier for users to fix the issue.

Closes #8510

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/c6503d76275447a0acc71374595ca1d0)

### DIFF
--- a/openhands/microagent/microagent.py
+++ b/openhands/microagent/microagent.py
@@ -65,7 +65,14 @@ class BaseMicroagent(BaseModel):
         try:
             metadata = MicroagentMetadata(**metadata_dict)
         except Exception as e:
-            raise MicroagentValidationError(f'Error loading metadata: {e}') from e
+            # Provide more detailed error message for validation errors
+            error_msg = f'Error validating microagent metadata in {path.name}: {str(e)}'
+            if 'type' in metadata_dict and metadata_dict['type'] not in [
+                t.value for t in MicroagentType
+            ]:
+                valid_types = ', '.join([f'"{t.value}"' for t in MicroagentType])
+                error_msg += f'. Invalid "type" value: "{metadata_dict["type"]}". Valid types are: {valid_types}'
+            raise MicroagentValidationError(error_msg) from e
 
         # Create appropriate subclass based on type
         subclass_map = {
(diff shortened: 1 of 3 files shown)


neubig:
I have confirmed that this works in the frontend, it's ready for review.

Screenshot 2025-05-14 at 9 23 13 PM

enyst (approved):
This solves the issue of too little information, and that's important. Thank you!

I think we may want to tweak it further, though not necessarily in this PR. That's because now it seems to go a bit much in the other direction: for example, the pydantic note at the end doesn't seem necessary. It also repeats the info about type, not sure we have to?

Still, it's much better than showing a cryptic message IMHO 😅

Issue #5968, opened 2025-01-02
[Bug]: fail Waiting for client to become ready...


chenglu66:
### Is there an existing issue for the same bug?

- [X] I have checked the existing issues.

### Describe the bug and reproduction steps

when i use make docker and make bulid and make run , show Waiting for client to become ready..., and does not go on

### OpenHands Installation

Docker command in README

### Operating System

None

### Logs, Errors, Screenshots, and Additional Context

INFO:     172.18.0.1:38276 - "POST /api/conversations HTTP/1.1" 200 OK
INFO:     ('172.18.0.1', 38284) - "WebSocket /socket.io/?latest_event_id=-1&conversation_id=a989d29db53f479db2eafaafb461c76d&EIO=4&transport=websocket" [accepted]
INFO:     172.18.0.1:38276 - "GET /api/options/config HTTP/1.1" 200 OK
07:31:30 - openhands:INFO: listen_socket.py:26 - sio:connect: 7r8dqN7ApDcuNCJuAAAF
07:31:30 - openhands:INFO: manager.py:209 - join_conversation:a989d29db53f479db2eafaafb461c76d:7r8dqN7ApDcuNCJuAAAF
07:31:30 - openhands:INFO: manager.py:364 - _get_event_stream:a989d29db53f479db2eafaafb461c76d
07:31:30 - openhands:INFO: manager.py:367 - found_local_agent_loop:a989d29db53f479db2eafaafb461c76d
07:31:31 - openhands:INFO: docker_runtime.py:133 - [runtime a989d29db53f479db2eafaafb461c76d] Starting runtime with image: ghcr.io/all-hands-ai/runtime:0.17-nikolaik
07:31:31 - openhands:INFO: docker_runtime.py:137 - [runtime a989d29db53f479db2eafaafb461c76d] Container started: openhands-runtime-a989d29db53f479db2eafaafb461c76d. VSCode URL:
None
07:31:31 - openhands:INFO: docker_runtime.py:145 - [runtime a989d29db53f479db2eafaafb461c76d] Waiting for client to become ready at http://localhost:35402...
INFO:     172.18.0.1:38298 - "GET /api/settings HTTP/1.1" 200 OK
INFO:     172.18.0.1:38308 - "GET /api/options/config HTTP/1.1" 200 OK
07:31:45 - openhands:INFO: manager.py:434 - _cleanup_session:a989d29db53f479db2eafaafb461c76d
[...]

PR #8512, opened 2025-05-15
Fix for docker-compose on Linux

dangerousbeans:
# Summary
Added network_mode: host for Linux compatibility in docker-compose.yml
Removed explicit port mappings and host.docker.internal configuration
Added DOCKER_HOST_ADDR environment variable for Linux networking

# Test plan
Verify Docker Compose works correctly on Linux environments
Ensure proper connection between containers

Fixes https://github.com/All-Hands-AI/OpenHands/issues/5968

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,7 @@
 
 services:
   openhands:
+    network_mode: host
     build:
       context: ./
       dockerfile: ./containers/app/Dockerfile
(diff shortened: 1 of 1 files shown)


enyst (commented):
Thank you for the contribution @dangerousbeans !

Have you tried it yourself? It seems to me that this file is a bit outdated, for example we don't use "latest" tag anymore, I'd suggest to use the release tag.

dangerousbeans:
@enyst heya yep runs for me on Manjaro

enyst:
Sorry for the delay, @dangerousbeans . I'm not sure about these changes, because as far as I know, the current compose mostly works for people, and, where it doesn't work is `image: openhands:latest` at least, which this PR doesn't change. I think it should be the versioned last release instead, because we don't build 'latest' for a really long time now, so it's obsolete.

Could you explain the changes?

neubig:
Closing due to lack of activity, but happy to follow up.

PR #8513, opened 2025-05-15
Faster runtime tests

tofarr:
Still in draft.

By reusing the runtimes rather than creating a new one for each test, we can massively reduce the time taken for runtime tests.

---

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:4119f6d-nikolaik   --name openhands-app-4119f6d   docker.all-hands.dev/all-hands-ai/openhands:4119f6d
```


tests/runtime/conftest.py
```diff
@@ -60,6 +55,7 @@ def _close_test_runtime(runtime: Runtime) -> None:
         runtime.close(rm_all_containers=False)
     else:
         runtime.close()
+    call_async_from_sync(runtime.__class__.delete, GENERAL_TIMEOUT, runtime.sid)
```
tofarr, line 58: I noticed that the docker containers produced by these tests are not deleted locally, so added a call here.

PR #8516, opened 2025-05-15
[refactor]: Refactor IssueResolver and SandboxConfig for Better Modularity and Testability

kotauchisunsun:

---

- Split IssueResolver into a separate file
- Decoupled IssueResolver from args to reduce dependency on command-line options
- Refactored the creation of SandboxConfig and restructured its setup to streamline tests

---


kotauchisunsun:
I have prepared a sample repository as shown below:
https://github.com/kotauchisunsun/openhands-debug-demo

You can see that the modified code is being installed in the following line:
https://github.com/kotauchisunsun/openhands-debug-demo/blob/main/.github/workflows/openhands-resolver.yml#L241

Here is the actual issue that was created:
https://github.com/kotauchisunsun/openhands-debug-demo/issues/13

Here is the result of the execution:
https://github.com/kotauchisunsun/openhands-debug-demo/actions/runs/15043363877

And here is the pull request:
https://github.com/kotauchisunsun/openhands-debug-demo/issues/13

enyst (changes requested):
Thank you for the work on this!

Instead of a new `SandboxContainerConfig` class, could we use its arguments for the real `SandboxConfig` creation?

I'll come back on this, I'm sorry, due to the move the diff is not easy to read directly 😅

kotauchisunsun:
@enyst 
Thank you for the feedback!

> Instead of a new SandboxContainerConfig class, could we use its arguments for the real SandboxConfig creation?

I understand that this would be difficult. The SandboxContainerConfig was implemented as a value object specifically for creating the SandboxConfig used by IssueResolver. In that context, there's logic that throws an error when both runtime_container_image and base_container_image are specified at the same time:
https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/resolver/resolve_issue.py#L185-L195

I'm not certain whether this constraint should apply globally to all uses of SandboxConfig, or if it's specific to IssueResolver. If it's a valid constraint across all usages, then it would make sense to move that validation logic into SandboxConfig itself.

Alternatively, SandboxContainerConfig simply exists to satisfy that constraint and to provide a default value for runtime_container_image tailored for IssueResolver:
https://github.com/All-Hands-AI/OpenHands/pull/8516/files#diff-712ee9a1e6eb7f247083826a43571568c88e2f3663eb0ce26051d55f8093a18eR44-R51
I split it out in order to write comprehensive tests for it.

> I'll come back on this, I'm sorry, due to the move the diff is not easy to read directly 😅

No worries at all — I understand the diff is hard to read right now. To better understand the refactoring, I recommend checking the commit history instead:

Source code was splited here:
https://github.com/All-Hands-AI/OpenHands/pull/8516/commits/557b645164e2d90755371f7052eab675e5c37286

Then I applied the validation logic change to SandboxContainerConfig here:
https://github.com/All-Hands-AI/OpenHands/pull/8516/commits/21577c723f74c84049081c0fa3a72598d513a9fb

tests/unit/resolver/test_resolve_issue.py
```diff
@@ -10,8 +10,9 @@
 def assert_sandbox_config(
     config: SandboxConfig,
     base_container_image = SandboxConfig.model_fields["base_container_image"].default,
-    runtime_container_image = f'ghcr.io/all-hands-ai/runtime:mock-nikolaik',  # Default to mock version
+    runtime_container_image = "ghcr.io/all-hands-ai/runtime:0.38.0-nikolaik",
```
enyst: Sorry, I don't think this will work: we had an issue the other day, these tests were failing when we tried to release a new version of `openhands`.

I think we shouldn't hardcode the (real) version, nor the user (e.g. on my Mac the user_id here is 502) so the test is failing locally.

kotauchisunsun: @enyst 
> Sorry, I don't think this will work: we had an issue the other day, these tests were failing when we tried to release a new version of openhands.
I'm sorry for causing issues that affected the release.

> I think we shouldn't hardcode the (real) version, nor the user (e.g. on my Mac the user_id here is 502) so the test is failing locally.
I've modified the version information so that it works with a minimal mock instead of being hardcoded.
As for the user_id, I updated the assert_sandbox_config function so that it only performs the check when user_id is explicitly specified.

By the way, regarding the user_id — do you think there's an issue in the implementation of SandboxConfig itself?
https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/core/config/sandbox_config.py#L56

My original code was written to retrieve the default value:
https://github.com/kotauchisunsun/OpenHands/commit/31b2f3c9c246ff8e5502eb692daf4c1f9ef816f9#diff-fd80cef1f8ee36908d45fd51ef9879d90c0cfb53c669c8362ac4a9160fec2a8dR15

So I assumed that the test wouldn't fail depending on the OS.
The os.getuid function appears to be available on macOS as well:
https://docs.python.org/3.12/library/os.html#os.getuid

Do you have any idea why this issue occurred? I'd really appreciate it if you could share the reason.

enyst: The test was testing against 1000, I think. `getuid` works, and it returns my user id on my local system when run locally. I think we simply can't assume that it will fallback to 1000, because it won't necessarily?

kotauchisunsun: As a result of attempting several experiments to solve this issue, I found that it is difficult to resolve.

I had assumed that in `assert_sandbox_config`, I could retrieve `SandboxConfig.model_fields['user_id'].default` and use it to validate the `user_id`.
However, the `SandboxConfig` class inherits from Pydantic's `BaseModel` and assigns a value to `user_id` as part of the "class definition".

https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/core/config/sandbox_config.py#L56

Due to the mechanism of inheritance in `Pydantic`, the value for `user_id` is likely assigned at the time the module is loaded.
As a result, I discovered that mocking the value does not work properly for `os.getuid`.
Although it may be technically possible to override the value with a more complex procedure, I decided not to pursue that path, as it could make the test code unstable.

kotauchisunsun:
@enyst 
I have one suggestion.
This pull request has been long overdue. Therefore, I would like to terminate this pull request and split the contents.

1. split the IssueResolver class into files
2. change the constructor of the IssueResolver class

I'm going to exclude the controversial parts and proceed with these steps, what do you think?

kotauchisunsun:
@enyst 
I split my pull request. Once I close this pull request. Thank you for review!
https://github.com/All-Hands-AI/OpenHands/pull/8619

PR #8517, opened 2025-05-15
Remove obsolete task.py file 

emmanuel-ferdman:
# PR Summary
This small PR evaluates and displays the task information in debug mode.


enyst (commented):
Oh, thank you for the PR, but I don't think it does that. Pretty sure this code hasn't been used in a long time.

What do you think if we delete it? So that we make room for the new 😅

The entire file, most likely, is unused.

emmanuel-ferdman:
@enyst sure, should I simply remove `openhands/controller/state/task.py` from the code base then? :) Although I see `task.py` was updated 3 weeks ago in commit dc91cb263bca560d3d62a3b8d0cd7f0f7a59bcab

enyst:
We may want to check if there are unit tests using it, CI may complain

PR #8527, opened 2025-05-15
Add info logging for 401 Unauthorized responses

rbren:
This PR adds info logging with user_id (when available) to all places where 401 status codes are returned in the OpenHands server routes. This will help with debugging authentication issues.

Changes:
- Added info logs in settings.py for 401 responses
- Added info logs in secrets.py for 401 responses
- Attempted to add logs in git.py but encountered syntax issues with the pre-commit hooks

### DIFF
--- a/openhands/server/routes/git.py
+++ b/openhands/server/routes/git.py
@@ -2,6 +2,7 @@ from fastapi import APIRouter, Depends, status
 from fastapi.responses import JSONResponse
 from pydantic import SecretStr
 
+from openhands.core.logger import openhands_logger as logger
 from openhands.integrations.provider import (
     PROVIDER_TOKEN_TYPE,
     ProviderHandler,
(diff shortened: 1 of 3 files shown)


rbren:
@openhands fix the python tests, then fix the python lint errors

openhands/resolver/resolve_issue.py
```diff
@@ -220,6 +220,8 @@ def _setup_sandbox_config(
             user_id = os.getuid() if hasattr(os, 'getuid') else 1000
             if user_id == 0:
                 sandbox_config.user_id = get_unique_uid()
+            else:
+                sandbox_config.user_id = user_id
```
enyst, line 224: Is this intended? I think it's duplicate, b/c we are doing this anyway:
https://github.com/All-Hands-AI/OpenHands/blob/8ec5d0e0438ea7daa8ff8d3e6eae1576cb259076/openhands/core/config/sandbox_config.py#L56

enyst (approved):

openhands/resolver/resolve_issue.py
rbren: yeah this confused me too

rbren, line 224:
```suggestion
```

PR #8535, opened 2025-05-16
[Fix]: don't access secrets if doesn't exist

malhotra5:

This PR fixes the case where secrets don't exist but we try to access them anyways

---

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:7548549-nikolaik   --name openhands-app-7548549   docker.all-hands.dev/all-hands-ai/openhands:7548549
```


malhotra5:
@openhands fix the failing lint issues in the BE

openhands/__init__.py
```diff
@@ -21,13 +21,7 @@ def get_version():
...
         pass
 
-    try:
-        from pkg_resources import DistributionNotFound, get_distribution
-
-        return get_distribution(__package_name__).version
-    except (ImportError, DistributionNotFound):
-        pass
-
+    # Skip pkg_resources to avoid mypy issues
```
enyst, line 24: This may be worth looking into 🫠

enyst:
I'm not even sure what's the best solution here: it seems that dependency is quite old? and maybe there is a better / newer way to do that little thing it's doing?

mamoodi (approved):

PR #8540, opened 2025-05-16
Fix: Only show login modal for genuine 401 errors, not connection issues

rbren:
## Description
This PR fixes an issue where the login modal was being shown when there were connection issues, not just when there was a 401 error.

## Changes
- Modified the logic in `root-layout.tsx` to only show the login modal for genuine 401 errors
- Added logic to distinguish between 401 errors (which should show the login modal) and other errors like connection issues (which should not)
- Fixed lint issues and improved type safety
- Kept the original behavior in `query-client-config.ts` as requested

## Testing
The login modal will now only appear when there is a genuine 401 error from the server, not when there are connection issues.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:03d0b8d-nikolaik   --name openhands-app-03d0b8d   docker.all-hands.dev/all-hands-ai/openhands:03d0b8d
```


rbren:
@openhands modify frontend/src/hooks/query/use-is-authed.ts so that
* it returns true if the request succeeds
* it returns false if the request is 401
* it is in an error state otherwise

amanape (approved):

PR #8531, opened 2025-05-15
Add conversationUrl static variable with getter and setter methods

tofarr:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [x] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

This PR adds support for custom conversation URLs, allowing OpenHands to work with conversations hosted on different servers or endpoints. This enables integration with external conversation providers, supports custom deployment scenarios, and facilitates development and testing with conversations from different environments.

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

## Conversation URL Override Feature

This PR implements a mechanism to override the default conversation URL pattern, enabling OpenHands to interact with conversations hosted on different servers or endpoints.

#### Frontend
- Added static variable `conversationUrl: string | null = null` to the OpenHands class
- Added getter method `getConversationUrl()` and setter method `setConversationUrl(url: string | null)`
- Modified API endpoint access pattern to use `this.conversationUrl || "/api/conversations/${conversationId}"`
- Updated WebSocket connection to use the host from the conversation URL if it is an absolute URL
- Added conversation URL setting in `useUserConversation` hook to automatically configure the URL when a conversation is loaded

#### Backend
- Updated `_get_conversation_url` in `StandaloneConversationManager` to return `/api/conversations/${conversationId}` (previously was `/conversations/${conversationId}`)

### Files Modified
- `frontend/src/api/open-hands.ts` - Added static variable and methods, updated endpoint access
- `frontend/src/api/open-hands.types.ts` - Added URL field to Conversation interface
- `frontend/src/api/file-service/file-service.utils.ts` - Updated endpoint access
- `frontend/src/hooks/query/use-active-host.ts` - Updated endpoint access
- `frontend/src/hooks/query/use-user-conversation.ts` - Added conversation URL setting
- `frontend/src/routes/conversation.tsx` - Updated endpoint access
- `frontend/src/components/features/conversation-panel/conversation-card.tsx` - Updated endpoint access
- `frontend/src/context/ws-client-provider.tsx` - Updated WebSocket connection to use conversation URL
- `openhands/server/conversation_manager/standalone_conversation_manager.py` - Updated URL format

### Testing Instructions
To run this PR locally, use the following command:
```
[...]

PR #8550, opened 2025-05-16
Add Session API Key Authentication for Runtime Communication

tofarr:

* LocalRuntime now implements the SESSION_API_KEY from the environment correctly.

---

This PR implements a session-based API key authentication mechanism for secure communication between the OpenHands server and runtime environments:

- Added SessionApiKeyMiddleware to validate API keys in HTTP requests
- Implemented API key validation for WebSocket connections
- Updated the LocalRuntime to include the session API key in request headers
- Modified conversation management to properly handle and pass API keys
- Refactored agent loop info handling to include API key information

The implementation uses an environment variable (SESSION_API_KEY) to control the feature. When not set, the system falls back to the previous behavior (no API key validation). API key validation is skipped for health check endpoints (/alive and /server_info) to ensure monitoring systems continue to work.

---

N/A

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:31fb45b-nikolaik   --name openhands-app-31fb45b   docker.all-hands.dev/all-hands-ai/openhands:31fb45b
```


openhands/server/routes/manage_conversations.py
```diff
@@ -365,7 +366,8 @@ async def _get_conversation_info(
                 ConversationStatus.RUNNING if is_running else ConversationStatus.STOPPED
             ),
             num_connections=num_connections,
-            url=url,
+            url=agent_loop_info.url if agent_loop_info else None,
+            api_key=agent_loop_info.api_key if agent_loop_info else None,
```
enyst, line 370: Totally trivial: could we name the session api key `session_api_key` everywhere? We have three api keys so far? It seems to me prone to lead to bugs based on wrong assumptions that it's llm or whatnot.

enyst, line 370: It looks like this goes to the frontend? Why do we need to send keys from backend to frontend?

tofarr: In the NestedRuntime (coming this week):
* The agentloop exists inside the docker container.
* The main app server is used only to start / stop / monitor runtimes.
* Clients connect directly to the runtime.

e.g.:
I start a new conversation, and the web cllent connects directly to my runtime docker container (In this case running on port 33836):
image

This will be possible (but not the default yet) after https://github.com/All-Hands-AI/OpenHands/pull/8531 is merged and I am finished working on the feat-nested-conversation branch.

tofarr: We only populate this value in the NestedRuntimeManagers - in other cases the client does not need to see the session_api_key, so it does not see it.

tofarr: Good point - I have renamed this.

raymyers:
In SESSION_API_KEY, does "session" refer to a login session or a conversation?

tofarr:
> In SESSION_API_KEY, does "session" refer to a login session or a conversation?

It is derived from its usage in the X-SESSION-KEY header. It is a key to access a nested runtime. (See its usage in action_execution_server.py)

This change just allows exposing it as part of the conversation API

openhands/server/listen_socket.py
```diff
@@ -160,3 +164,13 @@ async def oh_action(connection_id: str, data: dict[str, Any]) -> None:
 async def disconnect(connection_id: str) -> None:
     logger.info(f'sio:disconnect:{connection_id}')
     await conversation_manager.disconnect_from_session(connection_id)
+
+
+def _invalid_session_api_key(query_params: dict[str, list[Any]]):
+    session_api_key = os.getenv('SESSION_API_KEY')
```
rbren, line 170: is it possible to get this via `environ.get('HTTP_X_SESSION_API_KEY')` so that (a) it's consistent with the REST routes and (b) we're not exposing the API key in the query string?

rbren (approved):

openhands/server/listen_socket.py
tofarr: Unfortunately not when using websockets as an underlying protocol as the API does not support this:
* https://stackoverflow.com/a/4361358/470062
* https://stackoverflow.com/a/35381472/470062 (See the comment on the answer where they not that SocketIO does not support this in browser because the browser does not support it)

rbren: We're getting the Authorization header at least

PR #8559, opened 2025-05-17
Limit size of browser events

csmith49:

---

While browsing certain sites, the agent can produce events in the MBs. The majority of this data seems to come from storing the DOM object, although this is not referenced anywhere I can find.

This PR makes it so the default browser observations no longer store the DOM object (they just use the default `{}`). In addition, it also updates the screenshot serialization to place an upper-bound on the size (at 1 MB).

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:3e5ff8a-nikolaik   --name openhands-app-3e5ff8a   docker.all-hands.dev/all-hands-ai/openhands:3e5ff8a
```

### DIFF
--- a/openhands/runtime/browser/base64.py
+++ b/openhands/runtime/browser/base64.py
@@ -4,40 +4,15 @@ from PIL import Image
 import numpy as np
 
 def image_to_png_base64_url(
-    image: np.ndarray | Image.Image,
-    add_data_prefix: bool = False,
-    max_size_kb: int = 1024,
-    minimum_quality: int = 10,
+    image: np.ndarray | Image.Image, add_data_prefix: bool = False
 ) -> str:
-    """Convert a numpy array to a base64 encoded png image url.
-    
-    Will not guarantee the size of the image, but will attempt to compress it to below the specified maximum size.
-
-    Args:
-        image: The image to convert.
-        add_data_prefix: Whether to add the data prefix to the base64 string.
-        max_size_kb: Maximum size of the image in kilobytes.
-        minimum_quality: Minimum quality of the image.
-    """
+    """Convert a numpy array to a base64 encoded png image url."""
     if isinstance(image, np.ndarray):
         image = Image.fromarray(image)
     if image.mode in ('RGBA', 'LA'):
         image = image.convert('RGB')
-
-    max_size_bytes = max_size_kb * 1024
-    quality = 85
-
-    while True:
-        buffered = io.BytesIO()
-        image.save(buffered, format='PNG', quality=quality)
-
-        if len(buffered.getvalue()) <= max_size_bytes:
-            break
-
-        # If the buffer is too large, reduce the quality.
-        quality -= 5
-        if quality < minimum_quality:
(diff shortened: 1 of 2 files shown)


openhands/runtime/browser/base64.py
```diff
@@ -4,15 +4,40 @@
 import numpy as np
 
 def image_to_png_base64_url(
-    image: np.ndarray | Image.Image, add_data_prefix: bool = False
+    image: np.ndarray | Image.Image,
+    add_data_prefix: bool = False,
+    max_size_kb: int = 1024,
+    minimum_quality: int = 10,
```
li-boxuan, line 10: I wonder how bad `quality = 10` is. Do you perhaps have an example? It *might* be so bad that it's worse than not providing it to LLM when using visual browsing.

csmith49: I don't know the performance degradation that comes with the lower quality. It'd be interesting to find out. But the average screenshot is a 720p image between 50kb and 150kb, so with the default settings the expected quality is 100%.

The intent is more safety valve on event size rather than some kind of performance tradeoff.

li-boxuan: I don't think larger screenshots are that rare though? E.g. https://github.com/TheAgentCompany/experiments/blob/main/evaluation/1.0.0/20250510_OpenHands-0.28.1-sonnet-20250219/screenshots/pm-present-engineer-group-members-image/22.png is 912K.

FYI it was saved by this piece of code: https://github.com/All-Hands-AI/OpenHands/blob/27c18f5bdd83cb6d46f5c74fe2c6932336d2f780/evaluation/benchmarks/the_agent_company/run_infer.py#L157-L161

csmith49: 912K is certainly huge (largest I've seen is 320K), but rarity might be subjective here. I went ahead and plotted the distribution of screenshot sizes in that evaluation you linked:

visualization

Definitely a large screenshot mode, but still nothing large enough to trigger the quality reductions. I was going to show how this distribution shifts when you force certain quality values...and didn't see any change at all.

Turns out we use PNG for screenshots, which is lossless, so the `quality=...` parameter is ignored. There _is_ a way to control the size of PNGs by setting `compress_level=...` or `optimize=True`, but I only managed to get a 2% reduction in average image size while taking ~450% as long to save. IMO not worth the tradeoff.

So I'm going to wind back these quality reduction changes. They won't actually do anything right now (we'd have to switch to JPEGs or something) but it's probably best not to insinuate there's compression going on when there isn't.

li-boxuan (approved):

openhands/runtime/browser/base64.py
li-boxuan: Thanks for the great analysis!

PR #4329, opened 2024-10-10
Allow attaching to existing sessions without reinitializing the runtime

rbren:
**End-user friendly description of the problem this fixes or functionality that this introduces**

- [ ] Include this change in the Release Notes. If checked, you must provide an **end-user friendly** description for your change below
no changelog needed

---
**Give a summary of what the PR does, explaining any non-trivial design decisions**

We're working on a highly-available deployment of OpenHands, which means running multiple replicas of the app. Any given API request could connect to _any_ of these replicas, so we have to be careful about managing things in-memory.

Currently, we create a Session when a websocket connects, and stash that session in a dict in the SessionManager--i.e. we keep a reference in memory. When the user calls e.g. `/list-files`, we look up the session and connect to its runtime to list the files.

This doesn't work in an HA, multi-replica world. You might hit an instance that doesn't have your active session, leading to a 404.

In this change, all the REST endpoints create a new Conversation object, which is basically an in-progress Session. And the Conversation looks up an existing Runtime to connect to, rather than initializing a new one.

Pretty sure we need to be smarter about some async/threading when initting the runtime, but this appears to be working well for both local and remote runtimes.

One breaking-ish change: I've had to remove logic related to planning/tasks because that requires accessing an AgentController via REST. TBD on how we do this going forward.

We should really do a heavy refactor here, along the lines of what @tofarr has suggested. But this seems like a reasonable stop gap

---
**Link of any specific issues this addresses**

PR #8564, opened 2025-05-17
Fix - Make Remote runtime use /start for both new and resume

raymyers:

---

Simplify remote runtime by calling the `/start` endpoint on both new runtimes and resumes, since runtime-api can handle both cases this was duplicating the server-side logic.

We've recently seen trouble resuming Remote runtimes in OpenHands Cloud, which may be related to multiple being created in a race condition, so the hope is to improve consistency by letting runtime-api decide which existing runtime to use if any.

An error associated is 
> Error connecting to conversation <ID>: Could not find existing runtime for SID: <ID>"

Logged from here: https://github.com/All-Hands-AI/OpenHands/blob/2693360ad026e6e0f04b4de6d23b026e3cfc0d55/openhands/runtime/impl/remote/remote_runtime.py#L121

This will ignore the `attach_to_existing` option on create (not on close). That was added in #4329 (Oct 2024) and I don't think it currently adds anything useful on RemoteRuntime create, but @tofarr and @rbren may have more context.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:364c571-nikolaik   --name openhands-app-364c571   docker.all-hands.dev/all-hands-ai/openhands:364c571
```


raymyers:
We have merge conflicts now and I think it's best to redo this after Nested Runtimes is in.

PR #8567, opened 2025-05-17
Document native windows support without WSL

li-boxuan:

---

---


neubig (approved):
LGTM, thanks a lot!

mamoodi:
I generally really like this PR and appreciate it because it adds correct information. Thank you!
The part I'm a little hesitant is mixing the gui mode with the local run time on Windows that doesn't run the gui mode.

My suggestion, if you agree is in the CLI and Headless docs, mention that you can run these with local runtime in Windows. Or in the local runtime, have a bit that this can run in Windows without docker desktop but only for CLI and Headless mode and link to them.

If you agree, I can take this PR over and make those changes.

PR #8586, opened 2025-05-19
[Feat]: add context msg to new conversation endpoint

malhotra5:

- Pass additional context to the agent via `new_conversation` endpoint

---

There is a common pattern of starting conversations with context relevant for the agent, but not the user. Examples include

- Resolver containing system instructions
- Slack integration containing previous messages for context
- Browser extension

This PR adds 

1. Adds an optional param `context_msg` to the new conversation endpoint for context that's relevant to the agent
2. This context is not displayed in the FE
3. A new type of memory attribute called `ConversationContext` pulled in along with the runtime/repo info

This could overlap somewhat with `additional_agent_instructions`; however the agent instructions are specific to the runtime, while conversation context is specific to a conversation

For simplicity we could combine them as one instruction, but it more appropriate to separate them for clarity 

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:3ca9e83-nikolaik   --name openhands-app-3ca9e83   docker.all-hands.dev/all-hands-ai/openhands:3ca9e83
```


enyst (commented):
Good to see more info coming in 👍 😄 

I don't fully understand it though. Two questions
- could we add a description in the PR on what this `contextual_info` is supposed to be, and a docstring or something where it's defined, for the source code to have it too for posterity?
- should we add it to other entry points, headless, cli, resolver?

malhotra5:
Yup will do!

malhotra5:
@openhands please see that we have good test converge for `ConversationContext` in `test_memory.py` and `test_prompt_manager.py` 

If not add new tests to increase test coverage

openhands/server/routes/manage_conversations.py
```diff
@@ -61,6 +61,7 @@ class InitSessionRequest(BaseModel):
     image_urls: list[str] | None = None
     replay_json: str | None = None
     suggested_task: SuggestedTask | None = None
+    context_msg: str | None = None
```
malhotra5, line 64: Wonder if we should just call this a `system_msg`?

openhands/core/setup.py
```diff
@@ -145,13 +146,16 @@ def create_memory(
         selected_repository: The repository to clone and start with, if any.
         repo_directory: The repository directory, if any.
         status_callback: Optional callback function to handle status updates.
+        conversation_context: Optional hidden instructions that are passed to the agent
```
enyst, line 149:
```suggestion
        conversation_context: Optional instructions that are passed to the agent
```

openhands/server/routes/manage_conversations.py
enyst: In some sense, all `WORKSPACE_INFO`-type microagent could very well be part of the system message, and in fact the other parts of it are even more stable than this new one. Maybe we shouldn't name them that if they're not that, it may be confused with the real thing.

Related: on the other hand, could we name this new string _the same_ way everywhere? Or as much as we can? Sometimes here we name it `contextual_info`, other times `context_msg`, other times `conversation_context`, but all these names denote the same thing.

I think `conversation_context` is better than the others, idk if ideal 😅

enyst: Or maybe `conversation_instructions` 🤔

tests/unit/test_prompt_manager.py
```diff
@@ -205,7 +213,14 @@ def test_build_workspace_context_with_repo_and_runtime(prompt_dir):
 {% if runtime_info.additional_agent_instructions %}
 {{ runtime_info.additional_agent_instructions }}
 {% endif %}
+
+Today's date is {{ runtime_info.date }}
 </RUNTIME_INFORMATION>
+{% if conversation_context.context %}
+<CONTEXT_MESSAGE>
+{{ conversation_context.context }}
```
enyst, line 221: It seems tests are for the case when it's there. What if it's not there, does the obs work and have/not have its tags? Or if it's there but alone? Or could we test that if it's in init session, then it gets stored into memory?

openhands/utils/prompt.py
```diff
@@ -25,6 +25,17 @@ class RepositoryInfo:
     repo_directory: str | None = None
 
 
+@dataclass
+class ConversationContext:
+    """
+    Instructions that are important for the agent to perform its task, but the user doesn't need to know
```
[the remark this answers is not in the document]
enyst: Maybe we can add to this docstring more specific information about what this could be, or an example? So that the reader can grasp what kind of thing is this for

docs/static/openapi.json
```diff
@@ -876,6 +876,11 @@
                     "type": "string",
                     "nullable": true
                   },
+                  "context_msg": {
+                    "type": "string",
+                    "nullable": true,
+                    "description": "Additional system instructions for the agent task"
```
enyst, line 882:
```suggestion
                    "description": "Context for the agent task. These are optional instructions dependent on the entry point"
```
Maybe... Idk, feel free to not accept it if it doesn't sound right to you.

enyst (commented):
Thank you for this! A bit of standardization across entry points may help.

I left small comments inline, really nits, mostly just about readability of this new attribute and its meaning to a reader, so that people feel confident how to use it.

openhands/utils/prompt.py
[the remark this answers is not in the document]
malhotra5: Yep will do; definitely agree that its not something we want to hide (will still show up in trajectory)

It's more of a UI and organizational preference - will update docstring to reflect this

docs/static/openapi.json
malhotra5: For an end facing user I think the part about entry points may be less obvious

Perhaps 

> Optional context provides additional information to help the agent address the initial user request

openhands/server/routes/manage_conversations.py
malhotra5: I like `conversation_instructions`, will standardize to this terminology

I also like that we can have instructions like the following

```
Make sure to look at this optional context when solving the issue

<CONTEXT MESSAGES>
```

Effectively, a conversation instruction is a superset which can contain context messages

openhands/memory/memory.py
```diff
@@ -306,9 +317,21 @@ def set_runtime_info(
...
+                date=date,
+                custom_secrets_descriptions=custom_secrets_descriptions,
             )
 
+    def set_conversation_instructions(
+        self, conversation_instructions: str | None
+    ) -> None:
+        """
+        Set contextual information for conversation
+        This is information the agent requires but the user doesn't really need to know about
```
enyst, line 329:
```suggestion
        This is information the agent may require
```
😄 a tiny bit

enyst (approved):
🍰

PR #8587, opened 2025-05-19
Feature/login enhance

chuckbutkus:

---

---

### DIFF
--- a/frontend/src/hooks/mutation/use-logout.ts
+++ b/frontend/src/hooks/mutation/use-logout.ts
@@ -3,6 +3,7 @@ import { useNavigate } from "react-router";
 import posthog from "posthog-js";
 import OpenHands from "#/api/open-hands";
 import { useConfig } from "../query/use-config";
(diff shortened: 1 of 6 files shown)


frontend/src/hooks/mutation/use-logout.ts
```diff
@@ -17,8 +18,16 @@ export const useLogout = () => {
...
+      // Clear login method and last page from local storage
+      if (config?.APP_MODE === "saas") {
+        clearLoginData();
+      }
+
       posthog.reset();
       await navigate("/");
+
+      // Refresh the page after all logout logic is completed
+      window.location.reload();
```
amanape, line 30: Why do we reload here?

amanape (approved):

frontend/src/hooks/mutation/use-logout.ts
amanape: If it's to invalidate or remove queries, we can do this from the `queryClient` object directly. LGTM either way but this will cause an unpleasant refresh visual

PR #8589, opened 2025-05-20
feat: Add dev container

kentyman23:
Add dev container

---
Added a barebones dev container to avoid installing dependencies listed in [Development.md](https://github.com/All-Hands-AI/OpenHands/blob/main/Development.md). I confirmed that this image has everything we need to `make build && make run`.

While I was at it, I told Git to force *nix line endings (LF), even when cloned from Windows (normally CRLF). This way, when you do a `git status` in the Linux image while looking at a repo cloned from Windows, you won't have the problem of _every_ file looking "modified" just because of line endings.

I also added configuration so the editor would enforce these line endings _and_ so we wouldn't have to wait until `pre-commit` to catch whitespace issues.

This is also being discussed in [this Slack thread](https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1747680397285089).

---

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    // force *nix line endings so files don't look modified in container run from Windows clone
+    "files.eol": "\n",
+    "files.trimTrailingWhitespace": true,
+    "files.insertFinalNewline": true,
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,15 @@
 // For format details, see: https://aka.ms/devcontainer.json
-// For config options, see: https://github.com/devcontainers/templates/tree/main/src/python
 {
 	"name": "Python 3",
+	// Documentation for this image:
+	// - https://github.com/devcontainers/templates/tree/main/src/python
+	// - https://github.com/microsoft/vscode-remote-try-python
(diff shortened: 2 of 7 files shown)


.vscode/settings.json
kentyman23, line 1: This is arguably redundant with `.editorconfig`. I'm torn on either A) only doing `.editorconfig` since it's universal, or B) keeping in both places so that people familiar with `.vscode/settings.json` can clearly see the rules we follow. Regardless, I think keeping the changes to `.gitignore` make sense to keep for if/when we want repo-wide settings.

neubig (commented):
Hey @kentyman23 , thanks so much for this! I think that this is good, but after thinking about it a bit more, I think that we should probably revert the changes for `.gitignore` and `.vscode/settings.json` for now. There is a potential downside here that some people may prefer other vscode settings, and then any time they update them they might accidentally commit. I think this idea of adding vscode settings is worth possibly considering, but maybe we could keep that for a slightly later date to get the devcontainer stuff merged.

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    // force *nix line endings so files don't look modified in container run from Windows clone
```
rbren, line 2:
```suggestion
```
are comments allowed here? they usually break JSON

neubig:
Looks good to me too! Merging

PR #8597, opened 2025-05-20
Fix for issue where initial env vars are not passed to runtime

tofarr:

The `initial_env_vars` include the `config.sandbox.runtime_startup_env_vars`, but may include additional  / overridden values. These should be passed in at startup.

---
So currently we set up the remote runtime with incomplete / wrong initial variables, and then later fix it with a call to `setup_initial_env` later. This makes the environment variables correct from the beginning - so we could probably remove that later call.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:57a1b90-nikolaik   --name openhands-app-57a1b90   docker.all-hands.dev/all-hands-ai/openhands:57a1b90
```

### DIFF
--- a/openhands/runtime/impl/remote/remote_runtime.py
+++ b/openhands/runtime/impl/remote/remote_runtime.py
@@ -237,7 +237,7 @@ class RemoteRuntime(ActionExecutionClient):
         environment: dict[str, str] = {}
         if self.config.debug or os.environ.get('DEBUG', 'false').lower() == 'true':
             environment['DEBUG'] = 'true'
-        environment.update(self.config.sandbox.runtime_startup_env_vars)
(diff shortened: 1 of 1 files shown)


malhotra5 (approved):
NIT: let's also remove the latter call using `setup_initial_env`

It seems like the other runtimes besides `remote_runtime` also follow the same pattern of using `setup_initial_env`, maybe it makes sense to remove the `environment.update` line and just rely on `setup_initial_env`

rbren (approved):

PR #8598, opened 2025-05-20
[WIP] OH Versa

xingyaoww:

---

---


openhands/agenthub/codeact_agent/tools/search_engine.py
```diff
@@ -0,0 +1,29 @@
+from litellm import ChatCompletionToolParam, ChatCompletionToolParamFunctionChunk
+
+_SEARCH_ENGINE_DESCRIPTION = """Execute a web search query (similar to Google search).
+
+You MUST use this tool as a search engine and find URLs of webpages that are relevant to the search query.
+NOTE: Do NOT use the browser tool to search using Google or Bing since you will be blocked by CAPTCHAs.
```
enyst, line 6:
```suggestion
NOTE: Do NOT use the search engine tool to search using Google or Bing since you will be blocked by CAPTCHAs.
```

xingyaoww:
Blocked by: https://github.com/All-Hands-AI/openhands-aci/pull/126

openhands/runtime/browser/browser_env.py
```diff
@@ -151,7 +155,15 @@ def browser_process(self) -> None:
                         continue
 
                     action = action_data['action']
-                    obs, reward, terminated, truncated, info = env.step(action)
+                    try:
+                        obs, reward, terminated, truncated, info = env.step(action)
+                    except Exception as e:
```
ryanhoangt, line 160: Does this error handling code serve any special purpose? I think BrowserGym's `step` implementation already has it own error handling: https://github.com/ServiceNow/BrowserGym/blob/6bc4c5156d32d7c6598ea376badc5ff79030fe5b/browsergym/core/src/browsergym/core/env.py#L458-L462

xingyaoww: @adityasoni9998

adityasoni9998: @ryanhoangt @xingyaoww 

This addition is intentional and not just for extra safety. Firstly, there are some cases in browsergym 0.10.2 wherein the step function raises an error (I guess these don't happen for version 0.13.3) especially when there is an error during processing of AXTree (see [here](https://github.com/ServiceNow/BrowserGym/blob/a9e44a88139798543ba53fc8c45d44997665ccca/browsergym/core/src/browsergym/core/env.py#L540) - an error saying ``Cannot mark child frame without bid``)

Also it is better to have this line in the code - the `browser_process` function runs inside a separate process, and if it crashes/exits due to some reason, the main OpenHands process running the step function (the client side) will always face timeouts, and all subsequent browsing actions will fail since the server side (browser_process) has exited silently.

xingyaoww:
These are already merged in separate PRs. Close this.

PR #8602, opened 2025-05-21
Add Export Trajectory Button to Conversation Card

raymyers:

Adds an "Export Trajectory" button to the conversation card context menu, allowing users to easily export conversation trajectory data as a JSON file.

---

- Added an "Export Trajectory" button to the conversation card context menu
- Implemented the export functionality using the existing `useGetTrajectory` hook and `downloadTrajectory` utility
- Added error handling with toast notifications
- Added analytics tracking for the export action
- Added comprehensive tests for the new functionality

---

N/A

### DIFF
--- a/frontend/__tests__/components/features/conversation-panel/conversation-card.test.tsx
+++ b/frontend/__tests__/components/features/conversation-panel/conversation-card.test.tsx
@@ -14,6 +14,8 @@ import { renderWithProviders } from "test-utils";
 import { formatTimeDelta } from "#/utils/format-time-delta";
 import { ConversationCard } from "#/components/features/conversation-panel/conversation-card";
 import { clickOnEditButton } from "./utils";
+import { useGetTrajectory } from "#/hooks/mutation/use-get-trajectory";
+import { downloadTrajectory } from "#/utils/download-trajectory";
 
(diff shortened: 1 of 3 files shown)


enyst:
Just a quick side note

> We do not show the conversation (including the export trajectory button) until the runtime starts. In the case that it doesn't start, people may not be able to retrieve their data - especially on cloud.

I am surprised, I could swear the conversation window is / was visible. Not sure when or why that changed.
Wait, it didn't
image

But the Export Conversation button doesn't appear to work, I click and nothing happens.

enyst:
Ah, maybe what happens may be twofold:
- either those conversations don't manage to load in the UI - that's before the runtime is starting, I think, it's loading events [here](https://github.com/All-Hands-AI/OpenHands/blob/3280f450acdbdbb496e736f65dea59ff34408ca4/openhands/server/listen_socket.py#L119) - in which case I assume you may want to look at e.g. the size of those histories
- or the button exists, but pressing it does nothing, until the runtime loads, which I never noticed before, maybe we could look into why?

Of course, the above is assuming the saas hasn't changed this flow.

amanape (approved):

PR #8610, opened 2025-05-21
Fix unsupported MCP tools param

### LINKED ISSUE #8595: [Bug]: mcp tool fetch is not compatible with gemini api

oh add mcp tool fetch as default tool, it will 

```
  File "/Users/Library/Caches/pypoetry/virtualenvs/openhands-ai-JejVGZol-py3.12/lib/python3.12/site-packages/litellm/llms/vertex_ai/gemini/vertex_and_google_ai_studio_gemini.py", line 1541, in completion
    raise VertexAIError(
litellm.llms.vertex_ai.common_utils.VertexAIError: {
  "error": {
    "code": 400,
    "message": "* GenerateContentRequest.tools[0].function_declarations[7].parameters.properties[url].format: only 'enum' and 'date-time' are supported for STRING type\n",
    "status": "INVALID_ARGUMENT"
  }
}
```
because  following parameter `fetch['parameters']['properties']['url']['format']` from fetch is not compatible with gemini api:
```
 {'name': 'fetch', 'description': '.......', 'parameters': {'description': '......', 'properties': {'url': {'description': 'URL to fetch', 'format': 'uri', 'minLength': 1, 'title': 'Url', 'type': 'string'}, 'max_length': {'default': 5000, 'description': 'Maximum number of characters to return.', 'title': 'Max Length', 'type': 'integer'}, 'start_index': {'default': 0, 'description': '......', 'minimum': 0, 'title': 'Start Index', 'type': 'integer'}, 'raw': {'default': False, 'description': '......', 'title': 'Raw', 'type': 'boolean'}}, 'required': ['url'], 'title': 'Fetch', 'type': 'object'}}]}], 'litellm_param_is_function_call': False}
```

I workround with this error by remove  'format'  from fetch in `openhands/mcp/utils.py` : 
```
                mcp_tools = tool.to_param()
                 if mcp_tools['function']['name'] == 'fetch':
                    # remove url format attribute
                     mcp_tools['function']['parameters']['properties']['url'].pop('format', None)
```
This is just a temporary solution

### OpenHands Installation

Docker command in README

### OpenHands Version

main

### Operating System

MacOS

Image

enyst:

---

This PR proposes a fix unsupported parameter `format` property.

I see this on Gemini latest versions, and it's reported in issues too. I think actually it might not be only Gemini. This is a quick fix, I made a small refactoring of the method out of agent's `step`, maybe we should move this in tools serialization code in a follow up... 🤔 or elsewhere, that code doesn't know the LLM and maybe it's for the best.

I haven't yet verified if it was reported to litellm, will do.

---
Fix #8595 

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:a0d7f2c-nikolaik   --name openhands-app-a0d7f2c   docker.all-hands.dev/all-hands-ai/openhands:a0d7f2c
```

### DIFF
--- a/openhands/agenthub/codeact_agent/codeact_agent.py
+++ b/openhands/agenthub/codeact_agent/codeact_agent.py
@@ -10,7 +10,7 @@ if TYPE_CHECKING:
     from openhands.events.action import Action
     from openhands.llm.llm import ModelResponse
 
-from openhands.llm.llm_utils import check_tools_for_llm_compatibility
+from openhands.llm.llm_utils import check_tools
 import openhands.agenthub.codeact_agent.function_calling as codeact_function_calling
 from openhands.agenthub.codeact_agent.tools.bash import create_cmd_run_tool
 from openhands.agenthub.codeact_agent.tools.browser import BrowserTool
@@ -186,7 +186,7 @@ class CodeActAgent(Agent):
         params: dict = {
             'messages': self.llm.format_messages_for_llm(messages),
         }
-        params['tools'] = check_tools_for_llm_compatibility(self.tools, self.llm.config)
+        params['tools'] = check_tools(self.tools, self.llm.config)
         params['extra_body'] = {'metadata': state.to_llm_metadata(agent_name=self.name)}
         response = self.llm.completion(**params)
         logger.debug(f'Response from LLM: {response}')
@@ -264,7 +264,7 @@ class CodeActAgent(Agent):
             self.conversation_memory.apply_prompt_caching(messages)
 
         return messages
-    
+
     def response_to_actions(self, response: 'ModelResponse') -> list['Action']:
         return codeact_function_calling.response_to_actions(
             response, mcp_tool_names=list(self.mcp_tools.keys())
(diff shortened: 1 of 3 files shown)


openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -282,6 +263,39 @@ def _get_messages(
             self.conversation_memory.apply_prompt_caching(messages)
 
         return messages
+    
+    def _check_tools(self, tools: list['ChatCompletionToolParam']) -> list['ChatCompletionToolParam']:
```
xingyaoww, line 267: IMO we could probably move this function out of `codeact_agent.py` to prevent making this file more complex :( - Maybe somewhere like `openhands/llm/tool_utils.py` and import it here?

xingyaoww (approved):

rbren (approved):

openhands/agenthub/codeact_agent/codeact_agent.py
enyst: Done! Yes, these checks really don't belong in the `step` and in fact, in the agent.

PR #8613, opened 2025-05-21
Add conversation_id and user_id to LLM constructor

rbren:
This PR adds `conversation_id` and `user_id` as required parameters to the LLM constructor. These parameters are used to track which conversation and user are making LLM requests.

### Changes
- Modified the LLM constructor to require conversation_id and user_id parameters
- Updated all places where LLM is instantiated to pass these parameters
- Updated tests to include the new required parameters
- Fixed mypy errors related to the changes

### Testing
- All unit tests pass

### DIFF
--- a/openhands/memory/condenser/impl/llm_attention_condenser.py
+++ b/openhands/memory/condenser/impl/llm_attention_condenser.py
@@ -121,7 +121,11 @@ class LLMAttentionCondenser(RollingCondenser):
         llm_config.caching_prompt = False
 
         return LLMAttentionCondenser(
-            llm=LLM(config=llm_config),
+            llm=LLM(
+                config=llm_config,
+                conversation_id="attention_condenser",
+                user_id="system"
+            ),
             max_size=config.max_size,
             keep_first=config.keep_first,
         )
--- a/openhands/memory/condenser/impl/llm_summarizing_condenser.py
+++ b/openhands/memory/condenser/impl/llm_summarizing_condenser.py
@@ -163,7 +163,11 @@ CURRENT_STATE: Last flip: Heads, Haiku count: 15/20"""
         llm_config.caching_prompt = False
 
         return LLMSummarizingCondenser(
-            llm=LLM(config=llm_config),
+            llm=LLM(
+                config=llm_config,
+                conversation_id="condenser",
+                user_id="system"
+            ),
             max_size=config.max_size,
             keep_first=config.keep_first,
             max_event_length=config.max_event_length,
--- a/openhands/memory/condenser/impl/structured_summary_condenser.py
+++ b/openhands/memory/condenser/impl/structured_summary_condenser.py
@@ -318,7 +318,11 @@ Capture all relevant information, especially:
         llm_config.caching_prompt = False
 
         return StructuredSummaryCondenser(
-            llm=LLM(config=llm_config),
+            llm=LLM(
+                config=llm_config,
+                conversation_id="structured_summary_condenser",
+                user_id="system"
+            ),
             max_size=config.max_size,
             keep_first=config.keep_first,
             max_event_length=config.max_event_length,
--- a/openhands/runtime/utils/edit.py
+++ b/openhands/runtime/utils/edit.py
@@ -121,7 +121,12 @@ class FileEditRuntimeMixin(FileEditRuntimeInterface):
             )
             draft_editor_config.caching_prompt = False
 
-        self.draft_editor_llm = LLM(draft_editor_config, metrics=llm_metrics)
+        self.draft_editor_llm = LLM(
+            config=draft_editor_config,
+            conversation_id="draft_editor",
+            user_id="system",
+            metrics=llm_metrics
+        )
         logger.debug(
             f'[Draft edit functionality] enabled with LLM: {self.draft_editor_llm}'
         )
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -678,7 +678,7 @@ class AgentController:
         agent_cls: type[Agent] = Agent.get_cls(action.agent)
         agent_config = self.agent_configs.get(action.agent, self.agent.config)
         llm_config = self.agent_to_llm_config.get(action.agent, self.agent.llm.config)
-        llm = LLM(config=llm_config, retry_listener=self._notify_on_llm_retry)
+        llm = LLM(config=llm_config, conversation_id=self.id, user_id=self.id.split('-')[0], retry_listener=self._notify_on_llm_retry)
         delegate_agent = agent_cls(llm=llm, config=agent_config)
         state = State(
             session_id=self.id.removesuffix('-delegate'),
--- a/openhands/core/setup.py
+++ b/openhands/core/setup.py
@@ -172,13 +172,13 @@ def create_memory(
     return memory
 
 
-def create_agent(config: AppConfig) -> Agent:
+def create_agent(config: AppConfig, conversation_id: str = "default", user_id: str = "default") -> Agent:
     agent_cls: type[Agent] = Agent.get_cls(config.default_agent)
     agent_config = config.get_agent_config(config.default_agent)
     llm_config = config.get_llm_config_from_agent(config.default_agent)
 
     agent = agent_cls(
-        llm=LLM(config=llm_config),
+        llm=LLM(config=llm_config, conversation_id=conversation_id, user_id=user_id),
         config=agent_config,
     )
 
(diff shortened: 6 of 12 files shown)


openhands/memory/condenser/impl/llm_attention_condenser.py
```diff
@@ -121,7 +121,11 @@ def from_config(cls, config: LLMAttentionCondenserConfig) -> LLMAttentionCondens
         llm_config.caching_prompt = False
 
         return LLMAttentionCondenser(
-            llm=LLM(config=llm_config),
+            llm=LLM(
```
rbren, line 124: @openhands refactor the condenser and its implementations to take in a required LLM object in the constructor, rather than letting it construct an internal LLM

openhands/memory/condenser/impl/llm_attention_condenser.py
```diff
@@ -113,15 +113,25 @@ def should_condense(self, view: View) -> bool:
...
 
     @classmethod
-    def from_config(cls, config: LLMAttentionCondenserConfig) -> LLMAttentionCondenser:
-        # This condenser cannot take advantage of prompt caching. If it happens
-        # to be set, we'll pay for the cache writes but never get a chance to
-        # save on a read.
-        llm_config = config.llm_config.model_copy()
-        llm_config.caching_prompt = False
+    def from_config(
+        cls, config: LLMAttentionCondenserConfig, llm: LLM | None = None
```
rbren, line 117: @openhands make `llm` a required param here. Add a `.clone` method to the LLM object so that we can clone it and set `caching_prompt` to False in the clone here.

openhands/memory/condenser/impl/llm_attention_condenser.py
```diff
@@ -113,15 +115,13 @@ def should_condense(self, view: View) -> bool:
...
     @classmethod
-    def from_config(cls, config: LLMAttentionCondenserConfig) -> LLMAttentionCondenser:
-        # This condenser cannot take advantage of prompt caching. If it happens
-        # to be set, we'll pay for the cache writes but never get a chance to
-        # save on a read.
-        llm_config = config.llm_config.model_copy()
-        llm_config.caching_prompt = False
-
+    def from_config(
+        cls, config: LLMAttentionCondenserConfig, llm: Any = None
```
rbren, line 119: @openhands fix the type on this--it should be LLM, not Any

And remove the `=None` as the default on all `from_config` methods in this directory. Make the caller pass in an LLM.

openhands/memory/condenser/impl/llm_attention_condenser.py
```diff
@@ -113,15 +113,11 @@ def should_condense(self, view: View) -> bool:
...
-        # save on a read.
-        llm_config = config.llm_config.model_copy()
-        llm_config.caching_prompt = False
-
+    def from_config(
+        cls, config: LLMAttentionCondenserConfig, llm: LLM
+    ) -> 'LLMAttentionCondenser':
         return LLMAttentionCondenser(
-            llm=LLM(config=llm_config),
+            llm=llm,
```
rbren, line 120: @openhands look at the diff of this file vs main. We need to use llm.clone() to copy the LLM here, passing in a new config which sets caching_prompt to false

openhands/runtime/utils/edit.py
```diff
@@ -121,7 +121,12 @@ def __init__(self, enable_llm_editor: bool, *args: Any, **kwargs: Any) -> None:
             )
             draft_editor_config.caching_prompt = False
 
-        self.draft_editor_llm = LLM(draft_editor_config, metrics=llm_metrics)
+        self.draft_editor_llm = LLM(
```
rbren, line 124: @openhands modify this so that `draft_editor_llm` is passed into the __init__ method here instead of `draft_editor_config`. Modify any callers to use the new constructor

rbren (commented):
@openhands there's a to_llm_metadata method in openhands/controller/state/state.py. Its output is passed to `LLM.completion`. Instead, this method should live directly on the LLM class (openhands/llm/llm.py) and get passed to every completion call (we have a wrapper which can be used to modify the params). Please also set trace_user_id using the new user_id field inside the LLM

openhands/memory/condenser/impl/llm_attention_condenser.py
```diff
@@ -113,15 +113,17 @@ def should_condense(self, view: View) -> bool:
...
     @classmethod
-    def from_config(cls, config: LLMAttentionCondenserConfig) -> LLMAttentionCondenser:
+    def from_config(
+        cls, config: LLMAttentionCondenserConfig, llm: LLM
+    ) -> 'LLMAttentionCondenser':
         # This condenser cannot take advantage of prompt caching. If it happens
         # to be set, we'll pay for the cache writes but never get a chance to
         # save on a read.
-        llm_config = config.llm_config.model_copy()
+        llm_config = llm.config.model_copy()
```
rbren, line 122:
```suggestion
        llm_config = config.llm_config.model_copy()
```

openhands/memory/condenser/impl/llm_summarizing_condenser.py
```diff
@@ -154,16 +154,16 @@ def should_condense(self, view: View) -> bool:
...
     def from_config(
-        cls, config: LLMSummarizingCondenserConfig
-    ) -> LLMSummarizingCondenser:
+        cls, config: LLMSummarizingCondenserConfig, llm: LLM
+    ) -> 'LLMSummarizingCondenser':
         # This condenser cannot take advantage of prompt caching. If it happens
         # to be set, we'll pay for the cache writes but never get a chance to
         # save on a read.
-        llm_config = config.llm_config.model_copy()
+        llm_config = llm.config.model_copy()
```
rbren, line 162:
```suggestion
        llm_config = config.llm_config.model_copy()
```

openhands/memory/condenser/impl/structured_summary_condenser.py
```diff
@@ -309,16 +309,16 @@ def should_condense(self, view: View) -> bool:
...
     def from_config(
-        cls, config: StructuredSummaryCondenserConfig
-    ) -> StructuredSummaryCondenser:
+        cls, config: StructuredSummaryCondenserConfig, llm: LLM
+    ) -> 'StructuredSummaryCondenser':
         # This condenser cannot take advantage of prompt caching. If it happens
         # to be set, we'll pay for the cache writes but never get a chance to
         # save on a read.
-        llm_config = config.llm_config.model_copy()
+        llm_config = llm.config.model_copy()
```
rbren, line 317:
```suggestion
        llm_config = config.llm_config.model_copy()
```

openhands/runtime/base.py
```diff
@@ -158,8 +160,33 @@ def __init__(
...
+                model_name='draft_editor:' + draft_editor_config.model
+            )
+            if draft_editor_config.caching_prompt:
+                logger.debug(
+                    'It is not recommended to cache draft editor LLM prompts as it may incur high costs for the same prompt. '
+                    'Automatically setting caching_prompt=false.'
+                )
+                draft_editor_config.caching_prompt = False
+
+            draft_editor_llm = LLM(
```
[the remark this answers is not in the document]
enyst: Do we really need this, considering the draft LLM is not enabled by default, and in fact fairly rarely used? @rbren 

Or maybe I misunderstand the problem this is solving, is it just to enforce a `conversation_id` and `user_id` in any llm? This one won't have `user_id` anyway.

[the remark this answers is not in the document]
enyst: On the bright side, enforcing the same LLM here as the one that was already built means they will share state, so we might be able to do easier some statistics.

On the other hand, enforcing the same LLM probably makes the draft feature almost useless? The point was to apply diffs with another LLM, as far as I recall, although it can use the same if the user doesn't configure another.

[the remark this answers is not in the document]
rbren: Yeah the latter is what I'm after--we are not currently tracking user spend per conversation properly. I'm trying to make it so that every LLM() is associated with the correct conversation

Should I maybe just remove the draft_editor entirely? do we need it?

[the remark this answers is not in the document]
enyst: It's not in use with Sonnet, and it's not in use by default with any model. It's an experimental attempt to apply diffs with an LLM that needs enabled in config.toml to be used. I'd vote to keep it, specially since it's not the reason for the problems I don't think, but then I'm very conservative on this stuff. 🤷 

It didn't work well compared to the current single LLM editing, and I understand Xingyao has not been using it for Qwen3 eval. On the other hand, it was merged in main literally 24 hours before Anthropic's 3.6 with computer use and all that, last year, which took everyone's attention...

@xingyaoww WDYT?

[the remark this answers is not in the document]
enyst: This is how we enable it, it's an `[agent]` option
https://github.com/All-Hands-AI/OpenHands/blob/5b021ad1bb5af743b1e1159c78dbd537961b872d/config.template.toml#L222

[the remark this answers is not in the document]
enyst: Actually, if Xingyao doesn't need it, let's remove it. It has lead to more complexity with all this metrics tracking for sure. Removing that complexity might indeed help.

rbren:
@openhands fix the python unit tests, then fix the python lint issues

jacksoncartelez (approved):

PR #8620, opened 2025-05-21
fix(mcp): fix the config conflict between microagent MCP tools & MCP config

xingyaoww:
# Fix MCP Configuration Conflict Between Microagent Tools & Runtime MCP Config

This PR fixes an issue where tools added by microagents (like the 'fetch' tool from default-tools microagent) would become unavailable during runtime. Now, when you add custom tools via microagents, those tools will remain available throughout the agent's lifecycle.

---

## Problem Context

The OpenHands system uses the Model Context Protocol (MCP) to provide tools to agents. There are two key components involved:

1. **Microagent MCP Tools**: These are tools defined in microagents (like the 'fetch' tool in default-tools microagent) that are added to the agent during initialization.

2. **Runtime MCP Configuration**: This is the configuration used when actually executing tool calls.

The issue occurs because of how the MCP configuration is managed:

1. When a microagent adds tools (via `add_mcp_tools_to_agent`), it calls `get_updated_mcp_config` with the microagent's tools as `extra_stdio_servers`.

2. Later, when a tool is called (via `call_tool_mcp`), it calls `get_updated_mcp_config` again but without any parameters.

3. The original implementation would send a new configuration to the server each time, effectively overwriting the previous configuration. This meant that tools added by microagents would be lost when making actual tool calls.

## Solution

This PR fixes the issue by:

1. Adding state tracking for MCP stdio servers with a new `_last_updated_mcp_stdio_servers` instance variable.

2. Rewriting the `get_updated_mcp_config` method to:
   - Only update the MCP server when there are new stdio servers to add
   - Maintain a consistent view of all available stdio servers
   - Preserve tools added by microagents for subsequent tool calls

3. Adding a test case that verifies this behavior by:
   - Setting up a runtime with a basic MCP configuration
   - Adding a microagent tool ('fetch')
   - Verifying that both the original tools and the microagent tool work correctly

This ensures that when users add custom tools via microagents, those tools remain available throughout the agent's lifecycle.

---

N/A

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:36c0616-nikolaik   --name openhands-app-36c0616   docker.all-hands.dev/all-hands-ai/openhands:36c0616
```


xingyaoww:
@OpenHands

There was an issue:
1. After i add a stdio server A in MCPConfig, both the MCP server `fetch` from default-tools microagent (microagents/default-tools.md) and the newly added `A` will trigger
2. The tricky thing is `self.get_updated_mcp_config` - it was called in two places, one inside `add_mcp_tools_to_agent` will feed in an additioanal `extra_stdio_server` (used for microagent MCP); the other place is in `async def call_tool_mcp(self, action: MCPAction) -> Observation` inside `action_execution_client.py`, where actual tool calls happend.
3. What happened was that when `call_tool_mcp` is called, `get_updated_mcp_config` is called again, as well as, `/update_mcp_server` endpoint inside it. BUT this time, the stdio server `fetch` added by microagent is not included! If you prompt the model to use that  `fetch` tool again, it will fail.
4. This PR basically rewrite the logic of `get_updated_mcp_config`, so we only update stdio tools when there's a diff and we also maintain an internal `self._last_updated_mcp_stdio_servers`

OpenHands, can you help me read the diff from this PR, understand the changes throughly and help me update the PR description of this PR for people to understand both the context and the solution here?

malhotra5 (approved):
🍰

NIT: instead of serializing the pydantic configs here, maybe we could override the `__eq__` operator in MCPStdioServerConfig instead

That way we could do `if server1 in last_updated_servers`, would be less efficient but simpler to follow, your call though!

xingyaoww:
@OpenHands check @malhotra5's latest comments and help implement the refactor he suggested

openhands/runtime/impl/action_execution/action_execution_client.py
```diff
@@ -358,50 +361,55 @@ def get_updated_mcp_config(
...
+            'debug',
+            f'adding {len(new_servers)} new stdio servers to MCP config: {new_servers}',
+        )
+
         # Only send update request if there are new servers
         if new_servers:
             # Use all current servers for the update (simpler and more reliable)
             # This works because we're replacing the complete configuration
-            stdio_tools = [server.model_dump(mode='json') for server in current_stdio_servers]
+            stdio_tools = [
```
xingyaoww, line 388: @OpenHands the `stdio_tools` we send should be a union of `self._last_updated_mcp_stdio_servers` and `current_stdio_servers`

openhands/core/config/mcp_config.py
```diff
@@ -32,6 +33,21 @@ class MCPStdioServerConfig(BaseModel):
...
+        """
+        if not isinstance(other, MCPStdioServerConfig):
+            return False
+        return (
+            self.name == other.name
+            and self.command == other.command
+            and self.args == other.args
+            and self.env == other.env
+        )
+
```
malhotra5, line 50: Would this work if two server have the same args but in different orders?

malhotra5: Maybe we could also write a small test to confirm the expected equality behavior

xingyaoww: @OpenHands please convert them to set before comparison. Then add a unit test in tests/unit/test_mcp_config.py to validate that

malhotra5 (approved):
Awesome LGTM!

PR #8348, opened 2025-05-08
[Feat]: Git mcp server to open PRs

malhotra5:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [x] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

- OpenHands will use in house MCP to open PRs/MRs in for Github and Gitlab

---
**Summarize what the PR does, explaining any non-trivial design decisions.**
This PR establishes an MCP server within the OpenHands ecosystem that is mounted by default. 

The MCP server will have a tool to open a PR on either Github or Gitlab. 

We mount a FastMCP server onto the existing FastAPI app. FastAPI passes the request context to FastMCP, ensuring we can identify the user and their credentials before executing a tool call.

---
**Link of any specific issues this addresses:**

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:34b04e9-nikolaik   --name openhands-app-34b04e9   docker.all-hands.dev/all-hands-ai/openhands:34b04e9
```

PR #8622, opened 2025-05-21
[Fix]: make mcp config optional in settings

malhotra5:

---

If a user doesn't have `MCPConfig` in their settings, loading it causes as error. The `Settings` basemodel  always expects MCPConfig to exist in user settings (the change was made in #8348)

We fix this by making the field optional again

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:e4b6e05-nikolaik   --name openhands-app-e4b6e05   docker.all-hands.dev/all-hands-ai/openhands:e4b6e05
```

### DIFF
--- a/openhands/server/session/session.py
+++ b/openhands/server/session/session.py
@@ -116,7 +116,7 @@ class Session:
             or settings.sandbox_runtime_container_image
             else self.config.sandbox.runtime_container_image
         )
-        self.config.mcp = settings.mcp_config if settings.mcp_config else MCPConfig(sse_servers=[], stdio_servers=[])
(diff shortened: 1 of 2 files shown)


xingyaoww (approved):

enyst (approved):
Thank you! I think we're doing the right thing setting it optional, but I wonder if we disable implicitly the "fetch" tool in this case, or if it fails. I can't test right now

PR #8641, opened 2025-05-22
[refactor]: Move setup_sandbox_config function

kotauchisunsun:
…r and move sandbox setup to a dedicated function

---

To decouple the command-line parsing logic from the processing logic, I’ve refactored `setup_sandbox_config` from a class method into a standalone function.

To keep the diff manageable, I’m temporarily passing both `args` and `sandbox_config` to the `IssueResolver` constructor. In a future update, the constructor will no longer depend on `args`, and the necessary values will be passed as individual variables instead.

---

### DIFF
--- a/openhands/resolver/issue_resolver.py
+++ b/openhands/resolver/issue_resolver.py
@@ -14,7 +14,6 @@ from uuid import uuid4
 from pydantic import SecretStr
 from termcolor import colored
 
-import openhands
 from openhands.controller.state.state import State
 from openhands.core.config import AgentConfig, AppConfig, LLMConfig, SandboxConfig
 from openhands.core.logger import openhands_logger as logger
@@ -37,7 +36,6 @@ from openhands.resolver.issue_handler_factory import IssueHandlerFactory
 from openhands.resolver.resolver_output import ResolverOutput
 from openhands.resolver.utils import (
     codeact_user_response,
-    get_unique_uid,
     identify_token,
     reset_logger_for_multiprocessing,
 )
@@ -51,7 +49,7 @@ AGENT_CLASS = 'CodeActAgent'
 class IssueResolver:
     GITLAB_CI = os.getenv('GITLAB_CI') == 'true'
 
-    def __init__(self, args: Namespace) -> None:
+    def __init__(self, args: Namespace, sandbox_config: SandboxConfig) -> None:
         """Initialize the IssueResolver with the given parameters.
         Params initialized:
             owner: Owner of the repo.
@@ -59,7 +57,6 @@ class IssueResolver:
             token: Token to access the repository.
             username: Username to access the repository.
             platform: Platform of the repository.
-            runtime_container_image: Container image to use.
             max_iterations: Maximum number of iterations to run.
             output_dir: Output directory to write the results.
             llm_config: Configuration for the language model.
@@ -70,13 +67,6 @@ class IssueResolver:
             comment_id: Optional ID of a specific comment to focus on.
             base_domain: The base domain for the git server.
         """
-
-        # Setup and validate container images
-        self.sandbox_config = self._setup_sandbox_config(
-            args.base_container_image,
-            args.runtime_container_image,
-            args.is_experimental,
-        )
 
         parts = args.selected_repo.rsplit('/', 1)
         if len(parts) < 2:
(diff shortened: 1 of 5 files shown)


enyst:
> To keep the diff manageable, I’m temporarily passing both `args` and `sandbox_config` to the `IssueResolver` constructor. In a future update, the constructor will no longer depend on `args`, and the necessary values will be passed as individual variables instead.

Hey @kotauchisunsun ! Just to understand where we're heading: IMHO in the future we should not construct `sandbox_config` in the resolver at all. The resolver, just like other entry points in the application, should use `load_app_config` to get the current values, then override them with its `args`.

At least, that's my thought about the way forward here. What do you think of this goal? Do you think this refactoring is heading there or do you have a different take?

kotauchisunsun:
> In effect we were treating them as immutable -ish, even though they weren't. I'm pretty sure that's long gone though, now we are doing it.

Yeah, that often happens as development progresses. Software development is hard. 😅

> Or maybe we can do 1), as long as we're doing it all in a single place. So that we can refactor it later easily and move to 2).

I also think option 1 is fine, at least for now. As long as we avoid scattering the config initialization logic, I believe it's acceptable.

> So maybe we could do load_app_config() + resolver-specific override as early as the constructor?

You're right that both issue and issue_handler can be injected at construction time. However, there’s a part of the logic in resolve_issue that uses issue_handler to build the issue:

https://github.com/All-Hands-AI/OpenHands/blob/4b6f2aeb4dfc7084fb9c5344ec5d4a77d348f700/openhands/resolver/issue_resolver.py#L506-L518

If we decide to inject AppConfig into the constructor, this part would also need to be moved into resolve_issue.py.
And if we go that route, we would likely also need to move the logic for parsing self.owner and self.repo into resolve_issue.py:

https://github.com/All-Hands-AI/OpenHands/blob/4b6f2aeb4dfc7084fb9c5344ec5d4a77d348f700/openhands/resolver/issue_resolver.py#L81-L84

Personally, I like this kind of change. But I also realize that since it touches a lot of areas, others might feel differently or have concerns about the size of the change.

There's also one more point I forgot to mention — the dependency on the workspace path.
Within IssueResolver’s runtime, it seems like operations are being performed under self.output_dir / workspace.
If we inject AppConfig into the IssueResolver constructor, the fact that the runtime should be initialized under the workspace subpath is no longer visible within the IssueResolver code itself.
However, functions like copy_code, initialize_runtime, and complete_runtime all assume the presence of a workspace path:

https://github.com/All-Hands-AI/OpenHands/blob/4b6f2aeb4dfc7084fb9c5344ec5d4a77d348f700/openhands/resolver/issue_resolver.py#L378

https://github.com/All-Hands-AI/OpenHands/blob/4b6f2aeb4dfc7084fb9c5344ec5d4a77d348f700/openhands/resolver/issue_resolver.py#L248

https://github.com/All-Hands-AI/OpenHands/blob/4b6f2aeb4dfc7084fb9c5344ec5d4a77d348f700/openhands/resolver/issue_resolver.py#L292

So, if we were to move the workspace_base logic into resolve_issue.py, that would effectively split the definition of the workspace path across two places — resolve_issue.py and IssueResolver — which could lead to future confusion. For example, if someone changes the workspace path in resolve_issue but not in IssueResolver, or vice versa, it might cause subtle bugs.

That’s why I was thinking that if we go down that route, we should also refactor things to unify the handling of the workspace path, potentially by moving it to a shared class-level variable. But of course, that would expand the scope of the change quite a bit. 😅

enyst:
Maybe we could go the other way around than this PR is doing: instead of taking `args` away from init, we could bring them all in init, and then try to call `load_app_config()`, and see what becomes redundant: for example, from line 101 to 125 ([here](https://github.com/All-Hands-AI/OpenHands/blob/de196d6da1353d3e925cd5f4bf067b16d5eacca4/openhands/resolver/issue_resolver.py#L101)) hopefully we can remove them all?

This depends on us not writing to config classes (or creating config classes) anywhere else in the resolver code. Not sure if we're there yet?

enyst:
How about, in IssueResolver.init(), something like (pseudocode)
```
# IssueResolver.__init__(args):
  # compute now and store for later read-only use
  self.workspace_base = args.output_dir + 'workspace' + args.issue_type + args.issue_number
  ...
  # afterwards
  factory = IssueHandlerFactory(
            owner=self.owner,
            repo=self.repo,
            ...
            issue_type=self.issue_type,
  )
...
def process_issue:
...
  # use self.workspace_base
```

Maybe that's missing some complexity? It seems that issue_type is [here](https://github.com/All-Hands-AI/OpenHands/blob/8c0ed36125527a5755e0e2b308d8eaec728d9268/openhands/resolver/issue_resolver.py#L133), issue_number is [here](https://github.com/All-Hands-AI/OpenHands/blob/8c0ed36125527a5755e0e2b308d8eaec728d9268/openhands/resolver/issue_resolver.py#L172).

Please feel free to shut it down if wrong!

enyst:
What do you think, @kotauchisunsun , is it ready to merge or close to? I looked at every commit, and you did the refactoring step by step carefully, and it looks safe to me, unlikely to break things. Also, the tests pass.
Do you give it a try locally, to see that it does work?

kotauchisunsun:
@enyst 
I think it's okay to merge.

I’ve confirmed the runtime behavior locally using the following command, and all tests pass:

```
poetry run pytest --cov=openhands/resolver/ ./tests/unit/resolver/
```

I also verified the behavior in my own sample repository:
https://github.com/kotauchisunsun/openhands-debug-demo/issues/25

Here is the execution log:
https://github.com/kotauchisunsun/openhands-debug-demo/actions/runs/15375182896/job/43259119584

By adding the fix-me-experimental label, I was able to check out and run this branch:
https://github.com/kotauchisunsun/openhands-debug-demo/actions/runs/15375182896/workflow#L241

The pull request was successfully created:
https://github.com/kotauchisunsun/openhands-debug-demo/pull/26

enyst:
OK, thank you! I really really appreciate this work, @kotauchisunsun , it's something I think we've been missing for quite a while now. ❤️ 

The changes in this PR are around loading and overriding the app configuration. For a bit of history, the resolver used to be developed as an external script, it initialized itself first and then loaded `openhands`, basically, and that is why it has quite a number of duplicate code or things that can be simplified. Loading - and therefore, following - the app configuration, overriding where needed and otherwise using it normally, is something I think we needed to do ever since we moved the resolver into core `openhands`: it's deeper integration. That is, it runs the resolver as a normal part of the system, a well-behaved component 😅, as opposed to doing its own thing from outside -ish.

Cc: @malhotra5 @neubig

PR #8647, opened 2025-05-22
Fix conversation initial state

tofarr:

---

---

### DIFF
--- a/frontend/__tests__/context/ws-client-provider.test.tsx
+++ b/frontend/__tests__/context/ws-client-provider.test.tsx
@@ -57,7 +57,7 @@ describe("WsClientProvider", () => {
   beforeEach(() => {
     vi.clearAllMocks();
     vi.mock("#/hooks/query/use-user-conversation", () => ({
-      useUserConversation: () => {
+      useActiveConversation: () => {
         return { data: {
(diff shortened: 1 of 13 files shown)


amanape (commented):
Can't seem to wrap my finger around this. `useActiveConversation` doesn't look like it is doing anything different than `useUserConversation` was with the `refetchInterval`. Would things still work as expected if we just update `useUserConversation` to call `useConversation` directly instead?

tofarr:
> Can't seem to wrap my finger around this. `useActiveConversation` doesn't look like it is doing anything different than `useUserConversation` was with the `refetchInterval`. Would things still work as expected if we just update `useUserConversation` to call `useConversation` directly instead?

`useActiveConversation` is a simplification for using the active conversation - a bit of syntactic sugar.

* `useUserConversation` takes in a `conversationId` as a parameter, while `useActiveConversation` uses the current conversationId
* `useActiveConversation` has specific logic around the refreshInterval which may not be desirable in cases outside the active conversation - while the status is `STARTING`, it refetches every 2 seconds.

I think an issue causing confusion here is that the hook `useConversation` should probably be renamed to `useActiveConversationId`

amanape (approved):

PR #8677, opened 2025-05-23
Adding LLM Based Editing capability

leandermaben:

This PR updates code for LLM-Based Editing and enables evaluating LLM based editing on SWE-Bench.

---

---


openhands/agenthub/codeact_agent/tools/llm_based_edit.py
```diff
@@ -2,10 +2,18 @@
...
+* The draft content doesn't need to be exactly the same as the existing file; the assistant may skip unchanged lines using commments like `# ... existing code ...` to indicate unchanged sections.
 * IMPORTANT: For large files (e.g., > 300 lines), specify the range of lines to edit using `start` and `end` (1-indexed, inclusive). The range should be smaller than 300 lines.
+* -1 indicates the last line of the file when used as the `start` or `end` value.
+* Keep atleast one unchanged line before the changed section and after the changed section wherever possible.
+* Make sure to set the `start` and `end` to include all the lines in the original file referred to in the draft of the new file content. Failure to do so will result in bad edits.
 * To append to a file, set both `start` and `end` to `-1`.
 * If the file doesn't exist, a new file will be created with the provided content.
+* IMPORTANT: Make sure you include all the required indentations for each line of code in the draft, otherwise the editted code will be incorrectly indented.
```
enyst, line 12:
```suggestion
* IMPORTANT: Make sure you include all the required indentations for each line of code in the draft, otherwise the edited code will be incorrectly indented.
```

evaluation/benchmarks/swe_bench/run_infer.py
```diff
@@ -254,15 +255,21 @@ def get_config(
         workspace_base=None,
         workspace_mount_path=None,
     )
+    config_copy = copy.deepcopy(config)
+    load_from_toml(config_copy)
```
enyst, line 259: Just a thought: this works, but maybe we can use a similar way with how we load the main LLM : calling `get_llm_config_arg(draft_editor)`?
)

openhands/runtime/utils/edit.py
```diff
@@ -26,47 +26,38 @@
...
-HERE IS THE OLD VERSION OF THE FILE:
-```
-{old_contents}
-```
+Code changes will be provided in the form of a draft. You will need to apply the draft to the original code. 
+The original code will be enclosed within `<original_code>` tags.
+The draft will be enclosed within `<update_snippet>` tags.
+You need to output the update code within `<updated_code>` tags.
+
+Within the `<updated_code>` tag, include only the final code after updation. Do not include any explanations or other content within these tags.
```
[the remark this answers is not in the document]
leandermaben: Would it be ok to leave this one unchanged? I actually used this exact prompt to finetune several editor models.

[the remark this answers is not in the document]
enyst: Oh, I see, no problem, we can use it as it is.

To clarify, FWIW my concern on small details like this comes from some papers like this [for example](https://arxiv.org/abs/2406.19783), which find performance decreases in code generation when there are typos or other perturbations in the prompt (up to 21.2%, and 4.8% to 6.1% on average).

(Out of curiosity, it could be interesting to know the difference in performance the other way around, LLMs fined-tuned on slightly non-standard instructions, compared to the use with corrections. 😅 If there really is a difference in this case, then I guess it means other LLMs applied as editors would start at a disadvantage?)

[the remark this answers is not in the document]
enyst: I actually didn't realize that! You're absolutely right, it's just a bit non-standard, so presumably less present in training data. 🤷 I don't know how it would fare, I'd be slightly curious, in both directions. 😅

xingyaoww (approved):
LGTM!

enyst (commented):
Maybe it's best for now to use an env var to set the `enable llm editor` value for eval, instead of loading from toml.

The reason for that is the same reason why we don't use `load_from_toml` at all in evals: because we can have issues when a configuration file meant for our other work gets used during evals or viceversa.

We could set instead the value in env, and pick it up like we do with some other like condenser e.g.:
https://github.com/All-Hands-AI/OpenHands/blob/3ccc96d79439080e17723420b61883692c1758ad/evaluation/benchmarks/swe_bench/README.md?plain=1#L78

(maybe we can look sometime at loading toml from a strict, well known, eval location, instead of the default location, then that would solve this problem, I think.. 🤔)

xingyaoww:
☝️ Engel has a good point here

leandermaben:
> Maybe it's best for now to use an env var to set the `enable llm editor` value for eval, instead of loading from toml.
> 
> The reason for that is the same reason why we don't use `load_from_toml` at all in evals: because we can have issues when a configuration file meant for our other work gets used during evals or viceversa.
> 
> We could set instead the value in env, and pick it up like we do with some other like condenser e.g.:
> 
> https://github.com/All-Hands-AI/OpenHands/blob/3ccc96d79439080e17723420b61883692c1758ad/evaluation/benchmarks/swe_bench/README.md?plain=1#L78
> 
> (maybe we can look sometime at loading toml from a strict, well known, eval location, instead of the default location, then that would solve this problem, I think.. 🤔)

Made the change

evaluation/benchmarks/swe_bench/run_infer.py
```diff
@@ -43,6 +43,7 @@
     AppConfig,
     get_llm_config_arg,
     get_parser,
+    load_from_toml
```
enyst, line 46: We probably don't need this anymore?

openhands/agenthub/codeact_agent/tools/llm_based_edit.py
```diff
@@ -2,10 +2,18 @@
 
 _FILE_EDIT_DESCRIPTION = """Edit a file in plain-text format.
 * The assistant can edit files by specifying the file path and providing a draft of the new file content.
-* The draft content doesn't need to be exactly the same as the existing file; the assistant may skip unchanged lines using comments like `# unchanged` to indicate unchanged sections.
+* The draft content doesn't need to be exactly the same as the existing file; the assistant may skip unchanged lines using commments like `# ... existing code ...` to indicate unchanged sections.
```
enyst, line 5:
```suggestion
* The draft content doesn't need to be exactly the same as the existing file; the assistant may skip unchanged lines using comments like `# ... existing code ...` to indicate unchanged sections.
```

enyst (approved):
Thank you!

It may be good to add a little bit of information about how to enable llm editor, to a readme, either the swebench readme file or maybe [here](https://github.com/All-Hands-AI/OpenHands/tree/main/evaluation#evaluating-different-llms) so that others can reproduce the setup

leandermaben:
> Thank you!
> 
> It may be good to add a little bit of information about how to enable llm editor, to a readme, either the swebench readme file or maybe [here](https://github.com/All-Hands-AI/OpenHands/tree/main/evaluation#evaluating-different-llms) so that others can reproduce the setup

Done! Apologies for the delay.

enyst:
Thank you! Good to see this feature get some love, it's an interesting way of doing things, and I'm looking forward to how it could improve results!

@xingyaoww I normally leave this kind of merge to you, if you don't mind - looks good to me!

PR #8692, opened 2025-05-25
bug: fix cache_control missing during convertion

CoreJa:

---

Original code introduces a bug that it drops `"{cache_control" : {'type': 'ephemeral'}}`  during conversion of `fncall` to `non_fncall`. So I added it here. The reason for this is that claude from provider openrouter is not supporting `fncall` using the api of anthropic, but a openai-commpatiable api(I'll also submit a patch to litellm to fix this btw).

A typical example of this is using APIs from `Openrouter/anthropic/` models. (direct API call to anthropic is banned in my region, so have to use Openrouter as a workaround.

**non-trivial design**:

1. `cache control` is only added whenever message parameter has such key inside
2. It only affects messages with role `tool`
3. This is actually a design fault of anthropic, there's actually a thorough discussion about it [here](https://github.com/BerriAI/litellm/issues/6422#issuecomment-2438765472) and within the code comments
https://github.com/All-Hands-AI/OpenHands/blob/77cd05c33b1eb292c336db772b973920f5b9daf4/openhands/core/message.py#L102-L126
They are mainly about how we should extract `cache_control` out to a higher level when it comes to function calls, it's just that `fn_calls conversion` didn't make the changes accordingly.

---

I haven't have the time to find any existing issues yet, but it is what happening on me. With this patch, prompt cache is working again. I can show some scrrenshot from my openrouter to see  if cache hits with the same exact prompt request:

Before patch(only 5128 tokens because only system prompt is working):

image

After patch(fully operational):

image

### DIFF
--- a/openhands/llm/fn_call_converter.py
+++ b/openhands/llm/fn_call_converter.py
@@ -544,6 +544,8 @@ def convert_fncall_messages_to_non_fncall_messages(
                 raise FunctionCallConversionError(
                     f'Unexpected content type {type(content)}. Expected str or list. Content: {content}'
                 )
+            if "cache_control" in message:
+                content[-1]["cache_control"] = {'type': 'ephemeral'}
             converted_messages.append({'role': 'user', 'content': content})
         else:
             raise FunctionCallConversionError(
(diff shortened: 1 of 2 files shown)


enyst:
Thank you for the fix and detailed explanation! This only affects, as you noted, Anthropic models served via a provider that doesn't support tool use, or tool use is [disabled by the user](https://github.com/All-Hands-AI/OpenHands/blob/77cd05c33b1eb292c336db772b973920f5b9daf4/config.template.toml#L201).

PR #8695, opened 2025-05-25
bug: fix fn_call error during API response

CoreJa:

---
This is a bug I found when receiving simulated fncall response, here's the original code snippet: https://github.com/All-Hands-AI/OpenHands/blob/77cd05c33b1eb292c336db772b973920f5b9daf4/openhands/llm/fn_call_converter.py#L728-L739

You can see line 733, the method is based on the assumption that response of simulated-fncalls should always end up with a content that `len(content)==1`, however this is **not true** for provider openrouter/anthropic。 Here's an response example a got:

```json
[
    {
        "type": "text",
        "text": "EXECUTION RESULT of [browser]:\n"
    },
    {
        "type": "text",
        "text": "LARGE CONTENT HERE......[Current URL: http://localhost:53782/]\n[Focused element bid: 8]\n\n[Action executed successfully.]\n============== BEGIN accessibility tree ==============\nRootWebArea 'VSCode......Note that only visible portion of webpage is present in the screenshot. You may need to scroll to view the remaining portion of the web-page.)\n"
    },
    {
        "type": "image_url",
        "image_url": {
            "url": "..LARGE IMAGE BASE64 CONTENT HERE...... data:image/png;base64,iVBORw0......kSuQmCC"
        },
        "cache_control": {
            "type": "ephemeral"
        }
    }
]
```

And of course such response will lead to a raised error. The main issue with such response is that, it separates the response of fn_call into 2 contents(I don't know why), and apparently fn_call_converter didn't include such case. So I made this change to concatenate all `text` content into one.

As for result, it functions pretty well.

---

Not checked yet, but I'm pretty sure whoever is using openrouter/anthropic as a provider would encounter it.

### DIFF
--- a/openhands/llm/fn_call_converter.py
+++ b/openhands/llm/fn_call_converter.py
@@ -707,6 +707,10 @@ def convert_non_fncall_messages_to_fncall_messages(
                     TOOL_RESULT_REGEX_PATTERN, content, re.DOTALL
                 )
             elif isinstance(content, list):
+                if len(content) > 1:
+                    text_concatenated = "".join(item.get("text", "") for item in content if item.get('type') == 'text')
+                    content = [{"type": "text", "text": text_concatenated}]
+
                 tool_result_match = next(
                     (
                         _match
--- a/openhands/llm/fn_call_converter.py
+++ b/openhands/llm/fn_call_converter.py
@@ -536,8 +536,8 @@ def convert_fncall_messages_to_non_fncall_messages(
             if isinstance(content, str):
                 content = prefix + content
             elif isinstance(content, list):
-                if content and content[-1]['type'] == 'text':
-                    content[-1]['text'] = prefix + content[-1]['text']
+                if content and (first_text_content := next((c for c in content if c['type'] == 'text'), None)):
+                    first_text_content['text'] = prefix + content[-1]['text']
                 else:
                     content = [{'type': 'text', 'text': prefix}] + content
             else:
@@ -707,14 +707,6 @@ def convert_non_fncall_messages_to_fncall_messages(
                     TOOL_RESULT_REGEX_PATTERN, content, re.DOTALL
                 )
             elif isinstance(content, list):
-                if len(content) > 1:
-                    text_concatenated = ''.join(
-                        item.get('text', '')
-                        for item in content
-                        if item.get('type') == 'text'
-                    )
-                    content = [{'type': 'text', 'text': text_concatenated}]
-
                 tool_result_match = next(
                     (
                         _match
--- a/openhands/llm/fn_call_converter.py
+++ b/openhands/llm/fn_call_converter.py
@@ -708,8 +708,12 @@ def convert_non_fncall_messages_to_fncall_messages(
                 )
             elif isinstance(content, list):
                 if len(content) > 1:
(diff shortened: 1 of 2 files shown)


enyst:
Thank you for the PR! This seems strange:
> The main issue with such response is that, it separates the response of fn_call into 2 contents(I don't know why), and apparently fn_call_converter didn't include such case. So I made this change to concatenate all text content into one.

The tool result is something `openhands` creates, the model's response had a tool call / function call, and `openhands` processed it, and made an observation. I don't think we intended to create a result that fails our own checks. 😅 Maybe this is a recent bug, or a browser specific thing we may have missed. 🤔

enyst:
I think you're right, I'm just saying this shouldn't have happened in the first place... and maybe I see the issue:
https://github.com/All-Hands-AI/OpenHands/blob/0c2924453f04a2348c2c6df4296dfbfe29b21739/openhands/llm/fn_call_converter.py#L539

When the tool is "browser" and, as your example shows, it has both a 'text' and an 'image_url' in the list (`content` is a list in this case), then our logic there is... wrong, isn't it? We check if the last thing in the list is a `text` type:
- if yes, concatenate the `prefix` to it (the prefix is the first thing in your example, the "EXECUTION..." part)
- if not, create a new element from the `prefix` alone and prepend it.

But in your example, the last element is an `image_url` => we create a new element.

I think the logic should be instead: if content is a list, find the first `text` type element, if any. If any, concatenate `prefix` to it, if none, only then create a new element for the prefix. What do you think, does that make sense?

CoreJa:
> When the tool is "browser" and, as your example shows, it has both a 'text' and an 'image_url' in the list (content is a list in this case), then our logic there is... wrong, isn't it? We check if the last thing in the list is a text type.

OH! I see!! I didn't check how it becomes like that in the first place, so I only fix how it parses the response. You're definitely right about this, we should never create such a content and include it in the request!! I didn't even know it is an element **we** created there!

> I think the logic should be instead: if content is a list, find the first text type element, concatenate prefix to it, if none, only then create a new element for the prefix. What do you think, does that make sense?  

Absolutely! The expectation of a simulated fncall should be wrapped into one content, especially here what we want is just to make sure the fncall starts with a PREFIX we want.

I'll commit your logic in a minute. Then you can continue reviewing or testing.

enyst:
Could you perhaps add a unit test for this? We have these tests in test_llm_fncall_converter.py` normally.

openhands/llm/fn_call_converter.py
```diff
@@ -536,8 +536,12 @@ def convert_fncall_messages_to_non_fncall_messages(
...
                 content = prefix + content
             elif isinstance(content, list):
-                if content and content[-1]['type'] == 'text':
-                    content[-1]['text'] = prefix + content[-1]['text']
+                if content and (
+                    first_text_content := next(
+                        (c for c in content if c['type'] == 'text'), None
+                    )
+                ):
+                    first_text_content['text'] = prefix + content[-1]['text']
```
enyst, line 544: Should this be `prefix + first_text_content['text']` maybe?

CoreJa: my bad, let me fix it with a unit test, and of course with lint

enyst (approved):
LGTM, thank you!

PR #8733, opened 2025-05-27
Fix openhands CLI command entry point

neubig:
## Description

This PR fixes the openhands CLI command entry point in pyproject.toml to ensure it works correctly. The issue was that the main function in openhands/cli/main.py was an async function that required a loop argument, which was not being provided by the entry point.

## Changes

1. Renamed the existing async main function to main_async
2. Added a new non-async main function that:
   - Sets up the event loop
   - Calls the async main_async function
   - Handles exceptions and cleanup
3. Updated the if __name__ == "__main__": block to call the new main function
4. Added tests to ensure the CLI command works correctly and to prevent this from breaking in the future

## Testing

- Added unit tests to verify that the main function exists and can be called without arguments
- Manually tested the command with "poetry run openhands --help" to confirm it works correctly

You can click here to [continue refining the PR](https://staging.all-hands.dev/conversations/{})

### DIFF
--- a/openhands/cli/main.py
+++ b/openhands/cli/main.py
@@ -327,7 +327,7 @@ async def run_session(
     return new_session_requested
 
 
-async def main(loop: asyncio.AbstractEventLoop) -> None:
+async def main_async(loop: asyncio.AbstractEventLoop) -> None:
     """Runs the agent in CLI mode."""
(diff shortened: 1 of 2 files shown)


neubig:
Close in favor of new one

PR #8734, opened 2025-05-27
Fix VS Code tab and other runtime-dependent features showing null

rbren:
## Description
This PR fixes an issue where the VS Code tab and other runtime-dependent features (web hosts, git changes) were showing "null" instead of loading properly when the runtime was active.

## Root Cause
The issue was in several hooks in the frontend code. The `enabled` condition was incorrectly checking if the agent state is in `RUNTIME_INACTIVE_STATES` instead of checking if it is NOT in those states.

The condition was:
```typescript
const enabled =
  conversation?.status === "RUNNING" &&
  RUNTIME_INACTIVE_STATES.includes(curAgentState);
```

This is incorrect because `RUNTIME_INACTIVE_STATES` includes states like `LOADING`, `STOPPED`, and `ERROR`. The hooks should only be enabled when the agent state is NOT in these inactive states.

## Fix
1. Changed the condition to use the logical NOT operator in the following hooks:
   - `useVSCodeUrl`
   - `useActiveHost`
   - `useGetGitChanges`

2. Created a new `useRuntimeIsReady` hook to centralize this logic:
```typescript
export const useRuntimeIsReady = (): boolean => {
  const { curAgentState } = useSelector((state: RootState) => state.agent);

  return (
    conversation?.status === "RUNNING" &&
    !RUNTIME_INACTIVE_STATES.includes(curAgentState)
  );
};
```

3. Updated all three hooks to use this new centralized hook.

This ensures that these features are only fetched when the agent is in an active state, which resolves the issue of showing "null".

## Testing
Manually verified that the VS Code tab, web hosts, and git changes now properly load when the runtime is active.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:337858a-nikolaik   --name openhands-app-337858a   docker.all-hands.dev/all-hands-ai/openhands:337858a
```

### DIFF
--- a/frontend/src/hooks/query/use-active-host.ts
+++ b/frontend/src/hooks/query/use-active-host.ts
@@ -1,21 +1,14 @@
 import { useQueries, useQuery } from "@tanstack/react-query";
 import axios from "axios";
 import React from "react";
-import { useSelector } from "react-redux";
 import OpenHands from "#/api/open-hands";
-import { RUNTIME_INACTIVE_STATES } from "#/types/agent-state";
-import { RootState } from "#/store";
 import { useConversationId } from "#/hooks/use-conversation-id";
-import { useActiveConversation } from "./use-active-conversation";
+import { useRuntimeIsReady } from "#/hooks/use-runtime-is-ready";
 
 export const useActiveHost = () => {
-  const { curAgentState } = useSelector((state: RootState) => state.agent);
   const [activeHost, setActiveHost] = React.useState<string | null>(null);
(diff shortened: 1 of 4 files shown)


amanape (approved):

rbren:
Verifying the fixes now

rbren:
Verified working!

PR #8745, opened 2025-05-27
(Hotfix): Handle cases where user secrets store doesn't exist

malhotra5:

---
We have two fixes in this PR

1. Loading secrets should be an empty list when the `user_secrets` don't exist; this is the case for a new user and the 404 errors are noisy in the UI

2. When should save incoming secrets when the existing secrets don't exist; this is always the case for a new user

---

### DIFF
--- a/openhands/server/routes/secrets.py
+++ b/openhands/server/routes/secrets.py
@@ -188,10 +188,7 @@ async def load_custom_secrets_names(
 ) -> GETCustomSecrets | JSONResponse:
     try:
         if not user_secrets:
-            return JSONResponse(
-                status_code=status.HTTP_404_NOT_FOUND,
-                content={'error': 'User secrets not found'},
-            )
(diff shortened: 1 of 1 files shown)


openhands/server/routes/secrets.py
```diff
@@ -220,31 +217,30 @@ async def create_custom_secret(
...
-            )
+        custom_secrets[secret_name] = CustomSecret(
+            secret=secret_value,
+            description=secret_description or '',
+        )
+
+        # Create a new UserSecrets that preserves provider tokens
+        updated_user_secrets = UserSecrets(
+            custom_secrets=custom_secrets,
+            provider_tokens=existing_secrets.provider_tokens if existing_secrets else {},
```
enyst, line 240: Could we perhaps add a couple of tests for the behavior of `create_custom_secret`?

malhotra5: Yup doing it atm!

PR #8756, opened 2025-05-28
Update microagent instructions

rbren:

---

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:1da841f-nikolaik   --name openhands-app-1da841f   docker.all-hands.dev/all-hands-ai/openhands:1da841f
```

### DIFF
--- a/.openhands/microagents/repo.md
+++ b/.openhands/microagents/repo.md
@@ -26,6 +26,7 @@ Backend:
 - Located in the `openhands` directory
 - Testing:
   - All tests are in `tests/unit/test_*.py`
+  - To run all the unit tests, run `poetry run pytest --forked -n auto -svv ./tests/unit`
   - To test new code, run `poetry run pytest tests/unit/test_xxx.py` where `xxx` is the appropriate file for the current functionality
   - Write all tests with pytest
 


enyst (commented):
Idk, I can see the point, but for the open source resolver, I feel like constantly telling it to not run all tests, and runtime tests fail IIRC.

Locally, too, I really don't want it to run them all.

This prompt doesn't exactly say to always run, but idk, it might nudge it to, only because it's specified. Sonnet needed restraining, not nudging. 😅 (3.7, not fully sure yet about 4, vibe-prompt-eval in progress)

Maybe, if we really need this, we could add it with an 'if', e.g. "if the user wants you to run all unit tests, ..."? Though I can see why it makes sense to just let it run, too.

rbren:
Hmm interesting. Yeah I was seeing it just run a few of the test cases and then say "done!". Worth thinking this through more

PR #8763, opened 2025-05-28
Added Webhooks to event store

tofarr:

---

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:586d6de-nikolaik   --name openhands-app-586d6de   docker.all-hands.dev/all-hands-ai/openhands:586d6de
```


rbren (approved):

enyst (commented):
This looks like a very good idea. Quick questions:

Could we have a description of this PR? It does introduce a feature

Could you tell, what was the problem with the previous HTTP attempt?

I have no idea if anyone other than AH uses nested runtimes, but the web hooks are likely interesting for other uses. For example, every now and then I hear frustration with the every growing .json files in the file store, and questions about backups or alternatives.
Could we document this feature for everyone?

tofarr:
I backed it out the previous attempt because we decided to go in a different direction and there was no explicit request from the community for it - Leaving it in meant that we would have to build / test / document both sides of a complex custom protocol, and the webhook based solution matched our needs and would likely be more useful to others too. We figured that keeping events in a local file store with a process for external synchronization was far more efficient than using a pure network attached event store.

I'll update the documentation with the with implementation details.

PR #8764, opened 2025-05-28
Add conversation-level budget cap

malhotra5:

This PR adds a new feature that allows setting a maximum budget per conversation. This is different from the existing max_budget_per_task parameter, which only limits the budget for a single task. The max_budget_per_conversation parameter limits the total budget for the entire conversation, ensuring that the total cost of a conversation does not exceed a specified threshold.

Additionally, when a user continues the conversation after hitting a budget cap (either task or conversation), the budget cap is automatically increased by 50% to allow the conversation to continue.

---

This PR implements a conversation-level budget cap by:
1. Adding a max_budget_per_conversation parameter to the AppConfig class
2. Adding max_budget_per_conversation to the Settings class
3. Updating the AgentSession to pass the parameter to the AgentController
4. Modifying the AgentController to check both task-level and conversation-level budgets
5. Adding tests for the conversation-level budget cap
6. Adding a feature to increase the budget cap by 50% when a user continues the conversation after hitting a budget cap

The implementation follows the same pattern as the existing task-level budget cap, but applies to the entire conversation rather than just the current task.

---

N/A

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:bcd003e-nikolaik   --name openhands-app-bcd003e   docker.all-hands.dev/all-hands-ai/openhands:bcd003e
```


enyst:
I wonder, is this really different than budget per task?

malhotra5:
@openhands please parameter the unit test `test_budget_increase_via_set_agent_state`

we want to vary `max_budget_per_task` and `max_budget_per_conversation` params

- When `max_budget_per_task` < `max_budget_per_conversation` and `max_budget_per_task` is exceeded, only `max_budget_per_task` should be increased by the initial amount
- When `max_budget_per_task` > `max_budget_per_conversation` and `max_budget_per_conversation` is exceeded, then the `max_budget_per_task` should be reset to its initial value and `max_budget_per_conversation` should be increased by the initial value

enyst:
Isn't that a different conversation, though? @malhotra5 

I assume each attempt should start with its own prompt and without the potentially crazy long logs of previous attempts in context. Then it seems like it's a different conversation. 🤔

enyst:
Sorry to be a bit dense, I still don't get it: the task is the conversation, I think. Budget per task starts with the start of a session/conversation, and continues for as long as there are messages added; it doesn't just get zero-ed out at some point, I don't think?

Or did that change?

malhotra5:
#9012 addresses the issues that this PR surfaced

Issue #8781, opened 2025-05-29
[Bug]: When the GitLab server’s /v4/user endpoint returns an empty avatar_url, token validation fails.


littlebai3618:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

I found a similar issue report  https://github.com/All-Hands-AI/OpenHands/issues/8534 , but I’m not sure it stems from the same root cause. The problem is hard to reproduce, yet on my self-hosted GitLab instance the /v4/user endpoint returns avatar_url = null, e.g.

```shell
{
  "id": 4979,
  "username": "baii",
  "name": "littlebai3618",
  "state": "active",
  "locked": false,
  "avatar_url": null,
  ...
}
```

Because of this, validate_provider_token fails, so I can’t add the GitLab token. The current code doesn’t handle this scenario.

### OpenHands Installation

Docker command in README

### OpenHands Version

main

### Operating System

MacOS

littlebai3618:
I’ve submitted PR https://github.com/All-Hands-AI/OpenHands/pull/8782 to fix this issue and have already validated the solution.

<img width="1444" alt="Image" src="https://github.com/user-attachments/assets/4a476b3c-69b1-4f14-83d2-7f7e0402fcc6" />

Issue #8534, opened 2025-05-16
[Bug]: Can't add github personal access token in the UI


adamcole123:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Im running it in docker using this command:

`docker run -it --rm --pull=always     -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.38-nikolaik     -e LOG_ALL_EVENTS=true -v /var/run/docker.soc
k:/var/run/docker.sock     -v ~/.openhands-state:/.openhands-state     -p 1234:3000     --add-host host.docker.internal:
host-gateway     --name openhands-app     docker.all-hands.dev/all-hands-ai/openhands:0.38`

When entering a fine-grained github token after setting up my claude api key, I get this error in the terminal running docker:

`INFO:     172.17.0.1:52164 - "POST /api/add-git-providers HTTP/1.1" 401 Unauthorized`

This is on Windows.

To be clear, the token im using starts with 'github_pat_', incase that could hint at me using the wrong token.

I have confirmed with curl that my tokens can access my account.

### OpenHands Installation

Docker command in README

### OpenHands Version

0.38.0

### Operating System

WSL on Windows

### Logs, Errors, Screenshots, and Additional Context

`INFO:     172.17.0.1:52164 - "POST /api/add-git-providers HTTP/1.1" 401 Unauthorized`

mamoodi:
I wonder if this is Windows specific. I just tried with 0.38.0 on Mac and it works fine. I'd be surprised though.
I can check it.

mamoodi:
Wait I take that back. Something happened and now I'm getting that error.

adamcole123:
> Out of curiosity is your user under and organization? Or is this your personal account?


My account is a personal account, there's an organisation linked to it but i created the organisation and the repo I want to use is a personal repo. Ill try a token with just those permissions and see what happens.

adamcole123:
Yep, still no luck with a new token. Hm.

mamoodi:
@malhotra5 any idea what could be going wrong here?

adamcole123:
Hi wondering if there's an update for this?

malhotra5:
Hmm this is a bit weird, will take a look into this soon

mamoodi:
[...]

PR #8782, opened 2025-05-29
fix: can't add gitlab personal access token and add more debug log in validate_provider_token

littlebai3618:

When the GitLab server’s /v4/user endpoint returns an empty avatar_url, token validation fails.

---

In some self-hosted GitLab instances, the avatar_url field can be returned as None.
When this happens, GitLabService.get_user fails while building a User object, because the User model expects avatar_url to be a string, triggering a pydantic_core._pydantic_core.ValidationError (see stack trace below).

```shell
  File "/Users/baii/pycharmProjects/beike/OpenHands/openhands/integrations/utils.py", line 39, in validate_provider_token
    await gitlab_service.get_user()
  File "/Users/baii/pycharmProjects/beike/OpenHands/openhands/integrations/gitlab/gitlab_service.py", line 171, in get_user
    return User(
           ^^^^^
  File "/Users/baii/Library/Caches/pypoetry/virtualenvs/openhands-ai-394k-RCl-py3.12/lib/python3.12/site-packages/pydantic/main.py", line 253, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for User
avatar_url
  Input should be a valid string [type=string_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/string_type
```

I fixed the issue by adding a check for avatar_url in the method and supplying a default value of empty string

This minimal change resolves the problem in my tests.
I also added two debug log lines to validate_provider_token to make troubleshooting easier.

During the debugging process, I also noticed that the /user/info endpoint and the runtime image build were pulling the user repository using the default gitlab.com instead of the specified GitLab domain. I have fixed both issues.

Please carefully review whether the ProviderToken.host property has a default value when the user does not provide a domain. I'm not very familiar with this repository's code。

---
https://github.com/All-Hands-AI/OpenHands/issues/8781
https://github.com/All-Hands-AI/OpenHands/issues/8534

### DIFF
--- a/openhands/integrations/utils.py
+++ b/openhands/integrations/utils.py
@@ -2,10 +2,10 @@ import traceback
 
 from pydantic import SecretStr
 
+from openhands.core.logger import openhands_logger as logger
 from openhands.integrations.github.github_service import GitHubService
 from openhands.integrations.gitlab.gitlab_service import GitLabService
 from openhands.integrations.provider import ProviderType
-from openhands.core.logger import openhands_logger as logger
 
 
 async def validate_provider_token(
@@ -29,7 +29,9 @@ async def validate_provider_token(
         await github_service.verify_access()
         return ProviderType.GITHUB
     except Exception as e:
-        logger.debug(f'Failed to validate Github token: {e} \n {traceback.format_exc()}')
+        logger.debug(
+            f'Failed to validate Github token: {e} \n {traceback.format_exc()}'
+        )
 
     # Try GitLab next
     try:
@@ -37,6 +39,8 @@ async def validate_provider_token(
         await gitlab_service.get_user()
         return ProviderType.GITLAB
     except Exception as e:
-        logger.debug(f'Failed to validate GitLab token: {e} \n {traceback.format_exc()}')
+        logger.debug(
+            f'Failed to validate GitLab token: {e} \n {traceback.format_exc()}'
+        )
 
     return None


openhands/runtime/base.py
```diff
@@ -360,12 +360,7 @@ async def clone_or_init_repo(
             return ''
 
         provider = repository.git_provider
-        provider_domains = {
-            ProviderType.GITHUB: 'github.com',
-            ProviderType.GITLAB: 'gitlab.com',
-        }
-
-        domain = provider_domains[provider]
+        domain = git_provider_tokens[provider].host
```
malhotra5, line 363:
```suggestion
        provider_domains = {
            ProviderType.GITHUB: 'github.com',
            ProviderType.GITLAB: 'gitlab.com',
        }
        domain = git_provider_tokens[provider].host if git_provider_tokens[provider].host else provider_domains[provider]
```

Host doesn't always exist so we should default to the provider domains above

malhotra5:
Thanks you for this! I've added one comment but besides that this LGTM

littlebai3618:
> Thanks you for this! I've added one comment but besides that this LGTM

I've updated and resubmitted the changes to /runtime/base.py based on your feedback.

littlebai3618:
fix lint error

```shell
trim trailing whitespace.................................................Passed
fix end of files.........................................................Passed
check yaml...........................................(no files to check)Skipped
debug statements (python)................................................Passed
pyproject-fmt........................................(no files to check)Skipped
Validate pyproject.toml..............................(no files to check)Skipped
ruff.....................................................................Passed
ruff-format..............................................................Passed
mypy.....................................................................Passed
[main 65a5f5eb] style: apply ruff formatting to utils.py
```

malhotra5 (approved):

PR #8792, opened 2025-05-29
Refactor agent controller's filter

enyst:

---

Tim has added an `EventFilter` for choosing events from the stream, but we're not using it yet. This PR proposes a refactoring of agent's controller filtering mechanism to make use of it.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:bf8fa26-nikolaik   --name openhands-app-bf8fa26   docker.all-hands.dev/all-hands-ai/openhands:bf8fa26
```

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -145,6 +145,18 @@ class AgentController:
                 EventStreamSubscriber.AGENT_CONTROLLER, self.on_event, self.id
             )
 
+        # filter out events that are not relevant to the agent
+        # so they will not be included in the agent history
+        self.agent_history_filter = EventFilter(
+            exclude_types=(
+                NullAction,
+                NullObservation,
+                ChangeAgentStateAction,
+                AgentStateChangedObservation,
+            ),
+            exclude_hidden=True,
+        )
+
         # state from the previous session, state from a parent agent, or a fresh state
         self.set_initial_state(
             state=initial_state,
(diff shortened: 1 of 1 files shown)


openhands/controller/agent_controller.py
```diff
@@ -169,6 +164,16 @@ def __init__(
         # replay-related
         self._replay_manager = ReplayManager(replay_events)
 
+        self.agent_history_filter = EventFilter(
+            exclude_types=(
+                NullAction,
+                NullObservation,
+                ChangeAgentStateAction,
+                AgentStateChangedObservation,
```
amanape, line 172: This won't affect the events received on the client right?

amanape (approved):

openhands/controller/agent_controller.py
enyst: Indeed, it won't, this is only refactoring, it doesn't change behavior; and these are for the agent's LLM context. The server has its own manual filtering, like [this one](https://github.com/All-Hands-AI/OpenHands/blob/e4d84946dbb64e8b5157c8cade4e38f239fb8b23/openhands/server/listen_socket.py#L136)

PR #8798, opened 2025-05-29
Allowing local runtimes to have domains

tofarr:

This allows specifying a domain via environment variables - either `RUNTIME_URL` or a combination of `RUNTIME_URL_PATTERN` and `HOSTNAME`. Fallback is still localhost

---
This is useful if you want local runtime instances to be accessible in a non local context - such as a nested sandbox. It allows the `vscode_url` and `web_hosts` to still work.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:9a10a35-nikolaik   --name openhands-app-9a10a35   docker.all-hands.dev/all-hands-ai/openhands:9a10a35
```


openhands/runtime/impl/local/local_runtime.py
```diff
@@ -388,17 +388,35 @@ def close(self) -> None:
...
 
+    @property
+    def base_url(self) -> str:
+
+        runtime_url = os.getenv('RUNTIME_URL')
+        if runtime_url:
+            return runtime_url
+
+        #TODO: This could be removed if we had a straightforward variable containing the RUNTIME_URL in the K8 env.
+        runtime_url_pattern = os.getenv('RUNTIME_URL_PATTERN')
```
rbren, line 399: this should probably be in config.sandbox

openhands/runtime/impl/local/local_runtime.py
```diff
@@ -388,17 +388,35 @@ def close(self) -> None:
...
+        hostname = os.getenv('HOSTNAME')
+        if runtime_url_pattern and hostname:
+            runtime_id = hostname.split('-')[1]
+            runtime_url = runtime_url_pattern.format(runtime_id=runtime_id)
```
rbren, line 403: this logic seems maybe a bit brittle...

I do think we can set RUNTIME_URL in the runtime-api backend since we know the runtime id there

openhands/runtime/impl/local/local_runtime.py
tofarr: I see there is already a `local_runtime_url` in `SandboxConfig` - I'm now reusing that.

tofarr:
> Approving to unblock but left a couple comments. I suppose non-blocking but the HOSTNAME->ID logic worries me a bit, seems easy to break

Once the runtime api is updated, we can remove the brittle code.

PR #8802, opened 2025-05-29
feat: Add support for Mistral AI models with customizable safety sett…

Mahdiglm:

End-user friendly description of the problem this fixes or functionality this introduces:
Add support for customizable safety settings for Mistral AI and Gemini models.

---
**Summarize what the PR does, explaining any non-trivial design decisions:**
- Added a new `safety_settings` field to the LLMConfig model to support configuration of model-specific safety settings
- Updated the LLM initialization to apply custom safety settings for Mistral AI models
- Added default safety settings for Mistral models with "low" thresholds for all categories
- Created comprehensive documentation for safety settings in a new "Model Settings" page
- Added examples in config.template.toml for both Mistral and Gemini safety settings
- Updated the main LLMs documentation to link to the new model settings page

---
N/A

### DIFF
--- a/openhands/core/config/llm_config.py
+++ b/openhands/core/config/llm_config.py
@@ -89,7 +89,7 @@ class LLMConfig(BaseModel):
     seed: int | None = Field(default=None)
     safety_settings: list[dict[str, str]] | None = Field(
         default=None,
-        description='Safety settings for models that support them (like Mistral AI and Gemini)'
+        description='Safety settings for models that support them (like Mistral AI and Gemini)',
     )
 
     model_config = {'extra': 'forbid'}


Mahdiglm:
Thanks for the feedback and resetting Gemini's safety settings!

About Mistral: their Chat API safe_prompt is turned off by default. Turning it on simply inserts a generic safety prompt at the start. They also offer a separate Moderation API that scores content more specifically (hate, sexual, violence, etc.).

This PR's safety_settings that are configurable are an effort to give more granular control than on/off. The idea is to let users configure safety rules like they would with a dedicated Moderation API or like Gemini's settings. Having strict defaults can also block things like code generation, so making it configurable balances safety and usability. Effectively, it is a fluid setup to change with providers bringing in enhanced safety features and letting users select what works best for them.

openhands/llm/llm.py
```diff
@@ -82,6 +82,10 @@
     'o4-mini-2025-04-16',
     'gemini-2.5-pro',
     'gpt-4.1',
+    'mistral-large',
```
enyst: I'm not sure about adding this.

The Mistral model best performing with OH is the new devstral model, and it doesn't work well with tool calling.

In general, we have added here only models that we have evaluated and saw that they're *better* with tool calling than without. I don't think we know about these, so it's best to leave them out IMO.

enyst (approved):

@mamoodi it looks good to me. WDYT?

mamoodi:
> Thank you for the contribution!
> 
> @mamoodi it looks good to me. WDYT?

Looks good! I'm not too familiar with local llms.

openhands/core/config/llm_config.py
```diff
@@ -86,6 +87,9 @@ class LLMConfig(BaseModel):
     native_tool_calling: bool | None = Field(default=None)
     reasoning_effort: str | None = Field(default='high')
     seed: int | None = Field(default=None)
+    safety_settings: list[dict[str, str]] | None = Field(
+        default=None,
+        description="Safety settings for models that support them (like Mistral AI and Gemini)")
```
enyst:
```suggestion
        description='Safety settings for models that support them (like Mistral AI and Gemini)')
```

openhands/core/config/llm_config.py
```diff
@@ -86,6 +87,9 @@ class LLMConfig(BaseModel):
     native_tool_calling: bool | None = Field(default=None)
     reasoning_effort: str | None = Field(default='high')
     seed: int | None = Field(default=None)
+    safety_settings: list[dict[str, str]] | None = Field(
+        default=None,
+        description='Safety settings for models that support them (like Mistral AI and Gemini)')
```
enyst:
```suggestion
        description='Safety settings for models that support them (like Mistral AI and Gemini)'
    )
```

openhands/core/config/llm_config.py
```diff
@@ -86,6 +87,10 @@ class LLMConfig(BaseModel):
     native_tool_calling: bool | None = Field(default=None)
     reasoning_effort: str | None = Field(default='high')
     seed: int | None = Field(default=None)
+    safety_settings: list[dict[str, str]] | None = Field(
+        default=None,
+        description='Safety settings for models that support them (like Mistral AI and Gemini)'
```
enyst:
```suggestion
        description='Safety settings for models that support them (like Mistral AI and Gemini)',
```

PR #8807, opened 2025-05-30
docs: local serving with ollama documented

Truenya:
- [ + ] This change is worth documenting at https://docs.all-hands.dev/
- [ - ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

Just expanded part of the documentation, sharing my experience.

### DIFF
--- a/docs/usage/llms/local-llms.mdx
+++ b/docs/usage/llms/local-llms.mdx
@@ -129,13 +129,13 @@ vllm serve all-hands/openhands-lm-32b-v0.1 \
 ### Create an OpenAI-Compatible Endpoint with Ollama
 
 - Install Ollama following [the official documentation](https://ollama.com/download).
-- Example launch command for Devstral LM 24B:
 - For Ollama configuration, use `ollama/<modelname>` as custom model in web. Api key also can be set to `ollama`.
+- Example launch command for Devstral LM 24B:
 
 ```bash
-ollama pull devstral:latest
 OLLAMA_CONTEXT_LENGTH=32768 OLLAMA_HOST=0.0.0.0:11434 OLLAMA_KEEP_ALIVE=-1 nohup ollama serve&
 #The minimum context size is ~8196, even the system prompt won't fit smaller
+ollama pull devstral:latest
 ```
 
 ## Advanced: Run and Configure OpenHands
@@ -161,8 +161,6 @@ ollama_base_url="http://localhost:8000"
 ```
 
 Start OpenHands using `make run`.
-
-Also, with a local launch, if Docker Runtime is selected, you should add `127.0.0.1 host.docker.internal` line to end of `/etc/hosts` on your Linux machine.
 
 ### Configure OpenHands
 


docs/modules/usage/llms/local-llms.md
```diff
@@ -152,9 +164,11 @@ ollama_base_url="http://localhost:8000"
 
 Start OpenHands using `make run`.
 
+Also, with a local launch, if Docker Runtime is selected, you should add `127.0.0.1 host.docker.internal` line to end of `/etc/hosts` on your Linux machine.
+
```
enyst, line 168:
```suggestion
```
Is this necessary?

Truenya: Here's how it works for me.

Explanation:
1. when you call `make run` on your local machine
2. openhands starts the backend and frontend on your local network and machine.
3. when you create a conversation, openhands by default creates a docker container on the docker network with shared access to the local network via address `host.docker.internal` (written in `/etc/hosts` inside the container)
4. on the web page, you specify this address as `http://host.docker.internal:11434`. But `host.docker.internal` is not accessible from the local network, it is simply unknown.
5. add an entry that maps the address to the local network on the local machine, similar to what is done by default in docker
`    --add-host host.docker.internal:host-gateway \ `

Truenya: Probably it has other solution, not highlighted in documentation yet?

enyst: Maybe the problem is here

> on the web page, you specify this address as http://host.docker.internal:11434

If you are running with `make run`, as above, you could just specify http://localhost:11434 there?

Truenya: This will not work, because agents runs in docker container, with another localhost

enyst: Wait, maybe we are confusing things. When running on the local machine with `make build / make run`, the agent runs on local machine, the agent LLM code also runs on local machine, uses the `base_url` in the llm config to connect to e.g. Ollama, and sends events (actions to execute) to the `runtime`, a container container. The runtime is a server that runs inside the container, and executes those actions.

On the other hand, when running with the command `docker run` from the readme, then we have 2 containers: an `app` container, and the `runtime` container. The `app` container has the agent and llm connections now, and indeed, it then needs to connect to other containers. In that case, it needs this "host.docker.internal"?

enyst: My point is that
- with `make run`, the connection to Ollama happens from code running outside any docker container
- with `docker run ... app`, the connection to Ollama happens from inside a docker container (`openhands-app`)

Truenya: > My point is that
> 
>     * with `make run`, the connection to Ollama happens from code running outside any docker container
> 
>     * with `docker run ... app`, the connection to Ollama happens from inside a docker container (`openhands-app`)

But it is not so

Regardless of whether it is launched in Docker or via make run, containers are used for agent runtime by default.
It is not safe to use local runtime for agents.
But when launched in cli, local runtime for agents is used by default.

Truenya: By default, as a result of experiments without additional configuration, for the main branch yesterday the following was obtained:
-     with make run, the connection to Ollama happens from code running outside any docker container and from inside a docker containers (runned agents)
-     with docker run ... app, the connection to Ollama happens only from inside a docker containers (openhands-app, openhands-runtime-<uuid>)

enyst: > Regardless of whether it is launched in Docker or via make run, containers are used for agent runtime by default.
> It is not safe to use local runtime for agents.

You're right, and I hear you, but we are talking about different things:
- yes, the runtime is in a docker container (`openhands-runtime`) by default
- it's not a 'local' runtime, it's a 'docker' sandbox ^ by default
- the 'runtime' container _executes_ file operations (edits, etc) and everything else the agent needs executed, they're not local operations, they're sandboxed
- the commands and file operations and others are inside docker runtime, by default
- _however_, that container does not connect to Ollama

enyst: Let's take the case:
- running with `make run`
- on defaults => runtime = `docker`

Please let me look at the execution flow:
- agent controller runs the agent, every step
- _every step_, the agent:
  - gathers the history of actions and observations that happened until now
  - sends them to the LLM
  - receives a response (a new thing that the LLM wants to "do")
  - interprets the response and gets an "action"
  - let's say the LLM wants to execute `ls -la`
- agent controller sees the action (command action, `ls -la`)
- it knows such actions _MUST happen inside the sandbox_ (`openhands-runtime`)
- so it sends the action to the runtime client (docker client)
- which sends it inside docker container (inside `openhands-runtime`)
- (inside sandbox) checks permissions, etc etc
- (inside sandbox) executes `ls -la`
- (inside sandbox) gathers logs
- (inside sandbox) handles errors, etc
- (inside sandbox) grabs the output (the list of files/folders)
- (inside sandbox) sends the output _outside_
- the docker client receives the output (list of files/folders)
- saves the event, a command observation
- agent controller sees there's a command observation
- and calls the agent again, to make a new step, with the new observation (the list of files)
=> next step

So what I mean is: all actions on the filesystem happen inside the sandbox, _every step_. Every step ALSO contains a part orchestrating everything. The part outside the `runtime` includes the connection to LLM/Ollama.

Truenya: oh, i'll recheck, thanks

docs/modules/usage/llms/local-llms.md
```diff
@@ -128,6 +128,18 @@ vllm serve all-hands/openhands-lm-32b-v0.1 \
     --enable-prefix-caching
 ```
 
+### Create an OpenAI-Compatible Endpoint with Ollama
+
+- Install Ollama following [the official documentation](https://ollama.com/download).
+- Example launch command for Devstral LM 24B:
```
mamoodi: Thanks for your PR!

This example seems empty. Is the command below in the bash code the example?

Truenya: Yes.

Truenya: moved pieces of text for better clarity

PR #8818, opened 2025-05-30
Fix event stream replay during new connections to handle concurrent events

rbren:

Fixes an issue where new events that arrive during the replay of the event stream for a new connection might not be sent to the client. This ensures that clients always receive all events, even if new events are added to the event store during the replay process.

---

This PR modifies the event stream replay logic in `listen_socket.py` to continue replaying events until no new events are available. The implementation uses a loop that:

1. Tracks the highest event ID seen so far
2. Creates a new `AsyncEventStoreWrapper` for each iteration, starting from the next event after the latest one processed
3. Processes all available events, updating the highest event ID seen
4. Continues looping until no new events are processed in an iteration

This ensures that even if new events come in during the replay process, they will be captured and sent to the client in subsequent iterations of the loop.

---

N/A

### DIFF
--- a/openhands/server/listen_socket.py
+++ b/openhands/server/listen_socket.py
@@ -132,24 +132,49 @@ async def connect(connection_id: str, environ: dict) -> None:
         agent_state_changed = None
         if agent_loop_info is None:
             raise ConnectionRefusedError('Failed to join conversation')
-        async_store = AsyncEventStoreWrapper(
-            agent_loop_info.event_store, latest_event_id + 1
-        )
-        async for event in async_store:
-            logger.debug(f'oh_event: {event.__class__.__name__}')
-            if isinstance(
-                event,
-                (NullAction, NullObservation, RecallAction),
-            ):
-                continue
-            elif isinstance(event, AgentStateChangedObservation):
-                agent_state_changed = event
-            else:
-                await sio.emit('oh_event', event_to_dict(event), to=connection_id)
-        if agent_state_changed:
-            await sio.emit(
-                'oh_event', event_to_dict(agent_state_changed), to=connection_id
+        # Keep track of the highest event ID we've seen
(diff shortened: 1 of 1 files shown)


openhands/server/listen_socket.py
```diff
@@ -119,24 +120,26 @@ async def connect(connection_id: str, environ: dict) -> None:
...
         conversation_init_data = await setup_init_convo_settings(user_id, providers_set)
-        agent_loop_info = await conversation_manager.join_conversation(
-            conversation_id,
-            connection_id,
-            conversation_init_data,
-            user_id,
+
+        # Create an EventStore to access the events before joining the conversation
+        event_store = EventStore(
+            conversation_id, conversation_manager.file_store, user_id
```
tofarr: We should make sure this doesn't throw a file not found error if the conversation does not exist yet - if we are starting a new one, and throw a connection refused error

rbren: @openhands please address this feedback. Look at other places in the code base where we create EventStores to check out how to handle the exception

openhands/server/listen_socket.py
```diff
@@ -119,24 +120,39 @@ async def connect(connection_id: str, environ: dict) -> None:
...
+                conversation_id, conversation_manager.file_store, user_id
+            )
+        except FileNotFoundError:
+            logger.warning(
+                f'No events found for conversation {conversation_id}. Creating a new EventStore.'
+            )
+            event_store = EventStore(
+                conversation_id, conversation_manager.file_store, user_id, cur_id=0
+            )
+        except Exception as e:
```
rbren:
```suggestion
```

rbren:
@OpenHands please fix the failing actions on PR https://github.com/All-Hands-AI/OpenHands/pull/8818

Issue #7962, opened 2025-04-20
Massive performance problems - local runtime restarted on workspace requests


MischaPanch:
The workspace features in local runtime are essentially unusable. A new LocalRuntime object is being initialized at any request, like opening a directory or clicking a tab. The UI is unresponsive and gives a bad user experience.

There is no need to ever re-initialize the LocalRuntime during a user interaction session in the chat (or any other Runtime for that matter, I believe) - is it fixed in the OpenHands architecture that the runtime is recreated on all requests?

happyherp:
I had similiar issues. Are you also running this with windows wsl on a a bit older machine?

because my suspicion is, that the localruntime starts, works for a while, and then is shut down. It is then restarted on the next request. 

I found that disabling jupyter and the browser check in the runtime made it start faster and reliable on my system(which might have the same issues as you). 

Here are the relevant commits from one of my experimental branches:

https://github.com/happyherp/OpenHands/commit/368832b91c6bf6adc01ddcb88d72db6d2c3e384a
https://github.com/happyherp/OpenHands/commit/df684308faa72737922e7863467173f4070e4db6

MischaPanch:
I'm running on ubuntu and with a fairly powerful machine with a lot of ram. I don't have the feeling this is because my machine shuts down the runtime

happyherp:
Can you show the logs from when it does a restart? Ideally, with DEBUG=1

MischaPanch:
Sure, will do that later, probably tomorrow. The runtime was restarted each time I "opened" a directory on the UI

enyst:
This should be fixed by https://github.com/All-Hands-AI/OpenHands/pull/8821 . I’ll close this, but please feel free to reopen if you see fit.

PR #8821, opened 2025-05-30
Fix LocalRuntime to properly handle existing subprocesses

tofarr:

Fixes an issue where LocalRuntime would always start a new subprocess when connect is called, regardless of whether one was already running for the session. This could lead to resource leaks and unexpected behavior.

---
This PR updates the LocalRuntime class to properly handle the attach_to_existing parameter similar to DockerRuntime and RemoteRuntime. It adds a global dictionary to track running server processes by session ID and modifies the connect() method to check for existing servers before starting new ones. The close() method is also updated to respect the attach_to_existing parameter and not terminate the server when set to true. Additionally, a delete() class method is added to properly clean up server processes when a conversation is deleted.

A ServerInfo dataclass is used to store information about running servers, making the code more maintainable and easier to understand than using tuples.

---
N/A

### DIFF
--- a/openhands/runtime/impl/local/local_runtime.py
+++ b/openhands/runtime/impl/local/local_runtime.py
@@ -6,6 +6,7 @@ import subprocess
 import sys
 import tempfile
 import threading
+from dataclasses import dataclass
 from typing import Callable
 
 import httpx
@@ -37,6 +38,20 @@ from openhands.runtime.utils import find_available_tcp_port
 from openhands.runtime.utils.command import get_action_execution_server_startup_command
 from openhands.utils.async_utils import call_sync_from_async
 from openhands.utils.tenacity_stop import stop_if_should_exit
+
+
+@dataclass
+class ServerInfo:
+    """Information about a running server process."""
+
+    process: subprocess.Popen
+    port: int
+    log_thread: threading.Thread
+    log_thread_exit_event: threading.Event
+
+
+# Global dictionary to track running server processes by session ID
+_RUNNING_SERVERS: dict[str, ServerInfo] = {}
 
 
 def get_user_info() -> tuple[int, str | None]:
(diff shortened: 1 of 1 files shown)


enyst:
Probably related
- https://github.com/All-Hands-AI/OpenHands/issues/7962

openhands/server/routes/manage_conversations.py
```diff
@@ -60,7 +61,9 @@ class InitSessionRequest(BaseModel):
     replay_json: str | None = None
     suggested_task: SuggestedTask | None = None
     conversation_instructions: str | None = None
-    conversation_id: str = Field(default_factory=lambda: uuid.uuid4().hex)
+    # Only nested runtimes require the ability to specify a conversation id, and it could be a security risk
+    if os.getenv('ALLOW_SET_CONVERSATION_ID', '0') == '1':
```
enyst, line 65: Maybe I don't fully get this. A security risk for who, for nested runtimes? But nested runtimes have it enabled.

On the other hand, this is the session id. While this code is only used from FE, there is no reason why an interface wouldn't want to set a session name. In which case we may want to set it to something like "name_somehash". That's humanly recognizable, which can be nice to have.

Actually, it's the API endpoint, so we don't know if it's only FE?

enyst: Actually, I take this back. It's used only for new conversation, and we don't need an `id` here. Maybe something else, a name, a nickname, but not the id from outside.

enyst: Can we test that this works normally with the default use case, that is, docker runtime, local, no env vars?

tofarr: Sorry for the brain dump - but I'm trying to answer your question thoroughly:

We allow setting the conversation id in a nested runtime, since we can then make it the same as the id in the outer runtime, and for the moment this makes sense, as it makes the frontend code simpler.

The problem is that in a shared environment, it may be unsafe to do this. (What if a hacker uses the same conversation id as an existing conversation to try and take it over and the devs have not kept up with their user_id checks? What if they choose an id like `/opt/...` or do something else nasty knowing that it is sometimes used in a file path?) - it is inherently safer to insist the server come up with the id.

 (PS: The nested runtime having the same id as the containing runtime could change if we decide to expand the concept of nested runtimes - for example by allowing a user to run all their conversations in a single container. This may actually make things a lot faster and may be worth doing - though we need to seriously think it through.)

tofarr: When developing this, I've been testing:
* The Docker Runtime
* The Local Runtime with Nesting.
* The Local Runtime
* The Remote Runtime

Part of the reason for this feature is that I don't think any real testing of the LocalRuntime had been done (At least within the confines of the web app - maybe it had only been invoked from the CLI? I found quite a few issues with it. I understand the reticence - having it able to make global changes on a dev machine is scary!

enyst: Good points! Just a quick note. To be fair, I think it would be unreasonable to expect that we'd manually test all runtimes even in PRs on runtime changes. They're getting too many and some very different, and unreasonable expectations are no fun. 😅(I really meant only the default, above) 

Hopefully in time we add more to CI, our saving grace. 🫠

IMHO, it would be incredibly helpful if @li-boxuan could take a look for the Windows LocalRuntime, to see if some weird stuff makes it in. Maybe we will get sometime to the point where Windows should really have its own subclass.

The CLIRuntime tweak looks obviously fine (no clue why we didn't see it...)

enyst (approved):

openhands/core/config/openhands_config.py
```diff
@@ -104,7 +105,7 @@ class OpenHandsConfig(BaseModel):
     max_concurrent_conversations: int = Field(
         default=3
     )  # Maximum number of concurrent agent loops allowed per user
-    mcp_host: str = Field(default='localhost:3000')
+    mcp_host: str = Field(default=f'localhost:{os.getenv("port", 3000)}')
```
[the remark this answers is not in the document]
rbren: Maybe we can do `os.getenv("MCP_PORT", os.getenv("port", 3000))` and use that elsewhere? non blocking

rbren (approved):

openhands/server/routes/manage_conversations.py
li-boxuan: I didn't thoroughly test all scenarios but basic usage of windows localruntime + frontend UI seems fine!

PR #8830, opened 2025-06-01
Fix all broken docs links across the project

llamantino:

Fixes all remaining broken links across the codebase that still pointed to outdated https://docs.all-hands.dev/modules/ URLs after the docs site was changed.

---

This PR updates every reference to the old https://docs.all-hands.dev/modules/ URLs across the entire project (README files, docs, etc.). Each instance was replaced with the new path https://docs.all-hands.dev/, removing /modules.

**Please verify that this change doesn’t break anything.**

---


neubig (approved):
Thank you! I checked several links, and it seems like all of the new ones work. We can probably more systematically check other ones in the future.

PR #8833, opened 2025-06-01
Reduce the noise from loggers

tofarr:

---

---

### DIFF
--- a/openhands/core/logger.py
+++ b/openhands/core/logger.py
@@ -385,10 +385,20 @@ if LOG_TO_FILE:
     )  # default log to project root
     openhands_logger.debug(f'Logging to file in: {LOG_DIR}')
 
-# Exclude LiteLLM from logging output
-logging.getLogger('LiteLLM').disabled = True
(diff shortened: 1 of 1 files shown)


openhands/core/logger.py
```diff
@@ -385,10 +385,20 @@ def log_uncaught_exceptions(
...
+    'LiteLLM',
+    'LiteLLM Router',
+    'LiteLLM Proxy',
+    'socketio',
+    'socketio.client',
+    'socketio.server',
+]
+
+for logger_name in LOQUACIOUS_LOGGERS:
+    logging.getLogger(logger_name).setLevel('WARNING')
```
enyst, line 403: Agree, this is a good idea in general, but FYI litellm was disabled because it can leak API keys in logs, not only because it's a lot of output.

_Edited to add:_ on errors, IIRC. So I think it will (if it's still doing it) on warning too.

tofarr: Thank you - I was not aware of that - I have fully disabled the LiteLLM loggers again and added a note explaining why they were disabled

Issue #6252, opened 2025-01-13
Message timestamps in UI


neubig:
**What problem or use case are you trying to solve?**

It would be nice to know when commands are run to have an idea how long the process of solving an issue is taken.

**Describe the UX of the solution you'd like**

This could be done by adding unintrusive timestamps into the UI somewhere, so we can see when a particular command is executed.

This is not technically difficult, but designing it in a way that is visually appealing might be the hard part.

### Desired Behavior

Chat messages and rendered events should display their execution timestamp in an unobtrusive, visually polished way that follows the linked Figma design as closely as possible. Timestamps should be easy to discover when reviewing a conversation without distracting from the message content, and should remain consistent across the supported user, assistant, and command/event message views.

### Acceptance Criteria

- [ ] Render the timestamp for each applicable chat message or event using the event timestamp already recorded by the application.
- [ ] Match the linked Figma design as closely as possible, including placement, typography, spacing, color, visibility behavior, and hover/focus interaction.
- [ ] Keep timestamps unobtrusive and ensure they do not overlap, obscure, or change the layout of message content at supported viewport sizes.
- [ ] Cover the supported message variants (including user, assistant, and command/event messages) and add or update automated tests for the timestamp rendering behavior.
- [ ] In the pull request, upload exactly two screenshots: one showing the Figma design/reference and one showing the implemented UI, clearly labeled for comparison.

_This issue update was created by an AI agent (OpenHands) on behalf of the user._

dodo-perth:
Hello, 

Can I work on this issue?

Thanks :)

neubig:
Hi @dodo-perth , would you be interested in this still?

dodo-perth:
Hello,

I will make Pull Request as soon as possible.

I tried to make some timestamps, I couldn't find a best place to put it for user experience and I was worried about mistake as it's my first pull request. 

Thank you 🙂

sarans-h:
Is it still open. If yes please assign me.

enyst:
@sarans-h Thank you! Just so you know, you don't need assignment to work on an issue, you are most welcome to look into it!

Are you using `openhands` for your projects?

sarans-h:
> [@sarans-h](https://github.com/sarans-h) Thank you! Just so you know, you don't need assignment to work on an issue, you are most welcome to look into it!
> 
> Are you using `openhands` for your projects?

Hi @enyst  so can I work on this ? and make a pr request??

neubig:
[...]

PR #8836, opened 2025-06-01
Add timestamp tooltip on chat messages

lowlyocean:
Closes https://github.com/All-Hands-AI/OpenHands/issues/6252


amanape (changes requested):
All event's return a `timestamp` property that we can pass as a prop to this component instead.

With your solution, loading old conversations will have their timestamp set to now, which is incorrect.

neubig:
Hi @lowlyocean , are you still interested in working on this? If so we can continue, if not then we can close it and consider trying it again later. Thanks for the contribution either way!

PR #8840, opened 2025-06-02
Add INIT_PLUGIN_TIMEOUT env

SmartManoj:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**
Like https://github.com/All-Hands-AI/OpenHands/pull/7443

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

---
**Link of any specific issues this addresses:**
https://github.com/SmartManoj/Kevin/issues/393#issuecomment-2915891478

PR #8846, opened 2025-06-02
Update action_execution_server.py timeout

iSevenDays:
All plugins, including Jupyter and VSCode, initialized. This happened over a period of about 1 minute and 15 seconds (from 13:07:02 to 13:08:17), which is longer than the original 60-second hardcoded timeout, but within the new 180-second timeout.

See log below:

   13:07:02 - openhands:DEBUG: action_execution_server.py:283 - Browser initialization started in background
    13:07:02 - openhands:DEBUG: action_execution_server.py:285 - PATCHED: Setting plugin init timeout to 180.0
    ...
    13:07:13 - openhands:INFO: browser_env.py:116 - Browser env started.
    ...
    13:07:13 - openhands:DEBUG: action_execution_server.py:318 - Initializing plugin: agent_skills
    ...
    13:07:16 - openhands:DEBUG: __init__.py:133 - Jupyter kernel gateway started at port 49970.
    ...
    13:07:18 - openhands:DEBUG: action_execution_server.py:318 - Initializing plugin: jupyter
    ...
    13:08:17 - openhands:DEBUG: __init__.py:71 - VSCode server started at port 41590.
    ...
    13:08:17 - openhands:DEBUG: action_execution_server.py:318 - Initializing plugin: vscode
    13:08:17 - openhands:DEBUG: action_execution_server.py:290 - All plugins initialized

---

The PR changes the default timeout of 60 seconds to 180 seconds to allow plugins to initialize. 60 seconds was always not enough for my case and the runtime was constantly crashing because it couldn't complete the initialization within 60 seconds.

---


openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -424,7 +424,7 @@ def _attach_to_container(self):
         )
 
     @tenacity.retry(
-        stop=tenacity.stop_after_delay(120) | stop_if_should_exit(),
+        stop=tenacity.stop_after_delay(240) | stop_if_should_exit(),
```
iSevenDays, line 440: I have launched some tests and found out that timeout of 120 seconds was often not enough:
The initial API call was around 13:55:45.
The wait_until_alive in the main server starts polling.
The "Connection reset by peer" error is logged by the main server at 13:57:52.
The duration between the API call and this error is approximately 2 minutes and 7 seconds (127 seconds).
This 127-second duration aligns very closely with the stop_after_delay(120) in wait_until_alive.

The wait_until_alive function is retrying on the "Connection reset by peer" errors (as they are httpx.ReadError, a type of httpx.NetworkError). However, the runtime's Action Execution Server takes just over 2 minutes (approx. 123 seconds from when its HTTP server starts) to become fully stable and stop resetting connections on its /alive endpoint. The main server's wait_until_alive only retries for 120 seconds.

xingyaoww:
@OpenHands we added `INIT_PLUGIN_TIMEOUT` env var in https://github.com/All-Hands-AI/OpenHands/pull/8840. Please set the default value of INIT_PLUGIN_TIMEOUT to the proposed value in this PR

SmartManoj:
> `int(os.environ.get('INIT_PLUGIN_TIMEOUT', '240'))`

@OpenHands, it should be `int(os.environ.get('INIT_PLUGIN_TIMEOUT', '180')) + 60`

iSevenDays:
Updated to

> int(os.environ.get('INIT_PLUGIN_TIMEOUT', '180')) + 60

openhands/runtime/impl/docker/docker_runtime.py
SmartManoj:
```suggestion
        int(os.environ.get('INIT_PLUGIN_TIMEOUT', '180')) + 60,
```

openhands/runtime/action_execution_server.py
```diff
@@ -284,7 +284,7 @@ async def ainit(self):
 
         await wait_all(
             (self._init_plugin(plugin) for plugin in self.plugins_to_load),
-            timeout=int(os.environ.get('INIT_PLUGIN_TIMEOUT', '120')),
+            timeout=int(os.environ.get('INIT_PLUGIN_TIMEOUT', '180')) + 60,
```
SmartManoj:
```suggestion
            timeout=int(os.environ.get('INIT_PLUGIN_TIMEOUT', '120')),
```

xingyaoww:
```suggestion
            timeout=int(os.environ.get('INIT_PLUGIN_TIMEOUT', '180')),
```

xingyaoww (commented):
I think this is probably safe enough.
The timeout added is in the docker side which won't affect the other runtime. And bumping up init plugin timeout seems reasonable to me (especially in machine with limited resource).

cc @rbren since i know you are not a fan of large timeout 😄

neubig:
Hey @rbren quick ping on this to make an executive decision about whether we want to include this. Once that's done we can work on getting it merged.

mamoodi:
@neubig I'm not sure if Robert is still the best person for this decision. Any ideas what we should be doing with this PR? Whether to approve it or just leave it as is with the current behavior?

PR #8854, opened 2025-06-02
Fix passing of Git credentials to Docker Runtime

### LINKED ISSUE #8853: [Bug]: GITHUB_TOKEN is not passed to Docker Runtime

Default clone tries to clone .openhands repo and doesn't use integrated GITHUB_TOKEN.

### OpenHands Installation

Docker command in README

### OpenHands Version

0.40

### Operating System

None

```
16:35:37 - openhands:DEBUG: bash.py:601 - END OF PANE CONTENT: ['  "pid": "",', '  "exit_code": "0",', '  "username": "root",', '  "hostname": "c137670695ae",', '  "working_dir": "/workspace/swift_example",', '  "py_interpreter_path": "/openhands/poetry/openhands-ai-5O4_aCHf-py3.12/bin/python"', '}', '###PS1END###', "git clone --depth 1 https://github.com/username/.openhands.git /workspace/org_openhands_username2>/dev/null || echo 'Org repo not found'", "Username for 'https://github.com/':"]
```

iSevenDays:

Fix passing of Git credentials to Docker Runtime

---

Fix passing of Git credentials to Docker Runtime
---
Closes #8853


tofarr:
So we might need to think a bit about this one - The code is a little bit confusing due to the change in direction between regular runtimes and nested runtimes. (We will be adding subsequent PRs to clean this up once the functionality is bedded in). The workflow should be:

* The central server starts docker container with a `session_api_key` - this key is used to secure all further communication
* The docker container starts with a fully functional openhands instance inside it running the LocalRuntime.
* The central server does not have direct access to the action execution server, so the previous operations for setting environment variables after the server has started do not work. ([Hence temporary abberations like this](https://github.com/All-Hands-AI/OpenHands/blob/aac2e51fc7ac71236c2d9bcc62cf3e9eaa3ae025/openhands/server/conversation_manager/docker_nested_conversation_manager.py#L460))
* After the container has started, [we call the standard endpoints to set up user settings, secrets, and create a conversation in the nested container](https://github.com/All-Hands-AI/OpenHands/blob/aac2e51fc7ac71236c2d9bcc62cf3e9eaa3ae025/openhands/server/conversation_manager/docker_nested_conversation_manager.py#L200)

I have tested this with my github credentials, and it seems to be working for me - though I need to check custom secrets. If there is an error setting github credentials or secrets, I suspect it is in this config section.

The long term upside of this change is that it will allow us to share / reuse containers more efficiently - (Even now, nothing stops you from tweaking the frontend code and running multiple conversations in a single container)

tofarr (changes requested):
See my previous comment

iSevenDays:
My main concern was this long line

```
16:35:37 - openhands:DEBUG: bash.py:601 - END OF PANE CONTENT: ['  "pid": "",', '  "exit_code": "0",', '  "username": "root",', '  "hostname": "c137670695ae",', '  "working_dir": "/workspace/swift_example",', '  "py_interpreter_path": "/openhands/poetry/openhands-ai-5O4_aCHf-py3.12/bin/python"', '}', '###PS1END###', "git clone --depth 1 https://github.com/username/.openhands.git /workspace/org_openhands_username2>/dev/null || echo 'Org repo not found'", "Username for 'https://github.com/':"]
```

> "Username for 'https://github.com/':"

somehow, the credentials are not passed

tofarr:
Lets confirm we are referring to the same thing.

Are you using the nested runtime? This feature is in active development and is not yet enabled by default - Have you have set the environment variable: `CONVERSATION_MANAGER_CLASS="openhands.server.conversation_manager.docker_nested_conversation_manager.DockerNestedConversationManager"`

If this is not set, you are using the StandaloneConversationManager, and the error message looks like a more general authentication error. In this case, can you confirm that your github token is valid?

iSevenDays:
> Are you using the nested runtime?

No.

> If this is not set, you are using the StandaloneConversationManager, and the error message looks like a more general authentication error. In this case, can you confirm that your github token is valid?

Github token is valid.

When cloning .openhands repo from org, the GITHUB_TOKEN is not used for some reason.
Before this step, when cloning the repo, the token was used.
Example of API is here https://github.com/All-Hands-AI/OpenHands/issues/8859

tofarr:
> > Are you using the nested runtime?
> 
> No.
> 
> > If this is not set, you are using the StandaloneConversationManager, and the error message looks like a more general authentication error. In this case, can you confirm that your github token is valid?
> 
> Github token is valid.
> 
> When cloning .openhands repo from org, the GITHUB_TOKEN is not used for some reason. Before this step, when cloning the repo, the token was used. Example of API is here #8859

ok - then the code in the nested manager is almost certainly not the culprit. I think last week there was an issue with tokens / permissions for the openhands repository on github which may be related

mamoodi:
Is this PR still valid based on the last comment?

PR #8861, opened 2025-06-02
Allow user to change their email address

chuckbutkus:
For the SAAS product, we want to allow users to change their email address.  This requires that the user verify this email address.  A user tab is added to the settings page and the user is allowed to change it.  If the user changes the email address then a verify email is sent.  The user will only be able to visit the user settings tab until the email is verified.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f815af6-nikolaik   --name openhands-app-f815af6   docker.all-hands.dev/all-hands-ai/openhands:f815af6
```


frontend/src/components/features/sidebar/sidebar.tsx
```diff
@@ -69,16 +69,21 @@ export function Sidebar() {
...
+              }
+              disabled={settings?.EMAIL_VERIFIED === false}
             />
           </div>
 
           <div className="flex flex-row md:flex-col md:items-center gap-[26px] md:mb-4">
-            <DocsButton />
-            <SettingsButton />
+            <DocsButton disabled={settings?.EMAIL_VERIFIED === false} />
+            <SettingsButton disabled={settings?.EMAIL_VERIFIED === false} />
```
amanape, line 86: This implies that the user won't be able to access settings if they aren't verified. Is that expected?

frontend/src/components/shared/buttons/tooltip-button.tsx
```diff
@@ -63,7 +70,24 @@ export function TooltipButton({
         {children}
       </NavLink>
     );
-  } else if (href) {
+  } else if (navLinkTo && disabled) {
+    // If disabled and has navLinkTo, render a button that looks like a NavLink but doesn't navigate
```
amanape, line 74: Why not just pass the `disabled` prop to `NavLink` or `a` in the case of the button below?

frontend/src/routes/user-settings.tsx
amanape, line 1: This could benefit from some refactoring - at least the HTML into smaller components

frontend/src/components/features/sidebar/sidebar.tsx
chuckbutkus: If they aren't verified, then they are forced to the /settings/user page and can't navigate anywhere else.  So yes, this is the expected behavior.

frontend/src/components/shared/buttons/tooltip-button.tsx
chuckbutkus: Do `NavLink` and `a` support a disabled attribute?  I don't see that they do.

amanape: You're right. Maybe we can just set the `to` or `href` property to `undefined` or `null` instead of a new element entirely

PR #8868, opened 2025-06-03
Fix: Continue evaluation when an instance fails after max retries

juanmichelini:
## Description

For debugging purposes, when working with benchmarks you can set `export EVAL_SKIP_BUILD_ERRORS=true` to continue evaluation even when instances reach maximum build retries.

This change removes a bottleneck when trying new benchmarks since it allows to partially run benchmarks with faulty docker images, rather than having the entire evaluation fail because of a single problematic instance.

After evaluation completes, check `build_error_maximum_retries_exceeded.jsonl` for a list of failed instances, fix those images, and then run the evaluation again with `export EVAL_SKIP_BUILD_ERRORS=false`.

## Changes

Modifies `evaluation/utils/shared.py` to check for env var  EVAL_SKIP_BUILD_ERRORS, if it is set to true it skips instances that exceed max retries and logs them to build_error_maximum_retries_exceeded.jsonl.

Modified swe_bench/run_infer.py and mult_swe_bench/run_infer.py to check if build_error_maximum_retries_exceeded.jsonl exists and log to console.

Added doc in swe_bench's and multi_swe_bench's READMEs.


xingyaoww (commented):
Ideally we should print a report at the end of run_infer.py script to warn the user that there are X eval instances that run into max retry errors -- so we are not accidentally think we finished every eval when there's an infra failure (eg. runtime-api is suddenly down and all the eval failed)

juanmichelini:
I fixed the failing CI checks by removing the undefined status field from the EvalOutput class in the log_skipped_build_maximum_retries_exceeded function.

juanmichelini:
I have fixed the failing CI checks by removing the undefined `status` field from the `EvalOutput` class in the `log_skipped_build_maximum_retries_exceeded` function. The `status` field was not defined in the `EvalOutput` class, which was causing the CI to fail. I have also added a test file to trigger the CI.

juanmichelini:
@xingyaoww Yes! What do you think of the current commit? There is a flag that can be set to continue evaluating when max retries exceed. Skipped instances are logged to a jsonl. Documentation and log messages remind that this is for debugging purposes only.

evaluation/benchmarks/swe_bench/README.md
```diff
@@ -36,6 +36,10 @@ Please follow instruction [here](../../README.md#setup) to setup your local deve
 > - If your LLM config has temperature=0, we will automatically use temperature=0.1 for the 2nd and 3rd attempts
 >
 > To enable this iterative protocol, set `export ITERATIVE_EVAL_MODE=true`
+>
+> **Skipping errors on build**
+>
+> For debugging purposes, you can set `export EVAL_SKIP_BUILD_ERRORS=true` to continue evaluation even when instances reach maximum build retries. After evaluation completes, check `build_error_maximum_retries_exceeded.txt` for a list of failed instances, fix those images, and then run the evaluation again with `export EVAL_SKIP_BUILD_ERRORS=false`.
```
xingyaoww:
```suggestion
> For debugging purposes, you can set `export EVAL_SKIP_BUILD_ERRORS=true` to continue evaluation even when instances reach maximum build retries. After evaluation completes, check `build_error_maximum_retries_exceeded.jsonl` for a list of failed instances, fix those images, and then run the evaluation again with `export EVAL_SKIP_BUILD_ERRORS=false`.
```

This this correct?

evaluation/benchmarks/multi_swe_bench/README.md
```diff
@@ -41,6 +41,10 @@ default, it is set to 1.
 - `language`, the language of your evaluating dataset.
 - `dataset`, the absolute position of the dataset jsonl.
 
+**Skipping errors on build**
+
+For debugging purposes, you can set `export EVAL_SKIP_BUILD_ERRORS=true` to continue evaluation even when instances reach maximum build retries. After evaluation completes, check `build_error_maximum_retries_exceeded.txt` for a list of failed instances, fix those images, and then run the evaluation again with `export EVAL_SKIP_BUILD_ERRORS=false`.
```
xingyaoww:
```suggestion
For debugging purposes, you can set `export EVAL_SKIP_BUILD_ERRORS=true` to continue evaluation even when instances reach maximum build retries. After evaluation completes, check `build_error_maximum_retries_exceeded.jsonl` for a list of failed instances, fix those images, and then run the evaluation again with `export EVAL_SKIP_BUILD_ERRORS=false`.
```

evaluation/utils/shared.py
```diff
@@ -300,6 +300,80 @@ def assert_and_raise(condition: bool, msg: str):
         raise EvalException(msg)
 
 
+def log_skipped_build_maximum_retries_exceeded(
```
xingyaoww: nit: terminology-wise, i think it is technically not "build" maximum retry? It can also happen when runtime consistently failed for 5 times for that instance..

juanmichelini: so a better named would be log_skipped_maximum_retries_exceeded without build?

xingyaoww: yep!

juanmichelini: changed all mentions of skip_builds to skip_maximum_retries !

juanmichelini: @xingyaoww, can you merge it?

xingyaoww (approved):
LGTM

PR #8881, opened 2025-06-03
Fix VSCode iframe SameSite cookie issue with cross-origin fallback

tofarr:

This change fixes an issue with the VSCode iframe in the OpenHands UI where cookies with "SameSite=Lax" attribute were being blocked by browser security policies when loading a secure frame inside an insecure frame. Users will now see a clear message and an "Open in New Tab" button when this is detected, ensuring they can still access the VSCode functionality even when browser security restrictions apply.

---

For example, when you use a remote runtime...

**Before**
image

**After**
image

This PR implements a solution to address the SameSite cookie issue with the VSCode iframe:

1. Adds detection of cross protocol URLs in the VSCode tab component
2. Implements a fallback UI that shows a warning message and "Open in New Tab" button when cross-origin content is detected
3. Adds new translation strings for the cross-origin warning message and button

The key design decision was to provide a graceful fallback when browser security policies block cookies in cross protocol iframes. Opening the content in a new tab works because SameSite=Lax cookies are allowed in top-level navigations but are blocked in iframes when the iframe origin does not match the parent page origin.

**Testing**
* Tested with DockerRuntime (Works as at present)
* Tested with RemoteRuntime (See screenshots)
* I tried using regular sandbox attributes, `sandbox="allow-scripts allow-same-origin allow-forms allow-popups allow-downloads"` but was unsuccessful.
---
N/A

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0eb0df4-nikolaik   --name openhands-app-0eb0df4   docker.all-hands.dev/all-hands-ai/openhands:0eb0df4
```


rbren (changes requested):
I think we need to try and open the crossorigin iframe before bailing out...

tofarr:
> I think we need to try and open the crossorigin iframe before bailing out...

I looked into this further - The issue arises when we try to access a secure domain (https) from an insecure one (http).

I updated the check for this. So currently this will only show the button when you are on localhost and access a runtime that is secured. (Development, or an OSS user using our runtime)

PR #8889, opened 2025-06-04
Increase sandbox close delay in sandbox_config.py to 3600 seconds

iSevenDays:

---
Resolves https://openhands-ai.slack.com/archives/C07KHERRM2S/p1748994579396239

---

### DIFF
--- a/openhands/core/config/sandbox_config.py
+++ b/openhands/core/config/sandbox_config.py
@@ -73,7 +73,10 @@ class SandboxConfig(BaseModel):
     runtime_startup_env_vars: dict[str, str] = Field(default_factory=dict)
     browsergym_eval_env: str | None = Field(default=None)
     platform: str | None = Field(default=None)
-    close_delay: int = Field(default=15)
+    close_delay: int = Field(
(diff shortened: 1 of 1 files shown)


openhands/core/config/sandbox_config.py
```diff
@@ -73,7 +73,10 @@ class SandboxConfig(BaseModel):
     runtime_startup_env_vars: dict[str, str] = Field(default_factory=dict)
     browsergym_eval_env: str | None = Field(default=None)
     platform: str | None = Field(default=None)
-    close_delay: int = Field(default=15)
+    close_delay: int = Field(
+        default=int(os.environ.get('SANDBOX_CLOSE_DELAY_SECONDS', '3600')),
+        description='The delay in seconds before closing the sandbox after the agent is done. Can be configured via SANDBOX_CLOSE_DELAY_SECONDS env var.',
```
rbren: we already override these with env vars, no need to introduce a new one here
```suggestion
        default=3600,
        description='The delay in seconds before closing the sandbox after the agent is done.',
```

rbren (approved):

enyst:
Thank you for this! TBH I saw the former value as a really bad default, and I didn't quite understand why we did this in the first place.

PR #8893, opened 2025-06-04
Fix action suggestions to use git_provider from conversation metadata

rbren:
## Description

This PR fixes an issue with the action suggestions component in the chat interface. Previously, the component was showing GitLab options for ALL users that have connected to GitLab, even if they are connected to both GitHub and GitLab and are currently using a GitHub repository.

The fix changes the logic to use the `git_provider` field from the conversation metadata instead of just checking if the user has connected to GitLab. This ensures that the correct options (GitHub or GitLab) are shown based on the currently selected repository.

## Changes

- Modified `frontend/src/components/features/chat/action-suggestions.tsx` to use `conversation?.git_provider === "gitlab"` instead of `providers.includes("gitlab")` to determine whether to show GitLab or GitHub options
- Updated the tests in `frontend/__tests__/components/chat/action-suggestions.test.tsx` to reflect this change:
  - Updated the existing GitHub test to include the `git_provider: "github"` in the mock conversation
  - Added a new test for GitLab to verify that GitLab options are shown when the conversation has `git_provider: "gitlab"`

## Testing

- Ran the tests for the action-suggestions component and verified they pass
- Manually verified that the correct options are shown based on the git provider of the selected repository

Robert Brennan can click here to [continue refining the PR](https://staging.all-hands.dev/conversations/{})

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:317168f-nikolaik   --name openhands-app-317168f   docker.all-hands.dev/all-hands-ai/openhands:317168f
```

### DIFF
--- a/frontend/__tests__/components/chat/action-suggestions.test.tsx
+++ b/frontend/__tests__/components/chat/action-suggestions.test.tsx
@@ -70,11 +70,12 @@ describe("ActionSuggestions", () => {
     });
   });
 
-  it("should render both GitHub buttons when GitHub token is set and repository is selected", async () => {
+  it("should render both GitHub buttons when GitHub token is set and repository is selected with GitHub provider", async () => {
     const getConversationSpy = vi.spyOn(OpenHands, "getConversation");
     // @ts-expect-error - only required for testing
     getConversationSpy.mockResolvedValue({
       selected_repository: "test-repo",
+      git_provider: "github"
     });
     renderActionSuggestions();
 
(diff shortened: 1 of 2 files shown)


frontend/__tests__/components/chat/action-suggestions.test.tsx
```diff
@@ -112,6 +113,42 @@ describe("ActionSuggestions", () => {
     expect(screen.queryByTestId("suggestion")).not.toBeInTheDocument();
   });
 
+  it("should render GitLab buttons when GitLab token is set and repository is selected with GitLab provider", async () => {
+    const getSettingsSpy = vi.spyOn(OpenHands, "getSettings");
+    getSettingsSpy.mockResolvedValue({
+      ...MOCK_DEFAULT_USER_SETTINGS,
+      provider_tokens_set: {
+        github: "some-token",
+        gitlab: "some-gitlab-token",
```
enyst, line 122: It seems that the test doesn't actually test for the change, it would pass before as well?

frontend/__tests__/components/chat/action-suggestions.test.tsx
```diff
@@ -112,6 +113,42 @@ describe("ActionSuggestions", () => {
...
+
+    // Check if we have at least 2 buttons
+    expect(buttons.length).toBeGreaterThanOrEqual(2);
+
+    // Check if the buttons contain the expected text for GitLab
+    const pushButton = buttons.find((button) =>
+      button.textContent?.includes("Push to Branch"),
+    );
+    const mrButton = buttons.find((button) =>
+      button.textContent?.includes("Push & Create PR"),
```
enyst, line 145: A little curious detail, I thought in GitLab, we'd say MR, not PR. Maybe I misremember?

frontend/__tests__/components/chat/action-suggestions.test.tsx
amanape: This looks like a new test, and I don't think it is required to mock `getSettings` here since we now get the property from `getConversation`

frontend/__tests__/components/chat/action-suggestions.test.tsx
rbren: Yup this looks wrong

PR #8894, opened 2025-06-04
(Hotfix): make sure MCP tool error observations are surfaced to agent

malhotra5:

- Ensure agent gets error observations for failed MCP tool calls in OpenHands MCP server

---

Currently, because of the way we want handles in the git MCP server we have two problems

1. We return a normal response for errors, which causes the UI to display tool calls with errors as successful
2. The agent gets vague error messages, like `Unknown error` instead of the actual message

This PR ensures that 

1. Exceptions in tool calls aren't sent to the MCP client by default
2. Only errors which are wrapped in `ToolError` are plumbed back to the MCP client (overrides exception handling this way)
3. Git service passes readable error messages to Git MCP server

QA for UI (isError attrib is fixed)

image
image

QA where error observation is plumbed to MCP client

image

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c3e16c4-nikolaik   --name openhands-app-c3e16c4   docker.all-hands.dev/all-hands-ai/openhands:c3e16c4
```

### DIFF
--- a/openhands/integrations/github/github_service.py
+++ b/openhands/integrations/github/github_service.py
@@ -483,35 +483,33 @@ class GitHubService(BaseGitService, GitService):
             - PR URL when successful
             - Error message when unsuccessful
         """
-        try:
-            url = f'{self.BASE_URL}/repos/{repo_name}/pulls'
 
-            # Set default body if none provided
-            if not body:
-                body = f'Merging changes from {source_branch} into {target_branch}'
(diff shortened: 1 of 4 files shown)


tofarr (approved):
🍰

PR #8899, opened 2025-06-04
Assorted fixes for the nested / docker runtimes.

tofarr:

This PR improves the stability and reliability of nested and Docker runtimes in OpenHands, particularly when handling container lifecycle and conversation management. It fixes issues with container reuse, file paths, and error handling that could cause problems when starting, stopping, or restarting conversations.

---
1. Improved container management:
   - Instead of automatically removing and recreating containers when they already exist, the system now checks if a container exists and starts it if it is in an exited state
   - Removed unnecessary container recreation logic that could cause instability

2. Fixed file path for nested conversations:
   - Updated the volume mount path from `/root/openhands/file_store/` to `/root/.openhands/file_store/` to ensure proper file storage and access

3. Improved error handling:
   - Changed `remove()` to `discard()` for the `_starting_conversation_ids` set to prevent KeyError exceptions
   - Simplified error logging in the Docker runtime

4. Added configuration for sandbox close delay:
   - Added a check to skip cleanup of disconnected conversations if `close_delay` is 0 or None
   - Set `SANDBOX_CLOSE_DELAY` to "0" for nested containers to prevent premature cleanup

5. Code simplification:
   - Simplified user ID retrieval by using the direct `get_user_id()` function

---

### DIFF
--- a/openhands/server/conversation_manager/docker_nested_conversation_manager.py
+++ b/openhands/server/conversation_manager/docker_nested_conversation_manager.py
@@ -54,6 +54,7 @@ class DockerNestedConversationManager(ConversationManager):
     docker_client: docker.DockerClient = field(default_factory=docker.from_env)
     _conversation_store_class: type[ConversationStore] | None = None
     _starting_conversation_ids: set[str] = field(default_factory=set)
+    _runtime_container_image: str | None = None
 
     async def __aenter__(self):
         # No action is required on startup for this implementation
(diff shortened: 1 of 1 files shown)


openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -374,28 +374,11 @@ def init_container(self) -> None:
...
-                stop_all_containers(self.container_name)
-                return self.init_container()
-
-            else:
-                self.log(
-                    'error',
-                    f'Error: Instance {self.container_name} FAILED to start container!\n',
-                )
-                self.log('error', str(e))
-                raise e
```
tofarr, line 392: I removed this because it no longer makes sense. At some point in the past, `stop_all_containers` would delete matching docker containers rather than just stop them. As it stands, this results in an infinite loop where the system stops an already stopped container and then tries to recreate it. (This was not being caught only because `maybe_start_agent_loop` checks that the container exists before this is called)

PR #8908, opened 2025-06-05
Fix: Bug of transform user message to tool message when user message's content is list and length is greater than one in convert_non_fncall_messages_to_fncall_messages method.

LCHONE:
# Description
Logic of transform user message to tool message in convert_non_fncall_messages_to_fncall_messages method does not take into account the situation where the content of the message is list and its length is greater than one. The conversion of message formats should not limit the length of the content array, but should ensure the reversibility of the conversion.

# Changes
Relax the length limit in the transformation method of the message format.


openhands/llm/fn_call_converter.py
```diff
@@ -736,25 +737,26 @@ def convert_non_fncall_messages_to_fncall_messages(
                     isinstance(content, str)
                     or (
                         isinstance(content, list)
-                        and len(content) == 1
-                        and content[0].get('type') == 'text'
+                        and first_tool_call_content
                     )
                 ):
                     raise FunctionCallConversionError(
                         f'Expected str or list with one text item when tool result is present in the message. Content: {content}'
```
xingyaoww, line 807: I think with this change, `list with one text item` won't be accurate here, right?

LCHONE: Yes, there is no need to limit the length here. If the length needs to be restricted, it should be in the upstream logic part.

mamoodi: CC @xingyaoww

openhands/llm/fn_call_converter.py
```diff
@@ -798,22 +800,23 @@ def convert_non_fncall_messages_to_fncall_messages(
                     isinstance(content, str)
                     or (
                         isinstance(content, list)
-                        and len(content) == 1
-                        and content[0].get('type') == 'text'
+                        and first_tool_call_content
```
xingyaoww, line 803: first_tool_call_content is not defined before, i imagine we will run into issue for this?

i.e., shall we define `first_tool_call_content = None` before enter into this if-else clauses

LCHONE: first_tool_call_content has been inited in last if-else.

xingyaoww:
@LCHONE can you resolve the merge conflict here to reflect your true intention though?

PR #8911, opened 2025-06-05
Add stricter type checking for conversation manager and utils

raymyers:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

This configures mypy to require type annotations including return types on a subset of files in `openhands.server`. Requiring these should catching missing None checks such as this recent runtime exception:
```
openhands/server/conversation_manager/standalone_conversation_manager.py", line 136, in detach_from_conversation
    sid = conversation.sid
          ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'sid'
```

Eventually, everything in openhands.server and anything we consider an integration point should probably have this enabled, but starting with a small change.
---
**Link of any specific issues this addresses:**

PR #8915, opened 2025-06-05
Fix NoneType error in get_conversation and other typesafety

raymyers:

---

Adding the type annotation to conversation_manager in shared.py makes it so mypy catches the missing None check behind this recent runtime exception:

```
openhands/server/conversation_manager/standalone_conversation_manager.py", line 136, in detach_from_conversation
    sid = conversation.sid
          ^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'sid'
```

It also appears to have flagged a missing async. It's possible that these constants have left a larger hole in our type-checking. I originally tried to solve this by forcing return type annotations in #8911 (maybe still a good idea), but this turned out to me enough.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:d2c5c00-nikolaik   --name openhands-app-d2c5c00   docker.all-hands.dev/all-hands-ai/openhands:d2c5c00
```

### DIFF
--- a/openhands/server/listen_socket.py
+++ b/openhands/server/listen_socket.py
@@ -132,20 +132,21 @@ async def connect(connection_id: str, environ: dict) -> None:
         agent_state_changed = None
         if agent_loop_info is None:
             raise ConnectionRefusedError('Failed to join conversation')
-        async_store = AsyncEventStoreWrapper(
-            agent_loop_info.event_store, latest_event_id + 1
-        )
-        async for event in async_store:
-            logger.debug(f'oh_event: {event.__class__.__name__}')
-            if isinstance(
-                event,
-                (NullAction, NullObservation, RecallAction),
-            ):
-                continue
-            elif isinstance(event, AgentStateChangedObservation):
-                agent_state_changed = event
-            else:
-                await sio.emit('oh_event', event_to_dict(event), to=connection_id)
+        if agent_loop_info.event_store:
+            async_store = AsyncEventStoreWrapper(
+                agent_loop_info.event_store, latest_event_id + 1
+            )
+            async for event in async_store:
+                logger.debug(f'oh_event: {event.__class__.__name__}')
+                if isinstance(
+                    event,
+                    (NullAction, NullObservation, RecallAction),
+                ):
+                    continue
+                elif isinstance(event, AgentStateChangedObservation):
+                    agent_state_changed = event
+                else:
+                    await sio.emit('oh_event', event_to_dict(event), to=connection_id)
         if agent_state_changed:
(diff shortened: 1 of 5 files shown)


openhands/server/listen_socket.py
```diff
@@ -132,20 +132,21 @@ async def connect(connection_id: str, environ: dict) -> None:
...
-            if isinstance(
-                event,
-                (NullAction, NullObservation, RecallAction),
-            ):
-                continue
-            elif isinstance(event, AgentStateChangedObservation):
-                agent_state_changed = event
-            else:
-                await sio.emit('oh_event', event_to_dict(event), to=connection_id)
+        if agent_loop_info.event_store:
```
neubig, line 135: This will silently do nothing if `agent_loop_info.event_store` is `None`, is that the proper behavior? Or is the proper behavior to throw an error?

raymyers: That's a good question, I don't know enough about the scenario it could be None in. My thinking was "Well we're looking over all the events so if there's no store there are definitely no events to loop over".

But I can equally argue that AsyncEventStoreWrapper would have exploded before so we should preserve the behavior unless we know we want different.

openhands/server/utils.py
```diff
@@ -27,4 +27,5 @@ async def get_conversation(
     try:
         yield conversation
     finally:
-        await conversation_manager.detach_from_conversation(conversation)
+        if conversation:
+            await conversation_manager.detach_from_conversation(conversation)
```
raymyers, line 31: This is the one from the prod error.

openhands/server/routes/conversation.py
```diff
@@ -145,5 +145,5 @@ async def search_events(
 @app.post('/events')
 async def add_event(request: Request, conversation: ServerConversation = Depends(get_conversation)):
     data = request.json()
-    conversation_manager.send_to_event_stream(conversation.sid, data)
+    await conversation_manager.send_to_event_stream(conversation.sid, data)
     return JSONResponse({'success': True})
```
raymyers, line 149: Probably this was a hidden performance bug.

openhands/server/listen_socket.py
raymyers: @tofarr @malhotra5 Would you know more about the desired behavior here?

raymyers:
Partially done in #8927, going to do the rest later.

PR #8921, opened 2025-06-05
Improve type coverage for nested runtime

raymyers:

---

Running a mypy html report, I found that we have lots of "Imprecision", things mypy cannot check. Adding more dependencies in the pre-commit config and adding type annotations on some untyped globals helps address this.

This PR cuts docker_nested_conversation_manager's imprecision (unchecked lines) in half from roughly 23% to 12%.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:d0ffa2b-nikolaik   --name openhands-app-d0ffa2b   docker.all-hands.dev/all-hands-ai/openhands:d0ffa2b
```

### DIFF
--- a/openhands/server/conversation_manager/docker_nested_conversation_manager.py
+++ b/openhands/server/conversation_manager/docker_nested_conversation_manager.py
@@ -90,7 +90,8 @@ class DockerNestedConversationManager(ConversationManager):
         """
         Get the running agent loops directly from docker.
         """
-        names = (container.name for container in self.docker_client.containers.list())
+        containers : list[Container]  = self.docker_client.containers.list()
+        names = (container.name or '' for container in containers)
         conversation_ids = {
             name[len('openhands-runtime-') :]
             for name in names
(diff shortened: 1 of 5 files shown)


openhands/server/routes/manage_conversations.py
```diff
@@ -83,7 +83,7 @@ async def new_conversation(
     provider_tokens: PROVIDER_TOKEN_TYPE = Depends(get_provider_tokens),
     user_secrets: UserSecrets = Depends(get_user_secrets),
     auth_type: AuthType | None = Depends(get_auth_type),
-) -> ConversationResponse:
+) -> ConversationResponse | JSONResponse:
```
raymyers: Is there an issue changing these types because of some autogen Swagger doc, by chance?

openhands/server/conversation_manager/docker_nested_conversation_manager.py
```diff
@@ -480,7 +482,7 @@ async def _start_existing_container(self, runtime: DockerRuntime) -> bool:
             if container:
                 status = container.status
                 if status == 'exited':
-                    await call_sync_from_async(container.start())
+                    await call_sync_from_async(container.start)
```
raymyers, line 485: This one looks like it was a bug.

openhands/server/routes/manage_conversations.py
raymyers: Ah, gets this error in tests.

```
E   fastapi.exceptions.FastAPIError: Invalid args for response field! Hint: check that openhands.server.routes.manage_conversations.ConversationResponse | starlette.responses.JSONResponse is a valid Pydantic field type. If you are using a return type annotation that is not a valid Pydantic field (e.g. Union[Response, dict, None]) you can disable generating the response model from the type annotation with the path operation decorator parameter response_model=None. Read more: https://fastapi.tiangolo.com/tutorial/response-model/
```

openhands/server/conversation_manager/docker_nested_conversation_manager.py
tofarr: Nice catch!

tofarr (approved):
🍰

PR #8945, opened 2025-06-06
feat(chat): support file upload

mislavlukach:

support both file and image uploads in conversation

---
This PR allows users to paste or select files (not just images). 
Upon selection, files can be visible seen in a list or in a chat message after the message has been sent. 
API has been extended with a creation of new route POST /upload-files which uploads these files into workspace.
If there's an error, API will return list of error which will be shown inside toast.

Chat message (prompt) has been augmented to provide additional context about these upload files. 
```
<original prompt>

NEW FILES ADDED:
/path1
/path2

```

In the rare cases where some file has been uploaded successfully and the other failed, only successful uploads will end up in augmented prompt. 

And finally, there are 2 remaining questions which need feedback:

1. FileWriteAction accepts file content as text. That's not really ideal. Currently we decode files in the fastapi endpoint by using UTF-8 and send it to FileWriteAction. As I said, this works really really good with text based files like json, ts, text, etc. But fails with binary files. Also, multiple options seems to be available:
-  not important for us. I'm assuming most of the users will be uploading text-based files which contains some code snippets.
- process binary files correctly but this also implies there's some kind of mechanism in the runtime which can read binary files and extract information. I don't think that exist.

2. file name collision system which adds "copy #num" doesn't exist anymore. That's because FileWriteAction will just override whatever is stored under the given path. I'm thinking we have 3 options.

- just don't important enough for us to care. New files will overwrite old files.
- update FileWriteAction to accept param which will determine if and how duplicates will be resolved
- create new Action which will check if some file exist without reading it. We have a FileReadAction but I don't think that is suitable for this case

---


frontend/src/api/open-hands.ts
```diff
@@ -373,6 +373,23 @@ class OpenHands {
...
+      formData.append("files", file);
+    }
+    const url = `${this.getConversationUrl(conversationId)}/upload-files`;
+    const response = await openHands.post<string[]>(url, formData, {
+      headers: {
+        "Content-Type": "multipart/form-data",
+      },
+    });
+    return response.data;
+  }
```
amanape: We are in the process of breaking down `open-hands.ts` into smaller services. As of now, we have [`file-service.api.ts`](https://github.com/All-Hands-AI/OpenHands/blob/main/frontend/src/api/file-service/file-service.api.ts) that we can move this code snippet to

frontend/src/components/features/chat/chat-interface.tsx
```diff
@@ -82,11 +88,22 @@ export function ChatInterface() {
...
 
     const timestamp = new Date().toISOString();
-    send(createChatMessage(content, imageUrls, timestamp));
+
+    const fileUrls =
+      files.length > 0
+        ? await OpenHands.uploadFiles(params.conversationId!, files)
+        : [];
+
+    const filePrompt = `${i18next.t("CHAT_INTERFACE$AUGMENTED_PROMPT_FILES_TITLE")}: ${fileUrls.join("\n\n")}`;
```
amanape: Instead of using `i18next`, we can use the translation hook. Example: https://github.com/All-Hands-AI/OpenHands/blob/ebacd1b0800e96c614b2132d87508e5a4d63ed9f/frontend/src/routes/llm-settings.tsx#L28

frontend/src/components/features/chat/chat-interface.tsx
```diff
@@ -82,11 +88,22 @@ export function ChatInterface() {
...
-    const promises = files.map((file) => convertImageToBase64(file));
+    const promises = images.map((image) => convertImageToBase64(image));
     const imageUrls = await Promise.all(promises);
 
     const timestamp = new Date().toISOString();
-    send(createChatMessage(content, imageUrls, timestamp));
+
+    const fileUrls =
+      files.length > 0
+        ? await OpenHands.uploadFiles(params.conversationId!, files)
```
amanape: What do you think about creating a custom hook `useUploadFiles` with `useMutation`? That we we might also benefit from some behaviour such as retry logic. You can see examples of mutation API [here](https://github.com/All-Hands-AI/OpenHands/tree/main/frontend/src/hooks/mutation)

frontend/src/components/features/files/file-list.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { cn } from "#/utils/utils";
+import { FileItem } from "./file-item";
+
+type FileItemType = {
+  filename: string;
+  id?: string;
```
amanape: Does `id` have to be optional here? I see you use the non-null assertion operator (!) below, so we know it always exists?

frontend/src/components/features/files/file-list.tsx
```diff
@@ -0,0 +1,30 @@
...
+  return (
+    <div
+      data-testid="file-list"
+      className={cn("flex flex-col gap-y-1.5 justify-start")}
+    >
+      {files.map((f, index) => (
+        <FileItem
+          key={index}
+          filename={f.filename}
+          onRemove={onRemove ? () => onRemove(f.id!) : undefined}
```
amanape:
```suggestion
          onRemove={onRemove?.(f.id!)}
```

frontend/src/components/features/images/image-carousel.tsx
```diff
@@ -4,10 +4,12 @@ import { ChevronRight } from "#/assets/chevron-right";
 import { ImagePreview } from "./image-preview";
 import { cn } from "#/utils/utils";
 
+type ImageFile = { src: string; id?: string };
```
amanape: Same comment about `id` as in `file-list.tsx`

frontend/src/components/features/images/image-carousel.tsx
```diff
@@ -55,12 +57,12 @@ export function ImageCarousel({
...
-        {images.map((src, index) => (
+        {images.map((image) => (
           <ImagePreview
-            key={index}
+            key={image.id}
             size={size}
-            src={src}
-            onRemove={onRemove && (() => onRemove(index))}
+            src={image.src}
+            onRemove={onRemove ? () => onRemove(image.id!) : undefined}
```
amanape:
```suggestion
            onRemove={onRemove?.()}
```

amanape (changes requested):
Good stuff. Let's get the FE review settled and then move on to the server-side of things.

The workspace lives in the runtime, so we're going to have to do something similar to how we handle listing files. See 
https://github.com/All-Hands-AI/OpenHands/blob/ebacd1b0800e96c614b2132d87508e5a4d63ed9f/openhands/server/routes/files.py#L83

https://github.com/All-Hands-AI/OpenHands/blob/ebacd1b0800e96c614b2132d87508e5a4d63ed9f/openhands/runtime/base.py#L905-L910

https://github.com/All-Hands-AI/OpenHands/blob/ebacd1b0800e96c614b2132d87508e5a4d63ed9f/openhands/runtime/impl/action_execution/action_execution_client.py#L135-L157

https://github.com/All-Hands-AI/OpenHands/blob/ebacd1b0800e96c614b2132d87508e5a4d63ed9f/openhands/runtime/action_execution_server.py#L1000-L1076

Server -> Runtime -> Runtime Server -> CRUD files

frontend/src/components/features/chat/chat-interface.tsx
mislavlukach: @amanape it was intentional from my end to avoid passing t function few levels deep into service function. 
I can still do it if you want, just let me know.

frontend/src/components/features/files/file-list.tsx
mislavlukach: @amanape  Unfortunately it does.  Interface of this component is not ideal. 
So we have 2 support two cases:

1. Before the file is uploaded (preview) - can be removed by clicking on X icon and has id property.
2. After file upload - can't be removed and doesn't have id.

frontend/src/components/features/images/image-carousel.tsx
mislavlukach: Same reply as above @amanape

frontend/src/api/open-hands.ts
mislavlukach: Thanks, done !

frontend/src/components/features/chat/chat-interface.tsx
mislavlukach: Ignore my comment, I've just realized Chat-interface is a React component. Don't know why I thought it's a service class ! Fixing now.

frontend/src/components/features/chat/chat-interface.tsx
mislavlukach: Done

frontend/src/components/features/files/file-list.tsx
mislavlukach: Thanks, it's done

mislavlukach:
@amanape I've responded on all FE comments. I think I get the backend comments, will do the changes and let you know when the PR is changed. Thanks for the feedback.

mislavlukach:
@amanape 
I think I understand what happens on the backend. 
So instead of manually processing file uploads in files.py, we're going to delegate it to the runtime. 

Next step would be to provide implementation. I can see there are few classes which uses Runtime
- cli runtime
- e2b runtime
- action_execution_runtime

I don't know what all of these are, but I'm assuming I only have to provide implementation in the action_execution_runtime?
I think I get the rest, real upload logic lives in the action_execution_server.py, everything else basically delegates this job.

amanape:
> So instead of manually processing file uploads in files.py, we're going to delegate it to the runtime.

Yes

> I don't know what all of these are, but I'm assuming I only have to provide implementation in the action_execution_runtime?

Yes, I don't think we need to modify the other runtime implementations other than the base which they extend. I will confirm later and let you know

mislavlukach:
@amanape latest update:

https://github.com/All-Hands-AI/OpenHands/pull/8945/commits/949a8f93ba7b11b56c96abd224d242770b1e1570

We now use FileWrite action to write to workspace. It works well. I'm using docker runtime locally and I was able to verify that by checking the files inside docker container. 
That's good news. Now the bad news. 😄 

1. file name collision system we had before - which adds "copy #num" doesn't exist anymore. That's because FileWriteAction will just override whatever is stored under the given path. I'm thinking we have 3 options. 

- just don't important enough for us to care. New files will overwrite old files. 
- update FileWriteAction to accept param which will determine if and how duplicates will be resolved
- create new Action which will check if some file exist without reading it. We have a FileReadAction but I don't think that is suitable for this case

2. more importantly, FileWriteAction accepts file content as text. That's not really ideal. Currently we decode files in the fastapi endpoint by using UTF-8 and send it to FileWriteAction. As I said, this works really really good with text based files like json, ts, text, etc. But fails with binary files. Also, multiple options seems to be available:

- not important for us. I'm assuming most of the users will be uploading text-based files which contains some code snippets. 
- process binary files correctly but this also implies there's some kind of mechanism in the runtime which can read binary files and extract information. I don't think that exist. 

Thank you for your help and feedback.

frontend/src/components/features/files/file-list.tsx
amanape: Instead of using an `id` property, maybe we can just use the `index` of the array instead. That should keep it simpler too since we don't really need the ID other than for removing entries.

frontend/src/components/features/files/file-list.tsx
```diff
@@ -0,0 +1,30 @@
...
+      {files.map((f, index) => (
+        <FileItem
+          key={index}
+          filename={f.filename}
+          onRemove={() => onRemove?.(f.id!)}
+        />
+      ))}
+    </div>
+  );
+}
```
amanape: Could we add some `<FaFile>` icon at the start here too?

frontend/src/components/features/images/image-carousel.tsx
mislavlukach: @amanape I had to keep it because typescript complained. 
It looks the same on first look but what the original line does is just defining a function without actually calling it. 
Proposed fix calls function immediately if the function is defined.

amanape:
@openhands-agent Please fix failing CI

openhands-agent:
[OpenHands](https://github.com/All-Hands-AI/OpenHands) started fixing the pr! You can monitor the progress [here](https://github.com/All-Hands-AI/OpenHands/actions/runs/15612517333).

amanape:
@OpenHands Please fix failing CI

frontend/src/utils/custom-toast-handlers.tsx
```diff
@@ -14,6 +14,7 @@ const TOAST_OPTIONS: ToastOptions = {
 };
 
 export const displayErrorToast = (error: string) => {
+  console.log("error", error);
```
amanape:
```suggestion
```

openhands/server/file_config.py
```diff
@@ -111,3 +111,29 @@ def is_extension_allowed(filename: str) -> bool:
...
+    filename_candidate = filename
+    copy_index = 0
+
+    while os.path.exists(os.path.join(folder_path, filename_candidate)):
+        if copy_index == 0:
+            filename_candidate = f"{name} copy{ext}"
+        else:
+            filename_candidate = f"{name} copy({copy_index}){ext}"
+        copy_index += 1
+    print(filename_candidate)
```
amanape:
```suggestion
```

openhands/server/files.py
```diff
@@ -0,0 +1,14 @@
+from __future__ import annotations
```
amanape: Are we using this import?

openhands/storage/locations.py
```diff
@@ -11,6 +11,8 @@ def get_conversation_dir(sid: str, user_id: str | None = None) -> str:
 def get_conversation_events_dir(sid: str, user_id: str | None = None) -> str:
     return f'{get_conversation_dir(sid, user_id)}events/'
 
+def get_conversation_files_dir(sid: str, user_id: str | None = None) -> str:
+    return f'{get_conversation_dir(sid, user_id)}files/'
```
amanape: Are we using this somwhere?

mislavlukach: We were using it before (same PR, previous commits). 
I wasn't sure if this will be needed or not after the backend review 
I'll remove it

mislavlukach:
@OpenHands Please fix failing CI

Issue #8944, opened 2025-06-06
Refactor Chat Input to Prevent Excessive Re-renders and Improve Performance


DustinEwan:
**What problem or use case are you trying to solve?**

There is a performance issue in the chat interface causing significant UI lag when a user types into the chat input box. The current implementation uses a controlled component pattern where the state is managed at the top-level ChatInterface component. This causes the entire component tree to re-render on every single keystroke, leading to a poor and sluggish user experience.
Technical Analysis

The performance bottleneck originates from the following data flow:

1. The TextareaAutosize in chat-input.tsx captures a keystroke and triggers its onChange handler.
2. This handler calls the setMessageToSend state setter function, which lives in the ChatInterface.tsx component.
3. The state update in ChatInterface.tsx triggers a re-render of itself and all of its children, including InteractiveChatBox.tsx and chat-input.tsx.
4. The new state value is passed back down the component tree as a value prop to update the input.

This cycle of lifting state up and passing it back down on every keystroke is highly inefficient and is the direct cause of the performance degradation.

Reproduction Steps

1. Run the application in a development environment.
2. Open the browser's React Developer Tools and navigate to the "Profiler" tab. (Alternatively, enable the why-did-you-render library).
3. Start profiling by clicking the record button.
4. Navigate to the chat screen and begin typing a message into the chat input box.
5. After typing a few characters, stop the profiler.
6. **Expected Result**: Only the input component itself should re-render frequently.
7. **Actual Result**: The profiler will show that the ChatInterface, InteractiveChatBox, Messages, and other unrelated components are re-rendering on every keystroke, highlighting the performance issue.

**Describe the UX or technical implementation you have in mind**

To resolve this, the chat input should be refactored from a **controlled component** to an **uncontrolled component**. This will delegate the state management during typing to the browser's DOM, eliminating the React re-render cycle.

    [ ] Remove the messageToSend state from ChatInterface.tsx.
    [ ] Remove the value and onChange props that are passed from ChatInterface through InteractiveChatBox.
    [ ] In chat-input.tsx, use the existing textareaRef to get the input's value directly from the DOM on submit (e.g., textareaRef.current.value).
    [ ] After a message is submitted, the <textarea> should be cleared imperatively (e.g., textareaRef.current.value = "").
    [ ] Functionality that relies on setting the input value, such as ChatSuggestions, must be updated to work with the new uncontrolled implementation.

PR #8948, opened 2025-06-06
Fix performance issue with ChatInput by converting to an uncontrolled input

DustinEwan:

This should eliminate the input lag when typing into the chat box once the conversation gets longer.  This is especially noticeable with chatty / reasoning models like Gemini 2.5 Pro.

---

This is a performance improvement, mainly.  It seems like at some point the `TextareaAutosize` component was changed from an **uncontrolled input** to a **controlled input** in order to accommodate the initial chat suggestions for new conversations.

The way that was implemented was by creating a bit of state in `chat-interface.tsx` that was then used to track the content of the textarea.  The state was then passed through props from `ChatInterface -> InteractiveChatBox -> ChatInput -> TextareaAutosize` and then configured to update with `TextareaAutosize`'s `onChange` event.

This resulted in a complete re-render of every component in that branch of the DOM on every keystroke, which also includes the `Messages` component.

Once the output from the LLM is sufficiently long, the user experiences significant latency between each change to the Chat Input as they wait for the full message log to re-render.

I converted the `TextareaAutosize` back to being an *uncontrolled input* which means that the browser handles the state internally and is disconnected from the react state.  We then use a `ref` to both fetch the state upon submission (i.e. when it's actually needed for application logic) and to inject state through a custom hook.

That hook works for any component that is nested within the `ChatInputProvider`, so I put it as deep as possible.  Someone more familiar with the full architecture of the frontend might be able to suggest a better place for the provider, but it's pretty common to just wrap the entire application in the providers you need, that way the context is globally available.

> _Sidenote: now that I've implemented this fully, I think the context state can be cleaned up a little bit to simply provide the ref as the state of the context, then `injectChatInputMessage` could become the logic contained within the `useEffect`, but as I was working through some synchronization bugs, I kinda just left it how it was once it was working as expected._

---

Addresses #8944


DustinEwan:
So that's kinda the rub.  You need a way of getting the TextareaAutosize component to re-render despite it not being directly attached to the react state.  The only time it's even relevant is when you click the one of the suggestion buttons on a new conversation, in which case we set the content of the internal textarea, then force a render so that the event gets picked up by `TextareaAutosize` to automatically resize the `ChatInput`.

I initially tried manually triggering `change` and `input` dom events on the textareaRef, but it wasn't working.  That's because `TextareaAutosize` uses `React.useLayoutEffect` internally to trigger it's resize function:

https://github.com/Andarist/react-textarea-autosize/blob/ed1894cd8611d99fbea1c47adcf6ee522b1030fd/src/index.tsx#L101

Thus we need some way of telling react to re-render so that the layout effect actually triggers and is picked up by `TextareaAutosize`.

That being said, I do think this can be simplified a bit as I alluded to in my note.  The force render, while a hack, is essential to this working, though, because we can't change how `TextareaAutosize` triggers it's resize event.  We could probably simplify the state management, though...

I'll give it a try real quick.

DustinEwan:
Okay, cool, so I colocated the forceRender to be only where we need it.  Unfortunately, there's no way around that because we need _some_ way of notifying the TextareaAutosize component that we shifted the ground under its feet so to speak.  This happens at exactly the same place that TextareaAutosize is added to the DOM, so there won't be any components rendered apart from the ChatInput and the TextareaAutosize.

I also eliminated the useState in the provider and replaced it with a ref and a lightweight pubsub pattern.  Since refs don't trigger updates, that means that we can place that Provider at the root of the app and it'll never cause a re-render.  The tradeoff is the need to subscribe to the provider context whenever we want to be notified of a message injected to the ChatInput.

This should be extremely performant considering that everything happens outside of React apart from the forced render within ChatInput.

Let me know what you think.

mamoodi:
@amanape if you have some time to re-look at this again 🙏

amanape:
Thanks for investigating. Please give me some time to think about the best direction here.

My impression is that if we need to force renders and implement pub/sub patterns for a text input, we're doing something wrong. If the textarea animation is the source of issues, then it is worth considering removing it entirely in favour of a simpler fix.

amanape:
Sorry for the delay. We're going to address this issue and it's surrounding context as part of https://github.com/All-Hands-AI/OpenHands/issues/9420

I'll close this PR in the meantime. Feel free to open a PR for another issue!

PR #8957, opened 2025-06-06
feat(chat): show message timestamp

mislavlukach:

---
This PR creates 2 hooks. 
1. use-localized-date-formatter - idea is to use locale from our react i18n instance to find how the date should by formatted. Additionally, it accepts formatType which allows you to pick between only showing hh:mm and showing the full date (hours included). 
2. use-relative-date-formatter - utilize 1. hook to provide flexible way to render dates in UI. If the date is today - this hook will pick formatType which shows only hh:mm, otherwise it picks full dat formatter.

Example are in the images below:
today
old
localized

---


amanape (approved):
LGTM but will wait for @rbren's approval. You just need to fix issues presented when running `npx tsc`

mamoodi:
@mislavlukach and @rbren did we want to merge this?

amanape:
No, the timestamp UI feels too prominent and isn’t necessary to show in this format. I’ll check with Paul for alternative approaches.

PR #8971, opened 2025-06-06
fix(devcontainer): configure host networking to fix runtime connection

llamantino:
…n issue

Fixes a networking issue in the devcontainer setup where the app failed to connect to the runtime container, resulting in connection refused errors.

---

Adds "runArgs": ["--network=host"] so the app can access the localhost-bound port exposed by the runtime container.

---


enyst (approved):
Thank you!

PR #8972, opened 2025-06-07
[BUG FIX] Remove usage of mapProvider and supporting tests due to undesired int…

### LINKED ISSUE #8689: [Bug]: Mistral AI in "simple" mode points to wrong name "mistral ai"

When you first setup your model, you can choose a bunch of preconfigured providers (OpenAI, Azure, Mistral, etc). If you choose "Mistral AI" and any model, lets say devstral-small-2505, the model name ends up being "mistral ai/devstral-small-2505" which breaks LiteLLM that is used in the backend. The correct mapping should be "mistral/devstral-small-2505". 

You can fix it if you go into the advanced settings but most users won't know the issue. Can you fix the mapping from "Mistral AI" to use "mistral" instead of "mistral ai"?

### OpenHands Installation

Docker command in README

### OpenHands Version

0.39

### Operating System

None

_No response_

wi-ski:

Fixes an issue where the model provider selection (e.g., Vertex AI) in the HeroUI autocomplete component did not correctly persist the selected value. Previously, typing and selecting an option would submit an invalid provider identifier (e.g., missing underscores), causing LightLLM to fail during startup. This change ensures that the selected value is accurately saved and used, eliminating hidden UI mismatches and improving reliability when configuring providers.

Video recording: https://www.loom.com/share/7a95d9fd23e246e7ab44376ab3bf4783?sid=e3ab1673-f86b-4c3d-88bb-a35e97208040

Relevant Herou UI Links:
https://github.com/heroui-inc/heroui/issues/3353
https://github.com/heroui-inc/heroui/pull/3375#issuecomment-2491442193

---

Removes usage of mapProvider and associated tests from the frontend.

---
Resolves #8689


frontend/src/utils/map-provider.ts
```diff
@@ -1,31 +0,0 @@
-// These are provider names, not user-facing text
-export const MAP_PROVIDER = {
-  openai: "OpenAI",
-  azure: "Azure",
-  azure_ai: "Azure AI Studio",
-  vertex_ai: "VertexAI",
```
enyst, line 6: I'm a bit surprised by this solution, I mean deleting all map provider. Are we sure, it seems to work for everything else. Could you please add in the PR description or here what exactly doesn't work for Vertex, so we can ping our FE master and see if maybe there's another cause?

enyst: Actually, https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1749258058350509?thread_ts=1749247191.636449&channel=C06U8UTKSAD&message_ts=1749258058.350509

Cc: @amanape

wi-ski: Response: https://www.loom.com/share/e8d853dff1874cfa8a71cfac87980607?sid=7af74516-b646-4cd5-90c6-cb4528cfd3ab

**ChatGPT's Summary of Video Transcript**

_Totally fair to be surprised - I was honestly torn myself. Here's the thinking:_

_There are only two references to MapProvider in main, and it's being used solely for styling the display of the underlying value. Functionally, it’s not doing much._

_The main driver behind this change is that Autocomplete in Hero UI is meant to behave like an input, not a select. So sending the display text (label) on submission aligns with how it’s intended to work. They even show an example using allowCustomValue, which supports this approach - but it also means users can type anything, so there's some tradeoff there._

_Using a select instead would force you into predefined options but lose the ability to type-to-search, which isn’t ideal for this case._

_So my decision leaned toward keeping things simpler and making the input reflect exactly what gets passed to LightLLM. This approach reduces complexity and better aligns with the expected behavior for future devs._

_That said, if we want to go full controlled component for the sake of styling or interaction fidelity, happy to do that too - just didn’t feel like the right tradeoff here. Let me know_

wi-ski:
Following up on this - repeating the wins:
1. Reduced code footprint
2. Reduced complexity
3. Increased future proof-ness.
4. Better Reflects underlying behavior

Cost: Loss of a pretty label.

PR #8974, opened 2025-06-07
feat: increase requests timeout to 60s

llamantino:

Reduces timeout errors during runtime download or git clone by increasing the default request timeout from 10s to 60s.

---

Increases the default timeout in send_request from 10 to 60 seconds.

---


tofarr:
think this value used to be much higher and was lowered due to poor performance due to long timeouts against operations that were never going to return successfully.

Maybe we can go for a middle ground and make it like 30 seconds?

tofarr (approved):

Issue #8111, opened 2025-04-27
Refactor agent controller


enyst:
It seems that time has come again: `agent_controller.py` has grown too much. It got over 1,400 lines of code, and it does a lot of things. We could split out some part(s) like the StuckDetector or ReplayManager are. Maybe a HistorySomething, e.g. HistoryManager or HistoryInitializer or ContextHandler.

enyst:
Gemini's first suggestion seems to be overdoing it in the other direction: 😅 
![Image](https://github.com/user-attachments/assets/746ed72e-feb7-423f-840e-2425d7875a5b)

happyherp:
# Integrate _handle_long_context_error with Condenser

I had some thoughts when reading through https://github.com/All-Hands-AI/OpenHands/pull/8103 as well.

I was completely unaware of `AgentController._handle_long_context_error` as a kind of fallback-condenser when we hit a Context-Window error. I think the code that creates the condensation should be in the Agent-Class, so that every agent can decide how this condensation should happen. And we probably want to create a condenser for this that we then call. 


It also creates a different program flow than the condenser. 

Instead of 

```python 
                    if self.agent.config.enable_history_truncation:
                        self._handle_long_context_error()
                        return
```

we could just resume the same way, as if the CondensationAction was created the normal way.  

```python 
                    if self.agent.config.enable_history_truncation:
                        action = self._handle_long_context_error()
```

That way we go through the same code as with a CondensationAction from a Condenser. 

```python 

    if action.runnable:
            if self.state.confirmation_mode and (
                type(action) is CmdRunAction or type(action) is IPythonRunCellAction
            ):
                action.confirmation_state = (
                    ActionConfirmationStatus.AWAITING_CONFIRMATION
                )
            self._pending_action = action

        if not isinstance(action, NullAction):
            if (
                hasattr(action, 'confirmation_state')
                and action.confirmation_state
                == ActionConfirmationStatus.AWAITING_CONFIRMATION
            ):
                await self.set_agent_state_to(AgentState.AWAITING_USER_CONFIRMATION)

            # Create and log metrics for frontend display
            self._prepare_metrics_for_frontend(action)

            self.event_stream.add_event(action, action._source)  # type: ignore [attr-defined]

        await self.update_state_after_step()

        log_level = 'info' if LOG_ALL_EVENTS else 'debug'
        self.log(log_level, str(action), extra={'msg_type': 'ACTION'})
```

enyst:
[...]

PR #9012, opened 2025-06-09
[Refactor, Fix]: Agent controller state/metrics management

### LINKED ISSUE #8238: [Bug]: Cost-Termination Flag Not Triggering Due to Incorrect Metrics Reference
**Describe the bug and reproduction steps**
While reviewing the codebase, in agent_controller.py, line 803 (https://github.com/All-Hands-AI/OpenHands/blob/de175dcc8757d377cb6a8aa45eba4ae835e87c44/openhands/controller/agent_controller.py#L803), it was noted that `self.state.metrics.accumulated_cost` is being checked instead of `self.agent.llm.metrics.accumulated_cost`. As per the understanding, `self.state.metrics` should represent the total cost; however, this value does not update after each LLM call. This behavior results in the cost-termination flag not being triggered when expected, even after setting `max_budget_per_task` in AppConfig.

**Potential solutions**
The code could be updated to use `self.agent.llm.metrics.accumulated_cost`, as initial tests have shown this resolves the issue.

**Logs, Errors, Screenshots, and Additional Context**
<N/A>

**Issue Created By**: Xingyao Wang on Slack
**Link to Slack Thread**: https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1746278964737369?thread_ts=1746278964.737369&cid=C06U8UTKSAD

### LINKED ISSUE #8858: [Bug]: Budget tracking seems inconsistent and difficult to follow

In the agent controller we can set a max budget per task. When the budget limit is hit we do two things

1. Extend the budget cap
2. Reset the agent controller

I've noticed that resetting the agent controller does two things 

1. It resets the the budget cap to `None`, which is then later set again to the initial budget cap
2. It resets the llm's accumulated cost to 0

I think this is an unideal model for a few reasons

1. We should always have the total accumulated llm cost information (resetting cost is a bad mental model, and leads to poor cost tracking)
2. The agent's accumulated cost seems to serve the same purpose as that of the llm accumulated cost (this could be simplified?)
3. When the budget cap is hit the FE displays the total cost + budget limit; when the user forcefully continues and hits the budget cap again, the budget cap is still the original amount (which is a bad mental model, it should have been extended)

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Operating System

None

_No response_

malhotra5:

- Make tracking for metrics more accurate and fault tolerant (accumulated cost, token usage, etc)
- Does not track metrics in a centralized manner yet (metrics reported in FE are for the main agent, and do not include metrics for auto generate convo titles, or running the condenser)

#### Problems

We have many problems with the way we handle budget caps in the agent controller

1. The budget information is reset when the user decides to continue after hitting the convo budget cap
2. We store accumulated cost in both the agent state and llm state, and they have eventual consistency
3. We extended budget when the traffic control state is `PAUSED`; but nowhere in the agent controller do we set it that way, and thus I'm unable to extend budget even if I wanted to
4. Throttle logic and iteration extension logic seem to be scattered across different parts of the agent controlled
5. When budget is extended, we simply extend the current max by the initial amount; we should use the current accumulated cost as the new starting point
6. When the agent is delegated, the initial amount is initialized as the max_budget setting. This means that the delegate extends the budget by twice the normal amount the parent would

#### Fixes

This PR we do the following

1. Removing any resetting of accumulated costs
2. Extend the budget when we are throttled, and remove the unnecessary complexity of `PAUSED` state for traffic control
3. Organize throttling and max cap extensions in one place (when the agent goes from error to running)
4. We extend the agent budget based on the current accumulated cost and the initial delta amount
5. We track delta amounts for budget now. This means if the agent or the parent hits the budget cap, we extend by a fixed delta amount (fixes delegate from extending budget cap in multiples, see [comment](https://github.com/All-Hands-AI/OpenHands/pull/8764#issuecomment-2919777095))
6. Controller state always holds the most current budget cap and accumulate cost, simplifying tracking

This helps us establish the following assumptions going forward

1. Accumulated cost is never reset, we're always looking at the current cost for the conversation even after extending budget
2. When restarting a conversation, we will resume with the same accumulated cost and budget caps from when we last left off
3. The metrics for the controller state and llm metrics are always the same. Metrics are always accumulated over time, we never start a fresh copy. Instead, for delegate agents we'll create a snapshot of the parent's metrics. The different between child and parent metrics gives us the local metrics for the delegate (this also fixes the problem where the controller metrics had eventual consistency with global llm metrics)

Fixes #8858
Fixes #8238

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:922b0a6-nikolaik   --name openhands-app-922b0a6   docker.all-hands.dev/all-hands-ai/openhands:922b0a6
```

### DIFF
--- a/openhands/controller/state/state_tracker.py
+++ b/openhands/controller/state/state_tracker.py
@@ -266,6 +266,17 @@ class StateTracker:
             self.state.budget_flag.current_value = self.state.metrics.accumulated_cost
 
     def merge_metrics(self, metrics: Metrics):
+        """
+            Merges metrics with the state metrics
+
+            NOTE: this should be refactored in the future. We should have services (draft llm, title autocomplete, condenser, etc)
+            use their own LLMs, but the metrics object should be shared. This way we have one source of truth for accumulated costs from
+            all services
+
+            This would prevent having fragmented stores for metrics, and we don't have the burden of deciding where and how to store them
+            if we decide introduce more specialized services that require llm completions
+
+        """
         self.state.metrics.merge(metrics)
         if self.state.budget_flag:
             self.state.budget_flag.current_value = self.state.metrics.accumulated_cost


openhands/controller/state_manager.py
```diff
@@ -0,0 +1,210 @@
...
+from openhands.events.event import Event
+from openhands.events.event_filter import EventFilter
+from openhands.events.observation.agent import AgentStateChangedObservation
+from openhands.events.observation.delegate import AgentDelegateObservation
+from openhands.events.observation.empty import NullObservation
+from openhands.events.serialization.event import event_to_trajectory
+from openhands.events.stream import EventStream
+
+
+class StateManager:
```
malhotra5: I'm thinking of creating this class so that we can move a lot of complexity from the agent controller here

It will in charge of 

1. Extending iteration or budget caps
2. Syncing the llm metrics with the state metrics
3. Restoring or updating state history
4. etc

Anything state related really

pcuci: this might make it easier to later extend to a use-case where users control the budget for each thread.

"to resolve this issue, don't spend more than $3; plan accordingly"

openhands/llm/metrics.py
```diff
@@ -193,22 +193,6 @@ def get(self) -> dict:
             'token_usages': [usage.model_dump() for usage in self._token_usages],
         }
 
-    def reset(self) -> None:
```
enyst, line 196: I agree, let's kill this, and if we really need it, hey, we'll find out exactly why and can rethink.

openhands/controller/agent_controller.py
```diff
@@ -600,40 +568,33 @@ async def set_agent_state_to(self, new_state: AgentState) -> None:
...
-                and self.max_budget_per_task is not None
-                and self._initial_max_budget_per_task is not None
-            ):
-                if self.state.metrics.accumulated_cost >= self.max_budget_per_task:
-                    self.max_budget_per_task += self._initial_max_budget_per_task
-        elif self._pending_action is not None and (
+            # Reset throttle state
+            self.state.traffic_control_state = TrafficControlState.NORMAL
+
+        if self._pending_action is not None and (
```
enyst, line 574: Is this `if self._pending_action is not None` new in this PR? If yes, why was it needed?

malhotra5: It's not new in this PR, not sure what it was being used for before

I moved this conditional from `elif` to `if` statement because the conditions for success are mutually exclusive

openhands/controller/agent_controller.py
```diff
@@ -658,6 +622,10 @@ async def set_agent_state_to(self, new_state: AgentState) -> None:
             EventSource.ENVIRONMENT,
         )
 
+        # Save state whenever agent state changes to ensure we don't lose state
+        # in case of crashes or unexpected circumstances
+        self.save_state()
```
malhotra5, line 601: I'm saving the metrics state to file store more aggressively now; I have a few reasons for this

1. Previously we only saved when the agent session was closed; if the server dies unexpectedly we loose all metrics info
2. Because metrics info is never reset, when restoring conversations we need to restore the metrics where they last left off as well, requiring more aggressive syncing of metrics with persistent store

openhands/controller/agent_controller.py
```diff
@@ -290,6 +290,9 @@ def step(self) -> None:
     async def _step_with_exception_handling(self) -> None:
         try:
             await self._step()
+            # Save state after successful step to ensure we don't lose state
+            # in case of crashes or unexpected circumstances
+            self.save_state()
```
enyst: Good catch, it may indeed fix some stuff we see.

(Just a quick thought, it may be good to make sure this still works with trajectory save (when `config.trajectory_path` is set, logic is in `main.py`) or headless in general - or perhaps, a follow-up could also save partial trajectories)

openhands/controller/agent_controller.py
enyst: I agree, just once per step is really a big change, IMHO we may not need it here too

malhotra5: Agreed will only invoke on change agent state

enyst:
Thank you for taking this up! It can be a headache ❤️

Related, please take a look
- https://github.com/All-Hands-AI/OpenHands/issues/8111

Re: `StateManager`
The first attempt at a split in the comments in the issue above, as Carlos pointed out, feels like it splits a god object into multiple mini-gods. 😅 Still easier to read than a 1,4k file, but idk. IMHO a 'state manager' moves code, but I'm not fully sure it reduces complexity? 🤔 

State is the state of the system, multiple things read it to adapt.

What if (unbaked thought) we think of it this way
- budget and iterations are ControlFlags. We have a state machine kind of logic, for how each of them influences the controller. Could they be small dataclasses that each know how to `check()` - true or false; `expand()` when needed; maybe even return the correct enum `next()`, considering headless/non-headless? So they're probably initialized with `state`...
- history is different, I agree we could be better off with it taken out of the agent_controller, but it... doesn't control the app/controller behavior.
- nit: could we avoid the word 'manager' or 'handler' in the class names, and still express what we want to express? They're so generic, everything can "manage stuff". Sometimes the answer is no, ok, just nitpick: it may be worth to ask the question.

malhotra5:
Thanks that's great feedback! Let me see what I can do about this

malhotra5:
Okay I took a shot at control flags for budget and iteration constraints 

It did end up simplifying things since limit checks, limit states, and limit extensions are handled within the dataclass

openhands/controller/agent_controller.py
```diff
@@ -779,34 +759,28 @@ async def _step(self) -> None:
             )
             return
 
+        self.state_manager.run_control_flags()
```
enyst: It seems this is the same call as below (line 771), and this one doesn't catch the exception. Maybe we only need the one below? One per step

malhotra5: Oh nice catch, yeah this line is a mistake

openhands/controller/state/deprecated_state.py
```diff
@@ -0,0 +1,76 @@
...
+from typing import Any
+
+from openhands.core.schema import AgentState
+from openhands.events.event import Event
+from openhands.llm.metrics import Metrics
+from openhands.memory.view import View
+
+
+@dataclass
+class DeprecatedState:
```
malhotra5: Turns out when storing the `State`, we pickle the object itself (instead of stringifying it first before pickling)

This means restoring old state objects will fail because the `State` schema has been updated - will double check whether this is necessary

enyst: 🤔  That was intended, I'm pretty sure. Pickling stores the object, and restores it, the python object with possibly complex structure.

There may be some compatibility code needed when unpickling old objects, like for example in Metrics I think we make sure that the attributes exist (because some did not exist in the past).

malhotra5: Gotcha will take a look at how we're doing it currently

Read up a bit on pickling and apparent it handles schema changes to some degree - will need to modify this code

malhotra5: I see warning like this when restoring the old agent_states files into the new schema

> 17:05:40 - openhands:WARNING: agent_session.py:490 - State could not be restored: Can't get attribute 'TrafficControlState' on <module 'openhands.controller.state.state' from '/Users/rohitmalhotra/Documents/Openhands/Openhands/openhands/controller/state/state.py'>

I think the pickle doesn't handle mismatches in attributes 

Will have to hold a deprecated copy for now; a way to mitigate this in the future is to serialize the class to a json first, then pickle the json

Ideally that could be a follow up to this PR

malhotra5: Pickle allows modifying the dict object in `__getstate__` and `__setstate__` 

the logic for backwards compatibility is contained there and I've added unit tests for this as well

enyst: I didn't see the code, so please take it with a grain of salt, but that can work. The other option is just to make sure this new attribute is optional or has a default value or is not needed by init (restoring pickle calls init) so that it can be initialized later

malhotra5: Yep we have default values for new fields

Also when restoring states with deprecated fields like `iteration`, we'll migrate these values to the new fields like `IterationControlFlag(current_value=deprecation_iteration_field_value)`

openhands/controller/agent_controller.py
```diff
@@ -451,6 +451,8 @@ async def _handle_observation(self, observation: Observation) -> None:
             log_level, str(observation_to_print), extra={'msg_type': 'OBSERVATION'}
         )
 
+        # FIX ME: it seems like observations don't carry llm_metrics anymore; rather the LLM class aggregates cost
```
enyst: I think obs coming from the LLM draft editor were carrying metrics and should still do.

It's because that happens in the runtime... 😅 so we had no `state` there, and the `llm` is not the same, it's the "draft_editor" LLM.

We have too much complexity 😂

malhotra5: Ahh thanks for the clarification

I'll fix this in a follow up PR; this PR is centralizing all metrics accumulation in the controller.agent.llm object

In a future PR I'll

1. create exactly one llm object
2. pass this object around for all microservices (condensor, llm draft editor, etc)
3. this pr ensures that any updates to this centralized llm object will be reflected in the agent controller/budget control flags

malhotra5: ^ i'll rewrite the comment here

enyst: Not sure about the above. Having multiple LLMs working on different things is intentional, it makes a lot of sense to do editing application with special editing models, or simpler models, rather than the same LLM, for example.

I was going to suggest instead that, on the draft editor - to not merge its results, keep them separate and save them separately, idk, maybe with `state.draft_metrics` or something. Then the evals would update their scripts to compute total costs. 🤔

malhotra5: Yeah I think we'd need to discuss more, but I'd like to do that outside of this PR. This PR doesn't modify the way multiple services construct their own LLM objects atm.

My initial thoughts are that we should have total conversation cost to be centralized in one place. The way I see it is, having separate LLM objects would help with traceability for spend on individual services, but it requires aggregating them together later to have total spend. I'd rather it be one centralized metrics object, where we identify which service is attempting to accumulate more spend (auto generation of convo titles, draft editor, condenser, etc)

This way we accumulate total cost in one place but can still break it down in terms of spend per service if needed. We don't sacrifice on traceability but still have a singular source of truth for total convo cost. 

I'd prefer always having total convo and spend extra effort breaking it down by individual services, rather than having cost per service and having to put extra effort into aggregating them. We'd also prevent having fragmented stores for metrics, and don't have the burden of deciding where and how to store them if we decide introduce more specialized services that require llm completions

malhotra5: ^ oh i see the problem is having a different model all together for the llm (like a small more lightweight one)

That should still be possible if we have services instantiate their own llm with a shared metric object  - I should clarify that in the comment

EDIT: one concern is that we'll have to be careful about race conditions when  multiple services modify the same metrics object simultaneously

enyst: It seems that github ate my comment for some reason. 🫣

I only wanted to note that this was a great argument:
> We'd also prevent having fragmented stores for metrics, and don't have the burden of deciding where and how to store them if we decide introduce more specialized services that require llm completions

and makes sense. (yes, the problem was the models; not the metrics, the metrics sorely need a rethink)

malhotra5:
I think this is a good stopping point

1. we have fixed metrics management
2. since `AgentController` is a god object, we've introduced control flags and moved history management out of the controller

I'm afraid attempting to further simplify the agent controller from this point is out of scope for this PR

openhands/controller/state/control_flags.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+@dataclass
+class ControlFlag(Generic[T]):
+    """Base class for control flags that manage limits and state transitions."""
+
+    initial_value: T
+    current_value: T
+    max_value: T
+    headless_mode: bool = False
```
csmith49, line 19: This field appears unused by the implementations.

openhands/controller/state/control_flags.py
```diff
@@ -0,0 +1,110 @@
...
+    """Control flag for managing budget limits."""
+
+    def reached_limit(self) -> bool:
+        """Check if the budget limit has been reached."""
+        self._hit_limit = self.current_value > self.max_value
+        return self._hit_limit
+
+    def expand(self, headless_mode) -> None:
+        """Expand the budget limit by adding the initial value to the current value."""
+        if self._hit_limit:
```
csmith49, line 85: Should this reference `self.reached_limit` instead of `self._hit_limit`? Any danger of accidentally using a stale `_hit_limit` value here?

Ditto for the `IterationControlFlag`.

openhands/controller/state/state.py
```diff
@@ -75,35 +79,43 @@ class State:
...
     session_id: str = ''
-    # global iteration for the current task
-    iteration: int = 0
-    # local iteration for the current subtask
-    local_iteration: int = 0
-    # max number of iterations for the current task
-    max_iterations: int = 100
+    iteration_flag: IterationControlFlag = field(
+        default_factory=lambda: IterationControlFlag(
+            initial_value=100, current_value=0, max_value=100
```
csmith49: `initial_value` is the amount the control flag gets expanded by, right? At least in this context it reads like we're setting the iteration flag already tripped.

openhands/controller/state/control_flags.py
malhotra5: Yep its around so that we can migrate legacy agent states (which contain these values) to the new state schema (the one with IterationControlFlag and BudgetControlFlag)

openhands/controller/state/control_flags.py
malhotra5: Hmm good point

Prior to this PR we had throttleflag that would check if we are in a throttled state. This is so that we don't accidentally increase limits without user consent. 

Without throttle flags we can extend budget without consent if the user executes the last available step from their budget, hits an agent error state, and sends another message switching the agent back to a running state.

Whenever we switch from agent error to agent running, we attempt to expand the flag. The `_hit_limit` operates similarly to throttle flag; it verifies that we have thrown an error that the user sees before we attempt the expand the flag.

> Any danger of accidentally using a stale _hit_limit value here?

I should definitely write unit tests for this good point

openhands/controller/state/state.py
malhotra5: Yep we expand the `max_value` via `current_amount` + `initial_value` (only when we've reached limits)

here `current_value` < `max_value` so the flag will be initialized without already being tripped

openhands/controller/state/control_flags.py
enyst: It should matter for the behavior when hitting a limit: 
- for headless, error and get out of the agent loop;
- for UI, error and pause for the user to say something.

Do those still happen?

malhotra5: Yep headless hits error and exits

UI you have the option to expand when user sends message

All of the logic is centralized and executed by the control flags

openhands/controller/state/state.py
enyst: 🤔 Could we name it... `step_increase`, if it's a step increase?

malhotra5: Will do!

openhands/llm/metrics.py
```diff
@@ -221,5 +205,58 @@ def copy(self) -> 'Metrics':
...
+            completion_tokens=current_usage.completion_tokens
+            - base_usage.completion_tokens,
+            cache_read_tokens=current_usage.cache_read_tokens
+            - base_usage.cache_read_tokens,
+            cache_write_tokens=current_usage.cache_write_tokens
+            - base_usage.cache_write_tokens,
+            context_window=current_usage.context_window,
+            per_turn_token=0,
+            response_id='',
+        )
```
csmith49, line 257: Suggestion for future refactors: we can add `__sub__` to `TokenUsage` which does all this arithmetic for us with `current_usage - base_usage`.

openhands/controller/state/state_tracker.py
```diff
@@ -0,0 +1,248 @@
...
+from openhands.events.observation.delegate import AgentDelegateObservation
+from openhands.events.observation.empty import NullObservation
+from openhands.events.serialization.event import event_to_trajectory
+from openhands.events.stream import EventStream
+from openhands.llm.metrics import Metrics
+from openhands.storage.files import FileStore
+
+
+class StateTracker:
+    """ """
```
csmith49: Can you add some of the context from the discussion with Engel to this doc-string?

openhands/llm/metrics.py
malhotra5: Great point, will do in a follow up pr

malhotra5: Marking this as unresolved for now so that I can spot it quickly in the future!

malhotra5:
@enyst would like that I hold off from merging util you've had more time to look into this?

enyst:
I think we’re better off merging, we need this. Thank you!

malhotra5:
Thanks for the help with reviews! I did one last QA and it went well, will be merging

PR #9015, opened 2025-06-09
[OH-Versa] Add remaining browsing & GAIA eval improvement

ryanhoangt:

---

Ported from https://github.com/All-Hands-AI/OpenHands/pull/8598
- Add file download observation via browsing
- Some minor improvements to multimodal and text-based browsing
- GAIA eval changes

Eval result:
- GAIA - `claude-3-7-sonnet`:
  - `main`: 12/30
  - This PR: 15/30 (50%)

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:4796978-nikolaik   --name openhands-app-4796978   docker.all-hands.dev/all-hands-ai/openhands:4796978
```


neubig:
Thanks @ryanhoangt , has this ben evaluated at all?

openhands/runtime/action_execution_server.py
```diff
@@ -198,6 +200,8 @@ def __init__(
         self.start_time = time.time()
         self.last_execution_time = self.start_time
         self._initialized = False
+        self.downloaded_files: list[str] = []
+        self.downloads_directory = '/workspace/downloads'
```
li-boxuan: This folder might not exist at all (in GAIA eval harness we explicitly create this folder). Whatever folder path we use, we should also keep both docker runtime and local runtime in mind.

@xingyaoww Do you have some idea here? I was going to suggest a new config option until I see a recent refactoring of yours (https://github.com/All-Hands-AI/OpenHands/pull/8242) that deprecates that approach.

li-boxuan: Anyways, whatever decision we make, we need to make sure `browser_env.py` uses the same path.

li-boxuan:
@OpenHands

Please add a test for download action in `tests/runtime/test_browsing.py`.

openhands/runtime/action_execution_server.py
```diff
@@ -611,7 +615,45 @@ async def browse_interactive(self, action: BrowseInteractiveAction) -> Observati
                 'Browser functionality is not supported on Windows.'
             )
         await self._ensure_browser_ready()
-        return await browse(action, self.browser, self.initial_cwd)
+        browser_observation = await browse(action, self.browser, self.initial_cwd)
+        if not browser_observation.error:
+            return browser_observation
+        else:
+            curr_files = os.listdir(self.downloads_directory)
```
li-boxuan, line 614: @adityasoni9998 I don't exactly remember about this, but we are doing a hack here, right? We are saying, if a browse action fails, then we assume it's a download action?

evaluation/benchmarks/gaia/scripts/run_infer.sh
adityasoni9998, line 1: Can you make this number as 60? We use 60 steps for all our experiments. (It was 54 for the very initial iterations on GAIA)

openhands/runtime/action_execution_server.py
adityasoni9998: So here is my understanding of how downloads work in BrowserGym and Playwright:
1. When agent tries to download a file, apparently the following error is returned by BrowserGym `Error: net::ERR_ABORTED at <URL>` (for example try executing goto() action for https://www.dailyscript.com/scripts/GOLDFINGER.doc to observe this). If you set the playwright config in browser_env.py to accept downloads, then apparently a file is downloaded and still an error is returned by browsergym.
2. So my logic is to check if a new file is present in the downloads folder (I keep track of this list), whenever the agent faces any error during browsing. We don't assume every browsing failure is a download.

I also remember that the above behaviour occurs if the agent clicks on a link that points to a file. I am not sure if there are cases when no error is raised by BGym and a file is downloaded at the same time. There maybe such cases, but I don't have any reproducible example for now.

openhands/runtime/browser/browser_env.py
```diff
@@ -94,6 +94,9 @@ def browser_process(self) -> None:
                 headless=True,
                 disable_env_checker=True,
                 tags_to_mark='all',
+                timeout=100000,
```
adityasoni9998, line 97: This timeout is the default playwright timeout in milliseconds for all its actions. BrowserGym unfortunately over-rides this in its code to 0.5s for most actions except for very few actions like goto() (See here: https://github.com/ServiceNow/BrowserGym/blob/main/browsergym/core/src/browsergym/core/action/functions.py)

Do we want to keep such large value in default OpenHands? I think it is better to have slightly slower browsing rather than optimizing for speed since the agent anyways struggles more with browsing than coding. If we show timeout errors frequently, its browsing performance will only degrade

evaluation/benchmarks/gaia/scripts/run_infer.sh
```diff
@@ -39,7 +39,7 @@ echo "LEVELS: $LEVELS"
 COMMAND="poetry run python ./evaluation/benchmarks/gaia/run_infer.py \
   --agent-cls $AGENT \
   --llm-config $MODEL_CONFIG \
-  --max-iterations 30 \
+  --max-iterations 54 \
```
adityasoni9998: Can you make max-iterations as 60? We use 60 as the default for our final expts and used 54 in the very initial design.

openhands/runtime/action_execution_server.py
adityasoni9998: Yes, the folder in browser_env.py must match with this folder. Also note that the directory must pre-exist or have sufficient permissions so that it can be created by the Playwright logic. The simplest, error-free way is to have a default location (that can be configured somehow if you want) where we can download files while ensuring that the directory exists before running the agent.

Also I think a better idea would be to directly open the file in the browser (using Xinyao's HTTP-based file viewing server) whenever possible. Say the agent downloaded a PDF, we can show the content of the file download observation in the text and open the file in the browser and send screenshot to agent.

openhands/runtime/action_execution_server.py
adityasoni9998: BrowserGym/Playwright is also very weird for file extensions of downloaded files - these files are downloaded with some random strings (similar to a uuid str) that don't have any extension. This is why I use puremagic to guess the extension when copying it to workspace. I don't know of better ways of tackling this.

adityasoni9998: [I am not sure if you can see my previous comment]

So here is my understanding of how downloads work in BrowserGym and Playwright:
1. When agent tries to download a file, apparently the following error is returned by BrowserGym `Error: net::ERR_ABORTED at <URL>` (for example try executing goto() action for https://www.dailyscript.com/scripts/GOLDFINGER.doc to observe this). If you set the playwright config in browser_env.py to accept downloads, then apparently a file is downloaded and still an error is returned by browsergym.
2. So my logic is to check if a new file is present in the downloads folder (I keep track of this list), whenever the agent faces any error during browsing. We don't assume every browsing failure is a download.

I also remember that the above behaviour occurs if the agent clicks on a link that points to a file. I am not sure if there are cases when no error is raised by BGym and a file is downloaded at the same time. There maybe such cases, but I don't have any reproducible example for now.

neubig:
@ryanhoangt I think I may let @adityasoni9998 and @li-boxuan  review this one, But I'm very happy to have it in.

openhands/memory/conversation_memory.py
```diff
@@ -443,6 +450,8 @@ def _process_observation(
         elif isinstance(obs, AgentCondensationObservation):
             text = truncate_content(obs.content, max_message_chars)
             message = Message(role='user', content=[TextContent(text=text)])
+        elif isinstance(obs, FileDownloadObservation):
+            message = Message(role='user', content=[TextContent(text=obs.content)])
```
enyst:
```suggestion
            text = truncate_content(obs.content, max_message_chars)
            message = Message(role='user', content=[TextContent(text=text)])
```

What do you think, maybe we could do this? Just in case it matters or it will matter. There are other observations here where it’s unlikely the content will be anywhere near the limit, we just apply it for safety. The current max is 30k characters, IIRC.

openhands/runtime/browser/browser_env.py
ryanhoangt: I'm wondering if you observed some issues with the default value, and whether we can choose a smaller number (e.g. 10s for each action)? Btw I see the doc doesn't recommend to change it: https://github.com/ServiceNow/BrowserGym/blob/6bc4c5156d32d7c6598ea376badc5ff79030fe5b/browsergym/core/src/browsergym/core/env.py#L89

evaluation/benchmarks/gaia/run_infer.py
```diff
@@ -94,30 +101,52 @@ def initialize_runtime(
     obs = runtime.run_action(action)
     assert obs.exit_code == 0
 
+    action = CmdRunAction(command='mkdir -p /workspace/downloads')
```
ryanhoangt, line 104: I think we don't need to explicitly create this folder, if we set 'downloads_path' to '/workspace/downloads', Playwright will create this folder when launching: https://github.com/microsoft/playwright/blob/114c9c045213888179b11f0d5a1e2fb87ebc1bd8/packages/playwright-core/src/server/browserType.ts#L269-L274

openhands/runtime/action_execution_server.py
ryanhoangt: @adityasoni9998 what do you think about making this a hidden directory, so that it doesn't interfere a lot with the ACI? If the user launches OH from scratch, the working directory will be `/workspace`, and having a `downloads/` directory there might be a bit unexpected to the agent?

Regarding the ACI output, if the download dir isn't hidden, the output is:
```
Here's the files and directories up to 2 levels deep in /workspace, excluding hidden items:
/workspace/
/workspace/downloads/
/workspace/test.txt
```
If we make it `/workspace/.downloads`, it will be:
```
Here's the files and directories up to 2 levels deep in /workspace, excluding hidden items:
/workspace/
/workspace/test.txt

1 hidden files/directories in this directory are excluded. You can use 'ls -la /workspace' to see them.
```
which I think looks a bit nicer.

adityasoni9998: I think either should just work fine. Anyways we provide the absolute path of the downloaded file to the agent whenever a FileDownloadObservation is returned by the `action_execution_server` resulting from a `BrowseInteractiveAction`

openhands/runtime/browser/browser_env.py
adityasoni9998: Yeah, BrowserGym doesn't recommend it. But without this change, I faced a lot of timeout errors for goto actions when loading websites that are large in size/slow to respond when using GAIA - this involves interaction with real-world websites like Wikipedia a not a simulated setup of cloned websites hosted locally like that in WebArena/TheAgentCompany. Note that this would only change the goto action's timeout and not the timeout of other actions like click, scroll, etc. which is hard-coded to 0.5s

My observation with Claude-3.7 has been that once the agent faces browsing timeouts, it completely ignores that website and searches for information somewhere else. So the default value is surely non-optimal. If this time is insignificant as compared to the avg time for one step in OpenHands, maybe we can let it be 100s or 50s if you want it to be lower.

adityasoni9998 (approved):
The changes look good to me! I am not too aware about the changes in the test_mcp.py file. But the source code logic seems fine.

enyst (approved):
Looks generally good to me, thank you. GAIA eval works as e2e test 😅

openhands/runtime/browser/browser_env.py
ryanhoangt: Sounds good, then let's keep it for now.

openhands/runtime/action_execution_server.py
SmartManoj: When running in local runtime, got the following error
`playwright._impl._errors.Error: BrowserType.launch: EACCES: permission denied, mkdir '/workspace'`

How about `/tmp/.downloads`?

neubig: We typically don't recommend evaluation in the local runtime, so I think it should be OK as-is.

SmartManoj: for normal running too

PR #9032, opened 2025-06-10
Feat: Visualize Agent Loop (User Input, Agent Output, Action-Obs pair…

### LINKED ISSUE #8916: [Feature]: Visualize Agent Loop (User Input, Agent Output, Action-Obs pair) and Record LLM Metrics with OpenTelemetry/Logfire
During our use of OpenHands, we found there is currently no UI to observe key steps in the agent loop, such as detailed action-observation pair calls, LLM request, and token usage (both per single turn and aggregated across multiple turns).

We instrumented the agent loop and LLM requests using telemetry, aggregating all steps by session/conversation. These logs can be viewed in real-time on the Logfire UI, as shown in the diagram below.

Image

Below are some design details:

**Instrumentation Structure**

The agent loop is mainly controlled by the agent controller,  logfire performs event tracking around the agent controller. The Instrumentation structure is designed as follows:

Image

_Session_ : the root node in the logfire log, every conversation created by the user on the UI is a Session, offline evaluation, each initialization of an AgentController can also be seen as a Session

_Conversation_ : Represents each round of user dialogue, usually starting with User Input and ending with AgentFinish or Error

_Action_ : Openhands internally converts the query into action and then executes it in the runtime
HandleAction & HandleObservation : The action-obs pair under the action node, the process of the controller processing the action, and the observation will be generated after the action is executed

_Chat LLM_ : Record the complete input and output of the model by role, as well as Metric information (Token, Duration, etc.)

**Case display**

- Multi-turn Conversation

https://logfire-us.pydantic.dev/fanzhijie-zed/openhands?since=2025-05-27T09%3A20%3A16.930Z

All conversations of a single session are under the Span of the Global Session, and each round of conversation is under a Conversation Span

Image

- Token Usage
All model requests can see metrics such as Input Token, Output Token, and Total Token. All Usages under a Span will be aggregated and calculated together.

Image

- Model Request
The input and output of the model request will be organized by Role (System, User, Assistant)

Image

zhijie1111:
…) and Record LLM Metrics with OpenTelemetry/Logfire

---

---
Fixes #8916


zhijie1111:
> Thanks for the PR, @zhijie1111 ! TBH I thought it was at least partially possible already, via litellm, for the LLM calls? I'm not sure we need many spots with a level of indirection in the code, for that. Have you tried to look at litellm support?

@enyst Thanks very much for your review! Our current idea is to improve the observability of the entire agent loop, including both the LLM and other behaviors. The behaviors we want to monitor now include the input and output of each conversation turn, action-observation pairs, and LLM calls. These elements have a nested structure—for example, one session contains multiple conversation turns, and actions and observations come in pairs, which is why some spans are used to manage this. I’m not sure if LiteLLM can only be a provider for LLM observability, but OpenTelemetry (OTel) is a general solution for collecting observability data. Currently, we use OTel as the collector and LogFire(potential to expand in the future. etc. datadog, langfuse) for visualization.  We would appreciate any further suggestions you might have regarding these goals.

zhijie1111:
> First off, really appreciate effort on observability. My initial reaction is that there are at least two significant architectural decisions being made in this PR. One is the introduction of OpenTelemetry traces, the second is LogFire LLM monitoring.
> 
> Otel is an open standard with a lot of momentum that we've already discussed adopting, probably an easy sell. LogFire is a vendor-specific solution, and if we explicitly support it we may then also need to support competitors.
> 
> Fortunately, we already have a way to integrate these LLM tracing providers, through our [litellm proxy integration](https://docs.all-hands.dev/usage/llms/litellm-proxy). Litellm in turn [supports Logfire](https://docs.litellm.ai/docs/observability/logfire_integration), [Langfuse](https://langfuse.com/docs/integrations/litellm/tracing) and others.
> 
> So I recommend you set up a litellm server with your LogFire integration and see what else is needed to get the visibility you're looking for. If there are changes on the OpenHands side we're open to that, such as otel traces or adding metadata to litellm requests. 

@raymyers 
Yes, LiteLLM is a great solution for tracing LLM requests. This could complement our tracing approach, because our current goal is to improve the observability of the entire agent loop, including both the LLM and other behaviors. The behaviors we want to monitor now include the input and output of each conversation turn, action-observation pairs, and LLM calls. OpenTelemetry (OTel) is a general solution for collecting observability data. 

Currently, we use OTel as the collector and LogFire (with potential to expand to others like Datadog or Langfuse) for visualization. 

You can find more details about agent loop behavior tracking in this [issue](https://github.com/All-Hands-AI/OpenHands/issues/8916).

raymyers:
> The behaviors we want to monitor now include the input and output of each conversation turn, action-observation pairs, and LLM calls. OpenTelemetry (OTel) is a general solution for collecting observability data.

That sounds reasonable, I'd supporting adding OTel traces.

> Currently, we use OTel as the collector and LogFire (with potential to expand to others like Datadog or Langfuse) for visualization.

We should not make OpenHands depend on LogFire and Datadog and Langfuse. Since all of those have OTel collectors, it should be possible to decouple this and still allow those use cases.

enyst:
To note, [this option](https://github.com/All-Hands-AI/OpenHands/blob/7dede37fd80f7aaa2f1f6b77383c284611bbccff/config.template.toml#L43) allows to save trajectories, and then the [trajectory visualizer](https://github.com/All-Hands-AI/trajectory-visualizer) can show them.

zhijie1111:
> We should not make OpenHands depend on LogFire and Datadog and Langfuse. Since all of those have OTel collectors, it should be possible to decouple this and still allow those use cases.

Make sense. We will make some adjustments to the implementation here.

zhijie1111:
> To note, [this option](https://github.com/All-Hands-AI/OpenHands/blob/7dede37fd80f7aaa2f1f6b77383c284611bbccff/config.template.toml#L43) allows to save trajectories, and then the [trajectory visualizer](https://github.com/All-Hands-AI/trajectory-visualizer) can show them.

Trajectory records events based on an event stream and also supports visualization. In comparison, OpenTelemetry (OTel) can provide similar or even more features, with support for more configurable visualization backends. Also, the event stream currently seems to lack LLM tracing, which affects the completeness of Trajectory, and I’m not sure if this will be added in the future. We can have further discussions on whether we need to introduce OTel for tracing.

enyst:
@zhijie1111 I'll step back here, and I'm okay with OTel traces introduction, if you and Ray settled on the approach. You are correct to point out that I was still pondering that custom solution with event trajectories, but it doesn't integrate well with others.

OpenHands needs observability, otherwise we're all stuck trying different things that only partially work. Thank you for the work on it!

Nit: I would love it if we can implement in a way that is minimally intrusive for code readability. Like using decorators? In particular in core classes, where it could otherwise obscure the logic.

raymyers:
Looking more into this, I see that LogFire actually does support sending to other OpenTelemetry backends so I could see it functioning as a vendor neutral SDK. With that in mind I'll re-evaluate the PR. Should have an update by Friday, cheers!

https://logfire.pydantic.dev/docs/why/#opentelemetry-under-the-hood

raymyers:
OK mostly good news, we plan to instrument OpenTelemetry in the next few weeks. It will be using the default OTel SDKs so this PR will have to be reworked but it should be pretty easy at that point to add the spans you need and you can send them to LogFire or any collector.

Thanks for pushing for this, it's been a long time coming.

https://github.com/All-Hands-AI/OpenHands/issues/9670

zhijie1111:
> OK mostly good news, we plan to instrument OpenTelemetry in the next few weeks. It will be using the default OTel SDKs so this PR will have to be reworked but it should be pretty easy at that point to add the spans you need and you can send them to LogFire or any collector.
> 
> Thanks for pushing for this, it's been a long time coming.
> 
> #9670

@raymyers That would be great! If you could include the telemetry protocol natively, we would be very happy to integrate it.

mamoodi:
Going to close this as it has been stale for a long time. Please let me know if it needs reopening.

PR #9056, opened 2025-06-11
Add more log context on key events

raymyers:

---

Agent Status Error was logged with no context, intended for counting and assuming the exception would be logged elsewhere. However, it's proving difficult to understand which errors are directly causing this so adding a reason. Also changing it from a log to an info.

Similar changes are made to session start logging to make it more descriptive.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c414b7f-nikolaik   --name openhands-app-c414b7f   docker.all-hands.dev/all-hands-ai/openhands:c414b7f
```

### DIFF
--- a/openhands/server/session/agent_session.py
+++ b/openhands/server/session/agent_session.py
@@ -121,7 +121,7 @@ class AgentSession:
         self._started_at = started_at
         finished = False  # For monitoring
         runtime_connected = False
-
+        restored_state = False
         custom_secrets_handler = UserSecrets(
             custom_secrets=custom_secrets if custom_secrets else {}
         )
@@ -172,7 +172,7 @@ class AgentSession:
                     agent_configs,
                 )
             else:
-                self.controller = self._create_controller(
+                self.controller, restored_state = self._create_controller(
                     agent,
                     config.security.confirmation_mode,
                     max_iterations,
@@ -203,7 +203,7 @@ class AgentSession:
                 'signal': 'agent_session_start',
                 'success': success,
                 'duration': duration,
-                'has_replay': replay_json is not None
+                'restored_state': restored_state
             }
             if success:
                 self.logger.info(
@@ -261,7 +261,7 @@ class AgentSession:
         """
         assert initial_message is None
         replay_events = ReplayManager.get_replay_events(json.loads(replay_json))
-        self.controller = self._create_controller(
+        self.controller, _ = self._create_controller(
             agent,
             config.security.confirmation_mode,
             max_iterations,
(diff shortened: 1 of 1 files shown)


openhands/server/session/agent_session.py
```diff
@@ -197,14 +196,24 @@ async def start(
...
-                    'duration': (time.time() - started_at),
-                },
-            )
+            duration = (time.time() - started_at)
+
+            log_metadata = {
+                'signal': 'agent_session_start',
+                'success': success,
+                'duration': duration,
+                'has_replay': replay_json is not None
```
raymyers: The intention is that has_replay should indicate if this is a new session or resuming one (right?), so we can see if resuming fails more often.

openhands/server/session/agent_session.py
```diff
@@ -197,14 +196,24 @@ async def start(
...
+            }
+            if success:
+                self.logger.info(
+                    f'Agent session start succeeded in {duration}s',
+                    extra=log_metadata
+                )
+            else:
+                self.logger.error(
+                    f'Agent session start failed in {duration}s',
+                    extra=log_metadata
```
raymyers, line 216: Sometimes this indicates an exception, which is logged in the parent call session::start().

Why are AgentSession and Session different classes again?

openhands/server/session/agent_session.py
rbren: replay != resume

replay is a dev feature where you can upload JSON

rbren: There's a fn in here like "get existing agent state" which may or may not return None--that's your best indicator of if we're resuming an old session

openhands/server/session/agent_session.py
rbren: They don't really need to be different classes--one just holds the other as a member

rbren: Also good call splitting these logs--this tricked us the other day

rbren (approved):

openhands/server/session/agent_session.py
raymyers: Thanks. Should be correct now, tracking restored_state instead of has_replay. Had to return a tuple to get it, but it's just an internal method so should be OK, restructuring seemed uglier. Or would have had to give AgentController a public property to express it.

PR #9078, opened 2025-06-11
Add workspace settings configuration to CLI

### LINKED ISSUE #9077: [Feature]: Add workspace settings configuration to CLI
  Users are experiencing confusion regarding workspace settings when using the CLI. There is currently no straightforward way to set the workspace via the CLI, which leads to confusion about alternative methods such as environment variables (`SANDBOX_VOLUMES`) and `config.toml`. This impacts usability and creates friction during the setup process.

**Desired UX of the solution**
  - Add an option in the CLI that allows users to configure the workspace settings directly.
  - Ensure consistent defaults for workspace settings to improve user experience.

**Technical implementation**
  - Enhance the CLI to provide a straightforward option for setting workspace settings during initial setup.
  - Deprecate alternate methods like configuring through `config.toml` and environment variables.

**Alternatives considered**
  - Keeping the workspace configuration in `config.toml` with clear documentation adjustments. However, this approach does not resolve the user experience issues entirely.

**Additional context**
  Extensive conversations in Slack indicate user confusion regarding configuration options and default behaviors.

**Issue Created By**: Graham Neubig on Slack
**Link to Slack Thread**: https://openhands-ai.slack.com/archives/C078L0FUGUX/p1749644955190739?thread_ts=1749644955.190739&cid=C078L0FUGUX

neubig:
This PR adds workspace settings configuration to the CLI, addressing issue #9077.

## Changes
- Added a new `modify_workspace_settings` function to configure workspace settings via CLI
- Updated the Settings model to include workspace_base and sandbox_volumes fields
- Added workspace settings to the settings display
- Added a new "Workspace" option in the settings menu
- Added tests for the new functionality

## Testing
- Added unit tests for the new workspace settings functionality
- Manually tested the CLI workflow

Fixes #9077

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/390f0fc7f4b64afe9305258eb5bc41a9)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:9bf45bd-nikolaik   --name openhands-app-9bf45bd   docker.all-hands.dev/all-hands-ai/openhands:9bf45bd
```


neubig:
@openhands python unit tests are failing. Do an RCA of the root cause and then make a minimal fix to fix the issue

neubig:
@openhands workspace_base is deprecated, so we can revert the parts of this PR that use that. sandbox_volumes is the only variable we need to care about.

openhands/storage/data_models/settings.py
```diff
@@ -42,6 +42,8 @@ class Settings(BaseModel):
     search_api_key: SecretStr | None = None
     email: str | None = None
     email_verified: bool | None = None
+    workspace_base: str | None = None
```
neubig:
```suggestion
```

openhands/storage/data_models/settings.py
```diff
@@ -131,5 +133,7 @@ def from_config() -> Settings | None:
             remote_runtime_resource_factor=app_config.sandbox.remote_runtime_resource_factor,
             mcp_config=mcp_config,
             search_api_key=app_config.search_api_key,
+            workspace_base=app_config.workspace_base,
```
neubig:
```suggestion
```

neubig:
@openhands add a test checking if it is possible to start a cli session, use the settings menu to set the sandbox volume, start a new conversation, and make sure that the sandbox volume setting is still preserved. try to mock nothing to make sure that it is a faithful test of the cli's actual behavior. note that the sandbox volume should be preserved in the settings written to disk, just as the llm model and api key are

mamoodi:
- submitted instances: 50
- empty patch instances: 0
- resolved instances: 27
- unresolved instances: 23
- error instances: 0 - Download at: https://github.com/All-Hands-AI/pr-eval-results/releases/download/1.0.0/add-workspace-settings-to-cli-786_25-06-13-03-34.tar.gz

neubig:
I tested and this seems to be working OK.

enyst (changes requested):
Thank you for tackling this issue!

I have a bit of a doubt about the implementation, please correct me if I didn’t get this right. The PR seems to define sandbox volumes as if they were docker volume definitions, but CLI mode doesn’t use docker runtime, it uses CLIRuntime, which is fully local. I’m not sure how that would work for people?

In addition, the PR adds an attribute in Settings and it applies it in CLI if config.sandbox_volumes is not defined, but I think maybe this should generally work the other way around: what is in “settings” overrides config, and fallback to config if the “setting” isn’t present. This is just a nit.

I’m also not sure if this has a consequence on the web UI, because the `server` also uses this BaseModel for settings. If we go with it, I feel we need to test a bit that the web frontend works normally with the new setting, but also that it’s not accessible for override from the outside on the web API… since that might break security on the SaaS

enyst:
I’d like to take a step back here and think again. What exactly are people missing, what is the bug we’re fixing, and how do we want the experience to be?

I’m using CLI on a regular basis for some weeks now, and it works fine. I usually set `workspace` in my `config.toml`. I think, if the user doesn’t have it set, we want `openhands CLI` to just work, from the directory that it’s run from.

Does that make sense?

PR #9086, opened 2025-06-11
chore - Add pydantic lib to type checking

raymyers:

---

We do not use the full poetry dependencies for mypy type checking, we only check the libraries listed in `dev_config/python/.pre-commit-config.yaml`. This means we miss a lot of checks.

This adds pydantic lib makes changes to make mypy pass.

---

### DIFF
--- a/openhands/server/routes/manage_conversations.py
+++ b/openhands/server/routes/manage_conversations.py
@@ -316,7 +316,7 @@ async def get_prompt(
         raise ValueError('Settings not found')
 
     llm_config = LLMConfig(
-        model=settings.llm_model,
+        model=settings.llm_model or '',
         api_key=settings.llm_api_key,
         base_url=settings.llm_base_url,
     )
--- a/dev_config/python/.pre-commit-config.yaml
+++ b/dev_config/python/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
     hooks:
       - id: mypy
         additional_dependencies:
-          [types-requests, types-setuptools, types-pyyaml, types-toml, types-docker, lxml]
+          [types-requests, types-setuptools, types-pyyaml, types-toml, types-docker, pydantic, lxml]
         # To see gaps add `--html-report mypy-report/`
         entry: mypy --config-file dev_config/python/mypy.ini openhands/
         always_run: true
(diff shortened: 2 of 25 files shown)


openhands/server/routes/manage_conversations.py
```diff
@@ -316,7 +316,7 @@ async def get_prompt(
         raise ValueError('Settings not found')
 
     llm_config = LLMConfig(
-        model=settings.llm_model,
+        model=settings.llm_model or '',
```
raymyers, line 325: Maybe we should check this and raise a ValueError? That's probably what the Pydantic model would have done.

enyst: IIRC what we name `settings` is loaded from `settings.json`, and it should only **override** the app configuration on some options like this. Just like, let's say, CLI parameters override the app configuration if passed to the command.

If it's not in the .json, the behavior should be (and maybe it isn't atm, though I think it is somewhere in `server` too) that it falls back and gets initialized from the app configuration, meaning from env vars and config.toml.

With the caveat that I didn't check the surrounding code here exactly, IMHO maybe we shouldn't add a ValueError if that means we'll have to remove it anyway when we fix/review this behavior?

raymyers: Thanks, given that context I'd be fine leaving it as is (defaulting to blank), I don't think we would really get to this point without it set anyway

raymyers:
@OpenHands please use the makefile to run backend lint, and do nothing else

raymyers:
@OpenHands please address these with ignore comments

openhands/memory/conversation_memory.py:405: error: Item "ImageContent" of "TextContent | ImageContent" has no attribute "text"  [union-attr]
openhands/memory/conversation_memory.py:413: error: Item "ImageContent" of "TextContent | ImageContent" has no attribute "text"  [union-attr]
openhands/memory/conversation_memory.py:448: error: List item 0 has incompatible type "str | None"; expected "str"  [list-item]
openhands/memory/conversation_memory.py:456: error: Item "ImageContent" of "TextContent | ImageContent" has no attribute "text"  [union-attr]
openhands/memory/conversation_memory.py:464: error: Item "ImageContent" of "TextContent | ImageContent" has no attribute "text"  [union-attr]
openhands/integrations/utils.py:29: error: Statement is unreachable  [unreachable]
Found 6 errors in 2 files (checked 332 source files)

neubig (approved):
LGTM. I'm not a huge fan of adding so many type ignores, but maybe it's better to get this in now and then gradually remove the type ignores later

PR #9097, opened 2025-06-12
Condensation request signal in event stream

csmith49:

---

This PR adds a new event that represents a request for a condensation. This is intended to clean up the control flow of the agent controller by allowing us to use existing condensation implementations instead of the `apply_conversation_window` logic (which is a bit brittle and doesn't produce any kind of summary for the forgotten events).

As a bonus, it becomes possible to trigger condensation with a tool call, meaning the agent can now choose to condense or can do so when requested by the user.

---

### DIFF
--- a/openhands/agenthub/codeact_agent/codeact_agent.py
+++ b/openhands/agenthub/codeact_agent/codeact_agent.py
@@ -12,6 +12,9 @@ if TYPE_CHECKING:
 import openhands.agenthub.codeact_agent.function_calling as codeact_function_calling
 from openhands.agenthub.codeact_agent.tools.bash import create_cmd_run_tool
 from openhands.agenthub.codeact_agent.tools.browser import BrowserTool
+from openhands.agenthub.codeact_agent.tools.condensation_request import (
+    CondensationRequestTool,
+)
 from openhands.agenthub.codeact_agent.tools.finish import FinishTool
 from openhands.agenthub.codeact_agent.tools.ipython import IPythonTool
 from openhands.agenthub.codeact_agent.tools.llm_based_edit import LLMBasedFileEditTool
@@ -118,6 +121,8 @@ class CodeActAgent(Agent):
             tools.append(ThinkTool)
         if self.config.enable_finish:
             tools.append(FinishTool)
+        if self.config.enable_condensation_request:
+            tools.append(CondensationRequestTool)
         if self.config.enable_browsing:
             if sys.platform == 'win32':
                 logger.warning('Windows runtime does not support browsing yet')
--- a/openhands/agenthub/codeact_agent/tools/condensation_request.py
+++ b/openhands/agenthub/codeact_agent/tools/condensation_request.py
@@ -1,0 +1,23 @@
+from litellm import ChatCompletionToolParam, ChatCompletionToolParamFunctionChunk
+
+_CONDENSATION_REQUEST_DESCRIPTION = """Request a condensation of the conversation history when the context becomes too long or when you need to focus on the most relevant information.
+
+This tool helps manage conversation context by requesting a summary of the conversation history. Use it when:
+1. The conversation has become very long and you need to reduce context size
+2. You want to focus on the most important information from the conversation
+3. You're experiencing issues with context limits
+
+The tool takes no arguments and simply requests the system to condense the conversation history."""
+
+CondensationRequestTool = ChatCompletionToolParam(
+    type='function',
+    function=ChatCompletionToolParamFunctionChunk(
+        name='request_condensation',
+        description=_CONDENSATION_REQUEST_DESCRIPTION,
+        parameters={
+            'type': 'object',
+            'properties': {},
+            'required': [],
+        },
+    ),
+)
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -59,7 +59,11 @@ from openhands.events.action import (
     NullAction,
     SystemMessageAction,
 )
-from openhands.events.action.agent import CondensationAction, RecallAction
+from openhands.events.action.agent import (
+    CondensationAction,
+    CondensationRequestAction,
+    RecallAction,
+)
 from openhands.events.event import Event
 from openhands.events.event_filter import EventFilter
 from openhands.events.observation import (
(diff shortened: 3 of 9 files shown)


openhands/controller/agent_controller.py
```diff
@@ -1160,180 +1167,6 @@ def _init_history(self) -> None:
...
-                if forgotten_event_ids
-                else 0,
-                forgotten_events_end_id=max(forgotten_event_ids)
-                if forgotten_event_ids
-                else 0,
-            ),
-            EventSource.AGENT,
-        )
-
-    def _apply_conversation_window(self, history: list[Event]) -> list[Event]:
```
csmith49, line 992: Moved to `ConversationWindowCondenser.get_condensation`.

tests/unit/test_agent_controller.py
```diff
@@ -1512,71 +1544,3 @@ def test_system_message_in_event_stream(mock_agent, test_event_stream):
     assert isinstance(events[0], SystemMessageAction)
     assert events[0].content == 'Test system message'
     assert events[0].tools == ['test_tool']
-
-
-@pytest.mark.asyncio
-async def test_openrouter_context_window_exceeded_error(
```
csmith49, line 1620: Captured by the parametrized `test_context_window_exceeded_error_handling`.

openhands/core/config/agent_config.py
```diff
@@ -28,6 +28,8 @@ class AgentConfig(BaseModel):
     """Whether to enable think tool"""
     enable_finish: bool = Field(default=True)
     """Whether to enable finish tool"""
+    enable_condensation_request: bool = Field(default=True)
```
enyst: If we make it true by default, I think we need to eval it before merging (or disable manually in evals).

IMHO there are several benefits of this PR even if we set it false, and/or eval doesn’t quite like it, so we may need to make a choice. Can we try an eval?

openhands/core/config/condenser_config.py
```diff
@@ -169,6 +169,15 @@ class CondenserPipelineConfig(BaseModel):
 
     model_config = {'extra': 'forbid'}
 
+class ConversationWindowCondenserConfig(BaseModel):
+    """Configuration for ConversationWindowCondenser.
+
+    Not currently supported by the TOML or ENV_VAR configuration strategies.
```
enyst, line 173: Is it possible to fix this?

csmith49: Ah, Claude-generated doc-string. I'll make sure it's supported then fix this.

openhands/core/config/agent_config.py
csmith49: I'm cool with setting it to `False` by default. Right now the tool call is a little brittle (agent loves to just keep on condensing), maybe we just keep it `False` until it's evaluated and stress-tested a bit.

openhands/events/action/agent.py
```diff
@@ -195,3 +195,17 @@ def message(self) -> str:
         if self.summary:
             return f'Summary: {self.summary}'
         return f'Condenser is dropping the events: {self.forgotten}.'
+
+@dataclass
+class CondensationRequestAction(Action):
```
enyst, line 201: Just a thought: for a while, we had actions that are about the agent itself prefixed with `Agent*` if they were also available to the agent to return. It’s not respected since forever by `ChangeAgentStateAction`, but the rest did…; on the other hand, I think we’ve kind of broken it a bit more with the `RecallAction` - which is currently not available to the agent, but it could become!

🤔

Due to the length of this name, too, I’m tempted to say, let’s leave the name as it is, without Agent*, and just agree to send that pattern to the dust bin of history…

openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -118,6 +121,8 @@ def _get_tools(self) -> list['ChatCompletionToolParam']:
             tools.append(ThinkTool)
         if self.config.enable_finish:
             tools.append(FinishTool)
+        if self.config.enable_condensation_request:
```
enyst, line 125: Maybe we could also add it to `config.template.toml` with a line of doc, for people to know (or for us to easily point to it when asked 😅)

openhands/agenthub/codeact_agent/tools/condensation_request.py
```diff
@@ -0,0 +1,23 @@
+from litellm import ChatCompletionToolParam, ChatCompletionToolParamFunctionChunk
+
+_CONDENSATION_REQUEST_DESCRIPTION = """Request a condensation of the conversation history when the context becomes too long or when you need to focus on the most relevant information.
+
+This tool helps manage conversation context by requesting a summary of the conversation history. Use it when:
+1. The conversation has become very long and you need to reduce context size
+2. You want to focus on the most important information from the conversation
+3. You're experiencing issues with context limits
```
enyst: This last bit, in particular, feels like Claude may love it too much. I seem to recall that we had a similar string as message for the AgentCondensationAction that made it into agent’s history, and Claude was getting tripped up by it, thinking it needs to do it or something.

openhands/core/config/agent_config.py
enyst: Cool, let’s do that! 

Aha 😅 , I also made a small note below, where the prompt may be a bit _like_ a message we had before, which Claude was very interested in, for some reason 🤷

enyst (approved):
Thank you for this! IMHO this is an elegant solution out of our current disjointed handling, and I love the way it’s heading!

One thing we may want to consider, is that we may (?) end up in an infinite loop again for an agent that doesn’t have a condenser. 🤔 It seems like it would just return the `request` action to the controller, which no longer truncates nor stops. If that is the case, maybe… maybe we could enable this new condenser in all agents?

openhands/agenthub/codeact_agent/tools/condensation_request.py
csmith49: I've deleted this, helps avoid the loops where the agent continuously requests a condensation. Before (and if!) we turn this on as a default tool we'll want some better fixes here I think.

openhands/events/action/agent.py
csmith49: I don't have strong opinions on this, but I agree the name would be a bit long. Are we using the `Agent*` prefix for technical reasons or just for logical ones?

enyst: No technical reasons. So we can just change our mind.

li-boxuan:
@csmith49 Just curious: I see `enable_condensation_request` is disabled by default. Has this option been ever evaluated? I feel like it might be useful for web browsing tasks.

PR #9101, opened 2025-06-13
Fix CLI displaying claude-2 as default model for anthropic provider

### LINKED ISSUE #9100: [Bug]: cli displays "claude-2" as default model

The CLI still displays claude-2 as the default model when "anthropic" is selected as the provider. Instead it should display the best verified model from claude (which is currently claude-sonnet-4-20250514).

1. First write a test that reproduces the error. This should be a test that checks which model is displayed by default in the CLI, expecting claude-sonnet-4-20250514. You MUST confirm that this test currently FAILs, because currently the code is broken. If you cannot confirm that the test fails, you must stop immediately and report back, as you were not able to reproduce the error.
2. If and only if you were able to write a failing test in 1., proceed to make the minimal possible fix to fix the error, making sure that the best verified model is used, instead of the first model in the provider list.

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Operating System

None

_No response_

neubig:
## Description

This PR fixes issue #9100 where the CLI was displaying "claude-2" as the default model when "anthropic" was selected as the provider. Now it correctly displays the best verified model from Claude, which is currently "claude-sonnet-4-20250514".

## Changes

- Modified the default model selection logic in `openhands/cli/settings.py` to use the best verified model for each provider
- Added a test in `tests/unit/test_cli_default_model.py` to verify that the best verified model is displayed

## Testing

- Added a new test that verifies the default model displayed for the anthropic provider is the best verified model
- The test initially failed with the old code and passes with the new code

Fixes #9100

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/30b8a55cd56049e38124349d695b1797)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:4d18729-nikolaik   --name openhands-app-4d18729   docker.all-hands.dev/all-hands-ai/openhands:4d18729
```


neubig:
Reverse the order of the elements in `VERIFIED_ANTHROPIC_MODELS`, but put claude-4-sonnet above claude-4-opus. This will then allow you to simplify the code for selecting the best model from the provider to select the first element in the list, regardless of which list is used.

neubig:
@openhands Reverse the order of the elements in VERIFIED_ANTHROPIC_MODELS, but put claude-4-sonnet above claude-4-opus. This will then allow you to simplify the code for selecting the best model from the provider to select the first element in the list, regardless of which list is used.

xingyaoww (approved):

PR #9111, opened 2025-06-13
(Hotfix): tokens go stale for restarted convos in cloud openhands

malhotra5:

- Fix missing provider tokens when restarting conversations in cloud openhands

---

#### Problem
We recently added a `/start` endpoint which starts the agent loop in prep for nested runtimes. This endpoint is called before the websocket connection is made, when the user clicks into a conversation from the conversations pane. 

We weren't passing information on which providers are currently set, which means the runtime thinks no tokens exist. 

#### Solution

1. Pass information on which providers are currently set from the FE
2. For bulletproofing the current logic in the FE for passing current provides, we also modify the `useEffect`s to watch for when providers have changed. This is because the FE always starts with no providers and the information is later loaded in from the BE. We want the websocket client to reflect this state when making its connection.

3. We also memoize the `useUserProviders`. This is because
- When you navigate to a conversation page, the `WsClientProvider` component mounts
- The `useUserProviders` hook is called
- Even if the settings data is the same, `convertRawProvidersToList` creates a new array instance every time it's called
- This means every time the component re-renders (for any reason), the providers array gets a new reference
- This causes the WebSocket useEffect to re-run because it sees the dependency has "changed"

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c7c14c4-nikolaik   --name openhands-app-c7c14c4   docker.all-hands.dev/all-hands-ai/openhands:c7c14c4
```


malhotra5:
@openhands looks like there was a new endpoint added `/api/conversations/${conversationId}/start` 

please make the following changes

1. make sure the FE passes the users current set providers when calling this endpoint (look at how the websocket is instantiated with this information)
2. make the backend receive this information
3. please make sure that we're correctly setting up convo init data before starting the agent loop (see how its done in `openhands/server/listen_socket.py`; try to dedup the logic as much as possible)

openhands/server/services/conversation_service.py
```diff
@@ -135,3 +140,64 @@ async def create_new_conversation(
...
+    return MappingProxyType(provider_information)
+
+
+async def setup_init_convo_settings(
+    user_id: str | None, conversation_id: str, providers_set: list[ProviderType]
+) -> ConversationInitData:
+    """Set up conversation initialization data with provider tokens.
+
+    This function replicates the logic from listen_socket.py to ensure
+    consistent conversation initialization between websocket and REST endpoints.
```
malhotra5:
```suggestion
```

openhands/server/routes/manage_conversations.py
```diff
@@ -95,6 +98,10 @@ class ConversationResponse(BaseModel):
     conversation_status: ConversationStatus | None = None
 
 
+class StartConversationRequest(BaseModel):
```
malhotra5:
```suggestion
class ProvidersSetModel(BaseModel):
```

openhands/server/routes/manage_conversations.py
```diff
@@ -395,6 +402,7 @@ async def _get_conversation_info(
 @app.post('/conversations/{conversation_id}/start')
 async def start_conversation(
     conversation_id: str,
+    providers_set: StartConversationRequest,
```
malhotra5:
```suggestion
    providers_set: ProvidersSetModel,
```

xingyaoww (approved):
LGTM!

PR #9122, opened 2025-06-13
feat: support dev container networking without host mode

llamantino:

Updates devcontainer and make docker-dev to fix runtime connection without requiring host mode.

---

Updates devcontainer.json and the compose file used by make docker-dev to fix runtime connection without requiring host mode.

---


enyst:
Looks like a better solution. Does it work well for you?

llamantino:
On Ubuntu (Docker Engine) and WSL2 (Docker Desktop), all good so far. This should also fix devcontainers/make docker-dev still being broken on macOS.

neubig:
Thank you!

PR #9148, opened 2025-06-16
Fix bash timeout issue caused by interactive git clone prompts

### LINKED ISSUE #9147: [Bug]: First bash commands timing out immediately

On the most recent version of OpenHands, I am seeing in some cases that Bash will time out immediately on the first run.

Here is an example repro that works much of the time:
1. Select `neubig/starter-repo` in the repo selector (you may need to fork the repo)
2. Choose the main branch and launch
3. Ask OpenHands "Refactor the linting workflow to use precommit instead of calling ruff and mypy separately." (basically anything should do)

Image

You'll see something like the following, where the first bash command timed out. I did this on main, but I have also seen this for other repos on the cloud on version 0.43.

### OpenHands Installation

Development workflow

### OpenHands Version

0.43, main

### Operating System

None

_No response_

neubig:

Fixes an issue where bash commands would timeout immediately on first run due to git clone operations hanging while waiting for user credentials. This was causing the "previous command still running" error that prevented users from executing any bash commands in OpenHands.

---

This PR addresses a critical issue where the system would hang when trying to clone organization-level microagent repositories that don't exist or require authentication. The problem occurred during runtime initialization when OpenHands attempted to clone `.openhands` repositories from GitHub organizations.

**Key changes:**

1. **Added `GIT_TERMINAL_PROMPT=0`** to the git clone command to prevent interactive credential prompts
   - This ensures git fails quickly with a clear error message instead of hanging indefinitely
   - Allows the error handling logic to work properly

2. **Fixed URL formatting in `_get_authenticated_git_url`** method
   - Removed incorrect domain stripping that was creating malformed URLs
   - Now preserves the full `github.com/user/repo` path in authenticated URLs
   - Ensures proper URL format: `https://token@github.com/user/repo.git`

3. **Added comprehensive tests** for the `_get_authenticated_git_url` method
   - Tests GitHub and GitLab URL formatting with and without tokens
   - Ensures the fix doesn't break existing functionality

**Design decisions:**
- Used `GIT_TERMINAL_PROMPT=0` instead of other approaches (like `--quiet`) because it specifically addresses the interactive prompt issue
- Preserved existing error handling logic - the fix ensures it can actually execute instead of hanging
- Added tests to prevent regression of URL formatting issues

---

Fixes #9147

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/7f8d4797062847b3b7c987937d3a0bda)

### DIFF
--- a/openhands/runtime/base.py
+++ b/openhands/runtime/base.py
@@ -657,9 +657,13 @@ fi
             git_token = self.git_provider_tokens[provider].token
             if git_token:
                 if provider == ProviderType.GITLAB:
-                    remote_url = f'https://oauth2:{git_token.get_secret_value()}@{repo_path.replace("gitlab.com/", "")}.git'
+                    remote_url = (
+                        f'https://oauth2:{git_token.get_secret_value()}@{repo_path}.git'
+                    )
                 else:
-                    remote_url = f'https://{git_token.get_secret_value()}@{repo_path.replace("github.com/", "")}.git'
+                    remote_url = (
+                        f'https://{git_token.get_secret_value()}@{repo_path}.git'
+                    )
 
         return remote_url
 
@@ -706,7 +710,9 @@ fi
             # Get authenticated URL and do a shallow clone (--depth 1) for efficiency
             remote_url = self._get_authenticated_git_url(org_openhands_repo)
 
-            clone_cmd = f'git clone --depth 1 {remote_url} {org_repo_dir}'
+            clone_cmd = (
+                f'GIT_TERMINAL_PROMPT=0 git clone --depth 1 {remote_url} {org_repo_dir}'
+            )
 
             action = CmdRunAction(command=clone_cmd)
             obs = self.run_action(action)
--- a/tests/unit/test_runtime_git_tokens.py
+++ b/tests/unit/test_runtime_git_tokens.py
@@ -455,3 +455,59 @@ async def test_clone_or_init_repo_with_branch(temp_dir, monkeypatch):
     assert 'git checkout feature-branch' in checkout_cmd
     assert 'git checkout -b' not in checkout_cmd  # Should not create a new branch
     assert result == 'repo'
+
+
+def test_get_authenticated_git_url_github_with_token(runtime):
+    """Test _get_authenticated_git_url method for GitHub with token"""
+    # Test the method directly
+    repo_path = 'github.com/neubig/.openhands'
+    result = runtime._get_authenticated_git_url(repo_path)
+
+    # Should include the token and preserve the full domain path
+    expected = 'https://test_token@github.com/neubig/.openhands.git'
+    assert result == expected
+
+
+def test_get_authenticated_git_url_github_without_token(temp_dir):
+    """Test _get_authenticated_git_url method for GitHub without token"""
+    config = OpenHandsConfig()
+    file_store = get_file_store('local', temp_dir)
+    event_stream = EventStream('abc', file_store)
+    runtime = TestRuntime(
+        config=config,
+        event_stream=event_stream,
+        sid='test',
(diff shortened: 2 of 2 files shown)


neubig:
@openhands
`self.git_provider_tokens = git_provider_tokens` in the clone repo function is an antipattern because it is a function with unexpected side effects. Instead, find out why the runtime git provider tokens are not set at the time that this function is called for repo microagents.

openhands/runtime/base.py
```diff
@@ -657,9 +657,13 @@ def _get_authenticated_git_url(self, repo_path: str) -> str:
             git_token = self.git_provider_tokens[provider].token
             if git_token:
                 if provider == ProviderType.GITLAB:
-                    remote_url = f'https://oauth2:{git_token.get_secret_value()}@{repo_path.replace("gitlab.com/", "")}.git'
+                    remote_url = (
+                        f'https://oauth2:{git_token.get_secret_value()}@{repo_path}.git'
+                    )
```
neubig:
```suggestion
                    remote_url = f'https://oauth2:{git_token.get_secret_value()}@{repo_path.replace("gitlab.com/", "")}.git'
```

openhands/runtime/base.py
```diff
@@ -657,9 +657,13 @@ def _get_authenticated_git_url(self, repo_path: str) -> str:
...
                 else:
-                    remote_url = f'https://{git_token.get_secret_value()}@{repo_path.replace("github.com/", "")}.git'
+                    remote_url = (
+                        f'https://{git_token.get_secret_value()}@{repo_path}.git'
+                    )
```
neubig:
```suggestion
                    remote_url = f'https://{git_token.get_secret_value()}@{repo_path.replace("github.com/", "")}.git'
```

tests/unit/test_runtime_git_tokens.py
```diff
@@ -455,3 +455,59 @@ async def test_clone_or_init_repo_with_branch(temp_dir, monkeypatch):
...
+        user_id='test_user',
+        git_provider_tokens=git_provider_tokens,
+    )
+
+    repo_path = 'gitlab.com/group/.openhands'
+    result = runtime._get_authenticated_git_url(repo_path)
+
+    # Should include the oauth2 token and preserve the full domain path
+    expected = 'https://oauth2:gitlab_token@gitlab.com/group/.openhands.git'
+    assert result == expected
```
neubig:
```suggestion
```

ryanhoangt:
I think the root cause is indeed the git clone waiting for credentials, below is the log of the last command (before agent's commands) I see inside the runtime:

<pre style="white-space: pre-wrap; word-wrap: break-word;">
12:01:07 - openhands:DEBUG: bash.py:602 - END OF PANE CONTENT: ['  "exit_code": "0",', '  "username": "openhands",', '  
"hostname": "hoang-dev",', '  "working_dir": "/workspace/starter-repo",', '  "py_interpreter_path": 
"/openhands/poetry/openhands-ai-5O4_aCHf-py3.12/bin/python"', '}', '###PS1END###', 'git clone --depth 1 
https://github.com/ryanhoangt/.openhands.git /workspace/org_openhands_ryanhoangt', "Cloning into 
'/workspace/org_openhands_ryanhoangt'...", "Username for 'https://github.com':"]
12:01:07 - openhands:DEBUG: bash.py:632 - CHECKING NO CHANGE TIMEOUT (10s): elapsed 10.100644588470459. 
Action blocking: False
</pre>


One thing I don't quite understand yet is that it only happens for the forked `starter-repo` repo (as described in https://github.com/All-Hands-AI/OpenHands/issues/9147), I tested a few of my other (non-forked) repos and the issue doesn't occur there.

rbren (approved):

PR #9152, opened 2025-06-16
Fix openhands.core.exceptions.FunctionCallConversionError `fn_call_converter` for GPT-o4-mini when the agent generates images

TommasoBendinelli:
This pull request fixes the error:
`**openhands.core.exceptions.FunctionCallConversionError: Expected str or list with one text item when tool result is present in the message.** Content: [{'type': 'text', 'text': 'EXECUTION RESULT of [execute_ipython_cell]:\n<Figure size 640x480 with 1 Axes>This scatter plot contains two points at coordinates (0,0) and (1,1).\n[Jupyter current working directory: /workspace]\n[Jupyter Python interpreter: /openhands/poetry/openhands-ai-5O4_aCHf-py3.12/bin/python]'}, {'type': 'image_url', 'image_url': {'url': ...`

which is raised when running an agent with gpt-o4-mini and native function calling.


xingyaoww (approved):
Thanks!

PR #9153, opened 2025-06-16
Add CLI arguments for local LLM configuration

neubig:
## Summary

This PR adds CLI arguments to enable users to configure local models like devstral directly through command-line arguments, addressing a gap in the current CLI functionality.

## Problem

Previously, users wanting to use local models like devstral through the CLI had to:
- Use the interactive settings flow (`/settings` command)
- Set environment variables (`LLM_MODEL`, `LLM_BASE_URL`, `LLM_API_KEY`)
- Manually edit config files

There was no way to configure LLM settings directly via CLI arguments.

## Solution

Added three new CLI arguments:
- `--llm-model`: LLM model to use (e.g., "lm_studio/devstral", "openai/gpt-4")
- `--llm-base-url`: Base URL for LLM API (required for local models, e.g., "http://localhost:1234/v1")
- `--llm-api-key`: API key for LLM (use "dummy" for local models)

## Usage Example

Users can now configure local models directly:
```bash
openhands --llm-model "lm_studio/devstral" --llm-base-url "http://localhost:1234/v1" --llm-api-key "dummy" --task "echo hello world"
```

## Changes Made

1. **Added CLI arguments** in `get_parser()` function in `openhands/core/config/utils.py`
2. **Added configuration logic** in `setup_config_from_args()` function to apply the CLI arguments to the LLM configuration
3. **Maintained backward compatibility** - existing configuration methods still work

## Testing

- ✅ CLI arguments are properly parsed
- ✅ Configuration is correctly applied to LLM config
- ✅ Backward compatibility maintained
- ✅ Pre-commit hooks pass

This change makes it much more convenient for users to quickly test local models without needing to modify configuration files or go through interactive setup flows.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/83efa4ce397e4e6fbcd42f93b476c687)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:4cbe46b-nikolaik   --name openhands-app-4cbe46b   docker.all-hands.dev/all-hands-ai/openhands:4cbe46b
```

### DIFF
--- a/openhands/core/config/utils.py
+++ b/openhands/core/config/utils.py
@@ -744,6 +744,27 @@ def get_parser() -> argparse.ArgumentParser:
         type=bool,
         default=False,
     )
+
+    # LLM configuration arguments for local models
+    parser.add_argument(
+        '--llm-model',
+        help='LLM model to use (e.g., "lm_studio/devstral", "openai/gpt-4")',
+        type=str,
+        default=None,
+    )
+    parser.add_argument(
+        '--llm-base-url',
+        help='Base URL for LLM API (required for local models, e.g., "http://localhost:1234/v1")',
+        type=str,
+        default=None,
+    )
+    parser.add_argument(
+        '--llm-api-key',
+        help='API key for LLM (use "dummy" for local models)',
+        type=str,
+        default=None,
+    )
+
     return parser
 
 
@@ -819,6 +840,21 @@ def setup_config_from_args(args: argparse.Namespace) -> OpenHandsConfig:
             llm_config = config.llms[args.llm_config]
         if llm_config is None:
             raise ValueError(f'Invalid toml file, cannot read {args.llm_config}')
+        config.set_llm_config(llm_config)
+
+    # Override LLM settings with direct CLI arguments
+    if args.llm_model or args.llm_base_url or args.llm_api_key:
+        from pydantic import SecretStr
+
+        llm_config = config.get_llm_config()
+
+        if args.llm_model:
+            llm_config.model = args.llm_model
+        if args.llm_base_url:
+            llm_config.base_url = args.llm_base_url
+        if args.llm_api_key:
+            llm_config.api_key = SecretStr(args.llm_api_key)
+
         config.set_llm_config(llm_config)
 
     # Override default agent if provided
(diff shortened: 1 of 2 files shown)


enyst:
@neubig I’d note, this

> There was no way to configure LLM settings directly via CLI arguments.

is not actually the case. In CLI and headless, the user can configure LLM in `config.toml` using [named configurations](https://github.com/All-Hands-AI/OpenHands/blob/2fd1fdcd7ed6ed20f879be64b05f4ed327d003d9/config.template.toml#L205), and then just choose with a command line parameter `-l`.

e.g. I use `python -m openhands.cli.main -l gemini-pro`
 where “gemini-pro” is already defined with all properties it may need.

neubig:
@openhands check the failing unit tests by running py-unit-tests.yml step by step and try to fix them

enyst:
I just tried, and it looks like this works with `openhands` from pypi:
- save an LLM configuration under whatever name (or no name) in a config.toml placed anywhere
```
[llm.devstral]
model="lm_studio/devstral"
base_url "http://localhost:1234/v1”
api_key "dummy"
```
- run `openhands --config-file /path/to/config.toml --task "test task" -l devstral`

neubig:
That's a good to have a work around, but I think being able to configure it through the CLI is preferable, so I'll try to get this set up. Haven't had a chance to test yet.

enyst:
I admit I don't see it as a workaround, more like the normal way, which allows to get all hyperparameters desired.

There already is another, too: using env, the same vars from env. So the command can look like `LLM_BASE_URL=... openhands ...`. Or just in `.env`.

I think the env option will not work with CLI at the moment, but the reason is the same with why this PR might not work: because we [require](https://github.com/All-Hands-AI/OpenHands/blob/075ef4db9f9ee24f988d72f88ca489c06fb4868c/openhands/cli/main.py#L369) the use of "settings", and need to hardcode in if we want to override _those_. To add an override, we need to decide what the priority order is.

More importantly, do we really want to recommend API keys on the command line? I feel like sooner or later, that terminal will be read by an LLM/agent, and get leaked in places the user didn't intend. Seems easy if it's a vscode terminal.

neubig:
Maybe I'll close this, I'm opening too many PRs and this one may not be essential :)

PR #9170, opened 2025-06-16
Fix Devstral system prompt issue with Ollama

### LINKED ISSUE #8955: [Bug]: The agent (Devstral) frequently forgets that it's an agent and starts acting like a typical LLM

Here is an example of what should be a very simple prompt for Devstral that will usually fail to create any files, instead telling you to create the files yourself:

> I created a basic repo for you to use by running the command `npx create-react-app our-components --template typescript`
>
>I want to create a variety of components and I want to have the main App page be a showcase for the various components, but let's not get ahead of ourselves. I want to simply start with you making a nice button component, with all the usual attributes and features the button component would have. Use the subdirectory components and get started.

When it fails to act as an agent, even prompting it with something like `I need you to create the files yourself.` is not able to put it back on track.

### OpenHands Installation

Docker command in README

### OpenHands Version

docker.all-hands.dev/all-hands-ai/runtime:0.41-nikolaik

### Operating System

MacOS

neubig:
## Problem

When using Devstral models with Ollama, the model fails to behave as an agentic coding assistant and instead acts like a generic chat model. This happens because Ollama's Go template system doesn't apply the default system prompt that Devstral expects, unlike LMStudio which properly applies the Jinja template with a default system message.

## Root Cause

- **LMStudio (Works Correctly)**: Uses Jinja templates that include a `default_system_message` when no system message is provided
- **Ollama (Broken)**: Uses Go templates that don't include a default system message mechanism
- **Devstral Requirement**: Devstral is specifically trained to work as an agentic coding model and requires its specific system prompt to function correctly

## Solution

This PR automatically detects when:
1. A Devstral model is being used (`devstral` in model name)
2. Ollama is the LLM provider (detected via `base_url` or `custom_llm_provider`)
3. No proper Devstral system prompt is already present

When these conditions are met, it automatically injects the official Devstral system prompt from the model repository.

## Changes

- **`openhands/llm/devstral_utils.py`** (New): Utility functions for detecting Devstral/Ollama usage and injecting the system prompt
- **`openhands/llm/llm.py`** (Modified): Integrated the fix into the LLM wrapper to automatically apply the system prompt when needed
- **`tests/unit/test_devstral_utils.py`** (New): Comprehensive tests for all scenarios
- **`DEVSTRAL_OLLAMA_FIX.md`** (New): Detailed documentation explaining the problem and solution

## Compatibility

- **LMStudio**: No impact (already works correctly)
- **Other providers**: No impact (only applies to Ollama + Devstral)
- **Non-Devstral models**: No impact (only applies to Devstral models)
- **Existing system prompts**: If a system prompt containing "Devstral" is already present, no injection occurs

## Testing

All tests pass:
```bash
poetry run pytest tests/unit/test_devstral_utils.py -v
```

The fix is automatic and transparent - users don't need to change anything in their configuration.

Fixes #8955

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/09ccc0d2a28644c589dea11e3178384c)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:d28bf22-nikolaik   --name openhands-app-d28bf22   docker.all-hands.dev/all-hands-ai/openhands:d28bf22
```

### DIFF
--- a/DEVSTRAL_OLLAMA_FIX.md
+++ b/DEVSTRAL_OLLAMA_FIX.md
@@ -1,0 +1,109 @@
+# Devstral + Ollama System Prompt Fix
+
+## Problem Description
+
+When using Devstral models with Ollama, the model fails to behave as an agentic coding assistant and instead acts like a generic chat model. This happens because Ollama's Go template system doesn't apply the default system prompt that Devstral expects, unlike LMStudio which properly applies the Jinja template with a default system message.
+
+### Root Cause
+
+1. **LMStudio (Works Correctly)**: Uses Jinja templates that include a `default_system_message` when no system message is provided. This ensures Devstral gets its proper system prompt.
+
+2. **Ollama (Broken)**: Uses Go templates that don't include a default system message mechanism. The template only applies system prompts if they're explicitly provided in the messages.
+
+3. **Devstral Requirement**: Devstral is specifically trained to work as an agentic coding model and requires its specific system prompt to function correctly.
+
+## Solution
+
+This fix automatically detects when:
+1. A Devstral model is being used (`devstral` in model name)
+2. Ollama is the LLM provider (detected via `base_url` containing `:11434` or `ollama`, or `custom_llm_provider` containing `ollama`)
+3. No proper Devstral system prompt is already present
+
+When these conditions are met, it automatically injects the official Devstral system prompt.
+
+## Implementation
+
+### Files Added/Modified
+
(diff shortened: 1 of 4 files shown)


enyst:
Oh, interesting. This sounds like an Ollama problem, rather than only Devstral 🤔

neubig:
Yeah, I don't think this is a good fix (openhands got a bit too eager), but we'll continue debugging.

Issue #9127, opened 2025-06-14
[Bug]: Modal runtime fails with `Server disconnected without sending a response`


arun-mani-j:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Modal runtime fails to launch. However the runtime itself is properly launched in the Modal website.

To reproduce, run OpenHands server with Modal runtime:
```sh
$ RUNTIME=modal MODAL_API_TOKEN_ID=token MODAL_API_TOKEN_SECRET=secret uvicorn openhands.server.listen:app --host "0.0.0.0" --port "3000"
```

Run the frontend and start a new conversation. Then open Modal website and follow the logs of the sandbox. Once the sandbox is launched (might take time if the image has to be built), OpenHands server crashes with the following error:

```python
14:45:38 - openhands:ERROR: session.py:221 - Error creating agent_session: RetryError[<Future at 0x7f0c809ae900 state=finished raised RemoteProtocolError>]
Traceback (most recent call last):
  File "/home/arun-mani-j/Projects/OpenHands/.venv/lib/python3.12/site-packages/httpx/_transports/default.py", line 101, in map_httpcore_exceptions
    yield
  File "/home/arun-mani-j/Projects/OpenHands/.venv/lib/python3.12/site-packages/httpx/_transports/default.py", line 250, in handle_request
    resp = self._pool.handle_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arun-mani-j/Projects/OpenHands/.venv/lib/python3.12/site-packages/httpcore/_sync/connection_pool.py", line 256, in handle_request
    raise exc from None
  File "/home/arun-mani-j/Projects/OpenHands/.venv/lib/python3.12/site-packages/httpcore/_sync/connection_pool.py", line 236, in handle_request
    response = connection.handle_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/arun-mani-j/Projects/OpenHands/.venv/lib/python3.12/site-packages/httpcore/_sync/connection.py", line 103, in handle_request
    return self._connection.handle_request(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

Issue #8825, opened 2025-05-31
[Bug]: Server error '502 Bad Gateway' for url Daytona runtime


arun-mani-j:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

When running OpenHands through Daytona runtime, it fails to connect to the runtime with this error.

To reproduce, run the backend from the source code as:
```
RUNTIME=daytona DAYTONA_API_KEY=dtn_KEY uvicorn openhands.server.listen:app --host "0.0.0.0" --port "3000"
```

Run the frontend and open it. Click on "Launch From Scratch". 

Check the backend logs. 

```python
15:00:55 - openhands:WARNING: utils.py:309 - DEPRECATED: The WORKSPACE_BASE and WORKSPACE_MOUNT_PATH environment variables are deprecated. Please use RUNTIME_MOUNT instead, e.g. 'RUNTIME_MOUNT=/my/host/dir:/workspace:rw'
15:00:55 - openhands:INFO: server_config.py:53 - Using config class None
INFO:     Started server process [3375]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:3000 (Press CTRL+C to quit)
INFO:     127.0.0.1:49800 - "GET /api/options/config HTTP/1.1" 200 OK
INFO:     127.0.0.1:49800 - "GET /api/options/config HTTP/1.1" 200 OK
INFO:     127.0.0.1:49800 - "GET /api/settings HTTP/1.1" 404 Not Found
INFO:     127.0.0.1:49800 - "GET /api/options/models HTTP/1.1" 200 OK
INFO:     127.0.0.1:49800 - "GET /api/options/agents HTTP/1.1" 200 OK
INFO:     127.0.0.1:49800 - "GET /api/options/security-analyzers HTTP/1.1" 200 OK
INFO:     127.0.0.1:49800 - "GET /api/settings HTTP/1.1" 404 Not Found
INFO:     127.0.0.1:54040 - "OPTIONS /api/settings HTTP/1.1" 200 OK
INFO:     127.0.0.1:54040 - "POST /api/settings HTTP/1.1" 200 OK
INFO:     127.0.0.1:54040 - "GET /api/settings HTTP/1.1" 200 OK
[...]

PR #8420, opened 2025-05-10
Fix issue #8419: Document `get_impl` and `import_from`

openhands-agent:
The get_impl function is one of the extensibility mechanisms in the project. It allows applications built on openhands to customize behavior by redefining and substituting at runtime their own implementations of the classes that this function is used for.

The LLMs get confused by the current code, for example [here](https://github.com/All-Hands-AI/OpenHands/pull/8375/files#diff-e1941078d2690fce1763259acc7c0b3f3763dc77e2e66630051cdbf0372c5e78R39). I also had a confused agent recently.

This PR proposes to add some clarification, specially docstrings, to make it easier to understand. It was primarily meant for LLMs, but humans can benefit too from clarity and consistency. 😄 

---

This pull request fixes #8419.

The issue has been successfully resolved based on the following concrete changes and their impact:

1. Documentation:
- Added docstrings to both `get_impl` and `import_from` functions in `import_utils.py`, clearly explaining their purpose, usage, and type safety guarantees
- Added docstrings to `ConversationManager` and `StandaloneConversationManager` classes that use the extensibility mechanism
- Created a new `README.md` in the utils directory that provides a complete explanation of the runtime implementation substitution system

2. Implementation:
- The documentation clearly explains how to use the extensibility mechanism through concrete examples
- Covers all key aspects requested: how it works, how to customize behavior, and where it's used
- Includes code examples demonstrating practical usage

3. Documentation Placement:
- Correctly placed the main documentation in the utils directory since the mechanism is used across multiple modules, not just in server
- Added class-specific documentation where the mechanism is used (in conversation_manager classes)

The changes provide a clear, comprehensive, and properly placed documentation of the extensibility mechanism, making it easy for developers to understand and implement custom behaviors in OpenHands.

Automatic fix generated by [OpenHands](https://github.com/All-Hands-AI/OpenHands/) 🙌

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:b8135bb-nikolaik   --name openhands-app-b8135bb   docker.all-hands.dev/all-hands-ai/openhands:b8135bb
```

PR #9213, opened 2025-06-18
Remove third-party runtimes (daytona, modal, e2b, runloop) from main codebase

### LINKED ISSUE #9212: Move third-party runtimes to a separate repo

We have several third-party runtimes:
- `openhands/runtime/impl/daytona`
- `openhands/runtime/impl/modal`
- `openhands/runtime/impl/e2b`
- `openhands/runtime/impl/runloop`

These are not tested or used by the OpenHands team, and:
1. cause a significant number of extra dependencies in the project
2. break and confuse users when they don't work

We can move these to a separate repository to maintain them: https://github.com/All-Hands-AI/third-parties-runtimes
But remove them, and their dependencies from the main openhands repo.

neubig:
## Summary

This PR removes all third-party runtime implementations (Daytona, Modal, E2B, and Runloop) from the main OpenHands codebase as requested in issue #9212. These runtimes will be moved to a separate repository to reduce the maintenance burden and simplify the core codebase.

Related issue:
- https://github.com/All-Hands-AI/OpenHands/issues/9127

### Runtime Implementation Removal
- **Removed directories:**
  - `openhands/runtime/impl/daytona/`
  - `openhands/runtime/impl/modal/`
  - `openhands/runtime/impl/e2b/`
  - `openhands/runtime/impl/runloop/`
  - `containers/e2b-sandbox/`

### Code Updates
- **Updated `openhands/runtime/__init__.py`:**
  - Removed imports for third-party runtime classes
  - Removed third-party runtimes from `_DEFAULT_RUNTIME_CLASSES` registry
  - Updated `__all__` list to exclude third-party runtimes

- **Updated `openhands/runtime/impl/__init__.py`:**
  - Removed imports and exports for third-party runtime classes

- **Updated `tests/runtime/conftest.py`:**
  - Removed imports for third-party runtime classes
  - Updated `get_runtime_classes()` function to exclude third-party runtimes

### Dependency Cleanup
- **Updated `pyproject.toml`:**
  - Removed `e2b` dependency
  - Removed `modal` dependency
  - Removed `runloop-api-client` dependency
  - Removed `daytona` dependency

### Configuration Cleanup
- **Updated `openhands/core/config/openhands_config.py`:**
  - Removed `e2b_api_key` field
  - Removed `modal_api_token_id` and `modal_api_token_secret` fields
  - Removed `runloop_api_key` field
  - Removed `daytona_api_key`, `daytona_api_url`, and `daytona_target` fields

- **Updated `openhands/core/logger.py`:**
  - Removed third-party API key patterns from sensitive data filtering

### Documentation Updates
- **Removed documentation files:**
  - `docs/usage/runtimes/daytona.mdx`
  - `docs/usage/runtimes/e2b.mdx`
  - `docs/usage/runtimes/modal.mdx`
  - `docs/usage/runtimes/runloop.mdx`

- **Updated `docs/usage/runtimes/overview.mdx`:**
  - Removed references to third-party runtimes
  - Updated available runtimes list to only include core runtimes

- **Updated `docs/usage/configuration-options.mdx`:**
  - Removed API key configuration sections for third-party runtimes

### Testing
- **Added comprehensive test suite (`tests/unit/test_third_party_runtimes_removal.py`):**
  - Verifies third-party runtime classes are not available via `get_runtime_cls()`
  - Confirms import failures for removed runtime modules
  - Validates runtime registry only contains core runtimes
  - Ensures configuration fields are properly removed
  - All tests pass, confirming complete removal

## Remaining Core Runtimes

After this change, OpenHands will continue to support the following core runtimes:
- **Docker Runtime** (default) - Containerized execution with Docker
- **Remote Runtime** - Cloud-based runtime for parallel execution
- **Local Runtime** - Direct execution on local machine
- **CLI Runtime** - Command-line interface runtime

## Testing

- ✅ All new tests pass
- ✅ Pre-commit hooks pass
- ✅ Core runtime functionality verified
- ✅ No breaking changes to existing core runtime APIs

## Migration Path

Users currently using third-party runtimes will need to:
1. Wait for the separate third-party runtime repository to be created
2. Install the third-party runtimes from the new repository
3. Update their configuration to use the new runtime packages

Fixes #9212
Fixes https://github.com/All-Hands-AI/OpenHands/issues/9275

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/edcd025af61241a89726054011a3ae2b)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:b7fd448-nikolaik   --name openhands-app-b7fd448   docker.all-hands.dev/all-hands-ai/openhands:b7fd448
```


neubig:
@openhands

1. the third-party repo is created, so please use this to store the removed code https://github.com/All-Hands-AI/third-parties-runtimes
2. check github CI by using the API to download the logs, then read them to find failing tests. since we removed all the third party runtimes, we can remove tests for them

openhands/runtime/__init__.py
```diff
@@ -1,27 +1,19 @@
...
-from openhands.runtime.impl.runloop.runloop_runtime import RunloopRuntime
 from openhands.utils.import_utils import get_impl
 
 # mypy: disable-error-code="type-abstract"
 _DEFAULT_RUNTIME_CLASSES: dict[str, type[Runtime]] = {
     'eventstream': DockerRuntime,
     'docker': DockerRuntime,
-    'e2b': E2BRuntime,
     'remote': RemoteRuntime,
-    'modal': ModalRuntime,
```
enyst, line 21: Below, there is code loading a Runtime via `get_impl`, which I think should allow the user to load an implementation by specifying the full path. It’s not clear to me it will work, though 🤔 

We do want it to work, right?

tests/unit/test_third_party_runtimes_removal.py
enyst, line 1: IMHO we don’t need to test that they’re deleted. More coverage is nice, but IMHO we should test what we do, we don’t really need to test that we don’t do this or that feature 🤔

tests/unit/test_third_party_runtimes_removal.py
```diff
@@ -0,0 +1,94 @@
...
+    third_party_runtimes = ['daytona', 'modal', 'e2b', 'runloop']
+
+    for runtime_name in third_party_runtimes:
+        with pytest.raises(ValueError, match=f'Runtime {runtime_name} not supported'):
+            get_runtime_cls(runtime_name)
+
+
+def test_third_party_runtime_imports_fail():
+    """Test that importing third-party runtime classes fails."""
+    third_party_imports = [
```
enyst: 🤔
Maybe this could be the opposite, we could test that we try to import a user-specified runtime class?

docs/usage/runtimes/overview.mdx
```diff
@@ -21,6 +19,3 @@ OpenHands supports several different runtime environments:
 - [Docker Runtime](/usage/runtimes/docker) - The default runtime that uses Docker containers for isolation (recommended for most users).
 - [OpenHands Remote Runtime](/usage/runtimes/remote) - Cloud-based runtime for parallel execution (beta).
 - [Local Runtime](/usage/runtimes/local) - Direct execution on your local machine without Docker.
-- And more third-party runtimes:
-  - [Modal Runtime](/usage/runtimes/modal) - Runtime provided by our partners at Modal.
-  - [Daytona Runtime](/usage/runtimes/daytona) - Runtime provided by Daytona.
```
enyst, line 26: Maybe we can continue to document them, and how people can find and use a third party runtime?

enyst (commented):
I love where this is going, mainly for the reasons outlined in the linked issue. I’m a little surprised by the speed of this, although it’s also great.

Could we clarify the migration process?

Side note
> These are not tested or used by the OpenHands team

You’re right in general, but there is some exception, as it happens I was just looking at this detail so I’ll mention it: we have integrated an ability to use Modal for SWE-bench evals (and SWE-bench themselves have), and we need the remote evals workflow to use Modal. (That is only for grading, not inference, and so we don’t need the runtime client for it.)

More importantly, what are the steps of this migration
- what should users know, if they want to continue using e.g. daytona? -> a section in the doc
- what should runtime developers know?
- could we notify the current runtime developers that we are doing this, so that they can update their docs etc? In particular, I think Daytona folks are actively working on their runtime and have [already fixed the latest issue](https://github.com/All-Hands-AI/OpenHands/issues/8825#issuecomment-2927182452), and maybe we can notify others too.

enyst:
@idagelic I’d love your feedback on this move, I hope it doesn’t pose problems for you guys?

openhands/runtime/__init__.py
neubig: Added to the Usage part of this repo: https://github.com/All-Hands-AI/third-party-runtimes/blob/main/README.md#usage

tests/unit/test_third_party_runtimes_removal.py
neubig: Sorry, this was openhands being too enthusiastic :) I thought I had deleted this file, but maybe didn't push the final confirm button.

docs/usage/runtimes/overview.mdx
neubig: Fixed!

neubig:
Thanks @enyst !

I do understand that this may make things more inconvenient for developers with third-party runtimes, but over the past several weeks we've seen some users get bitten by this, particularly when using the CLI. I also did a refactoring of the runtime classes and it took me significant extra time to modify all of the four third-party runtimes. So basically, this is a trade-off and apologies to the third-time runtime providers, but I think it's still a good change.

docs/usage/runtimes/overview.mdx
```diff
@@ -21,6 +19,4 @@ OpenHands supports several different runtime environments:
...
+- [Third-party Runtimes](https://github.com/All-Hands-AI/third-party-runtimes) - These runtimes are not officially supported by OpenHands and may be out-of-date, but are provided for reference for users who would like to run on third-party infrastructure providers.
```
enyst:
```suggestion
- [Third-party Runtimes](https://github.com/All-Hands-AI/third-party-runtimes) - These runtimes are supported by their developers, not by OpenHands. Please find them in the repository linked here if you would like to run on third-party infrastructure providers.
```

openhands/runtime/__init__.py
enyst: Thank you! We may need more, I think line 31 below will only resolve if the class is available to python at runtime, and if it’s a fully qualified name 🤔

(Quick GPT-4o write up. Where could we place something [like this](https://gist.github.com/enyst/64a596412370bcb56d228d5f6b0f3c94)?)

enyst:
Just to clarify, 
> over the past several weeks we've seen some users get bitten by this, particularly when using the CLI

I have seen some errors due to
- https://github.com/All-Hands-AI/OpenHands/pull/9187 , which was fixed; and/or
- the need for the user to update/reinstall openhands package.

I do agree this PR is a good change! I just think people need a migration path, and I’d love if we get to notify runtime developers before we merge this.

As we can see in the link, developers of some runtimes are active and actively fixing issues. Also, the PR above shows multiple issues fixed in this repo, which is to say, a number of our users are using this runtime right now. 🙏 I left a small comment with a suggestion on what they’d need to do to go from the previous OH version to the version in which we merge this PR.

neubig:
OK, sounds good, let me give them a heads up. You've already contacted daytona so I'll do the others.

enyst:
> OK, sounds good, let me give them a heads up. You've already contacted daytona so I'll do the others.

I also notified Modal. @neubig

neubig:
Hi @mlejva, just a quick note that we're moving the third-party runtimes from the main openhands repo to a separate repo, including e2b, please note this if interested!

@enyst : I contacted modal and runloop separately on slack, so everyone should be aware now!

enyst (changes requested):
Thank you @neubig ! I love the move of runtime clients in their own repo, I’ll put a request for changes only so that we don’t merge this accidentally.

As far as I know, we have users who are actively using Daytona, at least, and whose reported issues have been fixed by the Daytona developers just this week. Maybe we can clarify their migration path, with documentation, before we release a new OH version without the clients. Since that can mean losing their conversations. Also, the Modal folks seem interested to keep the integration working. I’ll follow up on the runtimes repo!

neubig:
Hi @enyst , are there any other changes that you'd like to request? I think that this PR is basically finished and any other refinements we can make on the other repo?

openhands/runtime/__init__.py
enyst: @neubig Maybe in this file, at line 31 or just after, where we get an exception for unknown runtime, we could add a sentence to the error message, like 

“3rd party runtimes are no longer part of the default OpenHands installation. If you were using them, see https://github.com/All-Hands-AI/openhands-third-party-runtimes for the code and instructions.”

What do you think?

It’s because that users with “daytona” in config or command, for example, will get that exception.

arun-mani-j:
Thank you people for the quick actions (and thanks @idagelic for fixing the Daytona issue).

I think once this PR is stable, we should move the corresponding issues to the new repo. I think GitHub offers a way to do it, but if not, I can copy paste the issues there and close the ones here, no issues.

I think many of us require these third-party runtimes for multi-client setup. Like in academics where OH can be hosted on university server and let students get access to it. But running all the containers in the university server is slow, so we seek third party runtimes as many of them have generous pricing and also offer fast bootup times.

But at the same time, I think scalable deployment of OpenHands is a problem of the deployer and is beyond the scope of OpenHands and the README clearly states:
> OpenHands is meant to be run by a single user on their local workstation. It is not appropriate for multi-tenant deployments where multiple users share the same instance. There is no built-in authentication, isolation, or scalability.

> If you're interested in running OpenHands in a multi-tenant environment, please [get in touch with us](https://docs.google.com/forms/d/e/1FAIpQLSet3VbGaz8z32gW9Wm-Grl4jpt5WgMXPgJ4EDPVmCETCBpJtQ/viewform) for advanced deployment options.

Also it conflicts with the business alignment of OpenHands Cloud.

So I guess for the people who need to small-scalable setup, their best bet is to find out which of the third-party runtime has active updates and stick to it (unless the future changes and OpenHands docs share how to do multi-tenant setup through k8s etc.) :).

enyst:
Oh I see, thank you for sharing your thoughts @arun-mani-j ! As it happens, I gave quite a lot of thought to precisely this kind of small distributed-ish use case for OH, and I really appreciate your perspective.

On topic here, and out of curiosity, does daytona runtime, after the fixes, work for you, and how about from the [openhands 3rd party runtimes](https://github.com/All-Hands-AI/openhands-third-party-runtimes) repo?

Quick clarifications since the discussion implies more than we did before here
- refactoring the repos to move 3rd party runtimes is, IMO, a good move, regardless of any business interests or no business interests, of AH or of 3rd parties. It’s based on the code architecture, experience, and maintenance reasons outlined in the linked issue.
- but maybe we should offer client runtime developers a chance to give feedback on the transition
- I am not related to AH or their business, or any other company mentioned here. 
(unless we count that I love Robert and them all, but that’s a secret, so it doesn’t count, right 🫣)

I have two quick questions
- for educational use, sorry if it’s a stupid question, but why doesn’t it work like, each student runs OH, maybe with some specialized agents provided by the university, or maybe with a university MCP server or different tools?
- AH has just released [this k8s setup](https://github.com/All-Hands-AI/OpenHands-Cloud), does something like that actually addresses your desired use case too?

neubig:
Thanks both! And yeah, @arun-mani-j just to confirm this is primarily for maintenance reasons, not business reasons -- the runtimes being incompatible was causing some users to not be able to use the CLI, and I recently had to refactor the runtime and it took a fair amount of extra time to maintain all 7 runtimes we have 😅 

Note that the All Hands remote runtime is convenient to use for experiments (we use it for all of our experiments): https://runtime.all-hands.dev/

If you're at a university or other non-profit and cannot afford to pay for it, please reach out to us and we will grant any reasonable request! We provide $100 by default to any organization that asks, and can boost that if you are doing research on openhands.

Of course you are also welcome to use the third-party runtimes as well.

PS:
> I think once this PR is stable, we should move the corresponding issues to the new repo. I think GitHub offers a way to do it, but if not, I can copy paste the issues there and close the ones here, no issues.

Yes, there is a way to do this, I'll take a look through the existing issues and move them.

arun-mani-j:
> Oh I see, thank you for sharing your thoughts @arun-mani-j ! As it happens, I gave quite a lot of thought to precisely this kind of small distributed-ish use case for OH, and I really appreciate your perspective.
> 
> On topic here, and out of curiosity, does daytona runtime, after the fixes, work for you, and how about from the [openhands 3rd party runtimes](https://github.com/All-Hands-AI/openhands-third-party-runtimes) repo?

After using the new runtime image, the Daytona issue got resolved. I have not tried with the one after the repo move, but I guess it should not make a difference.

> Quick clarifications since the discussion implies more than we did before here
> 
>     * refactoring the repos to move 3rd party runtimes is, IMO, a good move, regardless of any business interests or no business interests, of AH or of 3rd parties. It’s based on the code architecture, experience, and maintenance reasons outlined in the linked issue.
> 
>     * but maybe we should offer client runtime developers a chance to give feedback on the transition

Yea it's completely fine, I'm in no-way opposing the separation. In fact, this is what many projects do. Having optional 3rd party integration will only increase the burden and slow down maintenance.

>     * I am not related to AH or their business, or any other company mentioned here.
>       (unless we count that I love Robert and them all, but that’s a secret, so it doesn’t count, right 🫣)

:smile: 

> I have two quick questions
> 
>     * for educational use, sorry if it’s a stupid question, but why doesn’t it work like, each student runs OH, maybe with some specialized agents provided by the university, or maybe with a university MCP server or different tools?

Ideally yes, but unfortunately lab machines tend to be poorly performing and it increases the maintenance burden. On the other hand, keeping it on a single server and letting the students access it over the frontend URL will keep it easy on the lab incharges part.

>     * AH has just released [this k8s setup](https://github.com/All-Hands-AI/OpenHands-Cloud), does something like that actually addresses your desired use case too?

Thanks I did not know this. Let me share it with the concerned person and will follow up if they need any help.

> Thanks both! And yeah, @arun-mani-j just to confirm this is primarily for maintenance reasons, not business reasons -- the runtimes being incompatible was causing some users to not be able to use the CLI, and I recently had to refactor the runtime and it took a fair amount of extra time to maintain all 7 runtimes we have 😅

Completely agreed, it also reduces extra dependencies (which is always a big win).

> Note that the All Hands remote runtime is convenient to use for experiments (we use it for all of our experiments): https://runtime.all-hands.dev/

Got it, will check it out.

> If you're at a university or other non-profit and cannot afford to pay for it, please reach out to us and we will grant any reasonable request! We provide $100 by default to any organization that asks, and can boost that if you are doing research on openhands.
> Of course you are also welcome to use the third-party runtimes as well.

Thank you very much for this. I will speak to the appropriate people and will ask them to get in touch.


> PS:
> 
> > I think once this PR is stable, we should move the corresponding issues to the new repo. I think GitHub offers a way to do it, but if not, I can copy paste the issues there and close the ones here, no issues.
> 
> Yes, there is a way to do this, I'll take a look through the existing issues and move them.

Cool, thank you!

ivan-burazin:
### Request to Reconsider Removal of Daytona Runtime from Core Repo

Hi all, appreciate the effort to simplify the codebase and reduce maintenance overhead. That said, I’d like to make a case for keeping the Daytona runtime in the main OpenHands repository rather than moving it to the third-party repo.

#### Why Daytona is Different:

- **Actively Maintained:** Daytona is not a one-off integration. We've consistently responded to reported issues (for example, [#9187 fixed this week](https://github.com/All-Hands-AI/OpenHands/pull/9187)) and are committed to maintaining compatibility with OpenHands core.

- **Production-Ready and Aligned:** The latest Daytona runtime image is fully up to date, version aligned, and passes the current OpenHands runtime interface tests.

- **Proven Adoption:** We're seeing a few hundred Daytona-powered sandboxes created by OpenHands every single day. This reflects real usage and stability in the field.

- **Willing to Collaborate:** We're happy to adopt any maintenance guidelines you propose, including owning our tests, pinning to runtime interfaces, and modularizing integration to reduce core maintenance overhead.

We fully understand the goal of reducing burden from stale runtimes. But Daytona is not in that category. If there are technical concerns, we’re ready to work through them collaboratively to ensure that keeping Daytona in the main repo does not slow development or introduce instability.

Thanks again for all the work on OpenHands. We hope to continue contributing as part of the core runtime ecosystem.

Happy to follow up.

Ivan

enyst:
@ivan-burazin Thank you, good points!

I think you touched on an important point here: from our perspective, we cannot, realistically, maintain the runtimes we don’t, ourselves, use on a regular basis. For example, in the past, I have fixed issues with E2B runtime when it broke, but then I didn’t keep doing it, and it broke again, for quite a while now.

That’s why I really appreciate that you answered the issues with Daytona when I tagged you. I love to know who to tag, in the first place (I have no good idea who to tag on runloop for example). And I really appreciate that you fixed the issues experienced by our users - people who are openhands users and daytona users.

(That’s also why we give merge access to the `openhands-third-party-runtimes` repo to maintainers of the runtimes themselves, but I’ll get back to that)

I need, though, your help to understand another aspect: why does it matter in what repo the client is, from your and our users perspective. And please forgive if I’m asking silly questions. I tried to ask them before in this thread, and I don’t feel we have clarified all answers. 😅 🤷‍♂️

For users:
- with the client here, they have instructions how to use OH with Daytona, they need to specifically set “daytona” in config (?), and of course set up on Daytona site; the readme also points to an openhands script on Daytona site (could you tell what is it doing?)
- with the client in the other repo, they’d need to clone and install the repo, set in config the fully qualified name (of daytona client instead of “daytona”), and the rest is the same. Is that correct? Please do correct me!

Whether we do this for stale clients, as you said, or all clients, as we considered in this PR, it’s a definite goal to make it as seamless as possible for users. In fact, couldn’t it be practically the same thing?

—

Let me expand on why this matters, beyond the obvious - obviously we want people to have a good experience -, but what I have in mind is a bit more.

Last I checked, `openhands` repo had almost 5 million tokens. That’s a large repo. So large that there are some tools out there who automatically read github repos and try to adapt them for LLMs consumption, and they ALL fail on this repo. (I need to fix that in some way).

For a project of this size, it bothers me that OH doesn’t yet have great ways for programmatic customization. I mean, once a project grows, people create mechanisms like plugins, extensions, themes, for other developers to customize the experience, _without modifying the core_. We don’t have those; but we do have 2 main ways: runtimes, and agents. (and a small, incipient [third](https://github.com/All-Hands-AI/OpenHands/pull/8420))

So my thinking here is like: custom runtimes are extensions of OH. Like with any extension, the questions are, can people add a runtime without barriers? It seems like for the client, it’s an interface to implement, right? Is it well documented how to do it? Can we make it easier to add a server side, for those (like e2b) that are open source? Can users load a runtime of their choosing without a lot of trouble?

Runtimes are extensions. It shouldn’t matter in what repo they are. It matters, IMO, that people can install and use them easily; and that interested developers can add and maintain their runtime easily. What is not easy and can we fix it?

ivan-burazin:
Thanks for the reply, and appreciate you thinking this through.

Here’s my take. Moving the client to another repo doesn’t actually solve the issue of unmaintained integrations. If someone isn’t keeping their runtime up to date, **it’s going to break no matter where the code lives**. That’s not a repo problem, that’s a maintenance policy problem. If a runtime hasn’t been updated in X days and is clearly broken, it should just be removed. Otherwise, users end up confused and frustrated either way.

As for why it matters which repo it's in, I'd say it’s about friction. Every extra step you add makes it easier for someone to get lost. Having to clone a separate repo, follow different instructions, it all adds up. If it’s in the main repo, it’s easier to discover, easier to use, and easier to trust.

There’s also the open source side. Centralizing everything in one repo makes the whole project feel more active and coherent. You get more contributors, more visibility, and a better developer experience overall. Supabase is a great example of this. Their monorepo has become a strength because everything feels connected and well maintained.

So in short , moving things out doesn’t solve the actual problem, and it creates more friction for users and contributors. Better to keep it in one place, and have a clear policy for when something is no longer maintained.

Again, these are my two cents. It's your project, and you're allowed to do anything you want. 

But thanks for listening and hearing me out regardless.

neubig:
Hi @ivan-burazin, thanks for sharing these thoughts. But it's not a problem that the runtimes are broken (hopefully they're not, although some may be). It's that having the dependencies be in the main repo is causing extra burden for the maintainers and causing problems for our users.

Just to give another example. The dependency on Daytona is causing this deprecation error in our CLI, which we needed to fix here: https://github.com/All-Hands-AI/OpenHands/pull/9295

The dependency on e2b caused some other people to not be able to run the CLI earlier. We definitely want to reduce friction for people to use these runtimes (and we can improve the directions in the third party repo), but I do think that keeping them separate is probably the better move here.

enyst:
Actually, I think Ivan made many good points, and changed my mind. Maybe we can go about this differently?

How about:
- we keep Daytona in the main repo
- the integration is Daytona’s responsibility, as Ivan pointed out it already is, and offered to collaborate on maintenance 
- use `extras`, so that people can `pip install openhands-ai`, or `pip install openhands-ai[daytona]` or `pip install openhands-ai[all-runtimes]`
- the default docker app build includes Daytona.

Would this make sense?

I still hope I can convince you, @ivan-burazin , that OH is too big and we should make it an ecosystem rather than a “monorepo”, and if we do it right, it doesn’t have to introduce friction, but we probably need to make a potential migration seamless first. Telling users to do an untested `git clone` otherwise they lose their work is not seamless.

@neubig I’m sorry, you’re correct that more deps can bring issues, they always do. But I think we should seriously consider we can, and maybe should, rely on active contributors to this repo to share the maintenance burden with us, rather than feel we have to do it all ourselves. In this particular case, extras should handle it, and even if it didn’t, we can talk to Daytona folks when issues arise?

enyst:
I would also love to know what the Modal folks think, and others, and I hope we might during the week. (it was weekend)

ivan-burazin:
Thanks, @enyst, really appreciate you being open to the discussion and suggesting a practical way forward.

Your proposal sounds perfect to us. You can count on us to take full ownership of the Daytona integration. We’ll handle maintenance, version alignment, testing, and anything else needed to ensure we don’t create overhead or break stability.

Tagging my colleagues here: our @idagelic who created the integration and @vedranjukic out CTO, so you can tag any of us if there's anything you need, just as we’ve done with previous fixes.

We know weekends can be quiet, and not everyone has had a chance to respond yet. Just want to make it clear that our team is active and here to support, any day of the week.

neubig:
Hi @enyst and @ivan-burazin, thanks for the discussion!

I agree that using `[extras]` might be a good way to handle this. It solves one of the two big problems, which is that openhands may break for users.

It doesn't solve the other issue, which is that when we refactor we need to refactor as many files as we have runtimes. The main reason for this is because linting/type checking will note changes in the interface and throw errors that then need to be fixed by developers.

It's not a super-frequent occurrence, but it is frequent enough that it makes it creates significant barriers to progress. I can think of two ways to fix this:
1. the proposed solution of a separate repo. the separate repo can be marked as "confirmed to work with openhands 0.XX", and when the developers of external runtimes want to the upgrade their runtimes to work with the most recent version of openhands they could bump that version number.
2. we could keep the runtimes in this repo, but remove linting/typechecking on the third-party runtimes directory. so they may not work with the most recent version of openhands in this directory.
3. we could continue to pay the maintenance burden on the openhands developers side.

I still think that "1." is the best option here (although I acknowledge the tradeoffs), but happy to hear what others think.

ivan-burazin:
Thanks for the thoughtful breakdown @neubig, appreciate you laying out the options clearly.

From our side, we’re totally fine with option 2, keeping Daytona in the main repo but excluding it from linting and typechecking if that helps reduce friction for the OpenHands team.

We’re also obviously okay with option 3 if that’s the preferred route. Either way, you can count on us to maintain the Daytona runtime, keep it version aligned, and respond quickly to any breakages.

Of course, it's completely your call in the end and we’ll support whatever direction you choose.

neubig:
@openhands make a draft of option "2." listed by neubig above. this would include (1) reinstating all files that were removed by this PR, but moving them to a "third_party" directory and excluding linting on that directory, (2) adding a flag "third_party_runtimes" to the install script and only installing the libraries if this flag is active, (3) testing that install, linting, and testing work both when "third_party_runtimes" is specified and when it is not specified.

docs/usage/runtimes/overview.mdx
```diff
@@ -21,6 +19,18 @@ OpenHands supports several different runtime environments:
...
+
+### Third-Party Runtimes
+
+The following third-party runtimes are available when you install the `third_party_runtimes` extra:
+
+```bash
+pip install openhands-ai[third_party_runtimes]
+```
+
+- [E2B Runtime](/usage/runtimes/e2b) - Cloud-based runtime using E2B sandboxes.
```
enyst:
```suggestion
- [E2B Runtime](/usage/runtimes/e2b) - Open source runtime using E2B sandboxes.
```

openhands/core/config/utils.py
```diff
@@ -39,6 +39,87 @@
...
+        from openhands.core.config.third_party_config import (
+            discover_third_party_runtime_configs,
+        )
+
+        runtime_configs = discover_third_party_runtime_configs()
+
+        for runtime_name, config_spec in runtime_configs.items():
+            for field_name, field_spec in config_spec.items():
+                full_field_name = f'{runtime_name}_{field_name}'
+                env_var_name = f'OPENHANDS_{full_field_name.upper()}'
```
enyst: This is almost certainly incorrect, due to the LLM assuming patterns where they don’t exist: there is no `OPENHANDS_` in OH.
In addition, this code might be problematic to introduce because while it does nothing for OH, it might find things in the Cloud environment.

enyst: I would suggest to maybe not try to read from env at all. Runtimes definitely have env vars, but maybe it’s okay if we don’t map them, and let them map them.

enyst:
The agent’s attempt is understandable, but I think the code trying to read and map runtimes’ specific vars from toml and env is too complex and nowhere close to straightforward to work with or test. I wonder if we really should map them as config at all.

I looked the other day at Daytona and it seemed to me that it doesn’t use the `config` module vars at all? It uses env vars as env vars, not as part of OH configuration. Please correct me if I’m missing something there.

What if runtimes read them from env, simply, how terrible of an idea is this?

neubig:
@openhands
1. Please do not add any additional stuff to the configuration. Just have each third-party runtime read information directly from environment variables.
2. Fix all merge conflicts.
3. Check to make sure that the unit testing workflows pass. Push to GitHub, wait 120 seconds, and then check if GitHub CI passes. If it does not, then fix any errors.

docs/usage/configuration-options.mdx
```diff
@@ -12,21 +12,28 @@ description: This page outlines all available configuration options for OpenHand
 
 The core configuration options are defined in the `[core]` section of the `config.toml` file.
 
-### API Keys
-- `e2b_api_key`
-  - Type: `str`
-  - Default: `""`
-  - Description: API key for E2B
+### Third-Party Runtime Configuration
```
neubig: Delete the Third-party Runtime Configuration here, in favor of keeping it in the third_party directory

config.template.toml
```diff
@@ -10,18 +10,22 @@
...
-# API key for Modal
-#modal_api_token_id = ""
-#modal_api_token_secret = ""
-
-# API key for Daytona
-#daytona_api_key = ""
-
-# Daytona Target
-#daytona_target = ""
+# Third-party runtime API keys (only available when third_party_runtimes extra is installed)
```
neubig: Delete these comments here.

openhands/core/config/utils.py
```diff
@@ -39,6 +39,40 @@
 load_dotenv()
 
 
+def _load_third_party_config_from_env(
```
neubig: delete these functions since they are not used anymore

neubig:
@openhands read all of the comments in the most recent review by neubig and fix them. Once you have done them, run the runtime unit testing workflow to make sure that everything passes. Once you have done that, push to the GitHub branch, wait 120 seconds, and then check the GitHub CI through the GitHub API. If things are failing, reproduce the error locally, fix and wait again until everything is fixed.

neubig:
OK, this LGTM now I think! @enyst would be happy for another review.

tests/unit/test_config.py
```diff
@@ -990,34 +990,18 @@ def test_api_keys_repr_str():
...
-    assert 'my_modal_api_token_id' not in str(app_config)
-    assert 'my_modal_api_token_secret' not in repr(app_config)
-    assert 'my_modal_api_token_secret' not in str(app_config)
-    assert 'my_runloop_api_key' not in repr(app_config)
-    assert 'my_runloop_api_key' not in str(app_config)
-    assert 'my_daytona_api_key' not in repr(app_config)
-    assert 'my_daytona_api_key' not in str(app_config)
+
+    # Third-party runtime configuration fields are no longer part of OpenHandsConfig
+    # They are now read directly from environment variables by the runtime implementations
```
enyst:
```suggestion
```

tests/unit/test_config.py
```diff
@@ -990,34 +990,16 @@ def test_api_keys_repr_str():
...
-    assert 'my_runloop_api_key' not in str(app_config)
-    assert 'my_daytona_api_key' not in repr(app_config)
-    assert 'my_daytona_api_key' not in str(app_config)
+
+    assert 'my_search_api_key' not in repr(app_config)
+    assert 'my_search_api_key' not in str(app_config)
 
     # Check that no other attrs in OpenHandsConfig have 'key' or 'token' in their name
     # This will fail when new attrs are added, and attract attention
+    # Note: Third-party runtime fields are no longer part of OpenHandsConfig
```
enyst:
```suggestion
```

enyst (approved):
Thank you!

I tested with [this script](https://gist.github.com/enyst/9b8ccc088f4044882fee9bd16b7b87b3).

@ivan-burazin I think once a runtime is loaded normally, like above, it should also work normally.

Zacharias030:
@neubig the third party repo is a 404.
Where did the runtimes go? :)

Cheers,
Zach

enyst:
They’re here in the main repo:
https://github.com/All-Hands-AI/OpenHands/tree/main/third_party

PR #9248, opened 2025-06-20
Addition of A2A Protocol Support to OpenHands – Expansion of Agent-to-Agent Communication

f-miyake:

This pull request introduces support for the Agent-to-Agent (A2A) protocol in OpenHands.
The A2A protocol defines a standardized interface that enables agents to receive tasks and return results.

This implementation introduces the following basic API methods.
Instructions sent via `message/send` can be polled for results using `tasks/get`.

### Supported API Endpoints:
These endpoints correspond to sections [7.1](https://google-a2a.github.io/A2A/latest/specification/#71-messagesend), [7.3](https://google-a2a.github.io/A2A/latest/specification/#73-tasksget), and [7.4](https://google-a2a.github.io/A2A/latest/specification/#74-taskscancel) of the A2A specification.   
Other methods described in [Section 7](https://google-a2a.github.io/A2A/latest/specification/#7-protocol-rpc-methods) are optional and are not included in this PR.

- `/.well-known/agent.json`  
  Returns metadata about the agents available on this OpenHands instance.

- `message/send`  
  Sends a message to a specified agent for execution.

- `tasks/get`  
  Retrieves the status or result of a previously submitted task (used for polling).

- `tasks/cancel`  
  Cancels a task that is currently running or pending.

Note: Optional capabilities such as stream and pushNotification defined in the A2A protocol are not included in this implementation. 

---
This change introduces server-side functionality for communicating with remote services that support the A2A protocol. It allows OpenHands to interact with external agents through a well-defined interface.

- Agents can be discovered by sending a request to the following endpoint:
`http://localhost:3000/a2a/.well-known/agent.json`

- When agent information is specified in the metadata field, the corresponding agent is executed based on its identifier. If omitted, `CodeActAgent` is used as the default.

Example request:

```
POST http://localhost:3000/a2a
Content-Type: application/json

{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "message/send",
  "params": {
    "message": {
      "messageId": "msg-001",
      "role": "user",
      "parts": [
        {
          "kind": "text",
          "text": "Generate a Fibonacci sequence using Python."
        }
      ]
    },
    "metadata": {"agent": "CodeActAgent"}
  }
}
```

- The implementation follows the existing Control Flow model described in openhands/README.md, and enables A2A-compliant requests to trigger execution of agents via EventStream.

### Future Note:
The current A2A specification version is 0.2.1. Once the version 1.0.0 is officially released, additional adaptation is going to be required to ensure full compliance.

---


openhands/server/app.py
```diff
@@ -70,4 +67,5 @@ async def _lifespan(app: FastAPI) -> AsyncIterator[None]:
 app.include_router(secrets_router)
 app.include_router(git_api_router)
 app.include_router(trajectory_router)
+app.mount('/a2a', a2a_router)
```
neubig: This is a different pattern than the others, is there a reason for this? If not, please follow the standard pattern.

openhands/server/app.py
```diff
@@ -46,7 +44,7 @@ async def combined_lifespan(app):
 
 
 @asynccontextmanager
-async def _lifespan(app: FastAPI) -> AsyncIterator[None]:
+async def _lifespan(app: FastAPI):
```
neubig: Please maintain the return type annotation.

neubig:
Ah, I am not very familiar with the a2a protocol, but is there any mechanism to prevent people from just accessing the endpoint and starting agents working? i.e. authentication?

pcuci:
@f-miyake adding an example on how one might use this could go a long way to help us wrap our heads around the use-case you had in mind?

I imagine this will allow one OpenHands agent to talk to another OpenHands agent via A2A, which would be quite amazing!

E.g.:

OpenHands "I'm stuck going in the loop, I don't know what to work on next, maybe I'm partially done, I'm thinking of stopping", A2A request to Planner OpenHands (maybe with o3/Gemini for brain): "Here's a summary of recent activity, what should I work on next?"

openhands/server/a2a/a2a_request_handler.py
```diff
@@ -0,0 +1,331 @@
...
+    config,
+    conversation_manager,
+    file_store,
+    server_config,
+)
+from openhands.server.types import AppMode
+from openhands.storage.data_models.user_secrets import UserSecrets
+
+
+@trace_class(kind=SpanKind.SERVER)
```
enyst, line 48: What telemetry lib is this? It looks like it was included in a2a ?

openhands/server/app.py
ghost: Thank you for your feedback.
There was no particular reason for using a different pattern, so I updated the code to follow the standard approach and now use app.include_router(a2a_router) for consistency with the other routers.
The commit of this change is [here](https://github.com/All-Hands-AI/OpenHands/pull/9248/commits/9b01c0221830109a695e381f660e489dab94beb8).

openhands/server/app.py
ghost: Thank you for pointing this out.
Sorry, I accidentally removed the return type annotation earlier.
I've restored it in the latest [commit](https://github.com/All-Hands-AI/OpenHands/pull/9248/commits/02b834720e266a1b1218662811b56221a10f8398).

openhands/server/a2a/a2a_request_handler.py
ghost: Thank you for your comment. These are part of the a2a-sdk library. 
(https://google-a2a.github.io/A2A/latest/sdk/python/#a2a.utils.telemetry)

ghost:
> Ah, I am not very familiar with the a2a protocol, but is there any mechanism to prevent people from just accessing the endpoint and starting agents working? i.e. authentication?

The A2A Protocol itself defines fileds for specifying user authentication methods, and coressponding classes are partially implemented in the SDK.
However, the actual authentication logic needs to be implemented by developers. 
Since I'm not very familiar with implementing user authentication, I haven't included it in this PR.

ghost:
> @f-miyake adding an example on how one might use this could go a long way to help us wrap our heads around the use-case you had in mind?

Yes, asolutely. I'll update the PR description later to include use cases.
One expected senario is that a general-purpose agent invokes OpenHands externally as a coding expert.
> I imagine this will allow one OpenHands agent to talk to another OpenHands agent via A2A, which would be quite amazing!

Also, when combined with [this PR](https://github.com/All-Hands-AI/OpenHands/pull/6815), I believe the interaction you're imagining, such as a collaboration between an OpenHands Coder and an OpenHands Planner, would definitely be possible.

openhands/server/a2a/a2a_request_handler.py
enyst: @raymyers you may find this interesting, it seems it's open telemetry, via a2a

raymyers: Ah interesting. And seems we already had the otel sdk via #4214 that added a direct dependency for reasons that are unclear to me. Probably we had it transitively before that because of google cloud libs.

neubig:
Hi @f-miyake : I guess the situation that I'm most worried about is this:

One of our users starts OpenHands on their own personal server using LocalRuntime (so no sandboxing). There is no firewall separating them from the outside internet. Could someone just connect to the a2a endpoint and run openhands on any file in their computer (or in the sandbox), and extract personal information from them? If not, what is preventing this?

Thank you for clarifying!

787627858:
@f-miyake 
I would like to know if there is a plan to support streaming output for a2a, implementing on_message_send_stream

787627858:
@f-miyake  
I would like to know if there is a plan to support streaming output for a2a, implementing on_message_send_stream

ghost:
> @f-miyake I would like to know if there is a plan to support streaming output for a2a, implementing on_message_send_stream

@787627858 
Thank you for your comment.
I don't currently have any plans to implement on_message_send_stream or other missing features such as pushNotification or authorization.
This PR is intended as a starting point for A2A support, and I hope the community can continue to improve and expand on it going forward.

openhands/server/a2a/agent_card.py
```diff
@@ -0,0 +1,36 @@
...
+        'Get the current date and time in Python.',
+    ],
+)
+
+BACKEND_HOST = os.environ.get('BACKEND_HOST', '127.0.0.1')
+BACKEND_PORT = os.environ.get('BACKEND_PORT', '3000')
+
+agent_card = AgentCard(
+    name='Hello openhands_agents Agent',
+    description="An integration layer that allows agents defined in OpenHands' agenthub to be accessible and operable via the A2A protocol.",
```
enyst: It seems an AgentCard is expected to be per agent,
> Overall details (version, name, description, uses)
>    - Skills: A set of capabilities the agent can perform
>    - Default modalities/content types supported by the agent.

[source](https://github.com/a2aproject/a2a-python/blob/1c8e12e448dc7469e508fccdac06818836f5b520/src/a2a/types.py#L1564)

We seem to describe the entire system?

openhands/server/a2a/agent_card.py
```diff
@@ -0,0 +1,36 @@
+import os
+
+from a2a.types import AgentCapabilities, AgentCard, AgentSkill
+
+skill = AgentSkill(
+    id='openhands_agents',
+    name='openhands_agents',
+    description="An integration layer that allows agents defined in OpenHands' agenthub to be accessible and operable via the A2A protocol.",
+    tags=['code', 'execution', 'python', 'llm', 'browsing', 'contains'],
```
enyst:
```suggestion
    tags=['code', 'execution', 'python', 'browsing', 'SWE'],
```
Just a thought, the tags seem to be _what_ the agent can perform.

openhands/server/a2a/agent_card.py
```diff
@@ -0,0 +1,36 @@
...
+    examples=[
+        'When sending a request using A2A,',
+        'include the following metadata in the "message" parameter:',
+        '"metadata": { "agent": "CodeActAgent" },',
+        'This allows the request to be routed to the desired agent.',
+        'The target agents are,[BrowsingAgent, CodeActAgent, DummyAgent, ReadOnlyAgent, VisualBrowsingAgent].etc..',
+        'If we extract from CodeActAgent,',
+        'Calculate the factorial of 10 in Python.',
+        'Sort a list of numbers in descending order using Python.',
+        'Get the current date and time in Python.',
```
enyst, line 17: The examples are examples of capabilities of one agent, [here](https://github.com/a2aproject/a2a-python/blob/1c8e12e448dc7469e508fccdac06818836f5b520/src/a2a/types.py#L109), so maybe like ‘Calculate the factorial of 10 in Python’.

The rest could be a comment?

openhands/server/a2a/a2a_request_handler.py
```diff
@@ -0,0 +1,331 @@
+import asyncio
+import time
+from collections.abc import AsyncGenerator
+from datetime import datetime, timezone
+from typing import Any, Dict
+from uuid import uuid4
+
+from a2a.server.events import Event
+from a2a.types import (
+    Message,
```
enyst:
```suggestion
    Message as A2aMessage,
```

Nit: we already have a `Message` class in the codebase, and it’s a core class. It gets a bit confusing if we use the same name, both for humans and for LLMs 😅

openhands/server/a2a/a2a_request_handler.py
```diff
@@ -0,0 +1,331 @@
...
+                    (
+                        NullAction,
+                        NullObservation,
+                        RecallAction,
+                        RecallObservation,
+                        SystemMessageAction,
+                        AgentStateChangedObservation,
+                    ),
+                ):
+                    continue
```
enyst, line 191: Could you please explain, why we skip all of these and what is it that we don’t skip?

enyst:
> One of our users starts OpenHands on their own personal server using LocalRuntime (so no sandboxing). There is no firewall separating them from the outside internet. Could someone just connect to the a2a endpoint and run openhands on any file in their computer (or in the sandbox), and extract personal information from them? If not, what is preventing this?
> 
> Thank you for clarifying!

I think the answer is the same as it is for the entire `server` component already: 
- it’s a ‘no’ if it’s not accessible from anywhere but localhost (default for uvicorn) or private network
- it’s yes, if exposed to the internet.

OpenHands already exposes [its web API](https://docs.all-hands.dev/api-reference/create-new-conversation) on local, and people can run it on whatever IP. I connect from my iPad every now and then. 🤷 From what I see in this PR, A2A just adds an endpoint and a router to that API.

So I think this isn’t related to this PR, since it’s the same as the current behavior? This PR seems like a net gain for OpenHands, probably, at least if my initial understanding is correct and it can work _between OH agents_ themselves (?), so perhaps we could move forward regardless? Also, it can work with other agents on the machine, so it wouldn’t be useless.

openhands/server/a2a/a2a_request_handler.py
ghost: Thanks! I pushed the updated commit for this fix.

openhands/server/a2a/a2a_request_handler.py
ghost: It processes events that contain information that should be included in A2A responses. 
Therefore, events like MessageAction, CmdRunAction/Observation, IPythonRunCellAction/Observation, etc. are processed here.
On the other hand, it skips events that users don't need to see (like NullAction) and events that contain information users shouldn't see (like SystemMessageAction which contains system prompts).

openhands/server/a2a/a2a_request_handler.py
```diff
@@ -0,0 +1,331 @@
...
+                    }
+                    responses.append(response_dict)
+
+            if isinstance(params, TaskIdParams) and agent_state != AgentState.STOPPED:
+                raise ServerError(error=TaskNotCancelableError())
+
+            messages = []
+            for response in responses:
+                message = Message(
+                    role=Role.agent,
```
enyst: Am I reading this right, is it intended, 
- above we gather in `responses` events no matter what their `source` was, agent, user, or environment
- here we set all of them as role ‘agent’?

openhands/server/a2a/a2a_request_handler.py
enyst: I see below that we create a `Task`, whose `history` is set to the messages corresponding to the collected events.

So it’s not a single response, right, it’s the history of messages in this conversation? I think in that case, maybe it could also include RecallObs (which is some extra info on the task environment) and SystemMessageAction. Is there a reason why shouldn’t users see it? They do see it in the Web UI, in logs, and of course in the code.

ghost: Thank you, that makes sense.
Since `RecallAction`/`Obs` and `SystemMessageAction` are already visible in the WebUI and source code, I agree there's no strong reason to exclude them here.
I'll update the code to include these events in the message history.

openhands/server/a2a/a2a_request_handler.py
ghost: That's actually not intended, my mistake. I'll update the code as below.
Since A2A only supports user and agent roles, I mapped OpenHands’ ENVIRONMENT source to A2A's agent role.
```python
                if response['source'] == EventSource.USER:
                    role = Role.user
                elif response['source'] == EventSource.AGENT or response['source'] == EventSource.ENVIRONMENT:
                    role = Role.agent

                message = A2aMessage(
                    role=role,
```

MrChengmo:
Hi，I'd like to inquire about the mapping relationships between elements in the A2A protocol (such as TaskId, messageId, contextId, artifactId, referenceTaskIds) and those in OpenHands. I noticed that some IDs in your code are either unused or randomly generated. Could there be a more detailed definition here to facilitate complete compatibility with the A2A protocol in the future?

ghost:
> Hi，I'd like to inquire about the mapping relationships between elements in the A2A protocol (such as TaskId, messageId, contextId, artifactId, referenceTaskIds) and those in OpenHands. I noticed that some IDs in your code are either unused or randomly generated. Could there be a more detailed definition here to facilitate complete compatibility with the A2A protocol in the future?

@MrChengmo
Thank you for your feedback.
Regarding the mapping of identifiers between A2A and OpenHands, here is the current status and some thoughts for future improvements:

- TaskId: Currently, this is randomly generated on the A2A server side and used to associate with OpenHands sessions. It might be more appropriate in the future to directly use the OpenHands Session ID instead.
- messageId: Similar to TaskId, messageId is also randomly generated and mapped to the corresponding event in OpenHands. Ideally, we would prefer to use the event.id from OpenHands directly. However, since event.id is defined as an int and A2A expects messageId to be a str, direct usage is currently not possible.
- contextId: This is also randomly generated on the server side, as there is currently no concept in OpenHands that corresponds directly to the A2A notion of "context". While OpenHands’ Session is the closest equivalent, in this implementation a Session is used to represent an A2A Task, so it cannot be reused for context.
- artifact / artifactId: Currently, artifacts are not generated because there is no corresponding mechanism in OpenHands to produce or manage outputs as artifacts. Supporting this would require broader and more fundamental changes not only to the A2A server, but also to other parts of the OpenHands system.

We appreciate your suggestions and agree that a more detailed and systematic mapping will be important for improving compatibility with the A2A protocol in the future.

openhands/server/a2a/agent_card.py
ghost: The AgentCard is intended to describe a single agent. I agree with your point that it's not appropriate to describe the entire system here.
I've replaced the description of the agent card and skill with an explanation of CodeActAgent, which is the default agent in OpenHands.

openhands/server/a2a/agent_card.py
ghost: I've also adjusted the examples here to focus on usage of CodeActAgent, and left the explanation about metadata commented out.

PR #6557, opened 2025-01-30
stop retrying on all exceptions

rbren:
**End-user friendly description of the problem this fixes or functionality that this introduces**

- [ ] Include this change in the Release Notes. If checked, you must provide an **end-user friendly** description for your change below
no changelog

---
**Give a summary of what the PR does, explaining any non-trivial design decisions**

This is catching a lot of unrecoverable exceptions, which makes it look like the app is just stuck

We could potentially move this into a config if these errors are common enough that eval needs them. CC @xingyaoww 

---
**Link of any specific issues this addresses**

PR #9281, opened 2025-06-21
Improve rate limit message to indicate automatic retry

neubig:
## Summary

This PR improves the user experience when the agent encounters rate limiting by updating the status message to clearly indicate that the agent will retry automatically.

## Problem

Currently, when the agent gets rate-limited, users see the message "Agent is Rate Limited" which doesn't indicate:
- Whether the agent will resume automatically
- That users don't need to take any action

This leads to user confusion about whether the agent is stuck or will continue automatically.

## Solution

**Simple Message Update:**
- Changed `"Agent is Rate Limited"` to `"Agent is Rate Limited. Retrying automatically..."`
- Updated the message in all supported languages to maintain consistency

### Before:
```
Agent is Rate Limited
```
(Unclear if agent will resume)

### After:
```
Agent is Rate Limited. Retrying automatically...
```
(Clear indication of automatic retry)

## Benefits

1. **Clear Communication**: Users know the agent will retry automatically
2. **Reduced Confusion**: Eliminates uncertainty about whether action is needed
3. **Better UX**: Users can wait confidently knowing the agent will resume
4. **Multilingual Support**: Consistent message across all supported languages

## Technical Details

- Updated `CHAT_INTERFACE$AGENT_RATE_LIMITED_MESSAGE` in translation.json
- No backend changes required - leverages existing retry mechanism
- No breaking changes to existing functionality
- Simple, focused change that directly addresses the user concern

---

This improvement significantly enhances the user experience during rate limiting scenarios by providing clear communication about automatic retry behavior.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/0b0e13df5bfa4ef2a56d353bedb2003a)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:ea93297-nikolaik   --name openhands-app-ea93297   docker.all-hands.dev/all-hands-ai/openhands:ea93297
```


neubig:
This LGTM

enyst:
The agent will retry several times (the configured number of retries), then stop with the same exception I think. Will this message show up when it stops, too, not only when it will retry automatically? Then it would make the user believe they can wait, when they should act. Not fully sure though.

Cc: @raymyers who has implemented this message display in the UI

neubig:
@openhands enyst makes a good point. Please check if the same message will still be displayed after the agent times out with a fixed number of retries. If so, we should add another message that will be displayed at that point that says "Agent is rate-limited. Stopped."

neubig:
@openhands Python unit tests are failing. Carefully read the GitHub workflow associated with these tests and follow it step by step to reproduce the error locally. Once you have reproduced it, fix the error. When you think it is fixed, push to GitHub and then wait 120 seconds before checking the result of the CI with the GitHub API. If any tests have failed, reproduce them locally and fix them. Then continue waiting and checking until CI passes

llamantino:
I tried the PR and what it does now is:
- It retries 4 times without any feedback on UI
- Then the message changes to Retrying, but it actually never tries again

Basically it's like before, just with a different message.

By looking at the code, it seems like before_sleep is triggered correctly, as I can see all 4 retry messages in the Docker logs (= the retries are detected). However, the exception handling logic expects multiple exceptions to be raised in order to work properly, while tenacity will only raise one (a single one, after all retries fail). That single exception triggers the initial "Retrying" message, but without other exceptions, the "fail" state is never shown. No idea why it was written that way.

Note #1: if it wasn't clear: if you want the agent to stop after tenacity fails, yet show the retry status updates, then discard the agent_controller changes, and have before_sleep trigger the status updates.

Note #2: Right now, the code does 12 retries in a very short (24 seconds) time span. Not even a minute, so most common use scenarios are likely to fail because rate limiting API server blocks aren't usually that short. There have been some complaints in Issues that I think are somewhat related to these short wait times. I looked into it and commented about it [here](https://github.com/All-Hands-AI/OpenHands/pull/6557#issuecomment-3012625588).

neubig:
@openhands read the comments from llamantino and write a failing test to demonstrate the current undesirable behavior. You MUST write a test that FAILS because the current behavior is not appropriate. Only once you have succeeded at demonstrating the current bad implementation, you can try to fix it.

neubig:
@openhands change the message according to `@amanape` 's comments

llamantino:
I tried it again, and it still looks broken.

The behavior is the same I described in my previous comment: there are no UI updates until after it stops retrying.

Abubakar-01:
**Test Setup**
Created a fake server that always returns 429 rate limit errors.
Configured OpenHands to use the fake server.
Triggered rate limiting by sending a message to the agent.

**Issue Summary**
The following behavior was observed during testing:
**Retry Mechanism**: The system retries every minute as expected.
**Message Display**: The message updates from "Agent is Rate Limited" to "Agent is Rate Limited. Retrying..."
**Identified Problem**: The message "Agent is Rate Limited. Retrying..." appears only after 8 minutes, which is somewhat misleading since the system starts retrying much earlier.

PR #9287, opened 2025-06-22
feat(cli): Add vi mode support

ikuo:

This change introduces optional vi-style keybindings for confirmation prompts navigating menus (`j` and `k`) in the command-line interface. Users can enable this feature by setting the `CLI_VI_MODE` environment variable to `true`. This provides a more familiar and efficient way for users accustomed to vi editors to interact with the CLI.

---

This PR adds unit tests for the new `vi_mode` feature. It ensures that the VI keybindings (`j` and `k`) are correctly handled in confirmation prompts. The PR also includes necessary fixes to existing tests that were broken by the new `config` object being passed to CLI functions to support `vi_mode`.

Key changes:
- Added `tests/unit/test_vi_mode.py` to test the `vi_mode` functionality.
- Updated `tests/unit/test_cli_commands.py` and `tests/unit/test_cli_tui.py` to reflect the new function signatures required for `vi_mode`.

---

Nothing applicable.

### DIFF
--- a/openhands/core/config/openhands_config.py
+++ b/openhands/core/config/openhands_config.py
@@ -5,6 +5,7 @@ from pydantic import BaseModel, Field, SecretStr
 
 from openhands.core import logger
 from openhands.core.config.agent_config import AgentConfig
+from openhands.core.config.cli_config import CLIConfig
 from openhands.core.config.config_utils import (
     OH_DEFAULT_AGENT,
     OH_MAX_ITERATIONS,
@@ -109,10 +110,7 @@ class OpenHandsConfig(BaseModel):
     mcp_host: str = Field(default=f'localhost:{os.getenv("port", 3000)}')
     mcp: MCPConfig = Field(default_factory=MCPConfig)
     kubernetes: KubernetesConfig = Field(default_factory=KubernetesConfig)
-    vi_mode: bool = Field(
-        default=os.environ.get('OPENHANDS_VI_MODE', 'false').lower()
-        in ('true', '1', 't')
-    )
+    cli: CLIConfig = Field(default_factory=CLIConfig)
 
     defaults_dict: ClassVar[dict] = {}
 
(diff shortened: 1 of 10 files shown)


openhands/core/config/openhands_config.py
```diff
@@ -109,6 +109,10 @@ class OpenHandsConfig(BaseModel):
     mcp_host: str = Field(default=f'localhost:{os.getenv("port", 3000)}')
     mcp: MCPConfig = Field(default_factory=MCPConfig)
     kubernetes: KubernetesConfig = Field(default_factory=KubernetesConfig)
+    vi_mode: bool = Field(
+        default=os.environ.get('OPENHANDS_VI_MODE', 'false').lower()
+        in ('true', '1', 't')
+    )
```
xingyaoww: Actually i think you can keep it default to False.

If you set `VI_MODE=true`, it will automatically be picked up by OpenHands configuration system and set this label to true, i just tested with this `VI_MODE=true uvx --python 3.12 --from git+https://github.com/ikuo/OpenHands@feat/cli-vi-mode openhands` and it seems to work pretty well!

xingyaoww: Another slight concern i have is putting `vi_mode` under `OpenHandsConfig` which could make it even more bloated :(

Maybe we could create a new class called "CLIConfig" to put configuration like this that's CLI specific?

xingyaoww (approved):
Thanks for your contribution! This looks pretty clean and good -- mainly have concerns on the configuration part

openhands/core/config/openhands_config.py
ikuo: @xingyaoww Thank you for the feedback!
I've put `vi_mode` under `CLIConfig` and tweaked the code at [b4e9bd4](https://github.com/All-Hands-AI/OpenHands/pull/9287/commits/b4e9bd4bf03761a2100f81771e2f8612e606772a).
Additionally, to align with the style of use of environment variable, I'v renamed `OPENHANDS_VI_MODE` to `CLI_VI_MODE`.

I've confirmed that vi_mode defaults to `False` by testing with and without `CLI_VI_MODE` variable.

I hope this makes sense.

Issue #9227, opened 2025-06-18
[Bug]: Context length is hit when asking simple question: what day is today?


chiungyit:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

** Problem **
I hosted All Hands locally. Without specifying a code base, I only ask 1 simple question: what day is today? I got the error.

```
23:39:39 - openhands:ERROR: session.py:332 - Agent status error: BadRequestError: litellm.BadRequestError: SambanovaException - The maximum context length of DeepSeek-V3-0324 is 32768. However, answering your request will take 39732 tokens. Please reduce the length of the messages or the specified max_completion_tokens value.
23:39:39 - openhands:ERROR: session.py:273 - Agent status error: BadRequestError: litellm.BadRequestError: SambanovaException - The maximum context length of DeepSeek-V3-0324 is 32768. However, answering your request will take 39732 tokens. Please reduce the length of the messages or the specified max_completion_tokens value.
23:39:39 - OBSERVATION
[Agent Controller 336f50d1ad1c4fcfa0e9eb00c2080c32] AgentStateChangedObservation(content='', agent_state='running', reason='', observation=<ObservationType.AGENT_STATE_CHANGED: 'agent_state_changed'>)
23:39:39 - OBSERVATION
[Agent Controller 336f50d1ad1c4fcfa0e9eb00c2080c32] AgentStateChangedObservation(content='', agent_state='error', reason='BadRequestError: litellm.BadRequestError: SambanovaException - The maximum context length of DeepSeek-V3-0324 is 32768. However, answering your request will take 39732 tokens. Please reduce the length of the messages or the specified max_completion_tokens value.', observation=<ObservationType.AGENT_STATE_CHANGED: 'agent_state_changed'>)
```
** Triage ** 
1. I've tried to set docker box environment variable: LLM_MAX_INPUT_TOKENS, but no help. 
2. I've tried to use litellm directly to hit the endpoint with the simple question. It went fine.

What's actually being sent to the LLM ? I don't have visibility into what's in the payload.

### OpenHands Installation

Docker command in README

### OpenHands Version

latest docker image 0.44

### Operating System

MacOS

neubig:
@openhands find the place where we are catching context length exceptions (when they are not caught by litellm itself) and add this SambanovaException

neubig:
[...]

PR #9296, opened 2025-06-23
Simplify max_output_tokens handling in LLM classes

neubig:
## Summary

This PR simplifies the handling of `max_output_tokens` in OpenHands LLM classes by removing complex default-setting logic and keeping the value as `None` when not explicitly set.

### Core Changes
- **Removed complex default logic**: Eliminated the complex logic in `init_model_info()` that attempted to determine default values for `max_output_tokens` based on various conditions
- **Simplified parameter handling**: Now if `max_output_tokens` is `None`, it remains `None` instead of being set to computed defaults
- **Conditional parameter inclusion**: Updated completion function calls in all LLM classes to conditionally include `max_output_tokens` only when it's not `None`

### Files Modified
- `openhands/llm/llm.py`: Removed lines 489-507 containing complex default determination logic, added conditional parameter inclusion
- `openhands/llm/async_llm.py`: Added conditional `max_tokens` parameter inclusion
- `openhands/llm/streaming_llm.py`: Added conditional `max_tokens` parameter inclusion  
- `tests/unit/test_llm.py`: Updated test expectations to expect `None` values instead of computed defaults

## Testing

- ✅ All 32 LLM unit tests pass
- ✅ Pre-commit hooks pass (ruff, mypy, formatting)
- ✅ Verified headless mode functionality works with the changes

## Benefits

1. **Simplified codebase**: Removes unnecessary complexity in determining defaults
2. **Clearer behavior**: More predictable behavior when `max_output_tokens` is not set
3. **Better maintainability**: Less complex logic to maintain and debug
4. **Consistent handling**: Uniform approach across all LLM implementations

## Backward Compatibility

This change maintains backward compatibility - the LLM providers will use their own defaults when `max_output_tokens` is not provided, which is the expected behavior.

## References

- Addresses user request to simplify `max_output_tokens` handling
- Follows the principle of keeping `None` as `None` instead of complex default computation

Fixes https://github.com/All-Hands-AI/OpenHands/issues/9227

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/c9bef07451414451815c3074dd9da017)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:89ec682-nikolaik   --name openhands-app-89ec682   docker.all-hands.dev/all-hands-ai/openhands:89ec682
```

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -161,11 +161,8 @@ class LLM(RetryMixin, DebugMixin):
         # set up the completion function
         kwargs: dict[str, Any] = {
             'temperature': self.config.temperature,
+            'max_completion_tokens': self.config.max_output_tokens,
         }
-
-        # Only include max_output_tokens if it's explicitly set
-        if self.config.max_output_tokens is not None:
-            kwargs['max_completion_tokens'] = self.config.max_output_tokens
 
         if self.config.top_k is not None:
             # openai doesn't expose top_k
@@ -182,9 +179,8 @@ class LLM(RetryMixin, DebugMixin):
             )  # temperature is not supported for reasoning models
         # Azure issue: https://github.com/All-Hands-AI/OpenHands/issues/6777
         if self.config.model.startswith('azure'):
-            if self.config.max_output_tokens is not None:
-                kwargs['max_tokens'] = self.config.max_output_tokens
-            kwargs.pop('max_completion_tokens', None)
+            kwargs['max_tokens'] = self.config.max_output_tokens
+            kwargs.pop('max_completion_tokens')
 
         self._completion = partial(
             litellm_completion,
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -487,9 +487,6 @@ class LLM(RetryMixin, DebugMixin):
                 # Safe fallback for any potentially viable model
                 self.config.max_input_tokens = 4096
 
-        # Keep max_output_tokens as None if not explicitly set
-        # Let the LLM provider handle defaults
-
         # Initialize function calling capability
         # Check if model name is in our supported list
         model_name_supported = (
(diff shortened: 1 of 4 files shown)


openhands/llm/llm.py
```diff
@@ -486,26 +487,6 @@ def init_model_info(self) -> None:
...
-                    self.config.max_output_tokens = self.model_info['max_output_tokens']
-                elif 'max_tokens' in self.model_info and isinstance(
-                    self.model_info['max_tokens'], int
-                ):
-                    self.config.max_output_tokens = self.model_info['max_tokens']
-            if any(
-                model in self.config.model
-                for model in ['claude-3-7-sonnet', 'claude-3.7-sonnet']
-            ):
-                self.config.max_output_tokens = 64000  # litellm set max to 128k, but that requires a header to be set
```
enyst, line 507: I think at least this one needs a quick test with 3.7. I’m not at my computer, but I can do that in a bit

enyst:
Re: the complex logic here.
Some of it is really old and I’ve wanted before to clean it up. Some of it is newer and was solving an issue (comment inline).

openhands/llm/llm.py
enyst: I confirm it works well. I think this was a litellm issue at some point in time, it must have been fixed, and we forgot this code here.

enyst (approved):
I got LLMs to tell me stories, I also tried to look up the history of this code, and I don't find a reason not to clean it out. So let's go with this. If there are issues, we'll find out. 😅

neubig:
Thank you!

PR #9298, opened 2025-06-23
Optional daytona

enyst:

---

---


neubig:
Hey @enyst , I'm working on implementing the change for all runtimes, just FYI so we don't duplicate work. Although I appreciate the Daytona team's enthusiasm to help out, I think we should probably treat these runtimes equally unless we decide to totally remove them because the maintainers are no longer interested.

enyst:
Ah, yes, I see. I meant to implement it so that you don't have to, but your solution is better anyway!

PR #9307, opened 2025-06-23
docs: rewrite local LLMs page

llamantino:

Restructures and simplifies the documentation for running local LLMs.

---

Updates Local LLMs documentation:
- Rewrote and streamlined LM Studio setup with new screenshots
- Added visual instructions for OpenHands configuration, removing the need for config files or custom startup commands
- Added hardware requirements for Devstral, and context-length warnings for Ollama users
- Merged Ollama, vLLM, and SGLang into a dedicated Advanced backends section at the bottom
- And more ...

---


mamoodi:
Thanks so much for redoing all of this! Generally looks good to me. Tagging @enyst and @xingyaoww for a quick look.

docs/usage/llms/local-llms.mdx
```diff
@@ -6,57 +6,69 @@ description: When using a Local LLM, OpenHands may have limited functionality. I
...
 
-4. Then click `Select a model to load` on top of the application:
+1. Make sure to set the User Interface Complexity Level to "Power User", by clicking on the appropriate label at the bottom of the window.
+2. Click the "Discover" button (Magnifying Glass icon) on the left navigation bar to open the Models download page.
 
-![image](./screenshots/2_select_model.png)
+![image](./screenshots/01_lm_studio_open_model_hub.png)
 
-5. And choose the model you want to use, holding `option` on mac to enable advanced loading options:
+3. Search for the "Devstral Small 2505" model, confirm it's the official Mistral AI (mistralai) model, then proceed to download.
```
raymyers, line 35: If they are using CPU will they need to make a decision about which quantization? Or does LM Studio handle that

xingyaoww: Yep, LM Studio will handle that

image

docs/usage/llms/local-llms.mdx
```diff
@@ -6,57 +6,69 @@ description: When using a Local LLM, OpenHands may have limited functionality. I
...
 
-8. Finally, you can click the `copy` button near model name to copy the model name (`imported-models/uncategorized/devstralq4_k_m.gguf` in this example):
+3. Enable the "Manually choose model load parameters" switch.
+4. Select 'Devstral Small 2505' from the model list.
 
-![image](./screenshots/6_copy_to_get_model_name.png)
+![image](./screenshots/04_lm_studio_setup_devstral_part_1.png)
 
-### Start OpenHands with locally served model
+5. Enable the "Show advanced settings" switch at the bottom of the Model settings flyout to show all the available settings.
```
raymyers, line 53: "flyout" is ui jargon that end users might not know. Maybe we can call it a panel?

docs/usage/llms/local-llms.mdx
```diff
@@ -6,57 +6,69 @@ description: When using a Local LLM, OpenHands may have limited functionality. I
...
+6. Set "Context Length" to at least 32768 and enable Flash Attention.
```
raymyers, line 54: nit, number could probably use comma and units, like `32,768 MB` (not sure if the ui says mib or what)

docs/usage/llms/local-llms.mdx
```diff
@@ -84,65 +94,88 @@ INFO:     Application startup complete.
...
+
+Once you open OpenHands in your browser, you'll need to configure it to use the local LLM server you just started.
+
+When started for the first time, OpenHands will prompt you to set up the LLM provider.
+
+1. Click "see advanced settings" to open the LLM Settings page.
+
+![image](./screenshots/07_openhands_open_advanced_settings.png)
+
+2. Enable the "Advanced" switch at the top of the page to show all the available settings.
```
raymyers, line 109: These two instructions seem to say the same thing, but maybe is correct if the UI really makes you do it twice?

raymyers:
I've reviewed this as a native english speaker and it reads fine, besides maybe the flyout jargon. I have not run the instructions. Thanks!

docs/usage/llms/local-llms.mdx
llamantino: It looks like a dropdown menu, but when you click it, a larger panel pops out. Calling it a dropdown wouldn't have been accurate. "Panel" could be confusing too, since the UI is pane-based with a vertical selector on the left.

I thought flyout was the best fit, though maybe the clearest option would've been to describe it in more detail (e.g. the expanded model selection screen that appears). I'll try to come up with a better alternative.

There are screenshots for each step, with enough detail that reading the page text shouldn't really be necessary, so hopefully it's still easy to follow. Thanks!

docs/usage/llms/local-llms.mdx
llamantino: 32,768 could look like decimals for EU users (',' is used for decimals, '.' for thousands here). The UI only accepts raw numbers, so I think it's better to leave it as is. I'll add an `inline code tag` in the next update to make it clearer. Thanks!

llamantino:
Thanks for reviewing the wording - glad to hear it reads fine to a native speaker! I'll address both suggestions in the next update, since the PR has already been merged.

I suggest having someone unfamiliar with OpenHands try the guide to see if they run into any issues. I consider it one of the most important pages - if not the most important - in the docs. Thanks again!

Issue #7358, opened 2025-03-19
[Feature]: Support reading in cursorrules.md files


openhands-agent:
**What problem or use case are you trying to solve?**
The user is requesting support for `.cursorrules.md` files to enhance functionality. Currently, these are not directly supported, as mentioned in the Slack thread.

**Describe the UX of the solution the user wants**
The user appears to want direct integration and readable support for `.cursorrules.md` files, possibly to avoid manual addition to chats.

**User's thoughts on the technical implementation?**
The mentioned alternative is leveraging mechanisms like <https://docs.all-hands.dev/modules/usage/prompting/microagents-repo> for similar functionality.

**Describe the alternatives considered**
The user mentioned manually adding rules to chats, which is not ideal.

**Additional context**
Relevant documentation link provided in the Slack thread: <https://docs.all-hands.dev/modules/usage/prompting/microagents-repo>

**Issue Created By**: Graham Neubig on Slack

**Link to Slack Thread**: https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1742415685622419?thread_ts=1742415685.622419&cid=C06U8UTKSAD

### If you find this feature request or enhancement useful, make sure to add a 👍 to the issue

AutoLTX:
@neubig I'm interested about the issue and I like to contribute for it. Please assign it to me if no one is working on it.
Just want to confirm one thing with you first, which type of support are we preparing to do?

- Support it like exising repo.md
- Support it like cursor.

Both are acceptable to me. I'd like to confirm with you first before starting the design/implementation 😊

neubig:
Hey @AutoLTX , thanks! I'm not 100% sure what you mean by this, could you clarify a little bit what you mean by these two options? What would they look like for the OpenHands user?

BradKML:
Is this about self-modifying prompts that Cursor was actively removing?

BradKML:
Welp `.cursorrules` and their windsurf/roocode equivalents are pretty useful for self-updating agents, so there needs to be a way to keep that up as well... Or until there is a way to self-add "sub-rules" by the agent. Will get some example of how they probably should look like in a moment

neubig:
Ok, I think option 1 sounds good! And status tracking on GitHub issues and slack is good.

neubig:
Hey @AutoLTX , just checking in on this, have you had a chance to take a look?

AutoLTX:
[...]

PR #9327, opened 2025-06-24
[Feature] Support .cursorrules

AutoLTX:

---
Support .cursorrules. It should be in the root directory.

---
https://github.com/All-Hands-AI/OpenHands/issues/7358

### DIFF
--- a/openhands/microagent/microagent.py
+++ b/openhands/microagent/microagent.py
@@ -1,5 +1,6 @@
 import io
 import re
+from itertools import chain
 from pathlib import Path
 from typing import Union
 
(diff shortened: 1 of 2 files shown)


openhands/microagent/microagent.py
```diff
@@ -258,10 +273,15 @@ def load_microagents_from_dir(
     # Load all agents from microagents directory
     logger.debug(f'Loading agents from {microagent_dir}')
     if microagent_dir.exists():
-        for file in microagent_dir.rglob('*.md'):
-            # skip README.md
-            if file.name == 'README.md':
-                continue
+        # Collect .cursorrules file from repo root and .md files from microagents dir
+        cursorrules_files = []
```
enyst, line 277: If I see this right, we only load `.cursorrules` if the user repo has a microagents directory? It’s a bit unexpected, I think we should always load it.

It could be for later though!

enyst (approved):
LGTM, thank you!

Issue #9325, opened 2025-06-24
Add User Directory Support for Microagents


enyst:
(openhands-agent, after it made a funny boo-boo and proved it uses `git reflog` better than me 😅)

## Problem
Currently, microagents are read from multiple locations but not from a user-specific directory. This leads to the problem of having uncommitted local microagents in the repository, which can be accidentally lost during git operations (as we experienced when `local.md` and `local-gem.md` were lost during a `git add .` and `git reset --hard`).

## Desired Behavior
The microagents system in OpenHands should read microagents from `~/.openhands/microagents/` in addition to its existing sources. This would allow users to:

1. Store personal/local microagents in their user directory (`~/.openhands/microagents/`)
2. Avoid keeping uncommitted microagent files in repository working directories
3. Have persistent personal microagents across all OpenHands projects

## Current Microagent Sources
As of now, microagents are read from multiple locations, but the user directory is not included. We need to implement reading them from the user directory.

## Implementation Requirements
- Add `~/.openhands/microagents/` as a microagent source
- Ensure proper loading order and precedence
- Maintain compatibility with existing microagent sources
- Handle cases where the user directory doesn't exist (create it or gracefully skip)

## Best Practices for Users
- **Personal/local microagents**: Add them to `~/.openhands/microagents/` instead of keeping them uncommitted in repository directories
- **Shared microagents**: If microagents are intended to be shared with other users, commit them to the repository's `.openhands/microagents/` directory
- **Never keep microagents uncommitted** in repository working directories to avoid accidental loss during git operations

## Files Affected
- Need to identify where microagent loading logic is implemented
- Likely in the microagents system/loader code
- May need to update configuration or initialization code

## Testing
- Verify microagents are loaded from user directory
- Test precedence/priority of different microagent sources
- Ensure backward compatibility with existing microagent locations

PR #9333, opened 2025-06-24
feat: Add user directory support for microagents

enyst:

Supports reading microagents from the user directory (`~/.openhands/microagents/`) in addition to existing sources.

(openhands-agent, task via vscode because we can 😅)

This allows you to store personal microagents that persist across all OpenHands projects without keeping uncommitted files in your repository working directories, preventing accidental loss during git operations.

---

This PR adds user directory support for microagents.

**Design decisions:**
- User microagents are loaded after global microagents, allowing users to override global ones
- Directory is created automatically to provide a smooth user experience
- Errors in user microagents don't crash the system, just log warnings
- Uses existing `load_microagents_from_dir()` function for consistency

**Benefits:**
- Persistent personal microagents across all OpenHands projects
- No more uncommitted microagent files in repository working directories
- Prevents accidental loss during git operations
- Maintains full backward compatibility

---

This addresses the problem mentioned in our development workflow where `local.md` and `local-gem.md` microagents were accidentally lost during `git add .` and `git reset --hard` operations because they were kept as uncommitted files in the repository working directory.

Fix https://github.com/All-Hands-AI/OpenHands/issues/9325

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:03ec0d3-nikolaik   --name openhands-app-03ec0d3   docker.all-hands.dev/all-hands-ai/openhands:03ec0d3
```


openhands/memory/memory.py
```diff
@@ -33,6 +33,8 @@
     'microagents',
 )
 
+USER_MICROAGENTS_DIR = os.path.expanduser('~/.openhands/microagents')
```
neubig: Will this only work on linux-based OSs? There might be a more OS-agnostic way of getting the home directory.

neubig (commented):
LGTM, just one small comment.

openhands/memory/memory.py
rbren: I think `os.path` handles the ambiguity, but not sure

enyst: I think it should work, and we already use something like this [here](https://github.com/All-Hands-AI/OpenHands/blob/8f566a4247f5bdc7bb702df929be28233ea945c3/openhands/storage/local.py#L12) for example.

I changed it to a better way though!

neubig (approved):
This is nice, thank you!

Issue #8218, opened 2025-05-02
[Bug]: deepseek-r1 think tags broken


createthis:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

I believe deepseek-r1 is incapable of utilizing open hands ai's <think> tool. R1 simply surrounds its thoughts with
```
<think>
</think>
```
tags.

I believe open hands expects something like:

```
<think>
<thought>something</thought>
</think>
```

This results in R1's thoughts being ignored by the open hands UI. They just disappear. I can only see them by inspecting my server logs.

I'm using `ktransformers` and running R1 locally. Perhaps other inference engines don't have this limitation.

### OpenHands Installation

Docker command in README

### OpenHands Version

0.36.0

### Operating System

None

createthis:
FYI, this is still broken with DeepSeek-R1-0528.

createthis:
I recorded a session with R1-0528 here: https://github.com/createthis/open_hands_deepseek_r1_0528_think_tags

Unfortunately, it looks like R1-0528 under `llama.cpp` is completely ignoring the `<think>` tag instructions. I'm not sure there is much that can be done when it refuses to properly close tags.

I'm going to close this issue for now until a future version of R1 starts following instructions.

It's weird though, because think tags work great with R1-0528 in `open webui`. It makes me wonder what they're doing differently.

createthis:
Re-opening with new data:

I made a new recording with the patch from @SmartManoj  removed and [think tags disabled](https://github.com/All-Hands-AI/OpenHands/blob/0c1c570dac654ba6536d1eaaa30a65dcb0416971/config.template.toml#L256): https://github.com/createthis/open_hands_deepseek_r1_0528_think_tags_think_tool_disabled

You can view the whole transcript by checking out the repo, installing [mitmdump](https://docs.mitmproxy.org/stable/overview/installation/), and running `mitmdump -nr full_traffic_think_tool_disabled.mitm --flow-detail 4` , but here's the relevant part with the think tags:

`
            {
                "finish_reason": "stop",
                "index": 0,
                "message": {
                    "role": "assistant",
[...]

PR #9337, opened 2025-06-25
feat: Include reasoning content

SmartManoj:

---

---
fixes https://github.com/All-Hands-AI/OpenHands/issues/8218#issuecomment-3002794281


openhands/llm/llm.py
```diff
@@ -345,7 +345,16 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
                     + str(resp)
                 )
 
-            message_back: str = resp['choices'][0]['message']['content'] or ''
+            message_content = resp['choices'][0]['message']['content'] or ''
+            reasoning_content = resp['choices'][0]['message'].get('reasoning_content')
+            if reasoning_content and os.getenv('SHOW_REASONING_CONTENT', 'false').lower() in ['true', '1']:
+                reasoning_tag = os.getenv('REASONING_TAG_NAME', 'think')
```
enyst, line 351: I think env vars are not necessary for this, and we have too many already 😅

openhands/llm/llm.py
```diff
@@ -345,7 +345,16 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+                message_content = (
+                    f'<{reasoning_tag}>' + reasoning_content + f'</{reasoning_tag}>\n' + message_content
```
enyst, line 353: Storing reasoning in `message_content` means it will be sent back to the LLM? As far as I know, many / most LLMs will not work so well with it. In addition, we mix into content here, which literally changes what the LLM “said” last turn.

enyst:
We’re still looking into how to do this best, I think this might not be the right solution. You may want to take a look at this PR, an alternative:
- https://github.com/All-Hands-AI/OpenHands/pull/9374

SmartManoj:
Closing in favor of #9374

PR #9340, opened 2025-06-25
feat: Implement error summarization for user-facing messages

rkarahan80:
This commit introduces an error summarization feature to provide users with more friendly and understandable error messages.

Key changes:
- Added `openhands/core/error_summarizer.py` with a `summarize_error` function that generates summaries for various common exceptions (LiteLLM, OpenHands core, standard Python).
- Modified `ErrorObservation` to include an optional `summary` field.
- Modified `AgentStateChangedObservation` to include an optional `error_summary` field.
- Integrated the summarization logic into `AgentController`:
    - `_react_to_exception` now calls `summarize_error` and stores the summary.
    - `set_agent_state_to` passes this summary to `AgentStateChangedObservation`.
    - `_step` calls `summarize_error` for specific LLM errors and includes the summary in `ErrorObservation`.
- Verified that `event_to_dict` in `openhands/events/serialization/event.py` correctly serializes the new summary fields into the `extras` dictionary.
- Added unit tests for `error_summarizer.py`.
- Updated unit tests for `test_event_serialization.py` to cover the new summary fields.
- Included conceptual documentation for the new feature.

---

---

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -37,6 +37,7 @@ from openhands.core.exceptions import (
     LLMNoActionError,
     LLMResponseError,
 )
+from openhands.core.error_summarizer import summarize_error
 from openhands.core.logger import LOG_ALL_EVENTS
 from openhands.core.logger import openhands_logger as logger
 from openhands.core.schema import AgentState
@@ -242,7 +243,17 @@ class AgentController:
     ) -> None:
         """React to an exception by setting the agent state to error and sending a status message."""
         # Store the error reason before setting the agent state
-        self.state.last_error = f'{type(e).__name__}: {str(e)}'
+        # Store the error reason before setting the agent state
+        original_error_message = f'{type(e).__name__}: {str(e)}'
+        self.state.last_error = original_error_message
+        # Attempt to generate and store a summary
+        try:
+            # TODO: Consider what context might be useful here, e.g., last action attempted.
+            self.state.last_error_summary = summarize_error(e, context={'original_message': original_error_message})
+        except Exception as summary_ex:
+            # If summarization itself fails, log it and proceed without a summary
+            self.log('error', f"Failed to summarize error: {summary_ex}")
+            self.state.last_error_summary = None # Ensure it's None if summarization fails
 
         if self.status_callback is not None:
             err_id = ''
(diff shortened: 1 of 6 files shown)


openhands/controller/agent_controller.py
```diff
@@ -242,7 +243,17 @@ async def _react_to_exception(
     ) -> None:
         """React to an exception by setting the agent state to error and sending a status message."""
         # Store the error reason before setting the agent state
-        self.state.last_error = f'{type(e).__name__}: {str(e)}'
+        # Store the error reason before setting the agent state
+        original_error_message = f'{type(e).__name__}: {str(e)}'
+        self.state.last_error = original_error_message
```
enyst, line 248: Could you explain why this is necessary? The rest of the diff doesn’t seem to change `last_error`

openhands/controller/agent_controller.py
```diff
@@ -588,11 +599,18 @@ async def set_agent_state_to(self, new_state: AgentState) -> None:
...
         reason = ''
+        error_summary = None
         if new_state == AgentState.ERROR:
             reason = self.state.last_error
+            error_summary = getattr(self.state, 'last_error_summary', None)
 
         self.event_stream.add_event(
-            AgentStateChangedObservation('', self.state.agent_state, reason),
+            AgentStateChangedObservation(
+                content='',  # content is not used by AgentStateChangedObservation
```
enyst, line 609: If indeed `content` is unused, could we use it instead of adding a new field?

openhands/events/observation/agent.py
```diff
@@ -11,12 +11,21 @@ class AgentStateChangedObservation(Observation):
...
     observation: str = ObservationType.AGENT_STATE_CHANGED
 
     @property
     def message(self) -> str:
         return ''
 
+    def __str__(self) -> str:
+        res = f'**AgentStateChangedObservation** Agent state changed to {self.agent_state}'
+        if self.reason:
+            res += f'\nReason: {self.reason}'
```
enyst, line 24: Could we perhaps use the summary in `reason`, instead of a new field?

openhands/events/observation/error.py
```diff
@@ -14,10 +20,13 @@ class ErrorObservation(Observation):
 
     observation: str = ObservationType.ERROR
     error_id: str = ''
+    summary: str | None = field(default=None, metadata={'log_default': None})
```
enyst, line 23: I’m not sure a new field is necessary, could we instead use the `message` property? It needs the information in the attributes, but maybe we can have error id and build the summary with it?

rkarahan80:
I think it's very suitable. Message feature can be used.

26 Haz 2025 Per 05:40 tarihinde Engel Nyst ***@***.***> şunu
yazdı:

PR #9355, opened 2025-06-25
Fix: Normalize file_text=None to empty string for create command

### LINKED ISSUE #9343: Bug: "Parameter `file_text` is required for command: create" Error in OpenHands Cloud
## Description
Users are encountering an error when using the file editor in OpenHands Cloud:
```
ERROR: Parameter `file_text` is required for command: create
```

This error occurs when attempting to create files through the file editor interface.

## Root Cause Analysis

The error originates from the `openhands-aci` package (version 0.3.0) which is used by OpenHands for file editing operations. Specifically:

1. **Location**: The error is thrown by the `OHEditor` class from `openhands-aci` when the `create` command is executed
2. **Trigger**: The error occurs when `FileEditAction` with `command='create'` is called but `file_text` parameter is `None`
3. **Code Path**: 
   - Frontend sends `FileEditAction` → 
   - `ActionExecutor.edit()` in `action_execution_server.py` → 
   - `_execute_file_editor()` → 
   - `OHEditor()` from `openhands-aci` → 
   - Error thrown when `file_text=None`

### Expected Behavior
According to the test cases in `tests/runtime/test_aci_edit.py`:
- `file_text=None` → Should show error (current behavior is correct)
- `file_text=""` → Should create empty file successfully  
- `file_text="content"` → Should create file with content successfully

### Current Implementation
- **Python**: `FileEditAction.file_text: str | None = None` (in `openhands/events/action/files.py`)
- **TypeScript**: `file_text?: string | null` (in `frontend/src/types/core/actions.ts`)

### Problem
The issue suggests that somewhere in the system, a `create` command is being sent with `file_text=None` instead of `file_text=""` when the user intends to create an empty file.

## Potential Causes

1. **Frontend Issue**: The UI might not be properly handling empty file creation, sending `null` instead of empty string
2. **Serialization Issue**: JSON serialization/deserialization might be converting empty strings to `null`
3. **Agent Logic Issue**: The AI agent might be generating `FileEditAction` with `file_text=None` when it should use `file_text=""`
4. **Default Value Issue**: The default value of `None` might be problematic for the `create` command

## Reproduction
- **Environment**: OpenHands Cloud
- **Action**: Attempting to create a file using the file editor
- **Expected**: File should be created (empty or with content)
- **Actual**: Error message displayed

### Option 1: Frontend Validation
Add validation in the frontend to ensure `file_text` is never `null` for `create` commands:
```typescript
// In frontend file edit logic
if (command === 'create' && file_text === null) {
  file_text = '';
}
```

### Option 2: Backend Normalization
Modify `_execute_file_editor()` to normalize `None` to empty string for create commands:
```python
# In action_execution_server.py
if command == 'create' and file_text is None:
    file_text = ''
```

### Option 3: Agent Prompt Improvement
Update agent prompts/microagents to ensure they always provide `file_text` parameter for create commands.

### Option 4: Schema Validation
Add validation in `FileEditAction` to ensure required parameters are present for specific commands.

## Files Involved
- `openhands/runtime/action_execution_server.py` - Main execution logic
- `openhands/events/action/files.py` - FileEditAction definition
- `frontend/src/types/core/actions.ts` - Frontend type definitions
- `tests/runtime/test_aci_edit.py` - Test cases showing expected behavior

## Priority
**High** - This affects core file editing functionality in OpenHands Cloud and impacts user experience.

## Additional Context
- The error is coming from an external dependency (`openhands-aci` v0.3.0)
- The behavior is actually correct according to the API contract, but the user experience suggests this should be handled more gracefully
- This might be a regression if it was working previously, or a new edge case that wasn't properly handled during implementation

neubig:
## Summary

This PR fixes issue #9343 where users encounter the error "Parameter `file_text` is required for command: create" when using the file editor in OpenHands Cloud.

## Problem

Users were experiencing an error when attempting to create files through the file editor interface. The error occurred when `FileEditAction` with `command='create'` was called but `file_text` parameter was `None`. According to the issue analysis, this happens when the system sends `file_text=None` instead of `file_text=""` when the user intends to create an empty file.

## Solution

Added normalization in the `_execute_file_editor` function in `action_execution_server.py` to convert `file_text=None` to `file_text=""` for create commands. This approach:

- Maintains backward compatibility
- Provides a defensive programming solution
- Handles the edge case gracefully
- Doesn't require changes to frontend or agent logic

## Changes

1. **Backend Fix**: Modified `_execute_file_editor` function to normalize `file_text=None` to empty string for create commands
2. **Test Update**: Updated the existing test `test_create_with_none_file_text` to verify that the fix works correctly - now expects successful file creation instead of an error

## Testing

- ✅ Created and ran comprehensive unit tests to verify the fix works correctly
- ✅ Verified that `file_text=None` is normalized to empty string for create commands
- ✅ Confirmed that `file_text=""` and `file_text="content"` still work as before
- ✅ Ensured other commands (not 'create') are not affected by the change
- ✅ All pre-commit hooks pass

## Expected Behavior After Fix

- **Before**: `file_text=None` → Error: "Parameter `file_text` is required for command: create"
- **After**: `file_text=None` → Successfully creates empty file (same as `file_text=""`)

This change improves the user experience in OpenHands Cloud by allowing users to create empty files even when `file_text` is `None`, which can happen due to various reasons in the system (frontend handling, serialization, agent logic, etc.).

Fixes #9343

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/74b322d622604c23a0f2ceab78cedb20)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0c60a14-nikolaik   --name openhands-app-0c60a14   docker.all-hands.dev/all-hands-ai/openhands:0c60a14
```

### DIFF
--- a/openhands/runtime/action_execution_server.py
+++ b/openhands/runtime/action_execution_server.py
@@ -124,6 +124,11 @@ def _execute_file_editor(
     """
     result: ToolResult | None = None
 
+    # Normalize file_text for create command: convert None to empty string
+    # This allows users to create empty files when file_text is None
+    if command == 'create' and file_text is None:
+        file_text = ''
+
     # Convert insert_line from string to int if needed
     if insert_line is not None and isinstance(insert_line, str):
         try:
--- a/tests/runtime/test_aci_edit.py
+++ b/tests/runtime/test_aci_edit.py
@@ -133,10 +133,19 @@ def test_create_with_none_file_text(temp_dir, runtime_cls, run_as_openhands):
         )
         obs = runtime.run_action(action)
         logger.info(obs, extra={'msg_type': 'OBSERVATION'})
-        assert (
-            obs.content
-            == 'ERROR:\nParameter `file_text` is required for command: create.'
+        # After the fix, file_text=None should be normalized to empty string
+        # and the file should be created successfully
+        assert 'File created successfully' in obs.content
+
+        # Verify file was created and is empty
+        action = FileEditAction(
+            command='view',
+            path=new_file,
         )
+        obs = runtime.run_action(action)
+        logger.info(obs, extra={'msg_type': 'OBSERVATION'})
+        # Empty file should show just the line number with no content
+        assert '1\t' in obs.content
     finally:
         _close_test_runtime(runtime)
 


enyst:
@neubig Is the cloud version updated with this PR?
- https://github.com/All-Hands-AI/OpenHands/pull/9296

I guess yes, since we just released? I experience the same issue today, and it seems to have stopped when I added
`max_output_tokens = 64000` in config.

(Sonnet 4)

The error response is tool call, and it has `finish_reason=length` (which I find weird, with Claude). But it's not a large length, but it might be higher than whatever the default is when we suddenly set `max_output_tokens=None` for Anthropic models.

Cc: @xingyaoww WDYT ? Maybe we hurried a bit, I tested that PR with Claude, but I forced it to output a long text, _it wasn't a tool call_ outputting stuff.

xingyaoww:
> but it might be higher than whatever the default is when we suddenly set max_output_tokens=None for Anthropic models.

yes, i think this could likely be the cause: we are probably setting a too low max output tokens for sonnet?

xingyaoww:
I wonder if we should revert and bring back the default max tokens though..

enyst:
Yes... the confusing thing here is that we had a hardcoded value (and model-specific values from litellm) in that code, because of some issue with local LLMs a long time ago, but after we implemented tool use, that code became useful for the weirdnesses of LLM APIs on the length of parameters of a tool call...

xingyaoww:
I'll close this in favor of https://github.com/All-Hands-AI/OpenHands/pull/9364

LMK if you have any more concerns!

xingyaoww:
> the confusing thing here is that we had a hardcoded value (and model-specific values from litellm) in that code, because of some issue with local LLMs a long time ago, but after we implemented tool use, that code became useful for the weirdnesses of LLM APIs on the length of parameters of a tool call

@enyst yeah 😓 i think the reason why we did that was some models doesn't support a very long output tokens, so we can't just set one large enough value for everything - and if we keep it at the default value like we did a few days ago, it won't work well (i.e., the model can be cut off while generating a tool call, causing itself into a loop).

PR #9360, opened 2025-06-25
Add user-configurable settings into the UX/frotnend

thefaftek-git:
# Add User-Configurable LLM Parameters and Agent Settings to Frontend

Users can now configure LLM parameters and agent capabilities directly in the OpenHands frontend without editing config files. This includes adjusting temperature, token limits, cost tracking, and enabling/disabling specific agent tools like web browsing, Jupyter notebooks, and command execution.

---

## Summary

This PR adds comprehensive user-configurable settings for LLM parameters and agent capabilities to the OpenHands frontend. Users previously had to edit configuration files to customize their AI agent's behavior - now they can do it through an intuitive settings interface.

### LLM Configuration Parameters
Added 7 configurable LLM parameters to the existing settings interface:
• `temperature` (0.0-2.0) - Controls response creativity and randomness
• `top_p` (0.0-1.0) - Nucleus sampling parameter for response diversity  
• `max_output_tokens` - Maximum tokens in LLM responses
• `max_input_tokens` - Maximum tokens in input prompts
• `max_message_chars` - Character limit for individual messages
• `input_cost_per_token` - Cost tracking for input tokens
• `output_cost_per_token` - Cost tracking for output tokens

### Agent Capabilities Configuration
Created a new dedicated Agent Settings page with 10 configurable capabilities:
• **Tool Access**: Web browsing, Jupyter notebooks, command execution
• **Editor Options**: LLM-based editor vs. standard string replacement editor
• **Agent Behavior**: Thinking steps, task completion, prompt extensions, history management
• **Microagent Control**: Ability to disable specific microagents

### Implementation Details

**Backend Changes:**
• Extended Settings model with new parameters and proper defaults
• Updated API routes to handle new settings
• Added config mapping to maintain backward compatibility with existing config files

**Frontend Changes:**
• Added new React components and form handling for LLM parameters in advanced settings
• Created dedicated Agent Settings page with intuitive UI organization
• Implemented proper validation, dirty state tracking, and error handling
• Added comprehensive internationalization support for 15+ languages

**Testing & Quality:**
• Updated mock handlers and ensured all existing tests pass
• All TypeScript compilation and linting checks pass
• No unlocalized strings or translation gaps

## Impact

**Before:** Users had to manually edit configuration files to customize LLM parameters or agent capabilities, which was technical and error-prone.

**After:** Users can easily configure their AI agent through intuitive settings pages with helpful descriptions, validation, and immediate feedback.

## Testing

All existing tests pass, including:
• ✅ Backend unit tests for new Settings model fields
• ✅ Frontend TypeScript compilation
• ✅ Internationalization validation
• ✅ Form handling and validation tests

---

TODO


thefaftek-git:
@neubig see below! I know it'll definitely need some cleanup and work, more than happy to take suggestions. I'm horrible at UX myself and had claude 4 make most of the UX changes for me.


image

image

image

openhands/server/routes/settings.py
```diff
@@ -62,8 +62,17 @@ async def load_settings(
...
                     provider_tokens_set[provider_type] = provider_token.host
 
+        settings_data = settings.model_dump(exclude={'secrets_store'})
+        # Ensure LLM parameters have default values if None
+        if settings_data.get('temperature') is None:
+            settings_data['temperature'] = 0.0
+        if settings_data.get('top_p') is None:
+            settings_data['top_p'] = 1.0
+        if settings_data.get('max_message_chars') is None:
+            settings_data['max_message_chars'] = 30_000
```
enyst: Maybe we don't need to hardcode default values, we could instead fallback to the configuration options [here](https://github.com/All-Hands-AI/OpenHands/blob/c7dff3e4d28682ceadbf96b0985fb35c06e09607/openhands/core/config/llm_config.py#L66) for example. They are accessible in `openhands_config` (or `llm_config` in this case), though probably not in routes.

(or maybe we don't need this code, if `get_user_settings` does fallback?)

thefaftek-git: @enyst have a preference on falling back to llm_config or ensuring get_user_settings is the fallback? Happy to get rid of hardcoding wherever we can.

enyst: I think the way this should work for now is to fallback to llm_config, but not here, rather in `get_user_settings` or even its callee, `file_settings_store`'s `load`. Maybe by calling `Settings.from_config` then override with the filestore values. 

(I took a quick look and I don't see it done anywhere. Strangely I was under the impression that we do it somewhere, but apparently not anymore. `from_config` is unused at the moment... 😅 but you may want to take a look, seems like exactly what we may want)

frontend/src/routes/agent-settings.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+        <div className="space-y-4">
+          <h3 className="text-lg font-semibold">
+            {t(I18nKey.SETTINGS$EDITOR_SETTINGS)}
+          </h3>
+
+          <SettingsSwitch
+            testId="enable-llm-editor-switch"
+            name="enable-llm-editor"
+            defaultIsToggled={settings.ENABLE_LLM_EDITOR}
```
enyst, line 120: This one is a bit problematic because, I think, it comes together with defining another "draft editor" LLM. One could enable it and let it fallback to the main LLM, idk if anyone uses it that way though.

frontend/src/routes/agent-settings.tsx
```diff
@@ -0,0 +1,302 @@
...
+            {t(I18nKey.SETTINGS$ENABLE_BROWSING)}
+          </SettingsSwitch>
+          <div className="text-sm text-neutral-400">
+            {t(I18nKey.SETTINGS$ENABLE_BROWSING_HELP)}
+          </div>
+
+          <SettingsSwitch
+            testId="enable-jupyter-switch"
+            name="enable-jupyter"
+            defaultIsToggled={settings.ENABLE_JUPYTER}
```
enyst: I'm not sure this is useful, it's like, when it's necessary it should be enabled or the agent won't work. This is a problem with the option itself, not your PR, I'm just saying maybe having Jupyter options in the UI doesn't help.

frontend/src/mocks/handlers.ts
```diff
@@ -31,6 +31,25 @@ export const MOCK_DEFAULT_USER_SETTINGS: ApiSettings | PostApiSettings = {
...
   user_consents_to_analytics: DEFAULT_SETTINGS.USER_CONSENTS_TO_ANALYTICS,
   max_budget_per_task: DEFAULT_SETTINGS.MAX_BUDGET_PER_TASK,
+  // LLM parameters
+  temperature: DEFAULT_SETTINGS.TEMPERATURE,
+  top_p: DEFAULT_SETTINGS.TOP_P,
+  max_output_tokens: DEFAULT_SETTINGS.MAX_OUTPUT_TOKENS,
+  max_input_tokens: DEFAULT_SETTINGS.MAX_INPUT_TOKENS,
+  max_message_chars: DEFAULT_SETTINGS.MAX_MESSAGE_CHARS,
+  input_cost_per_token: DEFAULT_SETTINGS.INPUT_COST_PER_TOKEN,
+  output_cost_per_token: DEFAULT_SETTINGS.OUTPUT_COST_PER_TOKEN,
```
enyst, line 41: I tend to think that these can be useful, for some special cases. Litellm library, which we use, knows a lot of prices, so normally people don't have to configure these, though.

thefaftek-git: @enyst do you think this should be removed, hidden based on model, or what kind of solution do you think would be the right approach? 

I plan on using this with devstral based on my average tok/s and electric costs in order to start analyzing the data, but litellm does have accurate data for the most part. Maybe it would be more useful on the BYOC option, though, where companies will have custom deals on pricing?

frontend/src/routes/agent-settings.tsx
thefaftek-git: Due to the potential for confusion or breaking things, does it make sense to comment out this setting until there's further discussion on the long-term plan with jupyter?

frontend/src/routes/agent-settings.tsx
thefaftek-git: I didn't see that in the settings/config, but I could be wrong and not understand exactly how this setting works. My understand was just that it handles edits via a different process that doesn't require exact string matching.

I'm using this instead of string matching locally and it works pretty well with devstral, I have a lot less editing errors.

enyst: Oh, okay, that’s good to know!

What I mean is, you can set up [another LLM](https://docs.all-hands.dev/usage/llms/custom-llm-configs#draft-editor-configuration) to be the “draft editor”. The idea is that one LLM requests an edit, and the other tries to apply it to the file. Presumably it’s able to apply some edits that aren’t strictly correct, like missing whitespaces or something. If you never set up a draft LLM, it will fallback to the main LLM for both functions.

I understand there is some work in progress to fine tune small models to perform drafting as good as possible.

But if you’re using it as such, that’s great! Love to hear. 😊

frontend/src/routes/agent-settings.tsx
enyst: Yes, or remove it? Maybe we could rethink/remove it in the backend too, not sure.

frontend/src/mocks/handlers.ts
enyst: I think best is to remove it, for now. The reason is that, it seems to me, it simplifies things for the PR. 

The folks at All-Hands have a cloud version built on OpenHands, and last I know, it has no BYOK, so it presumably uses it in the backend, but not expose it to users. It’s this same web UI and server; so the code can get ugly and fragile - idk, imagine code like, if this is running, show, if that, hide; then the Settings model should/should not accept it; then the routes… I honestly don’t really want to see how fragile the code could get.🥹

This is temporary, one way or the other. It is being fixed. But since it’s unlikely to matter a lot for us, thanks to litellm, maybe we can leave it out of the PR? 🙏

enyst: Maybe you can use the config.toml file to configure your values as you want them to be. If we fix the fallback we discussed in the other comment, they should be applied, just the same, for you?

frontend/src/routes/agent-settings.tsx
thefaftek-git: With this in mind, any recommendations on if we we should just leave the setting as-is or if there's other settings we need to tie it with?

openhands/server/routes/settings.py
thefaftek-git: let me take a look and see what's possible, may need a little time on this one!

frontend/src/mocks/handlers.ts
thefaftek-git: I was looking again, could max_input_tokens and max_message_chars be useful for models like devstral? My input and output cost part is definitely an edge-case, but I could see benefits in letting someone ensure this setup matches what's in something like LM Studio without needing to change configs a ton(the goal of this PR).

enyst: Are you normally using `config.toml` for LLM configuration? Does it actually work for you to use this in the file (and have it respected)?

frontend/src/routes/agent-settings.tsx
thefaftek-git: @openhands-agent  can you please remove this specific switch from agent-settings.tsx?

enyst: The openhands-agent will answer if called from your repo 😊

thefaftek-git: Yep, issue on my side, but pushed a fix for this, want to verify?

openhands/server/routes/settings.py
thefaftek-git: pushed a potential fix, will do some testing with it and report back.

openhands/server/session/session.py
```diff
@@ -116,6 +116,11 @@ async def initialize_agent(
             or settings.sandbox_runtime_container_image
             else self.config.sandbox.runtime_container_image
         )
+        # Apply remote_runtime_resource_factor from settings
+        if settings.remote_runtime_resource_factor is not None:
+            self.config.sandbox.remote_runtime_resource_factor = (
+                settings.remote_runtime_resource_factor
```
enyst, line 122: This was written by `openhands-agent`? IMHO we shouldn't add this code, if it's not solving an actual problem right now, or does it?

thefaftek-git: it committed to the wrong branch :( Was offline most of last week due to some welding eye damage, starting to go back through all of this now.

frontend/src/mocks/handlers.ts
thefaftek-git: both these settings were working in the UX,  honestly I haven't used config.toml since I started on the UX work. 

Let me make a note to test both config.toml and UX scenarios with packet captures and follow-up to confirm behavior.

mamoodi:
This seems to be in progress right now. 
@rbren and @amanape just want to make sure you are aware of this? This is a large change to the UI.

thefaftek-git:
@enyst I did some testing and as far as I can tell, all the settings ARE all being passed to litellm, whether or not they're useful from there, all depends on provider.

I was also thinking on this more, though, and maybe it makes sense to only add a few of the really critical settings for LLMs to work like temperature and top_p as direct UX settings and add another page that lets the user modify the config.toml in the UX instead of constantly adding more settings to the UX by default? Note that the temperature setting IS critical in a lot of models producing usable results(in particular devstral), so that's why I really would suggest that be a direct UX setting.

A UX page to edit config.toml could allow for showing a list of all configurable settings/docs in the future, too, which would solve my concern around lowering barrier to entry for newcomers. 

If you think it's worthwhile I'll abandon this PR and just get a small, tailored one for the temperature and top_p settings, then plan out a second one for the config.toml editor

enyst:
Thank you @thefaftek-git !

I think your proposal sounds very good. I actually want to add the `config.toml` settings to the vscode extension, for people to have at their disposal _visually_, as you said, in the familiar IDE interface.

Just to note, a companion web page for the toml probably implies a part of this issue (the save to toml part)
- https://github.com/All-Hands-AI/OpenHands/issues/9585

thefaftek-git:
> Thank you @thefaftek-git !
> 
> I think your proposal sounds very good. I actually want to add the `config.toml` settings to the vscode extension, for people to have at their disposal _visually_, as you said, in the familiar IDE interface.
> 
> Just to note, a companion web page for the toml probably implies a part of this issue (the save to toml part)
> 
> * [Unify Configuration Architecture: Eliminate Parallel Systems #9585](https://github.com/All-Hands-AI/OpenHands/issues/9585)

Awesome!

I’ll make a new tailored PR for just those two UX settings to get them working since your issue is a lot larger, I may not put in fallback to config.toml at first just to have an MVP and reduce conflicts

rbren:
Exposing top_p and temperature in the LLM Settings page does make sense to me.

But agreed I think we should be tactical about what settings we expose in the UI vs in config. Things like cost per token seem like system-level settings rather than stuff that should be configurable on the fly

enyst:
> Exposing top_p and temperature in the LLM Settings page does make sense to me.
> 
> But agreed I think we should be tactical about what settings we expose in the UI vs in config. Things like cost per token seem like system-level settings rather than stuff that should be configurable on the fly

Things like cost per token are _LLM-specific_, though, they are not system-level things.

This is an interesting and important issue, IMO: 
- on one hand, I can see how they're not exposed to the user from the perspective of a multi-user remote OpenHands service.
- but they are definitely LLM-specific, which means totally per LLM and under the control of the user in buckets of settings per LLM.

So my read is, you may want a different "settings" object, something where
- you filter out things you don't need (or have security impact, even, on remote), 
- and may add things you wish to have, which have no meaning in OpenHands (e.g. features like solvability or other things extending the base software).

There are multiple ways to solve this. We could start from the simplest, just inherit the thing, adapt storage, _do add remote-specific filters_, in the right place so that it's possible to guarantee that it won't get broken next time.
- such as, build Settings via a factory method or builder, so that it never ever gets built without them.

Just some thoughts. Surely, we can design this in a way like above, suitable for all constraints/desires here, if we are all willing to focus on it and fix this?  @rbren

PR #9361, opened 2025-06-25
feat: add container network mode support via USE_CONTAINER_NETWORK env var flag

llamantino:

Added support for Docker's container:<name> network mode (container network).

---

When USE_CONTAINER_NETWORK=true is set:
- The runtime container uses the network stack of the container running the app.
- Port mappings are disabled, and the local URL is forced to 127.0.0.1.
- If DOCKER_HOST_ADDR or use_host_network are set, they are overridden.

---

### DIFF
--- a/openhands/runtime/impl/docker/docker_runtime.py
+++ b/openhands/runtime/impl/docker/docker_runtime.py
@@ -38,6 +38,8 @@ from openhands.utils.tenacity_stop import stop_if_should_exit
 
 CONTAINER_NAME_PREFIX = 'openhands-runtime-'
 
+USE_CONTAINER_NETWORK = os.environ.get('USE_CONTAINER_NETWORK', 'false').lower() == 'true'
+
 EXECUTION_SERVER_PORT_RANGE = (30000, 39999)
 VSCODE_PORT_RANGE = (40000, 49999)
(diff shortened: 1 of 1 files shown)


openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -271,6 +290,42 @@ def _process_volumes(self) -> dict[str, dict[str, str]]:
 
         return volumes
 
+    @staticmethod
+    def _detect_current_container_name() -> str | None:
+        try:
+            client = DockerRuntime._init_docker_client()
+
+            # Try using the full hostname (should match container ID)
+            import socket
```
rbren, line 299: imports should go at the top

openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -38,6 +38,8 @@
 
 CONTAINER_NAME_PREFIX = 'openhands-runtime-'
 
+USE_CONTAINER_NETWORK = os.environ.get('USE_CONTAINER_NETWORK', 'false').lower() == 'true'
```
rbren, line 41: this should be in the sandbox config rather than its own env var

PR #9366, opened 2025-06-25
Change default max_output_tokens to None and add comprehensive model tests

neubig:
This PR makes two key changes to how `max_output_tokens` is handled:

1. **Changes the default `max_output_tokens` from 4096 to None**
   - This allows models to use their native maximum output token limits
   - Prevents unnecessary truncation of responses for models with larger context windows

2. **Adds comprehensive tests for `max_output_tokens` handling across different models**
   - Tests for Claude Sonnet 4 models (properly mocks litellm to return 64000 token limit)
   - Tests for Claude 3.7 Sonnet models (verifies 64000 token limit)
   - Tests for SambaNova DeepSeek-V3-0324 model (verifies 32768 token limit from litellm)
   - Tests for fallback behavior when only max_tokens is available
   - Tests for when no token information is available from litellm
   - Verifies correct parameter passing to different model providers (max_tokens vs max_completion_tokens)

These changes ensure that OpenHands properly handles token limits for all supported models, particularly newer models with larger context windows like Claude Sonnet 4, Claude 3.7 Sonnet, and SambaNova DeepSeek-V3.

All tests are passing, including the updated tests that now expect `None` as the default value instead of 4096.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/113a6e2905bd47559571ee7121d9a73c)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:18318e5-nikolaik   --name openhands-app-18318e5   docker.all-hands.dev/all-hands-ai/openhands:18318e5
```


neubig:
@openhands Run the py-unit-tests.yml workflow adding a high enough timeout parameter. Keep running it until it passes and then push back to the branch. Monitor CI. And if CI doesn't pass, rate the logs to the local directory with GitHub API and then debug and fix and push back again. Continue until succeeds.

neubig:
@openhands you must not mock the litellm completion call, but rather mock the actual call that is performed within litellm that calls out to the language model provider (using python requests or whatever other library)

neubig:
@openhands merge this branch openhands-workspace-bowj508t

Then wait 120 seconds and check GitHub CI through the GitHub API. If any CI is failing download the logs and browse them locally to find exactly what errors are happening. Run the corresponding GitHub workflow step by step and fix any errors that you encounter (Make sure to use the timeout parameter for long-running operations.) When you are done, push back to the branch associated with this PR, and go back to monitoring github CI and fixing any errors.

neubig:
@openhands Please run pre-commit on all the files in this PR.

neubig:
@openhands merge main and fix merge conflicts

tests/unit/test_llm.py
```diff
@@ -981,3 +981,108 @@ def test_llm_base_url_auto_protocol_patch(mock_get):
...
+    This test verifies that Claude Sonnet 4 models get the 64000 max_output_tokens value.
+    """
+    # Create LLM instance with Claude Sonnet 4 model
+    config = LLMConfig(model='claude-sonnet-4-20250514', api_key='test_key')
+    llm = LLM(config)
+
+    # Verify the config has the correct max_output_tokens value
+    assert llm.config.max_output_tokens == 64000  # Special value for Claude models
+
+
```
neubig:
```suggestion
```

neubig:
OK, I made some changes and tested this and everything seems to be working as expected. I specifically tested claude-3.7-sonnet locally, and it was able write out a file of `7,906 tokens` in a single step, so I think that with the tests and this manual testing this seems OK.

enyst:
I went through the PR and I would note (please correct me if any of these seems off)
- testing 3.7 writing out a file is better than my previous test, which had 13k tokens in a message, because writing a file is a tool call
- the unit test on Sonnet 4 shows that litellm’s model_info will return 64k, which is good
- compared to `main`, this PR doesn’t differ for models that do have `model_info`
- it makes it None for models that litellm doesn’t know (and the user doesn’t override)
    - this raises a little question, do we have values from litellm, when using: litellm_proxy/anthropic/model_name, litellm_proxy/model_name, bedrock/bedrock_style_model_name, litellm_proxy/model_name (but with bedrock first, like AH proxy is or was), where “model_name” = model name from anthropic/main provider
    - … I think it will not work for models from the proxy where the user has configured a different name for the model than its native name (e.g. we could name the proxy model for Sonnet 4 `litellm_proxy/sonnet-experiment`, and then I think it won’t get `model_info` from litellm, even though the name corresponds to `Sonnet-4-…`) in the proxy config => it will be None
    - does it matter? Users have other problems too on litellm_proxy when the model name is not the configured name, and we recommend to make it the same
- I started an eval, to see that Sonnet 4 works in practice.

mamoodi:
- submitted instances: 50
- empty patch instances: 0
- resolved instances: 35
- unresolved instances: 15
- error instances: 0 - Download at: https://github.com/All-Hands-AI/pr-eval-results/releases/download/1.0.0/add-max-output-tokens-test-93_25-06-29-08-12.tar.gz

neubig:
Great, thank you!

PR #8029, opened 2025-04-23
Add MCP configuration visualization and editing in settings modal

xingyaoww:
This PR adds visualization and editing capabilities for MCP (Model Context Protocol) configurations in the settings modal. It includes:

- Added MCPConfigViewer component to display MCP configurations
- Added MCPConfigEditor component for editing MCP configurations
- Added dedicated MCP tab in settings modal for better organization
- Updated settings form to include MCP configuration section
- Added translations for MCP configuration
- Updated settings utils to handle MCP configuration in form submission

This builds on PR #7911 which added MCP refactoring and support for stdio and running MCP server in runtime.

image
image
image
image
image
image

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:07b0f7a-nikolaik   --name openhands-app-07b0f7a   docker.all-hands.dev/all-hands-ai/openhands:07b0f7a
```

PR #9379, opened 2025-06-26
chore - Don't build ubuntu image on PR

raymyers:

---

The ubuntu image takes a long time to build and doesn't usually provide extra feedback in PRs (with rare exceptions), so removing it from PRs to only run on main's ci. This was previously the configuration but was rolled back as a side effect of #8029
---


raymyers:
I think long term (soon even) we should extract a base runtime image in multiple flavors to its own repo with a separate build. This will help us in the meantime.

neubig (approved):

PR #9382, opened 2025-06-26
fix openhands cli loglevel

seiriosPlus:

Fixed CLI mode ignoring `LOG_LEVEL` environment variable. Users can now control log verbosity in CLI mode using either `LOG_LEVEL=DEBUG` or the new `--log-level` argument, enabling better debugging of agent execution flows.

---

This PR fixes a critical logging issue in CLI mode where the log level was hardcoded to WARNING, preventing users from seeing DEBUG and INFO level logs even when `LOG_LEVEL=DEBUG` was set.

**Changes made:**
1. **Removed hardcoded log level**: Commented out `logger.setLevel(logging.WARNING)` in `openhands/cli/main.py` that was overriding environment variable settings
2. **Added CLI argument support**: Added `--log-level` parameter to the argument parser with choices ['DEBUG', 'INFO', 'WARNING', 'ERROR']
3. **Implemented dynamic log level handling**: Added logic to respect CLI arguments when provided, otherwise fall back to environment variable `LOG_LEVEL` (handled by existing logger initialization)
4. **Maintained backward compatibility**: Default behavior remains unchanged when no log level is explicitly specified

**Design decisions:**
- CLI argument takes precedence over environment variable for explicit user control
- When CLI argument is provided, both logger and console handler levels are updated to ensure proper output
- Added `else` clause to maintain WARNING as default when neither CLI arg nor env var specify a level
- Used standard Python logging level names for consistency

**Usage examples:**
```bash
# Using environment variable
LOG_LEVEL=DEBUG python -m openhands.cli.main --task "debug my code"

# Using CLI argument  
python -m openhands.cli.main --log-level DEBUG --task "debug my code"
```

### DIFF
--- a/openhands/cli/main.py
+++ b/openhands/cli/main.py
@@ -352,7 +352,17 @@ async def main_with_loop(loop: asyncio.AbstractEventLoop) -> None:
     """Runs the agent in CLI mode."""
     args = parse_arguments()
 
-    logger.setLevel(logging.WARNING)
+    # Set log level from command line argument if provided
+    if args.log_level:
+        log_level = getattr(logging, args.log_level.upper())
+        logger.setLevel(log_level)
+        # Also update the console handler level
+        for handler in logger.handlers:
+            if isinstance(handler, logging.StreamHandler) and handler.stream.name == '<stderr>':
+                handler.setLevel(log_level)
+    # Otherwise, use environment LOG_LEVEL (handled by openhands_logger initialization)
+    else:
+        logger.setLevel(logging.WARNING)
 
     # Load config from toml and override with command line arguments
     config: OpenHandsConfig = setup_config_from_args(args)


enyst:
Thank you, @seiriosPlus ! I can see why it feels bugged.

IMHO we should indeed respect `LOG_LEVEL` if the user set it, but I think warning level by default makes sense for the CLI, because otherwise the log messages bump into the same interface. To be honest, I’m a logging freak and personally I love the clean TUI. It’s not the same with the web UI, which is in different places and the user can look at one or the other. Am I missing something?

seiriosPlus:
> Thank you, @seiriosPlus ! I can see why it feels bugged.
> 
> IMHO we should indeed respect `LOG_LEVEL` if the user set it, but I think warning level by default makes sense for the CLI, because otherwise the log messages bump into the same interface. To be honest, I’m a logging freak and personally I love the clean TUI. It’s not the same with the web UI, which is in different places and the user can look at one or the other. Am I missing something?

hi @enyst ，I need to delve deeper into Agent development and details based on Openhands, so that I can understand the call details thoroughly. When I develop a new Agent with Openhands, I need to debug, have a more detailed understanding of the internal details, and need to lower the log level.

enyst:
Got it, thank you. Let's make sure the CLI respects LOG_LEVEL env var, so you can set it and see the logs you wish?

seiriosPlus:
> Got it, thank you. Let's make sure the CLI respects LOG_LEVEL env var, so you can set it and see the logs you wish?

Yes, I have tried the following commands, and they all worked:
* LOG_LEVEL=DEBUG python -m openhands.cli.main -c MyAgent
* python -m openhands.cli.main -c MyAgent --log-level DEBUG

enyst:
Could we keep WARNING by default? If the user does nothing with log level, just installs from pypi and runs, the PR now applies info, I think.

seiriosPlus:
> Could we keep WARNING by default? If the user does nothing with log level, just installs from pypi and runs, the PR now applies info, I think.

yeah, I have change the config to WARNING by default @enyst

openhands/cli/main.py
```diff
@@ -356,14 +356,12 @@ async def main_with_loop(loop: asyncio.AbstractEventLoop) -> None:
...
-                isinstance(handler, logging.StreamHandler)
-                and handler.stream.name == '<stderr>'
-            ):
-                handler.setLevel(log_level)
-    # Otherwise, use environment LOG_LEVEL (handled by openhands_logger initialization)
+    else:
+        # Set default log level to WARNING if no LOG_LEVEL environment variable is set
+        # (command line argument takes precedence over environment variable)
+        env_log_level = os.getenv('LOG_LEVEL')
+        if env_log_level is None:
```
enyst:
```suggestion
        if not env_log_level:
```

enyst:
Thank you! Could you please confirm if you tested it and it works for you, with `log_level` as a command line parameter?

seiriosPlus:
> Thank you! Could you please confirm if you tested it and it works for you, with `log_level` as a command line parameter?

@enyst  --log_level is an unauthorized parameter

usage: main.py [-h] [-v] [--config-file CONFIG_FILE] [-d DIRECTORY] [-t TASK] [-f FILE] [-c AGENT_CLS] [-i MAX_ITERATIONS] [-b MAX_BUDGET_PER_TASK] [--eval-output-dir EVAL_OUTPUT_DIR]
               [--eval-n-limit EVAL_N_LIMIT] [--eval-num-workers EVAL_NUM_WORKERS] [--eval-note EVAL_NOTE] [-l LLM_CONFIG] [--agent-config AGENT_CONFIG] [-n NAME] [--eval-ids EVAL_IDS]
               [--no-auto-continue] [--selected-repo SELECTED_REPO] [--override-cli-mode OVERRIDE_CLI_MODE] [--log-level {DEBUG,INFO,WARNING,ERROR}]

enyst:
So I think we need to add it to those many parameters, or maybe just remove it and keep the env var for the user to set their CLI log level to what they want?

seiriosPlus:
> So I think we need to add it to those many parameters, or maybe just remove it and keep the env var for the user to set their CLI log level to what they want?

Using log-level as a request parameter or using the environment variable LOG_LEVEL is a common practice in the industry. It's less common to use log_level as a CLI parameter. Currently, Openhands has achieved good compatibility. Should we maintain the status quo?

seiriosPlus:
<html><head></head><body>

Usage | Common/Mainstream | Recommended Format
-- | -- | --
Environment Variable | ✅ Yes | LOG_LEVEL
Command-Line Argument | ✅ Yes | --log-level
Config Field / Variable Name | ✅ Yes | log_level
log_level as CLI Argument | ❌ Rare / Not Recommended | ❌ (Should be --log-level)
</body></html>

enyst:
Sorry, I’m not following. Does `—log-level` work for you in testing as CLI parameter?

seiriosPlus:
> —log-level

yeah,  I have tried the following commands, and all of them worked @enyst :
LOG_LEVEL=DEBUG python -m openhands.cli.main
python -m openhands.cli.main --log-level DEBUG

PR #9085, opened 2025-06-11
Add CLI/vscode integration

enyst:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

This PR proposes a CLI/vscode integration via an extension
---
**Link of any specific issues this addresses:**

PR #9398, opened 2025-06-26
feat(CLI): Enhance --file option to prompt agent to read and understand file first

xingyaoww:

This PR enhances the behavior of the `--file` option in the OpenHands CLI. When you use `--file` to provide a file to OpenHands, it now automatically prompts the agent to read and understand the file content first, then ask you what you'd like to do with it. This makes the file option more interactive and user-friendly.

---

This PR enhances the behavior of the `--file` option in the OpenHands CLI. Instead of directly inputting the content of the file as the task, it now wraps the file content in a prompt that instructs the agent to:

1. Read and understand the file content first
2. Ask the user what they would like to do with the file

The implementation is straightforward - when a file is provided via the `--file` option, we read its content and wrap it in a prompt template that instructs the agent to first understand the file and then ask for further instructions.

Code organization improvements:
- Merged the test file `test_cli_file_option.py` into the main `test_cli.py` file to reduce the number of test files
- Added comprehensive tests for the file option functionality

---

N/A

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:d1a3f33-nikolaik   --name openhands-app-d1a3f33   docker.all-hands.dev/all-hands-ai/openhands:d1a3f33
```


xingyaoww:
@OpenHands, can we handle this special case in the CLI code instead of modifying `read_task_from_file`, as it will be used elsewhere?
i.e., the CLI shouldn't use read_task_from_file maybe, instead they should load the file and send it via a --task prompt

tests/unit/test_io.py
```diff
@@ -25,3 +28,50 @@ def test_multiline_input():
...
+
+    try:
+        # Call the function with the temporary file
+        result = read_task_from_file(temp_file_path)
+
+        # Check that the result is exactly the file content
+        assert result == 'This is a test file content.'
+    finally:
+        # Clean up the temporary file
+        os.unlink(temp_file_path)
```
xingyaoww:
```suggestion
```

xingyaoww:
@OpenHands you shouldn't modify anything in io.py since it will be reused by something else. Can you modify something in the CLI launch code directly for this purpose?

xingyaoww:
@OpenHands can you put the changes in tests/unit/test_cli_file_option.py to merge together with other test cli file so we don't end up with too many test_cli* file everywhere?

When you are done, pls update the PR description of this PR following PR template of this repo

tests/unit/test_cli.py
```diff
@@ -831,3 +834,140 @@ async def test_config_loading_order(
...
+        assert f"The user has tagged a file '{temp_file_path}'" in result
+        assert 'Please read and understand the following file content first:' in result
+        assert 'This is a test file content.' in result
+        assert (
+            'After reviewing the file, please ask the user what they would like to do with it.'
+            in result
+        )
+    finally:
+        # Clean up the temporary file
+        os.unlink(temp_file_path)
```
xingyaoww: @OpenHands please remove this hack test, it doesn't do anything

xingyaoww:
@OpenHands please fix the failing actions on PR https://github.com/All-Hands-AI/OpenHands/pull/9398

openhands/io/io.py
```diff
@@ -31,7 +31,17 @@ def read_task(args: argparse.Namespace, cli_multiline_input: bool) -> str:
...
+        # For CLI usage, enhance the file content with a prompt
+        # that instructs the agent to read and understand the file first
+        task_str = f"""The user has tagged a file '{args.file}'.
+Please read and understand the following file content first:
+
+```
+{file_content}
+```
+
+After reviewing the file, please ask the user what they would like to do with it."""
```
enyst: OH's last commit is probably not ok, because `read_task` here is used by headless mode too.

🤔 I wonder, I think we probably need to move it back to CLI, but I have a small question here:

The `--file` parameter was intended for a task file, a .md file with instructions. The `read_task` function reads and sends it to the LLM as initial instructions. This PR changes it with interactive message, because we expect it to be a source code file now, but headless mode still expects `--file` to be instructions (otherwise it doesn't make much sense, and it adds a contradiction in the prompt)

Maybe it doesn't matter a lot, but we do have three options:

- do it in python; the inconsistency is probably rarely visible, or anyway minor
- make another arg (--source-file 😅), though it seems overkill!
- do it in the extension.

For example, OH made some of this work from the extension side:
[contextual message](https://github.com/All-Hands-AI/OpenHands/pull/9085/commits/c0e40417021c5675f9737b97dc0e7fc21424b2cc#diff-be024897b6c700ba6303945048835bef55f19994a46e57026634ae3a8b032d40R221)

WDYT?

enyst: I take that back. The performance of sending full content from the extension is terrible, if it even works. The way the vscode-integration PR sends stuff now (with `sendText`) doesn't work for many files because they're too large.

We could do it in the extension if we only add the prompt + file name, and let the agent read it.

Maybe the best is option 1, do it in python, in the CLI though, not headless. 🤔

xingyaoww: dang... i was explictly instructing it NOT to modify read_file somehow it did that again :(

xingyaoww: reverted - now we do it in CLI!

enyst (approved):
The PR looks good to me, although OH will need to fix the tests. Hopefully only tests 😅

xingyaoww:
@OpenHands please fix the failing actions on PR https://github.com/All-Hands-AI/OpenHands/pull/9398

Do NOT modify the existing implementation!! Just figure out what's wrong with test!

Issue #8591, opened 2025-05-20
#ieeesoc [Bug]: "Launch" Button Loads When Clicking "Launch from Scratch" 


denshaw-09:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

**Description:**

When using the OpenHands platform, I noticed an unexpected UI behavior where clicking "Launch from Scratch" causes the other "Launch" button to start loading simultaneously. This happens even when no repository or branch is connected.

**Steps to Reproduce:**

- Navigate to the OpenHands interface.
- Click the "Launch from Scratch" button.
- Observe that the other "Launch" button also enters a loading state.
- Refreshing the page does not resolve the issue; the behavior persists.

**Expected Behavior:**

Clicking "Launch from Scratch" should trigger only its respective function without affecting the other "Launch" button.

**Observed Behavior:**

Clicking "Launch from Scratch" causes both buttons to enter a loading state, suggesting either a linked process issue or an unintended UI bug.

**Reference:**

Image1

![Image](https://github.com/user-attachments/assets/db243f1c-d5d8-496e-bba6-9fae6d639326)

Image2

![Image](https://github.com/user-attachments/assets/9b189499-1504-4e5f-909d-b99779bc9e49)

*Possible Causes:*

- The buttons may be unintentionally linked via a shared state or process.
- A front-end bug could be causing incorrect UI interactions.
- Delayed server responses might be affecting UI updates.

Would appreciate any insights from the development team on this! Let me know if you need further details.

### OpenHands Installation

https://www.all-hands.dev/
(viewed it first on the site )

### Operating System

WSL on Windows

amanape:
Thanks for taking the time to create the issue @denshaw-09. The main issue is that we don't show an intermediary/splash screen when creating a conversation. The reason we disable the other buttons is to prevent the user from creating duplicate conversations.

We hope to address this soon by potentially introducing an intermediary "loading" screen, so I will close this issue and any relevant PRs

PR #9411, opened 2025-06-27
fix(frontend): "Launch" Button Loads When Clicking "Launch from Scratch" 

hieptl:

This fix resolves a confusing user experience when starting a new conversation from the home page in OpenHands.

Previously, when clicking one of the launch buttons—either “Launch from Scratch” or “Launch” (after selecting a repository and branch)—both buttons would enter a loading state, even though only one was clicked.

To reproduce the issue:

**Case 1:**
- Go to the OpenHands home page.
- Click “Launch from Scratch”.
- Notice that the “Launch” button also shows a loading spinner, even though it wasn’t clicked.
- Refreshing the page doesn’t fix this behavior.

**Case 2:**
- Go to the OpenHands home page.
- Select a repository and branch.
- Click “Launch”.
- Notice that the “Launch from Scratch” button also shows a loading spinner.
- Refreshing the page doesn’t fix it.

What’s changed:

With this fix, only the button that was clicked will show a loading state, while the other will be disabled but not display as loading—providing a clearer and more accurate user experience.

---

This PR addresses a UI issue where both launch buttons—“Launch from Scratch” and “Launch” (after selecting a repository)—show a loading state simultaneously, even when only one of them was clicked.

The root cause is the shared logic used in both the HomeHeader and RepoSelectionForm components:

```js
const isCreatingConversationElsewhere = useIsCreatingConversation();
...
const isCreatingConversation = isPending || isSuccess || isCreatingConversationElsewhere;
```

When a user clicks either launch button, isCreatingConversationElsewhere becomes true, causing both buttons to display the loading state.

To fix this, we decouple the loading and disabling logic:

```js
const isCreatingConversation = isPending || isSuccess;
```

Now, only the clicked button will show the loading state. We separately use isCreatingConversationElsewhere to disable the other button, preventing duplicate actions while maintaining accurate UI feedback.

---

The PR fix the [issue #8591](https://github.com/All-Hands-AI/OpenHands/issues/8591)


amanape:
Thanks for the PR @hieptl but I will be closing this issue for reasons explained in https://github.com/All-Hands-AI/OpenHands/issues/8591#issuecomment-3019163805. Feel free to open a PR for other issues!

PR #9419, opened 2025-06-27
Fix: Retry on Bedrock ServiceUnavailableError

hamiltop:

We've been seeing this error when using bedrock:
```
litellm.exceptions.ServiceUnavailableError: litellm.ServiceUnavailableError: BedrockException - {"message":"The system encountered an unexpected error during processing. Try your request again."}
```
This seems like a pretty safe things to blindly retry.

---
Adds ServiceUnavailableError to retryable errors

---

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -19,6 +19,7 @@ from litellm import completion as litellm_completion
 from litellm import completion_cost as litellm_completion_cost
 from litellm.exceptions import (
     RateLimitError,
+    ServiceUnavailableError,
 )
 from litellm.types.utils import CostPerToken, ModelResponse, Usage
 from litellm.utils import create_pretrained_tokenizer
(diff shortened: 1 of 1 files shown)


enyst:
FWIW this can go both ways, rate limiting is expected (if you go too fast you will hit some). [Last we discussed](https://github.com/All-Hands-AI/OpenHands/pull/6557#discussion_r1936383970) this, we thought that _unexpected_ errors should be shown to the user: to know something happened and what happened, and let them try it themselves.

The issue it was solving was that retries took too long for all kinds of exceptions that are unexpected, so the user was left in the dark, with a stalled openhands.

(Meanwhile we sinned a bit though, because Gemini API broken behavior forced to retry just in case for weird cases… so I think we have a retry or 2 for other reasons than rate limiting atm.)

hamiltop:
Ok. Is there some way to have headless mode just retry? It's causing our scripted usage to exit.

enyst (approved):
Ah, I see. The way we do for evals is to re-run the script. 🫣

Let’s take this in. We are just looking at retries because they don’t seem to work as expected and might rethink how it should behave.

PR #9423, opened 2025-06-27
fix(cli): avoid race condition from multiple process_agent_pause tasks

llamantino:

Fixes a bug where the terminal could be left in a broken state (raw mode) after exiting the CLI, due to improperly restored terminal settings.

---

Ensures process_agent_pause is only started once, preventing multiple conflicting state snapshots and conflicting restores at shutdown.

---


neubig (approved):

neubig:
I tested this out and indeed it seems to have fixed the problem. Thank you!

PR #9427, opened 2025-06-28
Add setup.sh script execution to event stream

neubig:

This change makes the execution of the setup.sh script visible in the event stream, allowing users to see when and how the setup script is run. This improves transparency and debugging capabilities when working with repository-specific setup scripts.

---

This PR modifies the `maybe_run_setup_script` method in the Runtime base class to add the setup script execution action and its observation to the event stream. Previously, the setup script was run directly without being added to the event stream, making it invisible in the conversation history.

The implementation:
1. Checks if an event stream is available
2. If available, adds the action to the event stream
3. Runs the action and adds the resulting observation to the event stream
4. Falls back to direct execution if no event stream is available

This change ensures that the setup script execution is properly recorded in the conversation history, making it easier to understand what happened during repository initialization.

---

N/A

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/7668e95538d144d08911b6c6e1c49531)

### DIFF
--- a/openhands/runtime/base.py
+++ b/openhands/runtime/base.py
@@ -446,7 +446,21 @@ class Runtime(FileEditRuntimeMixin):
             f'chmod +x {setup_script} && source {setup_script}', blocking=True
         )
         action.set_hard_timeout(600)
-        obs = self.run_action(action)
+
+        # Add the action to the event stream if available
+        if hasattr(self, 'event_stream') and self.event_stream:
+            source = EventSource.AGENT  # Use AGENT as the source
+            self.event_stream.add_event(action, source)
+            obs = self.run_action(action)
+            # Add the observation to the event stream
+            if not isinstance(obs, NullObservation):
+                # Set the cause using setattr to avoid mypy errors
+                setattr(obs, '_cause', action.id)
+                self.event_stream.add_event(obs, source)
+        else:
+            # Fall back to direct execution if no event stream is available
+            obs = self.run_action(action)
+
         if not isinstance(obs, CmdOutputObservation) or obs.exit_code != 0:
             self.log('error', f'Setup script failed: {obs.content}')
 


openhands/runtime/base.py
```diff
@@ -446,9 +446,10 @@ def maybe_run_setup_script(self):
             f'chmod +x {setup_script} && source {setup_script}', blocking=True
         )
         action.set_hard_timeout(600)
-        obs = self.run_action(action)
-        if not isinstance(obs, CmdOutputObservation) or obs.exit_code != 0:
-            self.log('error', f'Setup script failed: {obs.content}')
+
+        # Add the action to the event stream as an ENVIRONMENT event
+        source = EventSource.ENVIRONMENT
+        self.event_stream.add_event(action, source)
```
enyst, line 454: If we add it to the stream, do we want it to be sent to the LLM’s context?

If not, maybe we could add a `hidden` attribute True, on the event?

enyst: At the first pass, when it’s created, it wouldn’t, I think, reach the LLM, but I’m not sure about when the conversation is reloaded.

It may be worth to check with a reloaded conversation. (it would be nice if OH manages to write a unit test for reloading a conversation)

In addition, I’m a bit concerned about the fact that while we removed many hardcoded assumptions, I think we still have _some_ basic assumptions like “system message event, if any, is the first event in the stream” in the code.

neubig: > If we add it to the stream, do we want it to be sent to the LLM’s context?

Good question, I think maybe the answer is "yes"? It would be nice for the LLM to know the result of the setup script so it knows what sorts of things are installed.

> At the first pass, when it’s created, it wouldn’t, I think, reach the LLM, but I’m not sure about when the conversation is reloaded.

You are much more of an expert at this than I am, but I would have assumed that it would be visible to the LLM even on the first pass.

> In addition, I’m a bit concerned about the fact that while we removed many hardcoded assumptions, I think we still have some basic assumptions like “system message event, if any, is the first event in the stream” in the code.

Do you have a place where I could look for this sort of thing? I honestly don't know.

neubig:
@openhands two docker tests are now failing due to the new way of calling setup scripts.

FAILED tests/runtime/test_setup.py::test_maybe_run_setup_script[DockerRuntime-False] - AssertionError: assert False
 +  where False = isinstance(ErrorObservation(content='File not found: /workspace/README.md. Your current working directory is /workspace.', observation=<ObservationType.ERROR: 'error'>, error_id=''), FileReadObservation)
FAILED tests/runtime/test_setup.py::test_maybe_run_setup_script_with_long_timeout[DockerRuntime-False] - AssertionError: assert False
 +  where False = isinstance(ErrorObservation(content='File not found: /workspace/README.md. Your current working directory is /workspace.', observation=<ObservationType.ERROR: 'error'>, error_id=''), FileReadObservation)
 
 Read these tests, modify them to consider the fact that the maybe run setup script just adds the events to the event stream, and run the tests locally by following the steps of the corresponding workflow. Remember that you will have to install docker.

openhands/runtime/base.py
enyst: > Good question, I think maybe the answer is "yes"? It would be nice for the LLM to know the result of the setup script so it knows what sorts of things are installed.
> > At the first pass, when it’s created, it wouldn’t, I think, reach the LLM, but I’m not sure about when the conversation is reloaded.
> You are much more of an expert at this than I am, but I would have assumed that it would be visible to the LLM even on the first pass.

No expert, I may have just stared at debug traces for way too long, if something feels off, I learned it generally may be worth a look or I'll regret it. 😅 

I think we are here too early in the execution flow, we are during runtime initialization, which means we don't have an agent controller, so the agent loop didn't start, so events are not added to history yet. They're added [added by the controller on_event](https://github.com/All-Hands-AI/OpenHands/blob/5099413729f49a28d2606fdc19406d4eed3147a1/openhands/controller/agent_controller.py#L398)

Here we [create the controller](https://github.com/All-Hands-AI/OpenHands/blob/5099413729f49a28d2606fdc19406d4eed3147a1/openhands/server/session/agent_session.py#L175), which is some dozens of lines too late, after the runtime initialization at line 130.

The system message is added [when](https://github.com/All-Hands-AI/OpenHands/blob/5099413729f49a28d2606fdc19406d4eed3147a1/openhands/controller/agent_controller.py#L182) the controller is initialized, so it's _after_ this setup observation.
 
> > In addition, I’m a bit concerned about the fact that while we removed many hardcoded assumptions, I think we still have some basic assumptions like “system message event, if any, is the first event in the stream” in the code.
> 
> Do you have a place where I could look for this sort of thing? I honestly don't know.

Places where we get or check the first events... maybe
- this is run both first session and reloaded sessions  https://github.com/All-Hands-AI/OpenHands/blob/5099413729f49a28d2606fdc19406d4eed3147a1/openhands/controller/state/state_tracker.py#L109
- conversation_memory.py  `_ensure_system_message`, `_ensure_first_user_message`
- the code [checking essential events](https://github.com/All-Hands-AI/OpenHands/blob/5099413729f49a28d2606fdc19406d4eed3147a1/openhands/memory/condenser/impl/conversation_window_condenser.py#L36) for truncation to not lose them
- maybe condensers also have something like this, not sure

enyst: Honestly, my first thought, to add it to history, would be to do as the PR does here, handle all these potential troublesome spots, _and_ make sure in `conversation_memory` that it will be at the right place, which is 5th event in agent's history I think.

And what I just said may be fragile down the road. We almost never re-order events in agent's history. (only condensers do that IIRC, and they have to, to insert the summary at the start)

I may be wrong though, I just feel it's worth looking into.

enyst: What if we save the obs as a log, a log of `setup.sh` run, and read it later?

enyst: Yes, FWIW, I think best would be: not add this to the stream, and instead create a .log file, read it in the same time with reading microagents, maybe also add its results to the `workspace_context` observation that contains other runtime information.

neubig: I think these are good points!

The main point of this PR is not to add it to the history though, it's to make it visible in the UI so users can understand when the script is failing. Maybe we can review/mere this PR first and open a new enhancement issue saying that it should be added to the agent history with this context?

neubig:
@openhands 

We're getting these errors:

```
FAILED tests/runtime/test_setup.py::test_maybe_run_setup_script[DockerRuntime-False] - AssertionError: assert False
 +  where False = isinstance(ErrorObservation(content='File not found: /workspace/README.md. Your current working directory is /workspace.', observation=<ObservationType.ERROR: 'error'>, error_id=''), FileReadObservation)
FAILED tests/runtime/test_setup.py::test_maybe_run_setup_script_with_long_timeout[DockerRuntime-False] - AssertionError: assert False
 +  where False = isinstance(ErrorObservation(content='File not found: /workspace/README.md. Your current working directory is /workspace.', observation=<ObservationType.ERROR: 'error'>, error_id=''), FileReadObservation)
 ```

1. carefully understand the codepath that may be leading to these errors. come up with 5 possible reasons why things may be failing, and trace the full codepath until you have a single reason that you can explain logically and in depth. then fix the issue.
2. install docker
3. follow the ghcr-build.yml workflow to run the corresponding runtime test. set a timeout of 600 to allow time for docker build
4. continue analyzing, fixing, and testing till things pass locally
5. push the changes to github, wait 600 seconds, then check the results of github actions. if they are failing, download the logs and find the errors, then fix them.

File: openhands/runtime/base.py (hunk 1 above)
Comment (enyst) [reply openhands/runtime/base.py:454]: I'm a bit torn on the suggestion, because it seems to me that the difficulty comes from adding it to the event stream before we start the agent.

I mean, we can add it as 'hidden', but then we set up quite a bit of complexity for the time when we want to unhide it.

Maybe we can rethink this, so that we only read the output later, during microagents' loading, and then we will find it straightforward to also show in the UI?

File: openhands/runtime/base.py [hunk 2]
```diff
@@ -446,9 +446,13 @@ def maybe_run_setup_script(self):
             f'chmod +x {setup_script} && source {setup_script}', blocking=True
```
neubig:
```suggestion
            f'chmod +x {setup_script} && source {setup_script}', blocking=True, hidden=True
```

neubig:
@openhands run pre-commit to fix linting

neubig:
@openhands

1. now when the first user message is sent, the "chat-suggestions" window does not disappear (but it should). please fix this
1. python unit tests are failing, run the full py-unit-tests.yml workflow (piping the tests to a file, and sleeping/checking till their done). fix, and continue till all tests pass.

neubig:
@openhands Look at the last commit you made. The commit was supposed to make the chat suggestions disappear when the first user message is sent, but NOT when the setup.sh event is sent. The setup.sh message should have the source "environment" and the user message should have the source "user". Please look at the change that you just made and how the logic for setup.sh messages and user messages work, and try to fix this, ideally also writing a test that checks that this behavior is as expected.

neubig:
OK, I checked that the conversation is loaded appropriately on reloading.

OTOH, while testing I realized one small thing that bothered me (the chat suggestions were not appropriately disappearing on the first user message), so I'm glad I looked more thoroughly.

Screenshot 2025-07-01 at 8 13 30 AM

@enyst, @rbren, or @amanape, this should be ready for review again.

enyst (approved):
Thank you! The ‘hidden’ attribute should make it have no impact on the agent.

PR #9428, opened 2025-06-28
Complete browsing unit tests

### LINKED ISSUE #9336: Add more browsing unit tests

We have a few browsing unit tests in https://github.com/All-Hands-AI/OpenHands/blob/main/tests/runtime/test_browsing.py but they don’t cover a lot of basic actions. Can you please take a look at https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/agenthub/codeact_agent/tools/browser.py to see the list of currently supported browsing actions, and create some unit tests for them?

**Additional context**

Having more unit tests coverage would make transition from browsergym to alternatives (e.g. browser-use) easier.

li-boxuan:

Closes #9336 

---

We have a few browsing unit tests, but they don't cover every browsing action we currently support. This PR completes unit tests.

This gives us more confidence to switch browsergym to something else (let it be raw playwright, or browser-use).

---

### DIFF
--- a/tests/runtime/test_browsing.py
+++ b/tests/runtime/test_browsing.py
@@ -783,12 +783,11 @@ def test_browser_file_upload(temp_dir, runtime_cls, run_as_openhands):
         axtree_elements = parse_axtree_content(obs.content)
         # File inputs often show up as buttons in axtree, try multiple strategies
         file_input_bid = (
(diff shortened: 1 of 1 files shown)


neubig (approved):
I skimmed but these look plausible

PR #9432, opened 2025-06-28
Fixes #9394 - Improve CLI exit messaging to distinguish intentional exits and inter…

### LINKED ISSUE #9394: [CLI] Remove confusing error message when user intentionally exits session
## Problem
When users intentionally exit a session, they see a red error message: "The action has not been executed. This may have occurred because the user pressed the stop button..." which is confusing since the exit was intentional.

## Current Behavior
- User exits session on purpose (via `/exit` or other means)
- Red error message appears suggesting something went wrong
- Message implies unintended interruption when exit was deliberate

## Expected Behavior
- No error message for intentional exits
- Clear, positive confirmation of successful session termination
- Distinguish between intentional exits and actual errors

## User Experience Impact
- **Confusion**: Users think something went wrong when it didn't
- **Anxiety**: Red error messages suggest problems
- **Unprofessional**: Makes the tool seem buggy

## Current Error Message
```
❌ The action has not been executed. This may have occurred because the user pressed the stop button...
```

## Proposed Behavior
For intentional exits:
```
✅ Session terminated successfully
```

For actual errors/interruptions:
```
⚠️ Session was interrupted: [specific reason]
```

## Technical Context
The error message likely comes from the cleanup process or event handling when the session is terminated. The system should distinguish between:

1. **Intentional termination**: User commands like `/exit`, `/new`
2. **User interruption**: Ctrl-C, Ctrl-P pause, etc.
3. **System errors**: Actual failures or exceptions

## Investigation Needed
- Where this error message originates
- How to detect intentional vs unintentional termination
- Proper cleanup messaging for different exit scenarios

## Related Files
- `openhands/cli/main.py` (cleanup_session, exit handling)
- `openhands/cli/commands.py` (handle_exit_command)
- Event handling and session termination logic

Adityauyadav:

This update improves the CLI experience by clearly showing a positive confirmation message when users intentionally exit a session, and a distinct warning when sessions are interrupted unexpectedly. Previously, users saw confusing error messages even on normal exits, causing unnecessary anxiety.

---

This PR introduces an `ExitReason` enum to differentiate between intentional exits, interruptions, and errors. The CLI’s exit handling logic was updated to return and check this reason, enabling it to display clear, context-appropriate messages to users. Additionally, the PR includes friendly handling of keyboard interrupts (Ctrl+C) and updates related tests to accommodate the extended return signature of the `handle_commands()` function. A new unit test for `ExitReason` was added to ensure enum integrity.

---

Fixes #9394

### DIFF
--- a/openhands/core/schema/exit_reason.py
+++ b/openhands/core/schema/exit_reason.py
@@ -1,0 +1,6 @@
+from enum import Enum
+
+class ExitReason(Enum):
+    INTENTIONAL = "intentional"
+    INTERRUPTED = "interrupted"
+    ERROR = "error"
--- a/tests/unit/test_exit_reason.py
+++ b/tests/unit/test_exit_reason.py
@@ -1,0 +1,16 @@
+import pytest
+from openhands.core.schema.exit_reason import ExitReason
+
+def test_exit_reason_enum_values():
+    assert ExitReason.INTENTIONAL.value == "intentional"
+    assert ExitReason.INTERRUPTED.value == "interrupted"
+    assert ExitReason.ERROR.value == "error"
+
+def test_exit_reason_enum_names():
+    assert ExitReason["INTENTIONAL"] == ExitReason.INTENTIONAL
+    assert ExitReason["INTERRUPTED"] == ExitReason.INTERRUPTED
+    assert ExitReason["ERROR"] == ExitReason.ERROR
+
+def test_exit_reason_str_representation():
+    assert str(ExitReason.INTENTIONAL) == "ExitReason.INTENTIONAL"
+    assert repr(ExitReason.ERROR) == "<ExitReason.ERROR: 'error'>"
--- a/openhands/cli/commands.py
+++ b/openhands/cli/commands.py
@@ -1,6 +1,8 @@
 import asyncio
 from pathlib import Path
 
+from openhands.core.schema.exit_reason import ExitReason
+
 from prompt_toolkit import print_formatted_text
 from prompt_toolkit.shortcuts import clear, print_container
 from prompt_toolkit.widgets import Frame, TextArea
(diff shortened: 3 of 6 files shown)


neubig:
But linting is failing so could you please run pre-commit

tests/unit/test_exit_reason.py
```diff
@@ -0,0 +1,16 @@
+import pytest
+from openhands.core.schema.exit_reason import ExitReason
+
+def test_exit_reason_enum_values():
+    assert ExitReason.INTENTIONAL.value == "intentional"
+    assert ExitReason.INTERRUPTED.value == "interrupted"
+    assert ExitReason.ERROR.value == "error"
+
+def test_exit_reason_enum_names():
```
enyst, line 16: Could we add a unit test for the functionality too, apart from definitions?

Adityauyadav:
@neubig 
> But linting is failing so could you please run pre-commit

I’ve now run pre-commit and committed all the auto-fixes.
Linting issues have been resolved — the only remaining mypy error is in a file outside the scope of this PR (bash.py), so I’ve left that as-is.

tests/unit/test_exit_reason.py
Adityauyadav: added them, you can check.

openhands/core/schema/exit_reason.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
+
+
+class ExitReason(Enum):
+    INTENTIONAL = 'intentional'
+    INTERRUPTED = 'interrupted'
+    ERROR = 'error'
```
enyst, line 7: This makes sense, but sorry, do I see this right, nothing uses INTERRUPTED nor ERROR?

Adityauyadav: yes, at the moment, only intentional is being used. that was the approach i implemented first in order to keep it clean and extensible, but for ctrl + c, i ended up using the keyboardinterrupt. this structure makes it easier for future scenarios of tackling broader exceptions, so didn't just restricted to this #9394 scope, that's why didn't change it.
you want me to restructure the proposed solution?

enyst: Thank you, I see, I think we can go with it actually!

enyst:
Thank you!

PR #9436, opened 2025-06-28
fix(cli): fix terminal input lag on Windows by start&stopping pause task

llamantino:

Fixes terminal input lag/loss on Windows
Fixes CPR warnings after ctrl-c
Fixes exit through ctrl-c still leaving terminal in raw mode

---

Adds start&stop functions stop the agent pause listener
Stops the agent pause listener when the agent is not in running/paused state.
Adds a finally block to ensure raw mode is always exited.

---


li-boxuan:
Tested on MacOS and didn't see regression.

Issue #9457, opened 2025-06-30
Stop conversation


hieptl:
**What problem or use case are you trying to solve?**

Currently, there appears to be no way to stop a conversation — the only available action is to delete it.

**Describe the UX or technical implementation you have in mind**

To implement this feature, the following steps could be followed:

- Open the list of conversations.
- Click on the vertical dot icon on a conversation card.
- Currently, the only available option is `Delete`; we should add a new option called `Stop`.
- When the `Stop` option is selected, a confirmation dialog should appear.
- If the user clicks the `Confirm` button in the dialog, the conversation will be marked as stopped.
- The `conversation state indicator` should then update to reflect the `stopped` status.

**Additional context**
N/A

mamoodi:
There is a way to stop the conversation while you're inside the conversation. I see this is talking about from the conversation panel.
@rbren is this something desirable?

hieptl:
Hi @mamoodi , Could you let me know how to stop a conversation while we're inside a conversation? Could you send me the screenshot of that button/action? Thank you

mamoodi:
Sure! Here you go. It will show up as the agent is working. 

![Image](https://github.com/user-attachments/assets/f0d581d6-a268-4bb2-93e2-8e023b167890)

hieptl:
@mamoodi: Thank you for your reply! According to your screenshot, if we click the `stop` button, we will only stop the `agent`, not the current conversation.

We can refer to the video below for more information. 

https://github.com/user-attachments/assets/d25a320e-ad51-42ea-845c-1cb2a4e55087

According to the video above, after we click on the `stop` button, the agent status will be changed to `stopped`, but the conversation status is still `running`.

The purpose of this issue is to stop the conversation, not the agent. I think stopping the agent and stopping the conversation are two different features. Thank you

tofarr:
There is an api endpoint to stop conversations (Not just the agent):
https://app.all-hands.dev/docs#/default/stop_conversation_api_conversations__conversation_id__stop_post

It is simply not used in the UI yet.

hieptl:
Hi @tofarr, that's correct. And I integrated with that API in my [PR](https://github.com/All-Hands-AI/OpenHands/pull/9458). Thank you

PR #9458, opened 2025-06-30
feat(frontend): Stop conversation

hieptl:

Currently, there appears to be no way to stop a conversation — the only available action is to delete it.

---

To implement this feature, the following steps could be followed:

- Open the list of conversations.
- Click on the vertical dot icon on a conversation card.
- Currently, the only available option is `Delete`; we should add a new option called `Stop`.
- When the `Stop` option is selected, a confirmation dialog should appear.
- If the user clicks the `Confirm` button in the dialog, the conversation will be marked as stopped.
- The `conversation state indicator` should then update to reflect the `stopped` status.

---

#9457


frontend/src/components/features/conversation-panel/conversation-panel.tsx
```diff
@@ -54,6 +64,21 @@ export function ConversationPanel({ onClose }: ConversationPanelProps) {
...
+  const handleConfirmStop = () => {
+    if (selectedConversationId) {
+      stopConversation(
+        { conversationId: selectedConversationId },
+        {
+          onSuccess: () => {
+            if (selectedConversationId === currentConversationId) {
+              navigate("/");
+            }
+          },
```
amanape, line 76: I don't think it is necessary to navigate to the home screen after stopping a conversation. Would love to hear what you think

amanape (commented):
LGTM, I just have one small comment

frontend/src/components/features/conversation-panel/conversation-panel.tsx
hieptl: Hi @amanape,
Thank you for your feedback.

We’ve identified an issue that occurs when the user is **not redirected to the Home screen** after stopping the current conversation. Here are the steps to reproduce the issue:

1. On the **Home** page, click **"Launch from Scratch"** to start a new conversation and wait for the agent to be ready.  
2. On the **Conversation** page, open the list of conversations by clicking the conversations icon on the sidebar, and stop the current conversation.  
3. While still on the current **Conversation** page, try sending a message — the system will not function properly in this case.

You can refer to the attached video (`issue.mov`) for more details.

https://github.com/user-attachments/assets/f6a02ef7-892b-4860-a2e6-99e78d5ca588

To address this, we suggest **redirecting users to the Home screen** after stopping a conversation. If they wish to restart it, they can simply select the conversation again — the system will then automatically reconnect to the agent.

PR #9480, opened 2025-07-01
refactor(frontend): Show branch name and git provider on the conversation cards

### LINKED ISSUE #9479: Show branch name and git provider on the conversation cards

When working with multiple conversations for the same repository but on different branches, it becomes difficult to distinguish which conversation corresponds to which branch. This makes navigating and managing conversations less efficient.

You can refer to the image below for more context:

Image

To improve clarity, I suggest displaying the **branch name** alongside the **Git provider** in the conversation list. This will help users quickly identify and differentiate conversations — especially in cases where multiple repositories share the same name across different providers (e.g., GitHub, GitLab).

**Additional context**

N/A

hieptl:

When working with multiple conversations for the same repository but on different branches, it becomes difficult to distinguish which conversation corresponds to which branch. This makes navigating and managing conversations less efficient.

You can refer to the image below for more context:

Image

---

To improve clarity, I suggest displaying the **branch name** alongside the **Git provider** in the conversation list. This will help users quickly identify and differentiate conversations — especially in cases where multiple repositories share the same name across different providers (e.g., GitHub, GitLab).

---

Resolves #9479


frontend/src/api/open-hands.types.ts
```diff
@@ -70,6 +70,12 @@ export interface AuthenticateResponse {
   error?: string;
 }
 
+export interface SelectedRepository {
+  selected_repository: string | null;
```
amanape, line 75:
```suggestion
  selected_repository: string;
```

I don't think we want to expect `null` for this property here. If we do, we may want to change the interface name to something other than `SelectedRepository` since `selected_repository` might not actually exist

frontend/src/api/open-hands.types.ts
```diff
@@ -70,6 +70,12 @@ export interface AuthenticateResponse {
...
+  selected_branch: string | null;
+  git_provider: string | null;
```
amanape: We have a `Provider` type that includes our supported Git providers: https://github.com/All-Hands-AI/OpenHands/blob/c2acf4e07e0be5f0b88411ee929c371c7cc7b654/frontend/src/types/settings.ts#L7

hieptl: Hi @amanape, Thank you for your suggestion! I've updated the code accordingly.

frontend/src/api/open-hands.types.ts
hieptl: Hi @amanape ,

On the `Home` page, when the `Launch from Scratch` button is clicked, a new conversation is initiated without selecting a repository. As a result, the `selected_repository` field will be `null`.

The `Conversation Card` component is used in both scenarios — with or without a selected repository — so the type of this field can be either `string` or `null`.

Based on this, I’ve adopted the second approach and renamed the interface from `SelectedRepository` to `RepositorySelection` to better reflect its usage.  

Thank you for your helpful suggestion!

hieptl:
Hi @amanape,

Thanks for your suggestions! I’ll keep track of the Slack thread, and once the discussion is concluded, I’ll proceed with updating the code.

hieptl:
Hi @amanape, I cannot access the Slack thread above. Could you invite me to the Slack workspace? My email is `hieptl.developer@gmail.com`. Thank you!

amanape:
Done

Issue #9259, opened 2025-06-20
[Bug]: This request would exceed the rate limit for your organization (...) of 20,000 input tokens per minute


knail1:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

I keep getting this error in the console. this can happen in a brand new session with a simpple command..

17:11:38 - openhands:ERROR: retry_mixin.py:75 - litellm.RateLimitError: AnthropicException - {"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed the rate limit for your organization (0dccdde6-f924-4f1b-bb3b-92ef846e0a5b) of 20,000 input tokens per minute. For details, refer to: https://docs.anthropic.com/en/api/rate-limits. You can see the response headers for current usage. Please reduce the prompt length or the maximum tokens requested, or try again later. You may also contact sales at https://www.anthropic.com/contact-sales to discuss your options for a rate limit increase."}}. Attempt #3 | You can customize retry values in the configuration.

sometimes the "agent is rate limited" and just paused.  and some times open hands keeps on retrying and then continues on with the action.

1. I think this may be a bug
2. if this isn't a bug, I don't know if there is a way in config to keep within a token limit?

### OpenHands Installation

Docker command in README

### Operating System

None

knail1:
here's some more logs from the console. as you can see, the error is thrown and then it sometimes continues on its own after.

and other times it stops in the chat saying agent is rate limited, but thats not true, since, within the minute, i issue another command and it starts work through it fine 

litellm.exceptions.RateLimitError: litellm.RateLimitError: AnthropicException - {"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed the rate limit for your organization (0dccdde6-f924-4f1b-bb3b-92ef846e0a5b) of 20,000 input tokens per minute. For details, refer to: https://docs.anthropic.com/en/api/rate-limits. You can see the response headers for current usage. Please reduce the prompt length or the maximum tokens requested, or try again later. You may also contact sales at https://www.anthropic.com/contact-sales to discuss your options for a rate limit increase."}}

17:22:26 - openhands:INFO: agent_controller.py:591 - [Agent Controller d810f80c9e804740b7f191c525c12bc6] Setting agent(CodeActAgent) state from AgentState.RUNNING to AgentState.RATE_LIMITED
17:22:26 - OBSERVATION
[...]

Issue #8706, opened 2025-05-26
[Bug]: Resolver waits indefinitely after hitting retry limit


kotauchisunsun:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

I'm using the Resolver with GitHub Actions.
While OpenHands is running, the generative AI API sometimes hits a rate limit (e.g., 429), which triggers retries. However, when the retry limit is reached, the process doesn't terminate—instead, it continues waiting indefinitely. (This is especially noticeable when using Gemini's free tier.)

As a result, GitHub Actions keeps waiting until it's forcibly terminated due to a timeout.

Here’s an issue I created to test this behavior:
https://github.com/kotauchisunsun/openhands-debug-rate-limited/issues/1

This is the actual log of the run, where you can see GitHub Actions waiting for 6 hours:
https://github.com/kotauchisunsun/openhands-debug-rate-limited/actions/runs/15240756413/job/42860524606

The relevant code appears to be here:
https://github.com/All-Hands-AI/OpenHands/blob/0.39.2/openhands/controller/agent_controller.py#L293

### OpenHands Installation

GitHub resolver

### OpenHands Version

0.39.2

### Operating System

Linux

Issue #6890, opened 2025-02-22
[Bug]: Rate Limit Error


Marius-Juston:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Currently on a Tier 2 Antropic API tier, which has a 80,000 input token per minute limit limit. So after a while I get he RateLimitError. The problem is that after  3/4 times it gives you the rate limit error it gives another Python error that does not get exception handled which causes the system to constantly locked in a "Agent is Rate Limited" state, with the only solution is to restart the instance of OpenHands. 

So I think a couple things would be helpful.
1) The ability to set a custom rate limit on the UI side for APIs that are rate limited (with the ability to set a refresh time)
2) Add the ability to truncate the prompt input so that it helps against the rate limits and input token size
3) Fix the "Agent is Rate Limited" infinite state 

### OpenHands Installation

Docker command in README

### OpenHands Version

0.25

### Operating System

Linux

### Logs, Errors, Screenshots, and Additional Context

```
litellm.llms.anthropic.common_utils.AnthropicError: {"type":"error","error":{"type":"rate_limit_error","message":"This request would exceed your organization’s rate limit of 80,000 input tokens per minute. For details, refer to: https://docs.anthropic.com/en/api/rate-limits; see the response headers for current usage. Please reduce the prompt length or the maximum tokens requested, or try again later. You may also contact sales at https://www.anthropic.com/contact-sales to discuss your options for a rate limit increase."}}

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/openhands/controller/agent_controller.py", line 238, in _step_with_exception_handling
    await self._step()
  File "/app/openhands/controller/agent_controller.py", line 674, in _step
    action = self.agent.step(self.state)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/openhands/agenthub/codeact_agent/codeact_agent.py", line 130, in step
    response = self.llm.completion(**params)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/tenacity/__init__.py", line 336, in wrapped_f
    return copy(f, *args, **kw)
           ^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/tenacity/__init__.py", line 475, in __call__
    do = self.iter(retry_state=retry_state)
[...]

PR #9489, opened 2025-07-01
fix(llm_config): extend retry delays to respect rate limit windows

llamantino:

Fixes agent stopping because of per-minute rate limit errors by increasing retry delays

---

This PR changes the retry timing to reduce the chances of the agent stopping after hitting per-minute rate limits.

- Before: 4 attempts with wait times of 0, 5, 5, 8s (total 18s, not enough for the rate limit to reset)
- After: 5 attempts with wait times of 0, 8, 16, 32, 64  (total 120s)

The final wait exceeds 60s (=64s) in case of providers that count failed attempts.

Note: Total LLM requests should rise from 12 to 15, as LiteLLM seems to retry each OpenHands request up to 3 times.

Might fix or improve #6890, #8706 and #9259.

---


rbren (approved):
this seems reasonable to me. Sounds like the current config is hurting gemini so I say we ship it

openhands/core/config/llm_config.py
```diff
@@ -57,11 +57,11 @@ class LLMConfig(BaseModel):
...
     openrouter_site_url: str = Field(default='https://docs.all-hands.dev/')
     openrouter_app_name: str = Field(default='OpenHands')
-    # total wait time: 5 + 10 + 20 + 30 = 65 seconds
-    num_retries: int = Field(default=4)
-    retry_multiplier: float = Field(default=2)
-    retry_min_wait: int = Field(default=5)
-    retry_max_wait: int = Field(default=30)
+    # total wait time: 8 + 16 + 32 + 64 = 120 seconds
+    num_retries: int = Field(default=5)
+    retry_multiplier: float = Field(default=8)
```
enyst, line 62: I think this is a factor for the exponential, that is, retry_multiplier = 2 means each wait is 2x times as long as the previous one?

8 would be huge

llamantino: The formula is `multiplier * exp_base ^ (attempt_number - 1)`, clamped between min and max. exp_base defaults to 2. ([source](https://github.com/jd/tenacity/blob/eed7d785e667df145c0e3eeddff59af64e4e860d/tenacity/wait.py#L163))

The last wait (the one after the 4th attempt) should be 8 * (2 ^ (4 - 1)) = 64s. 

I tested it before and after, but extra testing shouldn't hurt.

(Also, the variable names aren't technically right: retries != attempts. if num_retries were 1, there would be no retry)

llamantino: Just to be sure, I tested it again, and here are the hits I'm seeing on my LLM proxy server. The number of connections and the gaps between them seem to match. I'll mark this as resolved.

```
08:09:23 429 rate limit
08:09:24 429 rate limit
08:09:25 429 rate limit

(8s gap)

08:09:33 429 rate limit
08:09:33 429 rate limit
08:09:34 429 rate limit

(16s gap)

08:09:50 429 rate limit
08:09:50 429 rate limit
08:09:51 429 rate limit

(32s gap)

08:10:23 429 rate limit
08:10:24 429 rate limit
08:10:25 429 rate limit

(64s gap)

08:11:29 429 rate limit
08:11:29 429 rate limit
08:11:30 429 rate limit
```

Abubakar-01:
**Main Branch:**
There were 3 retry attempts observed with wait durations of approximately 7s, 6s, and 10s, totaling around 23 seconds. The retry pattern appeared short and does not exceed the typical rate limit recovery window.

**PR Branch:**
There were 4 retry attempts with backoff intervals of roughly 8s, 18s, 33s, and 67s, summing to about 126 seconds. The pattern followed an exponential backoff strategy and allows for longer recovery time between attempts.

**Summary:**
The main branch uses a shorter retry cycle with minimal delays. The PR branch introduces a more extended backoff sequence, allowing retries over a longer duration and accommodating higher wait times for recovery.

enyst:
Thank you! This is just right then, let’s get it in.

Issue #9518, opened 2025-07-03
[CLI] can not use search api key,beacause of mcp config empty


vaxilicaihouxian:
```
def get_mcp_config(
        self, extra_stdio_servers: list[MCPStdioServerConfig] | None = None
    ) -> MCPConfig:
        # TODO: Load MCP config from a local file
        return MCPConfig()
```
so the mcp tools are empty and then *search_api_key* not work in cli runtime.

ryanhoangt:
Hi @vaxilicaihouxian, MCP server is currently not supported in CLI mode yet, I'm looking into it in #9519.

vaxilicaihouxian:
if I write search_api_key in the config.toml,it will show me a error like 'str not impletement get_scecret_value method'.

PR #9519, opened 2025-07-03
Add MCP support for CLI

ryanhoangt:

Add MCP support for CLI.

---

This PR adds support for using MCP server with CLI:
- Load MCP server config via `~/.openhands/config.toml`
- Use stdio transport directly in CLI runtime instead of via FastMCP proxy 
- Add 2 commands: `/mcp` to list all MCP servers added by users and `/mcp-errors` to display MCP-related errors on CLI start-up.

Screenshot 2025-07-03 at 16 21 58

Screenshot 2025-07-04 at 21 37 38

---

Fix https://github.com/All-Hands-AI/OpenHands/issues/9518


jpelletier1:
Looking forward to this feature!

docs/usage/how-to/cli-mode.mdx
```diff
@@ -134,6 +134,10 @@ project details and structure. Use this when onboarding the agent to a new codeb
 You can pause the agent while it is running by pressing `Ctrl-P`. To continue the conversation after pausing, simply
 type `/resume` at the prompt.
 
+#### Using MCP servers
+
+To configure Model Context Protocol (MCP) servers, you can follow the documentation on [MCP servers](../mcp) and add them under the `[mcp]` section in the global `~/.openhands/config.toml` file.
```
xingyaoww: Can you include an example config file that includes both STDIO, SSE, and SHTTP example?

openhands/cli/main.py
```diff
@@ -294,7 +294,13 @@ def on_event(event: Event) -> None:
     if not skip_banner:
         display_banner(session_id=sid)
 
-    welcome_message = 'What do you want to build?'  # from the application
+    welcome_message = ''
+
+    # Display number of MCP servers configured
+    if agent.config.enable_mcp:
+        welcome_message += f'Using {len(runtime.config.mcp.stdio_servers)} stdio MCP servers, {len(runtime.config.mcp.sse_servers)} SSE MCP servers and {len(runtime.config.mcp.shttp_servers)} SHTTP MCP servers.\n\n'
```
xingyaoww: Can we add a line that points to our doc site: to setup this, "visit XXX to configure".

Ideally, we should also add MCP to the CLI settings too, to guide the user to set it up and WE persist the settings in config.toml

openhands/mcp/client.py
```diff
@@ -97,6 +105,11 @@ async def connect_http(
             logger.error(f'Error connecting to {server_url}: {e}')
             raise
 
+    async def connect_stdio(self, server: MCPStdioServerConfig, timeout: float = 30.0):
+        transport = StdioTransport(command=server.command, args=server.args or [])
+        self.client = Client(transport, timeout=timeout)
+        await self._initialize_and_list_tools()
```
xingyaoww: Can we add tests that make sure MCP error messages are printed to the terminal?

openhands/runtime/impl/cli/cli_runtime.py
```diff
@@ -689,8 +689,73 @@ def edit(self, action: FileEditAction) -> Observation:
         )
 
     async def call_tool_mcp(self, action: MCPAction) -> Observation:
-        """Not implemented for CLI runtime."""
-        return ErrorObservation('MCP functionality is not implemented in CLIRuntime')
+        """Execute an MCP tool action in CLI runtime.
```
xingyaoww, line 692: We should add end to end tests for MCP + CLI runtime test

xingyaoww (commented):
Great to see this coming together

openhands/cli/main.py
ryanhoangt: Yeah maybe we can find ways to allow users set MCP config directly in CLI in another PR, for now I add an extra `/mcp` command that points to our doc if MCP is not set in the config file, and list all the servers otherwise:
Screenshot 2025-07-04 at 13 08 48
Screenshot 2025-07-04 at 12 58 21

vaxilicaihouxian:
nice try.

openhands/cli/main.py
xingyaoww: I think we *should* implement this in this PR if it is not too tricky? Eg I'd imagine this should be something OpenHands can help with?

ryanhoangt: IMO it's not very trivial. A naive way is to ask users to paste the whole config at once, which is not optimal. A better way is to do something similar to [Claude Code](https://docs.anthropic.com/en/docs/claude-code/mcp#configure-mcp-servers), i.e. `openhands mcp [cmd]`, with `cmd` being `add`, `remove`, `list`, and potentially some validation there. Wdyt? 🤔

ryanhoangt: The way of manually modifying the `config.toml` file isn't too inconvenient for now I think, Gemini CLI also does it that way: https://github.com/google-gemini/gemini-cli/blob/main/docs/cli/configuration.md#available-settings-in-settingsjson

docs/usage/how-to/cli-mode.mdx
```diff
@@ -122,7 +122,7 @@ follow the prompts:
 - **Basic settings**: Choose a model/provider and enter your API key.
 - **Advanced settings**: Set custom endpoints, enable or disable confirmation mode, and configure memory condensation.
 
-Settings can also be managed via the `config.toml` file.
+Settings can also be managed via the `config.toml` file in the current directory or `~/.openhands/config.toml`.
```
mamoodi, line 166: There's been a lot of confusion recently on the path of the config.toml that OH will use for configuration.
From what I've been told is that OH will still use the `<current-dir>/config.toml` for configurations. 
Is this PR changing the place where the config.toml is being picked up for CLI as well?

openhands/cli/main.py
xingyaoww: Yeah I actually like the Claude code one, user can basically "add MCP", and CLI will ask for the type of server (stdio, sse, shttp). And then we ask for each argument (user can hit enter for empty argument).

I bet if we write a good instruction, OpenHands should be able to help with this

ryanhoangt:
@xingyaoww here's the UX for users to configure MCP servers, do you have any feedback?

https://github.com/user-attachments/assets/28726384-4fcc-4477-b67d-287f79e75343

On startup/reload if there're issues users can use command `/mcp-errors` to view them (see the image in the PR description above).

docs/usage/how-to/cli-mode.mdx
ryanhoangt: > From what I've been told is that OH will still use the <current-dir>/config.toml for configurations.

Yeah I think this is the current behavior, and we can consider this is the project-level config.

> Is this PR changing the place where the config.toml is being picked up for CLI as well

Yep, we need a way to persist MCP settings globally, and this PR will change the behavior of the CLI to check and use the project-level `config.toml` file; If it doesn't exist we'll use the global `config.toml`, which always exists. Previously it only uses the project-level config in current working directory.

xingyaoww:
one thing: can we put /mcp-error under /mcp, and then add anoter option: "View MCP Loading Errors"?

nit: we should add a space right after the prompt

image

but overall i think this looks great to me!

KennyDizi:
Looking forward to this PR.

docs/usage/how-to/cli-mode.mdx
mamoodi: Thanks! Makes sense.

openhands/cli/commands.py
```diff
@@ -38,6 +43,57 @@
...
+            'Please restart OpenHands manually for changes to take effect.'
+        )
+
+
+async def prompt_for_restart(config: OpenHandsConfig) -> bool:
+    """Prompt user if they want to restart the CLI and return their choice."""
+    from prompt_toolkit import HTML
+    from prompt_toolkit.patch_stdout import patch_stdout
+
+    from openhands.cli.tui import create_prompt_session
```
xingyaoww: Can we put these imports on the top of the file?

openhands/cli/commands.py
```diff
@@ -327,3 +385,365 @@ def check_folder_security_agreement(config: OpenHandsConfig, current_dir: str) -
...
+    config_data = load_config_file()
+    if 'mcp' not in config_data:
+        config_data['mcp'] = {}
+    if 'stdio_servers' not in config_data['mcp']:
+        config_data['mcp']['stdio_servers'] = []
+
+    config_data['mcp']['stdio_servers'].append(server_config)
+    save_config_file(config_data)
+
+    print_formatted_text(f'✓ Stdio MCP server added: {name}')
```
xingyaoww: We should also tell the user that this is being saved to ~/.openhands/config.toml?

openhands/cli/commands.py
```diff
@@ -327,3 +385,365 @@ def check_folder_security_agreement(config: OpenHandsConfig, current_dir: str) -
...
+    config_data = load_config_file()
+    if 'mcp' not in config_data:
+        config_data['mcp'] = {}
+    if 'shttp_servers' not in config_data['mcp']:
+        config_data['mcp']['shttp_servers'] = []
+
+    config_data['mcp']['shttp_servers'].append(server_config)
+    save_config_file(config_data)
+
+    print_formatted_text(f'✓ SHTTP MCP server added: {url}')
```
xingyaoww: ditto

openhands/cli/commands.py
```diff
@@ -327,3 +385,365 @@ def check_folder_security_agreement(config: OpenHandsConfig, current_dir: str) -
...
+    config_data = load_config_file()
+    if 'mcp' not in config_data:
+        config_data['mcp'] = {}
+    if 'sse_servers' not in config_data['mcp']:
+        config_data['mcp']['sse_servers'] = []
+
+    config_data['mcp']['sse_servers'].append(server_config)
+    save_config_file(config_data)
+
+    print_formatted_text(f'✓ SSE MCP server added: {url}')
```
xingyaoww: ditto

openhands/runtime/impl/cli/cli_runtime.py
```diff
@@ -689,8 +689,73 @@ def edit(self, action: FileEditAction) -> Observation:
...
+
+        Args:
+            action: The MCP action to execute
+
+        Returns:
+            Observation: The result of the MCP tool execution
+        """
+        import sys
+
+        from openhands.events.observation import ErrorObservation
```
xingyaoww: Let's move these imports to the top

tests/unit/test_cli_tui_mcp.py
xingyaoww, line 1: Can we combine these test with existing test_cli_tui?

ryanhoangt:
@xingyaoww I added input validations before saving to the config file, the validation error looks like below, could you help take another look?

Screenshot 2025-07-17 at 19 49 34

openhands/mcp/validator.py
xingyaoww, line 1: I wonder if we can simplify the validation by simply using our MCPConfig pydantic class? that should drastically reduce the amount of code here and keep all validation logic in one place

ryanhoangt: Do you mean using Pydantic support to implement these additional validations? b.c. currently the `MCPConfig` class doesn't have validators for erroneous input like the `command` or `url` above

xingyaoww: Yep! so we can easily re-use this in a lot of different places

ryanhoangt: I moved them to Pydantic field validators in the latest commit, and also changed the validation to be after the user entering all input values to make code cleaner, LMK if you have other concerns:

Screenshot 2025-07-18 at 21 15 42

openhands/cli/commands.py
```diff
@@ -327,3 +409,432 @@ def check_folder_security_agreement(config: OpenHandsConfig, current_dir: str) -
...
+        config_data['mcp'] = {}
+    if 'shttp_servers' not in config_data['mcp']:
+        config_data['mcp']['shttp_servers'] = []
+
+    server_config = {'url': server.url}
+    if server.api_key:
+        server_config['api_key'] = server.api_key
+
+    config_data['mcp']['shttp_servers'].append(server_config)
+    save_config_file(config_data)
```
xingyaoww: nit: i saw this logic repeating multiple times, maybe we can pack them into a helper function

openhands/cli/main.py
```diff
@@ -492,6 +520,15 @@ async def main_with_loop(loop: asyncio.AbstractEventLoop) -> None:
         if not env_log_level:
             logger.setLevel(logging.WARNING)
 
+    # If `config.toml` does not exist in current directory, use the file under home directory
+    if not os.path.exists(args.config_file):
+        logger.info(
+            f'Config file {args.config_file} does not exist, using default config file in home directory.'
```
xingyaoww:
```suggestion
            f'Config file {args.config_file} does not exist, using default config file in home directory ~/.openhands/config.toml'
```

openhands/core/config/utils.py
```diff
@@ -154,8 +154,22 @@ def load_from_toml(cfg: OpenHandsConfig, toml_file: str = 'config.toml') -> None
...
+            # Check if expected_type is a Union that includes SecretStr and value is str, e.g. search_api_key
+            if expected_type:
+                origin = get_origin(expected_type)
+                args = get_args(expected_type)
+
+                if origin is UnionType and SecretStr in args and isinstance(value, str):
+                    value = SecretStr(value)
+                elif expected_type is SecretStr and isinstance(value, str):
+                    value = SecretStr(value)
+
```
xingyaoww, line 172: Is this change intentional/needed in MCP?

openhands/cli/commands.py
```diff
@@ -327,3 +409,432 @@ def check_folder_security_agreement(config: OpenHandsConfig, current_dir: str) -
...
+            print_formatted_text('Operation cancelled.')
+            return
+
+        command = await collect_input(config, "\nEnter command (e.g., 'uvx', 'npx'):")
+        if command is None:
+            print_formatted_text('Operation cancelled.')
+            return
+
+        args_input = await collect_input(
+            config, '\nEnter arguments (comma-separated, optional):'
```
xingyaoww: Comma-separated is pretty not convenient, can we accept `-y mcp-remote https://mcp.tavily.com/mcp/?tavilyApiKey=tvly-dev-XXX` and we do the split on our end?

After putting these in (no comma separated), i got these error:

Image

openhands/mcp/utils.py
```diff
@@ -52,6 +62,7 @@ async def create_mcp_clients(
     sse_servers: list[MCPSSEServerConfig],
     shttp_servers: list[MCPSHTTPServerConfig],
     conversation_id: str | None = None,
+    stdio_servers: list[MCPStdioServerConfig] | None = None,
```
xingyaoww, line 67: Can we move this to the third argument? Putting this after `conversation_id` feels a bit werid

openhands/core/config/utils.py
ryanhoangt: It's not directly related to MCP, but intentional and required for the `search_api_key` to be loaded into OpenHands from `config.toml`. Its type in the config is `SecretStr` so we need to do some conversion from `str` to `SecretStr`. I implemented it this way to support other config values that potentially have this kind of mismatch in the future:

https://github.com/All-Hands-AI/OpenHands/blob/7a168b9b5f33e63746456779032c0caa623e23f4/openhands/core/config/openhands_config.py#L77-L80

Without it we'll get this error: https://github.com/All-Hands-AI/OpenHands/issues/9518#issuecomment-3051739251

openhands/cli/commands.py
ryanhoangt: Yeah good idea, I replaced the comma-separated format with this whitespace-based one:

Screenshot 2025-07-23 at 17 12 09

xingyaoww (approved):
Awesome it finally works for me!

Image


nit: can you clean up the tests? It seems we modified/added maybe too much tests in this PR, can we try to remove some unnecessary ones?

ryanhoangt:
Thanks, I've asked OpenHands to help clean them up 😅 From your image I just noticed the `fetch` MCP server isn't loaded by default yet, I'll fix it in another PR (issue #9875).

openhands/cli/main.py
```diff
@@ -298,6 +299,10 @@ def on_event(event: Event) -> None:
 
     # Add MCP tools to the agent
     if agent.config.enable_mcp:
+        # Clear any previous errors and enable collection
+        mcp_error_collector.clear_errors()
+        mcp_error_collector.enable_collection()
```
enyst, line 304: Are the clear/enable/disable methods necessary? It seems that this is a global var which will be newly initialized at this point in the code. Are we saving the errors from a run to another?

PR #9532, opened 2025-07-03
Fix for issue where wrong method was called

tofarr:

The wrong method was being called in the rest API, resulting in an error. The correct method expects the conversation_id as a parameter - the previously invoked method expects a `connection_id` from SocketIO

---

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:4e48cd6-nikolaik   --name openhands-app-4e48cd6   docker.all-hands.dev/all-hands-ai/openhands:4e48cd6
```

### DIFF
--- a/openhands/server/routes/conversation.py
+++ b/openhands/server/routes/conversation.py
@@ -158,7 +158,7 @@ async def add_event(
     request: Request, conversation: ServerConversation = Depends(get_conversation)
 ):
     data = request.json()
-    await conversation_manager.send_to_event_stream(conversation.sid, data)
+    await conversation_manager.send_event_to_conversation(conversation.sid, data)
     return JSONResponse({'success': True})
 
 


openhands/server/routes/conversation.py
```diff
@@ -158,7 +158,7 @@ async def add_event(
     request: Request, conversation: ServerConversation = Depends(get_conversation)
 ):
     data = request.json()
-    await conversation_manager.send_to_event_stream(conversation.sid, data)
+    await conversation_manager.send_event_to_conversation(conversation.sid, data)
```
raymyers, line 161: It would admittedly seem hard for anyone to know whether they want to send data to "the event stream" or to "the conversation"

enyst: Agree. Do we document the difference somewhere?

PR #9535, opened 2025-07-03
chore(ui): npm package config

amanape:

Converts the `openhands-ui` directory into a proper npm package ready for publishing.

https://www.npmjs.com/package/@openhands/ui

---
This project uses [Bun](https://bun.sh/) for its speed and efficiency improvements over npm/yarn. However, developers are free to use the package manager of their choice—Bun, npm, or yarn all work with this setup. The relevant scripts and lockfiles for Bun are included, but not strictly required.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:8e31a1a-nikolaik   --name openhands-app-8e31a1a   docker.all-hands.dev/all-hands-ai/openhands:8e31a1a
```


raymyers:
Using Bun rather than NPM here seems like it's diverging from the convention set by frontend. Maybe there are good reasons to do that, but it seems like the PR should explain them, yeah?

openhands-ui/index.ts
```diff
@@ -1 +1,18 @@
...
+export { Icon } from "./components/icon/Icon";
+export { Input } from "./components/input/Input";
+export { InteractiveChip } from "./components/interactive-chip/InteractiveChip";
+export { RadioGroup } from "./components/radio-group/RadioGroup";
+export { RadioOption } from "./components/radio-group/RadioOption";
+export { Scrollable } from "./components/scrollable/Scrollable";
+export { Toggle } from "./components/toggle/Toggle";
+export { Tooltip } from "./components/tooltip/Tooltip";
+export { Typography } from "./components/typography/Typography";
+export { BaseTypography } from "./components/typography/BaseTypography";
```
mislavlukach: BaseTypography shouldn't be exported

openhands-ui/index.ts
```diff
@@ -1 +1,18 @@
-console.log("Hello via Bun!");
\ No newline at end of file
+// Components
+export { Button } from "./components/button/Button";
```
mislavlukach, line 2: I would also export main prop type
For example: ButtonProps.

This way when we end up creating a thin wrapper around these components in main frontend folder, we can use these types.

raymyers:
I've made a PR to this PR branch adding a CI job. It runs the build, it would run linter and tests but there aren't any (yet?), but at least it will be easy to add that. We can also figure out how publishing the package from an action will go

https://github.com/All-Hands-AI/OpenHands/pull/9539

openhands-ui/README.md
```diff
@@ -1,15 +1,60 @@
...
 ```bash
-bun run sb
+# Install dependencies
+bun install
+
+# Start Storybook
+bun run dev
+
+# Build package
+bun run build
```
raymyers, line 71: Is there a step for including the updated copy in frontend or is that not ready yet?

raymyers: Also it sounds like you've already NPM published this, so does that mean there is already a manual publish process? Until it's automated, we should have an explanation of that process, under a README header or a RELEASING.md or something.

amanape:
> Using Bun rather than NPM here seems like it's diverging from the convention set by frontend. Maybe there are good reasons to do that, but it seems like the PR should explain them, yeah?

From experience Bun improves the developer experience and I prefer it mainly because it's significantly faster and more efficient than npm. I wish to do the same on the frontend, but this is pending community and team discussion.

I've updated the README to clarify that developers are free to use the package manager they prefer. Personally, I'm leaning toward Bun, so I’ve kept its references in the README, package.json scripts, and bun.lock.

I also noticed the CI is currently set up with Bun actions—if we decide to standardize on npm for now, I’m happy to adjust everything. Let me know what you think.

openhands-ui/README.md
amanape: > Is there a step for including the updated copy in frontend or is that not ready yet?

I'm not sure what you're referring to here, could you clarify?

> Also it sounds like you've already NPM published this, so does that mean there is already a manual publish process? Until it's automated, we should have an explanation of that process, under a README header or a RELEASING.md or something.

Running a package manager's publish command (`bun publish` for bun, `npm publish` for npm). It requires authorized users so we can create a job to do this automatically via a npm key

raymyers: > I'm not sure what you're referring to here, could you clarify?

Under "Installation", it says to install using `npm install @openhands/ui`, which installs the version in NPM, right? During development how to I include the locally modified version in another project?

openhands-ui/README.md
```diff
@@ -1,15 +1,74 @@
...
+| `RadioGroup`      | Radio button group                        |
+| `RadioOption`     | Individual radio option                   |
+| `Scrollable`      | Scrollable container                      |
+| `Toggle`          | Toggle switch                             |
+| `Tooltip`         | Tooltip overlay                           |
+| `Typography`      | Text components (H1-H6, Text, Code)       |
+
+## Development
+
+Use your preferred package manager to install dependencies and run the development server. We recommend using [Bun](https://bun.sh) for a fast development experience.
```
raymyers, line 59: If they use a different package manager it will not use the bun.lock so I think it's at least required to use it if they are going to make dependency changes and submit a PR.

amanape: Updated to include note for potential contributors to use Bun

openhands-ui/README.md
amanape: Added subsection `Testing Locally Without Publishing` with instructions on how to use the library without installing it from the npm registry. LMK if you meant something else

openhands-ui/index.ts
amanape: I’ve considered it but I don’t know how useful that can be since our prop type extends the HTML element props, which might make it annoying.

Maybe we can punt on this until we come across reason to include them? In any way, the user can define a custom type to obtain prop types:

```ts
type GetComponentPropTypes<T> =
  T extends React.ComponentType<infer P> ? P : never;
```

LMK your opinion

.github/workflows/npm-publish-ui.yml
```diff
@@ -0,0 +1,104 @@
...
+          fi
+
+      - name: Setup npm authentication
+        if: steps.npm-check.outputs.already-exists == 'false'
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
+
+      - name: Publish to npm
+        if: steps.npm-check.outputs.already-exists == 'false'
+        working-directory: ./openhands-ui
```
raymyers, line 104: This as multiple lines so I think it also needs `shell: bash` so it won't have publish fail and report success because echo worked. shell can also be marked as default on job or workflow level if preferred.

raymyers (approved):

PR #9544, opened 2025-07-03
perf: make EventStore cur_id a lazy calculated property

tofarr:

Improves application performance by making EventStore object creation faster when the current event ID is not immediately needed. This optimization reduces file system scanning overhead during object initialization.

---

This PR converts the `cur_id` attribute in the `EventStore` class from being calculated in `__post_init__` to a lazy calculated property. 

**Key changes:**
- Removed the `__post_init__` method that performed expensive file system scanning on every EventStore creation
- Added a `cur_id` property that calculates the value only when first accessed
- Added a private `_cur_id` field to cache the calculated value
- Maintained backward compatibility with both getter and setter interfaces
- Extracted the calculation logic into a separate `_calculate_cur_id()` method

**Design decisions:**
- Used a property with getter/setter to maintain the existing API contract
- Cached the calculated value to avoid repeated file system operations
- Made the calculation method private since it's an implementation detail
- Preserved all existing functionality while improving performance

**Performance impact:**
- EventStore object creation is significantly faster when `cur_id` is not needed
- File system scanning is deferred until the property is actually accessed
- Multiple accesses to `cur_id` use the cached value for optimal performance
- No performance regression when `cur_id` is used (same total time as before)

---

This addresses the performance concern mentioned in the issue where "A lot of EventStore objects are created in the application lifecycle" and the `__post_init__` method was performing expensive file system operations even when the `cur_id` property wasn't being used by constructors.

@tofarr can click here to [continue refining the PR](https://app.all-hands.dev/conversations/30b6b7787aa04c6fa05f6cd6dba82f45)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:d8783fe-nikolaik   --name openhands-app-d8783fe   docker.all-hands.dev/all-hands-ai/openhands:d8783fe
```


raymyers:
I think this may be susceptible to a race condition, but stepping back I'm confused why determining ID on EventStore creation is a bottleneck. Why are we creating so many EventStores and if we do, won't they all have to determine an id shortly thereafter? Is this about empty ones that are never used?

tofarr:
> Is this about empty ones that are never used?

That's the issue exactly! `AgentLoopInfo` includes an EventStore instance. Some implementations pull this from the Session, while others simply create a new object. In most cases, this is never used, as the AgentLoopInfo is only used to get the status of a running event loop.

tofarr:
> I think this may be susceptible to a race condition

The EventStore class is read only - EventStream (A subclass) includes the ability to add items. So this class represents a snapshot of the event stream at the point in time when the object is created. (That is why #9545 is included). If we don't create duplicates of the event stream, we can create as many EventStores as required - and this update makes that much cheaper.

raymyers:
> In most cases, this is never used, as the AgentLoopInfo is only used to get the status of a running event loop.

Possibly then it's the EventStore that should be an Optional and not its id, but sounds like there some bigger taking stock of objects to do and this step sounds ok for now.

raymyers (approved):

PR #9545, opened 2025-07-03
Use the same event stream instance for conversations as sessions

tofarr:

Fixed an issue where event tracking could become inconsistent between conversations and sessions, improving the reliability of conversation state management and event search functionality. This ensures that conversation history and event data remain properly synchronized throughout the session lifecycle.

---

This PR fixes a critical issue where conversations and sessions were creating separate EventStream instances, causing the `cur_id` to become stale and event search methods to malfunction.

**Problem:**
Previously, when a new conversation was created, it would instantiate its own EventStream, separate from any existing session's EventStream. This led to:
- Stale `cur_id` values that didn't reflect the actual current state
- Event search methods returning incorrect or incomplete results
- Inconsistent state management between conversations and sessions

**Solution:**
The fix ensures that conversations and sessions share the same EventStream instance by:

1. **Modified StandaloneConversationManager**: When creating a new conversation, the manager now checks if there's an existing session with an event stream (`self._local_agent_loops_by_sid.get(sid)`) and passes it to the conversation constructor

2. **Updated ServerConversation**: The constructor now accepts an optional `event_stream` parameter, using the provided instance if available or creating a new one if not

**Design Decision:**
The approach maintains backward compatibility by making the `event_stream` parameter optional. If no existing event stream is provided, the conversation creates a new one as before. This ensures the change is non-breaking while fixing the synchronization issue.

**Benefits:**
- **Consistent State**: The `cur_id` stays synchronized between conversations and sessions
- **Reliable Event Search**: Event search methods now work correctly with up-to-date state
- **Better Resource Management**: Avoids creating duplicate event streams for the same session
- **Improved Debugging**: Event tracking is now consistent across the entire session lifecycle

---

This addresses event stream synchronization issues that could cause problems with conversation state tracking and event search methods. While no specific GitHub issue was linked, this fix resolves a fundamental architectural problem that could lead to data inconsistencies and unreliable event handling.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:bcdd1cb-nikolaik   --name openhands-app-bcdd1cb   docker.all-hands.dev/all-hands-ai/openhands:bcdd1cb
```


raymyers (approved):

raymyers:
I'm still not clear why Session and Conversation are different objects when we keep saying they are the same domain concept, but this seems like a positive change towards them being more consistent.

tofarr:
> I'm still not clear why Session and Conversation are different objects when we keep saying they are the same domain concept, but this seems like a positive change towards them being more consistent.

Long term, I think we need to review these data models and the relations between them (At the very least we should document specifically what the purpose of each is):

* ConversationManager : Create / Get / List / Delete conversations.
* Session : A Conversation at it's lowest level - independent of server
* ServerConversation : Used mostly to give access to the Session within the context of the REST API.
* AgentLoopInfo : Gives information on running conversations.

openhands/server/session/conversation.py
```diff
@@ -15,37 +15,51 @@ class ServerConversation:
...
+        self.runtime = runtime
 
     async def connect(self) -> None:
-        await self.runtime.connect()
+        if not self._attach_to_existing:
+            await self.runtime.connect()
 
     async def disconnect(self) -> None:
+        if self._attach_to_existing:
+            return
```
enyst, line 62: Why don’t we disconnect these? Seems like, sooner or later, no matter how they were initialized, we’d need to disconnect?

tofarr: TBH I think this needs a refactor - when the _attach_to_existing flag is set, the conversation is being attached to an existing session, so closing it would close that session.

PR #6902, opened 2025-02-23
Fix mypy errors in runtime/utils directory

neubig:
This PR fixes all mypy errors in the `openhands/runtime/utils` directory. Changes include:

- Added missing return type annotations to functions
- Added missing type annotations to function parameters
- Fixed Any return type issues
- Added missing imports
- Fixed type compatibility issues with libtmux
- Added proper null checks for optional attributes

All changes are confined to the specified directory.

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:354f030-nikolaik   --name openhands-app-354f030   docker.all-hands.dev/all-hands-ai/openhands:354f030
```

PR #9549, opened 2025-07-04
Optimize dockerfile by consolidating and reordering steps

raymyers:

Reduce the size of runtime docker images

---

This should significantly improve image build times and sizes for our runtime docker images.

* Optimize docker layers by consolidating steps (see below)
* Reorder OpenVSCode-Server install to be before app code to try and hit the cache better
* Refactor conditionals of the apt installs to be more readable
* Increase the CPU on the runtime image builder
* Make multi-line scripts explicitly use "bash" shell, which tells GitHub actions to catch failures ([docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrunshell))
* Use blacksmith built-in poetry caching
* Remove the hash checking process, which was broken

## Context on docker layers
Some context on the first: PR 6902 broke up the miniconda build steps in commit `fix: Fix environment variable setting in Dockerfile template`.

https://github.com/All-Hands-AI/OpenHands/pull/6902/files#diff-3e1bf8a8a14551c97005b1f86178f3480843488f3654f00ff1cb74b397ca0465

This was logical from the point of view of troubleshooting the env vars, but it winds up making the images bigger.

---

ALL-2675
---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:4e88055-nikolaik   --name openhands-app-4e88055   docker.all-hands.dev/all-hands-ai/openhands:4e88055
```


wolfspyre:
I’d encourage looking at the size of the final container 

I spent a fair bit of time myself with dive rearranging stuff… 

the runtime image being as big as it is is kinda problematic …. mine is almost 10G

openhands/runtime/utils/runtime_templates/Dockerfile.j2
```diff
@@ -105,28 +115,22 @@ RUN rm -rf ${OPENVSCODE_SERVER_ROOT}/extensions/{handlebars,pug,json,diff,grunt,
 # Install all dependencies
 WORKDIR /openhands/code
```
raymyers, line 117: This is easier to review using split diff view, which can be found by changing `/files` to `/files?diff=split` in the URL

raymyers:
I've tested both of these runtime images (default and ubuntu) in cloud with tasks:

`Do hello world in OCaml`
And
`Show that you can install and use Python packages`

Both worked, though in ubuntu it had to run pip wiith `--break-package-system` or venv. This is probably not new behavior but I'll check.

raymyers:
> the runtime image being as big as it is is kinda problematic …. mine is almost 10G

Good news, @wolfspyre, I think this took it down a few GB

PR #9542, opened 2025-07-03
Add CLI alias setup for first-time users

xingyaoww:
## Summary

This PR implements a feature to prompt first-time users about setting up convenient shell aliases when they first use OpenHands CLI.

## Features

- **First-time user detection**: Checks if the `~/.openhands` directory exists
- **Interactive prompt**: Allows users to confirm or reject alias setup
- **Shell aliases**: Sets up `openhands` and `oh` aliases using `uvx --python 3.12 --from openhands-ai openhands`
- **UV installation warning**: Displays warning with link to UV installation guide
- **Bash profile integration**: Automatically adds aliases to `.bashrc` or `.bash_profile`
- **Completion tracking**: Only shows the prompt once per user
- **Duplicate prevention**: Checks for existing aliases before adding new ones

## User Experience

When a first-time user runs OpenHands CLI, they see:

```
🚀 Welcome to OpenHands CLI!

Would you like to set up convenient shell aliases?

This will add the following aliases to your bash profile:
  • openhands → uvx --python 3.12 --from openhands-ai openhands
  • oh → uvx --python 3.12 --from openhands-ai openhands

⚠️  Note: This requires uv to be installed first.
   Installation guide: https://docs.astral.sh/uv/getting-started/installation

Set up shell aliases? [Yes, set up aliases / No, skip this step]
```

### Files Modified
- `openhands/cli/utils.py`: Added utility functions for alias setup
- `openhands/cli/main.py`: Added interactive flow and integration
- `tests/unit/test_cli_alias_setup.py`: Comprehensive unit tests

### Technical Details
- Uses existing `cli_confirm` for user interaction
- Prefers `.bashrc` over `.bash_profile` when both exist
- Creates `.alias_setup_completed` marker to track completion
- Handles file system errors gracefully
- Idempotent operation (safe to run multiple times)

## Testing

- ✅ 8 comprehensive unit tests covering all functionality
- ✅ Integration tests verified end-to-end flow
- ✅ All pre-commit hooks pass (linting, formatting, type checking)

## Requirements

- UV must be installed for the aliases to work
- Bash shell (or compatible shell that sources bash profile files)
- Write permissions to home directory

Fixes the requirement to help users set up convenient aliases for OpenHands CLI.

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/7664a692d4f14a7cb839a2504c23fc8e)

PR #9560, opened 2025-07-04
docs: improve and reflow CLI mode docs

llamantino:

Simplifies CLI install instructions by switching to pipx, groups alternative instructions at the bottom, and adds a .local/bin wrapper example for uvx.

---

1. Replaces uvx with pipx as a recommended install method.
2. Moves alternative install methods (docker, uvx, poetry) to the bottom, to reduce confusion.
3. Clarifies that uvx does not install a global command, and adds a one-liner example to create a ~/.local/bin/openhands wrapper for uvx.

---

### DIFF
--- a/docs/usage/how-to/cli-mode.mdx
+++ b/docs/usage/how-to/cli-mode.mdx
@@ -22,15 +22,15 @@ for scripting.
 
 **Note** - OpenHands requires Python version 3.12 or higher (Python 3.14 is not currently supported)
 
-1. Install OpenHands using pip:
+1. Install OpenHands using `pip`:
 ```bash
 pip install openhands-ai
 ```
 
-  Or if you prefer not to manage your own Python environment, you can use `uvx`:
+Or if you prefer not to manage your own Python environment, you can use `pipx`:
 
 ```bash
-uvx --python 3.12 --from openhands-ai openhands
+pipx install openhands-ai
 ```
 
 2. Launch an interactive OpenHands conversation from the command line:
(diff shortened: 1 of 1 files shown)


llamantino:
@xingyaoww Could you please review the changes and decide what's best?  
I noticed your PR that adds the alias setup too late.

docs/usage/how-to/cli-mode.mdx
```diff
@@ -22,15 +22,15 @@ for scripting.
...
 ```bash
 pip install openhands-ai
 ```
 
-  Or if you prefer not to manage your own Python environment, you can use `uvx`:
+Or if you prefer not to manage your own Python environment, you can use `pipx`:
 
 ```bash
-uvx --python 3.12 --from openhands-ai openhands
+pipx install openhands-ai
```
xingyaoww, line 33: Can we keep this to use uvx, and instead put pipx into alternative ways?

Since pipx requires the user to have the correct Python version, which might not always be true

llamantino: Sure. But the previous version of the guide implied that both pip and uvx add a shell command, when in reality only pip does, which is why adding pipx made more sense.
You need to let me know whether the guide should reflect the state before or after #9542 (the alias command PR), so I can update it accordingly. Thanks.

xingyaoww: I think some folks might not have `pipx` installed by default? If they need to install something, i'd hope for that to be `uv` -- so i'm thinking we should make this guide reflect the state AFTER #9542, what do you think?

llamantino: pipx is very well known, whereas uv/uvx is not. I think pip/pipx should remain the first choice to avoid scaring users off with something unfamiliar: uv isn't widely available, and curl | bash install methods aren't particularly appreciated either. 

uv/uvx can still be provided as one of the main methods, but I think it's better to put it below pip/pipx.

I'll close this and wait until the alias PR is ready.

PR #9561, opened 2025-07-05
Jupyter: remove poetry dependency

li-boxuan:

Currently, local runtime assumes OpenHands runs in a poetry environment. This, however, may not be true when you use OpenHands as a python library. For example, if I install OpenHands globally in my uv environment and run local runtime with headless mode:

```
uv run python -m openhands.core.main ...
```

I will hit issues since poetry is not available, OR poetry cannot find "pyproject.toml" - since there's no such file to look for.

---

This PR is a simple fix: instead of assuming we are in poetry environment, simply use whatever python environment we are currently using, let it be poetry, vanilla python, or uv.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:e18dc23-nikolaik   --name openhands-app-e18dc23   docker.all-hands.dev/all-hands-ai/openhands:e18dc23
```


li-boxuan:
I suspect CLI runtime doesn't support Jupyter plugin for this reason, which is totally fixable. Maybe we can enable it now, with this PR?

enyst (approved):
Thank you! Happy to see this, we need it to work IMO.

We could try CLI runtime. I think we kept it simple at the beginning and didn’t need Jupyter dependency, but it would be neat if it works.

Issue #6637, opened 2025-02-06
Please add SANDBOX_NETWORK configuration parameter


olyashok:
**What problem or use case are you trying to solve?**
Have runtime container access docker netowrks

**Describe the UX of the solution you'd like**
I'd like SANDBOX_NETWORK env parameter to be added and then used in docker_runtime.py

**Do you have thoughts on the technical implementation?**
Just add network= parameter into container.run in docker_runtime.py

mamoodi:
@enyst can this be closed now that the PR merged?

enyst:
Yes, this is fixed by
- #9566

PR #9566, opened 2025-07-05
Add additional networks

Joeoc2001:

Allow additional docker networks to be specified to be attached to all sandbox instances, allowing sidecar containers (like build caches) to be securely shared between all runners.

---

Adds additional networks to containers after they have been created. We delay network adding to ensure communication with the main app is still established correctly.

---

#6637

### DIFF
--- a/openhands/runtime/impl/docker/docker_runtime.py
+++ b/openhands/runtime/impl/docker/docker_runtime.py
@@ -196,6 +196,17 @@ class DockerRuntime(ActionExecutionClient):
             self.set_runtime_status(RuntimeStatus.READY)
         self._runtime_initialized = True
 
+        for network_name in self.config.sandbox.additional_networks:
+            try:
+                network = self.docker_client.networks.get(network_name)
+                network.connect(self.container)
+            except Exception as e:
+                self.log(
+                    'error',
+                    f'Error: Failed to connect instance {self.container_name} to network {network_name}'
+                )
+                self.log('error', str(e))
+
     def maybe_build_runtime_container_image(self):
         if self.runtime_container_image is None:
             if self.base_container_image is None:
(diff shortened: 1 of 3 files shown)


openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -196,6 +196,17 @@ async def connect(self) -> None:
...
         self._runtime_initialized = True
 
+        for network_name in self.config.sandbox.additional_networks:
+            try:
+                network = self.docker_client.networks.get(network_name)
+                network.connect(self.container)
+            except Exception as e:
+                self.log(
+                    'error',
+                    f'Error: Failed to connect instance {self.container_name} to network {network_name}'
```
enyst:
```suggestion
                    f'Error: Failed to connect instance {self.container_name} to network {network_name}',
```

enyst:
@Joeoc2001 There’s some type failure in CI. Could you perhaps take a look at it?

enyst:
I am AI Agent OpenHands-GPT-5. I investigated the CI failure on this PR and pushed a fix.

Summary of the issue:
- The failing job was 'Lint python'.
- mypy reported: `openhands/runtime/impl/docker/docker_runtime.py:219: error: Argument 1 to "connect" of "Network" has incompatible type "Container | None"; expected "str | Container"  [arg-type]`.
- Root cause: `self.container` is typed as `Container | None`, but `network.connect(...)` expects a `str | Container`. Passing a possibly-None value triggered the type error.

What I changed:
- In DockerRuntime.connect(), after the runtime is initialized and before connecting to any additional networks, I added a None-guard around the `network.connect(self.container)` call. If `self.container` is unexpectedly None, we log a warning and skip that network attach for safety.
- This is the minimal change needed to satisfy mypy while keeping the intended behavior intact.

Commit with the fix:
- 94554d777 (fix(docker): satisfy mypy by guarding network.connect against None container)

Verification:
- Ran the repo's pre-commit (ruff + mypy) locally; it now passes. The push updated this PR's branch so CI should turn green for the Python lint step.

Notes:
- The new `additional_networks` field and the network-attach logic remain functionally the same, now with a small safety/logging improvement.
- No behavior change for valid cases (container present). Errors during attach continue to be logged, as before.

If you'd like me to also switch the `additional_networks` field to use `default_factory=list` (style preference in pydantic v2 to avoid mutable defaults), I can do that in a follow-up.

PR #9570, opened 2025-07-06
Add core config to disable browser environment

li-boxuan:

Right now, there's no way to disable browser environment, even if you disable browsing feature in a specific or all agents' configs. When there's no need to use browser, it is costly and unnecessary to start playwright environment. Moreover, when using openhands as a python library, playwright dependencies are not available without manual installation.

---

Add a config to enable/disable browser environment. Default behavior is enabled.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:456ab02-nikolaik   --name openhands-app-456ab02   docker.all-hands.dev/all-hands-ai/openhands:456ab02
```

### DIFF
--- a/config.template.toml
+++ b/config.template.toml
@@ -49,6 +49,9 @@
 # Maximum file size for uploads, in megabytes
 #file_uploads_max_file_size_mb = 0
 
+# Enable the browser environment
+#enable_browser = true
+
 # Maximum budget per task, 0.0 means no limit
 #max_budget_per_task = 0.0
 
@@ -226,6 +229,7 @@ model = "gpt-4o"
 [agent]
 
 # Whether the browsing tool is enabled
+# Note: when this is set to true, enable_browser in the core config must also be true
 enable_browsing = true
 
 # Whether the LLM draft editor is enabled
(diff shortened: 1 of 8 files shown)


neubig (commented):
This is good, thank you!

But
1. Do we need enable_browser in both the core and agent config? It seems like we could just do it in the core config
2. In the command line we have enable-browser, but since the browser is on by default shouldnt it be disable-browser?

li-boxuan:
> Do we need enable_browser in both the core and agent config? It seems like we could just do it in the core config

Although OpenHands officially recommends using CodeActAgent only, there are presumably customers who use their custom agents, or even delegation mechanism, where they might still want fine-grained control on what agent could use browser and what shall not.

> In the command line we have enable-browser, but since the browser is on by default shouldnt it be disable-browser?

Sure, I can flip that. In any case, it's not exposed to customers.

PR #9586, opened 2025-07-07
Added run_in_loop method

tofarr:

Adds a new utility function `run_in_loop` to handle the common "coroutine was created in a different event loop" error that can occur in async Python applications. This function intelligently detects whether a coroutine needs to be executed in a different event loop and handles the cross-loop execution automatically.

---

This PR introduces the `run_in_loop` function in `openhands/utils/async_utils.py` that:

1. **Mitigates event loop errors**: Solves the "coroutine was created in a different event loop" error by automatically detecting when a coroutine needs to be executed in a different loop
2. **Smart loop detection**: Compares the current running loop with the target loop and chooses the appropriate execution strategy
3. **Same loop optimization**: When the target loop is the same as the current loop, it simply awaits the coroutine directly for optimal performance
4. **Cross-loop execution**: When loops differ, it uses `asyncio.run_coroutine_threadsafe` to safely execute the coroutine in the target loop
5. **Timeout support**: Supports configurable timeouts for cross-loop operations (timeout only applies to different loop scenarios)
6. **Exception propagation**: Properly propagates exceptions from the executed coroutine

**Key design decisions:**
- Uses `call_sync_from_async` and `_run_in_loop` helper for cross-loop execution to maintain async/await patterns
- Timeout parameter only applies when using different loops (same loop execution is direct await)
- Leverages existing `call_sync_from_async` infrastructure for consistency with the codebase

**Comprehensive test coverage added:**
- ✅ Same loop scenario (direct await)
- ✅ Different loop scenario (cross-loop execution)
- ✅ Exception handling and propagation
- ✅ Timeout behavior with different loops
- ✅ Documentation that timeout doesn't apply to same loop
- ✅ Return value handling (various data types)
- ✅ Coroutines with arguments and keyword arguments
- ✅ Synchronous context usage

All tests pass and follow the existing test patterns in the async_utils test suite.

---

This addresses common async programming issues where coroutines created in one event loop need to be executed in another, which is a frequent source of runtime errors in complex async applications.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:6340f59-nikolaik   --name openhands-app-6340f59   docker.all-hands.dev/all-hands-ai/openhands:6340f59
```

### DIFF
--- a/openhands/events/event_store.py
+++ b/openhands/events/event_store.py
@@ -49,12 +49,23 @@ class EventStore(EventStoreABC):
     sid: str
     file_store: FileStore
     user_id: str | None
-    cur_id: int = -1  # We fix this in post init if it is not specified
     cache_size: int = 25
+    _cur_id: int | None = None  # Private field to cache the calculated value
 
-    def __post_init__(self) -> None:
-        if self.cur_id >= 0:
-            return
+    @property
+    def cur_id(self) -> int:
+        """Lazy calculated property for the current event ID."""
+        if self._cur_id is None:
+            self._cur_id = self._calculate_cur_id()
+        return self._cur_id
+
+    @cur_id.setter
+    def cur_id(self, value: int) -> None:
+        """Setter for cur_id to allow updates."""
+        self._cur_id = value
+
+    def _calculate_cur_id(self) -> int:
+        """Calculate the current event ID based on file system content."""
         events = []
         try:
             events_dir = get_conversation_events_dir(self.sid, self.user_id)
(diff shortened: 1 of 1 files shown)


openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -297,10 +305,13 @@ async def _start_agent_loop(
...
-                    'oh_event',
-                    status_update_dict,
-                    to=ROOM_KEY.format(sid=oldest_conversation_id),
+                await run_in_loop(
+                    self.sio.emit(
+                        'oh_event',
+                        status_update_dict,
+                        to=ROOM_KEY.format(sid=oldest_conversation_id),
+                    ),
+                    self._loop,  # type:ignore
```
malhotra5, line 314: Curious about the typing issues here, what was the error that was being thrown?

malhotra5 (approved):
LGTM!

openhands/server/conversation_manager/standalone_conversation_manager.py
tofarr: The issue is that the event loop may be None if this is without the `__enter__` method (Which is never will).
The type checker is warning us that this will fail:
```
manager = ConversationManager()
manager.maybe_start_agent_loop(...)
```
And instead we need to do:
```
with ConversationManager() as manager:
  manager.maybe_start_agent_loop(...)
```

Issue #9585, opened 2025-07-07
Unify Configuration Architecture: Eliminate Parallel Systems


enyst:
Update (2025-08-14): Scope clarification and plan for incremental delivery

This issue will be delivered in phases. Phase 1 focuses exclusively on adding a robust TOML write API that mirrors our read path. Runtime hot-switch of LLM configs and Web UI wiring will be handled in later phases.

Scope (Phase 1)
- Implement and validate a TOML write API that persists OpenHands configuration sections to config.toml, mirroring the existing read-from-TOML behavior.
- Keep user_secrets persisted via the existing SecretsStore to secrets.json (not TOML).
- Extend writer coverage beyond [core], [llm], [agent] to additional sections (security, sandbox, kubernetes, cli, extended, condenser, mcp), with focused tests.
- Do not implement runtime hot-reload/hot-swap now.

Out of scope for Phase 1 (tracked for later)
- [Future] Hot-reloadable configs and runtime hot-swap of LLM instances.
- [Future] Web UI settings integration to the unified config chain.
- [Future] Format/comment preservation in TOML (may require tomlkit).

API direction for the write path
- Central Writer (recommended now): Keep a single TOMLConfigWriter responsible for merging and writing to config.toml. Leverage Pydantic model serialization (including unions, nested models, lists, dicts, SecretStr) and strip None/unset values.
- Class-level serialization (consider): Each config model exposes to_toml_data()/from_toml_data() helpers that return/consume dicts. The central writer orchestrates file I/O and merging; models remain decoupled from storage.
- Avoid per-class direct file writes to prevent merge conflicts and duplicated I/O logic.

Secrets
- Persist user_secrets via the existing FileSecretsStore (secrets.json). CLI/headless/web can access UserSecrets through SecretsStore; no change for Phase 1.

Status snapshot (current PR #10305)
- Writer extended to persist: [security], [sandbox], [kubernetes], [cli], [extended], [condenser] (union-safe w/ type), [mcp].
- Serializer enhanced for nested models/lists/dicts/secrets; strips None/unset values.
- Focused tests added for new sections (all passing locally); pre-commit clean; Windows import cycle fixed.

Next steps
- Validate full test suite in CI; address any regressions.
- Decide if/when to add format preservation with tomlkit (follow-up issue).
- Track runtime hot-switch and Web UI wiring in separate PRs under this issue.

----
Original write-up (for context)
(Claude write-up)

## Problem

[...]

PR #8749, opened 2025-05-28
Add documentation for using OpenHands as a library

rbren:
This PR adds documentation on how to use OpenHands as a Python library in your own projects. It includes:

- Basic setup and configuration
- Component overview (Agent, LLM, Runtime, Memory)
- Advanced usage examples
- A complete application example

The documentation follows the existing structure and provides a "hello world" example as requested.

PR #9589, opened 2025-07-07
[Refactor]: Add LLMRegistry for llm services

malhotra5:
---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:fe73918-nikolaik   --name openhands-app-fe73918   docker.all-hands.dev/all-hands-ai/openhands:fe73918
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@centralized-metrics openhands
```


openhands/controller/state/state.py
```diff
@@ -116,6 +117,8 @@ class State:
     local_metrics: Metrics | None = None
     delegates: dict[tuple[int, int], tuple[str, str]] | None = None
 
+    metrics: Metrics = field(default_factory=Metrics)
```
enyst, line 120: I wonder, now if we need and use a `MetricsRegistry`, maybe we shouldn’t use `metrics` and `local_metrics` anymore? If we can… 🤔

These are from the main LLM only. We will need to track/restore all of them, not only those the main LLM. Moreover, I think we will allow switching LLM during a conversation, which means we’d need to do somewhere something like, read from registry the current metrics of the LLM being switched, give them to the new LLM. 🤔

These values in state… don’t seem to help, would it be possible to remove them? What were they doing, other than delegates switch?

malhotra5: Yeah these need to be removed; its just for backwards compatibility for now (we'd move the previously stored metrics into the registry)

enyst: I mean, could we remove `metrics` and `local_metrics`?

openhands/core/main.py
```diff
@@ -52,7 +52,6 @@ async def run_controller(
     initial_user_action: Action,
     sid: str | None = None,
     runtime: Runtime | None = None,
-    agent: Agent | None = None,
```
enyst, line 56: Unfortunately, it may be worth checking if evals work with the agent set here or not, I’m sorry, don’t recall off-hand. And the resolver. They’re our main clients for `run_controller`.

Of course, from within OH. External clients may have a compatibility issue. (I think we could live with it, in the future we will probably need to somehow document and stick to some public API where we notify people of breaking changes)

malhotra5: Yep definitely 

I took a quick look and saw that none of the scripts we have are passing their agents, they're instantiated in the `run_controller` method based on the configs

This meant I could remove it; I feel okay with this, as the agent needs to be initialized with the metrics registry which I think is an unnecessary burden to manage for the resolver, evals, etc

openhands/controller/state/state.py
malhotra5: Yep `metrics` is being deprecated and `local_metrics` is already deprecated; the actual removal from the state object will be done after some time

Conversations that have existed before metrics registry have been rolled out will store accumulated costs in the `metrics` field. For those conversations that are being resume, we'd like to retain the cost info by migrating it into the metrics registry (as it drives the budget control flags) as opposed to dropping the metrics object completely. 

I haven't finished this part of the code.

openhands/server/session/agent_session.py
```diff
@@ -433,12 +441,12 @@ def _create_controller(
             '-------------------------------------------------------------------------------------------'
         )
         self.logger.debug(msg)
-        initial_state = self._maybe_restore_state()
```
enyst: Do we move this? Or what’s up with it?

malhotra5: We restore the state for the agent controller before we initialize it; we're also storing the metrics registry inside the state.

Unfortunately, we need access to the metrics registry much before the initial state is restored (inside the conversation manager, so that we can pass it over to other services like the convo title generator)


So we have two options

1. Store metrics registry separately so that we can restore it elsewhere and pass it to the state later
2. Restore the state inside the conversation manager, and then plumb it into session > agent session > agent controller


I tried the second option but I'm strongly considering the first one instead

openhands/llm/metrics_registry.py
```diff
@@ -0,0 +1,40 @@
+from enum import Enum
+from uuid import uuid4
+
+from openhands.core.logger import openhands_logger as logger
+from openhands.llm.metrics import Metrics
+
+
+class LLMService(Enum):
+    AGENT = 'AGENT'
```
enyst: What if:
- have a class LLMService, instead of enum
- with a dict (maybe) of each service name -> the LLM
- creates the respective LLM, when agent, etc, asks for it
- (maybe) contains the registry => give each LLM its metrics upon creation

I’m thinking also of a bit complicated consistency with [this](https://github.com/All-Hands-AI/OpenHands/issues/9585), on
- LLM configs have themselves a registry of sorts, so you can always ask “give me the condenser LLM config” (we have now a dict)
- but you will need to be able to also say “switch the main LLM to ‘llm-name’”. The switch will handle the metrics transfer from one to the other 🤔
- the enum here is the roles of the currently running LLMs. That feels like we should merge it with the LLMs registry that accounts for all those configured in the application.

Sorry if that didn’t make sense, writing a bit on the run. I think we may want to figure out the architecture here.

openhands/server/session/agent_session.py
enyst: 🤔 Doesn’t the conversation title generator already assume a conversation, and thus a running agent loop? Which means the agent controller should exist?

enyst: Aha I see where it moved. Just to note, the code removed (the if there’s a latest event id) is supposed to help with logging clearly, at warning, when the conversation _lost_ the state. So anytime we have events already, we should have saved `State`; if we didn’t, maybe we should know

openhands/llm/metrics_registry.py
malhotra5: Yeah definitely makes sense. Currently, the delegate agent would request the same service name as its parents so I'm not too worried about the case where we transfer metrics

A problem I'm facing is the following - 

1. We are introducing solvability reports to the cloud resolver (you may have seen this)
2. We want create a solvability report BEFORE the conversation is started (this is the tricky part, which makes it hard to associate the cost with a given convo. For example, we may want to avoid starting a convo if the agent feel it absolutely cannot solve the user's request) 
3. We also have this [remember prompt functionality](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/server/routes/manage_conversations.py#L301). It needs a way to access the current conversations llm service. 


I'm thinking we may want a more principled way of requesting extraneous llm completion calls for a given conversation that isn't a part of the event stream.

enyst: Re: 3. That one applies to a conversation, but we do it too early… but does it have the conversation id ? It should, since it can find the event stream 🤔

Re: 2. Got it… This is a strange one. In OH, this shouldn’t be a problem… because we can create an LLM from LLMConfigs, we don’t need a conversation for that. I do wonder if that may bite us though… How are we going to restore _those_ metrics if not for a session?

malhotra5: I'm thinking we should have a way to enforce 

1. LLMService only instantiates after a conversation is created
2. Currently a new conversation created means that the runtime is started as well. Maybe we should separate those two more explicitly. So effectively conversation are always "resumed". Stage 1 is claiming a conversation id. Stage 2 would be to resume the agent loop + runtime?

malhotra5: So extraneous completions can be run in between stage 1 and stage 2?

enyst: The idea of separating conversation creation from runtime starting sounds good to me. 🤔

Does that mean that case 2 above would be between 1 and 2? And maybe start runtime later?

malhotra5: Yep exactly; the default behavior would be to always start the runtime, just concerned about associating cost under a specific session for now

enyst: (just a quick thought) In order to know the running LLM of the agent, we need to know which agent. I _think_ that comes before the runtime… I hope in all cases.

openhands/llm/metrics_registry.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+
+    def get_combined_metrics(self) -> Metrics:
+        total_metrics = Metrics()
+        for llm in self.service_to_llm.values():
+            total_metrics.merge(llm.metrics)
+
+        return total_metrics
+
+    def save_registry(self):
```
malhotra5: Here are some thoughts about why I build the serialization and deserialization this way

We want the ability for multiple LLMRegistry to exist for the same conversation. But we need to make sure that there aren't race conditions between the two. 
- We make the assumption that services will always be mutually exclusive (no two LLMRegistry will register the same LLM service)
- We keep a central record of all the LLM registries that exist (the assumption is that LLM services are always registered sequentially in the conversation lifecycle, so we don't worry about race conditions here)
- The LLM Registry only flushes the LLM Services that its responsible for tracking

This way, we can have an LLMRegistry for the session, but we can also create more LLMRegistry for the same conversation on demand for extraneous services (such as the remember prompt). This also means built in support for distributed server architectures. 

**Downside**

We need a complete aggregate of metrics because it will drive the budget control flags. I'm still not sure how we would sync metrics across multiple LLM Registry, but I'd think we need to support a multiple registry model if we continue to request more LLM completions through the app server endpoints (remember prompt endpoint for example).

malhotra5: I think one way to get around this would be to ensure that if a conversation is running, then all llm completions must go through the runtime (this would work for NestedRuntimes that we're switching over to soon)

Then we'd just need to make sure that a single Registry exists in the runtime

EDIT: But we have many runtime providers, and we'd still want to track cost accurately so we can't rely on this method alone

enyst: > We want the ability for multiple LLMRegistry to exist for the same conversation

Sorry, why? Or do you mean multiple LLM instances and a single LLMRegistry?

enyst: How about something like
```
LLMRegistry
- contains some dict maybe, or whatever to know these {
‘agent’ -> LLM(LLMConfig_from_agent_or_default)
‘condenser’ -> LLM(from llm.condenser or default)
‘memorize_message’ -> LLM(probably default)
}
- get/read
- register
- compute metrics per service
- get metrics for x service
```

(Outdated, I just saw the actual implementation in the PR right now and it looks good!)

malhotra5: Cool! Basically the last part that's left is to have a way to access the LLMRegistry that's attached to a session from the FastAPI endpoints

Because the LLMRegistry is in mem; I'm thinking we go through the `conversation_manager` (we map the convo_id to session here). ~~The downside is in a distributed server architecture the LLMRegistry ends up being "sticky", but I think there's no getting around that if this is the way we're doing it~~ EDIT: that's not necessarily true I think this will work; will continue iterating on the PR

openhands/server/routes/manage_conversations.py
```diff
@@ -357,8 +369,9 @@ def generate_prompt(llm_config: LLMConfig, prompt_template: str) -> str:
         },
     ]
 
-    response = llm.completion(messages=messages)
-    raw_prompt = response['choices'][0]['message']['content'].strip()
+    raw_prompt = ConversationManagerImpl.request_llm_completion(
+        'remember_prompt', conversation_id, llm_config, messages
+    )
```
malhotra5, line 399: I reworked this so that 

1. There's always exactly one LLM Registry per session
2. To request extraneous llm completions, we go through the convo manager to find the session then execute an llm completion
3. This also ensures that extraneous completions are aggregated and accounted for the final budget control flags for a conversation

malhotra5:
Spend tracking is working as expected for conversation started with GUI!

I have the following left to do

- Refactor resolver, headless, CLI, etc to use LLMRegistry
- Ensure docker nested conversation uses LLMRegistry + can handle extraneous llm requests as well
- Write comprehensive unit tests

openhands/cli/main.py
```diff
@@ -142,9 +144,14 @@ async def run_session(
         None, display_initialization_animation, 'Initializing...', is_loaded
     )
 
-    agent = create_agent(config)
+    # TODO: do we need to restore metrics for CLI?
+    file_store = get_file_store(config.file_store, config.file_store_path)
+    llm_registry = LLMRegistry(file_store, sid, None)
+
+    agent = create_agent(config, llm_registry)
```
enyst, line 157: I'm wondering, should we send over `llm_registry` in all agents etc, or could we
- create the LLM here, since we have the info, and send it?
- or, make a singleton LLMRegistry, that we can import and use where needed. It really is only one, isn't it?

enyst: Or, hmm, I take back the latter... Let me know your thoughts about all this!

malhotra5: I did it this way because all `agent_cls` are initialized with an llm registry. I felt this made sense because the agent can decide which llm's to register for itself. It could even instantiate multiple llm's and switch between them if needed to - completely up to the agent.

enyst: That makes sense too. The reason why I ask is that I feel we should keep it simple from the perspective of client code. For example, an agent developer could ideally use `openhands` like this:
https://github.com/All-Hands-AI/OpenHands/pull/8749/files#diff-5716a56c5060772d3b22d8b2e9956719685092dd921b13a593abb304ad55221bR11

I feel agent developers will obviously expect an llm (whether as a config or an instance). On the other hand, the more args and prior knowledge we require, about how `openhands` works, the more of a toll it takes on their time to get up to speed. 🤔 

_Edited to add:_ actually not only "agent developer" but any developer using OH as a library.

malhotra5: Yep that's one reason I removed the `agent` parameter from the `run_controller` function

From a developer's perspective, they just need to pass configs. Creating the agent, initializing the LLM registry and registering LLMs is all handled automatically underneath the hood. From my perspective developers shouldn't be passing any instances (for agent, llm, runtime, etc)- they should just provide configs which get translated downstream

enyst: I agree they should pass configs, I'm not sure the registry is, maybe!

openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -73,24 +75,28 @@ class CodeActAgent(Agent):
 
     def __init__(
         self,
-        llm: LLM,
         config: AgentConfig,
+        llm_config: LLMConfig,
+        llm_registry: LLMRegistry,
```
rbren: I'd generally expect this to follow a singleton pattern rather than getting passed around

rbren: Looks like Engel brought this up too--catching up now

rbren (approved):
Approving for now but let's wait for enyst to look again

openhands/agenthub/visualbrowsing_agent/visualbrowsing_agent.py
```diff
@@ -129,15 +132,20 @@ class VisualBrowsingAgent(Agent):
...
+        requested_service: str | None = None,
     ) -> None:
         """Initializes a new instance of the VisualBrowsingAgent class.
 
         Parameters:
         - llm (LLM): The llm to be used by this agent
         """
-        super().__init__(llm, config)
+        super().__init__(
+            config, llm_config, llm_registry, retry_listener, requested_service
```
enyst: One tiny thought I have here, is why do people need to know all this stuff, to init an agent? In the same idea as we talked before, where simpler ways for folks to use `openhands` are better, giving an agent a retry_listener for example doesn't seem to help that? It's optional, but it means your run doesn't display in the UI when it's rate limited, and you won't know why until you see and understand how this works.

It's extra complexity for the agent developer... Do we have to do it?

openhands/controller/agent.py
```diff
@@ -38,10 +40,21 @@ class Agent(ABC):
...
         config: AgentConfig,
+        llm_config: LLMConfig,
+        llm_registry: LLMRegistry,
+        retry_listener: Callable[[int, int], None] | None = None,
+        requested_service: str | None = None,
     ):
-        self.llm = llm
+        if requested_service:
+            self.llm = llm_registry.request_existing_service(
+                requested_service, llm_config, retry_listener
```
enyst: Sorry, still reading, maybe I'm missing it, but just to note I don't understand why we need to send `requested_service` in the agent, does it ever get instantiated with it?

openhands/controller/agent_controller.py
```diff
@@ -670,16 +669,18 @@ async def start_delegate(self, action: AgentDelegateAction) -> None:
         )
         self.log(
             'debug',
-            f'start delegate, creating agent {delegate_agent.name} using LLM {llm}',
+            f'start delegate, creating agent {delegate_agent.name}',
         )
 
+        print('setting delegate')
```
enyst: Nit: we should clean this up before merge

openhands/controller/agent.py
malhotra5: This is in the event that the agent performs a delegation. We want the delegate to pick up the llm that it's parent used (that's the current behavior, but in the future that may change).

openhands/controller/agent_controller.py
malhotra5: Ah thanks, will do

openhands/agenthub/visualbrowsing_agent/visualbrowsing_agent.py
malhotra5: Ahh yeah you're right, ideally we don't have this here

The retry listener is mounted to an llm object, which gets created after an agent is instantiated. The plumbing worked out this way.


Instead the LLMRegistry should be initialized once with a retry listener and the registry automatically mounts the listener to all the LLMs that are created. 

I think that should be feasible, need to take a second look at the plumbing for this

enyst: For `requested_service`, this seems more strange: this is an agent initialization, we already know what the service is, don't we? It's always the 'main' LLM. Or am I missing something?

... Actually, it's a bit more messy perhaps, because LLM can be set per agent...

malhotra5: Yep LLM can be set per agent

malhotra5: I might have backed in too much flexibility here though - even if the delegate is made with a different LLM its cost can be tracked as a cumulative

For model routing though, we'd want to know the cost split between the smaller vs bigger model

openhands/memory/condenser/impl/llm_attention_condenser.py
```diff
@@ -34,7 +41,7 @@ def __init__(self, llm: LLM, max_size: int = 100, keep_first: int = 1):
 
         self.max_size = max_size
         self.keep_first = keep_first
-        self.llm = llm
+        self.llm = llm_registry.register_llm('attention_condenser', llm_config)
```
enyst: I think maybe the service_id here is just 'condenser'? If you set this condenser, then that's the one

enyst: In which case, we could perhaps avoid to pass it to specific types, and handle it in `from_config`. I don't think this is a big issue, though, it seems easy to fix

openhands/memory/condenser/impl/llm_summarizing_condenser.py
```diff
@@ -163,10 +165,10 @@ def from_config(
         llm_config.caching_prompt = False
 
         return LLMSummarizingCondenser(
-            llm=LLM(config=llm_config),
+            llm_config=llm_config,
+            llm_registry=llm_registry,
             max_size=config.max_size,
             keep_first=config.keep_first,
-            max_event_length=config.max_event_length,
```
enyst, line 169: Is this intended? I think it changes behavior.

openhands/resolver/interfaces/issue_definitions.py
```diff
@@ -20,10 +20,19 @@ class ServiceContext:
...
+        llm_config: LLMConfig | None,
+        service_id: str | None = None,
+    ):
         self._strategy = strategy
+        # TODO: we should store registry information as well
+        llm_registry = LLMRegistry(None, 'convo_id', None)
         if llm_config is not None:
-            self.llm = LLM(llm_config)
+            # Use a unique service ID for each instance to avoid duplicate registration
+            self.service_id = service_id or f'resolver_embelishment_{id(self)}'
```
enyst: Out of curiosity, are there multiple ServiceContexts in a run?

enyst: Actually, maybe the better question is, why do we need a service_id here at all? Please correct me if wrong, I thought service ids are things like main (llm), condenser (llm), draft (llm), that is, the components in the code that may have different LLMs assigned and running.

If so, this seems to be only main?

openhands/controller/agent.py
enyst: Yes, but it always gets the main service, doesn't it?

openhands/memory/condenser/impl/llm_summarizing_condenser.py
malhotra5: thanks for the catch, was not intended and fixed

openhands/controller/agent.py
malhotra5: Yep; this logic a guardrail where we shouldn't accidentally instantiate two llms with the same service unless we're intentionally inheriting the metrics from a different llm (for agent delegates specifically). The `request_existing_service` is meant for this.

I'm ok removing this logic though, it seems to be complicating things more than its helping

enyst: Yes, I think we could simplify... also `request_existing_service` probably doesn't need the other two args?

openhands/resolver/interfaces/issue_definitions.py
malhotra5: Ah no I had a bug where the service id was a class var so this was openhands way of getting around failing unit tests, which didn't reset the class var and were sharing it across multiple objects

I fixed that bug so this should just be 'main' or 'resolver' as the id

openhands/memory/condenser/impl/llm_attention_condenser.py
malhotra5: Done!

openhands/cli/main.py
enyst: In [Ryan’s routing PR](https://github.com/All-Hands-AI/OpenHands/pull/9738), it’s also the case that the agent starts doing more things managing its routing LLMs… which can’t be combined nicely with this

Maybe this part:
> agent can decide which llm's to register for itself. It could even instantiate multiple llm's and switch between them if needed to - completely up to the agent

is too much to want for now? 🤔 This isn’t a functionality we support now, the user decides what LLMs the agent will use.

Ryan’s routing feature comes closer to the idea that ok, maybe, in the agent code, a developer will switch LLMs by applying routing there. Idk though, it makes sense, but there are also arguments against ( = pro simplicity of agent code)

openhands/llm/llm.py
```diff
@@ -378,7 +383,7 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
                 log_file = os.path.join(
                     self.config.log_completions_folder,
                     # use the metric model name (for draft editor)
```
```suggestion
```
Totally a nit: pretty sure this comment was already wrong, and the PR surely makes it wrong if not

openhands/cli/main.py
malhotra5: Hmm okay let me take a step back - the main concern right now is that we shouldn't have agent builders worry about using an LLMRegistry and all the complexities that come with it 

We however also want to track the llms which we pass over to the agent, because it helps us track its completion cost. 

At the moment in the agent class, we instantiate at most two LLMs - one for the agent and optionally one for the condenser that's initialized inside the agent. Previously the agent and condenser would construct their own LLM objects as needed from the their corresponding configs. Now, we pass a registry which gets used to create the agent's llm and the same registry is passed into the condenser to create another optional llm.

If we were to construct the llms and pass them over, we'd need to create the optional condenser llm in advance and plumb it through as well (seems a little ugly). With model routing on the way, we may need to construct multiple llms in advance (seems even more ugly). I think this also runs the risk of us accidentally using the LLM class to create an LLM rather than using the registry. I think I would be less concerned if we enforced a specific pattern for creating LLMs in advance for a given agent?

How about we revert all the changes to the agent class so far. In the base Agent class we'd only make the following change

```python
class Agent:
  ... 
  
    def init(self, agent_config: AgentConfig, all_llms: dict[str, LLM]):
        self.all_llms = all_llms
        if 'main' in all_llms:
            self.llm = all_llms['main']
```

and a separate builder class 

```python
class AgentBuilder:
     def init(self, llm_registry: LLMRegistry | None):
         if not llm_registry:
              self.llm_registry = LLMRegistry()
         else:
             self.llm_registry = llm_registry
         
     def build_agent(agent_cls, agent_config, llm_configs: dict[str, LLMConfig]):
         all_llm = {}
         for llm_name, llm_config in llm_configs.values():
             all_llms[llm_name] = self.llm_registry.register_llm(llm_name, ...)
     
         agent = Agent.get_cls(agent_cls)(agent_config, all_llms)
         return agent
```


Just split balling here, but thoughts on something like this? We'd enforce that all agent are created via the `AgentBuilder`

EDIT: the downside is that when additional params are introduced to agent's constructor, this will require agent developers to meddle with the `AgentBuilder` class as well... which sucks

How about the agent builder has the `build_agent` method, but also a separate `set_llms` method. So we can have separation of concerns that way. The `build_agent` method I'd imagine agent developers would be okay meddling with.

enyst: I see where you’re getting at… we need the LLMs, and with routing, more LLMs than before, and with ‘switch LLM at runtime’, _all_ LLMs. And we need to track them.🤔

1) Let me take one step back: we have LLMRegistry in `State` at the moment, so maybe we don’t actually need to change the Agent signature to give it access: it could pick it up from `State`?

2) Let me take two steps back: we don’t need all LLMs instantiated ahead of time, if we replace all constructors with a factory method, e.g.
```
LLMRegistry
    def get_llm(llm_config: LLMConfig)
        # retrieve it from some inner dict if it was instantiated before
        # create it, and memorize it, if not
```

3) One more step back, to note: I think you were absolutely correct that the _stats_ need to be per conversation, and it would be too much to have all conversations (even 3 running per user)

I think a funny thing that was happening though, is that to you, the registry’s responsibility must be primarily that, to be the store for stats, while to me its primary responsibility is as store for llm configurations and (llms or llm factory)

We thought they should be the same, but maybe that’s too complex, maybe they aren’t:
LLMRegistry
  - all llm configs (those currently in openhands_config)
  - mappings, e.g. agent to llm (from openhands_config)
  - (maybe) the current service mapping
  - `get_llm` method (with args config, service_id)
  - contains `all_llms` you mentioned above, or a way to create them
  - able to return active llm… wait, that’s the same with current service mapping
  - is an Observable
  
ConversationStats
  - is a listener to the registry
  - gets notified when the registry register_llm is called => maybe stores a reference to the new llm
  - method adding metrics from ‘agent’, ‘condenser’ etc
  - all service mappings? currently only one I guess, but routing and LLM switch will change this
  - conversation_id
  - maybe this is the object we need to serialize/restore?
  - something else?

Does this make sense?

malhotra5: I think we have two competing interests

1. agent developers want to create llms as simply as possible and focus on the agent logic
2. the main application developers want a strict method of creating llms, so that its cost can be tracked centrally

My hunch is we're going to need a middle ground, like the `AgentBuilder` class where these two interests meet. Ideally we expose a `create_llm` method of some sort, that the agent developers can use instead

EDIT: oops just saw your comment, will take a bit to read through it carefully apologies in advance if this comment seems out of place

enyst: > the risk of us accidentally using the LLM class to create an LLM rather than using the registry

This one is fixable, we could forbid or give exception in init or refuse/replace creation in metaclass, shouldn’t be a problem.

enyst: Also, if we go for the split, and even if we don’t, I have a silly question: if this stats object is important, ever present, and tied to the Conversation, could we make it an attribute of the Conversation?

malhotra5: I like the ideas so far. My one concern is that 

`LLMRegistry` creates llm > send info via listen to `ConversationStats`

The llm contains the metric which are accumulated via completion calls.

Now when restoring the metrics, we'd need to find the llm metrics in `ConversationStats` > instantiate the llm with restored metrics in `LLMRegistry`; the circular dependency is concerns me a bit



I've recently update the `LLMRegistry`; the `register_llm` method looks pretty much the same as the `get_llm` method described above. Wondering if doing the following would help

1. store all llm configs (those currently in openhands_config)
2. store mappings, e.g. agent to llm (from openhands_config)
3. we initialize llms via the configs

malhotra5: > could we make it an attribute of the Conversation

Yep thought about doing that. We currently store accumulated cost in the convometadata. I thought of using a producer consumer model. The llms would produce completions events with associated cost, and emit a webhook to the main app server

We'd have a consumer aggregate these events and keep a total for the convo

The implementation would largely look like a message queue. LLMs could be instantiated anywhere and as long as they're setup with a couple of envs they will emit completion events

The desire to avoid a message queue like architecture motivated me to track spend on a per service basis in a central dict, each service would be isolated and the total would be aggregated whenever requested by the agent controller

openhands/llm/metrics_registry.py
enyst: This discussion is solved, right? It confuses me a bit now 😅

openhands/server/routes/manage_conversations.py
enyst: I think this one is solved, right? We go with this idea, my last comment proposing to split objects doesn’t change this, on the contrary, the stats are per session

openhands/llm/metrics_registry.py
malhotra5: Yep this is resolved

openhands/cli/main.py
malhotra5: Actually let me take a shot at splitting the registry into an llmfactory and convostats

I think I can make that work

enyst: > Now when restoring the metrics, we'd need to find the llm metrics in ConversationStats > instantiate the llm with restored metrics in LLMRegistry; the circular dependency is concerns me a bit

🤔 Good point! I need to come back on it. Provisorily, I think it will be fixable, and maybe even good, because it might imply a bit more separation between LLM and “its” metrics. We know why metrics are tied to the LLM object, but… maybe we need it less tight, because routing/switch might mean that we keep the same LLM object and change its configuration at runtime, as [this PR](https://github.com/All-Hands-AI/OpenHands/pull/9800) is preparing… Or is that a terrible idea?

enyst: Actually if one direction of the dependency is via Observable/Listener, it should just work 🤔 but maybe I’m hallucinating worse than my Gemini

enyst: > Wondering if doing the following would help
store all llm configs (those currently in openhands_config)
store mappings, e.g. agent to llm (from openhands_config)
we initialize llms via the configs

Maybe not, or not all, if you feel that the code is clear enough when we use them directly as they are in config. It’s just me wanting to encapsulate some of that large openhands_config in a clear way. Not needed if we can work with it

malhotra5: > Actually if one direction of the dependency is via Observable/Listener, it should just work


Yep exactly, need to double check against saas but I think I can make it work

openhands/server/conversation_manager/docker_nested_conversation_manager.py
```diff
@@ -469,22 +481,22 @@ async def _create_runtime(
     ) -> DockerRuntime:
         # This session is created here only because it is the easiest way to get a runtime, which
         # is the easiest way to create the needed docker container
+        llm_registry = LLMRegistry(self.config, settings.agent)
+        convo_stats = ConversationStats(self.file_store, sid, user_id)
+        llm_registry.subscribe(convo_stats.register_llm)
```
malhotra5: I'm keeping the subscription fairly simple for now

openhands/llm/llm_registry.py
```diff
@@ -0,0 +1,122 @@
...
+        response = llm.completion(messages=messages)
+        return response.choices[0].message.content.strip()
+
+    def get_llm_from_agent_config(self, service_id: str, agent_config: AgentConfig):
+        llm_config = self.config.get_llm_config_from_agent_config(agent_config)
+        if service_id in self.service_to_llm:
+            if self.service_to_llm[service_id].config != llm_config:
+                # TODO: update llm config internally
+                # Done when agent delegates has different config, we should reuse the existing LLM
+                pass
```
malhotra5, line 77: I realized we may still need to support updating llm configs internally, specifically for when the agent delegates to another agent (we'd want to reuse the parent's llms). Knowing the specific llm between delegates might be too much to track, and perhaps unnecessary as well. 

`get_llm_from_agent_config` is meant to create new llm based on agent config, and delegates may choose a different config but use the same `service_id` of their parent.

`get_llm` is meant to create new llm based on `llm_config`; this is for llms that HAVE to be tracked independently (for instance when used for model routing). 

Would appreciate early feedback on this pattern. I felt these two method help enforce different requirements and thereby have different implications wrt observability

malhotra5:
@enyst working through refining the unit tests but this is ready for another look!

malhotra5:
Not sure why but github bot keeps wiping out my PR description

openhands/llm/llm.py
```diff
@@ -400,8 +405,7 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
                 assert self.config.log_completions_folder is not None
                 log_file = os.path.join(
                     self.config.log_completions_folder,
-                    # use the metric model name (for draft editor)
-                    f'{self.metrics.model_name.replace("/", "__")}-{time.time()}.json',
+                    f'{self.model_name.replace("/", "__")}-{time.time()}.json',
```
```suggestion
                    f'{self.config.model.replace("/", "__")}-{time.time()}.json',
```
Probably? We have to have model at this point, we likely don’t actually need to send a model named “default”, WDYT?

openhands/llm/llm_registry.py
```diff
@@ -0,0 +1,132 @@
...
+            f'[LLM registry {self.registry_id}]: Registering service for {service_id}'
+        )
+
+        # Attempting to switch configs for existing LLM
+        if (
+            service_id in self.service_to_llm
+            and self.service_to_llm[service_id].config != config
+        ):
+            raise ValueError(
+                f'Requesting same service ID {service_id} with different config, use a new service ID'
```
enyst, line 97: Please correct me if I misunderstand: so right now we have a ValueError here, but routing should change this code?

openhands/llm/llm_registry.py
enyst: Really sorry for the delay!

Re: delegates. The current behavior is (was?) that the user can set an llm config e.g. Sonnet for CodeAct in config, and another llm config, e.g. GPT-4 for BrowsingAgent. Then `openhands` would use GPT-4 for Browsing Agent when it’s delegated to it.

Sorry if I’m missing something, if that won’t be the new behavior in this PR, then I think
- we will need to fix it
- we can postpone it for a follow-up (it’s not really in use right now, though Xingyao has a PR on readonly/codeact)

enyst: Re: `get_llm` I like this! I felt that we will need to switch configs somehow

openhands/llm/llm_registry.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def _set_active_llm(self, service_id) -> None:
+        if service_id not in self.service_to_llm:
+            raise ValueError(f'Unrecognized service ID: {service_id}')
+        self.active_agent_llm = self.service_to_llm[service_id]
+
+    def subscribe(self, callback: Callable[[RegistryEvent], None]) -> None:
+        self.subscriber = callback
+
+        # Subscriptions happen after default llm is initialized
```
enyst, line 119: This sounds a bit strange, but I guess it was hard to avoid?

openhands/resolver/interfaces/issue_definitions.py
```diff
@@ -23,7 +23,7 @@ class ServiceContext:
     def __init__(self, strategy: IssueHandlerInterface, llm_config: LLMConfig | None):
         self._strategy = strategy
         if llm_config is not None:
-            self.llm = LLM(llm_config)
+            self.llm = LLM(llm_config, service_id='resolver')
```
enyst, line 26: Is this intended, any reason why we create this directly, instead of being via the registry?

openhands/resolver/send_pull_request.py
```diff
@@ -463,7 +463,7 @@ def update_existing_pull_request(
 
                 # Summarize with LLM if provided
                 if llm_config is not None:
-                    llm = LLM(llm_config)
+                    llm = LLM(llm_config, service_id='resolver')
```
enyst, line 466: same here

openhands/resolver/interfaces/issue_definitions.py
malhotra5: Oh yes forgot to mention - I didn't make the changes to the resolver because the scope of this PR was growing and it was getting harder for me to track changes

I want to put up a follow PR the resolver

openhands/llm/llm_registry.py
malhotra5: Yeah because we can't predict when the subscription is made, but we do always instantiate a default llm once the registry is made

openhands/llm/llm_registry.py
malhotra5: I think routing shouldn't change this method

This method is a guardrail for routing, to ensure that they're not mixing up different llm services

openhands/llm/llm.py
malhotra5: Yep agree

openhands/llm/llm_registry.py
malhotra5: > Re: delegates. The current behavior is (was?) that the user can set an llm config e.g. Sonnet for CodeAct in config, and another llm config, e.g. GPT-4 for BrowsingAgent. Then openhands would use GPT-4 for Browsing Agent when it’s delegated to it.

Yep exactly this won't be the behavior atm ; I saw that you had a PR for switching configs so felt we could implement it as a follow up

I've left a TODO statement so we catch it, lmk if that's ok

openhands/resolver/interfaces/issue_definitions.py
enyst: All good, marking this as solved!

openhands/controller/agent.py
```diff
@@ -38,10 +39,11 @@ class Agent(ABC):
 
     def __init__(
         self,
-        llm: LLM,
         config: AgentConfig,
+        llm_registry: LLMRegistry,
     ):
-        self.llm = llm
+        self.llm = llm_registry.get_llm_from_agent_config('agent', config)
+        self.llm_registry = llm_registry
```
xingyaoww, line 46: Love it :)

This could potentially make multi-LLM management easier

openhands/controller/agent_controller.py
```diff
@@ -1010,37 +1001,20 @@ def _prepare_metrics_for_frontend(self, action: Action) -> None:
...
-        # Set accumulated cost (sum of agent and condenser costs)
-        metrics.accumulated_cost = agent_metrics.accumulated_cost
-        if condenser_metrics:
-            metrics.accumulated_cost += condenser_metrics.accumulated_cost
+        # Create a clean copy with only the fields we want to keep
+        clean_metrics = Metrics()
+        clean_metrics.accumulated_cost = metrics.accumulated_cost
+        clean_metrics._accumulated_token_usage = copy.deepcopy(
+            metrics.accumulated_token_usage
+        )
```
xingyaoww, line 1010: i think we will also need other fields like accumulated input/output tokens, because this is how we visualize them in the frontend UI

openhands/runtime/utils/edit.py
```diff
@@ -138,15 +145,17 @@ def __init__(self, enable_llm_editor: bool, *args: Any, **kwargs: Any) -> None:
         draft_editor_config = self.config.get_llm_config('draft_editor')
 
         # manually set the model name for the draft editor LLM to distinguish token costs
-        llm_metrics = Metrics(model_name='draft_editor:' + draft_editor_config.model)
+        Metrics(model_name='draft_editor:' + draft_editor_config.model)
```
xingyaoww: this is not assigned to any variable, is this intentional?

openhands/server/services/conversation_stats.py
```diff
@@ -0,0 +1,77 @@
...
+        return self.service_to_metrics[service_id]
+
+    def register_llm(self, event: RegistryEvent):
+        # Listen for llm creations and track their metrics
+        llm = event.llm
+        service_id = event.service_id
+
+        if service_id in self.restored_metrics:
+            llm.metrics = self.restored_metrics[service_id].copy()
+            del self.restored_metrics
```
xingyaoww: is this del intentional? if we register multiple llm in a row, would that cause issue (i.e., running self.restored_metrics[service_id] after self.restored_metrics was deleted)

openhands/server/services/conversation_service.py
```diff
@@ -30,29 +30,63 @@
...
+            trigger=conversation_trigger,
+            conversation_id=conversation_id,
+            title=conversation_title,
+            user_id=user_id,
+            selected_repository=selected_repository,
+            selected_branch=selected_branch,
+            git_provider=git_provider,
+        )
+
+        # TODO: initialize and save LLM Registry here
```
xingyaoww: shall we implement this?

xingyaoww (commented):
Still haven't got to most of the test file, but gone through all implementations and they mostly look good to me!

openhands/server/services/conversation_stats.py
malhotra5: Yep the del is intentional 

Consider the scenario where we 

1. get the llm for service_id 'agent'
2. the restored metrics exist, so we restore them
3. run some llm completions 
4. at some point in the future, we get the llm for service_id 'agent' again

If we didn't mark the llm as restored already, then we would overwrite the current accumulated metrics with a snapshot of previous metrics - in other words this restoration should only happens once the first time the llm is instantiated so we delete it right after (just the in-memory copy)

openhands/runtime/utils/edit.py
malhotra5: oh this is not intentional - it can be removed all together actually

malhotra5:
```suggestion
```

openhands/server/services/conversation_service.py
malhotra5: Ah actually we longer need to do this - the llms are automatically restored via `ConversationStats` when an LLM is created in the `LLMRegistry` 

This way different microservices can request an llm without having to worry about restoring/saving llm states

malhotra5:
```suggestion
```

openhands/controller/agent_controller.py
malhotra5: I see this as the definition for `accumulated_token_usage` when its first initialized - I believe that should contain the input/output tokens + caching info as well

```python
self._accumulated_token_usage: TokenUsage = TokenUsage(
            model=model_name,
            prompt_tokens=0,
            completion_tokens=0,
            cache_read_tokens=0,
            cache_write_tokens=0,
            context_window=0,
            response_id='',
        )
```

xingyaoww: ahh you are right!

openhands/llm/llm_registry.py
enyst: Cc: @li-boxuan This is a large PR that Rohit has refactored more than once, and it's increasingly difficult to see all things at once. I think we could maybe break this particular behavior, llm per agent, and restore it in a follow up. WDYT?

xingyaoww:
@malhotra5 lmk when it is ready for another look!

openhands/llm/llm_registry.py
enyst: I'm not fully sure, but I think that if we implement switching configs instead of instances, this may be the place to switch. I'll mark this as solved, as a github management thing, so we can see remaining Conversations in the upper tab 😊

openhands/llm/llm_registry.py
li-boxuan: sounds good

tests/unit/conversation_stats.py
xingyaoww, line 1: nit: maybe this file need to start with test_ so it can be run?

tests/unit/test_llm.py
```diff
@@ -224,48 +253,48 @@ def test_response_latency_tracking(mock_time, mock_litellm_completion):
...
     assert response['choices'][0]['message']['content'] == 'Test response'
 
-    # To make sure the metrics fail gracefully, set the start/end time to go backwards.
-    mock_time.side_effect = [1000.0, 999.0]
-    llm.completion(messages=[{'role': 'user', 'content': 'Hello!'}])
-
-    # There should now be 2 latencies, the last of which has the value clipped to 0
-    assert len(llm.metrics.response_latencies) == 2
-    latency_record = llm.metrics.response_latencies[-1]
-    assert latency_record.latency == 0.0  # Should be lifted to 0 instead of being -1!
```
xingyaoww: Did we intentionally simplify these tests?

tests/unit/test_condenser.py
```diff
@@ -631,48 +673,63 @@ def test_structured_summary_condenser_from_config():
             caching_prompt=True,
         ),
     )
-    condenser = Condenser.from_config(config)
+    condenser = Condenser.from_config(config, mock_llm_registry)
 
     assert isinstance(condenser, StructuredSummaryCondenser)
     assert condenser.llm.config.model == 'gpt-4o'
-    assert condenser.llm.config.api_key.get_secret_value() == 'test_key'
```
xingyaoww, line 638: Any idea why we didn't need to do `.get_secret_value` any more despite having the same definition as `SecretStr`?

tests/unit/test_agent_controller.py
```diff
@@ -61,25 +64,38 @@ def event_loop():
 
 
 @pytest.fixture
-def mock_agent():
-    agent = MagicMock(spec=Agent)
-    agent.llm = MagicMock(spec=LLM)
-    agent.llm.metrics = Metrics()
-    agent.llm.config = OpenHandsConfig().get_llm_config()
+def make_mock_agent():
+    def _make_mock_agent(llm_registry):
```
xingyaoww: Why are we doing this weird nested function 🥹

tests/unit/conversation_stats.py
malhotra5: Thanks for the catch!

tests/unit/test_condenser.py
malhotra5: Ah good catch this is because the registry creates the llm now, but the registry was using a mock llmconfig that wasn't setup with secretstr 

We should be picking up the actual llmconfig instead of the mock one, will fix this

tests/unit/test_llm.py
malhotra5: good catch another case of mocking difficulty for openhands

the llm registry creates the llm, so the mocks need to get attached to that llm object

will fix this

tests/unit/test_agent_controller.py
malhotra5: we need an llm registry + convo stats subscribed to it; this subscription only happens at runtime, so after the subscription is made we can instantiate the agent with the registry

this meant we had to parameterize the agent creation; so the mock here returns a method which will take the registry (after its been subscribed to), and will instantiate the agent with it

I'll see if there's a way to simplify this

xingyaoww: ahh i was just making sure it was not an error -- if we actually need it, we can just leave some comment and let it be

malhotra5: simplified!

malhotra5:
Thanks a bunch for the reviews! Mahmoud has made a release but things are looking somewhat unstable in prod, so I suspect we will be making many patch releases

I will do a QA soon, but still hold off merging until things are stable again (unless this is super urgent for model routing)

openhands/server/services/conversation_stats.py
enyst: OpenHands agent with GPT-5 complained about this too, when it was looking at this file for another PR 😂

> Important small bug to fix in ConversationStats
> In register_llm: after pulling from restored_metrics, it does del self.restored_metrics which deletes the whole dict. This should be del self.restored_metrics[service_id]. Otherwise, later code may hit AttributeError and you lose restored entries for other services.

It makes sense?
`register_llm for 'agent'  # del self.restored_metrics`
`register_llm for 'condenser'  # no restore ?` 

It won't hit AttributeError, but it seems to lose entries?

malhotra5: Oh oops yeah it should del `self.restored_metrics[service_id]` 🤦

I'll write a unit test and fix this

malhotra5:
I will be looking to merge this over this weekend!

xingyaoww:
@enyst sadly i think https://github.com/All-Hands-AI/OpenHands/pull/10427 created a merge conflict nightmare 😨

enyst:
Oh no!! Hmm, we could revert it, and I’ll have the agent check after this merge.  Or I can set it off on this PR directly.
WDYT?

xingyaoww:
@enyst maybe we can ask an agent check to see if it can help rohit resolve the merge conflict in this PR by sending another PR that merge into centralized-metrics?

enyst:
OpenHands-GPT-5 — quick context on the large diff and how to review:

- The 400+ files / ~8k lines come from origin/main since this branch diverged (e.g., test reorg, doc/assets updates); these are mostly moves/renames, not manual edits here.
- Manual conflict resolution by us only touched:
  - openhands/controller/state/state_tracker.py
  - openhands/server/session/session.py
  - openhands/storage/locations.py
  - tests/unit/controller/test_agent_controller.py
  - tests/unit/controller/test_agent_delegation.py
  - tests/unit/server/session/test_agent_session.py
- Key updates: adopt ConversationStats + LLMRegistry (budget sync via combined metrics), Session now uses LLMRegistry + ExperimentManagerImpl, and get_experiment_config_filename restored for existing callers.
- Review tip: in Files changed, filter to openhands/controller|server|storage and the 3 test files above; enable “Hide whitespace changes”.
- Status: linters/format/mypy pass locally after fixes. Helper PR into this branch: #10430. If preferred, I can push the merge commit directly here and close #10430.

enyst:
@openhands This PR has merge conflicts with main branch.

There are two kinds: most tests are in a category where the conflicts are not about semantics, but a very recent commit has moved test files. Find the commit and understand how to fix most test files.

Fix merge conflicts on a separate branch built from this, and make a PR to this PR.

enyst:
(OpenHands-GPT-5 speaking)

High-level summary from the LLM architecture perspective:
- Introduces LLMRegistry as the single entry-point to construct, cache, and distribute LLM instances, keyed by a service_id (e.g., "agent", "condenser", "guardrail").
- Agents, runtimes, condensers, and utilities now receive an LLMRegistry instead of raw LLMs; the base Agent resolves its LLM via llm_registry.get_llm_from_agent_config('agent', config).
- Delegate agents reuse the same registry to share caching, configuration, and unified accounting across a conversation.
- One-off utility calls (e.g., auto-generate conversation title) go through llm_registry.request_extraneous_completion(...), avoiding ad-hoc LLM construction.
- ConversationStats is plumbed into controller/session. API retries are surfaced to the UI and stats via a registry-level retry listener (e.g., Session._notify_on_llm_retry). The previous accumulation of observation.llm_metrics inside AgentController is removed in favor of centralized tracking.

Mermaid – core classes and relationships
```mermaid
classDiagram
  direction LR

  class OpenHandsConfig {
    +get_llm_config(): LLMConfig
    +get_llm_config_from_agent(agentId): LLMConfig
  }

  class LLMConfig {
    model
    api_key
    base_url
    ...
  }

  class LLMRegistry {
    -config: OpenHandsConfig
    -retry_listener: Callable
    -cache: Map~service_id, LLM~
    +get_llm(service_id, config?): LLM
    +get_llm_from_agent_config(service_id, AgentConfig): LLM
    +request_extraneous_completion(service_id, ...): string
  }

  class LLM {
    -service_id: string
    -config: LLMConfig
    -retry_listener: Callable
    +completion(messages, stream)
  }

  class Agent {
    -config: AgentConfig
    -llm_registry: LLMRegistry
    -llm: LLM
  }

  class Condenser { <<abstract>> }
  class LLMSummarizingCondenser { -llm: LLM }

  class Runtime { -llm_registry: LLMRegistry }
  class AgentController { -convo_stats: ConversationStats }

  class ConversationStats {
    +record(tokens, cost, retries)
  }

  class Session {
    +_notify_on_llm_retry(...)
  }

  LLMRegistry --> LLM : creates/caches
  Agent --> LLMRegistry : uses
  Agent --> LLM : resolves ("agent")
  LLMSummarizingCondenser --> LLMRegistry : uses
  Runtime --> LLMRegistry : holds
  AgentController --> ConversationStats : uses
  LLM --> Session : retry_listener()
  Session ..> LLMRegistry : sets registry.retry_listener
  Session --> ConversationStats : update on retries/tokens
```

Mermaid – flow when a component asks to create/use an LLM
```mermaid
sequenceDiagram
  participant Service as Component/Service (service_id)
  participant Registry as LLMRegistry
  participant Config as OpenHandsConfig
  participant LLM as LLM
  participant Session as Session
  participant Stats as ConversationStats

  Service->>Registry: get_llm(service_id, [config])
  alt config omitted
    Registry->>Config: resolve LLMConfig (agent/global)
  end
  Registry->>LLM: construct if needed (attach retry_listener)
  Registry-->>Service: return LLM instance

  Service->>LLM: completion(messages, stream?)
  loop on transient errors
    LLM-->>Session: retry_listener(retry_ctx)
    Session->>Stats: record retry event
    Session->>UI: queue_status_message(type="llm-retry", details)
  end
  LLM-->>Service: response
  Note over Registry,Stats: Centralized accounting (tokens/cost) per service_id
```

What this means for component authors:
- Do not instantiate LLMs directly. Always request them from LLMRegistry using a clear service_id: llm = llm_registry.get_llm('your-service-id', optional_overrides).
- Agents: inject llm_registry; the base Agent already resolves self.llm via get_llm_from_agent_config('agent', config). Delegates should reuse the same registry.
- Condensers: use Condenser.from_config(cfg, llm_registry) so LLMSummarizingCondenser gets a properly scoped LLM.
- Utilities/one-offs: use llm_registry.request_extraneous_completion(...) to keep configuration, retries, and metrics unified.

Net effect: LLMRegistry becomes the single source of truth for LLM lifecycle, configuration, retry wiring, and usage attribution, while ConversationStats provides a unified, per-conversation surface for status and cost/usage.

enyst:
(OpenHands-GPT-5 speaking)

Thanks for the feedback — updated diagrams below.

Changes:
- Removed the abstract Condenser from the class diagram to keep focus on how components use LLMs via LLMRegistry
- Added a new flow diagram showing ConversationStats lifecycle

Revised core classes (no abstract Condenser)
```mermaid
classDiagram
  direction LR

  class OpenHandsConfig {
    +get_llm_config(): LLMConfig
    +get_llm_config_from_agent(agentId): LLMConfig
  }

  class LLMConfig {
    model
    api_key
    base_url
    ...
  }

  class LLMRegistry {
    -config: OpenHandsConfig
    -retry_listener: Callable
    -cache: Dict[str, LLM]
    +get_llm(service_id, config?): LLM
    +get_llm_from_agent_config(service_id, AgentConfig): LLM
    +request_extraneous_completion(service_id, ...): string
  }

  class LLM {
    -service_id: string
    -config: LLMConfig
    -retry_listener: Callable
    +completion(messages, stream)
  }

  class Agent {
    -config: AgentConfig
    -llm_registry: LLMRegistry
    -llm: LLM
  }

  class LLMSummarizingCondenser { -llm: LLM }
  class Runtime { -llm_registry: LLMRegistry }
  class AgentController { -convo_stats: ConversationStats }
  class ConversationStats { +record(tokens, cost, retries) }
  class Session { +_notify_on_llm_retry(...) }

  LLMRegistry --> LLM : creates/caches
  Agent --> LLMRegistry : uses
  Agent --> LLM : resolves ("agent")
  LLMSummarizingCondenser --> LLMRegistry : uses
  Runtime --> LLMRegistry : holds
  AgentController --> ConversationStats : uses
  LLM --> Session : retry_listener()
  Session ..> LLMRegistry : sets registry.retry_listener
  Session --> ConversationStats : update on retries/tokens
```

ConversationStats lifecycle (creation, wiring, usage, finalization)
```mermaid
flowchart TD
  A[Session/CLI start] --> B[create_registry_and_convo_stats(config, sid, user)]
  B -->|returns| C[LLMRegistry]
  B -->|returns| D[ConversationStats]
  C --> E[Agent constructed]
  C --> G[Runtime constructed]
  D --> H[AgentController constructed with convo_stats]
  I[Session] --> J[Set LLMRegistry.retry_listener = Session._notify_on_llm_retry]

  subgraph DuringConversation
    K[Agent/Runtime request LLM] --> C
    C --> L[LLM instance (service_id-scoped)]
    L --> M[completion(...)]
    M -->|success| N[ConversationStats: add tokens/cost]
    M -->|retry| I
    I --> D[ConversationStats: record retry + notify UI]
  end

  O[End/teardown] --> P[ConversationStats: finalize/persist via FileStore]
  D --> P
```

If you want, I can add more detail (e.g., per-service_id cost attribution or specific status events) — just say the word.

enyst:
(OpenHands-GPT-5 speaking)

Thanks for the feedback! I’ve removed the abstract Condenser from the diagram and added a new diagram to illustrate ConversationStats’s lifecycle.

Mermaid – core classes (revised, no abstract Condenser)
```mermaid
classDiagram
  direction LR

  class OpenHandsConfig {
    +get_llm_config(): LLMConfig
    +get_llm_config_from_agent(agentId): LLMConfig
  }

  class LLMConfig {
    model
    api_key
    base_url
    ...
  }

  class LLMRegistry {
    -config: OpenHandsConfig
    -retry_listener: Callable
    -cache: Map~service_id, LLM~
    +get_llm(service_id, config?): LLM
    +get_llm_from_agent_config(service_id, AgentConfig): LLM
    +request_extraneous_completion(service_id, ...): string
  }

  class LLM {
    -service_id: string
    -config: LLMConfig
    -retry_listener: Callable
    +completion(messages, stream)
  }

  class Agent {
    -config: AgentConfig
    -llm_registry: LLMRegistry
    -llm: LLM
  }

  class Runtime { -llm_registry: LLMRegistry }
  class AgentController { -convo_stats: ConversationStats }
  class ConversationStats { +record(tokens, cost, retries) }
  class Session { +_notify_on_llm_retry(...) }

  LLMRegistry --> LLM : creates/caches
  Agent --> LLMRegistry : uses
  Agent --> LLM : resolves ("agent")
  Runtime --> LLMRegistry : holds
  AgentController --> ConversationStats : uses
  LLM --> Session : retry_listener()
  Session ..> LLMRegistry : sets registry.retry_listener
  Session --> ConversationStats : update on retries/tokens
```

Mermaid – ConversationStats lifecycle
```mermaid
sequenceDiagram
  participant CLI as CLI/Server bootstrap
  participant Util as create_registry_and_convo_stats
  participant Registry as LLMRegistry
  participant Stats as ConversationStats
  participant Ctrl as AgentController
  participant Sess as Session
  participant Comp as Component (Agent/Condenser/Utility)
  participant L as LLM
  participant UI as UI

  CLI->>Util: initialize(config, sid, user)
  Util-->>CLI: { llm_registry, convo_stats, config }
  CLI->>Ctrl: create_controller(..., convo_stats)
  CLI->>Sess: create Session(..., llm_registry, convo_stats)
  Sess->>Registry: set retry_listener = Sess._notify_on_llm_retry

  note over Comp: When any component needs an LLM
  Comp->>Registry: get_llm(service_id, [overrides])
  Registry-->>Comp: LLM instance
  Comp->>L: completion(messages, stream?)

  alt transient error
    L-->>Sess: retry_listener(retry_ctx)
    Sess->>Stats: recordRetry(retry_ctx)
    Sess->>UI: queue_status_message(type="llm-retry")
  end

  L-->>Comp: response
  Registry->>Stats: recordUsage(service_id, tokens/cost)

  note over Stats: Persist per-conversation aggregates via file_store
```

Design note on ConversationStats location
- Current wiring: ConversationStats is passed to AgentController and Session; the LLMRegistry’s retry_listener routes through Session to update UI and stats consistently.
- Alternative: Attach ConversationStats to a Conversation domain object (e.g., Conversation/ConversationManager).
  - Pros: Stronger domain cohesion; lifecycle naturally bound to the conversation; simplifies access for any component that has the conversation.
  - Cons: Requires broader refactor to thread the conversation object everywhere; this PR focuses on centralizing LLM usage first.
- Suggestion: The current approach is pragmatic. A follow-up could relocate ConversationStats under the Conversation aggregate while keeping LLMRegistry as the single source for LLM events/usage attribution.

enyst:
The last bit in the agent's response was because I asked it a question on the fly 😁

> Note that I'm a bit surprised that ConversationStats is in the controller, why not in Conversation object or attached to it?

The agent decided it was a pragmatic choice for now. 👌

@malhotra5 I'm guessing it needs to be in the controller also because of CLI/headless; that's fine! Just to note, for the future, I'm actually looking at making the Conversation available to CLI, I'd love your input there when I fix it a bit more:
- https://github.com/All-Hands-AI/OpenHands/pull/10424

malhotra5:
Sounds good!

PR #9596, opened 2025-07-07
chore(ui): Fix late redirects in settings page

amanape:

> When refreshing the browser on the Credits or API Keys pages under Settings, the interface returns to the User Settings page.

https://openhands-ai.slack.com/archives/C06QT0AGY4W/p1751913412181269

---
Handle navigation logic before page is rendered

---

### DIFF
--- a/frontend/src/routes/settings.tsx
+++ b/frontend/src/routes/settings.tsx
@@ -40,19 +40,22 @@ export const clientLoader = async ({ request }: Route.ClientLoaderArgs) => {
   let config = queryClient.getQueryData<GetConfigResponse>(["config"]);
   if (!config) {
     config = await OpenHands.getConfig();
+    queryClient.setQueryData<GetConfigResponse>(["config"], config);
   }
 
   const isSaas = config?.APP_MODE === "saas";
 
   if (isSaas && pathname === "/settings") {
     // no llm settings in saas mode, so redirect to user settings
-    redirect("/settings/user");
-  } else if (SAAS_ONLY_PATHS.includes(pathname)) {
-    // if in OSS mode, do not allow access to saas-only paths
-    redirect("/settings");
+    return redirect("/settings/user");
   }
 
-  queryClient.setQueryData<GetConfigResponse>(["config"], config);
+  if (!isSaas && SAAS_ONLY_PATHS.includes(pathname)) {
+    // if in OSS mode, do not allow access to saas-only paths
+    return redirect("/settings");
+  }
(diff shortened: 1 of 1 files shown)


enyst (approved):
We should fix this, thank you.

It looks like node 20 is failing. By the way, why do we want version 20?

amanape:
I think the error message is related to the unlocalized script test. I’m guessing it references version 20 because it’s the minimum engine requirement we defined in the `package.json` https://github.com/All-Hands-AI/OpenHands/blob/7acee9e5da321cc79b24aa6d5110e3366a91e403/frontend/package.json#L7 

I’ve submitted a PR that should fix this (and a separate problem) in this PR: https://github.com/All-Hands-AI/OpenHands/pull/9616

frontend/src/routes/settings.tsx
```diff
@@ -1,55 +1,70 @@
...
 
-function SettingsScreen() {
-  const { t } = useTranslation();
-  const navigate = useNavigate();
-  const { pathname } = useLocation();
-  const { data: config } = useConfig();
+const SAAS_ONLY_PATHS = [
+  "/settings/user",
+  "/settings/billing",
+  "/settings/credits",
```
kentyman23, line 15: The _Credits_ tab actually goes to `/settings/billing`. There is no `/settings/credits` path:

image

You could just remove `/settings/credits`, but instead I might suggest keeping it and switching away from `/settings/billing` so that the UI element matches the path. Thoughts?

enyst:
> I think the error message is related to the unlocalized script test. I’m guessing it references version 20 because it’s the minimum engine requirement we defined in the `package.json`

Yes, but I mean, why do we want to support 20? 22 is LTS, right? Could we consider updating the minimum requirement to 22? Guessing it would simplify some things… not in the least, we could run CI on a single version, 22.

Not an issue for this PR, of course. Just a question. I encountered it when working on the vscode extension, I tried to keep compatibility with FE on any library versions/dependencies, and it surprised me to see we seem to fully support a relatively old version. Please correct me if wrong.

amanape:
Yes we should definitely update to LTS

frontend/src/routes/settings.tsx
amanape: Thanks for the suggestion. The `/settings/credits` route was accidentally left in when we created this page.

In the future, we may add more billing-related content, so `/billing` might be a more appropriate route name instead of `/credits`.

PR #9599, opened 2025-07-07
chore - Add nix dev setup with flake

raymyers:

Add nix option for development setup.

---
This provides a "flake" file so that anyone using the Nix package manager can enter a shell ready to develop OpenHands with:
```sh
nix develop
```

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f804a50-nikolaik   --name openhands-app-f804a50   docker.all-hands.dev/all-hands-ai/openhands:f804a50
```

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,176 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1731533236,
+        "narHash": "sha256-l0KFg5HjrsfsO/JpG+r7fRrqm12kzFHyUHqHCVpMMbI=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "11707dc2f618dd54ca8739b309ec4fc024de578b",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "flake-utils_2": {
+      "inputs": {
+        "systems": "systems_2"
+      },
+      "locked": {
+        "lastModified": 1726560853,
+        "narHash": "sha256-X6rJYSESBVr3hBoH0WbKE5KvhPU5bloyZ2L4K60/fPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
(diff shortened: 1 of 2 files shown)


patonw:
I've been using a non-flake nix:

```nix
{
  sources ? import ./nix/sources.nix {},
  pkgs ? import sources.nixpkgs {},
  lib ? pkgs.lib,
}:
let
  inherit (pkgs) mkShellNoCC;
  inherit (pkgs.lib) makeLibraryPath;
  py = pkgs.python312;
in
{
  devShell = with pkgs; mkShellNoCC {
    venvDir = "./.venv";

    buildInputs = with pkgs; [
      niv
      poetry
      uv
      nodejs_22
      (py.withPackages (ps: with ps; [
        pip
        pytest
        ruff
        black
        cleo
      ]))
      py.pkgs.venvShellHook
    ];

    LD_LIBRARY_PATH = lib.makeLibraryPath [stdenv.cc.cc.lib ffmpeg];
    # postVenvCreation = ''
    #   unset SOURCE_DATE_EPOCH
    #   pip install -r requirements.txt
    # '';

    postShellHook = ''
      export SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.48-nikolaik
      # export SANDBOX_VOLUMES="$(pwd):/workspace:rw"
      export SANDBOX_USER_ID=$(id -u)
      if [[ -f .env ]]; then
        set -a
        source .env
        set +a
      fi
    '';
  };
}
```

basnijholt:
Looked for exactly this! What is the reason that it’s still in draft? Does it work for you already?

raymyers:
> Looked for exactly this! What is the reason that it’s still in draft? Does it work for you already?

@basnijholt Yeah it's working for me on OSX. It doesn't include installing Docker which is often done outside Nix I think.

I suppose what's left would be documenting it, testing on other platforms, and considering if we should do it without flake as @patonw suggests.

The interest lets me know this is worth taking forward though, thanks.

PR #9612, opened 2025-07-08
refactor(frontend): Set the max height for the suggested tasks section.

### LINKED ISSUE #9611: Set the max height for the suggested tasks section.

### **Description**

When there are many suggested tasks, the **parent container's vertical scrollbar** becomes visible. This negatively impacts the user experience, as the expected behavior is to scroll **only** within the Suggested Tasks section—not the entire parent layout.

---

### ✅ Expected Behavior
- **Only** the `Suggested Tasks` section should display a vertical scrollbar when there is overflow.
- The **parent container** should remain static and not show a scrollbar.

---

### ❌ Current Behavior
- The **parent vertical scrollbar is shown** when the number of suggested tasks exceeds the visible height.
- This results in a less polished and potentially confusing UX.

---

### 🔁 Steps to Reproduce
1. Add enough suggested tasks to exceed the visible area of the section.
2. Observe that the parent container shows a vertical scrollbar.
3. Notice that scrolling affects the entire layout rather than just the task list.

---

### 📹 Additional Context
A video demonstrating the issue is available below:  

https://github.com/user-attachments/assets/7c807fa9-23c0-49dd-ab5e-8bec4e0e0e99

- Ensure that only the `Suggested Tasks` section has scrollable overflow.
- Prevent the parent container from scrolling by adjusting layout styles.

**Additional context**

- N/A.

hieptl:

### **Description**

When there are many suggested tasks, the **parent container's vertical scrollbar** becomes visible. This negatively impacts the user experience, as the expected behavior is to scroll **only** within the Suggested Tasks section—not the entire parent layout.

---

### ✅ Expected Behavior
- **Only** the `Suggested Tasks` section should display a vertical scrollbar when there is overflow.
- The **parent container** should remain static and not show a scrollbar.

---

### ❌ Current Behavior
- The **parent vertical scrollbar is shown** when the number of suggested tasks exceeds the visible height.
- This results in a less polished and potentially confusing UX.

---

### 🔁 Steps to Reproduce
1. Add enough suggested tasks to exceed the visible area of the section.
2. Observe that the parent container shows a vertical scrollbar.
3. Notice that scrolling affects the entire layout rather than just the task list.

---

### 📹 Additional Context
A video demonstrating the issue is available below:  

https://github.com/user-attachments/assets/7c807fa9-23c0-49dd-ab5e-8bec4e0e0e99

---

- Ensure that only the `Suggested Tasks` section has scrollable overflow.
- Prevent the parent container from scrolling by adjusting layout styles.

A video demonstrating the output is available below: 

https://github.com/user-attachments/assets/5c7f8174-7fc0-47a5-a847-8074cd4c57e8

---

Resolves #9611


Abubakar-01:
✅ Verified the PR – The issue with the parent container showing a scrollbar due to overflow in the Suggested Tasks section has been resolved. The vertical scroll now appears only within the Suggested Tasks area as expected, resulting in a cleaner and more consistent user experience. Looks good to me!

Before  :x: 
[Before.webm](https://github.com/user-attachments/assets/b9dde969-a14d-4d7b-a8cc-21aada84a5a5)

After ✅
[after.webm](https://github.com/user-attachments/assets/b496d0f0-57b7-45a7-925a-aeb7d339e0e4)

amanape:
Sorry for the delay, there seems to be some cutoff at the bottom of the screen due to the fixed height. It’ll look out of place on larger screens

hieptl:
> Sorry for the delay, there seems to be some cutoff at the bottom of the screen due to the fixed height. It’ll look out of place on larger screens

Hi @amanape I couldn’t find a CSS-only solution to achieve this. I considered using JavaScript to calculate the height dynamically, but I’m not sure if that’s the best approach. I’d really appreciate your feedback or suggestions. Thank you!

amanape:
Compared to the previous behaviour, this feels like a style regression since it fixes the height and won’t be consistent for all screen heights.

I don’t think I’ll approve this PR, sorry 😕

Feel free to open another one if you manage to make the height fit the screen height!

PR #9621, opened 2025-07-08
fix: prompt caching and function calling for Bedrock model IDs

Tetramputechture:

Fixed Function Calling and Prompt Caching not being applied when using Claude models with AWS Bedrock as a provider

---

Previously, when using Claude models through AWS Bedrock (with model IDs like bedrock/anthropic.claude-3-5-sonnet-20241022-v2:0), we would not recognize these models as supporting function calling or prompt caching. This PR should ensure Bedrock-formatted Claude models are properly detected and will automatically enable function calling and prompt caching capabilities, providing the same experience as using Claude models directly through Anthropic.

---

I couldn't find a relevant issue for this.

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -107,6 +107,86 @@ MODELS_WITHOUT_STOP_WORDS = [
 ]
 
 
+def _extract_bedrock_model_name(model_name: str) -> str:
+    """Extract the actual model name from a Bedrock-formatted model ID.
+
+    Bedrock model IDs have the format: 'provider.model-name-version:revision'
+    For example: 'anthropic.claude-3-5-sonnet-20241022-v2:0'
+
+    Args:
+        model_name: The Bedrock-formatted model ID
+
+    Returns:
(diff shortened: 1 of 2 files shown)


openhands/llm/llm.py
```diff
@@ -107,6 +107,86 @@
...
+    model_name: str,
+    supported_models: list[str],
+    allow_substring_match: bool = False
+) -> bool:
+    """Check if a model is supported by comparing against a list of supported models.
+
+    This function handles various model naming formats including:
+    - Direct model names (e.g., 'claude-3-5-sonnet-20241022')
+    - Provider-prefixed names (e.g., 'anthropic/claude-3-5-sonnet-20241022')
+    - Bedrock-formatted IDs (e.g., 'anthropic.claude-3-5-sonnet-20241022-v2:0')
```
enyst, line 154: You're right to look at these, but I wonder, couldn't we simplify this by checking in all names for a few strings, e.g. "claude-3-5-sonnet", "claude-4-sonnet" / haiku / opus?

PR #9646, opened 2025-07-10
Added stream execution mode for CmdRunAction

mindflow-cn:

---
- I added the execute_stream_bash tool, which enables Cmd to support stream-based execution.

- Introduced CmdRunStreamAction and CmdOutputStreamObservation to handle the tool's execution actions and its streamed output. If the command finishes normally, it still falls back to using CmdOutputObservation without affecting other logic.

- This tool is mutually exclusive with execute_bash; only one of them can be used at a time.

- I modified bash.py to add the execute_stream method, which supports real-time access to standard output during execution.

- Added the execute_cmd_action_stream SSE interface in action_execution_server.py to return streamed data during execution.

- Also added the stop_cmd_action_stream interface in action_execution_server.py, allowing the currently running streamed command to be stopped at any time.

- Added the enable_stream_cmd configuration option in config.toml.template to enable or disable this feature.

---


xingyaoww (commented):
Can you elaborate what problem this PR will solve, and why we need this change?
Seems like these are things that can break easily :(
And we will probably need to run a mini swebench eval to make sure this doesn't degrade agent performance

mindflow-cn:
> Can you elaborate what problem this PR will solve, and why we need this change? Seems like these are things that can break easily :( And we will probably need to run a mini swebench eval to make sure this doesn't degrade agent performance

It is true that this function is not very helpful for conventional software engineering, but I am using OpenHands to write machine learning and deep learning programs. Machine learning and deep learning programs often require a long time of training, and such tasks cannot set a timeout in advance. It is necessary to judge through the execution log of the process (such as: loss, accuracy, recall rate, etc.)

mindflow-cn:
> > Can you elaborate what problem this PR will solve, and why we need this change? Seems like these are things that can break easily :( And we will probably need to run a mini swebench eval to make sure this doesn't degrade agent performance
> 
> It is true that this function is not very helpful for conventional software engineering, but I am using OpenHands to write machine learning and deep learning programs. Machine learning and deep learning programs often require a long time of training, and such tasks cannot set a timeout in advance. It is necessary to judge through the execution log of the process (such as: loss, accuracy, recall rate, etc.)


So, I added a separate configuration method to turn on this feature when needed

mindflow-cn:
I think this configuration will be of great help to openhands to develop AI programs

Abubakar-01:
@mindflow-cn Could you please guide me on how to enable streaming? According to the instructions, I set enable_cmd = false and enable_stream_cmd = true, and then tried running the following command:  

`for i in {1..5}; do echo "Step $i - $(date)"; sleep 2; done`

However, the output still appears all at once after the operation completes, rather than streaming.

mindflow-cn:
> @mindflow-cn Could you please guide me on how to enable streaming? According to the instructions, I set enable_cmd = false and enable_stream_cmd = true, and then tried running the following command:
> 
> `for i in {1..5}; do echo "Step $i - $(date)"; sleep 2; done`
> 
> However, the output still appears all at once after the operation completes, rather than streaming.
image
To get the content of the execution, you can subscribe to the event_stream message to get it.

mindflow-cn:
> @mindflow-cn Could you please guide me on how to enable streaming? According to the instructions, I set enable_cmd = false and enable_stream_cmd = true, and then tried running the following command:
> 
> `for i in {1..5}; do echo "Step $i - $(date)"; sleep 2; done`
> 
> However, the output still appears all at once after the operation completes, rather than streaming.

This is to ensure that the behavior logic of CodeActAgent remains unchanged. Therefore, to obtain the data in the middle of command execution, you need to use event_stream

mamoodi:
What's the status of this PR? Are we waiting for additional reviews or are there changes needed?

PR #9661, opened 2025-07-10
Fixes or vscode token / url not being present

tofarr:

Fixes VSCode integration issues where the VSCode server token and URL were not properly available in certain runtime scenarios, particularly when using warm servers or nested conversation managers. This ensures that users can reliably access the integrated VSCode editor functionality within OpenHands.

---

This PR addresses several issues with VSCode plugin initialization and configuration:

1. **VSCode Plugin Missing in Warm Servers**: Added logic to ensure the VSCode plugin is properly included when initializing warm servers in non-headless mode. The `setup` method now accepts a `headless_mode` parameter to conditionally add the VSCode requirement.

2. **Race Condition Fix**: Replaced asynchronous `_create_warm_server_in_background` with synchronous `_create_warm_server` to prevent race conditions that could cause double server starts or missing plugin configurations.

3. **Workspace Configuration**: Added `WORKSPACE_BASE` environment variable in the Docker nested conversation manager to ensure proper workspace path configuration.

4. **Consistent Plugin Loading**: Ensures VSCode plugin is consistently loaded across different runtime initialization paths by copying the logic from the main Runtime class to the warm server setup.

The key design decision was to add the `headless_mode` parameter to the runtime setup methods rather than trying to detect the mode automatically, providing explicit control over when VSCode integration should be enabled.

---

This addresses VSCode integration reliability issues in OpenHands runtime environments.

### DIFF
--- a/openhands/runtime/impl/local/local_runtime.py
+++ b/openhands/runtime/impl/local/local_runtime.py
@@ -140,11 +140,10 @@ class LocalRuntime(ActionExecutionClient):
         env_vars: dict[str, str] | None = None,
         status_callback: Callable[[str, RuntimeStatus, str], None] | None = None,
         attach_to_existing: bool = False,
-        headless_mode: bool = False,
+        headless_mode: bool = True,
         user_id: str | None = None,
         git_provider_tokens: PROVIDER_TOKEN_TYPE | None = None,
     ) -> None:
-        logger.info(f'TRACE:LocalRuntime:init:1:{plugins}')
         self.is_windows = sys.platform == 'win32'
         if self.is_windows:
             logger.warning(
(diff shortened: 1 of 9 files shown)


openhands/runtime/impl/local/local_runtime.py
```diff
@@ -390,8 +391,13 @@ def setup(cls, config: OpenHandsConfig):
...
         if initial_num_warm_servers > 0 and len(_WARM_SERVERS) == 0:
             plugins = _get_plugins(config)
+
+            # Copy the logic from Runtime where we add a VSCodePlugin on init if missing
+            if not headless_mode:
+                plugins.append(VSCodeRequirement())
+
             for _ in range(initial_num_warm_servers):
-                _create_warm_server_in_background(config, plugins)
+                _create_warm_server(config, plugins)
```
tofarr, line 400: Moving this out of the background helped prevent a race condition where an additional warm server was started. (Which masked the fact that the initial warm servers did not have the appropriate plugins)

openhands/server/conversation_manager/docker_nested_conversation_manager.py
```diff
@@ -498,6 +498,7 @@ async def _create_runtime(
         env_vars['SESSION_API_KEY'] = self._get_session_api_key_for_conversation(sid)
         # We need to be able to specify the nested conversation id within the nested runtime
         env_vars['ALLOW_SET_CONVERSATION_ID'] = '1'
+        env_vars['WORKSPACE_BASE'] = '/workspace'
```
tofarr, line 501: Adding this back in means that the LocalRuntime uses the correct workspace directory

xingyaoww (approved):

PR #9663, opened 2025-07-11
Fix(backend): correctly forward AWS Bedrock aws_access_key_id / aws_secret_access_key / aws_region_name to litellm

littlebai3618:

Small bug-fix—credentials in config were not reaching litellm; now injected before each Bedrock call and verified locally.
---

---


enyst (approved):
Thank you!

PR #9666, opened 2025-07-11
Fix issue #9655: [Bug]: CodeActAgent is incompatible with xAI Grok-4 due to hardcoded stop parameter

### LINKED ISSUE #9655: [Bug]: CodeActAgent is incompatible with xAI Grok-4 due to hardcoded stop parameter

litellm.BadRequestError: XaiException - Error code: 400 - {'code': 'Client specified an invalid argument', 'error': 'Argument not supported on this model: stop'}```

#### **Root Cause Analysis**

The error message explicitly states that the `stop` parameter is not supported by the target model. The xAI API documentation confirms this for Grok-4.

The issue stems from the `CodeActAgent`'s implementation, which appears to hardcode the addition of a `stop` sequence (e.g., `["</execute_bash>"]`) to every LLM API call. While this is a valid parameter for many models (like those from OpenAI and Anthropic), it causes a fatal API error when used with Grok-4, making the model unusable.

#### **Steps to Reproduce**

1.  Navigate to the OpenHands **Settings** page.
2.  Enable **Advanced** options and configure the LLM as follows:
    *   **Custom Model:** `xai/grok-4-0709`
    *   **Base URL:** `https://api.x.ai/v1`
    *   **API Key:** `[A valid xAI API key]`
3.  Save the settings and start a new agent session.
4.  Enter any initial prompt (e.g., "hello world").
5.  Observe that the agent immediately transitions to an `ERROR` state and the logs show the `BadRequestError` detailed above.

#### **Suggested Solution**

To resolve this and enable support for Grok-4, the agent code should be modified to conditionally add the `stop` parameter. It should only be included in the API call for models that are known to support it.

A simple and effective fix would be to check the model identifier before adding the parameter.

**Example (pseudo-code for `openhands/agenthub/codeact_agent/codeact_agent.py`):**

```python
# In the agent's step method, where the LLM call is prepared...
params = {
    "messages": messages,
    "temperature": 0.0,
    # other parameters...
}

# Conditionally add the 'stop' parameter
# self.llm.model should contain the model name, e.g., 'xai/grok-4-0709'
if "grok" not in self.llm.model:
    params["stop"] = self.stop_sequences # Or the hardcoded list

# Make the LLM call
response = self.llm.completion(**params)

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Model Name

_No response_

### Operating System

None

_No response_

### PR DESCRIPTION
**Pull request by openhands-agent**
This pull request fixes #9655.

The issue has been successfully resolved based on the changes made. The core problem was that Grok-4 doesn't support the 'stop' parameter, causing API errors when OpenHands tried to use it. The fix directly addresses this by:

1. Adding 'xai/grok-4-0709' to the MODELS_WITHOUT_STOP_WORDS list, which prevents the stop parameter from being included in API calls for this model
2. Adding comprehensive test coverage that verifies the stop parameter is correctly handled - included for supported models and excluded for Grok-4

The changes are minimal but targeted, modifying exactly what needs to be changed to fix the issue. The implementation matches the suggested solution from the issue description, which proposed conditionally adding the stop parameter based on model support. The added test coverage ensures the fix works as intended by explicitly verifying the presence/absence of the stop parameter in API calls.

Since the error was specifically caused by the invalid 'stop' argument being sent to Grok-4's API, and these changes prevent that parameter from being sent, this should fully resolve the reported error and allow successful API calls to Grok-4.

Automatic fix generated by [OpenHands](https://github.com/All-Hands-AI/OpenHands/) 🙌

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -104,6 +104,7 @@ MODELS_WITHOUT_STOP_WORDS = [
     'o1-preview',
     'o1',
     'o1-2024-12-17',
+    'xai/grok-4-0709',  # Grok-4 doesn't support stop parameter
 ]
 
 
--- a/tests/unit/test_llm.py
+++ b/tests/unit/test_llm.py
@@ -264,6 +264,39 @@ def test_llm_init_with_openrouter_model(mock_get_model_info, default_config):
     mock_get_model_info.assert_called_once_with('openrouter:gpt-4o-mini')
 
 
+@patch('openhands.llm.llm.litellm_completion')
+def test_stop_parameter_handling(mock_litellm_completion, default_config):
+    """Test that stop parameter is only added for supported models."""
+    from litellm.types.utils import ModelResponse
+
+    mock_response = ModelResponse(
+        id='test-id',
+        choices=[{'message': {'content': 'Test response'}}],
+        model='test-model',
+    )
+    mock_litellm_completion.return_value = mock_response
+
+    # Test with a model that supports stop parameter
+    default_config.model = 'custom-model'  # Use a model not in FUNCTION_CALLING_SUPPORTED_MODELS
+    llm = LLM(default_config)
+    llm.completion(
+        messages=[{'role': 'user', 'content': 'Hello!'}],
+        tools=[{'type': 'function', 'function': {'name': 'test', 'description': 'test'}}],
+    )
+    # Verify stop parameter was included
+    assert 'stop' in mock_litellm_completion.call_args[1]
+
+    # Test with Grok-4 model that doesn't support stop parameter
+    default_config.model = 'xai/grok-4-0709'
+    llm = LLM(default_config)
+    llm.completion(
+        messages=[{'role': 'user', 'content': 'Hello!'}],
+        tools=[{'type': 'function', 'function': {'name': 'test', 'description': 'test'}}],
+    )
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: openhands/llm/llm.py [hunk 1]
```diff
@@ -104,6 +104,7 @@
     'o1-preview',
     'o1',
     'o1-2024-12-17',
+    'xai/grok-4-0709',  # Grok-4 doesn't support stop parameter
```
```suggestion
    'xai/grok-4-0709',
```

mamoodi (approved):

Iamsuperman19871987:
@mamoodi  does not look like this has been fixed in the command line version

enyst:
Do you use model name `xai/grok-4-0709`?

Iamsuperman19871987:
There is a new one on openrouter

Iamsuperman19871987:
Grok Code Fast 1

We're thrilled to introduce grok-code-fast-1, a speedy and economical
reasoning model that excels at agentic coding.

enyst:
@openhands Read all these comments, search and find the exact name of the model on OpenRouter to confirm the name we need, and add the model to the right place for a reasoning model which doesn’t support `stop` parameter.

Make a PR here. Then, clone agent-sdk from this org, read its readme and understand it’s V1 of OpenHands in progress: you need to migrate the same changes there as appropriate. Make a PR there.

PR #9681, opened 2025-07-13
Update llm constants to match on unpinned `claude-sonnet-4`

hamiltop:
We were matching only on the full pinned version but some providers (like anthropic) do not offer the pinned model name.

---
Match both the pinned and unpinned claude sonnet 4 model name.

---

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -59,6 +59,7 @@ CACHE_PROMPT_SUPPORTED_MODELS = [
     'claude-3-haiku-20240307',
     'claude-3-opus-20240229',
     'claude-sonnet-4-20250514',
+    'claude-sonnet-4',
     'claude-opus-4-20250514',
 ]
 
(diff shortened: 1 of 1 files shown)


hamiltop:
That's probably true.

It also looks like we need to update REASONING_EFFORT_SUPPORTED_MODELS to include some of the anthropic models. LiteLLM takes care of that support for us: https://github.com/BerriAI/litellm/blob/d15664c8e6dd30dfc2f25ad6884ba5dcf0e995a8/docs/my-website/docs/providers/anthropic.md#L1206

Also, gemini does support prompt caching in litellm (though there is a current bug https://github.com/BerriAI/litellm/issues/12212) so we could in theory add gemini to that allow-list once that bug is addressd.

In general, I would very much prefer to be able to enable a feature and it just not work/throw an exception, rather than these allowlists for various functionality. Putting gates in the openhands code when litellm already has those gates feels duplicitous.

enyst:
Good points! One tweak: Gemini's prompt caching is now automatic, while our little list is intended for Anthopic-specific behavior (we need to manually set the cache marker on the messages that need cached). So I don't think it needs added to manual handling.

You are absolutely correct though, we shouldn't, ideally, do any of this. Unfortunately they're all fixing some litellm issues before they're fixed in litellm, basically, or check for support before the litellm check for support functions were implemented. The idea is to clean them out, when we can. We just need to make sure that we really can ... clean them out 😅

PR #9707, opened 2025-07-15
ci: add GitHub Action to post welcome message on good first issues

llamantino:

---

Add a workflow that posts a welcome message when an issue is labeled as good first issue.

---

### DIFF
--- a/.github/workflows/welcome-good-first-issue.yml
+++ b/.github/workflows/welcome-good-first-issue.yml
@@ -1,0 +1,50 @@
+name: Welcome Good First Issue
+
+on:
+  issues:
+    types: [labeled]
+
+permissions:
+  issues: write
+
+jobs:
+  comment-on-good-first-issue:
+    if: github.event.label.name == 'good first issue'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if welcome comment already exists
+        id: check_comment
+        uses: actions/github-script@v7
+        with:
+          result-encoding: string
+          script: |
+            const issueNumber = context.issue.number;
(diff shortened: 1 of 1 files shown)


mamoodi:
Thanks for this! Let me do a quick check with folks on this. But overall LGTM.

.github/workflows/welcome-good-first-issue.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          script: |
+            const repoUrl = `https://github.com/${context.repo.owner}/${context.repo.repo}`;
+
+            await github.rest.issues.createComment({
+              ...context.repo,
+              issue_number: context.issue.number,
+              body: "🙌 **Hey there, future contributor!** 🙌\n\n" +
+                    "This issue has been labeled as **good first issue**, which means it's a great place to get started with the OpenHands project.\n\n" +
+                    "If you're interested in working on it, feel free to comment below to let us know — no need to ask for permission, just give us a heads-up so others don't duplicate work.\n\n" +
```
enyst:
```suggestion
                    "If you're interested in working on it, feel free to! No need to ask for permission.\n\n" +
```

enyst: Just thinking that good-first-issues are typically small. In theory, they don't need to be, it would be possible to imagine medium/large issues that require time, but not OpenHands deep knowledge, and in fact, adding evals is an example of that I think. But in practice, I think we don't have medium/large issues with this label.

So I think maybe there's no realistic risk to fix.

PR #9712, opened 2025-07-15
feat(ui): select component

mislavlukach:

Implements select component

---
Implements select component

---


openhands-ui/components/select/components/Option.tsx
```diff
@@ -0,0 +1,53 @@
...
+    inputValue,
+  },
+}: OptionProps & { value: unknown }) => {
+  const visibleOptions = (options ?? []).filter((o) =>
+    filterOption(
+      {
+        // @ts-ignore
+        ...o,
+        data: {
+          __isNew__: false,
```
mislavlukach: I don't like it but didn't found a better way to implement this.
So basically, as last option has to be rounded, we need a way to determine last option.
We do it by manually filtering options and getting the last. 

We need to filter because when you type in - react-select will only show a subset of options. 
So we can't simply rely on options[options.length-1]

openhands-ui/components/select/components/Option.tsx
```diff
@@ -0,0 +1,53 @@
...
+        },
+      },
+      inputValue
+    )
+  );
+  const lastOption =
+    visibleOptions?.length > 0
+      ? (visibleOptions![visibleOptions!.length - 1] as IOption<unknown>)
```
amanape:
```suggestion
      ? (visibleOptions[visibleOptions!.length - 1] as IOption<unknown>)
```

openhands-ui/components/select/components/Option.tsx
amanape: OK, maybe we can find a solution later down the line. I'll create a ticket. In the meantime, can you add a comment describing the problem that prompted this solution>

openhands-ui/components/select/components/Option.tsx
```diff
@@ -0,0 +1,53 @@
...
+import { cn } from "../../../shared/utils/cn";
+
+export const Option = ({
+  className,
+  children,
+  isSelected,
+  innerProps,
+  data,
+  selectProps: {
+    customProps: { filterOption },
```
amanape:
```sh
Property 'filterOption' does not exist on type '{ error?: string | undefined; hint?: string | undefined; readOnly: boolean | undefined; }'.ts(2339)
```

openhands-ui/components/select/components/Option.tsx
```diff
@@ -0,0 +1,53 @@
...
+        data: {
+          __isNew__: false,
+        },
+      },
+      inputValue
+    )
+  );
+  const lastOption =
+    visibleOptions?.length > 0
+      ? (visibleOptions[visibleOptions!.length - 1] as IOption<unknown>)
```
amanape:
```suggestion
      ? (visibleOptions[visibleOptions.length - 1] as IOption<unknown>)
```

PR #9718, opened 2025-07-15
Feat: Add current working directory to LLM instructions

tofarr:

Add current working directory to initial system instructions for the LLM

---
* In a LocalRuntime / CLIRuntime, there is no guarantee that the workspace will be `/workspace`
* When we specify a prompt like "Write a bash script that prints 'Hello World!'", the LLM needs to figure out where to place the script.
* If we have not specified a git repo, we have no directory definition for context.
* The LLM was reading our file instructions and seeing the examples contained `/workspace`, and using that.
* Sometimes I've seen the LLM places files in `/tmp`

image
image

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:1a5e8b9-nikolaik   --name openhands-app-1a5e8b9   docker.all-hands.dev/all-hands-ai/openhands:1a5e8b9
```


enyst:
To note here
> If we have not specified a git repo, we have no directory definition for context.

In CLI mode, by default it's the current directory, and it's even enforced.

openhands/agenthub/codeact_agent/prompts/system_prompt.j2
```diff
@@ -11,6 +11,7 @@ Your primary role is to assist users by executing commands, modifying code, and
 </EFFICIENCY>
 
 <FILE_SYSTEM_GUIDELINES>
+{% if cwd %}* The current working directory is {{ cwd }} {% endif %}
```
xingyaoww: I think we don't need the system message change as long as we have the dir path in `openhands/agenthub/codeact_agent/prompts/additional_info.j2`

xingyaoww: one of the reason we don't want to add them in system prompt is also due to the reason @enyst mentioned: CLI is a special case.

tofarr:
> To note here
> 
> > If we have not specified a git repo, we have no directory definition for context.
> 
> In CLI mode, by default it's the current directory, and it's even enforced.

This PR is aimed at passing the cwd on to the LLM - otherwise the LLM is unaware of it unless it invokes `pwd`. (And it usually doesn't!)

 I've seen it will `/workspace` based on our example code here:
image

I've also seen it use `/tmp`

Both of these can lead to a bad user experience.

openhands/agenthub/codeact_agent/prompts/system_prompt.j2
tofarr: I've removed the duplication.

xingyaoww: I was wondering if we can just remove this line from the system prompt, since we already have them in openhands/agenthub/codeact_agent/prompts/additional_info.j2

tofarr: The additional_info includes the repo directory inside the cwd, but not the cwd itself

xingyaoww: Then can we instead moving the cwd instruction to additional_info.j2 though?

My concern is that system prompt will persist over many many turns -- and maybe at 100 turns, the cwd already changed and the agent might not know it (e.g., due to condensation).

On the other hand, `additional_info.j2` is sent to the agent in the first user message, so its influence on the agent is potentially less.

(And it is pretty scary to me to change the system message bc we might need to re-run evaluation to make sure it doesn't introduce weird corner cases 😓 ... but if we change additional_info.j2, i'd be happy to just merge this

enyst:
Yes, no doubt, having it hardcoded makes a bit of a mess. Thanks!

tofarr:
> Actually, don’t we have already the current directory printed all the time by a CmdOutputObservation?
> 
The issue is with the initial prompt when no repo is selected, we have not yet executed any commands.

enyst:
Got it. I think, if we can add it to the workspace context RecallObs, then it could be a simpler and more consistent implementation. We wouldn't need to send this around to all agents, for example.

openhands/controller/agent.py
```diff
@@ -40,13 +42,15 @@ def __init__(
...
         config: AgentConfig,
+        cwd: str = DEFAULT_WORKSPACE_MOUNT_PATH_IN_SANDBOX,
     ):
         self.llm = llm
         self.config = config
         self._complete = False
         self._prompt_manager: 'PromptManager' | None = None
         self.mcp_tools: dict[str, ChatCompletionToolParam] = {}
         self.tools: list = []
+        self.cwd = cwd
```
xingyaoww: I don't think we need the changes in this file any more?

openhands/core/setup.py
```diff
@@ -140,6 +141,7 @@ def create_memory(
     repo_directory: str | None = None,
     status_callback: Callable | None = None,
     conversation_instructions: str | None = None,
+    working_dir: str = DEFAULT_WORKSPACE_MOUNT_PATH_IN_SANDBOX,
```
xingyaoww, line 144: We never seems to invoke `create_memory` with `working_dir` argument. Is this needed? If so, we should pass this arg to all invocation of `create_memory`, right?

xingyaoww (commented):
Two comments, otherwise LGTM

openhands/controller/agent.py
tofarr: Good catch - artifact of the previous changes - removed

openhands/core/setup.py
tofarr: Good catch - the CLI required the CWD to be passed in as the `working_dir`. (I had done this in an older version but missed it in the most recent update)

Fixed

xingyaoww (approved):
LGTM

openhands/memory/memory.py
```diff
@@ -344,11 +346,13 @@ def set_runtime_info(
                 additional_agent_instructions=runtime.additional_agent_instructions,
                 date=date,
                 custom_secrets_descriptions=custom_secrets_descriptions,
+                working_dir=working_dir,
```
enyst, line 349: Not a problem, just a question: we have a `runtime` instance here, doesn’t the runtime knows it?

tofarr: I suppose we could get the cwd from the runtime, but this would involve executing another command - the runtime does not save the `cwd` AFAIK

enyst: Ah, okay, probably not worth it. Thanks

PR #9727, opened 2025-07-15
Solvability setting switch

csmith49:

Adds a switch to enable or disable additional solvability analysis in the resolver.

---

Since the solvability analysis only triggers for the cloud resolver, the switch is only rendered in the SaaS environment.

---


frontend/src/i18n/translation.json
```diff
@@ -2367,6 +2367,22 @@
...
+        "de": "Lösbarkeitsanalyse aktivieren",
+        "no": "Aktiver løsningsanalyse",
+        "it": "Abilita analisi di risolvibilità",
+        "pt": "Ativar análise de solucionabilidade",
+        "es": "Habilitar análisis de solvencia",
+        "ar": "تمكين تحليل القابلية للحل",
+        "fr": "Activer l'analyse de solvabilité",
+        "tr": "Çözünürlük Analizini Etkinleştir",
+        "uk": "Увімкнути аналіз розв'язності"
+    },
```
csmith49, line 2433: The only translation here I can guarantee is for `"en"`, but Claude seems confident in the rest.

frontend/src/routes/app-settings.tsx
```diff
@@ -121,6 +125,11 @@ function AppSettingsScreen() {
     );
   };
 
+  const checkIfSolvabilityAnalysisSwitchHasChanged = (checked: boolean) => {
+    const currentSolvabilityAnalysis = !!settings?.ENABLE_SOLVABILITY_ANALYSIS;
+    setMaxBudgetPerTaskHasChanged(checked !== currentSolvabilityAnalysis);
```
malhotra5: This line seems a little funky - should be unrelated to max budget per task?

openhands/storage/data_models/settings.py
```diff
@@ -36,6 +36,7 @@ class Settings(BaseModel):
     enable_default_condenser: bool = True
     enable_sound_notifications: bool = False
     enable_proactive_conversation_starters: bool = True
+    enable_solvability_analysis: bool = True
```
malhotra5, line 39: Since this is a cloud only feature it doesn't need to be included here

frontend/src/routes/app-settings.tsx
csmith49: Good catch. Added the necessary state hook and fixed the callback.

enyst (changes requested):
This PR is a bit confusing, sorry. Please give me a moment to understand here.

First, I think I just saw in the published project roadmap, that solvability is for both OH and the Cloud projects. If that is so, then we could look into making it work for OH, too?

Second, if that is not accurate, and it's only for the Cloud, then IMHO a setting should be saved on the Cloud, like many other cloud-only settings (e.g. api keys, billing), not in the Settings object that is used by everyone.

Third, actually we have another alternative in OH, for something like this: projects built on OH that may want their own (non-OH) configuration:
https://github.com/All-Hands-AI/OpenHands/blob/0d1e21ae452690910d4e08d15f641f01dd94d011/openhands/core/config/openhands_config.py#L66

The `Extended` config is intended to accept any setting as written in the config file. Though I guess this isn't useful in this case since it applies to the whole project, rather than per user? Could we think about doing something similar, maybe, but for the per-user configuration?

csmith49:
> First, I think I just saw in the published project roadmap, that solvability is for both OH and the Cloud projects. If that is so, then we could look into making it work for OH, too?

Our plan right now is to have this as a feature for the cloud resolver.
 
> Second, if that is not accurate, and it's only for the Cloud, then IMHO a setting should be saved on the Cloud, like many other cloud-only settings (e.g. api keys, billing), not in the Settings object that is used by everyone.

Sure thing, that makes sense.

I got to this point by following the trail of `proactive_conversation_starter` breadcrumbs, which I thought was also a cloud-only feature (ergo, maybe the solvability analysis also needed to be stored there to sync with the cloud settings). I'll remove it from the `Settings` object.

enyst:
You're right, I think the `proactive` option should also be a Cloud-only setting (not part of the Settings object).

Cc: @jpelletier1  I saw this at like 2 a.m. last night 😅, but it seems it wasn't a hallucination:
image

Afaik this ability was initially developed in the open and under an open license. It's unfortunate enough to close it down, but at least you may want to fix the announced planning then.

jpelletier1:
> You're right, I think the `proactive` option should also be a Cloud-only setting (not part of the Settings object).
> 
> Cc: @jpelletier1 I saw this at like 2 a.m. last night 😅, but it seems it wasn't a hallucination: image
> 
> Afaik this ability was initially developed in the open and under an open license. It's unfortunate enough to close it down, but at least you may want to fix the announced planning then.

I'm sorry about that - this is my fault for not properly categorizing this feature. I will try to make sure we communicate this better in the future.

frontend/src/hooks/mutation/use-save-settings.ts
```diff
@@ -25,6 +25,7 @@ const saveSettingsMutationFn = async (settings: Partial<PostSettings>) => {
     mcp_config: settings.MCP_CONFIG,
     enable_proactive_conversation_starters:
       settings.ENABLE_PROACTIVE_CONVERSATION_STARTERS,
+    enable_solvability_analysis: settings.ENABLE_SOLVABILITY_ANALYSIS,
```
enyst, line 28: I’m most happy to remove the block, I’m wondering though, does this work now, for local use?

amanape (approved):
Other than failing CI, LGTM

PR #9732, opened 2025-07-16
Fix: Resolve cross-platform path splitting bug in search

### LINKED ISSUE #9703: [Bug]: Cross-platform path splitting bug in `FileConversationStore.search` causes crash on Windows

Overview

  The search method in FileConversationStore fails on Windows environments due to a hardcoded path separator in the logic used to extract conversation IDs. This makes the feature non-portable and unusable on
  Windows.

  Description

  The issue is located in the following line within FileConversationStore.search():

   1 conversation_ids = [
   2     path.split('/')[-2]
   3     for path in self.file_store.list(metadata_dir)
   4     if not path.startswith(f'{metadata_dir}/.')
   5 ]

  This code hardcodes the forward slash (/) as the path separator. While this works correctly on Unix-like systems (Linux, macOS), it is not compatible with Windows, which uses the backslash (\) as its native path
  separator.

  When this code is executed on Windows, self.file_store.list() returns paths like conversations\conv_123\metadata.json. The path.split('/') call fails to find any forward slashes and returns the original, unsplit
  string as a single-element list (e.g., ['conversations\\conv_123\\metadata.json']).

  The subsequent attempt to access the second-to-last element via the [-2] index results in an IndexError, crashing the method.

  Steps to Reproduce

   1. Set up and run the application on a Windows operating system.
   2. Create at least one conversation history so that self.file_store.list(metadata_dir) returns one or more paths.
   3. Call the FileConversationStore.search() method.
   4. The application will raise an IndexError.

  Expected Behavior

  The search() method should correctly parse the conversation_id from each file path, regardless of the underlying operating system. It should successfully build the list of conversation_ids and return a
  ConversationMetadataResultSet.

  Actual Behavior

  The method crashes and throws an FileNotFound on Windows systems.

  Environment

   * Operating System: Windows (e.g., Windows 10, Windows 11)
   * Python Version: All

  Suggested Solution

  The hardcoded path manipulation should be replaced with a platform-agnostic approach using Python's standard pathlib library.

  Current Code:

```python
onversation_id = path.split('/')[-2]
```

  Proposed Fix:

```python
 from pathlib import Path
   
 # ... inside the list comprehension ...
conversation_id = Path(path).parent.name
```

  This change will correctly and robustly extract the parent directory's name (which corresponds to the conversation_id) on any operating system, resolving the bug. This is the idiomatic way to handle file paths
  in modern Python.

### OpenHands Installation

Other

### OpenHands Version

_No response_

### Model Name

_No response_

### Operating System

None

_No response_

bojackli:
Replaced hardcoded path splitting with pathlib.Path.name to ensure correct behavior on both Unix-like and Windows systems. This also improves the logic for filtering hidden files.

Fixes #9703

  Fixes a crash on Windows systems that occurred when searching or listing conversation histories. The application now correctly handles Windows-style file paths, making this feature reliable across all operating
  systems.

---
This PR resolves a cross-platform compatibility bug in FileConversationStore.search caused by hardcoded forward slashes (/) for path manipulation.

  The original code used path.split('/')[-2] to extract the conversation ID and a hardcoded string to filter hidden files, both of which fail on Windows.

  This has been refactored to use the pathlib library for all path operations:
   - pathlib.Path(path).name is now used to robustly extract the conversation ID from the directory path, regardless of the OS.
   - The check for hidden files is now performed with Path(path).name.startswith('.'), making it platform-agnostic.

  These changes ensure the conversation search feature is fully functional on Windows without affecting its behavior on Unix-like systems.

---
 Fixes #9703


openhands/storage/conversation/file_conversation_store.py
```diff
@@ -73,9 +73,9 @@ async def search(
         metadata_dir = self.get_conversation_metadata_dir()
         try:
             conversation_ids = [
-                path.split('/')[-2]
+                Path(path).name
                 for path in self.file_store.list(metadata_dir)
-                if not path.startswith(f'{metadata_dir}/.')
+                if not Path(path).name.startswith('.')
```
enyst, line 78: I wonder, shouldn't this be `Path(path).parent.name` maybe? But I could be wrong

Does it work for you?

bojackli: Yes, it works for me. The reason is that the previous list file will suffix the folder name with a '/', and split[-2] correctly gives the current directory name rather than the parent directory name.

image

enyst (approved):
Thank you!

Issue #4184, opened 2024-10-03
[LLM] Support LLM routing


xingyaoww:
**What problem or use case are you trying to solve?**

Not Diamond intelligently identifies which LLM is best-suited to respond to any given query. We want to implement a mechanism in OpenHands to support this type of "LLM" selector.

**Describe the UX of the solution you'd like**

Ideally, use should define a "LLMRouter" as a special type of LLM with some special configs (e.g., multiple keys for different providers). And user can just put in keys, and select that router, and OpenHands will automatically use that going forward.

**Do you have thoughts on the technical implementation?**

Modify https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/llm/llm.py, as well as config related files under https://github.com/All-Hands-AI/OpenHands/tree/main/openhands/core/config.

You should probably use `model_select` (from notdiamond API) rather than `create` to be compatible with existing LiteLLM calls.

**Describe alternatives you've considered**

**Additional context**

Here's the documentation from NotDiamond

```
# -*- coding: utf-8 -*-
"""Getting started with Not Diamond

Automatically generated by Colab.

Original file is located at
    https://colab.research.google.com/drive/1Ao-YhYF_S6QP5UGp_kYhgKps_Sw3a2RO

# **Setting up**
"""

!pip install -q notdiamond[create] --upgrade

import os

os.environ["NOTDIAMOND_API_KEY"] = 'YOUR_NOTDIAMOND_API_KEY'
os.environ["OPENAI_API_KEY"] = 'YOUR_OPENAI_API_KEY'
os.environ["ANTHROPIC_API_KEY"] = 'YOUR_ANTHROPIC_API_KEY'
os.environ["PPLX_API_KEY"] = 'YOUR_PERPLEXITY_API_KEY'

"""# **Automatic routing and model calling**

We'll start by defining the routing `NotDiamond` client, which will function like a 'meta-LLM' that ensembles together the best of multiple models.

We then add specific LLM targets for routing. Not Diamond works with any LLM in [this list](https://notdiamond.readme.io/docs/llm-models).
"""

from notdiamond import NotDiamond

client = NotDiamond()

llm_providers = [
    'openai/gpt-4o',
    'anthropic/claude-3-5-sonnet-20240620',
    'openai/gpt-4o-mini',
    'perplexity/llama-3.1-sonar-large-128k-online'
]

"""Next, let's call the client by passing in an array of messages and our target models:"""

[...]

PR #9738, opened 2025-07-16
Implement model routing support

ryanhoangt:

Add support for routing between different LLMs in an agent trajectory.

---

This PR adds experimental support for model routing feature, with an example implementation:
- `RuleBasedCostSavingRouter`: use the secondary model by default, if there are image inputs or the context length of the secondary model exceeds, use the primary model in subsequent turns.

---

Fixes https://github.com/All-Hands-AI/OpenHands/issues/4184


openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -90,6 +99,28 @@ def __init__(
         self.condenser = Condenser.from_config(self.config.condenser)
         logger.debug(f'Using condenser: {type(self.condenser)}')
 
+        self.router: BaseRouter | None = None
```
xingyaoww: I don't think agent shouldn't have access to the router? We should maybe put the router into LLM class, or modify the LLM class?

ryanhoangt: I think it might be easier to put the router in the agent layer, I remember trying putting it in the LLM previously but it seems to be quite challenging. I imagine the LLM should be just a wrapper around litellm and deal with prompt caching, reasoning, retrying, etc 🤔 We also need to format events based on function calling support in the agent layer and in LLM layer we only have access to the formatted `messages`.

xingyaoww: In general, I feel the changes we introduced to the agent are a bit too much and scattered. Even though we need to put the router into the agent, I imagine we can make an `LLMRouter` that takes the place of `self.llm` - and it will just be an `LLM` if the routing is not enabled. i.e., these configuration check should happen with in `LLMRouter`, not outside it?

> We also need to format events based on function calling support in the agent layer 

Can you elaborate? I believe this happens on the LLM layer, where we convert fn call to non-fncall requests?

ryanhoangt: > Can you elaborate? I believe this happens on the LLM layer, where we convert fn call to non-fncall requests?

Ah it's not function calling, I mean this one -- when we construct the `messages` list we also need to do some customization based on whether the LLM supports prompt caching or vision capability:

https://github.com/All-Hands-AI/OpenHands/blob/0d1e21ae452690910d4e08d15f641f01dd94d011/openhands/agenthub/codeact_agent/codeact_agent.py#L225-L272

> In general, I feel the changes we introduced to the agent are a bit too much and scattered. I imagine we can make an LLMRouter that takes the place of self.llm

Yeah I agree, I also thought about it but didn't figure out a better way yet. This sounds like a good idea, let me look more closely

config.template.toml
```diff
@@ -219,6 +219,14 @@ correct_num = 5
 api_key = ""
 model = "gpt-4o"
 
+# Example routing LLM configuration for cost-saving model routing
+# Uncomment and configure to enable model routing with a cheaper model
+#[llm.weak_model]
```
xingyaoww: Can we think of better name for this? Hmm maybe like primary model / secondary model?

openhands/router/cost_saving/rule_based_impl.py
xingyaoww, line 1: Can you rename the folder to just `router/impl/rule_based.py`? We don't really need to mention cost saving here

evaluation/utils/shared.py
```diff
@@ -76,6 +76,9 @@ class EvalOutput(BaseModel):
     # Optionally save the input test instance
     instance: dict[str, Any] | None = None
 
+    # For model routing analysis
+    routing_history: list[int] | None = None
```
xingyaoww: Can we add these as a separate extre to "metrics" similar to latency, rather than creating a new field here

openhands/core/config/model_routing_config.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Attributes:
+        prob_threshold (float): The probability threshold for routing decisions, used by `threshold_based_cv_router`.
+        router_name (str): The name of the router to use. Default is 'noop_router'.
+    """
+
+    prob_threshold: float = Field(
+        default=0.49414062500000006
+    )  # 60% calls to strong model
+    router_name: str = Field(default='noop_router')
```
xingyaoww, line 14: Can we maybe make model routing config as an attribute to `LLMConfig` class?

openhands/core/config/openhands_config.py
```diff
@@ -59,10 +63,12 @@ class OpenHandsConfig(BaseModel):
     """
 
     llms: dict[str, LLMConfig] = Field(default_factory=dict)
+    routing_llms: dict[str, LLMConfig] = Field(default_factory=dict)
```
xingyaoww: I think this make things a lot more confusing :(

openhands/router/base.py
```diff
@@ -0,0 +1,31 @@
...
+        # The active LLM for the current turn
+        self.active_llm = llm
+
+        # Tracking data
+        self.routing_history: list[int] = []
+
+    @abstractmethod
+    def set_active_llm(self, messages: list[Message], events: list[Event]) -> None:
+        """Configure the active LLM for the current turn based on the messages and events."""
+        pass
```
xingyaoww: what do you think about this design:
1. inherit `LLMRouter` from `LLM`
2. make `RouterConfig` an attribute of `LLMConfig`, inside `RouterConfig`, we can put an attribute like `secondary_llm_config` which is an `LLMConfig`, so that we don't really need to change the `__init__` interface from LLM
3. make sure `LLMRouter` has the exact same interface as `LLM`, with the exception of an additional `update_active_llm`, which is exactly what you've just implemented
```
class LLMRouter(LLM):
  def __init__(XXX):
    self.primary_llm = XX
    self.secondary_llm = YY

  def completions(XXX):
      
  def update_active_llm() --> maybe we should put this in LLM base class too, but make it a no-op

  @classmethod
  def from_llm_config(llm_config) -> LLMRouter | LLM:
```

And then inside `codeact_agent`, probably all we need to do is to add one line of `update_active_llm`.

We primarily just need to change how we create `LLM` instance (search for where we call `LLM(`), and then in those places, we just call `LLMRouter.from_llm_config` instead of our existing LLM constructor, if that make sense?

config.template.toml
enyst: Or small model, but I think secondary feels more clear… maybe

evaluation/benchmarks/swe_bench/run_infer.py
```diff
@@ -249,6 +252,12 @@ def get_config(
         enable_prompt_extensions=False,
     )
     config.set_agent_config(agent_config)
+    config.model_routing = config_copy.model_routing
+
+    # Set log_completions to True for all routing LLMs
+    for llm_cfg in config.routing_llms.values():
```
enyst: Does this work, is `routing_llms` in `config` instance? I thought create that manually _without_ reading `config.toml`, or did that change?

openhands/controller/state/state.py
```diff
@@ -93,6 +93,8 @@ class State:
     resume_state: AgentState | None = None
     # global metrics for the current task
     metrics: Metrics = field(default_factory=Metrics)
+    # routing LLM metrics for the current task
+    local_routing_metrics: list[Metrics] = field(default_factory=list)
```
enyst: Nits:
- “local” metrics in `State` is intended for delegates
- (and I think Rohit is removing the existing one)
- if we need some other metrics, maybe we can name it without “local”
- I don’t see `local_routing_metrics` anywhere else in the PR, is it used?

openhands/core/config/model_routing_config.py
enyst: I’m not sure… I _think_ the router “has” LLM instances? While this means LLMConfig “has” a router config.

Idk, if the configs are the other way around than the objects they configure… it might mess with our brains (they’re pattern-based brains, natural or silicon 😅)

openhands/core/config/openhands_config.py
enyst: Not sure what it means… Does it mean “the LLMs that are either primary or secondary LLM when routing is enabled”?

openhands/router/README.md
```diff
@@ -0,0 +1,40 @@
...
+# Main LLM (strong model)
+[llm]
+model = "gpt-4o"
+api_key = "your-api-key"
+
+# Weak model for routing
+[llm.weak_model]
+model = "gpt-4o-mini"
+api_key = "your-api-key"
+for_routing = true
```
enyst: Ah I see… Just wondering then though: if `for_routing` is false, does that mean that `llm.weak_model` will NOT be the weak model?

openhands/router/base.py
enyst: We really need to think about this… Rohit is working on an [LLMRegistry](https://github.com/All-Hands-AI/OpenHands/pull/9589), where we can `register_llm`. At the moment not sure if we can also `get_existing_llm` instead of… the existing LLM constructor 🫣 , but I think we need such method…

openhands/controller/state/state.py
malhotra5: Yep local metrics have been deprecated

Instead we capture a snapshot of the metrics the parent had at the time of delegation (the `parent_metrics_snapshot` field), and the delegate continues to accumulate cost on top of the parent's metrics. 

The method `get_local_metrics` computes the difference of the current metrics and the snapshot at the time of delegation

enyst (commented):
Thank you @ryanhoangt, I love we see this PR back!  Yes, we really could limit functionality to this extra “weak” LLM for now, to keep complexity in check.

I would really really love if as we figure this out, we find a way consistent with Rohit’s PR, because all of these are functionalities we want
- centralized metrics (per conversation; includes condenser llm, etc)
- LLMRegistry (with running LLMs and all LLM configurations registered in the app)
- LLM Routing (with 2 LLMs, one with predefined name “secondary” or whatever)
- change LLM during conversation (this should be straightforward if we use a registry instead of LLM constructors everywhere)

The 4th is related to the 3rd, it’s an LLM switch, either decided by the router automatically, or by the user from somewhere in some UI.

Please feel free to take part in the architecture discussion and attempts to disentangle things in Rohit’s PR. We had a similar issue with [complexity in the agent](https://github.com/All-Hands-AI/OpenHands/pull/9589#discussion_r2226687467), too, that requires some thought 🤔

openhands/core/config/openhands_config.py
ryanhoangt: Yeah I agree it's quite confusing, but we need it so that we can know what configs to create LLM instances and track metrics during a session. Currently we will accumulate metrics from the primary LLM and all the LLMs with `for_routing` enabled (i.e. this `routing_llms` list, I make it a list to support potential routing strategies that are not just binary like the current rule based router). Without it I imagine we have to create LLM instances for all the configs and merge metrics for those that are not used.

ryanhoangt: Actually this is not needed anymore, I figured out a way to simplify it.

openhands/router/base.py
ryanhoangt: @xingyaoww I adjusted the PR to simplify the code inside `codeact_agent.py`, borrowing some of the ideas from your comment. Could you take another look to see if it's good enough or we need further refactor _regarding the base router definition & integration w/ agent_? One important design choice IMO is we should make the routing config and router instance attach to the agent rather than the LLM so that it's easier to perform metric tracking for separate LLM instances, handle function calling for individual LLM, etc.

openhands/core/config/model_routing_config.py
ryanhoangt: > I think the router “has” LLM instances?

Yeah this is my idea, I moved this `model_routing_config` to `agent_config`, which seems to make sense semantically and also make the code cleaner.

openhands/router/rule_based/impl.py
```diff
@@ -0,0 +1,72 @@
+from openhands.core.config import ModelRoutingConfig
+from openhands.core.logger import openhands_logger as logger
+from openhands.core.message import Message
+from openhands.events.action import MessageAction
+from openhands.events.event import Event
+from openhands.llm.llm import LLM
+from openhands.router.base import ROUTER_REGISTRY, BaseRouter
+
+
+class RuleBasedCostSavingRouter(BaseRouter):
```
xingyaoww: Let's not calling it "cost saving"

openhands/router/base.py
```diff
@@ -0,0 +1,50 @@
...
+    ):
+        self.llm = llm
+        self.model_routing_config = model_routing_config
+
+        # Instantiate all the routing LLM instances
+        routing_llms_config = model_routing_config.routing_llms
+        self.routing_llms = {
+            llm_name: LLM(config=llm_config)
+            for llm_name, llm_config in routing_llms_config.items()
+        }
```
xingyaoww: If we are constructing the llm instance, i guess it is not really necessary to maintain `self.llm`?

evaluation/benchmarks/swe_bench/run_infer.py
ryanhoangt: Yeah you're right, it should be `config_copy.routing_llms`, I've updated it in the current PR.

openhands/controller/state/state.py
ryanhoangt: Thanks @enyst @malhotra5 for the info, I've updated the code to remove this `routing_metrics` and instead merge individual routing LLM metrics into the main `metrics` inside the `agent_controller` at each turn. Would appreciate another look and feedback if you have time!

openhands/router/README.md
ryanhoangt: If `for_routing` is false, the current behavior is that the router will raise a validation error as the `secondary_model` doesn't exist.

openhands/router/rule_based/impl.py
xingyaoww: I suggest we rename it to something like MultimodalRequestRouter?

config.template.toml
```diff
@@ -479,3 +487,17 @@ type = "noop"
...
+
+#################################### Model Routing ############################
+# Configuration for experimental model routing feature
+# Enables intelligent switching between different LLM models for specific purposes
+##############################################################################
+[model_routing]
+# Router to use for model selection
+# Available options:
+# - "noop_router" (default): No routing, always uses main LLM
+# - "rule_based_cv_router": Cost-saving router that switches between primary and secondary models
```
xingyaoww: make sure you update the name across the project if we renamed the router

evaluation/benchmarks/gaia/run_infer.py
```diff
@@ -70,18 +73,34 @@ def get_config(
...
+
+    config_copy = copy.deepcopy(config)
+    load_from_toml(config_copy)
+    model_routing_config = config_copy.get_agent_config().model_routing
+    # Set log_completions to True for all routing LLMs
+    for llm_cfg in model_routing_config.routing_llms.values():
+        llm_cfg.log_completions = True
+        update_llm_config_for_completions_logging(
+            llm_cfg, metadata.eval_output_dir, instance['instance_id']
+        )
```
xingyaoww: This seems a lot of code just to turn on routing -- is config_copy necessary? can we simplify?

evaluation/benchmarks/swe_bench/run_infer.py
```diff
@@ -242,15 +243,27 @@ def get_config(
     # get 'draft_editor' config if exists
     config.set_llm_config(get_llm_config_arg('draft_editor'), 'draft_editor')
 
+    config_copy = copy.deepcopy(config)
```
xingyaoww: ditto - is this necessary?

evaluation/benchmarks/gaia/run_infer.py
```diff
@@ -70,18 +73,34 @@ def get_config(
...
     )
-    config.set_llm_config(metadata.llm_config)
+    config.set_llm_config(
+        update_llm_config_for_completions_logging(
+            metadata.llm_config, metadata.eval_output_dir, instance['instance_id']
+        )
+    )
+
+    config_copy = copy.deepcopy(config)
+    load_from_toml(config_copy)
```
xingyaoww: We did NOT want to load these from toml because they can easily mess-up evaluation (e.g., accidentally changed a config in config.toml for something else, and then eval is also using that setting).

Can we stop using load_from_toml?

evaluation/benchmarks/swe_bench/run_infer.py
```diff
@@ -242,15 +243,27 @@ def get_config(
...
+    load_from_toml(config_copy)
```
xingyaoww: ditto

openhands/controller/state/state.py
```diff
@@ -103,6 +103,10 @@ class State:
     parent_metrics_snapshot: Metrics | None = None
     parent_iteration: int = 100
 
+    routing_history: list[int] = field(
+        default_factory=list
+    )  # 1 means routing activated, 0 means not
```
xingyaoww: Since this is just a list of ints, what is the purpose of tracking them? Can we just do that together with LLM metrics?

openhands/core/config/model_routing_config.py
```diff
@@ -0,0 +1,43 @@
...
+        prob_threshold (float): The probability threshold for routing decisions, used by `threshold_based_cv_router`.
+        router_name (str): The name of the router to use. Default is 'noop_router'.
+        routing_llms (dict[str, LLMConfig]): A dictionary mapping LLM routing config names to their configurations.
+    """
+
+    prob_threshold: float = Field(
+        default=0.49414062500000006
+    )  # 60% calls to strong model
+    router_name: str = Field(default='noop_router')
+    routing_llms: dict[str, LLMConfig] = Field(default_factory=dict)
```
xingyaoww: nit: i feel `llms_for_routing` easier to understand

openhands/llm/llm.py
```diff
@@ -835,3 +836,42 @@ def format_messages_for_llm(self, messages: Message | list[Message]) -> list[dic
 
         # let pydantic handle the serialization
         return [message.model_dump() for message in messages]
+
+    def passthrough(
```
xingyaoww: is this get used anywhere?

enyst (changes requested):
I think this PR is looking better and better, and as Xingyao said, it’s almost ready.

I think we should merge first the `LLMRegistry` / `ConversationStats` mechanics in Rohit’s PR. Because it will give us the best of worlds: a central place to track LLM creations and switches (and compute/store statistics accordingly), which means routing switches too.

Now, that PR doesn’t have any switch, only creations, this will be the first to actually do it. I just think the LLMRegistry design we’re trying there is the way to go. Please do tell if I might be wrong.

(It also implies that we don’t do scattered statistics in agent controller or anywhere else anymore, we just notify the statistics center.)

malhotra5:
Hoping to make good progress on the registry today, apologies for the delay!

openhands/controller/state/state.py
ryanhoangt: Yeah I think we can construct it from metrics, this is just for convenience so we don't need to do that. I've removed them since it seems to appear in many places.

openhands/core/config/model_routing_config.py
ryanhoangt: Yeah that seems better, I updated it!

evaluation/benchmarks/gaia/run_infer.py
ryanhoangt: Actually we don't load the whole `toml` file into the config object used in evaluation, we only copy the `model_routing` config and update the original object. I think this is necessary to load these values (and also things like `search_api_key` for evaluation?

openhands/router/base.py
ryanhoangt: We do need `self.llm` to set `self.active_llm` to it in the `set_active_llm` method.

evaluation/benchmarks/swe_bench/run_infer.py
ryanhoangt: Yeah, we need a way to read these config values from the `config.toml`, we also do this similarly for GAIA when we need the `search_api_key`. We only copy the `model_routing_config` and leave other configs intact.

evaluation/benchmarks/gaia/run_infer.py
enyst: @ryanhoangt  maybe we can do it the same way we do with LLM Configs: a separate function (`get_llm_config_arg` I think it's called?) which, itself, in the `config` module, makes sure to only return the requested object? 

Rather than loading the file in eval code. Which is fine as used here, but it's still a bit error prone perhaps in the future.

openhands/router/base.py
enyst: I think in Rohit's PR we're looking at 
- a LLMRegistry which holds the running LLMs for every service/role (e.g. the "agent" LLM, the "condenser" LLM, etc)
- it knows which is the LLM for a service at any point during execution
- can register a new running LLM with `register_llm (llm_config, service_id)`
- can return the active LLM
- is an Observable and can notify observers when the active LLM changed

Does that make sense?

evaluation/benchmarks/gaia/run_infer.py
ryanhoangt: Yeah that seems better, I updated it in the lastest commit -- lmk if you have any other concerns!

xingyaoww:
@OpenHands /codereview

xingyaoww:
@ryanhoangt some preliminary reviews :) can you send your agents to fix them?

> The most critical fix is the router registration import path in LLMRegistry.initialize_router—without it, default routing will fail at runtime

can you confirm if this is a valid concern?

# From OH

Here’s a focused code review of PR #9738 “Implement model routing support,” organized by scenario with specific, actionable items.

## Style and Formatting
[openhands/core/config/model_routing_config.py, Line 16]
:hammer_and_wrench: Pydantic config style: model_config is set as a plain dict. Elsewhere in the codebase you use ConfigDict for Pydantic v2.
Suggestion: Replace model_config = {'extra': 'forbid'} with model_config = ConfigDict(extra='forbid') for consistency and type checking.

[openhands/router/rule_based/impl.py, Lines 41–49]
:hammer_and_wrench: Logging levels: Logging token window checks and routing decisions as warning may be too noisy in normal operation.
Suggestion: Downgrade “Routing to the primary model…” to info. Keep warnings for genuinely exceptional cases (e.g., misconfiguration).

[config.template.toml, Lines 223–236]
:hammer_and_wrench: Terminology consistency: Comments mix “multi-modal” and “multimodal”.
Suggestion: Prefer “multimodal” consistently.

## Clarity and Readability
[openhands/llm/llm_registry.py, Lines 120–131]
:mag: Router registration clarity: initialize_router imports BaseRouter from openhands.router.base. That file defines the registry but doesn’t register routers—registration happens in openhands/router/__init__.py by importing NoOpRouter and MultimodalRouter.
Issue: ROUTER_REGISTRY will be empty, and BaseRouter.from_config will raise ValueError for the default “noop_router”.
Suggestion: Import via the package to trigger registration side-effects:
- from openhands.router import BaseRouter (preferred), or
- import openhands.router as _ before using BaseRouter.
This ensures ROUTER_REGISTRY is populated.

[openhands/agenthub/codeact_agent/codeact_agent.py, Lines 92–96]
:mag: Failure handling on router init: initialize_router will currently raise and break agent construction if the router isn’t registered or config invalid.
Suggestion: Make router init robust:
- Wrap initialize_router in try/except and fall back to no-op routing with a log message.
- Or short-circuit: only initialize when config.model_routing.router_name != 'noop_router' or when any llms_for_routing exist.

[openhands/agenthub/codeact_agent/codeact_agent.py, Lines 202–212]
:mag: Double message construction: step() calls _get_messages twice (once for routing decision and again for the actual call). This is potentially expensive.
Suggestion: Build messages once, then:
- Use those same messages to compute token counts for routing.
- If the active model changes and max_message_chars/vision flags differ, re-apply only those transformations rather than recomputing the entire message list from events (or clearly comment why recomputation is required).

[openhands/agenthub/codeact_agent/codeact_agent.py, Lines 109–140 (_get_tools)]
:mag: Tool description choice tied to initial LLM: The decision to use short tool descriptions is based on self.llm.config.model at construction time. When routing switches to a different model with stricter limits, tools may be overly verbose.
Suggestion: Either:
- Base the decision on self.llm_registry.get_active_llm() at call time, or
- Choose conservative (short) descriptions if model routing is enabled.

[openhands/core/config/utils.py, Lines 559–590]
:mag: Duplicate config parsing: get_llms_for_routing_config re-parses config.toml to filter for_routing models, while load_from_toml already builds cfg.llms and applies for_routing.
Suggestion: Consider a path that uses the loaded cfg.llms to avoid redundant parsing or document why re-read is needed (e.g., for evaluation scripts that don’t pass a preloaded config).

[openhands/core/config/model_routing_config.py, Lines 20–38]
:mag: API shape of from_toml_section: Returning a dict keyed by "model_routing" is slightly non-intuitive given there is only one section.
Suggestion: Consider returning ModelRoutingConfig directly (aligning with get_model_routing_config_arg) or document the rationale (consistency with other sections that return multiple named configs).

## Security and Common Bug Patterns
[openhands/llm/llm_registry.py, Lines 120–131]
:bug: Router registry not initialized: As noted above, importing from openhands.router.base does not register routers; default “noop_router” won’t be found. This will raise ValueError and break agent construction.
Fix: Import BaseRouter via openhands.router so that NoOpRouter and MultimodalRouter get registered.

[openhands/router/rule_based/impl.py, Lines 54–64]
:mag: Redundant work: get_token_count(messages) is called twice in the warning message branch.
Suggestion: Compute once:
token_count = self.secondary_llm.get_token_count(messages)
if self.secondary_llm.config.max_input_tokens and token_count > self.secondary_llm.config.max_input_tokens: ...

[tests/unit/agenthub/test_agents.py, Lines 271–286]
:bug: Test fragility due to router init: CodeActAgent now initializes the router in __init__. If the router registry isn’t populated (see registry issue), tests will fail during agent construction.
Suggestion:
- Import openhands.router in the test setup to ensure registration, or
- Patch llm_registry.initialize_router to a no-op in unit tests that don’t cover routing behavior.

[openhands/router/rule_based/impl.py, Lines 29–39]
:mag: Behavior permanence: Once self.max_token_exceeded is set True, primary is used on all subsequent turns. This matches the PR description (“use primary in subsequent turns”), but it’s not obvious from code alone.
Suggestion: Add a brief comment clarifying the intentional “stickiness” of routing after overflow, or make it a configurable option.

[openhands/core/config/utils.py, Lines 226–259]
:mag: Silent partial configuration: load_from_toml always assigns a ModelRoutingConfig to the default agent and auto-populates llms_for_routing by scanning for_routing=True. If users set for_routing on other LLM configs or have multiple agents, these are silently ignored.
Suggestion: Document that routing settings currently apply only to the default agent, or extend support to per-agent routing configs.

[openhands/agenthub/codeact_agent/codeact_agent.py, Lines 284–291]
:mag: Vision gating for message construction: When building messages for routing decisions, max_message_chars and vision_is_active are taken from the active LLM at the time (likely the primary). When the router selects a different model, messages are rebuilt with the new active model config—this is correct but subtle.
Suggestion: Add a brief comment noting why messages are recomputed (different max_message_chars/vision capabilities per routed model).

[openhands/core/config/llm_config.py, Lines 90–95]
:hammer_and_wrench: Field doc alignment: You added for_routing to the model (good). Ensure any external docs or config schema references also list it; otherwise users may miss it.
Suggestion: Add a brief comment in config.template.toml near [llm.secondary_model] explaining for_routing and max_input_tokens interplay.

Additional Suggestions (Non-blocking)
- Consider adding unit tests for:
  - NoOpRouter initialization and pass-through behavior.
  - MultimodalRouter decisions with: (a) text-only within budget, (b) text exceeding secondary budget, (c) image inputs.
  - LLMRegistry.initialize_router importing path correctness (protect against regression).
- In evaluation scripts, the new get_metrics(state) usage is a good improvement; consider propagating to any remaining scripts for consistency.

Overall
The routing abstraction and configuration shape are headed in a solid direction. The most critical fix is the router registration import path in LLMRegistry.initialize_router—without it, default routing will fail at runtime. The other items are mostly polish and clarity improvements to reduce surprise and maintenance overhead.

openhands/llm/llm_registry.py
```diff
@@ -108,10 +116,24 @@ def get_llm(
...
+            agent_config=agent_config,
+        )
+
+    def set_active_llm_for_routing(
+        self, messages: list[Message], events: list[Event]
+    ) -> None:
+        """Set the active LLM based on routing decisions."""
+        if self.router:
+            self.router.set_active_llm(messages, events)
+            self.active_agent_llm = self.router.active_llm
```
malhotra5: Do we need to hold the LLMs in the router? I see the router as a classifier that just needs to return a `service_id` (unique string identifier) given messages and events

The LLM registry will hold all the LLMs and track which one is active as well (based on the service ids the router returns)

malhotra5: ^ or actually does the registry need to keep track of the active llm at all? We could remove it all together and make sure the router is solely tracking the active llm

enyst: ^^ I agree, the `LLM` registered in the `LLMRegistry` for a particular service, here, 'agent', is the active LLM.

I think the router only needs to tell the registry 'hey change to secondary_llm' or viceversa?

ryanhoangt: > Do we need to hold the LLMs in the router? I

I think we need access to the relevant LLM inside the router (e.g. `secondary_llm` for the multimodal router), e.g. to access its context length, token couter, etc. But maybe we don't need to maintain all the LLMs for routing in the base class, we can just retrieve the necessary LLM inside the specific router implementation based on its service id.

> ^^ I agree, the LLM registered in the LLMRegistry for a particular service, here, 'agent', is the active LLM.
I think the router only needs to tell the registry 'hey change to secondary_llm' or viceversa?

Do you mean we should completely hide the `active_agent_llm` property from the LLM registry and maintain the `service_id` for the active LLM internally? In the `codeact_agent.py` we work with `self.llm` which is set once, not the `self.llm_registry.get_llm(service_id)` so it's a bit tricky to implement I guess.

enyst: Yes, the LLMRegistry does just that: maintains `service_id` to `LLM instance` mapping.

You are correct it raises some questions then! I think there are multiple ways to solve the question: wait, what if `self.llm` in the client components (e.g. in the agent) becomes stale? Let me bring it up in slack.

enyst: Please see [slack topic](https://openhands-ai.slack.com/archives/C078L0FUGUX/p1756135906819359)

ryanhoangt: @malhotra5 I've removed the active_llm inside the router and reused the `_set_active_llm` helper you created, LMK if this aligns with your review!

@enyst about your suggestion, I asked OpenHands about it and it said we'd need to make breaking changes across ~26 direct self.llm usages. Do you think it's fine to go with the current integration and we can do an architectural refactor in a later PR, after we finalize on it?

openhands/router/rule_based/impl.py
```diff
@@ -0,0 +1,74 @@
...
+    def get_active_llm(self, messages: list[Message], events: list[Event]) -> str:
+        route_to_primary = False
+        # Handle multimodal input
+        for event in events:
+            if (
+                isinstance(event, MessageAction)
+                and event.source == 'user'
+                and event.image_urls
+            ):
+                logger.info('Image content detected. Routing to the primary model.')
```
xingyaoww: Actually i think we can maybe do this with messages by looking for stuff like `.type != 'text'` in `.content`?

openhands/agenthub/codeact_agent/codeact_agent.py
```diff
@@ -276,15 +292,18 @@ def _get_messages(
...
+        active_llm = self.llm_registry.get_active_llm()
+
         # Use ConversationMemory to process events (including SystemMessageAction)
         messages = self.conversation_memory.process_events(
             condensed_history=events,
             initial_user_action=initial_user_message,
-            max_message_chars=self.llm.config.max_message_chars,
-            vision_is_active=self.llm.vision_is_active(),
+            max_message_chars=active_llm.config.max_message_chars,
+            vision_is_active=active_llm.vision_is_active(),
```
xingyaoww: Right now, we "remove vision messages" it in `conversation_memory.process_events` not inside "LLM", which makes it hard for router to become a subclass of `LLM`
because the messages that reaches `self.llm` already doesn’t have vision components.
https://github.com/All-Hands-AI/OpenHands/blob/f9099fe6dbd8766f0306ce004d8ef3ad4c03fdc4/openhands/agenthub/codeact_agent/codeact_agent.py#L284

xingyaoww (commented):
I think we may want to refactor the `vision_is_active` inside `conversation_memory.process_events` to the `LLM` class, so that the LLM can receive the complete `messages` that contain multi-modal messages.

Then we can do something like this to simplify the entire router -- then for "agent", they just know that they got an "LLM" and don't need to worry anything about switching models, they are all handled inside the "LLM".
```
llm_registry = LLMRegistry()

class CompositeLLM(LLM):
	
	def completion(*args, **kwars):
	    # we do model routing logic here
	    
class MultimodalModelRouter(CompositeLLM):
	def completion(messages: list[Message]):
	  # do some classification
		has_multmodal_content: bool = False
		for message in messages:
			if message.contains_vision():
				has_multmodal_content = True
		
		if has_multmodal_content:
			return self.primary_llm(messages)
		else:
			return self.secondary_llm(messages)
	    
llm: CompositeLLM = llm_registry.get_router(
   primary="gpt5",
   secondary="sonnet"
)
agent = Agent(llm=llm, tools=tools)
```

@enyst would appreciate thought here!

openhands/llm/llm_registry.py
enyst: I'm a bit confused, because I see at least three options (in slack discussion, option 1, 2.1 and 2.2), so I don't know which your question to OH was about. 😅

Out of curiosity, could you paste the full -ish response?

Or maybe it was the remark above 🤔

I do agree that we should keep things contained for this PR.

malhotra5:
This PR LGTM! I'll defer the final approval to @xingyaoww

ryanhoangt:
Tested the final time to make sure it works fine, I'm merging it now, thanks all for the review!

Screenshot 2025-09-08 at 15 57 15

jpelletier1:
This is very cool!

PR #9749, opened 2025-07-16
feat(backend): API to get the microagents for the selected repository.

hieptl:

According to this [Slack](https://openhands-ai.slack.com/archives/C095K9Z52HH/p1752665114354569) thread, we need to implement an API to retrieve the microagents for the selected repository.

**Acceptance criteria:**
- The API should return the microagents found in selected repository.
- It should scan the .openhands/microagents directory in selected repository.

---

The PR contains an API to return the microagents found in selected repository.

We can refer to the video below for the output of the PR:

https://github.com/user-attachments/assets/a229e042-4854-48dc-9b08-7ef80b7934f0

---

Resolves [ALL-2747](https://linear.app/all-hands-ai/issue/ALL-2747/microagent-management-ui-develop-an-api-to-get-the-microagents-for-the)


openhands/server/routes/git.py
```diff
@@ -229,3 +239,268 @@ async def get_repository_branches(
...
+        List of microagents found in the repository's .openhands/microagents directory
+    """
+    if not provider_tokens:
+        logger.info(
+            f'Returning 401 Unauthorized - Git provider token required for user_id: {user_id}'
+        )
+        return JSONResponse(
+            content='Git provider token required. (such as GitHub).',
+            status_code=status.HTTP_401_UNAUTHORIZED,
+        )
```
malhotra5:
```suggestion
```

This is the confusing part of our codebase - we don't use provider tokens in cloud openhands (this the token the user passes to us)

We have two options, we can check `if (not provider_tokens and server_config.app_mode == AppMode.OSS) or (not user_id and server_config.app_mode == AppMode.SAAS)`

Alternatively, we can skip this logic all together and the ProviderHandler will throw an error which the try catch will handle

openhands/server/routes/git.py
```diff
@@ -229,3 +239,268 @@ async def get_repository_branches(
...
+
+    try:
```
malhotra5, line 514: NIT: instead of nested try catch maybe we can split them into separate chunks

openhands/server/routes/git.py
```diff
@@ -229,3 +239,268 @@ async def get_repository_branches(
...
+                    # App token format: username:app_password
+                    remote_url = f'https://{token_value}@{domain}/{repository_name}.git'
+                else:
+                    # Access token format: use x-token-auth
+                    remote_url = f'https://x-token-auth:{token_value}@{domain}/{repository_name}.git'
+            else:
+                # GitHub
+                remote_url = f'https://{token_value}@{domain}/{repository_name}.git'
+        else:
+            remote_url = f'https://{domain}/{repository_name}.git'
```
malhotra5: We have very similar logic in the [runtime](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/runtime/base.py#L604). Is there a way to dedup?

openhands/server/routes/git.py
```diff
@@ -229,3 +239,268 @@ async def get_repository_branches(
...
+
+def _get_default_domain(git_provider):
+    """Get the default domain for a git provider."""
+    from openhands.integrations.service_types import ProviderType
+
+    domain_map = {
+        ProviderType.GITHUB: 'github.com',
+        ProviderType.GITLAB: 'gitlab.com',
+        ProviderType.BITBUCKET: 'bitbucket.org',
+    }
```
malhotra5: NIT: this map also seems to be present in multiple parts of the codebase. Maybe we can have these maps and fetching logic be part of the `ProviderHandler` class

hieptl:
Hi @malhotra5, I updated the code based on your feedback. Thank you for your feedback!

openhands/integrations/provider.py
```diff
@@ -399,3 +399,59 @@ async def get_branches(
...
+            raise Exception('Git provider authentication issue when getting remote URL')
+
+        provider = repository.git_provider
+        repo_name = repository.full_name
+
+        provider_domains = {
+            ProviderType.GITHUB: 'github.com',
+            ProviderType.GITLAB: 'gitlab.com',
+            ProviderType.BITBUCKET: 'bitbucket.org',
+        }
```
malhotra5: NIT: let's make this a class variable

openhands/runtime/base.py
```diff
@@ -381,8 +380,8 @@ async def clone_or_init_repo(
                 )
             return ''
 
-        remote_repo_url = await self._get_authenticated_git_url(
-            selected_repository, git_provider_tokens
+        remote_repo_url = await self.provider_handler.get_authenticated_git_url(
```
malhotra5, line 383: For this method, do we still need to `git_provider_tokens` param?

hieptl: Hi @malhotra5, to be honest, I don’t have a definitive answer yet as I’m still getting familiar with the product. I moved the get_authenticated_git_url function to the ProviderHandler to make it reusable, and updated all the places that were using it to ensure nothing breaks.

If there’s anything you’d like me to adjust or improve, just let me know—happy to update it! Thank you!

openhands/server/routes/git.py
```diff
@@ -229,3 +241,326 @@ async def get_repository_branches(
...
+
+    logger.info(f'Found {len(microagents)} microagents in {repository_name}')
+    return microagents
+
+
+@app.get(
+    '/repository/{repository_name:path}/microagents',
+    response_model=list[MicroagentResponse],
+)
+async def get_repository_microagents(
```
malhotra5, line 490: Would we run this method in parallel across multiple repos?

openhands/integrations/provider.py
hieptl: Hi @malhotra5, I updated the code based on your feedback. Thank you for your feedback!

openhands/runtime/base.py
malhotra5: No worries looks good for now

openhands/server/routes/git.py
hieptl: @malhotra5: According to the discussion from [this Slack thread](https://openhands-ai.slack.com/archives/C095K9Z52HH/p1752665114354569), on the Microagent Management page, the list of repositories will be displayed as expandable sections (accordions). When a repository is expanded, we’ll call this API to fetch microagents for that specific repository only, not all repositories.

Hope I understood your question correctly—please let me know if I missed anything!

malhotra5: Gotcha makes sense, was concerned about the time it would take to populate the information

tests/unit/test_get_repository_microagents.py
```diff
@@ -0,0 +1,821 @@
...
+                    'openhands.server.routes.git.subprocess.run',
+                    return_value=mock_result,
+                ):
+                    with patch(
+                        'openhands.server.routes.git.load_microagents_from_dir',
+                        return_value=(
+                            mock_repo_agents,
+                            mock_knowledge_agents,
+                        ),
+                    ):
```
malhotra5: Overall this PR LGTM!

Could we try to reduce the deeply nested patching? Also if an exception is thrown before the assert statements are reached will this test pass successfully?

hieptl: @malhotra5: I updated the code based on your feedback. If an exception is thrown before the assert statements are reached, the tests will fail. The test will raise the exception and pytest will mark it as failed.

Issue #9392, opened 2025-06-26
[CLI] Improve confirmation UX with better options (like Claude's arrow interface)


neubig:
## Problem
The current confirmation interface is limited and forces users into a cumbersome workflow when they want to provide alternative instructions instead of just yes/no.

## Current Behavior
```
Proceed with action? (y)es/(n)o/(a)lways >
```

When users want to give different instructions:
1. Type "N" + Enter
2. Wait for prompt
3. Type their alternative instructions

## Expected Behavior (inspired by Claude)
```
Proceed with action?
→ Yes, proceed
  No, let me give different instructions
  Always proceed (don't ask again)

Or type your alternative instructions directly:
> 
```

## User Experience Issues
- **Workflow interruption**: Must say no, then explain what to do instead
- **Inefficient**: Extra steps when user wants to provide guidance
- **Unclear options**: Limited to yes/no/always

## Proposed Enhanced UX
1. **Arrow navigation**: Use arrow keys to select options
2. **Direct input**: Allow typing instructions directly at confirmation prompt
3. **More options**: 
   - Yes, proceed
   - No, skip this action
   - Always proceed (don't ask again)
   - Let me provide different instructions
4. **Inline editing**: Type alternative instructions directly

## Example Enhanced Interface
```
Command
┌─────────────────────────────────────┐
│ $ rm important_file.txt             │
└─────────────────────────────────────┘

Choose an option:
→ Yes, proceed
  No, skip this action  
  Always proceed
  Give different instructions

Or type alternative instructions:
> Instead, move the file to backup/ folder
```

## Technical Context
This would require enhancing the `read_confirmation_input()` function in `openhands/cli/tui.py` to support:
- Arrow key navigation
- Multiple choice selection
- Direct text input as alternative

## Related Files
- `openhands/cli/tui.py` (read_confirmation_input function)
- `openhands/cli/main.py` (confirmation handling in on_event_async)

chrismahoney:
Are we particularly attached to prompt_toolkit for the CLI interface? Not trying to reinvent wheels, but there are a couple of Terminal UI-focused projects that might be worth checking out while the CLI is still young. If prompt_toolkit is locked in, that’s all good too. Solid library.

Charm is the TUI toolkit that I’ve been absorbing over time, the only downer is that it’s built with Go and usually that doesn’t jive with JS/Python shops.

[...]

PR #9758, opened 2025-07-17
feat: [CLI] 9392 cli improve confirmation ux

VasilisAvgoustakis:

User gets a multiple choice menu which can be navigated using the arrow keys instead of a simple prompt to which he was asked to answer with a single char 'y', 'n' or 'a'. Additionally a new option  that was not previously available is provided that gives the user the possibility to give alternative instructions for the next operation the agent shall perfom instead of what the agent currently was planning.

---
Modified read_confirmation_input() in ./openhands/cli/tui.py:
I used cli_confirm() which alread provides the required functionality instead of prompt_async(). In order to keep the outer coroutine responsive I passed cli_confirm() in async.to_thread().

Modified on_event_async() in ./openhands/cli/main.py:
Added conditional for 'edit' that corresponds to the new option of the user providing alternative instructions.

Modified openhands/core/schema/agent.py added USER_EDITED state for 'edit' option!

---
#9392 
grafik


.gitignore
```diff
@@ -130,6 +130,9 @@ env.bak/
 .env.bak
 venv.bak/
 *venv/
+.repomix/
+repomix-output.txt
+
```
xingyaoww: Can we remove these? Why are these necessary
```suggestion
```

openhands/cli/main.py
```diff
@@ -231,12 +231,26 @@ async def on_event_async(event: Event) -> None:
...
+                if confirmation_status in ('yes', 'always'):
                     event_stream.add_event(
                         ChangeAgentStateAction(AgentState.USER_CONFIRMED),
                         EventSource.USER,
                     )
-                else:
+                elif confirmation_status == 'edit':
+                    # Tell the agent the proposed action was rejected
+                    event_stream.add_event(
+                        ChangeAgentStateAction(AgentState.USER_EDITED),
```
xingyaoww: Can we just return USER_REJECTED for simplicity? This way we don't need a new state.

And "user want a different action" do sounds like "reject" to me

xingyaoww (commented):
Mostly LGTM!

.gitignore
VasilisAvgoustakis: oh sorry! These are created by the repomix VSC extension here locally. I will remove them.

openhands/cli/main.py
VasilisAvgoustakis: Absolutely right will do so! Thanks for the feedback!

xingyaoww (approved):
LGTM! Thank you

enyst:
I’m not sure if this is just me, but this seems to block Vscode/Windsurf. When I use the new action, and enter some text, the agent starts a step, and then everything got locked. I had to restart the IDE completely. Happened three times, every time after using the new 'Let me provide different instructions' action.

~Do we need a thread there~, or could we add some text to the existing “n” option? The existing “n” option does the same thing, it waits for the user to enter text.

openhands/cli/main.py
```diff
@@ -231,12 +231,26 @@ async def on_event_async(event: Event) -> None:
...
+                        EventSource.USER,
+                    )
+                    # Notify the user
+                    print_formatted_text(
+                        HTML(
+                            '<skyblue>Okay, please tell me what I should do instead.</skyblue>'
+                        )
+                    )
+                    # Solicit replacement isntructions
+                    await prompt_for_next_task(AgentState.AWAITING_USER_INPUT)
```
enyst, line 252: Do we need this? I think ‘no’ does the same, the action is rejected and the prompt returns to the user. That code is not on the `else` branch below, so do we really need it here?

enyst: Without this `await`, I think the execution would continue with, eventually, a prompt for the user; with this code though, it’s not obvious to me, could it be that it double-handles `REJECTED`, if that made sense?

VasilisAvgoustakis:
Hey @enyst thank you very much for the comments. I haven't had the issue with the blocked IDE myself while testing but I will happily try to recreate the problem and fix it as well as the other points you mention in the other comments. The thing is I will probably be on it again on Monday though...

VasilisAvgoustakis:
@enyst @xingyaoww sorry for the extra text but I am new in contributing in OSS and I am not sure what I should do next here! Should I check the potential issues that @enyst mentioned above or no? (Cause I see here that the branch is merged and closed)
And if yes should I just go and do it or are you going to open a new issue for it from which I can start my new PR?

xingyaoww:
Hey @VasilisAvgoustakis, we'd appreciate if you can start a follow-up PR fixing these :), but no pressure, in the worst case we can revert this PR to fix it

VasilisAvgoustakis:
> I’m not sure if this is just me, but this seems to block Vscode/Windsurf. When I use the new action, and enter some text, the agent starts a step, and then everything got locked. I had to restart the IDE completely. Happened three times, every time after using the new 'Let me provide different instructions' action.
> 
> ~Do we need a thread there~, or could we add some text to the existing “n” option? The existing “n” option does the same thing, it waits for the user to enter text.

I haven't be able to recreate this problem in my VSC but I will continue to try the next few days. @enyst Did you notice the block happening after getting the agent to try some specific kind of command or did you notice any other pattern in general that will help me to recreate the problem?

As far "adding some additional text to the 'n' option" is concerned I would leave that decision to you. As far as I can tell it is indeed the same thing since in both cases the agent keeps the context of the discussion so far.

enyst:
Okay, thank you. Happened consistently, no special cases as far as I saw.

What if we remove the await for a user prompt, nevertheless, does it still work normally for you? I think it should because it's the same case as the 'no' option?

VasilisAvgoustakis:
> Okay, thank you. Happened consistently, no special cases as far as I saw.
> 
> What if we remove the await for a user prompt, nevertheless, does it still work normally for you? I think it should because it's the same case as the 'no' option?

Yes it does seem to end up asking for user input anyway...so I leave it out as you suggested...it seems to work normally!

PR #9759, opened 2025-07-17
Add support for the groq hosted kimi-k2-instruct model in the functio…

regismesquita:
Add support for the Groq hosted kimi-k2-instruct model in the function calling capability list, following the same pattern as kimi-k2-0711-preview.


enyst (approved):
Thank you!

I think maybe we could go through the list and simplify some, because if I got this right, "kimi-k2" actually might be enough to match both, for example, and I doubt it would risk being confused with something where it must be disabled?

regismesquita:
For investigation, it is breaking when using Groq with tool calling, this happened when using openhands both via LiteLLM Proxy and calling Groq directly:

Final part of the request:
```

  {
      "role": "user",
      "content": [
        {
          "text": "1+1?",
          "type": "text"
        }
      ]
    },
    {
      "role": "assistant",
      "content": [],
      "tool_calls": [
        {
          "id": "0",
          "type": "function",
          "function": {
            "name": "think",
            "arguments": "{\"thought\":\"The user is asking a simple math question: 1+1. This is straightforward and doesn't require any code execution or complex analysis. I should just provide the answer directly.\"}"
          }
        }
      ]
    },
    {
      "name": "think",
      "role": "tool",
      "content": [
        {
          "text": "Your thought has been logged.",
          "type": "text"
        }
      ],
      "tool_call_id": "0"
    }
  ],
  "temperature": 0
}
```
Error:
```
litellm.BadRequestError: GroqException - {"error":{"message":"'messages.5' : for 'role:assistant' the following must be satisfied[('messages.5.content' : one of the following must be satisfied[('messages.5.content' : value must be a string) OR ('messages.5.content' : minimum number of items is 1)])]","type":"invalid_request_error"}}
. Received Model Group=moonshotai/kimi-k2-instruct
```

enyst:
I see, thank you.

I think it's possible that forcing string serializer for Groq like we do for DeepSeek [here](https://github.com/All-Hands-AI/OpenHands/blob/1512a36b7a09970bf22a4971877b8b179e104250/openhands/llm/llm.py#L820) will work, without more headaches here. I know the support for "list serializer" (sending `content` as a list of strings, as opposed to sending `content` as a single string) is new in Groq, and the error message is clear: it wants either a string or a list with at least 1 element (it gets `[]`).

Maybe all litellm has to do is adapt these empty lists to an empty string. Maybe!

Or maybe the empty `content` that comes with a tool call should be dropped completely when sent back. I like this less, we used to try to do it in the past because of Bedrock, and it caused an endless array of side effects. (shouldn't apply to another provider, but also not when via litellm proxy, except actually sometimes urgh)

enyst:
Thank you!

PR #9760, opened 2025-07-17
Fix: Add navigation to conversation page after clicking Launch button on task suggestions

neubig:
## Issue
When clicking the "Launch" button on suggested tasks, the conversation was being created successfully in the backend, but the user was not being automatically navigated to the conversation page in the frontend.

## Solution
- Added the `useNavigate` hook from react-router to the TaskCard component
- Implemented an `onSuccess` callback in the `createConversation` call that navigates to the new conversation page using the conversation ID from the response
- Added a test to verify the navigation behavior

## Testing
- Created a test that verifies navigation occurs after clicking the Launch button
- Manually verified the fix works as expected
- Successfully built the application to ensure there were no compilation errors

This fix ensures that when a user clicks the "Launch" button on a suggested task, they will be automatically navigated to the conversation page, providing a seamless user experience.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/cae3949c7f1d4d979215b01c00d6562d)

### DIFF
--- a/frontend/__tests__/components/features/home/tasks/task-card.test.tsx
+++ b/frontend/__tests__/components/features/home/tasks/task-card.test.tsx
@@ -1,0 +1,100 @@
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { TaskCard } from "#/components/features/home/tasks/task-card";
+import { SuggestedTask } from "#/components/features/home/tasks/task.types";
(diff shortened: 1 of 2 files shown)


neubig:
@amanape I think this is a bug, but if it's expected behavior please tell me!

If it's expected behavior, then maybe at least we should change the link to "open conversation" (instead of automatically navigating).

amanape (approved):
Yes this was a bug, thanks!

PR #9764, opened 2025-07-17
feat(backend): Develop an API to fetch conversations by repository and conversation_trigger.

hieptl:

To support microagent creation, we initiate conversations with specific instructions—similar to the standard conversation flow.

As noted in [this Slack thread](https://openhands-ai.slack.com/archives/C095K9Z52HH/p1752665114354569), we need a way to filter conversations that originate specifically from the Microagent Management UI.

To achieve this, we should implement a new API that fetches conversations based on:
- The selected repository
- An optional conversation_trigger parameter

When conversation_trigger is set to "microagent_management", it indicates that the conversation was initiated via the Microagent Management interface.

**Acceptance Criteria:**

- A new API is available to fetch conversations filtered by:
  - Selected repository
  - Optional conversation_trigger parameter
- If conversation_trigger = "microagent_management", only conversations triggered from the Microagent Management UI are returned.

---

The PR provides an API to fetch conversations by repository and conversation_trigger.

https://github.com/user-attachments/assets/98efeb31-ce69-417b-9a7e-93862a91e17d

---

Resolves [ALL-2761
](https://linear.app/all-hands-ai/issue/ALL-2761/microagent-management-develop-an-api-to-fetch-conversations-by)


openhands/storage/conversation/file_conversation_store.py
```diff
@@ -95,6 +98,48 @@ async def search(
         next_page_id = offset_to_page_id(end, end < num_conversations)
         return ConversationMetadataResultSet(conversations, next_page_id)
 
+    async def search_by_filters(
```
malhotra5: Actually i think we should implement this method outside of the file conversation store

There are also other store implementations that exist. Maybe see if we can reuse the existing [`search_conversation`](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/server/routes/manage_conversations.py#L220) endpoint but with the added constraints you're looking for

hieptl: Hi @malhotra5, I have updated the code based on your feedback. Thank you for your feedback!

openhands/storage/conversation/file_conversation_store.py
```diff
@@ -68,8 +71,9 @@ async def search(
         self,
         page_id: str | None = None,
         limit: int = 20,
+        selected_repository: str | None = None,
+        conversation_trigger: ConversationTrigger | None = None,
```
malhotra5: I think these filters can be implemented in the `search_conversations` endpoint rather than here

malhotra5: The file store would paginate conversations but the filtering requirements would be handled by the endpoint - it would help enforce separation of concerns and we'd also avoid implementing the filtering logic for other types of conversation stores (aside from the file store)

hieptl: I've updated the code based on your feedback. Thank you for your feedback!

malhotra5 (approved):
LGTM! In the future we may want to move the filtering logic into the store itself but not blocking imo

Other stores (e.g sql based ones) would be better optimized that way, and we could also have stronger assumptions about the number of conversations returned from the `search_conversations` endpoint

PR #9765, opened 2025-07-17
feat(backend): Support CreateMicroagent in the “Create New Conversation” API

hieptl:

To create a new microagent, we need to initiate a conversation and pass in the appropriate instructions—similar to the existing flow on the conversation page.

As discussed in [this Slack thread](https://openhands-ai.slack.com/archives/C095K9Z52HH/p1752665114354569), we need a way to identify and handle conversations that originate from the Microagent Management UI.

To support this, we can introduce a new parameter: CreateMicroagent.

This would be a strongly-typed object (e.g., CreateMicroagent) that includes required attributes such as repo: string, and possibly others (e.g., owner, branch, etc.).

If this object is included in the create_conversation API request, the backend should automatically infer that the conversation_trigger is microagent_management.

**Acceptance Criteria:**

- A new optional parameter CreateMicroagent is supported by the Create New Conversation API.
- The CreateMicroagent object includes at least the following required field:
  - repo: string
- If the CreateMicroagent object is present in the request:
  - The conversation_trigger is automatically set to "microagent_management".
- Existing behavior for other conversation types remains unchanged.

---

The PR supports `CreateMicroagent` in the “Create New Conversation” API.

---

Resolves [ALL-2760](https://linear.app/all-hands-ai/issue/ALL-2760/microagent-management-support-conversation-trigger-as-an-optional).


openhands/server/routes/manage_conversations.py
```diff
@@ -85,6 +85,7 @@ class InitSessionRequest(BaseModel):
     replay_json: str | None = None
     suggested_task: SuggestedTask | None = None
     conversation_instructions: str | None = None
+    conversation_trigger: ConversationTrigger | None = None
```
malhotra5: We avoid doing this because the new conversation api is public; users should not be able to manipulate the conversation trigger value.

Is there a way to automatically detect that the trigger is `microagent_management` based on other params?

hieptl: @malhotra5: Instead of using the conversation_trigger parameter, I could introduce a new one—perhaps called created_by. This parameter would be stored in the conversation metadata.

Then, on the Microagent Management page, we could call a separate API that filters conversations based on this metadata field.

What do you think about this approach?

malhotra5: Is there a specific prompt template we will use for `microagent_management`? Maybe we could detect the trigger based off that?

Created_by would function very similarly to ConversationTrigger, so I'd like a method that utilizes the latter

malhotra5: Also we have an param `suggested_task: SuggestedTask | None = None`; we detect that the trigger is suggested task because we have a strongly typed param passed from the FE for it

Maybe we could do something similar for `microagent_management` if there are other params which get passed to the BE for it?

hieptl: @malhotra5 I checked the backend code and I think we could introduce a new `suggested task type` called `create_microagent`. If the suggested task type is set to `create_microagent`, we could automatically set the `conversation_trigger` to `create_microagent` as well.

Therefore, we do not need to support conversation_trigger as an optional parameter anymore.

What do you think about this approach?

https://github.com/All-Hands-AI/OpenHands/blob/5654e251a82786976e9932f3d6b55937469ebc8d/openhands/integrations/service_types.py#L19

malhotra5: Awesome thanks for digging in

hieptl: I’ve converted the PR to draft and will update the Linear ticket and PR title accordingly. I’ll let you know once it’s ready for review again. Thank you so much for your help!

hieptl: @malhotra5: I updated the code based on our discussion. Thank you!

openhands/server/routes/manage_conversations.py
```diff
@@ -129,8 +130,11 @@ async def new_conversation(
     conversation_trigger = ConversationTrigger.GUI
 
     if suggested_task:
-        initial_user_msg = suggested_task.get_prompt_for_task()
-        conversation_trigger = ConversationTrigger.SUGGESTED_TASK
+        if suggested_task.task_type == TaskType.CREATE_MICROAGENT:
+            conversation_trigger = ConversationTrigger.MICROAGENT_MANAGEMENT
```
malhotra5: Hmm not sure this would work, the suggested task model expected attribs like `issue_number` and `title` to exists; these also aren't relevant for microagent creation

It's clever making `create_microagent` a type of `suggested_task`! But then its odd to have its own type of trigger. 

I'm thinking one of two solutions

1. Modify the `SuggestedTask` class to have some optional params, implement the prompt template for create microagent, and have `create_microagent` be a `TaskType`. Then instead of having a separate trigger for `microagent_management`, we can track the `TaskType` as a separate field

2. Alternatively similar to `SuggestedTask` we can have a strongly typed object `CreateMicroagent` (which requires attribs like `repo: str`, etc); if this object is present in the params for the create convo endpoint, then automatically infer that a separate trigger is `microagent_management`

hieptl: Hi @malhotra5, I’ve updated the code and followed the second solution you suggested. Thank you so much for your help!

PR #9779, opened 2025-07-17
Fix MCP tool timeout causing agent to stall indefinitely

### LINKED ISSUE #9777: [Bug]: MCP Tool Timeout Causes Agent to Stall Indefinitely
## Description

When an MCP tool call times out, the agent stalls indefinitely instead of properly handling the error and continuing execution.

## Steps to Reproduce

1. Call an MCP tool that will timeout (e.g.,  with a long-running query)
2. Observe that the frontend stalls after showing 'Calling MCP Tool: tavily_tavily-search'
3. In the logs, you'll see: 

## Expected Behavior

The agent should handle the timeout gracefully, return an appropriate error observation to the agent, and continue execution.

## Actual Behavior

The agent stalls indefinitely, requiring a page refresh or restart to continue.

## Technical Details

The issue occurs in the  function in  which doesn't properly handle McpError exceptions when they occur during tool calls.

## Impact

This issue prevents agents from recovering from MCP tool timeouts, leading to a poor user experience where the agent appears to be stuck.

neubig:
This PR fixes an issue where MCP tool timeouts cause the agent to stall indefinitely.
The problem occurs when an MCP tool call times out, raising a McpError exception
that is not caught, preventing the agent from continuing its work.

## Problem
When an MCP tool call times out, it raises a McpError exception that is not caught in the call_tool_mcp function. This causes the agent to stall indefinitely because:
1. The exception propagates up and prevents the agent from receiving an observation
2. The agent controller remains in a state where it's waiting for the observation
3. No further processing can occur

## Solution
The fix:
1. Adds try/except block in call_tool_mcp to catch McpError exceptions
2. Returns an error observation instead of propagating the exception
3. This allows the agent to continue processing after a timeout

## Tests
Added tests:
- test_mcp_tool_timeout_error_handling: Verifies that MCP errors are properly handled
- test_mcp_tool_timeout_agent_continuation: Verifies the agent can continue after a timeout

Fixes #9777

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/a3d25fcdcbca4e23a4a9f18d934bdd3f)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:220af0f-nikolaik   --name openhands-app-220af0f   docker.all-hands.dev/all-hands-ai/openhands:220af0f
```

### DIFF
--- a/openhands/mcp/utils.py
+++ b/openhands/mcp/utils.py
@@ -5,14 +5,14 @@ if TYPE_CHECKING:
     from openhands.controller.agent import Agent
 
 
+from mcp import McpError
+
 from openhands.core.config.mcp_config import (
     MCPConfig,
     MCPSHTTPServerConfig,
     MCPSSEServerConfig,
 )
 from openhands.core.logger import openhands_logger as logger
-from mcp import McpError
-
 from openhands.events.action.mcp import MCPAction
 from openhands.events.observation.mcp import MCPObservation
 from openhands.events.observation.observation import Observation
@@ -192,11 +192,7 @@ async def call_tool_mcp(mcp_clients: list[MCPClient], action: MCPAction) -> Obse
     except McpError as e:
         # Handle MCP errors by returning an error observation instead of raising
         logger.error(f'MCP error when calling tool {action.name}: {e}')
-        error_content = json.dumps({
-            'isError': True,
-            'error': str(e),
-            'content': []
-        })
+        error_content = json.dumps({'isError': True, 'error': str(e), 'content': []})
         return MCPObservation(
             content=error_content,
             name=action.name,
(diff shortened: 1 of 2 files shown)


neubig:
LGTM

openhands/mcp/utils.py
```diff
@@ -5,6 +5,8 @@
     from openhands.controller.agent import Agent
 
 
+from mcp import McpError
```
malhotra5, line 8:
```suggestion
from fastmcp.exceptions import FastMCPError 
```

Could we try this from FastMCP instead?

malhotra5: EDIT: I think FastMCP still responds with `McpError` but it would be good to double check since our tool calls are based around FastMCP

malhotra5: Okay I dug into this `McpError` is the correct here

neubig: I had OpenHands confirm.

PR #9787, opened 2025-07-18
Fix issue #9786: Prevent LLM settings from being accessible in SaaS mode

### LINKED ISSUE #9786: [Bug]: In SaaS mode, clicking the settings button twice reveals LLM settings

In SaaS mode, LLM setting should be hidden. However, clicking the settings button twice reveals the LLM settings. This should be fixed.

https://github.com/user-attachments/assets/a29cd23d-ef1f-4447-8bba-f7ee2b76b58b

### OpenHands Installation

app.all-hands.dev

### OpenHands Version

_No response_

### Model Name

_No response_

### Operating System

None

_No response_

neubig:
## Description

This PR fixes issue #9786 where clicking the settings button twice in SaaS mode would reveal the LLM settings, which should be hidden in SaaS mode.

## Changes

- Modified the `clientLoader` function in `settings.tsx` to properly handle redirects from LLM settings in SaaS mode
- Added a constant `OSS_ONLY_PATHS` to clearly define which paths should be redirected in SaaS mode
- Updated the test to verify that LLM settings are not accessible in SaaS mode

## Testing

- Added a test case to verify that in SaaS mode, the LLM settings are not accessible
- Ran the test suite to ensure no regressions were introduced

Fixes #9786

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/2793d1b73d554fd2a67b4032e9c312e4)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:222d8ef-nikolaik   --name openhands-app-222d8ef   docker.all-hands.dev/all-hands-ai/openhands:222d8ef
```

### DIFF
--- a/frontend/__tests__/routes/settings-saas-mode.test.tsx
+++ b/frontend/__tests__/routes/settings-saas-mode.test.tsx
@@ -1,0 +1,109 @@
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import { QueryClientProvider } from "@tanstack/react-query";
+import { MemoryRouter } from "react-router";
+import SettingsScreen from "#/routes/settings";
+import OpenHands from "#/api/open-hands";
+
+// Mock the i18next hook
+vi.mock("react-i18next", async () => {
+  const actual =
+    await vi.importActual<typeof import("react-i18next")>("react-i18next");
+  return {
+    ...actual,
+    useTranslation: () => ({
+      t: (key: string) => {
+        const translations: Record<string, string> = {
+          SETTINGS$NAV_INTEGRATIONS: "Integrations",
+          SETTINGS$NAV_APPLICATION: "Application",
+          SETTINGS$NAV_CREDITS: "Credits",
+          SETTINGS$NAV_API_KEYS: "API Keys",
+          SETTINGS$NAV_LLM: "LLM",
+          SETTINGS$TITLE: "Settings",
+        };
+        return translations[key] || key;
+      },
+      i18n: {
+        changeLanguage: vi.fn(),
+      },
+    }),
+  };
+});
+
+// Mock react-router
+vi.mock("react-router", async () => {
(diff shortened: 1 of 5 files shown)


neubig:
@openhands you might want to revert everything and start over. Think very deeply about why the llm settings tab can still show up after clicking the settings button twice. Write a test that demonstrates how this happens by writing a test that clicks the button twice in saas mode and ends up on the llm page. until you can write this test, do not perform any fixes

neubig:
@openhands
revert the unnecessary change that checks for the "/settings/" since this is not the path that is used.

enyst:
Oof, I'm really surprised that this "hiding" was only client-side. I was assuming AH has some routing in place and filters stuff you don't want.

I think this FE PR will be a clear improvement, so that users don't inadvertently mess up their settings.

But it will still be possible, via API, documented or otherwise. I'd suggest to rethink the server side of `Settings` for the cloud, e.g.
- maybe add a middleware (some filters?)
- check before saving in the database! I'm really surprised that, according to user reports, it seems you saved.
- use a cloud-specific subclass of `Settings` (which doesn't set/save things you don't want)
- and/or we can use some creation pattern so that the cloud can have its own implementation without hardcoding.

Please let's not add more saas things to the OpenHands backend/server though.

We can think it over (the above are examples), and on OpenHands side we can offer a way to extend/plug in a cloud-specific implementation, but not implement it itself. 

Maybe it doesn't matter anymore for this particular issue, since I seem to understand you're re-opening LLM soon for user's choice (or only some of it?). Nevertheless, I think it matters for other options, for example [this PR](https://github.com/All-Hands-AI/OpenHands/pull/9360) which is blocked by the uncertainty about what directly affects the cloud. I would love to know your thoughts!

PR #9793, opened 2025-07-18
Fix timestamp corruption in EventStream secret replacement

xingyaoww:

Fixed a critical bug where secret replacement in event streams was corrupting timestamps, causing "Invalid isoformat string" errors. The system now properly protects timestamp and other system metadata fields from secret replacement while maintaining security for user content.

---

This PR fixes a bug in `EventStream._replace_secrets()` where secret values were being replaced in ALL string fields, including system metadata like timestamps, event type identifiers, and system messages. This caused timestamp corruption when secrets appeared in ISO format datetime strings.

**Key Changes:**
- Modified `_replace_secrets()` to distinguish between top-level system metadata and nested user content
- Added `is_top_level` parameter to protect system fields only at the top level
- Expanded protected fields to include: `timestamp`, `id`, `source`, `cause`, `action`, `observation`, `message`
- Added comprehensive test suite with 4 test cases covering all scenarios

**Design Decision:** Only protect system fields at the top level, allowing secrets to still be replaced in nested user content and data structures. This maintains security while preserving system integrity.

**Impact:**
- ✅ Fixes "Invalid isoformat string" errors in event stream processing
- ✅ Maintains full secret replacement functionality for user content
- ✅ Protects system metadata integrity while preserving security
- ✅ No breaking changes to existing functionality

---

Fixes timestamp parsing issues where secret replacement breaks ISO format strings in OpenHands event stream processing.

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:21f6c06-nikolaik   --name openhands-app-21f6c06   docker.all-hands.dev/all-hands-ai/openhands:21f6c06
```


openhands/events/stream.py
```diff
@@ -219,10 +219,26 @@ def set_secrets(self, secrets: dict[str, str]) -> None:
...
+        self, data: dict[str, Any], is_top_level: bool = True
+    ) -> dict[str, Any]:
+        # Fields that should not have secrets replaced (only at top level - system metadata)
+        TOP_LEVEL_PROTECTED_FIELDS = {
+            'timestamp',
+            'id',
+            'source',
+            'cause',
+            'action',
+            'observation',
```
malhotra5, line 232: NIT: in the future we should rename this to `observation_type`; at first I thought this was the observation contents but we have a separate field `content` for all observation classes

tests/unit/test_event_stream.py
```diff
@@ -735,3 +737,129 @@ def test_cache_page_with_missing_events(temp_dir: str):
...
+    file_store = get_file_store('local', temp_dir)
+    stream = EventStream('test_session', file_store)
+
+    # Set up a secret
+    stream.set_secrets({'api_key': 'secret123'})
+
+    # Create an event with the secret in the command
+    action = CmdRunAction(
+        command='curl -H "Authorization: Bearer secret123" https://api.example.com'
+    )
```
malhotra5, line 753: I'm guessing if the agent is running a command run action with a secret its already too late (i.e secret has already gotten out to the event stream)?

malhotra5: Oh but the secret masking logic is run across all events types and affects their protected fields as well, got it

malhotra5 (approved):
LGTM!

malhotra5:
We seem to have a flaky bash session unit test

openhands/events/stream.py
xingyaoww: yeah we should maybe just do "event_type":"action:XXX" :( but it is probably a big change that needs some backward compatibility support..

PR #9800, opened 2025-07-19
Refactor LLM class to support runtime configuration updates

enyst:
## Overview

This PR implements the **first step** of a larger refactoring to prepare OpenHands for runtime LLMConfig reloading. It decouples the creation of partial functions from LLM initialization, enabling hot-swapping of LLM configurations without restart.

### Core Refactoring
- **Decouple partial function creation from LLM initialization**
  - Extract `_build_completion_function()` method for reusable completion setup
  - Extract `_rebuild_completion_wrapper()` method to recreate retry decorator wrapper
  - Modify `__init__` to call both build methods separately

### New Runtime Configuration API
- **Add `update_config(new_config: LLMConfig)` method** for hot-swapping LLM configuration at runtime
  - Updates internal configuration with deep copy for independence
  - Handles model changes by resetting model info for re-initialization
  - Preserves metrics and retry listener instances during config updates
  - Supports custom tokenizer updates and log completion folder creation
  - Rebuilds completion function and wrapper with new parameters

### Comprehensive Testing
- **Add 9 new unit tests** covering all update scenarios:
  - Basic configuration updates (model, temperature, tokens, etc.)
  - Model change scenarios (resets model info vs preserves it)
  - Custom tokenizer updates and removal
  - Log completion folder creation and validation
  - Completion function rebuilding
  - Metrics and retry listener preservation
  - Deep copy independence to prevent external modifications

### Before
```python
# Partial function created once in __init__, couldn't be changed
class LLM:
    def __init__(self, config):
        # ... setup ...
        self._completion_unwrapped = functools.partial(litellm.completion, **completion_kwargs)
        # ... wrapper setup ...
```

### After
```python
# Partial function can be rebuilt anytime via update_config()
class LLM:
    def __init__(self, config):
        # ... setup ...
        self._build_completion_function()
        self._rebuild_completion_wrapper()
    
    def update_config(self, new_config: LLMConfig):
        # Update config, rebuild functions, handle model changes
        self._build_completion_function()
        self._rebuild_completion_wrapper()
```

## Testing Status

✅ All existing unit tests (40 tests) still pass  
✅ All new unit tests (9 tests) pass  
✅ Pre-commit hooks pass  
✅ Backward compatibility maintained  

## Next Steps

This prepares the LLM class for the upcoming unified configuration system phases:

1. **Phase 2**: Add write API to OpenHandsConfig
2. **Phase 3**: Connect web UI to use application config  
3. **Phase 4**: Unify secrets access across all entry points

## Related Issues

Part of the larger configuration system unification effort to enable:
- Runtime configuration updates without restart
- Unified configuration API across CLI, web UI, and headless modes
- Hot-reloadable LLM instances
- Connected configuration systems instead of fragmented ones

---

**Note**: This is a draft PR for the first step. The refactoring maintains full backward compatibility while adding the foundation for runtime config reloading.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c8d856c-nikolaik   --name openhands-app-c8d856c   docker.all-hands.dev/all-hands-ai/openhands:c8d856c
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@llm-init openhands
```

### DIFF
--- a/tests/unit/llm/test_llm.py
+++ b/tests/unit/llm/test_llm.py
@@ -1567,3 +1567,89 @@ def test_llm_reinit_resets_cost_flag(default_config):
     # Same model, but reinit should reset the flag to True
     llm.reinit(copy.deepcopy(default_config))
     assert llm.cost_metric_supported is True
+
+
+def test_llm_reinit_thread_safety_with_inflight_completion(default_config):
+    """Concurrent reinit during an in-flight completion should not raise,
+    and subsequent completions should use the new config.
+    """
+    import threading
+    import time
+    from unittest.mock import patch
+
+    llm = LLM(default_config)
+
+    calls = []
+
+    def fake_completion(*args, **kwargs):
+        # Simulate provider latency and record the model used
+        calls.append(kwargs.get('model'))
+        time.sleep(0.2)
+        return {
+            'id': 'test-1',
+            'choices': [{'message': {'content': 'ok'}}],
+            'usage': {'prompt_tokens': 1, 'completion_tokens': 1},
+        }
+
+    with patch('openhands.llm.llm.litellm_completion') as mock_completion:
+        mock_completion.side_effect = fake_completion
+
+        # Start an in-flight completion with the initial model
+        t = threading.Thread(
+            target=llm.completion, args=([{'role': 'user', 'content': 'hi'}],)
+        )
+        t.start()
+
+        # Reinit while the completion is in-flight
+        time.sleep(0.05)
+        new_cfg = copy.deepcopy(default_config)
+        new_cfg.model = 'claude-3-5-sonnet-20241022'
+        llm.reinit(new_cfg)
+
+        t.join()
+
+        # Subsequent completion should use the new config
+        llm.completion(messages=[{'role': 'user', 'content': 'again'}])
+
+        # Ensure the latest call used the new model and metrics reflect it
+        assert len(calls) >= 1
+        assert calls[-1] == 'claude-3-5-sonnet-20241022'
+        assert llm.metrics.model_name == 'claude-3-5-sonnet-20241022'
+
+
+@patch('openhands.llm.llm.litellm_completion')
+def test_llm_reinit_provider_mappings(mock_completion, default_config):
+    """Reinit should apply provider-specific mappings (openhands proxy, azure max_tokens)."""
+    # Return a minimal, valid response
+    mock_completion.return_value = {
+        'id': 'call',
+        'choices': [{'message': {'content': 'ok'}}],
+        'usage': {'prompt_tokens': 1, 'completion_tokens': 1},
+    }
+
+    llm = LLM(default_config)
+
+    # 1) OpenHands provider rewrite to litellm_proxy
+    cfg_proxy = copy.deepcopy(default_config)
+    cfg_proxy.model = 'openhands/qwen3-coder'
+    llm.reinit(cfg_proxy)
+    llm.completion(messages=[{'role': 'user', 'content': 'x'}])
+
+    model_arg = mock_completion.call_args.kwargs.get('model')
+    base_url_arg = mock_completion.call_args.kwargs.get('base_url')
+    assert model_arg == 'litellm_proxy/qwen3-coder'
+    assert base_url_arg == 'https://llm-proxy.app.all-hands.dev/'
+
+    # 2) Azure mapping: max_completion_tokens -> max_tokens
+    mock_completion.reset_mock()
+    cfg_azure = copy.deepcopy(default_config)
+    cfg_azure.model = 'azure/gpt-4o'
+    cfg_azure.max_output_tokens = 777
+    cfg_azure.api_version = '2024-06-01'
+    llm.reinit(cfg_azure)
+    llm.completion(messages=[{'role': 'user', 'content': 'y'}])
+
+    kwargs = mock_completion.call_args.kwargs
+    assert kwargs.get('model') == 'azure/gpt-4o'
+    assert 'max_tokens' in kwargs and kwargs['max_tokens'] == 777
+    assert 'max_completion_tokens' not in kwargs


openhands/llm/llm.py
```diff
@@ -245,6 +257,81 @@ def __init__(
 
         self._completion_unwrapped = self._completion
 
+    def update_config(self, new_config: LLMConfig) -> None:
+        """Update the LLM configuration and rebuild the completion function.
+
+        This method allows for runtime configuration changes. When called, it will:
+        1. Update the internal configuration
+        2. Update metrics model name if the model changed
```
enyst: @malhotra5 Not urgent at all, just a question:

Let's say we have an LLM used as 'agent', and for some reason (routing, user select) we should update it.
This PR aims to change the configuration in the LLM instance, and not the LLM instance (it doesn't create LLM)

That would allow us to not need to update client code that uses this LLM instance, they'll just continue to have a reference to the same object (`self.llm` or whatever), and continue to call `completion`, so `_completion_cost`, just the same. It will actually call a different endpoint etc.

If we update `metrics` minimally, just the name, input/output costs, and that's it, does that give `ConversationStats` what it needs?

Or you prefer to 
- get a notification when it happens (similar with original register llm, we could do an update_listeners() which sends `llm_updated(service_id, llm_config)` so that it knows)
- re-initialize metrics for the new LLM name (I'm assuming `ConversationStats` can keep history, so it has the old model with its metrics)
- don't switch config, create the new LLM instead. (to update all clients, I guess we may want to make them listen too)

malhotra5: I think it depends on how much observability we want

If we want a breakdown of cost per model, then we should create a new LLM object and have it registered

If we want just the agent's cost (and no breakdown of cost per model used by the agent), then changing the internal configs would be fine

As far as the user is concerned, they just want the agent cost

For a researcher, however, I'd imaging they want the breakdown cost per model if they're experimenting with routing

enyst:
@xingyaoww This isn't ready, but I'd love your thoughts about doing it this way. It's preparation so that, if we want, we'd be able switch the configuration of an LLM instance.
e.g. 
- conversation started with Sonnet 4
- user selects somewhere their GPT-5 to be 'agent' LLM 
- => switch configuration in the LLM instance registered in LLMRegistry as 'agent' to GPT-5
- (the registry will notify observers, like `ConversationStats`, to update its metrics calculations/history)
- continue with GPT-5

xingyaoww (commented):
Interesting.. i actually wonder if we can use @malhotra5 's LLM registry to do this LLM switch? 👀

enyst:
Hi everyone — I’m openhands-gpt-5 (the automation agent helping on this PR).

Quick take on the proposal: this refactor aims to let an existing LLM instance safely reinitialize itself at runtime so the registry can switch a service’s model/config without forcing callers to adopt a new instance. This is not mutually exclusive with a registry-driven swap; the registry stays the orchestrator, and can choose either to create a new LLM or to reinitialize an existing one.

What the proposal changes in LLM
- Add a reinit method (update_config kept as an alias for now) so an LLM can rebuild itself from a new LLMConfig.
- Add a lock around reinit to avoid races with in-flight completions.
- Centralize initialization in a helper (initialize core: model info, tokenizer, completion partial, retry-decorated wrapper) and call it from both __init__ and reinit to avoid code duplication.
- On reinit: deep-copy the new config, refresh model info when the model changes, update tokenizer if needed, ensure the log folder exists when logging is enabled, rebuild the completion function and the retry wrapper, and update metrics.model_name. Also reset cost_metric_supported so the new model can report cost.

Why this works with the registry
- The LLMRegistry remains the place that decides a switch is needed (by service_id like agent, condenser, etc.).
- When switching, the registry can either create and register a brand new LLM (best for strict per-model observability) or call llm.reinit(new_config) on the existing instance (best for agent-level simplicity).

Follow-ups to make the proposal work end-to-end
1) LLMRegistry
- Add reinit_llm(service_id, new_config): look up the LLM for the service, call llm.reinit(new_config), then notify subscribers again.
- Optionally extend RegistryEvent with an event_type field (created or updated) so observers can react differently on a config update.
- Keep get_llm strict for now (raise if config mismatches) and document that callers should use reinit_llm to intentionally reconfigure a service.

2) ConversationStats
- Bug fix: in register_llm, after restoring, delete only the restored entry for the service (del self.restored_metrics[service_id]) rather than deleting the whole dict.
- On updated events, simplest path is to keep the same Metrics object and continue accumulating “agent-level” cost. If we later want per-model breakdown, we can either rotate to a new Metrics per update, or compute breakdown by grouping TokenUsage entries by model (they already carry the model).

3) Capabilities recompute
- Ensure function-calling and similar capability flags are recomputed during reinit even when the model does not change, so config-only toggles are picked up. Easiest is to compute these in the centralized initializer so both __init__ and reinit run it.

4) Tests
- Add tests for the registry reinit_llm path and observer notifications.
- Keep the current LLM tests (they already check tokenizer/logs/model-info/metrics updates on reinit).
- Optionally add a simple concurrency test to validate the lock prevents races.

Summary
- The proposal keeps the registry as the control plane and adds a safe in-place reinitialization path in the LLM. It doesn’t block the alternative “new LLM per switch” approach for better observability—both can coexist.
- I’m happy to push the small follow-ups (registry reinit method + event_type, ConversationStats bug fix, and the capability recompute placement) either in this PR or a quick follow-up, depending on reviewer preference.

Thanks! — openhands-gpt-5

enyst:
Hi folks — I’m openhands-gpt-5, the automation agent helping on this PR.

What this proposal is
- Let an existing LLM instance safely reinitialize itself at runtime so the registry can switch a service’s model/config without forcing callers to replace their LLM reference.
- This is not mutually exclusive with a registry-driven swap: the registry remains the orchestrator and can choose either path.

What the proposal changes in LLM
- Add a reinit method.
- Add a lock around reinit to avoid races with in‑flight completions.
- Centralize initialization in one helper (model info, tokenizer, completion partial, retry wrapper) and call it from both __init__ and reinit to avoid duplication.
- On reinit:
  - deep‑copy the new config, 
  - refresh model info when the model changes, 
  - update tokenizer if needed, 
  - ensure the log folder exists when logging is enabled, 
  - rebuild the completion and wrapper, 
  - update metrics.model_name, 
  - reset cost_metric_supported so the new model can report cost.

Why this works well with the registry
- The LLMRegistry decides when a service (agent, condenser, etc.) should switch.
- On switch, the registry can either:
  • create a new LLM (cleanest for per‑model observability), or
  • call llm.reinit(new_config) on the existing instance (simple, keeps agent‑level continuity).

Follow‑ups to make the proposal complete end‑to‑end
1) LLMRegistry
- Add reinit_llm(service_id, new_config): look up the LLM, call llm.reinit(new_config), then notify subscribers again.
- Optionally extend RegistryEvent with an event_type field (created | updated) so observers can react differently on updates.
- Keep get_llm strict (raise on config mismatch) and document using reinit_llm for intentional reconfiguration.

2) ConversationStats
- Bug fix: in register_llm, delete only the restored entry for that service (del self.restored_metrics[service_id]), not the whole dict.
- On updated events, default to keeping the same Metrics object (agent‑level totals). If we later want per‑model breakdown, we can either rotate Metrics per update or compute it from TokenUsage entries (they already carry the model).

3) Capability recompute
- Ensure function‑calling and similar capability flags are recomputed during reinit even when the model doesn’t change (do this in the centralized initializer so both __init__ and reinit pick it up).

4) Tests
- Add tests for the registry reinit_llm path and observer notifications.
- Keep/extend LLM tests (tokenizer/logging/model‑info/metrics updates on reinit).
- Optionally add a simple concurrency test to confirm the lock prevents races.

Summary
- The proposal keeps the registry as the control plane and adds a safe in‑place reinit path in LLM. It doesn’t block creating a new LLM for better observability — both can coexist.
- I can push the small follow‑ups (registry reinit method + event_type, ConversationStats bug fix, capability recompute placement, and tests) here or in a quick follow‑up PR — whichever you prefer.

Thanks! — openhands‑gpt‑5

enyst:
Hi — I’m openhands-gpt-5, the automation agent helping on this PR.

Question: 
> review our PR, the changes in llm.py. Do we really capture EVERYTHING that __init_ was doing, so that it's a refactoring, and nothing really changes in behavior if LLMs are only just initialized and used as before?

Short answer: 
Yes. For plain construction-and-use, behavior is preserved. The refactor centralizes what __init__ used to do into _initialize_core(), and __init__ still executes the same steps in the same effective order.

What __init__ did before vs now
- Metrics and flags
  - Still sets self.metrics (defaults to model_name=config.model).
  - Still sets self.cost_metric_supported = True.
  - Still deep-copies config.

- Logging setup
  - Still enforces log_completions folder requirement and creates it if enabled.

- Model info and derived config
  - Still calls init_model_info() early, which:
    - Logs model info
    - Applies HF top_p normalization
    - Populates max_input_tokens/max_output_tokens from model_info when unset
  - Function-calling capability is still computed right after model-info is initialized. It was computed inline before; now we compute it via _compute_function_calling_active(). We also call the same helper from _initialize_core(), which is redundant but harmless and keeps behavior identical.

- Tokenizer
  - Still loads custom tokenizer via create_pretrained_tokenizer when set; else sets tokenizer=None.

- Completion pipeline
  - Still builds the litellm completion partial with the same parameters and special cases:
    - openhands/... -> litellm_proxy rewrite
    - reasoning model handling (thinking/reasoning_effort; temperature/top_p removed appropriately)
    - Azure max_tokens mapping
    - safety_settings for mistral/gemini
  - Still sets _completion_unwrapped and rebuilds the retry-decorated wrapper (self._completion) with the same RetryMixin settings.
  - Wrapper still handles:
    - modify_params toggle
    - prompt/response logging
    - non-native function-calling mocking
    - latency measurement
    - usage/cost recording, cache read/write tokens, and context window logging

What changed structurally, not behaviorally
- The above is now centralized in _initialize_core(), called by __init__.
- We added an RLock and a reinit() path, which doesn’t affect the first-time init behavior.

Small note
- We currently compute function-calling capability twice (once in init_model_info via the helper, and again in _initialize_core). That’s intentional for reinit robustness and is idempotent; it doesn’t change runtime behavior for regular init. We can drop one call later for tidiness if desired.

Verdict
- This is a refactor: initializing an LLM and using it as before should behave the same.

PR #9374, opened 2025-06-25
Support reasoning_content display in UI

neubig:
## Summary

This PR implements support for displaying reasoning content from LLM responses in the OpenHands UI, fixing issue #9370.

### Backend Changes
- **Added `reasoning_content` field** to all major action types:
  - `MessageAction`
  - `CmdRunAction` 
  - `IPythonRunCellAction`
  - `FileEditAction`
  - `FileReadAction`
  - `FileWriteAction`
- **Modified `response_to_actions` function** to extract `reasoning_content` from LLM responses
- **Added helper function `add_reasoning_content()`** to apply reasoning to the first action in multi-tool scenarios
- **Updated `__str__` and `__repr__` methods** for actions to display reasoning content when present

### Frontend Changes
- **Updated TypeScript interfaces** to include `reasoning_content` field for all action types
- **Created `ReasoningContent` component** with expandable UI using lightbulb icon
- **Modified `event-message.tsx`** to render reasoning content before messages and actions
- **Added helper function `hasReasoningContent()`** for type checking

### Testing
- **Added comprehensive backend tests** (`test_reasoning_content.py`, `test_reasoning_content_integration.py`) - 10 passing tests
- **Added frontend tests** (`reasoning-content.test.tsx`) - 5 passing tests
- **All tests verify** reasoning content extraction, preservation, and display

## Technical Details

- **Reasoning content is only applied to the first action** when multiple tool calls are made in a single LLM response
- **Backward compatibility maintained** - existing code continues to work with `reasoning_content` defaulting to `None`
- **UI design follows OpenHands patterns** with collapsible sections and consistent styling
- **Type safety ensured** with proper TypeScript interfaces and type guards

## Testing

```bash
# Backend tests
poetry run pytest tests/unit/test_reasoning_content.py tests/unit/test_reasoning_content_integration.py -v

# Frontend tests  
cd frontend && npm run test -- reasoning-content
```

All tests pass successfully.

## Screenshots

The reasoning content appears as an expandable section with a lightbulb icon above assistant messages and actions, providing users with insight into the LLM's thought process.

Fixes #9370

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/8d2403fb28fc45a0a18cf29169460a5d)

---

To run this PR locally, use the following command:

GUI with Docker:
```
[...]

PR #9809, opened 2025-07-19
fix grok4 return reasoning tokens with empty string stalling into awaiting mode during task execution

claysauruswrecks:

grok 4 no longer randomly goes idle when returning reasoning tokens with empty string.

---
MODELS_WITH_EMPTY_REASONING_RESPONSES keeps the agent from going into "awaiting user input" mode.

---

### DIFF
--- a/openhands/agenthub/codeact_agent/function_calling.py
+++ b/openhands/agenthub/codeact_agent/function_calling.py
@@ -44,10 +44,10 @@ from openhands.llm.llm import MODELS_WITH_EMPTY_REASONING_RESPONSES
 
 
 def combine_thought(action: Action, thought: str) -> Action:
-    if not hasattr(action, "thought"):
+    if not hasattr(action, 'thought'):
         return action
     if thought and action.thought:
-        action.thought = f"{thought}\n{action.thought}"
+        action.thought = f'{thought}\n{action.thought}'
     elif thought:
         action.thought = thought
     return action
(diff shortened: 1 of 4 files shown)


claysauruswrecks:
Correct! IIRC there's an empty string "command" in the prompts and grok either decides to use it and return '' to see more output or something, or reasoning tokens returned but '' for the message. Not 100% sure the concrete reason why the empty string is returned by grok, but seems to also happen after thinking for a while in cursor and it times out the grok4 thinking "waiting..." visual.

claysauruswrecks:
also saw some html encoded chars grok4 was returning, so I guess squash all the commits if you don't think we'll need them later

enyst:
Not sure about the first, because the empty command in tools is a tool call, for `execute_bash`, so I don't think it gets to this point in the code, it should be handled above. So I guess it's a 'reasoning' content 🤔 

What I wonder is: litellm, the library we use for making calls to LLM providers, does some processing to detect the reasoning from models, and returns it, I think, in a `reasoning_content` field. What if maybe we check for that?

Currently we don't read it, but [this PR](https://github.com/All-Hands-AI/OpenHands/pull/9374) should fix that: reads it and makes it a first class part of the events just like content.

claysauruswrecks:
grok4 doesn't return `reasoning_content`, just their quantity i think. ok, so the '' command means the model wants to see more context and history, so they should be shown the terminal output scrollback n lines or something?

claysauruswrecks:
I'm able to see the empty messages after I added metadata around each expandable, so could be it thinking about what to do next, but I'm losing the "work" of the reasoning tokens unless that is preserved for the next request in the convo. I was able to prompt grok4 to call the think tool with an OODA spec before every call which seemed to work until it said the instructions said to only call one tool at a time and it stopped.
image

claysauruswrecks:
Here is 4-Opus summary of the issue and the fix, could be off, but I'm unblocked at the moment from this workaround.

https://github.com/identikey/dcypher/blob/main/docs/openhands-fork-patches.md

enyst:
Thank you for the follow ups!

> grok4 doesn't return `reasoning_content`, just their quantity i think. 

Ah, OK, thank you. That’s different than others.😅 I looked it up in their docs, and it doesn’t say… anything else. Sigh. This also influences this:

> I'm losing the "work" of the reasoning tokens unless that is preserved for the next request in the convo.

Since we don’t receive those tokens, we can’t send them back. 🤔 There are a couple of ways that LLM providers “recognize” the previous convo (response ids, tool id in the request), but I didn’t yet find how does xAI do that.

> ok, so the '' command means the model wants to see more context and history, so they should be shown the terminal output scrollback n lines or something?

Yes, exactly! It shows more output from that command. (some really have a lot)

> I was able to prompt grok4 to call the think tool with an OODA spec before every call which seemed to work until it said the instructions said to only call one tool at a time and it stopped.

Ah, likely openhands/litellm at the moment default Grok-4 to non-native function calling, which means we use a prompting substitute. Maybe you could enable native function calling, either in env (`-e LLM_NATIVE_FUNCTION_CALLING=true`) or in `config.toml` like [this](https://github.com/All-Hands-AI/OpenHands/blob/95bda09cd9bfb8e6aaafadd7a0c576fb54514ad8/config.template.toml#L193). It should then support multiple tool calls too.

enyst:
I think this fix could work. Just to give you some context here, though, so Opus found several dicts in our LLM code, and it thought it was a pattern of how to do things, so it implemented with a new dict like those. That’s not wrong, the funny thing is that we intended those dicts to be just a quick fix here or there for stuff that `litellm` library, which does the actual LLM API calls, didn’t fix yet or didn’t implement yet, in other words, something temporary. 😅 The eternal story of software engineering: now the “temporary” things found their place in the code and won’t let go. 😂

We should be happy that LLMs respect the patterns in the code… hey, it’s what they should be doing best😅 

Let’s take this fix in. Maybe at some point xAI updates their API to send the reasoning content, then we can use that, one can dream!

openhands/agenthub/codeact_agent/function_calling.py
```diff
@@ -246,10 +247,28 @@ def response_to_actions(
...
     else:
+        content = str(assistant_msg.content) if assistant_msg.content else ''
+
+        # Check if this is a model that may return empty responses while reasoning
+        model_name = str(getattr(response, 'model', ''))
+        is_reasoning_model = any(
+            model in model_name for model in MODELS_WITH_EMPTY_REASONING_RESPONSES
+        )
+
+        # Don't wait for response if content is empty and it's a reasoning model
```
enyst:
```suggestion
        # Grok-4 can send the number of reasoning tokens, but NOT the reasoning content, nor content
        # Ref: https://github.com/All-Hands-AI/OpenHands/pull/9809
        # Don't wait for response if content is empty and it's Grok
```

openhands/agenthub/codeact_agent/function_calling.py
```diff
@@ -246,10 +247,30 @@ def response_to_actions(
...
+        # Check if this is a model that may return empty responses while reasoning
+        model_name = str(getattr(response, 'model', ''))
+        is_reasoning_model = any(
+            model in model_name for model in MODELS_WITH_EMPTY_REASONING_RESPONSES
+        )
+
+        # Grok-4 can send the number of reasoning tokens, but NOT the reasoning content, nor content
+        # Ref: https://github.com/All-Hands-AI/OpenHands/pull/9809
+        # Don't wait for response if content is empty and it's Grok
+        # This prevents getting stuck when reasoning models return empty responses while thinking
```
enyst:
```suggestion
        # This prevents getting stuck when Grok models return empty content while thinking
```

openhands/agenthub/codeact_agent/function_calling.py
```diff
@@ -246,10 +247,30 @@ def response_to_actions(
...
+        )
+
+        # Grok-4 can send the number of reasoning tokens, but NOT the reasoning content, nor content
+        # Ref: https://github.com/All-Hands-AI/OpenHands/pull/9809
+        # Don't wait for response if content is empty and it's Grok
+        # This prevents getting stuck when Grok models return empty content while thinking
+        wait_for_response = True
+        if is_reasoning_model and not content:
+            wait_for_response = False
+        elif content:  # For all other models, wait if there's content
```
enyst:
```suggestion
        elif content:
```

enyst (approved):

enyst:
@claysauruswrecks unit tests are failing, could you look into it?

claysauruswrecks:
Yes, I can take a look tonight. There are a few flaky tests I noticed. I can put flaky retry decorators on them.

mamoodi:
Now there's some conflicts :(

enyst:
@Openhands Read this PR carefully and understand what it aims to achieve.

Recently, a commit in main branch changed the location for this kind of variables in model_features.py. Track it down, understand it, and merge main into this branch in your local environment.

Then open another PR to this PR (switch branch to a new one, so you can open a new PR)

mamoodi:
Any chance we can get this PR ready to merge?

mamoodi:
Should we try to get this merged?

mamoodi:
@claysauruswrecks any chance you have time to resolve the conflicts so we can get this merged?

tofarr:
I'm going to close this because it is stale - if you feel that it should be reopened, please say so.

PR #9814, opened 2025-07-20
Fix CLI runtime invalid path error handling

### LINKED ISSUE #9813: [Bug]: CLI "invalid path" cause the agent to stop

When using the CLI, and "invalid path" causes the agent to stop. For instance, when the agent tried to perform an action outside of its permitted directory here:

```
Agent running... (Press Ctrl-P to pause)
13:28:48 - openhands:ERROR: base.py:348 - [runtime 9d0ea3c8-2fa4-4a1a-a43f-abc567c3444e-0c620f648141a83a] Unexpected error while running action: LLMMalformedActionError: Invalid path: /Users/gneubig/work/strudel. You can only work with files in /Users/gneubig/work/music-to-strudel.
13:28:48 - openhands:ERROR: base.py:349 - [runtime 9d0ea3c8-2fa4-4a1a-a43f-abc567c3444e-0c620f648141a83a] Problematic action: FileReadAction(path='/Users/gneubig/work/strudel', start=0, end=-1, thought='', action=<ActionType.READ: 'read'>, security_risk=None, impl_source=<FileReadSource.OH_ACI: 'oh_aci'>, view_range=None)
```

A better behavior would be that this error was passed back to the agent, so it could resume execution.

### OpenHands Installation

CLI

### OpenHands Version

0.49.1

### Model Name

_No response_

### Operating System

MacOS

_No response_

neubig:
## Summary

Fixes #9813: CLI "invalid path" cause the agent to stop

This PR fixes an issue where the CLI runtime would stop agent execution when encountering invalid file paths, instead of providing feedback to the agent so it can continue and potentially correct its behavior.

### Core Fix
- **Modified `openhands/runtime/base.py`**: Added special handling for `LLMMalformedActionError` in the `_handle_action` method
  - Import `LLMMalformedActionError` from core exceptions
  - Convert `LLMMalformedActionError` to `ErrorObservation` instead of treating it as a fatal runtime error
  - Other exceptions continue to be handled as before (causing runtime errors)

### Testing
- **Added `tests/unit/test_runtime_error_handling.py`**: Comprehensive tests to verify the fix
  - Test that `LLMMalformedActionError` is converted to `ErrorObservation`
  - Test both invalid absolute paths and path traversal attempts
  - Test that other exceptions still cause runtime errors as expected

## Problem Description

Previously, when an agent tried to access a file outside the workspace (e.g., `/Users/gneubig/work/strudel` when workspace is `/Users/gneubig/work/music-to-strudel`), the CLI runtime would:

1. Raise `LLMMalformedActionError` in `_sanitize_filename`
2. This error would bubble up to `base.py`'s `_handle_action` method
3. The error would be caught generically, logged, and cause the runtime to stop
4. No observation would be sent back to the agent
5. The agent would stop execution without feedback

## Solution

Now, when `LLMMalformedActionError` occurs:

1. The error is caught specifically in `_handle_action`
2. It's converted to an `ErrorObservation` with the error message
3. The observation is sent back to the agent via the event stream
4. The agent receives feedback about the invalid path and can continue execution

## Testing

All tests pass:
- ✅ New runtime error handling tests (3/3)
- ✅ Existing CLI workspace tests (6/6) 
- ✅ Agent controller tests continue to pass
- ✅ Pre-commit hooks pass

## Verification

The fix has been verified with a demo script that simulates the exact scenario from the issue. The agent now receives an `ErrorObservation` with the message:

```
Invalid path: /tmp/outside_workspace.txt. You can only work with files in /workspace/path.
```

This allows the agent to understand the constraint and potentially correct its behavior, rather than stopping execution entirely.

## Backward Compatibility

This change is fully backward compatible:
- Other exceptions continue to be handled exactly as before
- The agent controller already has proper handling for `LLMMalformedActionError` (converts to `ErrorObservation`)
- No changes to public APIs or interfaces

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/0cc6091674654e6e88b5dfa413e5ed9f)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:d209e26-nikolaik   --name openhands-app-d209e26   docker.all-hands.dev/all-hands-ai/openhands:d209e26
```


neubig:
@openhands review this PR. Do you think this is the best design, or would something else be better? In particular if similar types of permission errors or other errors are handled in a different place in the code base it would be good to handle all similar errors in a similar place.

enyst:
I was going to ask, do we really need to do this in base.py or is a solution in/with CLIRuntime possible?

The other runtimes are not where the problem in the reported issue happened. It happened in CLIRuntime, because I think that’s where it’s the code restricting the agent to a particular path. For other runtimes, things work well, and the issue isn’t happening.

What if we point this out, explicitly, to the agent?

neubig:
@enyst good question, but I think that it actually might be good to do it in the main agent loop. Currently this is only used in the CLI, but in the future we might want a version of the GUI that similarly can use the local file system, who knows? And catching permission errors in the agent loop seems to make sense.

I did change from LLMMalformedActionError to PermissionError, which makes more sense semantically though.

enyst:
Got it.

There’s another thing I don’t understand. Before this PR, we raised MalformedActionError in CLIRuntime. That should have arrived in the controller here:
https://github.com/All-Hands-AI/OpenHands/blob/bbfa37fd97a531478356186c6f9a9fdbf9fc341e/openhands/controller/agent_controller.py#L824

~where we create an `ErrorObservation`. I wonder if openhands agent can figure out and explain why it didn’t get there before? Maybe we don’t need actually, to do the logic in base.py, it’s … maybe not exactly duplicate, but kind of? Could we receive it there like every other exception that needs to go back to the agent as errorobs?~

Ah, scratch this, sorry. Of course, because it’s from the runtime, not from a step. Please ignore the noise. 😅

neubig:
I validated that this works in the CLI.

Screenshot 2025-07-21 at 4 32 17 PM

neubig:
@openhands run pre-commit on the files that were changed in this PR

tests/unit/test_runtime_error_handling.py
xingyaoww, line 1: I don't think we need this test?

xingyaoww (approved):
LGTM

PR #9819, opened 2025-07-21
fix(button): improve font-weight styling

mislavlukach:

Fix font-weight styling

---
Fix font-weight styling

---


openhands-ui/components/toast/Toast.stories.tsx
amanape, line 1: Modals pop up on every story when clicking on one notify button

openhands-ui/components/button/Button.tsx
amanape, line 1: Once a button is clicked, the font remains “active” variant even when hovering away. Should probably reset to the default font here

amanape (changes requested):
Thank you! Few comments

openhands-ui/components/toast/Toast.stories.tsx
mislavlukach: What do you mean? ToastManager is rendered directly in Toast.stories.tsx. 
So toast doesn't exist outside of this context

openhands-ui/components/button/Button.tsx
mislavlukach: When the button is active, font-weight is now back to 400.

amanape (approved):

PR #9824, opened 2025-07-21
feat: [CLI] 9392 cli improve confirmation ux - revisited

VasilisAvgoustakis:

User gets a multiple choice menu which can be navigated using the arrow keys instead of a simple prompt to which he was asked to answer with a single char 'y', 'n' or 'a'. Additionally a new option that was not previously available is provided that gives the user the possibility to give alternative instructions for the next operation the agent shall perfom instead of what the agent currently was planning.

---
Conditionally grouped 'n' and (former) 'edit' option in main.py and tui.py since the edit option did not have any difference as to how the agent was handling it compared to the 'n'. At the moment the user sees a 4th option but this is just mapped on the 'n' option aswell. If we prefer to leave the 'illusion' of a 4th option or add some text to the 2nd ('n') option to indicate to the user that he can continue providing instructions is still to be decided. 
Removed await prompt_for_next_task(AgentState.AWAITING_USER_INPUT) which potentially leading to double handling the USER_REJECTED state (during my tests I did not have any blocks but what I did notice before removing it is that I got the '>' agent awaiting input twice after selecting the 'edit' option, this is fixed now).
Removed test for 'edit' option since no longer relevant!

---
former 9392 - now closed
grafik

### DIFF
--- a/openhands/cli/tui.py
+++ b/openhands/cli/tui.py
@@ -705,7 +705,7 @@ async def read_confirmation_input(config: OpenHandsConfig) -> str:
         choices = [
             'Yes, proceed',
             'No (and allow to enter instructions)',
-            "Always proceed (don't ask again)"
+            "Always proceed (don't ask again)",
         ]
 
         # keep the outer coroutine responsive by using asyncio.to_thread which puts the blocking call app.run() of cli_confirm() in a separate thread
--- a/openhands/cli/tui.py
+++ b/openhands/cli/tui.py
@@ -714,7 +714,7 @@ async def read_confirmation_input(config: OpenHandsConfig) -> str:
             cli_confirm, config, 'Choose an option:', choices
         )
 
-        return {0: 'yes', 1: 'no', 2: 'always', 3: 'edit'}.get(index, 'no')
+        return {0: 'yes', 1: 'no', 2: 'always', 3: 'no'}.get(index, 'no')
 
     except (KeyboardInterrupt, EOFError):
         return 'no'
--- a/openhands/cli/tui.py
+++ b/openhands/cli/tui.py
@@ -704,9 +704,8 @@ async def read_confirmation_input(config: OpenHandsConfig) -> str:
     try:
         choices = [
             'Yes, proceed',
-            'No, skip this action',
-            "Always proceed (don't ask again)",
-            'Let me provide different instructions',
+            'No (and allow to enter instructions)',
+            "Always proceed (don't ask again)"
         ]
 
         # keep the outer coroutine responsive by using asyncio.to_thread which puts the blocking call app.run() of cli_confirm() in a separate thread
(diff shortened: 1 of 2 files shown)


openhands/events/stream.py
```diff
@@ -219,10 +219,26 @@ def set_secrets(self, secrets: dict[str, str]) -> None:
     def update_secrets(self, secrets: dict[str, str]) -> None:
         self.secrets.update(secrets)
 
-    def _replace_secrets(self, data: dict[str, Any]) -> dict[str, Any]:
+    def _replace_secrets(
+        self, data: dict[str, Any], is_top_level: bool = True
+    ) -> dict[str, Any]:
+        # Fields that should not have secrets replaced (only at top level - system metadata)
+        TOP_LEVEL_PROTECTED_FIELDS = {
```
enyst: Is this intended, is it part of the PR?

VasilisAvgoustakis: eehhmm not really, no, I never worked on or did any edits on this file. I did sync my feature branch with the upstream main before my last push though ... could this have come from there somehow?

tests/unit/test_event_stream.py
VasilisAvgoustakis, line 1: And the same goes for this file here ... never opened it in my IDE : / ... sorry for the confusion!

openhands/events/stream.py
enyst: Oh, okay, it must some merge issue

VasilisAvgoustakis: most likely right?
Should I do something?

enyst: Ah, found it:
- https://github.com/All-Hands-AI/OpenHands/pull/9793

Maybe update again from main, but I don't think it's a problem

openhands/cli/tui.py
```diff
@@ -601,7 +601,7 @@ async def read_confirmation_input(config: OpenHandsConfig) -> str:
             cli_confirm, config, 'Choose an option:', choices
         )
 
-        return {0: 'yes', 1: 'no', 2: 'always', 3: 'edit'}.get(index, 'no')
+        return {0: 'yes', 1: 'no', 2: 'always', 3: 'no'}.get(index, 'no')
```
enyst: Is the option 3 the same as 1 ?

VasilisAvgoustakis: At the moment yes. If we decide to 'merge' option 1 and 3 by adding some context text for the user to option 1 ('n') then of course this mapping is not needed.

openhands/events/stream.py
VasilisAvgoustakis: Allright will fetch and rebase from upstream main then...

VasilisAvgoustakis: I think those changes are now gone!

openhands/cli/tui.py
enyst: Let's merge it, FWIW I think in Claude Code there are 3 options, 'no' is `'no (and allow to enter instructions)'`, something like that

VasilisAvgoustakis: Allright then I will reduce it to three options and 'merge' the 'n' and 'enter further instructions' options. But it has to happen tomorrow, I'll be off the computer for the rest of the day (unfortunately) :)

enyst (approved):
Thank you!

PR #9851, opened 2025-07-22
Moved monitoring of last_execution_time to system_stats

tofarr:

Fixes issue where every URL hit is counted as an action_execution, which was causing inaccurate idle time tracking. Now, only actual action executions will reset the idle time counter.

---

This PR modifies the idle time tracking mechanism to only update the last execution time when an actual action is executed, rather than on every API request. This provides more accurate tracking of when the system is truly idle versus when it is just receiving regular health check or server info requests.

Specifically:
1. Moved the idle time tracking logic to the system_stats module
2. Added update_last_execution_time() calls in action execution code paths
3. Removed update_last_execution_time() calls from general request handling
4. Added comprehensive test coverage for the new behavior

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:af5dfc7-nikolaik   --name openhands-app-af5dfc7   docker.all-hands.dev/all-hands-ai/openhands:af5dfc7
```


tofarr:
Added comprehensive test coverage for the idle time tracking changes:

1. **Unit tests for system_stats.py**:
   - Tests for get_system_info() function
   - Tests for update_last_execution_time() function
   - Tests to verify idle time increases without updates
   - Tests to verify idle time calculation accuracy

2. **Tests for server health endpoints**:
   - Tests to verify server_info endpoint returns correct data
   - Tests to verify server_info endpoint does not update idle time

3. **Tests for action execution**:
   - Tests to verify action execution client updates idle time
   - Tests to verify action execution server updates idle time
   - Tests to verify update_last_execution_time() is called in finally blocks

4. **Integration tests**:
   - Tests to verify idle time increases without action execution
   - Tests to verify idle time resets after action execution
   - Tests to verify multiple server_info requests do not reset idle time

All tests are passing, providing good coverage for the new behavior.

openhands/runtime/impl/action_execution/action_execution_client.py
```diff
@@ -328,6 +329,8 @@ def send_action_for_execution(self, action: Action) -> Observation:
                 raise AgentRuntimeTimeoutError(
                     f'Runtime failed to return execute_action before the requested timeout of {action.timeout}s'
                 )
+            finally:
+                update_last_execution_time()
```
xingyaoww, line 333: Why are we updating these in both server and client?

tests/integration/test_idle_time_integration.py
xingyaoww, line 1: We seems to add too much tests at once. Can we consolidate them together into just one or two files, preferably to test files that already exists?

xingyaoww (commented):
Mostly LGTM! Just a question and a comment about tests..

openhands/runtime/impl/action_execution/action_execution_client.py
tofarr: * client is used in the local runtime.
* server is used on non local runtime (docker / non nested remote)

tests/integration/test_idle_time_integration.py
tofarr: I got openhands to do this and I think it went a bit overboard - I'll trim it.

openhands/runtime/impl/action_execution/action_execution_client.py
xingyaoww: make sense! thanks!

xingyaoww (approved):

PR #9865, opened 2025-07-23
Refactor get_microagents_from_org_or_user error handling

### LINKED ISSUE #9864: Refactor get_microagents_from_org_or_user

two issues with get_microagents_from_org_or_user

1.  it throws an error when a user doesn't have a ".openhands" repo in their github org, which is too high of an error level. it should say "org-level microagent directory xxx/.openhands not found" as a debug message
2. there are many over-broad catches of "Exception". We should fully understand which exceptions we expect to be thrown, and catch these specific exceptions.

The expected behavior is that:
* when the user opens a repo within an org, a debug message with the appropriate string is thrown
* otherwise, standard behavior is not changed

neubig:
## Summary

This PR refactors the `get_microagents_from_org_or_user` function to address the issues described in #9864.

## Changes Made

1. **Improved Error Logging**: Changed error-level logging to debug-level logging when org-level microagent directories don't exist
2. **Specific Exception Handling**: Replaced broad `Exception` catches with specific `AuthenticationError` handling
3. **Better Debug Messages**: Added appropriate debug messages that clearly indicate when org-level microagent directories are not found
4. **Comprehensive Tests**: Added thorough tests to verify the new error handling and logging behavior

## Issues Fixed

- When a user doesn't have a `.openhands` repo in their GitHub org, the function now logs a debug message instead of throwing an error
- Replaced over-broad `Exception` catches with specific exception handling for `AuthenticationError` and other exceptions
- The expected behavior is now implemented: when the user opens a repo within an org, a debug message with the appropriate string is logged if the org-level microagent directory doesn't exist

## Testing

- All existing tests continue to pass
- Added new comprehensive tests in `tests/unit/test_microagent_error_handling.py` that verify:
  - `AuthenticationError` is logged at debug level with appropriate message
  - Generic exceptions are handled correctly
  - Insufficient repository parts are handled properly

## Verification

- Pre-commit hooks pass successfully
- All related tests pass
- No breaking changes to existing functionality

Fixes #9864

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/f752dc8a5feb4d8fb9750b9f8f9c9d24)

### DIFF
--- a/tests/unit/test_microagent_error_handling.py
+++ b/tests/unit/test_microagent_error_handling.py
@@ -1,0 +1,192 @@
+"""Tests for microagent error handling and logging behavior."""
+
+import tempfile
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from openhands.core.config import OpenHandsConfig, SandboxConfig
+from openhands.events import EventStream
+from openhands.integrations.service_types import AuthenticationError
+from openhands.runtime.base import Runtime
+
+
+class MockRuntime(Runtime):
+    """Mock runtime for testing error handling."""
+
+    def __init__(self, workspace_root: Path):
+        # Create a minimal config for testing
+        config = OpenHandsConfig()
+        config.workspace_mount_path_in_sandbox = str(workspace_root)
+        config.sandbox = SandboxConfig()
+
+        # Create a mock event stream
+        event_stream = MagicMock(spec=EventStream)
+
+        # Initialize the parent class properly
+        super().__init__(
+            config=config, event_stream=event_stream, sid='test', git_provider_tokens={}
+        )
+
+        self._workspace_root = workspace_root
+        self._logs = []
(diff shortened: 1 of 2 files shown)


tests/unit/test_microagent_error_handling.py
xingyaoww, line 1: nit: can we move this to "test_microagent.py" maybe so we don't have too many separate files for test?

xingyaoww (approved):
Seems safe enough!

tests/unit/test_microagent_error_handling.py
```diff
@@ -0,0 +1,192 @@
+"""Tests for microagent error handling and logging behavior."""
```
rbren: I usually don't complain about new tests but these seem a bit silly 😂

neubig:
Thanks! Deleted the tests.

PR #9870, opened 2025-07-23
feat: Optimize git change detection with performance improvement and multi-repository support

tofarr:

This update significantly improves the performance of git change detection in OpenHands workspaces. Users will experience much faster file change detection (performance improvement) when working with git repositories. Additionally, OpenHands now supports detecting changes across multiple git repositories within a single workspace, making it more suitable for monorepo and multi-project workflows.

---

This PR introduces major optimizations and enhancements to the git change detection system:

1. **Performance Optimization**: Replaced multiple individual git commands with a single optimized bash command that processes all repositories in one execution.

2. **Multi-repository Support**: Enhanced `get_git_changes()` to scan all direct subdirectories of the workspace for git repositories, allowing OpenHands to work effectively with workspaces containing multiple projects.

3. **Single Command Architecture**: Refactored the implementation to use a single bash command that:
   - Lists all potential repository directories (current + subdirectories)
   - Checks each for git repository status
   - Extracts file changes using `git status --porcelain`
   - Returns structured data with repository path, file status, and file path

4. **Enhanced Data Structure**: Added `repository` field to change objects to identify which repository each change belongs to, while maintaining backward compatibility.

5. **Workspace Directory Consistency**: Always uses the workspace mount path as the working directory for consistent behavior across different runtime environments.

**Key Design Decisions:**
- Used bash scripting for cross-platform compatibility and performance
- Maintained backward compatibility by preserving existing `status` and `path` fields
- Added comprehensive test coverage for multi-repository scenarios
- Optimized for remote execution environments where minimizing command count is crucial

**Testing Scenarios**
* `export INIT_GIT_IN_EMPTY_WORKSPACE=0` and `Launch from Scratch`. Ask the agent to clone some repos
* `export INIT_GIT_IN_EMPTY_WORKSPACE=1` and `Launch from Scratch`. Ask the agent to clone some repos
* Launch with a repo/branch.

---

This addresses performance issues with git change detection in large workspaces and adds support for multi-repository workflows that are common in modern development environments.

@tofarr can click here to [continue refining the PR](https://app.all-hands.dev/conversations/d90a762b015c4aa1b11cd1f3b485fc82)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:e12460c-nikolaik   --name openhands-app-e12460c   docker.all-hands.dev/all-hands-ai/openhands:e12460c
```

### DIFF
--- a/openhands/runtime/mcp/proxy/manager.py
+++ b/openhands/runtime/mcp/proxy/manager.py
@@ -8,6 +8,7 @@ including initialization, configuration, and mounting to FastAPI applications.
 import logging
 from typing import Any, Optional
 
+from anyio import get_cancelled_exc_class
 from fastapi import FastAPI
 from fastmcp import FastMCP
 from fastmcp.utilities.logging import get_logger as fastmcp_get_logger
@@ -89,9 +90,29 @@ class MCPProxyManager:
         if not self.proxy:
             raise ValueError('FastMCP Proxy is not initialized')
 
+        def close_on_double_start(app):
+            async def wrapped(scope, receive, send):
+                start_sent = False
+
+                async def check_send(message):
+                    nonlocal start_sent
+                    if message['type'] == 'http.response.start':
+                        if start_sent:
+                            raise get_cancelled_exc_class()(
+                                'closed because of double http.response.start (mcp issue https://github.com/modelcontextprotocol/python-sdk/issues/883)'
+                            )
+                        start_sent = True
+                    await send(message)
+
+                await app(scope, receive, check_send)
+
+            return wrapped
+
         # Get the SSE app
         # mcp_app = self.proxy.http_app(path='/shttp')
-        mcp_app = self.proxy.http_app(path='/sse', transport='sse')
+        mcp_app = close_on_double_start(
+            self.proxy.http_app(path='/sse', transport='sse')
+        )
         app.mount('/mcp', mcp_app)
 
         # Remove any existing mounts at root path


openhands/runtime/utils/git_handler.py
```diff
@@ -75,124 +76,176 @@ def _verify_ref_exists(self, ref: str) -> bool:
...
+        # Single bash command that finds the closest git repository to the file and gets the ref content
+        cmd = f"""bash -c '
+        # Convert to absolute path
+        file_path="$(realpath "{file_path}")"
+
+        # Find the closest git repository by walking up the directory tree
+        current_dir="$(dirname "$file_path")"
+        git_repo_dir=""
+
+        while [[ "$current_dir" != "/" ]]; do
```
raymyers, line 104: Could we use:
```
git rev-parse --show-toplevel
```

raymyers: Oh I see that's used later... I don't really get why both methods are needed, I must not be understanding something

openhands/runtime/utils/git_handler.py
```diff
@@ -75,124 +76,176 @@ def _verify_ref_exists(self, ref: str) -> bool:
...
+        done
+
+        # If no git repository found, exit
+        if [[ -z "$git_repo_dir" ]]; then
+            exit 1
+        fi
+
+        # Get the file path relative to the git repository root
+        repo_root="$(cd "$git_repo_dir" && git rev-parse --show-toplevel)"
+        relative_file_path="${{file_path#${{repo_root}}/}}"
```
raymyers, line 119: An alternative but wouldn't be any better if we already need the root 
```
# Get the file path relative to the git repository root
git ls-files --full-name <file>
```

openhands/runtime/utils/git_handler.py
```diff
@@ -75,124 +76,176 @@ def _verify_ref_exists(self, ref: str) -> bool:
...
 
-        changes_list = self._get_changed_files()
-        result = parse_git_changes(changes_list)
+        # Single bash command that:
+        # 1. Creates a list of directories to check (current dir + direct subdirectories)
+        # 2. For each directory, checks if it's a git repo and gets status
+        # 3. Outputs in format: REPO_PATH|STATUS|FILE_PATH
+        cmd = """bash -c '
+        {
+            # Check current directory first
```
raymyers, line 204: Maybe we could do a python glob to simplify this?

```
 Path(root_dir).rglob('.git', max_depth=max_depth
```

tofarr: With DockerRuntime / RemoteRuntime the code is running on a completely different machine. Anything we run will have to be through `runtime.run_action` - and that incurs a penalty of about a second - regardless of the function. (This is why we instruct OpenHands to group multiple commands together where possible)

openhands/runtime/utils/git_handler.py
tofarr: The issue is that sometimes we need to handle nested git directories. Suppose we have a workspace containing a newly initialized git repo:
`/workspace/.git`

Then we do:
`git clone https://github.com/All-Hands-AI/OpenHands.git`.

We'll end up with:
`/workspace/OpenHands/.git`

Any `git status` / `git diff` we run will need to be in the context of the `/workspace/OpenHands` directory (Not `/workspace`!) 

Otherwise it will look like everything is a newly added file with all new lines.

openhands/runtime/utils/git_handler.py
raymyers: Gotcha. So it sounds like we've got a pattern where we don't have a crossplatform way to execute anything multi-step and that's a problem. This is the kind of thing Ansible, Chef, etc.. had to solve. It's outside this PR though. I think there are ways to do it in shorter bash but that's not a big deal.

openhands/runtime/utils/git_handler.py
raymyers: Ah ok. Git really should have an option for that but it doesn't.

raymyers (approved):

PR #9877, opened 2025-07-23
feat: Add configurable runtime support for issue resolver and fix: Kubernetes pod naming limits

ckavili:

  This PR introduces two key improvements for OpenHands:

  1. **Configurable Runtime for Issue Resolver**: The issue resolver now supports configurable runtime environments (docker, kubernetes, etc.) through command line arguments, config files, or environment variables, instead of being hardcoded to 'docker'.
   Users can now specify `--runtime kubernetes` when running the issue resolver.

  2. **Fixed Kubernetes Pod Naming Issues**: Resolved a critical bug where Kubernetes pod names could exceed the 63-character limit, causing pod creation failures. Session IDs are now optimized to ensure all Kubernetes resource names (pods, services, PVCs, ingresses) stay within the character limits.
---

  **Runtime Configuration Enhancement:**
  - Added `--runtime` argument to `resolve_issue.py` command line interface
  - Modified `IssueResolver.update_openhands_config()` to accept and process runtime parameter
  - Implemented fallback logic: CLI arg > config file > default ('docker')
  - Maintains full backward compatibility - existing scripts continue working unchanged

  **Session ID Length Optimization:**
  - Modified `generate_sid()` in `core/setup.py` to limit session IDs to exactly 32 characters
  - Implemented smart truncation logic:
    - Long session names (>16 chars): Use first 8 chars + 23-char hash
    - Short session names (≤16 chars): Use full name + remaining space for hash
  - Ensures Kubernetes resource names fit within 63-character limit:
    - Pod names: 50 chars (`openhands-runtime-` + 32-char session ID)
    - Service names: 54 chars (pod name + `-svc`)
    - Ingress names: 63 chars (pod name + `-ingress-code`)

  **Design Decisions:**
  1. **32-character limit**: Chosen to accommodate the longest Kubernetes resource suffix (`-ingress-code`) while staying within 63-char limit
  2. **Deterministic behavior**: Maintained same input → same output for session restoration
  3. **Backward compatibility**: Runtime defaults to 'docker', existing session restoration continues working
  4. **Configuration hierarchy**: Follows existing OpenHands pattern of CLI > env > config > defaults
 
---

### DIFF
--- a/openhands/core/setup.py
+++ b/openhands/core/setup.py
@@ -226,7 +226,7 @@ def create_controller(
 
 def generate_sid(config: OpenHandsConfig, session_name: str | None = None) -> str:
     """Generate a session id based on the session name and the jwt secret.
-    
+
     The session ID is kept short to ensure Kubernetes resource names don't exceed
     the 63-character limit when prefixed with 'openhands-runtime-' (18 chars).
     Total length is limited to 32 characters to allow for suffixes like '-svc', '-pvc'.
@@ -235,17 +235,17 @@ def generate_sid(config: OpenHandsConfig, session_name: str | None = None) -> st
     jwt_secret = config.jwt_secret
 
     hash_str = hashlib.sha256(f'{session_name}{jwt_secret}'.encode('utf-8')).hexdigest()
-    
+
     # Limit total session ID length to 32 characters for Kubernetes compatibility:
     # - 'openhands-runtime-' (18 chars) + session_id (32 chars) = 50 chars
     # - Leaves 13 chars for suffixes like '-svc' (4), '-pvc' (4), '-ingress-code' (13)
     if len(session_name) > 16:
         # If session_name is too long, use first 16 chars + 15-char hash for better readability
         # e.g., "vscode-extension" -> "vscode-extensio-{15-char-hash}"
-        session_id = f"{session_name[:16]}-{hash_str[:15]}"
+        session_id = f'{session_name[:16]}-{hash_str[:15]}'
     else:
         # If session_name is short enough, use it + remaining space for hash
         remaining_chars = 32 - len(session_name) - 1  # -1 for the dash
-        session_id = f"{session_name}-{hash_str[:remaining_chars]}"
-    
+        session_id = f'{session_name}-{hash_str[:remaining_chars]}'
+
     return session_id[:32]  # Ensure we never exceed 32 characters


openhands/core/setup.py
```diff
@@ -225,9 +225,26 @@ def create_controller(
...
 
     hash_str = hashlib.sha256(f'{session_name}{jwt_secret}'.encode('utf-8')).hexdigest()
-    return f'{session_name}-{hash_str[:16]}'
+    
+    # Limit total session ID length to 32 characters for Kubernetes compatibility:
+    # - 'openhands-runtime-' (18 chars) + session_id (32 chars) = 50 chars
+    # - Leaves 13 chars for suffixes like '-svc' (4), '-pvc' (4), '-ingress-code' (13)
+    if len(session_name) > 16:
+        # If session_name is too long (e.g., full UUID), use first 8 chars + hash
+        session_id = f"{session_name[:8]}-{hash_str[:23]}"
```
enyst: Nit: we could do it like this, just wondering if we could keep more of session_name and less of the hash? e.g. keep, idk, all 16

enyst (approved):
LGTM, thank you! Left a tiny nitpicky comment

openhands/core/setup.py
enyst: (Personally, I use human-readable session names, e.g. "vscode-integration", "vscode-extension", "vscode-runtime", and 8 would chop them to "vscode-i", "vscode-e" and "vscode-r" followed by hash... and that's lucky lol 😅)

ckavili: ahh that makes perfect sense 🙈 let's go with 16 chars of session name + 15 chars of hash then..so it'd be 16 (session name) + 1 (dash) + 15 (hash) = 32 chars total..should be alright I hope..and thanks a lot for the review! 🙏

enyst: Sorry, @ckavili , I think this still breaks user (manual) workflows, 16 is not enough for e.g. `vscode-integration-1` ... `vscode-integration-7`, to still differentiate between sessions and load the correct one. It will load the first all the time...
I actually don't fully understand why 32, maybe we can get away with doing this the other way around: to make 50, we can set aside 8 for the hash, and let the rest, the `session_name` be everything else?

Or perhaps, don't change this at all, because if the user doesn't set `session_name` then it's not set. Did we really have issues because of this particular "session_name"? It's not default.

ckavili: I personally hit this issue when I set a custom session name that was a bit long and exceeded 63-char limit due hash + the suffixes but I definitely understand it'd be hard to distinguish something like `vscode-integrati-hash1` from `vscode-integrati-hash2`. Setting 8 char aside for the uniqueness makes sense to me.

regarding why 32 character, it is because it adds `openhands-runtime-` as a prefix [here](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/runtime/impl/kubernetes/kubernetes_runtime.py#L56). And also adds suffixes like `-svc` or `-ingress-code` there too.

Actually, now that I think about, we don't have to add suffixes to differentiate the objects as they are different k8s kinds. It can save us a couple of characters as well 🙈

Issue #9488, opened 2025-07-01
[Bug]: Chat Width Limitation in Chat Window


openhands-agent:
**Describe the bug and reproduction steps**
   The chat window does not take up the full width but remains fixed-width instead. According to discussions, this fixed-width implementation might be a leftover behavior from before the container resize feature was reintroduced, which aimed to address a horizontal scrolling bug. However, this is reducing the usability of the chat window interface.

**Potential solutions**
   Ensure that the chat window dynamically takes up the full width of the available space while preventing any unwanted horizontal scrolling issues.

**Logs, Errors, Screenshots, and Additional Context**
   A relevant screenshot was attached in the Slack message, but it could not be uploaded successfully to S3. Original Slack file link: [View Screenshot](https://files.slack.com/files-pri/T06P212QSEA-F093GDZGWMD/download/screenshot_2025-07-01_at_11.24.06___am.png)

**Issue Created By**: Stephan Psaras on Slack
**Link to Slack Thread**: https://openhands-ai.slack.com/archives/C06QT0AGY4W/p1751383471131569?thread_ts=1751383471.131569&cid=C06QT0AGY4W

PR #9895, opened 2025-07-24
fix: Chat Width Limitation in Chat Window 

JamesZ12b:
This PR fixs #9488, allow chat window to use full width.

### DIFF
--- a/frontend/src/components/features/chat/chat-message.tsx
+++ b/frontend/src/components/features/chat/chat-message.tsx
@@ -86,26 +86,26 @@ export function ChatMessage({
         />
       </div>
 
-        <div
-            style={{
-                whiteSpace: 'normal',
-                wordBreak: 'break-word',
-            }}
+      <div
+        style={{
+          whiteSpace: "normal",
+          wordBreak: "break-word",
+        }}
+      >
+        <Markdown
+          components={{
+            code,
+            ul,
+            ol,
+            a: anchor,
+            p: paragraph,
+          }}
+          remarkPlugins={[remarkGfm, remarkBreaks]}
         >
-            <Markdown
-                components={{
-                    code,
-                    ul,
-                    ol,
-                    a: anchor,
-                    p: paragraph,
-                }}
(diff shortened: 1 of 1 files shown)


Abubakar-01:
**PR Review** 
**Before(Main Branch):**
image

**After (PR Branch):**
image

**Conclusion:**
 :x: The bug still persist, and need to be fixed!

JamesZ12b:
I think you've got Main Branch and PR Branch image output backwards.
My After (PR Branch):
屏幕截图 2025-08-01 154251
屏幕截图 2025-08-01 154327
屏幕截图 2025-08-01 154341

Main Branch: 

屏幕截图 2025-08-01 154937

Abubakar-01:
@JamesZ12b, for me, the width remained constant in both cases — regardless of how much space was available, it wasn't utilizing it. Anyway, I'll take another look and let you know.

JamesZ12b:
Okay, since the previous problem was that if you manually break a line (e.g. a sql query), the fixed width causes the line to break automatically, I've fixed it so that it won't break automatically because of the fixed width, and the width of the dialog box here will be determined by the width of the longest line (take up the full width).

Abubakar-01:
@JamesZ12b  

> Okay, since the previous problem was that if you manually break a line (e.g. a sql query), the fixed width causes the line to break automatically, I've fixed it so that it won't break automatically because of the fixed width, and the width of the dialog box here will be determined by the width of the longest line (take up the full width).

Thanks for the clarification. The issue description wasn't very detailed previously. Here's what I’m seeing on the **PR branch:**
image

On **Main Branch**
image


**Conclusion:**  :white_check_mark:  PR fixes the issue described by @JamesZ12b

amanape:
Long inline code text causes overflow

image

JamesZ12b:
> Long inline code text causes overflow
> 
> image

Now I've fixed it, it ensures that even long strings without spaces are correctly linefeeding to prevent overflows, thanks!

JamesZ12b:
@amanape, now I've solved the problem, it was a CSS prioritization issue, thanks!
image

PR #9905, opened 2025-07-25
Do not override user's git config in CLI mode or local machine

enyst:

---

Fix the system prompt to keep the user's git config in place. We have reports on slack about it overriding the local user, thus people have to spend time re-configuring it.

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c9368ea-nikolaik   --name openhands-app-c9368ea   docker.all-hands.dev/all-hands-ai/openhands:c9368ea
```

### DIFF
--- a/openhands/agenthub/codeact_agent/prompts/system_prompt.j2
+++ b/openhands/agenthub/codeact_agent/prompts/system_prompt.j2
@@ -31,9 +31,9 @@ Your primary role is to assist users by executing commands, modifying code, and 
 </CODE_QUALITY>
 
 <VERSION_CONTROL>
-* You MUST keep the same git config in place, if it already exists. Only if git has not been configured on the machine, then you are permitted to configure it with "openhands" as the user.name and "openhands@all-hands.dev" as the user.email by default, unless explicitly instructed otherwise.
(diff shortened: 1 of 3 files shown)


enyst:
Or maybe we can completely remove this phrase? Pretty sure docker runtime runs git config itself. 🤔

enyst:
Ah, hmm. Personally I use a microagent to
- stop messing my git config (this works)
- add openhands as co-author to the commits it does (this is spotty, sometimes it works, sometimes it doesn't)
    - with the name OpenHands-Claude for Claude, OpenHands-Gemini for Gemini etc. and the same email.

Let me try. I think your proposal means to add --author to each commit, and the "committer" would remain the configured user.

enyst:
Done! How about this, @amanape ?

enyst:
I see we also address the root problem in non-CLI environments (the runtime override) in https://github.com/All-Hands-AI/OpenHands/pull/9942 .

I'll merge this, since it improves the local user experience and fixes the slack reports, but we may need to address the system prompts again in 9942 or after. Maybe actually we should go back to telling the agent to simply not mess with local git config, and perhaps have a similar runtime override. 🤔

Issue #9177, opened 2025-06-17
[Bug]: CLI mode with Docker can't load `config.toml` file


KennyDizi:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Whenever I try to use OpenHands in CLI mode with Docker, it always asks about LLM settings.

Here is my `config.toml` file

```toml
[core]
search_api_key = "tvly-***"
reasoning_effort = "high"
platform = "linux/amd64" # Can be linux/amd64 or linux/arm64

[llm]
api_key = "sk-proj-***"
# API base URL (For Headless / CLI only -  In Web this is overridden by Session Init)
#base_url = ""
provider = "openai" # Can be openai, anthropic, deepseek, openrouter
# Model to use. (For Headless / CLI only -  In Web this is overridden by Session Init)
model = "openai/o3"
```

Here is the command that I used to run OpenHands CLI mode with Docker:

```bash
docker run -it --rm --pull=always \
        -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.43.0-nikolaik \
        -e LOG_ALL_EVENTS=true \
        -e SANDBOX_USER_ID=$(id -u) \
        -e SANDBOX_VOLUMES=$SANDBOX_VOLUMES \
        -e RUNTIME_MOUNT=$RUNTIME_MOUNT \
        -e LLM_API_KEY=$LLM_API_KEY \
        -e LLM_PROVIDER=$LLM_PROVIDER \
        -e LLM_MODEL=$LLM_MODEL \
        -e AGENT_MEMORY_ENABLED=$AGENT_MEMORY_ENABLED \
        -e LLM_CACHING_PROMPT=$LLM_CACHING_PROMPT \
        -e AGENT_ENABLE_THINK=$AGENT_ENABLE_THINK \
        -e LLM_NUM_RETRIES=$LLM_NUM_RETRIES \
        -e AGENT_ENABLE_MCP=$AGENT_ENABLE_MCP \
        -e CORE_PLATFORM=$CORE_PLATFORM \
        -v /var/run/docker.sock:/var/run/docker.sock \
        -v ~/.openhands-state:/.openhands-state \
        -v $RUNTIME_MOUNT \
        -w /workspace \
        -p 3080:3080 \
        --add-host host.docker.internal:host-gateway \
[...]

PR #9911, opened 2025-07-25
Fix configuration precedence in CLI mode

enyst:
When the user has a config.toml with an LLM configuration, it will be used unless overridden by the -l parameter The -l parameter now correctly finds and prioritizes the specified LLM configuration regardless of what settings.json says

---

This PR proposes to fix the precedence order for CLI:
(from low to high)
settings.json < workspace config.toml < command line parameters.

Before, we set `settings.json` as highest priority, which didn't allow to select an LLM from command line, regardless of user definitions in `config.toml`.

The configuration precedence order in CLI mode:
- When the user has a config.toml with an LLM configuration, it will be used 
- unless overridden by the -l parameter
- the -l parameter now correctly finds and prioritizes the specified LLM configuration
- only if none of these exists, reads the LLM config from settings.json

---

Fix https://github.com/All-Hands-AI/OpenHands/issues/9177

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:2e6b7c7-nikolaik   --name openhands-app-2e6b7c7   docker.all-hands.dev/all-hands-ai/openhands:2e6b7c7
```

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -302,6 +302,7 @@ class LLM(RetryMixin, DebugMixin):
 
             # log the entire LLM prompt
             self.log_prompt(messages)
+            print(self.config.model)
 
             # set litellm modify_params to the configured value
             # True by default to allow litellm to do transformations like adding a default message, when a message is empty
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -302,7 +302,6 @@ class LLM(RetryMixin, DebugMixin):
 
             # log the entire LLM prompt
             self.log_prompt(messages)
-            print(self.config.model)
 
             # set litellm modify_params to the configured value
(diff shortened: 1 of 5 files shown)


enyst:
Test with
-  `settings.json` with Claude Sonnet-4
- `config.toml` in current directory defining many other LLMs
- CLI started with: `openhands -n test -l gemini-pro-ah`
=> the model in use is Gemini
image

openhands/llm/llm.py
```diff
@@ -302,6 +302,7 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
 
             # log the entire LLM prompt
             self.log_prompt(messages)
+            print(self.config.model)
```
```suggestion
```

PR #9924, opened 2025-07-26
fix(docker-runtime): adjust default port ranges to avoid Windows ephemeral ports

### LINKED ISSUE #9202: [Bug]: Windows ports not available

Sometimes, when creating a new conversation, I get the following error:

Internal Server Error ("ports are not available: exposing port TCP 0.0.0.0:50281 -> 127.0.0.1:0: listen tcp4 0.0.0.0:50281: bind: An attempt was made to access a socket in a way forbidden by its access permissions.")

### OpenHands Installation

Docker command in README

### OpenHands Version

0.44

### Operating System

WSL on Windows

Internal Server Error ("ports are not available: exposing port TCP 0.0.0.0:50281 -> 127.0.0.1:0: listen tcp4 0.0.0.0:50281: bind: An attempt was made to access a socket in a way forbidden by its access permissions.")

llamantino:

Fixes "ports are not available" error message on Windows

---

Changes default port ranges on Windows to avoid overlap with ephemeral ports (49152 - 65535)

---
Closes #9202

### DIFF
--- a/openhands/runtime/impl/docker/docker_runtime.py
+++ b/openhands/runtime/impl/docker/docker_runtime.py
@@ -44,6 +44,12 @@ VSCODE_PORT_RANGE = (40000, 49999)
 APP_PORT_RANGE_1 = (50000, 54999)
 APP_PORT_RANGE_2 = (55000, 59999)
 
+if os.name == 'nt':
+    EXECUTION_SERVER_PORT_RANGE = (30000, 34999)
+    VSCODE_PORT_RANGE = (35000, 39999)
+    APP_PORT_RANGE_1 = (40000, 44999)
+    APP_PORT_RANGE_2 = (45000, 49151)
+
 
 def _is_retryablewait_until_alive_error(exception: Exception) -> bool:
     if isinstance(exception, tenacity.RetryError):
--- a/openhands/runtime/impl/docker/docker_runtime.py
+++ b/openhands/runtime/impl/docker/docker_runtime.py
@@ -1,4 +1,5 @@
 import os
+import platform
 import typing
 from functools import lru_cache
 from typing import Callable
@@ -44,7 +45,7 @@ VSCODE_PORT_RANGE = (40000, 49999)
 APP_PORT_RANGE_1 = (50000, 54999)
 APP_PORT_RANGE_2 = (55000, 59999)
 
-if os.name == 'nt':
+if os.name == 'nt' or platform.release().endswith('microsoft-standard-WSL2'):
     EXECUTION_SERVER_PORT_RANGE = (30000, 34999)
     VSCODE_PORT_RANGE = (35000, 39999)
     APP_PORT_RANGE_1 = (40000, 44999)


enyst:
Could we reduce redundancy by choosing common supported ranges?

openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -44,6 +44,12 @@
 APP_PORT_RANGE_1 = (50000, 54999)
 APP_PORT_RANGE_2 = (55000, 59999)
 
+if os.name == 'nt':
```
li-boxuan: I am not sure if I understand this... aren't we talking about WSL scenario here? When this runs in WSL environment, python interpreter shall see the OS as "posix", no?

llamantino: Yes, you're right. Thanks for catching this. There's also no way to detect it from inside the Docker container. Should I just make those values the default for all platforms then?

Edit: Wait, there seems to be a way: the kernel release string ends with WSL2 (for example: 6.6.87.2-microsoft-standard-WSL2). Will that cover all cases, though?

llamantino: Wait, there seems to be a way: the kernel release string ends with WSL2 (for example: `6.6.87.2-microsoft-standard-WSL2`). Will that cover all cases though?

li-boxuan: Maybe try this? https://github.com/scivision/detect-windows-subsystem-for-linux/blob/main/is_wsl.py

llamantino: Yes, that's the check that script does. docker will only use WSL2 though, so one check should be enough.
So, you don't want me to change or merge the defaults, like @enyst suggested. You just want me to add a condition for WSL2, right?
Like this: `if os.name == 'nt' or platform.release().endswith('microsoft-standard-WSL2')`

li-boxuan: I am neutral, but in general changing only for WSL2 is safer.

Do we really need `os.name == 'nt'` check though? We don't have docker_runtime in native windows yet, but assuming we have, do we also need this adjustment?

llamantino: Since it's not a WSL2‑specific issue but a Windows‑specific one, I don't think it would hurt to keep it. Also, there's already an import os in the file. If that's fine, I'll just update that line then.

li-boxuan (approved):

openhands/runtime/impl/docker/docker_runtime.py
li-boxuan: Sounds good!

li-boxuan:
Conditional approval: assuming https://github.com/All-Hands-AI/OpenHands/pull/9924#discussion_r2234051101 is addressed

llamantino:
Fixed the Windows detection logic with a WSL2 check as requested.

llamantino:
Just an additional note for future reference:

The same issue can also occur on Linux/macOS, as there's some overlap there too. It's more frequent on Windows because it keeps connections in TIME_WAIT (thus locking the ports) for longer by default[1].

If you ever see an "address in use" error when a runtime container is started, this could likely be the reason.

[1] Source: defaults for [Windows](https://learn.microsoft.com/it-it/biztalk/technical-guides/settings-that-can-be-modified-to-improve-network-performance) (120s), [MacOS](https://kaiwern.com/posts/2022/08/11/benchmarking-http-server-stuck-at-16k-requests/#tcp-time_wait-state) (30s), [Linux](https://www.ibm.com/docs/en/linux-on-systems?topic=tuning-tcpip-ipv4-settings) (60s).  Also, on Linux the ephemeral range starts 'earlier', so collisions are less likely to happen.

PR #9913, opened 2025-07-25
Investigate and fix Gemini 2.5 Pro performance issues

enyst:

PR #9925, opened 2025-07-27
perf(gemini): Apply Gemini 2.5 Pro performance optimizations from PR 9913

enyst:
Hi! I'm Claude, an AI agent helping with this repository.

## Changes Applied

Ported performance optimizations from PR #9913 to improve Gemini 2.5 Pro speed by ~2.4x:

### `llm_config.py`
- Changed `reasoning_effort` default from `'high'` to `None`
- Added logic to set `reasoning_effort='high'` for non-Gemini models only
- Gemini models now use optimized thinking budget when `reasoning_effort` is `None`

### `llm.py`
- Implemented Gemini-specific thinking budget logic (128 tokens)
- Maps `reasoning_effort=None` and `'low'` to `thinking: {budget_tokens: 128}`
- Passes through `'medium'`, `'high'`, `'none'` values to API as `reasoning_effort`
- Added debug print for model name

## Performance Impact
- ~2.4x speedup for Gemini 2.5 Pro based on testing in original PR
- Maintains backward compatibility for other models
- No breaking changes to existing configurations

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -193,7 +193,20 @@ class LLM(RetryMixin, DebugMixin):
             self.config.model.lower() in REASONING_EFFORT_SUPPORTED_MODELS
             or self.config.model.split('/')[-1] in REASONING_EFFORT_SUPPORTED_MODELS
         ):
-            kwargs['reasoning_effort'] = self.config.reasoning_effort
+            # For Gemini models, only map 'low' to optimized thinking budget
+            # Let other reasoning_effort values pass through to API as-is
+            if self.config.model == 'gemini-2.5-pro':
+                if self.config.reasoning_effort == 'low':
+                    kwargs['thinking'] = {'budget_tokens': 128}
+                elif self.config.reasoning_effort is None:
+                    # Default optimized thinking budget when not explicitly set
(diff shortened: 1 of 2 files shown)


openhands/llm/llm.py
```diff
@@ -302,6 +315,7 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
 
             # log the entire LLM prompt
             self.log_prompt(messages)
+            print(self.config.model)
```
```suggestion
```

openhands/llm/llm.py
```diff
@@ -302,6 +319,7 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
```
```suggestion
```

tests/unit/test_gemini_performance_optimizations.py
xingyaoww, line 1: nit: can we put these tests to "test_llm.py" so we don't have one more test?

enyst: Can do! I was a bit unsure because test_llm.py has more than 1.2k lines of code 😂

Maybe better, we could consider making subdirectories for each module, and split large test files like test_llm.py in some reasonable way, in `tests/unit/llm/`. Gemini performance is not very reasonable as criterion for a new file.

PR #9940, opened 2025-07-28
fix(conf): add cause to re-raised value-error to keep context.

happyherp:

Developer only: Show error that prevented config.toml from being processed. 

---

Add the cause to the thrown exception. The effect of this is, that the exact reason the config was rejected is visible in the console. 

---


openhands/core/config/utils.py
```diff
@@ -224,9 +224,9 @@ def load_from_toml(cfg: OpenHandsConfig, toml_file: str = 'config.toml') -> None
             logger.openhands_logger.warning(
                 f'Cannot parse [sandbox] config from toml, values have not been applied.\nError: {e}'
             )
-        except ValueError:
+        except ValueError as e:
             # Re-raise ValueError from SandboxConfig.from_toml_section
-            raise ValueError('Error in [sandbox] section in config.toml')
+            raise ValueError('Error in [sandbox] section in config.toml', e)
```
xingyaoww: I think we probably need to do this?

```suggestion
            raise ValueError('Error in [sandbox] section in config.toml') from e
```

xingyaoww (approved):
forgtt to approve 🤦

PR #9946, opened 2025-07-28
feat: redesign the text input

hieptl:

We need to redesign the chat input on the conversation page.

It’s a heavy component and is likely the root cause of a performance issue. We should rebuild it in a simpler, more modular way and ensure we avoid any performance pitfalls in the process.

---

The PR redesigns the chat input on the conversation page. We can refer to the video below for more information.

**Video 1: Loading state**

https://github.com/user-attachments/assets/424f5641-7794-45d8-8324-fe01f09bf1a4

**Video 2: Send a long message**

https://github.com/user-attachments/assets/63670c2f-e118-45c3-b2fa-73710302fb5e

**Video 3: Stop a conversation**

https://github.com/user-attachments/assets/8ca7c036-5c83-44db-b580-9dcac711dc05

**Video 4: Resume a conversation**

https://github.com/user-attachments/assets/c2a07b11-a27f-42d4-ab9e-56ed0d1f9c27

**Video 5: Sending files**

https://github.com/user-attachments/assets/46ec3a13-2bf2-451d-8965-f6e03a954448

We’ve also addressed the performance issue. Previously, the chat input triggered a re-render of the entire <ChatInterface /> component on every keystroke.

After redesigning the chat input, this no longer occurs. You can refer to the video below for confirmation—we added a console.log("ChatInterface rendered!!!") inside the component. As shown in the video, the log does not appear, which confirms that <ChatInterface /> is no longer being re-rendered unnecessarily.

We can refer to the video below to understand the performance issue.

https://github.com/user-attachments/assets/2f24c76a-708c-4ec7-b5c1-eac8ece77b2a

We can refer to the video below for the output after handling the performance issue.

https://github.com/user-attachments/assets/2cba54ac-ff76-4243-b63b-aab4f71b38fb

---


frontend/__tests__/components/interactive-chat-box.test.tsx
```diff
@@ -80,49 +129,90 @@ describe("InteractiveChatBox", () => {
...
+    expect(textarea).toHaveTextContent("Hello, world!");
+
+    // Ensure the text content is properly set in the contenteditable div
+    fireEvent.input(textarea, { target: { textContent: "Hello, world!" } });
+
+    // Also set the textContent directly to ensure it's available for innerText
+    textarea.textContent = "Hello, world!";
+
+    // Trigger input event to ensure the component updates properly
+    fireEvent.input(textarea, { target: { innerText: "Hello, world!" } });
```
amanape: I think we can use `userEvent` directly here to simulate expected state changes instead

frontend/__tests__/components/interactive-chat-box.test.tsx
```diff
@@ -136,55 +226,61 @@ describe("InteractiveChatBox", () => {
...
 
     // Verify text input was not cleared
-    expect(screen.getByRole("textbox")).toHaveValue("test message");
-    expect(onChange).not.toHaveBeenCalledWith("");
+    expect(screen.getByTestId("chat-input")).toHaveTextContent("test message");
+
+    // Ensure innerText is properly set for the contenteditable div
+    const textarea = screen.getByTestId("chat-input");
+
+    fireEvent.input(textarea, { target: { innerText: "test message" } });
```
amanape: Same comment as above

frontend/src/components/features/chat/chat-add-file-button.tsx
```diff
@@ -0,0 +1,29 @@
...
+  handleFileIconClick: () => void;
+  disabled?: boolean;
+}
+
+export function ChatAddFileButton({
+  handleFileIconClick,
+  disabled = false,
+}: ChatAddFileButtonProps) {
+  return (
+    <div
```
amanape: Use `button` element instead of `div`

frontend/src/components/features/chat/chat-play-button.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+export interface ChatResumeAgentButtonProps {
+  onAgentResumed: () => void;
+}
+
+export function ChatResumeAgentButton({
+  onAgentResumed,
+}: ChatResumeAgentButtonProps) {
+  return (
+    <div onClick={onAgentResumed} data-testid="play-button">
```
amanape: `button`

frontend/src/components/features/chat/chat-stop-button.tsx
```diff
@@ -0,0 +1,13 @@
+import PauseIcon from "#/icons/pause.svg?react";
+
+export interface ChatStopButtonProps {
+  handleStop: () => void;
+}
+
+export function ChatStopButton({ handleStop }: ChatStopButtonProps) {
+  return (
+    <div onClick={handleStop} data-testid="stop-button">
```
amanape: `button`

frontend/src/components/features/chat/custom-chat-input.tsx
```diff
@@ -0,0 +1,379 @@
...
+      chatInputRef.current.style.overflowY = "auto";
+    }
+  }, []);
+
+  useEffect(() => {
+    if (chatInputRef.current && value !== undefined) {
+      chatInputRef.current.textContent = value;
+      autoResize();
+    }
+  }, [value, autoResize]);
```
amanape: This is good logic, do you think we can move it into a custom hook so that we can re use it elsewhere?

frontend/src/components/features/chat/custom-chat-input.tsx
```diff
@@ -0,0 +1,379 @@
...
+      const rect = range.getBoundingClientRect();
+      const inputRect = chatInputRef.current.getBoundingClientRect();
+
+      // Scroll to keep cursor visible
+      if (rect.top < inputRect.top) {
+        chatInputRef.current.scrollTop -= inputRect.top - rect.top + 5;
+      } else if (rect.bottom > inputRect.bottom) {
+        chatInputRef.current.scrollTop += rect.bottom - inputRect.bottom + 5;
+      }
+    }, 0);
```
amanape, line 256: I am a little confused as to what this code here does, could you explain?

amanape (changes requested):
Thanks! Can we also get rid of the old `ChatInput` and `react-textarea-autosize` since you implemented custom resize logic? Also, have you been able to test the chat interface performance on auto resize as well?

amanape:
Overflow if there is a very long single line

image

frontend/src/components/features/chat/custom-chat-input.tsx
hieptl: Hi @amanape, this code automatically scrolls the text area to keep the cursor visible when the text area grows taller as you type.

**The Problem**
- When a text area auto-resizes, the cursor can scroll out of view, making it hard to see where you're typing.

**The Solution**
- Finds where your cursor is positioned
- Checks if it's still visible in the text area
- If not, automatically scrolls the text area to show the cursor

**Why It's Needed**
- Without this, in some cases, users may lose track of where they're typing and have to manually scroll to find their cursor - creating a poor user experience.
- This pattern is standard in all rich text editors and text editing components.

hieptl:
Hi @amanape, I’ve updated the code based on your feedback—thank you! 🙏

> Have you been able to test the chat interface performance on auto-resize as well?

Yes, I tested it with a long message and included a video demonstrating the behavior in the PR description.

frontend/src/components/features/chat/custom-chat-input.tsx
```diff
@@ -37,6 +38,14 @@ export function CustomChatInput({
   const fileInputRef = useRef<HTMLInputElement>(null);
   const chatContainerRef = useRef<HTMLDivElement>(null);
 
+  // Use the auto-resize custom hook
+  const {
+    autoResize,
+    handleInput: handleAutoResizeInput,
+    handlePaste: handleAutoResizePaste,
+    handleKeyDown: handleAutoResizeKeyDown,
```
amanape: I feel like we’re giving hook more responsibility than it needs. For example, these other handlers may not be useful if we want to reuse the resize logic for another input. Maybe we can define these input / paste / keydown handlers here and have them call `autoResize` only? Do you think that is possible or are there some things I may be missing that support defining these handlers within the hook?

amanape:
Another tiny comment and I think we can merge this:
image

Can we make the clip and send icons at the top when the textarea height is increased?

frontend/src/components/features/chat/custom-chat-input.tsx
hieptl: I've updated the hook. Thank you for your feedback! 🙏

hieptl:
Hi @amanape, I've updated the code based on your feedback. Thank you for your feedback! 🙏

amanape (approved):
Huge! Thanks

PR #9952, opened 2025-07-29
[CLI] Add default `fetch` MCP server & update doc to require `uvx`

### LINKED ISSUE #9875: [CLI] Enable `fetch` MCP server by default

MCP is supported for CLI mode in https://github.com/All-Hands-AI/OpenHands/pull/9519. We should enable the `fetch` MCP server by default to make it in feature parity with web GUI.

**Additional context**

ryanhoangt:

Add default `fetch` MCP server & require `uvx` for CLI.

---

This PR is to enable `fetch` MCP server by default for CLI & update CLI doc to require `uvx`.

---

Fix #9875


mamoodi:
Doc changes look fine. 👍  Thanks!

ryanhoangt:
> Maybe we can also conditionally add MCP (we only add it when uv is installed)

This makes sense, but there's an edge case as we saw in this [Slack thread](https://openhands-ai.slack.com/archives/C078L0FUGUX/p1753385133506379), if users work on the OH repo, they will have the `fetch` tool loaded automatically from microagent, and then the CLI gets stuck. I looked a bit, but it seems there's no way to handle this stuck behavior more gracefully (it comes from the Transport class provided by `mcp` package). Maybe a better option is to just error out and enforce user to install `uv`? Wdyt?

xingyaoww:
@ryanhoangt hmm, can we check the MCP `command` arguments before loading them and throw an error if `uvx` / `npx` exists in args but they are not available?

openhands/cli/main.py
```diff
@@ -304,6 +308,20 @@ def on_event(event: Event) -> None:
...
+                command='uvx',
+                args=['mcp-server-fetch'],
+            )
+            if fetch_mcp_config not in runtime.config.mcp.stdio_servers:
+                runtime.config.mcp.stdio_servers.append(fetch_mcp_config)
+        else:
+            logger.warning(
+                'uvx is not installed or not found in PATH. '
+                'The fetch MCP server will be disabled.'
+            )
```
xingyaoww: Since you add the check in openhands/mcp/utils.py, i don't think we need this anymore?

openhands/core/config/mcp_config.py
```diff
@@ -178,8 +178,7 @@ def __eq__(self, other):
         if not isinstance(other, MCPStdioServerConfig):
             return False
         return (
-            self.name == other.name
-            and self.command == other.command
+            self.command == other.command
```
xingyaoww: Can we revert this if we remove the changes in openhands/cli/main.py

openhands/mcp/utils.py
```diff
@@ -91,6 +92,14 @@ async def create_mcp_clients(
 
     for server in servers:
         if isinstance(server, MCPStdioServerConfig):
+            # Validate that the command exists before connecting
+            if not shutil.which(server.command):
+                logger.warning(
```
xingyaoww:
```suggestion
                logger.error(
```

xingyaoww (commented):
Mostly LGTM!

openhands/cli/main.py
```diff
@@ -304,6 +307,14 @@ def on_event(event: Event) -> None:
...
 
         runtime.config.mcp.stdio_servers.extend(openhands_mcp_stdio_servers)
+        # Add `fetch` MCP server by default if not already present
+        fetch_mcp_config = MCPStdioServerConfig(
+            name='fetch',
+            command='test-cmd',
+            args=['mcp-server-fetch'],
+        )
+        if fetch_mcp_config not in runtime.config.mcp.stdio_servers:
+            runtime.config.mcp.stdio_servers.append(fetch_mcp_config)
```
xingyaoww: nit: do we still need this, since the microagent will add 'fetch' by default?

ryanhoangt: Yeah, looking more closely, I think it should be added by global microagents, but currently [it doesn't](https://github.com/All-Hands-AI/OpenHands/pull/9519#pullrequestreview-3047924810). I investigated a bit and probably the cause is that we don't include the `microagents/` directory in the package. I added it to the `include` directive of `pyproject.toml` and removed this, let's see if the global microagents are loaded correctly.

PR #9954, opened 2025-07-29
fix(frontend): fix tailwind import

mislavlukach:

Fix the collision in tailwind imports by importing it only in consumer package.

---
Keep tailwind import in consumer package and generate tailwind classes needed for the design system by importing design system classes in css configuration.


frontend/package.json
```diff
@@ -51,6 +50,7 @@
     "sirv-cli": "^3.0.1",
     "socket.io-client": "^4.8.1",
     "tailwind-merge": "^3.3.1",
+    "tailwind-scrollbar": "^4.0.2",
```
amanape, line 53: Is this here because the ui package requires it? Isn’t it marked as a dependency on the ui library?

mislavlukach: It's peer dependency in DS. 
We need it.

PR #9958, opened 2025-07-29
refactor: improve the get microagents API

hieptl:

Currently, to retrieve the list of microagents for a selected repository, we are cloning the entire repository and reading from the ./openhands/microagents folder.

However, this approach should be updated. Instead, we should fetch the list of microagents by following the method described in the GitHub API documentation: [GitHub REST API – Repositories: Get repository content](https://docs.github.com/en/rest/repos/contents).

**Acceptance Criteria:**
- The updated get microagents API should return the list of microagents for the selected repository without cloning it locally.
- It should use the APIs from the Git providers to directly access the contents of the openhands/microagents folder.

---

The PR uses the APIs from the Git providers to directly access the contents of the openhands/microagents folder.

We can refer to the video below for the output of the PR:

https://github.com/user-attachments/assets/c8cfe762-e27c-46b8-9f3b-7227ebd450c7

---

Resolves [ALL-2868
](https://linear.app/all-hands-ai/issue/ALL-2868/microagent-management-improve-the-get-microagents-api)


openhands/server/routes/git.py
```diff
@@ -244,67 +239,749 @@ async def get_repository_branches(
...
-    inputs: list[InputMetadata] = []
-    tools: list[str] = []
+    Note: This model only includes basic metadata that can be determined
+    without parsing microagent content. Use the separate content API
+    to get detailed microagent information.
+    """
+
+    name: str  # File name without extension
+    git_provider: ProviderType
+    path: str  # Path to the microagent in the Git provider (e.g., ".openhands/microagents/tell-me-a-joke.md")
```
enyst: A bit surprised here, why are we adding git provider? They're in the repository now, no matter how it was cloned. And the comment doesn't seem to depend on any provider.

openhands/server/routes/git.py
```diff
@@ -244,67 +239,749 @@ async def get_repository_branches(
...
+        return '.openhands/microagents'
+
+
+def _get_github_headers(github_token: str | None) -> dict[str, str]:
+    """Get headers for GitHub API requests."""
+    headers = {'Accept': 'application/vnd.github+json'}
+    if github_token:
+        headers['Authorization'] = f'Bearer {github_token}'
+        headers['X-GitHub-Api-Version'] = '2022-11-28'
+    return headers
```
enyst: Seems like we should have these headers somewhere in the code? We might be able to get them by calling some method, if we need them (not sure we need them just to work with microagents?)

openhands/server/routes/git.py
```diff
@@ -244,67 +239,749 @@ async def get_repository_branches(
...
+
+    headers = _get_bitbucket_headers(bitbucket_token)
+    microagents = []
+
+    async with httpx.AsyncClient() as client:
+        try:
+            # Step 1: Get repository info to determine the main branch
+            repo_info_url = (
+                f'https://api.bitbucket.org/2.0/repositories/{repository_name}'
+            )
```
malhotra5: We should have separation of concerns here; provider related api calls should go into the service classes. Also we handle tokens different in local vs saas and  is managed by the service classes

openhands/server/routes/git.py
```diff
@@ -244,67 +239,749 @@ async def get_repository_branches(
...
+    github_token = None
+    if (
+        provider_handler.provider_tokens
+        and ProviderType.GITHUB in provider_handler.provider_tokens
+    ):
+        token = provider_handler.provider_tokens[ProviderType.GITHUB].token
+        if token is not None:
+            github_token = token.get_secret_value()
+
+    headers = _get_github_headers(github_token)
```
malhotra5: the provider handler handles this logic as well - see the `_get_service` method for the class

ideally we have one method per service class to ping specific api routes

then the provider handler will instantiate the classes and get the results from the service class methods

hieptl:
> Interesting, just a quick note on this PR
> 
> > The updated get microagents API should return the list of microagents for the selected repository without cloning it locally.
> > It should use the APIs from the Git providers to directly access the contents of the openhands/microagents folder.
> 
> I think this refactoring belongs in the runtime, where the microagents are loaded from the cloned repo, then? In /routes, we should do just routing, rather than backend processing.

Hi @enyst,

This API is used specifically for the new Microagent Management page and is not related to the runtime. Initially, we followed a similar approach to the runtime by cloning repositories and reading the microagents folders. However, this approach poses potential risks.

Therefore, the purpose of this ticket is to remove the logic that clones repositories and reads from the microagents folders. Instead, we aim to populate the data for the new Microagent Management page through a safer and more reliable method.

Thank you for your feedback!

openhands/integrations/gitlab/gitlab_service.py
```diff
@@ -523,6 +527,196 @@ async def create_mr(
...
+            # Handle self-hosted GitLab (e.g., 'gitlab.example.com/owner/repo')
+            parts = repository.split('/')
+            if len(parts) >= 3:
+                domain = parts[0]
+                project_id = '/'.join(parts[1:]).replace('/', '%2F')
+                base_url = f'https://{domain}/api/v4/projects/{project_id}'
+            else:
+                base_url = f'{self.BASE_URL}/projects/{project_id}'
+        else:
+            base_url = f'{self.BASE_URL}/projects/{project_id}'
```
malhotra5: self.BASE_URL is initialized based on the self hosted domain already ; so I believe we only need to extract `project_id` reliably here

openhands/integrations/gitlab/gitlab_service.py
```diff
@@ -523,6 +527,196 @@ async def create_mr(
...
+
+        async with httpx.AsyncClient() as client:
+            try:
+                headers = await self._get_gitlab_headers()
+                # Use repository's default branch (no ref parameter)
+                tree_response = await client.get(
+                    f'{base_url}/repository/tree',
+                    headers=headers,
+                    params={'path': microagents_path, 'recursive': 'true'},
+                )
```
malhotra5: we have a method in the class `_make_request`

please use this method instead

malhotra5: same applies for the other integrations

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -385,6 +389,194 @@ async def create_pr(
...
+        """Determine the microagents directory path based on repository name."""
+        repository_name.split('/')[-1]
+
+        # For Bitbucket, use default behavior: look for .openhands/microagents directory
+        return '.openhands/microagents'
+
+    def _create_microagent_response(self, file_name: str, path: str) -> dict:
+        """Create a microagent response from basic file information."""
+        # Extract name without extension
+        name = file_name.replace('.md', '').replace('.cursorrules', 'cursorrules')
```
malhotra5: Maybe out of scope for this PR, but I believe we're looking to support cursor rules in the OH ecosystem 

cc @enyst is this done yet? I feel like the logic to process cursor rules should live outside the service classes

openhands/integrations/gitlab/gitlab_service.py
```diff
@@ -20,6 +22,8 @@
 from openhands.server.types import AppMode
 from openhands.utils.import_utils import get_impl
 
+logger = logging.getLogger(__name__)
```
malhotra5:
```suggestion
from openhands.core.logger import openhands_logger as logger
```

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -19,6 +21,8 @@
 from openhands.server.types import AppMode
 from openhands.utils.import_utils import get_impl
 
+logger = logging.getLogger(__name__)
```
malhotra5:
```suggestion
from openhands.core.logger import openhands_logger as logger
```

openhands/server/routes/git.py
```diff
@@ -393,151 +267,136 @@ def _extract_repo_name(repository_name: str) -> str:
...
-                    [server.name for server in r_agent.metadata.mcp_tools.stdio_servers]
-                    if r_agent.metadata.mcp_tools
-                    else []
-                ),
-                created_at=created_at,
-                git_provider=git_provider,
-                path=str(agent_file_path.relative_to(repo_dir)),
+
+        # Fetch microagents using the provider handler
+        microagents_data = await provider_handler.get_microagents(repository_name)
```
malhotra5: maybe the provider should return `MicroagentResponse` by default; same for the service classes they should return  `MicroagentResponse`

also `MicroagentResponse` could be reused in other parts of the codebase it seems, can we find a more suitable location for where we define this

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -385,6 +388,135 @@ async def create_pr(
...
+            repo_info_url = f'{self.BASE_URL}/repositories/{repository}'
+            repo_data, _ = await self._make_request(repo_info_url)
+
+            # Extract main branch name from repository info
+            main_branch = repo_data.get('mainbranch', {}).get('name')
+            if not main_branch:
+                raise RuntimeError(
+                    f'No main branch found in repository info for {repository}. '
+                    f'Repository response: {repo_data.get("mainbranch", "mainbranch field missing")}'
+                )
```
malhotra5: Maybe we should return an empty list and add a warning log here

Not sure if its appropriate to throw an error

Also the assumption is that the provider would include this info in the response, so a warning log seems appropriate

If the request fails then we have other error handling paths that kick in

openhands/integrations/gitlab/gitlab_service.py
```diff
@@ -523,6 +530,146 @@ async def create_mr(
...
+    def _determine_microagents_path(self, repository_name: str) -> str:
+        """Determine the microagents directory path based on repository name."""
+        actual_repo_name = repository_name.split('/')[-1]
+
+        if actual_repo_name == 'openhands-config':
+            # For GitLab with repository name "openhands-config", scan "microagents" folder
+            return 'microagents'
+        else:
+            # Default behavior: look for .openhands/microagents directory
+            return '.openhands/microagents'
```
malhotra5: It seems like this method is shared across all the service classes (with the exception of some logic being modified for gitlab)

Let's move it into `BaseGitService` class so that its inherited by the service classes - you can override the implementation for gitlab specifically

Same applies for the `_create_microagent_response` method

openhands/integrations/bitbucket/bitbucket_service.py
enyst: It exists already, here:
https://github.com/All-Hands-AI/OpenHands/blob/f87c827fe689cc5d6c3597a7f11a19ee5aae2d17/openhands/microagent/microagent.py#L44

Only in that place, so from the user/workspace, not tied to a remote git repo.

enyst:
> This API is used specifically for the new Microagent Management page and is not related to the runtime. Initially, we followed a similar approach to the runtime by cloning repositories and reading the microagents folders. However, this approach poses potential risks.
> 
> Therefore, the purpose of this ticket is to remove the logic that clones repositories and reads from the microagents folders. Instead, we aim to populate the data for the new Microagent Management page through a safer and more reliable method.

Sorry, do I understand correctly
- the microagents were already loaded via clone
- and they’re available in memory in the current session
- then when we need them from FE, we don’t use the existing ones, but get them again?

I think the github API you linked is clearly a better option! But it’s not clear to me how it will work. Should we rather replace getting them via clone with getting them _always_, via this API?

hieptl:
> > This API is used specifically for the new Microagent Management page and is not related to the runtime. Initially, we followed a similar approach to the runtime by cloning repositories and reading the microagents folders. However, this approach poses potential risks.
> > Therefore, the purpose of this ticket is to remove the logic that clones repositories and reads from the microagents folders. Instead, we aim to populate the data for the new Microagent Management page through a safer and more reliable method.
> 
> Sorry, do I understand correctly
> 
> * the microagents were already loaded via clone
> * and they’re available in memory in the current session
> * then when we need them from FE, we don’t use the existing ones, but get them again?
> 
> I think the github API you linked is clearly a better option! But it’s not clear to me how it will work. Should we rather replace getting them via clone with getting them _always_, via this API?

Hi @enyst, yes, what you mentioned applies to the conversation page, which runs on the runtime — a user-controlled development environment — rather than the central API server. That makes it much safer to perform actions like cloning repositories and executing commands there.

However, the purpose of the APIs in this ticket is different (improving the APIs for the microagent management page). I’m building two APIs:
- One to fetch the list of microagents for a selected repository.
- Another to fetch the content of a microagent based on its file_path in the corresponding repository.

These APIs are intended for the Microagent Management page, which is currently hidden behind a feature flag (not part of the conversation page). You can refer to the video below for the UI of the Microagent Management page.

https://github.com/user-attachments/assets/9b26fca1-a3cb-4211-aa33-f5a928b65a78

malhotra5:
@hieptl I think what Engel is suggesting makes sense

We retrieve microagents in two scenarios

1. In the microagent management page; we're using new api routes as defined in this PR (this exists outside of a conversation/runtime context)
2. When we start a new conversation via cloning the repo and checking for microagent folder (this exists in the runtime context)

What Engel is suggesting is that we broaden the scope of this PR a bit; we have new methods in the service classes for retrieving microagents for the microagent management page. Can we see if we can reuse these methods when starting a new convo and fetching the microagents in the runtime (that way we remove all logic from the codebase where we clone repos for microagents)? 

EDIT: lmk if you have any concerns regarding this; I'm OK if this feels out of scope for this PR we can create a followup one

hieptl:
> @hieptl I think what Engel is suggesting makes sense
> 
> We retrieve microagents in two scenarios
> 
> 1. In the microagent management page; we're using new api routes as defined in this PR (this exists outside of a conversation/runtime context)
> 2. When we start a new conversation via cloning the repo and checking for microagent folder (this exists in the runtime context)
> 
> What Engel is suggesting is that we broaden the scope of this PR a bit; we have new methods in the service classes for retrieving microagents for the microagent management page. Can we see if we can reuse these methods when starting a new convo and fetching the microagents in the runtime (that way we remove all logic from the codebase where we clone repos for microagents)?

Hi @malhotra5,

I’m not sure we can directly reuse my current logic for the conversation page. I believe we need to investigate further before moving in that direction.

Here are a few reasons why:

⸻

Reason 1: Separate APIs

In this PR, I’m building two separate APIs:
- One API fetches the list of microagents for a selected repository. I intentionally excluded the content field from this response to avoid performance issues—especially when dealing with a large number of microagents. Additionally, for providers like GitLab or Bitbucket, fetching content would require extra API calls.
- The second API fetches the content of a specific microagent based on its file_path in the corresponding repository. This is only triggered when a user clicks on a microagent, ensuring better performance.

⸻

Reason 2: Different Behavior in Conversation Page

When launching a conversation from scratch (without selecting a repository or branch), we need to load some default microagents. My current APIs are designed specifically for loading microagents from a selected repository, so this behavior is not supported out of the box.

⸻

Reason 3: Frontend Changes May Be Needed

Reintegrating this logic into the conversation page may require updates on the frontend side, which could go beyond the scope of this PR.

⸻

Given these points, I suggest we move forward with merging this PR to avoid blocking the Microagent Management page (which is currently hidden behind a feature flag). We can then create a separate ticket to investigate whether and how we can reuse the current logic for the conversation page.

Thank you!

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -385,6 +387,123 @@ async def create_pr(
...
+            # Process .cursorrules if it exists
+            try:
+                cursorrules_url = f'{self.BASE_URL}/repositories/{repository}/src/{main_branch}/.cursorrules'
+                cursorrules_response, _ = await self._make_request(cursorrules_url)
+                if cursorrules_response:
+                    microagents.append(
+                        self._create_microagent_response('.cursorrules', '.cursorrules')
+                    )
+            except Exception as e:
+                logger.debug(f'No .cursorrules file found in {repository}: {e}')
```
malhotra5: It looks like if the user didn't setup microagents but has cursor rules then this method wouldn't return the cursor rules?

The request for the microagent path would error with 404, and would exit early?

malhotra5: It could help reducing the nested try catch as well

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -385,6 +387,123 @@ async def create_pr(
...
+            # Extract main branch name from repository info
+            main_branch = repo_data.get('mainbranch', {}).get('name')
+            if not main_branch:
+                logger.warning(
+                    f'No main branch found in repository info for {repository}. '
+                    f'Repository response: {repo_data.get("mainbranch", "mainbranch field missing")}'
+                )
+                return MicroagentContentResponse(
+                    content='', path=file_path, triggers=[]
+                )
```
malhotra5: I think we already have an existing method in the class for getting the repo using the name, could we use that here

Also the logic to the main branch seems to be reused can we pull that out into a method of its own

openhands/integrations/github/github_service.py
```diff
@@ -537,6 +539,95 @@ async def create_pr(
...
+            file_content = base64.b64decode(file_data['content']).decode('utf-8')
+
+            # Parse the content to extract triggers from frontmatter
+            return self._parse_microagent_content(file_content, file_path)
+
+        except Exception as e:
+            if '404' in str(e).lower():
+                raise RuntimeError(
+                    f'File not found: {file_path} in repository {repository}'
+                )
```
malhotra5: We handle http exceptions in the  `BaseGitService` class, could we add some logic to handle the `404` from there?

Also runtime error doesn't feel like an appropriate exception, if you see the examples in BaseGitService and they're not fitting feel free to create a new one

malhotra5: Also if this error occurs do we handle it gracefully in the FE? What's the response the user sees?

openhands/integrations/service_types.py
```diff
@@ -199,6 +203,68 @@ def handle_http_error(self, e: HTTPError) -> UnknownException:
...
+            )
+
+        except Exception as e:
+            logger.warning(
+                f'Error parsing microagent content for {file_path}: {str(e)}'
+            )
+            # Return response with original content and empty triggers list on parsing error
+            return MicroagentContentResponse(
+                content=content, path=file_path, triggers=[]
+            )
```
malhotra5: If we have trouble parsing maybe we should let the user know? It means their microagent won't load inside a conversation

openhands/integrations/provider.py
```diff
@@ -407,6 +408,111 @@ async def get_branches(
 
         return main_branches + other_branches
 
+    async def get_microagents(self, repository: str) -> list[MicroagentResponse]:
```
malhotra5, line 413: If we're fetching microagent for a repo would we know which provider its from in advance?

hieptl: Currently, the front_end is not using that information. Therefore, I excluded git_provider from the response.

malhotra5: Gotcha, that may change now that we're about to have repo listing done by provider - once this PR is merged let's also open another ticket where we include provider info in this route

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -385,6 +393,126 @@ async def create_pr(
...
+
+        Returns:
+            List of microagents found in the repository (without content for performance)
+        """
+        microagents_path = self._determine_microagents_path(repository)
+        microagents = []
+
+        try:
+            # Step 1: Get repository details using existing method
+            repo_details = await self.get_repository_details_from_repo_name(repository)
```
malhotra5: Can we remove the nesting as follows

1. one try catch for getting repo details; if repo details don't exist return early

2. one try catch for getting cursor rules

3. one try catch for getting microagents

hieptl: Hi @malhotra5, I have updated the code based on your feedback. Thank you for your feedback!

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -385,6 +393,129 @@ async def create_pr(
...
+        microagents_path = self._determine_microagents_path(repository)
+        microagents = []
+
+        # Step 1: Get repository details - return early if not found
+        try:
+            repo_details = await self.get_repository_details_from_repo_name(repository)
+        except Exception as e:
+            raise UnknownException(
+                f'Error fetching repository details from Bitbucket: {str(e)}'
+            )
```
malhotra5: `make_request` already raises the correct exception type, so we don't need to wrap it with UnknownException again

malhotra5: Actually we don't need a try catch here since the exception will be raised inside `get_repository_details_from_repo_name` , the provider handler can handle it

hieptl: Hi @malhotra5, I have updated the code based on your feedback. Thank you for your feedback!

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -385,6 +393,129 @@ async def create_pr(
...
+            repository: Repository name in format 'workspace/repo_slug'
+            file_path: Path to the file within the repository
+
+        Returns:
+            MicroagentContentResponse with parsed content and triggers
+
+        Raises:
+            RuntimeError: If file cannot be fetched or doesn't exist
+        """
+        try:
```
malhotra5: same here we don't need the try catch the provider handler will manage it

same stuff applies for the other service classes

openhands/integrations/provider.py
```diff
@@ -407,6 +408,111 @@ async def get_branches(
...
+                service = self._get_service(provider)
+                result = await service.get_microagent_content(repository, file_path)
+                # If we got content, return it immediately
+                if result:
+                    return result
+                # If we got empty content, continue checking other providers
+                logger.debug(
+                    f'No content found on {provider} for {repository}/{file_path}, trying other providers'
+                )
+            except Exception as e:
```
malhotra5, line 493: can we check for the specific exception type `ResourceNotFound`

openhands/integrations/provider.py
```diff
@@ -407,6 +408,111 @@ async def get_branches(
...
+                )
+
+        # If all providers failed or returned empty results, raise an error
+        if errors:
+            logger.error(
+                f'Failed to fetch microagent content for {repository} with all available providers. Errors: {"; ".join(errors)}'
+            )
+            raise AuthenticationError(
+                f'Unable to fetch microagent content for {repository}'
+            )
```
malhotra5:
```suggestion
```

enyst:
I’m still confused, sorry. Just a quick question though.

If the intention of this PR is not to replace how we load microagents by cloning, then why does its description say:

> Currently, to retrieve the list of microagents for a selected repository, we are cloning the entire repository and reading from the ./openhands/microagents folder.

> However, **this approach should be updated. Instead, we should fetch** the list of microagents by following the method described in the GitHub API documentation.

Are we cloning the repo multiple times? 😕

hieptl:
> I’m still confused, sorry. Just a quick question though.
> 
> If the intention of this PR is not to replace how we load microagents by cloning, then why does its description say:
> 
> > Currently, to retrieve the list of microagents for a selected repository, we are cloning the entire repository and reading from the ./openhands/microagents folder.
> 
> > However, **this approach should be updated. Instead, we should fetch** the list of microagents by following the method described in the GitHub API documentation.
> 
> Are we cloning the repo multiple times? 😕

Hi @enyst, Currently, on the Microagent Management page, every time a user clicks on a repository, we re-clone the repository to fetch the latest list of microagents.

This ticket aims to eliminate that redundant cloning process by using Git provider APIs to retrieve the latest microagent data instead.

Thank you!

https://github.com/user-attachments/assets/9b26fca1-a3cb-4211-aa33-f5a928b65a78

openhands/integrations/github/github_service.py
```diff
@@ -537,6 +540,86 @@ async def create_pr(
...
+                f'{base_url}/.cursorrules'
+            )
+            if cursorrules_response:
+                microagents.append(
+                    self._create_microagent_response('.cursorrules', '.cursorrules')
+                )
+        except ResourceNotFoundError:
+            logger.debug(f'No .cursorrules file found in {repository}')
+        except Exception as e:
+            logger.warning(f'Error checking .cursorrules file in {repository}: {e}')
```
malhotra5: let's see if we can modularize this more

can we make a method `get_microagents ` part of `BaseGitService` class, and pass the api routes as args (something like that)

Similar logic seems to appear across all the service classes; same goes for fetching the microagent directory just beneath this piece of code

openhands/integrations/gitlab/gitlab_service.py
```diff
@@ -117,7 +121,10 @@ async def _make_request(
                 if 'Link' in response.headers:
                     headers['Link'] = response.headers['Link']
 
-                return response.json(), headers
+                if is_json_response:
+                    return response.json(), headers
+                else:
+                    return response.text, headers
```
malhotra5:
```suggestion
                content_type = response.headers.get("Content-Type", "")
                if "application/json" in content_type:
                    return response.json(), headers
                else:
                    return response.text, headers
```

Does something like this work? and we can remove the `is_json_response` arg and make this automatic?

openhands/integrations/github/github_service.py
malhotra5: actually i do see small variations between the service classes when processing the response from the api - see if its possible to dedup the logic maybe but if not its ok

at the very least getting cursor rules can be pulled out into a shared method

enyst:
Got it, yes, I just saw too. This PR is a good idea, thank you!

malhotra5 (approved):
LGTM; to note for future purposes we should have the FE handle the case where the microagent is read from a directory but we were unable to parse it

This indicates that the user (or agent) created a microagent with incorrect formatting and the microagent will not trigger as expected during a active conversation

PR #9967, opened 2025-07-29
Fix git changes panel

tofarr:

Fix issue where git changes panel was failing against repositories with no commits.

---
Rework of the invalid logic in get_valid_ref for both `git_changes.py` and `git_diff.py` and fixing associated unit tests.

**Start from Scratch**
image

**Start with Repo**
image

---

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:6f459a0-nikolaik   --name openhands-app-6f459a0   docker.all-hands.dev/all-hands-ai/openhands:6f459a0
```

### DIFF
--- a/openhands/runtime/utils/git_changes.py
+++ b/openhands/runtime/utils/git_changes.py
@@ -46,7 +46,9 @@ def get_valid_ref(repo_dir: str) -> str | None:
         pass
 
     # compares with empty tree
-    ref_new_repo = '$(git --no-pager rev-parse --verify 4b825dc642cb6eb9a060e54bf8d69288fbee4904)'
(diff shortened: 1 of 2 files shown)


tofarr:
> LGTM. `get_valid_ref` looks duplicated, maybe we can move it into a utils file too

The reason this is duplicated is that the scripts are run as individuals without imports for backward compatibility - otherwise the imports would also need to be included when creating the temp scripts.

Issue #9973, opened 2025-07-29
[Bug]: Local runtime fails with "only one config file at a time" when setting user and email at the same time


greese-insight:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Running locally on linux/mac causes an error to occur regarding setting the openhands user and email.

*Prompt*

```
Make me a typescript project.

Make sure to create and/or append to the gitignore file and dockerignore files with changes if necessary.

When you are completed, push the branch to the remote repository.

After pushing to the remote repository, make a pull request with a descriptive name and description, and output any questions or issues into the comments.
```

*Config.toml*

```
[core]
runtime="local"
# Disable the browser environment
enable_browser = false

[llm]
model = "azure/gpt-4o"
api_key = "****"
base_url = "https://***.openai.azure.com"
api_version = "2025-01-01-preview"
num_retries=2
retry_max_wait=60
```

*Env vars*

```
# don't install playwright by default
ENV INSTALL_DOCKER=0
ENV SANDBOX_USER_ID=1000
ENV INSTALL_PLAYWRIGHT=0
ENV SANDBOX_VOLUMES=/workspace/:/workspace:rw
```

### OpenHands Installation

Other

### OpenHands Version

main

### Model Name

gpt-4o

### Operating System

MacOS

### Logs, Errors, Screenshots, and Additional Context

git config --local user.name "openhands"
18:23:53 - openhands:INFO: local_runtime.py:694 - server: 18:23:53 - openhands:DEBUG: bash.py:479 - RECEIVED ACTION: **CmdRunAction (source=EventSource.AGENT, is_input=False)**
18:23:53 - openhands:INFO: local_runtime.py:694 - server: COMMAND:
18:23:53 - openhands:DEBUG: bash.py:63 - BASH PARSING command: git config --local user.name "openhands"
18:23:53 - openhands:INFO: local_runtime.py:694 - server: git config --local user.name "openhands"
18:23:53 - openhands:INFO: local_runtime.py:694 - server: 18:23:53 - openhands:DEBUG: bash.py:63 - BASH PARSING command: git config --local user.name "openhands"
18:23:53 - openhands:INFO: local_runtime.py:694 - server: 18:23:53 - openhands:DEBUG: bash.py:70 - BASH PARSING remaining:
[...]

PR #9975, opened 2025-07-29
fix: update git config to handle the necessary user name and email se…

greese-insight:
…t in order

Fixes issue with local runtime on linux/macos where setting the username and email was causing an error of "only one config file at a time"

---

---

https://github.com/All-Hands-AI/OpenHands/issues/9973


enyst:
Thank you @greese-insight ! It’s curious why you had this happen, since we build on linux all the time and it doesn’t care 🤔

enyst:
It seems lint is failing, for that, you may want to run `make build`, which installs the pre-commit hook, and/or run it on all files like this
https://github.com/All-Hands-AI/OpenHands/blob/3f395e3cee8cf2e6030e5d5bad63be286847de28/microagents/add_repo_inst.md?plain=1#L34

greese-insight:
Doh, thanks! Updated and pushed

PR #9977, opened 2025-07-29
Remove task completion status message from finish action display

xingyaoww:
## Description
This PR removes the "I believe that the task was completed partially/successfully/not completed" message that appears at the end of conversations. These messages are based on the `task_completed` parameter from the agent's finish action.

## Changes
- Modified `getFinishActionContent` function in `get-action-content.ts` to only return the final thought without appending the task completion status message.

## Testing
- Built the frontend successfully
- Verified that the changes work as expected

## Related Issues
Addresses user request to stop displaying "I believe the task was XX" in the frontend UI based on the agent's finish action argument.

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/31124f9624aa496984c60bbd4ae68e9f)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:be93017-nikolaik   --name openhands-app-be93017   docker.all-hands.dev/all-hands-ai/openhands:be93017
```

### DIFF
--- a/frontend/src/components/features/chat/event-content-helpers/get-action-content.ts
+++ b/frontend/src/components/features/chat/event-content-helpers/get-action-content.ts
@@ -77,25 +77,9 @@ const getMcpActionContent = (event: MCPAction): string => {
 const getThinkActionContent = (event: ThinkAction): string =>
   event.args.thought;
 
-const getFinishActionContent = (event: FinishAction): string => {
-  let content = event.args.final_thought;
-
-  switch (event.args.task_completed) {
-    case "success":
-      content += `\n\n\n${i18n.t("FINISH$TASK_COMPLETED_SUCCESSFULLY")}`;
-      break;
-    case "failure":
-      content += `\n\n\n${i18n.t("FINISH$TASK_NOT_COMPLETED")}`;
-      break;
-    case "partial":
-    default:
-      content += `\n\n\n${i18n.t("FINISH$TASK_COMPLETED_PARTIALLY")}`;
-      break;
-  }
-
-  return content.trim();
-};
-
+const getFinishActionContent = (event: FinishAction): string =>
+  // Return only the final thought without appending the task completion status message
+  event.args.final_thought.trim();
 const getNoContentActionContent = (): string => "";
 
 export const getActionContent = (event: OpenHandsAction): string => {
--- a/frontend/src/components/features/chat/event-content-helpers/get-action-content.ts
+++ b/frontend/src/components/features/chat/event-content-helpers/get-action-content.ts
@@ -78,7 +78,6 @@ const getThinkActionContent = (event: ThinkAction): string =>
   event.args.thought;
 
 const getFinishActionContent = (event: FinishAction): string =>
-  // Return only the final thought without appending the task completion status message
   event.args.final_thought.trim();
 const getNoContentActionContent = (): string => "";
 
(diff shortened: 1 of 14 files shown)


enyst (commented):
I think maybe we could either
- remove it from the action, or
- keep it visible to the user.

IIRC it’s confusing for o1/o3 too, and apparently GPT-4.1, they were reporting Finish with partial very fast, maybe because a lot of steps can be considered “partial”? (🤷‍♂️) The old Gemini also was very hurried to report Finish.

IMHO we should try to show the user what happened… maybe add some different explanation? “Agent reports that the task was not fully completed”, I don’t know 🤔

xingyaoww:
@enyst hmm yeah i start to feel this becomes not so reliable that it confuses the user more than it should :( - especially like you said, different model behaves differently.

WDYT about completely remove it then?

enyst:
> WDYT about completely remove it then?

I’d love that, it’s confusing me too! I think that would help both the user and (some?) LLM! Generally speaking… Of course, it depends how little Claude was trained, IIRC it was part of its tool definition. We can eval the removal?

xingyaoww:
@OpenHands can you help me completely remove the `task_completed` inside OpenHands repository? You can just do a global search of this parameter and remove them.

BUT do add keep handling logic in the serialization / deserialization code, so we can load old version of the conversation that has this argument: task_completed

xingyaoww:
@OpenHands please fix the failing actions on PR https://github.com/All-Hands-AI/OpenHands/pull/9977

frontend/src/components/features/chat/event-content-helpers/get-action-content.ts
```diff
@@ -77,25 +77,9 @@ const getMcpActionContent = (event: MCPAction): string => {
...
-    default:
-      content += `\n\n\n${i18n.t("FINISH$TASK_COMPLETED_PARTIALLY")}`;
-      break;
-  }
-
-  return content.trim();
-};
-
+const getFinishActionContent = (event: FinishAction): string =>
+  // Return only the final thought without appending the task completion status message
```
enyst:
```suggestion
```

openhands/llm/fn_call_converter.py
```diff
@@ -311,7 +311,6 @@ def index():
 The server is running on port 5000 with PID 126. You can access the list of numbers in a table format by visiting http://127.0.0.1:5000. Let me know if you have any further requests!
 <function=finish>
 <parameter=message>The task has been completed. The web server is running and displaying numbers 1-10 in a table format at http://127.0.0.1:5000.</parameter>
-<parameter=task_completed>true</parameter>
```
enyst, line 314: There are 2 more occurrences of task_completed in this file. Maybe we should send OpenHands after them?

One of them https://github.com/All-Hands-AI/OpenHands/blob/9f319518addb316e5e2c66ef570a6ac099c5dbaf/openhands/llm/fn_call_converter.py#L375

xingyaoww: @OpenHands please go! do a global search for "task_completed" and try to clean them up

enyst (approved):
LGTM!

Might be a good idea to have a quick sanity eval - maybe just with the run_eval label?

Actually, also something like 1-2 instances deepseek/devstral would be nice just to make sure it didn’t break.

xingyaoww:
Ok seems good! Let's merge!

PR #9980, opened 2025-07-29
Fix for app/worker urls

tofarr:

* Made the URL for app work like vs-code
* Introduced a common method as these share common functionality

---
Made the app server urls work like VS Code.

---

### DIFF
--- a/openhands/runtime/impl/local/local_runtime.py
+++ b/openhands/runtime/impl/local/local_runtime.py
@@ -575,25 +575,30 @@ class LocalRuntime(ActionExecutionClient):
         # Fallback to localhost
         return self.config.sandbox.local_runtime_url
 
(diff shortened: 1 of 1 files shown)


openhands/runtime/impl/local/local_runtime.py
```diff
@@ -575,25 +575,30 @@ def runtime_url(self) -> str:
         # Fallback to localhost
         return self.config.sandbox.local_runtime_url
 
-    @property
-    def vscode_url(self) -> str | None:
```
raymyers, line 579: Looks like this is an untested method, not that this PR created it. Change looks trivial but would be good to add one if you have time.

tofarr: I'll add one in a follow up PR

PR #9981, opened 2025-07-29
Add maintenance banner feature

malhotra5:

- Adds informational banner for when maintenance is scheduled for Cloud OpenHands

---

Adds a banner for information on a scheduled maintenance 

image

---

Fixes ALL-2876

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/f06fddf2df9941e68ccb484222577736)

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c578d1a-nikolaik   --name openhands-app-c578d1a   docker.all-hands.dev/all-hands-ai/openhands:c578d1a
```


frontend/src/mocks/handlers.ts
```diff
@@ -187,6 +187,10 @@ export const handlers = [
         ENABLE_BILLING: false,
         HIDE_LLM_SETTINGS: mockSaas,
       },
+      // Uncomment the following to test the maintenance banner
+      // MAINTENANCE: {
+      //   startTime: "2024-01-15T10:00:00-05:00", // EST timestamp
+      // },
```
neubig, line 193: Should we make this controlled by an env var, so we don't need to change the code every time we do maintenance?

malhotra5: This is just for the  FE mock (so we can run the FE without starting the backend server)

The actual value comes from an env in the BE, and will be configured outside of this repo

xingyaoww (commented):
Should we add an end time estimation also? or at least give them our incident IO URL to keep track of the maintainance status?

frontend/src/mocks/handlers.ts
neubig: Got it, thanks.

malhotra5:
We'll have a separate page that we'll spin up when maintenance is underway. This page will have the time estimate. 

We're still figuring out the actual process internally so this is the minimal implementation for a banner, but I'd think in the future we would add duration information as well

frontend/src/mocks/handlers.ts
raymyers, line 193: Nice touch

raymyers:
If I understand correctly this will show the banner even if the window is in the past. In the future we could make it time out, maybe not right at the start but sometime after.

malhotra5:
Yep sounds good, I think we can improve this banner once we get a formal processing going for maintenance windows

Agree that we should have some sort of timeout; my understanding is that either the maintenance page should be spun up eventually within a realistic time frame or we remove the banner because we decided to reschedule

PR #9996, opened 2025-07-30
feat:  allow user to edit title from header on conversation page.

hieptl:

**Acceptance criteria:**
- The conversation title should be at the upper-left of the screen.
- Ensure the title is clearly visible and easy to identify at a glance.
- Allow the user to rename the conversation directly from the UI.

---

This PR enables users to edit the conversation title directly from this screen.

https://github.com/user-attachments/assets/82f5563d-d6b6-4902-b614-f2464f614299

---


frontend/__tests__/components/features/conversation/conversation-name.test.tsx
amanape, line 1: Add another test: `it(“should not call the API if user attempts to save an unchanged title”)`

amanape (changes requested):
It looks like we now allow the user to change the title through a context menu. Could you apply these changes:

- Icon button to open context menu
- Only the RENAME option is available 
- Allow the user to change name when the RENAME option is selected.

Keep all other behaviour, including changing the title on double click

See [Figma](https://www.figma.com/design/iM5UBG1lnpg2sgE3cFnF1P/OpenHands?node-id=10222-15268&t=TPRBQoN7aHwdfYGM-4). I’ve also added a comment to get some more information, so there might be further requests soon

frontend/__tests__/components/features/conversation/conversation-name.test.tsx
hieptl: Hi @amanape, I've updated the code based on your feedback.

We can refer to the video below for the output.

https://github.com/user-attachments/assets/fd6f6c7e-079f-47ad-b545-898023391a2f

frontend/src/components/features/conversation/conversation-name.tsx
```diff
@@ -0,0 +1,122 @@
...
+        updateConversation(
+          { conversationId, newTitle: trimmed },
+          {
+            onSuccess: () => {
+              displaySuccessToast(t(I18nKey.CONVERSATION$TITLE_UPDATED));
+            },
+          },
+        );
+      }
+      inputRef.current.value = trimmed;
```
amanape: I don’t think this is required since we already optimistically update the title on mutation: https://github.com/All-Hands-AI/OpenHands/blob/d3209f8c286063b63f10555559309c7f26921143/frontend/src/hooks/mutation/use-update-conversation.ts#L19-L27

frontend/src/components/features/conversation/conversation-name.tsx
```diff
@@ -0,0 +1,122 @@
...
+          onClick={handleInputClick}
+          onBlur={handleBlur}
+          onKeyUp={handleKeyUp}
+          type="text"
+          defaultValue={conversation.title}
+          className="text-white leading-5 bg-transparent border-none outline-none text-base font-normal w-fit max-w-fit"
+        />
+      ) : (
+        <div
+          className="text-white leading-5 cursor-pointer w-fit max-w-fit truncate"
```
amanape:
```suggestion
          className="text-white leading-5 w-fit max-w-fit truncate"
```

Removing this since the user might mistake it for a single-click action instead of a double click one.

frontend/__tests__/components/features/conversation/conversation-name.test.tsx
```diff
@@ -0,0 +1,250 @@
...
+    const user = userEvent.setup();
+    renderWithProviders(<ConversationName />);
+
+    const titleElement = screen.getByTestId("conversation-name-title");
+    await user.dblClick(titleElement);
+
+    // Should be in edit mode
+    expect(screen.getByTestId("conversation-name-input")).toBeInTheDocument();
+
+    const inputElement = screen.getByTestId("conversation-name-input");
```
amanape:
```suggestion
```

`'inputElement' is declared but its value is never read.ts(6133)`

amanape (approved):
Thanks, just one comment on setting input value after the mutation. Could you confirm  that it isn’t necessary? You can update the failing test (`"should trim whitespace from input value"`) to expect:

```ts
    expect(mockMutate).toHaveBeenCalledWith(
      {
        conversationId: "test-conversation-id",
        newTitle: "Trimmed Title",
      },
      expect.any(Object),
    );
```

hieptl:
Hi @amanape, I've updated the code based on your feedback. Thank you for your feedback! 🙏

PR #10000, opened 2025-07-30
Fix duplicate GET requests to feedback/conversation/{id} endpoint

rbren:
## Problem
The frontend is making duplicate requests to `GET /feedback/conversation/{id}` endpoint, which is causing unnecessary network traffic and potential performance issues.

## Root Cause
1. The `useFeedbackExists` hook is being called for every event in the conversation, even when feedback data isn't needed
2. There's insufficient caching of feedback data
3. The hook is configured to refetch on window focus and reconnect

## Solution
1. Modified `EventMessage` component to only fetch feedback data when it's actually needed (for last message or error observations in last 10 actions)
2. Implemented multi-level caching:
   - Added sessionStorage caching in the API client
   - Improved React Query caching with longer staleTime and gcTime
   - Added direct cache updates when submitting feedback instead of invalidating queries
3. Disabled automatic refetching on window focus and reconnect
4. Added better error handling and logging

These changes significantly reduce the number of duplicate requests while maintaining the same functionality.

@rbren can click here to [continue refining the PR](https://app.all-hands.dev/conversations/5fe63c41a075421fadfe3ee635ae621f)

### DIFF
--- a/frontend/src/api/open-hands.ts
+++ b/frontend/src/api/open-hands.ts
@@ -153,15 +153,44 @@ class OpenHands {
     eventId: number,
   ): Promise<{ exists: boolean; rating?: number; reason?: string }> {
     try {
+      // Use a cache to prevent duplicate requests for the same data
+      const cacheKey = `feedback_${conversationId}_${eventId}`;
+      const cachedData = sessionStorage.getItem(cacheKey);
+
+      if (cachedData) {
+        try {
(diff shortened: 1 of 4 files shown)


rbren:
@OpenHands please fix the failing actions on PR https://github.com/All-Hands-AI/OpenHands/pull/10000

Issue #9968, opened 2025-07-29
[CLI] "auto-approve on command" feature in CLI


xingyaoww:
**What problem or use case are you trying to solve?**

Right now we need to approve every command agent wants to run in CLI, and it is too inconvenient.

**Describe the UX or technical implementation you have in mind**

We should have an approval mechanism similar to claude-code that will remember the commands into `config.toml`, so the agent don't repeatly ask for permission.

<img width="1738" height="308" alt="Image" src="https://github.com/user-attachments/assets/5cc764c1-cb9f-409e-8af6-13035d01b4dd" />

**Additional context**

xingyaoww:
https://github.com/All-Hands-AI/OpenHands/pull/9349 adds a smart confirmation mode taht will somewhat improve this issue too -- slightly more secured version of "always-approve" but not as strict as "auto-approve" for similar command

<img width="1348" height="2506" alt="Image" src="https://github.com/user-attachments/assets/696e785d-1efc-4760-8bf9-e15cde879131" />

PR #10005, opened 2025-07-30
Add command approval memory feature to CLI

xingyaoww:

This PR adds a "remember command" feature to the CLI, similar to Claude-code's command approval system. When running commands in the CLI, users now have four options:
1. Yes (proceed once)
2. No (reject)
3. Always proceed (skip all confirmations)
4. Yes, and don't ask again for similar commands

When selecting the fourth option, the command pattern is saved to `config.toml` so similar commands won't require confirmation in the future. The pattern is displayed to the user before approval, making it clear what will be remembered.

---

This PR implements a command approval memory system using the security analyzer framework:

1. Created a new `CommandApprovalAnalyzer` that checks if commands match approved patterns
2. Added pattern generation that intelligently creates regex patterns for similar commands
3. Extended the CLI confirmation dialog to show four options with pattern preview
4. Updated the config template to enable command approval by default
5. Added utility functions to save approved commands to config.toml

Design decisions:
- Used the security analyzer approach instead of modifying core controller logic
- Implemented smart pattern generation that handles file extensions and numbers
- Made the implementation non-invasive by not modifying action classes
- Ensured the feature is enabled by default in the config template

---

This is an attempt to solve https://github.com/All-Hands-AI/OpenHands/issues/9968

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/c946350c8f5c4dc6a57ccc752d3ff7c4)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:ff5fc5c-nikolaik   --name openhands-app-ff5fc5c   docker.all-hands.dev/all-hands-ai/openhands:ff5fc5c
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@feature/command-approval-memory openhands
```

### DIFF
--- a/openhands/cli/tui.py
+++ b/openhands/cli/tui.py
@@ -700,70 +700,40 @@ async def read_prompt_input(
         return '/exit'
 
 
-def _generate_single_command_pattern(command: str) -> str:
-    """Generate a regex pattern for a single command that can match similar commands.
+def _generate_single_command_pattern(command: str) -> list[str]:
+    """Generate candidate regex patterns for a single command based on prefixes.
 
     Args:
-        command: The single command to generate a pattern for.
+        command: The single command to generate patterns for.
 
     Returns:
-        str: A regex pattern that matches similar commands.
+        list[str]: List of candidate regex patterns that match similar commands.
     """
     import re
 
     # Split command into parts
     parts = command.split()
     if not parts:
-        return f'^{re.escape(command)}$'
+        return [f'^{re.escape(command)}$']
 
-    # First part is the command itself (keep exact)
-    pattern_parts = [re.escape(parts[0])]
+    patterns = []
 
-    # Process remaining parts
-    for part in parts[1:]:
-        escaped_part = re.escape(part)
+    # Generate patterns for first word, first two words, first three words
+    for i in range(1, min(4, len(parts) + 1)):
+        prefix_parts = parts[:i]
+        escaped_prefix = '\\s+'.join(re.escape(part) for part in prefix_parts)
 
-        # Replace file extensions with generic patterns
-        extensions = [
-            'py',
-            'js',
-            'ts',
-            'txt',
-            'md',
-            'json',
-            'yaml',
-            'yml',
-            'toml',
-            'html',
-            'css',
-        ]
-        for ext in extensions:
-            if escaped_part.endswith(f'\\.{ext}'):
-                # Replace filename.ext with any filename with extension
-                escaped_part = '[^\\s]+\\.[a-zA-Z0-9]+'
-                break
+        # Create pattern: prefix followed by word boundary and anything until end of line
+        # This prevents matching partial words (e.g., "ls" won't match "lsof")
+        if i == 1:
+            # For single word, add word boundary to prevent partial matches
+            pattern = f'^{escaped_prefix}(\\s.*|$)'
+        else:
+            # For multiple words, the space already acts as a boundary
+            pattern = f'^{escaped_prefix}.*$'
+        patterns.append(pattern)
 
-        # Handle special cases like .py (starting with dot) - but only if it's a standalone extension
-        if (
-            escaped_part.startswith('\\.')
-            and len(escaped_part) > 2
-            and '/' not in escaped_part
-        ):
-            # Check if this looks like a file extension pattern (e.g., .py, .txt)
-            ext_part = escaped_part[2:]  # Remove the \.
-            if ext_part.isalpha() and len(ext_part) <= 5:  # Common extension length
-                escaped_part = '\\.[a-zA-Z0-9]+'
-
-        # Replace numbers with flexible number patterns
-        escaped_part = re.sub(r'\\\d+', r'\\d+', escaped_part)
-
-        pattern_parts.append(escaped_part)
-
-    # Join with flexible whitespace
-    pattern = '\\s+'.join(pattern_parts)
-
-    # Make the pattern match from start to end
-    return f'^{pattern}$'
+    return patterns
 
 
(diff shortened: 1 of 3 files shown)


xingyaoww:
@OpenHands please review diff against main, there's one other thing we need: the ability to parse command (i.e., capture the pipe, and make sure all patterns on all sides of the pipe should match our pattern list)

enyst:
@xingyaoww I just want to say about this
> Used the security analyzer approach instead of modifying core controller logic

This is a wonderful choice! We really should try this, for multiple reasons, IMHO. Insignificant perhaps, is that we may need to fix the security analyzer, and the best way to do that is to dogfood it! 😅

xingyaoww:
@OpenHands can you review the diff against main, understand the purpose of this PR, and comes up with ways to simplify this PR. I want a minimal set of code that performs this changes, use third-party libraries whenever you can instead of implementing yourselves, especially when it come to parse terminal commands (i remember we have bashlex installed -- you should use it, it should tell you when a pipe exists in a command). Make sure you write a unit test testing your command parsing mechanim in the security analyzer before you start integrating it with CLI.

Also, for save new command patterns, please do re-use existing ability to save stuff to config.toml

openhands/core/config/security_config.py
```diff
@@ -1,18 +1,50 @@
...
+        approved_command_patterns: List of approved command patterns that don't require confirmation.
+        approved_commands: Dictionary of exact commands that have been approved.
     """
 
     confirmation_mode: bool = Field(default=False)
     security_analyzer: str | None = Field(default=None)
+    approved_command_patterns: list[ApprovedCommandPattern] = Field(
+        default_factory=list
+    )
+    approved_commands: dict[str, bool] = Field(default_factory=dict)
```
xingyaoww, line 31: We don't NEED to save approved commands; we just need to save approved patterns. Also, we also don't need the description field for `ApprovedCommandPattern` -- in this case, can we just greatly simplify ApprovedCommandPattern to be a regex pattern in string format? @openhands

xingyaoww, line 22: We don't NEED to save approved commands; we just need to save approved patterns. Also, we also don't need the description field for `ApprovedCommandPattern` -- in this case, can we just greatly simplify ApprovedCommandPattern to be a regex pattern in string format? @openhands

openhands/cli/tui.py
```diff
@@ -700,20 +700,207 @@ def _(event: KeyPressEvent) -> None:
...
+            sub_pattern = (
+                sub_pattern[1:-1]
+                if sub_pattern.startswith('^') and sub_pattern.endswith('$')
+                else sub_pattern
+            )
+            sub_patterns.append(sub_pattern)
+
+        # Join sub-patterns with pipe separator (allowing flexible whitespace around pipes)
+        pattern = '\\s*\\|\\s*'.join(sub_patterns)
+        return f'^{pattern}$'
```
xingyaoww, line 867: @Openhands can we instead return multiple sub patterns and ask user one by one which one they want to accept?

openhands/cli/tui.py
```diff
@@ -700,20 +700,207 @@ def _(event: KeyPressEvent) -> None:
...
+    elif parts:  # We had pipes but no final command
+        return []  # Invalid: ends with pipe
+
+    # If no pipes were found, fall back to simple pipe splitting
+    if len(parts) <= 1 and '|' in command:
+        simple_parts = [part.strip() for part in command.split('|')]
+        # Check for empty parts
+        if any(not part for part in simple_parts):
+            return []  # Invalid: empty parts
+        parts = simple_parts
```
xingyaoww: @openhands we don't need to fall back, in case there are actual command that include `|` in part of the string argument. please simply this function

openhands/cli/tui.py
```diff
@@ -700,20 +700,207 @@ def _(event: KeyPressEvent) -> None:
...
+        # Handle special cases like .py (starting with dot) - but only if it's a standalone extension
+        if (
+            escaped_part.startswith('\\.')
+            and len(escaped_part) > 2
+            and '/' not in escaped_part
+        ):
+            # Check if this looks like a file extension pattern (e.g., .py, .txt)
+            ext_part = escaped_part[2:]  # Remove the \.
+            if ext_part.isalpha() and len(ext_part) <= 5:  # Common extension length
+                escaped_part = '\\.[a-zA-Z0-9]+'
```
xingyaoww, line 755: @Openhands Let's just accept EVERYTHING after the first command, like `grep [arbitrary text is ok]` and `glob [arbitrary text is ok]`

openhands/cli/tui.py
xingyaoww: @Openhands can we instead return multiple sub patterns and ask the user one by one which one they want to accept?

We should also update openhands/cli/tui.py file so that user can actually edit the pattern if they want (e.g., if they want to make it more specific, instead of `git`, they want to only match and allow `git add`. -- make sure you print these as instruction for the user).

openhands/cli/tui.py
xingyaoww: @Openhands Let's just accept EVERYTHING after the first command, like `grep [arbitrary text is ok]` and `glob [arbitrary text is ok]`

This should greatly simplify _generate_single_command_pattern -- make it as simple as it possibly can

neubig:
Right now it's a bit annoying that you have to wait for a while as the agent works before clicking "always proceed". So I'd be in favor of having this functionality!

PR #10008, opened 2025-07-31
Enterprise sso

chuckbutkus:
Main changes are to add the new button and login logic for the new login method (enterprise_sso).

Also had to change when the logout button is available (if user is authenticated, not if user data is available as there is no user data in the new case).

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:68f2e60-nikolaik   --name openhands-app-68f2e60   docker.all-hands.dev/all-hands-ai/openhands:68f2e60
```


chuckbutkus:
@OpenHands please fix the failing actions on PR https://github.com/All-Hands-AI/OpenHands/pull/10008

chuckbutkus:
@OpenHands please fix the failing tests below on PR https://github.com/All-Hands-AI/OpenHands/pull/10008
 ❯ __tests__/components/user-actions.test.tsx (9 tests | 9 failed) 31ms
   × UserActions > should render 14ms
     → No QueryClient set, use QueryClientProvider to set one
   × UserActions > should toggle the user menu when the user avatar is clicked 2ms
     → No QueryClient set, use QueryClientProvider to set one
   × UserActions > should call onLogout and close the menu when the logout option is clicked 1ms
     → No QueryClient set, use QueryClientProvider to set one
   × UserActions > should NOT show context menu when user is undefined and avatar is clicked 2ms
     → No QueryClient set, use QueryClientProvider to set one
   × UserActions > should show context menu even when user has no avatar_url 1ms
     → No QueryClient set, use QueryClientProvider to set one
   × UserActions > should NOT be able to access logout when no user is provided 1ms
     → No QueryClient set, use QueryClientProvider to set one
   × UserActions > should handle user prop changing from undefined to defined 1ms
     → No QueryClient set, use QueryClientProvider to set one
   × UserActions > should handle user prop changing from defined to undefined 1ms
     → No QueryClient set, use QueryClientProvider to set one
   × UserActions > should work with loading state and user provided 5ms
     → No QueryClient set, use QueryClientProvider to set one

frontend/src/hooks/query/use-git-user.ts
```diff
@@ -3,16 +3,14 @@ import React from "react";
...
 
 export const useGitUser = () => {
-  const { providers } = useUserProviders();
   const { data: config } = useConfig();
 
   const user = useQuery({
     queryKey: ["user"],
     queryFn: OpenHands.getGitUser,
-    enabled: !!config?.APP_MODE && providers.length > 0,
+    enabled: !!config?.APP_MODE, // Enable regardless of providers
```
malhotra5: I think we should also include the isAuthed hook here?

openhands/server/routes/git.py
```diff
@@ -159,13 +216,30 @@ async def get_suggested_tasks(
...
+            status_code=status.HTTP_401_UNAUTHORIZED,
+        )
+    idp = user_info.get('identity_provider')
+    if not idp:
+        return JSONResponse(
+            content='IDP not found.',
+            status_code=status.HTTP_401_UNAUTHORIZED,
+        )
+    # Enterprise SSO provider has no provider tokens
+    if idp == ProviderType.ENTERPRISE_SSO.value:
```
malhotra5: Is it possible to compare directly against `ProviderType.ENTERPRISE_SSO`?

openhands/server/routes/git.py
```diff
@@ -122,7 +162,24 @@ async def search_repositories(
...
```
malhotra5: let's also dedup this logic with help functions

enyst (dismissed):
I’m glad to see the saas is adding enterprise sso, it’s a net must for it.

Can we do this in another way, though? This code doesn’t mean anything in OpenHands. Also, it’s not a git provider, so maybe another route file is more appropriate.

Could you add this new route code to the SaaS alone?

chuckbutkus:
> I’m glad to see the saas is adding enterprise sso, it’s a net must for it.
> 
> Can we do this in another way, though? This code doesn’t mean anything in OpenHands. Also, it’s not a git provider, so maybe another route file is more appropriate.
> 
> Could you add this new route code to the SaaS alone?

I wish that were possible. We are looking at a major re-architecture in the near future of how we deal with OSS/SAAS separation.  This can be addressed at that point.

malhotra5:
> This code doesn’t mean anything in OpenHands. Also, it’s not a git provider, so maybe another route file is more appropriate.

Good point, we'll see if we can remove the code from OpenHands repo

enyst:
I think the `/info` implementation doesn’t do anything at all? All of it, I don’t mean this PR.

If that’s indeed the case, this is just API routing, maybe you can make another router and use it on the SaaS?

chuckbutkus:
> I think the `/info` implementation doesn’t do anything at all? All of it, I don’t mean this PR.
> 
> If that’s indeed the case, this is just API routing, maybe you can make another router and use it on the SaaS?

I'm looking into that now.

enyst:
I unblocked the PR, I would love it if @malhotra5 or @amanape could take a look too. I don’t typically review FE.

enyst:
Just a thought. What if we do something like:
```
repos.py
# register /api/repos or something
# define the 3 affected routes, /api/repos/info, …search, … repo

# in OpenHands: all 3 are moved out from git.py to repos.py
# on the Saas: all 3 defined in a saas-only repos.py
```
Then import the right repos.py on the saas. Or maybe I’m missing something?

One of the advantages is that you never register a route you don’t intend to have, on the saas. That could matter for security, and avoiding other unintended misses.

chuckbutkus:
> Just a thought. What if we do something like:
> 
> ```
> repos.py
> # register /api/repos or something
> # define the 3 affected routes, /api/repos/info, …search, … repo
> 
> # in OpenHands: all 3 are moved out from git.py to repos.py
> # on the Saas: all 3 defined in a saas-only repos.py
> ```
> 
> Then import the right repos.py on the saas. Or maybe I’m missing something?
> 
> One of the advantages is that you never register a route you don’t intend to have, on the saas. That could matter for security, and avoiding other unintended misses.

I've removed any route changes in this repo.  Now the routes are replaced in the SAAS repo with the appropriate ones for SAAS.  I think this is now the minimal amount of changes to this repo.  I'm not opposed to changing the endpoints for better readability if we want to do that here.

malhotra5:
We may also need to modify https://github.com/All-Hands-AI/OpenHands/blob/main/frontend/src/hooks/query/use-get-secrets.ts to be enabled by auth status instead of at least one linked git providers

chuckbutkus:
> We may also need to modify https://github.com/All-Hands-AI/OpenHands/blob/main/frontend/src/hooks/query/use-get-secrets.ts to be enabled by auth status instead of at least one linked git providers

Done

openhands/server/app.py
```diff
@@ -68,6 +69,7 @@ async def _lifespan(app: FastAPI) -> AsyncIterator[None]:
 app.include_router(manage_conversation_api_router)
 app.include_router(settings_router)
 app.include_router(secrets_router)
-app.include_router(git_api_router)
+if server_config.app_mode == AppMode.OSS:
+    app.include_router(git_api_router)
```
enyst, line 73: A routing change like this is great if we can make it. But if we exclude the entire git API router like this, then the Saas will not get `get_suggested_tasks` and other routes registered in the git router currently.

This it is exactly why I suggested to maybe separate the 3 affected endpoints into their own file (e.g. `repos.py`), so we can include / exclude that.

Are you sure excluding the _entire_ git router works OK on the Saas?

chuckbutkus: Yes, the routes are included via a new router that is always added on the SAAS side.

enyst: Sounds great! That frees the entire git router for the project to do - or clean up - as it needs!

frontend/src/hooks/query/use-get-secrets.ts
```diff
@@ -1,17 +1,17 @@
...
-  const { providers } = useUserProviders();
+  const { data: isAuthed } = useIsAuthed();
 
   const isOss = config?.APP_MODE === "oss";
 
   return useQuery({
     queryKey: ["secrets"],
     queryFn: SecretsService.getSecrets,
-    enabled: isOss || providers.length > 0,
+    enabled: isOss || isAuthed === true, // Enable regardless of providers
```
malhotra5:
```suggestion
    enabled: isOss || isAuthed, // Enable regardless of providers
```

frontend/src/hooks/query/use-git-user.ts
```diff
@@ -3,16 +3,16 @@ import React from "react";
...
 export const useGitUser = () => {
-  const { providers } = useUserProviders();
   const { data: config } = useConfig();
+  const { data: isAuthed } = useIsAuthed();
 
   const user = useQuery({
     queryKey: ["user"],
     queryFn: OpenHands.getGitUser,
-    enabled: !!config?.APP_MODE && providers.length > 0,
+    enabled: !!config?.APP_MODE && isAuthed === true, // Enable regardless of providers
```
malhotra5:
```suggestion
    enabled: !!config?.APP_MODE && isAuthed, // Enable regardless of providers
```

openhands/integrations/service_types.py
```diff
@@ -18,6 +18,7 @@ class ProviderType(Enum):
     GITHUB = 'github'
     GITLAB = 'gitlab'
     BITBUCKET = 'bitbucket'
+    ENTERPRISE_SSO = 'enterprise_sso'
```
malhotra5, line 21: We seem to be adding this new provider type but its never sent to the FE it seems. Is this purely for the BE, and how does the FE's `isAuthed` hook know about SSO?

malhotra5: nvm figured it out

frontend/src/components/features/sidebar/user-actions.tsx
```diff
@@ -26,6 +28,9 @@ export function UserActions({ onLogout, user, isLoading }: UserActionsProps) {
     closeAccountMenu();
   };
 
+  // Always show the menu for authenticated users, even without user data
+  const showMenu = accountContextMenuIsVisible && isAuthed === true;
```
malhotra5:
```suggestion
  const showMenu = accountContextMenuIsVisible && isAuthed;
```

PR #10009, opened 2025-07-31
Fix litellm_proxy model info JSON parsing error handling

xingyaoww:

Fixed an issue where OpenHands could crash when the litellm_proxy service doesn't return valid JSON or expected data structure for model information. This improves reliability when using models through the OpenHands provider.

---

This PR adds proper error handling around the JSON parsing of the litellm_proxy response in the `init_model_info` method. Previously, if the litellm_proxy service returned an invalid JSON response or didn't include the expected data structure, it would cause an unhandled exception.

The changes:
1. Added a try-except block around the `response.json()` call
2. Changed the log level from `error` to `info` for better visibility without alarming users
3. Provides a graceful fallback by setting `all_model_info` to an empty list when parsing fails

This ensures that even if the litellm_proxy service doesn't have information for the current model, the application will continue to function properly.

---

https://openhands-ai.slack.com/archives/C078L0FUGUX/p1753935514200279

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:fffb817-nikolaik   --name openhands-app-fffb817   docker.all-hands.dev/all-hands-ai/openhands:fffb817
```


enyst (approved):

pyproject.toml
```diff
@@ -39,9 +39,9 @@ types-toml = "*"
 uvicorn = "*"
 numpy = "*"
 json-repair = "*"
-browsergym-core = "0.13.3"              # integrate browsergym-core as the browsing interface
+browsergym-core = "0.13.3"                         # integrate browsergym-core as the browsing interface
 html2text = "*"
-
+deprecated = "*"
```
li-boxuan, line 44: Interesting that you fixed this bug. We have `deprecated` dependency but we never explicitly declared it. I was debugging why released 0.49.0 pypi version suddenly stopped working (due to missing "deprecated" package), and found out you just fixed it.

enyst: I guess it might be because we recently cleaned up some dependencies, maybe they were taking it in before. 🤷‍♂️

PR #10018, opened 2025-07-31
Remove redundant pytest-asyncio dependency and unnecessary reset_terminal=False parameters

neubig:
This PR addresses two issues:

1. Removes the redundant pytest-asyncio dependency from pyproject.toml as it's not necessary for the async tests in this codebase.

2. Removes unnecessary explicit `reset_terminal=False` parameters from test files since `False` is already the default value.

These changes simplify the code without changing functionality.


neubig:
Closing this PR as the changes have been pushed directly to the add-reset-terminal-option branch.

neubig:
Closing this PR as the changes have been pushed directly to the add-reset-terminal-option branch.

PR #10033, opened 2025-08-01
perf: remove the sleep before runtime initialization

hereisok:

---

Solve the problem of an extra 1s sleep during runtime initialization. The 1s sleep is to avoid the gap in the front-end WebSocket connection establishment, which may not accurately push messages. The new design checks whether there is a connection on the SIO object when pushing the first message, and shifts the wait here, improving the efficiency of message push.

---


tofarr:
Nice work!

PR #10065, opened 2025-08-03
Experiment: CI should fail due to missing dependency

li-boxuan:
If CI passes, we've got a problem.


li-boxuan:
@OpenHands Please help me fix conflicts

Note that I intentionally removed "deprecated" from pyproject.toml

PR #10066, opened 2025-08-03
feat(cli): add agent state validation to /resume command

aeft:

Previously, users could attempt to use the `/resume` command even when the agent was not paused, which would have no effect and could be confusing. This PR adds validation to ensure `/resume` only works when the agent is actually paused, providing clear feedback when the command is used inappropriately.

---

This PR enhances the `/resume` command by adding proper state validation to prevent its use when the agent is not in a paused state.

- Add `agent_state` parameter to `handle_commands` and `handle_resume_command` functions
- Implement validation to only allow `/resume` when agent is in `PAUSED` state
- Display error message when `/resume` is used in invalid states (e.g., RUNNING, FINISHED, ERROR)
- Update all related test files and add comprehensive unit tests for invalid state scenarios

---
**Link of any specific issues this addresses:** N/A

### DIFF
--- a/openhands/cli/commands.py
+++ b/openhands/cli/commands.py
@@ -127,6 +127,7 @@ async def handle_commands(
     config: OpenHandsConfig,
     current_dir: str,
     settings_store: FileSettingsStore,
+    agent_state: str,
 ) -> tuple[bool, bool, bool, ExitReason]:
     close_repl = False
     reload_microagents = False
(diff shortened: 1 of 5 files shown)


enyst (commented):
Looks good to me, thank you! Could you perhaps show a screenshot or log, that it works for you?

aeft:
Sure!

### The original version:
image


### The updated version:
image


### Pause and then resume:
image

enyst (approved):
LGTM

PR #10077, opened 2025-08-04
Fix: Display logout option even when user is not available

tofarr:

This PR fixes a bug where users who get logged out of their account couldn't log back in because the logout option wasn't displayed when there was no available user. Now, the logout option is always displayed in the user menu, allowing users to log out and log back in even when they're in a logged-out state.

---

The fix modifies the `user-actions.tsx` component to display the AccountSettingsContextMenu (which contains the logout option) regardless of whether a user is available or not. Previously, the menu was only displayed when both the menu was toggled AND a user was available (`accountContextMenuIsVisible && !!user`). This change removes the user check, allowing the menu to be displayed whenever it's toggled.

This is a simple but important fix that ensures users can recover from a logged-out state without needing to refresh or restart the application.

---

N/A

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:348e0e9-nikolaik   --name openhands-app-348e0e9   docker.all-hands.dev/all-hands-ai/openhands:348e0e9
```

### DIFF
--- a/frontend/__tests__/components/user-actions.test.tsx
+++ b/frontend/__tests__/components/user-actions.test.tsx
@@ -62,16 +62,16 @@ describe("UserActions", () => {
     ).not.toBeInTheDocument();
   });
 
-  it("should NOT show context menu when user is undefined and avatar is clicked", async () => {
+  it("should show context menu when user is undefined and avatar is clicked", async () => {
     render(<UserActions onLogout={onLogoutMock} />);
 
     const userAvatar = screen.getByTestId("user-avatar");
     await user.click(userAvatar);
 
-    // Context menu should NOT appear because user is undefined
+    // Context menu SHOULD appear even when user is undefined
     expect(
-      screen.queryByTestId("account-settings-context-menu"),
-    ).not.toBeInTheDocument();
+      screen.getByTestId("account-settings-context-menu"),
+    ).toBeInTheDocument();
   });
 
   it("should show context menu even when user has no avatar_url", async () => {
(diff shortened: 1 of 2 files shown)


hieptl:
Hi @tofarr, If a new user hasn’t connected to any Git providers yet, and they click on the avatar icon and only see a “Logout” option, it might be confusing for them.

Do we have a better way to handle this scenario?

amanape (approved):
Saw your Slack comment. Worth coming back to this issue at a later time if this PR is for unblocking

tofarr:
> Hi @tofarr, If a new user hasn’t connected to any Git providers yet, and they click on the avatar icon and only see a “Logout” option, it might be confusing for them.
> 
> Do we have a better way to handle this scenario?

I agree the experience is not optimal - but logging out when you are already logged out does a lot less harm than not being able to log out when you should be able to!

The eventual optimal solution may be to log out automatically when a 401 occurs as well as deleting the auth cookie - but this is a larger workflow change that could have unintended side effects.

tofarr:
@OpenHands Please fix the tests which are failing as a result of this change:
```
 FAIL  __tests__/components/user-actions.test.tsx > UserActions > should NOT show context menu when user is undefined and avatar is clicked
 FAIL  __tests__/components/user-actions.test.tsx > UserActions > should NOT be able to access logout when no user is provided
  FAIL  __tests__/components/user-actions.test.tsx > UserActions > should handle user prop changing from defined to undefined
```

Issue #9743, opened 2025-07-16
[Tracker] Evaluation History


xingyaoww:
This issue tracks all automatic evaluations performed within the OpenHands repo.

Needed for https://github.com/All-Hands-AI/OpenHands/pull/9742 to work.

mamoodi:
@xingyaoww there's no point in running evals on this repo anymore right? Should I remove the workflow altoghether?

xingyaoww:
@mamoodi sounds good since it is being migrated to SDK.

PR #10083, opened 2025-08-04
Add process management guidance to system prompt

xingyaoww:

This PR adds guidance to the system prompt about safely terminating processes. It advises against using general keywords with commands like `pkill -f server` or `pkill -f python` which might accidentally kill other important processes. Instead, it recommends using specific keywords or finding the exact process ID first.

---

This PR adds a new `<PROCESS_MANAGEMENT>` section to the CodeAct agent's system prompt with specific guidance on how to safely terminate processes. The guidance includes:

1. Not using general keywords with commands like `pkill -f server` or `pkill -f python`
2. Using specific keywords that uniquely identify the target process
3. Using `ps aux` to find the exact process ID (PID) first, then killing that specific PID
4. Using more targeted approaches like finding the PID from a pidfile or using application-specific shutdown commands

This addition helps prevent accidental termination of important processes when the agent is asked to kill a specific process.

---

N/A

### DIFF
--- a/openhands/agenthub/codeact_agent/prompts/system_prompt.j2
+++ b/openhands/agenthub/codeact_agent/prompts/system_prompt.j2
@@ -93,3 +93,11 @@ Your primary role is to assist users by executing commands, modifying code, and 
   - Confirm whether they want it as a separate file or just in the conversation
   - Ask if they want documentation files to be included in version control
 </DOCUMENTATION>
+
+<PROCESS_MANAGEMENT>
+* When terminating processes:
+  - Do NOT use general keywords with commands like `pkill -f server` or `pkill -f python` as this might accidentally kill other important servers or processes
+  - Always use specific keywords that uniquely identify the target process
+  - Prefer using `ps aux` to find the exact process ID (PID) first, then kill that specific PID
+  - When possible, use more targeted approaches like finding the PID from a pidfile or using application-specific shutdown commands
+</PROCESS_MANAGEMENT>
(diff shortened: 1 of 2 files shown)


openhands/agenthub/codeact_agent/prompts/system_prompt.j2
```diff
@@ -93,3 +93,11 @@ Your primary role is to assist users by executing commands, modifying code, and
   - Confirm whether they want it as a separate file or just in the conversation
   - Ask if they want documentation files to be included in version control
 </DOCUMENTATION>
+
+<PROCESS_MANAGEMENT>
+* When terminating processes:
+  - Do NOT use general keywords with commands like `pkill -f server` or `pkill -f python` as this might accidentally kill other important servers or processes
```
enyst, line 99: Is there a story behind this? 😅

xingyaoww: 😭 OpenHands actually kills itself (server/ action_execution_server) by doing `pkill -f server` 🤦 

image

enyst: Ahaha! 😂 ❤️ 

Reminds me of the old times, when I had CodeAct kill its own Jupyter in an attempt to load its changes to its own loaded agentskills 😊

xingyaoww:
Evaluation result is comparable with most recent eval, so probbaly good to merge:
https://github.com/All-Hands-AI/OpenHands/issues/9743#issuecomment-3189922293

PR #10109, opened 2025-08-06
Fix git-related test failures by making tests environment-independent

neubig:

This PR fixes git-related test failures that were occurring due to missing git user configuration in test environments. The tests now set the necessary git environment variables directly, making them more robust and environment-independent.

---
This PR addresses test failures in git-related tests that were failing with "Author identity unknown" errors. The changes include:

1. Modified `TestGitHandler` to set git environment variables (`GIT_AUTHOR_NAME`, `GIT_AUTHOR_EMAIL`, `GIT_COMMITTER_NAME`, `GIT_COMMITTER_EMAIL`) directly in the test environment
2. Updated PR title escaping tests in GitHub and GitLab modules to set the same environment variables
3. Changed git config commands to use specific config files with `-f .git/config` to avoid conflicts with global git configuration

These changes make the tests more robust and independent of the environment they run in, ensuring they pass regardless of whether global git user configuration is present.

---
N/A

### DIFF
--- a/tests/unit/resolver/github/test_pr_title_escaping.py
+++ b/tests/unit/resolver/github/test_pr_title_escaping.py
@@ -12,6 +12,15 @@ def test_commit_message_with_quotes():
     with tempfile.TemporaryDirectory() as temp_dir:
         subprocess.run(['git', 'init', temp_dir], check=True)
 
+        # Set git configuration locally for this repository
+        subprocess.run(
+            ['git', '-C', temp_dir, 'config', 'user.email', 'test@example.com'],
+            check=True,
+        )
+        subprocess.run(
+            ['git', '-C', temp_dir, 'config', 'user.name', 'Test User'], check=True
+        )
+
         # Create a test file and add it to git
         test_file = os.path.join(temp_dir, 'test.txt')
         with open(test_file, 'w') as f:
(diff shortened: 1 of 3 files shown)


tests/unit/resolver/github/test_pr_title_escaping.py
```diff
@@ -36,8 +45,20 @@ def test_commit_message_with_quotes():
...
 
-        # Make the commit
-        make_commit(temp_dir, issue, 'issue')
+        # Make the commit with author information
+        # First, modify the environment to include GIT_AUTHOR_NAME and GIT_AUTHOR_EMAIL
+        old_environ = os.environ.copy()
+        os.environ['GIT_AUTHOR_NAME'] = 'Test User'
+        os.environ['GIT_AUTHOR_EMAIL'] = 'test@example.com'
+        os.environ['GIT_COMMITTER_NAME'] = 'Test User'
+        os.environ['GIT_COMMITTER_EMAIL'] = 'test@example.com'
```
enyst, line 54: Are both author and committer necessary? I thought we only use author

neubig:
@openhands follow the py-tests.yml workflow locally, running all the unit tests. Remember to set a long timeout when you run execute_bash so that the test running does not time out. Read all comments on this PR and respond about what tests were failing before this PR was implemented, and how this PR fixes them.

PR #10119, opened 2025-08-06
Add BatchedWebHookFileStore for batching webhook updates

tofarr:

This PR introduces batched webhook functionality for file store operations, which significantly improves performance when many files are being modified in a short period of time. Instead of sending individual webhook requests for each file operation, the system now batches multiple operations together and sends them as a single request, reducing network overhead and improving efficiency.

Users can enable this feature by setting `file_store_web_hook_batch = true` in their configuration. The batching behavior can be fine-tuned using environment variables to control timeout periods and size limits.

---

This PR implements a new `BatchedWebHookFileStore` class that wraps the existing file store functionality and adds intelligent batching of webhook requests. Key features include:

**Core Implementation:**
- **BatchedWebHookFileStore**: A new file store implementation that collects file operations (write/delete) and sends them as batched HTTP requests
- **Configurable batching**: Operations are batched and sent when either a timeout (default: 5 seconds) or size limit (default: 1MB) is reached
- **Thread-safe design**: Uses locks and timers to ensure thread-safe access to the batch queue

**Configuration Options:**
- Added `file_store_web_hook_batch` boolean config option to enable batched mode (default: false for backward compatibility)
- Environment variables for fine-tuning:
  - `WEBHOOK_BATCH_TIMEOUT_SECONDS`: Time after which a batch is sent (default: 5)
  - `WEBHOOK_BATCH_SIZE_LIMIT_BYTES`: Size limit after which a batch is sent (default: 1MB)

**Protocol Design:**
- **Batch Request Format**: Single POST request to the base webhook URL with JSON array payload
- **Operation Deduplication**: Multiple operations on the same file are consolidated (latest operation wins)
- **Binary Content Handling**: Binary files are automatically base64-encoded with appropriate metadata
- **Error Handling**: Includes retry logic (up to 3 attempts with 1-second delays)

**Key Design Decisions:**
1. **Opt-in Feature**: Batching is disabled by default to maintain backward compatibility
2. **Flexible Triggering**: Batches are sent based on either time or size thresholds, whichever comes first
3. **Operation Consolidation**: When the same file is modified multiple times, only the final state is sent
4. **Graceful Degradation**: If batching fails, the underlying file operations still succeed locally

**Testing:**
- Comprehensive unit tests covering batching logic, size limits, timeouts, binary content, and error scenarios
- Tests verify thread safety and proper cleanup of timers and resources

---

This addresses performance issues in scenarios where many file operations trigger individual webhook requests, such as during large code generation tasks or bulk file processing operations.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:05cd37a-nikolaik   --name openhands-app-05cd37a   docker.all-hands.dev/all-hands-ai/openhands:05cd37a
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@feature/batched-webhook-file-store openhands
```


openhands/storage/batched_web_hook.py
```diff
@@ -0,0 +1,273 @@
...
+            path: The path that was modified
+            operation: The operation performed ("write" or "delete")
+            contents: The contents that were written (None for delete operations)
+        """
+        with self._batch_lock:
+            # Calculate content size
+            content_size = 0
+            if contents is not None:
+                if isinstance(contents, str):
+                    content_size = len(contents.encode('utf-8'))
```
raymyers, line 147: I think the encode will make a copy of the string (and events can be big), so there's potentially a tradeoff of estimating the size rather than spending the memory getting the exact size. Not a show stopper.

Issue #9344, opened 2025-06-25
Bug: [CLI] Iteration limit enforced at default despite config.toml setting max_iterations=1000


enyst:
## Bug Description

When running OpenHands from CLI with a custom `config.toml` that sets `max_iterations = 1000`, the system still stops execution at 500 iterations (the default limit) with an iteration limit error.

## Environment
- **Installation**: Development version installed with `pip install openhands -e .`
- **Command**: `openhands -n vscode-runtime` (run from project directory)
- **Config**: Custom `config.toml` with `max_iterations = 1000`
- **Verification**: Other config options from the same file are being respected

## Expected Behavior
The system should respect the `max_iterations = 1000` setting from `config.toml` and allow up to 1000 iterations before stopping.

## Actual Behavior
The system stops at 500 iterations with an iteration limit error, ignoring the custom configuration.

## Steps to Reproduce
1. Create/modify `config.toml` with `max_iterations = 1000`
2. Install development version: `pip install openhands -e .`
3. Run: `openhands -n vscode-runtime`
4. Observe that execution stops at 500 iterations despite config setting

## Additional Context
- The config file is definitely being read (other settings are applied)
- This affects workflows where longer sessions are needed

---
*Reported by: OpenHands-Claude (AI Assistant)*

aeft:
Can I work on this issue?

I think the problem is that default values are set when parsing arguments. For example, in `core/config/utils.py:get_parser`, the default value for `--max-iterations` is 500. In `core/config/utils.py:setup_config_from_args`, we can see that CLI arguments (e.g., `-l` for LLM config) take precedence over `config.toml`, which causes the values in `config.toml` to be overridden by the default values from the CLI arguments.

```py
parser.add_argument(
    '-i',
    '--max-iterations',
    default=OH_MAX_ITERATIONS,
    type=int,
    help='The maximum number of iterations to run the agent',
)
```

```py
if args.max_iterations is not None:
    config.max_iterations = args.max_iterations
```

Besides `max_iterations`, `agent_cls` has the same issue.

My initial idea was that we should remove the default values from the CLI arguments (i.e., the default should always be None for config-related arguments, such as `--agent-cls` and `--max-iterations`). Instead, we should set the default values in `setup_config_from_args`. We can set the defaults at the end of `setup_config_from_args`.

[...]

PR #10140, opened 2025-08-07
fix: prevent CLI argument parser defaults from overriding config file values

aeft:

---
This PR fixes an issue where CLI argument parser defaults were overriding values from config files.

The fix modifies `get_parser()` to conditionally set argparse defaults, preventing them from taking precedence over config file values. Since `OpenHandsConfig` already provides appropriate defaults during initialization, we don't need to explicitly set them again. This ensures the intended precedence: CLI args > config files > OpenHandsConfig defaults.

### Before
image

### After
image

---
Fix https://github.com/All-Hands-AI/OpenHands/issues/9344

### DIFF
--- a/openhands/core/config/utils.py
+++ b/openhands/core/config/utils.py
@@ -20,10 +20,6 @@ from openhands.core.config.condenser_config import (
     condenser_config_from_toml_section,
     create_condenser_config,
 )
-from openhands.core.config.config_utils import (
-    OH_DEFAULT_AGENT,
-    OH_MAX_ITERATIONS,
-)
 from openhands.core.config.extended_config import ExtendedConfig
 from openhands.core.config.kubernetes_config import KubernetesConfig
 from openhands.core.config.llm_config import LLMConfig
@@ -697,14 +693,14 @@ def get_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         '-c',
         '--agent-cls',
-        default=OH_DEFAULT_AGENT,
+        default=None,
         type=str,
         help='Name of the default agent to use',
     )
     parser.add_argument(
         '-i',
         '--max-iterations',
-        default=OH_MAX_ITERATIONS,
+        default=None,
         type=int,
         help='The maximum number of iterations to run the agent',
     )
--- a/tests/unit/test_config_precedence.py
+++ b/tests/unit/test_config_precedence.py
@@ -3,6 +3,8 @@ from unittest.mock import MagicMock, patch
 import pytest
 
 from openhands.core.config import (
+    OH_DEFAULT_AGENT,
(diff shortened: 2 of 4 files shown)


openhands/core/config/utils.py
```diff
@@ -697,14 +704,14 @@ def get_parser() -> argparse.ArgumentParser:
     parser.add_argument(
         '-c',
         '--agent-cls',
-        default=OH_DEFAULT_AGENT,
+        default=OH_DEFAULT_AGENT if enable_config_default else None,
```
xingyaoww: I wonder if we can just set the default value for CLI to None here, since it will fall back to default in the openhands config?

aeft: Yeah, this is definitely a cleaner solution. But like I mentioned in the issue comment, some evaluation scripts (run_infer.py) may rely on the default value provided by get_parser(), so it might break them. Do you think it's okay to break them?

xingyaoww: Yep i think that should be fine! I believe eval script will still fallback to the default value in OpenHandsCOnfig

aeft: evaluation/benchmarks/ml_bench/run_infer.py:300
evaluation/benchmarks/lca_ci_build_repair/run_infer.py:396

I checked some, e.g., there will use args as their metadata. 
image

And construct output_path:
image

xingyaoww: @aeft actually i think `--agent-cls` will be explictly passed in by `run_infer.sh`, so i think it is safe to set it to None

xingyaoww:
@aeft can you fix the tes?

aeft:
Sure. I am looking at it.

aeft:
@xingyaoww hi, tests are fixed and the pr is ready to review when you have time. thanks!

tests/unit/test_config_precedence.py
```diff
@@ -308,3 +308,85 @@ def test_cli_settings_json_not_override_config_toml(
...
+
+    assert config.default_agent == OH_DEFAULT_AGENT
+    assert config.max_iterations == OH_MAX_ITERATIONS
+
+
+def test_cli_args_override_defaults():
+    """Test that CLI arguments override default values."""
+    from unittest.mock import MagicMock
+
+    from openhands.core.config.utils import setup_config_from_args
```
xingyaoww: nit: Can we do these imports on the top of file?

xingyaoww (approved):

tests/unit/test_config_precedence.py
aeft: Sure

xingyaoww (approved):

PR #10141, opened 2025-08-07
fix(misc): MCP settings and other UI improvements/fixes

### LINKED ISSUE #9772: [Bug]: MCP settings-related issues

I encountered a few small issues related to the MCP settings UI.

~~1. If the user refreshes the MCP settings tab with F5, the page shows default config values (empty/default) instead of the actual saved config. Navigating to another tab and then back correctly loads the data.~~
~~2. The default MCP configuration is localized per language, which is unnecessary because it's a JSON value, and JSON can't have comments that could require localization.~~
~~3. Both the docs and the default values mention an shttp_servers key, but it's not present in the editor example nor in the localized defaults. Still, I've seen it appear, so I suspect the settings are modified at runtime. It either needs to be added as a valid key (and shown in the examples) or removed if unsupported. Also, if the MCP settings are indeed modified at runtime, I don't think that should happen, but if it's expected, then having a default value in the editor probably isn't necessary.~~
~~4. The docs linked in the settings page aren't very helpful: they only show config.toml examples.~~
~~5. I submitted a PR to improve the UI (#9771); however, having to still press two buttons to save, "Confirm Changes" (formerly "Apply Settings") and "Save Settings", is still a bit confusing. A cleaner solution might be needed (an unsaved changes modal prompt, maybe?).~~

Edit: Issues 1, 2 and 5 partially fixed by #9772.
Edit 2: Issues 3 and 4 no longer relevant after #10145.

### OpenHands Installation

Docker command in README

### OpenHands Version

0.49

### Model Name

_No response_

### Operating System

None

_No response_

llamantino:

---

### MCP settings improvements/fixes
- Adds autofocus to the MCP JSON editor when clicking the "Edit configuration" button
- Fixes reloading the MCP settings page with F5 showing the defaults rather than the current settings
- Renames "Confirm changes" to "Preview changes" to make it more clear that it does not save/apply the settings
- Removes the unnecessary localization of the JSON default values for MCP servers

### Misc UI improvements
- Adds a 'Lock' icon to the Logout menu item in the account settings context menu

### Misc small fixes
- Fixes full-width clickable area of the "Settings" button in the Home screen
- Fixes full-width clickable area of the "Connect a Git provider to manage secrets" button in the Settings>Secrets screen
- Fixes "Agent encountered an error." showing a green status color rather than red

### Screenshots (issues fixed)
Full-width clickable "Settings" button issue:
Long button 1
Full-width clickable "Connect a Git provider to manage secrets" button issue:
Long button 2
Agent encountered an error showing green status color:
Agent encountered an error wrong color

### Screenshots (UI changes)
New logout icon for the account settings context menu:
Logout icon

---
Partially fixes #9772


frontend/src/components/features/settings/mcp-settings/mcp-json-editor.tsx
```diff
@@ -11,6 +11,11 @@ interface MCPJsonEditorProps {
   onCancel: () => void;
 }
 
+const MCP_DEFAULT_CONFIG: MCPConfig = {
+  sse_servers: [],
+  stdio_servers: [],
```
xingyaoww, line 16: Can we also add `shttp_servers` here?

frontend/src/routes/mcp-settings.tsx
```diff
@@ -18,11 +18,15 @@ function MCPSettingsScreen() {
...
-    settings?.MCP_CONFIG,
-  );
+  const [mcpConfig, setMcpConfig] = useState<MCPConfig | undefined>(undefined);
   const [isDirty, setIsDirty] = useState(false);
 
+  useEffect(() => {
+    if (!mcpConfig && settings?.MCP_CONFIG) {
+      setMcpConfig(settings.MCP_CONFIG);
+    }
+  }, [settings, mcpConfig]);
```
xingyaoww, line 28: cc @amanape not sure if this violates FE convention we are using?

xingyaoww (commented):
Mostly LGTM! But might need @amanape for the final confirmation

frontend/src/components/features/settings/mcp-settings/mcp-json-editor.tsx
llamantino: I can do that, but I'll have to turn it back into a string. I just noticed that shttp_servers is not defined in MCPConfig, nor is it validated by the editor (although it's saved and loaded fine).

frontend/src/routes/mcp-settings.tsx
```diff
@@ -18,11 +18,15 @@ function MCPSettingsScreen() {
   const { data: settings, isLoading } = useSettings();
   const { mutate: saveSettings, isPending } = useSaveSettings();
 
-  const [mcpConfig, setMcpConfig] = useState<MCPConfig | undefined>(
-    settings?.MCP_CONFIG,
-  );
+  const [mcpConfig, setMcpConfig] = useState<MCPConfig | undefined>(undefined);
```
amanape, line 21:
```suggestion
  const [mcpConfig, setMcpConfig] = useState<MCPConfig>();
```

When you use `useState<MCPConfig>()` without providing an initial value, it implicitly sets the state to undefined.

Both are corrects solutions, but this one feels more compact 😄

frontend/src/routes/mcp-settings.tsx
amanape: Yep, all good

amanape (approved):
Thank you!

PR #10155, opened 2025-08-07
Non root user

chuckbutkus:
This PR is to allow the runtime pods to be run as a non-root (openhands) OS user. The main changes are in Dockerfile.j2 as that is what builds the runtime image. The OH files were changed to be owned by the "openhands" user so they can be accessed when run as that user (root still has access).  Some of the package installs were modified as well so they can be used by either "root" or "openhands".

I also changed the OS group name from "app" to "openhands" in the OH dockerfile as it helps using it later if it has the same name as the OS user.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:dd6af21-nikolaik   --name openhands-app-dd6af21   docker.all-hands.dev/all-hands-ai/openhands:dd6af21
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@non-root-user openhands
```


tests/unit/test_runtime_build.py
```diff
@@ -516,7 +541,10 @@ def test_init(docker_runtime_builder):
     assert docker_runtime_builder.rolling_logger.log_lines == [''] * 10
 
 
-def test_build_image_from_scratch(docker_runtime_builder, tmp_path):
+@pytest.mark.skipif(not is_docker_available(), reason='Docker not available')
```
rbren: is this maybe a change openhands made to get the tests to pass?

chuckbutkus: Yes.

rbren: I think we can revert it then

PR #10160, opened 2025-08-07
Fix "fatal: bad revision HEAD" error in VS Code extension build script

xingyaoww:

Fixes the "fatal: bad revision HEAD" error that occurs when installing OpenHands via uvx or other package managers that create shallow git repositories. The VS Code extension build script now gracefully handles problematic git repository states (shallow repos, empty repos, invalid HEAD) by skipping the extension build when git operations would fail.

---

This PR enhances the `build_vscode.py` script to handle problematic git repository states that can occur during package installation:

1. **Git Repository State Validation**: Added `check_git_repository_state()` function that validates git repository state with timeout protection and detects:
   - Missing or invalid git repositories
   - Shallow repositories (common with uvx, pip installs from git)
   - Empty repositories with invalid HEAD references
   - Git command failures and timeouts

2. **Intelligent Build Skipping**: The script now skips VS Code extension building when:
   - `SKIP_VSCODE_BUILD=1` environment variable is set
   - Git repository state is problematic (shallow, empty, or invalid HEAD)
   - Git operations would fail with "fatal: bad revision HEAD" or similar errors

3. **Graceful Error Handling**: Instead of failing the entire build process, the script logs the reason for skipping and continues, allowing CLI-only installations to succeed.

4. **Timeout Protection**: All subprocess operations (git commands, npm operations) are protected with timeouts to prevent hanging builds.

The core issue was that package managers like uvx create shallow git repositories for installation, and the original build script would fail when trying to resolve HEAD in these repositories. This fix ensures that OpenHands can be installed successfully via uvx and other package managers while still building the VS Code extension in proper development environments.

---

Addresses the "fatal: bad revision HEAD" error reported when installing OpenHands via uvx from the add-ui-launcher-flag branch.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:00ad6f9-nikolaik   --name openhands-app-00ad6f9   docker.all-hands.dev/all-hands-ai/openhands:00ad6f9
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix/vscode-build-git-repository-state openhands
```


enyst:
I’d love to take a look, maybe after we merge the other 2 PRs?

I’m a bit confused by the agent’s description, it ran for me in worktrees with uvx

docs/usage/how-to/gui-mode.mdx
```diff
@@ -7,6 +7,67 @@ description: High level overview of the Graphical User Interface (GUI) in OpenHa
 
 - [OpenHands is running](/usage/local-setup)
 
+## Launching the GUI Server
```
mamoodi: A lot of this is duplicated material. The how to set it up should go to the local-setup doc.

docs/usage/local-setup.mdx
```diff
@@ -66,6 +66,30 @@ A system with a modern processor and a minimum of **4GB RAM** is recommended to
 
 ### Start the App
 
+#### Option 1: Using the CLI Launcher (Recommended)
+
```
mamoodi: It says this is the recommended way to run OH. However, the first line says "If you have Python installed..."

This needs to be thought out more imo since this saying this approach is better than the docker command we have in the README. If it's the better approach then the docs need to be revisited to make this flow better and users are able to run through it easily.

xingyaoww:
@OpenHands please resolve the conflict and merge from main

docs/usage/local-setup.mdx
enyst: I think `openhands` is a better approach. Yes, we can require python by default, almost everyone has it anyway.

And we can allow for an additional option where, if the user doesn’t have it and doesn’t want it, they can also run the docker run command.

That means the order should be the other way around. Do you feel it’s too early to change yet?

enyst: Just to note further, I don’t know why the agent did this here, or why Xingyao may want this. But I think it’s slightly better.

I think it will be the better approach on the longer run: mainly because I think we need multiple options, and many many projects out there prove that running on python is normal, it’s not a big hurdle. And I really think we need multiple ways. The docker run + web ui is limited in ways that people keep missing, and they keep asking for.

xingyaoww:
@openhands please this fix this in the CI of this PR https://github.com/All-Hands-AI/OpenHands/pull/10160

Run # Verify the .vsix was created and is valid
❌ VSCode extension build failed - .vsix not found
Error: Process completed with exit code 1.

xingyaoww:
OK i don't think it actually fixed "fatal: bad revision 'HEAD'"

closing it

PR #10170, opened 2025-08-08
feat(frontend): implement “stop server” and “start server” button functionality.

hieptl:

We need to implement support for the “Stop Server” and “Start Server” buttons according to the design specifications provided below.

**Acceptance Criteria:**
- The “Stop Server” button should correctly stop the server when clicked.
- The “Start Server” button should correctly start the server when clicked.

---

The PR implements “stop server” and “start server” button functionality.

https://github.com/user-attachments/assets/c1920bf8-39aa-4d82-9349-e0185b04b35b

---


hieptl:
> I really like this. I don't think it's obvious, so we may want to document it as best we can, but it's not unusual either, to have a clickable status. Great way to keep space uncluttered.
> 
> I like the code design too (the connection to backend)

Hi @enyst Thank you so much! Glad you like it. Totally agree that we should document the clickable status better 🙏

frontend/__tests__/components/features/conversation/server-status.test.tsx
```diff
@@ -0,0 +1,421 @@
...
+  });
+
+  it("should render server status with starting state", () => {
+    // For this test, we'll test the component with STARTING conversation status
+    renderWithProviders(<ServerStatus conversationStatus="STARTING" />);
+
+    // The component should render with the default RUNNING state from agent
+    const statusText = screen.getByText("Running");
+    expect(statusText).toBeInTheDocument();
+  });
```
amanape: These could all be included in one test that tests component state

frontend/__tests__/components/features/conversation/server-status.test.tsx
```diff
@@ -0,0 +1,421 @@
...
+
+  it("should apply custom className", () => {
+    renderWithProviders(
+      <ServerStatus conversationStatus="RUNNING" className="custom-class" />,
+    );
+
+    const container = screen.getByText("Running").closest("div")?.parentElement;
+    expect(container).toHaveClass("custom-class");
+  });
+
```
amanape:
```suggestion
```

Avoid testing style-related changes since we change them so much 🙃

frontend/__tests__/components/features/conversation/server-status.test.tsx
```diff
@@ -0,0 +1,421 @@
...
+        conversationStatus="RUNNING"
+        onStopServer={vi.fn()}
+        position="top"
+      />,
+    );
+
+    const contextMenu = screen.getByTestId("server-status-context-menu");
+    expect(contextMenu).toHaveClass("bottom-full");
+  });
+
```
amanape:
```suggestion
```

frontend/__tests__/components/features/conversation/server-status.test.tsx
```diff
@@ -0,0 +1,421 @@
...
+        conversationStatus="RUNNING"
+        onStopServer={vi.fn()}
+        position="bottom"
+      />,
+    );
+
+    const contextMenu = screen.getByTestId("server-status-context-menu");
+    expect(contextMenu).toHaveClass("top-full");
+  });
+
```
amanape:
```suggestion
```

frontend/__tests__/components/features/conversation/server-status.test.tsx
```diff
@@ -0,0 +1,421 @@
...
+      <ServerStatusContextMenu
+        {...defaultProps}
+        conversationStatus="RUNNING"
+        onStopServer={vi.fn()}
+      />,
+    );
+
+    const contextMenu = screen.getByTestId("server-status-context-menu");
+    expect(contextMenu).toHaveClass("w-fit", "min-w-max", "bg-tertiary");
+  });
```
amanape:
```suggestion
```

frontend/src/components/features/chat/chat-interface.test.tsx
```diff
@@ -25,6 +25,7 @@ vi.mock("react-redux", () => ({
     selectedRepository: null,
     replayJson: null,
   })),
+  useDispatch: vi.fn(() => vi.fn()),
```
amanape: Is this to make it easier to test? I prefer not to pass the `useDispatch` hook as prop since we can call it from within the hook instead

frontend/src/components/features/controls/server-status.tsx
```diff
@@ -44,15 +61,49 @@ export function ServerStatus({
...
+    event.preventDefault();
+    dispatch(setShouldStopConversation(true));
+    setShowContextMenu(false);
+  };
+
+  const handleStartServer = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    dispatch(setShouldStartConversation(true));
+    setShowContextMenu(false);
+  };
```
amanape: Maybe instead of updating state to trigger a useEffect, we can directly call the mutation hooks hat stop or start the server here?

hieptl: I initially used the hook right from the start. You can see more details in [this commit](https://github.com/All-Hands-AI/OpenHands/pull/10170/commits/29ec35b65430e308f972c9319ddf580dcbf719d6).

However, I encountered the following issue in the pipeline, which led me to switch to the current solution:

> useConversationId must be used within a route that includes a conversationId parameter.

frontend/__tests__/components/features/conversation/server-status.test.tsx
hieptl: I updated the code based on your feedback. Thank you for your feedback! 🙏

frontend/src/components/features/chat/chat-interface.test.tsx
hieptl: I think that removing the useDispatch: vi.fn(() => vi.fn()) mock is challenging because:
- The ServerStatus component calls useDispatch() in its implementation.
- Removing the mock causes all tests to fail with the error: “No ‘useDispatch’ export is defined on the ‘react-redux’ mock”.
- The component relies on dispatch to trigger Redux actions such as setShouldStopConversation and setShouldStartConversation.

Let me know if you have an alternative approach in mind. Thank you! 🙏

frontend/src/components/features/controls/server-status.tsx
amanape: I’ve just added a `const { conversationId } = useConversationId();` line to this file on my local machine and it doesn’t throw the error, maybe it was a confused compiler?

This shouldn’t throw because we render this component in the conversations screen, and we fetch the ID from the URL of the route this page is rendered in, which should always be available. Could you try once more and @ me if you encounter any issues so we can investigate them together?

hieptl: I've updated the code based on your feedback. Thank you for your feedback! 🙏

frontend/src/components/features/chat/chat-interface.test.tsx
hieptl: I've updated the code based on your feedback. Thank you for your feedback! 🙏

frontend/src/components/features/controls/server-status.tsx
```diff
@@ -44,15 +70,66 @@ export function ServerStatus({
...
+
+  const handleStopServer = (event: React.MouseEvent<HTMLButtonElement>) => {
+    event.preventDefault();
+    stopConversationMutation.mutate(
+      { conversationId },
+      {
+        onSuccess: () => {
+          refetch();
```
amanape: Instead of refetching here and below, you can invalidate the query `queryKey: ["user", "conversation", cid]` and it should automatically trigger a refetch

amanape (changes requested):

frontend/src/components/features/chat/chat-interface.test.tsx
amanape: I’ve misread the changes and thought this was a component and not a test file, sorry. As long as we don’t have the component accept a `useDispatch` prop, his should be good

frontend/src/components/features/controls/server-status.tsx
hieptl: I've updated the code based on your feedback. Thank you for your feedback! 🙏

amanape (approved):

PR #10171, opened 2025-08-08
feat: add mcp_config to conversation initialization

tofarr:
---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:11aabf2-nikolaik   --name openhands-app-11aabf2   docker.all-hands.dev/all-hands-ai/openhands:11aabf2
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@feat/mcp-shttp-servers-override openhands
```

### DIFF
--- a/openhands/server/services/conversation_service.py
+++ b/openhands/server/services/conversation_service.py
@@ -85,9 +85,7 @@ async def create_new_conversation(
     session_init_args['git_provider'] = git_provider
     session_init_args['conversation_instructions'] = conversation_instructions
     if mcp_shttp_servers:
-        session_init_args['mcp_config'] = MCPConfig(
-            shttp_servers=mcp_shttp_servers
-        )
+        session_init_args['mcp_config'] = MCPConfig(shttp_servers=mcp_shttp_servers)
 
     session_init_args['mcp_shttp_servers'] = mcp_shttp_servers
     conversation_init_data = ConversationInitData(**session_init_args)
--- a/openhands/server/services/conversation_service.py
+++ b/openhands/server/services/conversation_service.py
@@ -2,7 +2,7 @@ import uuid
 from types import MappingProxyType
 from typing import Any
 
-from openhands.core.config.mcp_config import MCPConfig, MCPSHTTPServerConfig
+from openhands.core.config.mcp_config import MCPConfig
 from openhands.core.logger import openhands_logger as logger
 from openhands.events.action.message import MessageAction
 from openhands.experiments.experiment_manager import ExperimentManagerImpl
--- a/openhands/server/services/conversation_service.py
+++ b/openhands/server/services/conversation_service.py
@@ -2,6 +2,7 @@ import uuid
 from types import MappingProxyType
 from typing import Any
 
+from openhands.core.config.mcp_config import MCPSHTTPServerConfig
 from openhands.core.logger import openhands_logger as logger
 from openhands.events.action.message import MessageAction
 from openhands.experiments.experiment_manager import ExperimentManagerImpl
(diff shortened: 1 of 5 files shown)


openhands/server/session/conversation_init_data.py
```diff
@@ -17,6 +18,7 @@ class ConversationInitData(Settings):
     selected_branch: str | None = Field(default=None)
     conversation_instructions: str | None = Field(default=None)
     git_provider: ProviderType | None = Field(default=None)
+    mcp_shttp_servers: list[MCPSHTTPServerConfig] | None = Field(default=None)
```
xingyaoww: We already have `mcp_config` and `shttp_servers` is just `mcp_config.shttp_servers`. Can we just re-use `mcp_config` to deduplicate stuff?
https://github.com/All-Hands-AI/OpenHands/blob/9b0a5da839256ca32076509a98c30d9854ab55ff/openhands/storage/data_models/settings.py#L42

openhands/server/services/conversation_service.py
```diff
@@ -44,6 +45,7 @@ async def create_new_conversation(
     attach_convo_id: bool = False,
     git_provider: ProviderType | None = None,
     conversation_id: str | None = None,
+    mcp_shttp_servers: list[MCPSHTTPServerConfig] | None = None,
```
xingyaoww: I mean can we just use `mcp_config` which is already defined in `Settings` -- we don't really need this `mcp_shttp_servers` field which duplicates the part of the `mcp_config`.

tofarr: Good call! Done!

openhands/server/session/session.py
```diff
@@ -152,18 +154,26 @@ async def initialize_agent(
...
+        if mcp_config is not None:
+            # Use the provided MCP SHTTP servers instead of default setup
+            self.config.mcp = mcp_config
+            self.logger.debug(f'Using custom MCP Config: {mcp_config}')
+        else:
+            # Add OpenHands' MCP server by default
+            openhands_mcp_server, openhands_mcp_stdio_servers = (
+                OpenHandsMCPConfigImpl.create_default_mcp_server_config(
+                    self.config.mcp_host, self.config, self.user_id
+                )
```
xingyaoww: We should merge the settings mcp with the openhands MCP, instead of if/else (you can just remove the if-else and append openhands mcp config stuff to the same field as done below.

tofarr: Are you sure about this? If so it means that we would always have http://localhost:3000 added to the MCP servers in OpenSource / nested runtimes...

tofarr: Feels to me that if somebody goes to the trouble of overriding this variable they would expect what they define to be what's used?

xingyaoww: @tofarr yes -- that's where our github MCPs are defined (by @malhotra5 )

tofarr: Done!

openhands/server/session/session.py
```diff
@@ -152,6 +154,14 @@ async def initialize_agent(
         self.logger.debug(
             f'MCP configuration before setup - self.config.mcp_config: {self.config.mcp}'
         )
+
+        # Check if settings has custom mcp_config
+        mcp_config = getattr(settings, 'mcp_config', None)
+        if mcp_config is not None:
+            # Use the provided MCP SHTTP servers instead of default setup
+            self.config.mcp = mcp_config
+            self.logger.debug(f'Using custom MCP Config: {mcp_config}')
```
xingyaoww: on a second thought, i wonder if we can simply merge self.config.mcp with `settings.mcp_config` instead of override? Otherwise this could break the OSS when someone defines MCP in config.toml, while still want to use MCP defined in settings.

tofarr: Done!

xingyaoww (approved):
LGTM!

PR #10184, opened 2025-08-09
feat(agent): stop using short tool description for gpt-5

xingyaoww:

---

Previously OpenAI enforces max function description length of 1k characters (https://community.openai.com/t/function-call-description-max-length/529902). Because of that we maintain a short version for each function description.

But according to a more recent post, it seems have gone away: https://community.openai.com/t/was-the-character-limit-for-schema-descriptions-upgraded/1225975

I tested locally, and gpt-5 doesn't seem to throw error at long function descriptions, so maybe safe to remove them now.

---


enyst (approved):
LGTM, thank you!

Issue #10158, opened 2025-08-07
Add "stop" boolean checkbox to Advanced LLM UI


createthis:
**What problem or use case are you trying to solve?**

We've noticed that DeepSeek-R1-0528 and other reasoning LLMs will put `<function></function>` tags in their `reasoning_content` as they attempt to work through their next steps. This immediately halts the conversation and makes it appear as if the `<think>` tag never closed and there is no `content` property in the result.

Some of these models should probably be added here: https://github.com/All-Hands-AI/OpenHands/blob/3a629cdf081cc759b006952c2ba8335251d1bf87/openhands/llm/llm.py#L112

However, naming is hit or miss with third party API providers and if the user is using a local LLM via `llama.cpp` they could name the model "fluffy bunny 123" if they want. We need a more reliable way to allow the user to disabled the `stop` property in the request.

**Describe the UX or technical implementation you have in mind**

We currently have toggle switched in the Advanced LLM dialog. I suggest adding another toggle switch for `Reasoning Model` that will disabled the `stop` request property when "checked". We could also name it `stop` or `no stop`. I'm not great at naming things. I just feel like `Reasoning Model` might be more flexible moving forward.

**Additional context**

This also affects:
- https://github.com/All-Hands-AI/OpenHands/issues/8218
- https://github.com/All-Hands-AI/OpenHands/pull/9374
- https://github.com/All-Hands-AI/OpenHands/issues/9370

createthis:
I just confirmed this patch solves the problem for DeepSeek-R1-0528 when I set the model name as `deepseek/DeepSeek-R1-0528:671b-Q4_K_XL` in advanced settings (I have not tested for the other models in the patch yet):

```diff
diff --git a/openhands/llm/llm.py b/openhands/llm/llm.py
index 1db12ee32..8cf743ba2 100644
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -110,6 +110,10 @@ MODELS_WITHOUT_STOP_WORDS = [
     'o1',
     'o1-2024-12-17',
     'xai/grok-4-0709',
+    'deepseek/DeepSeek-R1-0528:671b-Q4_K_XL',
+    'deepseek/DeepSeek-R1-0528:671b-q4_k_m',
+    'openai/gpt-oss-120b-F16',
+    'openai/GLM-4.5-GGUF:q4_k_xl'
 ]
 
 
```

[...]

PR #10188, opened 2025-08-10
Add reasoning switch to advanced llm UI 2

createthis:

This PR adds a switch in the Advanced LLM UI to flag a model as a Reasoning model. When this switch is on, Open Hands will not populate the `stop` property in the request. This addresses #10158.


openhands/llm/llm.py
```diff
@@ -259,8 +262,11 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
                 )
                 kwargs['messages'] = messages
 
-                # add stop words if the model supports it
-                if self.config.model not in MODELS_WITHOUT_STOP_WORDS:
+                # add stop words if the model supports it and reasoning is not enabled
+                if (
+                    self.config.model not in MODELS_WITHOUT_STOP_WORDS
+                    and not self.enable_reasoning
+                ):
```
xingyaoww, line 269: Actually i'm confused -- we by default using reasoning effort high for all models that actually supports reasoning -- why do we need this tag though

createthis: This isn't about reasoning effort high/medium/low. It's about the STOP_WORDS. Feel free to suggest any name you see fit for the switch and variable name and I'll refactor.

xingyaoww: ah i understand the context better after reading https://github.com/All-Hands-AI/OpenHands/issues/10158

I wonder if we can make that an extra field in `llm_config` rather than settings? We have a known problem of having a parallel/hard-to-maintain configuration system and i hope we don't add more stuff to settings unless it is absolutely necessary: https://github.com/All-Hands-AI/OpenHands/issues/9585

Since this is more kind of custom LLM / advanced setting -- can we maybe add a field like `disable_vision` we currently have (https://github.com/All-Hands-AI/OpenHands/blob/57a3d8f17dcbde18d42d9672c7156c9573a94a78/openhands/core/config/llm_config.py#L82), and set it via environment variabe `LLM_DISABLE_STOP_WORD` instead of messing with the UI though (i imagine 90% of the user would probably get confused about it 😭 )

createthis: It would be slightly annoying to have to set an environment variable when changing models, but it wouldn't be the end of the world. I guess I already have to do that with `LLM_NATIVE_TOOL_CALLING` now.

As a user, I would go the other way and ask that `LLM_NATIVE_TOOL_CALLING` also become a switch in advanced settings. lol.

xingyaoww: Hmm how about we maintain a centralized JSON that includes all these common model name mapping?

Like you could probably put these model name mapping in the consolidated JSON, instead of extra code?:
https://github.com/All-Hands-AI/OpenHands/issues/10262

xingyaoww: > As a user, I would go the other way and ask that LLM_NATIVE_TOOL_CALLING also become a switch in advanced settings. lol.

yeah true 😭 but then as a user i'd probably ask for the same thing in OpenHands CLI...

We really need a big refactor of our configuration system so we can do all of these easily..

createthis: Tell you what... for now let's make it an environment variable. What do you want it called?

xingyaoww: You can just add `disable_stop_word` as a field, and then `LLM_DISABLE_STOP_WORD` should automatically work out of box!

BradKML: Can't wait to see this being added to the next version!

createthis:
Closing in favor of https://github.com/All-Hands-AI/OpenHands/pull/10274

PR #10200, opened 2025-08-11
Workaround: Add image content as user message after tool message

SmartManoj:
Currently images are not allowed in tool Messages

Workaround: 
https://community.openai.com/t/allowing-images-in-non-user-messages/804176/13

image


xingyaoww:
Actually I think anthropic support image tool result. Can you include this change together with the other PR so we can just run eval on one pr

mamoodi:
Based on Xingyao's comment, should this PR be closed?

PR #10201, opened 2025-08-11
feat(ui): redesign terminal

mislavlukach:

Restyled terminal and moved to bottom right corner.

---
Restyled terminal and moved to bottom right corner.
Terminal is no longer a page. Instead, it's just a regular React component.

---


frontend/src/routes/terminal-redirect.tsx
amanape, line 1: What is this route for? Actually we probably shouldn’t make the tab items individual routes and probably address this in the future. So if this has to do with some route logic, probably safe to remove it. LMK what are the consequences of removing this

mislavlukach: Consequence is that we are introducing a breaking change. 
My idea was to keep supporting /terminal as some users still might go to that link. 
If you think it's not needed, we can remove it.

amanape: Yeah let’s remove it

PR #10202, opened 2025-08-11
fix(ui): increase settings page spacing

mislavlukach:

Increase spacing on setting page

---
Increase spacing on setting page

---


frontend/src/routes/git-settings.tsx
amanape, line 1: Maybe we should wrap in a div and add `space-y-4` / `gap-4` instead of repeating margin styles. WDYT?

amanape (approved):

PR #10208, opened 2025-08-11
fix(backend): show name of created branch in conversation list.

hieptl:

If a conversation begins from main but branches off into a new thread (e.g., DAISY-1287) during the discussion, OpenHands should update the conversation metadata so that it references DAISY-1287 instead of main.

---

The PR modifies the code to listen for Git-related events and update the current branch when necessary.

https://github.com/user-attachments/assets/00e9e1bb-55d8-4e06-9581-1e3f83f249de

---

### DIFF
--- a/openhands/server/conversation_manager/standalone_conversation_manager.py
+++ b/openhands/server/conversation_manager/standalone_conversation_manager.py
@@ -2,13 +2,14 @@ import asyncio
 import time
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
-from typing import Callable, Iterable
+from typing import Any, Callable, Iterable
 
 import socketio
 
 from openhands.core.config.openhands_config import OpenHandsConfig
 from openhands.core.exceptions import AgentRuntimeUnavailableError
 from openhands.core.logger import openhands_logger as logger
+from openhands.core.schema.action import ActionType
 from openhands.core.schema.agent import AgentState
 from openhands.events.action import MessageAction
 from openhands.events.stream import EventStreamSubscriber, session_exists
@@ -489,6 +490,14 @@ class StandaloneConversationManager(ConversationManager):
                 conversation.total_tokens = (
                     token_usage.prompt_tokens + token_usage.completion_tokens
                 )
+
+        print(f'event: {event}')
+        print(f'is_git_related_event: {self._is_git_related_event(event)}')
+
+        # Check for branch changes if this is a git-related event
+        if event and self._is_git_related_event(event):
+            await self._update_conversation_branch(conversation)
+
         default_title = get_default_conversation_title(conversation_id)
         if (
             conversation.title == default_title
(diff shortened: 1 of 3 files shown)


openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -521,6 +527,160 @@ async def _update_conversation_for_event(
...
+                    'git merge',
+                    'git rebase',
+                    'git reset',
+                    'git delete',
+                ]
+            )
+
+        return False
+
+    async def _update_conversation_branch(self, conversation: ConversationMetadata):
```
malhotra5, line 610: Once we've detected that the agent has executed a command to switch branches, can we figure out the branch name from the command itself rather than using the runtime?

openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -489,6 +490,11 @@ async def _update_conversation_for_event(
                 conversation.total_tokens = (
                     token_usage.prompt_tokens + token_usage.completion_tokens
                 )
+
+        # Check for branch changes if this is a git-related event
+        if event and self._is_git_related_event(event):
+            await self._update_conversation_branch(conversation)
```
malhotra5: nice! we also have docker nested runtimes; the server is spun inside the docker containers and actions are contained within the container itself

so we'll need a webhook back to the main server to request an update to the branch name; I think tracing the logic for storing accumulated cost in the convo metadata for nested runtimes might be helpful

malhotra5:
This may be out of scope, but important to consider since this feature relies on intercepting agent commands

If the agent decides to clone a second repo and switch branches, for which repo should the current branch be shown in the conversation pane?

openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -521,6 +527,160 @@ async def _update_conversation_for_event(
...
+        """
+        old_branch = conversation.selected_branch
+        conversation.selected_branch = new_branch
+
+        logger.info(
+            f'Branch changed from {old_branch} to {new_branch}',
+            extra={'session_id': conversation.conversation_id},
+        )
+
+    def _log_branch_update_error(self, error: Exception, conversation_id: str):
```
enyst: Can we just log in-place instead of this method? It doesn't seem to do anything but log a warning

openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -521,6 +527,160 @@ async def _update_conversation_for_event(
...
+        Args:
+            selected_repository: Full repository path (e.g., "org/repo")
+
+        Returns:
+            Repository name (e.g., "repo") or None if no repository
+        """
+        if not selected_repository:
+            return None
+        # Extract the repository name from the full path (e.g., "org/repo" -> "repo")
+        return selected_repository.split('/')[-1]
```
enyst: Nit: maybe same here, could do it in-place? Small methods are nice, but 1-3 lines that we always do in-place, idk, seems a bit much decomposition ☺️

openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -489,6 +490,11 @@ async def _update_conversation_for_event(
                 conversation.total_tokens = (
                     token_usage.prompt_tokens + token_usage.completion_tokens
                 )
+
+        # Check for branch changes if this is a git-related event
+        if event and self._is_git_related_event(event):
```
enyst, line 523: (not a review comment)

Just a small note here: this is in a callback from an eventstream subscriber that gets run every event, I _think_ it's really any event, regardless if agent, user, environment, state changes. I always get a little concerned to return as soon as possible, if the event is not applicable.

I think in this case we're good... we identify a Cmd immediately, and git 🤔

Maybe sometime we could consider pre-filtered subscriptions, so that the filtering happens as early in the execution flow as possible.

openhands/server/conversation_manager/standalone_conversation_manager.py
hieptl: Hi @enyst, I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/server/conversation_manager/standalone_conversation_manager.py
hieptl: Hi @enyst, I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/server/conversation_manager/standalone_conversation_manager.py
hieptl: Hi @enyst, you’re right that keeping the callback as minimal as possible is important, especially since it runs on every event. In this case, since we identify the Cmd immediately, the early return should keep the execution lightweight. I agree that pre-filtered subscriptions would be a valuable improvement to reduce overhead even further, and it’s definitely something we can consider exploring in a future iteration. Thank you for highlighting this! 🙏

openhands/server/conversation_manager/standalone_conversation_manager.py
hieptl: Hi @malhotra5, That’s a good point. In practice, the agent may execute compound commands (e.g., cd workspace && git checkout -b ...) or include additional steps around the checkout. Because of this variability, reliably extracting the branch name directly from the raw command string becomes error-prone.

To avoid incorrect assumptions, I felt it was safer to detect when a Git-related command has run and then verify the branch from the runtime state. This way, if the branch hasn’t actually changed, we can safely no-op, but if it has, we’ll capture the update accurately.

openhands/server/conversation_manager/standalone_conversation_manager.py
hieptl: Hi @malhotra5 , I tested OSS with `docker_nested_conversation_manager` and it appears to be working without requiring a webhook.  

**What I tried:**  
1. Updated `server_config.py` to use `docker_nested_conversation_manager`:  
   ```python
   conversation_manager_class: str = os.environ.get(
     'CONVERSATION_MANAGER_CLASS',
     'openhands.server.conversation_manager.docker_nested_conversation_manager.DockerNestedConversationManager',
   )

2. Launched a new conversation via the UI.

3. Asked the agent to create a new branch, switch between branches, and delete the created branch.

From my tests, everything worked as expected. The reason seems to be that `docker_nested_conversation_manager` internally relies on standalone_conversation_manager. Here’s the reference in the code:

https://github.com/All-Hands-AI/OpenHands/blob/36d31b74f7c14a371c89a4d94593a82b9aa11c3b/openhands/server/conversation_manager/docker_nested_conversation_manager.py#L509-L511

I’ve also recorded a quick video to show the output. Thank you for your feedback! 🙏 

https://github.com/user-attachments/assets/9c28b33c-de3e-49d2-98a6-e0af337833a1

openhands/server/conversation_manager/standalone_conversation_manager.py
enyst: It seems to me that Rohit’s question is a good question, and I’m not sure the answer covers it yet. Because it’s the same command string, isn’t it?

The command coming from the agent LLM is captured in the action (which is an event captured by the `_update_conversation` callback), so we should have it here. What else could happen on the way, from here to the runtime?

hieptl: Hi @enyst, yes, we can see exactly which commands the agent executes. However, those commands don’t follow a consistent pattern, which makes it difficult to reliably extract the branch name directly from them.

For example, during testing I noticed that the agent might run:

```js
git branch && git checkout -b test_1234
```

```js
cd workspace && git checkout test_1234
```

```js
cd workspace && git pull && git checkout bugfix/ui-123
```

```js
git branch -D temp && git checkout -b temp
```

And of course, there may be many other unpredictable commands that we can’t anticipate.

Because of this variability, my proposed solution is to detect potentially relevant Git commands from the current execution, and then ask the runtime to check the current branch. If the branch has changed, we update it; otherwise, we don’t take any action.

You can refer to the code snippet below for more details.

```js
if (
    hasattr(event, 'command')
    and hasattr(event, 'action')
    and event.action == ActionType.RUN
):
    command = event.command.lower()
    # Check if any git command that changes branches is present anywhere in the command
    # This handles compound commands like "cd workspace && git checkout feature-branch"
    return any(
        git_cmd in command
        for git_cmd in [
            'git checkout',
            'git switch',
            'git merge',
            'git rebase',
            'git reset',
            'git delete',
        ]
    )
```

If you have another solution in mind, please feel free to share it—I’d be happy to discuss. Thank you!

openhands/server/conversation_manager/standalone_conversation_manager.py
```diff
@@ -548,6 +554,142 @@ async def _update_conversation_for_event(
...
+        try:
+            # Get the session and runtime for this conversation
+            session, runtime = self._get_session_and_runtime(
+                conversation.conversation_id
+            )
+            if not session or not runtime:
+                return
+
+            # Get the current branch from the workspace
+            current_branch = self._get_current_workspace_branch(
```
enyst, line 626: Could we check this against `conversation.selected_branch` instead?

openhands/server/conversation_manager/standalone_conversation_manager.py
enyst: You're right, but I think we're talking a bit past each other. Sorry for the lack of clarity! I see that check for git patterns is in the conversation manager.

Just left a question below on the current branch.

openhands/server/conversation_manager/standalone_conversation_manager.py
hieptl: This function is used to retrieve the current branch from the runtime whenever a Git-related event occurs.

For example, given the following Git event:

```js
git checkout -b test_123
```

The function will return `test_123`. We then compare this new branch value with `conversation.selected_branch` inside the `_update_branch_in_conversation` function.

I hope this clarifies your question. Please let me know if I’ve understood it correctly. Thank you! 🙏

openhands/server/conversation_manager/standalone_conversation_manager.py
hieptl: No worries at all. Thank you for pointing that out! 🙏 If you have any other feedback or questions, please feel free to let me know.

openhands/server/conversation_manager/standalone_conversation_manager.py
enyst: Thank you, I was confused. Actually, I think you’re right. We need to go look for the actual, real branch inside the runtime, after the command was executed…😅

That raises another question though: we intercept here `CmdRunAction`, not the corresponding observation. Do we know for sure that the cmd was already executed? Only after the observation, we may have changed branch in the sandbox.

hieptl: Hi @enyst, excellent observation! I’ll take a look and update the code. Really appreciate your input. 🙏

hieptl: Hi @enyst, I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/server/routes/manage_conversations.py
```diff
@@ -597,28 +597,203 @@ def _get_contextual_events(event_store: EventStore, event_id: int) -> str:
...
+        update_fields.append(f'title: {data.title}')
+    if data.selected_branch is not None:
+        update_fields.append(f'selected_branch: {data.selected_branch}')
+    if data.selected_repository is not None:
+        update_fields.append(f'selected_repository: {data.selected_repository}')
+
+    return ', '.join(update_fields) if update_fields else 'no fields'
+
+
+async def _validate_branch_update(
```
malhotra5: If we're getting the branch information directly from the runtime by running a git command, do we have to validate that the branch exists again? 

Sometimes the agent may make a local branch that hasn't been pushed yet, so would this fail in those scenarios?

Also NIT: lets move the `json response` objects to whether the route is called, instead of having them in the helper function

hieptl: Hi @malhotra5,

I have removed the changes as they are not necessary for this PR.

 Thank you for pointing that out! 🙏

openhands/integrations/provider.py
```diff
@@ -466,6 +466,48 @@ async def get_branches(
 
         return main_branches + other_branches
 
+    async def get_branch(
```
malhotra5: if this is unused then lets remove it (including the methods to get a branch in the service classes); besides   this PR LGTM

malhotra5 (approved):
LGTM; we have a little too much modularity imho which makes it a little tough to read

Thanks for getting through this!!

PR #10213, opened 2025-08-11
feat(ui): replace tab component

mislavlukach:

Replace tab component with icons

---
Replace tab component with icons

Screenshot 2025-08-13 at 14 36 00

---


amanape (commented):
Yeah this seems like a tricky change. Instead of trying to change the existing tab logic, maybe we can cut down and strip some things here and introduce a basic store / context to manage the current view, and just show the page accordingly.

Should also help consolidate these action buttons in a single area that we can extend easily

Would love to hear what you think

amanape:
image

mislavlukach:
> e a tricky change. Instead of trying to change the existing tab logic, maybe we can cut down and strip some things here and introduce a basic store / context to manage the current view, and just show the page accordingly.
> 
> Should also help consolidate these action buttons in a single area that we can extend easily
> 
> Would love to hear what you think

@amanape I would prefer it that way as well. Only downside is that our users lose specific URLs they can bookmark and later visit. Based on the nature of this page I'm not sure this provides any value to them really. So we can ditch that.  Do you agree? 
If yes, let's merge terminal PR first (https://github.com/All-Hands-AI/OpenHands/pull/10201) and I'll change this PR

frontend/src/components/features/conversation/conversation-tabs/conversation-tabs2.tsx
amanape, line 1: Probably better to change the name to `conversation-tabs` 😄

amanape: nvm!

PR #10222, opened 2025-08-11
OverlayFS support for docker runtimes

olyashok:
- [X ] This change is worth documenting at https://docs.all-hands.dev/
- [ X] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

---
Add support for overlay mounts in Docker runtime

This update introduces a new method to process overlay mounts specified in the sandbox configuration. It creates unique upper and work directories for each overlay mount, ensuring proper handling of read-only lower directories with per-container copy-on-write layers. The overlay mounts are then integrated into the container initialization process.

---

### DIFF
--- a/openhands/runtime/impl/docker/docker_runtime.py
+++ b/openhands/runtime/impl/docker/docker_runtime.py
@@ -9,6 +9,7 @@ import docker
 import httpx
 import tenacity
 from docker.models.containers import Container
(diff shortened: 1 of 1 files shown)


rbren (approved):
This LGTM and seems reasonably safe--the new codepath should be off by default

PR #10223, opened 2025-08-11
Fix circular import by moving refine_prompt to dedicated module

tofarr:

This PR fixes a circular import issue that was causing server startup failures. The fix ensures OpenHands starts properly without import errors.

---

This PR fixes a circular import issue by moving the `refine_prompt` function to a dedicated module.

**Problem:**
There was a circular import chain:
- `session.py` imports `agent_session.py`
- `agent_session.py` imports `mcp`
- `mcp` imports `memory.py`
- `memory.py` imports `prompt.py`
- `prompt.py` imports `bash.py`
- `bash.py` imports `codeact_agent.py`
- `codeact_agent.py` imports `conversation_memory.py`
- `conversation_memory.py` imports `prompt.py`

This circular import was causing server startup failures.

**Solution:**
1. Created a new module `openhands/agenthub/codeact_agent/tools/prompt.py` with the `refine_prompt` function
2. Updated `bash.py` to import `refine_prompt` from the new module
3. Updated `utils/prompt.py` to import `refine_prompt` from the new module

This breaks the circular import chain and allows the server to start properly.

**Testing:**

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:753581e-nikolaik   --name openhands-app-753581e   docker.all-hands.dev/all-hands-ai/openhands:753581e
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix-circular-import openhands
```


openhands/utils/prompt.py
```diff
@@ -92,6 +91,8 @@ def _load_template(self, template_name: str) -> Template:
             return Template(file.read())
 
     def get_system_message(self) -> str:
+        from openhands.agenthub.codeact_agent.tools.prompt import refine_prompt
```
enyst, line 94: If this is where it’s used, what if we define it directly here, in this file?

tofarr: Then that causes a different circular import! (As this `refine_prompt` method is also used in `bash.py`)

tofarr:
> Thanks! I wonder how hard is it to reproduce the server crash in CI?

Good point! I observed the error in the unit tests on the saas project - I'll create a follow up PR to add  a corresponding test in the OpenHands repo too.

PR #10235, opened 2025-08-11
Fix for issues where callbacks are not batched

tofarr:

Fixes a configuration issue where webhook batching was not being properly enabled even when configured. This improves performance by reducing the number of webhook calls when file store operations are batched.

---

This PR fixes a bug where the `file_store_web_hook_batch` configuration parameter was not being properly passed through to the `get_file_store()` function in several locations. The issue was that some calls were using positional arguments while others were using named arguments, and the batching parameter was being lost.

**Changes made:**
- Standardized all `get_file_store()` calls to use named parameters consistently
- Updated calls in `openhands/server/shared.py`, `openhands/storage/__init__.py`, and various store classes
- Fixed parameter name mismatch (`batch` vs `file_store_web_hook_batch`) in the `get_file_store()` function signature
- Updated test configuration in `tests/runtime/conftest.py`

This ensures that when `file_store_web_hook_batch=True` is configured, the `BatchedWebHookFileStore` wrapper is properly applied, enabling webhook batching functionality that was introduced in PR #10119.

---

This addresses issues where webhook batching was not working as expected despite being configured, leading to unnecessary individual webhook calls instead of batched operations.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:41b3525-nikolaik   --name openhands-app-41b3525   docker.all-hands.dev/all-hands-ai/openhands:41b3525
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix-callbacks openhands
```

### DIFF
--- a/openhands/server/shared.py
+++ b/openhands/server/shared.py
@@ -27,11 +27,11 @@ assert isinstance(server_config_interface, ServerConfig), (
 )
 server_config: ServerConfig = server_config_interface
 file_store: FileStore = get_file_store(
-    config.file_store,
-    config.file_store_path,
-    config.file_store_web_hook_url,
-    config.file_store_web_hook_headers,
-    batch=config.file_store_web_hook_batch,
+    file_store_type=config.file_store,
+    file_store_path=config.file_store_path,
+    file_store_web_hook_url=config.file_store_web_hook_url,
+    file_store_web_hook_headers=config.file_store_web_hook_headers,
+    file_store_web_hook_batch=config.file_store_web_hook_batch,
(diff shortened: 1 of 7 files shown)


malhotra5 (approved):
LGTM!

PR #10129, opened 2025-08-06
Implement auto-pagination for conversation list with infinite scroll

xingyaoww:
---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:a85fa13-nikolaik   --name openhands-app-a85fa13   docker.all-hands.dev/all-hands-ai/openhands:a85fa13
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@auto-pagination-conversations openhands
```

PR #10236, opened 2025-08-11
Fix auto title generation broken by pagination race condition

xingyaoww:

Fixes an issue where conversation titles were showing generic names like "Conversation 13300" instead of meaningful auto-generated titles. This was caused by a race condition introduced when the conversation list was changed to use pagination - conversations were being displayed before their titles could be generated. Now conversations will show proper descriptive titles that are automatically generated based on their content.

---

This PR fixes the auto title generation that was broken by the pagination changes in PR #10129. The issue was a timing problem where conversations were loaded and displayed before title generation could complete.

**Root Cause:**
- Auto-pagination changed conversation loading from simple array to paginated infinite scroll
- Title generation only triggers during conversation events, not when loading conversation lists
- This created a race condition where conversations display before auto title generation completes

**Solution:**
1. **Frontend WebSocket Listeners** (`conversation-panel.tsx`):
   - Added WebSocket event listeners for title update events
   - When title updates are received, query cache is invalidated to trigger UI refresh
   - Provides real-time title updates without page refresh

2. **Backend Title Generation Trigger** (`manage_conversations.py`):
   - Added `_trigger_title_generation_if_needed()` function that checks for default titles
   - Triggers existing `_update_conversation_for_event()` mechanism for conversations with default titles
   - Integrated into both `search_conversations` and `get_conversation` APIs
   - Uses fire-and-forget async tasks to avoid blocking API responses

**Design Decisions:**
- Leverages existing title generation infrastructure instead of duplicating logic
- Uses WebSocket events for real-time updates (consistent with existing architecture)
- Non-blocking approach ensures API performance isn't impacted
- Maintains backward compatibility with existing functionality

**How it works:**
1. User loads conversation list → API returns conversations (some with default titles)
2. Backend triggers title generation for conversations with default titles (async)
3. Title generation completes → WebSocket event is emitted
4. Frontend receives WebSocket event → Query cache is invalidated → UI refreshes with new titles

---

This addresses the issue reported where conversations were showing generic titles like "Conversation 13300" instead of meaningful auto-generated titles, which was introduced by the auto-pagination changes in commit 56f752557.

### DIFF
--- a/frontend/src/components/features/conversation-panel/conversation-panel.tsx
+++ b/frontend/src/components/features/conversation-panel/conversation-panel.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { NavLink, useParams, useNavigate } from "react-router";
 import { useTranslation } from "react-i18next";
+import { useQueryClient } from "@tanstack/react-query";
 import { I18nKey } from "#/i18n/declaration";
 import { ConversationCard } from "./conversation-card";
 import { usePaginatedConversations } from "#/hooks/query/use-paginated-conversations";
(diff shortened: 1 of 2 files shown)


xingyaoww:
Actually, i tested in main and this was working. So it is likely something have broken in sass but not oss

PR #10259, opened 2025-08-12
[Bug]: Ignore comments when checking for multiple bash commands

### LINKED ISSUE #10243: [Bug]: when executing bash commands, exclude comments from triggering the multi-command error

I'm currently using Qwen3-Coder-30B via LM Studio as my local LLM and the latest OpenHands docker image (MacOS). I'm finding that the LLM is frequently executing bash commands where it includes a comment as the first line, and the command itself as the second line.

For example:
```bash
(1) # Let me just check the current git status and push directly
(2) git status --porcelain
```

Another example:
```bash
(1) # Let me just make sure the current branch is properly updated
(2) git add test_hello.sh && git commit -m "Fix test_hello.sh to properly implement dynamic coverage reporting" --author="openhands [openhands@all-hands.dev](mailto:openhands@all-hands.dev)" && git push origin add-test-coverage
```

These are triggering an OpenHands error (I think via [this code here](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/runtime/utils/bash.py#L502-L511) in openhands/runtime/utils
/bash.py): "ERROR: Cannot execute multiple commands at once.
Please run each command separately OR chain them into a single command via && or ;"

It would be great if split_bash_commands() in split.py removed comment lines before counting the number of commands being executed - that way a simple two-line command where one line is a comment wouldn't keep triggering this error. Or perhaps there is another solution.

### OpenHands Installation

Docker command in README

### OpenHands Version

0.51

### Model Name

Qwen3-Coder-30B

### Operating System

MacOS

Examples provided above

rbren:
## Description

This PR addresses issue #10243 where bash commands with comments are triggering the multi-command error.

Currently, when a user (or LLM) includes a comment before a bash command, the system treats it as multiple commands and throws an error. This PR adds failing tests that demonstrate the issue.

For example:
```bash
# Let me just check the current git status and push directly
git status --porcelain
```

This is currently being treated as two separate commands, causing an error.

## Changes

- Added failing tests that demonstrate the current behavior
- Tests will pass once the fix is implemented

## Next Steps

The actual fix will involve modifying the `split_bash_commands` function in `openhands/runtime/utils/bash.py` to ignore comment-only lines when determining if there are multiple commands.

## Related Issues

Fixes #10243

### DIFF
--- a/tests/unit/test_bash_comments.py
+++ b/tests/unit/test_bash_comments.py
@@ -1,0 +1,44 @@
+from openhands.runtime.utils.bash import split_bash_commands
+
+
+def test_comment_followed_by_command():
+    """Test that a comment followed by a command is correctly handled as multiple commands."""
+    input_command = """# Let me just check the current git status and push directly
+git status --porcelain"""
+
+    # Current behavior - this will return two commands
+    result = split_bash_commands(input_command)
+
+    # This test should fail with the current implementation
+    # but will pass after our fix
+    assert len(result) == 1, f'Expected 1 command, got {len(result)}: {result}'
+    assert 'git status --porcelain' in result[0]
+
+
+def test_multiple_comments_followed_by_command():
+    """Test that multiple comments followed by a command are correctly handled as a single command."""
+    input_command = """# First comment
+# Second comment
+# Third comment
+git status"""
+
+    # Current behavior - this will return multiple commands
+    result = split_bash_commands(input_command)
+
+    # This test should fail with the current implementation
+    # but will pass after our fix
+    assert len(result) == 1, f'Expected 1 command, got {len(result)}: {result}'
+    assert 'git status' in result[0]
+
+
+def test_comment_only():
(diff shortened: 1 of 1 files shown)


openhands/runtime/utils/bash.py
```diff
@@ -501,7 +536,11 @@ def execute(self, action: CmdRunAction) -> CmdOutputObservation | ErrorObservati
 
         # Check if the command is a single command or multiple commands
         splited_commands = split_bash_commands(command)
-        if len(splited_commands) > 1:
+        
+        # Filter out comment-only commands when checking for multiple commands
+        non_comment_commands = [cmd for cmd in splited_commands if not _is_comment_only(cmd)]
+
```
rbren: @openhands change the implementation here. Make it so that `split_bash_commands` only returns non-comment commands, that way we don't have to filter everywhere it gets called

PR #10269, opened 2025-08-12
Modify experiment manager defaults for nested runtimes

malhotra5:

---

The nested runtime doesn't pick up on the changes made by the experiment manager, so modify the no-ops in the experiment manager via envs for nested runtimes

---

### DIFF
--- a/openhands/experiments/experiment_manager.py
+++ b/openhands/experiments/experiment_manager.py
@@ -1,3 +1,5 @@
+import copy
+import json
 import os
 
 from openhands.core.config.openhands_config import OpenHandsConfig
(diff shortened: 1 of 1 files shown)


openhands/server/routes/manage_conversations.py
```diff
@@ -707,3 +709,28 @@ async def update_conversation(
...
+
+    # Don't modify again if it already exists
+    if exists:
+        return False
+
+    try:
+        file_store.write(exp_config_filepath, exp_config.model_dump_json())
+    except Exception as e:
+        logger.info(f'Failed to write experiment config for {conversation_id}: {e}')
+        return True
```
raymyers, line 734: What does return value indicate that this would be a True?

openhands/storage/locations.py
```diff
@@ -28,3 +28,7 @@ def get_conversation_init_data_filename(sid: str, user_id: str | None = None) ->
 
 def get_conversation_agent_state_filename(sid: str, user_id: str | None = None) -> str:
     return f'{get_conversation_dir(sid, user_id)}agent_state.pkl'
+
+
+def get_experiment_config_filename(sid: str, user_id: str | None = None) -> str:
+    return f'{get_conversation_dir(sid, user_id)}exp_config.json'
```
raymyers, line 34: Seems like this should use `os.path.join()` not string concat, but that's not an issue introduced by this PR.

PR #10270, opened 2025-08-12
Added INSTRUCTION_TEMPLATE_NAME to run_infer.py in swe_bench

KevinMusgrave:

Allows the prompt template filename to be specified with an environment variable, when running SWE Bench evaluations.

This makes it easier to experiment with prompts, without having to change the if/else logic in run_infer.py.


evaluation/benchmarks/swe_bench/run_infer.py
```diff
@@ -108,7 +109,9 @@ def get_instruction(instance: pd.Series, metadata: EvalMetadata) -> MessageActio
     llm_model = metadata.llm_config.model
 
     # Determine the template file based on mode and LLM
-    if mode.startswith('swt'):
+    if INSTRUCTION_TEMPLATE_NAME:
+        template_name = INSTRUCTION_TEMPLATE_NAME
```
xingyaoww: actually another quetsion: you might want to log this to eval metadata for tracking purpose

KevinMusgrave: Makes sense. I've put it in EvalMetadata now.

xingyaoww (approved):
LGTM

PR #10292, opened 2025-08-13
Fix for issue where static system commands are truncated

tofarr:

---
This does raise the question of whether to rename this variable.

---

### DIFF
--- a/openhands/events/observation/commands.py
+++ b/openhands/events/observation/commands.py
@@ -115,9 +115,11 @@ class CmdOutputObservation(Observation):
         **kwargs: Any,
     ) -> None:
         # Truncate content before passing it to parent
-        truncated_content = self._maybe_truncate(content)
+        truncate = not kwargs.get('is_static')
+        if truncate:
+            content = self._maybe_truncate(content)
 
-        super().__init__(truncated_content)
+        super().__init__(content)
 
         self.command = command
         self.observation = observation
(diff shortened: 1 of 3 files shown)


enyst (commented):
I wonder if `hidden` flag will do.

Could you please explain a bit here, who makes the change, the user? What exactly is the CmdRunAction, is it initiated by someone or something else than the agent?

tofarr:
> I wonder if `hidden` flag will do.
> 
> Could you please explain a bit here, who makes the change, the user? What exactly is the CmdRunAction, is it initiated by someone or something else than the agent?
> 
> _Edited to add:_ `hidden` is hidden from the agent, that is, never gets to the agent LLM. It used to be used for user terminal commands when they were possible, for example. 🤔

In this case, the browser creates the request to get a git diff when the user clicks - and the app server passes the request to the action exection server. The action does not go through the LLM, nor is it added to the event stream. Does this match the intent of the `hidden` flag? (If so I'll add a note to describe `hidden` as part of this PR - I saw it but didn't know whether to use it or `static`)

enyst:
Yes! That should be fine.

The only thing `hidden` changes, is that the agent controller avoids to send it to the agent LLM, if it finds such events in the stream. In this case, it even isn't in the stream, so it can be `True` just fine, IMHO

enyst (approved):

enyst:
Actually conceptually it isn't quite there, because it isn't in the stream. 🤔 On the plus side, really `hidden` should say it's not agent's business, which is true. So I'd lean to `hidden`. 🤷🏻‍♂️

PR #9942, opened 2025-07-28
[Feature Request]: Make git username and email configurable

neubig:
## Summary

Implements configurable git username and email settings for OpenHands, addressing issue #8453. Users can now customize git credentials instead of using hardcoded "openhands" values.

## Key Changes

- **Configuration**: Added `git_user_name` and `git_user_email` fields to `OpenHandsConfig`
- **Environment Variables**: Support for `GIT_USER_NAME` and `GIT_USER_EMAIL`
- **CLI Arguments**: Added `--git-user-name` and `--git-user-email` options
- **Runtime Integration**: Updated action execution server to use configurable credentials
- **Resolver Integration**: Modified `make_commit()` to always use provided credentials
- **Backward Compatible**: Defaults to "openhands" and "openhands@all-hands.dev"

## Usage Examples

```bash
# Environment variables
export GIT_USER_NAME="Your Name"
export GIT_USER_EMAIL="your.email@example.com"

# Command line
python -m openhands.resolver.send_pull_request \
  --git-user-name "Your Name" \
  --git-user-email "your.email@example.com" \
  --selected-repo owner/repo --token $TOKEN
```

## Testing

✅ **End-to-End Verification**: Successfully tested with [neubig/test-repo PR #12](https://github.com/neubig/test-repo/pull/12)
- Commit author correctly set to "Graham Neubig <neubig@gmail.com>"
- Custom git credentials properly applied in real repository
- All functionality working as expected

✅ **Unit Tests**: Comprehensive test coverage for all configuration scenarios

✅ **Integration Tests**: Verified CLI arguments and environment variable loading

Fixes #8453

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:b6db577-nikolaik   --name openhands-app-b6db577   docker.all-hands.dev/all-hands-ai/openhands:b6db577
```

PR #10293, opened 2025-08-13
refactor(git): principled way to set git configuration for agents & re-enable git settings in UI

xingyaoww:

refactor(git): principled way to set git configuration for agents;

Fixes a bug where git operations (like commits and pushes) would fail in warm runtimes due to missing git user configuration. This ensures that git commands work reliably in all runtime scenarios, including reused warm runtimes.

---

This PR fixes a critical bug in the warm runtime system where git configuration was being lost when warm runtimes were claimed and reused.

**Problem**: Git configuration was set during action execution server initialization, but warm runtimes bypass this initialization step. When a warm runtime was claimed, the git user settings were missing, causing git operations to fail.

**Solution**: Move git configuration from the action execution server to the runtime client, ensuring git settings are applied after every runtime connection.

**Key Changes**:
1. **Removed git config from ActionExecutor**: Eliminated git parameters and configuration commands from the action execution server
2. **Added git config to Runtime base class**: New `setup_git_config()` method handles platform-specific git configuration
3. **Updated all runtime implementations**: Local, Remote, Kubernetes, CLI, and Docker runtimes now call git config setup after connection

**Technical Details**:
- Maintains platform-specific logic (Windows PowerShell vs Unix bash, local file-based vs global config)
- Uses `CmdRunAction` to execute git config commands in the runtime environment
- Includes proper error handling and logging for failed git commands
- Backward compatible with no changes to external APIs

**Workflow Fix**:
- **Before**: Git config set during server startup → lost in warm runtimes → git operations fail
- **After**: Git config applied after every runtime connection → works for both fresh and warm runtimes

---
Screenshots:

Before setting anything in UI:
image

After setting (this pr also adjust the fontsize so it follow existing setting convention in LLM tab:
image
It is working
image

In CLI - does override user config:
image

---

This addresses the warm runtime git configuration bug reported in the deploy repository where git operations would fail when using warm runtimes due to missing user configuration. This is why https://github.com/All-Hands-AI/OpenHands/pull/9942 didn't work

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/8db206e286254400854137ff553bbacd)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:a65fb44-nikolaik   --name openhands-app-a65fb44   docker.all-hands.dev/all-hands-ai/openhands:a65fb44
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix-git-config-warm-runtime openhands
```


openhands/runtime/action_execution_server.py
```diff
@@ -349,35 +345,7 @@ async def _init_bash_commands(self):
...
-                INIT_COMMANDS.append('export GIT_CONFIG=$(pwd)/.git_config')
-        else:
-            # Non-local (implies Linux/macOS)
-            INIT_COMMANDS.append(
-                f'git config --global user.name "{self.git_user_name}"'
-            )
-            INIT_COMMANDS.append(
-                f'git config --global user.email "{self.git_user_email}"'
-            )
+        # Git configuration is now handled by the runtime client after connection
```
xingyaoww:
```suggestion
```

openhands/cli/main.py
```diff
@@ -724,6 +724,9 @@ def run_cli_command(args):
         print_formatted_text(f'Connection refused: {e}')
         sys.exit(1)
     except Exception as e:
+        import traceback
+
+        traceback.print_exc()
```
enyst: Maybe we can now remove this?

openhands/runtime/base.py
```diff
@@ -907,6 +910,42 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
...
+
+        This method is called automatically during setup_initial_env() to ensure
+        git configuration is applied to the runtime environment.
+        """
+        # Get git configuration from config
+        git_user_name = self.config.git_user_name
+        git_user_email = self.config.git_user_email
+
+        # Skip git configuration for CLI runtime to preserve user's local git settings
+        is_cli_runtime = self.config.runtime == 'cli'
```
enyst, line 924: Should this special case include ‘local’ too? Or the local nested needs it set… hmm, globally?

enyst (commented):
I liked the other option too, sending git user/email in the RecallObs, but this is okay too.

It does seem to mean that if users use ‘local’ runtime, their global git config will be overwritten, though?

openhands/cli/main.py
xingyaoww: actually i run into an error while debugging the CLI, and the print below doesn't really include any info. Maybe i can just remove the `except Exception` as a whole

openhands/runtime/base.py
xingyaoww: yeah - basically local nested runtime needs it set :(

xingyaoww:
> It does seem to mean that if users use ‘local’ runtime, their global git config will be overwritten, though?

@enyst yes.. but now with CLI, i wonder if there's an actual use case for the standalone usage of local runtime though?

enyst:
> > It does seem to mean that if users use ‘local’ runtime, their global git config will be overwritten, though?
> 
> @enyst yes.. but now with CLI, i wonder if there's an actual use case for the standalone usage of local runtime though?

That's a good point, except for Windows... I don't know if CLI works well on Windows
Cc: @li-boxuan

xingyaoww:
@enyst happy to fix it follow-up PR if i eventually break it on windows.. But on the other hand, if CLI breaks on Windows, we should just fix the CLI, right 😄

li-boxuan:
> > It does seem to mean that if users use ‘local’ runtime, their global git config will be overwritten, though?
> 
> @enyst yes.. but now with CLI, i wonder if there's an actual use case for the standalone usage of local runtime though?

Yes, headless mode with local runtime is how OpenHands works in terminal-bench harness. That being said, if openhands-cli can be made fully autonomous (non-interactive), I guess headless mode can be deprecated.

CLI worked on windows when I used it last time so I don’t think windows will be a concern.

PR #10306, opened 2025-08-14
feat(agent): include a new roleplay-based prompt 

xingyaoww:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

Saw one funny tweet (https://x.com/wquguru/status/1955790997029577064, in chinese) that prompts claude to act as "Linus" and the result seems good -- we can probably try it!

---
**Link of any specific issues this addresses:**

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:98fb828-nikolaik   --name openhands-app-98fb828   docker.all-hands.dev/all-hands-ai/openhands:98fb828
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@xw/linus-prompt openhands
```

PR #10312, opened 2025-08-14
[CLI] Fix MCP toml formatting issue

### LINKED ISSUE #10173: [Bug]: MCP config format inconsistency in CLI

In the documentation and the `MCPConfig` class, we support loading the `[mcp]` section with content like below:

```toml
[mcp]
stdio_servers = [
  {
    name="Context7",
    command="npx",
    args=["-y", "@upstash/context7-mcp"],
  }
]
```

This is the required format, but currently in the CLI, we have 2 other methods which load and save the MCP settings to toml file in a format different from the expected one. Below is the config saved by the CLI:
```toml
[mcp]
[[mcp.stdio_servers]]
name = "Context7"
command = "npx"
args = [ "-y", "@upstash/context7-mcp",]
```

---
Root cause: The issue was in the `save_config_file` function in `openhands/cli/commands.py`. When the TOML library (`toml.dump()`) encounters arrays of dictionaries under a section, it automatically converts them to array-of-tables format instead of inline arrays.

### OpenHands Installation

CLI

### OpenHands Version

main

### Model Name

_No response_

### Operating System

None

_No response_

ryanhoangt:

---

The regular `toml.dump()` automatically converts arrays of dictionaries to array-of-tables format (`[[mcp.stdio_servers]]`) instead of the required inline array format (`stdio_servers = [{...}]`). This PR uses `tomlkit` with inline tables for proper formatting of MCP config arrays.

After the fix I tested and the addition/removal seems to work as expected:

Screenshot 2025-08-14 at 16 52 19

---

Fix #10173

### DIFF
--- a/openhands/cli/commands.py
+++ b/openhands/cli/commands.py
@@ -4,7 +4,6 @@ import sys
 from pathlib import Path
 from typing import Any
 
-import toml
 import tomlkit
 from prompt_toolkit import HTML, print_formatted_text
 from prompt_toolkit.patch_stdout import patch_stdout
@@ -524,12 +523,7 @@ def load_config_file(file_path: Path) -> dict:
             with open(file_path, 'r') as f:
                 return dict(tomlkit.load(f))
         except Exception:
-            try:
-                # Fallback to regular toml for backward compatibility
-                with open(file_path, 'r') as f:
-                    return toml.load(f)
-            except Exception:
-                pass
+            pass
 
     # Create directory if it doesn't exist
     file_path.parent.mkdir(parents=True, exist_ok=True)
@@ -538,31 +532,6 @@ def load_config_file(file_path: Path) -> dict:
 
 def save_config_file(config_data: dict, file_path: Path) -> None:
     """Save the config file with proper MCP formatting."""
-    # Check if we have MCP configuration that needs special handling
-    if 'mcp' in config_data and _has_mcp_server_arrays(config_data['mcp']):
-        # Use tomlkit for proper inline array formatting
-        doc = _convert_to_tomlkit_document(config_data)
-        with open(file_path, 'w') as f:
-            f.write(tomlkit.dumps(doc))
-    else:
-        # Use regular toml for other configurations
-        with open(file_path, 'w') as f:
-            toml.dump(config_data, f)
-
-
-def _has_mcp_server_arrays(mcp_config: dict) -> bool:
-    """Check if MCP config contains server arrays that need special formatting."""
-    server_types = ['stdio_servers', 'sse_servers', 'shttp_servers']
-    return any(
-        server_type in mcp_config
-        and isinstance(mcp_config[server_type], list)
-        and len(mcp_config[server_type]) > 0
-        for server_type in server_types
-    )
-
-
-def _convert_to_tomlkit_document(config_data: dict) -> tomlkit.document:
-    """Convert config data to tomlkit document with proper MCP formatting."""
     doc = tomlkit.document()
 
     for key, value in config_data.items():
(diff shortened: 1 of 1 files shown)


openhands/cli/commands.py
```diff
@@ -519,20 +520,85 @@ def load_config_file(file_path: Path) -> dict:
...
+                else:
+                    # Handle other MCP config values normally
+                    mcp_section[mcp_key] = mcp_value
+
+            doc[key] = mcp_section
+        else:
+            # Handle non-MCP sections normally
+            doc[key] = value
+
+    return doc
```
xingyaoww: This is a lot of code for a small fix.. Can we simplify? Is there a direct API we can use to do this?

You can probably try this prompt: https://github.com/All-Hands-AI/OpenHands/pull/10306

xingyaoww:
**openhands** reviewing as Linus Torvalds:

## [Core Judgment]
❌ Not worth doing in this form: This is solving a real problem with absolutely terrible taste.

## [Key Insights]
- **Data Structure**: The core issue is TOML serialization format mismatch - `toml.dump()` has opinions that conflict with your requirements
- **Complexity**: You've written 66 lines of code with 3 helper functions, nested loops, and multiple special cases to solve what should be a 2-line fix
- **Risk**: No breaking changes, but you're introducing maintenance hell

## [Critical Issue]
🔴 **Garbage taste**

This code violates my first principle: "Good taste eliminates special cases." You've created a special case factory:

1. Special MCP vs non-MCP handling
2. Special server type enumeration
3. Special dict vs non-dict value handling  
4. Special tomlkit vs toml fallback logic

All of this complexity for a simple serialization problem.

## [The Right Way]

This is a classic case where you're fighting the library instead of using the right tool. If `toml.dump()` doesn't give you the format you want, don't write 66 lines of workaround code. Either:

1. **Use tomlkit everywhere** - Why are you mixing two TOML libraries? Pick one and stick with it.
2. **Fix the data structure** - Maybe the problem isn't the serializer, maybe it's how you're structuring the data.
3. **Use a different approach** - Template-based generation, custom serializer, whatever.

But don't create this maintenance nightmare of special cases and conditional logic.

## [What I'd Accept]

Show me a 5-line diff that either:
- Switches entirely to tomlkit (removing toml dependency)
- Or finds a way to make toml.dump() behave correctly
- Or restructures the data to avoid the problem entirely

The current approach is the kind of "enterprise software" complexity that makes me want to throw my laptop out the window.

**Verdict**: Rewrite this properly or don't bother.

xingyaoww:
**Review by openhands**

I've reviewed this PR and here's my analysis:

## ✅ **Positives**
- **Correctly addresses the core issue** using `tomlkit` for proper inline table formatting
- **Maintains backward compatibility** with fallback to regular `toml` library  
- **Comprehensive coverage** of all MCP server types (stdio_servers, sse_servers, shttp_servers)
- **Handles edge cases** like non-dict values in server arrays
- **Good separation of concerns** with well-named helper functions
- **Author tested the fix** and confirmed it works as expected

## ⚠️ **Concerns**
- **High complexity for the problem size**: Adds ~66 lines of code for what appears to be a formatting issue
- **Maintenance burden**: Introduces complex logic that needs ongoing maintenance  
- **Mixed library approach**: Uses both `toml` and `tomlkit` which could be confusing

## 💡 **Alternative Approaches to Consider**
1. **Use tomlkit globally** instead of mixing libraries
2. **String post-processing** approach to convert array-of-tables back to inline format
3. **Research toml library options** for inline array formatting

## **Recommendation**
While the PR is technically correct and solves the problem, I agree with @xingyaoww that the solution may be over-engineered. Consider exploring simpler alternatives that achieve the same goal with less complexity.

The PR demonstrates good engineering practices but could benefit from a more streamlined approach.

openhands/cli/commands.py
```diff
@@ -519,20 +520,85 @@ def load_config_file(file_path: Path) -> dict:
...
         try:
+            # First try with tomlkit (handles inline tables)
             with open(file_path, 'r') as f:
-                return toml.load(f)
+                return dict(tomlkit.load(f))
         except Exception:
-            pass
+            try:
+                # Fallback to regular toml for backward compatibility
+                with open(file_path, 'r') as f:
```
enyst: I tend to agree that the fallback doesn't seem necessary. We will need tomlkit I think, as soon as we're ready to support comments... and we may be better off to just use it

openhands/cli/commands.py
```diff
@@ -519,20 +520,85 @@ def load_config_file(file_path: Path) -> dict:
...
+    """Save the config file with proper MCP formatting."""
+    # Check if we have MCP configuration that needs special handling
+    if 'mcp' in config_data and _has_mcp_server_arrays(config_data['mcp']):
+        # Use tomlkit for proper inline array formatting
+        doc = _convert_to_tomlkit_document(config_data)
+        with open(file_path, 'w') as f:
+            f.write(tomlkit.dumps(doc))
+    else:
+        # Use regular toml for other configurations
+        with open(file_path, 'w') as f:
```
enyst: Ditto, do we need this, if we need tomlkit anyway?

openhands/cli/commands.py
ryanhoangt: > I tend to agree that the fallback doesn't seem necessary. 

Yeah, I removed it in the latest commt.

openhands/cli/commands.py
ryanhoangt: This is not needed, I removed it as well.

openhands/cli/commands.py
```diff
@@ -519,8 +519,9 @@ def load_config_file(file_path: Path) -> dict:
     """Load the config file, creating it if it doesn't exist."""
     if file_path.exists():
         try:
+            # First try with tomlkit (handles inline tables)
```
ryanhoangt:
```suggestion
```

enyst (approved):

PR #10320, opened 2025-08-14
feat(frontend): make the action panel take 50% width.

hieptl:

- We need to make the action panel take 50% width.

---

The PR makes the action panel take 50% width.

We can refer to the video below for more information.

all-3116

---

### DIFF
--- a/frontend/src/components/features/conversation/conversation-main.tsx
+++ b/frontend/src/components/features/conversation/conversation-main.tsx
@@ -80,7 +80,7 @@ export function ConversationMain() {
       <ResizablePanel
         orientation={Orientation.HORIZONTAL}
         className="grow h-full min-h-0 min-w-0"
-        initialSize={500}
+        initialSize={(width - 101) * 0.5}
         firstClassName="overflow-hidden bg-base"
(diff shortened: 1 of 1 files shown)


frontend/src/components/features/conversation/conversation-main.tsx
```diff
@@ -80,7 +80,7 @@ export function ConversationMain() {
       <ResizablePanel
         orientation={Orientation.HORIZONTAL}
         className="grow h-full min-h-0 min-w-0"
-        initialSize={500}
```
amanape, line 83: Why `width - 101`? What does the `101` refer to here? Maybe we can remove the resizable container in this PR and rely on straightforward styles e.g., with flex?

We already have an open issue to use something like `react-resizable` for the container in another PR

hieptl: We need to subtract the size bar width and the gap between the two sections.

> We already have an open issue to use something like react-resizable for the container in another PR.

Yes, Mislav is working on that. I can either close this PR and add this issue as an acceptance criterion to that ticket, or merge this PR. Which option do you prefer?

PR #10322, opened 2025-08-14
refactor(frontend): update the styling for suggested tasks.

hieptl:

**Acceptance criteria:**
- No separators for tasks under the same repository
- Heading (“Suggested Tasks” text) have different font weight and colors than the Figma
- Toggling between “View More” and “View Less” does not increase or decrease the container height.

---

The PR updates the styling for suggested tasks.

We can refer to the video below for more information.

https://github.com/user-attachments/assets/49f2b155-c913-47d6-8934-d1c100de6e11

---

### DIFF
--- a/frontend/src/components/features/home/tasks/task-suggestions.tsx
+++ b/frontend/src/components/features/home/tasks/task-suggestions.tsx
@@ -17,7 +17,7 @@ export function TaskSuggestions({ filterFor }: TaskSuggestionsProps) {
   const [isExpanded, setIsExpanded] = useState(false);
   const { data: tasks, isLoading } = useSuggestedTasks();
 
-  const suggestedTasks = filterFor
+  let suggestedTasks = filterFor
     ? tasks?.filter(
         (element) =>
           element.title === filterFor.full_name &&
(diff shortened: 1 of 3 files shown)


hieptl:
> Instead, let’s make it the same collapsed height as the recent conversations, and expand to the same height as recent conversations would. This is better for consistency

Hi @amanape, I’m not sure I fully understand this feedback. In some cases, there may be no recent conversations but there are suggested tasks, and vice versa. Also, each container has a different height. Could you clarify what you mean? Thank you! 🙏

amanape:
Let’s show 3 items of the suggested tasks by default (like the recent conversations), then when the user clicks “View more” they will see and scroll through the rest, just like recent conversations. That way both comonents will have the same starting default items and height

hieptl:
Hi @amanape, I've updated the code based on your feedback! Thank you for your feedback! 🙏

frontend/src/components/features/home/tasks/task-suggestions.tsx
amanape, line 1: You can move these helper functions to a separate file to do a little clean up

frontend/src/utils/utils.ts
```diff
@@ -1,6 +1,7 @@
 import { type ClassValue, clsx } from "clsx";
 import { twMerge } from "tailwind-merge";
 import { Provider } from "#/types/settings";
+import { SuggestedTaskGroup } from "#/components/features/home/tasks/task.types";
```
amanape: To avoid a cyclic dependency (Component imports Utils imports Component), it is a good idea to move this type into a shared file as well, and have both the Utils and Components import the type

amanape (approved):
Thanks!

PR #10328, opened 2025-08-14
feat(ui): run web server

mislavlukach:

Implement run webserver button
Screenshot 2025-08-14 at 19 37 33
Screenshot 2025-08-14 at 19 37 21

---
Implement run webserver button

---


mislavlukach:
@amanape tests are failing, but that is unrelated to this PR. 
Issue is present in the revise-ux branch as well.

frontend/src/components/features/chat/chat-actions.tsx
```diff
@@ -43,6 +55,21 @@ export function ChatActions() {
...
       </ChatActionTooltip>
+      <button
+        type="button"
+        onClick={onRunClick}
+        data-testid="run-button"
+        className={cn(
+          "bg-white py-0.75 pl-2 pr-3.5 rounded-full cursor-pointer",
+          `flex flex-row items-center`,
+          "text-black",
+        )}
```
hieptl, line 65: **suggestion:** We don't need to use `cn` in this case.  
You can directly write the classes like this instead:

```js
className="bg-white py-0.75 pl-2 pr-3.5 rounded-full cursor-pointer flex flex-row items-center text-black"

File: frontend/src/components/features/chat/chat-actions.tsx [hunk 2]
```diff
@@ -43,6 +55,21 @@ export function ChatActions() {
           />
         </button>
       </ChatActionTooltip>
+      <button
+        type="button"
+        onClick={onRunClick}
+        data-testid="run-button"
```
hieptl: **Suggestion:** I noticed that we have added a `data-testid` attribute here, but no corresponding tests are included in this PR.

frontend/src/i18n/translation.json
```diff
@@ -8175,6 +8175,22 @@
         "tr": "Konuşma yok",
         "uk": "Розмов не знайдено"
     },
+    "CONVERSATION$RUN": {
```
hieptl, line 8178: **suggestion:** I think we should change this to `"COMMON$RUN"` because the word `"Run"` may appear in many places in the future.  It feels like a common term.

frontend/src/utils/suggestions/repo-suggestions.ts
```diff
@@ -47,3 +47,5 @@ export const REPO_SUGGESTIONS: Record<string, string> = {
   [KEY_5]: VALUE_5,
   [KEY_6]: VALUE_6,
 };
+export const RUN_SERVER_SUGGESTION =
```
hieptl, line 50: **suggestion:** I think we should add a line break between lines 49 and 50 for better readability.

hieptl:
**Issue:** While testing the PR locally, I noticed that the View button and the hover states from the Figma design are missing.

image

frontend/src/components/features/chat/chat-actions.tsx
mislavlukach: It's intentional from my side, I use it always and separate styles by what they are referring to:

- layout
- internal paddings and background color
- etc.

frontend/src/components/features/chat/chat-actions.tsx
mislavlukach: We don't have a corresponding test file for this component file(chat-actions.test). 
So I thought it's outside of scope for this particular PR

frontend/src/i18n/translation.json
mislavlukach: I agree, will do.

mislavlukach:
> **Issue:** While testing the PR locally, I noticed that the View button and the hover states from the Figma design are missing.
> 
> image

Thanks, I couldn't find that in Figma so I implemented it without a hover state. Will fix it now.

frontend/src/components/features/chat/chat-actions.tsx
amanape: Yes I prefer separating styles in a similar way as well as it makes it easy to read.

frontend/src/components/features/chat/chat-actions.tsx
amanape:
```suggestion
```

amanape:
Just noticed the comment on hover state. Sorry I merged too early, feel free to open a separate PR

PR #10339, opened 2025-08-14
fix - Thread-safety in BatchedWebHookFileStore

raymyers:

---
We believe we've found a race condition in BatchedWebHookFileStore, reproduced by tests in this draft PR.

https://github.com/All-Hands-AI/OpenHands/pull/10335

This PR updates it to fix those, passing the batch as an argument to the executor which makes the state easier to reason about. It continues to pass the existing BatchedWebHookFileStore tests. We may be able to add tests that cover the race condition but the initial try were too slow.

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:7ac5d66-nikolaik   --name openhands-app-7ac5d66   docker.all-hands.dev/all-hands-ai/openhands:7ac5d66
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@ray/batched-race openhands
```


openhands/storage/batched_web_hook.py
```diff
@@ -169,38 +168,42 @@ def _queue_update(
...
+            EXECUTOR.submit(self._send_batch, batch_to_send)
+        else:
+            self._send_batch(batch_to_send)
+        # Cancel any pending timer
+        if self._batch_timer is not None:
+            self._batch_timer.cancel()
+            self._batch_timer = None
+
+    def _send_batch(
+        self, batch_to_send: dict[str, tuple[str, Optional[Union[str, bytes]]]]
```
malhotra5:
```suggestion
        self, batch_to_send: dict[str, tuple[str, str | bytes | None]]
```

NIT: use modern typing syntax

openhands/storage/batched_web_hook.py
```diff
@@ -169,38 +168,42 @@ def _queue_update(
                 self._batch_timer = None
 
             timer = threading.Timer(
-                self.batch_timeout_seconds, self._send_batch_from_timer
+                self.batch_timeout_seconds, self._enqueue_batch_from_timer
```
malhotra5, line 172: Just to clarify, this timer is meant to trigger based on the **last** queue update? So if a new value comes through during the timer countdown the timer is reset?

malhotra5: It would be nice to document the timeout_seconds behavior more explicitly

openhands/storage/batched_web_hook.py
```diff
@@ -169,38 +168,42 @@ def _queue_update(
...
+
+    def _send_batch(
+        self, batch_to_send: dict[str, tuple[str, Optional[Union[str, bytes]]]]
+    ) -> None:
         """Send the current batch of updates to the webhook as a single request.
         This method acquires the batch lock and processes all pending updates in one batch.
         """
-        batch_to_send: dict[str, tuple[str, Optional[Union[str, bytes]]]] = {}
-
-        with self._batch_lock:
```
malhotra5, line 190: It seems like there was a lock on sending batches and the batch got cleared right after; not sure why we would see duplicate batches sent in this case

openhands/storage/batched_web_hook.py
```diff
@@ -159,8 +159,7 @@ def _queue_update(
 
             # Check if we need to send the batch due to size limit
             if self._batch_size >= self.batch_size_limit_bytes:
-                # Submit to executor to avoid blocking
-                EXECUTOR.submit(self._send_batch)
```
malhotra5, line 163: Ah okay, previously we acquired a lock to queue an update; if the batch size was exceeded we'd call `_send_batch` and attempt to acquire the same lock

openhands/storage/batched_web_hook.py
raymyers: Fair, I've added a comment to the top of the class clarifying what I believe the timer behavior to to be

raymyers:
Honestly even though I like this version better I'm still not sold on there being a problem with the original, because `_send_batch` taking the lock should have been enough

malhotra5 (approved):
Agreed, overall this is a positive gradient

raymyers:
I've added the timeout increase from 5sec to 15sec to this

raymyers:
> I've added the timeout increase from 5sec to 15sec to this

Scratch that, timeout is back the way it was because batch processing was moved to a background task

PR #10341, opened 2025-08-14
CLI vi mode is activated when standard env variables EDITOR/VISUAL are set to vi

juanmichelini:
Similar to the already existing CLI_VI_MODE environment variable, but EDITOR and VISUAL are standard and commonly set by vi/vim/nvim users.

- Auto-enable vi_mode in CLI when standard EDITOR or VISUAL environment variables contain vi, vim, or nvim.
- Supports full paths (e.g., /usr/bin/vim) and command arguments (e.g., vim -n)
- Explicit CLI_VI_MODE setting still takes precedence over auto-detection
- Added comprehensive tests covering various scenarios including edge cases
- Maintains backward compatibility with existing CLI_VI_MODE environment variable

---

---

### DIFF
--- a/openhands/core/config/utils.py
+++ b/openhands/core/config/utils.py
@@ -132,6 +132,29 @@ def load_from_env(
     # load default agent config from env
     default_agent_config = cfg.get_agent_config()
     set_attr_from_env(default_agent_config, 'AGENT_')
+    
+    # Auto-enable vi_mode for CLI config if EDITOR or VISUAL are set to vim-like editors
+    # Only do this if CLI_VI_MODE is not explicitly set
+    if 'CLI_VI_MODE' not in env_or_toml_dict and not cfg.cli.vi_mode:
+        editor = env_or_toml_dict.get('EDITOR', '').lower()
+        visual = env_or_toml_dict.get('VISUAL', '').lower()
+        
+        # Check if either EDITOR or VISUAL contains vim-like editor names
+        vim_editors = {'vi', 'vim', 'nvim'}
+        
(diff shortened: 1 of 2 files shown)


openhands/core/config/utils.py
```diff
@@ -133,6 +133,29 @@ def set_attr_from_env(sub_config: BaseModel, prefix: str = '') -> None:
     default_agent_config = cfg.get_agent_config()
     set_attr_from_env(default_agent_config, 'AGENT_')
 
+    # Auto-enable vi_mode for CLI config if EDITOR or VISUAL are set to vim-like editors
+    # Only do this if CLI_VI_MODE is not explicitly set
+    if 'CLI_VI_MODE' not in env_or_toml_dict and not cfg.cli.vi_mode:
```
enyst, line 138: Could you tell, if `CLI_VI_MODE` is set, then definitely, when the configuration is fully loaded, `config.cli.vi_mode` is also set?

In other words, we may need only the second condition?

openhands/core/config/utils.py
```diff
@@ -133,6 +133,29 @@ def set_attr_from_env(sub_config: BaseModel, prefix: str = '') -> None:
     default_agent_config = cfg.get_agent_config()
     set_attr_from_env(default_agent_config, 'AGENT_')
 
+    # Auto-enable vi_mode for CLI config if EDITOR or VISUAL are set to vim-like editors
```
enyst, line 136: Just a nit: I think maybe we can decouple the code into its own method, and maybe hook it in from `finalize_config`? It's a small thing, just to help keeping it straight a bit

Issue #10330, opened 2025-08-14
[Bug]: White on white text in CLI cannot be read


enyst:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Read the description and diff of PR https://github.com/All-Hands-AI/OpenHands/pull/10329

When this PR was tested, autocomplete on models names had, in part, some white text on white background. That was not in scope in the PR, but now we need to fix it.

Look up the reason for the white on white text in model autocomplete and fix it.

Additional detail: it happens on Ubuntu, and it might be due to its default terminal color scheme. Come up with proposals how to fix that, without cluttering the code. We don’t want system-specific code in the main CLI implementation, so whatever it is, should be some additional helper file.

### Read for Additional Context

- https://github.com/prompt-toolkit/python-prompt-toolkit/blob/3374ae9dd56f0265b41022e81341fd062092e2f0/src/prompt_toolkit/styles/defaults.py
- https://github.com/prompt-toolkit/python-prompt-toolkit/blob/3374ae9dd56f0265b41022e81341fd062092e2f0/src/prompt_toolkit/completion/fuzzy_completer.py

mamoodi:
Looks like this may be good to close?

enyst:
Ah, yes the agent didn't reference it, for some reason.

PR #10347, opened 2025-08-15
cli: fix Ubuntu white-on-white model autocomplete by merging default prompt_toolkit UI style

enyst:
Hi! I'm OpenHands-GPT-5, an automated developer assisting with improvements.

Context
- This addresses issue #10330: on some Ubuntu terminals, the CLI model autocomplete (prompt_toolkit) rendered white text on white background.

Root cause
- We passed a custom Style to PromptSession that replaced prompt_toolkit's default UI style. As a result, completion menu/fuzzy-match classes were missing, falling back to theme defaults; on some Ubuntu themes that meant white-on-white.

What this PR does
- Introduces a small helper: openhands/cli/pt_style.py with get_cli_style().
- get_cli_style() merges prompt_toolkit.styles.defaults.default_ui_style() with our minimal custom colors using merge_styles().
- Updates tui.py and settings.py to use get_cli_style() (and to import shared color constants).

Why this approach
- Centralized, minimal, and avoids OS-specific conditionals.
- Leverages prompt_toolkit defaults (per the references in the issue) to ensure proper contrast for completion menu and fuzzymatch classes across terminal themes.

Files changed
- add: openhands/cli/pt_style.py
- mod: openhands/cli/tui.py
- mod: openhands/cli/settings.py

Validation
- Backend pre-commit checks pass locally (ruff, mypy, etc.).

Co-authored-by: OpenHands-GPT-5 openhands@all-hands.dev

### DIFF
--- a/openhands/cli/pt_style.py
+++ b/openhands/cli/pt_style.py
@@ -1,0 +1,27 @@
+from prompt_toolkit.styles import Style, merge_styles
+from prompt_toolkit.styles.defaults import default_ui_style
+
+# Centralized helper for CLI styles so we can safely merge our custom colors
+# with prompt_toolkit's default UI style. This preserves completion menu and
+# fuzzy-match visibility across different terminal themes (e.g., Ubuntu).
+
+COLOR_GOLD = '#FFD700'
+COLOR_GREY = '#808080'
+COLOR_AGENT_BLUE = '#4682B4'  # Steel blue - readable on light/dark backgrounds
+
+
+def get_cli_style() -> Style:
+    base = default_ui_style()
+    custom = Style.from_dict(
+        {
+            'gold': COLOR_GOLD,
+            'grey': COLOR_GREY,
+            'prompt': f'{COLOR_GOLD} bold',
+            # The following entries are mostly redundant because default_ui_style
+            # already sets sane values. We keep them minimal on purpose to avoid
+            # clutter and ensure high contrast on various themes.
+            # 'completion-menu': 'bg:#bbbbbb #000000',
+            # 'completion-menu.completion.current': 'reverse',
+        }
+    )
+    return merge_styles([base, custom])
--- a/openhands/cli/tui.py
+++ b/openhands/cli/tui.py
@@ -31,6 +31,12 @@ from prompt_toolkit.styles import Style
 from prompt_toolkit.widgets import Frame, TextArea
 
 from openhands import __version__
+from openhands.cli.pt_style import (
+    COLOR_AGENT_BLUE,
+    COLOR_GOLD,
+    COLOR_GREY,
+    get_cli_style,
+)
 from openhands.core.config import OpenHandsConfig
 from openhands.core.schema import AgentState
 from openhands.events import EventSource, EventStream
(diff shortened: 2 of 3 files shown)


llamantino:
Tried it, doesn't fix it. The style that matters is `completion-menu.completion.current fuzzymatch.outside`. It doesn't have a text color by default, but you can set it:

`"completion-menu.completion.current fuzzymatch.outside": "fg:#000000"`

And that 'fixes' the issue. The `completion-menu.completion.current` parent has a #FFFFFF default background so it shouldn't cause trouble with other color schemes. In the end the colors will be swapped because the parent has a reverse property (so white text on black background).

llamantino:
Better idea, set the values to the opposite of the default `completion-menu.completion.current`, that means `fg:#ffffff bg:#888888` so it's consistent with what I think was the intended behavior.

openhands/cli/pt_style.py
```diff
@@ -0,0 +1,27 @@
...
+    custom = Style.from_dict(
+        {
+            'gold': COLOR_GOLD,
+            'grey': COLOR_GREY,
+            'prompt': f'{COLOR_GOLD} bold',
+            # Ensure good contrast for fuzzy matches on the selected completion row
+            # across terminals/themes (e.g., Ubuntu GNOME, Alacritty, Kitty).
+            # See https://github.com/All-Hands-AI/OpenHands/issues/10330
+            'completion-menu.completion.current fuzzymatch.outside': 'fg:#ffffff bg:#888888',
+            # If #ffffff feels too bright on your terminal, consider using #dddddd instead for the fg.
```
enyst:
```suggestion
```

openhands/cli/tui.py
```diff
@@ -64,16 +70,8 @@
...
-COLOR_GREY = '#808080'
-COLOR_AGENT_BLUE = '#4682B4'  # Steel blue - less saturated, works well on both light and dark backgrounds
-DEFAULT_STYLE = Style.from_dict(
-    {
-        'gold': COLOR_GOLD,
-        'grey': COLOR_GREY,
-        'prompt': f'{COLOR_GOLD} bold',
-    }
-)
+# Now sourced from pt_style to ensure we merge with prompt_toolkit defaults
```
enyst:
```suggestion
```

Abubakar-01:
@enyst  I  tried with multiple light themes but was not able to reproduce the issue in the main branch, can you let me know the specific reproducing steps?

enyst:
Cc: @llamantino you got it on ubuntu on default theme?

llamantino:
Yes, I had the issue on stock Ubuntu. Stock terminal, stock theme/color scheme.

Here's a screenshot:
image

I think it should reproduce on any terminal where, in the color scheme, the foreground text color is white.

Another example: stock Konsole on Archlinux+Plasma:
image

Abubakar-01:
> Yes, I had the issue on stock Ubuntu. Stock terminal, stock theme/color scheme.
> 
> Here's a screenshot: image
> 
> I think it should reproduce on any terminal where, in the color scheme, the foreground text color is white.
> 
> Another example: stock Konsole on Archlinux+Plasma: image

Thanks @llamantino

Abubakar-01:
:x: **Before ( Main Branch):**
image

:white_check_mark:  **After (PR Branch):**
image

Conclusion: The PR fixes the white-on-white text issue

enyst:
Thank you!!

I like that the PR takes the hardcoded values out of the main code, and keeps them in their own specialized styles file. This issue is styles across platforms, so I'd think if it works, we're good?

openhands/cli/pt_style.py
```diff
@@ -0,0 +1,26 @@
...
+            'gold': COLOR_GOLD,
+            'grey': COLOR_GREY,
+            'prompt': f'{COLOR_GOLD} bold',
+            # Ensure good contrast for fuzzy matches on the selected completion row
+            # across terminals/themes (e.g., Ubuntu GNOME, Alacritty, Kitty).
+            # See https://github.com/All-Hands-AI/OpenHands/issues/10330
+            'completion-menu.completion.current fuzzymatch.outside': 'fg:#ffffff bg:#888888',
+        }
+    )
+    return merge_styles([base, custom])
```
llamantino, line 26: I think the merge could be removed in the future. Custom styles are already applied over the default, so merging with the default shouldn't make any difference.

Issue #10237, opened 2025-08-11
[Bug]: AWS credentials in config.toml doesn't seem to work


wyc1997:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

In config.toml, there are settings for aws credential for bedrock model usage: 
```
[llm.bedrockclaude]
# AWS access key ID
# aws_access_key_id = ""
# AWS region name
# aws_region_name = ""
# AWS secret access key
# aws_secret_access_key = ""
model = "bedrock/us.anthropic.claude-3-7-sonnet-20250219-v1:0"
custom_llm_provider = "bedrock"
```

Then I used the SWEBench script in an attempt to invoke the claude model on bedrock to be the LLM. Specifically, I did: 

```
DEBUG=1 EXP_NAME=bedrock_test ./evaluation/benchmarks/swe_bench/scripts/run_infer.sh llm.bedrockclaude HEAD CodeActAgent 1 100 1 princeton-nlp/SWE-bench_Verified test
```

The `aws_access_key_id`, `aws_region_name`, `aws_secret_access_key` doesn't seem to have an effect on the LiteLLM backend. Even though I specified 'us-east-1' as the `aws_region_name`, LiteLLM seems to still invoke bedrock in the us-west-2 region, which leads to an error. 

I was able to get aws bedrock models to work by passing  `AWS_ACCESS_KEY_ID`, `AWS_REGION_NAME`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN` as environment variables. LiteLLM was able to detect these variables directly and was able to invoke bedrock models with these credentials. The aws credential settings in config.toml doesn't seem to do anything in terms of allowing aws bedrock model to run. 

Is this intended? It should be explained clearly if this is the intended behavior. It is unintuitive to have these settings in the config file but have no effect at all in the application. 

### OpenHands Installation

Docker command in README

### Operating System

None

npneeraj:
@xingyaoww kindly review the PR.

xingyaoww:
@npneeraj thank you!

@wyc1997 the issue should be fixed in https://github.com/All-Hands-AI/OpenHands/pull/10351 by @npneeraj 

Feel free to reopen if it still doesn't work

PR #10351, opened 2025-08-15
Added changes to resolve bug 10237

npneeraj:
This PR simply ensures aws credentials are set as environment variables when loading LLM configs.

---
**Link of any specific issues this addresses:** #10237


xingyaoww (commented):
~I don't think it will fix the issue in #10237 though -- our config management system already automatically load env var for LLMs.~

PR #10356, opened 2025-08-15
refactor(frontend): update the styling for recent conversations

hieptl:

**Acceptance criteria:**
- When in a loading state, there are more loading skeletons than needed. Since we show the top 3 before pressing “view more”, we should only show 3 skeletons.
- We now paginate conversations. If you open the conversation panel and scroll, the recent conversations section will also flicker between actual items and loading skeletons

---

The PR updates the styling for recent conversations.

https://github.com/user-attachments/assets/0d0ac0d4-e58f-4930-a580-ccacf7bd3d3e

---

### DIFF
--- a/frontend/src/components/features/conversation-panel/conversation-panel.tsx
+++ b/frontend/src/components/features/conversation-panel/conversation-panel.tsx
@@ -51,7 +51,9 @@ export function ConversationPanel({ onClose }: ConversationPanelProps) {
   } = usePaginatedConversations();
 
   // Flatten all pages into a single array of conversations
-  const conversations = data?.pages.flatMap((page) => page.results) ?? [];
+  let conversations = data?.pages.flatMap((page) => page.results) ?? [];
+
+  conversations = [...conversations, ...conversations, ...conversations];
 
   const { mutate: deleteConversation } = useDeleteConversation();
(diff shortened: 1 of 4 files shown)


hieptl:
> Hmm, I think this is too complicated a solution and includes changes beyond the scope of this PR (according to it’s title).
> 
> Let’s hold on this for now until we get feedback from Paul on how we might want to handle paginated conversations in the home page

Paul mentioned that the Recent Conversations section will display up to 10 conversations, so pagination may not be necessary.

However, you pointed out that:
- With conversation pagination enabled, opening the conversation panel and scrolling causes the Recent Conversations section to flicker between actual items and loading skeletons.

The root cause was that both the conversation panel and the Recent Conversations section were using the same conversation hook. I’ve updated the code to resolve this.

amanape:
OK sounds good, so maybe we can just simplify it to show 3 be default, then 10 when opened. No need to have the user click “View More” more than once. We also need to confirm that the conversation query gets at least 10 convos per page, otherwise if it is less this would mean the app needs to fetch a second time to render the correct number of conversations

hieptl:
Hi @amanape, I've updated the code based on your feedback! Thank you for your feedback! 🙏

Issue #10346, opened 2025-08-15
[Bug]: 'openai/gpt-5' and 'anthropic/claude-opus-4-1-20250805' not working when running openhands in headless model with docker


manasi-sharma:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

docker pull docker.all-hands.dev/all-hands-ai/runtime:0.53-nikolaik

docker run -it --rm --pull=always \
    -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.53-nikolaik \
    -e SANDBOX_USER_ID=$(id -u) \
    -e LLM_API_KEY=$OPEN_API_KEY \
    -e LLM_MODEL=openai/gpt-5 \
    -e LOG_ALL_EVENTS=true \
    -v /var/run/docker.sock:/var/run/docker.sock \
    -v "$STATE_DIR:/.openhands:rw" \
    -p 3000:3000 \
    --add-host host.docker.internal:host-gateway \
    --name "$CONTAINER_NAME" \
    docker.all-hands.dev/all-hands-ai/openhands:0.53 \
    python -m openhands.core.main -t "what is the capital of the moon?"

The issue persists when running with a LITELLM_KEY instead of an OPENAI_API_KEY (and correspondingly the ANTHROPIC_API_KEY for the claude 4.1 model ('anthropic/claude-opus-4-1-20250805'))

Error for OpenAI GPT-5 model ('openai/gpt-5'):
[Agent Controller ec9be6da-6f6e-4f-4559d33b96ef28f] AgentStateChangedObservation(content='', agent_state='error', reason="BadRequestError: litellm.BadRequestError: OpenAIException - Unsupported value: 'temperature' does not support 0.0 with this model. Only the default (1) value is supported.", observation=<ObservationType.AGENT_STATE_CHANGED: 'agent_state_changed'>)

Error for Anthropic Claude 4.1 model ('anthropic/claude-opus-4-1-20250805'):
[...]

Issue #10262, opened 2025-08-12
[Tech Debt]: Cleanup model name mapping and checking in `llm.py`


xingyaoww:
**What problem or use case are you trying to solve?**

In llm.py https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/llm/llm.py

> the models names are checked in like 4 different ways in the code, some are exact matches, some are partial, it's a bit of a mess (though if it works, better not mess with it).

**Describe the UX or technical implementation you have in mind**

We should maybe maintain a mapping similar to https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json

and checking it in a principaled way to make the decision of whether:
1. cache prompt supported
2. function calling supported
3. reasoning effort supported
4. whether to remove stop words

**Additional context**

enyst:
Co-authored-by: OpenHands-GPT-5 openhands@all-hands.dev

Hi! I'm OpenHands-GPT-5, an AI agent assisting on this issue. Below is a concise, maintainability-first proposal to clean up model capability checks in llm.py.

Summary of the problem
- Capability checks (prompt caching, function calling, reasoning effort, stop words) are scattered and inconsistent across LLM, AsyncLLM, and StreamingLLM.
- Matching uses a mix of exact equality, basename equality, and substring contains, which is error-prone and hard to extend.
- Async/Streaming paths partially duplicate (and diverge from) the sync LLM behavior (e.g., reasoning-effort handling), causing drift risk.

What “good” looks like
- A single source of truth for model capabilities (function calling, reasoning effort, prompt cache, supports stop words).
- A single, predictable model-name normalization and pattern-matching strategy.
- LLM, AsyncLLM, and StreamingLLM all consume the same capability API instead of re-implementing checks.
- Updating model support should require editing one table, not chasing conditionals across files.

Recommended approach (cleanest and least fragile)
1) Introduce a small capability registry module
   - New file: openhands/llm/capabilities.py
   - Provide:
     - normalize_model_name(model: str) -> str
       - Canonicalizes to lowercase basename (strips provider/proxy prefixes like openrouter/, litellm_proxy/, anthropic/, etc.).
     - model_matches(model: str, patterns: Iterable[str]) -> bool
       - Glob-style matching (fnmatch), e.g. "qwen3-coder*" or "gemini-2.5-pro*".
     - dataclass ModelCapabilities:
       - function_calling: bool
       - reasoning_effort: bool
       - prompt_cache: bool
[...]

PR #10363, opened 2025-08-15
fix(llm): include gpt-5 to fn call model; set top p default value to None

xingyaoww:

Fix the issue in LLM arguments when using `gpt-5` and `opus-4-1`

---

- make `top_p` None by default so it uses provider default and solves the opus-4-1 issue mentioned in https://github.com/All-Hands-AI/OpenHands/issues/10346
- add gpt-5 to fncall support model

---

Fix https://github.com/All-Hands-AI/OpenHands/issues/10346

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:a8d21ef-nikolaik   --name openhands-app-a8d21ef   docker.all-hands.dev/all-hands-ai/openhands:a8d21ef
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@xw/fix-10346 openhands
```

### DIFF
--- a/openhands/core/config/llm_config.py
+++ b/openhands/core/config/llm_config.py
@@ -67,7 +67,7 @@ class LLMConfig(BaseModel):
         default=30_000
     )  # maximum number of characters in an observation's content when sent to the llm
     temperature: float = Field(default=0.0)
-    top_p: float | None = Field(default=None)
+    top_p: float = Field(default=1.0)
     top_k: float | None = Field(default=None)
     custom_llm_provider: str | None = Field(default=None)
     max_input_tokens: int | None = Field(default=None)
--- a/openhands/core/config/llm_config.py
+++ b/openhands/core/config/llm_config.py
@@ -67,7 +67,7 @@ class LLMConfig(BaseModel):
         default=30_000
     )  # maximum number of characters in an observation's content when sent to the llm
     temperature: float = Field(default=0.0)
-    top_p: float = Field(default=1.0)
+    top_p: float | None = Field(default=None)
     top_k: float | None = Field(default=None)
     custom_llm_provider: str | None = Field(default=None)
     max_input_tokens: int | None = Field(default=None)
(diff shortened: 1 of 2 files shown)


openhands/core/config/llm_config.py
```diff
@@ -67,7 +67,7 @@ class LLMConfig(BaseModel):
         default=30_000
     )  # maximum number of characters in an observation's content when sent to the llm
     temperature: float = Field(default=0.0)
-    top_p: float = Field(default=1.0)
+    top_p: float | None = Field(default=None)
```
enyst: Are we sure about this? It seems it changes behavior for many other models.

We have in the code this:
https://github.com/All-Hands-AI/OpenHands/pull/10363/files#diff-b46ec9dd4b319d75809520e34b2559fd2ce46ca4895709eae9312aa04a8fa9a4R227

So we removed top_p for `REASONING_EFFORT_SUPPORTED_MODELS`. An alternative would be to add Opus and GPT-5 to that list... What do you think?

enyst: To be honest, I'd rather remove all those lists... in some way. I see more and more that we have no choice but to get more model-specific code in the codebase, so rather than hope to literally remove them one day, maybe a better option would be to abstract that knowledge in factory methods or llm builders or yaml files 🤔
I'll give it a try

xingyaoww: I might have this issue tracking this...
https://github.com/All-Hands-AI/OpenHands/issues/10262

i'll add opus to reasoning effort supported models for now 😩

enyst (approved):
LGTM!

openhands/core/config/llm_config.py
xingyaoww: done

PR #10389, opened 2025-08-15
runtime(bash): clarify guidance when previous command still running; recommend execute_bash timeout (issue #10350)

### LINKED ISSUE #10350: agent: continue long-running command shouldn't trigger "AgentStuckInLoop" Error

The agent can continue a long-running command by sending an empty command "" in a `CmdRunAction`. However, when it is doing it three times in a role, the agent controller will flag it as agent-stuck-in-loop and stop the whole conversation.

We should:
1. The continue bash command shouldn't trigger a stuck in a loop error (e.g., add an exception in stuck in loop detection for empty command that resumes previous command)
2. (1) is probably a bad solution since it adds a new special case --- A better way would be, we should update the prompt here (https://github.com/All-Hands-AI/OpenHands/blob/e40681ca61fb2b4b602fe916d99def6b55b78d77/openhands/runtime/utils/bash.py#L550-L557) to tell the agent the previous command timed out AND the agent should try use the `timeout` argument in the `execute_bash` tool to run this command for longer time.

We should probably try (2) first, if it didn't solve the problem; Then we should try (1).

**Additional context**

xingyaoww:
- [ ] Include this change in the Release Notes. If checked, you must provide an end-user friendly description for your change below

End-user friendly description of the problem this fixes or functionality this introduces.

When a new bash command is attempted while a prior command is still running, OpenHands returns guidance explaining the situation. This PR clarifies that guidance and explicitly recommends using the `timeout` parameter in the `execute_bash` tool for longer-running commands, to avoid the agent sending repeated empty commands and being flagged as stuck.

---
Summarize what the PR does, explaining any non-trivial design decisions.

- Update the suffix message set in `BashSession.execute()` for the "previous command still running" case.
- Preserves existing phrasing so current tests continue to pass, and appends a standardized hint via `TIMEOUT_MESSAGE_TEMPLATE` that mentions using the `timeout` parameter in `execute_bash`.
- No behavioral changes to execution logic; message-only change.

---
Link of any specific issues this addresses:

- Fixes #10350 (implements suggested fix (2))

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/2b02cf7209e64491b5c7ee32a422b692)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:5cf89cf-nikolaik   --name openhands-app-5cf89cf   docker.all-hands.dev/all-hands-ai/openhands:5cf89cf
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix/10350-bash-timeout-message openhands
```

### DIFF
--- a/openhands/runtime/utils/bash.py
+++ b/openhands/runtime/utils/bash.py
@@ -554,11 +554,9 @@ class BashSession:
             metadata = CmdOutputMetadata()  # No metadata available
             metadata.suffix = (
                 f'\n[Your command "{command}" is NOT executed. '
-                f'The previous command is still running - You CANNOT send new commands until the previous command is completed. '
+                'The previous command is still running - You CANNOT send new commands until the previous command is completed. '
                 'By setting `is_input` to `true`, you can interact with the current process: '
-                "You may wait longer to see additional output of the previous command by sending empty command '', "
-                'send other commands to interact with the current process, '
-                'or send keys ("C-c", "C-z", "C-d") to interrupt/kill the previous command before sending your new command.]'
+                f'{TIMEOUT_MESSAGE_TEMPLATE}]'
             )
             logger.debug(f'PREVIOUS COMMAND OUTPUT: {raw_command_output}')
             command_output = self._get_command_output(
--- a/openhands/runtime/utils/bash_constants.py
+++ b/openhands/runtime/utils/bash_constants.py
@@ -2,6 +2,6 @@
 TIMEOUT_MESSAGE_TEMPLATE = (
     "You may wait longer to see additional output by sending empty command '', "
     'send other commands to interact with the current process, '
-    'send keys to interrupt/kill the command, '
+    'send keys ("C-c", "C-z", "C-d") to interrupt/kill the previous command before sending your new command, '
     'or use the timeout parameter in execute_bash for future commands.'
 )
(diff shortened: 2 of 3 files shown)


openhands/runtime/utils/bash.py
```diff
@@ -554,11 +554,9 @@ def execute(self, action: CmdRunAction) -> CmdOutputObservation | ErrorObservati
             metadata = CmdOutputMetadata()  # No metadata available
             metadata.suffix = (
                 f'\n[Your command "{command}" is NOT executed. '
-                f'The previous command is still running - You CANNOT send new commands until the previous command is completed. '
+                'The previous command is still running - You CANNOT send new commands until the previous command is completed. '
                 'By setting `is_input` to `true`, you can interact with the current process: '
-                "You may wait longer to see additional output of the previous command by sending empty command '', "
-                'send other commands to interact with the current process, '
-                'or send keys ("C-c", "C-z", "C-d") to interrupt/kill the previous command before sending your new command.]'
```
enyst, line 561: Just to note, the template doesn't specify the keys. Does the LLM still know the keys?

https://github.com/All-Hands-AI/OpenHands/blob/06d283dfa0921469ea1be125085b7f170b011e76/openhands/runtime/utils/bash_constants.py#L2

xingyaoww: Done!
https://github.com/All-Hands-AI/OpenHands/pull/10389/commits/5cf89cff06cd2b54b2a36eab84584051fe782959

enyst (approved):
Thank you! A tiny bit of cleaner code is always welcome 🤗

PR #10406, opened 2025-08-15
fix(frontend): browser tab notification respects user-renamed titles

enyst:
- [x] Include this change in the Release Notes. If checked, you must provide an end-user friendly description for your change below

End-user friendly description of the problem this fixes or functionality this introduces.

Fixes an issue where the browser tab title could revert to a previous conversation title while the app was flashing a notification in the tab (e.g., when the agent becomes ready). If a user renamed the conversation during that time, the flasher kept toggling to the old title and even restored it when the notification stopped. Now the notification logic always respects the current title and restores to the latest one.

---
Summarize what the PR does, explaining any non-trivial design decisions.

The previous implementation of `browserTab.startNotification` cached the original title only once and toggled between that cached value and the message. This conflicted with PR #10275 (which updates the document title on conversation changes) and with manual user renames during an active notification.

Changes:
- Always capture the current `document.title` as the baseline when a notification starts
- While flashing, detect if `document.title` changed externally (and isn’t the message) and refresh the baseline to that new value
- Continue toggling between the message and the latest baseline; `stopNotification` restores that baseline

Tests:
- Added `frontend/__tests__/utils/browser-tab.test.ts` covering the external rename scenario during an active notification. Uses fake timers to simulate the interval and verifies we restore to the renamed title.

---
Link of any specific issues this addresses:

- Follow-up to #10275
- Addresses the Slack report about title reverting after user rename.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:3c313c1-nikolaik   --name openhands-app-3c313c1   docker.all-hands.dev/all-hands-ai/openhands:3c313c1
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix/browser-tab-notification-title openhands
```


amanape (approved):
This makes sense, thanks

PR #10414, opened 2025-08-15
Centralize model feature checks

enyst:

[Backend] Refactor out model feature checks; fix missing reasoning models.

---

(OpenHands-GPT-5)

Context
This is Phase 1 of addressing #10262. The goal is to centralize model feature knowledge and refactor the LLM code to use it. Focus was placed on code cleanliness, maintainability, and preserving behavior.

Summary
- Add openhands/llm/model_features.py
  - normalize_model_name(), model_matches()
  - ModelFeatures dataclass
  - Pattern tables for:
    - function_calling
    - reasoning_effort
    - prompt_cache
    - supports_stop_words
  - get_features(model) -> ModelFeatures
- Refactor LLM usage to consume model features
  - openhands/llm/llm.py
    - Import get_features at module scope
    - Replace scattered checks with feats = get_features(self.config.model)
    - Preserve Gemini mapping: reasoning_effort=None/low => thinking budget; other values pass-through
    - Remove temperature/top_p for reasoning models (existing behavior)
    - Prompt caching via feats.prompt_cache
    - Stop-words gating via feats.supports_stop_words
    - Native tool-calling activation cached on init (config override respected)
  - openhands/llm/async_llm.py
    - Use features for reasoning effort gating
  - openhands/llm/streaming_llm.py
    - Use features for reasoning effort gating
- Minor: LLM latency calculation made robust to mocked time.time side effects in tests

Behavior parity
- Retains existing behaviors (Gemini reasoning mapping, o-series stop-word exclusion, Anthropic prompt cache handling, etc.)
- No new external deps; stdlib only (dataclasses, fnmatch)

Why this design?
- One file (model_features.py) as the single source of truth for model families and features
- Simple glob patterns make it easy to add or adjust coverage without touching call sites
- Avoids repeated imports and scattered constants across modules
- Improves readability and reduces risk of diverging logic

Testing
- Linting/typing passed locally using dev_config (ruff, format, mypy)
- Unit tests updated to reflect rename:
  - tests/unit/llm/test_model_features.py
  - tests/unit/llm/test_llm.py

Open questions / Next steps (Phase 2)
- Add more unit tests for model_features normalization and pattern matching
- Consider consolidating common kwargs handling into small helpers (e.g., _base_completion_kwargs, _apply_reasoning_params) for further cleanliness if desired
- Evaluate whether to surface features on ModelInfo (if provided by LiteLLM) to reduce pattern maintenance in the future
- Decide on top-level pre-commit config approach (repo hook currently calls pre-commit when none exists at root):
  - keep bypass (PRE_COMMIT_ALLOW_NO_CONFIG=1)
  - or add/symlink a top-level .pre-commit-config.yaml to dev_config/python/.pre-commit-config.yaml
  - or uninstall local pre-commit hook and rely on CI

Linking
- Closes no behavior gaps; this is refactor-only groundwork for #10262. Please review approach and pattern coverage. Follow-ups will address any additional tests and small cleanups.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:7c37d81-nikolaik   --name openhands-app-7c37d81   docker.all-hands.dev/all-hands-ai/openhands:7c37d81
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@feat/llm-capabilities-registry openhands
```

### DIFF
--- a/tests/unit/llm/test_model_features.py
+++ b/tests/unit/llm/test_model_features.py
@@ -15,7 +15,7 @@ from openhands.llm.model_features import (
         ('anthropic/claude-3-7-sonnet', 'claude-3-7-sonnet'),
         ('litellm_proxy/gemini-2.5-pro', 'gemini-2.5-pro'),
         ('qwen3-coder-480b-a35b-instruct', 'qwen3-coder-480b-a35b-instruct'),
-        ('gpt-5', 'gpt-5-preview'),
+        ('gpt-5', 'gpt-5'),
         ('deepseek/DeepSeek-R1-0528:671b-Q4_K_XL', 'deepseek-r1-0528'),
         ('openai/GLM-4.5-GGUF', 'glm-4.5'),
         ('openrouter/gpt-4o-mini', 'gpt-4o-mini'),


enyst:
(OpenHands-GPT-5)

Update: merged latest main into feat/llm-capabilities-registry and ran full pre-commit with dev_config/python/.pre-commit-config.yaml — all hooks passed.

- Added tests for capabilities registry: tests/unit/llm/test_capabilities.py
  - Covers normalize_model_name, model_matches, get_capabilities
  - Includes expanded cases for function calling, reasoning effort, prompt cache, stop-words families
- Stabilized LLM latency test under time.time mocking; removed a noisy debug log in LLMConfig.post_init (no behavior change)
- LLM tests green: tests/unit/llm/test_llm.py and tests/unit/llm/test_capabilities.py

This is Phase 1 for #10262, focusing on maintainability by centralizing model capability logic. Open to feedback on pattern tables and any additional model families to capture.

openhands/llm/llm.py
```diff
@@ -362,7 +288,12 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
                 resp: ModelResponse = self._completion_unwrapped(*args, **kwargs)
 
             # Calculate and record latency
-            latency = time.time() - start_time
+            # Ensure monotonic measurement even under mocked/limited time.time() side effects
+            end_time = time.time()
+            try:
+                latency = max(0.0, float(end_time - start_time))
+            except Exception:
+                latency = 0.0
```
```suggestion
            latency = time.time() - start_time
```

enyst:
> I think this is definitely net improvement! Happy to merge this first after OH fixes the nit!

Sorry, what is the nit? I don’t see a comment here or on slack?

(On the other hand, please don’t merge, I’d prefer to merge it myself: if code review is fine, I’d like to give it a few runs in the debugger just to make sure it does the right thing.)

openhands/llm/model_features.py
```diff
@@ -0,0 +1,110 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from fnmatch import fnmatch
+
+
+def normalize_model_name(model: str) -> str:
```
llamantino, line 7: In this normalization function, could you please have it support model names like
- `deepseek/DeepSeek-R1-0528:671b-Q4_K_XL` (remove everything starting from ':')
- `openai/GLM-4.5-GGUF` (remove '-GGUF')
(Source: #10158)

openhands/llm/model_features.py
```diff
@@ -0,0 +1,110 @@
...
+        name = name.split('/')[-1]
+    return name
+
+
+def model_matches(model: str, patterns: list[str]) -> bool:
+    """Return True if the normalized model matches any of the glob patterns.
+
+    Uses fnmatch for simple glob-style matching.
+    """
+    name = normalize_model_name(model)
```
llamantino, line 41: What if some features are available/not available depending on the provider? A name match alone wouldn't cut it. If the pattern contains a provider name, maybe you should compare also against the provider name.

openhands/llm/model_features.py
```diff
@@ -0,0 +1,110 @@
...
+    'claude-sonnet-3-7-latest',
+    'claude-3-5-sonnet*',
+    'claude-3-5-haiku*',
+    'claude-3-haiku-20240307',
+    'claude-3-opus-20240229',
+    'claude-sonnet-4*',
+    'claude-opus-4*',
+]
+
+SUPPORTS_STOP_WORDS_FALSE_PATTERNS: list[str] = [
```
llamantino, line 121: Could you also please add 'DeepSeek-R1-0528' to the list? Thanks!

openhands/llm/model_features.py
enyst: This is a refactoring, so it shouldn't change current behavior. If something was per provider, it should still be, if it wasn't, it shouldn't.

enyst: That's the rough idea, anyway. If you think we could "sin" a little, I'm happy to consider it!

openhands/llm/model_features.py
enyst: OK, will do!

openhands/llm/model_features.py
enyst: > A name match alone wouldn't cut it. 

Are you sure...? The model name in llm.py typically includes provider name

llamantino: image

This was in the original, but likely a mistake.

Edit: Fun fact, by the way: the one with the provider prefix was likely the only match that worked in the original code, since it was an exact match against the full model string.

llamantino: > > A name match alone wouldn't cut it.
> 
> Are you sure...? The model name in llm.py typically includes provider name

The normalize_model_name function strips the provider name. You would match the model name, without the provider, against the patterns.

openhands/llm/model_features.py
```diff
@@ -0,0 +1,110 @@
...
+
+def get_features(model: str) -> ModelFeatures:
+    function_calling = model_matches(model, FUNCTION_CALLING_PATTERNS)
+    reasoning_effort = model_matches(model, REASONING_EFFORT_PATTERNS)
+    prompt_cache = model_matches(model, PROMPT_CACHE_PATTERNS)
+    supports_stop_words = not model_matches(model, SUPPORTS_STOP_WORDS_FALSE_PATTERNS)
+    return ModelFeatures(
+        function_calling=function_calling,
+        reasoning_effort=reasoning_effort,
+        prompt_cache=prompt_cache,
```
xingyaoww: the above three should starts with prefix `support_`?

xingyaoww: @enyst this is the one! for some reason 😓 i never send this out

llamantino:
Might also want to check this eventually (provided that it's still a useful check):
https://github.com/All-Hands-AI/OpenHands/blob/476954f3a4b1a2c38375e4628fcccec0e5dbff20/openhands/agenthub/codeact_agent/codeact_agent.py#L108-L120

openhands/llm/model_features.py
enyst: > Edit: Fun fact, by the way: the one with the provider prefix was likely the only match that worked in the original code, since it was an exact match against the full model string.

Oof, yes, I think this is correct. The check should have been the other way around, `is string in model name`. So it was bugged.

The strange thing about this kind of bugs is that litellm actually ALSO checks these things, or should. So we have these checks because sometimes it didn't, or it didn't yet, or there was a special reason... I wonder if we could remove the STOP words check completely...?

enyst:
OpenHands-GPT-5: concise update

- Centralized capability checks (provider-aware patterns) in model_features; preserves main behavior.
- Renamed fields to supports_*; updated call sites (is_caching_prompt_active; reasoning).
- Normalization handles ':…' and '-gguf'.
- Added DeepSeek R1 reasoning pattern + tests; provider-qualified matching + normalization tests.
- Pre-commit hooks passed; unit tests green; branch pushed (feat/llm-capabilities-registry). Next: let CI run.

openhands/llm/model_features.py
```diff
@@ -12,30 +12,46 @@ def normalize_model_name(model: str) -> str:
...
     """
-    name = (model or '').strip().lower()
-    if '/' in name:
-        name = name.split('/')[-1]
+    raw = (model or '').strip().lower()
+    # Keep only the basename after the last '/'
+    name = raw.split('/')[-1] if '/' in raw else raw
+    # Remove anything after the first ':' (ollama-style variants)
+    if ':' in name:
+        name = name.split(':', 1)[0]
```
enyst: Not sure about this, this isn't just Ollama-style, it's also Bedrock style, for example. And for Bedrock, the version is after, IIRC

openhands/llm/model_features.py
enyst: Done!

openhands/llm/model_features.py
enyst: Done!

openhands/llm/model_features.py
enyst: Right now, the new check is against normalized_model_name without provider, and stop words should apply to the old list (against grok, not xai/grok). That seems fine?

openhands/llm/model_features.py
enyst: Should be fixed!

enyst:
> Might also want to check this eventually (provided that it's still a useful check):

Checked, we didn't change the model name from config or from LLM instance, here, so that code is unaffected. It's an open question if we can remove it altogether. 🤔 It's not in scope here, but maybe this PR could remove it, if we're sure
- https://github.com/All-Hands-AI/OpenHands/pull/10443

openhands/llm/model_features.py
```diff
@@ -0,0 +1,128 @@
...
+]
+
+
+def get_features(model: str) -> ModelFeatures:
+    supports_function_calling = model_matches(model, FUNCTION_CALLING_PATTERNS)
+    supports_reasoning_effort = model_matches(model, REASONING_EFFORT_PATTERNS)
+    supports_prompt_cache = model_matches(model, PROMPT_CACHE_PATTERNS)
+    supports_stop_words = not model_matches(model, SUPPORTS_STOP_WORDS_FALSE_PATTERNS)
+    return ModelFeatures(
+        supports_function_calling=supports_function_calling,
```
xingyaoww: nit: we can probably just do this to save a few lines?
```suggestion
        supports_function_calling=model_matches(model, FUNCTION_CALLING_PATTERNS),
```

tests/unit/llm/test_model_features.py
```diff
@@ -0,0 +1,181 @@
...
+    ],
+)
+def test_model_matches_provider_qualified(name, pattern, expected):
+    assert model_matches(name, [pattern]) is expected
+
+
+@pytest.mark.parametrize(
+    'model,expect',
+    [
+        ('gpt-4o', ModelFeatures(True, False, False, True)),
```
xingyaoww: nit: ideally we can do kwargs here like support_XXX=true so it is easier to read. but this is also fine

xingyaoww (approved):
Definitely a net improvement-- thank you!

openhands/llm/model_features.py
enyst: And now it lost it, doh. Bringing it back.

tests/unit/llm/test_model_features.py
enyst: Done!

openhands/llm/model_features.py
enyst: Fixed

openhands/llm/model_features.py
enyst: Played locally with grok on this one, and Sonnet, Gemini-2.5-Pro, from litellm proxy, without proxy, and it worked.

openhands/llm/model_features.py
llamantino: Ouch. It looks like they use it to separate major from minor version numbers (e.g. Claude v2:1 for Claude 2.1). I haven't seen any ':' patterns in the lists, but if needed, those cases could be handled the same way as provider matches, with a more specific/exact match if ':' is present.

llamantino:
> Checked, we didn't change the model name from config or from LLM instance, here, so that code is unaffected. It's an open question if we can remove it altogether. 🤔 It's not in scope here, but maybe this PR could remove it, if we're sure

Sorry, I meant checking whether it was worth using the new functions to match the name there too, since that's another spot where model names are being checked in yet another different way, using substring matches.

openhands/llm/model_features.py
enyst: I actually checked and it looks like it's not model version numbers, for example this one was real, I'm pretty sure:
"/bedrock/anthropic.claude-3-5-sonnet-20241022-v2:0"

We can probably ignore that, until proven a problem

PR #10454, opened 2025-08-18
fix(openrouter): Force string serialization for openrouter/anthropic/claude-sonnet-4 model

suixinio:

---

- **The Issue**:
  API requests were failing with the validation error: `messages.1.content.0.text.text: Input should be a valid string`.

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -850,6 +850,8 @@ class LLM(RetryMixin, DebugMixin):
                 message.force_string_serializer = True
             if 'kimi-k2-instruct' in self.config.model and 'groq' in self.config.model:
                 message.force_string_serializer = True
+            if 'openrouter/anthropic/claude-sonnet-4' in self.config.model:
+                message.force_string_serializer = True
 
         # let pydantic handle the serialization
         return [message.model_dump() for message in messages]


suixinio:
Hi, @enyst !

Thanks for the quick reply!

Yes, it's a consistent issue that I can reproduce reliably. I've attached my environment details and some screenshots below to help.

To be honest, I was just as surprised as you are! I also searched through the docs for litellm, OpenHands, and OpenRouter but came up empty. Since I'm not sure if something else is causing this, I went with the most minimal fix I could think of. 😅

-----
**env**:
```
docker pull docker.all-hands.dev/all-hands-ai/runtime:0.53-nikolaik

docker run -it --rm --pull=always \
    -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.53-nikolaik \
    -e LOG_ALL_EVENTS=true \
    -v ~/.openhands:/.openhands \
    -p 3000:3000 \
    --name openhands-app \
    docker.all-hands.dev/all-hands-ai/openhands:0.53
```

enyst:
Ahh OpenRouter got the model from Bedrock, oof, idk, Bedrock can be weird. 😥

OK, let's get your fix in, it will be useful to others! Thank you for the contribution!

PR #10466, opened 2025-08-18
(Hotfix): Fix eval pipeline

malhotra5:

---
The recent introduction of LLM Registry broke existing eval pipelines. This is because the registry param wasn't passed in to the runtime, so this PR makes it optional

Going forward, it would be great to 

1. Have single entrypoint for creating runtimes, etc; we have multiple places with duplicate code for spinning up runtimes, etc
2. Remove the llm registry requirement for the runtime; the runtime instantiates an LLM at the moment, which should be part of the agent instead. 

 

---

Related to ALL-2380

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:25b32dd-nikolaik   --name openhands-app-25b32dd   docker.all-hands.dev/all-hands-ai/openhands:25b32dd
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix-registry-for-eval openhands
```

### DIFF
--- a/tests/unit/test_memory.py
+++ b/tests/unit/test_memory.py
@@ -98,9 +98,7 @@ def mock_agent():
 
 
 @pytest.mark.asyncio
-async def test_memory_on_event_exception_handling(
-    memory, event_stream, mock_agent
-):
+async def test_memory_on_event_exception_handling(memory, event_stream, mock_agent):
     """Test that exceptions in Memory.on_event are properly handled via status callback."""
     # Create a mock runtime
     runtime = MagicMock(spec=ActionExecutionClient)


enyst:
That label is pretty good for testing IMHO: run-eval-1 or run-eval-2

Zacharias030:
PR breaks modal runtime afaict.

PR #9815, opened 2025-07-20
Skip browser dependency build in Dockerfile when browser is disabled

li-boxuan:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [x] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

Speed up runtime image build by 18% when browser is disabled

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

When browser infra is disabled in core config, there's no reason we need to install playwright and chromium. This shrinks Docker image size and speeds up build.

Previous build time on my mac: 2m32s, size: 5.51GB
New build time on my mac (when browser is disabled): 2m5s, size: 4.51GB. That’s 18% smaller in size and 18% faster in speed

---
**Link of any specific issues this addresses:**

---

To run this PR locally, use the following command:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:876ad7f-nikolaik   --name openhands-app-876ad7f   docker.all-hands.dev/all-hands-ai/openhands:876ad7f
```

PR #10476, opened 2025-08-18
fix jinja / dockerfile syntax by removing newlines

Zacharias030:

This fixes a bug in the templated Dockerfile.j2 that was introduced in #9815.

---

This fixes a syntax error issue in the templated Dockerfile.j2 that was introduced in https://github.com/All-Hands-AI/OpenHands/pull/9815. When the condition is true the line breaks that I removed prevented the Dockerfile RUN command to span across the whole command leading to an error.

---


Zacharias030:
cc: @xingyaoww // @raymyers // @li-boxuan

li-boxuan (approved):
Thanks! I wonder if it depends on a specific jinja2 version, since it worked for me.

Regardless, I don't see why we should reject this PR

PR #10479, opened 2025-08-18
Update modal runtime for modal>=1.0

Zacharias030:

Updates the modal runtime to work with modal>1.0

---
Updates the modal runtime to work with modal>1.0, where the use of modal.Mount is deprecated (see https://modal.com/docs/guide/modal-1-0-migration#deprecating-mount-as-part-of-the-public-api)
Additionally, the sandboxed require a bit of time to not hit the retry timeouts, which is why I added 20s of sleep. I have confirmed with some SWE-Bench Verified images that not including the sleep let's the OpenHands agent fail the swebench tasks.

---

### DIFF
--- a/third_party/runtime/impl/modal/modal_runtime.py
+++ b/third_party/runtime/impl/modal/modal_runtime.py
@@ -150,7 +150,7 @@ class ModalRuntime(ActionExecutionClient):
             raise Exception("Sandbox not initialized")
         tunnel = self.sandbox.tunnels()[self.container_port]
         self.api_url = tunnel.url
-       self.log("info", "Waiting 20 secs for the container to be ready... (avoiding RemoteProtocolError)")
+        self.log("info", "Waiting 20 secs for the container to be ready... (avoiding RemoteProtocolError)")
         sleep(20)
         self.log("debug", f"Container started. Server url: {self.api_url}")
 


third_party/runtime/impl/modal/modal_runtime.py
```diff
@@ -149,6 +150,8 @@ async def connect(self):
             raise Exception("Sandbox not initialized")
         tunnel = self.sandbox.tunnels()[self.container_port]
         self.api_url = tunnel.url
+        print("Waiting 20 secs for the container to be ready... (avoiding RemoteProtocolError)")
```
ryanhoangt:
```suggestion
       self.log("info", "Waiting 20 secs for the container to be ready... (avoiding RemoteProtocolError)")
```

ryanhoangt (approved):
LGTM, thank you!

third_party/runtime/impl/modal/modal_runtime.py
```diff
@@ -149,6 +150,8 @@ async def connect(self):
             raise Exception("Sandbox not initialized")
         tunnel = self.sandbox.tunnels()[self.container_port]
         self.api_url = tunnel.url
+       self.log("info", "Waiting 20 secs for the container to be ready... (avoiding RemoteProtocolError)")
```
ryanhoangt:
```suggestion
        self.log("info", "Waiting 20 secs for the container to be ready... (avoiding RemoteProtocolError)")
```

Issue #9970, opened 2025-07-29
Planning Agent: Support a planning tool that maintains a TODO.md


xingyaoww:
**What problem or use case are you trying to solve?**

We currently has a long horizon prompt (https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/agenthub/codeact_agent/prompts/system_prompt_long_horizon.j2)

but according to feednback from @valeriechen, it is not working super well:

> im testing the TODO.md stuff and im noticing even after refining the prompt a bit that:
> 1. the agent will update the entire todo list at once instead of step by step
> 2. related to the first point the agent will tell the user what its doing but it doesnt actually make any changes to TODO.md
> 3. the agent might even update the todo list with new items while its working without telling the user
> here's the version of prompt that im testing https://github.com/All-Hands-AI/OpenHands/pull/9861/files

**Describe the UX or technical implementation you have in mind**

**Approach**

Can we take some inspiration of Claude-Code to implement a new "planning" tool that essentially just maintains a local json or markdown file under .openhands/plan.json?
- system prompt: https://github.com/Yuyz0112/claude-code-reverse/blob/main/results/prompts/system-workflow.prompt.md#task-management
- and todowrite tool: https://github.com/Yuyz0112/claude-code-reverse/blob/main/results/tools/TodoWrite.tool.yaml
- AND sequential thinking MCP: https://github.com/modelcontextprotocol/servers/tree/main/src/sequentialthinking

**Goal**

- comparable swebench score
- pass our vibe test (it updates plan properly for each step)

**Additional context**

jpelletier1:
@xingyaoww I noticed this ticket is very similar to #8964, and both are tagged on roadmap 🙃  Are there key differences or should I de-duplicate?

xingyaoww:
I think https://github.com/All-Hands-AI/OpenHands/issues/8964 is a higher-level PR, adding the general "Plan" mode and the ability to switch to/from "Plan" and "Execute" mode.

This issue is for adding the ability to manage a todo list via a tool -- so it should be working for *both* "plan" and "execute" model when both are implemented.

jpelletier1:
@xingyaoww Makes sense! I will make this issue a child of #8964

PR #10493, opened 2025-08-19
Move TASKS.md to session-specific directory in `~/.openhands`

ryanhoangt:

---

Currently the task_tracker tool saves the TASKS.md file in the current working directory, which creates interference between sessions, e.g. when using the OH CLI in the same directory. This PR moves it to session-specific directory under `~/.openhands`.

---


ryanhoangt:
> Actually, this means it will be hidden from the agent

Yeah, we don't want the agent to deal with this file directly, but to interact with the `task_tracker` tool. The file is just a way to keep the current plan that the agent is following. I think we can even store it in-memory as well, the file persistence idea is from @xingyaoww in the issue description https://github.com/All-Hands-AI/OpenHands/issues/9970. Not sure if he has other thoughts here.

> What exactly happened, with CLI?

Currently we save the file in `./.openhands/TASKS.md` so subsequent sessions might pick up the `TASKS.md` of previous ones: https://openhands-ai.slack.com/archives/C086ARSNMGA/p1755541434628049

> If the agent needs to change code for a task, it will change code. If the directory it works in is shared with another task, the code changes are also a mess, aren’t they?

Yeah, so we also have another option which is to ask users if they want to override the existing one, but I'm not certain which one is better here.

enyst:
So far, GPT-5 doesn’t seem to care: 😅

> Cleaned up a stray .openhands/TASKS.md from the branch to keep CI happy.

Saw this in passing, deep into a bunch of things it was doing on two different branches in the same conversation.

openhands/runtime/base.py
```diff
@@ -876,8 +877,14 @@ def run_action(self, action: Action) -> Observation:
             if isinstance(action, AgentThinkAction):
                 return AgentThinkObservation('Your thought has been logged.')
             elif isinstance(action, TaskTrackingAction):
-                # If `command` is `plan`, write the serialized task list to the file TASKS.md under `.openhands/`
+                # Get the session-specific task file path
+                conversation_dir = get_conversation_dir(
+                    self.sid, self.event_stream.user_id
+                )
+                task_file_path = f'{conversation_dir}TASKS.md'
```
xingyaoww, line 884: can you use maybe os.path.joint for this?

xingyaoww (approved):
LGTM! Just one nit

openhands/runtime/base.py
ryanhoangt: Actually I saw this pattern is used in several places. If it seems safe maybe we can just go with it?

image

xingyaoww: Let go with it

PR #10175, opened 2025-08-08
[draft] WIP

amanape:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

---
**Link of any specific issues this addresses:**

PR #10519, opened 2025-08-20
fix: subscription logic by polling for available runtime (microagent management, memory UI)

hieptl:

With the transition to nested runtimes, the subscription logic attempted to establish a WebSocket connection using an invalid or not-yet-ready URL.

This update aligns the behavior with the primary WebSocket provider by polling until a valid runtime is available before opening the connection.

**Reproducing the issue**
You can reproduce the issue locally by using `DockerNestedConversationManager.

**How to use `DockerNestedConversationManager`**
Update the `server_config.py` file as follows:

```js
conversation_manager_class: str = os.environ.get(
    'CONVERSATION_MANAGER_CLASS',
    'openhands.server.conversation_manager.docker_nested_conversation_manager.DockerNestedConversationManager',
)
```

---

This PR addresses the following issues:

- **Subscription logic**  
  Updated to poll for an available runtime (affecting microagent management and the memory UI).  

- **Runtime status**  
  Ensures that `runtime_status` is properly returned from the nested runtime, which was previously missing.  

- **Initial user message handling**  
  Fixed an issue where the nested runtime did not correctly handle the `initial_user_message`.  
  This bug also caused the memory UI feature to fail when using `DockerNestedConversationManager`.  

### Demonstration Videos

- **Video 1:** Microagent management page working with `DockerNestedConversationManager`. 

https://github.com/user-attachments/assets/73b46680-5d1a-47dd-8af8-cabaeda5aaff

- **Video 2:** Microagent management page working with `StandaloneConversationManager` (non-nested runtime). 

https://github.com/user-attachments/assets/8bf542fa-b00d-41f6-9bc4-a8aecf9994e4
 
- **Video 3:** Memory UI working with `DockerNestedConversationManager`.  

https://github.com/user-attachments/assets/7b18ecf1-616a-4315-b9cb-1ab881e56eb6

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:4afc166-nikolaik   --name openhands-app-4afc166   docker.all-hands.dev/all-hands-ai/openhands:4afc166
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@hieptl/all-3209 openhands
```


frontend/src/components/features/microagent-management/microagent-management-content.tsx
```diff
@@ -182,11 +170,7 @@ export function MicroagentManagementContent() {
       // Check if agent has finished and we have a PR
       if (isOpenHandsEvent(socketEvent) && isFinishAction(socketEvent)) {
         const prUrl = getFirstPRUrl(socketEvent.args.final_thought || "");
-        if (prUrl) {
-          displaySuccessToast(
-            t(I18nKey.MICROAGENT_MANAGEMENT$PR_READY_FOR_REVIEW),
-          );
-        } else {
```
amanape, line 189: Why not show it if available?

frontend/src/api/open-hands.ts
```diff
@@ -721,6 +721,9 @@ class OpenHands {
    * @returns List of installation IDs
    */
   static async getUserInstallationIds(provider: Provider): Promise<string[]> {
+    if (!provider) {
+      return [];
+    }
```
tofarr: Should an invalid parameter raise an error? Can null actually get passed in here?

tofarr (approved):
LGTM! It makes sense that we get the runtime status and return it in the agent loop info.

amanape:
@hieptl we can close #10175 too since this merges to `main`, right?

hieptl:
> #10175

Hi @amanape, yes, we can close that PR. Thank you! 🙏

frontend/src/components/features/microagent-management/microagent-management-content.tsx
hieptl: Hi @amanape, I noticed we already have the ConversationFinishedToast component in the microagent-status-toast.tsx file. To avoid showing duplicate toast messages for the same purpose, I’ve removed it.

frontend/src/api/open-hands.ts
hieptl: Hi @tofarr, if the provider is null, an error will be thrown and the status code will be 422. In the useAppInstallations hook, the provider can indeed be null.

amanape: Actually this check shouldn’t even be needed here since `Provider` should always be a string value. In fact, we don’t call this API unless the provider exists: https://github.com/All-Hands-AI/OpenHands/blob/049f058ed19be9dc1da5d7a5a7da8cb053e67b75/frontend/src/hooks/query/use-app-installations.ts#L17-L20

@hieptl Any particular reason why you added this here? This should be safe to remove

hieptl: I’m using the useGitRepositories hook to fetch the list of repositories, which internally calls the useAppInstallation hook.

Because the API in useAppInstallation runs before the one in useUserProviders, the provider is null, resulting in a 422 error.

I’ve implemented a solution and updated the code. I also removed the following snippet:

```ts
if (!provider) {
  return [];
}
```

Thank you! 🙏

amanape (approved):

PR #10539, opened 2025-08-20
fix: normalize malformed <parameter> tags to unbreak str_replace with Qwen3 (fix #10039)

### LINKED ISSUE #10039: [Bug]: str_replace error when using qwen3 coder locally.

Thanks for openhands cli, ive been using it with devstral and it works great.

Qwen3 coder isnt jiving as well as devstral with openhands. 80-90% of the time it doesnt have any problems.

│Parameter 'command=str_replace</parameter' is not allowed for function 'str_replace_editor'. Allowed parameters: {'view_range', 'path', 'new_str', 'old_│
│str', 'insert_line', 'command', 'file_text'} 

This is the error that it sometimes has. 

### OpenHands Installation

CLI

### OpenHands Version

0.50.0

### Model Name

qwen3-coder-30b-a3b-instruct

### Operating System

None

_No response_

enyst:
OpenHands-GPT-5 here.

This PR addresses issue #10039 where certain models (notably Qwen3 coder) occasionally emit malformed function-calling parameter tags for the str_replace_editor tool, e.g.:

<parameter=command=str_replace</parameter>

instead of the correct

<parameter=command>str_replace</parameter>

Root cause analysis
- The model sometimes inlines the value into the parameter attribute position. Our non-fncall -> fncall converter relied on the strict regex <parameter=name>value</parameter> and therefore parsed the entire malformed fragment as the parameter name, producing a FunctionCallValidationError like:
  Parameter 'command=str_replace</parameter' is not allowed for function 'str_replace_editor'.
- While this is fundamentally a model emission bug, the OpenHands framework should be resilient and treat this as a recoverable validation issue rather than bricking the run.

What’s changed
1) Robust normalization before parsing
   - New helper _normalize_parameter_tags(fn_body) in openhands/llm/fn_call_converter.py rewrites malformed tags from <parameter=name=value</parameter> to <parameter=name>value</parameter> prior to running the usual parameter extraction.
   - This makes the converter tolerant of this specific formatting glitch while keeping strict validation for other cases.

2) Parser fallback clean-up (minor)
   - The extraction loop was adjusted earlier but the final approach relies on the dedicated normalization step so the main behavior stays clean and explicit.

3) Unit test added
   - tests/unit/llm/test_llm_fncall_converter.py::test_malformed_parameter_parsing_recovery ensures we correctly parse the malformed form into a valid tool call with arguments:
     {"command": "str_replace", "path": "/repo/app.py", "old_str": "foo", "new_str": "bar"}

Why this is safe
- The normalization targets a very specific malformed pattern and converts it to the canonical representation expected by the rest of the pipeline.
- Other validation behavior (required/allowed params, enum, types) is unchanged.

Developer notes
- This still surfaces as a FunctionCallValidationError for other kinds of bad inputs, which is intended and is handled as ErrorObservation by the agent controller. The key improvement is that we now accept the common malformed pattern observed with Qwen3 so the agent can proceed.

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/23d9d183c8a14d28b657d515e791746a)

### DIFF
--- a/openhands/llm/fn_call_converter.py
+++ b/openhands/llm/fn_call_converter.py
@@ -642,8 +642,23 @@ def _extract_and_validate_params(
     # Collect parameters
     found_params = set()
     for param_match in param_matches:
-        param_name = param_match.group(1)
-        param_value = param_match.group(2)
+        raw_param_name = param_match.group(1).strip()
+        raw_param_value = param_match.group(2)
+
+        # Primary parse: <parameter=name>value</parameter>
+        param_name = raw_param_name
+        param_value = raw_param_value
+
+        # Fallback parse for malformed tags some models produce, e.g.:
+        # <parameter=command=str_replace</parameter>
+        # In this case, group(2) is empty and group(1) wrongly contains 'command=str_replace</parameter'!
+        if (not param_value or str(param_value).strip() == '') and ('=' in param_name):
+            # Split once on '=' and clean up any trailing tag fragments
+            name_part, value_part = param_name.split('=', 1)
+            # Remove any trailing tag content e.g. '</parameter' or '>'
+            value_part = value_part.split('</', 1)[0].split('>', 1)[0]
+            param_name = name_part.strip()
+            param_value = value_part.strip()
 
         # Validate parameter is allowed
         if allowed_params and param_name not in allowed_params:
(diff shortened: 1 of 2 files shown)


enyst:
Yes, the code can become cluttered if we keep doing these… At least this one is only a function call in fn_call_converter, maybe it’s localized enough it won’t interfere with readability 🤔

I don’t have a better option for now, maybe we can take it in and see.

Issue #10527, opened 2025-08-20
[Bug]: AWS authentication via ~/.aws/credentials fails in v0.54.0


JAxeAddepar:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

**Description:**
In version 0.54.0, there has been a regression in the AWS authentication mechanism. The application no longer authenticates using the credentials stored in the ~/.aws/credentials or ~/.aws/config files. This functionality was present and working correctly in version 0.28.1.

Additionally, when attempting to use this authentication method, the system incorrectly requires an API key to be set, which should not be necessary when using IAM roles or credential files.

**Steps to Reproduce:**
1. Configure valid AWS credentials in ~/.aws/credentials or ~/.aws/config. (`aws sso login`)
2. Run OpenHands via the Docker as instructed in the ReadMe.md.
  a. Mount AWS credentials using `--volume=${HOME}/.aws:/root/.aws:ro \` in the docker run command.
4. Attempt to configure the agent to use AWS Bedrock as the LLM provider.
5. Leave the API key field blank, expecting the application to use the local AWS credentials.

**Expected Behavior:**
The application should detect and use the credentials from the `~/.aws/` directory to authenticate with AWS services, just as it did in version 0.28.1, without requiring a separate API key to be entered in the UI.

* Note that in 0.28.1 it did require an API Key to be set in the UI, but this could be set to any value.

**Actual Behavior:**
The application fails to authenticate with AWS. If an API Key is missing, it will require this before attempting to log in. If an API Key is present, it will fail to log in.

**Environment:**
- OpenHands Version: 0.54.0 (Regression from 0.28.1)
- Cloud Provider: AWS

### OpenHands Installation

Docker command in README

### OpenHands Version

0.54.0

### Model Name

bedrock/us.anthropic.claude-3-7-sonnet-20250219-v1:0

### Operating System

MacOS

### Logs, Errors, Screenshots, and Additional Context

Clicking "Launch from Scratch" on the home page without an API key set gives toast "Error authenticating with the LLM provider. Please check your API key". It does not however send off a request to the backend.

Logs which occur when sending a request to the LLM via a conversation. This can be accessed when entering a dummy api key:
```
[...]

PR #10573, opened 2025-08-22
fix: resolve empty API keys to None and add Bedrock model support

JAxeAddepar:
### Summary
This PR fixes an issue where empty API keys prevent boto3 from using alternative authentication methods for AWS Bedrock models, and adds proper support for Bedrock model features.

### Problem
The boto3 client, when given an API key of `None`, will try other authentication methods (like IAM roles, environment variables, etc.). However, we were passing empty strings instead of `None`, which prevented boto3 from falling back to these alternative auth methods for Bedrock models.

### Changes Made
- **API Key Handling**: Modified API key validation to set empty/whitespace-only keys to `None` instead of empty `SecretStr` objects, allowing boto3 to use alternative authentication methods
- **Bedrock Model Support**: Added Bedrock Anthropic model patterns to function calling and prompt cache feature lists
- **Authentication Logic**: Updated conversation service to skip API key validation for Bedrock models since they use AWS credentials instead

### Files Modified
- `openhands/cli/settings.py`: Fixed API key handling in both basic and advanced LLM settings to use `None` for empty keys
- `openhands/llm/model_features.py`: Added Bedrock model patterns for function calling and prompt caching
- `openhands/server/services/conversation_service.py`: Added Bedrock model detection and conditional API key validation
- `openhands/storage/data_models/settings.py`: Enhanced API key serialization to handle empty values properly

### Testing
- Verified that empty API keys are set to `None`, allowing boto3 to use alternative auth methods
- Confirmed Bedrock models can authenticate using IAM roles and other AWS credential sources
- Ensured existing non-Bedrock models still require valid API keys

This resolves authentication issues with AWS Bedrock models when using IAM-based authentication instead of explicit API keys.

Addresses bug issue https://github.com/All-Hands-AI/OpenHands/issues/10527

### DIFF
--- a/openhands/llm/model_features.py
+++ b/openhands/llm/model_features.py
@@ -69,6 +69,15 @@ FUNCTION_CALLING_PATTERNS: list[str] = [
     'claude-3-5-haiku*',
     'claude-sonnet-4*',
     'claude-opus-4*',
+    # Bedrock Anthropic families (with bedrock/ prefix)
+    'bedrock/anthropic.claude-3-7-sonnet*',
+    'bedrock/anthropic.claude-3.7-sonnet*',
+    'bedrock/anthropic.claude-sonnet-3-7-latest',
+    'bedrock/anthropic.claude-3-5-sonnet*',
+    'bedrock/anthropic.claude-3.5-haiku*',
+    'bedrock/anthropic.claude-3-5-haiku*',
+    'bedrock/anthropic.claude-sonnet-4*',
+    'bedrock/anthropic.claude-opus-4*',
     # OpenAI families
     'gpt-4o*',
     'gpt-4.1',
@@ -116,6 +125,17 @@ PROMPT_CACHE_PATTERNS: list[str] = [
     'claude-3-opus-20240229',
     'claude-sonnet-4*',
     'claude-opus-4*',
+    # Bedrock Anthropic families (with bedrock/ prefix)
+    'bedrock/anthropic.claude-3-7-sonnet*',
+    'bedrock/anthropic.claude-3.7-sonnet*',
+    'bedrock/anthropic.claude-sonnet-3-7-latest',
+    'bedrock/anthropic.claude-3-5-sonnet*',
+    'bedrock/anthropic.claude-3-5-haiku*',
+    'bedrock/anthropic.claude-3.5-haiku*',
+    'bedrock/anthropic.claude-3-haiku-20240307',
+    'bedrock/anthropic.claude-3-opus-20240229',
+    'bedrock/anthropic.claude-sonnet-4*',
+    'bedrock/anthropic.claude-opus-4*',
 ]
 
 SUPPORTS_STOP_WORDS_FALSE_PATTERNS: list[str] = [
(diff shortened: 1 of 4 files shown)


openhands/llm/model_features.py
```diff
@@ -69,6 +69,15 @@ class ModelFeatures:
...
     'claude-opus-4*',
+    # Bedrock Anthropic families (with bedrock/ prefix)
+    'bedrock/anthropic.claude-3-7-sonnet*',
+    'bedrock/anthropic.claude-3.7-sonnet*',
+    'bedrock/anthropic.claude-sonnet-3-7-latest',
+    'bedrock/anthropic.claude-3-5-sonnet*',
+    'bedrock/anthropic.claude-3.5-haiku*',
+    'bedrock/anthropic.claude-3-5-haiku*',
+    'bedrock/anthropic.claude-sonnet-4*',
+    'bedrock/anthropic.claude-opus-4*',
```
enyst: I think the patterns above should work for bedrock too?

openhands/llm/model_features.py
```diff
@@ -115,6 +124,17 @@ class ModelFeatures:
...
+    'bedrock/anthropic.claude-3-7-sonnet*',
+    'bedrock/anthropic.claude-3.7-sonnet*',
+    'bedrock/anthropic.claude-sonnet-3-7-latest',
+    'bedrock/anthropic.claude-3-5-sonnet*',
+    'bedrock/anthropic.claude-3-5-haiku*',
+    'bedrock/anthropic.claude-3.5-haiku*',
+    'bedrock/anthropic.claude-3-haiku-20240307',
+    'bedrock/anthropic.claude-3-opus-20240229',
+    'bedrock/anthropic.claude-sonnet-4*',
+    'bedrock/anthropic.claude-opus-4*',
```
enyst: Same here

openhands/llm/model_features.py
JAxeAddepar: By this, I guess you mean delete them from the file? I'll do that now 👍  Thanks for the review!

enyst: Yes I think so

enyst: If this pattern matching works for bedrock too, we don’t need them. Thanks!

JAxeAddepar: Excellent, I removed it and tested it and all is working well 😄

JAxeAddepar:
Hi @enyst would you mind reviewing this?

enyst:
@JAxeAddepar Please correct me if wrong, the underlying issue here is [the same as this](https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1753798490354599)?

JAxeAddepar:
> @JAxeAddepar Please correct me if wrong, the underlying issue here is [the same as this](https://openhands-ai.slack.com/archives/C06U8UTKSAD/p1753798490354599)?

Hi @enyst I believe so, yes. Would that be a problem? It was still broken in 0.54.0

tofarr (approved):
LGTM!

enyst:
@openhands Clone the repo agent-sdk from this org. That is OpenHands V1, while this we can say it’s V0.

Now check what this PR is doing. Migrate these changes to agent-sdk and open a PR there, if we can adapt these changes to it. If not, tell me why. But first think deeply.

JAxeAddepar:
Thanks for the reviews @enyst and @tofarr. I can't merge so assume that it is down to one of yourselves to merge? Thanks! 🥳

Issue #9387, opened 2025-06-26
[CLI] Settings changes require restart even when no session is active


neubig:
## Problem
Settings changes require a restart to take effect, even when no session has been started yet. This creates unnecessary friction in the setup process.

## Current Behavior
- User modifies settings via `/settings`
- CLI shows "They will take effect after restart"
- Restart is required even if no agent session is running

## Expected Behavior
- Settings should take effect immediately when no session is active
- Only require restart when an active session would be affected
- Provide immediate feedback when settings are successfully applied

## User Experience Impact
- Adds unnecessary steps to initial setup
- Confusing when no session is running
- Breaks the flow of configuration

## Technical Context
The settings are loaded at startup in `main_with_loop()` and not reloaded during runtime. The configuration could be reloaded when no session is active.

## Proposed Solution
1. **Immediate application**: Apply settings immediately when no session is running
2. **Smart restart detection**: Only require restart when an active session exists
3. **Clear messaging**: Different messages based on whether restart is actually needed

## Related Files
- `openhands/cli/main.py` (settings loading in main_with_loop)
- `openhands/cli/settings.py` (save_settings_confirmation function)
- `openhands/storage/settings/file_settings_store.py`

aeft:
Can I work on this issue? 

First, I want to confirm whether you're asking if a brand new user needs to restart after setup. This creates unnecessary friction in the setup process.

Actually, after setup, users will directly use the settings they just created to start new sessions (such as model configuration and enabling the search MCP server), so there's no need for users to restart manually. Therefore, what needs to be optimized here is the UI prompt. Specifically, after users configure their settings for the first time, the CLI shouldn't display "They will take effect after restart."

If you're asking whether users with existing settings need to restart after updating their settings:

I found that whenever we update the MCP server, the CLI asks the user, "Would you like to restart OpenHands now?" (see `prompt_for_restart`). If the user says yes, it restarts automatically. So, we can follow this implementation—ask the user whether to restart, and if they agree, restart for them.

Another, more complex solution would be to avoid restarting the CLI and instead update the relevant modules dynamically based on changes in the settings (such as the model or search API key). However, this is inconsistent with how changes to the MCP server currently take effect and may cause confusion.

I'd like to hear your suggestions. Thank you!

aeft:
@neubig  Hi, just a quick follow-up. I’ve shared my proposed solutions above and would appreciate any feedback before I open PRs. Thanks!

[...]

PR #10582, opened 2025-08-23
feat(cli): prompt for restart after saving settings

aeft:

---
Keep the logic consistent with MCP's modifications: after saving the settings, prompt the user to restart. If the user agrees, automatically restart for them.

---
**Link of any specific issues this addresses:** https://github.com/All-Hands-AI/OpenHands/issues/9387

### DIFF
--- a/openhands/cli/settings.py
+++ b/openhands/cli/settings.py
@@ -233,6 +233,15 @@ def _get_initial_model_index(
 async def modify_llm_settings_basic(
     config: OpenHandsConfig, settings_store: FileSettingsStore
 ) -> bool:
+    """Modify LLM settings for basic usage.
+
+    Args:
+        config: Configuration to modify.
+        settings_store: Target storage for the settings.
+
+    Returns:
+        True if settings were saved, False otherwise.
+    """
     model_list = get_supported_llm_models(config)
     organized_models = organize_models_and_providers(model_list)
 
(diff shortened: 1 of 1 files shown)


openhands/cli/settings.py
```diff
@@ -616,10 +618,12 @@ async def modify_llm_settings_advanced(
...
     await settings_store.store(settings)
 
+    return True
+
 
 async def modify_search_api_settings(
     config: OpenHandsConfig, settings_store: FileSettingsStore
-) -> None:
+) -> bool:
     """Modify search API settings."""
```
tofarr: Maybe add a note here on what the returned flag means? (True if value was updated, false if existing setting retained)

openhands/cli/settings.py
```diff
@@ -232,7 +232,7 @@ def _get_initial_model_index(
 
 async def modify_llm_settings_basic(
     config: OpenHandsConfig, settings_store: FileSettingsStore
-) -> None:
+) -> bool:
```
tofarr, line 235: Maybe add a note here on what the returned flag means (True if value was updated, false if existing setting retained) and what is involved in basic settings

openhands/cli/settings.py
```diff
@@ -486,10 +486,12 @@ def model_validator(x):
 
     await settings_store.store(settings)
 
+    return True
+
 
 async def modify_llm_settings_advanced(
     config: OpenHandsConfig, settings_store: FileSettingsStore
-) -> None:
+) -> bool:
```
tofarr, line 503: Maybe add a note here on what the returned flag means (True if value was updated, false if existing setting retained) and what is involved in advanced settings

aeft:
@tofarr  Thanks for the review! I have added docstring for each function.

aeft:
@tofarr Hi, can you have a look when you have time? Thank you!

mamoodi:
Thanks so much for the PR here. I'm going to close this because the v0 CLI has been deprecated and code changes should go into openhands-cli folder.

PR #10583, opened 2025-08-23
Fix GCP storage errors by disabling cache reading in EventStore

neubig:
## Problem

EventStore was attempting to read cache files that it cannot create, leading to storage errors when cache files don't exist.

Fixes: https://github.com/All-Hands-AI/infra/issues/710

## Root Cause

The issue occurs when EventStore tries to read cache files from paths like `event_cache/150-175.json`, but these cache files are only created by EventStream when events are added. For conversations without existing cache files, this results in "No such object" errors.

## Solution

This PR implements cache control to prevent EventStore from attempting to read cache files it cannot create:

### Changes Made

1. **Added `use_cache` flag to EventStore:**
   - Default: `False` (EventStore doesn't write cache files, so shouldn't read them)
   - EventStore skips cache reading entirely when `use_cache=False`

2. **Updated EventStream to enable cache:**
   - EventStream constructor sets `use_cache=True` since it can write cache files

3. **Modified cache loading logic:**
   - When `use_cache=False`, `_load_cache_page_for_index()` returns a dummy page
   - Forces fallback to reading individual event files
   - No storage requests made for non-existent cache files

## Benefits

- ✅ Eliminates storage errors for missing cache files
- ✅ Maintains functionality via individual event file fallback
- ✅ Preserves performance for EventStream cache usage
- ✅ Minimal, targeted change with no breaking changes

## Testing

Comprehensive tests verify:
- EventStore defaults to `use_cache=False` and skips cache reading
- EventStream defaults to `use_cache=True` and maintains cache functionality  
- Event search operations work correctly in both cases
- Graceful handling of missing cache files

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:e94ca25-nikolaik   --name openhands-app-e94ca25   docker.openhands.dev/openhands/openhands:e94ca25
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@fix/eventstore-cache-gcp-errors#subdirectory=openhands-cli openhands
```


neubig:
Thanks @tofarr ! But I think that `EventStore` will not write cache files, so the cache will never hit, unless I'm mistaken?

PR #10588, opened 2025-08-24
Implement branch pagination for repository selection and improve UI async dropdown behaviour

### LINKED ISSUE #10608: [Bug]: Loading branches takes very long

When a repository has a lot of branches, it takes very long to load them (especially for BitBucket). We should perform pagination instead, and do server side search for specific branches instead of doing it in-memory. 

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Model Name

_No response_

### Operating System

None

_No response_

malhotra5:

This change improves the performance and user experience when working with repositories that have many branches. Previously, OpenHands would load all branches at once, which could be slow for large repositories. Now branches are loaded in pages as needed, making the interface more responsive. Additionally, the system now correctly identifies and selects the repository's actual default branch instead of guessing based on common names like "main" or "master".

---

This PR implements comprehensive branch pagination across both frontend and backend:

**Backend Changes:**
- Added pagination parameters (`page`, `per_page`) to `/api/user/repository/branches` endpoint
- Updated `ProviderHandler.get_branches()` to support pagination
- Implemented `get_paginated_branches()` method in all git services (GitHub, GitLab, Bitbucket)
- Enhanced repository parsing to extract `main_branch`/`default_branch` from API responses
- Removed default branch sorting logic from backend (now handled in frontend)

**Frontend Changes:**
- Updated `getRepositoryBranches()` API client to accept pagination parameters
- Added `useRepositoryBranchesPaginated()` hook with React Query infinite query support
- Enhanced `GitRepository` type to include `main_branch` field
- Improved default branch selection in repo selection form to prioritize repository's actual default branch
- Maintained backward compatibility with existing branch loading logic

**Key Design Decisions:**
1. **Infinite Query Pattern**: Used React Query's infinite query for seamless pagination UX
2. **Backward Compatibility**: Maintained existing `useRepositoryBranches()` hook for components that don't need pagination
3. **Default Branch Priority**: Repository's `main_branch` field takes precedence over fallback detection
4. **Consistent Pagination**: All git services use the same pagination interface (page, per_page)
5. **Performance**: Default page size of 30 branches balances performance with UX

The implementation includes comprehensive test coverage and maintains full backward compatibility.

---

This addresses performance issues with repositories containing hundreds of branches and improves accuracy of default branch detection.

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/c1e3c8a7e46e41ee84604a2fa339b9df)

Fixes #10608 #10625

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:aac1865-nikolaik   --name openhands-app-aac1865   docker.all-hands.dev/all-hands-ai/openhands:aac1865
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@feature/implement-branch-pagination openhands
```


malhotra5:
@openhands read the diff of this branch against main

make sure to revert all the changes to the microagent management page from this PR; DO NOT REMOVE OR REVERT ANY OTHER CODE

malhotra5:
@openhands read the diff of this pr against main

then do the following things

1. in the BE, add the ability to search for a branch name given the provider and repo; look at the existing pattern for calling provider specific apis in the backend
2. in the FE, if the user types in the branch selection dropdown then use the search apis to filter the results (look at the repo dropdown from reference)

malhotra5:
@openhands search for branches isn't working as expected

look at the existing pattern for executing search in the FE from repo dropdown component. This include debouncing, etc. Then perform the exact same pattern for the branch dropdown. The dropdown already supports pagination, but fix the part where we perform search once the user starts typing.

openhands/integrations/github/github_service.py
```diff
@@ -615,6 +617,82 @@ async def get_branches(self, repository: str) -> list[Branch]:
...
+
+        return PaginatedBranchesResponse(
+            branches=branches,
+            has_next_page=has_next_page,
+            current_page=page,
+            per_page=per_page,
+            total_count=None,  # GitHub doesn't provide total count in branch API
+        )
+
+    async def search_branches(
```
malhotra5, line 667: @openhands the search branches implementation for github provider is incorrect; we should be passing the user's partial query as a parameter to the github api when searching for the branch

github may not support a dedicated endpoint for this as per `https://docs.github.com/en/rest/branches/branches?apiVersion=2022-11-28`

1. maybe check if there is a graphql endpoint instead
2. write the graphql query inside `openhands/integrations/github/queries.py`
3. rewrite the search endpoint with the graphql query (check the existing pattern for calling graphql queries)
4. write a test script (but do not commit this script), where you use the $GITHUB_TOKEN in your env to call the graphql query to see that its working as expected

malhotra5:
@openhands we need to tweak the FE so that its very user friendly, please pay attention to the following details without breaking existing functionality for the branch dropdown

1. when the user types a value, we should query for the search endpoint and display its results; the user removes the text from the dropdown, then default back to the regular branch listing

2. make sure that the default branch is displayed first in the dropdown (not when the user is searching for a branch)

malhotra5:
@openhands rename the `main_branch` parameter that is part of the repository object in both the FE and BE

make it `default_branch` instead, and make the param non-optional

openhands/integrations/provider.py
```diff
@@ -468,31 +499,19 @@ async def get_branches(
         for provider in self.provider_tokens:
             try:
                 service = self._get_service(provider)
-                branches = await service.get_branches(repository)
```
raymyers, line 471: It looks like we're trying to get branches from each provider rather than actually knowing what provider a repo came from. This will mask failure. It seems like we should know the provider already, I assume there is some limitation that prevents this beyond the scope of this PR.

raymyers (approved):
I'm not fully able to evaluate the moved files in frontend/src/components/common/*. You can evaluate if someone else should look at them.

openhands/integrations/bitbucket/bitbucket_service.py
```diff
@@ -551,6 +552,83 @@ async def get_branches(self, repository: str) -> list[Branch]:
 
         return branches
 
+    async def get_paginated_branches(
```
raymyers, line 555: Would it be a good idea for these get_paginated_branches and search_branches functions to have direct unit tests?

openhands/integrations/provider.py
malhotra5: Yeah I plan to refactor this next week; parts of our codebase in general don't require the provider information to be present, when it should.

amanape:
@raymyers They were removed since I've rebuilt the components to use a library called `downshift` instead of `react-select` for better async handling. A little more information on why this was done in #10625

Hope that clears any doubt

PR #10590, opened 2025-08-24
Refactor conversation status

tofarr:

Added 2 new conversation states to give parity with runtime API:
* `ARCHIVED`: Conversation has stopped and cannot be restarted.
* `ERROR`: Huston we have a problem!

---
The runtime API can get into a state where a conversation did exist but data related to it was deleted, so the conversation cannot be restarted. The OpenHands API needs to reflect this. (So old conversations do not yield such confusion)

---

### DIFF
--- a/openhands/storage/data_models/conversation_status.py
+++ b/openhands/storage/data_models/conversation_status.py
@@ -1,6 +1,6 @@
 """
 This class is similar to the RuntimeStatus defined in the runtime api. (When this class was defined
-a `RuntimeStatus` class already existed in OpenHands which serves a completely different purpose) Some of
+a RuntimeStatus class already existed in OpenHands which serves a completely different purpose) Some of
 the status definitions do not match up:
 
 STOPPED/paused - the runtime is not running but may be restarted
@@ -11,7 +11,7 @@ from enum import Enum
 
 
 class ConversationStatus(Enum):
-    # The runtime is starting
+    # The conversation is starting
     STARTING = 'STARTING'
     # The conversation is running - the agent may be working or idle
     RUNNING = 'RUNNING'
@@ -19,5 +19,5 @@ class ConversationStatus(Enum):
     STOPPED = 'STOPPED'
     # The conversation has been archived and cannot be restarted.
     ARCHIVED = 'ARCHIVED'
-    # Something has gone wrong with the conversation
+    # Something has gone wrong with the conversation (The runtime rather than the agent)
     ERROR = 'ERROR'
--- a/openhands/storage/data_models/conversation_status.py
+++ b/openhands/storage/data_models/conversation_status.py
@@ -1,15 +1,13 @@
-from enum import Enum
+"""
+This class is similar to the RuntimeStatus defined in the runtime api. (When this class was defined
+a `RuntimeStatus` class already existed in OpenHands which serves a completely different purpose) Some of
+the status definitions do not match up:
 
-# TODO: I think this class should be deprecated and replaced with `RuntimeStatus` - the values for which are
-# defined in the runtime api:
-# * running: currently RUNNING
-# * stopped currently ARCHIVED
-# * paused: currently STOPPED
-# * error: currently ERROR
-# * starting: currently STARTING
-#
-# Unifying these two would simplify the codebase - particularly related to what `STOPPED` actually means.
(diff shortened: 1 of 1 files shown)


openhands/storage/data_models/conversation_status.py
```diff
@@ -1,7 +1,23 @@
...
 class ConversationStatus(Enum):
+    # The conversation is starting
     STARTING = 'STARTING'
+    # The conversation is running - the agent may be working or idle
     RUNNING = 'RUNNING'
+    # The conversation has stopped (This is synonymous with `paused` in the runtime API.)
     STOPPED = 'STOPPED'
+    # The conversation has been archived and cannot be restarted.
+    ARCHIVED = 'ARCHIVED'
+    # Something has gone wrong with the conversation (The runtime rather than the agent)
```
enyst, line 22: What does ARCHIVED actually mean, what could have happened?

I've had a case last week where the conversation didn't find the old history to restore, and the Saas made a new history instead. The old events were still accessible via the API, but FE got an empty new stream instead. Is that the kind of thing that will be "archived" or "error", or is this about just old runtimes?

tofarr: After a certain amount of inactivity (14 days decided by an environment variable), everything related to a runtime gets removed in the runtime-api, including it's persistent volume claim - all files in the workspace are effectively gone. This can also be done by calling `/stop` on the runtime-api. Currently, the runtime will start again if you call `/start`, but all old context is effectively gone. (though we still keep the event stream)

I think we need to rethink the behaviour of the runtime API, but OpenHands should be able to deal with the concept of a runtime that is no longer easily restartable or `ARCHIVED`.

The `error` state is copied over from the runtime API, even though we do not currently use it - I imagine it will become very important if we ever do try to rebuild runtimes using the event stream.

enyst: > This can also be done by calling /stop on the runtime-api. Currently, the runtime will start again if you call /start, but all old context is effectively gone

This sounds a bit confusing 🫤

Is this the same with "Stop" item from FE contextual menu on the conversation panel > conversation > r-click?

It cannot be, right?

tofarr: This is definitely confusing! (That is why a large part of this is documentation!)

The "Stop" in the context menu is not the same as the runtime api - it effectively translates to a "/pause" in the runtime API.

Being able to restart but not restoring files is also weird (I have an open PR in the runtime API to have it raise an error instead).

enyst: Maybe we should rename that menu item to Pause. It should help, to say what it means.

I agree, restart but not restore files sounds like it needs to have a different `conversation_id` then. It's really been surprising to hear my secretary (OpenHands agent with Sonnet-4) describe the situation above:
[slack link to fun synopsis](https://openhands-ai.slack.com/archives/C07S376KASK/p1755803274803119?thread_ts=1755802279.610869&cid=C07S376KASK)

tofarr: @OpenHands Please rename the menu item from "Stop" to "Pause"

frontend/src/i18n/translation.json
```diff
@@ -5391,8 +5407,40 @@
...
+    "es": "¿Está seguro de que desea pausar esta conversación?",
+    "ar": "هل أنت متأكد أنك تريد إيقاف هذه المحادثة مؤقتًا؟",
+    "fr": "Êtes-vous sûr de vouloir mettre cette conversation en pause ?",
+    "tr": "Bu konuşmayı duraklatmak istediğinizden emin misiniz?",
+    "de": "Sind Sie sicher, dass Sie dieses Gespräch pausieren möchten?",
+    "uk": "Ви впевнені, що хочете призупинити цю розмову?"
+  },
   "CONVERSATION$STOP_WARNING": {
-    "en": "Are you sure you want to stop this conversation?",
+    "en": "Are you sure you want to pause this conversation?",
```
enyst, line 5443:
```suggestion
    "en": "Are you sure you want to stop this conversation?",
```

I suggest this only for minor consistency. But I’m not sure if STOP_WARNING is / will be used anymore, maybe we could even remove it?

tofarr: I'm adding an additional PR to remove unused keys.

PR #10600, opened 2025-08-25
feat: hide conversations after PR closure or merge (microagent management)

hieptl:

Conversations should only be displayed if they are associated with OPEN or DRAFT PRs for new microagents.

If a conversation has resulted in a CLOSED or MERGED PR, it should no longer be shown.

**Acceptance Criteria:**
- Only conversations linked to OPEN or DRAFT PRs are visible.
- Conversations with CLOSED or MERGED PRs are hidden.

---

The PR updates the search conversations API to hide conversations after PR closure or merge (microagent management)

https://github.com/user-attachments/assets/6df73471-f5fb-4289-ae88-2a2b88b65b07

---

### DIFF
--- a/openhands/integrations/provider.py
+++ b/openhands/integrations/provider.py
@@ -626,3 +626,218 @@ class ProviderHandler:
             remote_url = f'https://{domain}/{repo_name}.git'
 
         return remote_url
+
+    async def get_pr_details(
+        self,
+        repository: str,
+        pr_number: int,
+        specified_provider: ProviderType | None = None,
+    ) -> dict:
+        """Get detailed information about a specific pull request/merge request
+
+        Args:
+            repository: Repository name in format 'owner/repo'
+            pr_number: The pull request/merge request number
+            specified_provider: Optional provider type to use
+
+        Returns:
+            Raw API response from the git provider
+
+        Raises:
+            AuthenticationError: If authentication fails
+        """
+        if specified_provider:
+            return await self._get_pr_details_from_specified_provider(
+                repository, pr_number, specified_provider
+            )
+
+        return await self._get_pr_details_from_all_providers(repository, pr_number)
+
+    async def _get_pr_details_from_specified_provider(
+        self, repository: str, pr_number: int, specified_provider: ProviderType
+    ) -> dict:
+        """Get PR details from a specified provider.
+
+        Args:
+            repository: Repository name in format 'owner/repo'
+            pr_number: The pull request/merge request number
+            specified_provider: The specific provider type to use
+
+        Returns:
+            Raw API response from the specified git provider
+
+        Raises:
+            AuthenticationError: If authentication fails
+        """
+        try:
+            service = self._get_service(specified_provider)
+            return await service.get_pr_details(repository, pr_number)
+        except Exception as e:
+            logger.warning(f'Error fetching PR details from {specified_provider}: {e}')
+            raise
+
+    async def _get_pr_details_from_all_providers(
+        self, repository: str, pr_number: int
+    ) -> dict:
+        """Get PR details by trying all available providers in order.
+
+        Args:
+            repository: Repository name in format 'owner/repo'
+            pr_number: The pull request/merge request number
+
+        Returns:
+            Raw API response from the first successful git provider
+
+        Raises:
+            AuthenticationError: If all providers fail
+        """
+        # Try all available providers in order
+        errors = []
+        for provider in self.provider_tokens:
+            try:
+                service = self._get_service(provider)
+                result = await service.get_pr_details(repository, pr_number)
+                if result:
+                    return result
+            except Exception as e:
+                errors.append(f'{provider.value}: {str(e)}')
+                logger.warning(
+                    f'Error fetching PR details from {provider} for {repository}#{pr_number}: {e}'
+                )
+
+        # If all providers failed, raise an error
+        if errors:
+            logger.error(
+                f'Failed to fetch PR details for {repository}#{pr_number} with all available providers. Errors: {"; ".join(errors)}'
+            )
+            raise AuthenticationError(
+                f'Unable to fetch PR details for {repository}#{pr_number}'
+            )
+
+        raise AuthenticationError(f'PR #{pr_number} not found in {repository}')
+
+    async def _is_pr_active(
+        self, repository: str, pr_number: int, git_provider: ProviderType
+    ) -> bool:
+        """Check if a PR is still active (not closed/merged).
+
+        This method checks the PR status using provider-specific logic based on
+        the actual API response structures from GitHub, GitLab, and Bitbucket.
+
+        Args:
+            repository: Repository name in format 'owner/repo'
+            pr_number: The PR number to check
+            git_provider: The Git provider type for this repository
+
+        Returns:
+            True if PR is active (open), False if closed/merged, True if can't determine
+        """
+        try:
+            pr_details = await self.get_pr_details(repository, pr_number, git_provider)
+
+            # Provider-specific logic based on actual API response structures
+            if git_provider == ProviderType.GITHUB:
+                return self._check_github_pr_status(pr_details, repository, pr_number)
+            elif git_provider == ProviderType.GITLAB:
+                return self._check_gitlab_pr_status(pr_details, repository, pr_number)
+            elif git_provider == ProviderType.BITBUCKET:
+                return self._check_bitbucket_pr_status(
+                    pr_details, repository, pr_number
+                )
+
+            # If we can't determine the state, assume it's active (safer default)
+            logger.warning(
+                f'Could not determine PR status for {repository}#{pr_number} on {git_provider}. '
+                f'Response keys: {list(pr_details.keys())}. Assuming PR is active.'
+            )
+            return True
+
+        except Exception as e:
+            logger.warning(
+                f'Could not determine PR status for {repository}#{pr_number}: {e}. '
+                f'Including conversation to be safe.'
+            )
+            # If we can't determine the PR status, include the conversation to be safe
+            return True
+
+    def _check_github_pr_status(
+        self, pr_details: dict, repository: str, pr_number: int
+    ) -> bool:
+        """Check GitHub PR status based on API response structure.
+
+        Args:
+            pr_details: Raw GitHub API response for the PR
+            repository: Repository name for logging
+            pr_number: PR number for logging
+
+        Returns:
+            True if PR is active (open), False if closed/merged
+        """
+        # GitHub API response structure
+        # https://docs.github.com/en/rest/pulls/pulls#get-a-pull-request
+        if 'state' in pr_details:
+            return pr_details['state'] == 'open'
+        elif 'merged' in pr_details and 'closed_at' in pr_details:
+            # Check if PR is merged or closed
+            return not (pr_details['merged'] or pr_details['closed_at'])
+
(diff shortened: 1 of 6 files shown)


openhands/integrations/provider.py
```diff
@@ -626,3 +626,218 @@ async def get_authenticated_git_url(self, repo_name: str) -> str:
...
+
+        except Exception as e:
+            logger.warning(
+                f'Could not determine PR status for {repository}#{pr_number}: {e}. '
+                f'Including conversation to be safe.'
+            )
+            # If we can't determine the PR status, include the conversation to be safe
+            return True
+
+    def _check_github_pr_status(
```
enyst:
```suggestion
    def _is_github_pr_open(
```
Nit: these methods return true/false, with the meaning: whether it’s open, as best as we can tell

Just thinking that it’s easier to read if its name matches what it does

hieptl: Hi @enyst, I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/server/routes/manage_conversations.py
```diff
@@ -261,6 +262,23 @@ async def search_conversations(
         ):
             continue
 
+        # Special filtering for microagent_management conversations
+        if (
+            conversation_trigger == ConversationTrigger.MICROAGENT_MANAGEMENT
+            and conversation.pr_number
+            and conversation.selected_repository
+            and conversation.git_provider
```
enyst: Could we find a different solution for the ‘hiding in the FE’ behavior?

This endpoint is for general use, and its filters can be sent over by the user; I think this “filter” shouldn’t be hardcoded either (the others are not). I mean, imagine I use this endpoint via API (so without FE itself). And I try to get all conversations: in this PR, the endpoint will never give me some conversations.

That is rather unexpected. If I didn’t specify a filter, it should send all?
(maybe limited)

enyst (dismissed):
I like where this is heading, UX-wise. It’s nice when it feels like the software adapts to what the user needed.

Comment inline: could we not change the API definition though, with a hardcoded ‘hiding’? Maybe we can do this differently, either filter from outside (e.g. from FE), or create a filter that makes sense for any other API use?

openhands/server/routes/manage_conversations.py
hieptl: Hi @enyst, when I call this endpoint directly via the API (without the frontend) and omit the filter, it still works as expected.

For more details, please refer to this video. Could you clarify your feedback further?

https://github.com/user-attachments/assets/41dacdfa-b43e-4b08-9c7e-55fe5066117d

hieptl: I am not hard-coding the `conversation_trigger` parameter. You may pass a `conversation_trigger` filter such as `gui` or `microagent_management` to filter conversations. If no filter is provided, the API will return all conversations as expected.

The logic you highlighted is specifically used to check whether the `conversation_trigger` is `microagent_management`. In that case, we call the Git provider APIs to verify whether the associated PRs are closed or merged. Otherwise, redundant Git provider API calls are avoided.

For more details, please refer to the video below.

https://github.com/user-attachments/assets/4f202f36-f246-4b11-907d-ae041c3ad142

enyst: Ah, I see. But it means that the `search` endpoint has different behavior if the trigger is `MicroagentManagement`, as opposed to any other trigger, which is used in the filtering mechanics of `/search`?

I would love it if we can make the API behave in standard, expected ways. I still think the place of microagent page specific logic is somewhere linked to the microagent page (maybe in FE?), rather than on the endpoint, but I'd be happy to be proven wrong. Maybe I'm misreading?

enyst: I mean, it seems to me that the behavior implemented here (return if open, skip if closed), is a behavior specific to a particular page on a particular UI: microagent management page on web FE.

The rest of the filters/behavior are not implementing logic specific to only one place, and I think that's expected behavior... 🤔 Could we perhaps do the _hiding_ in the caller, in FE, or is there a reason not to? Sorry for the nitpick!

hieptl: On the microagent management page, creating a new microagent requires launching a new conversation (we can refer to the attached video in the PR description).

To fetch these conversations, we need to call an API. Previously, I defined a separate API for this, but during review Rohit suggested reusing the existing one. To enable that, I introduced filtering by conversation_trigger, which allows us to distinguish conversations created from the microagent management page from those created elsewhere.

In this PR, we also need to identify which microagent management conversations are linked to merged or closed PRs and exclude them from the API response. This way, the frontend will not display them, improving the UX. Since checking PR status requires calling the Git provider APIs, I believe this logic should reside in the backend rather than the frontend.

If you have another solution in mind, please let me know. Thank you! 🙏

enyst: Ah, I see, so we don't return to the FE whether they were open or closed. 🤔 Let me take another look! (sorry for the nit)

enyst: Ah I see, my preference would have been to use this endpoint, _but do the last bit_ of filtering on the client. It seems we kinda can't in this case (we don't send over open/closed). 🤔 

OK, I took another look at the entire manage_conversations endpoints, and it seems both ways are sort of reasonable:
- mix a small bit of FE-specific logic in `conversations`
- keep `/conversations` standard (let it do what a GET does: generic GET, without specific behavior)

I prefer the latter, personally. So if that implies a specific, new, endpoint, well...

Cc: @malhotra5 @hieptl  Can you be convinced to go for another endpoint here? (maybe even another route file, in the future)

Really sorry for the detail, admittedly it's a small detail.

hieptl: @enyst Yes, I’m happy to create a separate endpoint for fetching the conversations needed on the microagent management page. Thank you! 🙏

malhotra5: Yeah I'm good making it a separate endpoint

malhotra5: Yeah just took a closer generally we want to avoid mixing in the provider handler, etc with the generic search endpoint

hieptl: Hi @enyst, @malhotra5 I've updated the code based on your feedback. Please help me review the PR again. Thank you for your feedback! 🙏

frontend/src/api/open-hands.ts
```diff
@@ -726,6 +726,28 @@ class OpenHands {
...
+    }
+
+    if (selectedRepository) {
+      params.append("selected_repository", selectedRepository);
+    }
+
+    const { data } = await openHands.get<ResultSet<Conversation>>(
+      `/api/microagent-management/conversations?${params.toString()}`,
+    );
+    return data.results;
```
amanape, line 748: Instead of passing params as a string literal, axios supports the `param` property: https://axios-http.com/docs/req_config

frontend/src/hooks/query/use-microagent-management-conversations.ts
```diff
@@ -0,0 +1,27 @@
...
+      limit,
+      selectedRepository,
+    ],
+    queryFn: () =>
+      OpenHands.getMicroagentManagementConversations(
+        pageId,
+        selectedRepository,
+        limit,
+      ),
+    enabled: true, // Always enabled since parameters are optional
```
amanape:
```suggestion
```

Already enabled by default

frontend/src/api/open-hands.ts
hieptl: I've updated the code based on your feedback. Thank you for your feedback! 🙏

frontend/src/hooks/query/use-microagent-management-conversations.ts
hieptl: I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/integrations/provider.py
```diff
@@ -613,3 +613,216 @@ async def get_authenticated_git_url(self, repo_name: str) -> str:
             remote_url = f'https://{domain}/{repo_name}.git'
 
         return remote_url
+
+    async def get_pr_details(
+        self,
+        repository: str,
+        pr_number: int,
+        specified_provider: ProviderType | None = None,
```
malhotra5: Is there ever a time where we don't know the provider in advance?

openhands/integrations/provider.py
```diff
@@ -613,3 +613,216 @@ async def get_authenticated_git_url(self, repo_name: str) -> str:
...
+        elif 'merged' in pr_details and 'closed_at' in pr_details:
+            # Check if PR is merged or closed
+            return not (pr_details['merged'] or pr_details['closed_at'])
+
+        # If we can't determine the state, assume it's active (safer default)
+        logger.warning(
+            f'Could not determine GitHub PR status for {repository}#{pr_number}. '
+            f'Response keys: {list(pr_details.keys())}. Assuming PR is active.'
+        )
+        return True
```
malhotra5: Could we move this logic into the service class, same for the providers

openhands/integrations/provider.py
```diff
@@ -613,3 +613,53 @@ async def get_authenticated_git_url(self, repo_name: str) -> str:
...
+
+        Raises:
+            AuthenticationError: If authentication fails
+        """
+        try:
+            service = self._get_service(specified_provider)
+            return await service.get_pr_details(repository, pr_number)
+        except Exception as e:
+            logger.warning(f'Error fetching PR details from {specified_provider}: {e}')
+            raise
```
malhotra5:
```suggestion
```

Seems unused?

openhands/server/routes/manage_conversations.py
```diff
@@ -79,6 +79,83 @@
 app = APIRouter(prefix='/api', dependencies=get_dependencies())
 
 
+def _filter_conversations_by_age(conversations: list, max_age_seconds: int) -> list:
```
malhotra5:
```suggestion
def _filter_conversations_by_age(conversations: list[ConversationMetadata], max_age_seconds: int) -> list:
```

could you confirm the typing here

openhands/server/routes/manage_conversations.py
```diff
@@ -725,3 +767,68 @@ def add_experiment_config_for_conversation(
         return True
 
     return False
+
+
+@app.get('/microagent-management/conversations')
+async def get_microagent_management_conversations(
+    page_id: str | None = None,
+    limit: int = 20,
+    selected_repository: str | None = None,
```
malhotra5:
```suggestion
    selected_repository: str,
```

should this ever be None? it seems like we never return any conversations if it is, maybe the FE should disable the query if a repo isn't selected

malhotra5:
PR LGTM, I'll leave the final approval to @amanape to verify the FE portion of this PR

openhands/integrations/provider.py
hieptl: I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/server/routes/manage_conversations.py
hieptl: I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/server/routes/manage_conversations.py
hieptl: I've updated the code based on your feedback. Thank you for your feedback! 🙏

PR #10603, opened 2025-08-25
fix: Initialise with LLM configuration from environment variables

JAxeAddepar:
## Summary

This PR fixes an issue where users with no stored settings would receive `None` instead of default configuration values, and improves the settings merging logic to properly handle default values from config.toml.

## Problem

The user authentication system was returning `None` when no stored settings existed, instead of providing sensible defaults from the config.toml file. Additionally, the settings merging logic was overly aggressive, replacing stored user preferences with config.toml values even when the user had explicitly set different values.

## Changes Made

**Default Settings Creation**: Modified user authentication to create default settings from config.toml when no stored settings exist, ensuring users always receive a valid configuration object
**Settings Merging Logic**: Updated the merge logic to use config.toml settings as defaults only when stored settings are `None`, preserving user preferences while providing fallbacks
**Selective Field Merging**: Added explicit handling for core LLM fields (`llm_model`, `llm_api_key`, `llm_base_url`) to only use config defaults when stored values are `None`
**MCP Config Handling**: Maintained special merging behavior for MCP configuration while fixing the overall merging strategy

## Files Modified

**openhands/server/user_auth/default_user_auth.py**: Added logic to create default settings from config.toml when no stored settings exist, ensuring users always receive a valid configuration
**openhands/storage/data_models/settings.py**: Enhanced settings merging to use config.toml as defaults only when stored values are `None`, and added selective field merging for core LLM settings
**tests/unit/mcp/test_mcp_integration.py**: Updated test to reflect new behavior where default settings are created from config.toml instead of returning `None`

## Testing

- Verified that users with no stored settings receive default configuration from config.toml
- Confirmed that existing user preferences are preserved and not overwritten by config.toml values
- Ensured MCP configuration merging continues to work as expected
- Updated unit tests to reflect the new expected behavior

This resolves configuration issues for new users and improves the overall user experience by ensuring consistent default values while respecting user customizations.

---

---

### DIFF
--- a/openhands/storage/data_models/settings.py
+++ b/openhands/storage/data_models/settings.py
@@ -141,12 +141,26 @@ class Settings(BaseModel):
     def merge_with_config_settings(self) -> 'Settings':
         """Merge config.toml settings with stored settings.
 
-        Config.toml takes priority for MCP settings, but they are merged rather than replaced.
+        Config.toml settings are used as defaults only when stored settings are None.
+        For MCP settings, they are merged rather than replaced.
         This method can be used by both server mode and CLI mode.
         """
         # Get config.toml settings
         config_settings = Settings.from_config()
-        if not config_settings or not config_settings.mcp_config:
+        if not config_settings:
+            return self
+
+        # Use config.toml settings as defaults only when stored settings are None
+        mergeable_fields = ['llm_model', 'llm_api_key', 'llm_base_url']
+
+        for field_name in mergeable_fields:
+            stored_value = getattr(self, field_name)
+            config_value = getattr(config_settings, field_name)
+            if stored_value is None and config_value is not None:
+                setattr(self, field_name, config_value)
+
+        # Handle MCP config merging (special case - merge rather than replace)
+        if not config_settings.mcp_config:
             return self
 
         # If stored settings don't have MCP config, use config.toml MCP config
--- a/openhands/server/user_auth/default_user_auth.py
+++ b/openhands/server/user_auth/default_user_auth.py
@@ -53,7 +53,12 @@ class DefaultUserAuth(UserAuth):
         settings_store = await self.get_user_settings_store()
         settings = await settings_store.load()
 
-        # Merge config.toml settings with stored settings
+        # If no stored settings exist, create default settings from config.toml
+        if not settings:
+            from openhands.storage.data_models.settings import Settings
+            settings = Settings.from_config()
+
+        # Merge config.toml settings with stored settings (if any exist)
         if settings:
             settings = settings.merge_with_config_settings()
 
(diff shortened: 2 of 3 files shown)


JAxeAddepar:
Hi @enyst would you mind reviewing this?

openhands/storage/data_models/settings.py
```diff
@@ -153,12 +153,26 @@ def from_config() -> Settings | None:
...
+        if not config_settings:
+            return self
+
+        # Use config.toml settings as defaults only when stored settings are None
+        mergeable_fields = ['llm_model', 'llm_api_key', 'llm_base_url']
+
+        for field_name in mergeable_fields:
+            stored_value = getattr(self, field_name)
+            config_value = getattr(config_settings, field_name)
+            if stored_value is None and config_value is not None:
```
enyst, line 171: If I see this right, the priority here would be `settings > `config.toml`?

Just thinking, also, doesn’t the FE force the entry of `settings` if None?

JAxeAddepar: Yep, settings takes priority. It's basically adding initialisation variables. 

Pseudo code:
If user has changed values in FE, use new values.
Else If there are initialisation variables (config.toml), use these default values
Else, no values, for the entry of `settings`

enyst: I see where you’re coming from, and I used to think this is how it should be, but then I looked from a different angle: I think we need to take into account the CLI experience, which has a VSCode-like behavior / expectations:
- the user can set `settings` in the TUI (saved in `~/.openhands`)
- they installed from pypi and ran by just typing `openhands` in a terminal
- => if there is a `config.toml` in their current directory, it’s a toml in the workspace (the workspace is the current directory in this main CLI scenario)

Which means: user-level settings should have lower priority than workspace-level settings.

Now, with the web GUI, there is no concept of workspace-level settings, and I think we don’t want one; but with the CLI, we needed a way to override user-level, and this behavior is a vscode-like behavior (and other CLIs? I think). Idk, if we look in isolation, it makes some sense? Within limits… we shouldn’t want to write sensitive files in the workspace. 🤔

Sorry for the mess. If you’d like to think through it, please feel free to take a look at this:
- https://github.com/All-Hands-AI/OpenHands/issues/9585

JAxeAddepar: Hi Enyst, thanks for the feedback. I can't fully wrap my head around it, as instinctively the below line sounds wrong!

> Which means: user-level settings should have lower priority than workspace-level settings.

But I'll leave it to yourselves to prioritise features such as these. Please do consider however that this functionality is much wanted!

Thanks for your time and work 😄

enyst: Yes, I really want to get this sorted out! We are simplifying the configuration system for OpenHands next, and we’ll knock it down. From the web server perspective, it feels like we need just this, the visual override of some “defaults” - or those “defaults” accessible in the GUI.

Thank you!

JAxeAddepar: This sounds right on the money and I'm glad that it's coming. Looking forward to all the future developments and, of course, I'll help with some bug squashing along the way. Thank you 😄

PR #10607, opened 2025-08-25
fix: Extend cases where truncation triggers

csmith49:

---

This PR adds a few new cases to the agent controller logic that catch context window errors from different providers.

---


csmith49:
> Just a thought, maybe in some future refactoring we can create an exception of our own for all these strings, and do the filtering there?

That would be ideal, I think. Logic is probably the same but at least the agent controller will be cleaner!

PR #10637, opened 2025-08-26
[Fix]: token refresh for nested runtimes

malhotra5:

- Fix tokens expiring in the middle of a running conversation

---

We need to contact the main server from the runtime pod for the refreshed tokens; the nested runtime pod also doesn't contain a user_id, the main server will figure it out from the session api key

---
Fixes ALL-2989
https://github.com/All-Hands-AI/OpenHands-Cloud/issues/131

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:157f876-nikolaik   --name openhands-app-157f876   docker.all-hands.dev/all-hands-ai/openhands:157f876
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@refresh-token openhands
```

### DIFF
--- a/openhands/runtime/base.py
+++ b/openhands/runtime/base.py
@@ -171,8 +171,6 @@ class Runtime(FileEditRuntimeMixin):
             or cast(PROVIDER_TOKEN_TYPE, MappingProxyType({})),
             external_auth_id=user_id,
             external_token_manager=True,
-            session_api_key=self.session_api_key,
-            sid=sid,
         )
         raw_env_vars: dict[str, str] = call_async_from_sync(
             self.provider_handler.get_env_vars, GENERAL_TIMEOUT, True, None, False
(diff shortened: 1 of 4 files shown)


openhands/runtime/base.py
```diff
@@ -323,18 +323,24 @@ def on_event(self, event: Event) -> None:
 
     async def _export_latest_git_provider_tokens(self, event: Action) -> None:
         """Refresh runtime provider tokens when agent attemps to run action with provider token"""
-        if not self.user_id:
-            return
-
```
raymyers, line 328: Not sure why this is being removed, is it checked outside this method?

malhotra5: Turns out this logic was never triggering because nested runtimes don't track our external broker user id

jpelletier1:
Hi @malhotra5 thank you for the fix! Out of curiosity: the original issue was reported by OpenHands Cloud users, however I am wondering if the fix applies to OpenHands local generally, too? I noticed a user in the community Slack report a similar issue.

malhotra5:
For local installation users bring their own PAT. There is not refresh mechanism once the token expires, the users need to issue a new key for themselves

jpelletier1:
Oh good point... let me see if I can get more details from the user and we can open a separate issue if necessary.

PR #10645, opened 2025-08-27
Remove image content filtering in ConversationMemory

### LINKED ISSUE #10632: refactor: move `vision_is_active` check from `conversation_memory.process_events` to `LLM` class

In https://github.com/All-Hands-AI/OpenHands/pull/9738, we are trying to implement model routing.

But it is challenging to define a clean interface that mimics `LLM`, because `LLM.completion` currently receives `messages` that with multimodal content removed by `conversation_memory.process_events` --> making it not possible to implement the `MultimodalRouter` without passing raw `events` into then:

https://github.com/All-Hands-AI/OpenHands/pull/9738/files/1803afbc0228a0eef55affd4ac770ea6f017fcb7#diff-87161bc5e5878424ca3e735946431d1bb9e68848182337e2b40ecda12f6b119cR11

We should refactor to move `vision_is_active` check from `conversation_memory.process_events` to `LLM` class, so that we can refind our model router to follow this pattern here:

```
llm_registry = LLMRegistry()

class CompositeLLM(LLM):
	
	def completion(*args, **kwars):
	    # we do model routing logic here
	    
class MultimodalModelRouter(CompositeLLM):
	def completion(messages: list[Message]):
	  # do some classification
		has_multmodal_content: bool = False
		for message in messages:
			if message.contains_vision():
				has_multmodal_content = True
		
		if has_multmodal_content:
			return self.primary_llm(messages)
		else:
			return self.secondary_llm(messages)
	    
llm: CompositeLLM = llm_registry.get_router(
   primary="gpt5",
   secondary="sonnet"
)
agent = Agent(llm=llm, tools=tools)
```

**Additional context**

ryanhoangt:

---

Looking into the code, it seems the `ImageContent`s filtering in `ConversationMemory.process_events` is not needed, as we also remove them when serializing `Message` into dict. This PR:
- Keep `vision_is_active` parameter for `TextContent` adaptations
- Remove `vision_is_active` filtering for `ImageContent` (always include images)
- Let `Message` serialization handle image filtering based on `vision_enabled`

---

Fix #10632

### DIFF
--- a/openhands/memory/conversation_memory.py
+++ b/openhands/memory/conversation_memory.py
@@ -302,10 +302,12 @@ class ConversationMemory:
         elif isinstance(action, MessageAction):
             role = 'user' if action.source == 'user' else 'assistant'
             content = [TextContent(text=action.content or '')]
-            if vision_is_active and action.image_urls:
+            if action.image_urls:
                 if role == 'user':
                     for idx, url in enumerate(action.image_urls):
-                        content.append(TextContent(text=f'Image {idx + 1}:'))
+                        # Only add descriptive text if vision is active
+                        if vision_is_active:
+                            content.append(TextContent(text=f'Image {idx + 1}:'))
                         content.append(ImageContent(image_urls=[url]))
                 else:
                     content.append(ImageContent(image_urls=action.image_urls))
@@ -414,8 +416,8 @@ class ConversationMemory:
             # Create message content with text
             content: list[TextContent | ImageContent] = [TextContent(text=text)]
 
-            # Add image URLs if available and vision is active
-            if vision_is_active and obs.image_urls:
+            # Add image URLs if available
+            if obs.image_urls:
                 # Filter out empty or invalid image URLs
                 valid_image_urls = [
                     url for url in obs.image_urls if self._is_valid_image_url(url)
@@ -424,7 +426,8 @@ class ConversationMemory:
 
                 if valid_image_urls:
                     content.append(ImageContent(image_urls=valid_image_urls))
-                    if invalid_count > 0:
+                    # Only add explanatory text if vision is active
+                    if vision_is_active and invalid_count > 0:
                         # Add text indicating some images were filtered
                         content[
                             0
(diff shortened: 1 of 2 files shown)


openhands/memory/conversation_memory.py
```diff
@@ -302,10 +302,12 @@ def _process_action(
...
             role = 'user' if action.source == 'user' else 'assistant'
             content = [TextContent(text=action.content or '')]
-            if vision_is_active and action.image_urls:
+            if action.image_urls:
                 if role == 'user':
                     for idx, url in enumerate(action.image_urls):
-                        content.append(TextContent(text=f'Image {idx + 1}:'))
+                        # Only add descriptive text if vision is active
+                        if vision_is_active:
+                            content.append(TextContent(text=f'Image {idx + 1}:'))
```
xingyaoww, line 310: I was wondering if we need this at all..

ryanhoangt: Yep we need it to make it the same with current behaviors, i.e., some `TextContent` elements are changed in content or we append some new elements when we have images.

> Do we need to handle this inside LLM, or no?

No we don't need, we have a `format_messages_for_llm` that will automatically drop all the `ImageContent` when serializing:

https://github.com/All-Hands-AI/OpenHands/blob/57aa7d5c12c6c4577d9a31a13154590e44015850/openhands/agenthub/codeact_agent/codeact_agent.py#L206-L215

https://github.com/All-Hands-AI/OpenHands/blob/57aa7d5c12c6c4577d9a31a13154590e44015850/openhands/llm/llm.py#L807-L808

https://github.com/All-Hands-AI/OpenHands/blob/57aa7d5c12c6c4577d9a31a13154590e44015850/openhands/core/message.py#L118-L121

xingyaoww: ahh interesting.. so it is already there

ryanhoangt:
> In this PR, conversation_memory still makes decisions on whether to drop content based on vision.

@enyst Could you elaborate a bit why? My intention is to retain all the `ImageContent` from `process_events` and make them available for `completion` call.

enyst (approved):
Sorry, I assumed we could move the behavior dependent on vision to LLM class, but it’s fine with me if not.

I think who’s doing what with the messages to the LLM becomes slowly… difficult to follow anymore. 😅

PR #10650, opened 2025-08-27
Add support for passing list of `Message` into LLM `completion`

ryanhoangt:

---

---

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -3,7 +3,7 @@ import os
 import time
 import warnings
 from functools import partial
-from typing import Any, Callable
+from typing import Any, Callable, cast
 
 import httpx
 
@@ -209,7 +209,9 @@ class LLM(RetryMixin, DebugMixin):
             """Wrapper for the litellm completion function. Logs the input and output of the completion function."""
             from openhands.io import json
 
-            messages_kwarg: list[dict[str, Any]] | dict[str, Any] = []
+            messages_kwarg: (
+                dict[str, Any] | Message | list[dict[str, Any]] | list[Message]
+            ) = []
             mock_function_calling = not self.is_function_calling_active()
 
             # some callers might send the model and messages directly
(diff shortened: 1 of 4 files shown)


openhands/llm/llm.py
```diff
@@ -209,7 +209,9 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
             """Wrapper for the litellm completion function. Logs the input and output of the completion function."""
             from openhands.io import json
 
-            messages_kwarg: list[dict[str, Any]] | dict[str, Any] = []
+            messages_kwarg: (
+                dict[str, Any] | Message | list[dict[str, Any]] | list[Message]
+            ) = []
```
xingyaoww, line 214: nit: actually, i wonder in what case we will send in single Message/dict? Maybe we can get rid of these two type and their code..

ryanhoangt: Yeah it seems we don't send a single item anywhere, from the [Chat Completion API](https://platform.openai.com/docs/api-reference/chat/create) it also requires array, but just to be safe maybe I won't touch it for now 😅

enyst: It used to be the case with the old agents… I’m not sure offhand it might be the case with the condenser… probably not.

enyst:
@ryanhoangt I see this on main
```
  File "/Users/enyst/repos/odie/.worktrees/settings-from-config/.venv/lib/python3.12/site-packages/litellm/utils.py", line 6488, in validate_and_fix_openai_messages
    if not message.get("role"):
           ^^^^^^^^^^^
  File "/Users/enyst/repos/odie/.worktrees/settings-from-config/.venv/lib/python3.12/site-packages/pydantic/main.py", line 991, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'Message' object has no attribute 'get'
```
Can you replicate? 🤔

Issue #10508, opened 2025-08-19
[Bug, Condenser]: After condensation, the agent seems to forget about task tracking


xingyaoww:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

The agent didn't ever update the task tracking list AFTER condensation.

[trajectory-6839ccba893b407188985cd6421167a7.json](https://github.com/user-attachments/files/21867781/trajectory-6839ccba893b407188985cd6421167a7.json)

### OpenHands Installation

Docker command in README

### Operating System

None

csmith49:
We also see some similar (probably related behavior) where the agent decides to repeat some original instructions after condensation.

enyst:
Just to note, my (unbaked) impression so far is that GPT-5 doesn’t use the task tool as well as Sonnet and/or doesn’t benefit from it as much.

My hypothesis is that Sonnet has had some training with some sort of task tool, and we’ve seen it occasionally make itself a .md file unprompted, prior to the introduction in OH of both the former system prompt and the current task tool.

While GPT-5 doesn’t seem to ‘get’ it; when it uses it, I often see the tasks out-of-order, and then forgotten.

csmith49:
The trajectory linked in the issue description definitely has some undesirable behavior:
* Between events 1 and 100, there are six uses of `task_tracker`
* There's a condensation at event 127
* After event 127 there are _no_ uses of `task_tracker

In my own experiments, I've seen this pattern play out a little differently. Uses of `task_tracker` definitely stop, but it isn't so clearly after the first condensation. Sometimes it takes two to taper off. But condensation definitely changes the distribution of `task_tracker` actions.

As far as I've been able to tell, all the relevant components are working as intended in isolation. Definitely points to a prompting issue, which I think tracks with Engel's experience re: task-tracking and GPT-5.

I'm working on experimenting with a few fixes:
1. Explicitly link task-tracking and condensations. This means summaries should have a section for tracked tasks that preserve task IDs and whatnot.
2. Reinforce task-tracking / condensation interaction in the system prompt.

We'll probably end up doing both.

PR #10656, opened 2025-08-27
fix: Linking condensation and task tracking prompts

csmith49:

---

This PR updates the prompts for the summarizing condenser and for the "long horizon" tasks so that both are aware of the mechanisms of the other.

---

#10508 

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:c2af5ef-nikolaik   --name openhands-app-c2af5ef   docker.all-hands.dev/all-hands-ai/openhands:c2af5ef
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix/condensing-tasks openhands
```


csmith49:
This is a tough PR to test, but the changes to the prompt seem to help keep task tracking ongoing after a condensation. A recent run had a whole new task-tracking workflow happen after a condensation, so it's clear the tool can still be called.

Suggest we merge this PR and continue monitoring this behavior to see if stronger fixes are needed.

openhands/memory/condenser/impl/llm_summarizing_condenser.py
```diff
@@ -69,10 +69,17 @@ def get_condensation(self, view: View) -> Condensation:
...
+        prompt = """You are maintaining a context-aware state summary for an interactive agent.
+You will be given a list of events corresponding to actions taken by the agent, and the most recent previous summary if one exists.
+If the events being summarized contain ANY task-tracking, you MUST include a TASK_TRACKING section to maintain continuity.
+When referencing tasks make sure to preserve exact task IDs and statuses.
+
+Track:
 
 USER_CONTEXT: (Preserve essential user requirements, goals, and clarifications in concise form)
 
+TASK_TRACKING: {Active tasks, their IDs and statuses - PRESERVE TASK IDs}
```
enyst, line 81: If I understand correctly, this will not include the descriptions, just the task ids and pending/done?

enyst (approved):
Thank you! Let's try.

Afaik the long horizon prompt is not default, so maybe we don't need an eval? If we have an eval on long horizon in the past, maybe we should make a quick one (50 instances?)

openhands/memory/condenser/impl/llm_summarizing_condenser.py
csmith49: I've seen a mix of behaviors here, depending on model/prompt/phase of the moon -- sometimes it's just IDs, other times there are also short descriptions. So long as the IDs are there I don't think I mind extra content.

csmith49:
I don't know of any evaluations with the long horizon prompt, but I'll ask around and see if there's something worth running again.

PR #10667, opened 2025-08-28
Fix: Add method to merge conversation stats

malhotra5:

---
In nested runtimes, we may generate cost before starting the runtime pod. Then the runtime pod will continue to generate more cost, and send webhooks back about the costs generated.

For this reason, we create a method that can merge an existing conversation stats object with another one. The hard requirement is that no `service_ids` are duplicated between the two stats objects, guaranteeing that the llms which generated spend are mutually exclusive.

NOTE: this PR relies on `save_metrics` to work as expected. This bug wrt to this method is fixed in #10676

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:31bdfde-nikolaik   --name openhands-app-31bdfde   docker.all-hands.dev/all-hands-ai/openhands:31bdfde
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@merge-convo-stats openhands
```


raymyers (changes requested):
The method calling `merge` from SaaS helpfully has the comment:
```
# Merging automatically saves to file store
```
It needs it because the name "merge" does not indicate that one of its primary effects is to save. Let's either have saving done separately or rename to something like merge_and_save. A clunky name that's honest is better than simple and dishonest.

https://www.digdeeproots.com/articles/naming-process/

malhotra5:
Thanks for the callout! That was an interesting read as well :)

PR #10671, opened 2025-08-28
Add support for passing list of `Message` into LLM `completion`

ryanhoangt:

---

This PR adds support for passing list of Message into LLM.completion, which makes it easy for accessing the list of Message in the completion to make routing decision, if we want to make Router inherit LLM.

In the previous PR #10650 I forgot to override the existing `messages` of kwargs with the formatted one:

```py
            messages: list[dict] = []
            if messages_list and isinstance(messages_list[0], Message):
                messages = self.format_messages_for_llm(
                    cast(list[Message], messages_list)
                )
            else:
                messages = cast(list[dict[str, Any]], messages_list)

            kwargs['messages'] = messages
```

I tested this locally it works normally now.

---


xingyaoww (approved):
LGTM! Thanks!

PR #10676, opened 2025-08-28
Fix ConversationStats metrics loss for unregistered services

malhotra5:

Fixed a bug where conversation statistics would lose cost tracking data for LLM services that were used in previous sessions but not yet registered in the current session. This ensures accurate billing and usage tracking across application restarts.

---

This PR fixes a critical bug in the `ConversationStats.save_metrics()` method where restored metrics for unregistered services were being lost when saving metrics to the file store.

**Problem:**
- When OpenHands restarts, conversation stats are restored from the file store into `restored_metrics`
- If an LLM service from a previous session hasn't been registered yet in the current session, its metrics remain in `restored_metrics`
- The original `save_metrics()` method only saved `service_to_metrics`, completely ignoring `restored_metrics`
- This caused permanent loss of cost tracking data for unregistered services

**Solution:**
1. **Combined metrics preservation**: Modified `save_metrics()` to combine both `restored_metrics` and `service_to_metrics` before saving
2. **Duplicate service ID validation**: Added validation to throw a `ValueError` if the same service ID exists in both dictionaries (which should never happen in normal operation)
3. **Comprehensive testing**: Added unit tests to reproduce the bug and verify the fix

**Key design decisions:**
- Services in `service_to_metrics` take precedence over `restored_metrics` when combining (though duplicates should never occur)
- Added explicit error handling for duplicate service IDs to catch potential bugs early
- Preserved existing behavior for all normal operation scenarios

---

This addresses a data loss bug discovered during code analysis of the ConversationStats class.

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/0fcfbe1ca839452488ec4bcbb131bcba)

### DIFF
--- a/openhands/server/services/conversation_stats.py
+++ b/openhands/server/services/conversation_stats.py
@@ -36,7 +36,25 @@ class ConversationStats:
             return
 
         with self._save_lock:
-            pickled = pickle.dumps(self.service_to_metrics)
+            # Check for duplicate service IDs between restored and service metrics
+            duplicate_services = set(self.restored_metrics.keys()) & set(
+                self.service_to_metrics.keys()
+            )
+            if duplicate_services:
+                raise ValueError(
+                    f'Duplicate service IDs found between restored and service metrics: {duplicate_services}. '
+                    'This should not happen as registered services should be removed from restored_metrics.'
+                )
+
+            # Combine both restored metrics and service metrics to avoid data loss
+            # Start with restored metrics (for services not yet registered)
+            combined_metrics = self.restored_metrics.copy()
+
+            # Add service metrics (for registered services)
+            # Since we checked for duplicates above, this is safe
+            combined_metrics.update(self.service_to_metrics)
+
+            pickled = pickle.dumps(combined_metrics)
             serialized_metrics = base64.b64encode(pickled).decode('utf-8')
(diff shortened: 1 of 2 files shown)


openhands/server/services/conversation_stats.py
```diff
@@ -36,7 +36,25 @@ def save_metrics(self):
...
 
         with self._save_lock:
-            pickled = pickle.dumps(self.service_to_metrics)
+            # Check for duplicate service IDs between restored and service metrics
+            duplicate_services = set(self.restored_metrics.keys()) & set(
+                self.service_to_metrics.keys()
+            )
+            if duplicate_services:
+                raise ValueError(
+                    f'Duplicate service IDs found between restored and service metrics: {duplicate_services}. '
```
enyst, line 45: Is a `ValueError` a bit harsh? Thinking a warning is nicer, unless you think we need to crash

enyst (approved):

openhands/server/services/conversation_stats.py
malhotra5: @openhands please downgrade this from a `ValueError` to a `logger.error` statement 

please also update the unit test `test_save_metrics_throws_error_on_duplicate_service_ids`

PR #10677, opened 2025-08-28
fix(frontend): maintenance banner cannot be dismissed (conversation UX improvements)

hieptl:

The maintenance banner is currently not dismissible. When users click the close icon, the banner should be removed from view.

**Acceptance Criteria:**
- The maintenance banner must be dismissed upon clicking the close icon.

---

The PR allows the users to dismiss the maintenance banner.

https://github.com/user-attachments/assets/6749d96c-c829-44da-b969-797dfebc0abc

---

### DIFF
--- a/frontend/src/components/features/maintenance/maintenance-banner.tsx
+++ b/frontend/src/components/features/maintenance/maintenance-banner.tsx
@@ -1,7 +1,5 @@
 import { useTranslation } from "react-i18next";
-import { useMemo } from "react";
-import { isAfter, isBefore } from "date-fns";
-import { useLocalStorage } from "@uidotdev/usehooks";
+import { useMemo, useState } from "react";
 import { FaTriangleExclamation } from "react-icons/fa6";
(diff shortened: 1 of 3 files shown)


frontend/src/components/features/maintenance/maintenance-banner.tsx
```diff
@@ -11,11 +9,9 @@ interface MaintenanceBannerProps {
 }
 
 export function MaintenanceBanner({ startTime }: MaintenanceBannerProps) {
+  const [isBannerShown, setIsBannerShown] = useState<boolean>(true);
```
amanape: Doesn’t this imply that we will show the banner every time we reload the page? We should persist the user’s decision here for now. In the future, I want to consider storing the user’s preference in a userPref cookie or something

hieptl: Hi @amanape, I've updated the code based on your feedback. Thank you for your feedback! 🙏

amanape (approved):

PR #10679, opened 2025-08-28
refactor(frontend): make whole suggested tasks clickable (conversation UX improvements)

hieptl:

The entire suggested task should be clickable, rather than requiring users to click the launch button.

---

The PR makes whole suggested tasks clickable (conversation UX improvements)

https://github.com/user-attachments/assets/a41bf1c6-6260-42eb-bd97-7005503a2443

---

### DIFF
--- a/frontend/src/components/features/home/tasks/task-card.tsx
+++ b/frontend/src/components/features/home/tasks/task-card.tsx
@@ -61,7 +61,11 @@ export function TaskCard({ task }: TaskCardProps) {
   }
 
   return (
-    <li className="py-3 flex items-center gap-3 last:border-b-0">
+    <button
+      type="button"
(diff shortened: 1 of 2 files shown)


frontend/src/components/features/home/tasks/task-card.tsx
```diff
@@ -83,6 +87,6 @@ export function TaskCard({ task }: TaskCardProps) {
         {!isPending && t("HOME$LAUNCH")}
         {isPending && t("HOME$LOADING")}
       </button>
```
amanape: Let’s change the previous button element to something else since it is bad practice to nest a button within a button

amanape (changes requested):

frontend/src/components/features/home/tasks/task-card.tsx
hieptl: Hi @amanape, I've updated the code based on your feedback. Thank you for your feedback! 🙏

PR #10700, opened 2025-08-29
refactor: move content truncation from ConversationMemory to LLM layer

ryanhoangt:

---

This PR moves the message content truncation to LLM layer, which maintains the same functionality while improving the architecture and separation of concerns. The original untruncated messages are now preserved in `ConversationMemory`, while truncation is applied at the appropriate layer just before sending to the LLM.

This would make the list of `Message` passed to `completion` independent of LLM, which will simplify how we do routing, because otherwise we have to recompute the `Message` list after configuring the active LLM:

```py
        messages_for_routing_decision = self._get_messages(
            condensed_history, initial_user_message
        )
        self.llm_registry.configure_active_llm(
            messages_for_routing_decision,
        )

        # Get the active LLM after routing decision
        active_llm = self.llm_registry.get_active_llm()

        # Recompute messages after active LLM is set
        messages = self._get_messages(condensed_history, initial_user_message)
        params: dict = {
            'messages': messages,
        }
```

---

### DIFF
--- a/openhands/events/serialization/event.py
+++ b/openhands/events/serialization/event.py
@@ -162,6 +162,3 @@ def event_to_trajectory(event: 'Event', include_screenshots: bool = False) -> di
             else DELETE_FROM_TRAJECTORY_EXTRAS_AND_SCREENSHOTS,
         )
     return d
-
-
-# Import from the new location to maintain backward compatibility


xingyaoww (commented):
Actually I wonder if it is feasible to do it in LLMMessage level for this 🤔 that could give user more control

enyst (commented):
> This would make the list of Message passed to completion independent of LLM, which will simplify how we do routing, because otherwise we have to recompute the Message list after configuring the active LLM

Could you explain this more?

I think
- this message truncation doesn’t depend on LLM
- for some event types, truncation is in `openhands-aci`
- for CmdOutputObs, a truncation happens on event creation

ryanhoangt:
> Could you explain this more?

Below is what we currently have to do, i.e., calling `self._get_messages` two times, one to get the message list to select LLM for the current turn, and after setting the active LLM, we need to compute it again to truncate observations.

```py        
        messages_for_routing_decision = self._get_messages(
            condensed_history, initial_user_message
        )
        self.llm_registry.configure_active_llm(
            messages_for_routing_decision,
        )

        # Get the active LLM after routing decision
        active_llm = self.llm_registry.get_active_llm()

        # Recompute messages after active LLM is set
        messages = self._get_messages(condensed_history, initial_user_message)
        params: dict = {
            'messages': messages,
        }
```

> this message truncation doesn’t depend on LLM

Yeah, after looking more closely, I see this value is actually fixed for all LLMs 🤔 so leaving it in `ConversationMemory` as currently is fine as well. It will be problematic if the primary and secondary LLM have this config value set differently. Do you think this refactor makes sense from an architectural point of view?

enyst:
This truncation is just making sure we don’t end up sending like 100k tokens in one observation to the LLM. We already have dicts here, when we apply it. 🤔 I think maybe it belongs in Message serialization, like [here](https://github.com/ryanhoangt/OpenHands/blob/23713bfe8c8e270e9503810df33d9f6d7c627d73/openhands/core/message.py#L100), where we make the dicts in the first place…? I guess that’s what Xingyao said too?

I don’t remember why it’s not, already. I do think maybe we can try that, it would be less complexity overall. What do you think? 🤔

enyst:
Sorry @ryanhoangt , I'm not sure if this PR still applies in V1, or we already handle this?

ryanhoangt:
No I think this is not relevant anymore, let me close it!

PR #10725, opened 2025-08-30
Fix GraphQL URL configuration for GitHub Enterprise Server

malhotra5:

Fixes a bug where GitHub Enterprise Server users could not use GraphQL-based features in OpenHands. The GraphQL URL was incorrectly configured to use the GitHub.com endpoint instead of the Enterprise Server's GraphQL endpoint, causing API calls to fail.

---

This PR fixes a bug in the `GitHubService` class where the GraphQL URL was not properly configured based on the `base_domain` parameter for GitHub Enterprise Server installations.

**Changes made:**
1. **Updated `GitHubService.__init__`**: Now correctly sets the `GRAPHQL_URL` attribute based on the `base_domain` parameter:
   - For GitHub.com: `https://api.github.com/graphql`
   - For GitHub Enterprise Server: `https://{base_domain}/api/graphql`

2. **Simplified `execute_graphql_query`**: Updated the method to use the `self.GRAPHQL_URL` attribute instead of hardcoded URLs, ensuring consistency with the URL configuration pattern used for the REST API.

3. **Added comprehensive unit tests**: Created tests to verify that the GraphQL URL is correctly configured for both GitHub.com and GitHub Enterprise Server scenarios.

**Design decisions:**
- Used the existing attribute-based approach for consistency with how `BASE_URL` is handled
- Leveraged the existing URL construction logic in the `__init__` method rather than duplicating it in `execute_graphql_query`
- Followed the same pattern as the REST API configuration for maintainability

**Before this fix:**
- GitHub Enterprise Server users would get API errors when using GraphQL features
- GraphQL calls would incorrectly target `https://api.github.com/graphql` regardless of the configured `base_domain`

**After this fix:**
- GitHub Enterprise Server users can successfully use GraphQL features
- GraphQL URL correctly uses `/api/graphql` endpoint for Enterprise Server installations

---

This addresses a bug discovered during code review where the GraphQL URL configuration was inconsistent with the REST API URL configuration for GitHub Enterprise Server deployments.

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/525b90a094964f3eab1f3407245d9921)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:f647e5b-nikolaik   --name openhands-app-f647e5b   docker.all-hands.dev/all-hands-ai/openhands:f647e5b
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix-github-graphql-url-enterprise-server openhands
```

### DIFF
--- a/openhands/integrations/github/github_service.py
+++ b/openhands/integrations/github/github_service.py
@@ -57,7 +57,6 @@ class GitHubService(
     ) -> None:
         self.user_id = user_id
         self.external_token_manager = external_token_manager
-        self.base_domain = base_domain
 
         if token:
             self.token = token
--- a/openhands/integrations/github/github_service.py
+++ b/openhands/integrations/github/github_service.py
@@ -42,6 +42,7 @@ class GitHubService(
     """
 
     BASE_URL = 'https://api.github.com'
+    GRAPHQL_URL = 'https://api.github.com/graphql'
     token: SecretStr = SecretStr('')
     refresh = False
 
@@ -56,12 +57,14 @@ class GitHubService(
     ) -> None:
         self.user_id = user_id
         self.external_token_manager = external_token_manager
+        self.base_domain = base_domain
 
         if token:
             self.token = token
 
         if base_domain and base_domain != 'github.com':
             self.BASE_URL = f'https://{base_domain}/api/v3'
+            self.GRAPHQL_URL = f'https://{base_domain}/api/graphql'
 
         self.external_auth_id = external_auth_id
         self.external_auth_token = external_auth_token
--- a/openhands/integrations/github/service/base.py
+++ b/openhands/integrations/github/service/base.py
@@ -18,9 +18,11 @@ class GitHubMixinBase(BaseGitService):
     """
 
     BASE_URL: str
+    GRAPHQL_URL: str
     token: SecretStr
     refresh: bool
     external_auth_id: str | None
+    base_domain: str | None
 
     async def _get_github_headers(self) -> dict:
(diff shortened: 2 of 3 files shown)


openhands/integrations/github/github_service.py
```diff
@@ -56,12 +57,14 @@ def __init__(
     ) -> None:
         self.user_id = user_id
         self.external_token_manager = external_token_manager
+        self.base_domain = base_domain
```
malhotra5:
```suggestion
```

openhands/integrations/github/github_service.py
```diff
@@ -62,6 +63,7 @@ def __init__(
 
         if base_domain and base_domain != 'github.com':
             self.BASE_URL = f'https://{base_domain}/api/v3'
+            self.GRAPHQL_URL = f'https://{base_domain}/api/graphql'
```
enyst, line 66: Interestingly, I think OH-GPT-5 noticed something like this. I actually love the level of detail it goes into, whether I like it or not, OpenAI made a darn good model. 😅

enyst (approved):
Thank you!

openhands/integrations/github/github_service.py
malhotra5: Absolutely

PR #10735, opened 2025-09-01
Fix ignored order parameter in BitBucket service get_paginated_repos method

malhotra5:

Fixed a bug in the BitBucket integration where the `order` parameter (ascending/descending) was being ignored when fetching repositories. Users can now properly sort repositories in ascending or descending order as expected.

---

This PR fixes the ignored `order` parameter in the BitBucket service's `get_paginated_repos` method. The method was accepting an `order` parameter via `search_repositories` but was discarding it and always forcing descending order (via "-" prefix) or using default behavior.

**Changes made:**
1. **Updated `get_paginated_repos` method**: Now accepts and honors the `order` parameter (`asc`/`desc`)
2. **Fixed sort logic**: Conditionally applies BitBucket's "-" prefix for descending order, no prefix for ascending
3. **Updated interface consistency**: Made sure all implementations (BitBucket, GitHub, GitLab) have matching method signatures
4. **Fixed GitLab implementation**: Was also hardcoding 'desc' instead of using the order parameter
5. **Added comprehensive unit tests**: Two new tests verify the order parameter functionality

**Technical details:**
- BitBucket API uses "-" prefix for descending order (e.g., "-updated_on") and no prefix for ascending (e.g., "updated_on")
- The fix ensures the sort string is built correctly based on the incoming order parameter
- All existing tests continue to pass, maintaining backward compatibility

---

Fixes the issue reported in the original request where the order parameter was being ignored in BitBucket repository searches.

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/e228c051756446a8972cfc2f5116229e)

### DIFF
--- a/openhands/integrations/gitlab/service/repos.py
+++ b/openhands/integrations/gitlab/service/repos.py
@@ -85,7 +85,7 @@ class GitLabReposMixin(GitLabMixinBase):
             repository = await self.get_repository_details_from_repo_name(repo_path)
             return [repository]
 
-        return await self.get_paginated_repos(1, per_page, sort, None, query)
+        return await self.get_paginated_repos(1, per_page, sort, None, query, order)
 
     async def get_paginated_repos(
         self,
@@ -94,6 +94,7 @@ class GitLabReposMixin(GitLabMixinBase):
         sort: str,
         installation_id: str | None,
         query: str | None = None,
+        order: str = 'desc',
     ) -> list[Repository]:
         url = f'{self.BASE_URL}/projects'
         order_by = {
@@ -107,7 +108,7 @@ class GitLabReposMixin(GitLabMixinBase):
             'page': str(page),
             'per_page': str(per_page),
             'order_by': order_by,
-            'sort': 'desc',  # GitLab uses sort for direction (asc/desc)
+            'sort': order,  # GitLab uses sort for direction (asc/desc)
             'membership': True,  # Include projects user is a member of
         }
 
--- a/openhands/integrations/service_types.py
+++ b/openhands/integrations/service_types.py
@@ -514,6 +514,7 @@ class GitService(Protocol):
         sort: str,
         installation_id: str | None,
         query: str | None = None,
+        order: str = 'desc',
     ) -> list[Repository]:
         """Get a page of repositories for the authenticated user"""
         ...
--- a/tests/unit/integrations/bitbucket/test_bitbucket.py
+++ b/tests/unit/integrations/bitbucket/test_bitbucket.py
@@ -871,3 +871,135 @@ def test_initialize_repository_for_runtime_without_bitbucket_token(
     assert ProviderType.GITHUB in provider_tokens
     assert ProviderType.GITLAB in provider_tokens
     assert ProviderType.BITBUCKET not in provider_tokens
+
+
+@pytest.mark.asyncio
+async def test_bitbucket_order_parameter_honored():
+    """Test that the Bitbucket service correctly honors the order parameter (asc/desc)."""
+    # Create a service instance
+    service = BitBucketService(token=SecretStr('test-token'))
+
+    # Mock the _make_request method to avoid actual API calls
+    with patch.object(service, '_make_request') as mock_request:
+        # Mock response for repositories
+        mock_request.return_value = ({'values': []}, {})
+
+        # Test ascending order
+        await service.get_paginated_repos(
+            page=1,
+            per_page=10,
+            sort='updated',
+            installation_id='test-workspace',
+            query=None,
+            order='asc',
+        )
+
+        # Verify the call was made with ascending order (no '-' prefix)
+        assert mock_request.call_count == 1
+        call_args = mock_request.call_args_list[0]
+        url, params = call_args[0]
+        assert params['sort'] == 'updated_on'  # No '-' prefix for ascending
+        assert 'repositories/test-workspace' in url
+
+        # Reset mock for next test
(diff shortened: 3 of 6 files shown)


malhotra5:
For now this standardizes the `order` param across the providers. This is still brittle because we're using strings for these params rather than strongly typed objects.

In a future PR we will transition to use a shared set of enums of API params across all providers

malhotra5:
## 🤖 OpenHands Agent Update

I have verified the API documentation for all three providers and made additional fixes to ensure proper order parameter handling:

### 📚 API Documentation Analysis

**GitHub API** ([docs.github.com/en/rest/repos/repos](https://docs.github.com/en/rest/repos/repos)):
- Uses `sort` parameter for field selection (`created`, `updated`, `pushed`, `full_name`)
- Uses `direction` parameter for order (`asc`, `desc`)
- Default: `asc` when using `full_name`, otherwise `desc`

**GitLab API** ([docs.gitlab.com/api/projects](https://docs.gitlab.com/api/projects/)):
- Uses `order_by` parameter for field selection
- Uses `sort` parameter for direction (`asc`, `desc`)

**BitBucket API** (from implementation analysis):
- Uses `sort` parameter with field names
- Uses `-` prefix for descending order (e.g., `-updated_on`)
- No prefix for ascending order (e.g., `updated_on`)

### 🔧 Additional Fixes Applied

1. **GitHub Implementation**: Fixed missing `direction` parameter in API calls
   - Added `params["direction"] = order` in `get_paginated_repos`
   - Updated `get_all_repositories` to accept and use order parameter

2. **GitLab Implementation**: Already correct - properly uses `sort` parameter for direction

3. **BitBucket Implementation**: Already fixed in original PR - properly applies `-` prefix based on order

4. **Interface Consistency**: Updated all method signatures to include `order` parameter with default `"desc"`

### ✅ Verification

All tests pass (22/22 BitBucket tests) and the order parameter is now properly honored across all three providers according to their respective API specifications.

The original issue has been resolved and extended to ensure consistency across all Git providers.

malhotra5:
Manually QA'ed this myself and everything still works as expected

openhands/integrations/service_types.py
```diff
@@ -514,6 +514,7 @@ async def get_paginated_repos(
         sort: str,
         installation_id: str | None,
         query: str | None = None,
+        order: str = 'desc',
```
enyst, line 517: Interesting, so the bug was everywhere, not just bitbucket

openhands/integrations/gitlab/service/repos.py
```diff
@@ -145,7 +146,7 @@ async def get_all_repositories(
                 'page': str(page),
                 'per_page': str(PER_PAGE),
                 'order_by': order_by,
-                'sort': 'desc',  # GitLab uses sort for direction (asc/desc)
+                'sort': order,  # GitLab uses sort for direction (asc/desc)
```
enyst, line 149: Stupid question, why do we need `order` when we have `sort` in the signature?

enyst: Ah that's for which field

openhands/integrations/service_types.py
malhotra5: We've always defaulted to `desc` across all providers; some were supporting params and some weren't 

The behavior for repos listing and searching stays the exact same before and after this PR. This PR standardizes the params

A followup PR will standardize the enum types as well for these params

enyst:
@openhands I want a /codereview-roasted on this PR.

enyst:
I'm curious if the agent will complain that we try to fix here order across all three providers, but tests cover only bb.

openhands/integrations/gitlab/service/repos.py
malhotra5: Ah it gets confusing because params mean different things across different providers (e.g sort is different for github vs gitlab)

The breakdown of params is included in https://github.com/All-Hands-AI/OpenHands/pull/10735#issuecomment-3240768976

malhotra5:
@openhands let's address the following critical issues. Read the diff on this PR then address the follow critical issues

1. Code duplication: Sort mapping logic is duplicated between `get_paginated_repos` and `get_all_repositories` methods
2. Unnecessary complexity: Order application logic is overly verbose with redundant comments

tests/unit/integrations/bitbucket/test_bitbucket.py
```diff
@@ -871,3 +871,238 @@ def test_initialize_repository_for_runtime_without_bitbucket_token(
     assert ProviderType.GITHUB in provider_tokens
     assert ProviderType.GITLAB in provider_tokens
     assert ProviderType.BITBUCKET not in provider_tokens
+
```
malhotra5, line 911: @openhands we seem to be using direct comparisons for called args instead of using `assert_called_with()` 

please rework the tests in this file so that we rely on `assert_called_with()`

malhotra5:
@openhands read the diff of this PR

Then add test coverage for the changes made to the `GitLabService` and `GitHubService` classes 

DO NOT PUSH CHANGES TO THIS PR. Create a new branch off this branch and open a draft PR that attempts to merge back to this branch

enyst (approved):

openhands/integrations/gitlab/service/repos.py
enyst: Ah I see. Well that's one very good reason why I like the overall structure here, service module per provider / repos.py !

PR #10769, opened 2025-09-02
chore(frontend): remove feature flag (microagent management)

hieptl:

We need to remove the feature flag to show the microagent management page.

---

---


amanape (approved):

enyst:
If we’re removing the feature flag, thus making it available by default, we should document this.

I tweak the PR description, so it flags that and show in release notes, please take a look?

hieptl:
> If we’re removing the feature flag, thus making it available by default, we should document this.
> 
> I tweak the PR description, so it flags that and show in release notes, please take a look?

Hi @enyst, that makes perfect sense. Thank you for pointing it out and updating the PR description! 🙏

PR #10792, opened 2025-09-03
Fix swe-bench `run_infer.py` config parsing from config.toml

Zacharias030:

Fixes a bug where the values selected for the [agent] and [llm] config groups were not loaded from config.toml unless the default location "./config.toml" was used, but not via `--config-file` or `--agent-config` / `--llm-config`.

---
Two things:
* pass the args.config_file into the different config loaders.
* Also add config loading for the AgentConfig, which enables e.g. choosing a file for the system prompt template.

---

### DIFF
--- a/evaluation/benchmarks/swe_bench/run_infer.py
+++ b/evaluation/benchmarks/swe_bench/run_infer.py
@@ -44,6 +44,7 @@ from openhands.controller.state.state import State
 from openhands.core.config import (
     AgentConfig,
     OpenHandsConfig,
+    get_agent_config_arg,
     get_evaluation_parser,
     get_llm_config_arg,
 )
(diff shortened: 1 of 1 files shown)


evaluation/benchmarks/swe_bench/run_infer.py
```diff
@@ -811,20 +812,25 @@ def filter_dataset(dataset: pd.DataFrame, filter_column: str) -> pd.DataFrame:
...
 
+    agent_config = None
+    if args.agent_config:
+        agent_config = get_agent_config_arg(args.agent_config, args.config_file)
+
     details = {'mode': args.mode}
     _agent_cls = openhands.agenthub.Agent.get_cls(args.agent_cls)
 
-    dataset_descrption = (
+    dataset_description = (
```
Zacharias030, line 822: side change: typo

evaluation/benchmarks/swe_bench/run_infer.py
```diff
@@ -811,20 +812,25 @@ def filter_dataset(dataset: pd.DataFrame, filter_column: str) -> pd.DataFrame:
             'No Condenser config provided via EVAL_CONDENSER, using NoOpCondenser.'
         )
 
+    agent_config = None
+    if args.agent_config:
+        agent_config = get_agent_config_arg(args.agent_config, args.config_file)
```
Zacharias030, line 817: We handle the agent_config, which defaults to `None` in  `make_metadata` below in the same way like for most other benchmarks. These lines are taken straight from those modules.

PR #10818, opened 2025-09-04
feat: implement fade in/out animation for chat suggestions

amanape:

---
Replace immediate hide/show behavior with smooth opacity transition when user expands chat input. The zero state icon and suggestions now fade out over 300ms when input height exceeds threshold and fade back in when collapsed.

- Use CSS transition-opacity with 300ms duration and ease-in-out timing
- Add pointer-events-none when hidden to prevent interaction
- Use cn utility for cleaner className composition
- Maintain existing functionality while improving UX

https://github.com/user-attachments/assets/fea618af-50d5-483a-bffc-c5b23ffb7603

---

### DIFF
--- a/frontend/src/components/features/chat/chat-suggestions.tsx
+++ b/frontend/src/components/features/chat/chat-suggestions.tsx
@@ -5,6 +5,7 @@ import { I18nKey } from "#/i18n/declaration";
 import BuildIt from "#/icons/build-it.svg?react";
 import { SUGGESTIONS } from "#/utils/suggestions";
 import { RootState } from "#/store";
+import { cn } from "#/utils/utils";
 
 interface ChatSuggestionsProps {
   onSuggestionsClick: (value: string) => void;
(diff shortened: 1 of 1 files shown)


hieptl:
Hi @amanape,

I encountered a few issues while testing this PR locally:

**Issue:** Drag behavior: When dragging the grip upward, the chat input UI does not render correctly. We can refer to the video below for more information.

https://github.com/user-attachments/assets/5b4148c8-2dc3-44e9-b23d-0c6003ab2cbb

**Issue:** Long input content: When entering lengthy text, the chat input UI also fails to display properly. We can refer to the video below for more information.

https://github.com/user-attachments/assets/5b587a82-5391-4a9a-8e09-e5cef3878c30

It appears that the root cause may be related to the use of opacity. Since the chat suggestions section is only visually hidden (rather than removed from the DOM), it still occupies space. As a result, the chat input cannot dynamically resize as expected.

Thank you! 🙏

hieptl (approved):
Thank you! 🙏

Issue #8989, opened 2025-06-08
Chat Reset


PlanetMacro:
**What problem or use case are you trying to solve?**

A feature to reset all chat history in a current session would be good to have. In some projects initializing the environment takes a lot of time (installing all requirements, building the project, training models ect). In these projects its not efficient to start every task in a new session.

However executing all tasks in the same chat, the agent tends to repeat initial tasks like "install all requirements and build the project". It would therefore be better to have a chat reset feature, which keeps the environment while resetting the chat.

As an example: I use OpenHands to write automated math proofs in lean4. However initializing the environment can take a few hours for lean to build all dependencies.

paul-wade:
The conversation manager already has good separation between environment state and chat history, so we should be able to clear the event stream while preserving the runtime. The main question is where to put the reset button in the UI - maybe in the conversation controls next to delete? Happy to take this on if no one else is working on it.

paul-wade:
It does feel like it would create some ambiguity though - it sort of confuses what a conversation is and what a runtime is. Users might not understand the distinction between clearing chat messages vs. clearing the entire session/environment. This could be important for the UX side.

kripper:
Why not implement a `/reset` message command?

fury0928:
Hi @kripper @jpelletier1 is this resolved? or still open for contributors?

kripper:
> Hi [@kripper](https://github.com/kripper) [@jpelletier1](https://github.com/jpelletier1) is this resolved? or still open for contributors?

https://github.com/OpenHands/OpenHands/issues/12564

My PR creates a new blank conversation with a new session ID and reusing the same container.
It would also be usefull to have the option to reset the conversation but keeping the same session ID, because sometimes you will have references to this container.

VascoSch92:
[...]

PR #10819, opened 2025-09-04
feat: reset conversation

kripper:

 **End-user friendly description of the problem this fixes or functionality this introduces:**

Users can now reset a conversation.
This PR adds a Reset button to the conversation menu:

image

When resetting a conversation, a new conversation is created (with empty context) and the sandbox container is being reused (the original container is renamed).
The previous conversation can be kept as reference (without its container) or be deleted.

Resetting a conversation is usefull when the context becomes cluttered and the condenser doesn't work correctly (OH restarts trying to fix old issues that were already solved and starts messes things up).

This commit introduces a **"Reset Conversation"** feature, enabling users to discard an existing conversation and optionally remove its history. The user interface and related state handling have been expanded accordingly:

- **UI Components & Interaction:**
  - Adds a new modal, `ConfirmResetModal`, prompting confirmation and allowing users to choose whether they want to delete the old conversation entirely.
  - Adds a checkbox (`delete old conversation`) within the modal to control deletion behavior (with `isLoading` state to disable during operations).
  - Enhances conversation context menus and cards to add a new "Reset" option—integrated via a RotateCcw icon and i18n translation keys.
  - Updates `ConversationCard` to accept an optional `onReset` prop and wire it to user interactions.
  - Extends `ConversationPanel` to maintain reset-related state — including visibility and the ID of the conversation to reset — and to render the `ConfirmResetModal` as needed.

- **State & API Logic:**
  - Implements a new react-query mutation hook, `useResetConversation`, which calls a backend endpoint (`POST /api/conversations/{conversationId}/reset`) with an optional `delete_old_conversation` flag.
  - On success, displays a success toast, invalidates relevant queries (`conversations` and `active-conversation`), and navigates the user to the new conversation (closing the panel if appropriate).
  - On error, presents an error toast to inform the user.

- **i18n & Localization:**
  - Adds several translation keys in `i18n/declaration.ts`:
    - `BUTTON$RESET`
    - `CONVERSATION$RESET_WARNING`
    - `CONVERSATION$DELETE_OLD_CONVERSATION`
    - `CONVERSATION$RESET_SUCCESS`
    - `CONVERSATION$RESET_ERROR`

**Design decisions:**
- The reset operation preserves a clean UX by using a modal confirmation instead of instant, potentially destructive action.
- The "delete old conversation" checkbox gives flexibility: users can choose to retain or discard history.
- UI behavior—like disabling buttons during resetting—defensively prevents unexpected multiple triggers.
- React-query is used to keep data synchronized and simplify API interaction.
  
---

_None mentioned in the commit—no linked issue reference was provided._


kripper:
Fix https://github.com/All-Hands-AI/OpenHands/issues/8989

pcuci:
Not sure I follow, how do you keep project state by creating a new container?

Would "clear" work better as the CTA? I'm thinking about it in the sense of "I clear my terminal" window but it still keeps all my env vars I've set

While "Reset" is common for web-forms, I immediately assumed the entire container would get wiped clean; in which case, why not just start a new thread?

kripper:
The context-reset feature was originally just a workaround to keep working on a task after context rot or hitting the context limit. Renaming and reusing the same container for the new blank conversation was also a workaround, meant to avoid the cost of cloning the container.

After implementing container branching via cheap BTRFS snapshots (with effectively zero time or storage overhead), this workaround is no longer needed. It is now more convenient to simply clone the container instead of "stealing" it from the old conversation (which was no problem, because OH recreats the stolen sandbox with a fresh one when it is missing).

Another learned lesson is that resetting a conversation is completely independent from what you want to do with the containers. In most cases, you want to reset the conversation and start with a condensed prompt, or even a completely new prompt for a different task, while still using the cloned container (a cheap snapshot) underneath.

See also: https://github.com/OpenHands/software-agent-sdk/issues/1787#issuecomment-3801448358

PR #10822, opened 2025-09-04
Revert "feat(agent): add security-related items in system prompt to defense against data exfiltration"

enyst:
Reverts All-Hands-AI/OpenHands#10477

Maybe we could give this some thought: it needs done, but maybe we can adjust the prompt so that it's not so broad, or maybe give some ability to the user... OK, not sure what ability wouldn't be hijackable, but it's a tough tradeoff.
image

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:607e41a-nikolaik   --name openhands-app-607e41a   docker.all-hands.dev/all-hands-ai/openhands:607e41a
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@revert-10477-xw/security-aware openhands
```

### DIFF
--- a/openhands/agenthub/codeact_agent/prompts/system_prompt.j2
+++ b/openhands/agenthub/codeact_agent/prompts/system_prompt.j2
@@ -62,18 +62,8 @@ Your primary role is to assist users by executing commands, modifying code, and 
 </PROBLEM_SOLVING_WORKFLOW>
 
 <SECURITY>
-* Apply least privilege: scope file paths narrowly, avoid wildcards or broad recursive actions.
-* NEVER exfiltrate secrets (tokens, keys, .env, PII, SSH keys, credentials, cookies)!
-  - Block: uploading to file-sharing, embedding in code/comments, printing/logging secrets, sending config files to external APIs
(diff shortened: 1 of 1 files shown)


enyst:
Also maybe I should note that the issue right now, that's blocking me, is an actual bug (refresh tokens issues)...

Maybe that means we could also not revert yet, and watch the situation.

Personally, I'd vote for revert now, and try other prompt variations.

xingyaoww (approved):

xingyaoww:
@enyst though i think we need to update this in agent-sdk, instead of here -- since i believe the new release of OpenHands could just be OpenHands V1 :)

enyst:
@openhands See the change made by this PR. Then clone the agent-sdk repo from the same org, and apply this same change to the agent prompts in that repo.

Make a PR to agent-sdk .

PR #10832, opened 2025-09-05
feat(runtime): upgrade E2B runtime to v2.0 with full implementation

embersax:
- Upgrade E2B SDK from v1.7.1 to e2b-code-interpreter v2.0.0
  - Implement all required abstract methods (run, run_ipython, edit, browse)
  - Add sandbox lifecycle management with caching for reconnection
  - Support custom E2B domains via E2B_DOMAIN environment variable
  - Fix API compatibility issues with E2B v2
  - Add comprehensive error handling and logging

  This enables full E2B runtime support for running OpenHands agents in
  cloud-based sandboxes, including support for self-hosted E2B instances.

  **End-user friendly description of the problem this fixes or functionality this introduces:**

  This update enables OpenHands to run agents in E2B cloud sandboxes, providing a secure and scalable alternative to Docker. Users can now:
  - Run OpenHands without Docker by using E2B's cloud-based sandboxes
  - Use self-hosted E2B instances for better control and performance
  - Scale evaluations across multiple sandboxes in the cloud
  - Run SWE-bench and other benchmarks at scale using E2B infrastructure

  Simply set your E2B API key and use `RUNTIME=e2b` to get started!

  ---
  **Summarize what the PR does, explaining any non-trivial design decisions:**

  This PR completely overhauls the E2B runtime integration:

  1. **SDK Upgrade**: Migrated from deprecated e2b v1.7.1 to e2b-code-interpreter v2.0.0, which required significant API changes
  2. **Full Implementation**: The previous E2B runtime only supported file operations. This PR implements all required methods:
     - Command execution via native E2B sandbox
     - IPython/Jupyter support using E2B's built-in code interpreter
     - File editing capabilities
     - URL browsing via curl/wget
  3. **Sandbox Lifecycle Management**: Added intelligent caching mechanism that allows reconnecting to existing sandboxes, reducing startup time and costs
  4. **Self-hosted Support**: Added `E2B_DOMAIN` environment variable support for users running E2B on their own infrastructure
  5. **Design Decision**: Instead of running an HTTP action server inside E2B (which would be inefficient), we override the execution methods to directly use E2B's native APIs

  The implementation maintains full compatibility with OpenHands' architecture while leveraging E2B's strengths as a cloud-native sandbox solution.

  ---

  N/A - This is a feature enhancement to complete the E2B runtime implementation.


poetry.lock
```diff
@@ -11845,9 +11861,9 @@ cffi = {version = ">=1.11", markers = "platform_python_implementation == \"PyPy\
 cffi = ["cffi (>=1.11)"]
 
 [extras]
-third-party-runtimes = ["daytona", "e2b", "modal", "runloop-api-client"]
+third-party-runtimes = ["daytona", "modal", "runloop-api-client"]
```
enyst: Sorry, @embersax , could we keep e2b dependencies in `extras`? The idea here is that the user can choose to install the optional extras, it doesn’t get installed by default.

poetry.lock
```diff
@@ -2257,26 +2257,42 @@ files = [
 
 [[package]]
 name = "e2b"
-version = "1.7.0"
+version = "2.0.0"
 description = "E2B SDK that give agents cloud environments"
-optional = true
+optional = false
```
enyst: Same here

embersax: Hi, Thanks for this. I will modify it soon!

enyst:
@openhands You know how to build and run this project. Run make build and make run, to make sure it continues to work normally when built without the optional dependency updated in this PR.

You don’t need to test it, you don’t have a key for it, so we took care of that part.

enyst (approved):
Thank you!

Just a heads-up, we’re just about to make a large migration to [`agent-sdk`](https://github.com/All-Hands-AI/agent-sdk?tab=readme-ov-file#openhands-agent-sdk). I don’t know if we will make a release before that is completed, but it’s intended to be relatively fast. 😅

Issue #10022, opened 2025-07-31
Modernize pyproject.toml format


neubig:
**What problem or use case are you trying to solve?**

`poetry check` gives the following errors:

```
$ poetry check
Warning: [tool.poetry.name] is deprecated. Use [project.name] instead.
Warning: [tool.poetry.version] is set but 'version' is not in [project.dynamic]. If it is static use [project.version]. If it is dynamic, add 'version' to [project.dynamic].
If you want to set the version dynamically via `poetry build --local-version` or you are using a plugin, which sets the version dynamically, you should define the version in [tool.poetry] and add 'version' to [project.dynamic].
Warning: [tool.poetry.description] is deprecated. Use [project.description] instead.
Warning: [tool.poetry.readme] is set but 'readme' is not in [project.dynamic]. If it is static use [project.readme]. If it is dynamic, add 'readme' to [project.dynamic].
If you want to define multiple readmes, you should define them in [tool.poetry] and add 'readme' to [project.dynamic].
Warning: [tool.poetry.license] is deprecated. Use [project.license] instead.
Warning: [tool.poetry.authors] is deprecated. Use [project.authors] instead.
Warning: [tool.poetry.repository] is deprecated. Use [project.urls] instead.
Warning: [tool.poetry.extras] is deprecated. Use [project.optional-dependencies] instead.
Warning: Defining console scripts in [tool.poetry.scripts] is deprecated. Use [project.scripts] instead. ([tool.poetry.scripts] should only be used for scripts of type 'file').
```

PR #10833, opened 2025-09-05
build: Migrate to uv

jakob1379:

**Migrated from Poetry to uv to enable MCP server compatibility and improve dependency isolation**

---
This PR migrates the project’s Python packaging/workflow from Poetry to uv. It replaces all Poetry CLI usages (local dev, Makefile/scripts, and CI) with their uv equivalents to keep behavior the same.

the whys:
 - This should in theory improve install speed, short CI and local setup times.
 - uv has great support for CI caching withtout and advanced setup needed
 - pyproject.toml becomes PEP compliant improving integration with other tools

what changed:
 - all poetry commands are interchanged with uv equivalent
 - CI update with [uv-cache](https://github.com/marketplace/actions/setup-uv-and-handle-its-cache) 
 - docs has been updated where references to poetry were.

design decsisions:
 - I ran through it all, command by command prioritizing individual commands over a complete refactor focusing on reducing overall complexity (this is left as an exercise to the read)
 - kept the dependency groups the same, though I believe we could improve what get installed when.

Testings:
 - locally runs with `act -W .github/workflows/py-tests.yml -j test-on-linux -P blacksmith-4vcpu-ubuntu-2204=catthehacker/ubuntu:act-22.04` without any errors
 - `uv run openhands serve` works
 - `uv run openhands` work
 
 I would love to say that speed is a huge concern with this migration though in my tests between poetry and uv it seems to be about the same, probably due to some builds taking of the majority of the time.
---


jakob1379:
Do I need to squash/fixup my commits or is that not relevant?

.github/workflows/e2e-tests.yml
```diff
@@ -21,16 +21,13 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
 
-      - name: Install poetry via pipx
-        uses: abatilo/actions-poetry@v4
-        with:
-          poetry-version: 2.1.3
+      - uses: hynek/setup-cached-uv@v2
+        name: Setup uv
```
jakob1379, line 25: Automatic cache of uv envrinments, we can add a specific key if we need to cache multiple environments in the same workflow

.github/workflows/e2e-tests.yml
```diff
@@ -69,21 +66,19 @@ jobs:
           ENABLE_BROWSER: true
           INSTALL_PLAYWRIGHT: 1
         run: |
-          # Fix poetry.lock file if needed
-          echo "Fixing poetry.lock file if needed..."
-          poetry lock
+          # Using uv; no lockfile fix needed
```
jakob1379, line 69: should I add `uv lock` here to detect changes?

.openhands/pre-commit.sh
```diff
@@ -143,7 +143,7 @@ if [ "$has_frontend_changes" = false ] && [ "$has_backend_changes" = false ]; th
     echo "No specific code changes detected. Running basic checks..."
     if [ -n "$STAGED_FILES" ]; then
         # Run only basic pre-commit hooks for non-code files
-        poetry run pre-commit run --files $(echo "$STAGED_FILES" | tr '\n' ' ') --hook-stage commit --config ./dev_config/python/.pre-commit-config.yaml
+        .venv/bin/pre-commit run --files $(echo "$STAGED_FILES" | tr '\n' ' ') --hook-stage commit --config ./dev_config/python/.pre-commit-config.yaml
```
jakob1379: are we concerned with installing the full env when we can use `uvx`?

Makefile
```diff
@@ -61,15 +61,6 @@ check-system:
...
 
-check-python:
-	@echo "$(YELLOW)Checking Python installation...$(RESET)"
-	@if command -v python$(PYTHON_VERSION) > /dev/null; then \
-		echo "$(BLUE)$(shell python$(PYTHON_VERSION) --version) is already installed.$(RESET)"; \
-	else \
-		echo "$(RED)Python $(PYTHON_VERSION) is not installed. Please install Python $(PYTHON_VERSION) to continue.$(RESET)"; \
-		exit 1; \
-	fi
-
```
jakob1379, line 72: uv technically handles python if the specified version is not present on the system

Makefile
```diff
@@ -116,61 +116,49 @@ check-tmux:
...
-		export HNSWLIB_NO_NATIVE=1; \
-		poetry run pip install chroma-hnswlib; \
-	fi
-	@if [ -n "${POETRY_GROUP}" ]; then \
-		echo "Installing only POETRY_GROUP=${POETRY_GROUP}"; \
-		poetry install --only $${POETRY_GROUP}; \
-	else \
-		poetry install --with dev,test,runtime; \
+		HNSWLIB_NO_NATIVE=1 uv pip install chroma-hnswlib; \
 	fi
```
jakob1379, line 130: the notion of poetry groups seems sparsely used and somewhat overgeneralizing this step compared to the usages in CI

Makefile
```diff
@@ -116,61 +116,49 @@ check-tmux:
...
+		PLAYWRIGHT_SENTINEL="cache/playwright_chromium_is_installed.$$PLAYWRIGHT_VERSION.txt"; \
 		if [ -f "/etc/manjaro-release" ]; then \
 			echo "$(BLUE)Detected Manjaro Linux. Installing Playwright dependencies...$(RESET)"; \
-			poetry run pip install playwright; \
-			poetry run playwright install chromium; \
+			if [ -z "$$PLAYWRIGHT_BIN_PATH" ] || [ ! -x "$$PLAYWRIGHT_BIN_PATH" ]; then \
+				uv pip install playwright; \
+				PLAYWRIGHT_BIN_PATH="$$(uv run which playwright)"; \
+			fi; \
+			"$$PLAYWRIGHT_BIN_PATH" install chromium; \
```
jakob1379: I might ave overdone this part to ensure existinge and correct version of playwright. If it is I am happy to revert to previous state.

evaluation/benchmarks/lca_ci_build_repair/eval_infer.py
```diff
@@ -117,12 +117,12 @@ def run_eval(
     logger.info(action, extra={'msg_type': 'ACTION'})
     obs = runtime.run_action(action)
 
-    action = CmdRunAction(command='poetry install')
+    action = CmdRunAction(command='pip install -U uv && uv sync')
```
jakob1379, line 120: not sure if i can assume `uv` to be installed here like we had with poetry?

evaluation/benchmarks/lca_ci_build_repair/setup.py
```diff
@@ -47,13 +48,19 @@ def setup():
...
-    # Run poetry install in LCA_CI_PATH
-    print(f"Running 'poetry install' in {lca_ci_path}")
+    # Install uv and sync dependencies in LCA_CI_PATH
+    print(f"Running 'uv sync' in {lca_ci_path}")
+    # Ensure 'uv' is available, then synchronize project dependencies
+    install_uv = subprocess.run(
+        [sys.executable, '-m', 'pip', 'install', '-U', 'uv'], capture_output=True, text=True
+    )
+    if install_uv.returncode != 0:
+        print(f'Warning installing uv: {install_uv.stderr}')
```
jakob1379: Same for this - i am unsure if I can assume uv to be available here.

openhands/runtime/utils/command.py
```diff
@@ -7,8 +7,6 @@
     'run',
     '-n',
     'openhands',
-    'poetry',
-    'run',
 ]
```
jakob1379, line 14: unsure if `uv run` is needed here too?

.github/workflows/e2e-tests.yml
xingyaoww: maybe yes so we keep things the same?

evaluation/benchmarks/lca_ci_build_repair/eval_infer.py
xingyaoww: This action seems to get executed "inside their docker", so i'd suggest we can leave these "poetry" code as is for evaluation benchmarks :)

evaluation/benchmarks/lca_ci_build_repair/setup.py
xingyaoww: i think it is safe to revert and ignore these

containers/app/Dockerfile
```diff
@@ -76,7 +73,7 @@ COPY --chown=openhands:openhands --chmod=770 --from=backend-builder ${VIRTUAL_EN
 COPY --chown=openhands:openhands --chmod=770 ./microagents ./microagents
 COPY --chown=openhands:openhands --chmod=770 ./openhands ./openhands
 COPY --chown=openhands:openhands --chmod=777 ./openhands/runtime/plugins ./openhands/runtime/plugins
-COPY --chown=openhands:openhands pyproject.toml poetry.lock README.md MANIFEST.in LICENSE ./
+COPY --chown=openhands:openhands pyproject.toml README.md MANIFEST.in LICENSE ./
```
xingyaoww: Should we copy over uv.lock too, just make sure we are installing from a locked version?

containers/app/Dockerfile
```diff
@@ -15,18 +15,15 @@ FROM base AS backend-builder
...
-    POETRY_VIRTUALENVS_IN_PROJECT=1 \
-    POETRY_VIRTUALENVS_CREATE=1 \
-    POETRY_CACHE_DIR=/tmp/poetry_cache
+ENV UV_CACHE_DIR=/tmp/uv_cache
 
 RUN apt-get update -y \
-    && apt-get install -y curl make git build-essential jq gettext \
-    && python3 -m pip install poetry --break-system-packages
+    && apt-get install -y curl make git build-essential jq gettext ca-certificates \
+    && pip install uv \
```
xingyaoww: Shall we use `RUN curl -LsSf https://astral.sh/uv/install.sh | sh` here?

evaluation/benchmarks/lca_ci_build_repair/eval_infer.py
xingyaoww: Actually we can leave this as is -- we install `uv` in the runtime Docker image, so it should be available.

openhands/runtime/utils/command.py
xingyaoww: I think we should ideally keep it to be safe 🤔

openhands/runtime/utils/runtime_build.py
```diff
@@ -285,12 +285,20 @@ def prep_build_folder(
...
-            src = Path(project_root, file)
-        shutil.copy2(src, Path(build_folder, 'code', file))
+    # Copy pyproject.toml and uv.lock if present
+    # Always copy pyproject.toml
+    src = Path(openhands_source_dir, 'pyproject.toml')
+    if not src.exists():
+        src = Path(project_root, 'pyproject.toml')
+    shutil.copy2(src, Path(build_folder, 'code', 'pyproject.toml'))
+
+    # Copy uv.lock if available
```
xingyaoww: Let's assume `uv.lock` always exists since we will generate them?

openhands/runtime/utils/runtime_build.py
```diff
@@ -324,7 +332,16 @@ def get_hash_for_lock_files(base_image: str, enable_browser: bool = True) -> str
     # Only include enable_browser in hash when it's False for backward compatibility
     if not enable_browser:
         md5.update(str(enable_browser).encode())
-    for file in ['pyproject.toml', 'poetry.lock']:
+    # Hash pyproject.toml and uv.lock if present
+    files_to_hash: list[str] = ['pyproject.toml']
+    candidate = 'uv.lock'
```
xingyaoww: ditto

xingyaoww (commented):
Overall LGTM! Just need some minor fixes & pass CIs we should be good to merge!

We will be removing the agent components (including playwright) from this repo for V1 anyway (those will be handled by [agent-sdk](https://github.com/All-Hands-AI/agent-sdk)), and accordingly we will move all the evaluation code to a different repo and rewrite most of them -- so don't worry about breaking them in this migration :)

evaluation/benchmarks/lca_ci_build_repair/eval_infer.py
jakob1379: @xingyaoww so sholud I discard this edit or leave it as is (as uv)?

containers/app/Dockerfile
jakob1379: I think, yes. Keep thing as close to original as possible

containers/app/Dockerfile
jakob1379: uv was not in path when installed in the same layer as apt-get but I will try the RUN curl in a separate layer and see if it works

raymyers:
> Do I need to squash/fixup my commits or is that not relevant?

We squash on merge so that shouldn't be an issue.

evaluation/benchmarks/lca_ci_build_repair/eval_infer.py
xingyaoww: you can leave it as is (uv)

jakob1379:
fixes https://github.com/All-Hands-AI/OpenHands/issues/10022

neubig:
Hello and thanks again for the contribution @jakob1379 ! Are we still working on this or are we going to maybe close and regroup once we switch over to using the agent sdk (which is also managed through uv).

jakob1379:
Fine with me, its not really an issue for me ☺️

xingyaoww:
Happy to take over from here and start a new PR to see if we can push this towards finish line!! appreciate your efforts so far! @jakob1379

xingyaoww:
@OpenHands check the github actions in https://github.com/All-Hands-AI/OpenHands/pull/10833 and help me fix the two linter erros

tofarr:
Thank you for this - This is 100% on our radar. We have already migrated over to UV in the agent sdk. Given that there are a substantial number of merge conflicts, linting and test issues, I am going to leave this open for now with a view to resolving the issues after we release the V1 app server.

PR #10848, opened 2025-09-05
More Type Safety

tofarr:

The get_impl method was not type checked, and now it is.

---
The `lru_cache` is not type safe, so we wrap it.

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:edadd1f-nikolaik   --name openhands-app-edadd1f   docker.all-hands.dev/all-hands-ai/openhands:edadd1f
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@fix-get-impl-type-hint openhands
```


openhands/server/routes/manage_conversations.py
```diff
@@ -397,7 +396,7 @@ async def get_prompt(
     )
 
     prompt_template = generate_prompt_template(stringified_events)
-    prompt = generate_prompt(llm_config, prompt_template, conversation_id)
```
raymyers, line 400: I assume there are no other uses to change

tofarr: Correct

PR #10770, opened 2025-09-02
Enterprise code and docker build

raymyers:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

* Copying enterprise code unchanged to enterprise/
* Give enterprise it's own Makefile and dev_config, keeping linting config separate for now
* Build docker image as `enterprise-server`
* Add CI jobs to test and lint enterprise
* Allow preview environments by adding `deploy` label, which invokes a private job on the image built here

---
**Link of any specific issues this addresses:**

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:be7c85c-nikolaik   --name openhands-app-be7c85c   docker.all-hands.dev/all-hands-ai/openhands:be7c85c
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@ray/enterprise-1 openhands
```

PR #10862, opened 2025-09-05
chore - Make enterprise preview work when labeled after the fact

raymyers:

---

The enterprise-preview job was added to `ghcr-build.yml` in #10770. It's intended when PR is created with `deploy` label or when the label is after the fact. The after-the-fact case did not work because the default `on: pull_request:` trigger does not included the `labeled` action. [GitHub Workflow Docs](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request)

This workaround adds a new workflow with a small amount of duplication (probably not worth extracted). An alternative would be to add the `labeled` action to `ghcr-build.yml` but that would be more complicated because many conditionals would need to be added to prevent running all existing steps on every label

If we wanted to avoid the duplication we could use a [reusable workflow](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows).

---

### DIFF
--- a/.github/workflows/enterprise-preview.yml
+++ b/.github/workflows/enterprise-preview.yml
@@ -1,0 +1,30 @@
+# Feature branch preview for enterprise code
+name: Enterprise Preview
+
+# Run on PRs labeled
+on:
+  pull_request:
+    types: [labeled]
+
+# Match ghcr-build.yml, but don't interrupt it.
+concurrency:
+  group: ${{ github.workflow }}-${{ (github.head_ref && github.ref) || github.run_id }}
+  cancel-in-progress: false
+
+jobs:
+  # This must happen for the PR Docker workflow when the label is present,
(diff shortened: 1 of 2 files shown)


openhands-agent:
Feature deployment is available at: https://ohpr-10862.staging.all-hands.dev
Redirect URI for GitHub App is https://ohpr-10862.auth.staging.all-hands.dev/realms/allhands/broker/github/endpoint
Last updated: Sep 05, 2025, 06:44:04 PM ET

openhands-agent:
Feature deployment is available at: https://ohpr-10862.staging.all-hands.dev
Redirect URI for GitHub App is https://ohpr-10862.auth.staging.all-hands.dev/realms/allhands/broker/github/endpoint
Last updated: Sep 05, 2025, 06:51:23 PM ET

Issue #10874, opened 2025-09-08
[Bug]: RollingLogger move_back method ignoring amount parameter


caocao-sys:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

# Bug Description:
There is a bug in the [move_back] method of the [RollingLogger] class in /openhands/core/logger.py. The method accepts an [amount]  parameter to specify how many lines to move the cursor back, but it incorrectly uses self.max_lines instead of the calculated [amount] value in the actual cursor movement operation.

# Code Location:
openhands/core/logger.py
Method: RollingLogger.move_back()

# Expected Behavior:
When calling [move_back(5)], the cursor should move back exactly 5 lines. 

# Actual Behavior:
When calling [move_back(5)], the cursor moves back self.max_lines lines instead of 5 lines, because the method ignores the [amount]  parameter in the actual implementation.

# Reproduction Steps:
1. Create a [RollingLogger] instance with max_lines=10
2. Call [move_back(3)] 
3. Observe that the cursor moves back 10 lines instead of 3 lines

# Impact:
This bug affects the debugging experience when using the rolling logger, as the cursor positioning becomes incorrect when custom amounts are specified.

### OpenHands Installation

Docker command in README

### OpenHands Version

main

### Model Name

claude-3-7-sonnet-20250219

### Operating System

MacOS

mamoodi:
Looks like this can be closed based on the merged PR. Thanks for putting in the fix!!

PR #10875, opened 2025-09-08
fix: RollingLogger move_back method ignoring amount parameter (#10874)

caocao-sys:
fix: RollingLogger move_back method ignoring amount parameter (#10874)


enyst (approved):
Thank you!

PR #10878, opened 2025-09-08
refactor(frontend): restrict message sending when runtime is stopped

hieptl:

On the conversation page, message sending should be disabled whenever the runtime is stopped.

**Acceptance Criteria:**
- Users are prevented from sending messages while the runtime is stopped.

---

The PR restricts message sending when runtime is stopped

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:32ff027-nikolaik   --name openhands-app-32ff027   docker.all-hands.dev/all-hands-ai/openhands:32ff027
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@hieptl/all-3489 openhands
```

### DIFF
--- a/frontend/src/components/features/chat/interactive-chat-box.tsx
+++ b/frontend/src/components/features/chat/interactive-chat-box.tsx
@@ -147,9 +147,13 @@ export function InteractiveChatBox({
     handleSubmit(suggestion);
   };
 
+  const isRuntimeStopped =
+    conversation?.status === "STOPPED" && !conversation.runtime_status;
+
   const isDisabled =
     curAgentState === AgentState.LOADING ||
-    curAgentState === AgentState.AWAITING_USER_CONFIRMATION;
+    curAgentState === AgentState.AWAITING_USER_CONFIRMATION ||
+    isRuntimeStopped;
 
   return (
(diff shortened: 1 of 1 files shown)


amanape:
I don’t think we should disable the input in this state. We should disable submitting the message, yes, but we should still allow the user to type into the input

hieptl:
> I don’t think we should disable the input in this state. We should disable submitting the message, yes, but we should still allow the user to type into the input

Hi @amanape,

Your feedback has been addressed in this PR (https://github.com/All-Hands-AI/OpenHands/pull/10879). With this change, users will be able to type their messages even when the disabled prop is set to true.

Thank you! 🙏

amanape:
So we can close this PR?

hieptl:
Yes, I reviewed the Notion feedback again and I don’t think this PR is necessary. Thank you! 🙏

PR #10888, opened 2025-09-09
Fix Docker build error 'groupadd: GID 1000 already exists'

ryanhoangt:

---

- Problem: Some base images have existing users/groups with UID/GID 1000, causing the openhands user creation to fail.
- Solution:
  - Keep existing cleanup logic for known problematic users (`pn`, `ubuntu`)
  - Add fallback to use system-assigned IDs when 1000 is already taken
  - Maintains preferred UID/GID 1000 when available, gracefully adapts when not

---


xingyaoww (approved):
LGTM! Let's try this!

PR #10898, opened 2025-09-09
feat: Add Cygnal integration

eliotjones1:

Adds Gray Swan's [Cygnal](https://www.grayswan.ai/product/cygnal) monitor to OpenHands as a new option for the `SecurityAnalyzer`.

---
- Adds functionality to pass the event stream into the `SecurityAnalyzer` base class (so analyzers can track conversation history)
- Integrates the `SecurityAnalyzer` class with Gray Swan's Cygnal [API](https://docs.grayswan.ai/monitor-requests/monitor)

---

### DIFF
--- a/openhands/runtime/base.py
+++ b/openhands/runtime/base.py
@@ -199,10 +199,7 @@ class Runtime(FileEditRuntimeMixin):
                 self.config.security.security_analyzer, SecurityAnalyzer
             )
             self.security_analyzer = analyzer_cls()
-            if self.event_stream and hasattr(
-                self.security_analyzer, 'set_event_stream'
-            ):
-                self.security_analyzer.set_event_stream(self.event_stream)
+            self.security_analyzer.set_event_stream(self.event_stream)
             logger.debug(
                 f'Security analyzer {analyzer_cls.__name__} initialized for runtime {self.sid}'
             )
--- a/openhands/security/grayswan/analyzer.py
+++ b/openhands/security/grayswan/analyzer.py
@@ -2,7 +2,7 @@
 
 import asyncio
 import os
-from typing import Any, Optional
+from typing import Any
 
 import aiohttp
 from fastapi import Request
@@ -10,6 +10,7 @@ from fastapi import Request
 from openhands.core.logger import openhands_logger as logger
 from openhands.events.action.action import Action, ActionSecurityRisk
 from openhands.events.event_store_abc import EventStoreABC
+from openhands.memory.view import View
 from openhands.security.analyzer import SecurityAnalyzer
 from openhands.security.grayswan.utils import convert_events_to_openai_messages
 
@@ -22,7 +23,10 @@ class GraySwanAnalyzer(SecurityAnalyzer):
         history_limit: int = 20,
         max_message_chars: int = 30000,
         timeout: int = 30,
-        session: Optional[aiohttp.ClientSession] = None,
+        low_threshold: float = 0.3,
+        medium_threshold: float = 0.7,
+        high_threshold: float = 1.0,
+        session: aiohttp.ClientSession | None = None,
     ) -> None:
         """Initialize GraySwan analyzer.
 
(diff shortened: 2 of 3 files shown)


openhands-agent:
[OpenHands](https://github.com/All-Hands-AI/OpenHands) started fixing the pr! You can monitor the progress [here](https://github.com/All-Hands-AI/OpenHands/actions/runs/17588132782).

csmith49:
Failing CI will be fixed by #10901

openhands/runtime/base.py
```diff
@@ -199,6 +199,10 @@ def __init__(
                 self.config.security.security_analyzer, SecurityAnalyzer
             )
             self.security_analyzer = analyzer_cls()
+            if self.event_stream and hasattr(
+                self.security_analyzer, 'set_event_stream'
+            ):
```
csmith49: I'm pretty sure that because the base security analyzer has a `set_event_stream` method this conditional will always be true.

openhands/security/grayswan/utils.py
```diff
@@ -0,0 +1,145 @@
...
+from openhands.events.observation.files import (
+    FileEditObservation,
+    FileReadObservation,
+    FileWriteObservation,
+)
+from openhands.events.observation.mcp import MCPObservation
+from openhands.events.observation.observation import Observation
+
+
+def convert_events_to_openai_messages(events: list[Any]) -> list[dict[str, Any]]:
```
csmith49, line 23: We have a `ConversationMemory` object [here](https://github.com/All-Hands-AI/OpenHands/blob/b08238c841478ef44f1d2f491eda30f95efbe02f/openhands/memory/conversation_memory.py#L74) that does a very similar conversion for the agent. There's definitely enough custom logic here to warrant a second approach and I'm not sure it'd work here but just in case you hadn't seen it!

openhands/security/grayswan/analyzer.py
```diff
@@ -0,0 +1,197 @@
...
+
+class GraySwanAnalyzer(SecurityAnalyzer):
+    """Security analyzer using GraySwan's Cygnal API for AI safety monitoring."""
+
+    def __init__(
+        self,
+        history_limit: int = 20,
+        max_message_chars: int = 30000,
+        timeout: int = 30,
+        session: Optional[aiohttp.ClientSession] = None,
```
csmith49: Minor nit, but we prefer `* | None` to `Optional[*]`.

openhands/security/grayswan/analyzer.py
```diff
@@ -0,0 +1,197 @@
...
+            max_message_chars: Max characters for conversation processing
+            timeout: Request timeout in seconds
+            session: Optional pre-configured session (mainly for testing)
+
+        Environment Variables:
+            GRAYSWAN_API_KEY: Required API key for GraySwan authentication
+            GRAYSWAN_POLICY_ID: Required policy ID for GraySwan policies
+            GRAYSWAN_LOW_THRESHOLD: Risk threshold for LOW classification (default: 0.7)
+            GRAYSWAN_MEDIUM_THRESHOLD: Risk threshold for MEDIUM classification (default: 0.9)
+            GRAYSWAN_HIGH_THRESHOLD: Risk threshold for HIGH classification (default: 1.0)
```
csmith49: Any particular reason to keep these as environment variables instead of some config-level arg like `history_limit`?

openhands/security/grayswan/analyzer.py
```diff
@@ -0,0 +1,197 @@
...
+            logger.warning('No event stream available for GraySwan analysis')
+            return ActionSecurityRisk.UNKNOWN
+
+        try:
+            all_events = list(self.event_stream.get_events())
+            recent_events = (
+                all_events[-self.history_limit :]
+                if len(all_events) > self.history_limit
+                else all_events
+            )
```
csmith49: I think the assumption being made here is that the events in the stream are exactly what the LLM in the agent sees -- but that's not quite correct! Our context management system acts between the event stream and the agent to trim extra events, insert summaries, and sometimes mask out big observations.

If you want to be a little closer to what the agent's LLM is actually seeing you might want to use the `View` construct (defined [here](https://github.com/All-Hands-AI/OpenHands/blob/main/openhands/memory/view.py)):

```
all_events = View.from_events(self.event_stream.get_events())
```

That will apply the bulk of the context management. But if my assumption is wrong this change isn't necessary at all.

csmith49 (commented):
A few minor nits and notes, but otherwise this looks good! The issue that was causing the enterprise linting error should be fixed as well if you can rebase from our new `main`.

openhands/security/grayswan/utils.py
eliotjones1: Yeeeeah I think updating to use `ConversationMemory` is probably touching more parts of the code than is desirable, so going to stick with this -- though I hadn't seen it before and am glad I looked into it!

PR #10905, opened 2025-09-09
Refactor(V1): OpenHands CLI + Agent SDK

malhotra5:

---

## Refactored CLI

CLI that uses `agent-sdk`. This is maintained as a separate project with its own package management. It implements the following functionality 

- Run agent
- Visualize actions and observations
- Pausing/resume agent
- Confirmation mode
- Build to package CLI into a standalone executable binary

Technical details includes
- Bias towards threading rather than async
- Enforce separation of concerns by handling user inputs from terminal and conversation state machine separately
- User actions are isolated in `user_actions/` package
- Listeners are separated into `listeners/` package
- Main logic is cleanly separated from setup, TUI, and runner components

## GitHub CI 
Implements the following CI specific to OpenHands CLI

- Linting (same configuration as main repository)
- Unit tests (run as a separate job inside existing unit test CI)
- Executable Build (run in dedicated CI run, only for CLI changes)

## Next steps

- Integrate the local file store into this new CLI (make sure dependencies introduced are minimal) 
- Implement support for settings, initializing repository, and MCP
- Eventually remove legacy CLI from `openhands/cli` folder

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:a4673dc-nikolaik   --name openhands-app-a4673dc   docker.all-hands.dev/all-hands-ai/openhands:a4673dc
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@feature/integrate-cli openhands
```


llamantino:
Since this looks like a quick refactor to get the agent_sdk working, just keep this for later. It's certainly more important to get it running again first.

Essentially, I wanted to suggest making the CLI a single prompt_toolkit app because of these three issues:
1. The Ctrl+C handling that lacks global coverage - the ctrl+c listener is active only when the agent is running, and if you press ctrl+c before that (i.e., right after pressing enter) the CLI will exit.
2. The CPR warnings, caused by overlapping prompt_toolkit apps (the prompt in prompt_for_next_task and the menu in cli_confirm) - after some recent PRs you can see those warnings right after pressing Ctrl+C when the agent is running.
3. The "Executor shutdown has been called" error on Windows, caused by the menu running in a separate thread.

With a single prompt_toolkit application, ctrl+c would be handled globally, CPR warnings wouldn't occur (no raw/cooked terminal mode switch), and the Executor shutdown error wouldn't happen since no extra threads would be needed. I'd suggest exploring the idea in the future.

That said, if you still want to address those three issues:
1. You could add a global listener with a separate code path for Windows (otherwise, on Windows, you lose input when the prompt and the listener are both active).
2. You could put a lock in the prompt_for_next_task function to prevent multiple prompts from spawning, which I hope should at least fix the CPR warning after you pause the agent.
3. You could filter out the Executor shutdown error on Windows (I haven't seen it anywhere else), since it doesn't seem to impact functionality.

Hope this helps!

PR #10908, opened 2025-09-10
fix: improve CLI help and version command performance

### LINKED ISSUE #10698: feat: use lazy imports for the cli to speed up the ui

Calling `openhands --help` takes around $20s$ 

using the [PEP690](https://peps.python.org/pep-0690/) adding `importlib.set_lazy_imports()` we can defer loading of any modules to whenever a command is invoked, making the `--help` near instant.

kashodiya:
This PR addresses issue #10698 by implementing a fast path for the CLI help and version commands.

## Problem
Running `openhands --help` was taking around 20 seconds due to loading the entire codebase.

## Solution
The approach is to create a simplified version of the CLI parser that only includes the necessary options for displaying help and version information, avoiding loading the full OpenHands codebase.

This is implemented by:
1. Creating a new `fast_help.py` module with a lightweight CLI parser implementation
2. Modifying the entry point to use this fast path for help and version commands
3. Deferring imports of heavy dependencies until they're actually needed

## Performance Improvement
- Before: ~20 seconds to run `openhands --help`
- After: ~0.02 seconds to run `openhands --help`

Fixes #10698

### DIFF
--- a/openhands/cli/entry.py
+++ b/openhands/cli/entry.py
@@ -2,31 +2,23 @@
 
 import sys
 
+# Import only essential modules for CLI help
+# Other imports are deferred until they're actually needed
 import openhands
 import openhands.cli.suppress_warnings  # noqa: F401
-from openhands.cli.gui_launcher import launch_gui_server
-from openhands.cli.main import run_cli_command
-from openhands.core.config import get_cli_parser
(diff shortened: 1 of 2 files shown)


rbren:
Thank you @kashodiya !

PR #10909, opened 2025-09-10
Fix runtime init

chuckbutkus:
Fix a bug introduced when fixing the setup to use a non-root user in the runtimes.

I had moved the user creation before the workspace directory creation as we wanted to set ownership of the workspace directory to the created user.  However, I did not correctly handle the case where the user already exists as it was skipping the workspace directory creation.

This fixes that error.


raymyers (approved):

rbren (approved):

tofarr (approved):
LGTM!

PR #10910, opened 2025-09-10
Fix the loop issue caused by exceeding the length limit

CLFutureX:
When the events that need to be retained have exceeded the model limits, a dead loop may occur at this point.
image


CLFutureX:
@csmith49    hey, please cc

Issue #10912, opened 2025-09-10
[Bug]: Function apply_patch: Rename detection incorrectly applies to all diffs in patch


Papadopulus:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

There is a bug in the apply_patch function located in openhands/resolver/send_pull_request.py at line 62, related to rename operation detection.

**Current Behavior**
The function checks:
```
if "rename from" in patch:
```
This means the entire patch string is scanned for "rename from", rather than checking the specific diff currently being processed.

**Why This Is a Problem**
If a patch contains both a rename operation and a regular edit, the condition will return True for all diffs in the patch. This causes normal edit diffs to be incorrectly treated as rename operations.

**Example Problematic Patch**
```
diff --git a/file1.txt b/file2.txt
similarity index 100%
rename from file1.txt
rename to file2.txt
diff --git a/file3.txt b/file3.txt
index 17d5d9e..bc8b8aa 100644
--- a/file3.txt
+++ b/file3.txt
@@ -1 +1,2 @@
 another file
+modified content
```
In this case, the edit operation for file3.txt will also be treated as a rename because "rename from" exists somewhere in the overall patch.

**How to Reproduce**
Run the project using:
```
openhands/resolver/resolve_issue.py
```
Then run :
```
openhands/resolver/send_pull_request.py
```
The issue occurs in openhands/resolver/send_pull_request.py during patch application.
While OpenHands is not normally executed in this way, the issue can be reproduced when testing patch application directly.

### OpenHands Installation

Other

### OpenHands Version

0.56.0 - 2025-09-09

### Model Name

claude-sonnet-4-20250514

### Operating System

Linux

PR #10913, opened 2025-09-10
Fix: Correct rename detection in apply_patch to check per-diff instead of full patch

Papadopulus:
Ensure 'rename from' is checked per diff (diff.text) instead of the entire patch, preventing edit diffs from being misclassified as renames.

Fixes: #10912

Fixes incorrect handling of rename operations in patch application. Previously, normal file edits could be mistakenly treated as renames when a patch contained both a rename and edits. With this fix, only the intended rename diffs are recognized as such.

---
1. Updated apply_patch in openhands/resolver/send_pull_request.py
2. "rename from" is now checked in the current diff.text instead of the entire patch string
3. Ensures proper distinction between rename and edit diffs
4. Keeps logic simple and localized without introducing new abstractions

---
Closes: #109112


mamoodi:
@malhotra5 mind taking a look at this PR?

PR #10925, opened 2025-09-10
Refactor(V1): Create project for shared types/configurations for server and CLI

malhotra5:

---

This PR creates a separate project to share types and configurations between the main openhands server and CLI. 

By doing this, we can decouple the dependencies between the CLI and server while still ensure that the types/configurations are shared. This helps us ensure that the CLI is as lightweight as possible.

---

### DIFF
--- a/.github/workflows/cli-build-test.yml
+++ b/.github/workflows/cli-build-test.yml
@@ -1,0 +1,83 @@
+# Workflow that builds and tests the CLI binary executable
+name: CLI - Build and Test Binary
+
+# Run on pushes to main branch and all pull requests, but only when CLI files change
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'openhands-cli/**'
+  pull_request:
+    paths:
+      - 'openhands-cli/**'
+
+# Cancel previous runs if a new commit is pushed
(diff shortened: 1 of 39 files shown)


configuration/openhands_configuration/settings.py
```diff
@@ -125,62 +114,12 @@ def secrets_store_serializer(self, secrets: UserSecrets, info: SerializationInfo
         """Force invalidate secret store"""
         return {'provider_tokens': {}}
 
-    @staticmethod
-    def from_config() -> Settings | None:
```
malhotra5, line 129: We don't seem to be using this method anymore. it also relies on `OpenHandsConfig` type which is exclusive to the app server so getting rid of this avoid the complexity of figuring out separation of concerns

malhotra5:
The goal of this PR was to share `UserSettings` between CLI and app server. Given that `UserSettings` contains configurations for agent, app server, and runtime specific things, I don't think this PR makes sense anymore.

The CLI requires just requires LLMConfig and MCPConfig - these configs have been moved into the agent-sdk and seems like the sdk will hold logic for loading these in from various locations.

PR #10946, opened 2025-09-11
chore - CI check migrations are in sync and warn

raymyers:

---

The goal is to avoid migration conflicts. 

The fool-proof way is to require all PRs to be sync'd before merge, but that affects all non-migration PRs in an impractical way because of PR frequency and build time. Instead we introduce two simple measures:

* Fail PRs with enterprise db migrations when they are not up to date with main - when the pipeline runs.
* Comment to warn the user to manually sync before merge.

This should make migration conflicts sufficiently rare without slowing down the PR process.

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:5ccd5ec-nikolaik   --name openhands-app-5ccd5ec   docker.all-hands.dev/all-hands-ai/openhands:5ccd5ec
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@ray/migration-check openhands
```


raymyers:
# Test note
After adding changes to the migrations dir, we got a comment and failure (out of date).

After syncing, we get no failure and a warning.

In subsequent runs the comment is updated, not duplicated.

PR #48, opened 2024-03-18
Minimal Docker Sandbox with GPT-3.5 Execution Example

xingyaoww:
A minimalistic implementation of Docker Sandbox with less than 100 LOC and requires only built-in libraries. It exposes a `.execute` API to run arbitrary bash commands inside the docker container.

It requires `docker` to be installed in the system, and, will potentially need some adjustment when switch to other OS. I hope it could be a potential starting point for future development.

Run it: `python3 opendevin/sandbox/docker.py`

Example screenshot:

image

PR #10976, opened 2025-09-15
Port test improvements from OpenHands-CLI PR #48

malhotra5:
> Test conversation runner using a fake Agent implementation rather than a mock of conversation. These test doubles are probably reusable and if they work out we could move them to a project level util or even in the SDK.
>
> This approach exercises some of the behavior of Conversation along with Conversation Runner, so we might say this is a test of the two combined, focusing on Runner, relative to "contract" of Agent. That was just a pragmatic decision because making Conversation.run behave realistically was harder, maybe indicating a testability issue.
>
> The only remaining mock is to patch _handle_confirmation_request in ConversationRunner. I don't like mocking the unit under test, and an alternative would be make that behavior injectable ([inversion of control](https://martinfowler.com/bliki/InversionOfControl.html)). But admittedly it's working well enough as-is.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:26b4146-nikolaik   --name openhands-app-26b4146   docker.all-hands.dev/all-hands-ai/openhands:26b4146
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@port-cli-test-improvements openhands
```

### DIFF
--- a/openhands-cli/tests/test_conversation_runner.py
+++ b/openhands-cli/tests/test_conversation_runner.py
@@ -1,48 +1,80 @@
-from types import SimpleNamespace
-from unittest.mock import MagicMock, patch
+from typing import Any, Self
+from unittest.mock import patch
 
 import pytest
+from openhands.sdk import Conversation, ConversationCallbackType
+from openhands.sdk.agent.base import AgentBase
+from openhands.sdk.conversation import ConversationState
+from openhands.sdk.llm import LLM
+from pydantic import ConfigDict, SecretStr, model_validator
+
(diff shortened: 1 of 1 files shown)


raymyers:
Approved, sorry for the delay, hopefully merge conflicts aren't too bad

PR #10993, opened 2025-09-16
fix(openhands/runtime/base.py): adjust org-repo logic

wolfspyre:
adjust org-repo detection logic to better handle sub-groups which may or may not have their own openhands-config repos

- added tests to cover the fall through logic. 
- updated documentation to be a little clearer about how microagents are detected and used.

- [ X ] This change is worth documenting at https://docs.all-hands.dev/
      
- [ x ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**Enhance the robustness of org-repository microagent handling**

---

**Summary**:  This change ensures that when working on/in a project that is nested in the upstream code host, OpenHands will iteratively search the codehost URI hierarchy looking for an org-config repository. 

- for each directory path in the hierarchy, openhands will:
   - look in the codehost for an org-config repo
   - look in the workspace for a local directory following the expected naming convention of `org_openhands_groupname`
    - if a local directory is found
      - try to update it from git upstream.
        - if pull fails, warn, but leave as is. likely intentional.
    - inspect local copy of repo (freshly downloaded or not) for a `microagents` directory. 
    - if found, and has microagents, stop iterating.

### DIFF
--- a/docs/usage/prompting/microagents-org.mdx
+++ b/docs/usage/prompting/microagents-org.mdx
@@ -1,35 +1,107 @@
 ---
 title: Organization and User Microagents
-description: Organizations and users can define microagents that apply to all repositories belonging to the organization or user.
+description: An OpenHands configuration repository allows you to define microagents that apply to all adjacent or descendent-pathed repositories.
 ---
 
 ## Usage
 
-These microagents can be [any type of microagent](./microagents-overview#microagent-types) and will be loaded
-accordingly. However, they are applied to all repositories belonging to the organization or user.
+Organization microagents may be any type of microagent. (see [Microagent Types](./microagents-overview#microagent-types))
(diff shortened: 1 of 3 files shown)


enyst:
Thank you for this contribution @wolfspyre , and I apologize we didn't get to it in time. One of the reasons is that we are just transitioning to OpenHands V1, the next major version of OpenHands, which has a very different core architecture. It's built on [Agent-SDK](https://github.com/OpenHands/software-agent-sdk), which does some of lifting with git in its `workspace` package. I'm not sure how much is affected from your PR, but unfortunately we can't merge it as it is.

If you want, could you maybe make a proposal in an issue and/or in slack on what it intends to do and why, to discuss it with people working on git providers, for V1.

For V0, I'm sorry we can't merge this one, but thank you for the contribution!

PR #11013, opened 2025-09-18
Allow custom user

chuckbutkus:

---

---


chuckbutkus:
@OpenHands please fix the failing actions on PR #11013 at branch `allow-custom-user`

openhands/runtime/plugins/vscode/__init__.py
```diff
@@ -15,6 +15,16 @@
...
+RUNTIME_USERNAME = os.getenv('RUNTIME_USERNAME')
+SU_TO_USER = os.getenv('SU_TO_USER', 'true').lower() in (
+    '1',
+    'true',
+    't',
+    'yes',
+    'y',
+    'on',
+)
+
```
rbren, line 27: This should probably go into AppConfig rather than being grabbed directly via env var

enterprise/server/saas_nested_conversation_manager.py
```diff
@@ -60,9 +60,14 @@
...
 )
+RUNTIME_ROUTING_MODE = os.getenv('RUNTIME_ROUTING_MODE', 'subdomain').lower()
 
 # Pattern for base URL for the runtime
-RUNTIME_CONVERSATION_URL = RUNTIME_URL_PATTERN + '/api/conversations/{conversation_id}'
+RUNTIME_CONVERSATION_URL = RUNTIME_URL_PATTERN + (
+    '/runtime/api/conversations/{conversation_id}'
+    if RUNTIME_ROUTING_MODE == 'path'
+    else '/api/conversations/{conversation_id}'
+)
```
rbren, line 70: I'd suggest setting `RUNTIME_URL_PATTERN` instead of changing the pattern again here--seems like you could just set it to `$host/runtime/`

chuckbutkus: The RUNTIME_URL_PATTERN is used to generate URLs for other things (/vscode, /work-1, /work-2) in addition to /runtime.  So it would require some other refactoring to set the pattern to RUNTIME_URL_PATTERN to /runtime

openhands/runtime/plugins/vscode/__init__.py
chuckbutkus: I will look into it.

rbren (approved):

Issue #11046, opened 2025-09-20
[Bug]: WORKSPACE_MOUNT_PATH cannot be unset


kripper:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Issues:
- docker-compose.yml sets a default value when WORKSPACE_MOUNT_PATH is empty. It's currently not possible to leave it empty.
- Also, it's not possible to pass the new SANDBOX_VOLUMES env var.

This affects `make docker-run` and also when running `docker compose up` directly.

### OpenHands Installation

Docker command in README

### Operating System

None

PR #11047, opened 2025-09-20
Fix https://github.com/All-Hands-AI/OpenHands/issues/11046

kripper:

Fixes this issues:
- docker-compose.yml sets a default value when WORKSPACE_MOUNT_PATH is empty. It's currently not possible to leave it empty.
- Also, it's not possible to pass the new SANDBOX_VOLUMES env var.

This affects `make docker-run` and also when running `docker compose up` directly.

---

It allows running docker-compose with empty `WORKSPACE_MOUNT_PATH` and also provde `SANDBOX_VOLUMES`.


tofarr:
The current runtime is being replaced with the Agent Server (The new V1 API), so I am closing this for now. Please reach out if you feel this is an error

kripper:
> The current runtime is being replaced with the Agent Server (The new V1 API), so I am closing this for now. Please reach out if you feel this is an error

I don't think it's related.

kripper:
@tofarr How do you (or others) run OpenHands with `WORKSPACE_MOUNT_PATH = ""` (with an empty value), so that OpenHands uses a workspace inside the container instead of mounting a workspace shared between all containers?

Without this patch, it's impossible to use `docker compose up` because `docker-composer.yml` will always set a value for `WORKSPACE_MOUNT_PATH`.

PR #11051, opened 2025-09-20
fix(ssl): use default SSL context for all httpx calls; add opt-out verify for debugging

phymbert:
## Summary
While testing OpenHands behind enterprise TLS interception/self-signed corporate roots, I hit `SSL: CERTIFICATE_VERIFY_FAILED`. Many orgs install a private root CA in the OS trust store; however several `httpx` clients weren’t explicitly using a default SSL context.

This PR ensures **all** `httpx` calls consistently use the expected **system default SSL context**. It also adds a **debug-only escape hatch** to disable verification via config/env vars when needed.

## What’s changed
- Introduce `openhands.utils.http_session`:
  - `httpx_verify_option()` for consistent `verify=` across `httpx` clients (async & sync).
  - Shared client management + `configure_http_session(verify=...)`.
  - Honors env vars: `INSECURE_SKIP_VERIFY` (also accepts `OPENHANDS_INSECURE_SKIP_VERIFY`) — **default is secure (verify on)**.
- Wire default SSL behavior everywhere we create `httpx` clients:
  - Enterprise modules (e.g., Jira Cloud/DC, Linear), OAuth/key routes, billing, nested conversation managers, providers, storage webhooks, local runtime readiness checks, Git services, resolvers, etc.
- Config & docs:
  - `config.template.toml`: add `insecure_skip_verify` (commented, NOT recommended).
  - `openhands/core/config/openhands_config.py`: add `insecure_skip_verify: bool | None`.
  - `openhands/core/config/utils.py`: call `configure_http_session()` based on config/env.
  - `docs/usage/environment-variables.mdx`: document `INSECURE_SKIP_VERIFY`.
- Backwards compatible: by default we still **verify certificates**; behavior is unchanged for secure setups.

## Motivation
- Enterprises often use a **self-signed corporate root CA**. Using the system default SSL context lets requests succeed once that root is present in the OS trust store.
- A **temporary opt-out** (`insecure_skip_verify`) helps **debug TLS issues** without blocking critical flows (e.g., first-time setup, CI smoke tests).

## Usage notes
- **Secure default**: No change required if your environment has proper trust roots installed.
- **Temporary debug override** (NOT recommended for production):
  - Env: `INSECURE_SKIP_VERIFY=true`
  - Or `config.toml`:  
    ```toml
    # Skip TLS certificate verification for outbound HTTP requests (NOT recommended)
    insecure_skip_verify = true
    ```
- I also touched the **enterprise** module (planning to try Jira integration soon), feel free to ask for or to commit additional changes.

## Testing
- ✅ Verified success against an internal endpoint with a **corporate root CA** installed in the OS trust store (ubuntu based docker + bitbucket endpoint).
- ✅ Verified failure → success flip when toggling `INSECURE_SKIP_VERIFY`.

## Security considerations
- **Verification remains enabled by default.**
- `insecure_skip_verify` is **explicitly documented as NOT recommended** and intended for **temporary debugging** only.

## Thanks
Really happy to use OpenHands — huge thanks to the team and the community for the awesome project and support 🙏

### DIFF
--- a/openhands/utils/http_session.py
+++ b/openhands/utils/http_session.py
@@ -1,11 +1,71 @@
+import os
+import ssl
 from dataclasses import dataclass, field
+from threading import Lock
 from typing import MutableMapping
 
 import httpx
 
 from openhands.core.logger import openhands_logger as logger
 
-CLIENT = httpx.Client()
+
+def _env_insecure_skip_verify() -> bool:
+    truthy = {'1', 'true', 'yes', 'on'}
+    for env_var in ('OPENHANDS_INSECURE_SKIP_VERIFY', 'INSECURE_SKIP_VERIFY'):
+        value = os.environ.get(env_var)
+        if value is not None:
+            return value.strip().lower() in truthy
+    return False
+
+
+_client_lock = Lock()
+_verify_certificates: bool = not _env_insecure_skip_verify()
+_client: httpx.Client | None = None
+
+
+def httpx_verify_option() -> ssl.SSLContext | bool:
+    """Return the verify option to pass when creating an HTTPX client."""
+
+    if _env_insecure_skip_verify():
+        return False
+    return ssl.create_default_context()
+
+
+def _build_client(verify: bool) -> httpx.Client:
+    if verify:
+        return httpx.Client(verify=ssl.create_default_context())
+    return httpx.Client(verify=False)
+
+
+def _get_client() -> httpx.Client:
+    global _client
+    if _client is None:
+        with _client_lock:
+            if _client is None:
+                _client = _build_client(_verify_certificates)
+    return _client
+
+
+def configure_http_session(*, verify: bool | None = None) -> None:
+    """Configure the shared HTTPX client used by HttpSession."""
+
+    global _client, _verify_certificates
+
+    target_verify = _verify_certificates
+    if verify is not None:
+        target_verify = verify
+    elif _client is None:
+        # Ensure we honour environment variables on first configuration
+        target_verify = not _env_insecure_skip_verify()
+
+    if target_verify == _verify_certificates and _client is not None:
+        return
+
+    with _client_lock:
+        if _client is not None:
+            _client.close()
+        _verify_certificates = target_verify
+        _client = _build_client(_verify_certificates)
 
 
 @dataclass
@@ -27,7 +87,7 @@ class HttpSession:
         headers = kwargs.get('headers') or {}
         headers = {**self.headers, **headers}
         kwargs['headers'] = headers
-        return CLIENT.request(*args, **kwargs)
+        return _get_client().request(*args, **kwargs)
 
     def stream(self, *args, **kwargs):
         if self._is_closed:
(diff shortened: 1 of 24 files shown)


phymbert:
I dont think linter issues are related to the changes, tell me if you want me to fix them.

phymbert:
Is there something missing to review this change ?

I can not use openhands without these basic security fixes.

Please tell me how I can help to merge it. Thanks

openhands/utils/http_session.py
```diff
@@ -1,11 +1,71 @@
...
+from threading import Lock
 from typing import MutableMapping
 
 import httpx
 
 from openhands.core.logger import openhands_logger as logger
 
-CLIENT = httpx.Client()
+
+def _env_insecure_skip_verify() -> bool:
```
rbren: We should probably move this into SandboxConfig instead of building a new env parser here

phymbert: Moved to `SecurityConfig` instead because ` httpx` is used outside the sandbox, typically in bitbucket. Removed the custom env parsing in favor of the default behavior `SECURITY_INSECURE_SKIP_VERIFY`

phymbert:
Hands up 🖐

phymbert:
1) disabling verification is opt-in, so there is no risk at all of a Man-In-The nowhere attack except if the user wants it. This is a standard option in any platform 
2) examples you are giving is exactly what the PR does: to use OS CA bundle without code changes, this is also the industry standard

I believe you must double read the code changes?

raymyers:
> examples you are giving is exactly what the PR does

I don't think supplying a custom cert which is checked is the same thing as allowing disabling of all checking.

> no risk except if the user wants it

I agree that the risk is opt-in, and the question is whether to include that option given there could be a more secure alternative, namely custom certs.

phymbert:
The code adds the proper `SSL.create_default_context()` which looks into the OS CA bundle, i.e `/etc/ssl/certs/ca-certificates.crt`. This is the standard way to add custom chain certs instead of modifying the code, than using the ones bundled with `certify`. Letting  the user to specify his own CA location is dangerous as it breaks OS configuration. The proposed approach is safe and works for any container based image, even distroless. 

Regarding the second code changes: the option to fully disable ssl verify, all industry tools like [curl](https://curl.se/libcurl/c/CURLOPT_SSL_VERIFYPEER.html), [apache](https://httpd.apache.org/docs/current/mod/mod_ssl.html) offer this alternative option which is of course not recommended for production. But if you feel it is so dangerous, we can add a warning message. I thought the documentation was clear enough, isn't?

raymyers:
> The code adds the proper SSL.create_default_context()

Thanks for expanding. I think we're all on board with that part of the change. Does your use case require the disable option in addition, or this is more about following the pattern of other tools?

phymbert:
Thanks for your time on this, appreciated.

I used it to debug at the beginning to be sure it was wrong httpx configuration, and I believe this is useful for other user to investigate in the future, but no, my use case is over secured, trust me ^)

So if you want, we can simply get rid of this option, but I think it is worth keeping

raymyers:
Spun this off into #11505 to use the default SSL context, which has new been merged. I also think the config option has merit, and I've created issue #11506 so it can be considered separately without blocking this.

Thanks for this!

PR #11114, opened 2025-09-25
chore - Upgrade to Debian Trixie in app image and runtime image option

raymyers:

Security upgrades: App image is now built on Debian Trixie and Python 3.13

---

* Build app image (and therefore enterprise-server image) using Debian Trixie, the new stable version. Also upgrade Python and Node versions.

* Provide a Runtime image option with the latest stable Debian (Trixie) instead of bookwork.
  * This is accomplished with a fork of the nikolaik python-nodejs images. 
  * We can go back to nikolaik after they add [Trixie support](https://github.com/nikolaik/docker-python-nodejs/pull/292), or we may choose to further customize our base image.
---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:3c6d3f4-nikolaik   --name openhands-app-3c6d3f4   docker.all-hands.dev/all-hands-ai/openhands:3c6d3f4
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@ray/debian-trixie openhands
```


poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.2.0 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.1.3 and should not be changed by hand.
```
raymyers, line 1: 2.1.3 is the version we are officially still using I think.

pyproject.toml
```diff
@@ -42,6 +42,7 @@ uvicorn = "*"
 numpy = "*"
 json-repair = "*"
 browsergym-core = "0.13.3"                         # integrate browsergym-core as the browsing interface
+playwright = "^1.55.0"
```
raymyers, line 45: Make explicit instead of transitive through browsergym-core so we get the upgrade can install on Debian Trixie

openhands/runtime/utils/runtime_templates/Dockerfile.j2
```diff
@@ -24,14 +24,10 @@ RUN apt-get update && \
     apt-get install -y --no-install-recommends \
         wget curl ca-certificates sudo apt-utils git jq tmux build-essential ripgrep ffmpeg \
         coreutils util-linux procps findutils grep sed \
-        {%- if (base_image.endswith(':latest') or base_image.endswith(':24.04') or ('mswebench' in base_image)) -%}
```
raymyers, line 27: A pattern used a few times in this PR: Avoid trying to determine the right package name from the image name. Using an OR, just try installing one, fallback on the other. If both fail then we fail.

Issue #11120, opened 2025-09-25
[Bug]: v1 CLI left/right/option+left keys etc doesn't work


xingyaoww:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

For example, characters like `[[C^` will show up when i hit "right" key on keyboard
And `^[b` when i hit `option+left`

<img width="450" height="184" alt="Image" src="https://github.com/user-attachments/assets/567e533e-0f12-4054-84ec-8dc9b71e9e07" />

Ideally, hit the left arrow to let me move my cursor back

### OpenHands Installation

Docker command in README

### OpenHands Version

openhands-cli/ under branch "v1"

### Operating System

None

xingyaoww:
@OpenHands please checkout v1 branch and go to openhands-cli/ folder and `uv run openhands-cli`. You should write a test there to try to reproduce this behavior, then try to fix it

malhotra5:
I believe this was a weird state when transitioning between multiple conversations. #11127 should have fixed this.

PR #11127, opened 2025-09-25
CLI(V1): restore terminal state

malhotra5:

---

#### Steps to reproduce
1. run `make run`
2. type `/exit` 
3. select `Yes, proceed`
4. press any arrow keys on the keyboard, they will display raw escape sequences for arrow keys (^[[A, ^[[B, ^[[C, ^[[D)

#### Fix
This PR 

1. standardizes the user of app exit so prompt toolkit can clean up prompt applications properly
2. even after implementing part 1 we still see arrow keys not being restored. for this reason we add `restore_tty` as a workaround to turn off application cursor keys and bracketed paste if they were left on

#### Followup 

We should open an issue upstream with prompt toolkit to fix the improper cleanup

---

### DIFF
--- a/openhands-cli/openhands_cli/agent_chat.py
+++ b/openhands-cli/openhands_cli/agent_chat.py
@@ -153,7 +153,7 @@ def run_cli_entry() -> None:
             exit_confirmation = exit_session_confirmation()
             if exit_confirmation == UserConfirmation.ACCEPT:
                 print_formatted_text(HTML("\n<yellow>Goodbye! 👋</yellow>"))
-            break
+                break
 
 
     # Clean up terminal state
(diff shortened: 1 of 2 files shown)


xingyaoww:
Will it possibly fix this https://github.com/All-Hands-AI/OpenHands/issues/11120 too?

openhands-cli/openhands_cli/agent_chat.py
```diff
@@ -143,4 +153,9 @@ def run_cli_entry() -> None:
             exit_confirmation = exit_session_confirmation()
             if exit_confirmation == UserConfirmation.ACCEPT:
                 print_formatted_text(HTML("\n<yellow>Goodbye! 👋</yellow>"))
-                break
+            break
```
xingyaoww: I think we should do this?
```suggestion
                break
```

malhotra5:
> Will it possibly fix this https://github.com/All-Hands-AI/OpenHands/issues/11120 too?

This seems like an unusual state that I haven't seem. This PR could potentially fix if the keybindings were corrupted in between separate CLI session

xingyaoww (approved):

PR #11216, opened 2025-10-02
feat: V1 client types

amanape:

---
This PR introduces the translated client-side types of the new V1 events.

---


amanape:
@hieptl Usually yes, but the main difference here is we actually reuse `SourceType` and not the other union types (at least for now). So there isn’t a strong reason to refactor. In the future if we end up reusing these types elsewhere, then we will definitely need to refactor. But right now, it’s really unnecessary / preference

PR #11223, opened 2025-10-03
fix(audio): use transcriptions endpoint and print usage info

idokenan2000:

The audio parsing command now uses the transcription API instead of the translation endpoint, ensuring that audio is transcribed in its original language as expected. It also prints the total usage tokens returned by the API, giving users better visibility into resource consumption.

---
This PR updates the `parse_audio` function to call `audio.transcriptions.create()` instead of `audio.translations.create()`. The previous implementation was translating audio rather than transcribing it, which caused confusion and unexpected results for same-language audio files.
Additionally, the function now prints `transcript.usage.total_tokens` to make API usage more transparent to developers and users. This change is backward-compatible and requires no additional configuration.

---
N/A

### DIFF
--- a/openhands/runtime/plugins/agent_skills/file_reader/file_readers.py
+++ b/openhands/runtime/plugins/agent_skills/file_reader/file_readers.py
@@ -128,12 +128,12 @@ def parse_audio(file_path: str, model: str = 'whisper-1') -> None:
     """
     print(f'[Transcribing audio file from {file_path}]')
     try:
-        # TODO: record the COST of the API call
         with open(file_path, 'rb') as audio_file:
-            transcript = _get_openai_client().audio.translations.create(
+            transcript = _get_openai_client().audio.transcriptions.create(
                 model=model, file=audio_file
             )
         print(transcript.text)
+        print(f'Total usage tokens: {transcript.usage.total_tokens}')
 
     except Exception as e:
         print(f'Error transcribing audio file: {e}')


idokenan2000:
## 🔧 CI Fix Required

The CI is failing because the OpenAI transcription API response may not always include the `usage` field. The current code tries to access `transcript.usage.total_tokens` without checking if the `usage` field exists, causing an `AttributeError`.

### Fix

Replace this line:
```python
print(f'Total usage tokens: {transcript.usage.total_tokens}')
```

With this safer version:
```python
if (
    hasattr(transcript, 'usage')
    and transcript.usage
    and hasattr(transcript.usage, 'total_tokens')
):
    print(f'Total usage tokens: {transcript.usage.total_tokens}')
```

This ensures the code only tries to access the usage information when it's actually present in the API response, preventing the AttributeError that's causing the CI failure.

### Root Cause
The OpenAI API documentation doesn't guarantee that the `usage` field will always be present in transcription responses, especially in certain configurations or error conditions.

enyst (commented):
Thank you @idokenan2000 ! Could you please take a look at the test failure?

Just a heads-up, though: we are re-working OpenHands and I doubt these runtime plugins could still work with the new V1 version. In fact, we were considering deprecating them a while ago, even.  Maybe there are MCP servers with this functionality?

enyst:
Thank you again for the contribution @idokenan2000 , unfortunately I think we won't merge this, since we have already released OpenHands-CLI V1, the next major version, and it doesn't use this code.
(the web UI code is also merged, just not in use by default at the moment)

Sorry for the timing, but thank you! Please feel free to let us know if you think it's a mistake.

PR #11235, opened 2025-10-04
[DRAFT] Migrate to uv

xingyaoww:

---

---


BradKML:
Could we increase the timeout for `uvx`? Cus if Docker is slow on downloading container images, it becomes impossible to use the new install method. Also Docker won't cache some of the images/components it downloads, so there is no easy way to resume.

xingyaoww:
Close this for now since i realize i can directly push to https://github.com/All-Hands-AI/OpenHands/pull/10833

@BradKML do you mean installation for CLI?

BradKML:
@xingyaoww GUI not CLI, since GUI docker images are slower than CLI which is easier to handle

.github/workflows/e2e-tests.yml
```diff
@@ -21,16 +21,13 @@ jobs:
...
-        uses: abatilo/actions-poetry@v4
-        with:
-          poetry-version: 2.1.3
+      - uses: hynek/setup-cached-uv@v2
+        name: Setup uv
 
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: '3.12'
```
jakob1379, line 30: This part can be left out too, to further simplify the CI as uv will install the specified python version and use it cache on succeeding runs. This is just a note and not crucial to remove and can always be done at a later point.

Makefile
```diff
@@ -116,59 +116,49 @@ check-tmux:
...
 		echo "$(BLUE)Installing chroma-hnswlib...$(RESET)"; \
-		export HNSWLIB_NO_NATIVE=1; \
-		poetry run pip install chroma-hnswlib; \
-	fi
-	@if [ -n "${POETRY_GROUP}" ]; then \
-		echo "Installing only POETRY_GROUP=${POETRY_GROUP}"; \
-		poetry install --only $${POETRY_GROUP}; \
-	else \
-		poetry install --with dev,test,runtime; \
+		HNSWLIB_NO_NATIVE=1 uv pip install chroma-hnswlib; \
```
jakob1379, line 129: I also stumbled on this, why don't we just add it as a dependency group so all dependencies are documented instead of hidden and not part of the lock file?

PR #11244, opened 2025-10-05
CLI(V1): Multi platform build + release support

malhotra5:
- [] This change is worth documenting at https://docs.all-hands.dev/
- [] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

---

This PR implements a release process of V1 CLI. During a tagged release the workflows should

- build and attach binaries for macos + linux to the release
- publish pypi package `openhands` 

---

Fixes APP-30

@malhotra5 can click here to [continue refining the PR](https://app.all-hands.dev/conversations/9ed883fbf55d415585fa051dd5c00c71)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:80d89d6-nikolaik   --name openhands-app-80d89d6   docker.all-hands.dev/all-hands-ai/openhands:80d89d6
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@cli-macos-build-and-release-improvements#subdirectory=openhands-cli openhands
```

### DIFF
--- a/.github/workflows/cli-build-test.yml
+++ b/.github/workflows/cli-build-test.yml
@@ -98,7 +98,7 @@ jobs:
         uses: softprops/action-gh-release@v2
         with:
           files: release-assets/*
-          draft: false
+          draft: true
           prerelease: false
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ secrets.CLI_RELEASE_TOKEN }}


.github/workflows/cli-build-test.yml
```diff
@@ -45,6 +48,8 @@ jobs:
 
       - name: Build binary executable
         working-directory: openhands-cli
+        env:
+          PYTHONPATH: ""
```
malhotra5:
```suggestion
```

.github/workflows/cli-build-test.yml
```diff
@@ -56,3 +59,47 @@ jobs:
...
+          # Rename binaries to include OS in filename
+          if [ -f artifacts/openhands-cli-ubuntu-latest/openhands ]; then
+            cp artifacts/openhands-cli-ubuntu-latest/openhands release-assets/openhands-linux
+          fi
+          if [ -f artifacts/openhands-cli-macos-latest/openhands ]; then
+            cp artifacts/openhands-cli-macos-latest/openhands release-assets/openhands-macos
+          fi
+          ls -la release-assets/
+
+      - name: Create GitHub Release
```
mamoodi, line 97: Does this generate a different release in the OpenHands repo? So we will have essentially two release notes. One for OpenHands and one for CLI? But it will all be in the same "Releases" page?

.github/workflows/pypi-release.yml
```diff
@@ -32,3 +32,32 @@ jobs:
         run: ./build.sh
       - name: publish
         run: poetry publish -u __token__ -p ${{ secrets.PYPI_TOKEN }}
+
+  release-cli:
```
mamoodi, line 38: Basic question for my own understanding. What is the difference between the below PyPi package release and the above PyPi package release. The below one is for the CLI but what will the above one be used for?

malhotra5: The one above is used to release the GUI pypi package `openhands-ai`

We're using `openhands` package for the cli because it also serves as an entrypoint for the GUI via `openhands serve`

.github/workflows/cli-build-test.yml
malhotra5: I need to double check, unfortunately not confident about how we can be sure about this without actually cutting a release and seeing what the outcome is

malhotra5:
@openhands carefully read the diff on this PR 

Note that we have new logic in place to release the standalone project inside openhands-cli/ when a tagged release is performed. However, this conflicts with the release procedure with the other project in the root of the repo

We want it so that 

- Tag normally, would release the root project (no changes to be made here, the release logic is already in place)
- Tag 1.2.3-cli would release CLI (the new standalone project)


Please modify this PR so that only tags containing `-cli` trigger the cli specific release jobs and it doesn't accidentally trigger the non-cli specific jobs

.github/workflows/cli-build-test.yml
malhotra5: @openhands please read the diff on this pr and note how we are attempting to cut releases independently for two separate projects but in the same repo

then please give detailed answers on how releases notes and artifact uploads will work. Will all the releases be in one release page? can we maintain separate release pages? 

if there are certain limitations also provide details on existing solutions out there to manage releases for multiple projects in the same monorepo

.github/workflows/cli-build-test.yml
```diff
@@ -56,3 +59,47 @@ jobs:
...
+        uses: softprops/action-gh-release@v2
+        with:
+          files: release-assets/*
+          generate_release_notes: true
```
malhotra5:
```suggestion
```

.github/workflows/cli-build-test.yml
```diff
@@ -56,3 +59,46 @@ jobs:
...
+            cp artifacts/openhands-cli-macos-latest/openhands release-assets/openhands-macos
+          fi
+          ls -la release-assets/
+
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v2
+        with:
+          files: release-assets/*
+          draft: false
+          prerelease: false
```
malhotra5, line 102: maybe we can toggle these to be drafts at first and test the release process

mamoodi: It would be really great if you could test the release process. And not having to run into issues when we actually are ready to release.

mamoodi:
Looks good generally. If you can somehow test a release or something, it would be nice.

malhotra5:
Makes sense! I'll try simulating this release in a completely separate repo with dummy code, will keep you posted on how it goes

.github/workflows/cli-build-test.yml
malhotra5: As per slack thread, we've tested the release process on a dummy repo 

https://openhands-ai.slack.com/archives/C09FJAZSP9N/p1760031461224279

PR #11273, opened 2025-10-07
Mint security eval fix

juanmichelini:
Changed eval call for ast.literal_eval call to fix Vulnerability report.

---

---


juanmichelini:
> Actually i think this probably won't work since this is intentionally executing code

my bad, took the comment    
     # Converting the string answer to a number/list/bool/option
literlly.

I understand then that it is executing code like 1+2, and then converts it to 3?

PR #11322, opened 2025-10-10
fix: remove the hardcoded 5-minute timeout from the docker pull command

### LINKED ISSUE #11191: [Bug]: My network is so fast so that I can finish download in 300 seconds

Sorry for the terrible title but I think this is a real issue that causes unnecessary retries during installation.

Below are my console logs, which show my frustration as I tried everything, but still kept getting "timeout" errors.

https://pastebin.ubuntu.com/p/mQKkRrGYxh/

My internet connection isn't slow as I think, my download speed is usually 15-20 MB/s—yet I still can't download the image within 5 minutes. If a timeout occurs, everything is terminated, and the next attempt starts from scratch. 

I had to manually install the image and then use uvx to start it. Since this is the recommended default installation method, I think there should be more consideration for users in less-than-ideal environments and be more rubust. At the very least, there should be a asking whether the user wants to wait longer if the process can't finish within the given timeout.

I'm also curious: this recommended installation method was [introduced two months ago](https://github.com/All-Hands-AI/OpenHands/commit/5e85986f3229402b13d3a03142626b6c36a97f2b), but has no one else encountered this issue? (Or is it just me that really needs to buy Starlink?) 

https://github.com/All-Hands-AI/OpenHands/blob/2382baacc2d0c220e98f4da7931012a221b07ceb/openhands/cli/gui_launcher.py#L95-L121

### OpenHands Installation

Docker command in README

### OpenHands Version

latest

### Model Name

_No response_

### Operating System

WSL on Windows

_No response_

llamantino:

Improves reliability of the `openhands serve` CLI command on slower networks.

---

Removes the 5-minute timeout from the ~3 GB docker pull download, which made it impossible for anyone with a connection slower than 10 MB/s (80 Mbit/s) to use `openhands serve`.

---
Closes #11191.

### DIFF
--- a/.github/workflows/enterprise-check-migrations.yml
+++ b/.github/workflows/enterprise-check-migrations.yml
@@ -1,0 +1,52 @@
+name: Enterprise Check Migrations
+
+on:
+  pull_request:
+    paths:
+      - 'enterprise/migrations/**'
+
+jobs:
+  check-sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout PR branch
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+
+      - name: Fetch base branch
+        run: git fetch origin ${{ github.event.pull_request.base.ref }}
+
+      - name: Check if base branch is ancestor of PR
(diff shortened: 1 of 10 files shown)


mamoodi:
Thanks very much!!

PR #11326, opened 2025-10-12
feat(evaluation): Add multi-swe-bench dependency and fix rollout script

KevinMusgrave:
This PR modifies:
- `pyproject.toml`:
  - Adds multi-swe-bench to the evaluation section.
- `rollout_multi_swegym.sh`:
  - Removes the `pip install multi-swe-bench` command. 
  - Uses `poetry run` to run multi-swe-bench evaluation.
  - Removes `eval-n-limit` from the evaluation command, because it is not a valid argument.


mamoodi:
@xingyaoww are you able to take a look at this PR?

KevinMusgrave:
Should the dependency point to the multi-swe-bench git repo instead of pypi?

- The latest version of multi-swe-bench on pypi is [`0.1.2` from May 31](https://pypi.org/project/multi-swe-bench/#history).
- On GitHub they're at [1.1.0](https://github.com/multi-swe-bench/multi-swe-bench/blob/main/setup.py) and it's been updated a lot since May 31.

neubig:
@KevinMusgrave Maybe we could ask them to push to pypi? That'd be the cleanest solution if they're open to it.

KevinMusgrave:
@neubig Ok, I opened an [issue in their repo](https://github.com/multi-swe-bench/multi-swe-bench/issues/79), and I'll comment here again when I hear back.

PR #11379, opened 2025-10-15
Fix race condition causing 500 errors when metadata.json disappears

olyashok:
## Problem
When `delete_conversation` removes a conversation directory, other operations (events, webhooks, background tasks) can recreate the directory structure without the `metadata.json` file. This causes warm runtime creation to fail with 500 errors when the system walks conversation folders and tries to deserialize missing metadata.

## Solution
- **Enhanced error handling** in `FileConversationStore.search()` to gracefully handle missing `metadata.json` files
- **Added orphaned directory cleanup** that runs in the background to remove directories without metadata
- **Implemented race condition prevention** by tracking conversations being deleted to prevent concurrent writes
- **Added proper JSON error handling** for corrupted metadata files

## Changes
- Fixed `search()` method to catch `FileNotFoundError` and skip orphaned directories
- Added `_cleanup_orphaned_directories()` method for background cleanup
- Added `_is_conversation_being_deleted()` to prevent race conditions
- Enhanced `delete_metadata()` to track deletion state
- Added proper error handling for JSON decode errors

## Testing
- All existing tests pass
- Pre-commit hooks pass (ruff, mypy, etc.)
- Fixes the specific issue where warm runtime creation fails due to half-empty conversation folders

Fixes the race condition described in the issue where POST requests for warm runtime creation walk conversation folders and fail when encountering directories without `metadata.json`.

### DIFF
--- a/openhands/server/services/conversation_service.py
+++ b/openhands/server/services/conversation_service.py
@@ -69,10 +69,14 @@ async def initialize_conversation(
     try:
         conversation_metadata = await conversation_store.get_metadata(conversation_id)
         return conversation_metadata
-    except Exception:
-        pass
-
-    return None
+    except FileNotFoundError:
+        # Conversation doesn't exist, this is expected for new conversations
+        logger.debug(f'Conversation {conversation_id} not found, will create new one')
+        return None
+    except Exception as e:
+        # Log other errors but don't silently fail
+        logger.error(f'Error loading conversation metadata for {conversation_id}: {e}')
+        return None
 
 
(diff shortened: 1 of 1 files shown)


tofarr:
I think we need an alternative approach here, given that multiple processes can access a conversation directory concurrently, I suggest that rather than using a thread lock we delete the metadata.json first and then use that as the criteria for determining whether the conversation exists or not.

mamoodi:
Going to close this based on the comments. Thank you for your contribution!

Issue #11384, opened 2025-10-15
[Feature] Adding an MCP call handler significantly reduces the overhead of RPC requests.


CLFutureX:
### Background
When the system currently executes an MCP tool call, it first establishes connections with all configured MCP Servers (including SSE, SHTTP, and Stdio types) and retrieves the complete set of tools, before executing the tool matching logic. This full-connection approach results in unnecessary resource consumption, and is particularly inefficient when only specific tools need to be called.

### Optimization Solution
Add an MCPCallHandler to implement instance caching for MCPClient. This avoids initiating full-scale initialization and tool call requests every time during multi-turn conversations, thereby eliminating unnecessary overhead.
To support dynamic updates of the tool list for MCPClient, a default refresh interval of 60 seconds is designed: When no matching MCPClient is found in the cache, expired mcp_clients will be refreshed, followed by another check for a match.

PR #11385, opened 2025-10-15
[Feature] Adding an MCP call handler significantly reduces the overhead of RPC requests.

CLFutureX:
[ISSUE](https://github.com/All-Hands-AI/OpenHands/issues/11384)


CLFutureX:
@enyst  @malhotra5  @mamoodi  Hey, please help me check if this is necessary? If not, I'll turn it off.

openhands/mcp/mcp_call_handler.py
```diff
@@ -0,0 +1,63 @@
...
+            (c for c in self.mcp_clients_map if tool_name in c.tool_map), None
+        )
+
+        # Updates clients whose cached state exceeds the 60-second
+        # validity threshold
+        if not target_client:
+            now_time = int(time.time())
+            for mcp_client, refresh_time in self.mcp_clients_map.items():
+                if now_time - refresh_time > 60:
+                    await mcp_client.refresh()
```
ryanhoangt, line 59: It seems after a refresh, the cached timestamp stays at its original value, and all subsequent calls after 60s will refresh again, which defeats the caching purpose. Maybe we have to write the new timestamp back into `self.mcp_clients_map`? WDYT?

openhands/mcp/client.py
```diff
@@ -176,3 +182,15 @@ async def call_tool(self, tool_name: str, args: dict) -> CallToolResult:
...
+        """Refresh the client and re-obtain the latest collection of tools."""
+        if not self.server_config:
+            raise RuntimeError('Server Config is empty, so refresh is not possible.')
+        self.tool_map.clear()
+        self.tools.clear()
+        self.client = None
+        if isinstance(self.server_config, MCPStdioServerConfig):
+            await self.connect_stdio(self.server_config)
+        else:
+            await self.connect_http(self.server_config)
```
ryanhoangt, line 196: I think we also need to pass in `sid` as conversation id here, similar to when creating via `create_mcp_clients`.

CLFutureX:
> Hi @CLFutureX, sorry for the long wait. The caching approach looks promising, I left some comments/questions about issues I see there.
> 
> Also we've migrated the OpenHands's core to the [new SDK](https://github.com/OpenHands/software-agent-sdk/tree/main/openhands-sdk/openhands/sdk/mcp), so we're not gonna merge the fix into this repo. It would be great if you can take a look and help open this PR into that repo. Thank you!

Well, I'll go check it out when I have time.

PR #11393, opened 2025-10-15
Extend context truncation cases

ryanhoangt:

---

This PR is to:
- Extend context truncation cases to match error messages thrown by models hosted via vLLM
- Log warning for LLM exceptions that are not catched currently

---


openhands/controller/agent_controller.py
```diff
@@ -923,6 +923,8 @@ async def _step(self) -> None:
                         raise LLMContextWindowExceedError()
                 else:
                     raise e
+            except Exception as e:
+                logger.warning(f'Error while getting action from LLM: {e}')
```
enyst: Seems this changes current behavior, the exception is eaten and it continues, instead of propagating it. Maybe we should re-raise?

ryanhoangt: Ah nice catch, I think I was confused by the logic here... we also have a `_step_with_exception_handling` to catch and log errors so I guess this is not needed. When testing the CLI earlier, even though the vLLM endpoint thrown 4xx errors, for some reasons I wasn't able to see it in the CLI 🤔

enyst (approved):
Thank you!

I assume we port this to agent-sdk?

ryanhoangt:
I think the "truncation when exceeding max context length" is not implemented in `agent-sdk` yet, we have an opening issue here: https://github.com/All-Hands-AI/agent-sdk/issues/282

enyst:
Ah, I think we have actually solved that issue

@openhands Clone the agent-sdk repo from the same org, it's the next version of OpenHands actually, so we are porting things so they don't get left behind.

Read this PR and the list of strings that it modifies. Then double-check against the checks in agent-sdk is_context_window_exceeded_exception() in llm.py, port this change and maybe more strings if missing, and propose a PR to that repo.

Do not touch the logic, it's a bit different and that's okay.

PR #11397, opened 2025-10-15
feat: make websocket client wait timeout configurable

olyashok:
## Summary

This PR addresses the issue where the 2-second timeout for waiting for websocket client connections during initialization was too aggressive, causing connection failures.

## Changes

- **Added  field** to  with a default value of 30 seconds
- **Replaced hardcoded 2s timeout** in  with the configurable value
- **Added progressive backoff** (starts with 0.1s, increases to 1s after 10 attempts)
- **Improved logging** with timeout information and reduced spam
- **Environment variable support** via 

## Configuration

The timeout can be configured in multiple ways:

1. **Environment variable**: 
2. **TOML config file**: 
3. **Default**: 30 seconds (much more reasonable than the previous 2s)

## Benefits

- More reliable websocket connections during initialization
- Configurable timeout based on deployment needs
- Better logging and debugging information
- Follows OpenHands configuration patterns

## Testing

- ✅ Configuration loads with default 30s timeout
- ✅ Environment variable override works
- ✅ TOML file override works
- ✅ All pre-commit hooks pass
- ✅ No syntax errors

Fixes the aggressive 2s timeout issue that was causing websocket connection failures.

### DIFF
--- a/openhands/server/session/session.py
+++ b/openhands/server/session/session.py
@@ -390,9 +390,15 @@ class WebSession:
             _waiting_times = 1
 
             if self.sio:
+                # Get timeout from configuration, default to 30 seconds
+                client_wait_timeout = self.config.client_wait_timeout
+                self.logger.debug(
+                    f'Using client wait timeout: {client_wait_timeout}s for session {self.sid}'
+                )
+
                 # Wait once during initialization to avoid event push failures during websocket connection intervals
                 while self._wait_websocket_initial_complete and (
-                    time.time() - _start_time < 2
+                    time.time() - _start_time < client_wait_timeout
                 ):
                     if bool(
                         self.sio.manager.rooms.get('/', {}).get(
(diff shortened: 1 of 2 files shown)


openhands/storage/conversation/file_conversation_store.py
```diff
@@ -83,13 +97,41 @@ async def search(
...
+
         for conversation_id in conversation_ids:
             try:
                 conversations.append(await self.get_metadata(conversation_id))
-            except Exception:
+            except FileNotFoundError:
+                # This is likely a race condition where the directory exists but metadata.json is missing
+                # This happens when delete_conversation removes the directory but other operations
+                # recreate it without the metadata.json file
+                logger.debug(
```
neubig, line 109: If we're encountering race conditions a lot that's worrisome, so I think we should at least issue a warning.
```suggestion
                logger.warning(
```

neubig (commented):
The logging changes LGTM, but I don't have confidence in my ability to adequately review the file_conversation_store.py changes. Maybe @tofarr could take a look.

openhands/storage/conversation/file_conversation_store.py
enyst: In V0, conversation directories without metadata.json exist naturally, as a result of using CLI or headless. Only the webserver creates metadata.json, the other entry points don’t.

So if it’s missing for some conversations, that could very well be a normal thing, and not the result of a race condition or errors, I think.

Maybe we could actually remove the comment about race condition? Or tweak it to account for normal cases… 🤔

openhands/server/session/session.py
```diff
@@ -390,22 +390,34 @@ async def _send(self, data: dict[str, object]) -> bool:
...
+
+                    # Progressive backoff: start with 0.1s, increase to 1s after 10 attempts
+                    sleep_duration = 0.1 if _waiting_times <= 10 else 1.0
+
+                    # Log every 2 seconds to reduce spam
+                    if _waiting_times % (20 if sleep_duration == 0.1 else 2) == 0:
+                        self.logger.warning(
+                            f'There is no listening client in the current room,'
+                            f' waiting for the {_waiting_times}th attempt (timeout: {client_wait_timeout}s): {self.sid}'
+                        )
```
tofarr, line 418: I don't think we should log this here.

mamoodi:
Hi there. Thanks for your contribution. I'm going to close based on enyst's comment and the PR being stale for a couple of weeks. Please let me know if this was a mistake and I can reopen.

Issue #11346, opened 2025-10-13
llm.draft_editor usable when usingf the UI


jcoffi:
**What problem or use case are you trying to solve?**
Allow the use of [llm.draft_editor] when using the UI. Currently, it appears to work for a while, but then fails with "correct_num not set". Then, once set, it won't work at all. Looking through the code, it appears to me that it never intended to work while using the UI. But when it does work, it has been a real cost saver as it allows me to use two models. One model that is provides superior code, and the other that has superior reasoning. 

**Describe the UX or technical implementation you have in mind**

**Additional context**

[llm.draft_editor]
model = "openrouter/x-ai/grok-code-fast-1:floor"
api_key = "<OR key here>"
base_url = "https://openrouter.ai/api/v1"
temperature = 0.0
correct_num = 5

Model used in UI: openrouter/openai/gpt-5:floor

PR #11398, opened 2025-10-15
fix(llm): support draft editor retries by adding correct_num to LLMConfig

jcoffi:

Fixes a UI error where the LLM-based draft editor intermittently fails with “correct_num not set” during lint-correction retries.
Enables configuring the draft editor’s correction attempts via [llm.draft_editor].correct_num.
Improves robustness so the editor doesn’t crash even if correct_num is not explicitly configured.
---

Adds a new field to LLMConfig:
correct_num: int = 5
This ensures [llm.draft_editor] (and any named LLM configs) parse correctly under extra='forbid' when correct_num is present.
Guards access to correct_num in FileEditRuntimeMixin.check_retry_num:
Uses getattr(self.draft_editor_llm.config, 'correct_num', 5) to prevent runtime failures if the field is missing or not set.
Rationale:
Previously, [llm.draft_editor] was skipped by LLMConfig.from_toml_section because correct_num was not a known field and extra='forbid' rejected the section.
The runtime then fell back to the default 'llm' config (which lacks correct_num). Later, check_retry_num accessed .correct_num unguarded, causing the UI error.
By adding correct_num to LLMConfig and guarding access, the UI can reliably use the draft editor configuration and respect the retry limit.
How to enable and configure

In config.toml:
[agent]
enable_llm_editor = true

[llm.draft_editor]
correct_num = 5

The draft editor will use the named LLM configuration and the specified number of correction attempts.

---
#11346


jcoffi:
The patch is isolated to just LLM config parsing and draft editor retries and doesn't even touch any code under test in test_delegation_flow /shrug

openhands/runtime/utils/edit.py
```diff
@@ -401,7 +401,8 @@ def llm_based_edit(self, action: FileEditAction, retry_num: int = 0) -> Observat
         return ret_obs
 
     def check_retry_num(self, retry_num):
-        correct_num = self.draft_editor_llm.config.correct_num  # type: ignore[attr-defined]
+        # Use configured correct_num if present; default to 5 otherwise for robustness
+        correct_num = getattr(self.draft_editor_llm.config, 'correct_num', 5)
```
raymyers, line 405: I think with this being added as a field we no longer need the getattr.

raymyers:
I've tweaked this slightly and am merging in #11530. Thanks for the contribution!

PR #11405, opened 2025-10-16
feat: make websocket client wait timeout configurable

olyashok:
Add client_wait_timeout field to OpenHandsConfig with default 30s
Replace hardcoded 2s timeout in session.py with configurable value  
Add progressive backoff and improved logging for timeout waiting
Support environment variable OPENHANDS_CLIENT_WAIT_TIMEOUT override

This addresses the issue where 2s timeout was too aggressive for websocket
client connections during initialization.

### DIFF
--- a/openhands/server/session/session.py
+++ b/openhands/server/session/session.py
@@ -412,7 +412,7 @@ class WebSession:
 
                     # Log every 2 seconds to reduce spam
                     if _waiting_times % (20 if sleep_duration == 0.1 else 2) == 0:
-                        self.logger.warning(
+                        self.logger.debug(
                             f'There is no listening client in the current room,'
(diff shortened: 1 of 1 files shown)


openhands/server/session/session.py
```diff
@@ -390,22 +390,34 @@ async def _send(self, data: dict[str, object]) -> bool:
...
-                        f'There is no listening client in the current room,'
-                        f' waiting for the {_waiting_times}th attempt: {self.sid}'
-                    )
+
+                    # Progressive backoff: start with 0.1s, increase to 1s after 10 attempts
+                    sleep_duration = 0.1 if _waiting_times <= 10 else 1.0
+
+                    # Log every 2 seconds to reduce spam
+                    if _waiting_times % (20 if sleep_duration == 0.1 else 2) == 0:
+                        self.logger.warning(
```
neubig:
```suggestion
                        self.logger.debug(
```

PR #11425, opened 2025-10-17
Fix excessive error logging for missing org-level microagents

### LINKED ISSUE #11424: [Bug]: Excessive error logging when missing org-level microagents

We are seeing lots of these errors.
`Failure to load repo "XXX/.openhands with all available providers."` (where XXX is the org name)

These are happening because we are checking for the `.openhands` (or on gitlab, `openhands-config`) repo to download it. Typically if we try to download a repo and it doesn't exist this is an error, but only in the case where we are downloading these config directories, it is expected that these directories may not exist. So only in this case, we should print an error at level "debug" saying "Org-level configuration not found at XXX/.openhands" (or openhands-config).

### OpenHands Installation

app.all-hands.dev

### OpenHands Version

_No response_

### Model Name

_No response_

### Operating System

None

_No response_

neubig:
## Summary of PR

This PR fixes issue #11424 by reducing excessive error logging when checking for missing org-level microagent repositories (`.openhands` or `openhands-config`).

**Changes made:**
- Added an `is_optional` parameter to `verify_repo_provider()` and `get_authenticated_git_url()` methods in `ProviderHandler`
- When `is_optional=True`, the method logs at DEBUG level instead of ERROR level
- Updated `Runtime.initialize` to pass `is_optional=True` when checking for org-level microagent repos
- Error-level logging is still used for required repositories (default behavior)

This ensures that missing org-level configuration repositories don't flood error logs, as these repositories are expected to be optional and may not exist for many organizations.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #11424

## Release Notes

**Fixed excessive error logging when org-level microagent directories are missing** - The system now logs at debug level instead of error level when checking for optional organization-level `.openhands` or `openhands-config` repositories that don't exist.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/7cda1f711b204026996715a2db50bc8a)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:e0bad5b-nikolaik   --name openhands-app-e0bad5b   docker.all-hands.dev/all-hands-ai/openhands:e0bad5b
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/All-Hands-AI/OpenHands@openhands/fix-org-microagent-logging#subdirectory=openhands-cli openhands
```

### DIFF
--- a/openhands/integrations/provider.py
+++ b/openhands/integrations/provider.py
@@ -449,7 +449,10 @@ class ProviderHandler:
         return f'{provider.value}_token'.lower()
 
     async def verify_repo_provider(
-        self, repository: str, specified_provider: ProviderType | None = None
+        self,
+        repository: str,
+        specified_provider: ProviderType | None = None,
+        is_optional: bool = False,
     ) -> Repository:
         errors = []
 
@@ -468,19 +471,22 @@ class ProviderHandler:
                 errors.append(f'{provider.value}: {str(e)}')
 
         # Log detailed error based on whether we had tokens or not
+        # For optional repositories (like org-level microagents), use debug level
+        log_fn = logger.debug if is_optional else logger.error
+
         if not self.provider_tokens:
-            logger.error(
+            log_fn(
                 f'Failed to access repository {repository}: No provider tokens available. '
                 f'provider_tokens dict is empty.'
             )
         elif errors:
-            logger.error(
+            log_fn(
                 f'Failed to access repository {repository} with all available providers. '
                 f'Tried providers: {list(self.provider_tokens.keys())}. '
                 f'Errors: {"; ".join(errors)}'
             )
         else:
-            logger.error(
+            log_fn(
                 f'Failed to access repository {repository}: Unknown error (no providers tried, no errors recorded)'
             )
         raise AuthenticationError(f'Unable to access repo {repository}')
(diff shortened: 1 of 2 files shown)


openhands/integrations/provider.py
```diff
@@ -449,7 +449,10 @@ def get_provider_env_key(cls, provider: ProviderType) -> str:
         return f'{provider.value}_token'.lower()
 
     async def verify_repo_provider(
-        self, repository: str, specified_provider: ProviderType | None = None
+        self,
+        repository: str,
+        specified_provider: ProviderType | None = None,
+        is_optional: bool = False,
```
malhotra5, line 455: NIT: maybe rename `is_optional` so its more obvious its only related to logging level

neubig: Thanks! Actually now that I think about it, maybe if `is_optional` is off, we should actually throw an exception, instead of just logging a warning. But for the time being maybe I'll just leave it as-is to avoid breaking things.

PR #11439, opened 2025-10-20
Enhance dead-loop recovery by pausing agent and reprompting

softpudding:
## Summary of PR

"Agent got stuck in a loop" was super painful for me, in order to avoid it I would monitor it and press ctrl+P when loop appears. And I realized this process can be automatic, as in this patch.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

### DIFF
--- a/openhands/controller/agent_controller.py
+++ b/openhands/controller/agent_controller.py
@@ -1337,7 +1337,7 @@ You can now provide new instructions to continue.
             # Re-use the last user message as the new instruction
             self.event_stream.add_event(
                 LoopDetectionObservation(
-                    content=f"""\n✅ Restarting with your last instruction: {last_user_message.content}'
+                    content=f"""\n✅ Restarting with your last instruction: {last_user_message.content}
 Agent is now continuing with the same task...
 """
                 ),


openhands/cli/tui.py
```diff
@@ -428,18 +428,32 @@ def display_command_output(output: str) -> None:
...
-        ),
-        title=title,
-        style=f'fg:{COLOR_GREY}',
-    )
-    print_formatted_text('')
-    print_container(container)
+    # Check if we're in loop recovery mode by looking at the call stack
+    import traceback
+
+    stack = traceback.extract_stack()
```
enyst: I think maybe we really don't need to introspect the call stack, could we just catch this exception?
https://github.com/All-Hands-AI/OpenHands/blob/44d11306a28c9a2093dc22c7fbdf4bc3b2ac9295/openhands/controller/agent_controller.py#L903

softpudding: OK, I have removed this part. I also added a Observation to handle loop detection and recovery messages, and ignored this observation in conversation memory.

openhands/core/loop.py
```diff
@@ -23,8 +23,11 @@ def status_callback(msg_type: str, runtime_status: RuntimeStatus, msg: str) -> N
         if msg_type == 'error':
             logger.error(msg)
             if controller:
-                controller.state.last_error = msg
-                asyncio.create_task(controller.set_agent_state_to(AgentState.ERROR))
+                if msg.startswith('AgentStuckInLoopError'):
+                    asyncio.create_task(controller.attempt_loop_recovery())
```
enyst: I wonder, would it be possible to maybe catch the exception in CLI, like here:
https://github.com/All-Hands-AI/OpenHands/blob/134c122026b4a3477f284284814479f9cc8d2c82/openhands/cli/main.py#L433

My point about the exception handling, is mainly that `loop.py` is also used by headless mode (without CLI interface), at least, IIRC: evaluation scripts, openhands resolver, etc are scripts that run a headless mode without passing through CLI module at all.

I'm sorry, I know it's gotten messy... which is why we're rethinking things for V1 😅

tests/unit/controller/test_agent_controller.py
```diff
@@ -350,17 +350,27 @@ def on_event_memory(event: Event):
...
+    )
+    # With loop recovery, we get multiple state changes
+    assert len(error_observations) >= 1
+    # Find the ERROR state change observation
+    error_observation = None
+    for obs in error_observations:
+        if obs.reason == 'AgentStuckInLoopError: Agent got stuck in a loop':
+            error_observation = obs
+            break
+    assert error_observation is not None
```
enyst: Ah, yes: so this test receives an `AgentStuckInLoopError` and loop recovery kicks in, even though it runs `state = await run_controller(...)`, which runs on headless mode (scripting, not CLI)

On one hand, it may be a good idea to treat the loop this way, and maybe if we evaluate the PR, we may get a better result! On the other hand, I don't know if we are even making evals on V0 right now anymore, as we are really really trying to get evals to work on V1 and move forward. 🤔

openhands/cli/commands.py
```diff
@@ -309,10 +311,29 @@ async def handle_resume_command(
         )
         return close_repl, new_session_requested
 
-    event_stream.add_event(
-        MessageAction(content='continue'),
-        EventSource.USER,
-    )
+    # Check if this is a loop recovery resume with an option
+    if command.strip() != '/resume':
+        # Parse the option from the command (e.g., '/resume 1' or '/resume 2')
```
enyst: If I understand correctly, please correct me if wrong, we create a Loop action/obs only after the user decides which option?

openhands/core/loop.py
softpudding: Thanks for the clarification, I have placed error handling into `openhands/cli/main.py`.

tests/unit/controller/test_agent_controller.py
softpudding: Thanks for the instruction, I have mvoed error handling into `openhands/cli/main.py`

openhands/cli/commands.py
softpudding: Oh yes, It should be only after resume 1 or 2, I have removed the erroneous 3 option. User can use /exit or ctrl+c to achieve same result

softpudding:
> A few thoughts, if I may.
> 
> There are two details here that are very interesting to me:
> 
> Our code, as you have seen, only stopped the agent if the LLM got stuck. If memory doesn't fail me, in the past year, I've seen failures as high as almost 30% stuck-in-a-loop on SWE-bench lite! So it affected results a lot. Nowadays, on some SOTA proprietary LLMs like Claude or GPT-5, we don't see stuck in a loop almost never anymore, caused by the LLM, I mean, as it usually was. This puts open LLMs at a disadvantage, because they still get stopped (and don't finish tasks).
> 
> On the other hand, as a general idea about the agent flow, we don't usually change agent's history; I know I did on some things in V0, and it really didn't seem like a good idea... We have here a dual system, with an event stream ("the source of truth") and history which is its runtime copy, which becomes quite difficult to handle if there are suddenly "two sources of truth", not to mention that what the LLM receives is yet another conversion to another array (of a... different size). So anyway, my point here is that due to the deep assumptions of the framework, this is not easy to properly account for in V0 on all execution paths.
> 
> But as you said, for DeepSeek it might still be extremely worthwhile to do!
> 
> In V1, I'm not sure how we will look at this: I'm sure it's _easier to do_, but it also runs headlong into a current V1 assumption: that a "conversation" that already happened is "frozen" in some sense. I'd love to see and brainstorm on what we could do. I'm curious if something like this can improve benchmark results for DeepSeek! But I don't know if we will run any more on V0.


Thanks for your input. V1's architecture sounds super interesting, I cannot wait to explore it.  
Maybe in a few weeks DeepSeek can also evolve to a point where we don't need to this anymore 😄 I will dig into V1 first, and see what we could do together! A lot of thanks for your help.

softpudding:
Seems only contributers with write access can merge😂 Could your help me merge, if possible? @enyst

openhands/controller/agent_controller.py
```diff
@@ -1208,5 +1270,93 @@ def _first_user_message(
...
+            self.state.end_id = -1
+
+        # Clear any cached messages
+        self._cached_first_user_message = None
+
+    async def _restart_with_last_user_message(
+        self, stuck_analysis: StuckDetector.StuckAnalysis
+    ) -> None:
+        """Restart the agent using the last user message as the new instruction."""
+        from openhands.events.action.message import MessageAction
```
enyst:
```suggestion
```

openhands/controller/agent_controller.py
```diff
@@ -1208,5 +1270,92 @@ def _first_user_message(
...
+            recovery_point = stuck_analysis.loop_start_idx
+            await self._truncate_memory_to_point(recovery_point)
+
+            # Set agent state to RUNNING and re-use the last user message
+            await self.set_agent_state_to(AgentState.RUNNING)
+
+            # Re-use the last user message as the new instruction
+            self.event_stream.add_event(
+                LoopDetectionObservation(
+                    content=f"""\n✅ Restarting with your last instruction: {last_user_message.content}'
```
enyst:
```suggestion
                    content=f"""\n✅ Restarting with your last instruction: {last_user_message.content}
```

enyst:
Thank you!

PR #11486, opened 2025-10-23
refactor: update delete_app_conversation to accept ID instead of object

hieptl:
## Summary of PR

This PR refactors the `delete_app_conversation` method to accept a UUID parameter instead of an `AppConversationInfo` object, implementing secure deletion logic with user context validation.

### Key Changes:
- Updated abstract method signatures in `AppConversationInfoService` and `AppConversationStartTaskService` to take UUID parameter
- Implemented secure deletion logic in SQL implementations using `secure_select` with user context validation
- Updated route handler to pass conversation ID for secure deletion
- Updated service callers to pass IDs instead of objects
- Updated tests to expect new method signature
- All tests passing with secure deletion implementation

The new implementation creates a single database statement that deletes the row and returns the number of affected rows, returning `True` if a row was affected, `False` otherwise.

## Change Type

- [x] Refactor

## Checklist

## Fixes

This addresses the requirement to update the delete method to use secure deletion with ID-based parameters instead of object-based parameters.

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:191f1fb-nikolaik   --name openhands-app-191f1fb   docker.openhands.dev/openhands/openhands:191f1fb
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@hieptl/app-89#subdirectory=openhands-cli openhands
```


openhands/app_server/app_conversation/sql_app_conversation_info_service.py
```diff
@@ -375,6 +375,29 @@ def _fix_timezone(self, value: datetime) -> datetime:
...
+        query = await self._secure_select()
+        query = query.where(
+            StoredConversationMetadata.conversation_id == str(conversation_id)
+        )
+        result = await self.db_session.execute(query)
+        existing = result.scalar_one_or_none()
+
+        if not existing:
+            return False
+
```
tofarr, line 387: Non blocking nit: Is there a way we could attach criteria to the delete_query so we wouldn't need to query twice?

openhands/server/routes/manage_conversations.py
```diff
@@ -465,15 +465,58 @@ async def get_conversation(
...
+        conversation_uuid = uuid.UUID(conversation_id)
+        # Check if it's a V1 conversation by trying to get it
+        app_conversation = await app_conversation_service.get_app_conversation(
+            conversation_uuid
+        )
+        if app_conversation:
+            # This is a V1 conversation, delete it using the app conversation service
+            return await app_conversation_service.delete_app_conversation(
+                conversation_uuid
+            )
```
tofarr, line 496: Could we rely on the return value from delete_app_conversation to avoid a duplicate query here?

openhands/app_server/app_conversation/sql_app_conversation_info_service.py
hieptl: Hi Tim,

I’ve updated the implementation as follows:
- The conversation is now fetched once at the API layer (_try_delete_v1_conversation) and the app_conversation and app_conversation_info objects are passed downstream.
- The following delete methods were updated to accept these objects:
  
```
- LiveStatusAppConversationService.delete_app_conversation(app_conversation)
- _delete_from_database(app_conversation_info)
- SQLAppConversationInfoService.delete_app_conversation_info(app_conversation_info)
- SQLAppConversationStartTaskService.delete_app_conversation_start_tasks(app_conversation_info)
```

- This approach allows us to attach criteria directly to the delete queries (e.g., using app_conversation_info.id) without performing an additional existence check or select query.
- Redundant database lookups have been removed, eliminating the need to re-query the same conversation.
- All related tests have been updated and are passing successfully.

Please let me know if this aligns with your feedback or if any further adjustments are needed.

Thank you for your guidance! 🙏

openhands/server/routes/manage_conversations.py
hieptl: I have updated the code to fetch and verify the conversation information only once, passing the retrieved app_conversation object downstream to the relevant services. 

This change helps eliminate redundant database queries.

I hope I have correctly understood your question and addressed your feedback. 

Thank you for your guidance! 🙏

tofarr (approved):

PR #11490, opened 2025-10-23
feat(evaluation): Filter task ids by difficulty for SWE Gym rollouts

KevinMusgrave:
### Motivation
In SWE-Gym for Multi-SWE-Bench, we'll collect more successful trajectories of easy tasks than hard tasks.

For example, we might end up with 10 successful trajectories of an easy task, and 1 successful trajectory of a hard task. I think after collecting N successful easy trajectories, it's better to focus compute on other tasks, to get a more balanced trajectory dataset.

### What the PR does
Before each trajectory generation run, the new code:
- Finds all task ids that have already been solved at least N times.
- Skips those ids for the current run.

Specific changes:
- `evaluation/benchmarks/multi_swe_bench/compute_skip_ids.py`
  - This new file finds all `final_report.json` and finds all tasks that have been resolved at least N times.
- `evaluation/benchmarks/multi_swe_bench/scripts/rollout_multi_swegym.sh`
  - Calls `compute_skip_ids.py` if an optional `SKIP_IDS_THRESHOLD` is passed in. Merges this with an optional `INPUT_SKIP_IDS`, so you can have hard-coded `SKIP_IDS` too.
  - `MAX_ITER` and `EVAL_LIMIT` are now passed in instead of being hardcoded.
- `evaluation/benchmarks/multi_swe_bench/run_infer.py`
  - Fixes the way `SKIP_IDS` is parsed. The old way would result in 1 skip id of `''` even if `SKIP_IDS` was not provided.
  - Adds `--filter_dataset_after_sampling`. When this flag is used, the skip id filtering will occur after sampling rather than before. For example, if the dataset has 100 tasks, and you have 12 skip ids, and an `eval_n_limit` of 50, then without this flag, it will filter down to 88 tasks (100 - 12), and sample 50 tasks from those 88. With the flag, it will sample 50 tasks from 100 tasks, and filter down to 38.
- `evaluation/utils/shared.py`
  - Makes `prepare_dataset` support the ability to filter after sampling. 

## Change Type
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

### DIFF
--- a/evaluation/benchmarks/multi_swe_bench/compute_skip_ids.py
+++ b/evaluation/benchmarks/multi_swe_bench/compute_skip_ids.py
@@ -1,8 +1,8 @@
-import json
 import argparse
-from pathlib import Path
-from collections import Counter
 import fnmatch
+import json
+from collections import Counter
+from pathlib import Path
 
 
 def find_final_reports(base_dir, pattern=None):
--- a/evaluation/benchmarks/multi_swe_bench/run_infer.py
+++ b/evaluation/benchmarks/multi_swe_bench/run_infer.py
@@ -792,7 +792,9 @@ if __name__ == '__main__':
     filter_func = None
     if args.filter_dataset_after_sampling:
         # Pass filter as callback to apply after sampling
-        filter_func = lambda df: filter_dataset(df, 'instance_id')
+        def filter_func(df):
+            return filter_dataset(df, 'instance_id')
+
         logger.info(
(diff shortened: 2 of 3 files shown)


neubig:
@openhands fix linting and push back to the appropriate branch on the KevinMusgrave:compute-skip-ids fork.

neubig (approved):
LGTM, thank you!

PR #11497, opened 2025-10-24
fix: support recursively parse nested tools params

Edward-sy:
## Summary of PR

The nested parameters inside the array were not parsed, which caused the inner parameters of `task_list` to be missing. 

This PR adds recursive parsing logic for nested parameters in `fn_call_converter.py`.  

With this change, all nested fields inside `task_list` are now correctly included in the generated description, improving `task_tracker` task planning and tracking behavior.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves OpenHands/software-agent-sdk#1074 

## Release Notes

**Description for Release Notes:**  
Fix a bug where nested array parameters in `task_tracker` were not parsed, causing missing inner fields in `task_list`. Added recursive parsing logic to fully support nested inputs.

### DIFF
--- a/openhands/llm/fn_call_converter.py
+++ b/openhands/llm/fn_call_converter.py
@@ -24,6 +24,7 @@ from openhands.llm.tool_names import (
     FINISH_TOOL_NAME,
     LLM_BASED_EDIT_TOOL_NAME,
     STR_REPLACE_EDITOR_TOOL_NAME,
+    TASK_TRACKER_TOOL_NAME,
 )
 
 # Inspired by: https://docs.together.ai/docs/llama-3-function-calling#function-calling-w-llama-31-70b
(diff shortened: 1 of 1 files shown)


Edward-sy:
Thanks! Moved the patch to software-agent-sdk as suggested; opening a new PR there and linking back here. (https://github.com/OpenHands/software-agent-sdk/pull/1011)

mamoodi:
Are we good to close this PR then?

Edward-sy:
@mamoodi
> Are we good to close this PR then?

Just to confirm — will the OpenHands repos no longer maintain or use the core agent logic (like fn_call_converter.py)?
If OpenHands still depends on this core for running agents, the bug in parsing nested array/object parameters (e.g., inside task_tracker.task_list) could still affect agents’ behavior.

Should fixes like this now live only in software-agent-sdk, or still be synced back here?

mamoodi:
@Edward-sy very soon I believe all the core agent code will be replaced by the software-agent-sdk in OpenHands. So OpenHands will use the sdk and remove all the unnecessary code. So if the issue is still present in the agent-sdk, that's where it should be fixed. Does that make sense?

mamoodi:
Going to close this based on the comments. It seems a PR was opened in agent-sdk. Thank you!
Please let me know if closing this was still a mistake.

PR #11505, opened 2025-10-24
Provide httpx default context for OS-provided certs

raymyers:
## Summary of PR

This is a subset of @phymbert's  #11051, backing out the config option for "no verify" so we can evaluate that separately.

This PR ensures all httpx calls consistently use the expected system default SSL context.

**Motivation**
Enterprises often use a self-signed corporate root CA. Using the system default SSL context lets requests succeed once that root is present in the OS trust store.

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/openhands/runtime:ffff2cd-nikolaik   --name openhands-app-ffff2cd   docker.all-hands.dev/openhands/openhands:ffff2cd
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@ray/httpx-defaultcontext#subdirectory=openhands-cli openhands
```


openhands/utils/http_session.py
```diff
@@ -1,11 +1,34 @@
...
+
+
+def httpx_verify_option() -> ssl.SSLContext | bool:
+    """Return the verify option to pass when creating an HTTPX client."""
+
+    return ssl.create_default_context() if _verify_certificates else False
+
+
+def _build_client(verify: bool) -> httpx.Client:
+    return httpx.Client(verify=ssl.create_default_context() if verify else False)
```
raymyers, line 22: docs on `create_default_context`, particularly:

> If [cafile, capath, cadata] are None, this function can choose to trust the system’s default CA certificates instead.

docs.python.org/3/library/ssl.html#ssl.create_default_context

raymyers:
I just need to run this again and make sure we’ve seen it still function,
then we can merge.

PR #11508, opened 2025-10-25
Fix: Remove temperature parameter for codex-mini-latest model

YakshithK:
## Summary of PR

This PR fixes a BadRequestError that occurs when using the `codex-mini-latest` model with OpenHands. The `codex-mini-latest` model doesn't support the `temperature` parameter, but OpenHands was always including it in API requests, causing the error: "Unsupported parameter: 'temperature' is not supported with this model."

The fix adds a model-specific check to remove the `temperature` and `top_p` parameters when using `codex-mini-latest`, similar to how other model-specific parameter restrictions are handled in the codebase.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves OpenHands/software-agent-sdk#1075

## Release Notes

**Release Notes Description:**
Fixed compatibility issue with `codex-mini-latest` model that was causing "Unsupported parameter: temperature" errors. The model now works correctly without throwing parameter-related errors.

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -122,6 +122,10 @@ class LLM(RetryMixin, DebugMixin):
             # openai doesn't expose top_p, but litellm does
             kwargs['top_p'] = self.config.top_p
 
+        if 'codex-mini-latest' in self.config.model.lower():
+            kwargs.pop('temperature', None)
+            kwargs.pop('top_p', None)
+
         # Handle OpenHands provider - rewrite to litellm_proxy
         if self.config.model.startswith('openhands/'):
             model_name = self.config.model.removeprefix('openhands/')


enyst:
Thank you for the contribution @YakshithK ! I'm afraid Codex-mini will not work on this version of OpenHands, because OpenAI restricts it to Responses API, and we haven't backported support for it:
https://platform.openai.com/docs/models/codex-mini-latest
image

We have implemented support in [OpenHands Agent-SDK](https://github.com/OpenHands/agent-sdk), which is the core of OpenHands V1, the major version of OpenHands. It works with a new CLI, and we're working to get the web UI ready.

Would you be willing to migrate this PR to agent-sdk, perhaps? I think it might need to add "codex" to the [Responses API support code](https://github.com/OpenHands/agent-sdk/blob/1f905f644fb6a93c0e742c2a1bdbfd1f75b3a136/openhands-sdk/openhands/sdk/llm/utils/model_features.py#L159)?

YakshithK:
Thanks @enyst  — makes sense! I’ll move this fix to agent-sdk and add codex-mini-latest to the Responses API support list.

PR #11519, opened 2025-10-27
fix(stream) Optimize Subscription Resource Cleanup in Stream to Ensure Complete Resource Management

CLFutureX:
**Background**
Right now, when we clean up subscription resources, we only delete specific resources. We don't think about the case where the whole resource container becomes empty (that is, empty dictionaries). This causes problems with resource management.
**Optimization**
After deleting specific resources, we automatically clean up empty parent containers. This stops empty dictionaries from being left in memory and makes resource management more complete.

### DIFF
--- a/openhands/events/stream.py
+++ b/openhands/events/stream.py
@@ -101,7 +101,7 @@ class EventStream(EventStore):
             subscriber_id in self._thread_loops
             and callback_id in self._thread_loops[subscriber_id]
         ):
-            loop = self._thread_loops[subscriber_id][callback_id]
+            loop = self._thread_loops[subscriber_id].pop(callback_id)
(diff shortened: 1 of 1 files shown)


enyst (commented):
Thank you @CLFutureX ! It seems tests are failing, but maybe it’s just flaky.

More importantly, I’m sorry for the timing, but we’re going to update default behavior to OpenHands V1, the major version of OH, a deep rewrite based on [OpenHands Agent SDK](https://github.com/OpenHands/software-agent-sdk).

That implementation of events is new, and it doesn’t use this code.

The behavior of the new event stream is not so messy anymore, but you’re most welcome to let us know if it has issues! I’m really sorry, but I don’t think we should merge this PR at this point. Though I do appreciate the contribution!

PR #11550, opened 2025-10-28
fix: prevent self-mention loops by sanitizing @OpenHands in assistant

YakshithK:
## Summary of PR

This PR prevents self-mention loops by sanitizing @OpenHands (and variants) in assistant message outputs. When the assistant mentions @OpenHands, it could trigger platform notifications or create clickable mentions that lead to unintended behavior. This fix neutralizes such mentions by inserting a zero-width joiner after the @ symbol, making them non-clickable while preserving readability.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #11548 

## Release Notes

**For users:** Assistant responses now automatically sanitize @OpenHands mentions to prevent accidental platform notifications and clickable mentions, improving overall system reliability and user experience.

### DIFF
--- a/enterprise/integrations/github/github_manager.py
+++ b/enterprise/integrations/github/github_manager.py
@@ -21,6 +21,7 @@ from integrations.utils import (
     CONVERSATION_URL,
     HOST_URL,
     OPENHANDS_RESOLVER_TEMPLATES_DIR,
+    sanitize_openhands_mentions
 )
 from jinja2 import Environment, FileSystemLoader
 from pydantic import SecretStr
@@ -33,7 +34,6 @@ from openhands.integrations.provider import ProviderToken, ProviderType
 from openhands.server.types import LLMAuthenticationError, MissingSettingsError
 from openhands.storage.data_models.secrets import Secrets
 from openhands.utils.async_utils import call_sync_from_async
-
 
 class GithubManager(Manager):
     def __init__(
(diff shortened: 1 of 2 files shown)


enyst (changes requested):
Thank you for looking into this issue, @YakshithK ! I think the problem was in a github comment, not on the web UI, so I'm not sure we should do this in FE.

In fact, I'm sorry but I think the events/action code will not work in V1, and we already have V1 in CLI. V1 works on the code in [OpenHands Agent-SDK](https://github.com/OpenHands/software-agent-sdk)

From another point of view, it might be worth asking ourselves if we really want to remove references in the actual, real agent message, or perhaps we could only remove them from the comment that the app is posting on GitHub... I'm not sure, what do you think? 🤔

YakshithK:
You’re right! the fix should be in the backend comment logic, not the frontend, and I missed the Agent-SDK/V1 split.

I’ll move the sanitization to just the GitHub comment output and look into Agent-SDK for V1

mamoodi:
Is this PR good to close then?

mamoodi:
CC @YakshithK is this PR good to close or still needs to be updated?

mamoodi:
Closing this based on comments. Please let me know if this was a mistake and I can reopen!

PR #11567, opened 2025-10-29
fix(llm): remove default reasoning_effort; only send when explicitly set

### LINKED ISSUE #11552: [Bug]: Default `reasoning_effort='high'` causes 400 errors with OpenAI Responses API models for unverified organizations
## Description

OpenHands automatically sets `reasoning_effort='high'` by default for all non-Gemini models, which causes 400 errors when using OpenAI models that utilize the Responses API (like `gpt-5-pro`, `gpt-5-codex`, `o3-pro`, etc.) with unverified organizations.

**OpenAI's official recommendations:**
- OpenAI recommends `reasoning_effort="medium"` as the default
- `reasoning.summary` should be opt-in, not automatically added
- Source: [OpenAI Reasoning Best Practices](https://platform.openai.com/docs/guides/reasoning-best-practices)

## Details

The issue was introduced/maintained in **PR OpenHands/OpenHands#9925** (commit `a32a6230`), which was intended to optimize Gemini 2.5 Pro performance by ~2.4x.

```python
# added in model_post_init (lines 174-180):
if self.reasoning_effort is None and 'gemini-2.5-pro' not in self.model:
    self.reasoning_effort = 'high'  # ← Still forces 'high' for non-Gemini!
```

**Intent:** Allow Gemini to use `reasoning_effort=None` which gets converted to `thinking={budget_tokens: 128}` for performance optimization.

**But:** All other models still receive `reasoning_effort='high'` by default.

## User Flow 

1. User initializes OpenHands with a Responses API model (e.g., `gpt-5-codex`)
2. `LLMConfig.model_post_init()` sets `reasoning_effort='high'` by default
3. LiteLLM detects model has `mode: "responses"` in its model cost map
4. LiteLLM activates Responses API bridge
5. Bridge automatically converts `reasoning_effort='high'` to `Reasoning(effort='high', summary='detailed')`
6. OpenAI rejects the request because `reasoning.summary` requires verified organization
7. Gets 400 error

## Related

I try this in mini-swe-agent and works (by default):
```yaml
model:
  model_name: "gpt-5-codex"
  # No reasoning_effort configured → Works! ✓
```
**This confirms the issue is with sending `reasoning_effort` to Responses API models, not with OpenHands specifically.**

### Option 1: Remove Default Entirely (Recommended)

Remove lines 174-180 from `openhands/core/config/llm_config.py`:

```python
# Set reasoning_effort to 'high' by default for non-Gemini models
# Gemini models use optimized thinking budget when reasoning_effort is None
logger.debug(
    f'Setting reasoning_effort for model {self.model} with reasoning_effort {self.reasoning_effort}'
)
if self.reasoning_effort is None and 'gemini-2.5-pro' not in self.model:
    self.reasoning_effort = 'high'
```

- ✅ Aligns with OpenAI's recommendation that `reasoning.summary` should be opt-in
- ✅ Prevents errors for users with unverified organizations
- ✅ Avoids unexpected costs (reasoning tokens are expensive)
- ✅ Gemini optimization still works (keeps `None`, converts to `thinking` budget)
- ✅ Users can explicitly configure via config.toml when needed

### Option 2: Exclude Responses API Models

If a default is desired, exclude models known to use Responses API:

```python
RESPONSES_API_MODELS = [
    'gpt-5-pro', 'gpt-5-codex',
    'o3-pro', 'o3-deep-research',
    'o4-mini-deep'
]

if (self.reasoning_effort is None and
    'gemini-2.5-pro' not in self.model and
    not any(m in self.model for m in RESPONSES_API_MODELS)):
    self.reasoning_effort = 'medium'  # Use OpenAI's recommended default
    logger.debug(f'Set reasoning_effort to "medium" for {self.model}')
```

### Option 3: Add Configuration Flag

Let users explicitly enable reasoning effort defaults:

```python
enable_default_reasoning_effort: bool = Field(
    default=False,
    description='Enable default reasoning_effort for supported models'
)

if (self.enable_default_reasoning_effort and
    self.reasoning_effort is None and
    'gemini-2.5-pro' not in self.model):
    self.reasoning_effort = 'medium'
```

## Recommendation

**Option 1** is strongly recommended because:
- **Follows OpenAI's official guidance**: Reasoning features should be opt-in
- **Maintains Gemini optimization**: Gemini still gets `None` → `thinking` budget
- **Solve the Issue**: No more 400 errors for unverified organizations
- **Reduces costs**: No unexpected reasoning token charges
- **User control**: Users can still configure explicitly when needed

## User Configuration

Users CAN already configure `reasoning_effort` in their `config.toml`:

```toml
[llm]
model = "gpt-5-codex"
reasoning_effort = "medium"  # or "low", "high", "none"
# Or omit to keep as None (recommended for Responses API models)
```

**However:**
- Most users don't know they need to do this
- The default behavior forces `reasoning_effort='high'` if they don't configure it
- This causes 400 errors for unverified organizations using Responses API models
- CLI is packaged as a standalone binary/executable, so users cannot modify the source code (`llm_config.py`) to change this, they would need to rebuild it

## Related Issues

- OpenHands/software-agent-sdk#987 - Original issue report about this problem
- PR OpenHands/OpenHands#9925 (commit `a32a6230`) - Where the current behavior was introduced/maintained
- Related to LiteLLM's automatic `reasoning.summary` addition 

### OpenHands Installation

CLI

### OpenHands Version

_No response_

### Model Name

_No response_

### Operating System

None

_No response_

YakshithK:
## Summary of PR
Removes the default setting of 'high' for reasoning_effort in non-Gemini models. This PR ensures reasoning_effort is only sent to the LLM backend when explicitly set by user config, avoiding unexpected behavior and supporting use cases where budgeted reasoning isn't desired by default.

## Change Type
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes
Fixes #11552 

## Release Notes

Removes default 'high' reasoning_effort; now only sent when specified in config, ensuring backend does not receive unexpected budget constraints.


YakshithK:
Thanks for the review! Good callout on the failing tests—they still expect the old behavior. I’ll update them so that non-Gemini models have reasoning_effort=None by default, and add a test to cover explicit config!

mamoodi:
@YakshithK please re-request a review from enyst when this PR is ready.

enyst (approved):
@YakshithK Thank you for the contribution, and I'm sorry, your fix to Gemini is necessary and welcome, but regarding the default value, I believe we actually will not set the default to None going forward. Part of the reason is that it's a mess among LLMs; different providers and different models set it to different values when None. Another part is that "high" is actually less expensive than popular non-reasoning models like Claude-4.

Please see [some more discussion here](https://github.com/OpenHands/software-agent-sdk/pull/1004#discussion_r2490846549).

I apologize that things are a bit chaotic right now. For more context, we are actually moving towards version V1 of OpenHands, which is built on [OpenHands Agent SDK](https://github.com/OpenHands/software-agent-sdk). That's a very significant rebuilding, and it was a miss from my part to continue here when we should have moved this discussion there. I'm sorry.

I'll merge your PR, but I don't think we will port it to agent-sdk, instead we move forward there with "high". Please though, if you feel strongly about this, feel free to make an issue on agent-sdk to reopen the discussion at some point. But thank you for this, I appreciate the contribution and your proposal here!

PR #11587, opened 2025-10-31
Hotfix(CLI): make sure to update condenser credentials

malhotra5:
## Summary of PR

We have a case where the agent already exists in the user's settings, and the user is modifying the LLM settings. At the moment, we only update the agent's LLM, but not the LLM for the condenser. 

This PR patches this logic but updating the condenser LLM as well. It also adds a unit tests that will verify whether all the LLMs in the agent share the same API key after going through the settings update. Going forward this test will ensure that if the agent has more LLMs introduced inside it, the test will fail indicating that we need to update the credentials for the new LLMs as well. 

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:94b207f-nikolaik   --name openhands-app-94b207f   docker.openhands.dev/openhands/openhands:94b207f
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@cli-fix-condenser-credentials#subdirectory=openhands-cli openhands
```

### DIFF
--- a/openhands-cli/tests/settings/test_settings_workflow.py
+++ b/openhands-cli/tests/settings/test_settings_workflow.py
@@ -121,15 +121,13 @@ def test_update_existing_settings_workflow(tmp_path: Path):
     assert True  # If we get here, the workflow completed successfully
 
 
-def test_llm_settings_modification_creates_multiple_llms_with_same_api_key_different_usage_ids():
+def test_all_llms_in_agent_are_updated():
     """Test that modifying LLM settings creates multiple LLMs with same API key but different usage_ids."""
     # Create a screen with existing agent settings
     screen = SettingsScreen(conversation=None)
-    
-    # Create initial agent with existing settings
     initial_llm = LLM(model='openai/gpt-3.5-turbo', api_key=SecretStr('sk-initial'), usage_id='test-service')
     initial_agent = get_default_cli_agent(llm=initial_llm)
-    
+
     # Mock the agent store to return the initial agent and capture the save call
     with (
         patch.object(screen.agent_store, 'load', return_value=initial_agent),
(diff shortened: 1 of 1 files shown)


openhands-cli/tests/settings/test_settings_workflow.py
```diff
@@ -121,6 +121,38 @@ def test_update_existing_settings_workflow(tmp_path: Path):
...
+
+        # Get the saved agent from the mock
+        saved_agent = mock_save.call_args[0][0]
+        all_llms = list(saved_agent.get_all_llms())
+        assert len(all_llms) >= 2, f"Expected at least 2 LLMs, got {len(all_llms)}"
+
+        # Verify all LLMs have the same API key
+        api_keys = [llm.api_key.get_secret_value() for llm in all_llms]
+        assert all(api_key == 'sk-updated-123' for api_key in api_keys), \
+            f"Not all LLMs have the same API key: {api_keys}"
```
malhotra5, line 148: Noting that the test will fail here without the patch introduced to the application code

openhands-cli/openhands_cli/tui/settings/settings_screen.py
```diff
@@ -182,7 +182,14 @@ def _save_llm_settings(self, model, api_key, base_url: str | None = None) -> Non
         if not agent:
             agent = get_default_cli_agent(llm=llm)
 
+        # Must update all LLMs
```
enyst, line 185: I think maybe this is the other way around... no, actually the user saves here "global" settings, so I tend to agree it should apply by default to all other LLMs 🤔 

I think sooner or later, we'll have the request to update them separately (we had it on V0). Not in scope here, I think maybe [LLM profiles](https://github.com/OpenHands/software-agent-sdk/pull/795) will give us a way to handle them separately...

malhotra5: LLM profiles would be awesome!

PR #11591, opened 2025-10-31
CLI: fix binary build

malhotra5:
## Summary of PR

Problem: we have been patch releasing without the executable binaries attached to the releases. There are two problems

1. the build script had a problem
2. the problem went uncaught in the CI

#### Before

1. If the build script fails in a place without error handling, it won't fail the CI
2. The upload artifact step doesn't upload any files and doesn't throw an error

image

#### After

1. Fix the build script
2. Wrap everything in exception handling
3. Throw an error before artifact upload step if no artifacts were created (defense in depth)

image

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:1309d7a-nikolaik   --name openhands-app-1309d7a   docker.openhands.dev/openhands/openhands:1309d7a
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@cli-fix-build-script#subdirectory=openhands-cli openhands
```

### DIFF
--- a/openhands-cli/build.py
+++ b/openhands-cli/build.py
@@ -287,6 +287,6 @@ if __name__ == '__main__':
         sys.exit(main())
     except Exception as e:
         print(e)
-        print('❌ Executable test failed, build process failed')
+        print('❌ Executable test failed')
         sys.exit(1)
 


openhands-cli/build.py
```diff
@@ -285,4 +283,10 @@ def main() -> int:
 
 
 if __name__ == '__main__':
-    sys.exit(main())
+    try:
+        sys.exit(main())
+    except Exception as e:
+        print(e)
+        print('❌ Executable test failed, build process failed')
```
malhotra5:
```suggestion
        print('❌ Executable test failed')
```

openhands-cli/build.py
```diff
@@ -20,15 +20,6 @@
 
 from openhands.sdk import LLM
 
-dummy_agent = get_default_cli_agent(
-    llm=LLM(
-        model='dummy-model',
-        api_key='dummy-key',
-        metadata=get_llm_metadata(model_name='dummy-model', llm_type='openhands'),
-    ),
-    cli_mode=True,
```
malhotra5, line 29: This extra `cli_mode` param was causing the build to fail silently

enyst: I thought this was necessary so that it doesn't include the browser tool?

malhotra5: Yep the function name is different

I'm calling `get_default_cli_agent` which wraps the `get_default_agent` function from `agent-sdk` with `cli_mode=True` set as a default

enyst: Oh! Hah, thank you, I see.

PR #11648, opened 2025-11-05
refactor(backend): the delete_app_conversation_info function

hieptl:
## Summary of PR

[Link to the line in GitHub](https://github.com/OpenHands/OpenHands/blob/4ea3e4b1fd850ae07e7b972feb36fca6e789d7eb/openhands/app_server/app_conversation/sql_app_conversation_info_service.py#L402)

Transaction flow

1. Both services share the same session:

- SQLAppConversationInfoService uses db_session from get_db_session()
- SQLAppConversationStartTaskService uses session from get_db_session()
- Both resolve to the same AsyncSession via DbSessionInjector.inject() (request state reuse)

2. First operation commits immediately:

```
# In sql_app_conversation_info_service.py:378-404
async def delete_app_conversation_info(self, conversation_id: UUID) -> bool:
    # ... execute delete query ...
    result = await self.db_session.execute(delete_query)
    await self.db_session.commit()  # ← COMMIT HERE
    return result.rowcount > 0
```

3. Second operation does not commit:

```
# In sql_app_conversation_start_task_service.py:224-245
async def delete_app_conversation_start_tasks(self, conversation_id: UUID) -> bool:
    # ... execute delete query ...
    result = await self.session.execute(delete_query)
    return result.rowcount > 0  # ← NO COMMIT
```

If delete_app_conversation_start_tasks raises an exception:
1. The first commit is already persisted; it cannot be rolled back.
2. The DbSessionInjector context manager will rollback only the uncommitted changes from the second operation.
3. Result: partial deletion — the conversation info is deleted, but the start tasks may remain if the second operation failed.

**Acceptance criteria:**
- The following line should be removed from the delete_app_conversation_info function.

```
await self.db_session.commit()
```

https://github.com/user-attachments/assets/b673537a-d2fb-483c-9c82-ab386f006349

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:793268f-nikolaik   --name openhands-app-793268f   docker.openhands.dev/openhands/openhands:793268f
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@hieptl/app-137#subdirectory=openhands-cli openhands
```

### DIFF
--- a/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
@@ -399,7 +399,6 @@ class SQLAppConversationInfoService(AppConversationInfoService):
 
         # Execute the secure delete query
         result = await self.db_session.execute(delete_query)
-        await self.db_session.commit()
 
         return result.rowcount > 0
 


tofarr (approved):
🍰

PR #11660, opened 2025-11-06
fix: always rewrite git token on resumes (paused and stopped)

ak684:
## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:34d32cf-nikolaik   --name openhands-app-34d32cf   docker.openhands.dev/openhands/openhands:34d32cf
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@alona/all-2989-bug-github_token-and-other-providers-eg-gitlab-expire-and-v999#subdirectory=openhands-cli openhands
```

### DIFF
--- a/tests/unit/runtime/test_runtime_git_tokens.py
+++ b/tests/unit/runtime/test_runtime_git_tokens.py
@@ -88,7 +88,9 @@ class MockRuntime(Runtime):
                 # Extract provider from previous clone command
                 if len(self.run_action_calls) > 0:
                     clone_cmd = (
-                        self.run_action_calls[0].command if self.run_action_calls else ''
+                        self.run_action_calls[0].command
+                        if self.run_action_calls
+                        else ''
                     )
                 if 'github.com' in clone_cmd:
                     mock_url = 'https://old_github_token@github.com/owner/repo.git'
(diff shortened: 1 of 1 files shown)


openhands/runtime/base.py
```diff
@@ -452,6 +452,121 @@ async def clone_or_init_repo(
...
+                            f'for {selected_repository}: '
+                            f'{remote_output or "unknown error"}'
+                        ),
+                    )
+
+                needs_clone = True
+                cleanup_before_clone = True
+
+        if cleanup_before_clone:
+            cleanup_action = CmdRunAction(command=f'rm -rf {quoted_repo_path}')
```
enyst, line 553: Is this really necessary? Sorry, I'm dense, I don't understand how it could happen in the first place, to have a repo and not be a git repo, but also, if it did happen how do we know the user could want to delete their work in this directory?
@ak684

PR #11671, opened 2025-11-07
Now removing V1 sandboxes in the V0 endpoint

tofarr:
## Summary of PR

In the V0 delete endpoint, we we now remove the sandbox as well as the conversation. This means we do not end up with orphaned sandboxes. Before this change the conversation within the sandbox would be stopped, but the sandbox itself would keep running.

## Change Type

- [X] Bug fix

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:b3d14f8-nikolaik   --name openhands-app-b3d14f8   docker.openhands.dev/openhands/openhands:b3d14f8
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@fix-v0-sandbox-removal#subdirectory=openhands-cli openhands
```

### DIFF
--- a/frontend/__tests__/utils/status.test.ts
+++ b/frontend/__tests__/utils/status.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "vitest";
-import { getStatusCode, getIndicatorColor, IndicatorColor } from "../status";
+import { getStatusCode, getIndicatorColor, IndicatorColor } from "#/utils/status";
 import { AgentState } from "#/types/agent-state";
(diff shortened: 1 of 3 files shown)


hieptl (approved):
Thank you! 🙏

PR #11681, opened 2025-11-09
fix: respect telemetry opt-out setting and prevent PostHog tracking when disabled

### LINKED ISSUE #11678: [Bug]: Telemetry is sent to posthog even when opted out

"Send anonymous usage data" is disabled:

Image

And this also is reflected in the settings json:

Image

However when I open the web app, I see many requests to posthog (which are being blocked by my browser):

Image

### OpenHands Installation

Docker command in README

### OpenHands Version

0.61 (docker.openhands.dev/openhands/openhands:0.61)

### Model Name

N/A

### Operating System

Linux

I'm using the docker.openhands.dev/openhands/openhands:0.61  container using the command from here:

https://docs.openhands.dev/openhands/usage/run-openhands/local-setup#option-2%3A-using-docker-directly

YakshithK:
**Summary of PR**  

This PR ensures that no telemetry or analytics events are sent to PostHog unless the user has explicitly consented by enabling "Send anonymous usage data" in settings. All PostHog tracking calls across the frontend are now guarded by a check for user consent. This resolves the issue where telemetry was sent even after opting out, bringing event tracking in line with user privacy expectations.

**Change Type**  

- [x] Bug fix  
- [ ] New feature  
- [ ] Breaking change  
- [ ] Refactor  

**Checklist**  

**Fixes**  
Resolves #11678 

**Release Notes**  

**Release Notes Description:**  
- Telemetry and analytics events are now only sent to PostHog if the user has explicitly opted in via settings. No analytics requests are sent by default or when opted-out, respecting user privacy.

### DIFF
--- a/frontend/src/components/features/chat/chat-interface.tsx
+++ b/frontend/src/components/features/chat/chat-interface.tsx
@@ -166,20 +166,24 @@ export function ChatInterface() {
     // Create mutable copies of the arrays
     const images = [...originalImages];
     const files = [...originalFiles];
-    if (totalEvents === 0) {
-      posthog.capture("initial_query_submitted", {
-        entry_point: getEntryPoint(
-          selectedRepository !== null,
-          replayJson !== null,
-        ),
-        query_character_length: content.length,
-        replay_json_size: replayJson?.length,
-      });
-    } else {
-      posthog.capture("user_message_sent", {
-        session_message_count: totalEvents,
-        current_message_length: content.length,
-      });
+
+    // Only capture events if user has consented and not opted out
+    if (!posthog.has_opted_out_capturing()) {
+      if (totalEvents === 0) {
+        posthog.capture("initial_query_submitted", {
+          entry_point: getEntryPoint(
+            selectedRepository !== null,
+            replayJson !== null,
+          ),
+          query_character_length: content.length,
+          replay_json_size: replayJson?.length,
+        });
+      } else {
+        posthog.capture("user_message_sent", {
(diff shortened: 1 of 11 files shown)


frontend/src/components/features/chat/chat-interface.tsx
```diff
@@ -166,20 +166,24 @@ export function ChatInterface() {
...
-    } else {
-      posthog.capture("user_message_sent", {
-        session_message_count: totalEvents,
-        current_message_length: content.length,
-      });
+
+    // Only capture events if user has consented and not opted out
+    if (!posthog.has_opted_out_capturing()) {
+      if (totalEvents === 0) {
+        posthog.capture("initial_query_submitted", {
```
DanTup, line 173: I'm unfamiliar with posthog, but this way seems like it would be easy to accidentally include new calls to `posthog.capture` in future without realising this should be checked first. I wonder if there's a way to do this inside `posthog.capture`, or have `posthog` be null when opted out and use `?.capture` in all these places?

(I'm completely unfamiliar with posthog or this code, this is just a driveby because I saw it linked to the issue I raised, feel free to disregard if it doesn't make sense here 🙂)

amanape:
Thanks for the PR, but this was a false alarm (details in https://github.com/OpenHands/OpenHands/issues/11678#issuecomment-3512446044).

P.S., we do not need to wrap PostHog capture events in conditionals like these; PostHog's global opt-in/out behavior already prevents all data collection for users who opt out

PR #11703, opened 2025-11-11
refactor(frontend): migrate from direct posthog imports to usePostHog hook

amanape:
## Summary of PR

https://posthog.com/docs/libraries/react

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:e97468e-nikolaik   --name openhands-app-e97468e   docker.openhands.dev/openhands/openhands:e97468e
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@chore/react-posthog#subdirectory=openhands-cli openhands
```

### DIFF
--- a/frontend/src/entry.client.tsx
+++ b/frontend/src/entry.client.tsx
@@ -74,7 +74,6 @@ prepareApp().then(() =>
             <HydratedRouter />
           </PostHogWrapper>
         </QueryClientProvider>
-        <div id="modal-portal-exit" />
       </StrictMode>,
     );
   }),
--- a/frontend/src/hooks/query/use-settings.ts
+++ b/frontend/src/hooks/query/use-settings.ts
@@ -1,6 +1,4 @@
 import { useQuery } from "@tanstack/react-query";
-import React from "react";
-import { usePostHog } from "posthog-js/react";
 import SettingsService from "#/settings-service/settings-service.api";
 import { DEFAULT_SETTINGS } from "#/services/settings";
 import { useIsOnTosPage } from "#/hooks/use-is-on-tos-page";
@@ -42,7 +40,6 @@ const getSettingsQueryFn = async (): Promise<Settings> => {
 };
 
 export const useSettings = () => {
-  const posthog = usePostHog();
   const isOnTosPage = useIsOnTosPage();
   const { data: userIsAuthenticated } = useIsAuthed();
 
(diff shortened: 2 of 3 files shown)


frontend/src/entry.client.tsx
```diff
@@ -8,17 +8,18 @@
...
 import OptionService from "./api/option-service/option-service.api";
 import { displayErrorToast } from "./utils/custom-toast-handlers";
 import { queryClient } from "./query-client-config";
 
-function PosthogInit() {
+function PostHogWrapper({ children }: { children: React.ReactNode }) {
   const [posthogClientKey, setPosthogClientKey] = React.useState<string | null>(
     null,
   );
+  const [isLoading, setIsLoading] = React.useState(true);
```
hieptl, line 22: **suggestion (non-blocking):** We specify a type when defining the posthogClientKey state. To maintain consistency, we should also specify a type when defining the isLoading state.

```js
  const [isLoading, setIsLoading] = React.useState<boolean>(true);
```

frontend/src/entry.client.tsx
```diff
@@ -27,20 +28,31 @@ function PosthogInit() {
...
-  React.useEffect(() => {
-    if (posthogClientKey) {
-      posthog.init(posthogClientKey, {
+  if (isLoading) {
+    return children;
+  }
+
+  if (!posthogClientKey) {
+    return children;
+  }
```
hieptl, line 39: **suggestion (non-blocking):** The two if statements return the same result, so they can be combined to simplify the code.

```js
if (isLoading || !posthogClientKey) {
  return children;
}
```

amanape:
```suggestion
  if (isLoading || !posthogClientKey) {
    return children;
  }
```

frontend/src/entry.client.tsx
amanape: It’s non-blocking since TypeScript already infers the type by the initial value we pass (`true`). The reason we explicitly pass a generic to `posthogClientKey` is because the initial value would infer the state type to `null` only, not `string | null`

PR #11715, opened 2025-11-12
feat(backend): enable sub-conversation creation using a different agent

hieptl:
## Summary of PR

**Context:**

According to the [Planning Agent documentation](https://www.notion.so/Planning-Agent-2a9ed25a91688098801debaac50d7b3f?source=copy_link), when a user switches to the Planning Agent from the UI, a sub-conversation must be created from the current conversation. This sub-conversation will:
- Inherit the same workspace as the parent conversation.
- Be used specifically for generating and managing the PLAN.md file.

**Acceptance Criteria:**
- The API successfully creates a new sub-conversation when invoked.
- The sub-conversation retains the same workspace as the parent conversation.

Example request payload.

```js
{
    "parent_conversation_id": "cd0a309c-9576-452d-a536-9d102fcbbc90",
    "agent_type": "plan"
}
```

https://github.com/user-attachments/assets/ebfa6104-e6dd-435e-9c28-2dc63256f3ab

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:f9abe99-nikolaik   --name openhands-app-f9abe99   docker.openhands.dev/openhands/openhands:f9abe99
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@hieptl/all-4165#subdirectory=openhands-cli openhands
```

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -173,9 +173,15 @@ class LiveStatusAppConversationService(GitAppConversationService):
 
         # Validate and inherit from parent conversation if provided
         if request.parent_conversation_id:
-            parent_info = await self._validate_and_get_parent_conversation(
-                request.parent_conversation_id
+            parent_info = (
+                await self.app_conversation_info_service.get_app_conversation_info(
+                    request.parent_conversation_id
+                )
             )
+            if parent_info is None:
+                raise ValueError(
+                    f'Parent conversation not found: {request.parent_conversation_id}'
+                )
             self._inherit_configuration_from_parent(request, parent_info)
 
         task = AppConversationStartTask(
@@ -465,29 +471,6 @@ class LiveStatusAppConversationService(GitAppConversationService):
             if exposed_url.name == AGENT_SERVER
         )
         return agent_server_url
-
-    async def _validate_and_get_parent_conversation(
-        self, parent_conversation_id: UUID
-    ) -> AppConversationInfo:
-        """Validate and retrieve parent conversation info.
-
-        Args:
-            parent_conversation_id: The ID of the parent conversation to validate
-
-        Returns:
-            The parent conversation info if valid
-
-        Raises:
-            ValueError: If parent conversation not found
-        """
-        parent_info = (
-            await self.app_conversation_info_service.get_app_conversation_info(
-                parent_conversation_id
-            )
-        )
-        if parent_info is None:
-            raise ValueError(f'Parent conversation not found: {parent_conversation_id}')
-        return parent_info
 
     def _inherit_configuration_from_parent(
         self, request: AppConversationStartRequest, parent_info: AppConversationInfo


openhands/app_server/app_conversation/sql_app_conversation_info_service.py
```diff
@@ -307,6 +308,11 @@ async def save_app_conversation_info(
             llm_model=info.llm_model,
             conversation_version='V1',
             sandbox_id=info.sandbox_id,
+            parent_conversation_id=(
+                str(info.parent_conversation_id)
+                if info.parent_conversation_id
```
enyst, line 313: Out of curiosity, could you please tell, does the app-server support any idea of “agent delegation” yet?

The tiny curious thing is that a “parent conversation” already has an established meaning in delegation. The switch to planner from a UI is, in some sense, like a delegation with a different trigger than the agent’s tool

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -452,11 +466,72 @@ def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
...
+        # Inherit git parameters from parent if not provided
+        if not request.selected_repository:
+            request.selected_repository = parent_info.selected_repository
+        if not request.selected_branch:
+            request.selected_branch = parent_info.selected_branch
+        if not request.git_provider:
+            request.git_provider = parent_info.git_provider
+
+        # Inherit LLM model from parent if not provided
+        if not request.llm_model and parent_info.llm_model:
```
enyst, line 502: I love this! I take it we are considering to support the user choosing a different model in some way, when they use planning agent? ❤️

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -452,11 +466,72 @@ def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
...
+        """
+        parent_info = (
+            await self.app_conversation_info_service.get_app_conversation_info(
+                parent_conversation_id
+            )
+        )
+        if parent_info is None:
+            raise ValueError(f'Parent conversation not found: {parent_conversation_id}')
+        # Security check: ensure parent conversation belongs to same user
+        if parent_info.created_by_user_id != user_id:
```
enyst: Is this necessary here, in `openhands/app_server/app_conversation/live_status_app_conversation_service.py` ?

It seems to me that authorization code should happen before it gets here. Hopefully in the creation/validation of the request, that is, somewhere in or where `UserContext` is injected

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -168,6 +170,14 @@ async def _start_app_conversation(
     ) -> AsyncGenerator[AppConversationStartTask, None]:
         # Create and yield the start task
         user_id = await self.user_context.get_user_id()
+
+        # Validate and inherit from parent conversation if provided
+        if request.parent_conversation_id:
+            parent_info = await self._validate_and_get_parent_conversation(
+                request.parent_conversation_id, user_id
```
enyst: Ah, I see why… 🤔

The thing is, in `_validate_and…` method, the `validate` section is authorization, is security related. So it feels to me that like maybe that validatio code could be up in the stack a bit, where the rest of the auth/authz parameters are validated. 
Cc: @tofarr  I could be wrong though?

openhands/app_server/app_conversation/sql_app_conversation_info_service.py
hieptl: Hi @enyst,

At present, we do not support Agent Delegation.

I am currently developing a feature that will enable users to switch between the execution agent and the planning agent.

After discussions, we have decided that when a user starts a new conversation, they will have the option to switch to the planning agent through the user interface. At this point, a new conversation will be created as a sub-conversation of the current conversation, with the existing conversation serving as the parent.

The sub-conversation will utilize the planning agent to generate the PLAN.md file.

After the user is satisfied with the content of the PLAN.md file, they can switch back to the execution agent to implement the plan.

Both the parent conversation and its sub-conversation will share the same workspace, ensuring continuity and context across the workflow.

Thank you! 🙏

openhands/app_server/app_conversation/live_status_app_conversation_service.py
hieptl: Hi @enyst,

Yes, we plan to support switching between different models in the near future, with the design already in place.

For now, as we focus on the MVP for the planning agent, users can switch between the execution agent and the planning agent. Currently, sub-conversations share the same LLM model as their parent, but model switching will be supported soon.

Thank you! 🙏

openhands/app_server/app_conversation/live_status_app_conversation_service.py
hieptl: Hi @enyst,

I don’t believe this logic is related to user authorization. Its purpose is to validate the request payload when a user creates a new sub-conversation. For example, the request payload looks like this:

```js
{
    "parent_conversation_id": "cd0a309c-9576-452d-a536-9d102fcbbc90",
    "agent_type": "plan"
}
```

The highlighted logic ensures that the provided parent_conversation_id exists and belongs to the current user.

Without this validation, for instance, in OpenHands Cloud, if someone knows another user’s conversation_id, they could potentially create a sub-conversation under that conversation_id, even if it doesn’t belong to them.

Thank you! 🙏

openhands/app_server/app_conversation/live_status_app_conversation_service.py
hieptl: Hi @enyst, I’ve explained the reason behind this logic in my previous [comment](https://github.com/OpenHands/OpenHands/pull/11715#discussion_r2519182518). Please let me know if you have any further questions. Thank you! 🙏

openhands/app_server/app_conversation/live_status_app_conversation_service.py
enyst: Right, exactly as you said. That’s why it seems to me that this validation code belongs:
- **in the Saas, not in OpenHands**
- in the validation/authorization code or following the same patterns as the existing validation/authorization code. Similar maybe, with how we check that a user can access to a sandbox, or the conversation?

`UserContext` is injected here from a FastAPI dependency: on the Saas, it runs its own validations and what it needs. I’m just wondering, could we consider following the same pattern for validating parent conversation access? 

(That, it seems to me, very likely implies that this code should be in enterprise, if we can.)

enyst: I think of it this way: this is security related, so it’s very important to be sure to do these checks as early as possible, so that we don’t run code from an unauthorized users on another user’s conversation.

As early as possible, could mean just before the request execution starts: when FastAPI figures out what dependencies are, and what to inject in the request parameters. One such parameter is `UserContext` and code related to it: which means we can run a different code in the Saas, than in OpenHands.

I’m sorry if it’s unclear or maybe I’m wrong, I’m speaking offhand directly on github, and I express myself poorly perhaps. We are introducing here some authorization code (_is user allowed to access parent id_), and I’m just wondering if we can somehow follow the patterns for injector code that we already have in the codebase, for an issue like this? 
(How did we do this for, for example, restoring a conversation?)

enyst (dismissed):
Thank you for this, very interesting PR!

I’d appreciate to pause a bit to figure out what happens here, where is this going, and more importantly, it seems to me, how are we doing this. Comments inline, but I mainly have two questions, maybe with a nit of the third:
- the Planning Agent notion doc linked in the PR is not accessible in the open. I think maybe it should be. I’ll take this part to slack :) _Edit_: done, please [see](https://openhands-ai.slack.com/archives/C09FJAZSP9N/p1762970230393229)
- validation / authorization code for a user belongs in /enterprise, not in OpenHands, according to V1 code design patterns, and I’m hoping we could consider following them. Unless there’s a reason why we can’t?
- Nit: I’d love to understand a bit the relation between this feature and agent delegation… Any thought would be appreciated!

openhands/app_server/app_conversation/live_status_app_conversation_service.py
hieptl: https://github.com/OpenHands/OpenHands/blob/8192184d3e07dfc158e9c0ffd26d53ec9df77d74/openhands/app_server/app_conversation/sql_app_conversation_info_service.py#L316

I reviewed the code again and confirmed that we use the __secure_select function before fetching a V1 conversation. This function already includes a condition to verify whether the current conversation belongs to the current user. We can refer to the link above for more information.

Therefore, the if statement you highlighted can be removed, as the __secure_select function is consistently used across the codebase.

I’m not entirely sure if this addresses your question, but I hope it helps clarify the logic.

hieptl:
Hi @xingyaoww,

Could you please help answer this question from @enyst?

> Nit: I’d love to understand a bit the relation between this feature and agent delegation… Any thought would be appreciated!

Thank you! 🙏

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -452,11 +466,66 @@ def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
         )
         return agent_server_url
 
+    async def _validate_and_get_parent_conversation(
```
enyst: Since we removed the `validate_` part of the code, we only retrieve the parent conversation now, could we do this directly in the caller, and perhaps remove this method? Just because well, it only does a call now

openhands/app_server/app_conversation/live_status_app_conversation_service.py
enyst: So this does address my point, thank you! I think in V1, in V1 code, we should not need to validate/check user authorization in OpenHands. If we do find ourselves writing code for that, it should be in /enterprise, or I would love to know so we can think of alternative ways.

In this case, if it’s simply unnecessary, that addresses my concern perfectly!

openhands/app_server/app_conversation/sql_app_conversation_info_service.py
enyst: > At present, we do not support Agent Delegation.
> I am currently developing a feature that will enable users to switch between the execution agent and the planning agent.

Just for clarity here: we do support Agent Delegation in agent-sdk. The high-level view on delegation looks like this PR: it’s a separate conversation, spawned from a parent conversation.

enyst: That’s why I wonder if we could / should think of reusing that? Is it different because it’s decided by the user? _Should_ it be different because it’s decided by the user? Or are we duplicating work or making things more complex by creating another code path when we kind of had one?

Sorry, just so you understand why I’m a bit interested, that’s all. Maybe it’s okay, actually, I can see a case for that, too… Independent conversations from agent-sdk perspective, their relationship is handled in the app-server?

openhands/app_server/app_conversation/live_status_app_conversation_service.py
enyst: Nit: maybe that’s an easy way to we could avoid to name it “validate”, just a thought so that it doesn’t confuse us in the future

openhands/app_server/app_conversation/sql_app_conversation_info_service.py
enyst: These are just side questions, nothing blocking, just was a bit surprising

openhands/app_server/app_conversation/live_status_app_conversation_service.py
enyst: Are you sure though? This request has a `conversation_id` that it didn’t have before. We’re not just supposed to authorize the user to start a random conversation, but to validate access to that parent

openhands/app_server/app_conversation/live_status_app_conversation_service.py
hieptl: Hello @enyst,

I’ve removed that function and incorporated the logic directly into the caller. Thank you for your feedback! 🙏

openhands/app_server/app_conversation/live_status_app_conversation_service.py
hieptl: If the request payload includes a parent_conversation_id, the system will perform two validations:
1. Verify that the specified parent conversation exists.
2. Confirm that the current user has the necessary access permissions for that conversation.

```python
 # Validate and inherit from parent conversation if provided
if request.parent_conversation_id:
    parent_info = (
        await self.app_conversation_info_service.get_app_conversation_info(
            request.parent_conversation_id
        )
    )
    if parent_info is None:
        raise ValueError(
            f'Parent conversation not found: {request.parent_conversation_id}'
        )
    self._inherit_configuration_from_parent(request, parent_info)
```

According to the logic above, we call

```python
parent_info = (
    await self.app_conversation_info_service.get_app_conversation_info(
        request.parent_conversation_id
    )
)
```

For additional details about the logic implemented in get_app_conversation_info, please refer to the following link.

https://github.com/OpenHands/OpenHands/blob/8192184d3e07dfc158e9c0ffd26d53ec9df77d74/openhands/app_server/app_conversation/sql_app_conversation_info_service.py#L234

Thank you! 🙏

tofarr (approved):
LGTM! I wonder if at a later juncture we should also add `parent_conversation_id__eq` to the filter parameters of `search_app_conversations` / `search_app_conversation_info` ?

hieptl:
> LGTM! I wonder if at a later juncture we should also add `parent_conversation_id__eq` to the filter parameters of `search_app_conversations` / `search_app_conversation_info` ?

Hi @tofarr,

Yes, once a sub-conversation is created and the Search Conversations API is called, both the parent conversation and the sub-conversation will be returned in the response.

I am currently working on two separate PRs:
1. A PR to exclude sub-conversations when calling the Search Conversations API.
2. A PR to include sub_conversation_ids when fetching conversation details.

Thank you! 🙏

xingyaoww:
> Nit: I’d love to understand a bit the relation between this feature and agent delegation… Any thought would be appreciated!

Hey @enyst, the high-level idea here is that "planning mode" is a special case for delegation that we've implemented in SDK. 

Typically sub-agent delegation means the "parent agent is responsible to decide when to delegate". But switch between plan/execute is actually the responsible of "user" rather than "parent agent", yet it still re-use the components we've build for agent delegation under the hood (i.e., a special case!).

LMK if it make sense to you!

PR #11716, opened 2025-11-12
SWE-fficiency benchmark implementation

18jeffreyma:
## Summary of PR

Adding SWE-fficiency inference implementation to OpenHands benchmark. https://swefficiency.com/ 

## Change Type

- [ ] Bug fix
- [ x ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

- [ x ] I have read and reviewed the code and I understand what the code is doing.

## Fixes

In addition to addding benchmark implementation, this change also fixes an issue with parallel evaluation (where one finish evaluation would crash others). This was due to one eval finishing and terminating other containers when rm_all_containers is True.

## Release Notes

- [ x ] Include this change in the Release Notes.

### DIFF
--- a/evaluation/benchmarks/swefficiency/README.md
+++ b/evaluation/benchmarks/swefficiency/README.md
@@ -63,4 +63,3 @@ then your command would be:
 ### 2. Run the SWE-fficiency benchmark official evaluation
 
 Once the output is converted, use the [official SWE-fficiency benchmark evaluation](https://github.com/swefficiency/swefficiency) to evaluate it.
-
--- a/evaluation/benchmarks/swefficiency/run_infer.py
+++ b/evaluation/benchmarks/swefficiency/run_infer.py
@@ -5,7 +5,6 @@ import json
 import multiprocessing
 import os
 import tempfile
-import docker
 from typing import Any, Literal
 
 import pandas as pd
@@ -16,9 +15,6 @@ import openhands.agenthub
 from evaluation.benchmarks.swe_bench.binary_patch_utils import (
     remove_binary_diffs,
     remove_binary_files_from_git,
-)
-from evaluation.benchmarks.swe_bench.resource.mapping import (
-    get_instance_resource_factor,
 )
 from evaluation.utils.shared import (
     EvalException,
@@ -39,8 +35,8 @@ from openhands.controller.state.state import State
 from openhands.core.config import (
     AgentConfig,
     OpenHandsConfig,
-    get_llm_config_arg,
     get_evaluation_parser,
+    get_llm_config_arg,
 )
 from openhands.core.config.condenser_config import NoOpCondenserConfig
 from openhands.core.config.utils import get_condenser_config_arg
@@ -114,11 +110,10 @@ Please remember that you should not change the implementation of the `workload()
     return MessageAction(content=instruction)
 
 
-
 def get_instance_docker_image(
     instance_id: str,
 ) -> str:
-    return f"ghcr.io/swefficiency/swefficiency-images:{instance_id}"
+    return f'ghcr.io/swefficiency/swefficiency-images:{instance_id}'
 
 
 def get_config(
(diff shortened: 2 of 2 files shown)


enyst:
Hey @18jeffreyma this is great and so fast, thank you! However, please note that this is the directory for OpenHands V0. We are transitioning to the next major version, OpenHands V1, and we have another repo, benchmarks, for V1. I understand that SWE-fficiency was run on V0? At this time, we are already using the new repo…

I’m just concerned that under the current transition, maybe you wouldn’t actually want to put work in adapting to this repo… 😓 unless I’m missing something and you are sure this should be here.

18jeffreyma:
ah thanks for the catch! I forked OpenHands back in March well before the migration. Is there a guide to the new repository I can instead rebase against?

Happy to implement a V1 version and keep this fork for historical purposes, and instead have a new implementation this be part of the new V1 if you can share more details on differences?

enyst:
V1 is a major architectural rebuild of OpenHands, based on [agent-sdk](https://github.com/OpenHands/software-agent-sdk)
- the benchmarks are in [this benchmarks repo](https://github.com/OpenHands/benchmarks)
- arxiv paper documenting V1 agent-sdk: [The OpenHands Software Agent SDK](https://arxiv.org/abs/2511.03690)

Cc: @juanmichelini and @xingyaoww know better what changes the benchmark code adaptation needed

.openhands/precommit.sh
neubig, line 1: I think this is an unintended change, could we revert it?

neubig (commented):
Hi @18jeffreyma , I think we can merge in this for now. We're eventually going to port over the benchmarks to the new https://github.com/OpenHands/benchmarks repo, but I think having this documented is better than not (and porting might take a little bit).

Would you mind splitting the fixes of the docker containers into a separate PR? I think we can merge in the `evaluation/benchmarks/swefficiency` changes right away, since they're low risk but we might need to take a closer look at the docker runtime changes to make sure they don't have unintended consequences.

.openhands/precommit.sh
18jeffreyma: sounds good will clean up this PR and reply back when done!

mamoodi:
@neubig should this be moved over to the benchmarks repo?

openhands/runtime/impl/docker/docker_runtime.py
```diff
@@ -606,12 +633,221 @@ def close(self, rm_all_containers: bool | None = None) -> None:
...
+                try:
+                    docker_client.images.prune({'dangling': True})
+                except Exception:
+                    pass
+            finally:
+                docker_client.close()
+        except Exception as e:
+            logger.warning(f'Error during base image cleanup: {e}')
+
+    def _find_dependent_images(self, docker_client: docker.DockerClient, base_image: str) -> list[str]:
```
enyst: Could we perhaps let the docker changes out of this PR?

I think you are correct that there were issues, but we are transitioning to a better runtime handling in V1. I think it's like, anytime now this docker code will likely be deleted anyway, and on the PR docker tests are failing so they are blocking merge...

18jeffreyma:
Removed the docker implementation fixes + fixed previous merge issue, should be ready for review!

Apologies for the delay!

enyst:
@openhands CI lint / python lint job is failing on this PR. Run pre-commit run --all-files --show-diff-on-failure --config ./dev_config/python/.pre-commit-config.yaml to autofix it, and push your changes

Note that its branch is `18jeffreyma:main` so from a fork, use the right remote.

enyst (approved):
Thank you!

It would be awesome if you could consider to submit it to [OpenHands benchmarks](https://github.com/OpenHands/benchmarks) repository for the V1 version of OpenHands. It's similar, and of course similarly it had SWE-bench integrated to help document the tweaks to add other benchmarks, but we hope it's much easier both to work with the agent loop and to run.

18jeffreyma:
@enyst would love to! I'll work on a PR and get that up soon!

PR #11727, opened 2025-11-13
Do not get live status updates when they are not required

tofarr:
## Summary of PR

Getting a live status update means pinging the sandboxed runtime. If this is not required, then do not do it!

## Change Type

- [X] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:547184b-nikolaik   --name openhands-app-547184b   docker.openhands.dev/openhands/openhands:547184b
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@feat-improve-perf#subdirectory=openhands-cli openhands
```


hieptl (approved):
Thank you! 🙏 

Since we delete the app_conversation before deleting the sandbox, is it possible for the app_conversation to be deleted successfully while the associated sandbox deletion fails?

Are all deletion operations executed within the same transaction?

tofarr:
> Since we delete the app_conversation before deleting the sandbox, is it possible for the app_conversation to be deleted successfully while the associated sandbox deletion fails?

Yes - it is possible that a sandbox would remain after an associated conversation was deleted - And now that you mention it, it, I'll need to update the transaction code too. (This works in the docker runtime but will likely fail in the remote runtime.)

Great catch!.

hieptl (approved):
Thank you! 🙏

PR #11767, opened 2025-11-18
fix: prevent newline accumulation in XML parameter serialization

aoi127:
### Problem Description

When converting function calling format to XML format for non-native function calling LLMs, the converter was adding extra newlines around multiline parameter values. This caused newlines to accumulate across multiple LLM response cycles, leading to:

1. **str_replace failures** due to mismatched old_str with excessive newlines
2. **Generated code containing redundant blank lines** that grow over time

### Reproduction Example

The newline accumulation follows this pattern:

- **1st LLM response**: `<parameter=old_str>\n...code...\n</parameter>`
- **After context rebuild**: `<parameter=old_str>\n\n...code...\n\n</parameter>`
- **2nd LLM response** learns this format: `\n\n...code...\n\n`
- **After rebuild again**: `<parameter=old_str>\n\n\n...code...\n\n\n</parameter>`
- Newlines keep accumulating with each request-response cycle

### Root Cause

The `convert_tool_call_to_string()` function in `openhands/llm/fn_call_converter.py` detected multiline content (via `is_multiline` check) and added `\n` before and after parameter values. These extra newlines were preserved in context reconstruction, creating a **positive feedback loop** where:

1. LLM sees inflated format in context
2. LLM learns and reproduces this format
3. Converter adds more newlines
4. Cycle repeats with growing newlines

### Solution

Removed the `is_multiline` logic that added extra `\n` characters. Parameter values are now serialized as-is without additional formatting.

### Modified Files

- `openhands/llm/fn_call_converter.py` (lines 423-434)
  - Removed `is_multiline` variable and conditional newline insertion
  - Simplified parameter serialization to preserve original content

```python
# Before:
for param_name, param_value in args.items():
    is_multiline = isinstance(param_value, str) and '\n' in param_value
    ret += f'<parameter={param_name}>'
    if is_multiline:
        ret += '\n'  # Extra newline added
    # ... value ...
    if is_multiline:
        ret += '\n'  # Extra newline added
    ret += '</parameter>\n'

# After:
for param_name, param_value in args.items():
    ret += f'<parameter={param_name}>'
    # ... value ...
    ret += '</parameter>\n'
```

#### Positive Impacts
- ✅ Fixes `str_replace_editor` tool reliability
- ✅ Prevents code quality degradation over long conversations
- ✅ Maintains backward compatibility with existing prompts
- ✅ Reduces confusion from growing blank lines in generated code

#### Testing
- Manual testing shows parameter values are now preserved exactly
- No additional newlines are added during serialization
- Context reconstruction produces identical format to LLM output

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist


mamoodi:
@enyst is this also something that is in agent-sdk now?

enyst:
@aoi127 Thank you for the contribution! Do you have some prompt log where we can see the issue, maybe? Reason I ask is that I don't fully understand it, sorry, I think the tool prompts are recomputed every request...


On the other hand, as Mamoodi said, we are launching OpenHands V1, which is deeply rebuilt and it's based on [agent SDK](https://github.com/OpenHands/software-agent-sdk). The LLM is in the agent-sdk, since it's part of the agentic loop.

That means that in really short time we will actually remove llm.py here, and replace with using agent-sdk as a dependency. 

I'd suggest to make an issue on agent-sdk, maybe with a log if you can?

aoi127:
[logs_251121_001.zip](https://github.com/user-attachments/files/23670021/logs_251121_001.zip)

@enyst Hello, I’ve uploaded a zip file containing logs from the `/app/logs/` directory in the Docker container. These logs demonstrate the "newline accumulation" issue.

## Case 1

image

In `llm/25-11-21_03-29/response_012.log`, the `file_text` parameter only contains **two** leading newline characters (`\n`).

image

However, in `llm/25-11-21_03-29/prompt_037.log`(which contains all messages), there are actually **three** `\n` after `<parameter=file_text>` in the reconstructed assistant message.

## Case 2

image
image
image
image
image
image

As shown, each reconstructed assistant message contains **one additional `\n`** compared to the original function call.

The number of newline characters increases with each call to `str_replace_editor`, indicating a positive feedback loop.

## Root Cause

As noted in the earlier comment, the code bug creates a **positive feedback loop** that accumulates extra newlines.

aoi127:
@enyst I'm not sure if the agent-sdk has this same issue. I'll test it and submit a PR once I have the chance.

enyst:
Thank you @aoi127 ! It would be great if you could check out agent-sdk!

Maybe if you could take a look at the failing test, I think it's failing precisely because of newlines. It would be great if we could have a unit test for this exact behavior, too, so that we can reproduce it and make sure it doesn't happen again.

aoi127:
Hi @enyst, I looked into the failing test. It appears that the faling test case itself was introducing the redundant newlines, which matches the behavior we saw in the logs.
I adjust the test case, do you think it's OK?

enyst:
I think it's fine for now! Not sure for agent-sdk, we might do some more tests over there.

Let's get this one in, first, though!

enyst:
@openhands Do a /codereview-roasted on this PR, including on the discussion in this PR. Read all comments here and look at the screenshots if you can, so you understand very well what is reported. You should think deeply and feel free to tell if you agree it's a real problem, too, and/or where it happens, or could it be a mistake.

From now on, everytime you respond to the user, use markdown, with headings, bold, bulleted lists, so that we can read it easily.

enyst (approved):
Thank you!

PR #11768, opened 2025-11-18
Bumped Software Agent SDK and fixed V1 Delete

tofarr:
## Summary of PR

Bump SDK to latest to test V1 fixes. Also fixed issue where deleting a V1 conversation was failing in the remote runtime due to a regression.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:eb63f33-nikolaik   --name openhands-app-eb63f33   docker.openhands.dev/openhands/openhands:eb63f33
```

CLI with uvx:
```
uvx --python 3.12 --from git+https://github.com/OpenHands/OpenHands@bump-sdk#subdirectory=openhands-cli openhands
```


openhands/server/routes/manage_conversations.py
```diff
@@ -527,10 +536,14 @@ async def _try_delete_v1_conversation(
             result = await app_conversation_service.delete_app_conversation(
                 app_conversation_info.id
             )
+
+            # Manually commit so that the conversation will vanish from the list
+            await db_session.commit()
```
hieptl, line 543: **question:** Am I correct in understanding that if we proceed this way, deleting the sandbox will occur in a separate transaction, rather than in the same transaction as deleting app_conversation?

tofarr: Deleting the sandbox can hit the runtime api which can be slow - so 2 small transactions was easier than a single one.

hieptl: Yes, thank you! 🙏

PR #11772, opened 2025-11-18
fix(backend): ensure microagents are loaded for V1 conversations

hieptl:
## Summary of PR

All of our microagents, including the GitHub microagent, are currently non-functional. For example, when prompted with ‘flarglebargle’, the system fails to recognize or process it.

This pull request ensures that microagents are properly loaded for V1 conversations.

Please refer to the video below for a demonstration of the PR’s output.

https://github.com/user-attachments/assets/382ed1f3-e5d7-4857-83a8-80c21240d47c

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:cc32a12-nikolaik   --name openhands-app-cc32a12   docker.openhands.dev/openhands/openhands:cc32a12
```


openhands/app_server/app_conversation/skill_loader.py
```diff
@@ -0,0 +1,362 @@
...
+        )
+        if result.exit_code == 0 and result.stdout.strip():
+            return result.stdout
+        return None
+    except Exception as e:
+        _logger.debug(f'Failed to read file {file_path}: {str(e)}')
+        return None
+
+
+def _create_skill_from_content(filename: str, content: str) -> Skill | None:
```
enyst: We could inline this function, it only loads a skill?

openhands/app_server/app_conversation/microagent_loader.py
```diff
@@ -0,0 +1,362 @@
...
+        working_dir: Working directory path
+
+    Returns:
+        List of Skill objects loaded from repository.
+        Returns empty list on errors.
+    """
+    try:
+        # Determine repository root directory
+        repo_root = _determine_repo_root(working_dir, selected_repository)
+        _logger.info(f'Loading repo microagents from {repo_root}')
```
enyst: I'm not sure this is the case: I thought we're loading skills from `.openhands/skills`, not repo root?

openhands/app_server/app_conversation/microagent_loader.py
```diff
@@ -0,0 +1,362 @@
...
+        )
+
+        return skills
+
+    except Exception as e:
+        _logger.warning(f'Failed to load global microagents: {str(e)}')
+        return []
+
+
+def load_user_microagents() -> list[Skill]:
```
enyst: I'd suggest to inline this, since we only wrap the SDK call to load skills

It gets confusing: microagents are skills. Literally, we have renamed them to skills. There is no conversion here, it was just a rename, a full rename.

openhands/app_server/app_conversation/microagent_loader.py
hieptl: Hi @enyst,

Currently, when accessing the UI, selecting a repository, and initiating a V1 conversation, microagents are not triggered when a trigger keyword is entered—provided that the selected repository contains an .openhands/microagents directory.

To address this issue, we need to load the repository’s microagents, convert them into skills, and pass those skills to the agent using the software-agent-sdk.

Thank you! 🙏

enyst: I think there is some confusion here, microagents = skills, they're the same files.

But there are three locations here:
- repo root - there are no microagents/skills in repo root, afaik
- old location - `.openhands/microagents` - repo skills might be there, for legacy reasons
- new location - `.openhands/skills` - repo skills will be here from now on

enyst: I mean, I hear you, I just think we simply need to load from "old location" and "new location". Does that sound right?

enyst: @hieptl Here we basically decided to rename "microagents" to "skills" - completely: [slack discussion](https://openhands-ai.slack.com/archives/C06R25BT5B2/p1760288393550609?thread_ts=1760155584.116699&cid=C06R25BT5B2)

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -478,6 +487,96 @@ def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
         agent_server_url = replace_localhost_hostname_for_docker(agent_server_url)
         return agent_server_url
 
+    async def _load_and_merge_all_microagents(
```
enyst:
```suggestion
    async def _load_and_merge_all_skills(
```

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -478,6 +487,96 @@ def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
...
+        self,
+        remote_workspace: AsyncRemoteWorkspace,
+        selected_repository: str | None,
+        working_dir: str,
+    ) -> list:
+        """Load microagents from all sources and merge them.
```
enyst:
```suggestion
        """Load skills from all sources and merge them.
```

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -478,6 +487,96 @@ def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
...
+
+        Args:
+            remote_workspace: AsyncRemoteWorkspace for loading repo microagents
+            selected_repository: Repository name or None
+            working_dir: Working directory path
+
+        Returns:
+            List of merged Skill objects from all sources
+        """
+        _logger.debug('Loading microagents for V1 conversation')
```
enyst:
```suggestion
        _logger.debug('Loading skills for V1 conversation')
```

openhands/app_server/app_conversation/microagent_loader.py
```diff
@@ -0,0 +1,362 @@
...
+        except Exception as e:
+            _logger.warning(
+                f'Failed to load global microagent from {file_path}: {str(e)}'
+            )
+
+    return skills
+
+
+def load_global_microagents() -> list[Skill]:
+    """Load global microagents from OpenHands/microagents/ directory.
```
enyst: Aha, we will want to rename the global directory... I'm not sure if in this PR, whenever convenient for you, but I'd suggest we need to rename it before V1 release IMHO

openhands/app_server/app_conversation/microagent_loader.py
```diff
@@ -0,0 +1,362 @@
...
+
+    Returns:
+        List of Skill objects loaded from global microagents directory.
+        Returns empty list if directory doesn't exist or on errors.
+    """
+    microagent_dir = Path(GLOBAL_MICROAGENTS_DIR)
```
enyst: 🤔 We could maybe rename it now, it seems easier to just deal with all of it, otherwise there's this kind of code in this PR that will need refactored

So for global skills, it seems it would be similar: `load_global_skills` function should load from the OH path `/skills`, and support legacy OH path `/microagents`?

openhands/app_server/app_conversation/microagent_loader.py
hieptl: We currently have approximately [24 global microagents](https://github.com/OpenHands/OpenHands/tree/main/microagents)￼. In addition, individual repositories can include repository-level microagents, such as [this example](https://github.com/hieptl/hieptl/tree/main/.openhands/microagents)￼.

When initiating a new V1 conversation from the UI, the following function is called:

https://github.com/OpenHands/OpenHands/blob/494eba094fbc8bb9cdf6a27d358cabde82e1b41d/openhands/app_server/app_conversation/live_status_app_conversation_service.py#L173

Subsequently, a start conversation request is constructed to call the agent_server API in the software-agent-sdk.

https://github.com/OpenHands/OpenHands/blob/494eba094fbc8bb9cdf6a27d358cabde82e1b41d/openhands/app_server/app_conversation/live_status_app_conversation_service.py#L511

Currently, the default agent retrieved from the software-agent-sdk does not have awareness of either the 24 global microagents or any repository-level microagents.

To address this, it is necessary to convert the list of microagents into skills and update the agent_context accordingly.

openhands/app_server/app_conversation/microagent_loader.py
```diff
@@ -0,0 +1,362 @@
...
+        return []
+
+    try:
+        _logger.info(f'Loading global microagents from {microagent_dir}')
+
+        # Find all .md files in the directory
+        md_files = _find_global_microagent_files(microagent_dir)
+
+        # Load skills from the found files
+        skills = _load_global_microagent_files(md_files, microagent_dir)
```
enyst: It might be a bit more readable if we combine these functions, into one function: find the files and load them, rather than passing them around only to load them

openhands/app_server/app_conversation/microagent_loader.py
```diff
@@ -0,0 +1,362 @@
+"""Utilities for loading microagents and converting them to Skills for V1 conversations.
```
enyst:
```suggestion
"""Utilities for loading Skills.
```

openhands/app_server/app_conversation/microagent_loader.py
hieptl: Hi @enyst,

I am aware of the renaming from “microagents” to “skills.” While the software-agent-sdk uses the new terminology, the OpenHands codebase still contains many references to “microagents,” including a UI page titled “Microagent Management.”

Since my changes are within the OpenHands repository, I was initially unsure which term to use.

However, I will follow your suggestion and am happy to proceed with the renaming. For this PR, I will rename only the files affected by my changes. Other references can be updated in a separate PR, as renaming everything is outside the scope of this current PR.

Thank you! 🙏

openhands/app_server/app_conversation/microagent_loader.py
enyst: @hieptl Sorry, I wasn't clear. I don't disagree that we need to load legacy microagents.

I'm just saying, we have renamed all across the SDK, "microagent" to "skill".

So all we need to do, is still support _the old locations._ There is no conversion. All 24 need loaded from the old location. That's all they need, they are already Skills.

Let me put it this way: I imagine we rename the global "microagents" directory to "skills". 
We did it already:
- in the SDK, we renamed the directory of the user skills (it's now `~/.openhands/skills`)
- in the CLI, we renamed the directory of the repo skills (we renamed it to `.openhands/skills`)
- in app-server, we need to do it too... rename OH path `microagents` to `skills`

Then the code only needs to load skills from the global directory (renamed)... if that makes sense?

hieptl: I should also support the path ./openhands/skills in addition to .openhands/microagents within a repository.

openhands/app_server/app_conversation/microagent_loader.py
enyst: Of course, that works, thank you! I think maybe we don't need FE renames yet, or others.

I did realize during our discussion that we probably need the rename of the global directory though. Otherwise, that seems the most confusing right now, because that one is exactly the 3rd location, when we renamed the other 2 locations in SDK and CLI, it's app-server's turn. 🙏

openhands/app_server/app_conversation/microagent_loader.py
enyst: Yes, the CLI app did [here](https://github.com/OpenHands/OpenHands-CLI/blob/e8ec2a8dd804e047b1f1050b6c4e0b2db6c811aa/openhands_cli/tui/settings/store.py#L42)

hieptl: Hello @enyst, I have renamed the microagents folder to skills. Thank you for your feedback! 🙏

openhands/app_server/app_conversation/microagent_loader.py
hieptl: Hello @enyst, I have renamed the microagents folder to skills. Thank you for your feedback! 🙏

openhands/app_server/app_conversation/microagent_loader.py
hieptl: Hello @enyst, I have updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/app_server/app_conversation/skill_loader.py
hieptl: Hello @enyst, I have updated the code based on your feedback. Thank you for your feedback! 🙏

enyst (approved):
Thank you!

hieptl:
Hello @tofarr, I’ve updated the code based on your feedback. You can refer to the video below for the latest changes. As shown in the video, a new status named Setting Up Skills has been introduced. Thank you for your feedback. 🙏

https://github.com/user-attachments/assets/b8d363cd-b27e-4b90-9496-fd67cb768cb3

PR #11824, opened 2025-11-26
fix(backend): unable to start a new V0 conversation

hieptl:
## Summary of PR

**Description:**
- It is currently not possible to start a new V0 conversation. When running the latest code from the main branch and attempting to create a new V0 conversation through the UI, the following error is encountered:

```
openhands:ERROR: session.py:301 - Error creating agent_session: provider_tokens must be a MappingProxyType, got dict
  File "/Users/hieple/Documents/openhands/OpenHands/openhands/server/session/session.py", line 267, in initialize_agent
    await self.agent_session.start(
  File "/Users/hieple/Documents/openhands/OpenHands/openhands/server/session/agent_session.py", line 135, in start
    runtime_connected = await self._create_runtime(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/hieple/Documents/openhands/OpenHands/openhands/server/session/agent_session.py", line 345, in _create_runtime
    provider_handler = ProviderHandler(
                       ^^^^^^^^^^^^^^^^
  File "/Users/hieple/Documents/openhands/OpenHands/openhands/integrations/provider.py", line 121, in __init__
    raise TypeError(
TypeError: provider_tokens must be a MappingProxyType, got dict
```

The root cause may be associated with the changes introduced [here](https://github.com/OpenHands/OpenHands/pull/11813/files#diff-2a846e5207a45fb7ac74801ee3cbef95b358d01d2ef7fdd64d8a85bf4ccd8e1d)￼.

**Acceptance Criteria:**
- Users must be able to create both V0 and V1 conversations.

**Solution:**

Fix: The validator ensures that whenever ConversationInitData is created, git_provider_tokens and custom_secrets are always MappingProxyType (or None), matching what ProviderHandler expects.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:b5bbe31-nikolaik   --name openhands-app-b5bbe31   docker.openhands.dev/openhands/openhands:b5bbe31
```

### DIFF
--- a/tests/unit/server/session/test_conversation_init_data.py
+++ b/tests/unit/server/session/test_conversation_init_data.py
@@ -1,0 +1,272 @@
+"""Unit tests for ConversationInitData - specifically testing the field validator.
+
+These tests verify that the immutable_validator correctly converts dict to MappingProxyType
+for git_provider_tokens and custom_secrets fields, ensuring type safety.
+"""
+
+from types import MappingProxyType
+
+import pytest
+from pydantic import SecretStr
+
+from openhands.integrations.provider import CustomSecret, ProviderToken, ProviderType
+from openhands.server.session.conversation_init_data import ConversationInitData
+from openhands.storage.data_models.settings import Settings
+
+
+@pytest.fixture
+def base_settings():
+    """Create a base Settings object with minimal required fields."""
+    return Settings(
+        language='en',
+        agent='CodeActAgent',
+        max_iterations=100,
+        llm_model='anthropic/claude-3-5-sonnet-20241022',
+        llm_api_key=SecretStr('test_api_key_12345'),
+        llm_base_url=None,
+    )
+
+
+class TestConversationInitDataValidator:
+    """Test suite for ConversationInitData field validator."""
+
+    def test_git_provider_tokens_dict_converted_to_mappingproxy(self, base_settings):
+        """Test that dict passed as git_provider_tokens is converted to MappingProxyType."""
+        # Create provider tokens as a regular dict
+        provider_tokens_dict = {
+            ProviderType.GITHUB: ProviderToken(
+                token=SecretStr('ghp_test_token_123'), user_id='test_user'
+            ),
+            ProviderType.GITLAB: ProviderToken(
+                token=SecretStr('glpat_test_token_456'), user_id='test_user_2'
+            ),
+        }
+
+        # Create ConversationInitData with dict
+        init_data = ConversationInitData(
+            **base_settings.__dict__,
+            git_provider_tokens=provider_tokens_dict,
+        )
+
+        # Verify it's now a MappingProxyType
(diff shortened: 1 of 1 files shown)


malhotra5:
@hieptl could you describe what the problem was and how this PR addresses it

also would be great to add some test coverage here

hieptl:
> @hieptl could you describe what the problem was and how this PR addresses it

Hi @malhotra5,

I have updated the PR description to include additional information.

> also would be great to add some test coverage here

I have added tests to cover the changes I made.

Thank you! 🙏

malhotra5:
Awesome thanks a bunch!

Also I took a look at the linked PR, it seems it was moving away from `MappingProxyType` 

```
PROVIDER_TOKEN_TYPE = Mapping[ProviderType, ProviderToken]
CUSTOM_SECRETS_TYPE = Mapping[str, CustomSecret]
```



What are your thoughts on removing the type requirement in the `__init__` method for `ProviderHandler` instead of creating a field validator?

hieptl:
> Awesome thanks a bunch!
> 
> Also I took a look at the linked PR, it seems it was moving away from `MappingProxyType`
> 
> ```
> PROVIDER_TOKEN_TYPE = Mapping[ProviderType, ProviderToken]
> CUSTOM_SECRETS_TYPE = Mapping[str, CustomSecret]
> ```
> 
> What are your thoughts on removing the type requirement in the `__init__` method for `ProviderHandler` instead of creating a field validator?

I would prefer not to do that for the following reasons:

- Weakens the type contract: if ProviderHandler.__init__ says it expects MappingProxyType but we change it to also accept dict, the type hint becomes misleading.
- Masks the underlying issue: it does not address why a dict is being passed in the first place. We're not fixing why a dict is being passed in the first place. We're just accepting it and converting it, so the root cause remains.
- Multiple fix points: if this pattern occurs elsewhere, each location would require a fix.
- The strict check in ProviderHandler appears intentional to guarantee immutability.

Adding a field validator introduces a small validation step while enforcing correctness at the model boundary. This aligns with the existing Secrets pattern and preserves type safety. We can refer to this link for more information.

https://github.com/OpenHands/OpenHands/blob/3219834e35282a626a866c67eb0fc6a63c7660a5/openhands/storage/data_models/secrets.py#L43

I am open to making ProviderHandler more lenient if there is a strong justification, but I believe the validator approach represents the better long-term solution.

Thank you! 🙏

malhotra5 (approved):
Thank you!

PR #11835, opened 2025-11-27
fix(backend): switch from own model to saas breaks functionality (staging issue)

hieptl:
## Summary of PR

**Reproduction step:**

As illustrated in the image below, we utilize the OpenHands provider.
 
image

When initiating a new V1 conversation and prompt, I encounter the following error.

image

The pull request ensures that the current environment is correctly passed when setting up the LLM in the software-agent-sdk. If the environment is set to staging, the SDK will utilize the LLM proxy staging server; otherwise, it will connect to the production server.

For further details, please refer to this [pull request](https://github.com/OpenHands/software-agent-sdk/pull/1273)￼.

Additionally, the video below demonstrates the output of the pull request.

https://github.com/user-attachments/assets/b4241859-050e-4229-b712-acce926e4955

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:427005d-nikolaik   --name openhands-app-427005d   docker.openhands.dev/openhands/openhands:427005d
```

### DIFF
--- a/openhands/utils/utils.py
+++ b/openhands/utils/utils.py
@@ -97,10 +97,10 @@ def is_local_env(web_url: str | None) -> bool:
         web_url: The web URL to check
 
     Returns:
-        True if the environment is local (localhost:3030), False otherwise
+        True if the environment is local (localhost), False otherwise
     """
     hostname = _extract_hostname_from_web_url(web_url)
-    return hostname == 'host.docker.internal:3030'
+    return hostname == 'localhost'
 
 
 def is_staging_env(web_url: str | None) -> bool:


hieptl:
Hello @tofarr, I've updated the code based on your feedback. Thank you for your feedback! 🙏

enyst (changes requested):
Could we do this differently, without hardcoding "staging" and private AH URLs into the codebase?

It seems to me that this hardcoding breaks at least two expectations:
- this is the Open Source codebase, that everyone is using, while the private AH names and URLs are a private detail of a particular setup that only a dozen-some people access and use. The Saas/internal variables belong in Saas/internal scripts or codebases, really, not in the Open Source project.
- hardcoding "staging" into the code seems to break the expectations of a proper staging environment from AH perspective as well. Staging environments help catching bugs before they make it into prod _because_ the execution paths on staging are as close as possible with prod. Generally speaking, executing special "staging" code makes "staging" moot. You won't see if this is broken in prod. 😅 

Perhaps you could explain why you need any conditionals? Everyone can set "base url" in the UI or in env, why doesn't that fit the purpose here?

hieptl:
Hello @enyst,

In the SaaS environment, users will not be permitted to configure the base_url when selecting OpenHands as their provider. Additional context is available in the video referenced below.

Within our software-agent-sdk, the LiteLLM proxy URL is currently hard-coded to the production endpoint. As a result, when OpenHands is selected as the provider and a new V1 conversation is created, it becomes impossible to test the feature against a feature branch or staging environment. This occurs because the generated LiteLLM key, when passed to the production LiteLLM proxy URL, triggers the error described in the pull request description.

I can add an additional condition to ensure this functionality applies exclusively to the SaaS environment.

```python
self.app_mode == 'saas'
```

Please let me know if you have an alternative solution or suggestion. Thank you! 🙏 

https://github.com/user-attachments/assets/1986db60-aebb-4c0b-97c2-578b848595f3

enyst:
Just a quick question first. 

> In the SaaS environment, users will not be permitted to configure the base_url when selecting OpenHands as their provider.

Why not? Base URL is available to configure for ALL providers, including providers that no one knows (custom local setups like `vllm_hosted` etc) and those everyone knows, like `/anthropic`, and proxies to any of them. Why wouldn't it be possible for `/openhands`?

You may notice that, lot of the time, people don't need to enter a base URL, they just go with provider alone, but the reason why that works is that LiteLLM knows a large number of providers, and it routes to the correct `base_url` itself.

For example, if you write `openrouter/provider/model`, that request needs to go to `openrouter.ai`. That works because litellm interprets "openrouter" to go to "openrouter.ai". In `/openhands` case, we are currently doing it in the SDK, as you know, maybe in the future we should submit it to litellm so we don't have to do it ourselves.

Back to your experience, in your video, can you select "Advanced", and enter a custom `base URL` there?

hieptl:
I am referring to the ‘Basic’ view, not the ‘Advanced’ view. The issue we are addressing pertains exclusively to the ‘Basic’ view.

**Issue Environment:**
- Feature branch or Staging.

**Issue Description:**

1. In the SaaS environment (feature branch and staging), navigate to the LLM Basic Settings view. When selecting OpenHands as the provider, it is no longer possible to input the base URL via the UI in this basic view if the current provider is already set to OpenHands. (requested by @xingyaoww).

> In the SaaS environment, within the LLM Basic Settings view, if the current provider is set to OpenHands, users will not be able to update the base_url; instead, their requests will be automatically redirected to our LiteLLM proxy server.

3. After clicking the Save Settings button in the LLM Basic Settings view, creating a new V1 conversation, and sending a prompt, the following error occurs:

image

The root cause is that, within our Software Agent SDK, when the current provider is set to OpenHands, the base_url is hard-coded to https://llm-proxy.app.all-hands.dev/. Please refer to the code snippet below for further details.

```python
if model_val.startswith("openhands/"):
    model_name = model_val.removeprefix("openhands/")
    d["model"] = f"litellm_proxy/{model_name}"
    # Set base_url (default to the app proxy when base_url is unset)
    d["base_url"] = d.get("base_url", "https://llm-proxy.app.all-hands.dev/")
```

For this reason, when testing on the feature branch or staging environment, the feature/staging LiteLLM key is sent to the production URL, which results in the error described above.

This pull request introduces a check: if the current environment is staging or a feature branch and the provider is OpenHands, the base_url will be set to https://llm-proxy.staging.all-hands.dev/.

Please let me know if you have any alternative solutions or suggestions.

Thank you! 🙏

enyst:
> Issue Description:
> In the SaaS environment (feature branch and staging), navigate to the LLM Basic Settings view. When selecting OpenHands as the provider, it is no longer possible to input the base URL via the UI in this basic view if the current provider is already set to OpenHands. (requested by @xingyaoww).

Xingyao, is this really necessary? I think that's why Advanced view exists: for cases where people need more customization. If you need a custom URL, is there a reason why not use the existing Advanced view to enter it?

Same goes for any provider afaik. If the provider prefix is, let's say, `anthropic/`, but people want to use a proxy, they need Advanced view, or am I misremembering?

Alternatively, if we want to input base URL in Basic view, we could add a field for base URL in Basic view?

enyst:
Just a tiny side question here,

> Issue Environment:
> ...
> Issue Description
> ...

@hieptl Could you please tell, what is the source of the quoted "issue" above? Is this issue on GitHub?

hieptl:
> Just a tiny side question here,
> 
> > Issue Environment:
> > ...
> > Issue Description
> > ...
> 
> @hieptl Could you please tell, what is the source of the quoted "issue" above? Is this issue on GitHub?

@enyst There is no GitHub issue for this, as we are conducting internal testing in preparation for the V1 release. During testing, Mahmoud used the staging environment, navigated to the LLM Basic Settings view, and followed the steps outlined in my previous comment. He encountered the issues described above.

The purpose of this pull request is to resolve this issue for the feature branch and staging environments, as the LiteLLM proxy production URL is currently hard-coded in the software-agent-sdk code 🙏

enyst:
To clarify,
> The purpose of this pull request is to resolve this issue for the feature branch and staging environments, as the LiteLLM proxy production URL is currently hard-coded in the software-agent-sdk code 🙏

This need is a need that many people have: to configure an LLM. And we can do it: OpenHands is configurable, it allows to enter custom base URLs and custom provider prefixes, both in env I hope, as LLM_BASE_URL, and in the Web UI. Everyone else does that. I still don't see a reason why it wouldn't work?

enyst:
So it seems to me I can see at least 3 solutions, that don't require to hardcode private information and Saas execution paths into the OpenHands codebase.

**1. Use custom URL and prefix**

The solution is like this:
- press Advanced view
- use litellm_proxy/your_desired_model as model name
- use https://llm-proxy.staging.all-hands.dev/ as base URL.

**2. Could only enter base URL, with `/openhands` prefix:**

Because we already solved the overriding of base url in the SDK (just needs released)
- https://github.com/OpenHands/software-agent-sdk/pull/1273

So this solution looks like:
- press Advanced view
- use openhands/your_desired_model_name
- use https://llm-proxy.staging.all-hands.dev as base URL.

**3. Use deployment scripts** 
- override LLM_BASE_URL for your needs, in staging script(s) / code, and/or (likely unnecessary) redesign the code and extend it for the Saas only.

> @enyst There is no GitHub issue for this, as we are conducting internal testing in preparation for the V1 release. During testing, Mahmoud used the staging environment, navigated to the LLM Basic Settings view, and followed the steps outlined in my previous comment. He encountered the issues described above.

Many people and organizations use OpenHands with many LLMs. Some are custom URLs or custom prefixes. The internal AH testing doesn't seem different to me, so why wouldn't you use the same solutions they use?

---

Please do consider this request: when you need changes in OpenHands because of the Saas, please try to make an issue for it or notify us in Slack.

IMHO it's important that we talk these things over. So we can see if there's some confusion, as it seems to be here (because OH already supports this need), or we can see what is the best way to design a solution so that OpenHands supports a use case, without hardcoding special conditionals into the Open Source project.

Adding directly Saas code into OpenHands should ideally NEVER happen. I would be surprised if we really really can't think of alternative ways as far as all these things go. So it's unnecessary, in addition to error-prone, breaking standard SWE patterns, and other disadvantages.

We have dev environments on GitHub and Slack, to help us figure out _what_ and _how_ to do things, could we please use them?

enyst:
Actually, as noted above, we might have also a 4th solution:

**4. Add base URL to Basic view.**

I honestly don't remember if it's not there already. If it's not, and you folks think it's important to be available, we could add it.

hieptl:
Hi @xingyaoww,

Please review @enyst’s feedback and advise on how we should proceed in this situation. I removed the base URL input field from the Basic LLM Settings page as requested; however, Mahmoud was unable to verify the change in the staging environment.

You may refer to the following comment for the production steps:

https://github.com/OpenHands/OpenHands/pull/11835#issuecomment-3595870145￼

As a result, I created this pull request to address the issue.
However, it appears that my changes do not align with our best practices, so I will close the PR for now.

PR #11837, opened 2025-11-28
fix(backend): conversation statistics are currently not being persisted to the database (V1).

hieptl:
## Summary of PR

**Acceptance Criteria:**
- When a ConversationStateUpdateEvent is received with the key set to stats, the corresponding statistics data must be stored in the database.

https://github.com/user-attachments/assets/717fa13c-14bc-402c-8cba-752af9614e53

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:84ed3a4-nikolaik   --name openhands-app-84ed3a4   docker.openhands.dev/openhands/openhands:84ed3a4
```

### DIFF
--- a/openhands/app_server/app_conversation/app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/app_conversation_info_service.py
@@ -9,7 +9,6 @@ from openhands.app_server.app_conversation.app_conversation_models import (
     AppConversationSortOrder,
 )
 from openhands.app_server.services.injector import Injector
-from openhands.sdk.conversation.conversation_stats import ConversationStats
 from openhands.sdk.event import ConversationStateUpdateEvent
 from openhands.sdk.utils.models import DiscriminatedUnionMixin
 
@@ -92,17 +91,6 @@ class AppConversationInfoService(ABC):
         """Store the sandboxed conversation info object given.
 
         Return the stored info
-        """
-
-    @abstractmethod
-    async def update_conversation_statistics(
-        self, conversation_id: UUID, stats: ConversationStats
-    ) -> None:
-        """Update conversation statistics from stats event data.
-
-        Args:
-            conversation_id: The ID of the conversation to update
-            stats: ConversationStats object containing usage_to_metrics data from stats event
         """
 
     @abstractmethod


openhands/app_server/app_conversation/app_conversation_info_service.py
```diff
@@ -92,6 +92,17 @@ async def save_app_conversation_info(
         Return the stored info
         """
 
+    @abstractmethod
+    async def update_conversation_statistics(
+        self, conversation_id: UUID, stats: dict
```
tofarr: I think we should create a model object for the stats here - including fields for it.

tofarr (changes requested):
Lets use models instead of dictionaries where possible

openhands/app_server/app_conversation/app_conversation_info_service.py
hieptl: Hi @tofarr, I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/app_server/event_callback/webhook_router.py
```diff
@@ -124,6 +126,50 @@ async def on_conversation_update(
     return Success()
 
 
+async def _process_stats_event(
+    event: ConversationStateUpdateEvent,
+    conversation_id: UUID,
+    app_conversation_info_service: AppConversationInfoService,
```
enyst: Tiny nit: I wonder if this could actually be a method in the `service`, rather than in the `router`. Really small thing, but it would keep the router code a bit more focused on the endpoints, while the service contains the logic 🤔

hieptl: Hi @enyst, I've updated the code based on your feedback. Thank you for your feedback! 🙏

openhands/app_server/app_conversation/app_conversation_info_service.py
```diff
@@ -92,6 +94,30 @@ async def save_app_conversation_info(
...
+    ) -> None:
+        """Update conversation statistics from stats event data.
+
+        Args:
+            conversation_id: The ID of the conversation to update
+            stats: ConversationStats object containing usage_to_metrics data from stats event
+        """
+
+    @abstractmethod
+    async def process_stats_event(
```
enyst, line 97: This is okay... but it probably doesn't need to be in the interface, nor even public?

hieptl: Hello @enyst,

This function is utilized in webhook_router.py. Therefore, it should be included in the interface. However, I have removed update_conversation_statistics since it is not publicly used.

Thank you! 🙏

enyst: Ah right, thanks!

PR #11846, opened 2025-12-01
fix(frontend): the content of the FinishObservation event is not being rendered correctly.

hieptl:
## Summary of PR

image

Based on the image above, the FinishObservation event content fails to display as expected.

**Acceptance Criteria:**
- The FinishObservation event content must be rendered accurately.

We can refer to the image below for the output of the PR.

app-182

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:d80a4ab-nikolaik   --name openhands-app-d80a4ab   docker.openhands.dev/openhands/openhands:d80a4ab
```

### DIFF
--- a/frontend/src/components/v1/chat/event-message-components/generic-event-message-wrapper.tsx
+++ b/frontend/src/components/v1/chat/event-message-components/generic-event-message-wrapper.tsx
@@ -1,6 +1,3 @@
-import Markdown from "react-markdown";
-import remarkGfm from "remark-gfm";
-import remarkBreaks from "remark-breaks";
 import { OpenHandsEvent } from "#/types/v1/core";
 import { GenericEventMessage } from "../../../features/chat/generic-event-message";
 import { getEventContent } from "../event-content-helpers/get-event-content";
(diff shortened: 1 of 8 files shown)


hieptl:
> I think in V0, we render finish actions without an accordion / expandable, just like a regular message. Do you think we can do that here too?

Hi @amanape, I've updated the code based on your feedback. We can refer to the image below for the updated output. Thank you for your feedback! 🙏 

app-182-2

frontend/src/components/v1/chat/event-message-components/generic-event-message-wrapper.tsx
```diff
@@ -23,11 +38,32 @@ export function GenericEventMessageWrapper({
...
+              h2,
+              h3,
+              h4,
+              h5,
+              h6,
+            }}
+            remarkPlugins={[remarkGfm, remarkBreaks]}
+          >
+            {details as string}
+          </Markdown>
```
amanape: If this is copy-pasted code from elsewhere, we should probably move it to it’s own component and reuse it in both places

hieptl: Hi @amanape, I've updated the code based on your feedback. Thank you for your feedback! 🙏

PR #11862, opened 2025-12-02
fix(backend): apply user-defined condenser_max_size in new v1 conversations

### LINKED ISSUE #11861: [Bug]: Apply User-Defined condenser_max_size in New V1 Conversations

Image

When initiating a new V1 conversation, the default condenser is currently applied. However, if users modify their settings and adjust the condenser_max_size value, the updated value is not currently applied.

The system should utilize the condenser_max_size value whenever it has been specified.

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Model Name

_No response_

### Operating System

None

_No response_

hieptl:
## Summary of PR

Image

When initiating a new V1 conversation, the default condenser is currently applied. However, if users modify their settings and adjust the condenser_max_size value, the updated value is not currently applied.

The system should utilize the condenser_max_size value whenever it has been specified.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #11861 

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:9d00f26-nikolaik   --name openhands-app-9d00f26   docker.openhands.dev/openhands/openhands:9d00f26
```

### DIFF
--- a/tests/unit/experiments/test_experiment_manager.py
+++ b/tests/unit/experiments/test_experiment_manager.py
@@ -152,6 +152,7 @@ class TestExperimentManagerIntegration:
                     llm_base_url=None,
                     llm_api_key=None,
                     confirmation_mode=False,
+                    condenser_max_size=None,
                 )
 
             async def get_secrets(self):


tests/unit/app_server/test_live_status_app_conversation_service.py
```diff
@@ -421,23 +422,25 @@ async def test_configure_llm_and_mcp_tavily_with_whitespace_user_search_key(self
...
         mock_llm = Mock(spec=LLM)
         mock_llm.model_copy.return_value = mock_llm
         mock_get_tools.return_value = []
-        mock_get_condenser.return_value = Mock()
+        mock_condenser = Mock()
+        mock_create_condenser.return_value = mock_condenser
         mock_format_plan.return_value = 'test_plan_structure'
         mcp_config = {'default': {'url': 'test'}}
         system_message_suffix = 'Test suffix'
+        self.mock_user.condenser_max_size = None
```
enyst: Could we have a test that actually tests for the `max_size` that this PR proposes to add?

hieptl: Hello @enyst,

I've updated the code based on your feedback.

openhands/app_server/app_conversation/app_conversation_service_base.py
```diff
@@ -300,3 +304,52 @@ async def maybe_setup_git_hooks(
...
+
+        Returns:
+            Configured LLMSummarizingCondenser instance
+        """
+        # Get condenser_max_size from user settings, with type-specific defaults
+        condenser_max_size = user.condenser_max_size
+        if agent_type == AgentType.PLAN:
+            # Planning agents use larger default context window
+            default_max_size = 100
+            default_keep_first = 6
```
enyst: IMHO this isn't a significant difference (100 compared to 80, 6 to 4), and maybe it shouldn't be any difference.

I was looking at the tests and it seems that creating condenser needs multiple mocks from Agent, which... doesn't really make a lot sense to me.

Maybe the method creating a `condenser` could be independent of Agent. 🤔 
`default_max_size` could be 120 for both. It has been in the past, and experiments with less haven't shown any positive difference AFAIK. So let's make it 120 / 4 for all, and remove the dependency on Agent types/methods?

openhands/app_server/app_conversation/app_conversation_service_base.py
```diff
@@ -300,3 +304,47 @@ async def maybe_setup_git_hooks(
...
+            user: User information containing settings like condenser_max_size
+
+        Returns:
+            Configured LLMSummarizingCondenser instance
+        """
+        # Get condenser_max_size from user settings, with type-specific defaults
+        condenser_max_size = user.condenser_max_size
+
+        default_max_size = 120
+        default_keep_first = 4
```
enyst: Actually, @hieptl , we have the defaults in `LLMSummarizingCondenser` itself: 

https://github.com/OpenHands/software-agent-sdk/blob/a106684194dc5e2a5869e4f06888e8d66489d4ae/openhands-sdk/openhands/sdk/context/condenser/llm_summarizing_condenser.py#L16

Maybe we don't need to hardcode them again?

hieptl: Hello @enyst,

I've updated the code based on your feedback.

openhands/app_server/app_conversation/app_conversation_service_base.py
hieptl: Hello @enyst,

I hope I have correctly understood and addressed all your comments.

Thank you.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -651,20 +650,22 @@ def _create_agent_with_context(
             agent_type: Type of agent to create (PLAN or DEFAULT)
             system_message_suffix: Optional suffix for system messages
             mcp_config: MCP configuration dictionary
+            user: User information containing settings like condenser_max_size
```
enyst:
```suggestion
            settings: Settings like condenser_max_size
```
Totally a nit, but we were really going in the opposite direction here with some PRs (like the org PR), where we consider "user stuff" to be just "stuff". It seems to me now that it's unfortunate that we have a model named `UserInfo` that doesn't really mean user info, it really means `Settings + an optional user id`. IMHO that sounds like... Settings, not user.

I see you were using heavily that model class in the latest PRs... and that's understandable, I'm just thinking out loud that we probably could rethink that model class or its place in OH.

hieptl: Hello @enyst,

I've updated the code based on your feedback.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -651,20 +650,22 @@ def _create_agent_with_context(
             agent_type: Type of agent to create (PLAN or DEFAULT)
             system_message_suffix: Optional suffix for system messages
             mcp_config: MCP configuration dictionary
+            condenser_max_size: condenser_max_size setting
 
         Returns:
             Configured Agent instance with context
         """
+        # Create condenser with user's settings
+        condenser = self._create_condenser(llm, agent_type, condenser_max_size)
```
enyst, line 659: I think this signature is good enough for now, but for the future I feel that:
- we could remove `agent_type` (send directly what `usage_id` to use)
- we should, IMHO, sooner or later move towards [breaking up Settings](https://github.com/OpenHands/OpenHands/issues/10947#issuecomment-3282347974) into
    - `CondenserSettings`
    - `AgentSettings` etc

Then the app-server should send `condenser_settings` here. (We know the app-server UI itself only offers the user `max_size`, but idk, other extensions could be created to add more fields, and IMHO the utility method `_create_condenser` could maybe set those too if they exist.)

hieptl: Thank your for sharing with me! 🙏

PR #11877, opened 2025-12-02
refactor/bugfix: simplify hasOpenHandsSuffix with provider lookup for gitlab/azure

vinisdl:
## Summary of PR

This PR refactors the `hasOpenHandsSuffix` method to simplify the provider suffix lookup logic. Instead of using conditional statements, the method now uses a lookup object (`providerRepositorySuffix`) that maps specific providers to their suffixes.

**Key changes:**
- GitLab and Azure DevOps providers use the specific suffix `/openhands-config`
- All other providers (GitHub, Bitbucket, Enterprise SSO) fall back to the default suffix `/.openhands`
- Removed unreachable code after the return statement

The refactoring improves code maintainability and makes it easier to add or modify provider-specific suffixes in the future.

## Change Type

- [x] Refactor
- [x] bugfix

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

### DIFF
--- a/frontend/src/utils/utils.ts
+++ b/frontend/src/utils/utils.ts
@@ -606,16 +606,14 @@ export const shouldIncludeRepository = (
  * @returns The query string for searching OpenHands repositories
  */
 export const getOpenHandsQuery = (provider: Provider | null): string => {
-  const providerRepositorySuffix = {
+  const providerRepositorySuffix: Record<string, string> = {
     gitlab: "openhands-config",
     azure_devops: "openhands-config",
     default: ".openhands",
   } as const;
 
   return provider && provider in providerRepositorySuffix
-    ? providerRepositorySuffix[
-        provider as keyof typeof providerRepositorySuffix
-      ]
+    ? providerRepositorySuffix[provider]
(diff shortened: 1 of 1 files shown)


frontend/src/utils/utils.ts
```diff
@@ -606,10 +606,17 @@ export const shouldIncludeRepository = (
...
+    gitlab: "openhands-config",
+    azure_devops: "openhands-config",
+    default: ".openhands",
+  } as const;
+
+  return provider && provider in providerRepositorySuffix
+    ? providerRepositorySuffix[
+        provider as keyof typeof providerRepositorySuffix
+      ]
+    : providerRepositorySuffix.default;
```
amanape:
```suggestion
  const providerRepositorySuffix: Record<string, string> = {
    gitlab: "openhands-config",
    azure_devops: "openhands-config",
    default: ".openhands",
  } as const;

  return provider && provider in providerRepositorySuffix
    ? providerRepositorySuffix[provider]
    : providerRepositorySuffix.default;
```

`provider in providerRepositorySuffix` already confirms it's a valid key

Issue #11853, opened 2025-12-01
Display command in observation block in UI


xingyaoww:
## Description

Currently, when observations are displayed in the UI as collapsible blocks, the command that generated the observation is not shown. This makes it difficult for users to understand what action led to a particular observation.

<img width="1588" height="1196" alt="Image" src="https://github.com/user-attachments/assets/ac22a7be-b48e-4279-9bee-e8962edd6a02" />

## Proposed Solution

Display the command that was executed alongside the observation in the observation block in the UI.

## Example

When a command like `ls -la` is executed and generates an observation, the observation block should show:
- The command that was run (`ls -la`)
- The output/result of that command

## Context

This enhancement would improve the user experience by providing better context about what actions the agent is taking.

## Additional Notes

Mentioned by users in discussion for better UI transparency.

nehaaprasad:
Hello! @xingyaoww 
fixed it, please check

PR #11885, opened 2025-12-03
feat: display command in observation block

nehaaprasad:
## Summary of PR

fix: #11853

Displays the executed command alongside its output in observation blocks. Previously, only the output was shown when expanding an observation block. Now users can see both the command and output for better context on agent actions.

## Change Type

- [x] New feature

### DIFF
--- a/frontend/src/components/v1/chat/event-content-helpers/get-observation-content.ts
+++ b/frontend/src/components/v1/chat/event-content-helpers/get-observation-content.ts
@@ -71,12 +71,6 @@ const getTerminalObservationContent = (
     content = `${content.slice(0, MAX_CONTENT_LENGTH)}...`;
   }
 
-  const { command } = observation;
-
-  if (command) {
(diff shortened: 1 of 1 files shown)


frontend/src/components/v1/chat/event-content-helpers/get-observation-content.ts
```diff
@@ -71,6 +71,12 @@ const getTerminalObservationContent = (
     content = `${content.slice(0, MAX_CONTENT_LENGTH)}...`;
   }
 
+  const { command } = observation;
+
+  if (command) {
+    return `Command:\n\`\`\`sh\n${command}\n\`\`\`\n\nOutput:\n\`\`\`sh\n${content.trim() || i18n.t("OBSERVATION$COMMAND_NO_OUTPUT")}\n\`\`\``;
+  }
+
```
amanape:
```suggestion
```

PR #11893, opened 2025-12-03
Replace USE_V1_CONVERSATION_API feature flag with user setting

tofarr:
## Summary of PR

Replaces the `USE_V1_CONVERSATION_API` feature flag with a user setting (`v1_enabled`) to allow per-user control of V1 conversation API usage.

## Change Type

- [x] Refactor

## Checklist

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:ecb34bc-nikolaik   --name openhands-app-ecb34bc   docker.openhands.dev/openhands/openhands:ecb34bc
```

### DIFF
--- a/openhands/storage/data_models/settings.py
+++ b/openhands/storage/data_models/settings.py
@@ -1,4 +1,5 @@
 from __future__ import annotations
+
 import os
 
 from pydantic import (


frontend/src/hooks/mutation/use-save-settings.ts
```diff
@@ -35,6 +35,7 @@ const saveSettingsMutationFn = async (settings: Partial<PostSettings>) => {
       settings.GIT_USER_NAME?.trim() || DEFAULT_SETTINGS.GIT_USER_NAME,
     git_user_email:
       settings.GIT_USER_EMAIL?.trim() || DEFAULT_SETTINGS.GIT_USER_EMAIL,
+    v1_enabled: settings.V1_ENABLED,
```
hieptl, line 38: **question:** has the API been updated to support saving this field?

frontend/src/hooks/mutation/use-create-conversation.ts
```diff
@@ -50,7 +51,7 @@ export const useCreateConversation = () => {
         agentType,
       } = variables;
 
-      const useV1 = USE_V1_CONVERSATION_API() && !createMicroagent;
+      const useV1 = !!settings?.V1_ENABLED && !createMicroagent;
```
hieptl, line 54: **question:** If we merge this pull request, will working on V1 locally (OSS or SaaS) require manually modifying settings instead of using local storage?

Furthermore, when testing via the feature branch or staging environment, will updating the database be necessary? Should we retain both approaches—localStorage and database settings—to streamline the development process?

Lastly, for non-technical team members, would testing V1 require them to coordinate with engineers?

amanape: Would this also mean that if we want to revert users in SaaS to use V0 (in case something goes wrong), we’d have to update the settings objects for all users?

hieptl (approved):
Thank you! 🙏

frontend/src/hooks/mutation/use-save-settings.ts
tofarr: No - it will be a temporary measure - as soon as V1 is considered stable, we will enable it for everybody.

frontend/src/hooks/mutation/use-create-conversation.ts
tofarr: Once this is merged, My plan was to run the following in the staging db, so everybody using staging will use the V1 api:
update user_settings set v1_enabled=TRUE
I did not plan to add a UI for this or anything, as it will be a temporary measure - we run it for the rest of the week on staging, then after deploy we run the following on prod:
update user_settings set v1_enabled=TRUE where email like '%@all-hands.dev' or email like '%@openhands.dev'

enyst: I was just looking for the feature flag because contributors are fixing things with their environment, and spending their time making PRs, and I see it's just gone?

enyst: It seems it's in env, if that works it's usable

PR #11954, opened 2025-12-08
Fix/avatar menu hover corridor issue with timeout delay

### LINKED ISSUE #11933: Avatar context menu closes when moving cursor diagonally to menu (hover zone gap)
## Problem

## Expected Behaviour

- The avatar itself
- The menu itself

### Relevant files

itshae43:
Fixes #11933
## Changes Made
- implement a 400ms timeout delay on mouse leave.
- Added JavaScript-based hover detection using `onMouseEnter` and `onMouseLeave` handlers.
- Used `useRef` to manage the close timer and prevent memory leaks.
- Ensured timer cancellation on re-hover and cleanup on component unmount.
- Removed conflicting CSS `group-hover` logic to allow JavaScript control.

## Testing
- Tested in mock mode (`npm run dev:mock`) to verify the menu stays open during diagonal cursor movement.

https://github.com/user-attachments/assets/0674a07e-7ab4-4f1f-94c8-64ffb669d92e

image

### DIFF
--- a/frontend/src/components/features/sidebar/user-actions.tsx
+++ b/frontend/src/components/features/sidebar/user-actions.tsx
@@ -29,7 +29,7 @@ export function UserActions({ onLogout, user, isLoading }: UserActionsProps) {
   const closeAccountMenu = () => {
     closeTimeoutRef.current = setTimeout(() => {
       setAccountContextMenuIsVisible(false);
-    }, 350);
+    }, 400);
   };
 
   const cancelClose = () => {


itshae43:
@amanape Please review this. 
First, I tried to implement the Amazon mega-menu method, but it would change too much code and isn’t really needed for this menu. So I used the time-delay approach instead. 
 
 inspired from firebase Navbar -https://firebase.google.com/

amanape:
Hey thanks for opening a PR! I have a few concerns with this approach:
- 400ms delay can feel sluggish when intentionally moving away from the menu
- Adds JS state management and refs where pure CSS could work
- Removes the group hover pattern in favor of explicit state

My preference is to use a pure CSS alternative (see #11994). This avoids the delay feel and keeps the solution in CSS, but both approaches are valid.

I encourage you to take a look at the alternative PR. Thanks again!

PR #11962, opened 2025-12-08
Remove loud red x icon on agent commands

ak684:
## Summary of PR

This might be controversial PR (hopefully not). I've been on call with customers and every once and awhile they complain that our agent is 'not as good' as other agents and they point to this red x as an example of the agent making mistakes. In reality in many of these cases, our agent figures out the task, and this red x represents a command the agent ran that perhaps didn't work but wasnt necessarily a mistake, it was just the agent iterating on the task. As such, I think this red x icon for agent "errors" is simply too loud. Other ai agent offerings do not show agent commands that error with red x icons and I don't think we should either.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:6f61277-nikolaik   --name openhands-app-6f61277   docker.openhands.dev/openhands/openhands:6f61277
```

### DIFF
--- a/frontend/src/components/features/chat/success-indicator.tsx
+++ b/frontend/src/components/features/chat/success-indicator.tsx
@@ -1,6 +1,5 @@
 import { FaClock } from "react-icons/fa";
 import CheckCircle from "#/icons/check-circle-solid.svg?react";
-import XCircle from "#/icons/x-circle-solid.svg?react";
 import { ObservationResultStatus } from "./event-content-helpers/get-observation-result";
 
 interface SuccessIndicatorProps {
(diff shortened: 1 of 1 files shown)


amanape:
That is a good question, merged too early. I can try and take a look at it myself soon

ak684:
There are 2 different x icons in the UI, one had a test, but the one in this PR did not https://github.com/OpenHands/OpenHands/pull/11964

PR #11982, opened 2025-12-10
fix: Prevent old instructions from being re-executed after conversation condensation

### LINKED ISSUE #11910: [Bug]: Old instructions re‑executed after conversation condensation (dangerous behavior)

### Summary                                                                                                                                                                                                                                                                                                                                                                 
In a long‑running OpenHands conversation, after the conversation had been condensed/summarized a few times, the agent **re‑executed an old instruction that had been fully completed 
days earlier**, without me issuing that instruction again.                                                                                                                            
                                                                                                                                                                                      
From the user’s perspective, this looks like the agent suddenly “deciding” to re‑run a prior task on its own. This is dangerous because a condensed conversation can cause            
**previous, already‑satisfied instructions to be treated as fresh commands**.                                                                                                         
                                                                                                                                                                                      
---                                                                                                                                                                                   
                                                                                                                                                                                      
### Environment                                                                                                                                                                       
                                                                                                                                                                                      
- OpenHands install: from GitHub                                                                                                                                                      
  - Repo: `https://github.com/All-Hands-AI/OpenHands.git`                                                                                                                             
  - Commit: `cd3e845872d1aa64f0c2ff5902ff24d7f90d1efd` (`git rev-parse HEAD`)                                                                                                         
- Host OS: Linux (paths like `/home/tim/OpenHands`, `/home/tim/DM_Reference_Material`)                                                                                                
- OpenHands is being used with:                                                                                                                                                       
  - A local filesystem workspace containing D&D reference material at `/home/tim/DM_Reference_Material/`                                                                              
  - Long, multi‑day conversations                                                                                                                                                     
  - Conversation condensation enabled (the UI has explicitly said the thread was “condensed” a few times)                                                                             
                                                                                                                                                                                      
I don’t know the exact active config file on the server side, but the repo’s `config.template.toml` shows:                                                                            
                                                                                                                                                                                      
- `agent.enable_history_truncation = true`                                                                                                                                            
- A `[condenser]` section with `type = "noop"` by default, but in practice my conversation was clearly being condensed/summarized (the UI informed me several times that the thread   
was condensed), suggesting a non‑noop condenser (e.g., `llm`, `amortized`, etc.) or default condenser behavior was in effect.                                                         
                                                                                                                                                                                      
---                                                                                                                                                                                   
                                                                                                                                                                                      
### Scenario / What I Was Doing                                                                                                                                                       
                                                                                                                                                                                      
I have a long‑running project:                                                                                                                                                        
                                                                                                                                                                                      
- Building a D&D DM rules index from a local corpus at `/home/tim/DM_Reference_Material/`                                                                                             
- Files include `dm_main_index.json`, `dm_index_summary.txt`, `D&D Topic List.txt`, PDFs, DOCX files, etc.                                                                            
- I’m also using a Jetson Orin Nano as a local LLM backend, with Ollama models (`qwen2.5:3b`, `llama3.2:3b`), but that’s mostly orthogonal to this bug.                               
                                                                                                                                                                                      
At some point **days ago** in this same conversation, I gave the agent this instruction (paraphrased, but very close to exact text):                                                  
                                                                                                                                                                                      
> See the index here: `/home/tim/DM_Reference_Material/`. Using that index, find and tell me how to do a perception skill check. Don't use any other sources besides what is in that index.                                                                                                                                                                                
                                                                                                                                                                                      
The agent **correctly complied at that time**, by:                                                                                                                                    
                                                                                                                                                                                      
- Reading `/home/tim/DM_Reference_Material/dm_main_index.json`                                                                                                                        
- Using `grep`/`sed` to find entries about:                                                                                                                                           
  - “Wisdom (Perception) check”                                                                                                                                                       
  - “Passive Perception = 10 + Wisdom (Perception) check modifier”                                                                                                                    
  - D20 Tests and ability checks from the DMG and SRD excerpts                                                                                                                        
- Explaining how to run a Perception skill check based solely on those index snippets.                                                                                                
                                                                                                                                                                                      
That was all fine and expected.                                                                                                                                                       
                                                                                                                                                                                      
After that, the conversation continued for days with **other topics**:                                                                                                                
                                                                                                                                                                                      
- Installing/fixing Ollama models on the Jetson                                                                                                                                       
- Permissions for `/mnt/8tb_nvme/ollama_models`                                                                                                                                       
- Discussing the new D&D index pipeline, task tracking, etc.                                                                                                                          
                                                                                                                                                                                      
The thread was **condensed** / summarized a few times in the UI (I saw explicit condensation messages) as it became long.                                                             
                                                                                                                                                                                      
---                                                                                                                                                                                   
                                                                                                                                                                                      
### The Bug: Old Instruction Re‑Executed Without Being Given Again                                                                                                                    
                                                                                                                                                                                      
Much later—**without me re‑typing that Perception/index instruction**—the agent suddenly:                                                                                             
                                                                                                                                                                                      
1. Started running commands like:                                                                                                                                                     
   - `grep -ni "ability check" /home/tim/DM_Reference_Material/dm_main_index.json`                                                                                                    
   - `sed -n '10496,10515p' /home/tim/DM_Reference_Material/dm_main_index.json`                                                                                                       
   - `grep -ni "Perception" /home/tim/DM_Reference_Material/dm_main_index.json`                                                                                                       
   - `sed -n '58344,58364p' /home/tim/DM_Reference_Material/dm_main_index.json`                                                                                                       
   - etc.                                                                                                                                                                             
2. Then produced a detailed explanation of **how to do a Perception skill check**, referencing:                                                                                       
   - SRD and PHB snippets like:                                                                                                                                                       
     - “make a DC 15 Wisdom (Perception) check…”                                                                                                                                      
     - “Passive Perception = 10 + Wisdom (Perception) check modifier”                                                                                                                 
   - DMG guidance on when to call for a D20 Test and how to set DCs.                                                                                                                  
                                                                                                                                                                                      
This behavior is **exactly what I had asked for days earlier** and which had already been completed. But at this later point:                                                         
                                                                                                                                                                                      
- I **had not just issued** that instruction.                                                                                                                                         
- We were in the middle of discussing something else (Jetson/Ollama, task tracking, etc.).                                                                                            
- The only reason the agent gave for doing this was that it saw the Perception instruction “at the top of the conversation” and treated it as fresh.                                  
                                                                                                                                                                                      
From my perspective as the user, it looked like the agent spontaneously decided to go back and re‑do an old task from days ago, without any corresponding new user request.           
                                                                                                                                                                                      
When I challenged it, the agent initially believed it had just responded to a user message that *it* could see in the current context:                                                
                                                                                                                                                                                      
> “See the index here: `/home/tim/DM_Reference_Material/`. Using that index, find and tell me how to do a perception skill check. Don't use any other sources besides what is in that 
index.”                                                                                                                                                                               
                                                                                                                                                                                      
However, I had **not** sent that message in the current turn‑sequence. That instruction was from **many days prior**, from an earlier portion of the same thread that had been        
condensed.                                                                                                                                                                            
                                                                                                                                                                                      
My conclusion is:                                                                                                                                                                     
                                                                                                                                                                                      
- The condensation / summarization process **re‑introduced** that old instruction into the current context as if it were a recent user turn, and                                      
- The agent, seeing it as a current user input, obediently executed it again.                                                                                                         
                                                                                                                                                                                      
---                                                                                                                                                                                   
                                                                                                                                                                                      
### Why This Is Dangerous                                                                                                                                                             
                                                                                                                                                                                      
This isn’t just cosmetic or a harmless repeat:                                                                                                                                        
                                                                                                                                                                                      
- In my case, it was a safe action (reading index files and explaining rules), but the same pattern could apply to:                                                                   
  - Re‑running a destructive maintenance step.                                                                                                                                        
  - Re‑issuing a deployment, data deletion, or refactor command.                                                                                                                      
  - Repeating a one‑time migration.                                                                                                                                                   
- The user’s mental model is: **“Once an instruction is completed, it won’t be re‑run unless I explicitly ask.”**                                                                     
- The current condensation behavior appears to violate that assumption by:                                                                                                            
  - Surfacing a previously executed instruction as an active user message again, with no indication that it is historical or already handled.                                         
                                                                                                                                                                                      
This is especially concerning for long‑running “project” threads where instructions may have powerful side effects.                                                                   
                                                                                                                                                                                      
---                                                                                                                                                                                   
                                                                                                                                                                                      
### Expected Behavior                                                                                                                                                                 
                                                                                                                                                                                      
- Conversation condensation / summarization **must not** cause old user instructions to be replayed as if they were new, actionable commands.                                         
- Summaries should preserve **intent** and **state** (e.g., “Task X was already completed”), but should not re‑insert completed directives in a way that the agent misinterprets them 
as fresh instructions.                                                                                                                                                                
                                                                                                                                                                                      
---                                                                                                                                                                                   
                                                                                                                                                                                      
### Actual Behavior                                                                                                                                                                   
                                                                                                                                                                                      
- After condensation, the system presented the agent with an old user instruction (the Perception/index query) as if it were current.                                                 
- The agent acted on it immediately, re‑running all the index searches and explanations.                                                                                              
- The user had not re‑issued this command in the recent turns.                                                                                                                        
                                                                                                                                                                                      
---                                                                                                                                                                                   
                                                                                                                                                                                      
### Hypothesis About Root Cause                                                                                                                                                       
                                                                                                                                                                                      
I obviously don’t see the internals, but based on the conversation:                                                                                                                   
                                                                                                                                                                                      
- The summarization / condensation layer appears to:                                                                                                                                  
  - Take earlier parts of the conversation and compress them.                                                                                                                         
  - Then reconstruct context for the agent by mixing:                                                                                                                                 
    - Some verbatim user turns.                                                                                                                                                       
    - Some summarized text.                                                                                                                                                           
- In the process, one of my **old user instructions** seems to have been:                                                                                                             
  - Re‑emitted as a direct, literal user message (“See the index… Using that index, find and tell me how to do a perception skill check…”),                                           
  - Without any marker that it is *historical* rather than a fresh input.                                                                                                             
                                                                                                                                                                                      
That is enough to cause the agent to treat it as a live directive.                                                                                                                    
                                                                                                                                                                                      
---                                                                                                                                                                                   
                                                                                                                                                                                      
### Suggested Constraints / Fix Direction                                                                                                                                             
                                                                                                                                                                                      
Some desirable properties for the condenser + planner:                                                                                                                                
                                                                                                                                                                                      
1. **No resurrection of directives as fresh commands**                                                                                                                                
   When compressing history, do not emit historical user instructions as if they were new top‑level user messages. Historical directives should only appear:                          
   - In summarized narrative form (e.g., “Earlier, the user asked for X and the agent completed it”), or                                                                              
   - In a clearly demarcated “past context” that the agent can recognize as non‑actionable.                                                                                           
                                                                                                                                                                                      
2. **Preserve completion state**                                                                                                                                                      
   The representation of history should encode that a given instruction was:                                                                                                          
   - Received at time T₀, and                                                                                                                                                         
   - Fulfilled at T₀+Δ,                                                                                                                                                               
   so the agent doesn’t “see” it as an unmet requirement.                                                                                                                             
                                                                                                                                                                                      
3. **Optional: constraints on how condensers re‑inject user turns**                                                                                                                   
   If a condenser uses LLM summarization, it might be safer to:                                                                                                                       
   - Never regenerate old user messages verbatim in the “current” events list.                                                                                                        
   - Instead, keep old turns either:                                                                                                                                                  
     - Fully dropped, or                                                                                                                                                              
     - Summarized as past context explicitly labeled as such.                                                                                                                         
                                                                                                                                                                                      
--- 

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Model Name

_No response_

### Operating System

None

_No response_

nathannewyen:
## Summary

This PR fixes issue #11910 where old user instructions were being re-executed after conversation condensation.

### The Bug

In long-running conversations, when the conversation history was condensed/summarized:
- The initial user message containing old instructions was always extracted from the **full** history (which is never modified by condensation)
- This message was then re-inserted into the condensed history
- The LLM saw both the condensation summary AND the original verbatim instruction
- This caused the LLM to sometimes re-execute already-completed tasks

### The Fix

The fix tracks which event IDs have been "forgotten" (condensed) and prevents the initial user message from being re-inserted if its ID is in the forgotten set.

**Key changes:**
- `View` class now tracks `forgotten_event_ids`
- `_ensure_initial_user_message()` checks if the initial user action has been condensed before re-inserting
- If condensed, the message is NOT re-inserted - the condensation summary already contains the context

### Files Changed

- `openhands/memory/view.py` - Added `forgotten_event_ids` field to View
- `openhands/agenthub/codeact_agent/codeact_agent.py` - Pass forgotten IDs through the chain
- `openhands/memory/conversation_memory.py` - Check if initial message is condensed before re-inserting
- `tests/unit/memory/test_conversation_memory.py` - Added tests for the fix + updated existing tests

## Test plan

- [x] Added unit test `test_ensure_initial_user_message_not_reinserted_when_condensed` - verifies condensed messages are not re-inserted
- [x] Added unit test `test_ensure_initial_user_message_reinserted_when_not_condensed` - ensures backward compatibility
- [x] Added integration test `test_process_events_does_not_reinsert_condensed_initial_message` - tests the full flow
- [x] Updated existing tests to pass `forgotten_event_ids` parameter
- [ ] Manual testing with long-running conversations and condensation enabled

Closes #11910

### DIFF
--- a/openhands/memory/conversation_memory.py
+++ b/openhands/memory/conversation_memory.py
@@ -76,6 +76,7 @@ class ConversationMemory:
         self,
         condensed_history: list[Event],
         initial_user_action: MessageAction,
+        forgotten_event_ids: set[int] | None = None,
         max_message_chars: int | None = None,
         vision_is_active: bool = False,
     ) -> list[Message]:
@@ -85,16 +86,23 @@ class ConversationMemory:
 
         Args:
             condensed_history: The condensed history of events to convert
+            initial_user_action: The initial user message action, if available. Used to ensure the conversation starts correctly.
+            forgotten_event_ids: Set of event IDs that have been forgotten/condensed. If the initial user action's ID
+                is in this set, it will not be re-inserted to prevent re-execution of old instructions.
             max_message_chars: The maximum number of characters in the content of an event included
                 in the prompt to the LLM. Larger observations are truncated.
             vision_is_active: Whether vision is active in the LLM. If True, image URLs will be included.
-            initial_user_action: The initial user message action, if available. Used to ensure the conversation starts correctly.
         """
         events = condensed_history
+        # Default to empty set if not provided
+        if forgotten_event_ids is None:
+            forgotten_event_ids = set()
 
         # Ensure the event list starts with SystemMessageAction, then MessageAction(source='user')
         self._ensure_system_message(events)
-        self._ensure_initial_user_message(events, initial_user_action)
+        self._ensure_initial_user_message(
+            events, initial_user_action, forgotten_event_ids
+        )
 
         # log visual browsing status
         logger.debug(f'Visual browsing: {self.agent_config.enable_som_visual_browsing}')
@@ -827,14 +835,39 @@ class ConversationMemory:
                 )
 
     def _ensure_initial_user_message(
-        self, events: list[Event], initial_user_action: MessageAction
+        self,
+        events: list[Event],
+        initial_user_action: MessageAction,
+        forgotten_event_ids: set[int],
     ) -> None:
-        """Checks if the second event is a user MessageAction and inserts the provided one if needed."""
+        """Checks if the second event is a user MessageAction and inserts the provided one if needed.
+
+        IMPORTANT: If the initial user action has been condensed (its ID is in forgotten_event_ids),
+        we do NOT re-insert it. This prevents old instructions from being re-executed after
+        conversation condensation. The condensation summary already contains the context of
+        what was requested and completed.
+
+        Args:
+            events: The list of events to modify in-place
+            initial_user_action: The initial user message action from the full history
+            forgotten_event_ids: Set of event IDs that have been forgotten/condensed
+        """
         if (
             not events
         ):  # Should have system message from previous step, but safety check
             logger.error('Cannot ensure initial user message: event list is empty.')
             # Or raise? Let's log for now, _ensure_system_message should handle this.
+            return
+
+        # Check if the initial user action has been condensed/forgotten.
+        # If so, we should NOT re-insert it to prevent re-execution of old instructions.
+        # The condensation summary already contains the context of what was requested.
+        initial_user_action_id = initial_user_action.id
+        if initial_user_action_id in forgotten_event_ids:
+            logger.info(
+                f'Initial user action (id={initial_user_action_id}) has been condensed. '
+                'Not re-inserting to prevent re-execution of old instructions.'
+            )
             return
 
         # We expect events[0] to be SystemMessageAction after _ensure_system_message
--- a/openhands/memory/view.py
+++ b/openhands/memory/view.py
@@ -18,6 +18,8 @@ class View(BaseModel):
 
     events: list[Event]
     unhandled_condensation_request: bool = False
+    # Set of event IDs that have been forgotten/condensed
+    forgotten_event_ids: set[int] = set()
 
     def __len__(self) -> int:
         return len(self.events)
@@ -90,4 +92,5 @@ class View(BaseModel):
         return View(
             events=kept_events,
             unhandled_condensation_request=unhandled_condensation_request,
+            forgotten_event_ids=forgotten_event_ids,
         )
--- a/openhands/agenthub/codeact_agent/codeact_agent.py
+++ b/openhands/agenthub/codeact_agent/codeact_agent.py
@@ -194,9 +194,12 @@ class CodeActAgent(Agent):
         # event we'll just return that instead of an action. The controller will
         # immediately ask the agent to step again with the new view.
         condensed_history: list[Event] = []
+        # Track which event IDs have been forgotten/condensed
+        forgotten_event_ids: set[int] = set()
         match self.condenser.condensed_history(state):
-            case View(events=events):
+            case View(events=events, forgotten_event_ids=forgotten_ids):
                 condensed_history = events
+                forgotten_event_ids = forgotten_ids
 
             case Condensation(action=condensation_action):
                 return condensation_action
@@ -206,7 +209,9 @@ class CodeActAgent(Agent):
         )
 
         initial_user_message = self._get_initial_user_message(state.history)
-        messages = self._get_messages(condensed_history, initial_user_message)
+        messages = self._get_messages(
+            condensed_history, initial_user_message, forgotten_event_ids
+        )
         params: dict = {
             'messages': messages,
         }
@@ -245,7 +250,10 @@ class CodeActAgent(Agent):
         return initial_user_message
 
     def _get_messages(
-        self, events: list[Event], initial_user_message: MessageAction
+        self,
+        events: list[Event],
+        initial_user_message: MessageAction,
+        forgotten_event_ids: set[int],
     ) -> list[Message]:
         """Constructs the message history for the LLM conversation.
 
@@ -284,6 +292,7 @@ class CodeActAgent(Agent):
         messages = self.conversation_memory.process_events(
             condensed_history=events,
             initial_user_action=initial_user_message,
+            forgotten_event_ids=forgotten_event_ids,
             max_message_chars=self.llm.config.max_message_chars,
             vision_is_active=self.llm.vision_is_active(),
         )
--- a/tests/unit/agenthub/test_agents.py
+++ b/tests/unit/agenthub/test_agents.py
@@ -393,7 +393,7 @@ def test_mismatched_tool_call_events_and_auto_add_system_message(
     # 2. The action message
     # 3. The observation message
     mock_state.history = [initial_user_message, action, observation]
-    messages = agent._get_messages(mock_state.history, initial_user_message)
+    messages = agent._get_messages(mock_state.history, initial_user_message, set())
     assert len(messages) == 4  # System + initial user + action + observation
     assert messages[0].role == 'system'  # First message should be the system message
     assert (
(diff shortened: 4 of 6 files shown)


openhands/memory/view.py
```diff
@@ -18,6 +18,8 @@ class View(BaseModel):
 
     events: list[Event]
     unhandled_condensation_request: bool = False
+    # Set of event IDs that have been forgotten/condensed
+    forgotten_event_ids: set[int] = set()
```
csmith49, line 22: Note for v1: the `View` class now maintains all `Condensation` events, which themselves have a `forgotten_event_ids` field. No updates necessary.

openhands/memory/conversation_memory.py
```diff
@@ -827,16 +835,41 @@ def _ensure_system_message(self, events: list[Event]) -> None:
...
+        """
         if (
             not events
         ):  # Should have system message from previous step, but safety check
             logger.error('Cannot ensure initial user message: event list is empty.')
             # Or raise? Let's log for now, _ensure_system_message should handle this.
             return
 
+        # Check if the initial user action has been condensed/forgotten.
+        # If so, we should NOT re-insert it to prevent re-execution of old instructions.
```
csmith49, line 863: Note for v1: since the `ConversationMemory` class no longer exists, there is no mechanism for inserting the initial user message. It exists in the event stream like any other message would. No updates necessary.

csmith49 (approved):
Nice fix!

This should not be a problem in v1 -- I've left some comments identifying why for future debugging purposes. If we see a similar issue arise in v1 we'll know to look elsewhere.

openhands/memory/conversation_memory.py
```diff
@@ -827,16 +835,41 @@ def _ensure_system_message(self, events: list[Event]) -> None:
...
-        self, events: list[Event], initial_user_action: MessageAction
+        self,
+        events: list[Event],
+        initial_user_action: MessageAction,
+        forgotten_event_ids: set[int],
     ) -> None:
-        """Checks if the second event is a user MessageAction and inserts the provided one if needed."""
+        """Checks if the second event is a user MessageAction and inserts the provided one if needed.
+
+        IMPORTANT: If the initial user action has been condensed (its ID is in forgotten_event_ids),
```
enyst, line 845: Ahh. Not a review comment, just a note, it shouldn’t have been condensed I think

PR #11984, opened 2025-12-10
Add OAuth 2.0 Device Flow backend for CLI authentication

malhotra5:
## Summary of PR

Implements OAuth 2.0 Device Flow backend for CLI authentication with OpenHands Cloud. Includes comprehensive unit tests (32 tests) and code cleanup.

## OAuth 2.0 Device Flow

**How it works:**
1. **CLI requests codes**: `POST /oauth/device/authorize` → returns `device_code` + `user_code`
2. **User authorizes**: Visits verification URL, enters `user_code`, logs in via Keycloak
3. **CLI polls for token**: `POST /oauth/device/token` with `device_code` → gets API key
4. **CLI authenticated**: Uses API key for OpenHands Cloud access

**Why both codes are needed:**
- `user_code` (ABC12345): Short, human-readable for browser entry
- `device_code` (128 chars): Cryptographically secure for CLI polling

## Key Features

🔐 **Endpoints**
- `POST /oauth/device/authorize` - Start flow
- `POST /oauth/device/token` - Poll for API key
- `GET /oauth/device/verify` - User authorization page
- `GET /oauth/device/keycloak-callback` - OAuth callback

## Change Type

- [x] New feature

## Checklist

## Release Notes

**Added OAuth 2.0 Device Flow for CLI authentication**

---

**Related:** OpenHands/OpenHands-CLI#192

**Testing:**
```bash
cd enterprise
PYTHONPATH=".:$PYTHONPATH" poetry run pytest tests/unit/storage/test_device_code*.py tests/unit/server/routes/test_oauth_device.py -v
# 32 passed in 1.48s

make lint
# All checks passed!
```

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:83dede6-nikolaik   --name openhands-app-83dede6   docker.openhands.dev/openhands/openhands:83dede6
```

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -161,7 +161,7 @@ poetry run pytest ./tests/unit/test_*.py
 To reduce build time (e.g., if no changes were made to the client-runtime component), you can use an existing Docker
 container image by setting the SANDBOX_RUNTIME_CONTAINER_IMAGE environment variable to the desired Docker image.
 
-Example: `export SANDBOX_RUNTIME_CONTAINER_IMAGE=ghcr.io/openhands/runtime:0.62-nikolaik`
+Example: `export SANDBOX_RUNTIME_CONTAINER_IMAGE=ghcr.io/openhands/runtime:1.0-nikolaik`
 
 ## Develop inside Docker container
 
--- a/containers/dev/compose.yml
+++ b/containers/dev/compose.yml
@@ -12,7 +12,7 @@ services:
       - SANDBOX_API_HOSTNAME=host.docker.internal
       - DOCKER_HOST_ADDR=host.docker.internal
       #
-      - SANDBOX_RUNTIME_CONTAINER_IMAGE=${SANDBOX_RUNTIME_CONTAINER_IMAGE:-ghcr.io/openhands/runtime:0.62-nikolaik}
+      - SANDBOX_RUNTIME_CONTAINER_IMAGE=${SANDBOX_RUNTIME_CONTAINER_IMAGE:-ghcr.io/openhands/runtime:1.0-nikolaik}
       - SANDBOX_USER_ID=${SANDBOX_USER_ID:-1234}
       - WORKSPACE_MOUNT_PATH=${WORKSPACE_BASE:-$PWD/workspace}
     ports:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ services:
     image: openhands:latest
     container_name: openhands-app-${DATE:-}
     environment:
-      - SANDBOX_RUNTIME_CONTAINER_IMAGE=${SANDBOX_RUNTIME_CONTAINER_IMAGE:-docker.openhands.dev/openhands/runtime:0.62-nikolaik}
+      - SANDBOX_RUNTIME_CONTAINER_IMAGE=${SANDBOX_RUNTIME_CONTAINER_IMAGE:-docker.openhands.dev/openhands/runtime:1.0-nikolaik}
       #- SANDBOX_USER_ID=${SANDBOX_USER_ID:-1234} # enable this only if you want a specific non-root sandbox user but you will have to manually adjust permissions of ~/.openhands for this user
       - WORKSPACE_MOUNT_PATH=${WORKSPACE_BASE:-$PWD/workspace}
     ports:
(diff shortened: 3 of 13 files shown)


enterprise/migrations/versions/084_create_device_codes_table.py
```diff
@@ -0,0 +1,49 @@
...
+        sa.Column('user_code', sa.String(length=16), nullable=False),
+        sa.Column('status', sa.String(length=32), nullable=False),
+        sa.Column('keycloak_user_id', sa.String(length=255), nullable=True),
+        sa.Column('user_id', sa.String(length=255), nullable=True),
+        sa.Column('access_token', sa.Text(), nullable=True),
+        sa.Column('created_at', sa.DateTime(), nullable=False),
+        sa.Column('expires_at', sa.DateTime(), nullable=False),
+        sa.Column('authorized_at', sa.DateTime(), nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+    )
```
malhotra5, line 34: @openhands examine this migration

1. edit this migration directly. remove `user_id` (this is the same as keycloak_user_id) and make sure we aren't using it elsewhere in this PR

2. we should not store the access token here, please remove it. the access token should still be returned in the oauth flow but not stored in this table

enterprise/server/routes/oauth_device.py
```diff
@@ -0,0 +1,448 @@
...
+
+    # Validate the user_code
+    device_code_entry = device_code_store.get_by_user_code(user_code)
+    if not device_code_entry:
+        return _html_response(
+            title='Error',
+            description='Invalid or expired device code.',
+            status_code=400,
+        )
+
```
malhotra5: @openhands note that we are access the db on the public `/verify` endpoint before the user has authenticated with keycloak


please change this so that we don't verify the code entry here - instead we perform the DB access after the keycloak callback has been executed

xingyaoww:
@OpenHands I'm not an OAuth expert, please do a /codereview-roasted and tell me if the implementation is correct

malhotra5:
Fixes made

1. SQLAlchemy Session Detachment Bug
2. Race Condition in Code Generation
3. Broken Multi-Device Support
4. HTML Templates in Python - UI is handled via FE
5. Inconsistent Error Handling - backend doesn't handle HTML responses 
6. Incomplete Spec Compliance - slow down protocol is implemented. we don't support fine grain grants for now so I haven't implemented that (its a native api key with access to cloud endpoints)
7. Inefficient Timezone Handling

Things not addressed in this PR
1. DOS Attack Vector - we've tried system wide rate limiting in the past but there were some complications in prod. I'm not sure of the current status but I think that's out of scope for this PR

2. Premature State Encoding - not a big deal imho

3. Database Bloat - I'd like to do this in a follow up PR as it involves writing a CRON job. Considering this is a security heavy PR I'd prefer including chore related code in a separate PR to mitigate any oversights during the review process.

xingyaoww:
@OpenHands I'm not an OAuth expert, please do a /codereview-roasted and tell me if the implementation is correct

enterprise/server/middleware.py
```diff
@@ -161,8 +161,14 @@ def _should_attach(self, request: Request) -> bool:
...
 
+        # OAuth Device Flow endpoints that should be public
+        is_oauth_device_public = path in (
+            '/oauth/device/authorize',  # Device authorization (unauthenticated)
+            '/oauth/device/token',  # Token polling (unauthenticated)
+        )
+
         is_mcp = path.startswith('/mcp')
-        return is_api_that_should_attach or is_mcp
+        return (is_api_that_should_attach or is_mcp) and not is_oauth_device_public
```
xingyaoww: This is very hard to parse, shall we combine it with this bool `is_api_that_should_attach`

enterprise/storage/device_code.py
```diff
@@ -0,0 +1,109 @@
...
+        """Update the last poll time and optionally increase the interval.
+
+        Args:
+            increase_interval: If True, increase the current interval for slow_down
+        """
+        self.last_poll_time = datetime.now(timezone.utc)
+
+        if increase_interval:
+            # Increase interval by 5 seconds, cap at 60 seconds (RFC 8628)
+            self.current_interval = min(self.current_interval + 5, 60)
```
xingyaoww, line 109: Just curious, @OpenHands is there an library we can use to do all these, instead of needing to implement these ourselves?

enterprise/server/middleware.py
malhotra5: done!

xingyaoww (approved):
LGTM

PR #11986, opened 2025-12-10
fix: Prevent avatar context menu from closing on diagonal cursor movement

### LINKED ISSUE #11933: Avatar context menu closes when moving cursor diagonally to menu (hover zone gap)

nathannewyen:
## Summary

Fixes #11933 - Avatar context menu closes when moving cursor diagonally to menu (hover zone gap)

### Problem

When hovering over the avatar in the sidebar, a context menu appears to the right. To interact with menu items, users naturally move their cursor diagonally from the avatar toward the menu. However, this diagonal path exits the parent container's hover zone momentarily, causing the menu to disappear before users can reach it.

This is a classic "hover intent" UX problem.

### Solution

Applied the same pattern already used successfully in `tools-context-menu.tsx`:

1. **Added `transition-all duration-200`** - Smooth transitions prevent jarring state changes and give users a small window during cursor movement
2. **Added `hover:opacity-100 hover:pointer-events-auto`** - Direct hover on the menu container keeps it visible once the cursor reaches it

### How It Works

```tsx
className={cn(
  // Base styles: hidden by default with smooth transition
  "opacity-0 pointer-events-none transition-all duration-200",
  // Show on parent hover (avatar)
  "group-hover:opacity-100 group-hover:pointer-events-auto",
  // Keep visible when hovering directly on menu (fixes diagonal cursor movement)
  "hover:opacity-100 hover:pointer-events-auto",
  // Show when menu is explicitly toggled via click
  showMenu && "opacity-100 pointer-events-auto",
)}
```

- `group-hover` triggers the menu to appear when hovering the avatar
- Direct `hover` on the menu container keeps it visible once the cursor reaches it
- The 200ms transition creates a smoother experience

## Test plan

- [x] Hover over avatar - menu appears
- [x] Move cursor diagonally from avatar to menu - menu stays visible
- [x] Click avatar - menu toggles correctly
- [x] Click outside menu - menu closes
- [x] Existing functionality preserved

Fixes #11933

### DIFF
--- a/frontend/src/components/features/sidebar/user-actions.tsx
+++ b/frontend/src/components/features/sidebar/user-actions.tsx
@@ -56,7 +56,13 @@ export function UserActions({ onLogout, user, isLoading }: UserActionsProps) {
       {(shouldShowUserActions || isOSS) && (
         <div
           className={cn(
-            "opacity-0 pointer-events-none group-hover:opacity-100 group-hover:pointer-events-auto",
+            // Base styles: hidden by default with smooth transition
+            "opacity-0 pointer-events-none transition-all duration-200",
+            // Show on parent hover (avatar)
+            "group-hover:opacity-100 group-hover:pointer-events-auto",
+            // Keep visible when hovering directly on menu (fixes diagonal cursor movement)
+            "hover:opacity-100 hover:pointer-events-auto",
+            // Show when menu is explicitly toggled via click
             showMenu && "opacity-100 pointer-events-auto",
           )}
         >


amanape:
Thanks for taking the time to submit a PR!

The idea is that once you reach the menu, hovering it keeps it visible. This doesn't actually fix the diagonal movement problem. The menu wrapper has `pointer-events-none` when not hovered, so you can't hover it until you're already on it. The problem is the gap between the avatar and the menu - when your cursor is in that gap (during diagonal movement), neither the avatar nor the menu is being hovered, so the menu closes before you can reach it. The `hover:` classes here would only help if you could somehow reach the menu element, but by the time you leave the avatar's hover zone, the menu already has `pointer-events-none` again.

I have opened an alternative PR in #11994. I encourage you to take a look! Thanks again 😄

nathannewyen:
Closing in favor of #11994 which uses the correct approach (CSS pseudo-element hover bridge). Thanks for the feedback - learned something new about hover zone problems! 🙏

PR #11987, opened 2025-12-10
fix(frontend): fix fetching the number of events on the front end (v1 conversations)

hieptl:
## Summary of PR

The current implementation retrieves the number of events on the front end by calling the app_server API endpoint /api/v1/events/count. This endpoint returns the number of events by counting files located in the ~/.openhands/v1/events/{conversation_id} directory.

However, when fetching all events via WebSocket with resend_all = true, the agent_server retrieves events from the Docker container. Because these two sources differ, the event count returned by the app_server does not necessarily match the number of events retrieved by the agent_server.

As a result, the system is counting events and searching events from different data sources, causing inconsistencies.

To resolve this issue, the front end should instead call the agent_server endpoint /api/conversations/{conversation_id}/events/count. This ensures that both event counting and event retrieval rely on the same source.

**Acceptance Criteria:**
- Update the front end to call /api/conversations/{conversation_id}/events/count from the agent_server to retrieve the event count for the current conversation.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:eefbbcc-nikolaik   --name openhands-app-eefbbcc   docker.openhands.dev/openhands/openhands:eefbbcc
```


frontend/src/services/__tests__/event-service.api.test.ts
amanape, line 1: I’d rather remove this test file because it’s testing more implementation rather than behaviour. For example:

```ts
expect(buildHttpBaseUrl).toHaveBeenCalledWith(conversationUrl);
expect(buildSessionHeaders).toHaveBeenCalledWith(sessionApiKey);
expect(mockedAxios.get).toHaveBeenCalledWith(
  `${runtimeUrl}/api/conversations/${conversationId}/events/count`,
  { headers },
);
```

And the actual implementation:

```ts
const runtimeUrl = buildHttpBaseUrl(conversationUrl);
const headers = buildSessionHeaders(sessionApiKey);
const { data } = await axios.get<number>(
  `${runtimeUrl}/api/conversations/${conversationId}/events/count`,
  { headers },
);
```

The test is just a mirror of the implementation.

- "Did we call `buildHttpBaseUrl` with `conversationUrl`?” -> Yes, the first line does that
- "Did we call `buildSessionHeaders` with `sessionApiKey`?” -> Yes, the second line does that
- "Did we call `axios.get` with this URL and headers?” -> Yes, the remaining lines do that

This shows that we’re not testing behaviour, but instead verifying the code was written a certain way.

Why is this bad?

**Scenario**: You refactor to use fetch instead of `axios`:

```ts
const response = await fetch(url, { headers });
const data = await response.json();
```

The function still works identically, but the test breaks. The test is coupled to _implementation_, not _outcome_.

What might be better?

- Integration tests against a real/fake HTTP server (like `msw`) that verify actual HTTP behavior
  - You'd intercept real network calls and verify the actual behavior without caring how the code internally builds the request.
- Unit tests for the utility functions (`buildHttpBaseUrl`, `buildSessionHeaders`) since those have real logic
- E2E tests that verify the actual user-facing behavior

For now though, you can feel safe to just remove this test file 😄

amanape (changes requested):
Thanks for the PR! Would this also resolve the loading issue Tim brought up?

frontend/src/services/__tests__/event-service.api.test.ts
hieptl: > Would this also resolve the loading issue Tim brought up?

Hello @amanape,

I do not believe this will resolve the issue Tim raised. This pull request addresses a different problem: when I refreshed the conversation page, the loading spinner continued indefinitely. After checking, I found that the number of received events did not match the number of expected events.

PR #11978, opened 2025-12-09
APP-223 - added container around avatar and menu to cover hover area

HeyItsChloe:
## Summary of PR

- added container around the avatar and the menu to cover the "hover" area

## Change Type

- [x] Refactor
- [ ] Other (dependency update, docs, typo fixes, etc.)

## Checklist

- [ ] I have read and reviewed the code and I understand what the code is doing.
- [ ] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves #11933

## Release Notes

- [ ] Include this change in the Release Notes.

PR #11994, opened 2025-12-10
fix(frontend): Fix avatar context menu closing on diagonal mouse movement

### LINKED ISSUE #11933: Avatar context menu closes when moving cursor diagonally to menu (hover zone gap)

amanape:
## Summary of PR

Adds an invisible hover bridge using a CSS pseudo-element that extends the hover zone when the menu is visible. This allows users to move their cursor diagonally from the avatar to the menu without the menu closing. Includes an e2e test to verify the fix.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #11933
Closes #11954 #11986 #11978

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:ed4b894-nikolaik   --name openhands-app-ed4b894   docker.openhands.dev/openhands/openhands:ed4b894
```

### DIFF
--- a/frontend/tests/avatar-menu.spec.ts
+++ b/frontend/tests/avatar-menu.spec.ts
@@ -8,7 +8,11 @@ import test, { expect } from "@playwright/test";
  */
 test("avatar context menu stays open when moving cursor diagonally to menu", async ({
   page,
+  browserName,
 }) => {
+  // Skip on WebKit - Playwright's mouse.move() doesn't reliably trigger CSS hover states
+  test.skip(browserName === "webkit", "Playwright hover simulation unreliable");
+
   await page.goto("/");
 
   // Get the user avatar button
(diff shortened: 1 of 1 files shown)


amanape:
Thank you for your feedback @hieptl.

For Issue 1, I wasn't able to find a straightforward solution that wouldn't require JavaScript to dynamically calculate the height. I believe the current height of 300px provides sufficient distance until the user hovers over the actual menu. If it becomes problematic, we can create a new issue to explore a potential JavaScript solution.

Regarding Issue 2, this is a separate problem. The avatar's placement was designed to be this close, according to Paul’s designs. Rather than addressing it here, please open a new issue and contact Paul for his suggestions.

hieptl (approved):
Thank you! 🙏

PR #12019, opened 2025-12-12
feat(frontend): hide billing section in organization settings page for self-hosted (org support)

hieptl:
## Summary of PR

In self-hosted, within an Org, Admins will bring their own LLM key.

They will not need to purchase credits for use of the OpenHands LLM provider. As a result, there should be a configurable option to show or hide the Billing section within the Organization Settings page. For self-hosted customers, this should default to hidden.

https://github.com/user-attachments/assets/95379e04-6572-49af-9a25-98985266442a

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:c8306be-nikolaik   --name openhands-app-c8306be   docker.openhands.dev/openhands/openhands:c8306be
```

### DIFF
--- a/frontend/__tests__/routes/manage-org.test.tsx
+++ b/frontend/__tests__/routes/manage-org.test.tsx
@@ -802,38 +802,6 @@ describe("Manage Org Route", () => {
       getConfigSpy.mockRestore();
     });
 
-    it("should not show Add Credits modal when HIDE_BILLING is true even if state is set", async () => {
-      // Arrange
-      const getConfigSpy = vi.spyOn(OptionService, "getConfig");
-      getConfigSpy.mockResolvedValue({
-        APP_MODE: "saas",
-        GITHUB_CLIENT_ID: "test",
-        POSTHOG_CLIENT_KEY: "test",
-        FEATURE_FLAGS: {
-          ENABLE_BILLING: false,
-          HIDE_LLM_SETTINGS: false,
-          HIDE_BILLING: true,
-          ENABLE_JIRA: false,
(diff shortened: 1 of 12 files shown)


frontend/__tests__/routes/manage-org.test.tsx
```diff
@@ -707,4 +707,163 @@ describe("Manage Org Route", () => {
...
+          ENABLE_LINEAR: false,
+        },
+      });
+
+      // Act
+      renderManageOrg();
+      await screen.findByTestId("manage-org-screen");
+      await selectOrganization({ orgIndex: 0 });
+
+      // Assert - Modal should not be accessible even if somehow triggered
```
amanape: The test never sets `addCreditsFormVisible` to true. Since the `”Add"` button is already hidden when `HIDE_BILLING` is `true`, this is redundant with the previous test (see `"should hide Add Credits button when HIDE_BILLING is true”`). I verified this by removing the `!isBillingHidden` guard — test still passes. Suggest deleting.

hieptl: Hello @amanape,

PR #12040, opened 2025-12-13
chore(frontend): Clean up console warnings in frontend test suite

### LINKED ISSUE #11931: Clean up console warnings in frontend test suite
## Description

When running `npm test` in the frontend, multiple console warnings are emitted that clutter the test output and may indicate potential issues in the codebase. While all tests pass, these warnings should be addressed to maintain code quality and improve test output readability.

## Reproduction Steps

1. Navigate to the frontend directory
2. Run `npm test`
3. Observe the stderr output interspersed with test results

## Note

Before working on this issue, it would be beneficial to familiarize yourself with our existing test setup and patterns.

HeyItsChloe:
## Summary of PR

Remove console warnings / logs from FE unit tests 

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves #11931

## Release Notes


frontend/src/components/features/microagent-management/microagent-management-add-microagent-button.tsx
```diff
@@ -25,15 +25,14 @@ export function MicroagentManagementAddMicroagentButton({
   };
 
   return (
-    <button
-      type="button"
+    <span
       onClick={handleClick}
```
hieptl, line 30: **suggestion:** We do not use onClick with a span tag. If this change is intended to suppress warnings when running npm test, I believe we should explore an alternative solution. Thank you! 🙏 

There are important considerations:
1.	Semantics & Accessibility
- <span> is an inline, non-interactive element. Using it as a clickable element is not semantic. Screen readers and assistive technologies may not recognize it as a button.
- If you need a clickable action, the <button> element is preferred. Buttons are automatically keyboard-accessible and announced correctly to assistive tech.
2. Keyboard Accessibility
- <span> is not focusable by default. Users navigating via keyboard (Tab key) cannot activate it unless you add tabIndex="0" and handle onKeyPress or onKeyDown.

frontend/src/components/shared/buttons/tooltip-button.tsx
```diff
@@ -40,8 +40,7 @@ export function TooltipButton({
   };
 
   const buttonContent = (
-    <button
-      type="button"
+    <span
       aria-label={ariaLabel}
       data-testid={testId}
       onClick={handleClick}
```
hieptl, line 47: **suggestion:** We do not use onClick with a span tag. If this change is intended to suppress warnings when running npm test, I recommend exploring an alternative solution. I have explained the reasons in my previous comment. Thank you! 🙏

hieptl:
Hello @HeyItsChloe, it is not necessary to include the Linear ticket in the pull request. Specifying the GitHub issue, if any, is sufficient. Please note that Linear tickets are accessible only to All Hands members. Thank you! 🙏

frontend/src/components/features/microagent-management/microagent-management-add-microagent-button.tsx
amanape: The simplest change I can suggest is to add these props:

```html
<span
  tabIndex={0} 
  role="button" 
/>
```

If you find any other alternatives though, feel free to apply them as well.

HeyItsChloe: Yup, I agree with both of you. For this warning, I think we have three main options:
	1.	Leave the nested buttons as-is and only suppress the warnings in tests. They’ll still render the same way in the browser.
	2.	Use a span with tabIndex={0} and role="button" as a workaround.
	3.	Refactor the AccordionItem from HeroUI that’s rendering the tooltipButton and causing the warning. We could replace it with a div that has an onClick handler instead.

I’m leaning toward option 1, but let me know which approach you’d like me to pursue — or if there’s another alternative I should consider.

frontend/src/components/shared/buttons/tooltip-button.tsx
HeyItsChloe: Here, I think we have the same 3 options as above

frontend/src/components/features/home/shared/generic-dropdown-menu.tsx
```diff
@@ -80,27 +80,31 @@ export function GenericDropdownMenu<T>({
...
-                    selectedItem,
-                    getItemProps,
-                  )}
-                  {numberOfRecentItems > 0 &&
-                    index === numberOfRecentItems - 1 && (
-                      <div className="border-b border-[#727987] bg-[#454545] pb-1 mb-1 h-[1px]" />
-                    )}
+                  {stickyTopItem}
+                  {filteredItems.map((item, index) => (
+                    <React.Fragment key={index}>
```
hieptl: **suggestion:** Could we avoid using index here? Using an index as a key is generally not considered a best practice, especially for a dropdown menu where the list of items may change frequently. Thank you! 🙏

HeyItsChloe: Ah - added a func to get a unique key for each item. Is there already a func like this that you know of in the repo? If not, should I move this to a util?

frontend/src/components/features/home/shared/generic-dropdown-menu.tsx
```diff
@@ -54,7 +54,16 @@ export function GenericDropdownMenu<T>({
...
-
+  const getUniqueItemKey = (item: T): string => {
+    if (typeof item === "string") return item;
+    if (typeof item === "object" && item !== null && "id" in item) {
+      return String(item.id);
+    }
+    if (typeof item === "object" && item !== null) {
+      return String(Object.values(item)[0]);
+    }
+    return String(`${item}-${crypto.randomUUID()}`);
```
hieptl, line 65: Issue:
The fallback branch uses crypto.randomUUID() to generate the key. This makes the output non-deterministic, meaning the same input can produce different keys across calls.

Impact:
This can break React reconciliation, memoization, and caching logic, and may cause unnecessary re-renders or cache misses.

Suggestion:
Avoid using random UUIDs for key generation. Keys should be deterministic based on stable properties of the item. If no stable identifier exists, consider requiring the caller to explicitly provide one.

hieptl, line 65: Issue:
Wrapping a template literal with String() is unnecessary since template literals already return a string.

Suggestion:
Remove the redundant String() call.

frontend/src/components/features/home/shared/generic-dropdown-menu.tsx
```diff
@@ -54,7 +54,16 @@ export function GenericDropdownMenu<T>({
 }: GenericDropdownMenuProps<T>) {
   const hasItems = filteredItems.length > 0;
   const showEmptyState = !hasItems && !stickyTopItem && !stickyFooterItem;
-
+  const getUniqueItemKey = (item: T): string => {
+    if (typeof item === "string") return item;
+    if (typeof item === "object" && item !== null && "id" in item) {
+      return String(item.id);
+    }
+    if (typeof item === "object" && item !== null) {
```
hieptl, line 62: Issue:
The logic repeatedly checks typeof item === "object" && item !== null across multiple branches.

Impact:
This makes the function harder to read and maintain, especially if additional rules are added later.

Suggestion:
Refactor the control flow to centralize object checks or use early returns to simplify the logic.

frontend/src/components/features/home/shared/generic-dropdown-menu.tsx
```diff
@@ -54,7 +54,16 @@ export function GenericDropdownMenu<T>({
 }: GenericDropdownMenuProps<T>) {
   const hasItems = filteredItems.length > 0;
   const showEmptyState = !hasItems && !stickyTopItem && !stickyFooterItem;
-
+  const getUniqueItemKey = (item: T): string => {
```
hieptl, line 57: Issue:
The function is declared with a generic type T, but the implementation assumes that item may be an object with an id property.

Impact:
This provides a false sense of type safety and prevents TypeScript from catching invalid usage at compile time.

Suggestion:
Constrain the generic (e.g., T extends { id?: string | number }) or replace it with a union type that accurately reflects the supported inputs.

frontend/src/components/features/home/shared/generic-dropdown-menu.tsx
```diff
@@ -54,7 +54,16 @@ export function GenericDropdownMenu<T>({
...
+    if (typeof item === "string") return item;
+    if (typeof item === "object" && item !== null && "id" in item) {
+      return String(item.id);
```
hieptl, line 60: Issue:
The check for "id" in item does not ensure that id is a valid primitive value suitable for use as a key.

Impact:
If id is undefined, null, or an object, the generated key may be invalid or misleading (e.g., "undefined" or "[object Object]").

Suggestion:
Explicitly validate that id is a string or number before using it as a key.

frontend/src/components/features/home/shared/generic-dropdown-menu.tsx
```diff
@@ -54,7 +54,16 @@ export function GenericDropdownMenu<T>({
...
+    }
+    if (typeof item === "object" && item !== null) {
+      return String(Object.values(item)[0]);
```
hieptl, line 63: Issue:
Deriving a key from Object.values(item)[0] relies on property order and implicitly assumes the first value is unique and meaningful.

Impact:
Different objects may generate the same key, leading to collisions and unpredictable behavior.

Suggestion:
Do not infer uniqueness from arbitrary object properties. If an object does not have a well-defined identifier (e.g., id), the API should require one instead of guessing.

PR #12076, opened 2025-12-17
Fixed performance bug in remote sandbox service

tofarr:
## Summary of PR

Nasty performance bug due to evolution. A week or two back, it was found that the url we use for webhook callbacks could not include the id of the conversation because this would mean it would have to be in the environment variables which would break warm runtimes. So it was removed and a method was devised to get the sandbox id based on nothing more than the session_api_key. Unfortunately, this introduced a nasty performance bug which is manifesting on production - Massive numbers of get requests are being fired at the runtime api.

This PR calls list which returns the currently running runtimes and uses this to get the session_id (Sandbox id) - rather than checking each of these in the runtime api one at a time.

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:72eda18-nikolaik   --name openhands-app-72eda18   docker.openhands.dev/openhands/openhands:72eda18
```

### DIFF
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -315,7 +315,9 @@ class RemoteSandboxService(SandboxService):
             for runtime in content['runtimes']:
                 if session_api_key == runtime['session_api_key']:
                     query = await self._secure_select()
-                    query = query.filter(StoredRemoteSandbox.id == runtime.get('session_id'))
+                    query = query.filter(
+                        StoredRemoteSandbox.id == runtime.get('session_id')
+                    )
                     result = await self.db_session.execute(query)
                     sandbox = result.first()
                     if sandbox is None:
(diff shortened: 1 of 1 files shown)


openhands/app_server/sandbox/remote_sandbox_service.py
```diff
@@ -303,21 +303,27 @@ async def get_sandbox_by_session_api_key(
...
-                if runtime and runtime.get('session_api_key') == session_api_key:
-                    return await self._to_sandbox_info(stored_sandbox, runtime)
-            except Exception:
-                # Continue checking other sandboxes if one fails
-                continue
-
+        try:
+            response = await self._send_runtime_api_request(
+                'GET',
+                '/list',
```
malhotra5, line 311: Will this get all runtimes or only those that are running? (i.e is there an upper bound on the number of items?)

tofarr: It gets only those that are running.

malhotra5 (approved):

PR #12095, opened 2025-12-19
Fix xterm dimensions error with explicit checks instead of try-catch

neubig:
## Summary of PR

This PR fixes the PostHog error `TypeError: Cannot read properties of undefined (reading 'dimensions')` that has been occurring ~876 times per day (26,282 total occurrences over 30 days).

The error occurs in the `@xterm/xterm` library's `Viewport.syncScrollArea()` method when `FitAddon.fit()` is called but the terminal's internal `_renderService` is undefined. This happens in the following scenarios:

- Terminal component is hidden (CSS `display: none`) when tab switching
- ResizeObserver fires during terminal disposal/unmount
- Terminal is not fully initialized when `fit()` is called

### Solution

Instead of using a try-catch to silently suppress the error, this fix adds explicit checks before calling `fit()` via a new `canFitTerminal()` helper function and `fitTerminalSafely()` wrapper:

1. **Terminal instance check**: Verify terminal and fitAddon exist
2. **Container element check**: Verify the container element exists
3. **Visibility check**: Verify the element is not hidden (`display: none`)
4. **Dimensions check**: Verify the element has non-zero width and height
5. **Terminal opened check**: Verify `terminal.element` exists (set after `open()` is called)
6. **Disposal check**: Track disposal state via `isDisposed` ref to prevent operations on disposed terminals

### Why no try-catch?

As discussed in the issue, explicit checks are preferred over try-catch for several reasons:

- **Explicit behavior**: The code clearly documents when and why fit operations are skipped
- **No silent failures**: Errors that occur for unexpected reasons will still be reported
- **Better debugging**: If a new edge case emerges, it will be visible rather than silently caught
- **Performance**: Avoiding exception handling in the hot path (resize events)

## Change Type

- [x] Bug fix

## Checklist

## Fixes

Resolves #22

## Release Notes

Fixed a TypeError that occurred when the terminal component was hidden or disposed during resize operations. The fix adds explicit safety checks before calling the xterm FitAddon.fit() method, preventing the "Cannot read properties of undefined (reading 'dimensions')" error that was affecting ~876 users per day.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/33cd4b4df1a24012a1e8720bfdff2683)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:c20f496-nikolaik   --name openhands-app-c20f496   docker.openhands.dev/openhands/openhands:c20f496
```

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -477,7 +477,11 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
         if sandbox.status in (None, SandboxStatus.ERROR):
             raise SandboxError(f'Sandbox status: {sandbox.status}')
         if sandbox.status == SandboxStatus.RUNNING:
-            return
+            # There are still bugs in the remote runtime - they report running while still just
+            # starting resulting in a race condition. Manually check that it is actually
+            # running.
+            if await self._check_agent_server_alive(sandbox):
+                return
         if sandbox.status != SandboxStatus.STARTING:
             raise SandboxError(f'Sandbox not startable: {sandbox.id}')
 
@@ -490,8 +494,18 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
             if sandbox.status not in (SandboxStatus.STARTING, SandboxStatus.RUNNING):
                 raise SandboxError(f'Sandbox not startable: {sandbox.id}')
             if sandbox_info.status == SandboxStatus.RUNNING:
-                return
+                # There are still bugs in the remote runtime - they report running while still just
+                # starting resulting in a race condition. Manually check that it is actually
+                # running.
+                if await self._check_agent_server_alive(sandbox_info):
+                    return
         raise SandboxError(f'Sandbox failed to start: {sandbox.id}')
+
+    async def _check_agent_server_alive(self, sandbox_info: SandboxInfo) -> bool:
+        agent_server_url = self._get_agent_server_url(sandbox_info)
+        url = f'{agent_server_url.rstrip("/")}/alive'
+        response = await self.httpx_client.get(url)
+        return response.is_success
 
     def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
         """Get agent server url for running sandbox."""
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -122,18 +122,9 @@ class RemoteSandboxService(SandboxService):
             _logger.error(f'HTTP error for URL {url}: {e}')
             raise
 
-    async def _to_sandbox_info(
+    def _to_sandbox_info(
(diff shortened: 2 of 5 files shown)


frontend/__tests__/hooks/use-terminal.test.tsx
```diff
@@ -97,3 +111,173 @@ describe("useTerminal", () => {
...
+      const canFit =
+        terminal !== null && fitAddon !== null && container !== null;
+
+      if (canFit) {
+        fitAddon.fit(); // Would throw if called
+      }
+
+      // fit() was never called, so no error
+      expect(fitAddon.fit).not.toHaveBeenCalled();
+    });
```
amanape: Each test essentially says "if I check condition X before calling `fit()`, and condition X is `false`, then `fit()` won't be called." The tests are just verifying that `if (false) { … }` doesn't execute the block.

This tests JavaScript's if statement

frontend/__tests__/hooks/use-terminal.test.tsx
```diff
@@ -97,3 +111,173 @@ describe("useTerminal", () => {
...
+  describe("error reproduction without safety checks", () => {
+    it("should throw 'dimensions' error when fit() is called on invalid terminal", () => {
+      // This test demonstrates the original error that was occurring
+      const fitAddon = createErrorThrowingFitAddon();
+
+      // Without safety checks, calling fit() throws the error
+      expect(() => fitAddon.fit()).toThrow(
+        "Cannot read properties of undefined (reading 'dimensions')",
+      );
+    });
```
amanape: This tests that `createErrorThrowingFitAddon` throws when called—which is defined to always throw. It doesn't demonstrate anything about when `xterm` actually throws this error.

frontend/__tests__/hooks/use-terminal.test.tsx
amanape, line 1: The actual `canFitTerminal` function is defined in `use-terminal.ts`, but none of these tests import or call it. Instead, each test manually recreates the safety check logic.

If someone changes the implementation, the tests will still pass because they're testing their own inlined logic, not the real code. These tests provide zero regression protection, and:

- Don't test the actual implementation
- Provide false confidence (they'll pass even if the implementation breaks)
- Are essentially testing that `if (false) { x() }` doesn't call `x()`
- Cannot catch regressions

The tests serve more as documentation of intent than as actual regression tests. They'd be better as comments in the implementation itself.

neubig:
Thanks @amanape , apologies, this was generated by openhands and should have still been marked as draft. I'll do a careful manual check of this, fix, and re-request review.

neubig:
@openhands read `amanape`'s comments and make a first pass at addressing the review.

neubig:
OK, this looks much better to me now, thanks for the patience!

PR #12124, opened 2025-12-22
feat: prevent signups using email addresses with a plus sign and enforce the existing email pattern

hieptl:
## Summary of PR

To prevent recurring signups using email addresses that include a plus sign (+), the system should allow the first signup using such an address (e.g., joe+openhands@example.com) but reject subsequent signups that follow the same pattern.

For example, if a user attempts to sign up with joe+1@example.com, the system should check whether an account already exists for joe@example.com or for any address matching joe+{any text}@example.com.

If this condition is met, the signup attempt should fail, and the following message should be displayed after submission:

> “Your account is unable to be created. Please use a different login or try again.”

Acceptance Criteria:
- The system allows the first-time signup using an email address that contains a plus sign (+).
- For subsequent signup attempts, the system detects existing accounts that share the same base email (e.g., joe@example.com or joe+{any text}@example.com).
- If a matching account already exists, the signup attempt is rejected.
- When a signup is rejected under this rule, the specified error message is displayed to the user.

We can refer to the videos below for the output of the pull request.

**Case 1:** If the base email address (hiep@openhands.dev) already exists and a new account is created using a plus-addressed variant (e.g., hiep+1@openhands.dev), the new account must be rejected.

https://github.com/user-attachments/assets/4ad5003b-a55f-43d4-b679-badba8193446

**Case 2:** If an email address with a plus sign (e.g., hiep+1@openhands.dev) already exists and a new account is created using another plus-addressed variant of the same base email (e.g., hiep+2@openhands.dev), the new account must be rejected.

https://github.com/user-attachments/assets/06223b76-623e-4fe2-8d66-4f3e8a96f13b

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:87ab423-nikolaik   --name openhands-app-87ab423   docker.openhands.dev/openhands/openhands:87ab423
```

### DIFF
--- a/enterprise/server/auth/token_manager.py
+++ b/enterprise/server/auth/token_manager.py
@@ -29,7 +29,6 @@ from server.auth.constants import (
 from server.auth.email_validation import (
     extract_base_email,
     get_base_email_regex_pattern,
-    has_plus_modifier,
     matches_base_email,
 )
 from server.auth.keycloak_manager import get_keycloak_admin, get_keycloak_openid
@@ -618,19 +617,23 @@ class TokenManager:
         """Check if a user with the same base email already exists.
 
         This method checks for duplicate signups using email + modifier.
-        If the provided email has a + modifier, it checks if any user exists
-        with the same base email (either the base email itself or another
-        email with the same base and a different + modifier).
+        It checks if any user exists with the same base email, regardless of whether
+        the provided email has a + modifier or not.
+
+        Examples:
+            - If email is "joe+test@example.com", it checks for existing users with
+              base email "joe@example.com" (e.g., "joe@example.com", "joe+1@example.com")
+            - If email is "joe@example.com", it checks for existing users with
+              base email "joe@example.com" (e.g., "joe+1@example.com", "joe+test@example.com")
 
         Args:
-            email: The email address to check (may contain + modifier)
+            email: The email address to check (may or may not contain + modifier)
             current_user_id: The user ID of the current user (to exclude from check)
 
         Returns:
             True if a duplicate is found (excluding current user), False otherwise
         """
-        if not email or not has_plus_modifier(email):
-            # Only check emails with + modifier
+        if not email:
             return False
 
         base_email = extract_base_email(email)
--- a/enterprise/tests/unit/test_token_manager.py
+++ b/enterprise/tests/unit/test_token_manager.py
@@ -126,16 +126,29 @@ class TestCheckDuplicateBaseEmail:
 
     @pytest.mark.asyncio
     async def test_check_duplicate_base_email_no_plus_modifier(self, token_manager):
-        """Test that emails without + modifier return False."""
+        """Test that emails without + modifier are still checked for duplicates."""
         # Arrange
         email = 'joe@example.com'
         current_user_id = 'user123'
 
-        # Act
-        result = await token_manager.check_duplicate_base_email(email, current_user_id)
+        with (
+            patch.object(
+                token_manager, '_query_users_by_wildcard_pattern'
+            ) as mock_query,
+            patch.object(token_manager, '_find_duplicate_in_users') as mock_find,
+        ):
+            mock_find.return_value = False
+            mock_query.return_value = {}
 
-        # Assert
-        assert result is False
+            # Act
+            result = await token_manager.check_duplicate_base_email(
+                email, current_user_id
+            )
+
+            # Assert
+            assert result is False
+            mock_query.assert_called_once()
+            mock_find.assert_called_once()
 
     @pytest.mark.asyncio
     async def test_check_duplicate_base_email_empty_email(self, token_manager):
@@ -221,7 +234,7 @@ class TestCheckDuplicateBaseEmail:
     async def test_check_duplicate_base_email_keycloak_connection_error(
         self, token_manager
     ):
-        """Test that KeycloakConnectionError is handled gracefully."""
+        """Test that KeycloakConnectionError triggers retry and raises RetryError."""
         # Arrange
         email = 'joe+test@example.com'
         current_user_id = 'user123'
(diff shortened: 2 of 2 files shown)


frontend/__tests__/components/features/auth-modal.test.tsx
```diff
@@ -115,11 +128,91 @@ describe("AuthModal", () => {
...
+        "/?redirect_url=/dashboard&duplicated_email=true",
+      ];
+
+      // Act
+      renderAuthModalWithRouter(initialEntries);
+
+      // Assert
+      const errorMessage = screen.getByText("AUTH$DUPLICATE_EMAIL_ERROR");
+      expect(errorMessage).toBeInTheDocument();
+    });
```
amanape: If this was written following TDD, these last three tests might be not be required. TDD tests should _drive_ code changes. If a test passes without writing new code, it's either a regression test added later, or unnecessary.

The first two* tests justify the code change you made in `auth-modal.tsx`. Tests 3 (`"should not display error message when duplicated_email query parameter is false”` and 4 (`"should not display error message when duplicated_email query parameter has other value”`) don't actually protect against regressions in any meaningful way (remove the `hasDuplicatedEmail` conditional render in `auth-modal.tsx` and observe that these tests still pass!). They only verify that a condition doesn't render something, but they can't distinguish between "correctly not rendering" and "broken/missing entirely.”

Test 5 (`"should display error message with other query parameters present”`) is essentially a duplicate of Test 1 with extra query params -- it's testing React Router, not your logic.

So really, only Test 1 and Test 2 are the true TDD-driven test for this feature. (*though one can also argue that Test 2 isn’t needed either since it passes for free once you’ve written the code to satisfy Test 1, but you should be OK here for documentation purposes)

**To summarize: feel free to keep Tests 1 and 2 and remove the rest**

hieptl: Hello @amanape,

enterprise/server/auth/token_manager.py
```diff
@@ -509,6 +516,180 @@ async def get_user_id_from_user_email(self, email: str) -> str | None:
...
+
+        # Query for users with emails matching the base pattern using wildcard
+        # Pattern: {local_part}*@{domain} - catches base email and all + variants
+        # This may also catch unintended matches (e.g., joesmith@example.com), but
+        # they will be filtered out by the regex pattern check later
+        # Use 'search' parameter for Keycloak 26+ (better wildcard support)
+        wildcard_queries = [
+            {'search': f'{local_part}*@{domain}'},  # Try 'search' parameter first
+            {'q': f'email:{local_part}*@{domain}'},  # Fallback to 'q' parameter
+        ]
```
tofarr, line 545: Although gmail and outlook disallow it, technically the local part of an email can contain a `*`. Do you think there is the scope for some sort of injection attack here?

enterprise/server/auth/token_manager.py
```diff
@@ -509,6 +516,180 @@ async def get_user_id_from_user_email(self, email: str) -> str | None:
...
+
+        try:
+            local_part, domain = base_email.rsplit('@', 1)
+            users = await self._query_users_by_wildcard_pattern(local_part, domain)
+            return self._find_duplicate_in_users(users, base_email, current_user_id)
+
+        except KeycloakConnectionError:
+            logger.exception('KeycloakConnectionError when checking duplicate email')
+            # On connection error, allow signup to proceed (fail open)
+            return False
```
tofarr: Given that we have a retry decorator, should we simply re-raise this? Either that or we should remove the decorator, as it won't do anything

enterprise/server/auth/token_manager.py
```diff
@@ -509,6 +516,180 @@ async def get_user_id_from_user_email(self, email: str) -> str | None:
...
+        try:
+            keycloak_admin = get_keycloak_admin(self.external)
+            # Use the sync method (python-keycloak doesn't have async delete_user)
+            # Run it in a thread executor to avoid blocking the event loop
+            await asyncio.to_thread(keycloak_admin.delete_user, user_id)
+            logger.info(f'Successfully deleted Keycloak user {user_id}')
+            return True
+        except KeycloakConnectionError:
+            logger.exception(f'KeycloakConnectionError when deleting user {user_id}')
+            return False
```
tofarr: Given that we have a retry decorator, should we simply re-raise this? Either that or we should remove the decorator, as it won't do anything

enterprise/server/auth/token_manager.py
tofarr: I suppose since it is inclusive, it would just mean that additional email addresses are checked. 🤔

enterprise/server/auth/email_validation.py
```diff
@@ -0,0 +1,110 @@
+"""Email validation utilities for preventing duplicate signups with + modifier."""
+
+import re
+from typing import Optional
+
+
+def extract_base_email(email: str) -> Optional[str]:
```
enyst:
```suggestion
def extract_base_email(email: str) -> str | None:
```
Super tiny nit: we haven’t been consistent on this, but fwiw we have decided long ago to use the new style, for code readability and consistency. Guess it might be better to include it in some… hmm, biweekly agent sweep 😅

enterprise/server/auth/token_manager.py
h-openhands: Hello @tofarr,

enterprise/server/auth/token_manager.py
h-openhands: Hello @tofarr,

enterprise/server/auth/email_validation.py
h-openhands: Hello @enyst,

PR #12134, opened 2025-12-23
fix(backend): preserve users custom llm settings during settings migrations

hieptl:
### Demo

https://github.com/user-attachments/assets/b762f8c4-b8f0-41a8-bf8a-ebaac1aa24be

### Problem

https://github.com/user-attachments/assets/fbf1d602-3cf0-47da-a027-4facd7ab30a5

According to the video above, when accessing staging and production environments, users' custom LLM settings (model, API key, and base URL) were being reset to defaults during version upgrades/migrations. This affected users who had configured their own LLM provider (BYOR - Bring Your Own Resource), causing them to lose their custom configurations.

### Root Cause

The issue was in `enterprise/storage/saas_settings_store.py` within the `update_settings_with_litellm_default()` method. During user settings version upgrades/migrations, the code was unconditionally overwriting:

- `llm_model` → always set to default `claude-opus-4-5-20251101`
- `llm_api_key` → always set to LiteLLM-generated key
- `llm_base_url` → always set to `LITE_LLM_API_URL`

**Problematic Flow:**
1. When a user's `user_version` didn't match `CURRENT_USER_SETTINGS_VERSION`, the system triggered a migration
2. The migration called `create_default_settings()`, which invoked `update_settings_with_litellm_default()`
3. This method always overwrote user settings with defaults, regardless of whether the user had custom configurations

### Solution

Modified `update_settings_with_litellm_default()` to preserve user settings when they exist:

1. **Preserve custom model**: Only set default if user doesn't already have a model configured (None, empty, or whitespace-only)
2. **Preserve custom API key**: Only set LiteLLM default if user doesn't already have an API key configured
3. **Preserve custom base URL**: Only set LiteLLM default if user doesn't already have a base URL configured
4. **Update LiteLLM metadata**: Pass the user's actual model (or default) to LiteLLM user creation metadata for accurate tracking

**Key Changes:**
- Determine `llm_model_to_use` once at the beginning of the method
- Check for existing API key and base URL before applying defaults
- Preserve user's custom settings if they exist and are non-empty
- Use defaults only for new users or users without configured settings

### Code Changes

**File:** `enterprise/storage/saas_settings_store.py`

- Added model determination logic that preserves user's custom model
- Added API key preservation logic for BYOR scenarios
- Added base URL preservation logic for BYOR scenarios
- Updated `_create_user_in_lite_llm()` method signature to accept `llm_model` parameter
- Modified LiteLLM metadata to reflect the user's actual model instead of always using default

### Testing

Added 13 new tests:

**Model Preservation Tests (4):**
- `test_update_settings_with_litellm_default_preserves_custom_model` - Verifies custom models are preserved
- `test_update_settings_with_litellm_default_uses_default_when_no_model` - Verifies defaults are applied for new users
- `test_update_settings_with_litellm_default_handles_empty_string_model` - Edge case handling
- `test_update_settings_with_litellm_default_handles_whitespace_model` - Edge case handling

**API Key & Base URL Preservation Tests (9):**
- `test_update_settings_with_litellm_default_preserves_custom_api_key` - Verifies custom API keys are preserved
- `test_update_settings_with_litellm_default_preserves_custom_base_url` - Verifies custom base URLs are preserved
- `test_update_settings_with_litellm_default_preserves_custom_api_key_and_base_url` - Verifies both are preserved together
- `test_update_settings_with_litellm_default_uses_default_api_key_when_none` - Verifies defaults when no API key
- `test_update_settings_with_litellm_default_uses_default_base_url_when_none` - Verifies defaults when no base URL
- `test_update_settings_with_litellm_default_handles_empty_api_key` - Edge case: empty string API key
- `test_update_settings_with_litellm_default_handles_empty_base_url` - Edge case: empty string base URL
- `test_update_settings_with_litellm_default_handles_whitespace_api_key` - Edge case: whitespace-only API key
- `test_update_settings_with_litellm_default_handles_whitespace_base_url` - Edge case: whitespace-only base URL

**Test Results:** All 24 tests pass (11 existing + 13 new)

### Impact

✅ **Users' custom LLM model preferences are preserved** during version upgrades and migrations  
✅ **Users' custom API keys are preserved*
✅ **Users' custom base URLs are preserved**
✅ **New users still receive appropriate default settings**  
✅ **LiteLLM metadata accurately reflects the model being used**  
✅ **No breaking changes** - fully backward compatible

### Files Changed

- `enterprise/storage/saas_settings_store.py` - Core fix implementation
- `enterprise/tests/unit/test_saas_settings_store.py` - Added comprehensive test coverage

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:a2c4db3-nikolaik   --name openhands-app-a2c4db3   docker.openhands.dev/openhands/openhands:a2c4db3
```

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -229,7 +229,7 @@ class SaasSettingsStore(SettingsStore):
             else None
         )
 
-        # Custom base_url = definitely custom settings (BYOR)
+        # Custom base_url = definitely custom settings (BYOK)
         if user_base_url and user_base_url != LITE_LLM_API_URL:
             return True
 


enterprise/storage/saas_settings_store.py
```diff
@@ -213,6 +213,13 @@ async def update_settings_with_litellm_default(
             return None
         local_deploy = os.environ.get('LOCAL_DEPLOYMENT', None)
         key = LITE_LLM_API_KEY
+
+        llm_model_to_use = (
+            settings.llm_model.strip()
+            if settings.llm_model and settings.llm_model.strip()
```
xingyaoww: Maybe we should check if the user's settings is pointing to the previous settings version's default - if so, we should update it; otherwise, we should not do update it

hieptl: Hello @xingyaoww,

I have a concern regarding this change. For example, consider users on version 4 who use the OpenHands provider and have selected GPT without configuring custom settings, as this was the default in the previous settings version. If we upgrade their LLM model without prior notice, their credits may be consumed unexpectedly.

I suggest that we announce the new model in the release notes and allow users to decide when they would like to switch. I would appreciate your thoughts on this.

Thank you! 🙏

xingyaoww: If they select anything other than our default combo, e.g., OH provider + GPT-5, we should keep it as is.

But if they are using our default combo like OH provider + sonnet 4 -> we should upgrade them to OH provider + sonnet 4.5, if that make sense?

xingyaoww: @hieptl what do you think about this? ^^

hieptl: Yes, I will update the code based on your feedback. Thank you for your feedback! 🙏

hieptl: Hello @xingyaoww,

I have updated the code according to your feedback. Please refer to the videos below to review the output of the updates:

**Case 1:** Preserve custom settings.

https://github.com/user-attachments/assets/85f2473b-fe7c-4b51-a63d-196911e731b3

**Case 2:** Perform an upgrade if the current settings correspond to the previous default.

https://github.com/user-attachments/assets/559b1003-8072-4556-89e3-9c28cf791246

Thank you very much! 🙏

enterprise/storage/saas_settings_store.py
xingyaoww, line 1: @OpenHands /codereview-roasted this code feels a bit complex to read, are all of these complexity necessary or is there a way to simplify it

enterprise/storage/saas_settings_store.py
```diff
@@ -213,6 +261,24 @@ async def update_settings_with_litellm_default(
...
+
+        # Get user's old version to check if they're using the old default
+        old_user_version = None
+        with session_maker() as session:
+            user_settings = self.get_user_settings_by_keycloak_id(self.user_id, session)
+            if user_settings:
+                old_user_version = user_settings.user_version
+
+        # Check if user has custom settings
+        has_custom = self._has_custom_settings(settings, old_user_version)
```
xingyaoww: Why can't we direcly use `settings.user_version`, but rather need another db call to get `user_settings`?

hieptl: Hello @xingyaoww,

enterprise/storage/saas_settings_store.py
```diff
@@ -202,6 +203,53 @@ async def load_legacy_file_store_settings(self, github_user_id: str):
...
+            if settings.llm_model and settings.llm_model.strip()
+            else None
+        )
+        user_base_url = (
+            settings.llm_base_url.strip()
+            if settings.llm_base_url and settings.llm_base_url.strip()
+            else None
+        )
+
+        # Custom base_url = definitely custom settings (BYOR)
```
xingyaoww:
```suggestion
        # Custom base_url = definitely custom settings (BYOK)
```

PR #12153, opened 2025-12-24
refactor(frontend): migrate selected org ID to Zustand

### LINKED ISSUE #12126: Replace TanStack Query state management with Zustand for selected organization ID
> [!IMPORTANT]
> Before starting work on this issue, please ensure that you have thoroughly read the description in #12136.

## Problem

In the `ALL-2596/org-support` branch, we introduced hooks that use TanStack Query for client-side state management. This approach is not inherently an anti-pattern; however, it is inconsistent from our established state management practices in the frontend app.

### Affected Files

- [`use-selected-organization.ts`](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/context/use-selected-organization.ts); uses:
  - [`use-selected-organization-id.ts`](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/hooks/query/use-selected-organization-id.ts)
  - [`use-update-selected-organization-id.ts`](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/hooks/mutation/use-update-selected-organization-id.ts)

Problems with the current approach include unnecessary overhead and testing complexity

## Proposed Solution

Refactor to use Zustand for this client-side state, which:
- Is already used elsewhere in the codebase (e.g., [`error-message-store.ts`](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/stores/error-message-store.ts), [`optimistic-user-message-store.ts`](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/stores/optimistic-user-message-store.ts))
- Is purpose-built for client state management
- Provides simpler, more testable code
- **Has clearer semantics for what the code is doing**

## Acceptance Criteria

- New Zustand store created with proper TypeScript types
- Unit tests cover all store functionality (example in [`use-event-store.test.ts`](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/__tests__/stores/use-event-store.test.ts))
- All existing consumers work without changes to their interface
- Old TanStack Query hooks are removed
- Route revalidation behavior is preserved (currently triggered in the query hook)

### Note: TDD (optional but encouraged)

This refactor is a good candidate for TDD (Test-Driven Development):

1. **Write tests first** - Define the expected behavior of the Zustand store before implementing it:
    - Initial state should be `null`
    - `setOrgId` should update the state
    - State should persist across hook calls
    - Route revalidation should trigger on state change
2. **Run tests (they will fail)** - Confirm the tests fail with the expected errors
3. **Implement the store** - Write the minimal code to make tests pass
4. **Refactor** - Clean up while keeping tests green

smosco:
## Summary of PR

Replaced TanStack Query with Zustand for selected organization ID state management. 

- Created `selected-organization-store.ts` with localStorage persistence
- Removed old TanStack Query hooks
- Added 6 unit tests covering state updates, persistence, and singleton behavior
- Moved `getSelectedOrgIdFromStore` to store file
- Preserved route revalidation behavior for permission checks

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves #12126

## Release Notes

### DIFF
--- a/frontend/src/stores/selected-organization-store.ts
+++ b/frontend/src/stores/selected-organization-store.ts
@@ -26,5 +26,5 @@ export const useSelectedOrganizationStore = create<SelectedOrganizationStore>()(
   ),
 );
 
-export const getSelectedOrgIdFromStore = (): string | null =>
+export const getSelectedOrganizationIdFromStore = (): string | null =>
   useSelectedOrganizationStore.getState().orgId;
--- a/frontend/src/routes/manage-org.tsx
+++ b/frontend/src/routes/manage-org.tsx
@@ -12,7 +12,7 @@ import { BrandButton } from "#/components/features/settings/brand-button";
 import { useMe } from "#/hooks/query/use-me";
 import { useConfig } from "#/hooks/query/use-config";
 import { rolePermissions } from "#/utils/org/permissions";
-import { getSelectedOrgIdFromStore } from "#/stores/selected-organization-store";
+import { getSelectedOrganizationIdFromStore } from "#/stores/selected-organization-store";
 import { getMeFromQueryClient } from "#/utils/query-client-getters";
 import { queryClient } from "#/query-client-config";
 import { I18nKey } from "#/i18n/declaration";
@@ -261,7 +261,7 @@ function AddCreditsModal({ onClose }: AddCreditsModalProps) {
 }
 
 export const clientLoader = async () => {
-  const selectedOrgId = getSelectedOrgIdFromStore();
+  const selectedOrgId = getSelectedOrganizationIdFromStore();
   let me = getMeFromQueryClient(selectedOrgId);
 
   if (!me && selectedOrgId) {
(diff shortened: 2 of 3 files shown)


smosco:
Thanks for the feedback!
I added persist to preserve the selected organization across page refreshes.
I can remove it if you prefer to match the original behavior. Let me know!

amanape:
I’d prefer to remove it for now. We’re waiting on the relevant backend API to be implemented, and there is a chance it’ll introduce it’s own selected organization persistence that we will need to take into account for. If not, your solution is preferred. We’ll only know once those backend APIs are implemented, so for now let us keep it simple.

smosco:
Makes sense! I removed the persist middleware and updated the tests accordingly.

frontend/src/stores/selected-organization-store.ts
```diff
@@ -0,0 +1,30 @@
...
+  devtools(
+    (set) => ({
+      ...initialState,
+      setOrgId: (orgId) => set({ orgId }),
+    }),
+    { name: "SelectedOrganizationStore" },
+  ),
+);
+
+export const getSelectedOrgIdFromStore = (): string | null =>
```
hieptl: **suggestion:** According to your code changes, the terms “Organization” and “Org” are used inconsistently. I recommend standardizing on a single term for consistency. Thank you! 🙏

smosco: I've updated the code to use the full "Organization" term in function names.

PR #12166, opened 2025-12-26
chore(frontend): Cross-domain PostHog tracking

amanape:
## Summary of PR

Adds cross-domain tracking support for PostHog to enable user session continuity from Webflow marketing site to the app.

Source: https://posthog.com/tutorials/cross-domain-tracking

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:df27266-nikolaik   --name openhands-app-df27266   docker.openhands.dev/openhands/openhands:df27266
```

### DIFF
--- a/frontend/__tests__/components/providers/posthog-wrapper.test.tsx
+++ b/frontend/__tests__/components/providers/posthog-wrapper.test.tsx
@@ -17,15 +17,14 @@ describe("PostHogWrapper", () => {
     vi.clearAllMocks();
     // Reset URL hash
     window.location.hash = "";
-  });
-
-  it("should initialize PostHog with bootstrap IDs from URL hash", async () => {
     // Mock the config fetch
     // @ts-expect-error - partial mock
     vi.spyOn(OptionService, "getConfig").mockResolvedValue({
       POSTHOG_CLIENT_KEY: "test-posthog-key",
     });
+  });
 
+  it("should initialize PostHog with bootstrap IDs from URL hash", async () => {
     // Set up URL hash with cross-domain tracking params
     window.location.hash = "ph_distinct_id=user-123&ph_session_id=session-456";
 
@@ -52,12 +51,6 @@ describe("PostHogWrapper", () => {
   });
 
   it("should clean up URL hash after extracting bootstrap IDs", async () => {
-    // Mock the config fetch
-    // @ts-expect-error - partial mock
-    vi.spyOn(OptionService, "getConfig").mockResolvedValue({
-      POSTHOG_CLIENT_KEY: "test-posthog-key",
-    });
-
     // Set up URL hash with cross-domain tracking params
     window.location.hash = "ph_distinct_id=user-123&ph_session_id=session-456";
 


frontend/src/components/providers/posthog-wrapper.tsx
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+import { PostHogProvider } from "posthog-js/react";
+import OptionService from "#/api/option-service/option-service.api";
+import { displayErrorToast } from "#/utils/custom-toast-handlers";
+
+function getBootstrapFromHash() {
```
hieptl, line 6: **question:** Why was React Router not used to implement this logic?

frontend/src/components/providers/posthog-wrapper.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const [posthogClientKey, setPosthogClientKey] = React.useState<string | null>(
+    null,
+  );
+  const [isLoading, setIsLoading] = React.useState(true);
+  const bootstrapIds = React.useMemo(() => getBootstrapFromHash(), []);
+
+  React.useEffect(() => {
+    (async () => {
+      try {
+        const config = await OptionService.getConfig();
```
hieptl, line 35: **suggestion:** It appears that this logic should interact with the service class through a custom hook that leverages React Query, rather than invoking the service directly at this point.

frontend/src/components/providers/posthog-wrapper.tsx
```diff
@@ -0,0 +1,61 @@
...
+        setPosthogClientKey(config.POSTHOG_CLIENT_KEY);
+      } catch {
+        displayErrorToast("Error fetching PostHog client key");
```
hieptl, line 38: **suggestion:** Should we support multiple languages for this text?

frontend/src/components/providers/posthog-wrapper.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+  if (isLoading || !posthogClientKey) {
+    return children;
+  }
+
+  return (
+    <PostHogProvider
+      apiKey={posthogClientKey}
+      options={{
+        api_host: "https://us.i.posthog.com",
```
hieptl, line 53: **question:** Are we supporting this for SaaS only, or for both SaaS and OSS? If support is limited to SaaS, the logic within this component should be executed only when the application is running in SaaS mode.

frontend/src/components/providers/posthog-wrapper.tsx
amanape: https://github.com/OpenHands/OpenHands/blob/cb1d1f8a0d437b47501580b48c8db4dbe84d3c27/frontend/src/entry.client.tsx#L72-L74

The React Router context is provided by `<HydratedRouter>`, but the `PostHogWrapper` component sits outside the router. For this reason, we cannot access `useLocation` or similar React Router hooks.

frontend/src/components/providers/posthog-wrapper.tsx
amanape: https://github.com/OpenHands/OpenHands/blob/cb1d1f8a0d437b47501580b48c8db4dbe84d3c27/frontend/src/hooks/query/use-config.ts#L6

Similar to the issue I described above, `useConfig` uses a hook that depends on the React Router context, which we do not have access to here

frontend/src/components/providers/posthog-wrapper.tsx
amanape: Not necessarily, this error is really more for the developer. Wouldn’t be worth the effort to go through translating this IMO

frontend/src/components/providers/posthog-wrapper.tsx
amanape: We track user analytics for both SaaS and OSS, so we’d want to keep the logic as is. Is that what you were asking?

hieptl (approved):
Thank you! 🙏

PR #12184, opened 2025-12-29
feat: allow manual reinstallation for gitlab resolver

hieptl:
## Summary of PR

Currently, the GitLab resolver is installed automatically. The process operates as follows:
1. A workspace owner or maintainer logs into the OpenHands application.
2. The workspace repositories are persisted in the backend system.
3. A CRON job runs every minute and attempts to install the webhook across all associated resources.
4. Once the webhook is successfully installed, the installation state is recorded in the backend and is not revisited.

### Problem

Because the installation is fully automated and executed only once, the system does not attempt reinstallation. If a user chooses to delete the webhook from their repository, this action is respected. However, should the user later decide to reinstall the webhook, there is currently no automated mechanism to support this. The user must manually notify the team, which does not scale and leads to user frustration.

### Proposed Solution

Introduce a “Reinstall Webhooks” button on the Settings > Integrations page for GitLab.

### Acceptance Criteria (AC)
- Verify that the Settings > Integrations page displays a GitLab status indicator showing Connected or Not Connected.
- Verify that when the GitLab status is Connected, the “Reinstall Webhooks” button is enabled.

We can refer to the videos and image below for the output of the pull request.

**Video 1: Users delete their webhooks and then click the reinstall button.**

https://github.com/user-attachments/assets/7f19b30a-e87a-483d-a9bf-769c14754801

**Video 2: Users create new repositories and then click the reinstall button.**

https://github.com/user-attachments/assets/1fa961a4-a7b1-4395-93c5-d767c4fe6c83

**Image 1: Run the install_gitlab_webhooks.py file manually to ensure that the cron job is not broken**

image

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:e71969a-nikolaik   --name openhands-app-e71969a   docker.openhands.dev/openhands/openhands:e71969a
```

### DIFF
--- a/openhands/app_server/sandbox/docker_sandbox_service.py
+++ b/openhands/app_server/sandbox/docker_sandbox_service.py
@@ -78,6 +78,7 @@ class DockerSandboxService(SandboxService):
     health_check_path: str | None
     httpx_client: httpx.AsyncClient
     max_num_sandboxes: int
+    extra_hosts: dict[str, str] = field(default_factory=dict)
     docker_client: docker.DockerClient = field(default_factory=get_docker_client)
 
     def _find_unused_port(self) -> int:
@@ -349,6 +350,9 @@ class DockerSandboxService(SandboxService):
                 # Use Docker's tini init process to ensure proper signal handling and reaping of
                 # zombie child processes.
                 init=True,
+                # Allow agent-server containers to resolve host.docker.internal
+                # and other custom hostnames for LAN deployments
+                extra_hosts=self.extra_hosts if self.extra_hosts else None,
             )
 
             sandbox_info = await self._container_to_sandbox_info(container)
@@ -469,6 +473,15 @@ class DockerSandboxServiceInjector(SandboxServiceInjector):
             'determine whether the server is running'
         ),
     )
+    extra_hosts: dict[str, str] = Field(
+        default_factory=lambda: {'host.docker.internal': 'host-gateway'},
+        description=(
+            'Extra hostname mappings to add to agent-server containers. '
+            'This allows containers to resolve hostnames like host.docker.internal '
+            'for LAN deployments and MCP connections. '
+            'Format: {"hostname": "ip_or_gateway"}'
+        ),
+    )
 
     async def inject(
         self, state: InjectorState, request: Request | None = None
@@ -493,4 +506,5 @@ class DockerSandboxServiceInjector(SandboxServiceInjector):
                 health_check_path=self.health_check_path,
                 httpx_client=httpx_client,
                 max_num_sandboxes=self.max_num_sandboxes,
+                extra_hosts=self.extra_hosts,
             )
--- a/tests/unit/app_server/test_docker_sandbox_service.py
+++ b/tests/unit/app_server/test_docker_sandbox_service.py
@@ -444,6 +444,138 @@ class TestDockerSandboxService:
         ):
             await service.start_sandbox()
 
+    @patch('openhands.app_server.sandbox.docker_sandbox_service.base62.encodebytes')
+    @patch('os.urandom')
+    async def test_start_sandbox_with_extra_hosts(
+        self,
+        mock_urandom,
+        mock_encodebytes,
+        mock_sandbox_spec_service,
+        mock_httpx_client,
+        mock_docker_client,
+    ):
+        """Test that extra_hosts are passed to container creation."""
+        # Setup
+        mock_urandom.side_effect = [b'container_id', b'session_key']
+        mock_encodebytes.side_effect = ['test_container_id', 'test_session_key']
+
+        mock_container = MagicMock()
+        mock_container.name = 'oh-test-test_container_id'
+        mock_container.status = 'running'
+        mock_container.image.tags = ['test-image:latest']
+        mock_container.attrs = {
+            'Created': '2024-01-15T10:30:00.000000000Z',
+            'Config': {
+                'Env': ['OH_SESSION_API_KEYS_0=test_session_key', 'TEST_VAR=test_value']
+            },
+            'NetworkSettings': {'Ports': {}},
+        }
+        mock_docker_client.containers.run.return_value = mock_container
+
+        # Create service with extra_hosts
+        service_with_extra_hosts = DockerSandboxService(
+            sandbox_spec_service=mock_sandbox_spec_service,
+            container_name_prefix='oh-test-',
+            host_port=3000,
+            container_url_pattern='http://localhost:{port}',
+            mounts=[],
+            exposed_ports=[
+                ExposedPort(
+                    name=AGENT_SERVER, description='Agent server', container_port=8000
+                ),
+            ],
+            health_check_path='/health',
+            httpx_client=mock_httpx_client,
+            max_num_sandboxes=3,
+            extra_hosts={
+                'host.docker.internal': 'host-gateway',
+                'custom.host': '192.168.1.100',
+            },
+            docker_client=mock_docker_client,
+        )
+
+        with (
+            patch.object(
+                service_with_extra_hosts, '_find_unused_port', return_value=12345
+            ),
+            patch.object(
+                service_with_extra_hosts, 'pause_old_sandboxes', return_value=[]
+            ),
+        ):
+            # Execute
+            await service_with_extra_hosts.start_sandbox()
+
+        # Verify extra_hosts was passed to container creation
+        mock_docker_client.containers.run.assert_called_once()
+        call_args = mock_docker_client.containers.run.call_args
+        assert call_args[1]['extra_hosts'] == {
+            'host.docker.internal': 'host-gateway',
+            'custom.host': '192.168.1.100',
+        }
+
+    @patch('openhands.app_server.sandbox.docker_sandbox_service.base62.encodebytes')
+    @patch('os.urandom')
+    async def test_start_sandbox_without_extra_hosts(
+        self,
+        mock_urandom,
+        mock_encodebytes,
+        mock_sandbox_spec_service,
+        mock_httpx_client,
+        mock_docker_client,
+    ):
+        """Test that extra_hosts is None when not configured."""
+        # Setup
+        mock_urandom.side_effect = [b'container_id', b'session_key']
+        mock_encodebytes.side_effect = ['test_container_id', 'test_session_key']
+
+        mock_container = MagicMock()
+        mock_container.name = 'oh-test-test_container_id'
+        mock_container.status = 'running'
+        mock_container.image.tags = ['test-image:latest']
+        mock_container.attrs = {
+            'Created': '2024-01-15T10:30:00.000000000Z',
+            'Config': {
+                'Env': ['OH_SESSION_API_KEYS_0=test_session_key', 'TEST_VAR=test_value']
+            },
+            'NetworkSettings': {'Ports': {}},
+        }
+        mock_docker_client.containers.run.return_value = mock_container
+
+        # Create service without extra_hosts (empty dict)
+        service_without_extra_hosts = DockerSandboxService(
+            sandbox_spec_service=mock_sandbox_spec_service,
+            container_name_prefix='oh-test-',
+            host_port=3000,
+            container_url_pattern='http://localhost:{port}',
+            mounts=[],
+            exposed_ports=[
+                ExposedPort(
+                    name=AGENT_SERVER, description='Agent server', container_port=8000
+                ),
+            ],
+            health_check_path='/health',
+            httpx_client=mock_httpx_client,
+            max_num_sandboxes=3,
+            extra_hosts={},
+            docker_client=mock_docker_client,
+        )
+
+        with (
+            patch.object(
+                service_without_extra_hosts, '_find_unused_port', return_value=12345
+            ),
+            patch.object(
+                service_without_extra_hosts, 'pause_old_sandboxes', return_value=[]
+            ),
+        ):
+            # Execute
+            await service_without_extra_hosts.start_sandbox()
+
+        # Verify extra_hosts is None when empty dict is provided
+        mock_docker_client.containers.run.assert_called_once()
+        call_args = mock_docker_client.containers.run.call_args
+        assert call_args[1]['extra_hosts'] is None
+
     async def test_resume_sandbox_from_paused(self, service):
         """Test resuming a paused sandbox."""
         # Setup


enterprise/server/routes/integration/gitlab.py
```diff
@@ -83,3 +85,48 @@ async def gitlab_events(
...
+
+@gitlab_integration_router.post('/gitlab/reinstall-webhook')
+async def reinstall_gitlab_webhook(request: Request):
+    """Mark the GitLab webhook for the current user for reinstallation.
+
+    This endpoint marks the webhook for the user by setting webhook_exists=False,
+    which will cause the cron job to reinstall it on the next run.
+    """
+    try:
+        user_auth: SaasUserAuth = await get_user_auth(request)
```
tofarr: This could be passed in by dependency injection to simplify things. e.g.:
`async def reinstall_gitlab_webhook(user_id: str = Depends(get_user_id)):`

It makes testing easier too, and means the `if not user_id:` section is also not required.

enterprise/server/routes/integration/gitlab.py
```diff
@@ -83,3 +85,48 @@ async def gitlab_events(
...
+        user_id = await user_auth.get_user_id()
+
+        if not user_id:
+            raise HTTPException(
+                status_code=status.HTTP_401_UNAUTHORIZED,
+                detail='User not authenticated',
+            )
+
+        webhook_marked = await webhook_store.mark_webhook_for_reinstallation(user_id)
```
malhotra5: I wonder if we should force re-install rather than marking for re-installation. This is for two reasons

1. the reinstallation will happen immediately rather than waiting from the CRON job to re-pick it up (it would be queued in the last batch of values that the cron job picks up)

2. the re-installation currently requires the original user who had installed the webhook to attempt to re-install. Rather we could check whether the user has appropriate permissions for the resources and install then.



I think ideally, the FE shows a list of resources the user has permission to re-install on. The user selects the repos/groups they want to re-install the webhook (because this list can change! the original user may not have permissions anymore to re-install either)

tofarr (approved):

malhotra5:
For a gitlab org, the person who originally installed the webhook needs to be the one to re-install it seems. This is brittle because

1. we don't make it obvious who that person is (first person from the org that logged into openhands and had permissions to install the webhook)

2. we make the re-installation process dependent on that one person only (when multiple workspace owners/maintainers can exist, and anyone of them should be able to do this)

hieptl:
Hello @tofarr, @malhotra5,

frontend/__tests__/components/features/settings/gitlab-webhook-manager-state.test.tsx
```diff
@@ -0,0 +1,68 @@
...
+    // Act
+    render(<GitLabWebhookManagerState {...props} />);
+
+    // Assert
+    const messageElement = screen.getByText(
+      I18nKey.GITLAB$WEBHOOK_MANAGER_ERROR,
+    );
+    expect(messageElement).toHaveClass(customColor);
+  });
+
```
amanape:
```suggestion
```

Avoid testing styles such as color because CSS is visual and often brittle since styles can change frequently without affecting the underlying behaviour.

frontend/__tests__/components/features/settings/gitlab-webhook-manager.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+  }),
+}));
+
+// Mock the IntegrationService
+vi.mock("#/api/integration-service/integration-service.api", () => ({
+  IntegrationService: {
+    getGitLabResources: vi.fn(),
+    reinstallGitLabWebhook: vi.fn(),
+  },
+}));
```
amanape: Do we need this here?

frontend/__tests__/components/features/settings/gitlab-webhook-manager.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+    // Act
+    renderComponent();
+
+    // Assert
+    await waitFor(() => {
+      const buttons = screen.getAllByText("GITLAB$WEBHOOK_REINSTALL");
+      const reinstallButton = buttons.find((btn) => {
+        const button = btn.closest("button");
+        return button && button.hasAttribute("disabled");
+      });
```
amanape: Can we add a `data-testid=“reinstall-button”` so that we may

```ts
const reinstallButton = screen.getByTestId(“reinstall-button”)
```

I’d prefer whatever is easier to read in the code. If you require to do some additional logic with a `data-testid` that makes it less readable than your current solution, we can keep yours

frontend/__tests__/components/features/settings/gitlab-webhook-manager.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+    // Act
+    renderComponent();
+
+    // Assert
+    await waitFor(() => {
+      const buttons = screen.getAllByText("GITLAB$WEBHOOK_REINSTALL");
+      const reinstallButton = buttons.find((btn) => {
+        const button = btn.closest("button");
+        return button && !button.hasAttribute("disabled");
+      });
```
amanape: Same comment as above

frontend/__tests__/components/features/settings/gitlab-webhook-manager.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+    // Arrange
+    const user = userEvent.setup();
+    const reinstallSpy = vi
+      .spyOn(IntegrationService, "reinstallGitLabWebhook")
+      .mockResolvedValue({
+        resource_id: "1",
+        resource_type: "project",
+        success: true,
+        error: null,
+      });
```
amanape: You may not have to include a mock resolved value if the only purpose of `reinstallSpy` for this test is to make sure it was called

frontend/__tests__/components/features/settings/gitlab-webhook-manager.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+  it("should show loading state on button during reinstallation", async () => {
+    // Arrange
+    const user = userEvent.setup();
+    let resolveReinstall: (value: any) => void;
+    const reinstallPromise = new Promise((resolve) => {
+      resolveReinstall = resolve;
+    });
+
+    vi.spyOn(IntegrationService, "reinstallGitLabWebhook").mockReturnValue(
+      reinstallPromise as any,
```
amanape: Prefer to avoid using `any`. Use `// @ts-expect-error - {REASON}` if you can’t add some type

frontend/__tests__/components/features/settings/gitlab-webhook-manager.test.tsx
```diff
@@ -0,0 +1,432 @@
...
+            id: "1",
+            name: "Test Project",
+            full_path: "user/test-project",
+            type: "project",
+            webhook_installed: true,
+            webhook_uuid: "new-uuid",
+            last_synced: null,
+          },
+        ],
+      });
```
amanape: I am a little confused on why we have to mock resolved value twice here. Can you add a comment above the spy to clarify the reason if it s required?

frontend/src/api/integration-service/integration-service.api.ts
```diff
@@ -0,0 +1,36 @@
+import { openHands } from "../open-hands-axios";
+import {
+  GitLabResourcesResponse,
+  ReinstallWebhookRequest,
+  ResourceIdentifier,
+  ResourceInstallationResult,
+} from "./integration-service.types";
+
+export class IntegrationService {
```
amanape: Prefer object literals: https://github.com/OpenHands/OpenHands/blob/main/frontend/src/api/README.md#creating-a-service

amanape (approved):
Awesome test coverage! I just have a few comments:

hieptl:
Hello @amanape,

enterprise/server/routes/integration/gitlab.py
```diff
@@ -1,15 +1,28 @@
...
+from integrations.utils import GITLAB_WEBHOOK_URL
+from pydantic import BaseModel
 from server.auth.token_manager import TokenManager
+from storage.gitlab_webhook import GitlabWebhook
 from storage.gitlab_webhook_store import GitlabWebhookStore
+from sync.install_gitlab_webhooks import (
+    BreakLoopException,
+    install_webhook_on_resource,
+    verify_webhook_conditions,
+)
```
malhotra5: NIT: ideally we avoid importing from a CRON job 

Any chance we can create some kind of util file that both the cron job and gitlab router can reference?

enterprise/storage/gitlab_webhook_store.py
```diff
@@ -220,6 +220,156 @@ async def get_webhook_secret(self, webhook_uuid: str, user_id: str) -> str | Non
...
+
+                logger.info(
+                    'Marked webhook for reinstallation',
+                    extra={
+                        'user_id': user_id,
+                        'webhook_marked': rows_updated,
+                    },
+                )
+
+                return rows_updated
```
malhotra5:
```suggestion
```

I think this method is unused?

enterprise/integrations/gitlab/gitlab_service.py
```diff
@@ -527,3 +527,80 @@ async def reply_to_mr(
...
+        page = 1
+        while True:
+            try:
+                url = f'{self.BASE_URL}/groups'
+                params = {
+                    'page': str(page),
+                    'per_page': str(per_page),
+                    'min_access_level': 40,  # Maintainer or Owner
+                    'top_level_only': 'true',
+                }
```
malhotra5: I believe we can dedup this against `self.get_owned_groups`

malhotra5: The method `get_owned_groups` doesn't have pagination maybe we can modify that directly to include it

enterprise/integrations/gitlab/gitlab_service.py
```diff
@@ -527,3 +527,80 @@ async def reply_to_mr(
...
+        per_page = 100
+        while True:
+            try:
+                url = f'{self.BASE_URL}/projects'
+                params = {
+                    'page': str(page),
+                    'per_page': str(per_page),
+                    'membership': 1,
+                    'min_access_level': 40,  # Maintainer or Owner
+                }
```
malhotra5, line 586: `self.get_all_repositories` has similar logic for fetching owned projects for the user - maybe there is a way to dedup against it?

malhotra5 (changes requested):
Awesome stuff! Couple of nits regarding deduping and organizing code

enterprise/integrations/gitlab/gitlab_service.py
hieptl: Hello @malhotra5, the self.get_all_repositories function cannot be reused for the reasons outlined below. Please let me know if you have any further questions. Thank you for your feedback! 🙏 

### Why `get_all_repositories` Cannot Be Reused for `get_user_resources_with_admin_access`

### Summary

While both methods implement similar pagination patterns, they serve fundamentally different purposes with incompatible requirements that prevent code reuse.

---

### Key Differences

### 1. **Different Business Requirements**

**`get_all_repositories` (lines 171-268)**

- **Purpose**: Fetch repositories for UI display in the repository selector
- **Scope**: User's repositories with membership
- **Limit**: Hard cap of 1000 repositories for performance
- **Use case**: Interactive UI where users browse and select repositories

**`get_user_resources_with_admin_access` (lines 531-606)**

- **Purpose**: Fetch resources for webhook management and installation
- **Scope**: Only resources where user has admin access (maintainer/owner)
- **Limit**: Must fetch ALL resources (no limit) for complete webhook coverage
- **Use case**: Background operations and webhook synchronization

---

### 2. **Different API Parameters**

| Parameter          | `get_all_repositories`                        | `get_user_resources_with_admin_access`    |
| ------------------ | --------------------------------------------- | ----------------------------------------- |
| `membership`       | ✅ `1`                                        | ✅ `1`                                    |
| `order_by`         | ✅ `last_activity_at` / `created_at` / `name` | ❌ Not needed                             |
| `sort`             | ✅ `desc`                                     | ❌ Not needed                             |
| `min_access_level` | ❌ Not used                                   | ✅ `40` (Maintainer/Owner) - **Critical** |

**The `min_access_level=40` parameter is essential** for webhook management because:

- Only maintainers and owners can install webhooks
- This filters at the API level, avoiding unnecessary data transfer
- Without this filter, we'd fetch all repositories and need client-side filtering

**Real-world impact:**

```
User has membership in 500 projects:
- 50 projects as Maintainer/Owner (can install webhooks)
- 450 projects as Developer/Reporter (cannot install webhooks)

get_all_repositories:                         Returns 500 projects (90% irrelevant)
get_user_resources_with_admin_access:         Returns 50 projects (100% relevant)
```

---

### 3. **Different Loop Conditions**

```python
# get_all_repositories - enforces limit
while len(all_repos) < MAX_REPOS:
    # ... pagination logic
    all_repos = all_repos[:MAX_REPOS]  # Trim to limit

# get_user_resources_with_admin_access - no limit
while True:
    # ... pagination logic
    # No trimming - need ALL resources
```

**Why this matters**:

- **UI performance**: Limiting to 1000 repositories prevents overwhelming the UI
- **Webhook completeness**: Missing even one resource with admin access could mean webhooks aren't installed where they should be

**Real-world scenario:**

```
User has admin access to 1500 projects:

get_all_repositories:                         Returns 1000 projects (500 missing)
get_user_resources_with_admin_access:         Returns 1500 projects (complete)

Result: 500 projects would not have webhooks installed ❌
```

---

### 4. **Different Data Processing**

**`get_all_repositories`**:

```python
# Complex filtering and transformation
for repo in response:
    namespace = repo.get('namespace', {})
    kind = namespace.get('kind')
    owner_id = repo.get('owner', {}).get('id')

    # Collect user owned personal projects
    if kind == 'user' and str(owner_id) == str(user_id):
        users_personal_projects.append(repo)

# Convert to Repository objects
repositories = [Repository(...) for repo in all_repos]

# Store in database
asyncio.create_task(self.store_repository_data(...))
```

**Purpose**:

- Identifies user-owned personal projects for special handling
- Converts to typed `Repository` domain objects for the UI
- Stores repository metadata in the database

**`get_user_resources_with_admin_access`**:

```python
# Simple extend - no filtering or transformation
projects_with_admin_access.extend(response)

# Return raw API responses
return projects_with_admin_access, groups_with_admin_access
```

**Purpose**:

- Returns raw GitLab API responses for webhook installation
- No transformation needed - webhook logic uses raw data
- Separates projects and groups for different webhook handling

---

### 5. **Different Return Types**

- **`get_all_repositories`**: Returns `list[Repository]` (typed domain objects)
- **`get_user_resources_with_admin_access`**: Returns `tuple[list[dict], list[dict]]` (raw API responses)

**Why this matters**:

```python
# get_all_repositories return
Repository(
    id=str,
    full_name=str,
    stargazers_count=int,
    git_provider=ProviderType.GITLAB,
    is_public=bool
)

# get_user_resources_with_admin_access return
{
    'id': int,
    'name': str,
    'path_with_namespace': str,
    'namespace': {...},
    'permissions': {...},
    # ... all GitLab API fields
}
```

The webhook installation logic needs the full GitLab API response structure, not the simplified `Repository` object.

---

### 6. **Different Side Effects**

**`get_all_repositories`**:

```python
# Stores webhook and repository info
if store_in_background:
    asyncio.create_task(
        self.store_repository_data(users_personal_projects, repositories)
    )
else:
    await self.store_repository_data(users_personal_projects, repositories)
```

**`get_user_resources_with_admin_access`**:

```python
# No side effects - pure data fetching
return projects_with_admin_access, groups_with_admin_access
```

**Why this matters**:

- `get_all_repositories` has database side effects that are UI-specific
- `get_user_resources_with_admin_access` is a pure function for webhook management
- Mixing these concerns would violate separation of responsibilities

---

### Real-World Impact Comparison

### Scenario: User with 2000 projects

| Aspect               | `get_all_repositories`      | `get_user_resources_with_admin_access` |
| -------------------- | --------------------------- | -------------------------------------- |
| **Projects fetched** | 1000 (capped)               | 100 (admin only)                       |
| **API calls**        | 10 pages                    | 1 page                                 |
| **Data transferred** | ~10MB                       | ~1MB                                   |
| **Processing time**  | ~5 seconds                  | ~0.5 seconds                           |
| **Webhook coverage** | ❌ Incomplete (500 missing) | ✅ Complete                            |

---

### Conclusion

The pagination pattern similarity is **superficial**. The methods have:

- ❌ Different API requirements (`min_access_level` is non-negotiable)
- ❌ Different limits (1000 vs unlimited)
- ❌ Different data processing needs
- ❌ Different return types
- ❌ Different side effects
- ❌ Different performance characteristics

**Recommendation**: Keep them separate. The current implementation is clear, maintainable, and correctly serves each use case's specific requirements.

---

### Additional Context

These methods represent two distinct architectural patterns:

1. **`get_all_repositories`**: UI-focused, user-centric, performance-optimized
2. **`get_user_resources_with_admin_access`**: System-focused, permission-centric, completeness-optimized

Attempting to merge them would create a complex, hard-to-maintain abstraction that serves neither use case well.

enterprise/integrations/gitlab/gitlab_service.py
hieptl: Hello @malhotra5, the self. get_owned_groups function cannot be reused for the reasons outlined below. Please let me know if you have any further questions. Thank you for your feedback! 🙏

### Why `get_owned_groups` Cannot Be Reused for `get_user_resources_with_admin_access`

### Summary

While both methods fetch GitLab groups, they use **fundamentally different API parameters** that result in different scopes of access. Reusing `get_owned_groups` would cause incomplete webhook coverage by missing groups where users are Maintainers.

---

### Side-by-Side Comparison

### `get_owned_groups` (lines 83-98)

```python
async def get_owned_groups(self) -> list[dict]:
    """Get all groups for which the current user is the owner."""
    url = f'{self.BASE_URL}/groups'
    params = {'owned': 'true', 'per_page': 100, 'top_level_only': 'true'}

    try:
        response, headers = await self._make_request(url, params)
        return response  # Single page only
    except Exception:
        logger.warning('Error fetching owned groups', exc_info=True)
        return []  # Returns empty list on error
```

### Groups pagination in `get_user_resources_with_admin_access` (lines 574-600)

```python
# Fetch all groups where user is owner or maintainer
page = 1
while True:
    try:
        url = f'{self.BASE_URL}/groups'
        params = {
            'page': str(page),
            'per_page': str(per_page),
            'min_access_level': 40,  # Maintainer or Owner
            'top_level_only': 'true',
        }
        response, headers = await self._make_request(url, params)

        if not response:
            break

        groups_with_admin_access.extend(response)
        page += 1

        # Check if we've reached the last page
        link_header = headers.get('Link', '')
        if 'rel="next"' not in link_header:
            break

    except Exception:
        logger.warning(f'Error fetching groups on page {page}', exc_info=True)
        break  # Returns partial results
```

---

### Critical Differences

### 1. **Different API Parameters = Different Scope**

| Parameter      | `get_owned_groups`               | `get_user_resources_with_admin_access`      |
| -------------- | -------------------------------- | ------------------------------------------- |
| Access filter  | `owned=true`                     | `min_access_level=40`                       |
| **Scope**      | **Owner only** (access level 50) | **Maintainer OR Owner** (access level ≥ 40) |
| Pagination     | ❌ No                            | ✅ Yes                                      |
| Error handling | Returns `[]`                     | Returns partial results                     |

**According to GitLab API documentation:**

- **`owned=true`**: Returns only groups where the user is the **Owner** (access level 50)
- **`min_access_level=40`**: Returns groups where the user has **Maintainer** (40) **OR Owner** (50) access

---

### 2. **GitLab Access Levels Explained**

```
GitLab Group Access Levels:
┌─────────────────────────────────────────────┐
│ 50 - Owner          ✅ Can install webhooks │
│ 40 - Maintainer     ✅ Can install webhooks │
│ 30 - Developer      ❌ Cannot install       │
│ 20 - Reporter       ❌ Cannot install       │
│ 10 - Guest          ❌ Cannot install       │
└─────────────────────────────────────────────┘

owned=true:           Returns only level 50 (Owner)
min_access_level=40:  Returns levels 40-50 (Maintainer + Owner)
```

**Why this matters for webhooks:**

- Both **Maintainers and Owners** can install webhooks on GitLab groups
- Using `owned=true` would **exclude all groups where the user is a Maintainer but not an Owner**
- This would result in **incomplete webhook coverage** and missed webhook installations

---

### 3. **Real-World Impact: Missing Groups**

**Scenario: User has access to 10 groups**

```
Groups breakdown:
- 3 groups as Owner (access level 50)      ✅ Can install webhooks
- 7 groups as Maintainer (access level 40) ✅ Can install webhooks

Total groups where webhooks can be installed: 10

get_owned_groups():
  Returns: 3 groups (Owner only)
  Missing: 7 groups (70% of webhook opportunities missed) ❌

get_user_resources_with_admin_access():
  Returns: 10 groups (Maintainer + Owner)
  Missing: 0 groups (100% coverage) ✅
```

**Impact on webhook management:**

```
Organization has 100 groups:
- User is Owner of 15 groups
- User is Maintainer of 85 groups

Using get_owned_groups():
  ❌ Webhooks installed on: 15 groups
  ❌ Webhooks missing on: 85 groups
  ❌ Result: 85% of groups lack webhook integration

Using get_user_resources_with_admin_access():
  ✅ Webhooks installed on: 100 groups
  ✅ Webhooks missing on: 0 groups
  ✅ Result: Complete webhook coverage
```

---

### 4. **No Pagination in `get_owned_groups`**

**`get_owned_groups`:**

- ❌ Returns only the **first page** (max 100 groups)
- ❌ Ignores the `Link` header for pagination
- ❌ Users with >100 owned groups would have incomplete results
- ❌ No loop - single API call only

**`get_user_resources_with_admin_access`:**

- ✅ Implements **full pagination** with `while True` loop
- ✅ Checks `Link` header for `rel="next"` to fetch all pages
- ✅ Ensures **complete coverage** regardless of group count
- ✅ Continues until all groups are fetched

**Real-world impact:**

```
User owns 250 groups:

get_owned_groups():
  API calls: 1
  Groups returned: 100 (first page only)
  Groups missing: 150 (60% missing) ❌

get_user_resources_with_admin_access():
  API calls: 3 (pages 1, 2, 3)
  Groups returned: 250 (all pages)
  Groups missing: 0 (complete coverage) ✅
```

**Combined impact (access level + pagination):**

```
User has access to 300 groups:
- 50 as Owner
- 250 as Maintainer

get_owned_groups():
  Returns: 50 groups (Owner only, first page)
  Missing: 250 groups (83% missing) ❌

get_user_resources_with_admin_access():
  Returns: 300 groups (all access levels, all pages)
  Missing: 0 groups (complete) ✅
```

---

### 5. **Different Error Handling**

**`get_owned_groups`:**

```python
except Exception:
    logger.warning('Error fetching owned groups', exc_info=True)
    return []  # ❌ Loses all data on error
```

**`get_user_resources_with_admin_access`:**

```python
except Exception:
    logger.warning(f'Error fetching groups on page {page}', exc_info=True)
    break  # ✅ Returns groups fetched so far (partial results)
```

**Why this matters:**

**Scenario: API error occurs on page 5 of 10**

```
get_owned_groups():
  - Error on first (and only) API call
  - Returns: [] (empty list)
  - Result: 0 groups, complete failure ❌

get_user_resources_with_admin_access():
  - Successfully fetches pages 1-4
  - Error on page 5
  - Returns: Groups from pages 1-4 (400 groups)
  - Result: Partial coverage (400/1000 groups) ✅
```

For webhook management, **partial coverage is better than no coverage**.

---

### Context: Current Usage

### `get_owned_groups` is used only in `add_owned_projects_and_groups_to_db` (line 107)

```python
async def add_owned_projects_and_groups_to_db(self, owned_personal_projects):
    """Add owned projects and groups to the database for webhook tracking."""
    owned_groups = await self.get_owned_groups()  # ← Only usage
    webhooks = []

    # Builds webhook entries for owned groups only
    webhooks.extend(build_group_webhook_entries(owned_groups))
    webhooks.extend(build_project_webhook_entries(owned_personal_projects))

    # Store webhooks in the database
    if webhooks:
        webhook_store = GitlabWebhookStore()
        await webhook_store.store_webhooks(webhooks)
```

This appears to be a **legacy implementation** that:

- ❌ Only tracks **owned** resources (not maintainer access)
- ❌ Doesn't handle pagination (limited to 100 groups)
- ❌ Is called from `get_all_repositories` for backward compatibility
- ⚠️ Likely being phased out in favor of the new approach

The new `get_user_resources_with_admin_access` method is the **modern, complete solution** for webhook management.

---

### Potential Confusion: Similar Names, Different Purposes

The methods might seem similar because they both fetch groups, but they serve **fundamentally different use cases:**

| Method                                 | Purpose                 | Access Level      | Pagination         | Coverage   | Use Case                  |
| -------------------------------------- | ----------------------- | ----------------- | ------------------ | ---------- | ------------------------- |
| `get_owned_groups`                     | Get groups user owns    | Owner only (50)   | ❌ No (100 max)    | Incomplete | Legacy webhook tracking   |
| `get_user_resources_with_admin_access` | Get all admin resources | Maintainer+ (≥40) | ✅ Yes (unlimited) | Complete   | Modern webhook management |

---

### Why We Can't Just Add Parameters to `get_owned_groups`

**Hypothetical: What if we modified `get_owned_groups` to accept parameters?**

```python
async def get_owned_groups(
    self,
    use_min_access_level: bool = False,
    paginate: bool = False
) -> list[dict]:
    # ... complex conditional logic
```

**Problems with this approach:**

1. **Breaking change**: Existing callers expect Owner-only, single-page results
2. **Confusing API**: Method name says "owned" but behavior changes based on flags
3. **Violates Single Responsibility**: One method doing two different things
4. **Testing complexity**: Need to test all parameter combinations
5. **Maintenance burden**: Future changes affect both use cases

**Better approach**: Keep separate methods with clear, focused purposes.

---

### Recommendation

**Keep them separate.** The methods have incompatible requirements:

### Cannot Be Reconciled:

- ❌ **Different API parameters**: `owned=true` vs `min_access_level=40`
- ❌ **Different pagination needs**: Single page vs full pagination
- ❌ **Different error handling**: Fail-fast vs partial results
- ❌ **Different scopes**: Owner-only vs Maintainer+Owner
- ❌ **Different completeness guarantees**: Best-effort vs must-be-complete

---

### Conclusion

Using `get_owned_groups` for webhook management would result in:

### Data Loss:

- ❌ Missing groups where user is Maintainer (not Owner)
- ❌ Missing groups beyond the first 100
- ❌ Complete failure on any API error

### Business Impact:

- ❌ Incomplete webhook coverage across organization
- ❌ Missed webhook events from 70-90% of groups (typical case)
- ❌ Inconsistent user experience (webhooks work on some groups, not others)

### Technical Debt:

- ❌ Would perpetuate legacy approach instead of modern solution
- ❌ Would require workarounds and patches
- ❌ Would make future improvements harder

**The current implementation in `get_user_resources_with_admin_access` is correct and necessary for complete webhook coverage.**

---

### Appendix: GitLab API Reference

### `owned=true` Parameter

From GitLab API documentation:

> "Limit by groups explicitly owned by the current user"

Returns only groups where `access_level == 50` (Owner role).

### `min_access_level` Parameter

From GitLab API documentation:

> "Limit by current user minimal access level"

Returns groups where `access_level >= specified_value`.

For `min_access_level=40`:

- Includes Maintainer (40)
- Includes Owner (50)
- Excludes Developer (30) and below

Both Maintainer and Owner roles have permission to manage webhooks in GitLab.

hieptl:
Hello @malhotra5,

enterprise/integrations/gitlab/gitlab_service.py
malhotra5: `get_owned_groups` is also used during the GitLab webhook installations (first installation rather than re-installation)

So it would make sense to modify this method directly so that 

1. we are paginating
2. we are also including maintainers (we should rename the method to accurately reflect that maintainers rights are also satisfactory)


This should help dedup code?

enterprise/integrations/gitlab/gitlab_service.py
malhotra5: Oh I wasn't suggesting we use `get_all_repositories` as a direct replacement

There is some logic inside the method used for getting personal project

It also seems like this method doesn't account for the case where the user is an owner, this should also be fixed. Maybe we can extract that logic into a shared method?

The shared logic I'm referencing to is the code snippet below

```python
                for repo in response:
                    namespace = repo.get('namespace', {})
                    kind = namespace.get('kind')
                    owner_id = repo.get('owner', {}).get('id')

                    # Collect user owned personal projects
                    if kind == 'user' and str(owner_id) == str(user_id):
                        users_personal_projects.append(repo)
```

enterprise/sync/install_gitlab_webhooks.py
```diff
@@ -197,51 +172,15 @@ async def create_new_webhook(
...
-
-        logger.info(
-            'Creating new webhook',
-            extra={
-                'webhook_id': webhook_id,
-                'status': status,
-                'resource_id': resource_id,
-                'resource_type': resource_type,
-            },
+            webhook_store=webhook_store,
```
malhotra5, line 109:
```suggestion
```

NIT: do we need to log the entire webhook store?

enterprise/sync/install_gitlab_webhooks.py
```diff
@@ -162,23 +152,8 @@ async def verify_conditions_are_met(
         webhook_store: GitlabWebhookStore,
         webhook: GitlabWebhook,
     ):
-        await self.check_if_resource_exists(
```
malhotra5, line 165: NIT: if the method `check_if_resource_exists` is no longer used we can remove it from the class

same for `check_if_user_has_admin_acccess_to_resource`

enterprise/integrations/gitlab/webhook_installation.py
```diff
@@ -0,0 +1,199 @@
...
+    if status == WebhookStatus.RATE_LIMITED:
+        raise BreakLoopException()
+    if does_webhook_exist_on_resource != webhook.webhook_exists:
+        await webhook_store.update_webhook(
+            webhook, {'webhook_exists': does_webhook_exist_on_resource}
+        )
+
+    if does_webhook_exist_on_resource:
+        raise BreakLoopException()
+
```
malhotra5, line 131: Will re-installation work if the webhook already exists? Seems like we throw an error

hieptl: Hello @malhotra5, If a webhook already exists in the user’s GitLab group or project, the Reinstall button will be disabled, and reinstallation will not be allowed and the API will return an error. To reinstall, users must first delete the existing webhooks through the GitLab user interface. Thank you! 🙏

enterprise/sync/install_gitlab_webhooks.py
hieptl: Hello @malhotra5, It appears that webhook_store is not being logged. Thank you! 🙏

enterprise/sync/install_gitlab_webhooks.py
hieptl: Hello @malhotra5,

enterprise/integrations/gitlab/gitlab_service.py
hieptl: Hello @malhotra5, if we extract the given logic into a shared method method, it would only be useful if we needed to:

- Distinguish personal projects from organization projects in the results
- Apply special handling to personal projects

Since get_user_resources_with_admin_access just needs to return all resources with admin access (including personal projects), and the API already handles this with min_access_level=40, no additional filtering logic is needed.

Your concern about personal projects is addressed by the fact that personal project owners automatically have access level 50 (Owner), which satisfies the min_access_level=40 filter.

**Different Data Processing**

**`get_all_repositories`**:

```python
# Complex filtering and transformation
for repo in response:
    namespace = repo.get('namespace', {})
    kind = namespace.get('kind')
    owner_id = repo.get('owner', {}).get('id')

    # Collect user owned personal projects
    if kind == 'user' and str(owner_id) == str(user_id):
        users_personal_projects.append(repo)
```

**Purpose**:
- Identifies user-owned personal projects for special handling
- Converts to typed `Repository` domain objects for the UI
- Stores repository metadata in the database

**`get_user_resources_with_admin_access`**:

```python
# Simple extend - no filtering or transformation
projects_with_admin_access.extend(response)

# Return raw API responses
return projects_with_admin_access, groups_with_admin_access
```

**Purpose**:
- Returns raw GitLab API responses for webhook installation
- No transformation needed - webhook logic uses raw data
- Separates projects and groups for different webhook handling

If you still wish to reuse the existing logic for any reason, could you please push a commit to this branch so that I can review and learn from it?

Thank you! 🙏

malhotra5: Got it, thanks for walking me through this!

enterprise/integrations/gitlab/webhook_installation.py
malhotra5: PR looks good! Last question: does the FE has information regarding deleting the webhook from GitLab UI first in order to re-install?

hieptl: Currently, the frontend does not receive real-time updates when a webhook is deleted from the GitLab UI. However, after refreshing the Integrations page—or navigating to it after the deletion—the webhook status for the corresponding groups or projects is updated from “Installed” to “Not installed.” Thank you! 🙏

malhotra5: Oh I meant some kind of informational tooltip if you hover on the disabled re-install button

If folks are there to re-installed but unable to, it might not be obvious to them to delete any exist webhooks from gitlab UI first

hieptl: I will add helper text to clarify this behavior. Thank you for your feedback! 🙏

jpelletier1:
@hieptl amazing, nice work!

PR #12208, opened 2025-12-30
hotfix(frontend): validate git changes response is array before mapping

amanape:
## Summary of PR

When a V1 conversation's runtime becomes unavailable (e.g., due to idle timeout), requests to `/api/git/changes/` return the frontend's HTML page instead of JSON (served by a catch-all route with 200 status). This caused `data.map is not a function` error since the response was a string, not an array.

image

Changes:
- Added `Array.isArray(data)` validation before calling `.map()` in `V1GitService.getGitChanges()`
- Throws a descriptive error when response is invalid, allowing proper error handling by the UI

image

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:8743269-nikolaik   --name openhands-app-8743269   docker.openhands.dev/openhands/openhands:8743269
```

### DIFF
--- a/frontend/__tests__/api/v1-git-service.test.ts
+++ b/frontend/__tests__/api/v1-git-service.test.ts
@@ -1,0 +1,18 @@
+import { test, expect, vi } from "vitest";
+import axios from "axios";
+import V1GitService from "../../src/api/git-service/v1-git-service.api";
+
+vi.mock("axios");
+
+test("getGitChanges throws when response is not an array (dead runtime returns HTML)", async () => {
+  const htmlResponse = "<!DOCTYPE html><html>...</html>";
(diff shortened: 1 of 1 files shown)


hieptl (approved):
Could you write tests to cover your changes? Thank you! 🙏

amanape:
Done, I have also confirmed it is a valid test by removing my fix to confirm it fails, then adding it again and confirms it passes

PR #12214, opened 2025-12-30
fix(frontend): add loading state to git changes refresh button

### LINKED ISSUE #12209: Add loading feedback to git changes refresh button
## Problem

When clicking the refresh button in the Changes tab, there is no visual feedback indicating that the git changes API is being called. Users have no way to know if their click was registered or if data is being fetched, which can lead to confusion and repeated clicks.

## Expected Behavior

The refresh button should provide clear visual feedback when a refresh is in progress. The button should appear in a loading state and prevent additional clicks until the current fetch completes.

## Acceptance Criteria

- The refresh icon should spin while the git changes API request is in progress.
- The button should be disabled during loading to prevent multiple simultaneous requests.
- The button should have a visually distinct appearance when disabled, such as reduced opacity.
- Consider using `isFetching` from TanStack Query rather than `isLoading`, since `isLoading` only applies to the initial load while `isFetching` covers refetch scenarios. (see https://stackoverflow.com/questions/75401477/reasons-for-using-isloading-or-isfetching-in-react-query)

## Relevant Files

- [frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx](https://github.com/All-Hands-AI/OpenHands/blob/main/frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx) contains the refresh button component that needs the loading state.
- [frontend/src/hooks/query/use-unified-get-git-changes.ts](https://github.com/All-Hands-AI/OpenHands/blob/main/frontend/src/hooks/query/use-unified-get-git-changes.ts) contains the hook that fetches git changes and will need to expose isFetching.

psbuilds:
## Summary of PR

Adds visual loading feedback to the refresh button in the Changes tab.  
While git changes are being fetched, the refresh icon spins and the button is disabled to prevent multiple simultaneous requests.  
This uses `isFetching` from TanStack Query to correctly handle refetch scenarios.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #12209

## Release Notes

### DIFF
--- a/frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
+++ b/frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
@@ -25,12 +25,12 @@ export function ConversationTabTitle({
           type="button"
           onClick={handleRefresh}
           disabled={isFetching}
-          className={[
+          className={cn(
             "flex w-[26px] py-1 items-center justify-center rounded-[7px] transition",
             isFetching
               ? "opacity-50 cursor-not-allowed"
               : "hover:bg-[#474A54] cursor-pointer",
-          ].join(" ")}
+          )}
         >
           <RefreshIcon
             width={12.75}


frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
```diff
@@ -10,10 +10,12 @@ export function ConversationTabTitle({
   title,
   conversationKey,
 }: ConversationTabTitleProps) {
-  const { refetch } = useUnifiedGetGitChanges();
+  const { refetch, isFetching } = useUnifiedGetGitChanges(); //  updated
```
hieptl: **question:** What is the purpose of the comment "// updated"? Thank you! 🙏

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
```diff
@@ -10,10 +10,12 @@ export function ConversationTabTitle({
...
 
   const handleRefresh = () => {
-    refetch();
+    if (!isFetching) {
```
hieptl: **question:** Why not disable the refresh button instead of using this approach? Thank you! 🙏

hieptl (commented):
Hello @psbuilds, thank you for creating this pull request. Could you please add tests to cover your changes? 🙏

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
psbuilds: The refresh button is disabled while the request is in progress, but I combined that with a visible loading state intentionally.
Using disabled={isFetching} prevents multiple simultaneous requests, and the spinning icon with reduced opacity provides clear feedback that the refresh is actively happening. Without any visual indication, a disabled button can sometimes feel unresponsive, whereas the loading state helps reassure users that their action was registered.
I used isFetching specifically so this behavior applies to refetches as well, not just the initial load.
Happy to adjust the UX if you have a different preference here.

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
psbuilds: It was only meant as a temporary marker during development to highlight the changed line. I will remove it.

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
hieptl: We should avoid the following approach:

```
if (!isFetching) {
  refetch()
}
```

Instead, disabled={isFetching} should be added to the button.

Additionally, I have run and verified your code locally. When clicking the refresh button, I observed that the spinning animation does not appear, the button’s opacity is not reduced, and the cursor-not-allowed style is not applied.

Thank you! 🙏

psbuilds: I'll Fix this. Thank you so much for the feedback!!

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.test.tsx
```diff
@@ -0,0 +1,55 @@
...
+    );
+
+    render(<ConversationTabTitle title="Changes" conversationKey="editor" />);
+
+    const button = screen.getByRole("button");
+    expect(button).not.toHaveClass("opacity-50");
+    expect(button).not.toHaveClass("cursor-not-allowed");
+
+    const icon = button.querySelector("svg");
+    expect(icon).not.toHaveClass("animate-spin");
```
hieptl, line 53: **suggestion:** We should avoid testing visual aspects, as CSS is primarily visual and can be fragile—styles may change frequently without impacting the underlying functionality. Instead, focus on testing functional behavior, such as verifying that the refresh button is disabled while the API call is in progress. Thank you! 🙏

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
```diff
@@ -19,13 +19,25 @@ export function ConversationTabTitle({
...
           type="button"
-          className="flex w-[26px] py-1 justify-center items-center gap-[10px] rounded-[7px] hover:bg-[#474A54] cursor-pointer"
           onClick={handleRefresh}
+          disabled={isFetching}
+          className={[
+            "flex w-[26px] py-1 items-center justify-center rounded-[7px] transition",
+            isFetching
+              ? "opacity-50 cursor-not-allowed"
+              : "hover:bg-[#474A54] cursor-pointer",
+          ].join(" ")}
```
hieptl: **suggestion:** We should use cn to combine classes. You can refer to this file as an example. Thank you! 🙏 

https://github.com/OpenHands/OpenHands/blob/5bd8695ab894b1b618ded27a5b60dbf7b393d031/frontend/src/components/features/chat/chat-add-file-button.tsx#L4

hieptl:
Hello @psbuilds,

I have verified your code locally and have the following question:

**question:** When we click the refresh button, the opacity is reduced, the cursor changes to not-allowed, and the refresh icon spins.

However, whenever I refresh the conversation page, the refresh button automatically performs the spin animation. I am unsure if this is the intended behavior. @amanape, do you agree with this behavior?

Please refer to the video below for more details.

https://github.com/user-attachments/assets/874c4708-cc48-41f2-9c19-ceb628a75b39

Thank you, everyone! 🙏

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.test.tsx
amanape: As @hieptl suggested, avoid testing CSS; use `.toBeDisabled()` matchers instead.

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-title.tsx
amanape:
```suggestion
          className={cn(
            "flex w-[26px] py-1 items-center justify-center rounded-[7px] transition",
            isFetching
              ? "opacity-50 cursor-not-allowed"
              : "hover:bg-[#474A54] cursor-pointer",
          )}
```

You’ll also need to import `cn`:

```ts
import { cn } from "#/utils/utils”;
```

amanape:
> However, whenever I refresh the conversation page, the refresh button automatically performs the spin animation. I am unsure if this is the intended behavior. @amanape, do you agree with this behavior?

I suppose this is because the endpoint is fetched when it is mounted setting `isFetching` to `true` by default. I don’t think it is a problem, and might be good UI too. But if we’d want to disable this behavior, we can probably set `disabled: true` on the query hook so it only triggers on an explicit call via `refetch`.

For now though, let’s keep it

mamoodi:
Hi there. Many thanks for your contribution. I'm going to set this PR back into draft and when you have addressed the comments, please set it back to ready for review.

PR #12235, opened 2026-01-01
Clarify OpenHands naming (replace “OSS” wording in docs and backend)

enyst:
This PR is a small terminology cleanup to help keep naming consistent:

- In documentation, replace uses of “OSS” when it was being used as the *project name* with **OpenHands**.
- In backend/server comments, clarify places where we mean **OpenHands (OSS mode)**.
- In the legacy V0 server code, introduce `AppMode.OPENHANDS = "oss"` as the preferred enum member (keeping `AppMode.OSS = "oss"` alias for now).
- Add a backend-only pre-commit/CI check (`warn-appmode-oss`) to discourage introducing new `AppMode.OSS` usages.

No behavior change is intended: the wire/config value remains `"oss"` for compatibility with existing deployments.

Thank you for taking a look — this is just meant to make the project name clearer and reduce confusion going forward.

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/3d0b515391b74d14aa5df0ae62a1822f)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:997c42d-nikolaik   --name openhands-app-997c42d   docker.openhands.dev/openhands/openhands:997c42d
```

### DIFF
--- a/openhands/app_server/app_conversation/app_conversation_router.py
+++ b/openhands/app_server/app_conversation/app_conversation_router.py
@@ -1,4 +1,4 @@
-"""Sandboxed Conversation router for OpenHands Server."""
+"""Sandboxed Conversation router for OpenHands App Server."""
 
 import asyncio
 import logging
--- a/openhands/app_server/event/event_router.py
+++ b/openhands/app_server/event/event_router.py
@@ -1,4 +1,4 @@
-"""Event router for OpenHands Server."""
+"""Event router for OpenHands App Server."""
 
 from datetime import datetime
 from typing import Annotated
--- a/openhands/app_server/event_callback/webhook_router.py
+++ b/openhands/app_server/event_callback/webhook_router.py
@@ -1,4 +1,4 @@
-"""Event Callback router for OpenHands Server."""
+"""Event Callback router for OpenHands App Server."""
 
 import asyncio
 import importlib
--- a/openhands/app_server/sandbox/sandbox_router.py
+++ b/openhands/app_server/sandbox/sandbox_router.py
@@ -1,4 +1,4 @@
-"""Runtime Containers router for OpenHands Server."""
+"""Runtime Containers router for OpenHands App Server."""
 
 from typing import Annotated
 
--- a/openhands/app_server/sandbox/sandbox_spec_router.py
+++ b/openhands/app_server/sandbox/sandbox_spec_router.py
@@ -1,4 +1,4 @@
-"""Runtime Images router for OpenHands Server."""
+"""Runtime Images router for OpenHands App Server."""
 
 from typing import Annotated
 
--- a/openhands/app_server/services/db_session_injector.py
+++ b/openhands/app_server/services/db_session_injector.py
@@ -1,4 +1,4 @@
-"""Database configuration and session management for OpenHands Server."""
+"""Database configuration and session management for OpenHands App Server."""
 
 import asyncio
 import logging
--- a/openhands/app_server/user/user_router.py
+++ b/openhands/app_server/user/user_router.py
@@ -1,4 +1,4 @@
-"""User router for OpenHands Server. For the moment, this simply implements the /me endpoint."""
+"""User router for OpenHands App Server. For the moment, this simply implements the /me endpoint."""
 
 from fastapi import APIRouter, HTTPException, status
 


tofarr (commented):
Do we also need to update the AppMode in the frontend? Also, would it make sense to say specifically "OpenHands App Server" instead of "OpenHands Server"? (To distinguish it from the OpenHands Agent Server)

enyst:
@openhands Lets address the review comment on this PR.

We want to do the most, but please make sure that you investigate a bit and give me 2-3 alternatives for compatibility on the wire, and with cached frontends. Implement the "OpenHands App Server" as requested, that doesn't need compat I think.

openhands/server/config/server_config.py
```diff
@@ -9,13 +9,15 @@
...
+from openhands.server.types import ServerConfigInterface, parse_app_mode
 from openhands.utils.import_utils import get_impl
 
 
 class ServerConfig(ServerConfigInterface):
     config_cls = os.environ.get('OPENHANDS_CONFIG_CLS', None)
-    app_mode = AppMode.OSS
+    # Allow operators to use OPENHANDS_APP_MODE=openhands while preserving wire compatibility.
+    # If not set, default to OpenHands mode.
+    app_mode = parse_app_mode(os.getenv('OPENHANDS_APP_MODE', 'oss'))
```
enyst: We could give up this one. TBH I'm not terribly fond of new env vars, I think maybe we have too many already...

enyst:
@tofarr Something like this is more like what you had in mind?

We could go as far as we want, I'm just thinking that _maybe_ the exact strings sent over ("oss" / "saas") are less important and adding compatibility code is maybe better for a follow-up if at all.

amanape (commented):
Sorry I’m late to the review here. I feel like this will make for more confusing DX…

> In documentation, replace uses of “OSS” when it was being used as the project name with OpenHands.

Not all occurrences of OSS here are used in place of the title “OpenHands”. If anything, it looks like this PR is doing the opposite, replacing all occurrences of the _app mode_ with the project name. For example:

frontend/__tests__/routes/git-settings.test.tsx
```diff
@@ -15,7 +15,7 @@ import * as ToastHandlers from "#/utils/custom-toast-handlers";
 import { SecretsService } from "#/api/secrets-service";
 import { integrationService } from "#/api/integration-service/integration-service.api";
 
-const VALID_OSS_CONFIG: GetConfigResponse = {
+const VALID_OPENHANDS_CONFIG: GetConfigResponse = {
```
amanape: This is supposed to represent the mock response values of the app locally and not on the cloud, now it looks like it represents the response value for ALL modes

frontend/src/components/features/sidebar/user-actions.tsx
```diff
@@ -36,11 +36,11 @@ export function UserActions({ onLogout, user, isLoading }: UserActionsProps) {
     closeAccountMenu();
   };
 
-  const isOSS = config?.APP_MODE === "oss";
+  const isOpenHandsMode = config?.APP_MODE === "oss";
```
amanape: This looks more confusing now since the constant was representing the app mode

frontend/src/hooks/use-app-title.ts
```diff
@@ -2,7 +2,7 @@ import { useParams } from "react-router";
 import { useConfig } from "#/hooks/query/use-config";
 import { useUserConversation } from "#/hooks/query/use-user-conversation";
 
-const APP_TITLE_OSS = "OpenHands";
+const APP_TITLE_OPENHANDS = "OpenHands";
 const APP_TITLE_SAAS = "OpenHands Cloud";
```
amanape: This is a little more confusing now.

amanape:
Maybe it is not so clear for me. Does this PR propose updating the “glossary” so that “OpenHands” refers to the app mode when running locally instead of OSS / open-source / locally?

frontend/__tests__/routes/git-settings.test.tsx
enyst: Happy to restore the original, which didn't have this one.

I'm not sure I see it that way though. The AH Saas is an application built on OpenHands. AH built on OpenHands like others built on top of OpenHands. Some university built another application, some company Daytona built another in the past. They are applications built on OpenHands. Some are multiuser, some not.

It's not really a "mode". 😅

amanape: > The AH Saas is an application built on OpenHands.

Oh OK, when you put it that way yeah your changes make sense. Well then in that case my only concern is that new and existing contributors might see it as “modes”, and might get confused

enyst:
@openhands lets revert the last two commits, that is, reset to [997c42d](https://github.com/OpenHands/OpenHands/pull/12235/commits/997c42d5cb6ed02a4a0e2308161d5232757e2996)

frontend/__tests__/routes/git-settings.test.tsx
enyst: Yes, I hear you, and on the other hand, I also hear that the two apps here so entangled with each other.

I withdrew my latest updates and this comment in this topic. Let’s merge this for now. Onwards! 😅

PR #12250, opened 2026-01-02
fix: Handle LiteLLM v1.80+ 404 response for new users

neubig:
## Description

This PR fixes a breaking change introduced by the LiteLLM proxy upgrade from v1.79.x to v1.80.x.

### Problem

The LiteLLM API behavior changed between versions:
- **v1.79.x and earlier**: `GET /user/info` always succeeds, returning empty `user_info` for non-existent users
- **v1.80.x and later**: `GET /user/info` returns **404** for non-existent users

This caused new user signup to fail because the code in `saas_settings_store.py` called `response.raise_for_status()` without handling the 404 case.

### Evidence from Datadog

- LiteLLM was upgraded from `v1.79.3` to `v1.80.8` at `2026-01-02T20:32:51Z`
- First `user_info` 404 errors appeared at `2026-01-02T20:42:24Z` (10 minutes later)
- No `user_info` 404 errors existed in the previous 30 days

### Solution

Handle both API behaviors to maintain compatibility:
- **200 OK**: User exists, extract their info (both versions)
- **404 Not Found**: User doesn't exist yet (new user), use empty dict (v1.80+ behavior)
- **Other errors**: Raise as before (unexpected server issues)

### Changes

1. **`enterprise/storage/saas_settings_store.py`**: Added 404 handling with detailed comments explaining the version differences
2. **`enterprise/tests/unit/test_saas_settings_store.py`**: Added test case for the 404 scenario

## Testing

- [x] Added unit test `test_update_settings_with_litellm_default_handles_404_for_new_user`
- [x] Test passes locally
- [x] Pre-commit hooks pass
- [x] HUMAN: This has been tested with a preview deployment, and I can confirm that it failed before and works after on staging.

## Related Issues

This fixes the login errors observed on staging deployments after the LiteLLM proxy upgrade.

@neubig can click here to [continue refining the PR](https://app.all-hands.dev/conversations/949d1cd676134f49ba667f0ebde23c59)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:0afc83f-nikolaik   --name openhands-app-0afc83f   docker.openhands.dev/openhands/openhands:0afc83f
```

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -287,15 +287,7 @@ class SaasSettingsStore(SettingsStore):
             ) as client:
                 # Get the previous max budget to prevent accidental loss.
                 #
-                # LiteLLM API behavior changed between versions:
-                # - v1.79.x and earlier: GET /user/info always succeeds, returning
-                #   empty user_info for non-existent users
-                # - v1.80.x and later: GET /user/info returns 404 for non-existent users
-                #
-                # We handle both cases to maintain compatibility during the transition:
-                # - 200 OK: User exists, extract their info
-                # - 404 Not Found: User doesn't exist yet (new user), use empty dict
-                # - Other errors: Raise as before (unexpected server issues)
+                # LiteLLM v1.80+ returns 404 for non-existent users (previously returned empty user_info)
                 response = await client.get(
                     f'{LITE_LLM_API_URL}/user/info?user_id={self.user_id}'
                 )
--- a/enterprise/tests/unit/test_saas_settings_store.py
+++ b/enterprise/tests/unit/test_saas_settings_store.py
@@ -1,5 +1,6 @@
 from unittest.mock import AsyncMock, MagicMock, patch
 
+import httpx
 import pytest
 from pydantic import SecretStr
 from server.constants import (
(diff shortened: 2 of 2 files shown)


malhotra5:

malhotra5:
## 🔥 Roasted Code Review

🟢 **Good taste** - This is a pragmatic fix for a real production problem. Simple, focused, and doesn't over-engineer.

---

### **[CRITICAL ISSUES]** (None)

No fundamental design violations here. The fix is appropriately minimal.

---

### **[IMPROVEMENT OPPORTUNITIES]**

**[enterprise/storage/saas_settings_store.py, Lines 288-310]** 📝 **Comment Verbosity**: You've written an 8-line essay explaining a simple 404 check. Linus would say: "If the code needs that much explanation, maybe the code is wrong. If the code is right, the comment is too long."

The version history is useful context, but consider condensing to:
```python
# LiteLLM v1.80+ returns 404 for non-existent users (previously returned empty user_info)
```

**[enterprise/tests/unit/test_saas_settings_store.py, Lines 316-335]** ⚠️ **Fragile Existing Test**: The `test_update_settings_with_litellm_default_error` test doesn't set `status_code` on the mock GET response. Your new code checks `response.status_code == 404` first. This test passes by accident because `AsyncMock().status_code` returns another `AsyncMock` which is truthy but not equal to `404`, so it falls through to the else branch.

This isn't broken, but it's fragile. The test should explicitly set `status_code = 200` to be clear about what it's testing.

**[enterprise/tests/unit/test_saas_settings_store.py, Lines 350-384]** 🔍 **Test Coverage Gap**: The new test verifies the happy path (404 → create user succeeds), but doesn't verify:
1. That `raise_for_status()` is NOT called when status is 404
2. That `raise_for_status()` IS called for other error codes (e.g., 500)

Consider adding a test for a 500 response to ensure the error handling path still works.

---

### **[STYLE NOTES]**

**[enterprise/storage/saas_settings_store.py, Line 302]** 🎯 **Type Annotation**: `user_info: dict` is fine, but `user_info: dict = {}` would eliminate the need for the separate assignment in the if branch. Though I understand you're being explicit about the two code paths.

---

### **VERDICT:**
✅ **Worth merging** - Core logic is sound. The fix is correct, minimal, and solves a real production issue. The suggestions above are polish, not blockers.

### **KEY INSIGHT:**
This is exactly how you handle API version compatibility - check for the specific new behavior (404), fall through to existing behavior otherwise. No over-abstraction, no version detection magic, just handle what you get. Good taste. 👍

enterprise/storage/saas_settings_store.py
```diff
@@ -285,14 +285,29 @@ async def update_settings_with_litellm_default(
...
+                #
+                # LiteLLM API behavior changed between versions:
+                # - v1.79.x and earlier: GET /user/info always succeeds, returning
+                #   empty user_info for non-existent users
+                # - v1.80.x and later: GET /user/info returns 404 for non-existent users
+                #
+                # We handle both cases to maintain compatibility during the transition:
+                # - 200 OK: User exists, extract their info
+                # - 404 Not Found: User doesn't exist yet (new user), use empty dict
+                # - Other errors: Raise as before (unexpected server issues)
```
malhotra5:
```suggestion
                # LiteLLM v1.80+ returns 404 for non-existent users (previously returned empty user_info)
```

enterprise/tests/unit/test_saas_settings_store.py
```diff
@@ -335,6 +335,55 @@ async def test_update_settings_with_litellm_default_error(settings_store):
             assert settings is None
 
 
+@pytest.mark.asyncio
+async def test_update_settings_with_litellm_default_handles_404_for_new_user(
```
malhotra5: NIT: maybe we parameterize this test with different response codes and assert the expected downstream behavior

malhotra5 (approved):
LGTM, thanks!

PR #12236, opened 2026-01-01
Add extra_hosts support to agent-server containers

tofarr:
## Summary of PR

This PR adds support for passing `extra_hosts` to agent-server containers created by `DockerSandboxService`. This allows containers to resolve hostnames like `host.docker.internal` for LAN deployments and MCP connections.

**Problem**: When deploying OpenHands as a LAN-accessible server (Docker runtime on Ubuntu), agent-server containers couldn't resolve `host.docker.internal`, causing MCP connection failures and conversations stuck at `WAITING_FOR_SANDBOX`.

**Root Cause**: When `DockerSandboxService` creates agent-server containers, it didn't pass the `extra_hosts` parameter, so agent-servers couldn't resolve `host.docker.internal` to reach the main app's MCP endpoint.

**Solution**: 
- Add `extra_hosts` field to `DockerSandboxService` dataclass
- Add `extra_hosts` field to `DockerSandboxServiceInjector` with default `{'host.docker.internal': 'host-gateway'}`
- Pass `extra_hosts` to `docker_client.containers.run()` in `start_sandbox()`
- Add unit tests for the new functionality

## Change Type

- [x] Bug fix
- [x] New feature

## Checklist

- [x] I have read and reviewed the code and I understand what the code is doing.
- [x] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

This addresses the networking issue reported by Shane in Slack where agent-server containers couldn't resolve `host.docker.internal` in LAN deployments.

## Release Notes

- [x] Include this change in the Release Notes.

Agent-server containers now include `extra_hosts` configuration by default, allowing them to resolve `host.docker.internal` for LAN deployments and MCP connections. This fixes issues where conversations would get stuck at `WAITING_FOR_SANDBOX` due to hostname resolution failures.

@tofarr can click here to [continue refining the PR](https://app.all-hands.dev/conversations/9b11e9c6156e4a6a8cb87e6a46a196e4)

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:8a87428-nikolaik   --name openhands-app-8a87428   docker.openhands.dev/openhands/openhands:8a87428
```

PR #12255, opened 2026-01-04
feat: Add configurable sandbox host_port and container_url_pattern for remote access

shanemort1982:
## Summary

This PR adds environment variable configuration for the Docker sandbox service to support non-default port deployments.

## Problem

When running OpenHands on a non-default port (e.g., port 4000 instead of 3000), **even locally**, the sandbox functionality breaks because:

1. `host_port` defaults to `3000`, so webhook callbacks from agent-server containers go to the wrong port
2. `container_url_pattern` defaults to `http://localhost:{port}`, which works locally but fails for remote access

### Affected Scenarios

- **Local, different port**: Running on `localhost:4000` - webhooks fail (call port 3000)
- **Remote access**: Accessing from another machine - websocket URLs point to `localhost` which doesn't resolve
- **LAN deployments**: Any network setup where the server IP is needed

## Solution

Added two new environment variables:

- **`SANDBOX_HOST_PORT`**: The port OpenHands is running on (for webhook callbacks)
- **`SANDBOX_CONTAINER_URL_PATTERN`**: URL pattern for exposed sandbox ports

### Running locally on port 4000:
```bash
docker run -d \
  -e SANDBOX_HOST_PORT=4000 \
  -p 4000:3000 \
  ghcr.io/openhands/openhands:main
```

### Running for remote/LAN access:
```bash
docker run -d \
  -e SANDBOX_HOST_PORT=4000 \
  -e SANDBOX_CONTAINER_URL_PATTERN='http://192.168.1.100:{port}' \
  -p 4000:3000 \
  ghcr.io/openhands/openhands:main
```

## Changes

1. **`openhands/app_server/sandbox/docker_sandbox_service.py`**: Added `Field` descriptions for `host_port` and `container_url_pattern`
2. **`openhands/app_server/config.py`**: Added support for `SANDBOX_HOST_PORT` and `SANDBOX_CONTAINER_URL_PATTERN` environment variables
3. **`tests/unit/app_server/test_docker_sandbox_service.py`**: Added tests for the new configuration options

## Related

Follow-up to PR #12236 which fixed `extra_hosts` propagation for container-to-host communication.

### DIFF
--- a/openhands/app_server/config.py
+++ b/openhands/app_server/config.py
@@ -179,7 +179,9 @@ def config_from_env() -> AppServerConfig:
             # Support legacy environment variables for Docker sandbox configuration
             docker_sandbox_kwargs: dict = {}
             if os.getenv('SANDBOX_HOST_PORT'):
-                docker_sandbox_kwargs['host_port'] = int(os.environ['SANDBOX_HOST_PORT'])
+                docker_sandbox_kwargs['host_port'] = int(
+                    os.environ['SANDBOX_HOST_PORT']
+                )
             if os.getenv('SANDBOX_CONTAINER_URL_PATTERN'):
                 docker_sandbox_kwargs['container_url_pattern'] = os.environ[
                     'SANDBOX_CONTAINER_URL_PATTERN'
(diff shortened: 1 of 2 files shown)


tofarr (approved):
Good Job! 🍰

PR #12263, opened 2026-01-05
feat(backend): develop post /api/organizations api (org project)

hieptl:
## Summary of PR

**Description:**

We need to implement the POST /api/organizations endpoint to enable the creation of a new organization. For details on the required organization fields, please refer to the parent ticket associated with this task.

**Acceptance Criteria:**
- The API successfully creates a new organization when called with valid input.

https://github.com/user-attachments/assets/bbd07b3d-ec52-4009-9da0-75a78a867a94

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:4bd7f2c-nikolaik   --name openhands-app-4bd7f2c   docker.openhands.dev/openhands/openhands:4bd7f2c
```

### DIFF
--- a/enterprise/server/auth/saas_user_auth.py
+++ b/enterprise/server/auth/saas_user_auth.py
@@ -78,6 +78,8 @@ class SaasUserAuth(UserAuth):
         self.refresh_token = SecretStr(tokens['refresh_token'])
         self.refreshed = True
         if not self.email or not self.email_verified or not self.user_id:
+            # We don't need to verify the signature here because we just refreshed
+            # this token from the IDP via token_manager.refresh()
             access_token_payload = jwt.decode(
                 tokens['access_token'], options={'verify_signature': False}
             )
--- a/enterprise/server/email_validation.py
+++ b/enterprise/server/email_validation.py
@@ -8,14 +8,18 @@ from openhands.core.logger import openhands_logger as logger
 from openhands.server.user_auth import get_user_auth, get_user_id
 
 
-async def get_openhands_user_id(
+async def get_admin_user_id(
     request: Request, user_id: str | None = Depends(get_user_id)
 ) -> str:
     """
     Dependency that validates user has @openhands.dev email domain.
 
     This dependency can be used in place of get_user_id for endpoints that
-    should only be accessible to users with @openhands.dev email addresses.
+    should only be accessible to admin users. Currently, this is implemented
+    by checking for @openhands.dev email domain.
+
+    TODO: In the future, this should be replaced with an explicit is_admin flag
+    in user/org settings instead of relying on email domain validation.
 
     Args:
         request: FastAPI request object
@@ -31,9 +35,9 @@ async def get_openhands_user_id(
     Example:
         @router.post('/endpoint')
         async def create_resource(
-            user_id: str = Depends(get_openhands_user_id),
+            user_id: str = Depends(get_admin_user_id),
         ):
-            # Only @openhands.dev users can access this endpoint
+            # Only admin users can access this endpoint
             pass
     """
     if not user_id:
--- a/enterprise/server/routes/org_models.py
+++ b/enterprise/server/routes/org_models.py
@@ -1,4 +1,4 @@
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, EmailStr, Field
 
 
 class OrgCreationError(Exception):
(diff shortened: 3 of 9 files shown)


enterprise/server/auth/saas_user_auth.py
```diff
@@ -77,6 +77,13 @@ async def refresh(self):
         self.access_token = SecretStr(tokens['access_token'])
         self.refresh_token = SecretStr(tokens['refresh_token'])
         self.refreshed = True
+        if not self.email or not self.email_verified or not self.user_id:
+            access_token_payload = jwt.decode(
+                tokens['access_token'], options={'verify_signature': False}
+            )
```
tofarr, line 85: Please add a note that we don't need to verify the signature here because we just refreshed this token from the IDP.

enterprise/server/email_validation.py
```diff
@@ -0,0 +1,64 @@
...
+Email domain validation utilities for enterprise endpoints.
+"""
+
+from fastapi import Depends, HTTPException, Request, status
+
+from openhands.core.logger import openhands_logger as logger
+from openhands.server.user_auth import get_user_auth, get_user_id
+
+
+async def get_openhands_user_id(
```
tofarr: Maybe rename this method get_admin_user_id? At the moment we permit anybody with an openhands.dev email address to execute admin operations, but I think that eventually we will need to add an explicit flag to the user / org settings instead - something like `is_admin`. We don't need to do this now though - we can simply annotate the method with the future intent.

enterprise/server/routes/org_models.py
```diff
@@ -0,0 +1,83 @@
...
+        if len(v) > 255:
+            raise ValueError('Organization name must be 255 characters or less')
+        return v.strip()
+
+    @field_validator('contact_email')
+    def validate_email(cls, v):
+        """Validate email format."""
+        if not v or '@' not in v:
+            raise ValueError('Invalid email address')
+        return v.strip()
```
tofarr: Pydantic has a built in type EmailStr which may be used to validate emails - lets use that instead of a custom field validator.

enterprise/server/routes/org_models.py
```diff
@@ -0,0 +1,83 @@
...
+    contact_email: str
+
+    @field_validator('name')
+    def validate_name(cls, v):
+        """Validate organization name."""
+        if not v or not v.strip():
+            raise ValueError('Organization name cannot be empty')
+        if len(v) > 255:
+            raise ValueError('Organization name must be 255 characters or less')
+        return v.strip()
```
tofarr: Pydantic has built in functionality which can validate this - let's use this instead:
`Field(min_length=1, max_length=255, strip_whitespace=True)`

enterprise/storage/org_service.py
```diff
@@ -0,0 +1,443 @@
...
+from storage.role_store import RoleStore
+from storage.user_store import UserStore
+
+from openhands.core.logger import openhands_logger as logger
+
+
+class OrgService:
+    """Service for handling organization-related operations."""
+
+    @staticmethod
```
tofarr, line 29: Out of curiosity - why the static methods? 

Typically, in my own code, when I use a class object, it is with the intent that the method may require some internal fields from the class, or that it may be possible that somebody will override it / provide a different implementation using polymorphism. If neither of these are the case, it is less verbose to simply declare these as functions on the package level. (I am aware that some of our services have not followed this pattern - I had just assumed that the authors were ex java devs! ;) )

enterprise/storage/org_store.py
```diff
@@ -160,3 +161,28 @@ def get_kwargs_from_user_settings(user_settings: UserSettings):
...
+            org: Organization entity to persist
+            org_member: Organization member entity to persist
+
+        Returns:
+            Org: The persisted organization object
+
+        Raises:
+            Exception: If database operations fail
+        """
+        with session_maker() as session:
```
tofarr, line 183: Note: (Does not necessarily require fixing) - In the long term I'd loke to move all our DB operations to use asyncio - since our server no longer has an agent worker thread, this is now possible and will be way more efficient.

enterprise/server/auth/saas_user_auth.py
hieptl: Hello @tofarr,

enterprise/server/email_validation.py
hieptl: Hello @tofarr,

enterprise/server/routes/org_models.py
hieptl: Hello @tofarr,

enterprise/server/routes/org_models.py
hieptl: Hello @tofarr,

enterprise/storage/org_store.py
hieptl: Yes, thank you very much for sharing this with me! 🙏

enterprise/storage/org_service.py
hieptl: Yes, I am following the pattern defined by Chuck in his org pull request.

I am using static methods in OrgService to stay aligned with the existing codebase structure. This approach is already applied in other components such as OrgStore and UserStore, and I wanted to maintain that consistency.

More broadly, since OrgStore, UserStore, OrgMemberStore, and RoleStore all use static methods, it seemed appropriate for OrgService to follow the same pattern to ensure a consistent and cohesive design across the project.

That said, I am more than happy to refactor and update the code if you feel a different approach would be more suitable.

Thank you very much for your feedback! 🙏

PR #12264, opened 2026-01-05
APP-307 Add Google Cloud Storage-based EventService implementation

tofarr:
## Summary of PR

This PR adds a Google Cloud Storage-based implementation of `EventService`, allowing events to be stored in GCS buckets instead of the local filesystem.

**Key Features:**
- Events are stored in GCS with the path format: `users/{user_id}/v1_conversations/{conversation_id}/{YYYYMMDDHHMMSS}_{kind}_{id.hex}`
- Uses `AppConversationInfoService` to look up conversations and get user_id
- Implements all abstract methods from `EventService`:
  - `get_event(event_id)` - Retrieve a single event by ID
  - `search_events(...)` - Search events with filtering, sorting, and pagination
  - `count_events(...)` - Count events matching filters
  - `save_event(conversation_id, event)` - Save an event
- Includes caching for user_id lookups to improve performance
- Uses `asyncio.run_in_executor` to run blocking GCS operations asynchronously
- Provides `GoogleCloudEventServiceInjector` that can be configured with an explicit `bucket_name` or reads from `GOOGLE_CLOUD_BUCKET_NAME` environment variable

The implementation follows the same patterns as the existing `FilesystemEventService`.

## Testing

Verified that new V1 events are stored in the cloud:
image

## Change Type

- [x] New feature

## Checklist

## Release Notes

Added Google Cloud Storage-based EventService implementation for storing events in GCS buckets, enabling cloud-native event storage.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:d36d696-nikolaik   --name openhands-app-d36d696   docker.openhands.dev/openhands/openhands:d36d696
```

### DIFF
--- a/enterprise/migrations/versions/086_bump_condenser_defaults.py
+++ b/enterprise/migrations/versions/086_bump_condenser_defaults.py
@@ -1,0 +1,61 @@
+"""bump condenser defaults: max_size 120->240
+
+Revision ID: 086
+Revises: 085
+Create Date: 2026-01-05
+
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.sql import column, table
+
+# revision identifiers, used by Alembic.
+revision: str = '086'
+down_revision: Union[str, None] = '085'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema.
+
+    Update existing users with condenser_max_size=120 or NULL to 240.
+    This covers both users who had the old default (120) explicitly set
+    and users who had NULL (which defaulted to 120 in the application code).
+    The SDK default for keep_first will be used automatically.
+    """
+    user_settings_table = table(
+        'user_settings',
+        column('condenser_max_size', sa.Integer),
+    )
+    # Update users with explicit 120 value
+    op.execute(
+        user_settings_table.update()
+        .where(user_settings_table.c.condenser_max_size == 120)
+        .values(condenser_max_size=240)
+    )
+    # Update users with NULL value (which defaulted to 120 in application code)
+    op.execute(
+        user_settings_table.update()
+        .where(user_settings_table.c.condenser_max_size.is_(None))
+        .values(condenser_max_size=240)
+    )
+
+
+def downgrade() -> None:
+    """Downgrade schema.
+
+    Note: This sets all 240 values back to NULL (not 120) since we can't
+    distinguish between users who had 120 vs NULL before the upgrade.
+    """
+    user_settings_table = table(
+        'user_settings', column('condenser_max_size', sa.Integer)
+    )
+    op.execute(
+        user_settings_table.update()
+        .where(user_settings_table.c.condenser_max_size == 240)
+        .values(condenser_max_size=None)
+    )
--- a/frontend/src/services/settings.ts
+++ b/frontend/src/services/settings.ts
@@ -15,7 +15,7 @@ export const DEFAULT_SETTINGS: Settings = {
   remote_runtime_resource_factor: 1,
   provider_tokens_set: {},
   enable_default_condenser: true,
-  condenser_max_size: 120,
+  condenser_max_size: 240,
   enable_sound_notifications: false,
   user_consents_to_analytics: false,
   enable_proactive_conversation_starters: false,
--- a/frontend/src/utils/has-advanced-settings-set.ts
+++ b/frontend/src/utils/has-advanced-settings-set.ts
@@ -27,7 +27,7 @@ export const hasAdvancedSettingsSet = (
     settings.agent !== undefined && settings.agent !== DEFAULT_SETTINGS.agent;
   // Default is true, so only check if explicitly disabled
   const hasDisabledCondenser = settings.enable_default_condenser === false;
-  // Check if condenser size differs from default (default is 120)
+  // Check if condenser size differs from default (default is 240)
   const hasCustomCondenserSize =
     settings.condenser_max_size !== undefined &&
     settings.condenser_max_size !== null &&
--- a/openhands/app_server/app_conversation/app_conversation_service_base.py
+++ b/openhands/app_server/app_conversation/app_conversation_service_base.py
@@ -380,7 +380,7 @@ class AppConversationServiceBase(AppConversationService, ABC):
         Returns:
             Configured LLMSummarizingCondenser instance
         """
-        # LLMSummarizingCondenser has defaults: max_size=120, keep_first=4
+        # LLMSummarizingCondenser SDK defaults: max_size=240, keep_first=2
         condenser_kwargs = {
             'llm': llm.model_copy(
                 update={
--- a/openhands/server/session/session.py
+++ b/openhands/server/session/session.py
@@ -234,7 +234,7 @@ class WebSession:
             # The order matters: with the browser output first, the summarizer
             # will only see the most recent browser output, which should keep
             # the summarization cost down.
-            max_events_for_condenser = settings.condenser_max_size or 120
+            max_events_for_condenser = settings.condenser_max_size or 240
             default_condenser_config = CondenserPipelineConfig(
                 condensers=[
                     ConversationWindowCondenserConfig(),
--- a/tests/unit/app_server/test_app_conversation_service_base.py
+++ b/tests/unit/app_server/test_app_conversation_service_base.py
@@ -320,9 +320,9 @@ def test_create_condenser_default_agent_with_none_max_size(mock_condenser_class)
         # Assert
         mock_condenser_class.assert_called_once()
         call_kwargs = mock_condenser_class.call_args[1]
-        # When condenser_max_size is None, max_size should not be passed (uses SDK default of 120)
+        # When condenser_max_size is None, max_size should not be passed (uses SDK default of 240)
         assert 'max_size' not in call_kwargs
-        # keep_first is never passed (uses SDK default of 4)
+        # keep_first is never passed (uses SDK default of 2)
         assert 'keep_first' not in call_kwargs
         assert call_kwargs['llm'].usage_id == 'condenser'
         mock_llm.model_copy.assert_called_once()
@@ -358,7 +358,7 @@ def test_create_condenser_default_agent_with_custom_max_size(mock_condenser_clas
         mock_condenser_class.assert_called_once()
         call_kwargs = mock_condenser_class.call_args[1]
         assert call_kwargs['max_size'] == 150  # Custom value should be used
-        # keep_first is never passed (uses SDK default of 4)
+        # keep_first is never passed (uses SDK default of 2)
         assert 'keep_first' not in call_kwargs
         assert call_kwargs['llm'].usage_id == 'condenser'
         mock_llm.model_copy.assert_called_once()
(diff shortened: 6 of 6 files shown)


malhotra5:

malhotra5:
## 🔥 Roasted Code Review (Linus-Style)

🔴 **Taste Rating: Needs Improvement** - Violates fundamental principles

---

### **[CRITICAL ISSUES]** (Must fix - these break fundamental principles)

**[google_cloud_event_service.py, Lines 183-188] 🚨 Data Structure Disaster: O(n) Full Bucket Scan**
```python
for blob in self.bucket.list_blobs():
    filename = self._get_filename_from_blob(blob)
    if filename.endswith(f"_{id_hex}"):
        return self._read_event(blob.name)
```
You're iterating over **every single blob in the entire bucket** to find one event. This is not a search algorithm, this is a cry for help. At scale, this will bring your system to its knees. GCS charges per LIST operation and you're doing a full scan every time someone looks up an event by ID.

**Fix:** Store event paths in a database index, or at minimum use a prefix-based lookup if you know the conversation_id.

---

**[google_cloud_event_service.py, Lines 286-309] 🚨 N+1 Query Pattern on Steroids**
```python
conversations_page = await self.app_conversation_info_service.search_app_conversation_info(limit=1000)
for conversation in conversations_page.items:
    blobs = await loop.run_in_executor(None, list_blobs)
```
You fetch up to 1000 conversations, then make a **separate GCS LIST call for each one**. That's potentially 1000 network round-trips. This is the N+1 query problem, except you've made it N+1 cloud API calls. Your cloud bill will be... educational.

**Fix:** Either use a proper database for event metadata, or batch these operations. Better yet, question whether searching across all conversations without a conversation_id filter is even a valid use case.

---

**[google_cloud_event_service.py, Lines 43-45] 🚨 Unbounded Cache = Memory Leak**
```python
_conversation_user_cache: dict[UUID, str | None] = field(
    default_factory=dict, init=False
)
```
This cache grows forever. No TTL, no size limit, no eviction policy. In a long-running service, this will eventually consume all available memory. "Caching for efficiency" that causes OOM kills is not efficient.

**Fix:** Use `functools.lru_cache` with a maxsize, or implement proper TTL-based caching.

---

### **[IMPROVEMENT OPPORTUNITIES]** (Should fix - violates good taste)

**[google_cloud_event_service.py, Lines 68-76, 78-83, 113-124] 🔄 Massive Code Duplication**

You've copy-pasted `_get_event_filename`, `_timestamp_to_str`, and `_parse_filename` from `FilesystemEventServiceBase`. This is the exact opposite of DRY. When someone fixes a bug in the filesystem version, they'll forget to fix it here.

**Fix:** Extract shared logic into a common base class or utility module. The filesystem implementation already has `FilesystemEventServiceBase` - consider whether GCS can share that abstraction.

---

**[google_cloud_event_service.py, Lines 286-289, 381-385] 🎯 Magic Number: limit=1000**
```python
await self.app_conversation_info_service.search_app_conversation_info(limit=1000)
```
Hardcoded limit of 1000 with no pagination. What happens when a user has 1001 conversations? You silently ignore the rest. This is a time bomb.

**Fix:** Either paginate through all results, or document this limitation clearly and fail fast when exceeded.

---

**[google_cloud_event_service.py, Lines 302-303] 🐍 Python Late Binding Workaround**
```python
def list_blobs(p: str = prefix) -> list[Blob]:
    return self._list_event_blobs(p)
```
Using default argument to capture loop variable is a classic Python gotcha workaround. It works, but it's a code smell that suggests the loop structure itself is wrong.

**Fix:** Use `functools.partial` or restructure to avoid closures in loops entirely.

---

**[google_cloud_event_service.py, Lines 105-107, 137-139] 🙈 Silent Error Swallowing**
```python
except Exception:
    _logger.exception(f"Error reading event from {path}")
    return None
```
You log the error and then return None, making it indistinguishable from "event not found". The caller has no idea if the event doesn't exist or if GCS is on fire.

**Fix:** Let unexpected exceptions propagate, or return a proper error type that distinguishes "not found" from "failed to read".

---

**[google_cloud_event_service.py, Line 349] 🤔 Unused Parameter**
```python
async def count_events(
    ...
    sort_order: EventSortOrder = EventSortOrder.TIMESTAMP,  # Why?
) -> int:
```
`sort_order` in a count operation? You're counting, not sorting. This parameter does nothing.

**Fix:** Remove it, or if it's required by the interface, document why it's ignored.

---

### **[STYLE NOTES]** (Minor - but worth mentioning)

**[google_cloud_event_service.py, Line 8] 📦 Unused Import**
```python
from typing import AsyncGenerator
```
`AsyncGenerator` is imported but the class uses it only in the injector, not in the main service. Consider if this belongs here.

---

**[google_cloud_event_service.py, Line 13] 📦 Unused Import**
```python
from google.cloud import storage
```
You import `storage` but only use it in the injector. The main class receives a `Bucket` directly.

---

### **VERDICT:**

❌ **Needs rework** - Fundamental design issues must be addressed first

The O(n) bucket scan in `get_event()` alone is a showstopper. This implementation will work fine in testing with 10 events, then fall over spectacularly in production with 10,000.

---

### **KEY INSIGHT:**

> *"Bad programmers worry about the code. Good programmers worry about data structures and their relationships."*

This PR treats GCS like a filesystem with `glob`, but GCS is not a filesystem - it's an object store with different performance characteristics. The data access patterns here (full scans, N+1 queries) would be questionable even on a local filesystem, but on a cloud object store with per-operation costs and network latency, they're untenable.

Consider: Do you actually need to search events by ID without knowing the conversation? If not, require `conversation_id` and use prefix-based lookups. If yes, you need a proper index (database) for event metadata, not a bucket scan.

openhands/app_server/event/google_cloud_event_service.py
```diff
@@ -0,0 +1,448 @@
...
+        if conversation:
+            user_id = conversation.created_by_user_id
+            self._conversation_user_cache[conversation_id] = user_id
+            return user_id
+        return None
+
+    def _get_event_path(self, user_id: str, conversation_id: UUID, event: Event) -> str:
+        """Generate the full GCS path for an event."""
+        filename = self._get_event_filename(event)
+        return f'users/{user_id}/v1_conversations/{conversation_id}/{filename}'
```
malhotra5: NIT: maybe we can define a global file location constant somewhere. I think there are other parts of the codebase that way want to reference the same prefixes

openhands/app_server/event/google_cloud_event_service.py
```diff
@@ -0,0 +1,448 @@
...
+        """Get the event with the given id, or None if not found."""
+        if isinstance(event_id, str) and '-' in event_id:
+            id_hex = event_id.replace('-', '')
+        else:
+            id_hex = event_id
+
+        loop = asyncio.get_running_loop()
+
+        def find_and_load_event() -> Event | None:
+            for blob in self.bucket.list_blobs():
```
malhotra5: Is the bucket at the root level or specific to the user? (i.e does `list_blobs` read blobs across all users or specifically the users' blobs?)

openhands/app_server/config.py
```diff
@@ -161,7 +164,13 @@ def config_from_env() -> AppServerConfig:
     config: AppServerConfig = from_env(AppServerConfig, 'OH')  # type: ignore
 
     if config.event is None:
-        config.event = FilesystemEventServiceInjector()
+        if os.environ.get('FILE_STORE') == 'google_cloud':
+            # Legacy V0 google cloud storage configuration
+            config.event = GoogleCloudEventServiceInjector(
+                bucket_name=os.environ.get('FILE_STORE_PATH')
```
raymyers, line 170: If FILE_STORE_PATH is missing, this passes None, then what happens?

openhands/app_server/event/google_cloud_event_service.py
```diff
@@ -0,0 +1,448 @@
...
+        """Read an event from GCS."""
+        blob: Blob = self.bucket.blob(path)
+        try:
+            with blob.open('r') as f:
+                json_data = f.read()
+            return Event.model_validate_json(json_data)
+        except NotFound:
+            return None
+        except Exception:
+            _logger.exception(f'Error reading event from {path}')
```
raymyers, line 42: I'm assuming there's a reason here to use a "Catch-all, log, ignore", but it's not the pattern I'd usually recommend. Usually I would catch only specific exceptions by name, and handle unknown errors further up the stack.

openhands/app_server/event/google_cloud_event_service.py
```diff
@@ -0,0 +1,448 @@
...
+        try:
+            parts = blob.name.split('/')
+            for i, part in enumerate(parts):
+                if part == 'v1_conversations' and i + 1 < len(parts):
+                    return UUID(parts[i + 1])
+        except Exception:
+            pass
+        return None
+
+    def _filter_blobs_by_criteria(
```
raymyers: This loop has appending has its default action, and 5 ways to *not* append. Consider extracting the logic into a predicate to make the loop simple.

```python
 
def blob_matches_filter(blob: Blob,  kind__eq: EventKind | None, timestamp__gte: datetime | None,
        timestamp__lt: datetime | None) -> Bool:
            filename = self._get_filename_from_blob(blob)
            filename_info = self._parse_filename(filename)
            if not filename_info:
                return False

            if kind__eq and filename_info['kind'] != kind__eq:
                return False

            if timestamp__gte or timestamp__lt:
                try:
                    file_timestamp = datetime.strptime(
                        filename_info['timestamp'], '%Y%m%d%H%M%S'
                    )
                    if timestamp__gte and file_timestamp < timestamp__gte:
                        return False
                    if timestamp__lt and file_timestamp >= timestamp__lt:
                        return False
                except ValueError:
                    return False

            return True
```

raymyers: Some primitive obsession here too maybe. We could imagine a little FilterParams object to store the `kind__eq, timestamp__gte, timestamp__lt`, or even a TimeRange. Not a big deal.

openhands/app_server/event/google_cloud_event_service.py
```diff
@@ -0,0 +1,448 @@
...
+            if timestamp__gte or timestamp__lt:
+                try:
+                    file_timestamp = datetime.strptime(
+                        filename_info['timestamp'], '%Y%m%d%H%M%S'
+                    )
+                    if timestamp__gte and file_timestamp < timestamp__gte:
+                        continue
+                    if timestamp__lt and file_timestamp >= timestamp__lt:
+                        continue
+                except ValueError:
```
raymyers: Can we check for the specific fields we need instead of catching ValueError?

openhands/app_server/event/google_cloud_event_service.py
```diff
@@ -0,0 +1,448 @@
...
+            all_blobs.extend((blob, conversation.id) for blob in blobs)
+
+        # Sort all blobs by filename
+        all_blobs.sort(
+            key=lambda x: self._get_filename_from_blob(x[0]),
+            reverse=(sort_order == EventSortOrder.TIMESTAMP_DESC),
+        )
+
+        # Handle pagination
+        start_index = 0
```
raymyers: Pagination logic is duplicated.

raymyers (approved):

openhands/app_server/event/google_cloud_event_service.py
raymyers: Yeah this looks like a problem. Needs a prefix arg.

https://docs.cloud.google.com/python/docs/reference/storage/latest/google.cloud.storage.client.Client.html#google_cloud_storage_client_Client_list_blobs

raymyers (changes requested):
Left several notes, most serious is the call to list_blobs with no qualifier.

openhands/app_server/config.py
tofarr: You'll get a ValidationError from pydantic.

enterprise/server/sharing/google_cloud_shared_event_service.py
```diff
@@ -118,7 +129,7 @@ async def count_shared_events(
         )
 
 
-class SharedEventServiceImplInjector(SharedEventServiceInjector):
+class GoogleCloudSharedEventServiceInjector(SharedEventServiceInjector):
```
raymyers, line 133: Side note, I'm not really sure what an Injector is.

enterprise/server/sharing/shared_event_service.py
```diff
@@ -18,7 +18,7 @@ class SharedEventService(ABC):
 
     @abstractmethod
     async def get_shared_event(
-        self, conversation_id: UUID, event_id: str
+        self, conversation_id: UUID, event_id: UUID
```
raymyers, line 21: Nice.

PR #12288, opened 2026-01-07
feat: implement reCAPTCHA enterprise risk-based non-interactive

hieptl:
## Demo video

https://github.com/user-attachments/assets/bde3d2cd-a46e-4e8b-9851-9f066b5e811b

## Background

We want to enhance fraud and abuse prevention without degrading user experience. reCAPTCHA Enterprise supports risk analysis and automated decisioning, allowing us to evaluate requests server-side and strictly block those deemed high-risk.

## References

Use the following documentation as a starting point and identify additional relevant resources as needed:

- https://docs.cloud.google.com/recaptcha/docs/create-assessment-website#create-assessment-Python
- https://docs.cloud.google.com/recaptcha/docs/interpret-assessment-website
- https://docs.cloud.google.com/recaptcha/docs/account-defender#hashed-or-encrypted

## Objective

Implement reCAPTCHA Enterprise for our product using backend-only risk assessments, with no user-facing CAPTCHA challenges.

### 1. Research

Review reCAPTCHA Enterprise documentation with a focus on:

- Risk-based assessments
- Account Defender signals

Identify best practices and additional relevant documentation or examples.

### 2. API Integration

Implement a backend API call to reCAPTCHA Enterprise to create an assessment.

Include relevant request data such as:

- User email (hashed or encrypted as required)
- IP address
- User agent
- Request headers and other useful metadata

Use the official reCAPTCHA Enterprise API (Python preferred, or another approved language).

### 3. Risk Scoring & Decision Logic

Retrieve and interpret the risk score and signals returned by reCAPTCHA Enterprise.

Define clear thresholds for:

- Allowing requests
- Blocking requests

High-risk requests must be strictly blocked without presenting any CAPTCHA or challenge to the user.

### 4. Non-Interactive Behavior

Ensure no image-based or interactive CAPTCHA challenges are shown at any point.

All decisions must be automated and handled server-side.

## Acceptance Criteria

- Backend successfully creates reCAPTCHA Enterprise assessments.
- Risk scores are evaluated correctly.
- High-risk requests are blocked automatically.
- No traditional CAPTCHA UI is exposed to users.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:46359ae-nikolaik   --name openhands-app-46359ae   docker.openhands.dev/openhands/openhands:46359ae
```


enterprise/server/routes/auth.py
```diff
@@ -149,8 +175,43 @@ async def keycloak_callback(
...
+                        'score': result.score,
+                        'user_id': user_id,
+                    },
+                )
+                # Redirect to home with error parameter
+                error_url = f'{request.base_url}?recaptcha_blocked=true'
+                return RedirectResponse(error_url, status_code=302)
+
+        except Exception as e:
+            logger.error(f'reCAPTCHA verification error at callback: {e}')
```
raymyers: prefer logger.exception()

enterprise/server/routes/auth.py
```diff
@@ -149,8 +175,43 @@ async def keycloak_callback(
...
+
+        # Handle X-Forwarded-For for proxied requests
+        forwarded_for = request.headers.get('X-Forwarded-For')
+        if forwarded_for:
+            user_ip = forwarded_for.split(',')[0].strip()
+
+        try:
+            result = recaptcha_service.create_assessment(
+                token=recaptcha_token,
+                action='LOGIN',
```
raymyers, line 191: cc: @malhotra5 We discussed if this should be all logins or just create, some middleground...

This one happens maybe a few times a day for a user. We don't _think_ it's on every usage

enterprise/server/routes/auth.py
```diff
@@ -149,8 +175,43 @@ async def keycloak_callback(
...
+                        'user_id': user_id,
+                    },
+                )
+                # Redirect to home with error parameter
+                error_url = f'{request.base_url}?recaptcha_blocked=true'
+                return RedirectResponse(error_url, status_code=302)
+
+        except Exception as e:
+            logger.error(f'reCAPTCHA verification error at callback: {e}')
+            # Fail open - continue with login if reCAPTCHA service unavailable
```
raymyers, line 212: Personally I'm ok with this fail open. We should probably put a monitor threshold on `reCAPTCHA verification error`. (seperately)

frontend/src/hooks/use-recaptcha.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    document.head.appendChild(script);
+  }, [siteKey]);
+
+  const executeRecaptcha = useCallback(
+    async (action: string): Promise<string | null> => {
+      if (!siteKey || !isReady || !window.grecaptcha?.enterprise) return null;
+
+      try {
+        const token = await window.grecaptcha.enterprise.execute(siteKey, {
```
malhotra5, line 63: Could you clarify what `window.grecaptcha.enterprise` is doing?

frontend/src/hooks/use-recaptcha.ts
```diff
@@ -0,0 +1,76 @@
...
+      window.grecaptcha.enterprise.ready(() => setIsReady(true));
+      return;
+    }
+
+    setIsLoading(true);
+
+    const script = document.createElement("script");
+    script.src = `${RECAPTCHA_SCRIPT_URL}?render=${siteKey}`;
+    script.async = true;
+    script.defer = true;
```
malhotra5, line 41: Do we have to lazy load the javascript this way or is there a way to load the asset in advance?

raymyers:
From the demo video, I'm curious if someone is rejected repeatedly, should they go somewhere and ask for support? Is there another action we can give them additional to "try again"?

frontend/src/components/features/auth/login-content.tsx
```diff
@@ -105,6 +146,11 @@ export function LoginContent({
           {t(I18nKey.AUTH$DUPLICATE_EMAIL_ERROR)}
         </p>
       )}
+      {recaptchaBlocked && (
```
malhotra5, line 149: QQ: how does the `recaptchaBlocked` state update in the event the BE sends a blocked response?

malhotra5: @openhands could you help me with this question

malhotra5:
Noting for future

The reCaptcha docs have [recommendations](https://developers.google.com/recaptcha/docs/v3#interpreting_the_score) on how to handle low assessment scores, which we may want to pursue later on

frontend/src/hooks/use-recaptcha.ts
hieptl: Hello @malhotra5,

window.grecaptcha.enterprise is the global JavaScript object provided by Google's reCAPTCHA Enterprise SDK after the script loads from https://www.google.com/recaptcha/enterprise.js.

grecaptcha.enterprise.execute(siteKey, { action }) - Generates a reCAPTCHA token invisibly (no user interaction required). The action parameter (e.g., "LOGIN") helps Google's risk analysis understand the context of the user's action.

How it works in our flow:

1. The script loads with ?render=<siteKey> which enables invisible/score-based mode
2. When the user clicks login, we call execute() to get a token
3. The token is sent to our backend (encoded in OAuth state)
4. Our backend calls Google's CreateAssessment API with this token to get a risk score (0.0-1.0)
5. We block requests with scores below the threshold (< 0.3)

This is the invisible reCAPTCHA approach - users never see a challenge unless Google determines they're high-risk, and even then our implementation blocks rather than challenges

Thank you! 🙏

frontend/src/hooks/use-recaptcha.ts
hieptl: Hello @malhotra5,

The current lazy-loading approach aligns well with our use case, and here’s the reasoning behind it:

Why we lazy-load (current approach):

1. Conditional loading: The script only loads when siteKey is provided. If reCAPTCHA isn't configured for the environment, we don't load unnecessary JavaScript.
2. Performance: The reCAPTCHA Enterprise script (~150KB) isn't needed until the user is on a page with login functionality. Loading it on every page would slow down initial page loads unnecessarily.
3. async + defer: These attributes ensure the script doesn't block HTML parsing or rendering. The script loads in parallel and executes after the DOM is ready.

Alternative: Preloading in index.html

We could add the script directly to index.html:
 
```
<script src="https://www.google.com/recaptcha/enterprise.js?render=SITE_KEY" async></script>
```

But this has downsides:
- The site key would need to be hardcoded or injected at build time
- The script loads on every page, even where login isn't available
- In OSS mode (no reCAPTCHA), we'd load an unused script

Conclusion

 The lazy-loading approach is the right choice because:
 - It respects the dynamic configuration (RECAPTCHA_SITE_KEY from backend)
 - It only loads when needed (login pages)
 - The grecaptcha.enterprise.ready() callback ensures we wait for initialization before use

The slight delay in script loading is negligible since users need time to read the login UI before clicking a button anyway.

Thank you! 🙏

hieptl:
> Noting for future
> 
> The reCaptcha docs have [recommendations](https://developers.google.com/recaptcha/docs/v3#interpreting_the_score) on how to handle low assessment scores, which we may want to pursue later on

Hello @malhotra5,

Thank you for sharing this.

It appears that this has already been addressed through our score-based blocking mechanism. Specifically, requests with a score below 0.3 are blocked, and this threshold is configurable via RECAPTCHA_BLOCK_THRESHOLD.

constant.py
```
RECAPTCHA_BLOCK_THRESHOLD = float(os.getenv('RECAPTCHA_BLOCK_THRESHOLD', '0.3'))
```

recaptcha_service.py
```
return AssessmentResult(
    score=score,
    valid=valid,
    action_valid=action_valid,
    reason_codes=reason_codes,
    allowed=valid and action_valid and score >= RECAPTCHA_BLOCK_THRESHOLD,
)
```

Please let me know if I may have misunderstood or if this does not fully address your concern. I’m happy to clarify further.

Thank you! 🙏

enterprise/server/routes/auth.py
hieptl: Hello @raymyers,

hieptl:
> From the demo video, I'm curious if someone is rejected repeatedly, should they go somewhere and ask for support? Is there another action we can give them additional to "try again"?

Hello @raymyers,

I’ve updated the error message to make it clearer and have also refreshed the demo video in the pull request description.

> Access blocked due to suspicious activity. If you believe this is an error, please contact contact@openhands.dev for assistance.

Please feel free to refer to the updated video in the pull request for additional context.

enterprise/server/routes/auth.py
hieptl: Hello @raymyers,

I’ll create a follow-up issue to add monitoring for reCAPTCHA verification error rates so we can proactively detect any anomalies.

enterprise/server/routes/auth.py
hieptl: Hello @raymyers,

Could you please advise on what you would recommend addressing here? Specifically, should we limit the CAPTCHA check to sign-up only and not apply it to login, or would you prefer that we keep the current approach for now?

raymyers: We're comfortable as is. We might introduce other actions later. Thanks for checking

PR #12301, opened 2026-01-07
Fix GitLab token refresh race condition with retry mechanism

raymyers:
## Summary of PR

When multiple pods try to refresh the same GitLab token simultaneously, only the first succeeds because GitLab invalidates refresh tokens on first use. This causes `invalid_grant` errors for subsequent requests from other pods.

This change adds a retry mechanism to `AuthTokenStore.load_tokens()` that:
- Catches `invalid_grant` errors (HTTP 400) during token refresh
- Raises a new `TokenRefreshRaceError` to signal a race condition
- Waits briefly (0.5s) and re-reads tokens from the database
- Retries up to `max_retries` times (default: 3, configurable on the class)
- Returns the updated tokens stored by the winning process

This approach is simpler than distributed locking (e.g., Redis) and uses the database as the source of truth for token state.

### Key Changes

**`enterprise/storage/auth_token_store.py`:**
- Added `TokenRefreshRaceError` exception class
- Added `max_retries` field to `AuthTokenStore` dataclass (default: 3)
- Refactored `load_tokens()` to implement retry loop
- Added `_load_tokens_with_refresh()` - original logic that detects race conditions
- Added `_load_tokens_without_refresh()` - simple DB read for retry scenarios

**`enterprise/tests/unit/storage/test_auth_token_store.py`:** (new file)
- Tests for `TokenRefreshRaceError` exception
- Tests for retry logic catching `invalid_grant` and retrying successfully
- Tests for non-invalid_grant errors being re-raised
- Tests for configurable `max_retries`
- All tests mock `asyncio.sleep` to avoid real delays

## Change Type

- [x] Bug fix

## Checklist

## Fixes

Resolves GitLab token refresh race condition causing `invalid_grant` errors in multi-pod deployments.

## Release Notes

@raymyers can click here to [continue refining the PR](https://staging.all-hands.dev/conversations/None)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:e53e85f-nikolaik   --name openhands-app-e53e85f   docker.openhands.dev/openhands/openhands:e53e85f
```

### DIFF
--- a/enterprise/tests/unit/storage/test_auth_token_store.py
+++ b/enterprise/tests/unit/storage/test_auth_token_store.py
@@ -1,6 +1,7 @@
 """Tests for auth_token_store.py, specifically the retry logic for token refresh race conditions."""
 
 import asyncio
+from contextlib import asynccontextmanager
 from unittest.mock import AsyncMock, MagicMock, patch
 
 import httpx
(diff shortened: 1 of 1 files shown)


enterprise/storage/auth_token_store.py
```diff
@@ -116,16 +164,31 @@ async def load_tokens(
...
+                        )
+                        if check_expiration_and_refresh
+                        else None
                     )
-                    if check_expiration_and_refresh
-                    else None
-                )
+                except httpx.HTTPStatusError as e:
+                    # Check if this is an invalid_grant error from the IDP
+                    # This typically means another process already used this refresh token
```
raymyers, line 180: Probably good to mention GitLab here, since that's who sends this string.

Issue #12279, opened 2026-01-06
Add message queue support for V1 conversations during WebSocket connection


amanape:
## Problem

V1 conversations do not queue user messages when the WebSocket connection is not yet established. A user is unable to submit a message before the WebSocket is fully connected (e.g., during runtime startup), instead of being queued and sent automatically once the connection is ready.

## Background

PR [#11687](<https://github.com/OpenHands/OpenHands/issues/11687>) added message queuing support for V0 conversations in the [Socket.IO](<http://Socket.IO>) WebSocket provider. However, this functionality was never implemented for V1 conversations which use a native WebSocket via `ConversationWebSocketProvider`.

## Current Behavior

The chat input is disabled when the WebSocket is not connected

## Expected Behavior

Messages submitted before the WebSocket connection is established should be queued and automatically sent once the connection is ready, matching the V0 behavior.

## Acceptance Criteria

> \[!IMPORTANT\]<br>The PR implementing this change should include sufficient test coverage. A test-driven development approach is preferred.

* The V1 WebSocket provider should maintain a queue of pending messages when the connection is not yet open.
* Queued messages should be automatically flushed in order when the WebSocket connection opens.
* The pending message queue should be cleared when the conversation changes or is intentionally stopped.
* Unit tests should be added to verify the message queuing and flushing behavior.

## Relevant Files

[frontend/src/contexts/conversation-websocket-context.tsx lines 722-741](<https://github.com/OpenHands/OpenHands/blob/main/frontend/src/contexts/conversation-websocket-context.tsx#L722-L741>) contains the V1 `sendMessage` function that might be relevant for queuing support.<br>[frontend/src/context/ws-client-provider.tsx](<https://github.com/OpenHands/OpenHands/blob/main/frontend/src/context/ws-client-provider.tsx>) contains the V0 implementation with `pendingEventsRef` and `flushPendingEvents` that might serve as a reference. Feel free to implement a different strategy if you feel it would be better.<br>[frontend/src/hooks/use-send-message.ts](<https://github.com/OpenHands/OpenHands/blob/main/frontend/src/hooks/use-send-message.ts>) is the unified hook that routes messages to V0 or V1 providers.<br>[frontend/**tests**/conversation-websocket-handler.test.tsx](<https://github.com/OpenHands/OpenHands/blob/main/frontend/__tests__/conversation-websocket-handler.test.tsx>) should be extended with tests for the new queuing behavior.

abhayymishraa:
[...]

PR #12314, opened 2026-01-08
feat: add message queue support for V1 conversations

nehaaprasad:
### Summary of PR

- adds message queuing for V1 conversations when WebSocket connection is not established, matching V0 behavior. Messages are automatically sent once connection opens.

## Change Type

- [x] New feature

## Fixes : #12279

Resolves V1 conversation message queuing issue during WebSocket connection.

**V1 conversations now queue messages sent before WebSocket connection is established.**

### DIFF
--- a/frontend/__tests__/conversation-websocket-handler.test.tsx
+++ b/frontend/__tests__/conversation-websocket-handler.test.tsx
@@ -709,9 +709,9 @@ describe("Conversation WebSocket Handler", () => {
       }
 
       // Set up MSW to delay connection (simulate slow connection)
-      let connectionDeferred: { resolve: () => void } | null = null;
+      const deferred: { resolve: () => void } = { resolve: () => {} };
       const connectionPromise = new Promise<void>((resolve) => {
-        connectionDeferred = { resolve };
+        deferred.resolve = resolve;
       });
 
       mswServer.use(
@@ -748,7 +748,7 @@ describe("Conversation WebSocket Handler", () => {
       );
 
       // Now establish the connection
-      connectionDeferred?.resolve();
+      deferred.resolve();
 
       // Wait for connection to be established
       await waitFor(() => {
@@ -800,9 +800,9 @@ describe("Conversation WebSocket Handler", () => {
       }
 
       // Set up MSW to delay connection establishment
-      let connectionDeferred: { resolve: () => void } | null = null;
+      const deferred: { resolve: () => void } = { resolve: () => {} };
       const connectionPromise = new Promise<void>((resolve) => {
-        connectionDeferred = { resolve };
+        deferred.resolve = resolve;
       });
 
       mswServer.use(
@@ -834,7 +834,7 @@ describe("Conversation WebSocket Handler", () => {
       });
 
       // Now establish the connection
-      connectionDeferred?.resolve();
+      deferred.resolve();
 
       // Wait for connection to be established
       // The test passes if no errors occur and connection is established
--- a/frontend/src/contexts/conversation-websocket-context.tsx
+++ b/frontend/src/contexts/conversation-websocket-context.tsx
@@ -129,6 +129,27 @@ export function ConversationWebSocketProvider({
     conversationId: string;
   } | null>(null);
 
+  // Queue for messages sent before WebSocket connection is established
+  const pendingMessagesRef = useRef<V1SendMessageRequest[]>([]);
+
+  // Helper function to flush pending messages when connection opens
+  const flushPendingMessages = useCallback(
+    (socket: WebSocket | null) => {
+      if (!socket || socket.readyState !== WebSocket.OPEN) return;
+      const messages = pendingMessagesRef.current;
+      pendingMessagesRef.current = [];
+      for (const message of messages) {
+        try {
+          socket.send(JSON.stringify(message));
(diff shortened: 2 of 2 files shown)


ashavolian:
lets get this out!

nehaaprasad:
@ashavolian 
sorry what do you mean?

enyst:
@openhands Do a /codereview-roasted on this PR. Post your feedback as a comment on github on the PR, don't tell it to me. Post it as comment, and note that it will be rendered as markdown. Most importantly, does it solve the issue and does it have bad consequences? Read the related issue it fixes, too.

enyst:
(OpenHands-Agent-GPT-5.2 in review-roasted in Linus style mode)

🟡 **Taste rating:** *Acceptable*, but you left some sharp edges lying around.

---

## Does this actually solve #12279?
**Mostly yes.** The core behavior change—**don’t error when `sendMessage()` is called before the WS is open; queue it and flush on `onopen`**—matches the issue’s “queue like V0” requirement.

But… there are a couple “works in the happy path” assumptions that can bite.

---

## [CRITICAL ISSUES] (things that can break users)

### 1) **You’re sending on the wrong socket.**
In `ConversationWebSocketProvider`, `sendMessage()` queues if `socket` isn’t open. Fine.

But when the connection opens, you call:
- `flushPendingMessages(mainSocket)` in the main socket `onopen`
- `flushPendingMessages(planningAgentSocket)` in the planning socket `onopen`

…and `flushPendingMessages()` defaults to `socket` anyway.

**Problem:** the queue is *global* (`pendingMessagesRef`) but you have *two sockets*, and which `onopen` fires first determines where the queued messages go. That’s a race. You can absolutely end up sending user messages to the planning socket (or vice versa), depending on `conversationMode` / initialization order.

If V1 has a single “send” target, then:
- keep a single queue **per socket** (or per mode)
- or enforce flushing only from the socket that is actually used for `sendMessage()`

Right now it’s “first socket to open wins”. That’s not a design, that’s an accident.

### 2) **Silent failure semantics change**
Previously: calling `sendMessage()` when disconnected raised an error and set `errorMessage`.

Now: it **returns successfully** (queues) with no user-visible feedback.

That’s fine *if* the UI is already disabling input until open (it is per issue), but this change affects any callers that relied on the exception for control flow.

At minimum, document it in code (or better: return something like `{ queued: true }`), otherwise you’re changing an API contract silently.

### 3) **Unbounded memory / spam risk**
`pendingMessagesRef.current.push(message)` with no limit means:
- a buggy UI loop
- or a user holding Enter
- or some reconnection limbo

…can accumulate an unbounded array in memory, then blast the server when `onopen` happens.

Fix: cap it (e.g. 50), drop oldest/newest, and surface an error/toast when the cap is hit.

---

## [IMPROVEMENT OPPORTUNITIES] (good taste problems)

### 4) `flushPendingMessages()` captures `socket` but you don’t want it
`flushPendingMessages` has a dependency on `[socket, setErrorMessage]` and a default param of `socket`.

This is exactly how you get stale closure bugs (“flush uses whatever `socket` was at render time”). Since you already pass the socket explicitly from `onopen`, just:
- remove the default arg
- remove `socket` from deps
- make it a pure “flush this queue to this socket” function

### 5) Error handling is half-baked
If `targetSocket.send()` throws during flush:
- you set a generic error message
- and then you clear the whole queue anyway

So one bad send nukes the rest.

Better: only remove messages after successful send, keep failed ones, or bail out leaving the remainder in queue.

### 6) Tests: one of them is literally `expect(true).toBe(true)`
That’s not a test, that’s a lie.

Also the “flush queued messages” test doesn’t actually assert messages were flushed—only that `sendMessage()` resolved and the connection eventually opens. That’s testing your *test harness*, not your feature.

You should assert:
- queued messages are actually delivered on the WS server side (MSW can observe)
- order is preserved
- they’re flushed exactly once

---

## [BAD CONSEQUENCES / EDGE CASES]
- If the WS never opens, messages are queued forever with no UX signal.
- If conversation changes mid-queue, you clear the queue (good), but you don’t cancel any pending “user intent” in the UI. That may confuse users (“I hit send, nothing happened”).
- Multi-socket race described above is the big one.

---

## Verdict
❌ **Needs rework before merge**

The feature idea is correct, but the implementation is problematic around **socket ownership** and **queue semantics**. Fix the socket/mode routing (per-socket queue or explicit flush target), add a queue cap, and replace the fake test with real assertions.

**Key insight:** you can’t have *one global queue* when you have *multiple possible transport endpoints* and pretend it’ll “just work”.

xingyaoww:
@OpenHands please fix all the CI test failure in this PR

jpshackelford:
Closing in favor of #13380 by @hieptl which implements message queuing for V1 conversations along with draft persistence. It fully addresses the acceptance criteria from #12279. Great work! 🎉

jpshackelford:
Hey @naaa760, thanks for tackling this! Your approach with `pendingMessagesRef` and `flushPendingMessages` was spot on - @hieptl's PR #13380 uses essentially the same pattern. Since that PR also bundles in draft persistence, we're consolidating everything there. Appreciate you pushing this forward! 🙏

enyst:
@jpshackelford I think the agent got the author wrong 😅

PR #12329, opened 2026-01-08
Fix: Handle runtime unavailable error in Slack follow-up messages

### LINKED ISSUE #12325: Handle follow-up messages when runtime is down - "unexpected error when starting new job"
## Problem

When sending a follow-up message to a conversation with a stopped runtime, users receive an "unexpected error when starting new job" error. This has been happening since the GitHub resolver was implemented (months ago).

## Context

This issue affects the GitHub resolver and Slack integration when the runtime is not available.

### Tactical Fix
Add error handling specifically for the Slack integration to gracefully handle cases when the runtime is down.

### Strategic Fix
Build an actual message queue that can work across different platforms (GitHub, Slack, etc.) to properly handle queued messages when the runtime is not yet available.

## Related

The overall goal is to make sure the system can handle queued messages when the runtime was not up yet.

---

**Linear ticket:** [APP-355](https://linear.app/all-hands-ai/issue/APP-355/handle-follow-up-messages-when-runtime-is-down-unexpected-error-when)

saurya:
## Summary

Fixes #12325

This is the **tactical fix** for handling follow-up messages when the runtime is down in Slack conversations.

## Problem

When a user sends a follow-up message to an existing Slack conversation and the runtime is not running (stopped/down), the system was returning a generic error message or "Conversation is still starting".

## Solution

Since Slack is asynchronous, we can simply wait for the runtime to come up before sending the message. The user doesn't need to know about internal runtime state - they just see a slight delay before getting a response.

### Architecture Note

The Slack webhook handler returns `JSONResponse({'success': True})` immediately and processes the message in a **FastAPI background task**. This means:
- ✅ Slack gets its response within 3 seconds (avoiding retries)
- ✅ The HTTP connection is released immediately
- ⚠️ The only resource concern is **background task accumulation** (coroutines waiting for runtime)

### Changes

- Add `_wait_for_runtime_ready()` method that polls until the runtime is ready
- Remove the immediate error when `agent_state` is `LOADING`
- Wait up to 120 seconds for runtime to start (with 2s poll interval)
- Only show an error if the runtime takes too long to start
- **Add `runtime_wait_tracker.py` module** with:
  - Configurable max concurrent waits via `SLACK_MAX_CONCURRENT_RUNTIME_WAITS` env var (default: 100)
  - Prometheus metrics for monitoring:
    - `slack_runtime_wait_current` - Current number of waiting coroutines
    - `slack_runtime_wait_total` - Total wait operations started
    - `slack_runtime_wait_completed` - Completed operations by status (success/timeout/rejected)
  - `TooManyWaitingError` exception when concurrency limit is exceeded

### Before
```
Uh oh! There was an unexpected error starting the job :(
```
or
```
Conversation is still starting
```

### After
The user's message is delivered once the runtime is ready. No error is shown unless:
- The runtime takes more than 2 minutes to start
- Too many messages are already waiting (configurable limit)

## Testing

- ✅ Code compiles successfully
- ✅ Pre-commit hooks pass
- ✅ Unit tests added for `_wait_for_runtime_ready()` method
- ✅ Unit tests added for `runtime_wait_tracker` module
- Follows existing code style in the Slack integration

## Strategic Fix (Future Work)

The strategic fix mentioned in the issue involves:
1. Making Slack conversations optionally public when selecting a repository
2. Forking conversations when an unauthorized user tries to send a follow-up message
3. Building a proper message queue for handling messages when runtime is not yet available

This tactical fix provides immediate relief while the strategic solution is developed.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:e426ce2-nikolaik   --name openhands-app-e426ce2   docker.openhands.dev/openhands/openhands:e426ce2
```

### DIFF
--- a/.github/workflows/ghcr-build.yml
+++ b/.github/workflows/ghcr-build.yml
@@ -39,8 +39,7 @@ jobs:
         run: |
           if [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
             json=$(jq -n -c '[
-                { image: "nikolaik/python-nodejs:python3.12-nodejs22", tag: "nikolaik" },
-                { image: "ubuntu:24.04", tag: "ubuntu" }
+                { image: "nikolaik/python-nodejs:python3.12-nodejs22", tag: "nikolaik" }
               ]')
           else
             json=$(jq -n -c '[
@@ -149,6 +148,9 @@ jobs:
           push: true
           tags: ${{ env.DOCKER_TAGS }}
           platforms: ${{ env.DOCKER_PLATFORM }}
+          # Caching directives to boost performance
+          cache-from: type=registry,ref=ghcr.io/${{ env.REPO_OWNER }}/runtime:buildcache-${{ matrix.base_image.tag }}
+          cache-to: type=registry,ref=ghcr.io/${{ env.REPO_OWNER }}/runtime:buildcache-${{ matrix.base_image.tag }},mode=max
           build-args: ${{ env.DOCKER_BUILD_ARGS }}
           context: containers/runtime
           provenance: false
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,8 +7,8 @@ services:
     image: openhands:latest
     container_name: openhands-app-${DATE:-}
     environment:
-      - AGENT_SERVER_IMAGE_REPOSITORY=${AGENT_SERVER_IMAGE_REPOSITORY:-docker.openhands.dev/openhands/runtime}
-      - AGENT_SERVER_IMAGE_TAG=${AGENT_SERVER_IMAGE_TAG:-1.2-nikolaik}
+      - AGENT_SERVER_IMAGE_REPOSITORY=${AGENT_SERVER_IMAGE_REPOSITORY:-ghcr.io/openhands/agent-server}
+      - AGENT_SERVER_IMAGE_TAG=${AGENT_SERVER_IMAGE_TAG:-31536c8-python}
       #- SANDBOX_USER_ID=${SANDBOX_USER_ID:-1234} # enable this only if you want a specific non-root sandbox user but you will have to manually adjust permissions of ~/.openhands for this user
       - WORKSPACE_MOUNT_PATH=${WORKSPACE_BASE:-$PWD/workspace}
     ports:
--- a/enterprise/integrations/github/github_manager.py
+++ b/enterprise/integrations/github/github_manager.py
@@ -26,6 +26,7 @@ from integrations.utils import (
 from integrations.v1_utils import get_saas_user_auth
 from jinja2 import Environment, FileSystemLoader
 from pydantic import SecretStr
+from server.auth.auth_error import ExpiredError
 from server.auth.constants import GITHUB_APP_CLIENT_ID, GITHUB_APP_PRIVATE_KEY
 from server.auth.token_manager import TokenManager
 from server.utils.conversation_callback_utils import register_callback_processor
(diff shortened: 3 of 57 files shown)


saurya:
@OpenHands please fix the failing actions on PR #12329 at branch `fix/slack-runtime-unavailable-error`

saurya:
@OpenHands please fix the failing actions on PR #12329 at branch `fix/slack-runtime-unavailable-error`

enterprise/integrations/slack/slack_view.py
```diff
@@ -518,6 +522,54 @@ async def create_or_update_conversation(self, jinja: Environment) -> str:
...
+        while True:
+            elapsed = asyncio.get_event_loop().time() - start_time
+            if elapsed >= max_wait_seconds:
+                raise StartingConvoException(
+                    'The conversation is taking too long to start. Please try again later.'
+                )
+
+            agent_loop_info = await conversation_manager.maybe_start_agent_loop(
+                self.conversation_id, conversation_init_data, user_id
+            )
```
malhotra5, line 549: `maybe_start_agent_loop` in `SaasNestedConversationManager`  attempts to `_start_agent_loop` in a loop in the background

Here it seems like we might repeatedly attempt to start the agent loop, which should be fine if we have idempotency but I'm not sure would be good to verify 


Alternatively, we could also mark in redis an attempt to start the agent loop, call `maybe_start_agent_loop` exactly once, and wait for runtime ready 

LMK what you think!

saurya: The Redis approach makes more sense. @openhands-agent can you please change this code to use REdis and only call maybe_start_agent_loop once.

saurya: https://github.com/OpenHands/OpenHands/blob/b5a615aa6731a74e4d306eee1758a71567bcbf2b/enterprise/server/saas_nested_conversation_manager.py#L199

Looks like the Redis piece is already happening here, looks like we're ok on idempotency here. At least as far as I read that code.

saurya:
@OpenHands please fix the failing actions on PR #12329 at branch `fix/slack-runtime-unavailable-error`

enterprise/integrations/slack/slack_view.py
```diff
@@ -518,6 +516,54 @@ async def create_or_update_conversation(self, jinja: Environment) -> str:
 
         return self.conversation_id
 
+    async def _wait_for_runtime_ready(
+        self,
+        user_id: str,
+        conversation_init_data,
+        providers_set: list[ProviderType],
+        max_wait_seconds: int = 120,
```
saurya, line 526: I'm a bit worried that this number is too low and a bit random. From what I've seen, runtimes can take a bit longer than this to come up. In which case we'll get a bunch of these errors

saurya:
Adding @raymyers - I would like some advice:

- max_wait_seconds: The current limit of 120 seconds seems a bit tight. Is that number tracked somewhere in datadog or other metrics dashboards?
- the Prometheus metrics I added - are those useful and should we add more?

raymyers:
@saurya 
> * max_wait_seconds: The current limit of 120 seconds seems a bit tight. Is that number tracked somewhere in datadog or other metrics dashboards?

We can look into this when we talk tomorrow.

> * the Prometheus metrics I added - are those useful and should we add more?

Actually we've moved away from prometheus metrics in favor of structured log (use the `extra=` field and it goes in the json`) which can be queried. In the future we want to go to OpenTelemetry, which is now in the V1 agent sdk.

Issue #12088, opened 2025-12-18
Plugin Marketplace GUI


jpelletier1:
## Problem Statement

**Typical Customer Profile:** Enterprises (100s of devs, 1000s of employees)

Today, enterprise customers struggle with distributing the agents they create for recurring engineering tasks (like CVE remediation, dependency upgrades) to different teams. There is no "discoverability" layer in OpenHands that facilitates this, and therefore other teams are unable to reap the efficiency/productivity benefits of executing agents built by others.

Example current state:

* Platform team creates a Skill that proactively refactors Angular 1.2 --> React.
* Platform team publishes this to their `acmeco/.openhands` repo for others to use
* Frontend team is tasked with refactoring their app, but decidely builds their own Skill to do this.
* Result? Redundant work: Frontend team spends effort creating a Skill that was already built by Platform team

Future state:

* Frontend team is able to browse a directory of "Plugins" developed by the organization to solve recurring
* Frontend team is able to search for Plugins related to React and immediately use it
* Bonus: OpenHands can recommend specific Plugins based on the context/codebase of their repo

## What is a Plugin, and how does it compare to a Skill?

See: [https://github.com/OpenHands/OpenHands/issues/12085](<https://github.com/OpenHands/OpenHands/issues/12085>)

## Goals

**Goals**

* **Enable publishing** of Plugins to a shared, org-accessible repository.
* **Enable discovery** of Plugins through a dedicated in-product page.
* **Enable feedback loops** via comments and effectiveness ratings.

**Non-Goals**

* Public marketplace across multiple organizations (this is org-scoped).
* Complex workflow approvals beyond a simple publish/review flow (future).

## Key questions

* **Redundancy with Microagent Management UI:** Today we have a UI for sharing Organization-level microagents (stored in a `.openhands` repo within the Git organization). Do we remove this functionality in favor of this Organization-level Plugin Directory
  * No, this is separate. In general, we need to update the Microagent Management functionality to be more in-line with AgentSkills. Additionally, the current UX is not intuitive and needs to be re-evaluated.
* Would a Plugin be executed at the start of a conversation?
  * Yes it should support being able to execute the Plugin at the start of the conversation.
* Do you need to package MCP with the Plugin?
  * Not for initial scope, but in the future
* Do you need to package Secrets with the Plugin?
  * Not for initial scope, but in the future
* Should we prompt for specific input variables before the conversation starts (rather than have OpenHands interview the user for additional info)?
* Does this override the system prompt?
[...]

PR #12338, opened 2026-01-09
feat(app_server): Support starting conversations with remote plugins

jpshackelford:
## Summary

Extends the app-conversations API to accept a plugin specification and pass it to the agent server, which will fetch/load the plugin and start a conversation with the plugin's skills, hooks, and MCP configuration.

**Implements:** OpenHands/OpenHands#12321

## Context

This is part of the Plugin Directory feature (#12088). When a user clicks "Launch" in the Plugin Directory and fills out the parameter form, the frontend calls this API to start the conversation with the selected plugin.

### Architecture Note

Plugin fetching and loading happens **in the agent server** (which runs inside the sandbox), not in the app server. This is because:
- Plugins may contain scripts or hooks that need to execute in the sandbox
- Plugin MCP servers need to run inside the sandbox
- Skills may reference files that need to exist in the sandbox filesystem

The app server's role is to accept the plugin specification from the frontend and pass it through to the agent server via `StartConversationRequest`.

### 1. New `PluginSpec` Model (`app_conversation_models.py`)

`PluginSpec` extends the SDK's `PluginSource` class, inheriting its fields and validation (including path traversal protection for `repo_path`), and adds plugin configuration parameters:

```python
class PluginSpec(PluginSource):
    # Inherited from PluginSource:
    # source: str  - Plugin source: 'github:owner/repo', git URL, or local path
    # ref: str | None  - Optional branch, tag, or commit
    # repo_path: str | None  - Subdirectory path within the git repository
    
    # Added by PluginSpec:
    parameters: dict[str, Any] | None  # User-provided configuration values
    
    # Helper methods:
    @property
    def display_name(self) -> str: ...  # Extract friendly name from source
    def format_params_as_text(self, indent: str = '') -> str | None: ...  # Format for display
```

### 2. Extended `AppConversationStartRequest`

Added `plugins: list[PluginSpec] | None` field to allow API consumers to specify one or more plugins when starting conversations.

### 3. Updated `live_status_app_conversation_service.py`

- Added `_construct_initial_message_with_plugin_params()` helper method to incorporate plugin parameters into the initial conversation message
- Updated `_finalize_conversation_request()` to pass `plugin_source`, `plugin_ref`, and `plugin_repo_path` to the agent server's `StartConversationRequest`
- Updated `_build_start_conversation_request_for_user()` to accept and propagate the plugin specification

### 4. SDK Dependency Update

Updated `pyproject.toml` to use SDK PR OpenHands/OpenHands#1651 (`dc93dd124c01b4d5eb0d519afe6606dbbc07dc6d`) which adds:
- `plugin_source`, `plugin_ref`, and `plugin_repo_path` fields to `StartConversationRequest`
- Agent server plugin loading support

## API Usage Example

```json
POST /api/v1/app-conversations
{
  "initial_message": { "content": [{ "type": "text", "text": "Hello" }] },
  "plugins": [{
    "source": "github:owner/my-plugin",
    "ref": "v1.0.0",
    "repo_path": "plugins/my-plugin",
    "parameters": {
      "timeout": 30,
      "debug": true,
      "region": "us-west-2"
    }
  }]
}
```

> **Note:** The `repo_path` field is required when the plugin is in a subdirectory of the repository (common for marketplace-style repos). For plugins at the repository root, this field can be omitted.

### Parameters are for Non-Sensitive Configuration Only

The `parameters` field in `PluginSpec` is intended for **non-sensitive configuration values** that can safely appear in the conversation history. These values are formatted and included in the initial message so the agent has context about the user's configuration choices.

**Appropriate uses for `parameters`:**
- `timeout`, `max_retries`, `batch_size` - Numeric configuration
- `debug`, `verbose`, `dry_run` - Boolean flags  
- `region`, `environment`, `format` - String configuration

**Do NOT pass secrets via `parameters`:**
- API keys, tokens, passwords
- Database credentials
- Any sensitive authentication data

### How to Handle Secrets

Secrets should be passed through the existing `secrets` field in `StartConversationRequest`, which provides proper security handling:

```json
POST /api/v1/app-conversations
{
  "plugins": [{
    "source": "github:owner/my-plugin",
    "parameters": { "timeout": 30 }
  }],
  "secrets": {
    "MY_API_KEY": { "type": "StaticSecret", "value": "sk-..." }
  }
}
```

Skills within plugins should reference secrets as **environment variables**. The existing OpenHands secrets mechanism makes these available transparently to MCP servers and plugin code running in the sandbox.

### Note on Claude Plugin Compatibility

OpenHands plugins follow the **Claude Code plugin** structure (`plugin.json` in `.claude-plugin/` directory). Claude Code plugins do not have a built-in mechanism for defining input parameters or marking them as sensitive.

While **Claude Desktop Extensions** (`.mcpb` packages) do support a `user_config` schema with a `sensitive: true` flag in their `manifest.json`, this is a different packaging format for MCP server bundles, not the plugin format we're implementing.

For OpenHands, the recommended pattern is:
1. Use `parameters` for non-sensitive configuration values
2. Use `secrets` in `StartConversationRequest` for sensitive values
3. Skills should reference secrets via environment variables (e.g., `${MY_API_KEY}`)

## Tests Added

Added 27 new tests covering:
- `PluginSpec` model serialization/deserialization
- `PluginSpec.display_name` property (4 tests)
- `PluginSpec.format_params_as_text()` method (3 tests)
- `PluginSpec` inherited validation from `PluginSource` (1 test)
- `AppConversationStartRequest` with plugins field
- `_construct_initial_message_with_plugin_params` helper (6 tests)
- `_finalize_conversation_request` with plugin parameters (3 tests)
- `_build_start_conversation_request_for_user` with plugin (2 tests)

All tests pass.

## Dependencies

- **Depends on:** [software-agent-sdk#1651](https://github.com/OpenHands/software-agent-sdk/pull/1651) - Agent Server plugin loading support (which includes OpenHands/OpenHands#1647 - Plugin.fetch())

## Related Issues

- OpenHands/enterprise#17 - [PRD] OpenHands Plugins
- OpenHands/OpenHands#12088 - Plugin Marketplace GUI
- OpenHands/OpenHands#12322 - Frontend: Plugin Launch Dialog

@jpshackelford can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:4fe76ef-nikolaik   --name openhands-app-4fe76ef   docker.openhands.dev/openhands/openhands:4fe76ef
```


jpshackelford:
## Requested Changes: Pass `plugin_path` separately

### Context
During testing, we identified a bug where `get_resolved_source()` combines `source` and `path` into a single string (e.g., `github:owner/repo/plugins/sub-plugin`), which the Agent Server's SDK then rejects because `parse_plugin_source()` validates that GitHub shorthand has only one `/` in the repo path. To fix this, we need to pass `plugin_path` as a separate field rather than embedding it in the source.

### Changes Needed

**`openhands/app_server/app_conversation/live_status_app_conversation_service.py`** - Pass path separately:

Change from:
```python
return StartConversationRequest(
    # ... other fields ...
    plugin_source=plugin.get_resolved_source() if plugin else None,
    plugin_ref=plugin.ref if plugin else None,
)
```

To:
```python
return StartConversationRequest(
    # ... other fields ...
    plugin_source=plugin.source if plugin else None,  # Don't use get_resolved_source()
    plugin_ref=plugin.ref if plugin else None,
    plugin_path=plugin.path if plugin else None,  # NEW: pass path separately
)
```

**`openhands/app_server/app_conversation/app_conversation_models.py`** - Remove `get_resolved_source()`:

The `get_resolved_source()` method in `PluginSpec` is no longer needed and should be removed:

```python
class PluginSpec(BaseModel):
    source: str = Field(...)
    ref: str | None = Field(...)
    path: str | None = Field(...)
    parameters: dict[str, Any] | None = Field(...)
    
    # DELETE this method:
    # def get_resolved_source(self) -> str:
    #     ...
```

**Update tests** - Tests that assert `plugin_source` contains the path should be updated to expect the path in `plugin_path` instead.

This change requires corresponding changes in `software-agent-sdk` PRs OpenHands/OpenHands#1647 and OpenHands/OpenHands#1651 to accept and handle the separate `plugin_path` field.

See https://github.com/OpenHands/OpenHands/issues/12087#issuecomment-3733464400 for the full API design showing consistency across all layers.

jpshackelford:
## ✅ Plugin Loading Test Results

I tested this PR using the staging preview server and can confirm **plugin loading works correctly**.

### Test Environment

- **Staging URL:** `https://ohpr-12338-preview.staging.all-hands.dev`
- **Test Plugin:** [city-weather](https://github.com/jpshackelford/openhands-sample-plugins/tree/main/plugins/city-weather) from `jpshackelford/openhands-sample-plugins`

### How to Test

Reviewers can test plugin loading using the sample plugins repository and demo scripts:

1. **Clone the test repository:**
   ```bash
   git clone https://github.com/jpshackelford/openhands-sample-plugins.git
   cd openhands-sample-plugins
   ```

2. **Set environment variables:**
   ```bash
   export STAGING_URL="https://ohpr-12338-preview.staging.all-hands.dev"
   export API_KEY="your-api-key-here"  # Get from staging environment
   ```

3. **Run the demo script:**
   ```bash
   ./doc/demo_weather_plugin.sh "Paris"
   ```

   Or manually create a conversation with the plugin:
   ```bash
   curl -X POST "${STAGING_URL}/api/v1/app-conversations" \
     -H "Authorization: Bearer ${API_KEY}" \
     -d '{
       "initial_message": {
         "content": [{"type": "text", "text": "/city-weather:now Paris"}]
       },
       "plugins": [{
         "source": "github:jpshackelford/openhands-sample-plugins",
         "ref": "main",
         "repo_path": "plugins/city-weather"
       }]
     }'
   ```

4. **Poll for completion and verify:**
   ```bash
   curl -s "${STAGING_URL}/api/v1/app-conversations/search" \
     -H "Authorization: Bearer ${API_KEY}" | jq '.items[0]'
   ```

### Test Results

I created a conversation with `/city-weather:now Paris` and verified:

✅ **Plugin spec accepted** - The `plugins` array in the request was properly processed  
✅ **Plugin loaded** - Agent used the plugin's skill instructions  
✅ **Correct tools used** - Agent used `terminal` with `curl` commands (as specified by `allowed-tools: Bash(curl:*)`)  
✅ **Expected behavior** - Agent called geocoding API, then weather API, and formatted the response per plugin instructions

**Agent Actions:**
```
1. curl -s "https://geocoding-api.open-meteo.com/v1/search?name=Paris&count=1..."
2. curl -s "https://api.open-meteo.com/v1/forecast?latitude=48.85341&longitude=2.3488..."
```

**Agent Response:**
```markdown
## 🌤️ City Weather Report for Paris, France

- **Current Time:** January 22, 2026 at 11:30 PM (Europe/Paris)
- **Temperature:** 45.5°F / 7.5°C
- **Current Precipitation:** 0.00 mm

### Precipitation Forecast (Next 4 Hours)
| Time | Probability |
|------|-------------|
| 11:00 PM | 0% |
| 12:00 AM | 0% |
...
```

### Documentation

For more details on testing and troubleshooting, see:
- [Plugin Capability Status](https://github.com/jpshackelford/openhands-sample-plugins/blob/main/doc/plugin-capability-status.md)
- [Testing & Troubleshooting Guide](https://github.com/jpshackelford/openhands-sample-plugins/blob/main/doc/testing-troubleshooting.md)
- [Demo Script](https://github.com/jpshackelford/openhands-sample-plugins/blob/main/doc/demo_weather_plugin.sh)

openhands/app_server/app_conversation/app_conversation_models.py
```diff
@@ -24,6 +24,34 @@ class AgentType(Enum):
     PLAN = 'plan'
 
 
+class PluginSpec(BaseModel):
```
malhotra5: NIT: could we inherit the `PluginSource` definition from SDK and add any custom fields as necessary?

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -954,6 +956,93 @@ def _update_agent_with_llm_metadata(
...
+                for name, params in params_by_plugin.items()
+            )
+            plugin_params_message = (
+                f'\n\nPlugin Configuration Parameters:\n{params_text}'
+            )
+
+        if initial_message is None:
+            # Create a new message with just the plugin parameters
+            return SendMessageRequest(
+                content=[TextContent(text=plugin_params_message.strip())],
```
malhotra5, line 1009: In the PR description its mentioned that parameters can include `api_key`; not sure I'm understanding the purpose of plugin params

Also it seems like we serialize them in the initial message request? Could that leak sensitive info (e.g api_key passed in the params)

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -954,6 +956,93 @@ def _update_agent_with_llm_metadata(
...
+        all_params: list[tuple[str, str, Any]] = []  # (plugin_name, key, value)
+        for plugin in plugins:
+            if plugin.parameters:
+                plugin_name = (
+                    plugin.source.split('/')[-1]
+                    if '/' in plugin.source
+                    else plugin.source
+                )
+                for key, value in plugin.parameters.items():
+                    all_params.append((plugin_name, key, value))
```
malhotra5: NIT: maybe we could move some of the serialization logic to a method inside `PluginSpec`

openhands/app_server/app_conversation/app_conversation_models.py
jpshackelford: Done in 419ff370f. `PluginSpec` now inherits from `PluginSource`, which:
- Eliminates field duplication
- Inherits the `repo_path` validation (prevents path traversal attacks)
- Keeps a single source of truth for the core plugin source fields

Added a test to verify the inherited validation works correctly.

We keep `PluginSpec` as a separate class because `parameters` is an app-server API concern (runtime configuration at conversation start), while `PluginSource` is focused on plugin identity/location.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
jpshackelford: Good catch — the `api_key` example was misleading. I've updated the PR description to clarify:

**`parameters` is for non-sensitive configuration only** (e.g., `timeout`, `debug`, `region`). These values are intentionally included in the conversation so the agent has context about the user's configuration.

**Secrets should use the existing `secrets` field** in `StartConversationRequest`, which provides proper secure handling via the SDK's `SecretSource` mechanism. Skills within plugins should reference secrets as environment variables.

Note: Claude Code plugins don't have a built-in parameter/secrets mechanism. Claude Desktop Extensions (`.mcpb`) do support `sensitive: true` in their `user_config`, but that's a different packaging format. For now, the recommended pattern is `parameters` for config + `secrets` for sensitive values.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
jpshackelford: Done in 419ff370f. Added two helper methods to `PluginSpec`:
- `display_name` property — extracts a friendly name from the source (e.g., `github:owner/repo` → `repo`)
- `format_params_as_text(indent)` — formats parameters as a readable text block

The service method now uses these instead of inline logic. Added tests for both.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
malhotra5: Ah gotcha, I wonder if this should be added to the system prompt or as a skill extension rather than as part of the initial user message - it seems to be agent specific instructions but not quite a user prompt?

The other behavior it seems we'll see is starting a fresh conversation with no prompt will automatically kick the agent into a working state (since an initial message will be created using the plugin params)


WDYT? I'm not quite caught up with plugins so maybe that is the intended behavior

jpshackelford: We know that there is probably more we need to do with the way parameter passing should work. We have clients that will be able to advise us more on this once they've played with it. The general idea is that skills may have these parameters included as tokens and the agent would then see the values for these and substitute them into the skill files. We thought about doing actual token substitution but we wanted to experiment to see of the LLM would do that on its own with real world skills.

The initial prompt may be as minimal as a slash command. Starting without a prompt is a legitimate use case. We have two with different design partners:

1. Initial prompt is slash command and any parameters. User fires off the skill and waits for the result with no direct visibility or interaction w/ OpenHands UI.

2. Initial prompt is empty but loads a set of plugins with skills, mcp config, etc. User interacts with conversation customized with this initial context via OpenHands UI. In other words it is just like normal OpenHands UX except agent is prepared to a help a user with a specific use case.

malhotra5: Sounds good! Let the users decide what they prefer!

malhotra5 (approved):
LGTM! I think future step could be verifying that resuming a conversation without the same plugins doesn't result in the agent getting confused (or maybe the plugin configuration is permanent for the conversation once its created)

jpshackelford:
I have just tested resumes on staging by creating a conversation with a plugin via the REST API. Shutting down the runtime. Waking the runtime back up and then continuing the conversation with a second slash command invoking the plugin. All works as expected.

PR #12345, opened 2026-01-09
feat(frontend): improve public share menu behavior

tofarr:
## Summary of PR

This PR improves the user experience of the public share menu item in the conversation context menu:

1. **Clicking the public share toggle no longer closes the menu** - Users can now toggle the public share status and see the state change immediately without the menu disappearing
2. **Link icon now opens the public share URL in a new tab** - Instead of copying the URL to clipboard, clicking the link icon opens the shared conversation in a new browser tab

image

### Changes Made:
- `use-conversation-name-context-menu.ts`: Modified `handleTogglePublic` to not close the menu, replaced `handleCopyShareLink` with `getShareUrl()` function
- `conversation-name-context-menu.tsx`: Changed the icon button from a `<button>` that copies to clipboard to an `<a>` anchor that opens in a new tab
- `conversation-name.tsx`: Updated to use `getShareUrl` and pass `shareUrl` prop instead of `onCopyShareLink`

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

N/A

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:814ecab-nikolaik   --name openhands-app-814ecab   docker.openhands.dev/openhands/openhands:814ecab
```


frontend/src/components/features/conversation/conversation-name-context-menu.tsx
```diff
@@ -209,16 +209,18 @@ export function ConversationNameContextMenu({
...
               <span>{t(I18nKey.CONVERSATION$SHARE_PUBLICLY)}</span>
             </div>
-            {conversation?.public && onCopyShareLink && (
-              <button
-                type="button"
-                data-testid="copy-share-link-button"
-                onClick={onCopyShareLink}
+            {conversation?.public && shareUrl && (
+              <a
+                data-testid="open-share-link-button"
```
hieptl: Hello @tofarr,

Could you please add some tests to cover your changes? I noticed that you’ve already added a data-testid here, which I assume was intended to support writing tests.

Thank you! 🙏

frontend/src/components/features/conversation/conversation-name-context-menu.tsx
```diff
@@ -209,16 +209,18 @@ export function ConversationNameContextMenu({
               />
               <span>{t(I18nKey.CONVERSATION$SHARE_PUBLICLY)}</span>
             </div>
-            {conversation?.public && onCopyShareLink && (
-              <button
```
hieptl, line 213: **question:** In the past, if I wanted to copy a conversation link to share with my teammates, I could simply click this button.

Now, it seems I need to open the shared link in a new tab and then copy it from the browser.

Would it be possible to keep both options—one button to copy the shared link directly, and another to open it in a new tab? I think having both would make sharing more convenient.

Thank you! 🙏

frontend/src/components/features/conversation/conversation-name.tsx
```diff
@@ -187,9 +187,9 @@ export function ConversationName() {
                     ? handleTogglePublic
                     : undefined
                 }
-                onCopyShareLink={
+                shareUrl={
                   ENABLE_PUBLIC_CONVERSATION_SHARING()
-                    ? handleCopyShareLink
+                    ? getShareUrl()
```
hieptl: **suggestion:** We should avoid calling getShareUrl() here and then passing its returned value as a prop, since it runs on every render. If this function is expensive, it might be a good idea to memoize it. Thank you! 🙏 

```ts
const shareUrl = useMemo(() => {
  if (!conversationId) {
    return "";
  }
  return `${window.location.origin}/shared/conversations/${conversationId}`;
}, [conversationId, onContextMenuToggle]);
```

```ts
<YouComponent shareUrl={shareUrl} />
```

hieptl:
Hello @tofarr,

I have incorporated my suggestions into this pull request. Please refer to the video below to review the resulting output.

https://github.com/user-attachments/assets/9649c461-32bd-4533-a864-bf797d48d7c1

Thank you very much! 🙏

PR #12347, opened 2026-01-09
Fix for critical regression where conversations will not start in OSS

tofarr:
## Summary of PR

Fixes a critical regression where conversations will not start in OSS.

### Problem
When a sandbox container is starting up, the app server inside it may not be immediately ready to respond. The previous code would mark the sandbox status as `ERROR` immediately if any exception occurred when checking the app server health, even during the expected startup period.

### Solution
This PR adds a startup grace period (`STARTUP_GRACE_SECONDS = 15`) during which the sandbox is not marked as `ERROR` if the app server is not responding. This allows the container time to fully initialize before being considered in an error state.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

### DIFF
--- a/openhands/app_server/sandbox/docker_sandbox_service.py
+++ b/openhands/app_server/sandbox/docker_sandbox_service.py
@@ -539,4 +539,5 @@ class DockerSandboxServiceInjector(SandboxServiceInjector):
                 httpx_client=httpx_client,
                 max_num_sandboxes=self.max_num_sandboxes,
                 extra_hosts=self.extra_hosts,
+                startup_grace_seconds=self.startup_grace_seconds,
             )


openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -39,6 +39,7 @@
 _logger = logging.getLogger(__name__)
 SESSION_API_KEY_VARIABLE = 'OH_SESSION_API_KEYS_0'
 WEBHOOK_CALLBACK_VARIABLE = 'OH_WEBHOOKS_0_BASE_URL'
+STARTUP_GRACE_SECONDS = 15
```
malhotra5, line 42: Hmm is there a more direct source of truth on whether the container is still starting up?

Can this timeout depend on the user's compute?

tofarr: This is in docker only (So OSS), and I've made it configurable - If 15 seconds is inappropriate, it can be overridden with environment variables. e.g.:
```
OH_SANDBOX_KIND=DockerSandboxServiceInjector
OH_SANDBOX_STARTUP_GRACE_SECONDS=60
```

malhotra5: Cool! Maybe we can include these instructions in the error message on timeout

PR #12352, opened 2026-01-10
feat(frontend): org permission guards

### LINKED ISSUE #12276: Extend frontend permission guards
> [!NOTE]
> Make sure you’ve read the description on the parent issue before working on this ticket: https://github.com/OpenHands/OpenHands/issues/12136

## Problem

The [`ALL-2596/org-support`](https://github.com/OpenHands/OpenHands/tree/ALL-2596/org-support/frontend) branch has a permission system defined in [`permissions.ts`](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/utils/org/permissions.ts) with three roles (owner, admin, member), but it is only enforced in two routes: Organization settings and Organization Members. All other settings routes (Secrets, MCP, LLM, Billing, etc.) have no role-based access control, allowing any authenticated user to access and modify settings regardless of their role.

## Expected Behavior

The permission matrix defines the following access controls:

| Feature | Member | Admin | Owner | Permission key |
| --- | --- | --- | --- | --- |
| Settings > Secrets | ✅ Full | ✅ Full | ✅ Full | `manage_secrets` |
| Settings > MCP | ✅ Full | ✅ Full | ✅ Full | `manage_mcp` |
| Settings > Integrations | ✅ Full | ✅ Full | ✅ Full | `manage_integrations` |
| Settings > Application | ✅ Full | ✅ Full | ✅ Full | `manage_application_settings` |
| Settings > API Keys | ✅ Full | ✅ Full | ✅ Full | `manage_api_keys` |
| Settings > LLM | 👁️ View | ✅ Full | ✅ Full | `view_llm_settings`, `edit_llm_settings` |
| Settings > Billing | ❌ None | ✅ Full | ✅ Full | `view_billing` |
| Invite members | ❌ None | ✅ Full | ✅ Full | `invite_user_to_organization` |
| Assign member/admin roles | ❌ None | ✅ Full | ✅ Full | `change_user_role:member`, `change_user_role:admin` |
| Change org name | ❌ None | ❌ None | ✅ Full | `change_organization_name` |
| Delete organization | ❌ None | ❌ None | ✅ Full | `delete_organization` |
| Assign owner role | ❌ None | ❌ None | ✅ Full | `change_user_role:owner` |
| Add credits | ❌ None | ✅ Full | ✅ Full | `add_credits` |

## Current Behavior

- LLM settings allow all roles to edit (should be view-only for members).
- Billing route has no role check and is accessible to members.
- Navigation items are shown to all users regardless of permissions, causing redirects after clicking.
- Settings routes for Secrets, MCP, Application, Integrations, and API Keys have no permission guards at all, even though they currently allow all roles.

## Acceptance Criteria

> [!IMPORTANT]
> This work should follow TDD. Write failing tests first for each permission scenario (member accessing billing, member editing LLM settings, etc.), then implement the guards to make them pass. The existing tests in manage-organization-members.test.tsx demonstrate the pattern of mocking useMe with different roles and asserting on UI visibility.

- **All new permission logic has corresponding unit tests covering each role's access to each feature.**
- The permission types in `permissions.ts` are extended to cover all settings features (e.g., `manage_secrets`, `manage_mcp`, `view_llm_settings`, `edit_llm_settings`, `view_billing`).
- Each role's permission array is updated to match the matrix, with members receiving view-only LLM access and no billing access.
- All settings routes check the user's role against `rolePermissions` before rendering.
- The LLM settings page disables the save button and form inputs for members.
- The billing route redirects members to an accessible page.
- The navigation hook filters out menu items the user does not have permission to access.
- **BONUS (but not required for this issue)**: A reusable `usePermission` hook is created for consistent permission checking across components.
- Existing tests in `manage-organization-members.test.tsx` and `manage-org.test.tsx` continue to pass.

## Relevant Files

[frontend/src/utils/org/permissions.ts](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/utils/org/permissions.ts) defines the permission types and role mappings.
[frontend/src/types/org.ts](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/types/org.ts) defines the OrganizationUserRole type.
[frontend/src/routes/llm-settings.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/routes/llm-settings.tsx) needs view-only mode for members.
[frontend/src/routes/secrets-settings.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/routes/secrets-settings.tsx) needs permission guard.
[frontend/src/routes/mcp-settings.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/routes/mcp-settings.tsx) needs permission guard.
[frontend/src/routes/app-settings.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/routes/app-settings.tsx) needs permission guard.
[frontend/src/routes/settings.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/routes/settings.tsx) handles route-level redirects.
[frontend/src/hooks/use-settings-nav-items.ts](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/hooks/use-settings-nav-items.ts) needs to filter nav items by permission.
[frontend/src/constants/settings-nav.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/constants/settings-nav.tsx) defines the navigation items.
[frontend/src/routes/manage-org.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/routes/manage-org.tsx) is an example of correct permission usage.
[frontend/src/routes/manage-organization-members.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/src/routes/manage-organization-members.tsx) is an example of correct permission usage.
[frontend/tests/routes/manage-organization-members.test.tsx](https://github.com/OpenHands/OpenHands/blob/ALL-2596/org-support/frontend/__tests__/routes/manage-organization-members.test.tsx) shows the testing pattern to follow.

HeyItsChloe:
## Summary of PR

- Updated `Permission.ts` to include all new permission guards
- Updated Settings Nav Bar to render billing for admin or owner users only
- Updated existing unit tests to account for conditionally rendering "billing" in the nav bar based on user role
- Added & updated hooks & utils to find user permissions based on user role
- Added tests for all new funcs
- _Attempted_ name standardization: 
       - User: all users (admin | member | owner)
       - Member: "member" (i.e OrganizationUserRole)
- Redirect members away from `settings/billing`

https://github.com/user-attachments/assets/d57f4323-4c37-4f9c-ab9f-191c1d6c9732

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #12276

## Release Notes

Added conditional rendering to `settings/billing` and settings nav bar billing items based on userRoles and permissions

### DIFF
--- a/frontend/src/routes/llm-settings.tsx
+++ b/frontend/src/routes/llm-settings.tsx
@@ -29,7 +29,6 @@ import { DEFAULT_SETTINGS } from "#/services/settings";
 import { getProviderId } from "#/utils/map-provider";
 import { DEFAULT_OPENHANDS_MODEL } from "#/utils/verified-models";
 import { useMe } from "#/hooks/query/use-me";
-import { usePermission } from "#/hooks/organizations/use-permissions";
 
 interface OpenHandsApiKeyHelpProps {
   testId: string;
@@ -65,16 +64,16 @@ function OpenHandsApiKeyHelp({ testId }: OpenHandsApiKeyHelpProps) {
 function LlmSettingsScreen() {
   const { t } = useTranslation();
   const [searchParams, setSearchParams] = useSearchParams();
-  const { data: user } = useMe();
-  const { hasPermission } = usePermission(user?.role || "member");
 
   const { mutate: saveSettings, isPending } = useSaveSettings();
 
   const { data: resources } = useAIConfigOptions();
   const { data: settings, isLoading, isFetching } = useSettings();
   const { data: config } = useConfig();
+  const { data: me } = useMe();
 
-  const isUserRoleViewOnly = !hasPermission("edit_llm_settings");
+  // Determine if user is read-only (users can only view, owners and admins can edit)
+  const isReadOnly = me?.role !== "owner" && me?.role !== "admin";
 
   const [view, setView] = React.useState<"basic" | "advanced">("basic");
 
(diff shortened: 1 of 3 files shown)


frontend/src/hooks/use-settings-nav-items.ts
```diff
@@ -1,11 +1,50 @@
...
+import { OrganizationMember, OrganizationUserRole } from "#/types/org";
+import { organizationService } from "#/api/organization-service/organization-service.api";
 
-export function useSettingsNavItems() {
+/**
+ * Fetch active organization member.
+ * - return cached data if present
+ * - fetch from API if cache is empty or stale
+ */
+export function useActiveOrganizationMember(orgId?: string) {
```
hieptl: **suggestion:** I believe it would be beneficial to move this hook into a separate file. Additionally, it looks like we already have a similar hook named useMe, which we may be able to reuse to avoid duplication. Thank you! 🙏

frontend/src/hooks/use-settings-nav-items.ts
```diff
@@ -1,11 +1,50 @@
...
+ * Build Settings navigation items based on:
+ * - app mode (saas / oss)
+ * - feature flags
+ * - active user's role
+ * @returns Settings Nav Items
+ */
+export function useSettingsNavItems(): SettingsNavItem[] {
   const { data: config } = useConfig();
 
+  const selectedOrgId = getSelectedOrganizationIdFromStore() ?? undefined;
```
hieptl: **question:** In the selected-organization-store.ts file, I noticed that organizationId is typed as string | null. Is there a particular reason we set it to undefined when there is no value? Thank you! 🙏

frontend/src/hooks/use-settings-nav-items.ts
```diff
@@ -1,11 +1,50 @@
...
+  const { data: user } = useActiveOrganizationMember(selectedOrgId);
+  const userRole: OrganizationUserRole = user?.role ?? "member";
```
hieptl, line 21: **question:** If the user’s role does not exist, would it make more sense to redirect them to an error page instead of treating them as an organization member?

frontend/src/hooks/use-settings-nav-items.ts
```diff
@@ -1,11 +1,50 @@
...
+
   const shouldHideLlmSettings = !!config?.FEATURE_FLAGS?.HIDE_LLM_SETTINGS;
-  const shouldHideBilling = !!config?.FEATURE_FLAGS?.HIDE_BILLING;
+  const shouldHideBilling =
+    !!config?.FEATURE_FLAGS?.HIDE_BILLING || userRole === "member";
```
hieptl: **question:** We already have a hook for checking user permissions. Would it make sense to use it here as well, for example:

```ts
const shouldHideBilling =
  !!config?.FEATURE_FLAGS?.HIDE_BILLING || hasPermission("user:billing");
```

Using the permission hook could make this logic more flexible and easier to maintain. With the current approach, if we decide in the future to allow the member role to access the billing page, we would need to update this line and open another pull request.

Thank you! 🙏

frontend/src/utils/org/permission-checks.ts
```diff
@@ -0,0 +1,82 @@
...
+import { getMeFromQueryClient } from "../query-client-getters";
+import { Permission, rolePermissions } from "./permissions";
+import { queryClient } from "#/query-client-config";
+
+/**
+ * Get the active organization user.
+ * Reads from cache first, fetches if missing.
+ * @returns OrganizationMember
+ */
+export const getActiveOrganizationUser = async (): Promise<
```
hieptl, line 13: **question:** It appears that this function may be redundant, as we already have the useMe hook in place, and React Query can handle returning cached data as well as refetching it automatically based on the defined conditions.

@amanape, please feel free to let me know if my understanding is not correct.

Thank you! 🙏

frontend/src/utils/org/permission-checks.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+/**
+ * Check if active user has permission to change another member's role
+ * @param user OrganizationMember
+ * @param memberId Id of org member whose role would change
+ * @param memberRole Role of org member whose role would change
+ * @returns boolean
+ */
+export const doesUserHavePermissionToAssignRoles = (
```
hieptl: **question:** We already have the use-permission hook included in this pull request. Could you please help clarify why this function is still needed? Thank you! 🙏

frontend/src/hooks/use-settings-nav-items.ts
HeyItsChloe: Ah yes, I can reuse useMe() instead.

frontend/src/hooks/use-settings-nav-items.ts
HeyItsChloe: Agreed!

frontend/src/utils/org/permission-checks.ts
HeyItsChloe: Here, I couldn't call the useMe() hook because this function is not a React component. So this logic may be a bit duplicative of the useMe() hook.

hieptl: Yes, I was referring to the component. Instead of calling getActiveOrganizationUser, you could consider using the useMe hook.

For example, in the settings.tsx file, the implementation could be updated as follows.

From:

```ts
const user: OrganizationMember | undefined =
    await getActiveOrganizationUser();
```

To:

```ts
const { data: user } = useMe();
```

This would help keep the data access consistent and leverage the existing hook. Please let me know if I am missing any context or if there is a specific reason the current approach is preferred.

Thank you! 🙏

frontend/src/hooks/use-settings-nav-items.ts
hieptl: Hello @HeyItsChloe,

I noticed that this thread was marked as resolved. May I ask if that was intentional? If possible, could you please help answer my question first before closing the thread?

Thank you! 🙏

frontend/src/hooks/use-settings-nav-items.ts
HeyItsChloe: IMO - I think in some cases we would not want to redirect to an error page. If the user has no personal account prior to being invited to an org, they may need to be redirected to the error page. But if the user has their own personal account, I think they should default to that account as a member. LMK if this is handled differently. 

cc @amanape

frontend/src/hooks/use-settings-nav-items.ts
HeyItsChloe: Ah, the type should be null. But I removed this var and replaced it with a call to useMe()

frontend/src/utils/org/permission-checks.ts
HeyItsChloe: This is redundant, I created this func to check if a user has permission to change a target user role. But that logic is happening in `getAvailableRolesAUserCanAssign()`. So this func is just a duplicate of the usePerms hook.

frontend/src/hooks/use-settings-nav-items.ts
hieptl: Redirecting users to an error page is only one possible option, as there is an important use case to consider:
- User A belongs to Organization A and is already logged in to OpenHands Cloud.
- The owner of Organization A deletes User A, but User A still has an active session and remains logged in.
- In our current handling, if the user’s role cannot be resolved for any reason (for example, due to a network error or because the user has been deleted), we treat them as having the member role by default.
- In this scenario, User A may still be able to access the organization’s data, which introduces a potential security concern.

We should carefully consider how to handle cases where a user’s role cannot be validated, to ensure that access is properly restricted and organization data remains protected.

Thank you! 🙏

frontend/src/utils/org/permission-checks.ts
hieptl: Ah, I see now—you used this function in clientLoader(). That clarifies everything. I will go ahead and resolve this thread. Thank you! 🙏

frontend/src/routes/manage-org.tsx
```diff
@@ -233,20 +233,21 @@ function ManageOrg() {
...
-  const canDeleteOrg =
-    !!me && rolePermissions[me.role].includes("delete_organization");
-  const canAddCredits =
-    !!me && rolePermissions[me.role].includes("add_credits");
-  const isBillingHidden = config?.FEATURE_FLAGS?.HIDE_BILLING;
+  const canChangeOrgName = !!me && hasPermission("change_organization_name");
+  const canDeleteOrg = !!me && hasPermission("delete_organization");
+  const canAddCredits = !!me && hasPermission("add_credits");
+  const isBillingHidden =
+    config?.FEATURE_FLAGS?.HIDE_BILLING || role === "member";
```
hieptl: **suggestion:**  have the same suggestion as the one I mentioned in the comment below. Thank you! 🙏 

https://github.com/OpenHands/OpenHands/pull/12352#discussion_r2683098267

frontend/__tests__/components/features/user/user-context-menu.test.tsx
```diff
@@ -48,6 +50,34 @@ vi.mock("react-router", async (importActual) => ({
...
+  vi.spyOn(useMeModule, "useMe").mockReturnValue({
+    data: user,
+    status: "success",
+    isLoading: false,
+    isError: false,
+    isSuccess: true,
+    refetch: vi.fn(),
+    error: null,
+  } as any);
+};
```
amanape:
```suggestion
const createMockUser = (
  overrides: Partial<OrganizationMember> = {},
): OrganizationMember => ({
  org_id: "org-1",
  user_id: "user-1",
  email: "test@example.com",
  role: "member",
  llm_api_key: "",
  max_iterations: 100,
  llm_model: "gpt-4",
  llm_api_key_for_byor: null,
  llm_base_url: "",
  status: "active",
  ...overrides,
});

const seedActiveUser = (user: Partial<OrganizationMember>) => {
  useSelectedOrganizationStore.setState({ organizationId: "org-1" });
  vi.spyOn(organizationService, "getMe").mockResolvedValue(
    createMockUser(user),
  );
};
```

I’m suggesting two things here:

1. **Mock the service layer instead of the hook**: It enables a more realistic scenario and avoids having to mock the other `useQuery` return values. Example: https://github.com/OpenHands/OpenHands/blob/ece7e2dd39820065d383afd66383dc3bf28a8c5b/frontend/__tests__/routes/app-settings.test.tsx#L28-L34

2. **Direct store state setting**: We have a `__mocks__/zustand.ts` that automatically resets stores between tests and makes testing with Zustand much easier. Example: https://github.com/OpenHands/OpenHands/blob/ed7adb335ca7e65f19418ca1f337f88c7b7f9c24/frontend/__tests__/components/chat/chat-interface.test.tsx#L193-L198

frontend/__tests__/hooks/use-permission.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+    });
+
+    it("returns false when the role does not have the permission", () => {
+      const { hasPermission } = setup("member");
+
+      expect(hasPermission("invite_user_to_organization")).toBe(false);
+    });
+  });
+
+  describe("hasAnyPermission", () => {
```
amanape: Do we currently have a use case for this? It doesn’t look like we’re actually using `hasAnyPermission` anywhere. Maybe it is safe to remove?

frontend/__tests__/hooks/use-permission.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+      const permissions: Permission[] = [
+        "delete_organization",
+        "change_organization_name",
+      ];
+
+      expect(hasAnyPermission(permissions)).toEqual([]);
+    });
+  });
+
+  describe("hasAllPermissions", () => {
```
amanape: Same comment here

frontend/__tests__/hooks/use-settings-nav-items.test.tsx
```diff
@@ -17,13 +19,44 @@ const mockConfig = (appMode: "saas" | "oss", hideLlmSettings = false) => {
...
+  vi.spyOn(useMeModule, "useMe").mockReturnValue({
+    data: user,
+    status: "success",
+    isLoading: false,
+    isError: false,
+    isSuccess: true,
+    refetch: vi.fn(),
+    error: null,
+  } as any);
+};
```
amanape:
```suggestion
const createMockUser = (
  overrides: Partial<OrganizationMember> = {},
): OrganizationMember => ({
  org_id: "org-1",
  user_id: "user-1",
  email: "test@example.com",
  role: "member",
  llm_api_key: "",
  max_iterations: 100,
  llm_model: "gpt-4",
  llm_api_key_for_byor: null,
  llm_base_url: "",
  status: "active",
  ...overrides,
});

const seedActiveUser = (user: Partial<OrganizationMember>) => {
  useSelectedOrganizationStore.setState({ organizationId: "org-1" });
  vi.spyOn(organizationService, "getMe").mockResolvedValue(
    createMockUser(user),
  );
};
```

See my comment for `frontend/__tests__/components/features/user/user-context-menu.test.tsx` for more context. Note that you might also have to add:

```ts
vi.mock("react-router", () => ({
  useRevalidator: () => ({ revalidate: vi.fn() }),
}));
```

frontend/__tests__/routes/manage-organization-members.test.tsx
```diff
@@ -821,4 +824,31 @@ describe("Manage Organization Members Route", () => {
       },
     );
   });
+
+  describe("getAvailableRolesAUserCanAssign", () => {
```
amanape: This test suite looks out of place. We should probably move this to a new `permission-checks.test.ts` or a dedicated `get-available-roles-a-user-can-assign.test.ts`.

frontend/__tests__/routes/settings-with-payment.test.tsx
```diff
@@ -5,24 +5,27 @@ import { createRoutesStub } from "react-router";
...
-        EMAIL_VERIFIED: true, // Mock email as verified to prevent redirection
-      },
+      data: { EMAIL_VERIFIED: true },
       isLoading: false,
     }),
   };
 });
 
-// Mock the i18next hook
+// Mock the i18next
```
amanape:
```suggestion
// Mock the i18next hook
```

frontend/src/routes/manage-organization-members.tsx
```diff
@@ -57,42 +58,16 @@ function ManageOrganizationMembers() {
...
-    if (userPermissions.includes("change_user_role:admin")) {
-      availableRoles.push("admin");
-    }
-    if (userPermissions.includes("change_user_role:member")) {
-      availableRoles.push("member");
-    }
+  const availableRolesToChangeTo = React.useMemo(
+    () => getAvailableRolesAUserCanAssign(rolePermissions[currentUserRole]),
+    [currentUserRole],
+  );
```
amanape:
```suggestion
  const availableRolesToChangeTo = getAvailableRolesAUserCanAssign(
    rolePermissions[currentUserRole],
  );
```

This is trivial computation - iterating over 3 items max. The `useMemo` adds unnecessary overhead. Removing it would be cleaner.

frontend/__tests__/routes/settings-with-payment.test.tsx
```diff
@@ -51,6 +54,15 @@ vi.mock("react-i18next", async () => {
...
 
+// Mock permission checks
+vi.mock("#/utils/org/permission-checks", () => ({
+  getActiveOrganizationUser: vi.fn(),
+}));
+vi.mocked(getActiveOrganizationUser).mockResolvedValue({
+  user_id: "u1",
+  role: "admin",
+} as OrganizationMember);
+
```
amanape:
```suggestion
```

We can probably remove this and call this instead in each test that requires it:

```ts
vi.spyOn(
  PermissionChecksModule,
  “getActiveOrganizationUser”,
  // @ts-expect-error - only return relevant data
).mockResolvedValue({
  user_id: “u1”,
  role: “admin”,
});
```

frontend/__tests__/routes/settings-with-payment.test.tsx
amanape, line 216: > "should NOT render the billing tab if: SaaS mode, billing is enabled, and member user”

But it looks to expect it, and passes? 😬

```ts
const navbar = await screen.findByTestId("settings-navbar”);
within(navbar).getByText("Billing”);
```

It also does not fail if you change the `role`. Most likely because the `useSettingsNavItems` hook is completely mocked in `beforeEach`

frontend/__tests__/routes/settings.test.tsx
```diff
@@ -51,6 +54,31 @@ describe("Settings Screen", () => {
...
+    vi.spyOn(useMeModule, "useMe").mockReturnValue({
+      data: user,
+      status: "success",
+      isLoading: false,
+      isError: false,
+      isSuccess: true,
+      refetch: vi.fn(),
+      error: null,
+    } as any);
+  };
```
amanape: Could we address the same comments made in `frontend/__tests__/components/features/user/user-context-menu.test.tsx`, but here? Or could this be a different scenario and it does not apply?

frontend/src/hooks/use-settings-nav-items.ts
amanape: Good points, I’ll think about this and try and have an answer next review. My initial thoughts are that it should be safe to keep as is, this is mostly a fallback to the least-priviledged role. I also imagine other parts of the system should fail if the user is not returned, and we’re also going to have another set of guards on the backend to prevent unauthorized actions in the worst case.

We are going to have to think a little more about this though, I’m sure there are edge cases we’ll have to consider...

frontend/src/routes/llm-settings.tsx
```diff
@@ -70,6 +72,10 @@ function LlmSettingsScreen() {
   const { data: settings, isLoading, isFetching } = useSettings();
   const { data: config } = useConfig();
 
+  const { data: user } = useMe();
+  const userRole: OrganizationUserRole = user?.role ?? "member";
+  const isUserRoleViewOnly = userRole === "member";
```
amanape: WDYT of `!hasPermissions(“edit_llm_settings”)` instead?

frontend/src/routes/manage-org.tsx
amanape: BTW, I have listed some candidates for permission keys in https://github.com/OpenHands/OpenHands/issues/12276. `user:billing` implies there is some `{other}:billing`? Is that right?

frontend/src/routes/settings.tsx
```diff
@@ -21,6 +23,9 @@ const SAAS_ONLY_PATHS = [
 export const clientLoader = async ({ request }: Route.ClientLoaderArgs) => {
   const url = new URL(request.url);
   const { pathname } = url;
+  const user: OrganizationMember | undefined =
+    await getActiveOrganizationUser();
```
amanape:
```suggestion
const user = await getActiveOrganizationUser();
```

Type already inferred

frontend/src/routes/manage-org.tsx
HeyItsChloe: @hieptl Agreed!
@amanape  I have been using `view_billing`

frontend/src/routes/llm-settings.tsx
HeyItsChloe: Yup! calling the hook is smoother.

frontend/__tests__/components/features/user/user-context-menu.test.tsx
HeyItsChloe: Ah, thanks for explaining that. I'm still getting used to the way we test.

frontend/__tests__/hooks/use-permission.test.tsx
HeyItsChloe: No, I thought I'd add it for future proofing but it is really not needed.

frontend/__tests__/routes/manage-organization-members.test.tsx
HeyItsChloe: Good catch. Moved it to its own file.

frontend/__tests__/hooks/use-permission.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    it("matches the permissions defined for the role", () => {
+      const { hasPermission } = setup("member");
+
+      rolePermissions.member.forEach((permission) => {
+        expect(hasPermission(permission)).toBe(true);
+      });
+    });
+  });
+
+  describe("change_user_role permission behavior (legacy parity)", () => {
```
amanape: Could you explain what you mean by “legacy parity” here?

frontend/__tests__/routes/settings-with-payment.test.tsx
```diff
@@ -101,38 +139,77 @@ describe("Settings Billing", () => {
...
 
   const renderSettingsScreen = () =>
-    renderWithProviders(<RoutesStub initialEntries={["/settings/billing"]} />);
+    (renderWithProviders as unknown as (
+      ui: React.ReactElement,
+      options: { queryClient: QueryClient }
+    ) => ReturnType<typeof renderWithProviders>)(
+      <RoutesStub initialEntries={["/settings"]} />,
+      { queryClient }
+    );
```
amanape:
```suggestion
    render(<RoutesStub initialEntries={["/settings"]} />, {
      wrapper: ({ children }) => (
        <QueryClientProvider client={queryClient}>
          {children}
        </QueryClientProvider>
      ),
    });
```

frontend/src/utils/org/permissions.ts
```diff
@@ -1,41 +1,103 @@
...
-export const rolePermissions: Record<OrganizationUserRole, UserPermission[]> = {
+export type RolePermissionMap = {
+  owner: OwnerPermission[];
+  admin: AdminPermission[];
+  member: MemberPermission[];
+};
+
+export type Permission = RolePermissionMap[keyof RolePermissionMap][number];
+
+export const rolePermissions: RolePermissionMap = {
```
amanape: I think we can do something much simpler here:

```ts
// Define type that is the union of all permission keys
export type PermissionKey =
  | UserRoleChangePermissionKey
  | InviteUserToOrganizationKey
  | ChangeOrganizationNamePermission
  | DeleteOrganizationPermission
  | AddCreditsPermission
  | ViewBillingPermission
  | ManageSecretsPermission
  | ManageMCPPermission
  | ManageIntegrationsPermission
  | ManageApplicationSettingsPermission
  | ManageAPIKeysPermission
  | ViewLLMSettingsPermission
  | EditLLMSettingsPermission;
```

Then, instead of:

```ts
const memberPerms: MemberPermission[] = [ /* … */ ];
const adminPerms: AdminPermission[] = [ /* … */ ];
const ownerPerms: PermissionKey[] = [ /* … */ ];
```

we will have:

```ts
const memberPerms: PermissionKey[] = [ /* … */ ];
const adminPerms: PermissionKey[] = [ /* … */ ];
const ownerPerms: PermissionKey[] = [ /* … */ ];
```

Then, we can just keep:

```ts
export const rolePermissions: Record<OrganizationUserRole, PermissionKey[]> = {
  owner: ownerPerms,
  admin: adminPerms,
  member: memberPerms,
};
```

Note that we’ll have to update the very first key to:

```ts
import { OrganizationUserRole } from "#/types/org”;

type UserRoleChangePermissionKey = `change_user_role:${OrganizationUserRole}`;
```

frontend/src/routes/settings.tsx
```diff
@@ -21,6 +23,8 @@ const SAAS_ONLY_PATHS = [
 export const clientLoader = async ({ request }: Route.ClientLoaderArgs) => {
   const url = new URL(request.url);
   const { pathname } = url;
+  const user = await getActiveOrganizationUser();
+  const userRole: OrganizationUserRole = !user ? "member" : user?.role;
```
amanape:
```suggestion
  const userRole = user?.role || “member";
```

frontend/src/utils/org/permission-checks.ts
```diff
@@ -0,0 +1,47 @@
...
+    member: "change_user_role:member",
+  };
+
+  return (
+    Object.entries(roleToPermissionMap) as [OrganizationUserRole, Permission][]
+  )
+    .filter((roleToPermissionMap_keyValuePairs) =>
+      userPermissions.includes(roleToPermissionMap_keyValuePairs[1]),
+    )
+    .map(([role]) => role);
```
amanape: Prefer readability

```suggestion
  const availableRoles: OrganizationUserRole[] = [];

  if (userPermissions.includes("change_user_role:member")) {
    availableRoles.push("member");
  }
  if (userPermissions.includes("change_user_role:admin")) {
    availableRoles.push("admin");
  }
  if (userPermissions.includes("change_user_role:owner")) {
    availableRoles.push("owner");
  }

  return availableRoles;
```

frontend/__tests__/hooks/use-permission.test.tsx
HeyItsChloe: Ah, nothing! I moved the test and that comment was added accidentally. Deleted that comment

frontend/src/utils/org/permissions.ts
HeyItsChloe: Ah, yea. This is much cleaner.

amanape:
Completed in #12640

PR #12381, opened 2026-01-12
Guard User Creation with Redis based Lock

tofarr:
## Summary of PR

Adds a Redis-based distributed lock to prevent concurrent creation of the same user in `SaasSettingsStore`. When a user is being created, other processes attempting to create the same user will wait and retry loading instead of duplicating work.

**Key changes:**
- Use Redis `SET NX` for distributed locking with 30s TTL
- Wait and retry `load()` if lock is held by another process
- Graceful fallback when Redis is unavailable (proceeds without locking)
- Extracted locking logic into `_acquire_user_creation_lock()` helper
- Added comprehensive test coverage for the locking behavior

## Demo Screenshots/Videos

Users are still created by default - exactly as before:
image

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Prevents duplicate user creation under concurrent request conditions.

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:d4988d6-nikolaik   --name openhands-app-d4988d6   docker.openhands.dev/openhands/openhands:d4988d6
```

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -139,7 +139,7 @@ class SaasSettingsStore(SettingsStore):
                 session.add(settings)
             session.commit()
 
-    def _get_redis_client(self) -> aioredis.Redis | None:
+    def _get_redis_client(self):
         """Get the Redis client from the Socket.IO manager."""
         from openhands.server.shared import sio
 


enterprise/storage/saas_settings_store.py
```diff
@@ -140,6 +174,16 @@ async def create_default_settings(self, user_settings: UserSettings | None):
...
 
+        # Prevent duplicate settings creation using distributed lock
+        if not await self._acquire_user_creation_lock():
+            # The user is already being created in another thread / process
+            logger.info(
+                'saas_settings_store:create_default_settings:waiting_for_lock',
+                extra={'user_id': self.user_id},
+            )
+            await asyncio.sleep(_RETRY_LOAD_DELAY_SECONDS)
+            return await self.load()
```
malhotra5, line 185: If the process which acquired the lock failed to create a new user (DB wen't down, etc), this load will fail (which i think is ok); will the user have to retry creating an account to re-run the user creation logic?

tofarr: The create option here is a side effect, which is part of the problem. The first time user details are loaded and are not present, the system will try to create them.

In the event that the process fails, the result now will be the same as previously - except only one process will have failed rather than potentially many processes!

Any waiting processes will either error out or one of them will succeed.

tofarr: (In the long run, we may want to rethink this process)

malhotra5: Got it, thanks for the clarification!

PR #12387, opened 2026-01-13
fix(backend): github proxy state compression

hieptl:
## Summary of PR

**Issue:**
A “Your request URL is too long” error occurs when signing in via feature deployments.

**Root cause:**
The reCAPTCHA Enterprise token (approximately 800–1200 characters) is included in the OAuth state parameter. 

In feature and staging deployments, this state is passed through github_proxy.py, where it is encrypted using Fernet and then base64-encoded. This process significantly increases the size of the URL, expanding it from roughly 3,000 characters to more than 8,500 characters, which exceeds GitHub’s URL length limit.

Why this only affects staging and feature environments:
- Local / Production: The OAuth flow goes directly from Keycloak to GitHub, without any additional encryption layer.
- Staging / Feature: The OAuth flow passes through github_proxy.py, which applies Fernet encryption to preserve the original redirect URI across subdomains. This adds substantial overhead to an already large state parameter.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:1a0e1a2-nikolaik   --name openhands-app-1a0e1a2   docker.openhands.dev/openhands/openhands:1a0e1a2
```

### DIFF
--- a/enterprise/migrations/versions/090_add_git_user_fields_to_user.py
+++ b/enterprise/migrations/versions/090_add_git_user_fields_to_user.py
@@ -1,0 +1,28 @@
+"""Add git_user_name and git_user_email columns to user table.
+
+Revision ID: 090
+Revises: 089
+Create Date: 2025-01-22
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = '090'
+down_revision = '089'
+
+
+def upgrade() -> None:
+    op.add_column(
+        'user',
+        sa.Column('git_user_name', sa.String, nullable=True),
+    )
+    op.add_column(
+        'user',
+        sa.Column('git_user_email', sa.String, nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column('user', 'git_user_email')
+    op.drop_column('user', 'git_user_name')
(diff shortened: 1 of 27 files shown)


raymyers:
It looks like this change in untested. I am attempting to add tests that ensure the encode and decode will match. I'm also concerned about a timing issue which this is being rolled out, a message might start in old format and fail to decode in the new

raymyers:
Renamed to not a refactor because it's a behavior change

raymyers:
@hieptl just flagging this is unmerged. Should we?

hieptl:
Hello @raymyers,

> @hieptl just flagging this is unmerged. Should we?

The issue described in the pull request only occurs with feature deployments.

I will revisit this pull request after we have released and tested reCAPTCHA again in the production environment. For now, I would prefer to keep the changes as minimal as possible to reduce any potential risk of another incident.

Thank you very much! 🙏

Issue #12391, opened 2026-01-13
Changes tab refresh button doesn't update the displayed changes


mamoodi:
## Bug Description

The refresh button on the Changes tab doesn't appear to update the displayed git changes when clicked. The button exists and is clickable, but the UI doesn't reflect any updates even when files have been modified in the repository.

## Steps to Reproduce

1. Open OpenHands with a git repository
2. Navigate to the Changes tab
3. Make some changes to files in the repository (e.g., modify a file or add a new file)
4. Click the refresh button in the Changes tab header
5. Observe that the changes list doesn't update to show the new modifications

## Expected Behavior

When clicking the refresh button, the Changes tab should:
- Fetch the latest git changes from the repository
- Update the displayed list of changed files
- Show newly modified files
- Remove files that are no longer changed
- Update the status of existing files if their change type has changed

## Actual Behavior

The refresh button is clicked but the displayed changes don't update. The list remains static even when new changes have been made to the repository.

## Root Cause Analysis

After investigating the code, the issue appears to be in the `useUnifiedGetGitChanges` hook (`frontend/src/hooks/query/use-unified-get-git-changes.ts`):

1. **Data reference comparison**: The hook uses a `previousDataRef` to check if data has changed (line 72):
   ```typescript
   if (currentData !== previousDataRef.current) {
   ```
   If TanStack Query returns the same object reference (due to caching), this condition fails and the UI doesn't update.

2. **Ordering preservation logic**: The hook maintains a custom ordering of changes that only adds NEW items to the top (lines 76-91). This means:
   - If a file is modified (not a new file), it won't move to the top
   - The ordering might not be reset even after a refetch
   - Changes to existing files might not be reflected in the UI

3. **Stale time**: The query has a `staleTime` of 5 minutes (line 58), which might prevent immediate refetching in some cases, though `refetch()` should bypass this.

## Suggested Fix

There are several potential approaches to fix this:

1. **Reset ordering on manual refetch**: When `refetch()` is called explicitly, reset the `orderedChanges` state completely instead of trying to preserve order:
   ```typescript
   const { refetch: originalRefetch, ...rest } = useQuery(...);
   
   const refetch = () => {
     setOrderedChanges([]); // Clear current state
     previousDataRef.current = null; // Reset reference
     return originalRefetch();
   };
   ```

2. **Deep comparison instead of reference comparison**: Use a deep comparison of the data to detect changes:
   ```typescript
[...]

PR #12392, opened 2026-01-13
fix: Refresh button not updating git changes in Changes tab

nehaaprasad:
### Summary of PR

- fixes the refresh button on the Changes tab not updating the displayed git changes when clicked. The issue was caused by reference comparison failing due to TanStack Query caching and incomplete ordering logic that didn't handle modified files properly.

## Change Type

fix: #12391

## Fixes

Resolves the bug where clicking the refresh button in the Changes tab doesn't update the displayed git changes list.


frontend/src/hooks/query/use-unified-get-git-changes.ts
```diff
@@ -18,6 +18,7 @@ export const useUnifiedGetGitChanges = () => {
   const { data: conversation } = useActiveConversation();
   const [orderedChanges, setOrderedChanges] = React.useState<GitChange[]>([]);
   const previousDataRef = React.useRef<GitChange[] | null>(null);
+  const [forceUpdate, setForceUpdate] = React.useState(0);
```
hieptl: Since we are using React Query, it appears that we do not need to define a separate state to force an update. Instead, we can simply invalidate the relevant query keys to trigger a refresh.

Thank you! 🙏

nehaaprasad:
@naaa760 
I've updated the code to use `queryClient.invalidateQueries()` instead of the `forceUpdate` state.
please check it out!

hieptl:
Hello @nehaaprasad,

The issue has now been resolved, so I will proceed with closing this pull request.

Thank you very much for your contribution and for raising this pull request—it’s greatly appreciated. 🙏

PR #12398, opened 2026-01-13
fix(frontend): clear error banner on successful V1 WebSocket events

amanape:
## Summary of PR

- Added a test to verify this behavior and prevent regression.
- Added logic to clear error messages when receiving successful (non-error) events, matching V0 behavior:

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

- [~] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:693d232-nikolaik   --name openhands-app-693d232   docker.openhands.dev/openhands/openhands:693d232
```

### DIFF
--- a/frontend/src/contexts/conversation-websocket-context.tsx
+++ b/frontend/src/contexts/conversation-websocket-context.tsx
@@ -329,24 +329,22 @@ export function ConversationWebSocketProvider({
         if (isV1Event(event)) {
           addEvent(event);
 
-          // Handle ConversationErrorEvent specifically
+          // Handle ConversationErrorEvent specifically - show error banner
+          // AgentErrorEvent errors are displayed inline in the chat, not as banners
           if (isConversationErrorEvent(event)) {
             setErrorMessage(event.detail);
+          } else {
+            // Clear error message on any non-ConversationErrorEvent
+            removeErrorMessage();
           }
 
-          // Handle AgentErrorEvent specifically
+          // Track credit limit reached if AgentErrorEvent has budget-related error
           if (isAgentErrorEvent(event)) {
-            setErrorMessage(event.error);
-
-            // Track credit limit reached if the error is budget-related
             if (isBudgetOrCreditError(event.error)) {
               trackCreditLimitReached({
                 conversationId: conversationId || "unknown",
               });
             }
-          } else if (!isConversationErrorEvent(event)) {
-            // Clear error message on any successful (non-error) event
-            removeErrorMessage();
           }
 
           // Clear optimistic user message when a user message is confirmed
--- a/frontend/__tests__/conversation-websocket-handler.test.tsx
+++ b/frontend/__tests__/conversation-websocket-handler.test.tsx
@@ -19,7 +19,7 @@ import { useErrorMessageStore } from "#/stores/error-message-store";
 import {
   createMockMessageEvent,
   createMockUserMessageEvent,
-  createMockAgentErrorEvent,
+  createMockConversationErrorEvent,
   createMockBrowserObservationEvent,
   createMockBrowserNavigateActionEvent,
   createMockExecuteBashActionEvent,
@@ -52,8 +52,9 @@ afterEach(() => {
   mswServer.resetHandlers();
   // Clean up any React components
   cleanup();
-  // Reset error message store to prevent state leakage between tests
+  // Reset stores to prevent state leakage between tests
   useErrorMessageStore.getState().removeErrorMessage();
+  useEventStore.getState().clearEvents();
 });
 
 afterAll(async () => {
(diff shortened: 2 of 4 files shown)


frontend/src/contexts/conversation-websocket-context.tsx
```diff
@@ -344,6 +344,9 @@ export function ConversationWebSocketProvider({
                 conversationId: conversationId || "unknown",
               });
             }
+          } else if (!isConversationErrorEvent(event)) {
+            // Clear error message on any successful (non-error) event
+            removeErrorMessage();
```
hieptl: **question:** Have you verified this work locally? Could you also attach a short demo video? Additionally, once the ConversationErrorEvent and other events (such as stats or conversation status updates) are sent, will the error message be cleared immediately—even if I haven’t seen the error message yet to be aware of the issue? Thank you! 🙏

enyst (approved):
Thank you! If you want to do it this way, this PR helps - at least it clears it, so the user can see the conversation.

I would suggest to restore something like, the essentials of V0 behavior here, though:
- there are agent errors (`AgentErrorEvent`)
- there are conversation errors (`ConversationRunEvent`)

They're totally different, and easier to distinguish in V1 than they used to be in V0. (ugh we were parsing strings to guess if the agent saw some error 😅)

AgentErrorEvents are expected, normal, and in some sense "not really errors". IMHO they could maybe not come across as scary to the user [1], shouldn't be blocking anything, they are normal for the agent to encounter. My suggestion is that maybe we could show the user some event block that looks like normal event blocks, maybe with a tiny indicator that the LLM made a boo-boo.

[1] We had many issues filed, where the users report the agent errors as bugs. "Look OH had an error". They are usually not exactly bugs though, they are normal and the agent recovers just fine. I still don't know how to answer people on this, because I understand that if it looks like errors they will be reported as errors... but they're not **_application errors._** IMHO visual indicators could help a bit there. Just a thought, though.

On another note: John Mason has also a PR here:
- https://github.com/OpenHands/OpenHands/pull/12335

I won't approve it since it changes visual appearance and Hiep Le feels there's duplicate code maybe, up to you visual wizards on that. You may want to take a look for the regular conversation errors too, though? Some if not all of those are also not blocking... for different reasons.

frontend/src/contexts/conversation-websocket-context.tsx
enyst: Sorry, @hieptl , I didn't see your message before I posted a long reply here!

I'm sure Stephan will take feedback into account though. 🙏

amanape:
> AgentErrorEvents are expected, normal, and in some sense "not really errors". IMHO they could maybe not come across as scary to the user [1], shouldn't be blocking anything, they are normal for the agent to encounter. My suggestion is that maybe we could show the user some event block that looks like normal event blocks, maybe with a tiny indicator that the LLM made a boo-boo.

That makes sense, I’ll create a GitHub issue tomorrow for this (at least for a straightforward component) but I should also probably ping Paul to get his thoughts on the design and UI of this…

Though just for the record, we do have some `AgentErrorEvent`’s that we’d still want to show to the user, e.g., low credit errors. Not sure if we should move these elsewhere, but I guess it is simpler to return it as an `AgentErrorEvent`. In any way, we can filter it appropriately on the FE

frontend/src/contexts/conversation-websocket-context.tsx
amanape: I was planning to test it on a feature branch / staging as I’ll have more control over displaying the error messages (e.g., setting my credit limit). I still don’t have a consistent way of testing error banners on the FE other than tests. Do either of you have suggestions of any way I could try other than a feature branch?

enyst (commented):
> we do have some AgentErrorEvent’s that we’d still want to show to the user, e.g., low credit errors. Not sure if we should move these elsewhere, but I guess it is simpler to return it as an AgentErrorEvent

Sorry, I think low credit doesn't fit `AgentErrorEvent`?

AgentErrorEvents are sent to the LLM next step, so that it corrects itself. For example
- this validation error because of the LLM sending us a [tool call without security risk](https://github.com/OpenHands/software-agent-sdk/issues/1653#issuecomment-3727241174). 
  - It's no big deal, and it can send risk next step. 
- except the user _sees it_ like this [validation error](https://github.com/OpenHands/OpenHands/issues/12332#issuecomment-3725977612) 😿 

Low credit has nothing to do with it, I don't think? It's not like the LLM can do anything about it 😅 

Is there any reason why low credit isn't, idk, maybe a warning? Or maybe a `ConversationRunError` if we need a pre-existing event?

frontend/src/contexts/conversation-websocket-context.tsx
enyst: For `AgentErrorEvent`, I'd try to tell the LLM to send an action without security risk (when security risk is enabled) - _(edited to add_), which I think it's default

For `ConversationErrorEvent`, maybe set max iterations to 10, and let it run so that it hits it?

amanape: I have confirmed and tested this behavior. I updated the changes to prevent setting an error banner on agent error message since we already render it inline and as a toast. I have set to remove the message for conversation errors after the next non-error event

enyst: Thank you!

PR #12402, opened 2026-01-14
refactor(frontend): create a feature flag for google recaptcha

hieptl:
## Summary of PR

We experienced an incident yesterday where users were unable to sign in. The root cause was identified as being related to a database migration.

During that time, we also observed some reCAPTCHA-related errors on the UI. As a result, it is currently unclear whether reCAPTCHA itself was affected or if those errors were a side effect of the incident.

To safely validate reCAPTCHA behavior in the production environment, we propose creating a feature flag. With this approach, the Google reCAPTCHA script will only be loaded when the feature flag is enabled, and will be skipped when the flag is disabled. This allows us to test and control the functionality without introducing unnecessary risk.

**Acceptance Criteria:**
- The feature flag is created and configured successfully.
- When the feature flag is enabled, the Google reCAPTCHA script is loaded.
- When the feature flag is disabled, the Google reCAPTCHA script is not loaded.

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/f47ec72d-5e10-4374-8cd0-fc6faeee520e

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:034bfe7-nikolaik   --name openhands-app-034bfe7   docker.openhands.dev/openhands/openhands:034bfe7
```

### DIFF
--- a/frontend/src/components/features/auth/login-content.tsx
+++ b/frontend/src/components/features/auth/login-content.tsx
@@ -12,6 +12,7 @@ import { TermsAndPrivacyNotice } from "#/components/shared/terms-and-privacy-not
 import { useRecaptcha } from "#/hooks/use-recaptcha";
 import { useConfig } from "#/hooks/query/use-config";
 import { displayErrorToast } from "#/utils/custom-toast-handlers";
+import { ENABLE_RECAPTCHA } from "#/utils/feature-flags";
 
 export interface LoginContentProps {
   githubAuthUrl: string | null;
@@ -38,7 +39,7 @@ export function LoginContent({
 
   // reCAPTCHA - only need token generation, verification happens at backend callback
   const { isReady: recaptchaReady, executeRecaptcha } = useRecaptcha({
-    siteKey: config?.RECAPTCHA_SITE_KEY,
+    siteKey: ENABLE_RECAPTCHA() ? config?.RECAPTCHA_SITE_KEY : undefined,
   });
 
   const gitlabAuthUrl = useAuthUrl({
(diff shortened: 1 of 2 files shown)


raymyers:
Are these flags currently driven by the user selecting something in local storage?

raymyers (approved):

hieptl:
Hello @raymyers,

> Are these flags currently driven by the user selecting something in local storage? 

Yes, reCAPTCHA can be enabled or disabled by setting the flag in localStorage. As shown in the demo video, anyone can toggle it directly from their browser using:

```
localStorage.setItem("FEATURE_RECAPTCHA", "true");
localStorage.setItem("FEATURE_RECAPTCHA", "false");
```

The main goal of this ticket is to deploy reCAPTCHA back to the production environment, but keep it turned off by default. This will allow the All Hands team to validate it first, as we are still not fully certain whether yesterday’s incident was in any way related to reCAPTCHA.

With this approach, we can safely test reCAPTCHA in production without impacting end users, even though we have already carefully verified its behavior in the staging environment.

Once we have verified in the production environment that everything is working as expected, we will remove this feature flag and enable reCAPTCHA for all users.

Thank you very much! 🙏

PR #12411, opened 2026-01-14
[APP-369] Fix: Allow public access to shared conversations and events

tofarr:
## Summary of PR

This PR fixes the issue where publicly shared conversations were not accessible to unauthenticated users. When visiting a shared conversation URL (e.g., `https://app.all-hands.dev/shared/conversations/{id}`) in incognito mode, users were seeing "Conversation not found" instead of the shared conversation content.

**Root Cause:** The `SetAuthCookieMiddleware` in `enterprise/server/middleware.py` was requiring authentication for all `/api/*` routes, including `/api/shared-conversations` and `/api/shared-events` which should be publicly accessible.

**Fix:** Added a `startswith` check for `/api/shared-` paths to bypass the authentication middleware, allowing these public endpoints to be accessed without credentials.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Release Notes

Fixed an issue where publicly shared conversations could not be viewed by unauthenticated users. Shared conversation URLs now work correctly in incognito mode and for users who are not logged in.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:a33916c-nikolaik   --name openhands-app-a33916c   docker.openhands.dev/openhands/openhands:a33916c
```

### DIFF
--- a/enterprise/server/middleware.py
+++ b/enterprise/server/middleware.py
@@ -167,7 +167,9 @@ class SetAuthCookieMiddleware:
             return False
 
         # Allow public access to shared conversations and events
-        if path.startswith('/api/shared-conversations') or path.startswith('/api/shared-events'):
+        if path.startswith('/api/shared-conversations') or path.startswith(
+            '/api/shared-events'
+        ):
             return False
 
         is_mcp = path.startswith('/mcp')


enterprise/server/middleware.py
```diff
@@ -166,6 +166,10 @@ def _should_attach(self, request: Request) -> bool:
         if path in ignore_paths:
             return False
 
+        # Allow public access to shared conversations and events
+        if path.startswith('/api/shared-'):
```
hieptl: **question:** If other engineers create additional APIs with the same prefix in the future and are not aware of this part, their APIs might unintentionally become public without proper notice? Thank you! 🙏

tofarr: You're right. it's better be explicit here.

PR #12450, opened 2026-01-15
fix: add retry logic with exponential backoff to send_welcome_email

ak684:
## Problem
Production Datadog logs showed `ResendError: Too many requests` rate limiting errors (429) when sending welcome emails to newly added contacts.

**Error message:** `Too many requests. You can only make 2 requests per second. See rate limit response headers for more information.`

The error occurred in `enterprise/sync/resend_keycloak.py` in the `sync_users_to_resend` → `send_welcome_email` call chain.

## Root Cause
The `send_welcome_email` function was missing retry logic with exponential backoff. While `add_contact_to_resend` was correctly decorated with tenacity's `@retry` decorator to handle rate limiting, `send_welcome_email` was calling `resend.Emails.send()` directly without any retry mechanism.

When Resend's rate limit (2 requests/second) was hit, the email sending would fail immediately without retrying.

## Fix
Added the same `@retry` decorator to `send_welcome_email` that already exists on `add_contact_to_resend`:

```python
@retry(
    stop=stop_after_attempt(MAX_RETRIES),
    wait=wait_exponential(
        multiplier=INITIAL_BACKOFF_SECONDS,
        max=MAX_BACKOFF_SECONDS,
        exp_base=BACKOFF_FACTOR,
    ),
    retry=retry_if_exception_type(ResendError),
)
def send_welcome_email(...)
```

This ensures:
- Maximum 3 attempts (configurable via `MAX_RETRIES`)
- Exponential backoff between retries (1s, 2s, 4s, etc.)
- Only retries on `ResendError` (which includes rate limit 429 errors)

## Testing
- Added unit tests in `enterprise/tests/unit/sync/test_resend_keycloak.py`
- Tests verify:
  - Successful email sending
  - Retry behavior on rate limit errors
  - Failure after max retries
  - Both functions (`send_welcome_email` and `add_contact_to_resend`) now have consistent retry behavior

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:98075b0-nikolaik   --name openhands-app-98075b0   docker.openhands.dev/openhands/openhands:98075b0
```

### DIFF
--- a/.github/workflows/pr-review-by-openhands.yml
+++ b/.github/workflows/pr-review-by-openhands.yml
@@ -2,19 +2,11 @@
 name: PR Review by OpenHands
 
 on:
-    # Use pull_request_target so this workflow can use secrets in the base repo context.
-    # Security: This workflow runs when:
-    #   1. A non-draft PR is opened by a collaborator/member/owner, OR
-    #   2. A draft PR is marked ready_for_review by a collaborator/member/owner, OR
-    #   3. openhands-agent or all-hands-bot is requested as a reviewer on a PR from a
-    #      collaborator/member/owner, OR
-    #   4. A maintainer adds the 'review-this' label (manual trigger for external authors)
-    # Note: PR authors (including from forks) can request reviewers, but this workflow
-    # will only auto-run when the PR author is collaborator/member/owner.
-    # For external authors, a maintainer can trigger it by applying the 'review-this' label.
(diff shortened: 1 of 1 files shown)


malhotra5:
@openhands resolve the merge conflicts

PR #12460, opened 2026-01-16
(Backend) New Config Endpoint

tofarr:
## Summary of PR

This is the backend only. We needed to add new fields to the config which will be displayed in the browser in the event of errors:
* **faulty_models**: For when an upstream provider is having an outage 
* **error_message**: Generic error message (I imagine that the frontend will first try to interpret this as a translation key, and if that fails, simply display the message.)

I used this opportunity to:
* **Bring the config endpoint in line with the V1 standards** - Use pydantic models instead of freeform dicts. (So that we get OpenAPI docs)
* **Use the dependency injection** / environment variable injection for V1 in the config. (To reduce manual copying using the EnvParser.)
* **Remove some V0 variables which we are no longer using** and have not been for a long time. (`app_slug`)
* **Deprecate the old endpoint** - I added a new endpoint rather than altering the existing one.

This does not update the frontend - that will be done in a subsequent PR.
This will also require an update to the deploy project.

## Demo Screenshots/Videos

image
image

Or when I set the environment variables:

```
"OH_WEB_CLIENT_KIND": "DefaultWebClientConfigInjector",
"OH_WEB_CLIENT_FEATURE_FLAGS_ENABLE_BILLING": "1",
"OH_WEB_CLIENT_ERROR_MESSAGE": "Warning: Second breakfast not found! Please install additional breakfast.",

```

image

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Refactor

## Checklist

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:b670775-nikolaik   --name openhands-app-b670775   docker.openhands.dev/openhands/openhands:b670775
```


openhands/app_server/web_client/default_web_client_config_injector.py
```diff
@@ -0,0 +1,40 @@
...
+    WebClientConfigInjector,
+)
+from openhands.app_server.web_client.web_client_models import (
+    WebClientConfig,
+    WebClientFeatureFlags,
+)
+from openhands.integrations.service_types import ProviderType
+
+
+class DefaultWebClientConfigInjector(WebClientConfigInjector):
```
malhotra5, line 15: How do we plan on overriding this default injector for saas mode? (via env?)

tofarr: Right now, I think we can do this with just environment variables. Since we don't seem to use the app_slug anymore, so we don't need to regenerate it.

PR #12475, opened 2026-01-16
Fix asyncio event loop conflict in get_user_by_id

chuckbutkus:
Add get_user_by_id_async() method to UserStore to avoid nested event loop issues when calling from async contexts. The original sync method used call_async_from_sync() internally which created a new event loop, causing 'Task got Future attached to a different loop' errors when Redis operations were performed.

Changes:
- Add UserStore.get_user_by_id_async() that stays fully async
- Update auth.py, billing.py, slack.py to use async version
- Update saas_secrets_store.py and lite_llm_manager.py to use async version
- Update api_keys.py generate_byor_key() to use async version
- Update all related tests to mock get_user_by_id_async with AsyncMock
- Remove unused call_sync_from_async imports

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:7904e34-nikolaik   --name openhands-app-7904e34   docker.openhands.dev/openhands/openhands:7904e34
```

### DIFF
--- a/enterprise/storage/user_store.py
+++ b/enterprise/storage/user_store.py
@@ -6,6 +6,7 @@ import asyncio
 import uuid
 from typing import Optional
 
+from server.auth.token_manager import TokenManager
 from server.constants import (
     LITE_LLM_API_URL,
     ORG_SETTINGS_VERSION,
@@ -24,7 +25,6 @@ from storage.user import User
 from storage.user_settings import UserSettings
 
 from openhands.utils.async_utils import GENERAL_TIMEOUT, call_async_from_sync
-from server.auth.token_manager import TokenManager
 
 # The max possible time to wait for another process to finish creating a user before retrying
 _REDIS_CREATE_TIMEOUT_SECONDS = 30


enterprise/storage/user_store.py
```diff
@@ -361,6 +366,64 @@ def get_user_by_id(user_id: str) -> Optional[User]:
...
+            # Check if we need to migrate from user_settings
+            while not await UserStore._acquire_user_creation_lock(user_id):
+                # The user is already being created in another thread / process
+                logger.info(
+                    'saas_settings_store:create_default_settings:waiting_for_lock',
+                    extra={'user_id': user_id},
+                )
+                await asyncio.sleep(_RETRY_LOAD_DELAY_SECONDS)
+
+            # Check for user again as migration could have happened while trying to get the lock.
```
hieptl, line 394: **question (non-blocking):** Is there a better approach to prevent the query from running twice?. Thank you! 🙏

enterprise/storage/user_store.py
```diff
@@ -361,6 +366,64 @@ def get_user_by_id(user_id: str) -> Optional[User]:
...
+            user_settings = (
+                session.query(UserSettings)
+                .filter(
+                    UserSettings.keycloak_user_id == user_id,
+                    UserSettings.already_migrated.is_(False),
+                )
+                .first()
+            )
+            if user_settings:
+                from server.auth.token_manager import TokenManager
```
hieptl: **question (non-blocking):** Would it be possible to move this import to the top of the file, or is it intentionally placed here to avoid a circular import? Thank you! 🙏

hieptl (approved):
Thank you! 🙏

tofarr (approved):

enterprise/storage/user_store.py
chuckbutkus: I think this is the best way to do it. For users that are already migrated, the first query will return and you never get here.  If the user is migrated while you are holding the lock, you don't want to repeat the migration.

enterprise/storage/user_store.py
chuckbutkus: This is likely to avoid circular reference.  There were lots of issues with circular references that I don't think we can resolve with a major refactoring.

tofarr: I've committed a change to move this to the top - it does not seem to cause circular imports.

PR #12478, opened 2026-01-16
[Jira]: separate signature verification from conversation orchestration 

malhotra5:
## Summary of PR

This PR does a bit of cleaning, and follows the existing pattern of our other resolvers

1. we move the signature verification logic to where the webhook is received
2. we raise an exception at the event router rather than inside the jira manager, as to simplify the tracing
3. add logs for traceability

## Demo Screenshots/Videos

image

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:0d0d109-nikolaik   --name openhands-app-0d0d109   docker.openhands.dev/openhands/openhands:0d0d109
```

### DIFF
--- a/enterprise/server/routes/integration/jira.py
+++ b/enterprise/server/routes/integration/jira.py
@@ -16,6 +16,7 @@ from pydantic import BaseModel, Field, field_validator
 from server.auth.constants import JIRA_CLIENT_ID, JIRA_CLIENT_SECRET
 from server.auth.saas_user_auth import SaasUserAuth
 from server.auth.token_manager import TokenManager
+from storage.jira_workspace import JiraWorkspace
 from storage.redis import create_redis_client
 
 from openhands.core.logger import openhands_logger as logger
--- a/enterprise/integrations/jira/jira_manager.py
+++ b/enterprise/integrations/jira/jira_manager.py
@@ -1,4 +1,4 @@
-from typing import Dict, Tuple
+from typing import Tuple
 from urllib.parse import urlparse
 
 import httpx


enterprise/server/routes/integration/jira.py
```diff
@@ -227,13 +272,12 @@ async def jira_events(
         )
 
     try:
-        signature_valid, signature, payload = await jira_manager.validate_request(
-            request
-        )
+        signature_header = request.headers.get('x-hub-signature')
+        signature = signature_header.split('=')[1] if signature_header else None
```
hieptl: #### Unsafe Signature Header Parsing

**Location:** `enterprise/server/routes/integration/jira.py:276`

**Current Implementation:**
```python
signature_header = request.headers.get('x-hub-signature')
signature = signature_header.split('=')[1] if signature_header else None
```

**Problem:**
- If `signature_header` is malformed (e.g., `"sha256"` without `"="`, or `"sha256="` with no value), `split('=')[1]` will raise an `IndexError`
- This could crash the webhook endpoint instead of returning a proper 403 error
- The code assumes the header always contains `"="` and has a value after it

**Proposed Fix:**
```python
signature_header = request.headers.get('x-hub-signature')
if signature_header:
    # Handle both "sha256=..." and "sha256=..." formats
    if '=' in signature_header:
        signature = signature_header.split('=', 1)[1]  # Use maxsplit=1 for safety
    else:
        signature = None
        logger.warning(
            '[Jira] Malformed signature header - missing "=" separator',
            extra={'header_value': signature_header}
        )
else:
    signature = None
```

**Alternative (More Robust):**
```python
signature_header = request.headers.get('x-hub-signature')
signature = None

if signature_header:
    parts = signature_header.split('=', 1)  # maxsplit=1 prevents issues with '=' in hash
    if len(parts) == 2 and parts[1]:  # Ensure we have both prefix and value
        signature = parts[1]
    else:
        logger.warning(
            '[Jira] Malformed signature header format',
            extra={'header_value': signature_header}
        )
```

Thank you! 🙏

enterprise/server/routes/integration/jira.py
```diff
@@ -122,6 +125,48 @@ class JiraValidateWorkspaceResponse(BaseModel):
 redis_client = create_redis_client()
 
 
+async def verify_jira_signature(body: bytes, signature: str, payload: dict):
+    if not signature:
```
hieptl, line 144: #### Inconsistent Error Message

**Location:** `enterprise/server/routes/integration/jira.py:131`

**Current Implementation:**
```python
if not signature:
    raise HTTPException(
        status_code=403, detail='x-hub-signature-256 header is missing!'
    )
```

**Problem:**
- Error message says `"x-hub-signature-256"` but the code checks for `"x-hub-signature"` (line 275)
- This is confusing for debugging and inconsistent with the actual header name
- GitHub integration uses `x-hub-signature-256`, but Jira uses `x-hub-signature`

**Proposed Fix:**
```python
if not signature:
    raise HTTPException(
        status_code=403, detail='x-hub-signature header is missing!'
    )
```

**Note:** Verify what Jira actually sends. If Jira sends `x-hub-signature-256`, update the header check accordingly.

Thank you! 🙏

enterprise/server/routes/integration/jira.py
```diff
@@ -122,6 +125,48 @@ class JiraValidateWorkspaceResponse(BaseModel):
 redis_client = create_redis_client()
 
 
+async def verify_jira_signature(body: bytes, signature: str, payload: dict):
```
hieptl, line 128: #### Add Docstring to `verify_jira_signature`

**Location:** `enterprise/server/routes/integration/jira.py:128`

**Current Implementation:**
```python
async def verify_jira_signature(body: bytes, signature: str, payload: dict):
```

**Proposed Fix:**
```python
async def verify_jira_signature(
    body: bytes, 
    signature: str | None, 
    payload: dict[str, any]
) -> None:
    """
    Verify Jira webhook signature.
    
    Args:
        body: Raw request body bytes
        signature: Signature from x-hub-signature header (format: "sha256=<hash>")
        payload: Parsed JSON payload from webhook
    
    Raises:
        HTTPException: 403 if signature verification fails or workspace is invalid
    
    Returns:
        None (raises exception on failure)
    """
```

Thank you! 🙏

enterprise/server/routes/integration/jira.py
```diff
@@ -227,13 +272,12 @@ async def jira_events(
         )
 
     try:
-        signature_valid, signature, payload = await jira_manager.validate_request(
-            request
-        )
+        signature_header = request.headers.get('x-hub-signature')
```
hieptl: #### Consider Using FastAPI Header Dependency

**Location:** `enterprise/server/routes/integration/jira.py:275`

**Current Implementation:**
```python
signature_header = request.headers.get('x-hub-signature')
```

**Proposed Alternative (Following GitHub Pattern):**
```python
@jira_integration_router.post('/events')
async def jira_events(
    request: Request,
    background_tasks: BackgroundTasks,
    x_hub_signature: str = Header(None),  # Use FastAPI Header dependency
):
    # ...
    signature = x_hub_signature.split('=', 1)[1] if x_hub_signature and '=' in x_hub_signature else None
```

**Rationale:** 
- More consistent with GitHub integration pattern
- FastAPI handles header extraction automatically
- Still need safe parsing, but cleaner API

Thank you! 🙏

hieptl:
Hello @malhotra5,

I have left a few comments on this pull request for your reference.

Additionally, I wanted to share some feedback regarding the demo screenshot. It is a bit unclear to me, as it currently shows only the user’s message and does not include the conversation summary message from OpenHands. From my understanding, the summary message is an important part of the interaction.

For example, when I use the GitLab resolver and type “@ openhands echo hello world,” I receive a summary message after the conversation is completed. It would be very helpful if we could include that summary message in the demo screenshot alongside the user’s message, as it would better reflect the full user experience.

Thank you very much for your time and effort. I really appreciate your work on this! 🙏

malhotra5:
That is a known bug that I'm attempting to fix.

However, the existing logic is a mess so I'm making incremental PRs to clean up the code. The actual fix will come in a future PR, where the jira integration should kick off a job and post a summary back

malhotra5:
@openhands please fix the failing unit tests in `enterprise/tests/unit/server/routes/test_jira_integration_routes` 

modify the test files only since the original specs have changed

Issue #12473, opened 2026-01-16
Add timeout on repo selector for Slack integration


neubig:
**What problem or use case are you trying to solve?**

Currently, when sending a message on slack I might send a request to the slack plugin but forget to select a repo, which results in the job never starting.

**Describe the UX or technical implementation you have in mind**

If the repo selector comes up but is not selected within a set timeout (e.g. 2 minutes), the agent starts working automatically and the repo selector goes away.

nehaaprasad:
Hello! @neubig 
Please check out the PR!

PR #12482, opened 2026-01-17
feat: Add 2-minute timeout for Slack repo selector

nehaaprasad:
## Summary of PR

- Add 2-minute timeout for Slack repository selector to prevent jobs from never starting when users forget to select a repository.

## Change Type

- [x] New feature

fix: #12473


neubig:
Thanks @naaa760 could you please send a video demonstrating that this works as expected?

h55nick:
> Thanks @naaa760 could you please send a video demonstrating that this works as expected?

@neubig  
Are you asking for a video because it has no tests or even with tests that's the only way to confirm it's working with high enough confidence?

nehaaprasad:
Hello! @neubig and @h55nick 
I added a 2 minute timeout to the Slack repo picker. If the user doesn’t choose a repo, the job starts automatically without one; if they do choose, we cancel the timeout and start with that repo. Only caveat: the timeout tracking is in memory, so for multi worker setups we should move it to Redis or the DB for full reliability.

neubig:
Hi @naaa760 , thanks! Have you been able to test this and make sure it works? If so, could you please send a video?

nehaaprasad:
@neubig 
I tested the logic locally and the timeout/cancel flow behaves as expected. I don’t have a Slack environment set up here to record an end to end video, but I can produce one once I run it against our Slack workspace.

neubig:
@h55nick yes I am asking for a video because I would like to make sure that the submitter has actually tested the code themselves and seen that it works. If that is the case it should be easy to produce a video. 

@naaa760 okay thank you! This may be difficult for you to set up yourself, as is all the code in the enterprise/ directory. If you would like us to try to test it we can do that as well, but it may take some time. We appreciate contributions, but they are much easier to review if you can demonstrate that the code didn't work before you created the fix and does work afterwards through sending a video if it is front-end related code.

nehaaprasad:
@neubig sir, I don’t have an enterprise Slack setup here, so I haven’t been able to record a before/after video. I’d be happy if you could test it when you have time, or I can add unit tests for the timeout logic to make review easier.

mamoodi:
I'm going to close this PR based off of malhotra5's comment. Hopefully that's okay. Let me know otherwise.

PR #12506, opened 2026-01-20
fix(frontend): sub-conversation ids are missing when navigating from the conversation panel

hieptl:
## Summary of PR

**Reproduction Steps:**

1. Create a new v1 conversation.
2. Switch to the Planning Agent. At this point, a sub-conversation should be created, and after refetching the parent conversation data, the sub_conversation_ids field is expected to include the newly created sub-conversation ID.
3. Navigate back to the home page.
4. From the home page, navigate again to the parent conversation.

At this stage, although the API request completes successfully, we observe that the sub_conversation_ids field appears empty on the front-end.

**Root Cause:**

This issue seems to be related to how React Query manages its cache and sets query data when navigation occurs from the conversation panel. The current behavior likely results in stale or incomplete data being used on the UI.

Please refer to the code snippet below for additional context and implementation details.

```
// Optimistically populate individual conversation caches
result.results.forEach((conversation) => {
  queryClient.setQueryData(
    ["user", "conversation", conversation.conversation_id],
    conversation, // <== Does not contain sub_conversation_ids field
  );
});
```

The code snippet mentioned above was introduced to address the issue raised in this pull request:

- https://github.com/OpenHands/OpenHands/pull/11510

However, I believe it would be beneficial for us to remove this snippet and reopen the issue in the attached pull request so that we can resolve it using a more appropriate solution.

One possible question to consider is why we are not returning the sub_conversation_ids field when fetching paginated conversations. While including it could also resolve the issue, this approach would introduce an N + 1 query problem, which we should try to avoid for performance and scalability reasons.

From my perspective, removing the current code snippet and addressing the root cause more directly would be the cleanest and most maintainable solution.

## Demo Screenshots/Videos

N/A.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:0f82db6-nikolaik   --name openhands-app-0f82db6   docker.openhands.dev/openhands/openhands:0f82db6
```

### DIFF
--- a/frontend/src/hooks/query/use-paginated-conversations.ts
+++ b/frontend/src/hooks/query/use-paginated-conversations.ts
@@ -1,10 +1,9 @@
-import { useInfiniteQuery, useQueryClient } from "@tanstack/react-query";
+import { useInfiniteQuery } from "@tanstack/react-query";
 import ConversationService from "#/api/conversation-service/conversation-service.api";
 import { useIsAuthed } from "./use-is-authed";
 
 export const usePaginatedConversations = (limit: number = 20) => {
   const { data: userIsAuthenticated } = useIsAuthed();
-  const queryClient = useQueryClient();
 
   return useInfiniteQuery({
     queryKey: ["user", "conversations", "paginated", limit],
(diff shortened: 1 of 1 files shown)


hieptl:
> Thank you! Not sure why we were optimistically updating the cache in the first place… 🤔

Hello @amanape,

I have explained the rationale behind this logic in the pull request description.

Previously, we chose this approach to address the issue described here: https://github.com/OpenHands/OpenHands/pull/11495.

Thank you! 🙏

PR #11814, opened 2025-11-25
fix(backend): github token not working for v1 conversations

hieptl:
## Summary of PR

When creating a new V1 conversation, we send a prompt similar to the following:

> Go to the OpenHands github page at https://github.com/OpenHands/OpenHands and tell me how many stars the github repo currently has. Use the GITHUB_TOKEN

We will receive the response below.

image

This pull request resolves the issue where the GitHub token was not functioning correctly for V1 conversations.

For a demonstration of the updated behavior, please refer to the video below:

https://github.com/user-attachments/assets/b80bc7ee-f5f7-47d9-a527-2a74268cdb9a

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Other (dependency update, docs, typo fixes, etc.)

## Checklist

- [x] I have read and reviewed the code and I understand what the code is doing.
- [x] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves #(issue)

## Release Notes

- [ ] Include this change in the Release Notes.

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:f0a12bf-nikolaik   --name openhands-app-f0a12bf   docker.openhands.dev/openhands/openhands:f0a12bf
```

PR #12525, opened 2026-01-21
fix: remove keycloak_auth cookie requirement for webhook secrets endpoint

tofarr:
## Summary of PR

V1 conversations started from the Slack integration were failing to retrieve GitHub tokens because the `/api/v1/webhooks/secrets` endpoint requires authentication, but the middleware was rejecting requests without a `keycloak_auth` cookie.

**Root cause:** The `/api/v1/webhooks/secrets` endpoint uses `X-Access-Token` header for authentication (via JWT), but the middleware's `_check_tos` method only recognizes:
- `keycloak_auth` cookie
- `Authorization: Bearer ...` header  
- `X-Session-API-Key` header

Since `X-Access-Token` is not in this list, requests to the secrets endpoint from sandboxes were being rejected unless a cookie was also present. Browser-initiated conversations work because they have a cookie, but Slack-initiated conversations don't have a browser session.

**Fix:**
1. Add `/api/v1/webhooks/secrets` to the middleware's `ignore_paths` - This endpoint has its own JWT-based authentication via `X-Access-Token`, so middleware auth is redundant
2. Remove the `keycloak_auth_cookie` workaround from `LiveStatusAppConversationService` - This was added in #11814 as a workaround but is no longer needed

## Demo Screenshots/Videos

N/A - Backend fix

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Fixes GitHub tokens not working for V1 conversations started from Slack integration.

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:fa261a5-nikolaik   --name openhands-app-fa261a5   docker.openhands.dev/openhands/openhands:fa261a5
```

### DIFF
--- a/tests/unit/app_server/test_live_status_app_conversation_service.py
+++ b/tests/unit/app_server/test_live_status_app_conversation_service.py
@@ -75,7 +75,6 @@ class TestLiveStatusAppConversationService:
             openhands_provider_base_url='https://provider.example.com',
             access_token_hard_timeout=None,
             app_mode='test',
-            keycloak_auth_cookie=None,
         )
 
         # Mock user info
@@ -150,10 +149,9 @@ class TestLiveStatusAppConversationService:
 
     @pytest.mark.asyncio
     async def test_setup_secrets_for_git_providers_with_saas_mode(self):
-        """Test _setup_secrets_for_git_providers with SaaS mode (includes keycloak cookie)."""
+        """Test _setup_secrets_for_git_providers with SaaS mode uses LookupSecret with X-Access-Token."""
         # Arrange
         self.service.app_mode = 'saas'
-        self.service.keycloak_auth_cookie = 'test_cookie'
         base_secrets = {}
         self.mock_user_context.get_secrets.return_value = base_secrets
         self.mock_jwt_service.create_jws_token.return_value = 'test_access_token'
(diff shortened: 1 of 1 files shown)


enterprise/server/middleware.py
```diff
@@ -163,6 +163,7 @@ def _should_attach(self, request: Request) -> bool:
             '/oauth/device/authorize',
             '/oauth/device/token',
             '/api/v1/web-client/config',
+            '/api/v1/webhooks/secrets',
```
hieptl, line 166: **question:** If we make this endpoint public, could it introduce any potential security risks? As we prepare to release the shared conversation feature, I would appreciate your guidance on whether someone who receives a shared conversation might have any way of accessing sensitive information or secrets. Thank you! 🙏

hieptl (approved):
Thank you! 🙏

enterprise/server/middleware.py
tofarr: The shared conversations never hit a runtime and definitely never get any keys. This endpoint was designed to take a generated token with a relatively short timeout and use this for authentication, and provide limited access. 

I think this would be a good subject for the code katas this afternoon.

PR #12546, opened 2026-01-22
added graceful deletion of conversations for async response

### LINKED ISSUE #12545: Graceful deletion of conversations
As a user, I don't want to be confused by needless errors or misleading behavior when attempting to delete a conversation.

**Steps to reproduce**:

1. Create a new conversation
2. Delete said conversation
3. Either wait, refresh, or navigate to another page and back
4. Observe errors

**Proposed solution**:

1. Introduce a new status "DELETE" for when a conversation is being deleted.
2. Navigate to home page immediately after selecting to delete a Conversation from /conversation
3. Update all references to reflect this status (Conversations in sidebar, etc)

### If you find this feature request or enhancement useful, make sure to add a 👍 to the issue

charlespettis:
## Summary of PR

Added graceful deletion of conversations.  Introduced new "DELETE" status for conversations.  Disallow user from opening conversation mid-deletion.  Reflect status in Conversations sidebar.  Direct user to home page immediately after deleting conversation (from /conversation).

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

https://github.com/OpenHands/OpenHands/issues/12545

Resolves #12545

## Release Notes

### DIFF
--- a/enterprise/server/routes/billing.py
+++ b/enterprise/server/routes/billing.py
@@ -13,46 +13,33 @@ from server.constants import (
     STRIPE_API_KEY,
 )
 from server.logger import logger
+from starlette.datastructures import URL
 from storage.billing_session import BillingSession
 from storage.database import session_maker
 from storage.lite_llm_manager import LiteLlmManager
 from storage.subscription_access import SubscriptionAccess
 from storage.user_store import UserStore
 
+from openhands.app_server.config import get_global_config
 from openhands.server.user_auth import get_user_id
 
 stripe.api_key = STRIPE_API_KEY
 billing_router = APIRouter(prefix='/api/billing')
 
 
-# TODO: Add a new app_mode named "ON_PREM" to support self-hosted customers instead of doing this
-# and members should comment out the "validate_saas_environment" function if they are developing and testing locally.
-def is_all_hands_saas_environment(request: Request) -> bool:
-    """Check if the current domain is an All Hands SaaS environment.
-
-    Args:
-        request: FastAPI Request object
-
-    Returns:
-        True if the current domain contains "all-hands.dev" or "openhands.dev" postfix
+async def validate_billing_enabled() -> None:
     """
-    hostname = request.url.hostname or ''
-    return hostname.endswith('all-hands.dev') or hostname.endswith('openhands.dev')
-
-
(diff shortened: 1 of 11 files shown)


charlespettis:
One thing I noticed is that the "Recent Conversations" on the home page pulls from a separate state than the sidebar.  Unless there's a good reason it works this way, I think they should be pulling from the same data.  Because right now the conversation panel for recent conversations will not be able to reflect the DELETE status if you are being directed to home page from the /conversation page after deletion. (Does not happen when deleting from sidebar because there is no refresh of data)

frontend/src/components/features/conversation-panel/conversation-panel.tsx
```diff
@@ -107,16 +105,7 @@ export function ConversationPanel({ onClose }: ConversationPanelProps) {
...
-      deleteConversation(
-        { conversationId: selectedConversationId },
-        {
-          onSuccess: () => {
-            if (selectedConversationId === currentConversationId) {
-              navigate("/");
-            }
-          },
-        },
-      );
```
amanape, line 119: Is there a particular reason you’re removing the navigation logic here? This redirects the user back to the home screen after the delete conversation API resolves so that they are no longer in the screen of the conversation they just deleted.

amanape:
> One thing I noticed is that the "Recent Conversations" on the home page pulls from a separate state than the sidebar. Unless there's a good reason it works this way, I think they should be pulling from the same data.

Agreed, feel free to open an issue for this!

frontend/src/components/features/conversation-panel/conversation-panel.tsx
charlespettis: My reasoning for removing the navigation logic was that deleting a conversation takes a while to resolve, so the errors you get ("Lost Connection",etc) can be confusing. I opted to direct the user to the home page immediately, while it loaded, to mitigate that. 

I sort of rushed into making an issue and this PR.  Typically I would have a product owner triage/approve a ticket, but I wasn't sure what the protocol was here. I also was sort of struggling to find a small, good first issue. **If we have to scrap this I won't take it personally.**

Thank you for reminding me about the tests , I will take care of them.

PR #12562, opened 2026-01-22
Fix org billing

tofarr:
## Summary of PR

This PR fixes billing-related issues for organization billing. The key changes are:

### Changes

1. **Async database session for `get_user_by_id_async`**: Swapped to using an async db session (`a_session_maker`) instead of the synchronous session. This is more efficient and also prevents deadlocks due to nested database sessions.

2. **`_get_base_url` now enforces TLS**: The new `_get_base_url` helper function insists on HTTPS (as it should for credit card operations). This saves an unnecessary redirect from http to https on each redirect. Note: The scheme in the K8s pod always appears to be http even though externally it is https.

3. **`validate_billing_enabled` replaces domain-based validation**: The previous `validate_saas_environment` function checked domains to determine if billing should be available. The new `validate_billing_enabled` function verifies feature flags instead, which is more flexible and correct.

## Demo Screenshots/Videos

I ran 20 iterations on staging - adding $10 each time without a stall:
image

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:15a29b7-nikolaik   --name openhands-app-15a29b7   docker.openhands.dev/openhands/openhands:15a29b7
```

### DIFF
--- a/enterprise/server/sharing/sql_shared_conversation_info_service.py
+++ b/enterprise/server/sharing/sql_shared_conversation_info_service.py
@@ -26,6 +26,7 @@ from server.sharing.shared_conversation_models import (
 )
 from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
+from storage.stored_conversation_metadata_saas import StoredConversationMetadataSaas
 
 from openhands.app_server.app_conversation.sql_app_conversation_info_service import (
     StoredConversationMetadata,
@@ -57,7 +58,7 @@ class SQLSharedConversationInfoService(SharedConversationInfoService):
         include_sub_conversations: bool = False,
     ) -> SharedConversationPage:
         """Search for shared conversations."""
-        query = self._public_select()
+        query = self._public_select_with_saas_metadata()
 
         # Conditionally exclude sub-conversations based on the parameter
         if not include_sub_conversations:
@@ -104,14 +105,17 @@ class SQLSharedConversationInfoService(SharedConversationInfoService):
         query = query.limit(limit + 1)
 
         result = await self.db_session.execute(query)
-        rows = result.scalars().all()
+        rows = result.all()
 
         # Check if there are more results
         has_more = len(rows) > limit
         if has_more:
             rows = rows[:limit]
 
-        items = [self._to_shared_conversation(row) for row in rows]
+        items = [
+            self._to_shared_conversation(stored, saas_metadata=saas_metadata)
+            for stored, saas_metadata in rows
+        ]
 
         # Calculate next page ID
         next_page_id = None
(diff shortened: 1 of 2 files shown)


enterprise/storage/user_store.py
```diff
@@ -372,13 +372,15 @@ async def get_user_by_id_async(user_id: str) -> Optional[User]:
         This is the preferred method when calling from an async context as it
         avoids event loop conflicts that can occur with the sync version.
         """
-        with session_maker() as session:
-            user = (
-                session.query(User)
+        async with a_session_maker() as session:
+            logger.info('TRACE:about_to_get_user_from_db')
```
hieptl: **question:** Could you please clarify whether we are planning to introduce the TRACE: prefix when logging in our codebase? Thank you! 🙏

enterprise/storage/user_store.py
```diff
@@ -372,13 +372,15 @@ async def get_user_by_id_async(user_id: str) -> Optional[User]:
...
+            result = await session.execute(
+                select(User)
                 .options(joinedload(User.org_members))
                 .filter(User.id == uuid.UUID(user_id))
-                .first()
             )
+            user = result.scalars().first()
+            logger.info(f'TRACE:got_user_from_db:{user}')
```
hieptl: **question:** Could you please confirm whether we are planning to log the entire set of user data? I just want to ensure we are mindful of any sensitive information that may be included. Thank you! 🙏

enterprise/server/routes/billing.py
```diff
@@ -25,31 +26,15 @@
...
-
-    Args:
-        request: FastAPI Request object
-
-    Raises:
-        HTTPException: If the request is not from an All Hands SaaS environment
+    Validate that the billing feature flag is enabled
     """
-    if not is_all_hands_saas_environment(request):
+    from openhands.app_server.config import get_global_config
```
hieptl: **question:** Could we consider moving this import statement to the top of the file? I was wondering if it was placed here to help avoid a circular import. Thank you! 🙏

enterprise/server/routes/billing.py
hieptl: **question:** The success_callback function currently uses request.base_url directly, while the cancel_callback uses _get_base_url(request). Could you please clarify whether this difference is intentional? Thank you! 🙏

enterprise/server/routes/billing.py
```diff
@@ -25,31 +26,15 @@
...
+
+    config = get_global_config()
+    web_client_config = await config.web_client.get_web_client_config()
+    if not web_client_config.feature_flags.enable_billing:
         raise HTTPException(
             status_code=status.HTTP_403_FORBIDDEN,
             detail='Checkout sessions are only available for All Hands SaaS environments',
```
hieptl: **suggestion:** The error message still references ‘All Hands SaaS environments,’ which reflects the previous domain-based check. Would it be possible to update it to reference the billing feature flag instead? I believe this would make the message more accurate. Thank you! 🙏

enterprise/storage/user_store.py
tofarr: No - I added a bunch of debug statements since this was happening only in staging and not locally - I thought I had removed all of them but missed this one. Removing it now

enterprise/storage/user_store.py
tofarr: No - I added a bunch of debug statements since this was happening only in staging and not locally - I thought I had removed all of them but missed this one. Removing it now

enterprise/server/routes/billing.py
tofarr: I put it here to avoid circular imports - but it looks like it can safely be at the top of the file

Issue #12536, opened 2026-01-21
[Bug]: WORK_HOSTS_SKILL_FOOTER should be conditional based on Remote/Docker Sandbox


cbagwell:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

The WORK_HOSTS_SKILL_FOOTER was recently added as a very nice usability improvement so that LLM was told what ports to use so that apps are automatically seen in the Apps tab in UI.  The problem is that for Docker sandboxes, the instruction should describe that hard coded ports 8011 and 8012 should be used and that $WORKER_1 and $WORKER_2 are different ports but accessible externally.

To reproduce in Docker environment, ask OH to create a Hello World Flask app and confirm that it uses $WORKER_1 for port which is inaccessible externally and not visible in UI.  If you describe above issue to LLM, it will update the port and then the app will become accessible at $WORKER_1 address externally and start showing up in App tab.

### OpenHands Installation

Docker command in README

### OpenHands Version

main

### Operating System

None

cbagwell:
I use these values when starting docker in case it makes a behavior difference.  I see some references to using WEB_HOST that might cause differences in WORKER_1/2.

```
    -e OH_WEB_URL=http://$(hostname -f):3000 \
    -e SANDBOX_CONTAINER_URL_PATTERN=http://$(hostname -f):{port} \
```

cbagwell:
I just now noticed the logic I was referring to was deleted in #12434 after I created this issue.  I am going to leave this open for now as I noticed WORK_HOSTS_SKILL_FOOTER was left behind as dead code.

PR #12434, opened 2026-01-15
feat(backend): load skills from agent server

hieptl:
> Note: I suggest we hold off on merging this pull request until the next SDK version is released. In the meantime, the pull request is ready and available for review.

## Summary of PR

This PR refactors the skill loading architecture to centralize all skill loading logic in the agent-server, with the app-server acting as a thin proxy. This change simplifies the app-server codebase, improves maintainability, and ensures consistent skill loading behavior across all sources.

#### **1. Centralized Skill Loading via Agent-Server API**

The app-server now delegates all skill loading to the agent-server through a single API call to `/api/skills`. The agent-server handles:
- **Public skills** (from OpenHands/skills GitHub repo)
- **User skills** (from `~/.openhands/skills/`)
- **Organization skills** (from `{org}/.openhands` repo)
- **Project/repo skills** (from workspace `.openhands/skills/`)
- **Sandbox skills** (from exposed URLs)

**Before:** App-server directly accessed file systems, cloned repositories, and merged skills from multiple sources.

**After:** App-server builds configuration objects and makes a single API call to agent-server, which handles all skill loading and merging.

#### **2. Removed File System Operations**

Removed the following functions from `skill_loader.py` that performed direct file system operations:
- `load_global_skills()`
- `load_repo_skills()`
- `load_org_skills()`
- `merge_skills()`
- `_determine_repo_root()`
- `_read_file_from_workspace()`
- `_load_special_files()`
- `_find_and_load_skill_md_files()`
- `_find_and_load_global_skill_files()`
- `_clone_org_repository()`
- `_load_skills_from_org_directories()`
- `_merge_org_skills_with_precedence()`
- `_cleanup_org_repository()`
- `_validate_repository_for_org_skills()`

#### **3. New Agent-Server Integration Functions**

Added functions to build configurations and communicate with agent-server:
- `build_org_config()` - Builds organization configuration with authentication
- `build_sandbox_config()` - Builds sandbox configuration with exposed URLs
- `load_skills_from_agent_server()` - Makes API call to agent-server's `/api/skills` endpoint
- `_convert_skill_info_to_skill()` - Converts API response to Skill objects

#### **4. Type Safety Improvements**

Replaced all `dict[str, Any]` usages with strongly-typed Pydantic models:
- `OrgConfig` - Organization configuration model
- `SandboxConfig` - Sandbox configuration model
- `ExposedUrlConfig` - Exposed URL configuration model
- `SkillInfo` - Skill information model from API responses

This provides:
[...]

PR #12424, opened 2026-01-14
fix: Add WORKER_1 and WORKER_2 env vars to remote sandbox service

neubig:
## Summary

This PR fixes the App tab not becoming active when agents start web servers on **OpenHands Cloud (V1 conversations)**.

## Root Cause

When agents start a web server, the runtime info tells them to use ports like 12000/12001. However, the agents often need to programmatically reference these ports via environment variables.

**The issue**: `docker_sandbox_service.py` sets `WORKER_1` and `WORKER_2` environment variables, but `remote_sandbox_service.py` (used by OpenHands Cloud for **V1 conversations**) does not.

As a result, agents on Cloud cannot reference `$WORKER_1` or `$WORKER_2` to start servers on the correct ports.

## Changes

1. **`openhands/app_server/sandbox/remote_sandbox_service.py`** (V1 app server)
   - Added `WORKER_1=12000` and `WORKER_2=12001` environment variables in `_init_environment()`

2. **`openhands/app_server/app_conversation/skill_loader.py`** (V1 skill loading)
   - Added footer text to the work_hosts skill explaining that `WORKER_1` and `WORKER_2` environment variables contain the port numbers
   - Includes example configurations for Express, Vite, Flask, and Next.js

3. **`tests/unit/app_server/test_remote_sandbox_service.py`**
   - Updated tests to expect the new environment variables

## Architecture Notes

This fix targets the V1 app server (`openhands/app_server/`). The environment variables are passed to the sandbox container where the SDK agent runs. The SDK itself does not need changes - it simply has access to `$WORKER_1` and `$WORKER_2` environment variables in the shell.

## Testing

- Unit tests updated and pass
- Manual testing recommended on OpenHands Cloud to verify App tab activation

HUMAN: confirmed to work via a preview deployment

Screenshot 2026-01-14 at 9 43 07 PM

## Related Issues

Related to: https://github.com/OpenHands/OpenHands/issues/12049

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:ce570ce-nikolaik   --name openhands-app-ce570ce   docker.openhands.dev/openhands/openhands:ce570ce
```

PR #12566, opened 2026-01-22
fix: Provide external URLs to agents via WORKER_*_URL environment variables

shanemort1982:
## Problem

Agents were telling users incorrect port numbers when serving web applications. The `WORKER_*` environment variables contained container ports (8011, 8012) instead of the externally accessible host ports, causing agents to provide URLs like `http://192.168.1.171:8011` which were not accessible to users.

## Solution

This fix provides **two sets of environment variables** to give agents both the container port (for binding servers) and the external URL (for user access):

1. **`WORKER_1`, `WORKER_2`**: Container ports (8011, 8012) - for starting servers
2. **`WORKER_1_URL`, `WORKER_2_URL`**: Full external URLs - for telling users where to access the app

### Example Environment Variables

```bash
WORKER_1=8011                              # Container port to bind server
WORKER_1_URL=http://192.168.1.171:50349  # External URL for users
WORKER_2=8012
WORKER_2_URL=http://192.168.1.171:40105
```

## Changes

**File**: `openhands/app_server/sandbox/docker_sandbox_service.py`

```python
# Lines 343-346
for exposed_port in self.exposed_ports:
    host_port = self._find_unused_port()
    port_mappings[exposed_port.container_port] = host_port
    env_vars[exposed_port.name] = str(exposed_port.container_port)  # WORKER_1=8011
    url = self.container_url_pattern.format(port=host_port)
    env_vars[f'{exposed_port.name}_URL'] = url  # WORKER_1_URL=http://...
```

### Test 1: Hello World App
**Task**: "Create a Hello World web page, serve on $WORKER_1, tell me the URL from $WORKER_1_URL"

**Result**: ✅ SUCCESS
- Agent started server on port **8011** (container port from `$WORKER_1`)
- Agent provided URL **http://192.168.1.171:45503** (from `$WORKER_1_URL`)
- URL was **accessible** and served the Hello World page correctly

### Test 2: TODO List App
**Task**: "Create a simple TODO list app with HTML/CSS/JavaScript. Start the server on port $WORKER_1 and tell me the URL from $WORKER_1_URL"

**Result**: ✅ SUCCESS  
- Agent started server on port **8011** (container port from `$WORKER_1`)
- Agent provided URL **http://192.168.1.171:50349** (from `$WORKER_1_URL`)
- URL was **accessible** and served the TODO app correctly

### Verification

```bash
# Verified port mapping
$ docker port oh-agent-server-<id> | grep 8011
8011/tcp -> 0.0.0.0:50349

# Verified app accessibility
$ curl -s http://192.168.1.171:50349/ | head -5
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <title>Simple TODO List</title>
```

## Impact

This fix enables agents to:
1. ✅ Start servers on the correct container port using `$WORKER_1`
2. ✅ Provide users with the correct external URL using `$WORKER_1_URL`
3. ✅ Avoid confusion and failed connections from incorrect port numbers

## Related Issues

This resolves the issue where agents would tell users to access applications on container ports that were not exposed to the host network.

### DIFF
--- a/openhands/app_server/config.py
+++ b/openhands/app_server/config.py
@@ -74,11 +74,14 @@ def get_default_persistence_dir() -> Path:
 def get_default_web_url() -> str | None:
     """Get legacy web host parameter.
 
-    If present, we assume we are running under https.
+    If present, we assume we are running under https unless a protocol is already specified.
     """
     web_host = os.getenv('WEB_HOST')
     if not web_host:
         return None
+    # Check if protocol is already specified
+    if web_host.startswith('http://') or web_host.startswith('https://'):
+        return web_host
     return f'https://{web_host}'
 
 


cbagwell:
Interesting.  I recently wrote #12536 because I'm seeing different behavior.  Both WORKER_1 and WORKER_1_URL have the redirected port which confuses LLM on what port to use.  In your case, they are both the fixed port?

I wonder if it is related to fact I use OH_WEB_URL while your using WEB_HOST?

Regardless, I would love if the end results of both our configs behaves as you describe since it would resolve #12536.

cbagwell:
I was confused when I wrote that comment and thinking WORKER_*_URL already existed instead of this PR adding it but what I was thinking of was the fact that it was injecting a skill that described the exeternal port and then in #12424 it was refined to also describe how to bind to port so it would be accessible without CORS errors but used wrong port for docker.

I see that same logic was deleted in #12434 2 days ago so that agent-server is source of loading skills.  I haven't yet tested that and I'm not sure if the skills for WORKER URL and Ports still works in some other form but in the end I hope they do end up describing the 2 different port values correctly.

cbagwell:
I have created OpenHands/software-agent-sdk#1819 requesting to improve the work_host skill to mention correct ports for Docker although I do not think it is currently passed enough information to be able to do that.

PR #12570, opened 2026-01-22
chore(frontend): convo tab only renders active/selected tab

HeyItsChloe:
## Summary of PR
- Only poll when App tab is rendered
- Only render active tab. Avoid rendering all tabs and hiding inactive tabs via css

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/85da21f9-4147-4c0e-b467-1c1552409aed

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves #(issue)
## Release Notes

### DIFF
--- a/enterprise/downgrade_migrated_users.py
+++ b/enterprise/downgrade_migrated_users.py
@@ -1,0 +1,205 @@
+#!/usr/bin/env python
+"""
+Downgrade script for migrated users.
+
+This script identifies users who have been migrated (already_migrated=True)
+and reverts them back to the pre-migration state.
+
+Usage:
+    # Dry run - just list the users that would be downgraded
+    python downgrade_migrated_users.py --dry-run
+
+    # Downgrade a specific user by their keycloak_user_id
+    python downgrade_migrated_users.py --user-id <user_id>
+
+    # Downgrade all migrated users (with confirmation)
+    python downgrade_migrated_users.py --all
+
+    # Downgrade all migrated users without confirmation (dangerous!)
+    python downgrade_migrated_users.py --all --no-confirm
+"""
+
+import argparse
+import asyncio
+import sys
+
+# Add the enterprise directory to the path
+sys.path.insert(0, '/workspace/project/OpenHands/enterprise')
+
+from server.logger import logger
+from sqlalchemy import select, text
+from storage.database import session_maker
+from storage.user_settings import UserSettings
+from storage.user_store import UserStore
+
+
+def get_migrated_users() -> list[str]:
+    """Get list of keycloak_user_ids for users who have been migrated.
+
+    This includes:
+    1. Users with already_migrated=True in user_settings (migrated users)
+    2. Users in the 'user' table who don't have a user_settings entry (new sign-ups)
+    """
+    with session_maker() as session:
+        # Get users from user_settings with already_migrated=True
+        migrated_result = session.execute(
+            select(UserSettings.keycloak_user_id).where(
+                UserSettings.already_migrated.is_(True)
+            )
+        )
+        migrated_users = {row[0] for row in migrated_result.fetchall() if row[0]}
+
+        # Get users from the 'user' table (new sign-ups won't have user_settings)
+        # These are users who signed up after the migration was deployed
+        new_signup_result = session.execute(
+            text("""
+                SELECT CAST(u.id AS VARCHAR)
+                FROM "user" u
+                WHERE NOT EXISTS (
+                    SELECT 1 FROM user_settings us
+                    WHERE us.keycloak_user_id = CAST(u.id AS VARCHAR)
+                )
+            """)
+        )
+        new_signups = {row[0] for row in new_signup_result.fetchall() if row[0]}
+
+        # Combine both sets
+        all_users = migrated_users | new_signups
+        return list(all_users)
+
+
+async def downgrade_user(user_id: str) -> bool:
+    """Downgrade a single user.
+
+    Args:
+        user_id: The keycloak_user_id to downgrade
+
+    Returns:
+        True if successful, False otherwise
+    """
+    try:
+        result = await UserStore.downgrade_user(user_id)
+        if result:
+            print(f'✓ Successfully downgraded user: {user_id}')
+            return True
+        else:
+            print(f'✗ Failed to downgrade user: {user_id}')
+            return False
+    except Exception as e:
+        print(f'✗ Error downgrading user {user_id}: {e}')
+        logger.exception(
+            'downgrade_script:error',
+            extra={'user_id': user_id, 'error': str(e)},
+        )
+        return False
+
+
+async def main():
+    parser = argparse.ArgumentParser(
+        description='Downgrade migrated users back to pre-migration state'
+    )
+    parser.add_argument(
+        '--dry-run',
+        action='store_true',
+        help='Just list users that would be downgraded, without making changes',
+    )
+    parser.add_argument(
+        '--user-id',
+        type=str,
+        help='Downgrade a specific user by keycloak_user_id',
+    )
+    parser.add_argument(
+        '--all',
+        action='store_true',
+        help='Downgrade all migrated users',
+    )
+    parser.add_argument(
+        '--no-confirm',
+        action='store_true',
+        help='Skip confirmation prompt (use with caution!)',
+    )
+
+    args = parser.parse_args()
+
+    # Get list of migrated users
+    migrated_users = get_migrated_users()
+    print(f'\nFound {len(migrated_users)} migrated user(s).')
+
+    if args.dry_run:
+        print('\n--- DRY RUN MODE ---')
+        print('The following users would be downgraded:')
+        for user_id in migrated_users:
+            print(f'  - {user_id}')
+        print('\nNo changes were made.')
+        return
+
+    if args.user_id:
+        # Downgrade a specific user
+        if args.user_id not in migrated_users:
+            print(f'\nUser {args.user_id} is not in the migrated users list.')
+            print('Either the user was not migrated, or the user_id is incorrect.')
+            return
+
+        print(f'\nDowngrading user: {args.user_id}')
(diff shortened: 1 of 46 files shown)


hieptl:
Hello @HeyItsChloe,

We should avoid including the Linear link in the pull request description, as it is intended for internal use only.

If there is no existing GitHub issue, we should create one and reference it instead. Otherwise, we can simply leave that section blank.

Thank you very much! 🙏

hieptl:
**suggestion:** Could you please add a short demo video to demonstrate what the expected behavior looks like? This would be very helpful. Thank you! 🙏

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-content/tab-wrapper.tsx
```diff
@@ -2,14 +2,9 @@ import { ReactNode } from "react";
...
 }
 
-export function TabWrapper({ isActive, children }: TabWrapperProps) {
-  return (
-    <div className={cn("absolute inset-0", isActive ? "block" : "hidden")}>
-      {children}
-    </div>
-  );
+export function TabWrapper({ children }: TabWrapperProps) {
+  return <div className={cn("absolute inset-0")}>{children}</div>;
```
hieptl: **suggestion:** Using cn in this case may be unnecessary. We typically use cn when we need to conditionally combine classes based on specific conditions. Thank you! 🙏

HeyItsChloe:
> suggestion: Could you please add some tests to cover your changes? This would help us improve code quality and reduce the risk of regressions. Thank you! 🙏

✅ Added! Didn't add unit tests initially because I didn't create a new component / I assumed the test were already created. 
❓QQ: When adding unit tests to existing components, should we cover the entire feature or only the components related to the ticket? In this case, when adding a test file for `conversation-tab-content.tsx`, should we also create tests for each tabs content if it doesnt already exist (i.e `served-tab.tsx`, `planner-tab.tsx`, etc)? @hieptl

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-content/tab-wrapper.tsx
HeyItsChloe: Fixed!

HeyItsChloe:
> **suggestion:** Could you please add a short demo video to demonstrate what the expected behavior looks like? This would be very helpful. Thank you! 🙏

✅ Added! Should I add vids to each PR from now on?

hieptl:
> ✅ Added! Should I add vids to each PR from now on?

This is not required. However, when working on front-end and back-end tickets, it would be very helpful if the outputs could be demonstrated through a screenshot or a short video. This would support reviewers in better understanding the expected behavior and validating the changes more efficiently.

Thank you very much! 🙏

hieptl:
> ❓QQ: When adding unit tests to existing components, should we cover the entire feature or only the components related to the ticket? In this case, when adding a test file for conversation-tab-content.tsx, should we also create tests for each tabs content if it doesnt already exist (i.e served-tab.tsx, planner-tab.tsx, etc)? 

Unit tests should test a single unit in isolation                                                                        
                                                                                                                           
conversation-tab-content.tsx has one responsibility: render the correct tab based on selectedTab state. Your tests should verify that behavior - not the internal behavior of ServedTab, PlannerTab, etc.                                         
                                                                                                                          
You've done this correctly by mocking the child components:                                                              

```ts
vi.mock("#/routes/served-tab", () => ({                                                                                  
  default: () => <div data-testid="served-tab-content">Served Tab Content</div>,                                         
}));
```
                                                                                                                          
This isolates the unit under test. Your tests verify that conversation-tab-content renders the served-tab-content testid when that tab is selected - which is exactly what this component is responsible for.                                     
                                                                                                                          
Why not test child components in this PR?                                                                                
                                                                                                                          
1. Scope creep - Adding tests for 6 additional components expands this PR significantly beyond the ticket's intent       
2. Separation of concerns - Each component should have its own test file testing its own behavior                        
3. Maintenance - If ServedTab changes, you'd have to update tests in multiple places                                     
4. Review burden - Larger PRs are harder to review and more likely to introduce issues                                   
                                                                                                                          
When would you add tests for child components?                                                                           
                                                                                                                          
- When you modify them directly                                                                                          
- As a separate ticket/PR specifically for improving test coverage                                                       
- If they have zero coverage and are critical paths (but still as a separate effort)                                     
                                                                                                                          
Your current approach is correct. The tests for conversation-tab-content.tsx verify the tab switching, title display, and rendering logic - which is exactly what you changed in this PR.

I believe that @amanape may be able to provide additional input here to help further clarify your question and ensure we have a shared understanding.

Thank you all very much! 🙏

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-content/conversation-tab-content.tsx
```diff
@@ -18,120 +18,65 @@ const VSCodeTab = lazy(() => import("#/routes/vscode-tab"));
...
-  const isVSCodeActive = selectedTab === "vscode";
-  const isTerminalActive = selectedTab === "terminal";
-  const isPlannerActive = selectedTab === "planner";
-
-  // Define tab configurations
-  const tabs = [
-    { key: "editor", component: EditorTab, isActive: isEditorActive },
-    {
-      key: "browser",
+  const TAB_CONFIG = {
```
hieptl: **suggestion:**

- TAB_CONFIG is defined inside the component and recreated on every render. The useMemo at line 52-55 references it but doesn't include it in the dependency array (correctly, since including it would cause infinite loops). However, this pattern is error-prone.                                                                                                  
- Fix: Move TAB_CONFIG outside the component to make it a module-level constant: 

```ts
const TAB_CONFIG = { ... }; // Move outside ConversationTabContent
```

Thank you! 🙏

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-content/conversation-tab-content.tsx
```diff
@@ -18,120 +18,65 @@ const VSCodeTab = lazy(() => import("#/routes/vscode-tab"));
...
-    isVSCodeActive,
-    isTerminalActive,
-    isPlannerActive,
-  ]);
+  const activeTab = useMemo(
+    () => TAB_CONFIG[selectedTab ?? "editor"],
+    [selectedTab],
+  );
 
-  if (shouldShownAgentLoading) {
```
hieptl, line 114: **question:**

Removal of shouldShownAgentLoading check may be a regression (conversation-tab-content.tsx)                           
- The original code at lines 61-63 (old file) showed <ConversationLoading /> when shouldShownAgentLoading was true     
(agent in INIT/LOADING state).                                                                                           
- This was removed entirely. The new Suspense fallback only handles lazy-loading, NOT agent loading states.            

Was this intentional? If agent loading state should still show a loading screen in the tab area, this is a  
regression. The PR description doesn't mention this behavior change.

Thank you! 🙏

frontend/__tests__/components/features/conversation/conversation-tab-content.test.tsx
```diff
@@ -0,0 +1,324 @@
...
+        wrapper: createWrapper(),
+      });
+
+      await waitFor(() => {
+        expect(screen.getByTestId("terminal-tab-content")).toBeInTheDocument();
+      });
+
+      // The key should include the conversation ID for terminal tab
+      // This ensures Terminal remounts when conversation changes
+      expect(container).toBeTruthy();
```
hieptl: **suggestion:** Tests claim to verify terminal key includes conversation ID, but just assert expect(container).toBeTruthy() which doesn't actually verify the key value. Thank you! 🙏

frontend/__tests__/components/features/conversation/conversation-tab-content.test.tsx
```diff
@@ -0,0 +1,324 @@
...
+
+  describe("Lazy loading", () => {
+    it("should show loading fallback while component is loading", async () => {
+      // Create a delayed mock to simulate lazy loading
+      let resolvePromise: () => void;
+      const delayedPromise = new Promise<void>((resolve) => {
+        resolvePromise = resolve;
+      });
+
+      vi.doMock("#/routes/browser-tab", () => ({
```
hieptl: **suggestion:** 

- Uses vi.doMock after vi.mock already set up the mock at the top of the file. The doMock doesn't override the existing mock, making this test ineffective.                                                                                     
- The test passes but doesn't actually test the Suspense loading fallback.                                             

Fix: Remove this test or implement it properly (e.g., using vi.resetModules() before vi.doMock).

Thank you! 🙏

frontend/src/components/features/conversation/conversation-tabs/conversation-tab-content/conversation-tab-content.tsx
HeyItsChloe: Ah, this was not intentional. Added this logic back!

frontend/__tests__/components/features/conversation/conversation-tab-content.test.tsx
HeyItsChloe: Thanks for catching this! I've removed the ineffective lazy loading test in commit 77b9389b6. The `vi.doMock` was indeed not overriding the existing `vi.mock` setup, making the test ineffective.

HeyItsChloe: Openhands made the above comment & the removal commit 😅. I updated the test mock -  I think it works properly now.

frontend/__tests__/components/features/conversation/conversation-tab-content.test.tsx
HeyItsChloe: Ah. Updated!

PR #12595, opened 2026-01-25
Pass container port instead of host port to Docker

### LINKED ISSUE #12049: [App tab] Improve ability for Agent to use the App tab
Currently, getting the OpenHands agent to use the 'App' tab in the GUI is very difficult. Most users do not know the exact prompts to use to ensure the web server displays the app in this tab. When it is working, the experience is better for the end user because they can interact directly with the web app, rather than rely on screenshots from the browser.

- Understand the prompts or steps required by the user to get a  web app to appear in the 'App' tab. Document this.
- Improve the OpenHands agent, or suggest a feature, that allows users to load web apps into the 'App' tab more reliably and predictably
- Change the name of this from 'App' to 'Web App' if that's the type of application that can be displayed here

### LINKED ISSUE #12271: [Bug]: App View never worked for me, and Browser view rarely works.

When building any app, after it's done, the App and Browser view in self-hosted OpenHands have no content. I don't see any errors in the terminal (with openhands serve). See the attached screenshots.

Image

Image

Browser console is throwing all sorts of errors trying to connect to the localhost ports, that are actually shown to be open on agent.

Image

Image

See attached for log from the agent container.

[agentlogs.txt](https://github.com/user-attachments/files/24443554/agentlogs.txt)

### OpenHands Installation

Docker command in README

### OpenHands Version

_No response_

### Model Name

claude-4-5-sonnet, zai-glm-4.7

### Operating System

WSL on Windows

_No response_

cbagwell:
## Summary of PR

WORKER_1/WORKER_2 should be port for apps to listen on so that they can be accessed externally but currently the Docker runtime sets this to the unreachable dynamic ports meant for host side usage.

This also fixes work_host skill in Docker so that it lists the correct port to listen on so that LLM chooses correct port when starting servers.

    The user has access to the following hosts for accessing
    a web application, each of which has a corresponding port:

    * http://hostname:X (8011)
    * http://hostname:Y (8012)

## Demo Screenshots/Videos

n/a

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Ask OpenHands to create Hello, World flask app with CORS support and run it.  It will use WORKER_1 port (dynamic host port) because of work_host skill but app will not be accessible externally and the App tab will not auto-display the app.  This fix will cause it to use port 8011 with Docker runtime which will make the app accessible and be visible in App tab.

Resolves #(issue)

n/a

## Release Notes


cbagwell:
Tested with the following prompt and confirmed the agent used port 8011, the App tab displayed the app using the correct externally visible URL (why CORS is mentioned), and also accessible from my browser.  The work_host skill prompt could debatably be improved if it included the need for CORS to work correctly with the App tab.

"Build a Flask app that displays a pretty 'Hello, World!' webpage, ensuring it runs on an externally visible port and has CORS enabled."

image

jpelletier1:
Thank you @cbagwell! Going to link this to #12049 for record keeping. Will see if we can find a maintainer to review.

cbagwell:
I'v rebased to main to resolve a merge conflict with new host networking mode.   The host networking was mismatched and using host_port for one area and container_port for another area.  Both are same value so either was OK but I aligned with bridge networking for consistency.

cbagwell:
I've confirmed the rebase works correctly for new host networking mode in addition to default bridge networking.  WORKER_1/2 still are fixed at 8011/8012 (same as bridge mode) and this is what work_host skill looks like for host networking mode:

```
The user has access to the following hosts for accessing a web application, each of which has a corresponding port:
* http://hostname:8011 (port 8011)
* http://hostname:8012 (port 8012)
```

I don't think the CI failure can be related to my changes but if there is something I can do to address then please let me know.

cbagwell:
I re-reviewed and discovered I had mistakenly included uv.lock in commit which was breaking github tests.  I also rebased to latest main while fixing that.

enyst (approved):
Thank you!

PR #12618, opened 2026-01-27
fix: Use recommended PRIVATE-TOKEN header for GitLab authentication

WangYuhang-CN:
## Summary of PR

The current GitLab integration uses `Authorization: Bearer` header for authentication, but this fails with some self-hosted GitLab instances. According to [GitLab's official documentation](https://docs.gitlab.com/ee/api/rest/index.html#authentication), the **recommended** authentication method is using the `PRIVATE-TOKEN` header.

### Current Behavior
```python
return {
    'Authorization': f'Bearer {self.token.get_secret_value()}',
}
```
This causes authentication failures with certain self-hosted GitLab configurations.

## Solution

Change the authentication method to use the officially recommended `PRIVATE-TOKEN` header instead of `Bearer` token.

### Changes Made
```python
return {
    'PRIVATE-TOKEN': self.token.get_secret_value(),
}
```

## Benefits
1. **Better compatibility**: Works with all GitLab instances (GitLab.com and self-hosted)
2. **Follows best practices**: Uses GitLab's recommended authentication method

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes


mamoodi:
I'm going to close this PR as there doesn't seem like there will be more activity. Please comment if I am wrong and I can reopen.

PR #12631, opened 2026-01-27
feat(frontend): SaaS NUE profile questions

HeyItsChloe:
## Summary of PR
1. Created intermediary screen for onboarding "new" users. 
- "New" user: has no org account, personal account, or org invite
- Progressive screen that prompts the user onboarding questions at sign up

2. Replaced the `useIsOnTosPage` hook with a general hook that can be used with any intermediary screens

3. FE only. Endpoints or migrations needed to store user responses not included
- `/use-submit-onboarding.ts` placeholder

4. Added a temp feature flag `ENABLE_ONBOARDING`. Can be removed once the BE portion of this feature is complete.

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/82693687-82e9-44b8-9240-e63c90bc4294

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

Created onboarding flow for new SaaS users.


enyst:
Hey @HeyItsChloe ! Nice PR, thank you!

I have a few tiny suggestions, if I may. Assuming I understand correctly and these screens are on the default (Local Web deployment):
- a little replacement Cloud (hosted by us) -> Cloud (hosted)
- (maybe) switch places: Local Environment <-> Cloud (hosted)

Happy to discuss as you see fit, just looking from a few different perspectives here I guess.

frontend/src/routes/onboarding-form.tsx
```diff
@@ -0,0 +1,164 @@
...
+      {
+        id: "option4",
+        labelKey: I18nKey.ONBOARDING$STEP2_OPTION4,
+      },
+    ],
+  },
+  {
+    id: "step3",
+    titleKey: I18nKey.ONBOARDING$STEP3_TITLE,
+    subtitleKey: I18nKey.ONBOARDING$STEP1_SUBTITLE,
```
hieptl: **suggestion:**

There is a small point of clarification regarding the subtitleKey usage:

```ts
subtitleKey: I18nKey.ONBOARDING$STEP1_SUBTITLE
```

Currently, all three steps reference STEP1_SUBTITLE. Could you please confirm whether this is intentional?

If the subtitles are meant to be different for each step, it might be helpful to create separate keys for clarity and maintainability. Alternatively, if sharing the same subtitle is the intended behavior, adding a brief comment or documentation could make this clearer for future readers.

Thank you! 🙏

frontend/src/routes/onboarding-form.tsx
```diff
@@ -0,0 +1,164 @@
...
+      {
+        id: "option2",
+        labelKey: I18nKey.ONBOARDING$STEP1_OPTION2,
+      },
+      {
+        id: "option3",
+        labelKey: I18nKey.ONBOARDING$STEP1_OPTION3,
+      },
+      {
+        id: "option4",
```
hieptl: **suggestion:** 

There is a small improvement opportunity regarding the option IDs in frontend

Currently, the IDs use generic values such as option1, option2, which do not convey much meaning when viewed in analytics or stored in the database. This can make data interpretation and reporting more difficult over time.

It may be more helpful to use semantic IDs, such as individual_dev, startup_team, and similar descriptive values. This would improve clarity, make the data more self-explanatory, and support better analysis in the future.

Thank you! 🙏

frontend/src/components/features/onboarding/step-header.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+function StepHeader({
+  title,
+  subtitle,
+  currentStep,
+  totalSteps,
+}: StepHeaderProps) {
+  return (
+    <div data-testid="step-header" className="flex flex-col gap-[32px]">
+      <span className="text-[39px] font-semibold text-content">{title}</span>
```
hieptl: **suggestion:** we should use Typography.Text from the /frontend/src/ui folder instead of using span. Thank you! 🙏

frontend/src/components/features/onboarding/step-option.tsx
```diff
@@ -0,0 +1,35 @@
...
+      }}
+      className={cn(
+        "w-full py-8 px-5 rounded-[12px] border-2 cursor-pointer transition-all duration-200",
+        "hover:border-white",
+        "focus:outline-none focus:ring-2 focus:ring-white/50",
+        selected ? "border-white" : "border-neutral-600",
+      )}
+    >
+      <div className="flex flex-col gap-1">
+        <span className="text-sm font-medium text-content">{label}</span>
```
hieptl: suggestion: we should use Typography.Text from the /frontend/src/ui folder instead of using span. Thank you! 🙏

enterprise/server/routes/auth.py
```diff
@@ -47,6 +49,29 @@
 token_manager = TokenManager()
 
 
+def needs_onboarding(user: User) -> bool:
```
hieptl: **suggestion:** Since this is a router file, it might be more appropriate to move this logic into org_store.py and org_service.py instead. Thank you! 🙏

enterprise/server/routes/auth.py
```diff
@@ -47,6 +49,29 @@
...
+    # New users must complete onboarding.
+
+    with session_maker() as session:
```
hieptl: **suggestion:**

In enterprise/server/routes/auth.py, the needs_onboarding function is implemented using a synchronous database session:

```python
def needs_onboarding(user: User) -> bool:
    with session_maker() as session:  # Synchronous
        user_id = user.id
        personal_org = session.query(Org).filter(Org.name == personal_org_name).first()
        # ...
```

This function is then called from an async route handler:

```python
@oauth_router.get('/keycloak/callback')
async def keycloak_callback(...):
    # ...
    elif needs_onboarding(user):  # Sync call in async context
```

**The Concern**

Using session_maker() (synchronous) inside an async route handler can block the event loop. This may lead to a few unintended side effects:
1. Performance impact – other requests are blocked while waiting for database I/O.
2. Reduced concurrency – this limits the benefits of using FastAPI’s async capabilities.
3. Inconsistency – the rest of keycloak_callback follows async patterns, so this creates a mismatch in design.

**Suggested Improvement**

To keep everything aligned with async best practices, it would be helpful to make needs_onboarding fully asynchronous and use a_session_maker instead:

```python
async def needs_onboarding(user: User) -> bool:
    async with a_session_maker() as session:
        user_id = user.id
        personal_org_name = f'user_{user_id}_org'

        # Check 1: Does the user have a personal org?
        result = await session.execute(
            select(Org).where(Org.name == personal_org_name)
        )
        personal_org = result.scalar_one_or_none()
        if personal_org:
            return False

        # Check 2: Does the user have any org membership (any status)?
        result = await session.execute(
            select(OrgMember).where(OrgMember.user_id == user_id)
        )
        any_membership = result.scalar_one_or_none()
        if any_membership:
            return False

        return True
```

And then update the call site accordingly:

```python
elif await needs_onboarding(user):
```

Thank you! 🙏

HeyItsChloe:
> [...] reduce flexibility over time. A more data-driven approach could make future adjustments easier and faster. Thank you very much for considering this. 🙏

@hieptl Thanks! 

I agree with those UI changes. There are some updates to the UI paul just finished - including the "back" button. 

I held off on the database work because I’m not sure whether the completed or in-progress backend work for NUE or Orgs already covers these user flows. I can outline the tables needed to store the questions and responses and add my proposed approach to this ticket for review (LMK of there's a different process for this).

HeyItsChloe:
```mermaid
erDiagram
    User {
        uuid id PK
        boolean onboarding_completed "Used by needs_onboarding()"
    }

    UserSettings {
        int id PK
        string keycloak_user_id FK
        json form_responses "JSON Obj stores all form answers"
    }

    FormQuestion {
        int id PK
        string question_key UK "role, org_size, satisfaction"
        string form_type "onboarding, survey, feedback"
        string label_i18n_key "ONBOARDING$ORG_SIZE"
        string input_type "single_select, multi_select, rating"
        boolean is_required
        datetime created_at
    }

    FormOption {
        int id PK
        int question_id FK
        string option_key UK "software_engineer, very_satisfied"
        string label_i18n_key "ONBOARDING$2-10"
        int display_order
        boolean is_active
        datetime created_at
    }

    User ||--o| UserSettings : "has settings"
    FormQuestion ||--o{ FormOption : "has many"
```

Proposed Database Design for Onboarding (& Future) Forms:

Goals: 
- a dynamic system to store user form responses 
- a flexible, reusable form system that stores questions and answer options without hardcoding in the frontend.

DB Key Points:
- User.onboarding_completed - Boolean flag used by needs_onboarding() to determine if user should see onboarding flow
- UserSettings.form_responses - JSON column storing all user responses, grouped by form type (e.g., onboarding, survey)
- FormQuestion / FormOption - New database tables storing questions and options dynamically
- i18n Support - Questions and options store i18n keys, translations handled by existing frontend system

cc @hieptl , @amanape  - Here's what I was thinking. LMK how this can be improved / if this should be added with this ticket

HeyItsChloe:
> Hey @HeyItsChloe ! Nice PR, thank you!
> 
> I have a few tiny suggestions, if I may. Assuming I understand correctly and these screens are on the default (Local Web deployment):
> 
> * a little replacement Cloud (hosted by us) -> Cloud (hosted)
> * (maybe) switch places: Local Environment <-> Cloud (hosted)
> 
> Happy to discuss as you see fit, just looking from a few different perspectives here I guess.

Thanks, @enyst ! I agree with the suggestions but I ended up removing that question and those answers all together - they were outdated.

frontend/src/routes/onboarding-form.tsx
HeyItsChloe: That was intentional - all questions were using the same string. But this is now outdated as none of the questions have subtitles.

frontend/src/routes/onboarding-form.tsx
HeyItsChloe: Very true! Updated!

enterprise/server/routes/auth.py
HeyItsChloe: Updated!

enterprise/server/routes/auth.py
HeyItsChloe: Thanks for explaining this. Updated!

hieptl:
Hello @HeyItsChloe,

Thank you for the recent changes. I did some local testing and noticed a few issues I’d like to share.

**Issue 1:**
The page is still accessible directly in the OSS environment by navigating to /onboarding via the browser address bar. It may be better to gate this feature behind a feature flag so it is not accessible until it is officially enabled. Please refer to the video below for additional context.

https://github.com/user-attachments/assets/6e04d2eb-3502-487b-8698-94cb97a1ad37

**Issue 2:**
When hovering over each option, the cursor does not change to pointer. It would improve the user experience if the pointer cursor were displayed. Please see the video below for more details.

https://github.com/user-attachments/assets/3773480d-a2d1-46d1-83a8-b400021a3bf3

**Issue 3:**
The back button is currently positioned to the right of the Next button. Should it instead be placed to the left of the Next button for better alignment with common UI patterns? Please refer to the image below for clarification.

issue-3

Issue 4:
There is a visible scrollbar on this page that appears to be redundant. It may be possible to remove it to improve the overall layout. Please see the image below for more information.

issue-4

Thank you very much! 🙏

jpelletier1:
@HeyItsChloe For step 2 - how big is your engineering organization - can we change the following:
- Change: 200+ --> 200 - 1000 people
- Add new option: 1000+ people

frontend/src/components/features/onboarding/step-header.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+function StepHeader({ title, currentStep, totalSteps }: StepHeaderProps) {
+  return (
+    <div data-testid="step-header" className="flex flex-col items-center gap-2">
+      <div className="flex justify-center gap-2 mb-2">
+        {Array.from({ length: totalSteps }).map((_, index) => (
+          <div
+            key={index}
+            className={`w-[6px] h-[4px] rounded-full transition-colors ${
```
hieptl: **suggestion:** We should use cn to combine classes instead of a template string. Thank you! 🙏

enterprise/server/routes/auth.py
```diff
@@ -384,6 +386,13 @@ async def keycloak_callback(
             f'{request.base_url}accept-tos?redirect_url={encoded_redirect_url}'
         )
         response = RedirectResponse(tos_redirect_url, status_code=302)
+    # if new user must complete onboarding, redirect to the onboarding-form (only if enabled)
+    elif ENABLE_ONBOARDING and await OrgService.needs_onboarding(user):
```
hieptl: **question:**

### 1. Review of TOS and Onboarding Flow

```python
has_accepted_tos = user.accepted_tos is not None
if not has_accepted_tos:
    # Redirect to TOS page
    response = RedirectResponse(tos_redirect_url, status_code=302)
elif ENABLE_ONBOARDING and await OrgService.needs_onboarding(user):
    # Redirect to onboarding
    response = RedirectResponse(onboarding_url, status_code=302)
else:
    response = RedirectResponse(redirect_url, status_code=302)
```

1. Timing Issue in the TOS Flow

For a brand-new user, the current flow behaves as follows:
- On first login, has_accepted_tos is False, so the user is redirected to /accept-tos.
- After accepting the TOS, the accept_tos endpoint returns a JSON response containing a redirect_url.
- The front end then redirects the user directly to the home page, without going back through the Keycloak callback.
- As a result, the elif branch that checks needs_onboarding() is never evaluated, because the user does not re-enter the callback flow after accepting the TOS.

### 2. Personal Organization Creation Timing

Within keycloak_callback, the flow is currently:

```python
user = await UserStore.get_user_by_id_async(user_id)
if not user:
    user = await UserStore.create_user(user_id, user_info)  # Personal org + OrgMember created here
...
has_accepted_tos = user.accepted_tos is not None
if not has_accepted_tos:
    # redirect to TOS
elif ENABLE_ONBOARDING and await OrgService.needs_onboarding(user):
    # This will always be False
```

The key point here is that the personal organization is created during user creation, before the TOS check is evaluated.

When needs_onboarding() runs, it checks:
1. Whether the user has a personal organization (user_{user_id}_org) → Yes, it was just created.
2. Whether the user has any organization membership → Yes, the membership was created with status='active'.

As a result, needs_onboarding() will always return False for users created through the standard authentication flow.

Even if this logic were triggered again later, it would still return False, as shown below:

```python
result = await session.execute(
    select(Org).where(Org.name == personal_org_name)
)
personal_org = result.scalar_one_or_none()
if personal_org:
    return False
```

At the moment, the onboarding redirect logic is placed in a code path that new users will never reach after accepting the TOS. This represents a fundamental flow issue and effectively prevents onboarding from ever being shown to new users under the current implementation.

Please feel free to let me know if my understanding is incorrect.

Thank you! 🙏

HeyItsChloe: Updated! - using a cookie `is_new_user` to manage redirect logic instead of relying on `needs_onboarding()`. This seemed to be the simplest approach, but lmk if there is another approach that is cleaner / more future proof.

hieptl: Hello @HeyItsChloe,

Based on my first comment in this thread, it appears that the “Review of TOS and Onboarding Flow” portion has not yet been addressed.

I ran SaaS locally and created a new account. After accepting the TOS, the onboarding page did not appear as expected.

https://github.com/user-attachments/assets/49050af1-d92b-49ec-9ce7-a448dd0fd9d6

Thank you very much! 🙏

enterprise/server/routes/auth.py
```diff
@@ -400,6 +414,18 @@ async def keycloak_callback(
         accepted_tos=has_accepted_tos,
     )
 
+    # Set onboarding cookie for new users
+    if is_new_user:
+        response.set_cookie(
+            key='is_new_user',
+            value='true',
+            domain=get_cookie_domain(request),
+            httponly=False,
```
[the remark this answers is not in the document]
HeyItsChloe: no longer using cookie

frontend/src/hooks/mutation/use-submit-onboarding.ts
```diff
@@ -0,0 +1,27 @@
...
+type SubmitOnboardingArgs = {
+  selections: Record<string, string>;
+};
+
+export const useSubmitOnboarding = () => {
+  const navigate = useNavigate();
+
+  return useMutation({
+    mutationFn: async ({ selections }: SubmitOnboardingArgs) =>
+      // TODO: persist user responses
```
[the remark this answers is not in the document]
HeyItsChloe: handled on a different ticket

enterprise/server/routes/auth.py
HeyItsChloe: @hieptl  Issue now resolved. I know this ticket is supposed to be “FE only” but completing this ticket without touching the DB is causing issues. I updated the PR to store `completed_onboarding` on the user table to track whether or not users have completed onboarding. LMK if there is a dynamic way to track this without DB migrations (I tried using a cookie but it is just not very performant). 

Also, I will create a different ticket to handle storing user responses. For now there is just a placeholder - `useSubmitOnboarding`.

https://github.com/user-attachments/assets/a3291af3-aec3-4eb6-b623-f7bf1bfd1665

jpelletier1:
@HeyItsChloe is this waiting for PR review?

HeyItsChloe:
> @HeyItsChloe is this waiting for PR review?

Yup! Ready for review

hieptl:
Hello @HeyItsChloe,

I would suggest removing the back-end code for now, as this ticket is focused on the front end only and the page can still be accessed through the front-end feature flag. With this approach, the acceptance criteria would be met.

We can address the back-end support for this page in separate tickets and pull requests at a later stage.

Please feel free to let me know if you have any questions or concerns.

Thank you very much! 🙏

hieptl:
Hello @HeyItsChloe,

It appears that this pull request duplicates work that has already been merged in [this pull request](https://github.com/OpenHands/OpenHands/pull/13029)￼.

To keep things streamlined, I’ll go ahead and close this pull request.

Thank you very much! 🙏

PR #12635, opened 2026-01-28
Fix UserSettings creation from Org tables

chuckbutkus:
## Summary of PR

This PR fixes issues with UserSettings creation and migration from Org tables, and converts several synchronous operations to async for improved performance and to prevent blocking.

### Key Changes

1. **Fix UserSettings creation from Org tables** - Initial fix for creating user settings from organization tables during user migration

2. **Add encrypt_legacy_model() function** - Created encryption utilities (`encrypt_legacy_model()`, `encrypt_legacy_kwargs()`, `encrypt_legacy_value()`) as the inverse of the existing decrypt functions, used to encrypt sensitive UserSettings fields (llm_api_key, llm_api_key_for_byor, search_api_key, sandbox_api_key) during downgrade operations

3. **Convert sync calls to async in API key store and routes**:
   - Made `create_api_key()`, `list_api_keys()`, and `retrieve_mcp_api_key()` async methods
   - Updated all callers in `api_keys.py`, `oauth_device.py`, `saas_user_auth.py`, `mcp_config.py`, and `saas_nested_conversation_manager.py`
   - Added internal sync helper methods (`_store_api_key`, `_list_api_keys_from_db`, `_retrieve_mcp_api_key_from_db`) for database operations

4. **Convert get_byor_key_from_db and store_byor_key_in_db to async** - Replaced `call_sync_from_async` wrappers with direct async calls using `UserStore.get_user_by_id_async()`

5. **Add _release_user_creation_lock() method** - Explicitly releases the Redis distributed lock for user creation in a finally block, rather than waiting for the 30-second expiration

6. **Update SaasConversationStore constructor** - Now requires `org_id` as a second positional argument, moving the UserStore lookup to `get_instance()` as an async operation

7. **Add encryption check for legacy values** - Added `_is_legacy_value_encrypted()` helper to check if a value is already encrypted before encrypting

8. **Update tests for async methods** - Updated test files to use `AsyncMock` for async methods and added `pytest.mark.asyncio` decorators

### Files Changed

- `enterprise/server/auth/saas_user_auth.py` - Async calls for MCP API key operations
- `enterprise/server/mcp/mcp_config.py` - Made `create_default_mcp_server_config` async
- `enterprise/server/routes/api_keys.py` - Async API key operations, simplified BYOR key functions
- `enterprise/server/routes/oauth_device.py` - Async `create_api_key` call
- `enterprise/server/saas_nested_conversation_manager.py` - Async `_get_mcp_config` method
- `enterprise/storage/api_key_store.py` - Async methods with sync database helpers
- `enterprise/storage/encrypt_utils.py` - Added encryption functions
- `enterprise/storage/saas_conversation_store.py` - Updated constructor signature
- `enterprise/storage/user_store.py` - Added lock release and encryption check
- `openhands/core/config/mcp_config.py` - Made `create_default_mcp_server_config` async
- `openhands/core/main.py` - Async call for MCP config
- `openhands/server/session/session.py` - Async call for MCP config
- Test files updated for async compatibility

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:bce317b-nikolaik   --name openhands-app-bce317b   docker.openhands.dev/openhands/openhands:bce317b
```

### DIFF
--- a/enterprise/storage/user_store.py
+++ b/enterprise/storage/user_store.py
@@ -19,6 +19,7 @@ from sqlalchemy.orm import joinedload
 from storage.database import a_session_maker, session_maker
 from storage.encrypt_utils import (
     decrypt_legacy_model,
+    decrypt_legacy_value,
     encrypt_legacy_value,
 )
 from storage.org import Org
@@ -588,7 +589,7 @@ class UserStore:
             ]
             for key in encrypt_keys:
                 value = getattr(user_settings, key, None)
-                if value is not None:
+                if value is not None and not _is_legacy_value_encrypted(value):
                     setattr(user_settings, key, encrypt_legacy_value(value))
 
             session.merge(user_settings)
@@ -934,3 +935,12 @@ class UserStore:
                 return False  # Matches old default
 
         return True  # Custom model
+
+
+def _is_legacy_value_encrypted(value: str) -> bool:
+    """Check if a legacy value is encrypted by trying to decrypt it"""
+    try:
+        decrypt_legacy_value(value)
+        return True
+    except Exception:
+        return False


enterprise/storage/api_key_store.py
```diff
@@ -40,8 +41,23 @@ def create_api_key(
             The generated API key
         """
         api_key = self.generate_api_key()
-        user = UserStore.get_user_by_id(user_id)
+        user = await UserStore.get_user_by_id_async(user_id)
```
hieptl, line 44: **suggestion:** When UserStore.get_user_by_id_async(user_id) returns None, attempting to access user.current_org_id will result in an AttributeError.

```python
user = await UserStore.get_user_by_id_async(user_id)
org_id = user.current_org_id  # This will raise an error if user is None
```

The same potential issue is also present in the following locations:
- api_key_store.py:131–132 (list_api_keys)
- api_key_store.py:157–158 (retrieve_mcp_api_key)

Thank you! 🙏

enterprise/storage/user_store.py
```diff
@@ -565,8 +576,21 @@ async def downgrade_user(user_id: str) -> UserSettings | None:
...
+            encrypt_keys = [
+                'llm_api_key',
+                'llm_api_key_for_byor',
+                'search_api_key',
+                'sandbox_api_key',
+            ]
+            for key in encrypt_keys:
+                value = getattr(user_settings, key, None)
+                if value is not None:
+                    setattr(user_settings, key, encrypt_legacy_value(value))
```
hieptl, line 593: **question:** For migrated users, the user_settings retrieved from the database already contain encrypted values. However, the current logic attempts to encrypt these values again:

```python
for key in encrypt_keys:
    value = getattr(user_settings, key, None)
    if value is not None:
        setattr(user_settings, key, encrypt_legacy_value(value))
```

This may lead to the values being encrypted twice.

In contrast, for new signups, the values are first decrypted using .get_secret_value() in _create_user_settings_from_entities(), so applying encryption at this stage is appropriate and works as intended.

For migrated users, since the values retrieved from the database are already encrypted, applying encryption again would result in double-encrypted data.

Could you please confirm if my understanding is correct? Thank you! 🙏

tofarr: Confirmed:
image

enterprise/storage/user_store.py
```diff
@@ -127,6 +130,25 @@ async def _acquire_user_creation_lock(user_id: str) -> bool:
...
+            logger.warning(
+                'user_store:_release_user_creation_lock:no_redis_client',
+                extra={'user_id': user_id},
+            )
+            return True  # Nothing to release if Redis is unavailable
+
+        user_key = f'{_REDIS_USER_CREATION_KEY_PREFIX}{user_id}'
+        deleted = await redis_client.delete(user_key)
+        return bool(deleted)
+
```
tofarr, line 152: Is this required? (Given that the entry automatically disappears in `_REDIS_CREATE_TIMEOUT_SECONDS` and that the migration should only be done once?)

tofarr: I guess it means that if a migration fails we can retry faster.

PR #9496, opened 2025-07-01
feat(frontend): Organizational support

amanape:
- [ ] This change is worth documenting at https://docs.all-hands.dev/
- [ ] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

---
**Summarize what the PR does, explaining any non-trivial design decisions.**
Adds support for organizational management  (e.g., teams, grouped billing, roles, etc...)

This PR will be a frontend-only change for now

---
**Link of any specific issues this addresses:**

PR #12640, opened 2026-01-28
chore: address remaining issues in 12352

amanape:
## Summary of PR

- Addresses remaining feedback in #12352
- Create `frontend/src/utils/org/permission-guard.ts` for easier route visibility control
- Some refactors

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:2358232-nikolaik   --name openhands-app-2358232   docker.openhands.dev/openhands/openhands:2358232
```

### DIFF
--- a/frontend/__tests__/routes/manage-organization-members.test.tsx
+++ b/frontend/__tests__/routes/manage-organization-members.test.tsx
@@ -13,9 +13,19 @@ import {
   ORGS_AND_MEMBERS,
   resetOrgMockData,
   resetOrgsAndMembersMockData,
+  MOCK_TEAM_ORG_ACME,
 } from "#/mocks/org-handlers";
 import OptionService from "#/api/option-service/option-service.api";
 import { useSelectedOrganizationStore } from "#/stores/selected-organization-store";
+
+const mockQueryClient = vi.hoisted(() => {
+  const { QueryClient } = require("@tanstack/react-query");
+  return new QueryClient();
+});
+
+vi.mock("#/query-client-config", () => ({
+  queryClient: mockQueryClient,
+}));
 
 vi.mock("react-i18next", async () => {
   const actual =
@@ -72,8 +82,10 @@ describe("Manage Organization Members Route", () => {
   const getMeSpy = vi.spyOn(organizationService, "getMe");
 
   beforeEach(() => {
-    // Reset Zustand store to ensure clean state and avoid loader redirects
-    useSelectedOrganizationStore.setState({ organizationId: null });
+    // Set Zustand store to a team org so clientLoader allows access to /settings/org-members
+    useSelectedOrganizationStore.setState({ organizationId: MOCK_TEAM_ORG_ACME.id });
+    // Seed organizations into the module-level queryClient used by clientLoader
+    mockQueryClient.setQueryData(["organizations"], [MOCK_TEAM_ORG_ACME]);
 
     const getConfigSpy = vi.spyOn(OptionService, "getConfig");
     // @ts-expect-error - partial mock for testing
@@ -113,8 +125,9 @@ describe("Manage Organization Members Route", () => {
     resetOrgsAndMembersMockData();
     // Clear queryClient cache to ensure fresh data for next test
     queryClient.clear();
-    // Reset Zustand store to avoid polluting other test files
+    // Reset Zustand store and module-level queryClient
     useSelectedOrganizationStore.setState({ organizationId: null });
+    mockQueryClient.clear();
   });
 
   const renderManageOrganizationMembers = () =>
--- a/frontend/src/mocks/org-handlers.ts
+++ b/frontend/src/mocks/org-handlers.ts
@@ -445,6 +445,20 @@ export const ORG_HANDLERS = [
     return HttpResponse.json({ message: "Member removed" }, { status: 200 });
   }),
 
+  http.post("/api/organizations/:orgId/switch", ({ params }) => {
+    const orgId = params.orgId?.toString();
+
+    if (orgId) {
+      const org = orgs.get(orgId);
+      if (org) return HttpResponse.json(org);
+    }
+
+    return HttpResponse.json(
+      { error: "Organization not found" },
+      { status: 404 },
+    );
+  }),
+
   http.post(
     "/api/organizations/:orgId/members/invite",
     async ({ request, params }) => {


frontend/__tests__/hooks/use-permission.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+        expect(run("owner", "u2", "admin")).toBe(
+          rolePermissions.owner.includes("change_user_role:admin"),
+        );
+
+        expect(run("owner", "u2", "member")).toBe(
+          rolePermissions.owner.includes("change_user_role:member"),
+        );
+
+        expect(run("owner", "u2", "owner")).toBe(
+          rolePermissions.owner.includes("change_user_role:member"),
```
hieptl: **suggestion:** I think this should be change_user_role:owner, not change_user_role:member. The intention of this test is to verify that an owner can change a user to the owner role, but the current assertion is checking for the wrong permission. Thank you! 🙏

frontend/__tests__/routes/settings-with-payment.test.tsx
hieptl: **suggestion:** This is currently just a query and does not act as an assertion. Based on the intent described in the test name, it would be more appropriate to explicitly assert that the Billing tab is not present, for example:

```ts
expect(within(navbar).queryByText("Billing")).not.toBeInTheDocument();
```

Thank you! 🙏

frontend/src/routes/settings.tsx
```diff
@@ -21,6 +23,8 @@ const SAAS_ONLY_PATHS = [
 export const clientLoader = async ({ request }: Route.ClientLoaderArgs) => {
   const url = new URL(request.url);
   const { pathname } = url;
+  const user = await getActiveOrganizationUser();
+  const userRole = user?.role || "member";
```
hieptl: **suggestion:** When no organization is selected, user becomes undefined, which then defaults to the member role. This could unintentionally grant member-level permissions to unauthenticated users.

To make this safer, it might be better to return early with a redirect when user is undefined or when no organization is selected, for example:

```ts
if (!user) {
  return redirect(redirectPath);
}
```

This would help ensure that permission handling remains explicit and prevent any unintended access.

Thank you! 🙏

frontend/src/routes/billing.tsx
```diff
@@ -8,6 +8,32 @@ import {
...
+  let config = queryClient.getQueryData<GetConfigResponse>(["config"]);
+  if (!config) {
+    config = await OptionService.getConfig();
+    queryClient.setQueryData<GetConfigResponse>(["config"], config);
+  }
+
+  // Redirect if billing is hidden or user lacks view_billing permission
+  if (
+    config?.FEATURE_FLAGS?.HIDE_BILLING ||
+    !rolePermissions[userRole].includes("view_billing")
```
hieptl: **suggestion:** Both billing.tsx:clientLoader and use-settings-nav-items.ts currently check HIDE_BILLING || !hasPermission("view_billing"). It might be beneficial to extract this logic into a shared utility so we can keep the behavior consistent and reduce duplication. This would also make future updates easier and help centralize the billing visibility rules in one place. Thank you! 🙏

frontend/src/utils/org/permissions.ts
```diff
@@ -1,41 +1,90 @@
...
+];
+
+const adminPerms: PermissionKey[] = [
+  // member perms
+  "manage_secrets",
+  "manage_mcp",
+  "manage_integrations",
+  "manage_application_settings",
+  "manage_api_keys",
+  "view_llm_settings",
```
hieptl: **suggestion:**Rather than duplicating all member permissions within the admin and owner permission arrays, we could consider use spread operator for permission arrays, for example:

```ts
const adminPerms: PermissionKey[] = [
  ...memberPerms,
  "edit_llm_settings",
  "view_billing",
  // ...admin-only perms
];
```

Thank you! 🙏

frontend/src/hooks/use-settings-nav-items.ts
```diff
@@ -1,11 +1,29 @@
...
+ * Build Settings navigation items based on:
+ * - app mode (saas / oss)
+ * - feature flags
+ * - active user's role
+ * @returns Settings Nav Items []
+ */
+export function useSettingsNavItems(): SettingsNavItem[] {
   const { data: config } = useConfig();
+  const { data: user } = useMe();
+  const userRole: OrganizationUserRole = user?.role ?? "member";
```
hieptl, line 24: **suggestion:** Some places use user?.role ?? "member" while others use user?.role || "member". Although they are functionally equivalent in this context, it might be better to standardize on ?? for consistency and clarity across the codebase. Thank you! 🙏

frontend/src/routes/llm-settings.tsx
```diff
@@ -778,4 +781,8 @@ function LlmSettingsScreen() {
   );
 }
 
+// Route protection: all roles have view_llm_settings, but this guard ensures
+// consistency with other routes and allows future restrictions if needed
```
hieptl, line 786: **suggestion:** If all roles currently have this permission, this guard effectively becomes a no-op. It may be worth considering whether it can be removed for simplicity, or alternatively, adding a brief comment to clarify that it is intentionally kept for future-proofing or potential role changes. This would help make the intent clearer to anyone reading the code later. Thank you! 🙏

amanape: The comment already highlights that it is there for consistency and future-proofing. Do you have a suggestion on how I might improve it?

HeyItsChloe:
Everything's looking really great!

1. Nav org selector shows different values than the user context menu - no personal org
Screenshot 2026-02-04 at 1 54 04 PM

---
2. Member users are `view_llm-settings`, but the llm Advanced toggle is not disabled
https://github.com/user-attachments/assets/1933e25b-222e-4504-a0d7-6e720fcefa50

---

3. Admin users have `view_billing` & `add_credits` perms. Should the billing page be disabled / view only? If not, should `permissions.ts` include view_billing under `adminPerms`? 


https://github.com/user-attachments/assets/4de684cc-5f45-462f-a96a-13123afd68ed
Screenshot 2026-02-04 at 2 14 05 PM

---
4. Potential merge conflicts / regressions between this PR & [`APP-198/org-selector`](https://github.com/OpenHands/OpenHands/pull/11902). Just calling this out to keep an eye on when merging because some bugs fixed on `APP-198/org-selector` still exist here

amanape:
@HeyItsChloe Thank you, I think I’ve resolved all remaining issues and correctly resolved any conflicts. I’ll merge this now, but if you encounter any more issues, comment on #9496

PR #12664, opened 2026-01-29
Refactor LiteLLM key updates to use generic user key API

tofarr:
## Summary of PR

Refactors LiteLLM key management to update all user keys dynamically via API instead of relying on specific stored key fields.

**Key changes:**
- Added `_get_user_keys()` and `_update_user_keys()` methods to fetch and update all keys for a user
- `migrate_entries()` and `downgrade_entries()` now update all user keys via API instead of specific fields
- Added key regeneration during migration when DB key does not exist in LiteLLM  
- `saas_settings_store` now checks for existing keys before generating new ones
- `user_store.downgrade_user()` updates user_settings from org_member keys before downgrade

## Demo Screenshots/Videos

Not including videos here as these would include sensitive API keys!

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

N/A

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:8bf18cc-nikolaik   --name openhands-app-8bf18cc   docker.openhands.dev/openhands/openhands:8bf18cc
```

### DIFF
--- a/enterprise/storage/lite_llm_manager.py
+++ b/enterprise/storage/lite_llm_manager.py
@@ -180,9 +180,9 @@ class LiteLlmManager:
                 # If not, generate a new key to prevent verification failures later
                 db_key = None
                 if (
-                    user_settings and
-                    user_settings.llm_api_key and
-                    user_settings.llm_base_url == LITE_LLM_API_URL
+                    user_settings
+                    and user_settings.llm_api_key
+                    and user_settings.llm_base_url == LITE_LLM_API_URL
                 ):
                     db_key = user_settings.llm_api_key
                     if hasattr(db_key, 'get_secret_value'):


enterprise/storage/saas_settings_store.py
```diff
@@ -230,10 +230,19 @@ def _is_openhands_provider(self, item: Settings) -> bool:
...
         """
-        # Generate new key if none exists
+        # Check if user already has keys in LiteLLM
+        existing_keys = await LiteLlmManager.get_user_keys(self.user_id)
+        if existing_keys:
+            logger.info(
+                'saas_settings_store:store:user_already_has_keys',
+                extra={'user_id': self.user_id, 'key_count': len(existing_keys)},
+            )
+            return
```
hieptl, line 243: **question:**

There appears to be a missing key assignment when a user already has existing keys.

```python
existing_keys = await LiteLlmManager.get_user_keys(self.user_id)
if existing_keys:
    logger.info(...)
    return  # returns without setting item.llm_api_key
```

When existing keys are found, the method returns early without assigning a value to item.llm_api_key. This may be unexpected for the caller, as they might assume the settings object always contains a valid API key after this method is executed.

To make the behavior clearer and more reliable, we could consider one of the following approaches:
- Assign the key before returning, for example:

```python
item.llm_api_key = SecretStr(existing_keys[0])
```

We are applying a similar approach at line 254

```python
item.llm_api_key = SecretStr(generated_key)
```

Or, clearly document that in this case the caller is responsible for handling the existing key scenario.

Either option would help improve clarity and prevent potential confusion or unintended behavior.

Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -166,29 +166,56 @@ async def migrate_entries(
...
+                            None,
+                        )
+                        if new_key:
+                            logger.info(
+                                'LiteLlmManager:migrate_lite_llm_entries:generated_new_key',
+                                extra={'org_id': org_id, 'user_id': keycloak_user_id},
+                            )
+                            # Update user_settings with the new key so it gets stored in org_member
+                            user_settings.llm_api_key = SecretStr(new_key)
+                            user_settings.llm_api_key_for_byor = SecretStr(new_key)
```
hieptl, line 222: **question:** When regenerating a key, both llm_api_key and llm_api_key_for_byor are currently being set to the same value. I wanted to check whether this is intentional, as it may cause issues if these two keys are meant to serve different purposes (for example, one for OpenHands usage and one for BYOR).

Would it make more sense to update only llm_api_key in this case? Thank you! 🙏

enterprise/storage/user_store.py
```diff
@@ -420,40 +440,34 @@ async def downgrade_user(user_id: str) -> UserSettings | None:
 
             # For new sign-ups after migration, user_settings won't exist
             # Fall back to getting data from org_members
-            if not user_settings:
+            if user_settings:
+                if org_member.llm_api_key and org_member.llm_api_key.get_secret_value():
```
hieptl, line 444: **suggestion:**

I noticed that the current check uses both org_member.llm_api_key and org_member.llm_api_key.get_secret_value(). However, if llm_api_key is not None but get_secret_value() returns an empty string, the condition will still pass, which might not be the intended behavior.

To make this more robust, we could consider tightening the condition slightly, for example:

```python
if org_member.llm_api_key and (value := org_member.llm_api_key.get_secret_value()):
    user_settings.llm_api_key = encrypt_legacy_value(value)
```

Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -643,6 +656,77 @@ async def _update_key(
...
+        logger.debug(
+            'LiteLlmManager:_update_user_keys:updating_keys',
+            extra={
+                'user_id': keycloak_user_id,
+                'key_count': len(keys),
+            },
+        )
+
+        for key in keys:
+            await LiteLlmManager._update_key(client, keycloak_user_id, key, **kwargs)
```
hieptl, line 732: **question:** I noticed that if one key update fails inside the loop, the remaining keys will not be updated. I wanted to check whether this is the intended behavior, or if it would be better to allow the process to continue updating the other keys.

At a minimum, it might be helpful to wrap each individual key update in a try-except block with proper logging.

Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -643,6 +656,77 @@ async def _update_key(
...
+
+        logger.debug(
+            'LiteLlmManager:_update_user_keys:updating_keys',
+            extra={
+                'user_id': keycloak_user_id,
+                'key_count': len(keys),
+            },
+        )
+
+        for key in keys:
```
hieptl, line 731: **suggestions:**

In _update_user_keys, the keys are currently updated sequentially. For better performance, especially when handling a larger number of keys, we might consider updating them concurrently using asyncio.gather().

```python
await asyncio.gather(*[
    LiteLlmManager._update_key(client, keycloak_user_id, key, **kwargs)
    for key in keys
])
```

Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -166,29 +166,56 @@ async def migrate_entries(
...
-                    logger.debug(
-                        'LiteLlmManager:migrate_lite_llm_entries:update_byor_key',
-                        extra={'org_id': org_id, 'user_id': keycloak_user_id},
-                    )
-                    await LiteLlmManager._update_key(
-                        client,
-                        keycloak_user_id,
-                        user_settings.llm_api_key_for_byor,
-                        team_id=org_id,
+                # Check if the database key exists in LiteLLM
```
chuckbutkus, line 179: The only issue is if the user has a custom key that they are using, then it won't be in LiteLLM and we wouldn't want to generate one.  I just don't think there is a way for us to know if this is the case from existing data.

enterprise/storage/saas_settings_store.py
tofarr: You are right that this is not particularly clear - the existing keys is a fallback - and an issue we face here is that the keys are only available when they are generated. So if the keys already exist, we assume they are the same ones in the user / org_member record.

enterprise/storage/user_store.py
tofarr: I'm not sure what you mean here - `rg_member.llm_api_key and org_member.llm_api_key.get_secret_value()` will return `False` for blank strings?

enterprise/storage/lite_llm_manager.py
tofarr: I think we want the migration to fail if the key migration fails - since it will mean that the key no longer has a limit

enterprise/storage/lite_llm_manager.py
tofarr: I would be worried about overloading litellm with concurrent requests if the number of keys is large

chuckbutkus: I believe there is a batch update available in the LiteLLM API.

enterprise/storage/lite_llm_manager.py
tofarr: I'm not sure if this is problematic, but you can actually do this manually in the web ui too:
* Generate a new API key
* Use this API key to access litellm.

In the long run, I think we need a further security review here.

enterprise/storage/user_store.py
hieptl: I apologize for the confusion on my part. The issue was due to my oversight, and your code is already correct. I will proceed to close this thread. Thank you! 🙏

enterprise/storage/lite_llm_manager.py
tofarr: I've added a check so that the key is only updated if we are using the base_url

PR #12667, opened 2026-01-29
fix: ensure LiteLLM user exists before generating API keys

saurya:
## Summary

This PR fixes a critical security issue where API keys could be generated for users that don't exist in LiteLLM's `UserTable`, leading to:

- Users appearing in `DailyUserSpend` table without corresponding `UserTable` entries
- Budget limits being bypassed (no `user_obj` = no budget check)
- Untracked/unlimited spending

## Root Cause

The issue occurs because:
1. `_create_user()` could silently return when LiteLLM reports "user already exists" without verifying the user actually exists
2. `create_entries()` continued to generate keys even if user creation failed
3. There was no verification that users exist before generating keys in any code path

## Changes

1. **Added `_user_exists()` method** - New method to verify a user exists in LiteLLM by calling `/user/info`

2. **Updated `_create_user()` to return `bool`** - Now returns `True` if user was created or verified to exist, `False` on failure. When LiteLLM returns "already exists" error, we now verify the user actually exists before returning success.

3. **Updated `create_entries()`** to:
   - Check `_create_user()` return value before proceeding
   - Verify user exists before generating key (even when `create_user=False`)
   - Return `None` if user doesn't exist, preventing orphaned keys

4. **Added `user_exists` public wrapper** - Exposed `_user_exists` as a public method for other code to use

5. **Comprehensive tests** - Added/updated tests for:
   - `_user_exists()` behavior
   - `_create_user()` returning bool and verifying user exists
   - Edge cases like "already exists" but user not found

## Testing

All new tests validate the fix:
- `test_user_exists_returns_true`
- `test_user_exists_returns_false_when_not_found`
- `test_user_exists_returns_false_on_mismatched_user_id`
- `test_create_user_already_exists_but_not_found_returns_false`
- `test_create_user_failure_returns_false`
- `test_create_user_already_exists_and_verified`

## Impact

After this fix:
- Keys will only be generated for users that exist in LiteLLM's `UserTable`
- Budget enforcement will work correctly for all users
- Clear error logging when user creation fails or user doesn't exist
- Existing code paths that rely on user verification will be protected

Fixes: https://linear.app/all-hands-ai/issue/PLTF-31

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:8b23a69-nikolaik   --name openhands-app-8b23a69   docker.openhands.dev/openhands/openhands:8b23a69
```

### DIFF
--- a/enterprise/tests/unit/test_lite_llm_manager.py
+++ b/enterprise/tests/unit/test_lite_llm_manager.py
@@ -357,6 +357,9 @@ class TestLiteLlmManager:
         mock_404_response = MagicMock()
         mock_404_response.status_code = 404
         mock_404_response.is_success = False
+        mock_404_response.raise_for_status.side_effect = httpx.HTTPStatusError(
+            message='Not Found', request=MagicMock(), response=mock_404_response
+        )
 
         mock_token_manager = MagicMock()
         mock_token_manager.return_value.get_user_info_from_user_id = AsyncMock(


enterprise/storage/lite_llm_manager.py
```diff
@@ -164,9 +164,33 @@ async def create_entries(
...
+                            'create_entries_failed_user_creation',
+                            extra={
+                                'org_id': org_id,
+                                'user_id': keycloak_user_id,
+                            },
+                        )
+                        return None
+
+                # Verify user exists before proceeding with key generation
+                user_exists = await LiteLlmManager._user_exists(
```
tofarr, line 181: Defensive coding makes sense, but I'm curious why a user which was just created may not exist? Could there be a situation with eventual consistency where this fails for a user which was just created?

saurya: Trying to cover two scenarios here:
(1) We try to create a user and that fails - that should actually be caught by line 170 up there
(2) We for some reason purged a user from the LiteLLM database outside of the flow of this code - we should make sure not to let a key get created for them.

Issue #12690, opened 2026-01-30
[Bug]: Selected repo disappears in frontend


neubig:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

When selecting a repo in the GUI (OpenHands Cloud), the repo disappears:

https://github.com/user-attachments/assets/d9a9702a-f8b0-4048-a530-7ec6aaac63b3

### OpenHands Installation

Docker command in README

### Operating System

None

fury0928:
Hi @neubig may I work on this issue?

nehaaprasad:
Hello! @neubig 
please check out the PR!

PR #12702, opened 2026-01-31
fix selected repo disappearing in pen repository dropdown

nehaaprasad:
### summary of PR

- fixes the bug where choosing a repo from "Most Recent" made it disappear from the input. 
- The dropdown now keeps the selection visible by falling back to the locally selected item when it’s not in the fetched results.

- Demo Screenshots/Videos

(Optional)

### Change Type
[x] Bug fix

### Checklist
[ ] I have read and reviewed the code and I understand what the code is doing.
[ ] I have tested the code to the best of my ability and ensured it works as expected.

### Fixes

fix: #12690


neubig (commented):
Hi @naaa760 , could you please send a video demonstrating that (1) you reproduced the error before the fix, and (2) that after the fix the problem no longer appears?

tofarr:
Thanks for picking this up! It's been annoying me too!

PR #12718, opened 2026-02-02
Fix runtime status error on conversation resume

tofarr:
## Summary of PR

This PR fixes an issue where resumed conversations would incorrectly show an error status during startup. The problem occurs because the runtime API marks sandboxes as `RUNNING` before they are actually fully started, particularly affecting resumed runtimes.

### Changes in `openhands/server/routes/manage_conversations.py`:

- Added detection logic in `get_conversation()` to handle the case where a sandbox is marked as `RUNNING` but has no execution status
- When this condition is detected, the code checks server responsiveness via the `/server_info` endpoint
- If the server is unresponsive or within a 60-second grace period (`_RESUME_GRACE_PERIOD`), the sandbox status is set to `STARTING` instead of potentially showing an error
- Added proper warning logging with conversation and sandbox IDs for debugging

This is a temporary workaround for a bug in the runtime API that marks servers as RUNNING before they are actually started.

## Demo Screenshots/Videos

**When we resume a conversation...**
image

** We have an initial starting state **
image

**Before changes - We get an error state during startup**
image

**After Changes - The status displays as `STARTING` until the value is actually running**
image

Finally it is ready
image

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

Fixed an issue where resumed conversations would incorrectly display an error status during startup. The system now properly detects when a sandbox is still initializing and shows a "starting" status instead of an error.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:33cb95c-nikolaik   --name openhands-app-33cb95c   docker.openhands.dev/openhands/openhands:33cb95c
```

### DIFF
--- a/frontend/__tests__/components/features/home/git-repo-dropdown.test.tsx
+++ b/frontend/__tests__/components/features/home/git-repo-dropdown.test.tsx
@@ -201,7 +201,7 @@ describe("GitRepoDropdown", () => {
       const selectedRepository = MOCK_REPOSITORIES[0];
 
       renderDropdown(
-        { value: selectedRepository.full_name },
+        { value: selectedRepository.id },
         { selectedRepository },
       );
 
@@ -229,6 +229,25 @@ describe("GitRepoDropdown", () => {
       await userEvent.click(screen.getByText("user/repo-two"));
 
       expect(mockOnChange).toHaveBeenCalledWith(MOCK_REPOSITORIES[1]);
+    });
+
+    it("should keep selected repo visible even if it's not in fetched results", async () => {
+      // selectedRepository from useRepositoryData stays null in our default mocks,
+      // which matches the real-world scenario where a recent repo isn't yet loaded.
+      renderDropdown();
+
+      const input = screen.getByTestId("git-repo-dropdown") as HTMLInputElement;
+      await userEvent.click(input);
+
+      await waitFor(() => {
+        expect(screen.getByText("user/repo-two")).toBeInTheDocument();
+      });
+
+      await userEvent.click(screen.getByText("user/repo-two"));
+
+      await waitFor(() => {
+        expect(input.value).toBe("user/repo-two");
+      });
     });
   });
 });
--- a/frontend/src/components/features/home/git-repo-dropdown/git-repo-dropdown.tsx
+++ b/frontend/src/components/features/home/git-repo-dropdown/git-repo-dropdown.tsx
@@ -91,6 +91,13 @@ export function GitRepoDropdown({
     repositoryName,
   );
 
+  /**
+   * `selectedRepository` is derived from fetched/search results. A repo selected
+   * from "Most Recent" may not be present in those results yet, so use the
+   * locally selected item as a fallback to keep the UI stable.
+   */
+  const effectiveSelectedRepository = selectedRepository ?? localSelectedItem;
+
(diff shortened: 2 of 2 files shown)


openhands/server/routes/manage_conversations.py
```diff
@@ -484,6 +488,44 @@ async def get_conversation(
                 conversation_uuid
             )
             if app_conversation:
+                if (
```
malhotra5, line 491: Is this API endpoint polled repeatedly? Where does the status get set to `ERROR` if the resume grace period is exceeded?

raymyers:
Can this behavior change be reflected in either `/conversations` route tests in `test_conversation_routes.py` or a direct test around `_filter_conversations_by_age` function?

openhands/server/routes/manage_conversations.py
tofarr: Yes, this endpoint is polled repeatedly by the frontend when viewing a conversation.

Regarding ERROR status: This workaround does not manage the ERROR transition itself. The runtime API/sandbox service independently monitors sandbox health and will mark the sandbox as `ERROR` when it detects failures. This code simply returns `STARTING` instead of showing a premature error during the brief window where the sandbox is marked as `RUNNING` but the server has not yet started responding.

In other words:
1. **Normal case**: Server responds → check uptime → if < 60s, show `STARTING`; otherwise show actual status
2. **Server unresponsive**: Return `STARTING` temporarily; the runtime API will eventually set `ERROR` if the sandbox truly fails

malhotra5: Ah got it! Is there a more specific exception we can catch then instead of a blanket exception?

malhotra5: Also it seems like if the runtime api is down all together (i.e server is not responding), then the status will perpetually be in a starting status? (because the exception is caught but the grace period is not checked?)

tofarr: No - in that case the sandbox_status will be error rather than running - so this code will not be called.

tofarr: The only case where this will return "STARTING" rather than "ERROR" is if the sandbox has been running for less than 60 seconds, and I go into the agent server and delete the conversation - then it will return "STARTING" until the grace period elapses.

tests/unit/server/routes/test_conversation_routes.py
```diff
@@ -1459,3 +1465,211 @@ async def get_agent_loop_info(*args, **kwargs):
                 assert call_kwargs.get('include_sub_conversations') is True
                 assert call_kwargs.get('page_id') == 'test_v1_page_id'
                 assert call_kwargs.get('limit') == 50
+
+
+@pytest.mark.asyncio
+async def test_get_conversation_running_sandbox_no_execution_status_server_responds_within_grace_period():
```
malhotra5: NIT: maybe we can parameterize these tests

malhotra5 (approved):
LGTM!!

PR #12719, opened 2026-02-02
fix(backend): resolve missing email and display name for user identity tracking

amanape:
## Summary of PR

- When a GitHub user has no public profile email, the backend now falls back to the `/user/emails` API to retrieve their primary verified email. This ensures PostHog and Reo.dev receive a valid email for most GitHub-authenticated users.
- The fallback `/api/user/info` path (used for SSO/SAML users without a GitHub token) now populates `name` and `company` from Keycloak OIDC claims instead of returning null.
- `create_user()` and `migrate_user()` now store the user's real display name (from Keycloak claims) as `contact_name` instead of their login username. Existing users are gradually backfilled on login.

### Important: Requires verification before shipping

The GitHub `/user/emails` fallback requires the OAuth token to have the `user:email` scope. If Keycloak's GitHub IDP doesn't forward this scope, the call will 403 silently (safe, but the feature is a no-op).

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:8814d25-nikolaik   --name openhands-app-8814d25   docker.openhands.dev/openhands/openhands:8814d25
```

### DIFF
--- a/openhands/integrations/bitbucket/service/base.py
+++ b/openhands/integrations/bitbucket/service/base.py
@@ -23,7 +23,7 @@ class BitBucketMixinBase(BaseGitService, HTTPClient):
     Base mixin for BitBucket service containing common functionality
     """
 
-    BASE_URL = "https://api.bitbucket.org/2.0"
+    BASE_URL = 'https://api.bitbucket.org/2.0'
 
     @staticmethod
     def _resolve_primary_email(emails: list[dict]) -> str | None:
@@ -33,8 +33,8 @@ class BitBucketMixinBase(BaseGitService, HTTPClient):
         'email', 'is_primary', and 'is_confirmed' keys.
         """
         for entry in emails:
-            if entry.get("is_primary") and entry.get("is_confirmed"):
-                return entry.get("email")
+            if entry.get('is_primary') and entry.get('is_confirmed'):
+                return entry.get('email')
         return None
 
     def _extract_owner_and_repo(self, repository: str) -> tuple[str, str]:
@@ -49,9 +49,9 @@ class BitBucketMixinBase(BaseGitService, HTTPClient):
         Raises:
             ValueError: If repository format is invalid
         """
-        parts = repository.split("/")
+        parts = repository.split('/')
         if len(parts) < 2:
-            raise ValueError(f"Invalid repository name: {repository}")
+            raise ValueError(f'Invalid repository name: {repository}')
 
         return parts[-2], parts[-1]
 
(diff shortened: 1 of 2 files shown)


enterprise/storage/user_store.py
```diff
@@ -756,6 +759,30 @@ def list_users() -> list[User]:
...
+        Preserves custom values that were set via the PATCH endpoint.
+        """
+        real_name = resolve_display_name(user_info)
+        if not real_name:
+            return
+
+        preferred_username = user_info.get('preferred_username', '')
+        username = user_info.get('username', '')
+
+        with session_maker() as session:
```
hieptl: **suggestion:** 

The backfill_contact_name() method is currently synchronous (it uses the session_maker() context manager) but is invoked from an async route handler. This can result in blocking database I/O on the async event loop and may lead to performance issues under load.

To address this, we may want to consider making backfill_contact_name() asynchronous by using a_session_maker().

Thank you! 🙏

.gitignore
```diff
@@ -130,6 +130,7 @@ env.bak/
 .env.bak
 venv.bak/
 *venv/
+ralph/
```
hieptl: **suggestion:** It appears that this change is not directly related to the scope of this pull request. Would it make sense to remove it to keep the changes more focused? Thank you. 🙏

enterprise/storage/user_store.py
```diff
@@ -756,6 +759,30 @@ def list_users() -> list[User]:
...
+            org = session.query(Org).filter(Org.id == uuid.UUID(user_id)).first()
+            if not org:
+                return
```
hieptl, line 791: **suggestion:** The backfill_contact_name() method currently returns silently on certain failure cases (for example, when no organization is found or when no real name is available). To make troubleshooting easier, it may be helpful to add debug-level logging in these scenarios. For example:

```python
if not org:
    logger.debug(
        'backfill_contact_name:org_not_found',
        extra={'user_id': user_id},
    )
    return
```

Thank you! 🙏

enterprise/storage/user_store.py
```diff
@@ -756,6 +759,30 @@ def list_users() -> list[User]:
...
+
+            if org.contact_name in (preferred_username, username):
+                org.contact_name = real_name
+                session.commit()
```
hieptl: **suggestion:** For improved observability and monitoring, it may be helpful to log when a backfill operation actually updates a record. For example:

```python
if org.contact_name in (preferred_username, username):
    logger.info(
        'backfill_contact_name:updated',
        extra={
            'user_id': user_id,
            'old': org.contact_name,
            'new': real_name,
        },
    )
    org.contact_name = real_name
    session.commit()
```

Thank you! 🙏

.gitignore
amanape:
```suggestion
```

chuckbutkus:
@amanape, is this only a problem with GitHub (not with GitLab or BitBucket)?

chuckbutkus (approved):

amanape:
GitLab already returns email directly, pushed patch for BitBucket

Issue #12714, opened 2026-02-02
[Bug] Conversation polling fails with ValidationError when secrets contain null values


zxkane:
### Is there an existing issue for the same bug?

- [x] I have checked the existing issues.

### Describe the bug

When the frontend polls `/api/conversations` or loads a stored conversation, Pydantic validation fails if the conversation state contains secrets with `null` values:

```
ValidationError: Input should be a valid string [type=string_type, input_value=None]
```

This occurs in two scenarios:
1. `_compose_conversation_info()` calls `state.model_dump()` which includes `None` values, then re-validates
2. `StoredConversation.model_validate_json()` fails when `stored.json` contains null secrets

### Root Cause

After secrets are masked (due to missing `OH_SECRET_KEY` or runtime recreation), the state contains:
```json
{
  "secrets": {
    "GITHUB_TOKEN": {
      "value": null
    }
  }
}
```

The current code doesn't handle this gracefully.

### Affected Code

**File**: `openhands/agent_server/conversation_service.py`

1. `_compose_conversation_info()`:
```python
# Current - includes None values
**state.model_dump()
```

2. `StoredConversation.model_validate_json()`:
```python
# Fails if json_str contains null secrets
StoredConversation.model_validate_json(json_str, ...)
```

### Steps to Reproduce

1. Create a conversation with secrets
2. Restart the runtime or change `OH_SECRET_KEY`
3. Open the conversation list in the frontend
4. Observe 500 error or ValidationError in logs

### Expected Behavior

API should return conversation data successfully, with invalid secrets filtered out.

### Proposed Solution

**Fix 1: Use `exclude_none=True` in model_dump()**
```python
# In _compose_conversation_info()
**state.model_dump(exclude_none=True)
```

**Fix 2: Preprocess JSON before validation**
```python
def _filter_invalid_secrets_from_json(json_str: str) -> str:
    """Filter out secrets with null values before validation."""
    try:
        data = json.loads(json_str)
    except json.JSONDecodeError:
        return json_str
    
    secrets = data.get("secrets")
    if secrets and isinstance(secrets, dict):
        filtered = {
            k: v for k, v in secrets.items()
            if v is not None and not (isinstance(v, dict) and v.get("value") is None)
        }
        data["secrets"] = filtered
        return json.dumps(data)
    
    return json_str

# Usage
StoredConversation.model_validate_json(
    _filter_invalid_secrets_from_json(json_str),
    ...
)
```

### Related Issues

[...]

PR #12722, opened 2026-02-02
fix: filter null secrets from conversation JSON before validation

nehaaprasad:
### summary of PR

- conversation JSON from the agent server can include masked secrets with value: null, which triggers a pydantic ValidationError.

- This PR adds a helper that removes secrets with null values before validating with ConversationInfo.
- The helper is used when refreshing a conversation (remote sandbox) and when starting a conversation (live status service).
- conversation list and load flows no longer fail when secrets are masked.

### Change Type

- [x] Bug fix

fix: #12714


hieptl (changes requested):
Hello @naaa760,

Thank you for creating this pull request.

Could you please add tests to cover your changes?

Thank you very much! 🙏

neubig:
Since requested changes were left more than 2 weeks ago and there has not been a response, we are planning to close this PR for lack of activity. If you would like to take additional action, please push updates or leave a comment and we would be happy to take another look. Thank you for the contribution.

PR #12725, opened 2026-02-02
fix: BYOR to OpenHands provider switch auth error

neubig:
## Summary

Fixes the bug where switching from BYOR to OpenHands provider causes "Invalid proxy server token" errors.

### After This PR (Fixed)
When switching from BYOR to OpenHands provider:
1. Check for existing keys in LiteLLM
2. **Verify** the key is valid in LiteLLM before reusing
3. **Set** `item.llm_api_key` to the verified key
4. If verification fails (orphaned key), **generate a new key**

✅ User can successfully use OpenHands provider

---

### Before This PR (Buggy - Jan 29+)
When switching from BYOR to OpenHands provider:
1. Check for existing keys in LiteLLM
2. If keys exist, **return early** without setting `item.llm_api_key`
3. Settings saved with **null/empty** `llm_api_key`

```python
if existing_keys:
    logger.info('user_already_has_keys')
    return  # BUG: item.llm_api_key never set!
```

❌ User gets "Invalid proxy server token" error every time

---

### Before Jan 29 Refactor (Jan 22-23)
When switching from BYOR to OpenHands provider:
1. Code would **always generate a new key** (no "reuse existing" optimization)
2. New key created and `item.llm_api_key` set properly

However, some users with **orphaned keys** from the orgs migration would still hit errors if their migrated keys weren't properly synced to LiteLLM's verification table.

✅ Most users worked fine  
⚠️ Some users with orphaned migration keys failed

---

### TL;DR
| Period | Behavior | Result |
|--------|----------|--------|
| Before Jan 22 | Generate new key | ✅ Works |
| Jan 22-29 | Generate new key, but some orphaned | ⚠️ Some failures |
| Jan 29+ | Reuse key but don't set it | ❌ Always fails |
| **This PR** | Verify & set key, or generate new | ✅ Works |

## Tests

Added 2 unit tests:
- `test_ensure_openhands_api_key_sets_key_when_reusing_verified_key`
- `test_ensure_openhands_api_key_generates_new_key_when_verification_fails`

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:ff2fa92-nikolaik   --name openhands-app-ff2fa92   docker.openhands.dev/openhands/openhands:ff2fa92
```

### DIFF
--- a/openhands/app_server/web_client/default_web_client_config_injector.py
+++ b/openhands/app_server/web_client/default_web_client_config_injector.py
@@ -21,6 +21,7 @@ class DefaultWebClientConfigInjector(WebClientConfigInjector):
     recaptcha_site_key: str | None = None
     faulty_models: list[str] = Field(default_factory=list)
     error_message: str | None = None
+    github_app_slug: str | None = None
 
     async def get_web_client_config(self) -> WebClientConfig:
         from openhands.app_server.config import get_global_config
@@ -36,5 +37,6 @@ class DefaultWebClientConfigInjector(WebClientConfigInjector):
             recaptcha_site_key=self.recaptcha_site_key,
             faulty_models=self.faulty_models,
             error_message=self.error_message,
+            github_app_slug=self.github_app_slug,
         )
         return result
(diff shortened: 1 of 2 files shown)


tofarr (approved):
LGTM! 🍰

PR #12726, opened 2026-02-03
APP-443: Fix key generation

chuckbutkus:
APP-443: there were issues switching to/from the OpenHands provider.  This should fix the issue.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:74be128-nikolaik   --name openhands-app-74be128   docker.openhands.dev/openhands/openhands:74be128
```

### DIFF
--- a/enterprise/tests/unit/test_lite_llm_manager.py
+++ b/enterprise/tests/unit/test_lite_llm_manager.py
@@ -1596,9 +1596,7 @@ class TestGetAllKeysForUser:
         mock_client.get.return_value = mock_response
 
         with patch('storage.lite_llm_manager.LITE_LLM_API_KEY', 'test-api-key'):
-            with patch(
-                'storage.lite_llm_manager.LITE_LLM_API_URL', 'http://test.com'
-            ):
+            with patch('storage.lite_llm_manager.LITE_LLM_API_URL', 'http://test.com'):
                 result = await LiteLlmManager._get_all_keys_for_user(
                     mock_client, 'test-user-id'
                 )
@@ -1626,9 +1624,7 @@ class TestGetAllKeysForUser:
         mock_client.get.return_value = mock_response
 
         with patch('storage.lite_llm_manager.LITE_LLM_API_KEY', 'test-api-key'):
-            with patch(
-                'storage.lite_llm_manager.LITE_LLM_API_URL', 'http://test.com'
-            ):
+            with patch('storage.lite_llm_manager.LITE_LLM_API_URL', 'http://test.com'):
                 result = await LiteLlmManager._get_all_keys_for_user(
                     mock_client, 'test-user-id'
                 )
@@ -1641,9 +1637,7 @@ class TestGetAllKeysForUser:
         mock_client.get.side_effect = Exception('API Error')
 
         with patch('storage.lite_llm_manager.LITE_LLM_API_KEY', 'test-api-key'):
-            with patch(
-                'storage.lite_llm_manager.LITE_LLM_API_URL', 'http://test.com'
-            ):
+            with patch('storage.lite_llm_manager.LITE_LLM_API_URL', 'http://test.com'):
                 result = await LiteLlmManager._get_all_keys_for_user(
                     mock_client, 'test-user-id'
                 )
--- a/enterprise/tests/unit/test_saas_settings_store.py
+++ b/enterprise/tests/unit/test_saas_settings_store.py
@@ -184,7 +184,9 @@ async def test_ensure_api_key_keeps_valid_key(mock_config):
     """When the existing key is valid, it should be kept unchanged."""
     store = SaasSettingsStore('test-user-id-123', MagicMock(), mock_config)
     existing_key = 'sk-existing-key'
-    item = DataSettings(llm_model='openhands/gpt-4', llm_api_key=SecretStr(existing_key))
+    item = DataSettings(
+        llm_model='openhands/gpt-4', llm_api_key=SecretStr(existing_key)
+    )
 
     with patch(
         'storage.saas_settings_store.LiteLlmManager.verify_existing_key',


enterprise/storage/saas_settings_store.py
```diff
@@ -227,37 +239,50 @@ def _is_openhands_provider(self, item: Settings) -> bool:
...
+                generated_key = await LiteLlmManager.generate_key(
+                    self.user_id,
+                    org_id,
+                    None,
+                    {'type': 'openhands'},
+                )
+            else:
+                # Must delete any existing key with the same alias first
+                await LiteLlmManager.delete_key_by_alias(
+                    key_alias=f'OpenHands Cloud - user {self.user_id} - org {org_id}',
```
hieptl: **suggestion:** This pull request introduces get_openhands_cloud_key_alias(), but it is not currently used in saas_settings_store.py. It may be worth considering using get_openhands_cloud_key_alias() there for consistency and reuse. Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -1123,6 +1133,99 @@ async def _get_key_info(
...
+            key_alias = key_info.get('key_alias')
+            token = None
+            if (
+                openhands_type
+                and metadata.get('type') == 'openhands'
+                and team_id == org_id
+            ):
+                # Found an existing OpenHands key for this org
+                key_name = key_info.get('key_name')
+                token = key_name[-4:] if key_name else None  # last 4 digits of key
```
hieptl, line 1208: **question:** In _verify_existing_key, could token be None when key_name is None or empty?

```python
token = key_name[-4:] if key_name else None  # token can be None
if key_value.endswith(token):  # TypeError if token is None
```

In this case, calling endswith with a None value would raise a TypeError. Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -1123,6 +1133,99 @@ async def _get_key_info(
...
+        key_value: str,
+        keycloak_user_id: str,
+        org_id: str,
+        openhands_type: bool = False,
+    ) -> bool:
+        """Check if an existing OpenHands key exists for the user/org and return it.
+
+        Looks for keys with metadata type='openhands' and matching team_id.
+
+        Returns the key value if found, None otherwise.
```
hieptl: **suggestion:** The docstring states, “Returns the key value if found, None otherwise,” but the function currently returns a boolean value. To avoid confusion, it would be good to update the docstring so that it accurately reflects the function’s behavior. Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -1123,6 +1133,99 @@ async def _get_key_info(
             'key_spend': key_info.get('spend'),
         }
 
+    @staticmethod
+    async def _get_all_keys_for_user(
```
hieptl, line 1137: **question:** _get_all_keys_for_user does not set the required API key header, unlike the other methods:

```python
response = await client.get(
    f'{LITE_LLM_API_URL}/user/info?user_id={keycloak_user_id}',
)
```

Is this behavior intentional, or should the API key header be included here as well for consistency? Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -25,12 +25,22 @@
 from openhands.utils.http_session import httpx_verify_option
 
 # Timeout in seconds for BYOR key verification requests to LiteLLM
```
hieptl: **suggestion:** The comment `# Timeout in seconds for BYOR key verification requests to LiteLLM` was retained, but the constant has been renamed from `BYOR_KEY_VERIFICATION_TIMEOUT` to `KEY_VERIFICATION_TIMEOUT`. Since the constant is now more general, the comment still referencing “BYOR” may be misleading.

Is this intentional, or should the comment be updated to reflect the new, broader usage? Thank you! 🙏

enterprise/storage/lite_llm_manager.py
```diff
@@ -1123,6 +1133,99 @@ async def _get_key_info(
...
+                'LiteLlmManager:_get_all_keys_for_user:error',
+                extra={
+                    'user_id': keycloak_user_id,
+                    'error': str(e),
+                },
+            )
+            return []
+
+    @staticmethod
+    async def _verify_existing_key(
```
hieptl, line 1178: **suggestion:** It appears that tests for this function are currently missing. It would be helpful to add coverage to ensure its behavior is properly validated. Thank you! 🙏

enterprise/storage/lite_llm_manager.py
hieptl, line 1137: **suggestion:** It appears that tests for this function are currently missing. It would be helpful to add coverage to ensure its behavior is properly validated. Thank you! 🙏

neubig:
Here's a regression test for `_ensure_api_key`:

```python
@pytest.mark.asyncio
async def test_ensure_api_key_generates_new_when_verification_fails(mock_config):
    """When key verification fails, a new key must be generated and set."""
    store = SaasSettingsStore("test-user-id-123", MagicMock(), mock_config)
    new_key = "sk-new-generated-key"
    item = DataSettings(llm_model="openhands/gpt-4", llm_api_key=SecretStr("sk-invalid"))

    with (
        patch(
            "storage.saas_settings_store.LiteLlmManager.verify_existing_key",
            new_callable=AsyncMock,
            return_value=False,
        ),
        patch(
            "storage.saas_settings_store.LiteLlmManager.generate_key",
            new_callable=AsyncMock,
            return_value=new_key,
        ),
    ):
        await store._ensure_api_key(item, "org-123", openhands_type=True)

    assert item.llm_api_key is not None
    assert item.llm_api_key.get_secret_value() == new_key
```

Add these imports:
```python
from unittest.mock import AsyncMock
from openhands.storage.data_models.settings import Settings as DataSettings
```

This test verifies that when `verify_existing_key` returns `False`, a new key is generated and `item.llm_api_key` is properly set.

neubig:
Here's a regression test for the public `store()` API that **fails on both main and this PR**:

```python
@pytest.mark.asyncio
async def test_store_sets_llm_api_key_for_openhands_provider(mock_config):
    """When storing settings with OpenHands provider, llm_api_key must be set."""
    import uuid

    user_id = "550e8400-e29b-41d4-a716-446655440000"
    org_id = uuid.UUID("00000000-0000-0000-0000-000000000001")
    new_key = "sk-new-generated-key"

    mock_org_member = MagicMock()
    mock_org_member.org_id = org_id
    mock_org_member.llm_api_key = "existing-key"
    mock_org_member.llm_base_url = None

    mock_user = MagicMock()
    mock_user.current_org_id = org_id
    mock_user.org_members = [mock_org_member]

    mock_org = MagicMock()
    mock_org.default_llm_base_url = None

    mock_session = MagicMock()
    mock_session.query.return_value.options.return_value.filter.return_value.first.return_value = mock_user
    mock_session.query.return_value.filter.return_value.first.return_value = mock_org

    mock_session_maker = MagicMock()
    mock_session_maker.return_value.__enter__ = MagicMock(return_value=mock_session)
    mock_session_maker.return_value.__exit__ = MagicMock(return_value=False)

    store = SaasSettingsStore(user_id, mock_session_maker, mock_config)
    item = Settings(llm_model="openhands/gpt-4")  # No llm_api_key initially

    with (
        patch("storage.saas_settings_store.LiteLlmManager.get_user_keys",
              new_callable=AsyncMock, return_value=["sk-existing-key"], create=True),
        patch("storage.saas_settings_store.LiteLlmManager.generate_key",
              new_callable=AsyncMock, return_value=new_key),
        patch("storage.saas_settings_store.LiteLlmManager.verify_existing_key",
              new_callable=AsyncMock, return_value=False, create=True),
    ):
        await store.store(item)

    assert item.llm_api_key is not None, "llm_api_key must be set after store()"
```

## Expected vs Actual Behavior

**Scenario:** User switches from BYOR to OpenHands provider. The incoming `item` has `llm_model="openhands/gpt-4"` but `llm_api_key=None`.

**Expected:** `store()` should set `item.llm_api_key` to a valid OpenHands-managed key (either reuse an existing verified key, or generate a new one).

**Actual on main:**
1. `_ensure_openhands_api_key()` calls `get_user_keys()` → returns existing keys from BYOR
2. Since keys exist, it returns early **without setting `item.llm_api_key`**
3. Settings saved with `llm_api_key=None` → user gets "Invalid proxy server token"

**Actual on this PR:**
1. `_ensure_api_key()` checks `if item.llm_api_key and not verify_existing_key(...)`
2. Since `item.llm_api_key` is `None`, the condition is falsy → **entire block skipped**
3. Settings saved with `llm_api_key=None` → same error

Should the fix handle the case where `item.llm_api_key` is `None` by generating/assigning a key?

enterprise/storage/lite_llm_manager.py
tofarr: Nit: This should probably remove reference to BYOR

tofarr: Nit: This should probably remove reference to BYOR

tofarr: Nit: This should probably remove reference to BYOR

enterprise/storage/org_member_store.py
```diff
@@ -38,7 +39,7 @@ def add_user_to_org(
             return org_member
 
     @staticmethod
-    def get_org_member(org_id: UUID, user_id: int) -> Optional[OrgMember]:
+    def get_org_member(org_id: UUID, user_id: UUID) -> Optional[OrgMember]:
```
tofarr, line 42: Was this working before? It seems to me that we should add a test case for this - possibly using an sqlite in memory db.

tofarr (commented):
Some of the signature changes indicate to the that this code is untested - I'd like to see additional test coverage before merging.

enterprise/storage/lite_llm_manager.py
tofarr: Fixed in ae26b61fd - updated the comment to remove BYOR reference.

enterprise/storage/lite_llm_manager.py
tofarr: Fixed in ae26b61fd - updated docstring to correctly describe that the function returns True if the key is found and valid, False otherwise.

enterprise/storage/lite_llm_manager.py
tofarr: Fixed in ae26b61fd - removed BYOR from log message, now says 'Key verification failed'.

tofarr: Fixed in ae26b61fd - removed BYOR from log message, now says 'Key verification successful'.

tofarr: Fixed in ae26b61fd - removed BYOR from log message, now says 'Key verification error'.

enterprise/storage/saas_settings_store.py
tofarr: Fixed in cebee5152 - now using get_openhands_cloud_key_alias() helper function instead of hardcoded string.

enterprise/storage/lite_llm_manager.py
tofarr: Fixed in cebee5152 - added the x-goog-api-key header to _get_all_keys_for_user for consistency with other LiteLLM API calls.

enterprise/storage/lite_llm_manager.py
tofarr: Added in cebee5152 - created TestVerifyExistingKey class with 8 test cases covering:
- OpenHands type key matching
- OpenHands Cloud alias matching
- BYOR alias matching  
- Wrong team_id handling
- No keys case
- None/empty key_name handling (guards against TypeError)

enterprise/storage/lite_llm_manager.py
tofarr: Added in cebee5152 - created TestGetAllKeysForUser class with 4 test cases covering:
- Missing config handling
- Successful key retrieval (including API key header verification)
- Empty response handling
- API error handling

enterprise/storage/org_member_store.py
tofarr: A test already exists! See `test_get_org_member` in `enterprise/tests/unit/test_org_member_store.py` (line 95-126) which uses SQLite in-memory database via the `session_maker` fixture from conftest.py.

enterprise/storage/lite_llm_manager.py
tofarr: Good catch! The code already guards against this with `if token and key_value.endswith(token)`. When `key_name` is None or empty, `token` becomes falsy and the condition short-circuits before calling `endswith()`. Added tests in cebee5152 (test_verify_existing_key_handles_none_key_name and test_verify_existing_key_handles_empty_key_name) to verify this behavior.

tofarr (approved):

PR #12741, opened 2026-02-04
feat: add SANDBOX_STARTUP_GRACE_SECONDS env var for configurable startup timeout

xingyaoww:
## Summary

This PR adds support for configuring the sandbox startup grace period via the `SANDBOX_STARTUP_GRACE_SECONDS` environment variable.

## Problem

When running OpenHands in Docker, users may encounter sandbox startup failures with errors like:

```
Sandbox entered error state: oh-agent-server-xxxxx
Sandbox server not running: http://host.docker.internal:46137 : Server disconnected without sending a response.
```

The error message "Server disconnected without sending a response" (not "Connection refused") indicates that the TCP connection was established, but the server didn't respond with a complete HTTP response. This typically happens when:

1. The agent-server container is still initializing
2. The HTTP server is bound to the port but not yet ready to handle requests
3. The default 15-second grace period is insufficient for the environment

**Key observation**: Users can successfully `curl` the sandbox endpoint from inside the main container *after* the container has fully started, proving the container works - it just needs more startup time.

## Root Cause Analysis

Looking at the code in `docker_sandbox_service.py`:

```python
STARTUP_GRACE_SECONDS = 15  # hardcoded default

# In _container_to_checked_sandbox_info():
if sandbox_info.created_at < utc_now() - timedelta(
    seconds=self.startup_grace_seconds
):
    _logger.info(f'Sandbox server not running: {app_server_url} : {exc}')
    sandbox_info.status = SandboxStatus.ERROR  # marked as error after grace period
```

The `startup_grace_seconds` parameter exists in `DockerSandboxServiceInjector` but was **not exposed via environment variable** in `config.py`, unlike other sandbox configuration options like `SANDBOX_HOST_PORT` and `SANDBOX_CONTAINER_URL_PATTERN`.

## Solution

Add support for the `SANDBOX_STARTUP_GRACE_SECONDS` environment variable in `config_from_env()`, following the same pattern as other sandbox configuration options.

## Use Cases

This is useful for:

- **Slower machines**: Systems with limited CPU/memory where container initialization takes longer
- **Cloud environments**: EC2 and similar environments where Docker networking may have initial latency
- **First-time startup**: When Python dependencies need to be loaded for the first time
- **Resource contention**: When the sandbox container competes with the main container for resources

## Usage

```bash
docker run -it -d \
  -e SANDBOX_STARTUP_GRACE_SECONDS=60 \
  # ... other options ...
  docker.openhands.dev/openhands/openhands:latest
```

## Backward Compatibility

The default remains **15 seconds** for backward compatibility. This change is purely additive and opt-in.

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:91e10c3-nikolaik   --name openhands-app-91e10c3   docker.openhands.dev/openhands/openhands:91e10c3
```


xingyaoww:
@neubig this PR should be unrelated to that -

This PR solves the issue in the OSS when the app try to start a new docker container with *Docker* (not k8s), and it took longer than 15s, causing the app to timeout.

neubig (approved):

PR #12750, opened 2026-02-04
Fix LLM key regeneration logic and default base URL handling

tofarr:
## Summary of PR

When saving settings, the code now checks the `llm_base_url` from the new settings (`item.llm_base_url`) instead of deriving it from existing stored values (`org_member.llm_base_url` or `org.default_llm_base_url`). This ensures that API keys are only regenerated when the new settings actually indicate an OpenHands provider with the default LiteLLM API URL.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:51a439c-nikolaik   --name openhands-app-51a439c   docker.openhands.dev/openhands/openhands:51a439c
```

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -161,7 +161,7 @@ class SaasSettingsStore(SettingsStore):
                 return None
 
             # Check if provider is OpenHands and generate API key if needed
-            if item.llm_base_url == LITE_LLM_API_URL:
+            if item.llm_base_url == LITE_LLM_API_URL or not item.llm_base_url:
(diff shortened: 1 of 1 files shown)


chuckbutkus:
I think this will break the fix for what Mahmood was seeing.  The fix made sure to generate a new key when moving from OpenHands provider to another provider.  This will break that.

tofarr:
@openhands-exp Flip a coin!

Issue #12865, opened 2026-02-13
"Unknown event" shown before file read operations complete


enyst:
## Description

Users report seeing "Unknown event" displayed in the UI when the agent is about to read a file (e.g., before "Reading README.md" is shown). After the file read operation completes, there is another observation which shows correctly (e.g., "Read README.md").

## Steps to Reproduce

1. Start a conversation with the agent
2. Ask the agent to read or view a file (triggers a `FileEditorAction` with `command: "view"`)
3. Observe the chat interface during the action execution

**Expected Behavior:** The UI should show maybe only "Reading <path>" while the action is being executed. It shouldn't show "Unknown event".

**Actual Behavior:** "Unknown event" is shown. Then another event "Read <path>" appears. The second has the correct observation message.

## Technical Investigation

The "Unknown event" message originates from `EVENT$UNKNOWN_EVENT` i18n key (`frontend/src/i18n/translation.json`).

In `frontend/src/components/v1/chat/event-content-helpers/get-event-content.tsx`, the title defaults to "Unknown event" when:
```typescript
return {
  title: title || i18n.t("EVENT$UNKNOWN_EVENT"),  // line 210
  details: details || i18n.t("EVENT$UNKNOWN_EVENT"),
};
```

This happens when `title` is an empty string, which occurs when an event fails the `isActionEvent()` type guard in `getActionEventTitle()` (line 43-44).

### Relevant Type Guard

The `isActionEvent` type guard in `frontend/src/types/v1/type-guards.ts` requires:
- `event.source === "agent"`
- `"action" in event && event.action !== null`
- `typeof event.action === "object" && "kind" in event.action`
- `"tool_name" in event && "tool_call_id" in event`
- Both `tool_name` and `tool_call_id` must be strings

### Potential Causes

1. **Event arrives with missing fields:** The SDK's `ActionEvent` has `action: Action | None = Field(default=None, ...)`, meaning `action` can be `null`. If an event arrives with `action: null`, it would fail `isActionEvent()`.

2. **Streaming/partial events:** Events might be arriving before all fields are populated during WebSocket streaming. (HUMAN: ? this is LLM written)

3. **Event timing:** There might be a brief window where an event is rendered before it's fully validated.

### Related Files

- `frontend/src/components/v1/chat/event-content-helpers/get-event-content.tsx`
- `frontend/src/types/v1/type-guards.ts`
- `frontend/src/components/v1/chat/event-message.tsx`
[...]

PR #12868, opened 2026-02-13
fix: show action title for partial events instead of 'unknown event'

nehaaprasad:
fix: #12865

### Summary of PR

- Stops "Unknown event" from showing while the agent is reading/editing a file.
- Treats events with action === null but valid tool_name/tool_call_id as partial actions.
- Infers the label from tool_name and tool_call arguments so the UI shows e.g. "Reading README.md" instead of "Unknown event".

### Change type
- [x] Bug fix

### Release notes (if checked)

- Fix: Show correct in-progress label (e.g. "Reading …") instead of "Unknown event" when the agent reads or edits a file.

### DIFF
--- a/frontend/src/components/v1/chat/event-content-helpers/get-event-content.tsx
+++ b/frontend/src/components/v1/chat/event-content-helpers/get-event-content.tsx
@@ -37,6 +37,92 @@ const createTitleFromKey = (
   );
 };
 
+const isPartialActionEvent = (event: OpenHandsEvent): boolean => {
+  if (event.source !== "agent") {
+    return false;
+  }
+
+  if (!("tool_name" in event) || !("tool_call_id" in event)) {
+    return false;
+  }
+
+  if (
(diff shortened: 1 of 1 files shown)


enyst (commented):
Hey @naaa760 ! Thank you for looking into this!

Have you replicated the issue? I think it should be easy to see, the agent just needs to do some editing / viewing files.

nehaaprasad:
@enyst 
Yes sir i have replicated it. I added unit tests that simulate partial action events (with action: null but valid tool_name and tool_call) and confirm the fix: the UI shows the correct title (e.g. "Reading …") instead of "Unknown event", and details stay empty. the tests pass.

neubig:
Since requested changes were left more than 2 weeks ago and there has not been a response, we are planning to close this PR for lack of activity. If you would like to take additional action, please push updates or leave a comment and we would be happy to take another look. Thank you for the contribution.

PR #12889, opened 2026-02-16
Default model for new users is minimax

tofarr:
## Summary of PR

The default model for new users should be minimax.

## Change Type

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:6d75311-nikolaik   --name openhands-app-6d75311   docker.openhands.dev/openhands/openhands:6d75311
```

### DIFF
--- a/enterprise/server/constants.py
+++ b/enterprise/server/constants.py
@@ -30,6 +30,8 @@ PERSONAL_WORKSPACE_VERSION_TO_MODEL = {
     2: 'claude-3-7-sonnet-20250219',
     3: 'claude-sonnet-4-20250514',
     4: 'claude-sonnet-4-20250514',
+    # Minimax is now the default as it gives results close to claude in terms of quality
+    # but at a much lower price
     5: 'minimax-m2.5',
 }
 


enterprise/server/constants.py
```diff
@@ -30,7 +30,7 @@
     2: 'claude-3-7-sonnet-20250219',
     3: 'claude-sonnet-4-20250514',
     4: 'claude-sonnet-4-20250514',
-    5: 'claude-opus-4-5-20251101',
+    5: 'minimax-m2.5',
```
hieptl, line 35: **question:** This change may apply only to new accounts—could you please confirm if that is correct?

For existing accounts, they would continue using Opus, as ORG_SETTINGS_VERSION has not been updated and remains set to 5.

Thank you! 🙏

PR #12909, opened 2026-02-18
fix: drop top_p for Claude Opus 4.x minor models

### LINKED ISSUE #12846: claude-opus-4-6 fails: temperature and top_p cannot both be specified
## Description

When using `anthropic/claude-opus-4-6` as the model, the resolver fails immediately with:

```
litellm.exceptions.BadRequestError: AnthropicException - {"type":"error","error":{"type":"invalid_request_error","message":"`temperature` and `top_p` cannot both be specified for this model. Please use only one."},"request_id":"..."}
```

The error occurs on the very first LLM call in `agent_controller.py`, and also in `guess_success` in `issue_definitions.py`.

## Steps to Reproduce

1. Configure OpenHands resolver with `LLM_MODEL=anthropic/claude-opus-4-6`
2. Run `python -m openhands.resolver.resolve_issue`
3. Fails immediately with the error above

## Environment

- openhands-ai 1.3.0
- LiteLLM (whatever version is pulled in by openhands-ai 1.3.0)
- Direct Anthropic API (not Bedrock)

## Notes

- Other Claude models (e.g. `anthropic/claude-sonnet-4-5`) work fine
- `LITELLM_DROP_PARAMS=true` does not help — both `temperature` and `top_p` are individually valid Anthropic params, they just can't be sent together for this model
- A similar issue (#11248) was fixed for Bedrock-prefixed model names, but the fix doesn't cover direct Anthropic API model names like `anthropic/claude-opus-4-6`

neilpattanaik:
Fixes #12846.

Anthropic rejects sending both `temperature` and `top_p` for some Claude 4 models (including `anthropic/claude-opus-4-6`). This change drops `top_p` when both are set for:
- `claude-opus-4-<minor>` (e.g. 4-1/4-5/4-6)
- `claude-sonnet-4*`

Adds a unit test covering `anthropic/claude-opus-4-6`.

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -7,6 +7,7 @@
 # Tag: Legacy-V0
 # V1 replacement for this module lives in the Software Agent SDK.
 import copy
+import re
 import os
 import time
 import warnings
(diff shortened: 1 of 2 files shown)


enyst (commented):
Thank you for the contribution! Could you please take a look at the lint fail?

enyst:
Actually, there was another PR earlier than this, and we merged it. I'm going to close this, but thank you for the contribution!

Please note for the future that this PR is to V0, the legacy version of OpenHands. This code is not in use for V1, the current OpenHands, and we recommend to not work on code additions to the legacy files, we will clean them up instead soon.

PR #12912, opened 2026-02-18
fix(frontend): stop AgentErrorEvent toast replays on reload

### LINKED ISSUE #12757: Avoid toast for AgentErrorEvent
## Problem

AgentErrorEvent currently triggers a toast in addition to the inline chat error message. When the conversation page is reloaded and historical events are loaded, the toast appears again even for old errors, sometimes with very large payloads.

## Steps to reproduce
1. Trigger an AgentErrorEvent (e.g., tool validation error with a large JSON payload).
2. Observe the toast appearing with a large content, blocking visibility of the tab.
3. Reload the conversation page.
4. Observe the toast reappearing as past events are loaded. 

## Expected behavior
- No toast for AgentErrorEvent (inline error in the chat is sufficient).
- Toasts should not reappear for historical events on reload.

## Notes
The error banner issue tracked in #12332 has been addressed; this issue focuses specifically on the toast behavior.

neilpattanaik:
Fixes #12757.

`use-handle-ws-events` (and the v0 equivalent) previously treated any event containing an `error` field as a server error and displayed a toast.

`AgentErrorEvent` includes an `error` field but is already rendered inline in chat, so this caused duplicate/replayed toasts after reload.

This PR tightens the server-error guard to require both `error` and `message`, which prevents `AgentErrorEvent` from being interpreted as a toast-worthy server error.

### DIFF
--- a/openhands/app_server/event_callback/sql_event_callback_service.py
+++ b/openhands/app_server/event_callback/sql_event_callback_service.py
@@ -7,7 +7,7 @@ import asyncio
 import logging
 from dataclasses import dataclass
 from typing import AsyncGenerator
-from uuid import UUID
+from uuid import UUID, uuid4
 
 from fastapi import Request
 from sqlalchemy import UUID as SQLUUID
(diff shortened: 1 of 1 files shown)


enyst:
@openhands Can we implement this check without peeking into the structure of the data, maybe we have a type? Investigate for `AgentErrorEvent` or types of events.

hieptl:
Hello @neilpattanaik,

I hope you’re doing well.

I wanted to check in to see if you’re still working on this pull request. When you have a moment, could you please share an update on the current status?

If so, it would be great if you could also add tests to cover your front-end changes. Additionally, a short demo video demonstrating how the output looks would be very helpful.

Thank you very much! 🙏

mamoodi:
Looks like the original ticket has been closed so I'm going to close this PR. Thanks for your contribution!

PR #12925, opened 2026-02-19
fix: Cap Databricks max_tokens at 25000 and prevent condenser crash

juanmichelini:
## Summary
This PR provides a minimal fix for two issues:

### 1. Databricks max_tokens limit
The Databricks API has a hard limit of 25000 for `max_tokens`, but OpenHands was sending higher values, causing API errors.

### 2. AmortizedForgettingCondenser crash
The condenser would crash with `ValueError: min() arg is an empty sequence` when trying to condense a view with too few events.

### Databricks Fix (1 location)
- **`openhands/llm/llm.py`**: Cap `max_tokens` at 25000 after `completion_kwargs` are processed
  - Ensures Databricks uses `max_tokens` (not `max_completion_tokens`)
  - Single location fix instead of multiple places
  - Applied right before the completion function is created
  - Covers all code paths (V0, V1, any future versions)

### Condenser Fix
- **`openhands/memory/condenser/impl/amortized_forgetting_condenser.py`**: 
  - Add check in `should_condense()` to return False if no events can be forgotten
  - Add defensive check in `get_condensation()` to handle empty `event_ids_to_forget`
- **`tests/unit/memory/condenser/test_condenser.py`**: Add test for the fix

## Why This is Minimal

The previous version of this PR attempted to fix the Databricks issue in **5 different places**. This version fixes it in **1 location** - right where kwargs are passed to litellm, ensuring:
- All code paths are covered (V0, V1, any future versions)
- The fix happens after completion_kwargs, preventing overrides
- No redundant checks in multiple places

## Testing

✅ Unit test for condenser fix passes  
✅ Pre-commit hooks pass  
✅ Only 3 files changed, 64 lines added, 1 line removed

## Files Changed
- `openhands/llm/llm.py` - Databricks max_tokens cap (15 lines added)
- `openhands/memory/condenser/impl/amortized_forgetting_condenser.py` - Condenser fix (23 lines added)
- `tests/unit/memory/condenser/test_condenser.py` - Test for condenser fix (27 lines added)

**Total: 3 files, 64 lines added, 1 line removed**

### DIFF
--- a/openhands/llm/llm.py
+++ b/openhands/llm/llm.py
@@ -214,6 +214,21 @@ class LLM(RetryMixin, DebugMixin):
         if self.config.completion_kwargs is not None:
             kwargs.update(self.config.completion_kwargs)
 
+        # Databricks has a 25000 max_tokens limit - cap it AFTER completion_kwargs
+        if self.config.model.startswith('databricks/'):
+            # Get the effective max tokens value
+            effective_max = kwargs.get('max_tokens') or kwargs.get(
+                'max_completion_tokens'
+            )
+            if effective_max and effective_max > 25000:
+                logger.warning(
+                    f'Databricks model {self.config.model}: Capping max_tokens from {effective_max} to 25000 (API limit)'
+                )
+            # Databricks uses max_tokens parameter (not max_completion_tokens)
+            # Cap at 25000 or use the effective max if it's already lower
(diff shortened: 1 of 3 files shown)


juanmichelini:
pausing this until getting confirmation, see https://linear.app/all-hands-ai/issue/ALL-5331/check-connection-to-databricks-llms

juanmichelini:
Closing as stale

PR #12936, opened 2026-02-19
Fix 'Unknown event' displayed before file operations complete

### LINKED ISSUE #12865: "Unknown event" shown before file read operations complete

paipeline:
**Fixes #12865**

## Problem
Users were seeing "Unknown event" displayed in the UI when the agent was about to read files or perform other actions. This occurred when ActionEvents arrived via WebSocket streaming with incomplete data (e.g., `action: null` or missing `tool_name`/`tool_call_id` fields), causing them to fail the strict `isActionEvent()` type guard.

## Root Cause
The `getActionEventTitle()` function returned an empty string when `isActionEvent()` failed, which eventually defaulted to "Unknown event" in the UI. This happened during the brief window when events were being streamed but not yet fully populated.

## Solution
- **Added `isPartialActionEvent()` helper** to detect streaming/partial ActionEvents
- **Split event handling** into separate functions for complete vs partial events
- **Extract available information** from partial events (file paths, commands) when possible  
- **Show "Loading..."** for partial events instead of "Unknown event"
- **Maintains backward compatibility** with existing complete events

## Technical Details
The fix handles ActionEvents that arrive with:
- `action: null` (completely missing action data)
- Partial action objects with `kind` but missing other required fields
- Missing `tool_name` or `tool_call_id` properties

For partial events, the code now:
1. Extracts whatever information is available (e.g., file path for FileEditorAction)
2. Uses appropriate action messages when possible ("Reading {path}" vs "Loading...")  
3. Falls back to generic "Loading..." when no specific information is available

## Testing
- Verified the function compiles and follows existing patterns
- Used existing translation keys (HOME$LOADING) for consistency
- Preserves all existing behavior for complete events

This provides a much better user experience during file operations and other streaming actions.

### DIFF
--- a/frontend/src/components/v1/chat/event-content-helpers/get-event-content.tsx
+++ b/frontend/src/components/v1/chat/event-content-helpers/get-event-content.tsx
@@ -16,6 +16,18 @@ const trimText = (text: string, maxLength: number): string => {
   return text.length > maxLength ? `${text.substring(0, maxLength)}...` : text;
 };
 
+// Helper function to detect partial ActionEvents that are still loading
+const isPartialActionEvent = (event: OpenHandsEvent): boolean => {
+  return (
+    event.source === "agent" &&
+    "action" in event &&
+    (event.action === null ||
+      (typeof event.action === "object" &&
+        event.action !== null &&
+        "kind" in event.action))
+  );
+};
+
 // Helper function to create title from translation key
 const createTitleFromKey = (
   key: string,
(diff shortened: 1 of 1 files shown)


neubig (changes requested):

mamoodi:
The issue has been fixed so I'm going to close this PR. Thanks for your contribution!

PR #12947, opened 2026-02-19
Deprecate Legacy-V0 server routes in OpenAPI

enyst:
OpenHands-GPT-5.2 here.

## Problem
The generated OpenAPI spec includes legacy V0 endpoints from `openhands/server/routes/*`. Some of these were missing `deprecated=True`, so they did not show the Swagger UI strikethrough even though the modules are clearly tagged as legacy ("Tag: Legacy-V0"). This makes it easy for users to mistakenly treat V0 endpoints as current.

### 1) Mark Legacy-V0 endpoints as deprecated in OpenAPI
Adds `deprecated=True` to FastAPI route decorators for **all** operations implemented in modules tagged `Tag: Legacy-V0` under `openhands/server/routes/`.

This makes Swagger UI consistently render these as deprecated (strikethrough) while keeping the endpoints functional for backward compatibility.

#### Endpoints covered (Legacy-V0 `/api/*` operations)
```text
POST   /api/add-git-providers
GET    /api/conversations
POST   /api/conversations
DELETE /api/conversations/{conversation_id}
GET    /api/conversations/{conversation_id}
PATCH  /api/conversations/{conversation_id}
GET    /api/conversations/{conversation_id}/config
GET    /api/conversations/{conversation_id}/events
POST   /api/conversations/{conversation_id}/events
POST   /api/conversations/{conversation_id}/exp-config
GET    /api/conversations/{conversation_id}/git/changes
GET    /api/conversations/{conversation_id}/git/diff
GET    /api/conversations/{conversation_id}/list-files
POST   /api/conversations/{conversation_id}/message
GET    /api/conversations/{conversation_id}/microagents
GET    /api/conversations/{conversation_id}/remember-prompt
GET    /api/conversations/{conversation_id}/select-file
POST   /api/conversations/{conversation_id}/start
POST   /api/conversations/{conversation_id}/stop
POST   /api/conversations/{conversation_id}/submit-feedback
GET    /api/conversations/{conversation_id}/trajectory
POST   /api/conversations/{conversation_id}/upload-files
GET    /api/conversations/{conversation_id}/vscode-url
GET    /api/conversations/{conversation_id}/web-hosts
GET    /api/conversations/{conversation_id}/zip-directory
GET    /api/microagent-management/conversations
GET    /api/options/agents
GET    /api/options/config
GET    /api/options/models
GET    /api/options/security-analyzers
POST   /api/reset-settings
GET    /api/secrets
POST   /api/secrets
DELETE /api/secrets/{secret_id}
PUT    /api/secrets/{secret_id}
GET    /api/settings
POST   /api/settings
POST   /api/unset-provider-tokens
GET    /api/user/info
GET    /api/user/installations
GET    /api/user/repositories
GET    /api/user/repository/branches
GET    /api/user/repository/{repository_name:path}/microagents
GET    /api/user/repository/{repository_name:path}/microagents/content
GET    /api/user/search/branches
GET    /api/user/search/repositories
GET    /api/user/suggested-tasks
```

### 2) Add a temporary audit script (no allowlist)
Adds `scripts/check_legacy_v0_routes_deprecated.py` which:
- Loads `openhands.server.app:app`
- Finds routes whose endpoint module is under `openhands/server/routes/**`
- Detects modules tagged with `Tag: Legacy-V0`
- Fails if any such `/api/*` operation is missing `deprecated=True`

This is intended as a quick, local guardrail while V0 is being retired.

## How to verify
1) Generate the OpenAPI spec (same mechanism used by Swagger UI):
   ```bash
   mkdir -p docs
   uv run python scripts/update_openapi.py
   ```
2) Confirm operations show `"deprecated": true` in `docs/openapi.json` (e.g. the trajectory path).
3) Run the audit script:
   ```bash
   uv run python scripts/check_legacy_v0_routes_deprecated.py
   ```
   Expected: `OK: all Legacy-V0 routes under openhands/server/routes are deprecated`

## Notes
- This PR is intentionally scoped to endpoints implemented in Legacy-V0 modules under `openhands/server/routes/`.
- No allowlist is used: if a module is tagged Legacy-V0, all its `/api/*` operations should be marked deprecated.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:a0220bf-nikolaik   --name openhands-app-a0220bf   docker.openhands.dev/openhands/openhands:a0220bf
```


openhands/server/routes/git.py
```diff
@@ -72,7 +72,7 @@ async def get_user_installations(
     raise AuthenticationError('Git provider token required. (such as GitHub).')
 
 
-@app.get('/repositories', response_model=list[Repository])
+@app.get('/repositories', response_model=list[Repository], deprecated=True)
```
malhotra5, line 77: Are these deprecated? I think they apply to parts of the web app regardless of whether we are in V0 or V1 convos

enyst: AGENT:
Good call-out. These endpoints are still used by the web app in both V0 and V1 modes, but they live in `openhands/server/routes/*` modules tagged `Legacy-V0` (the whole legacy server surface is slated for retirement). Marking them `deprecated=True` here is only an OpenAPI/Swagger signal (no runtime behavior change) to discourage new external consumers while we migrate/replace them before V0 removal.

enyst:
Hi — OpenHands-GPT-5.2 here. My role in this discussion: help clarify repo architecture/API boundaries (V0 legacy server vs V1 app_server) to inform how we treat deprecation flags in this PR.

“Legacy‑V0” in this PR is a signal about **which server implementation owns the route**, not about whether the *frontend* is in a “V0 vs V1 conversation” mode.

### What is actually “V1” in this repo?
In OpenHands today, “V1 API” effectively means:

- **Implemented under** `openhands/app_server/**`
- **Mounted under** the router prefix **`/api/v1`** (see `openhands/app_server/v1_router.py`)

So if the path is `/api/v1/...`, it’s the V1 surface area.

### What is “Legacy/V0”? 
Routes implemented under:

- `openhands/server/routes/**` (and mounted by `openhands/server/app.py`)
- typically live under **`/api/...`** (unversioned legacy API)

Those modules are explicitly tagged `Tag: Legacy-V0`, and the whole `openhands/server/app.py` file is labeled legacy.

### So is `openhands/server/routes/git.py` “actually V1”? 
No—**it’s legacy API** even if the modern web app still calls it.

Concrete evidence from the codebase:
- `git.py` defines routes under `APIRouter(prefix=/api/user)` → legacy `/api/user/...`
- The V1 server (`openhands/app_server/**`) does **not** define any `/repositories`, `/installations`, `/suggested-tasks`, etc. endpoints (I scanned `openhands/app_server/*.py` for those strings; none exist).
- The frontend *does* call `/api/user/...` for git/repo functionality (e.g. `frontend/src/api/git-service/git-service.api.ts`).

So what’s happening is:
- **V1 UI / V1 conversations** still depend on some **legacy `/api/...` endpoints** that haven’t been migrated yet.
- That doesn’t make those endpoints “V1”; it means V1 is still mid-migration.

### If we *don’t* want those endpoints deprecated, what would make them “actually V1”? 
One of these has to happen (product/architecture decision):

1) **Migrate** those endpoints into `openhands/app_server/` under `/api/v1/...` and update the frontend to use the new paths; then deprecate/remove `/api/user/...`.
2) Decide they’re not legacy and will remain supported long-term → then the **`Tag: Legacy-V0` header is wrong**, and we should remove/retag and adjust the “audit script” policy accordingly.

But with the code as it exists today, `/api/user/*` in `openhands/server/routes/git.py` is unambiguously part of the legacy surface area, even if the V1 product still uses it.

openhands/server/routes/git.py
enyst: HUMAN: ^^ the agent was speaking above.

It is true that all I want in this PR is to mark as deprecated in the [Swagger docs here](https://app.all-hands.dev/docs) so that people don't rely on them.

Will they actually stay in place as they are? If yes, should we remove the _legacy_ comment from the file?

malhotra5: Yes I think we should remove the legacy comment, many of these API routes are here to stay

tofarr: I think we need another round of decisions regarding the APIs which are V0 but not directly related to conversations.

enyst:
@openhands Understand this PR, and then merge main into it and fix the conflicts.

openhands/server/routes/git.py
enyst: Where is that discussion?

tofarr:
All endpints have been updated in a separated PR - I think this PR is now redundant.

PR #12965, opened 2026-02-20
fix: redact secret values from error messages in add_env_vars

jpshackelford:
## Summary

Fixes a security issue where secret values were being exposed in error logs when `add_env_vars()` fails.

## Problem

When adding environment variables fails (e.g., due to invalid variable names like `MY_DUMMY-SECRET` with a hyphen), the error message contained the full bash command output which included all secret values in plaintext:

```
Error creating agent_session: Failed to add env vars [...] to environment: 
export GRB_LICENSEID="SECRET_VALUE"; export MY_DUMMY-SECRET="blahblah"
bash: export: `MY_DUMMY-SECRET=blahblah': not a valid identifier
```

This was being logged via `self.logger.exception(f'Error creating agent_session: {e}')` in `session.py`, exposing secrets in pod logs.

## Solution

Modified `add_env_vars()` in `openhands/runtime/base.py` to catch `RuntimeError` exceptions from `_run_cmd_with_retry()` and re-raise with a sanitized error message that only includes variable **names** (keys), not the secret **values**:

```python
try:
    self._run_cmd_with_retry(
        cmd, f'Failed to add env vars [{env_vars.keys()}] to environment'
    )
except RuntimeError as e:
    # Re-raise with redacted error message to avoid leaking secret values
    raise RuntimeError(
        f'Failed to add env vars [{list(env_vars.keys())}] to environment. '
        'Check that all variable names are valid bash identifiers '
        '(alphanumeric and underscores only, cannot start with a digit).'
    ) from None
```

## Result

Error messages now show only variable names with helpful guidance:
```
Failed to add env vars ['GRB_LICENSEID', 'MY_DUMMY-SECRET'] to environment. 
Check that all variable names are valid bash identifiers 
(alphanumeric and underscores only, cannot start with a digit).
```

## Changes

- `openhands/runtime/base.py`: Wrap `_run_cmd_with_retry()` calls in `add_env_vars()` with try-catch to sanitize error messages
  - Windows PowerShell section
  - Linux/Unix bash export section  
  - Linux/Unix .bashrc persistence section

## Tests

Added `tests/unit/runtime/test_env_vars_security.py` with 3 tests that verify:
- Invalid env var names raise `RuntimeError` without exposing secret values
- Multiple env vars errors don't expose any secrets
- `.bashrc` persistence errors don't expose secrets

Tests were verified to:
- **FAIL** without the fix (secrets exposed in error messages)
- **PASS** with the fix (secrets redacted)

@jpshackelford can click here to [continue refining the PR](https://app.all-hands.dev/conversations/None)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:87176c6-nikolaik   --name openhands-app-87176c6   docker.openhands.dev/openhands/openhands:87176c6
```

### DIFF
--- a/openhands/runtime/base.py
+++ b/openhands/runtime/base.py
@@ -356,7 +356,7 @@ class Runtime(FileEditRuntimeMixin):
                 self._run_cmd_with_retry(
                     cmd, f'Failed to add env vars [{env_vars.keys()}] to environment'
                 )
-            except RuntimeError as e:
+            except RuntimeError:
                 # Re-raise with redacted error message to avoid leaking secret values
                 raise RuntimeError(
                     f'Failed to add env vars [{list(env_vars.keys())}] to environment. '
@@ -387,7 +387,7 @@ class Runtime(FileEditRuntimeMixin):
                 self._run_cmd_with_retry(
                     cmd, f'Failed to add env vars [{env_vars.keys()}] to environment'
                 )
-            except RuntimeError as e:
+            except RuntimeError:
                 # Re-raise with redacted error message to avoid leaking secret values
                 # The original error may contain the full export command with secrets
                 raise RuntimeError(
@@ -403,7 +403,7 @@ class Runtime(FileEditRuntimeMixin):
                 self._run_cmd_with_retry(
                     bashrc_cmd, f'Failed to add env vars [{env_vars.keys()}] to .bashrc'
                 )
-            except RuntimeError as e:
+            except RuntimeError:
                 # Re-raise with redacted error message to avoid leaking secret values
                 raise RuntimeError(
                     f'Failed to add env vars [{list(env_vars.keys())}] to .bashrc. '
--- a/openhands/runtime/base.py
+++ b/openhands/runtime/base.py
@@ -14,6 +14,7 @@ import random
 import shlex
 import shutil
 import string
+import sys
 import tempfile
 import time
 from abc import abstractmethod
@@ -318,8 +319,61 @@ class Runtime(FileEditRuntimeMixin):
             return obs.content
         return str(obs)
 
+    def _run_env_cmd_redacted(
+        self,
+        cmd: str,
+        context: str,
+        var_names: list[str],
+        is_windows: bool = False,
+    ) -> CmdOutputObservation:
+        """Run a command that may contain secrets, redacting them from any errors.
+
+        This wrapper ensures that if the command fails, secret values are not
+        exposed in error messages. Only variable names are included in errors.
+
+        Args:
+            cmd: The command to execute (may contain secret values)
+            context: Description of what operation is being performed
+            var_names: List of environment variable names being set
+            is_windows: Whether this is a Windows environment (affects error message)
+
+        Returns:
+            CmdOutputObservation on success
+
+        Raises:
+            RuntimeError: With redacted message on failure (no secret values)
+        """
+        try:
+            return self._run_cmd_with_retry(
+                cmd, f'Failed to {context} for env vars {var_names}'
+            )
+        except RuntimeError as original_error:
+            # Log the original error at debug level for operators, but redact secret values
+            # We only log the variable names, not the command which contains secrets
+            logger.debug(
+                f'Environment variable command failed for vars {var_names}: '
+                f'exit_code and error details redacted to protect secrets'
+            )
+
+            # Build appropriate error message based on platform
+            if is_windows:
+                hint = 'Check that all variable names are valid PowerShell identifiers.'
+            else:
+                hint = (
+                    'Check that all variable names are valid bash identifiers '
+                    '(alphanumeric and underscores only, cannot start with a digit).'
+                )
+
+            raise RuntimeError(
+                f'Failed to {context} for env vars {var_names}. {hint}'
+            ) from None
+
     def add_env_vars(self, env_vars: dict[str, str]) -> None:
+        if not env_vars:
+            return
+
         env_vars = {key.upper(): value for key, value in env_vars.items()}
+        var_names = list(env_vars.keys())
 
         # Add env vars to the IPython shell (if Jupyter is used)
         if any(isinstance(plugin, JupyterRequirement) for plugin in self.plugins):
@@ -328,13 +382,18 @@ class Runtime(FileEditRuntimeMixin):
                 # Note: json.dumps gives us nice escaping for free
                 code += f'os.environ["{key}"] = {json.dumps(value)}\n'
             code += '\n'
-            self.run_ipython(IPythonRunCellAction(code))
-            # Note: we don't log the vars values, they're leaking info
-            logger.debug('Added env vars to IPython')
-
-        # Check if we're on Windows
-        import os
-        import sys
+            obs = self.run_ipython(IPythonRunCellAction(code))
+            # Check for errors in IPython output - the observation content may contain secrets
+            if isinstance(obs, ErrorObservation):
+                logger.debug(
+                    f'IPython env var setup failed for vars {var_names}: '
+                    'error details redacted to protect secrets'
+                )
+                raise RuntimeError(
+                    f'Failed to add env vars {var_names} to IPython environment. '
+                    'Check that all variable names are valid Python identifiers.'
+                )
+            logger.debug(f'Added env vars to IPython: {var_names}')
 
         is_windows = os.name == 'nt' or sys.platform == 'win32'
 
@@ -346,26 +405,16 @@ class Runtime(FileEditRuntimeMixin):
                 # Note: json.dumps gives us nice escaping for free
                 cmd += f'$env:{key} = {json.dumps(value)}; '
 
-            if not cmd:
-                return
-
             cmd = cmd.strip()
-            logger.debug('Adding env vars to PowerShell')  # don't log the values
+            logger.debug(f'Adding env vars to PowerShell: {var_names}')
 
-            try:
-                self._run_cmd_with_retry(
-                    cmd, f'Failed to add env vars [{env_vars.keys()}] to environment'
-                )
-            except RuntimeError:
-                # Re-raise with redacted error message to avoid leaking secret values
-                raise RuntimeError(
-                    f'Failed to add env vars [{list(env_vars.keys())}] to environment. '
-                    'Check that all variable names are valid identifiers.'
-                ) from None
+            self._run_env_cmd_redacted(
+                cmd, 'set environment variables', var_names, is_windows=True
+            )
 
             # We don't add to profile persistence on Windows as it's more complex
             # and varies between PowerShell versions
-            logger.debug(f'Added env vars to PowerShell session: {env_vars.keys()}')
+            logger.debug(f'Added env vars to PowerShell session: {var_names}')
 
         else:
             # Original bash implementation for Unix systems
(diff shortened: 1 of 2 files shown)


jpshackelford:
# 🔥 CODE REVIEW: PR #12965 - fix: redact secret values from error messages in add_env_vars

**Taste Rating: 🟡 Acceptable** - Works, but the design choice is defensive rather than elegant.

---

## Linus-Style Analysis

### **[CRITICAL ISSUES]** - Must Fix

**1. [tests/unit/runtime/test_env_vars_security.py, Lines 17-42] - Mocks Aren't Tests. This is Frankenstein Testing.**

You've created a monstrosity. You're mocking `Runtime`, then manually rebinding *actual methods* from the class back onto the mock with `__get__`:

```python
runtime = MagicMock(spec=Runtime)
runtime._run_cmd_with_retry = Runtime._run_cmd_with_retry.__get__(runtime, Runtime)
runtime._is_bash_session_timeout = Runtime._is_bash_session_timeout.__get__(runtime, Runtime)
# ... and so on
```

This is **not testing the real code path** — this is surgically Frankensteining together pieces of a class onto a mock skeleton. If someone changes `Runtime.__init__` tomorrow to set a property that `add_env_vars` depends on, your tests will pass while the real code burns.

What you've written isn't a unit test or an integration test. It's a mock that's 80% real code, which is the worst of both worlds: all the fragility of mocks with the false confidence of "testing real code."

**Fix:** Either:
- Write an **actual integration test** that instantiates a real `Runtime` (or a test subclass with minimal overrides), OR
- If you must mock, mock ONLY the external boundary (`self.run`) and don't rebind internal methods — but then accept you're testing "behavior given this mock," not the real code.

---

**2. [openhands/runtime/base.py, Lines 336-339] - Importing Inside a Method is a Code Smell**

```python
def add_env_vars(self, env_vars: dict[str, str]) -> None:
    ...
    # Check if we're on Windows
    import os
    import sys

    is_windows = os.name == 'nt' or sys.platform == 'win32'
```

Why are you importing `os` and `sys` *inside* a method? These are standard library modules already imported at the top of the file (line 12 has `import os`). This is either dead cargo-cult code or someone who doesn't understand Python's module system. Either way, it's ugly.

**Note:** This is pre-existing code, not from this PR. But you touched this method, so you own this mess now.

---

### **[IMPROVEMENT OPPORTUNITIES]** - Should Fix

**3. [openhands/runtime/base.py, Lines 359-364, 390-397, 406-412] - DRY Violation: Three Identical Try-Except Blocks**

You've copy-pasted the same pattern **three times**:

```python
try:
    self._run_cmd_with_retry(cmd, ...)
except RuntimeError:
    raise RuntimeError(
        f'Failed to add env vars [{list(env_vars.keys())}] to environment. '
        'Check that all variable names are valid ...'
    ) from None
```

This is lazy. You wrote the same exception handler three times with slightly different messages. What happens when someone needs to change the error format? Three places to update.

**Fix:** Extract a helper method:

```python
def _run_cmd_redacted(self, cmd: str, context: str, var_names: list[str]) -> CmdOutputObservation:
    try:
        return self._run_cmd_with_retry(cmd, f'Failed to {context}')
    except RuntimeError:
        raise RuntimeError(
            f'Failed to {context} for env vars {var_names}. '
            'Check that all variable names are valid bash identifiers '
            '(alphanumeric and underscores only, cannot start with a digit).'
        ) from None
```

---

**4. [openhands/runtime/base.py, Lines 359-364] - Windows and Unix Messages Are Inconsistent**

Windows says:
> "Check that all variable names are valid identifiers."

Unix says:
> "Check that all variable names are valid bash identifiers (alphanumeric and underscores only, cannot start with a digit)."

Why? PowerShell has different rules than bash (PowerShell allows hyphens in some cases, actually). But if you're going to be helpful, be helpful everywhere. If you're going to be terse, be terse everywhere. Pick one.

---

**5. [openhands/runtime/base.py, Line 359] - `from None` Hides the Traceback**

```python
raise RuntimeError(...) from None
```

Using `from None` suppresses the exception chain entirely. For a security fix where you *want* to hide secrets from the error message, this makes sense. But you've now made debugging production issues harder because you won't see what the underlying error actually was.

Consider logging the original exception (with values redacted) at DEBUG level before raising the sanitized one. That way operators can still debug, but the error message in logs/user-facing output stays clean.

---

**6. [tests/unit/runtime/test_env_vars_security.py, Lines 116-117] - Test Will Break on Dict Ordering (Pre-3.7 Mindset, But Still)**

```python
# Should contain the variable names (keys)
assert 'API_KEY' in error_message
assert 'MY-BAD-VAR' in error_message
```

You're asserting both keys appear in the message. Fine. But you're not testing the format. The error message says `[list(env_vars.keys())]` — what if someone changes the format to not include brackets? Or changes list to sorted? Fragile.

This is nitpicking, but since I'm here: tests should assert *behavior*, not incidental formatting.

---

### **[TESTING GAPS]** - This Is Not Optional

**7. You're Testing the Happy Path of Failure, Not the Edge Cases**

Your tests verify:
- Invalid var name → redacted error
- Multiple vars → redacted error
- .bashrc failure → redacted error

What about:
- **What if `_run_cmd_with_retry` raises something other than `RuntimeError`?** (e.g., `ValueError`, network error) — does that get propagated or swallowed?
- **What if `env_vars` is empty?** Does it return early before hitting any try-except? (It does, but there's no test.)
- **What if `env_vars` contains a valid var followed by an invalid one?** Does the bash export for the valid one succeed but `.bashrc` fail?

---

**8. [tests/unit/runtime/test_env_vars_security.py] - No Test for IPython/Jupyter Path**

The `add_env_vars` method has three code paths:
1. IPython (if Jupyter plugin present)
2. Windows PowerShell
3. Unix bash

You test #3 (bash export and .bashrc). You don't test #1 or #2. If someone adds secrets to IPython and that fails, are secrets leaked? Unknown — because you didn't test it.

---

## **VERDICT:**

**✅ Worth merging** — the fix is correct and addresses a real security vulnerability. But the tests are questionable mock-surgery, and the implementation has DRY violations.

---

## **KEY INSIGHT:**

The fix solves the problem, but the test strategy of rebinding real methods onto a mock is a maintenance trap waiting to snap. Either test the real thing or commit fully to mocking — this Frankenstein approach gives you neither confidence nor isolation.

jpshackelford:
Thanks for the thorough self-analysis! I've addressed all the feedback in commit 28d300ade:

## Changes Made

### ✅ DRY Violation Fixed (Issue #3)
Extracted `_run_env_cmd_redacted()` helper method that:
- Handles the try-except pattern in one place
- Takes `context`, `var_names`, and `is_windows` parameters for appropriate error messages
- Consistent error message format across all code paths

### ✅ Debug Logging Added (Issue #5)
Added debug-level logging in the helper that logs variable names and indicates error details are redacted:
```python
logger.debug(
    f'Environment variable command failed for vars {var_names}: '
    f'exit_code and error details redacted to protect secrets'
)
```
This gives operators visibility without exposing secrets.

### ✅ IPython Error Handling Added (Testing Gap #8)
Added proper error handling for the IPython/Jupyter code path - if `run_ipython()` returns an `ErrorObservation`, we now raise a sanitized `RuntimeError` instead of silently continuing.

### ✅ Tests Completely Rewritten (Critical Issue #1)
Replaced the "Frankenstein mock" approach with a proper `StubRuntime` test subclass that:
- Only overrides the external boundary methods (`run`, `run_ipython`)
- Keeps the real implementation of `add_env_vars`, `_run_cmd_with_retry`, `_run_env_cmd_redacted`
- Implements abstract methods with stubs (not used in tests)

This ensures we test **actual code paths**, not mock surgery.

### ✅ Added Test Coverage (Testing Gaps #7)
New tests added:
- `test_ipython_error_does_not_expose_secrets` - IPython path errors
- `test_empty_env_vars_returns_early` - Early return verification
- `test_non_runtime_error_propagates` - Non-RuntimeError exceptions propagate correctly
- `test_successful_env_vars_do_not_raise` - Happy path
- `test_var_names_are_uppercased_in_error` - Verifies uppercasing behavior

### ✅ Pre-existing Code Left Alone (Issue #2)
The redundant `import os; import sys` inside the method was pre-existing code. I moved `sys` to the top-level imports since we now use it at module scope, but left the rest out of scope for this security fix.

All 8 tests pass locally. Ready for review!

jpshackelford:
@openhands please address ci

nidhishgajjar:

## PR #12965: fix: redact secret values from error messages in add_env_vars

Thanks for addressing this security concern! This is an important fix.

### What I reviewed
This PR prevents secret values from being exposed in error messages when `add_env_vars` fails (e.g., due to invalid variable names containing hyphens). It introduces a `_run_env_cmd_redacted` wrapper and adds proper error handling for IPython, bash, and PowerShell paths.

### Findings

**Well done:**
- ✅ **Real security fix**: Without this, bash errors like `'bash: export: MY-SECRET=secret_value: not a valid identifier'` would expose secrets in error output
- ✅ `raise RuntimeError(...) from None` properly suppresses the exception chain that could contain secrets
- ✅ Platform-specific error hints (bash vs PowerShell identifier rules) are helpful for users
- ✅ IPython path now checks for `ErrorObservation` and raises a redacted error
- ✅ Debug logging only includes variable names, never values
- ✅ Import cleanup: `import os` moved to top of file
- ✅ Early return for empty `env_vars` dict

**Excellent test coverage:**
- ✅ Tests for bash, PowerShell (.ps1), and IPython error paths
- ✅ Tests multiple env vars scenario
- ✅ Tests `.bashrc` persistence errors are also redacted
- ✅ Tests non-RuntimeError exceptions propagate correctly
- ✅ Tests successful case doesn't raise
- ✅ Tests variable name uppercasing in errors

**One minor consideration:**
- 📝 The `from None` pattern means operators debugging production issues lose the original error details. The debug log mentions an error occurred but without specifics. This is a reasonable security trade-off, but operators should be aware they may need to check at a lower level for debugging.

### Summary
Well-crafted security fix that prevents secret values from leaking in error messages when environment variable operations fail. Thorough test coverage validates the fix across all code paths (bash, PowerShell, IPython).

### Assessment: approve

jpshackelford:
Reopening — this PR was auto-closed by the stale bot on 2026-06-06 despite carrying the `security` label and containing an actual fix (redacting secret values from `add_env_vars` error messages). The underlying class of leakage is still tracked by [`OpenHands#12959`](https://github.com/OpenHands/OpenHands/issues/12959) (also recently reopened) and the SDK-side issue [`software-agent-sdk#2152`](https://github.com/OpenHands/software-agent-sdk/issues/2152), so closing this for inactivity drops a concrete piece of in-flight remediation on the floor.

The branch (`fix/redact-secrets-from-env-var-errors` @ `87176c6`) is still present. Will follow up on rebasing/refreshing CI separately.

---
_This comment was added by an AI agent (OpenHands) on behalf of @jpshackelford._

simonrosenberg:
Closing — `openhands/runtime/base.py` was deleted from main in #14117 ("Removed V0 runtime"). This patch targets dead code that no longer exists in V1. The V1 equivalent (`add_env_vars` in the SDK/agent-server layer) doesn't have the same exposure pattern.

PR #12982, opened 2026-02-21
feat(frontend): display Agent Skills and Commands in slash menu

### LINKED ISSUE #12623: [GUI] Feature: Display Agent Skills and Commands in slash menu
We should display Agent Skills and any available `commands/` in a slash command menu in the GUI:

- When the CLI starts up, it should look look for available Agent Skills. Those Agent Skills should then be user invocable through a Slash command based on the `name` attribute in the frontmatter of that Agent Skill. 
  - For example, an Agent Skill named `random-number` should be invocable as `/random-number`
  - If the Agent Skill has frontmatter `user-invocable: false` set, then the Agent Skill should NOT appear in the slash command menu

- If the CLI is also started with a Plugin, any slash commands in that `commands/` directory should also be made available in the slash command menu.

Image

eureka0928:
## Summary of PR

Adds a slash command autocomplete menu to the chat input. When the user types `/`, a dropdown appears showing available agent skills as slash commands with a "Commands" header, bold skill names, command labels, and first-sentence descriptions.

**Filtering logic:**
- Skills with explicit `/` triggers (TaskTrigger) use those triggers directly
- AgentSkills without `/` triggers get a derived `/<skill-name>` command
- Skills with keyword-only triggers (e.g., `docker`, `container`) and repo-type skills are excluded from the menu

**Features:**
- Real-time filtering as the user types (e.g., `/in` filters to `/init`)
- Keyboard navigation (Arrow Up/Down, Enter/Tab to select, Escape to close)
- Mouse selection with mouseDown to prevent input blur
- Human-readable skill names (e.g., `code-search` → "Code search")
- First-sentence content descriptions extracted from skill content
- Full i18n support for the "Commands" header
- ARIA accessibility (`role="listbox"`, `role="option"`, `aria-selected`)

## Demo Screenshots/Videos

N/A.

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #12623

## Release Notes

Adds a slash command menu to the chat input that displays available agent skills and commands when typing `/`, with real-time filtering, keyboard navigation, and first-sentence descriptions.

image

### DIFF
--- a/enterprise/saas_server.py
+++ b/enterprise/saas_server.py
@@ -47,6 +47,7 @@ from server.routes.org_invitations import (  # noqa: E402
 from server.routes.orgs import org_router  # noqa: E402
 from server.routes.readiness import readiness_router  # noqa: E402
 from server.routes.user import saas_user_router  # noqa: E402
+from server.routes.user_app_settings import user_app_settings_router  # noqa: E402
 from server.routes.verified_models import (  # noqa: E402
     api_router as verified_models_router,
 )
@@ -79,6 +80,7 @@ base_app.include_router(api_router)  # Add additional route for github auth
 base_app.include_router(oauth_router)  # Add additional route for oauth callback
 base_app.include_router(oauth_device_router)  # Add OAuth 2.0 Device Flow routes
 base_app.include_router(saas_user_router)  # Add additional route SAAS user calls
+base_app.include_router(user_app_settings_router)  # Add routes for user app settings
 base_app.include_router(
     billing_router
 )  # Add routes for credit management and Stripe payment integration
--- a/enterprise/server/routes/user_app_settings.py
+++ b/enterprise/server/routes/user_app_settings.py
@@ -1,0 +1,115 @@
+"""Routes for user app settings API.
+
+Provides endpoints for managing user-level app preferences:
+- GET /api/users/app - Retrieve current user's app settings
+- POST /api/users/app - Update current user's app settings
+"""
+
+from fastapi import APIRouter, Depends, HTTPException, status
+from server.routes.user_app_settings_models import (
+    UserAppSettingsResponse,
+    UserAppSettingsUpdate,
+    UserNotFoundError,
+)
+from server.services.user_app_settings_service import (
+    UserAppSettingsService,
+    UserAppSettingsServiceInjector,
+)
+
+from openhands.core.logger import openhands_logger as logger
+
+user_app_settings_router = APIRouter(prefix='/api/users')
+
+# Create injector instance and dependency at module level
+_injector = UserAppSettingsServiceInjector()
+user_app_settings_service_dependency = Depends(_injector.depends)
+
+
+@user_app_settings_router.get('/app', response_model=UserAppSettingsResponse)
+async def get_user_app_settings(
+    service: UserAppSettingsService = user_app_settings_service_dependency,
+) -> UserAppSettingsResponse:
+    """Get the current user's app settings.
+
+    Returns language, analytics consent, sound notifications, and git config.
+
+    Args:
+        service: UserAppSettingsService (injected by dependency)
+
+    Returns:
+        UserAppSettingsResponse: The user's app settings
+
+    Raises:
+        HTTPException: 401 if user is not authenticated
+        HTTPException: 404 if user not found
+        HTTPException: 500 if retrieval fails
+    """
+    try:
+        return await service.get_user_app_settings()
+
+    except ValueError as e:
+        # User not authenticated
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail=str(e),
+        )
+    except UserNotFoundError as e:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=str(e),
+        )
+    except Exception as e:
+        logger.exception(
+            'Unexpected error retrieving user app settings',
+            extra={'error': str(e)},
+        )
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail='Failed to retrieve user app settings',
+        )
+
+
+@user_app_settings_router.post('/app', response_model=UserAppSettingsResponse)
+async def update_user_app_settings(
+    update_data: UserAppSettingsUpdate,
+    service: UserAppSettingsService = user_app_settings_service_dependency,
+) -> UserAppSettingsResponse:
+    """Update the current user's app settings (partial update).
+
+    Only provided fields will be updated. Pass null to clear a field.
+
+    Args:
+        update_data: Fields to update
+        service: UserAppSettingsService (injected by dependency)
+
+    Returns:
+        UserAppSettingsResponse: The updated user's app settings
+
+    Raises:
+        HTTPException: 401 if user is not authenticated
+        HTTPException: 404 if user not found
+        HTTPException: 500 if update fails
+    """
+    try:
+        return await service.update_user_app_settings(update_data)
+
+    except ValueError as e:
+        # User not authenticated
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail=str(e),
+        )
+    except UserNotFoundError as e:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=str(e),
+        )
+    except Exception as e:
+        logger.exception(
+            'Failed to update user app settings',
+            extra={'error': str(e)},
+        )
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail='Failed to update user app settings',
+        )
--- a/enterprise/server/routes/user_app_settings_models.py
+++ b/enterprise/server/routes/user_app_settings_models.py
@@ -1,0 +1,57 @@
+"""
+Pydantic models for user app settings API.
+"""
+
+from pydantic import BaseModel, EmailStr
+from storage.user import User
+
+
+class UserAppSettingsError(Exception):
+    """Base exception for user app settings errors."""
+
+    pass
+
+
+class UserNotFoundError(UserAppSettingsError):
+    """Raised when user is not found."""
+
+    def __init__(self, user_id: str):
+        self.user_id = user_id
+        super().__init__(f'User with id "{user_id}" not found')
+
+
+class UserAppSettingsUpdateError(UserAppSettingsError):
+    """Raised when user app settings update fails."""
+
+    pass
+
+
+class UserAppSettingsResponse(BaseModel):
+    """Response model for user app settings."""
+
+    language: str | None = None
+    user_consents_to_analytics: bool | None = None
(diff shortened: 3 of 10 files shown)


fury0928:
I think we need to pass description for `item.skill` for summary of every skill, currently crop the whole content to show on the command.

I can put the following PR

Looking forward to your review, @jpelletier1

hieptl:
Hello @eureka928,

Thank you very much for creating this pull request. I truly appreciate your effort and contributions.

While running the code locally, I noticed a couple of issues that I would like to share:

**Issue 1:** The descriptions for each skill do not display as expected, and some of them appear to be truncated. I understand that you have already added a comment and tagged @jpelletier1 regarding this, and we can refer to the image below for additional context.

issue1

Issue 2: After selecting a skill, I am unable to trigger a slash command again. Please refer to the video below for more details.

https://github.com/user-attachments/assets/6744dfc0-8884-4286-a92b-5cd776635097

Thank you very much! 🙏

frontend/src/components/features/chat/components/slash-command-menu.tsx
```diff
@@ -0,0 +1,88 @@
...
+      selectedItem.scrollIntoView({ block: "nearest" });
+    }
+  }, [selectedIndex]);
+
+  if (items.length === 0) return null;
+
+  return (
+    <div
+      role="listbox"
+      aria-label="Slash commands"
```
hieptl: **suggestion:** Could we please consider supporting multiple languages for this text? Thank you! 🙏

frontend/src/components/features/chat/components/slash-command-menu.tsx
```diff
@@ -0,0 +1,88 @@
...
+            index === selectedIndex ? "bg-[#383b45]" : "hover:bg-[#2a2d37]",
+          )}
+          onMouseDown={(e) => {
+            // Use mouseDown instead of click to fire before input blur
+            e.preventDefault();
+            onSelect(item);
+          }}
+        >
+          <div className="flex items-center gap-2">
+            <span className="text-sm font-semibold text-white">
```
hieptl: **suggestion:** We should consider using Typography.Text from the /frontend/src/ui folder instead of the span tag. Thank you! 🙏

frontend/src/components/features/chat/components/slash-command-menu.tsx
```diff
@@ -0,0 +1,88 @@
...
+      role="listbox"
+      aria-label="Slash commands"
+      className="absolute bottom-full left-0 w-full mb-1 bg-[#1e2028] border border-[#383b45] rounded-lg shadow-lg max-h-[300px] overflow-y-auto custom-scrollbar z-50"
+      data-testid="slash-command-menu"
+    >
+      <div className="px-3 py-2 text-xs text-[#9ca3af] border-b border-[#383b45]">
+        {t("CHAT_INTERFACE$COMMANDS")}
+      </div>
+      {items.map((item, index) => (
+        <button
```
hieptl: **suggestion:** Could we consider breaking this part down into a smaller component, such as SlashCommandMenuItem? Thank you! 🙏

enyst:
> Issue 2: After selecting a skill, I am unable to trigger a slash command again. Please refer to the video below for more details.

This is an interesting question. Should we insert two slash commands on the same line? It’s a command, it tells the agent to do a thing. 

I think maybe it makes sense to have only one, and only at the start of the user message, otherwise the agent might get confused, the handling of potential arguments would get complex, and the user wouldn’t be able to use / inside the message 🤔 WDYT?

hieptl:
> > Issue 2: After selecting a skill, I am unable to trigger a slash command again. Please refer to the video below for more details.
> 
> This is an interesting question. Should we insert two slash commands on the same line? It’s a command, it tells the agent to do a thing.
> 
> I think maybe it makes sense to have only one, and only at the start of the user message, otherwise the agent might get confused, the handling of potential arguments would get complex, and the user wouldn’t be able to use / inside the message 🤔 WDYT?

Hello @enyst,

I’m currently running the latest code from the main branch and tested triggering two skills within a single prompt. Below is the prompt I used:

> I’m attempting to trigger two skills within a single message. The first skill is /onboard, and the second is /vercel.

Based on the prompt and the attached video:
- I was able to enter the slash (/) in the middle of the message; it does not need to appear at the start of the user input.
- Both skills were successfully triggered.

https://github.com/user-attachments/assets/0a88bb68-ffcf-4c1b-9e79-520dc3a8e740

Please let me know if you need any additional details.

Thank you! 🙏

enyst:
I see, OK, thank you. I think maybe it's a bit confusing, and we probably over-trigger right now: 
- [`Vercel` trigger](https://github.com/OpenHands/extensions/tree/main/skills/vercel#triggers) doesn't have slash
- [`/onboard`](https://github.com/OpenHands/extensions/tree/main/skills/onboarding-agent#triggers) does. 🤔 

The point is that I see a case for the expected UX here to be: a slash command is a command, it feels like a command, it probably starts the message, it can have inputs (which used to open a window or the agent asks for them, that wouldn't work so well with two commands, it seems to me). But I could be wrong.

Thank you for fixing it, @eureka928 ! We'll see how it feels like, in real use?

fury0928:
> I see, OK, thank you. I think maybe it's a bit confusing, and we probably over-trigger right now:
> 
> * [`Vercel` trigger](https://github.com/OpenHands/extensions/tree/main/skills/vercel#triggers) doesn't have slash
> * [`/onboard`](https://github.com/OpenHands/extensions/tree/main/skills/onboarding-agent#triggers) does. 🤔
> 
> The point is that I see a case for the expected UX here to be: a slash command is a command, it feels like a command, it probably starts the message, it can have inputs (which used to open a window or the agent asks for them, that wouldn't work so well with two commands, it seems to me). But I could be wrong.
> 
> Thank you for fixing it, @eureka928 ! We'll see how it feels like, in real use?

Vercel doesn't work? cc: @hieptl  I think you tested that?

fury0928:
Hi @enyst @hieptl would you give me update on this?
Thank you

enyst:
I withdraw my concern @eureka928 , I’ll be happy to try it as it is now, if Hiep Le is OK with it!

fury0928:
> I withdraw my concern @eureka928 , I’ll be happy to try it as it is now, if Hiep Le is OK with it!

Hi @hieptl would you give me your opinion? cc: @jpelletier1 
If looks good, maybe good to merge?

jpelletier1:
I am not sure we need human-readable skill names. 

As a first pass, it might be better for Skills in the menu to be listed like this:
**{{slash command based on `name` attribute}}**
{{`description` attribute, with trailing dots if it's longer than 1 line}}

Example:
**/code-review**
Structured code review covering style, readability, and security concerns with actionable...

fury0928:
@jpelletier1 I updated the PR, would you review?

hieptl:
Hello @eureka928,

While testing the pull request locally, I noticed a few points that I would like to share.

**Issue 1:**
It appears that the command description is written in Markdown. As a result, it does not display optimally in the slash command menu. The video below provides additional context for this behavior.

https://github.com/user-attachments/assets/81095424-59c3-4621-83a4-a85fb7ddc37f

**Issue 2:**
At the moment, the command is only triggered when the slash (/) is entered at the beginning of the message. Could you please clarify whether there is a reason it cannot be triggered when the slash is typed in the middle of the prompt?

**Issue 3:**
If I type some text first, move the cursor back to the beginning of the message, and then enter a slash (/), the menu does not appear. This scenario is also demonstrated in the video below.

https://github.com/user-attachments/assets/8ed4730a-782e-43cd-8faf-39c7466f8492

**Issue 4:**
When resizing the chat input and then typing a slash (/), the UI alignment of the slash command menu does not look correct. Ideally, the menu position should adjust appropriately when the input is resized. Please refer to the video below for more details.

https://github.com/user-attachments/assets/3da840a1-e0bb-4d5e-ab6e-e33f0bf79fa3

Thank you very much for your time and effort. 🙏

⸻

Hello @enyst and @jpelletier1,

While testing these changes, I made the following assumptions, and I would appreciate your confirmation:

**Assumption 1:**
The slash command menu can only be triggered at the start of the message.

**Assumption 2:**
Only one slash command can be triggered within a single message.

Please feel free to let me know if these assumptions are accurate.

Thank you very much! 🙏

fury0928:
Hi @hieptl, thank you for the thorough testing and detailed feedback! 🙏

**Regarding Issues #2 and #3 (slash trigger behavior):**

These are working as intended. The slash command menu is designed to only trigger when `/` is at the very start of the message — the regex is `^\/(\S*)$`, which requires the entire input to be just the slash command.

This is a deliberate design choice to:
- Prevent accidental triggers when typing URLs, file paths, or other text containing `/`
- Keep the interaction model simple: one slash command per message, always at the start

So your assumptions are correct:
- **Assumption 1** ✅ — The slash command menu can only be triggered at the start of the message.
- **Assumption 2** ✅ — Only one slash command can be triggered within a single message.

Issue #3 (cursor-to-start) follows from the same constraint — once there's other text in the input, the full `innerText` won't match the `^/(\S*)$` pattern regardless of cursor position.

I'll look into fixing **Issue #1** (Markdown in descriptions) and **Issue #4** (menu misalignment on resize) separately.

hieptl:
Hello @eureka928,

Could you please also take a look at issue 3? Based on the requirements, it is still considered valid. While typing / at the very beginning works as expected, the slash command does not trigger if a user enters text first and then moves the cursor back to the starting position.

> once there's other text in the input, the full innerText won't match the ^/(\S*)$ pattern regardless of cursor position.

This appears to be related to the current technical approach and would need to be addressed as part of the fix. For example, if a user types a longer message and then moves back to the beginning to insert a command, they currently have to delete all existing text before the slash command will work again, which is not an ideal experience.

Thank you very much! 🙏

fury0928:
[Screencast from 2026-02-26 22-30-18.webm](https://github.com/user-attachments/assets/f79e6dc8-a46e-4580-bced-1816e6bdab16)

jpelletier1:
Video looks good to me. I haven't been able to test this branch, but video seems to implement the feature correectly.

fury0928:
> Video looks good to me. I haven't been able to test this branch, but video seems to implement the feature correectly.

Would you approve on this so that @hieptl can test with branch and merge when he is available?
Thank you for review

enyst:
Oh, no worries, eureka! For visibility for everyone: I mentioned in DM, personally I prefer to leave it to our Frontend wizards - Hiep Le, Stephan -, to approve/merge (it's a FE PR), but that's just a detail and I'm sure they'll take a look soon.

Whether or not they may want to test it manually again in this case, I don't know; that, too, is up to them.

In my reading, Joe doesn't have concerns either.

We have 201 PRs in this repo (!), though, that's a very high maintainer load.

I just think we could give some space to our FE do their magic, they know better how FE is like. Hitting some 'approved' button or not doesn't matter compared to that. 😅

jpelletier1:
Thanks @enyst :)

@eureka928 - overall the video looks like it implements the requirements, so thanks! I am not part of the OpenHands engineering team, so I don't have the authority to review the code ;-) So agree we'll leave it to FE code owners to review next!

PR #13003, opened 2026-02-23
fix: detect team/org-level budget errors in error banner

amanape:
## Summary of PR

  - Expand backend budget error detection to match both litellm proxy formats ("ExceededBudget" 
  for user-level and "Budget has been exceeded" for team/org/key-level)
  - Add `isBudgetOrCreditError` check to V1 ConversationErrorEvent handler as defense-in-depth (the
  `AgentErrorEvent` handler already had this)

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:f9b1e79-nikolaik   --name openhands-app-f9b1e79   docker.openhands.dev/openhands/openhands:f9b1e79
```


openhands/controller/agent_controller.py
```diff
@@ -343,7 +343,9 @@ async def _react_to_exception(
             elif isinstance(e, InternalServerError):
                 runtime_status = RuntimeStatus.ERROR_LLM_INTERNAL_SERVER_ERROR
                 self.state.last_error = runtime_status.value
-            elif isinstance(e, BadRequestError) and 'ExceededBudget' in str(e):
+            elif isinstance(e, BadRequestError) and (
+                'ExceededBudget' in str(e) or 'Budget has been exceeded' in str(e)
+            ):
```
hieptl: **question:** Could you please help clarify why this file needed to be modified? Is this file part of v0? Thank you! 🙏

frontend/src/contexts/conversation-websocket-context.tsx
```diff
@@ -358,7 +358,14 @@ export function ConversationWebSocketProvider({
               },
               posthog,
             });
-            setErrorMessage(event.detail);
+            if (isBudgetOrCreditError(event.detail)) {
```
hieptl, line 361: **question:** Do we need to handle the same case for handlePlanningMessage as well? Thank you! 🙏

openhands/controller/agent_controller.py
amanape: Budget errors for team-based litellm users are different than individual litellm users. Because the backend did not take into account this fact, we were no longer passing the right error to the frontend. This change takes into account team-based litellm budget errors

hieptl: However, agent_controller.py is associated with V0 conversations rather than V1. This pull request addresses the issue for V1 only. Is my understanding correct?

Thank you very much! 🙏

amanape: Oh, that completely missed my mind… then it is just the frontend change is required

amanape:
```suggestion
            elif isinstance(e, BadRequestError) and 'ExceededBudget' in str(e):
```

PR #11052, opened 2025-09-20
feat(bitbucket): supports cloud and server APIs

phymbert:
## Summary
This PR introduces a first working version of **Bitbucket Server REST API 1.0 support**, in addition to the existing Bitbucket Cloud integration.  

It allows OpenHands to work with self-hosted Bitbucket instances that expose the older `1.0` REST endpoints, expanding compatibility beyond just cloud users.  

## Motivation
Most organizations running Bitbucket use the **on-premise edition** ([Bitbucket Server / Data Center](https://www.atlassian.com/software/bitbucket)) instead of the cloud service. Supporting the [Bitbucket Server REST API 1.0](https://developer.atlassian.com/server/bitbucket/rest/) ensures OpenHands can integrate with these enterprise setups.  

## Notes
- This is an **initial implementation**; it will be updated and refined once [#11051](https://github.com/All-Hands-AI/OpenHands/pull/11051) is merged.  
- Current focus is on getting a minimal working version so that contributors using Bitbucket Server can start testing.  

## Thanks
Excited to contribute this — first version working! Big thanks to the community for the support 🙏

PR #13033, opened 2026-02-24
feat(enterprise): Bitbucket Data Center SaaS OAuth2 / Keycloak integration

jlav:
## Summary

Adds enterprise SaaS support for Bitbucket Data Center via OAuth2/Keycloak, stacked on top of the OSS integration.

**Depends on:** #13032 (OSS PR — must merge first)

> **Note for reviewers:** This PR is intentionally based on `jl/bitbucket-dc-oss` so the diff shows only enterprise-specific changes. After the OSS PR merges, update the base: `gh pr edit <this-pr-number> --base main`

### Changes (enterprise/ only)

- **`enterprise/integrations/bitbucket_data_center/bitbucket_data_center_service.py`** — `SaaSBitbucketDataCenterService` subclass that exchanges Keycloak tokens for Bitbucket DC access tokens
- **`enterprise/server/auth/constants.py`** — new OAuth2 env vars (`BITBUCKET_DC_*`) and `BITBUCKET_DC_ENABLED` feature flag
- **`enterprise/server/auth/token_manager.py`** — token refresh logic via Keycloak identity provider
- **`enterprise/server/routes/auth.py`** — OAuth2 flow fixes for Bitbucket DC provider
- **`enterprise/server/config.py`** — registers `SaaSBitbucketDataCenterService` when feature flag is set
- **`enterprise/allhands-realm-github-provider.json.tmpl`** — Keycloak identity provider config template
- **`enterprise/enterprise_local/convert_to_env.py`** — env var conversion support
- **`enterprise/tests/unit/integrations/bitbucket_data_center/`** — unit tests for `SaaSBitbucketDataCenterService`

## After OSS PR Merges

```bash
gh pr edit <this-pr-number> --base main
```

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:3d3a031-nikolaik   --name openhands-app-3d3a031   docker.openhands.dev/openhands/openhands:3d3a031
```


enyst (changes requested):
Thank you for looking into bitbucket @jlav ! Sorry I’m getting a bit confused with suddenly multiple PRs, when we already have PRs trying to fix Bitbucket integration(s).

Could you please tell how does this compare with the previous
- https://github.com/OpenHands/OpenHands/pull/11052
- https://github.com/OpenHands/OpenHands/pull/12389

I’m not familiar with BB, since you’re looking into this though, any help reviewing the old PRs would be deeply appreciated! Perhaps a synopsis updated in the description could be useful too?

To be sure, it’s understandable if so it happens that the older PRs might not do what we need!

I’m sure we could agree though that the work people put in them deserves a fair review.

I would love to make sure we avoid that unintently, that work would be wasted work, and yours duplicate work. 😥 Again, please let me know what I am missing, by all means I might, I got a bit confused with all these. 🙏

jlav:
Thanks for the comment @enyst. I apologize, I definitely should have been more communicative about how all this work relates to the existing PRs. I'm going to get the descriptions updated with a more detailed synopsis later today when I'm ready to take these out of draft, and I will make sure I tag those PRs.

But, for some quick background: I initially started with adding support for signing in with Bitbucket Data Center on the enterprise side of the integration, then started looking into how the resolvers work.

Bitbucket Data Center and Bitbucket have substantially different API contracts, which lead me on the enterprise side to break it out as an integration distinct from the existing Bitbucket cloud support. When trying to mesh that with https://github.com/OpenHands/OpenHands/pull/11052 and https://github.com/OpenHands/OpenHands/pull/12389, I ended up making the same distinction on the OSS side as well.

I'll circle back on those PRs to discuss this approach.

enyst:
> I'll circle back on those PRs to discuss this approach.

Thank you for the answers. I don't fully follow, I'm sorry. Could we review the existing PRs before making others?

Could we take them over and continue them towards what we think works?

Just for the record, that is how things normally go. We don't usually start a PR that already exists, and when we find out, we stop and consider it, review it, and maybe continue it, _closing the new ones if they happened by accident_ @jlav 

I'm just trying to say that, for example, I know Philippe has been working a lot on the Cloud PR, and I'm sure you would agree duplicate work is not great nor let work go to waste.

jlav:
@enyst I'll close my PRs and follow up with a review on #11052.

PR #13040, opened 2026-02-25
fix: support self-hosted GitLab URLs in enterprise token manager

### LINKED ISSUE #13031: [Bug]: gitlab.com is hardcoded

### Bug Description

Hello,

gitlab.com is hardcoded in the enterprise token manager preventing the usage of a self hosted gitlab server.

https://github.com/OpenHands/OpenHands/blob/3afeccfe7f8f4eafd24e02f902d186fba11df3cc/enterprise/server/auth/token_manager.py#L389

### Expected Behavior

allow custom gitlab server endpoint

### Actual Behavior

_No response_

### Steps to Reproduce

_No response_

CLI (uv tool install)

Helm chart

### OpenHands Version

latest

### Version Confirmation

### Model Name

_No response_

### Operating System

None

### Browser (if using web UI)

_No response_

### Logs and Error Messages

_No response_

_No response_

stakeswky:
## Problem

The GitLab OAuth token refresh URL in the enterprise token manager is hardcoded to `gitlab.com` (line 389 of `enterprise/server/auth/token_manager.py`):

```python
url = 'https://gitlab.com/oauth/token'
```

This prevents self-hosted GitLab instances from refreshing OAuth tokens, as reported in #13031.

## Solution

- Add a `GITLAB_BASE_DOMAIN` environment variable in `enterprise/server/auth/constants.py`, defaulting to `gitlab.com` for backward compatibility.
- Use this variable to construct the OAuth token URL dynamically in `_refresh_gitlab_token()`.

Self-hosted GitLab users can now set `GITLAB_BASE_DOMAIN=gitlab.example.com` to point token refresh at their own instance.

## Changes

- `enterprise/server/auth/constants.py`: Added `GITLAB_BASE_DOMAIN` constant
- `enterprise/server/auth/token_manager.py`: Import and use `GITLAB_BASE_DOMAIN` instead of hardcoded `gitlab.com`

Fixes #13031


stakeswky:
Thanks for the review, @malhotra5. You're right — this PR only addresses the token refresh URL, which is just one piece of the puzzle.

If Keycloak is configured to use `gitlab.com` as the identity provider, then the OAuth authorization flow (initial redirect, callback URLs, etc.) would also need to be configurable for true self-hosted support.

I'd appreciate it if you could open that issue detailing the full requirements. In the meantime, should I close this PR since the fix is incomplete, or would it be useful to keep it as a partial improvement (it does fix the token refresh for cases where Keycloak is already pointed at the self-hosted instance)? Happy to follow your guidance.

malhotra5:
Yes please feel free to close the PR, I'll work towards detailing the requirements

stakeswky:
Thanks for the review @malhotra5! Closing as requested — happy to revisit once the requirements for true self-hosted GitLab support are detailed in an issue.

L3o-pold:
I think you can edit Keycloak config to use self hosted gitlab instance, but it's not the case for the refreshing token here.

PR #13046, opened 2026-02-25
feat: add user-configurable enable/disable of default global skills w…

statxc:
## Summary of PR

Users have no way to turn off default global skills today - everything loads automatically. The `disabled_microagents` field already existed in `AgentConfig` but nothing in the UI or API ever used it.

This PR wires it up end-to-end:

- `Settings` model now stores `disabled_microagents`, persisted per user
- That list gets threaded through session startup into `Memory`, which skips disabled skills when building repo instructions, matching knowledge triggers, and loading MCP tools
- New `GET /api/skills` endpoint so the frontend knows what skills exist
- New Skills page under Settings with toggles for each skill
- Works through both web and CLI paths

27 files changed: mostly plumbing through existing layers, the usual frontend boilerplate (types, hook, route, i18n, page component), and enterprise DB schema. Tests cover Memory filtering, the API endpoint, and settings persistence

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor


statxc:
@enyst would be really thankful to give me any feedback!

statxc:
Ok. I will revert changes legacy version V0. 
I will attach screenshot

statxc:
[test.webm](https://github.com/user-attachments/assets/cf1e2496-231d-4038-aa4a-8bcc73828f6c)

@enyst could you please review the pr again? Thanks!

statxc:
@enyst any update for me?

openhands/server/routes/skills.py
```diff
@@ -0,0 +1,98 @@
+from fastapi import APIRouter
+from pydantic import BaseModel
+
+from openhands.core.logger import openhands_logger as logger
+from openhands.memory.memory import GLOBAL_MICROAGENTS_DIR, USER_MICROAGENTS_DIR
```
enyst: I think code here, including this, has a lot of legacy V0 😢

openhands/runtime/utils/bash.py
```diff
@@ -530,11 +530,14 @@ def execute(self, action: CmdRunAction) -> CmdOutputObservation | ErrorObservati
         # Check if the command is a single command or multiple commands
         splited_commands = split_bash_commands(command)
         if len(splited_commands) > 1:
+            cmd_list = '\n'.join(
+                f'({i + 1}) {cmd}' for i, cmd in enumerate(splited_commands)
+            )
```
enyst: I think this file is legacy V0, your agent just went ahead for some reason

openhands/runtime/utils/windows_bash.py
```diff
@@ -1014,11 +1014,14 @@ def execute(self, action: CmdRunAction) -> CmdOutputObservation | ErrorObservati
                 splited_cmds = [
                     str(s.Extent.Text) for s in statements
                 ]  # Try to get text
+                cmd_list = '\n'.join(
```
enyst: Same here, not in use

frontend/src/api/skills-service.ts
```diff
@@ -0,0 +1,18 @@
+import { openHands } from "./open-hands-axios";
+import { SkillInfo } from "#/types/settings";
+
+interface SkillListResponse {
+  skills: SkillInfo[];
+}
+
+class SkillsService {
+  /**
+   * Get the list of available skills (global + user microagents)
```
enyst: The fact that the LLM says "user microagents" (V0) seems to me a bit of a warning flag. Unfortunately, we have some chaos in this codebase, with the legacy version + the current version + a bunch of things on top

It's maybe not so "agent ready" as we may want it to be 😓 Sorry about that

enyst:
If I may suggest, @statxc , maybe try to focus on issues labeled good-first-issue, or please note: this functionality is discussed in multiple issues across V1 agent (software-agent-sdk repository), CLI repo, and this repo, which is actually the Web application repo. Please feel free to participate to discussions and let's figure out what we want.

On this PR, I like the video, it makes sense, maybe that's how we want it on the web! I think we should show it to people working on the web interface. Could you maybe join Slack and propose it in #proj-gui channel?

---

On another note, I'm still curious: what LLM, what agent, and what prompts you used? 😅 It's becoming the more important and fun stuff these days.

statxc:
@enyst Sorry for the inconvenience. I’ve updated it again based on your feedback. thanks a lot. 🙏

I also reviewed the legacy V0 on my hand and re-tested everything, and I confirmed it’s working properly. Could you please review and test it again?

About the LLM agent: I’m using Claude Opus 4.6, but I always review the changes and test them myself. I only submit a PR after I’ve confirmed all tests pass. 😊

statxc:
@enyst please any update for me. Hope to merge this pr if it has no problem.

enyst:
@statxc Thank you for the PR! It looks really useful to me, this is pretty much what I had in mind!

Please join our Slack and take part to the [discussion here](https://openhands-ai.slack.com/archives/C06QT0AGY4W/p1772947515783259)

jamiechicago312:
@openhands can you review this PR and give feedback to make sure this is following bet practices please?

jamiechicago312:
## Code Review: Skill Enable/Disable Feature

Thanks for this useful feature! I have reviewed the implementation and have some feedback on best practices:

### Positive Aspects

1. **Clean Architecture**: Good separation of concerns with `disabled_skills` field properly added to Settings model
2. **Proper Error Handling**: The `/api/skills` endpoint gracefully handles missing directories
3. **Test Coverage**: Good tests for API endpoint and settings persistence
4. **Frontend Patterns**: Correctly uses TanStack Query for data fetching

### Suggestions for Improvement

1. **Settings API Edge Case** (`openhands/server/routes/settings.py:200-202`)
   - The current logic treats `null` and "not provided" the same way. Consider if users should be able to explicitly clear their disabled skills list (set to empty array `[]`) vs not providing the field at all (keep existing).
   - Current: `if settings.disabled_skills is None` - this makes it impossible to clear the list via API

2. **Frontend Save Pattern** (`frontend/src/routes/skills-settings.tsx`)
   - This uses the "manual save" pattern (Save Changes button), while other similar settings like MCP servers use "immediate save" (immediate toggle). Consider which UX is more consistent with the rest of the application.

3. **Naming Consistency**
   - The backend uses `disabled_skills` while the original `AgentConfig` uses `disabled_microagents`. This is intentional for the migration, but ensure documentation reflects this naming transition.

4. **Skills Visibility**
   - The Skills settings page is nested under Settings. Consider if it deserves top-level navigation for better discoverability.

Overall the implementation looks solid and follows the existing patterns in the codebase. The main actionable feedback is the settings API edge case for clearing the list.

statxc:
> ## Code Review: Skill Enable/Disable Feature
> Thanks for this useful feature! I have reviewed the implementation and have some feedback on best practices:
> 
> ### Positive Aspects
> 1. **Clean Architecture**: Good separation of concerns with `disabled_skills` field properly added to Settings model
> 2. **Proper Error Handling**: The `/api/skills` endpoint gracefully handles missing directories
> 3. **Test Coverage**: Good tests for API endpoint and settings persistence
> 4. **Frontend Patterns**: Correctly uses TanStack Query for data fetching
> 
> ### Suggestions for Improvement
> 1. **Settings API Edge Case** (`openhands/server/routes/settings.py:200-202`)
>    
>    * The current logic treats `null` and "not provided" the same way. Consider if users should be able to explicitly clear their disabled skills list (set to empty array `[]`) vs not providing the field at all (keep existing).
>    * Current: `if settings.disabled_skills is None` - this makes it impossible to clear the list via API
> 2. **Frontend Save Pattern** (`frontend/src/routes/skills-settings.tsx`)
>    
>    * This uses the "manual save" pattern (Save Changes button), while other similar settings like MCP servers use "immediate save" (immediate toggle). Consider which UX is more consistent with the rest of the application.
> 3. **Naming Consistency**
>    
>    * The backend uses `disabled_skills` while the original `AgentConfig` uses `disabled_microagents`. This is intentional for the migration, but ensure documentation reflects this naming transition.
> 4. **Skills Visibility**
>    
>    * The Skills settings page is nested under Settings. Consider if it deserves top-level navigation for better discoverability.
> 
> Overall the implementation looks solid and follows the existing patterns in the codebase. The main actionable feedback is the settings API edge case for clearing the list.

1. Settings API edge case (clearing disabled_skills)
Sending `disabled_skills: []` already works - it's not `None`, so it passes through and clears the list. `None` just means "field not provided in the request body." This is the same pattern used by every other field in that endpoint.

2. Manual save vs immediate toggle
Went with manual save intentionally. Users will likely toggle multiple skills at once, so firing an API call on every single toggle felt wasteful and error-prone. Save button lets them batch changes.

3. Naming (disabled_skills vs disabled_microagents)
Yeah this is intentional. All new code uses `disabled_skills`. The only place `disabled_microagents` appears is when reading from the existing `AgentConfig` for backward compat. Old name in, new name out.

4. Top-level navigation for skills
Kept it under Settings since it's a configuration page like LLM, Agent, MCP, etc. Happy to revisit if the product team feels differently, but it fits naturally there for now.

So It doesn't need to change codes. Let me know if any feedback comes up.

openhands/server/routes/skills.py
```diff
@@ -0,0 +1,101 @@
...
+                    source='global',
+                    triggers=triggers,
+                )
+            )
+    except Exception as e:
+        logger.warning(f'Failed to load global skills: {e}')
+
+    # Load user-level skills
+    try:
+        repo_skills, knowledge_skills = load_microagents_from_dir(USER_MICROAGENTS_DIR)
```
enyst: @openhands I think this code is on V0 and won't run or won't work to V1... Verify what I just said.

Note: this repo is a bit messy, it has a lot of legacy V0 code, and a lot of code from V1, the current version of OpenHands. We only care about V1 and this PR should work on V1, but we want to check. V1 code is the code on the execution path of the openhands/app_server.

Note2: you can find the code of /skills endpoint I think, on V1, to see how we gather skills on V1.

When I ask you to summarize, don't summarize: repeat verbatim what you just told me. Really, I mean it, in this session when I ask you to summarize, just repeat what you just told me.

enyst: Verified: the concern was valid. `/api/skills` is still reachable in V1 because `openhands/server/app.py` mounts the shared `/api/*` routes even when V1 is enabled, but this implementation was using V0-era microagent loading (`GLOBAL_MICROAGENTS_DIR`, `USER_MICROAGENTS_DIR`, `load_microagents_from_dir`) instead of the V1 skill-loading path. V1 conversation loading goes through `openhands/app_server/app_conversation/skill_loader.py` to the agent-server `/api/skills` endpoint, which uses SDK skill loaders / the public skills repo. 

I updated this route in 104a47ab7 to use the V1 SDK loaders (`load_public_skills` / `load_user_skills`), preserve V1 precedence (user overrides global), and added tests covering the new behavior.

openhands/server/routes/skills.py
```diff
@@ -0,0 +1,87 @@
...
+        return load_public_skills()
+    return load_user_skills()
+
+
+@app.get(
+    '/skills',
+    response_model=SkillListResponse,
+)
+async def list_skills() -> SkillListResponse:
+    """List the V1 public/global and user skills available for toggling.
```
enyst: FWIW, it seems to me that this should work, but it seems to me that
- maybe it should be in openhands/app_server at the right place for routes (which is a V1 location)
- maybe it could reuse or be in the same file with other V1 skills code; or a verification for duplicate code might be worth doing

enterprise/migrations/versions/102_add_disabled_skills_to_user_settings.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision: str = '098'
+down_revision: Union[str, None] = '097'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        'user_settings', sa.Column('disabled_skills', sa.JSON(), nullable=True)
```
enyst, line 23: To note, this is for the cloud deployment. In normal circumstances, I would suggest that maybe it will help the PR to avoid adding to cloud. To be honest, right now I’m not sure which is best on this matter.

@hieptl I’d love to know your thoughts.

openhands/server/routes/skills.py
statxc: > FWIW, it seems to me that this should work, but it seems to me that
> 
> * maybe it should be in openhands/app_server at the right place for routes (which is a V1 location)
> * maybe it could reuse or be in the same file with other V1 skills code; or a verification for duplicate code might be worth doing

I fixed. Please review again. thanks

statxc:
@hieptl @enyst Please any updates for me...🙏

openhands/app_server/user/skills_router.py
```diff
@@ -0,0 +1,101 @@
+from fastapi import APIRouter
+from pydantic import BaseModel
+
+from openhands.core.logger import openhands_logger as logger
+
+# NOTE: These imports still use the legacy "microagent" naming from shared modules.
+# The underlying APIs will be renamed to "skills" as part of the broader migration.
+from openhands.memory.memory import GLOBAL_MICROAGENTS_DIR, USER_MICROAGENTS_DIR
+from openhands.microagent import load_microagents_from_dir
```
enyst: These are V0, same as below for listing skills. Maybe when your agent moved the file, it did actually restore older legacy code?

statxc:
@enyst Sorry for ping again. Please review once more. 
thanks

openhands/storage/data_models/settings.py
```diff
@@ -162,6 +163,10 @@ def from_config() -> Settings | None:
         if hasattr(app_config, 'mcp'):
             mcp_config = app_config.mcp
 
+        # Get disabled_skills from agent config if available
+        agent_config = app_config.get_agent_config()
```
enyst: I don’t know if this is in use in V1, I think it’s a V0 feature

statxc: Yes, good point. I reviewed it carefully and confirmed this is a V0 config-file mechanism: from_config() reads disabled_microagents from AgentConfig (config.toml), and that’s only used by V0’s conversation_memory.py. In V1, disabled skills are managed per user via the UI/settings store, so this code path isn’t used. I’ll remove it.

statxc:
@enyst Fixed! please review again. thanks

openhands/app_server/user/skills_router.py
```diff
@@ -0,0 +1,128 @@
...
+from openhands.core.logger import openhands_logger as logger
+from openhands.server.dependencies import get_dependencies
+
+router = APIRouter(prefix='/skills', tags=['Skills'], dependencies=get_dependencies())
+
+# V1 skill directory paths
+GLOBAL_SKILLS_DIR = Path(
+    os.path.join(os.path.dirname(os.path.dirname(openhands.__file__)), 'skills')
+)
+USER_SKILLS_DIR = Path.home() / '.openhands' / 'microagents'
```
enyst: I wonder, does this work with docker?

Maybe we could not do this, I think it doesn’t work with some use cases at least

statxc:
@enyst Updated! Testing errors don't seem to relate to my changes.

statxc:
@enyst Sorry to bother you. What else should I need to update more? I submitted this PR a month ago. I want to wrap up this PR. Thanks

enterprise/migrations/versions/102_add_disabled_skills_to_user_settings.py
```diff
@@ -0,0 +1,28 @@
+"""add disabled_skills to user_settings
```
tofarr: There have been other migrations since this - I think the number will need to be 102 to prevent collisions

openhands/app_server/user/skills_router.py
```diff
@@ -0,0 +1,125 @@
...
+        return None
+
+    end = text.find('---', 3)
+    if end == -1:
+        return None
+
+    try:
+        return yaml.safe_load(text[3:end])
+    except yaml.YAMLError:
+        return None
```
tofarr, line 56: I think it would be useful to log an error here if the yaml is invalid.

openhands/app_server/user/skills_router.py
```diff
@@ -0,0 +1,125 @@
...
+            logger.warning(f'Failed to parse skill file {md_file}: {e}')
+
+    return skills
+
+
+@router.get(
+    '',
+    response_model=SkillListResponse,
+)
+async def list_skills() -> SkillListResponse:
```
tofarr: In our API we have generally steered away from having `list` methods - all endpoints are designed with pagination in mind up front. Even if it is not likely that there will be a prohibitive number of skills it maintains consistency - the first question is always "What if there are 10,000 of them?" We can change the internal structure more easily to accommodate large numbers but once we have exposed an unpaginated API endpoint is is much harder to put that toothpaste back in the tube.

see: https://github.com/search?q=repo%3AOpenHands%2FOpenHands+search_+path%3A*_router.py&type=code

* search implies we are searching - list implies we get them all.
* search methods return a page object with a `next_page_id` and `items`

statxc:
@tofarr  Could you please review again? I updated backend following your feedback

tofarr:
Backend looks really good!
I have a few requests for the frontend code:

* Please remove the second `Skills` header
* Please move the description `Enable or disable default skills. Disabled skills will not be loaded into agent context.` outside the scroll div below the first `Skills` header.
* Please make sure that skills handle special characters correctly. e.g.: `Triggers: &#x2F;add_repo_inst global / knowledge` should be `Triggers: /add_repo_inst global / knowledge`

image

PR #13048, opened 2026-02-25
fix: prevent automatic sandbox resume on WebSocket disconnect

tofarr:
## Summary of PR

This PR fixes the issue where sandboxes were automatically resumed whenever a WebSocket disconnect was detected. The fix ensures sandboxes are only resumed when the user explicitly shows intent (by focusing on the tab or refreshing the page).

### Problem
The client was automatically resuming sandboxes via the `/resume` endpoint whenever a WebSocket disconnect was detected in `use-websocket-recovery.ts`. This was incorrect because:
- The server pauses sandboxes after 20 minutes of inactivity
- Sandboxes should only be resumed when the user explicitly shows intent

### Solution
Modified `frontend/src/hooks/use-websocket-recovery.ts`:

1. **Removed automatic resume on WebSocket disconnect**: The `handleDisconnect` callback is now intentionally empty.

2. **Added visibility change detection**: Added a `visibilitychange` event listener that resumes the sandbox when:
   - The tab gains focus (user focuses on the tab after it has lost focus)
   - AND the conversation status is "STOPPED"

3. **Page refresh handling**: Already handled by existing auto-start logic in `conversation.tsx`.

### Behavior after fix
- **WebSocket disconnect**: Does NOT automatically resume the sandbox
- **Tab gains focus**: Resumes the sandbox if it was paused/stopped
- **Page refresh**: Resumes the sandbox if it was paused/stopped

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Release Notes

This change fixes the sandbox auto-resume behavior so that sandboxes are not unnecessarily resumed when the WebSocket disconnects (e.g., due to the 20-minute inactivity timeout), but only when the user explicitly shows intent by focusing on the tab or refreshing the page.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:9391af4-nikolaik   --name openhands-app-9391af4   docker.openhands.dev/openhands/openhands:9391af4
```

### DIFF
--- a/enterprise/server/routes/auth.py
+++ b/enterprise/server/routes/auth.py
@@ -549,6 +549,10 @@ async def keycloak_offline_callback(code: str, state: str, request: Request):
         user_id=user_info['sub'], offline_token=keycloak_refresh_token
     )
 
+    logger.warning(
+        'auth_keycloak_offline_callback', extra={'base_url': request.base_url}
+    )
+
     return RedirectResponse(state if state else request.base_url, status_code=302)
 
 


tofarr:
I've created a Linear ticket for tracking: https://linear.app/all-handsai/issue/PRD-413f/sandbox-auto-resumes-on-websocket-disconnect

This ticket documents the issue and links to this PR.

PR #13051, opened 2026-02-25
fix: sanitize file_path in git diff shell commands to prevent command injection

raymyers:
## Human summary

Using Shlex quote to protect against shell injection in git diff endpoint. Test via the utility function to avoid a git subprocess in unit tests.

Thanks to Eran Shimony of Palo Alto (@ESPanda666) for raising this.

## Agent Summary

The `GET /api/conversations/{id}/git/diff` endpoint accepted a user-supplied `path` query parameter and passed it directly into a shell command string without sanitization. Because this is a GET endpoint, it was exploitable via CSRF: an attacker could embed `` on a malicious page, and any developer running OpenHands who visited that page would have the payload execute on their machine.

The injection occurred in two places:

**`git_handler.py`** (primary) — the `GIT_DIFF_CMD` template wrapped `{file_path}` in double-quotes, so a payload like `"; calc; #` produced:
```
python3 git_diff.py ""; calc; #"
```

**`git_diff.py`** (secondary) — the internal `git show` command was also manually double-quoted, making filenames with shell metacharacters exploitable.

### Fix

Apply `shlex.quote()` to `file_path` before it is embedded into any shell command string, in both files. `shlex.quote()` wraps the value in single-quotes and escapes embedded single-quotes, so the payload becomes the literal argument `'"; calc; #'` — harmless.

Also updates:
- `GIT_DIFF_CMD` template to remove the now-redundant hardcoded double-quotes (quoting is handled by `shlex.quote` at call time)
- The fallback `git_diff_cmd` template set when uploading a script to a legacy runtime
- Test `setUp` to match the updated template
- Adds `test_get_git_diff_command_injection_is_sanitized`: uses a sentinel file to assert the classic CSRF/RCE payload no longer executes shell commands

### Related endpoints checked

| Endpoint | User param fed to shell? | Verdict |
|---|---|---|
| `GET /git/diff` | ✅ `path` → **fixed** | Was vulnerable |
| `GET /git/changes` | ❌ uses runtime `cwd` only | Not exploitable |
| `GET /list-files` | ❌ file system walk, no shell | Not vulnerable |
| `GET /select-file` | ❌ file read, no shell | Not vulnerable |
| `POST /upload-files` | ❌ POST, no shell | Not vulnerable |

## Demo Screenshots/Videos

N/A — security fix with automated regression test.

## Change Type

- [x] Bug fix

## Checklist

## Fixes

Resolves the CSRF / Drive-by RCE reported against the `GET /git/diff` endpoint.

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:2bfc1ec-nikolaik   --name openhands-app-2bfc1ec   docker.openhands.dev/openhands/openhands:2bfc1ec
```

### DIFF
--- a/.agents/skills/update-sdk/SKILL.md
+++ b/.agents/skills/update-sdk/SKILL.md
@@ -1,0 +1,123 @@
+---
+name: update-sdk
+description: This skill should be used when the user asks to "update SDK", "bump SDK version", "pin SDK to a commit", "test unreleased SDK", "update agent-server image", "bump the version", "prepare a release", "what files change for a release", or needs to know how SDK packages are managed in the OpenHands repository. For detailed reference material, see references/docker-image-locations.md and references/sdk-pinning-examples.md in this skill directory.
+---
+
+# Update SDK
+
+Bump SDK packages (`openhands-sdk`, `openhands-agent-server`, `openhands-tools`), pin them to unreleased commits for testing, and cut an OpenHands release.
+
+## Quick Summary — How Many Files Change?
+
+| Activity | Manual edits | Auto-regenerated | Total |
+|----------|:------------:|:----------------:|:-----:|
+| **SDK bump** (released PyPI version) | 2 | 3 | **5** |
+| **SDK pin** (unreleased git commit) | 3 | 3 | **6** |
+| **Release commit** (version bump) | 3 | 0 | **3** |
+
+The 3 auto-regenerated files are always: `poetry.lock`, `uv.lock`, `enterprise/poetry.lock`.
+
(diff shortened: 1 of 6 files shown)


ESPanda666:
Hi, the PR is merged and I appreciate the quick response. Could you please request a CVE via the GitHub security advisory interface? This is a Critical RCE and it should be tracked so security scanners can notify users to update. I've drafted the advisory content -- happy to assist if needed.

PR #13061, opened 2026-02-26
fix(frontend): resolve 403 error after organization deletion caused by stale cache race condition

hieptl:
## Summary of PR

Fix race condition causing 403 error after organization deletion

When an organization is deleted, useAutoSelectOrganization was still running with a stale cached currentOrgId before the cache had been refreshed. As a result, the hook attempted to re-select the recently deleted organization, which triggered a 403 Forbidden error on GET /api/organizations/{deleted_org_id}.

This update addresses the issue by adjusting the order of operations and replacing invalidateQueries with removeQueries, ensuring that stale data is cleared before any state updates trigger re-renders.

Changes included:
- Fixed a race condition where useAutoSelectOrganization could use a stale currentOrgId after organization deletion
- Replaced invalidateQueries with removeQueries to immediately clear stale cache data
- Reordered operations to clear cache before triggering state update re-renders

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/c43b56a5-d6e3-4c8d-9ed7-54434b3146c9

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:cb1bd6d-nikolaik   --name openhands-app-cb1bd6d   docker.openhands.dev/openhands/openhands:cb1bd6d
```

### DIFF
--- a/frontend/src/hooks/mutation/use-delete-organization.ts
+++ b/frontend/src/hooks/mutation/use-delete-organization.ts
@@ -14,11 +14,22 @@ export const useDeleteOrganization = () => {
       return organizationService.deleteOrganization({ orgId: organizationId });
     },
     onSuccess: () => {
-      setOrganizationId(null);
-      queryClient.invalidateQueries({
+      // Remove stale cache BEFORE clearing the selected organization.
+      // This prevents useAutoSelectOrganization from using the old currentOrgId
+      // when it runs during the re-render triggered by setOrganizationId(null).
+      // Using removeQueries (not invalidateQueries) ensures stale data is gone immediately.
+      queryClient.removeQueries({
         queryKey: ["organizations"],
         exact: true,
       });
+      queryClient.removeQueries({
+        queryKey: ["organizations", organizationId],
+      });
+
+      // Now clear the selected organization - useAutoSelectOrganization will
+      // wait for fresh data since the cache is empty
+      setOrganizationId(null);
+
       navigate("/");
     },
   });


hieptl:
> The double query remove confused me at first, since we have 2 different queries using the same key (I would have thought we would use something like `listOrgs` and `getOrg`)

The query key structure follows a commonly adopted React Query convention, in which related queries share a common prefix:

```
┌──────────────────┬──────────────────────────┬────────────────────────────────┐
│       Hook       │        Query Key         │          API Endpoint          │
├──────────────────┼──────────────────────────┼────────────────────────────────┤
│ useOrganizations │ ["organizations"]        │ GET /api/organizations         │
├──────────────────┼──────────────────────────┼────────────────────────────────┤
│ useOrganization  │ ["organizations", orgId] │ GET /api/organizations/{orgId} │
└──────────────────┴──────────────────────────┴────────────────────────────────┘
```

This hierarchical naming is intentional, as it enables more flexible and targeted cache operations. For example:

```
// Invalidate only the list (exact match)
queryClient.invalidateQueries({ queryKey: ["organizations"], exact: true });

// Invalidate all organization-related queries (list and individual organizations)
queryClient.invalidateQueries({ queryKey: ["organizations"] }); // without exact: true
```

In the context of our fix, we need to ensure that both of the following are addressed:
1. The list cache is cleared to remove any stale currentOrgId.
2. The cache for the specific deleted organization is removed to prevent stale single-organization data.

You are absolutely right that using names such as ["organizationList"] and ["organization", id] would be more explicit and immediately clear. That said, the existing convention closely mirrors the REST API URL structure and aligns with a widely used React Query pattern. Adjusting this approach would require changes across multiple files and could potentially impact existing cache invalidation logic elsewhere in the codebase.

Thank you! 🙏

PR #13082, opened 2026-02-26
fix(frontend): prevent auto sandbox resume behavior

HeyItsChloe:
## Summary of PR

This PR fixes the issue where sandboxes were automatically resumed whenever a WebSocket disconnect was detected. The fix ensures sandboxes are only resumed when the user explicitly shows intent (by focusing on the tab or refreshing the page).

Expected Behavior
- WebSocket disconnect: Does NOT automatically resume the sandbox
- Tab gains focus: Resumes the sandbox if it was paused/stopped
- Page refresh: Resumes the sandbox if it was paused/stopped

Actual Behavior
- The client was automatically resuming sandboxes via the /resume endpoint whenever a WebSocket disconnect was detected in use-websocket-recovery.ts. This was incorrect because:
- The server pauses sandboxes after 20 minutes of inactivity
- Sandboxes should only be resumed when the user explicitly shows intent

Additional Context
- Context & Original fix: 
     - https://github.com/OpenHands/OpenHands/pull/13048
     - https://github.com/OpenHands/OpenHands/pull/12677/changes
- Original Bug: https://allhandsai.slack.com/archives/C07ELRWQR3P/p1769714982118359?thread_ts=1769699843.305159&cid=C07ELRWQR3P

Steps to Reproduce:
1. Start a conversation
2. Leave conversation inactive for extended period (20+ minutes)
3. Return to conversation
4. View inspector -> network tab -> see at least 1 call to `/resume`. The app auto resumed

## Demo Screenshots/Videos

After inactivity, logs show the page only resumed when the user returned to the tab
Screenshot 2026-02-26 at 10 01 22 AM

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

WebSocket disconnect: Does NOT automatically resume the sandbox


frontend/src/hooks/use-sandbox-recovery.ts
```diff
@@ -0,0 +1,127 @@
...
+  const { mutate: resumeSandbox, isPending: isResuming } =
+    useUnifiedResumeConversationSandbox();
+
+  // Track if we've already attempted recovery for this conversation on initial load
+  const hasAttemptedRecoveryRef = React.useRef<string | null>(null);
+  const isInitialLoadRef = React.useRef(true);
+
+  const attemptRecovery = React.useCallback(() => {
+    // Only recover if conversation is stopped and not already resuming
+    if (!conversationId || conversationStatus !== "STOPPED" || isResuming) {
```
hieptl: **question:**

Missing handling for other stopped states
- The hook currently checks only for conversationStatus === "STOPPED", while the accompanying comment indicates that it is intended to handle both paused and stopped sandboxes.
- It may be worth considering whether the "ERROR" status should also trigger a recovery flow, or alternatively clarifying in the documentation why this state is intentionally excluded.

Thank you! 🙏

frontend/src/hooks/use-sandbox-recovery.ts
```diff
@@ -0,0 +1,127 @@
...
+  React.useEffect(() => {
+    if (!conversationId) return undefined;
+
+    const handleVisibilityChange = () => {
+      // Only trigger when tab becomes visible (user returns to tab)
+      if (document.visibilityState === "visible") {
+        if (conversationStatus === "STOPPED") {
+          attemptRecovery();
+        }
+      }
```
hieptl: **suggestion:** Could you share the reasoning behind the use of nested if statements here? This logic could potentially be simplified to the following, which may improve readability:

```ts
if (document.visibilityState === "visible" && conversationStatus === "STOPPED") {
  attemptRecovery();
}
```

Thank you! 🙏

frontend/src/hooks/use-sandbox-recovery.ts
```diff
@@ -0,0 +1,127 @@
...
+      isInitialLoadRef.current = false;
+      hasAttemptedRecoveryRef.current = conversationId;
+
+      if (conversationStatus === "STOPPED") {
+        attemptRecovery();
+      }
+    }
+  }, [conversationId, conversationStatus, attemptRecovery]);
+
+  // Handle tab focus (visibility change)
```
hieptl: **suggestion:** The logic for checking and handling the visibilitychange event could be extracted into a separate hook, which would allow us to reuse it in other parts of the codebase in the future. Thank you! 🙏

frontend/src/hooks/use-sandbox-recovery.ts
```diff
@@ -0,0 +1,127 @@
...
+ *    and sandboxes should only be resumed when the user explicitly shows intent)
+ *
+ * @param options.conversationId - The conversation ID to recover
+ * @param options.conversationStatus - The current conversation status
+ * @param options.onSuccess - Callback when recovery succeeds
+ * @param options.onError - Callback when recovery fails
+ * @returns isResuming - Whether a recovery is in progress
+ * @returns attemptRecovery - Function to manually trigger recovery
+ */
+export function useSandboxRecovery({
```
hieptl, line 41: **question:** Could you please share a recorded video demonstrating how the conversation automatically resumes after 20–30 minutes of inactivity without requiring a page refresh? At the moment, the changes have not been deployed to a feature environment, so I’m unable to verify this behavior directly. Thank you! 🙏

frontend/src/hooks/use-sandbox-recovery.ts
HeyItsChloe: Good catch - updated the comment. I intentionally left ERROR status handling unchanged since there are no reported bugs there. I want to avoid introducing edge cases where the app errors → user refreshes → app repeatedly hits the same error in a loop.

Currently, error states are only handled via UI indicators and error badges. I think any changes to auto-recovery behavior for ERROR status should be tracked in a separate ticket. LMK what you think.

frontend/src/hooks/use-sandbox-recovery.ts
HeyItsChloe: Ah, there was no reason. Just didn't catch it. Updated.

frontend/src/hooks/use-sandbox-recovery.ts
HeyItsChloe: @hieptl Yea, the staging deploys keep timing out. 

I've been testing for a while today, but after waiting ~20+ minutes, the convo status always remains`running`. I will record a video when I am able to inspect a stopped convo again. Meanwhile, here is a screen shot showing the app only attempts refresh on tab focus. Whereas production made multple calls to resume

Local: it checks if convo has stopped (it hasn't) and will only call /resume if it needs to.
Screenshot 2026-02-27 at 1 19 20 PM

Production made multiple calls while inactive
Screenshot 2026-02-27 at 1 29 19 PM

HeyItsChloe: Tested again - you can see the timestamps after waiting show the connection closed, reopened, and only called `/resume` once.
Screenshot 2026-03-01 at 10 19 18 AM
Screenshot 2026-03-01 at 1 24 25 PM

PR #11324, opened 2025-10-10
feat: Clarifai Integration as LLM Provider

mogith-pn:
- [x] This change is worth documenting at https://docs.all-hands.dev/
- [x] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**]

Integrated [Clarifai](https://www.clarifai.com/) as LLM provider in openhands. Users can access LLM models for development using openhands with clarifai platform.
Clarifai models are `openai compatible` and also supported by `litellm` .

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

Added Clarifai as provider in OpenHands.

---
**Link of any specific issues this addresses:**

PR #11181, opened 2025-09-29
Lemonade Provider Integration

eddierichter-amd:
- [x] This change is worth documenting at https://docs.all-hands.dev/
- [x] Include this change in the Release Notes. If checked, you **must** provide an **end-user friendly** description for your change below

**End-user friendly description of the problem this fixes or functionality this introduces.**

Adding [Lemonade](https://lemonade-server.ai/) as a provider to OpenHands. This allows OpenHands to leverage local LLMs running on laptops and workstations using Lemonade. 

---
**Summarize what the PR does, explaining any non-trivial design decisions.**

Added `Lemonade` as a provider in OpenHands. 

---
**Link of any specific issues this addresses:**

PR #13094, opened 2026-02-27
feat: add Avian as LLM provider

avianion:
## Summary
- Add [Avian](https://avian.io) as an LLM provider for OpenHands
- Avian provides an OpenAI-compatible API (`https://api.avian.io/v1`) with Bearer token authentication via `AVIAN_API_KEY`
- Available models: DeepSeek V3.2 (164K context), Kimi K2.5 (131K context), GLM-5 (131K context), MiniMax M2.5 (1M context)
- Follows the same integration pattern as Clarifai (#11324) and Lemonade (#11181)

## Changes
- **`openhands/utils/llm.py`**: Add `AVIAN_MODELS` list and include in `get_supported_llm_models()`
- **`frontend/src/utils/map-provider.ts`**: Add `avian: "Avian"` provider display name
- **`frontend/src/utils/verified-models.ts`**: Add `"avian"` to `VERIFIED_PROVIDERS`
- **`frontend/__tests__/utils/map-provider.test.ts`**: Add test for Avian provider mapping
- **`config.template.toml`**: Add example Avian LLM configuration section

## Usage
```toml
[llm]
model = "avian/deepseek/deepseek-v3.2"
api_key = "<your-avian-api-key>"
base_url = "https://api.avian.io/v1"
```

## Test plan
- [ ] Verify `avian/` prefixed models appear in the model selector dropdown
- [ ] Verify Avian provider shows correct display name ("Avian") in the UI
- [ ] Verify `map-provider.test.ts` passes with new Avian assertion
- [ ] Verify chat completions work with Avian API key and base URL configured

cc @amanape @neubig @xingyaoww

### DIFF
--- a/config.template.toml
+++ b/config.template.toml
@@ -227,6 +227,13 @@ model = "gpt-4o"
 #for_routing = true
 #max_input_tokens = 128000
 
+# Example Avian LLM configuration (OpenAI-compatible API)
+# Get your API key at https://avian.io
+#[llm.avian]
+#model = "avian/deepseek/deepseek-v3.2"
+#api_key = ""
+#base_url = "https://api.avian.io/v1"
+
 
 #################################### Agent ###################################
 # Configuration for agents (group name starts with 'agent')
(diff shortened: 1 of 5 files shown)


enyst (changes requested):

I think we add to "verified models" models that are verified by us 😅 and typically well known and long-running providers. Maybe we can drop that for now?

On another note, though: is Avian added to litellm? We use litellm for provider support.

neubig:
Hi @avianion, thanks for sharing this! I think we will not be adding special support for Avian, but we will support it as an openai-compatible provider and/or through litellm.

PR #13110, opened 2026-02-28
fix(llm): Improve Bedrock support — default credential chain, bearer token, inference profiles, max_output_tokens

zxkane:
## Summary

Improve AWS Bedrock integration to work with all standard AWS credential methods — including the new Bedrock API key (bearer token) authentication — and fix model discovery/token limit issues.

### Problem

1. **Model discovery requires explicit `aws_access_key_id` + `aws_secret_access_key`** — blocks IAM roles (EC2/ECS/Lambda), SSO, `AWS_PROFILE`, STS tokens, and Bedrock API keys
2. **Cross-region inference profiles not listed** — profiles like `us.anthropic.claude-3-5-sonnet-*` are invisible in model picker
3. **Non-Anthropic Bedrock models hit `max_tokens` errors** — DeepSeek, Kimi, GLM etc. on Bedrock fail because `max_output_tokens` is `None` and gets sent as `max_completion_tokens=None`
4. **boto3 missing `AWS_DEFAULT_REGION`** — when `aws_region_name` is set in config but not in env

### Changes

| File | Change |
|------|--------|
| `openhands/llm/bedrock.py` | `_create_bedrock_client()` helper: passes explicit creds when available, otherwise uses boto3 default chain (which includes bearer token auth via `AWS_BEARER_TOKEN_BEDROCK`). Added paginated `list_inference_profiles()` (optional, non-fatal). Deduplication via `set`. |
| `openhands/utils/llm.py` | Relaxed credential gate: try listing when explicit creds exist **OR** `aws_region` set **OR** AWS env vars hint at credentials (`AWS_ACCESS_KEY_ID`, `AWS_PROFILE`, `AWS_ROLE_ARN`, `AWS_WEB_IDENTITY_TOKEN_FILE`, `AWS_BEARER_TOKEN_BEDROCK`, `AWS_DEFAULT_REGION`) |
| `openhands/llm/llm.py` | Strip cross-region prefix (`us.`, `eu.`, `apac.`, `global.`) and version suffix (`:0`, `:1`) when looking up model info. For unknown Bedrock models, **omit `max_completion_tokens`** from the request so the [Bedrock API uses the model's own maximum](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InferenceConfiguration.html) (e.g. DeepSeek-R1: 32,768, Claude: up to 128K). |
| `openhands/core/config/llm_config.py` | `os.environ.setdefault('AWS_DEFAULT_REGION', ...)` for boto3 compatibility |
| `tests/unit/llm/test_bedrock.py` | 13 new tests: credential chain, inference profiles, deduplication, API failures |
| `tests/unit/llm/test_llm.py` | 5 new tests: Bedrock max_output_tokens handling, cross-region prefix stripping, context-window-as-output-limit detection |

### How Bearer Token Auth Works

boto3 natively supports Bedrock API keys via the `AWS_BEARER_TOKEN_BEDROCK` environment variable. The Bedrock service model declares `smithy.api#httpBearerAuth` as a supported auth scheme, and botocore's `ScopedEnvTokenProvider` automatically:

1. Derives the env var name from the signing name: `AWS_BEARER_TOKEN_BEDROCK`
2. Reads the token from environment
3. Uses `BearerAuth` to add `Authorization: Bearer <token>` header

No explicit credential parameters are needed — a plain `boto3.client('bedrock')` with `AWS_BEARER_TOKEN_BEDROCK` set just works. This means `_create_bedrock_client()` requires **no code changes** to support bearer tokens; we only need to detect the env var in the gate check so model discovery triggers.

See [AWS docs: Use an Amazon Bedrock API key](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-use.html).

### Backward Compatibility

| Scenario | Before | After |
|----------|--------|-------|
| Explicit ak/sk/region in config.toml | Works | Works (same path) |
| IAM role / SSO / AWS_PROFILE | No model listing | Works |
| Bedrock API key (`AWS_BEARER_TOKEN_BEDROCK`) | No model listing | Works |
| EC2/Fargate with IAM role + `AWS_DEFAULT_REGION` | No model listing | Works |
| Bedrock DeepSeek/Kimi/GLM | `max_tokens` error | API uses model's own max |
| Cross-region inference profiles | Not listed | Listed |
| Known Bedrock models (e.g. Claude) | Works | Works (litellm resolves max_output_tokens) |
| Non-Bedrock providers | Unaffected | Unaffected |

### Why omit `max_tokens` instead of hardcoding a default?

Per [AWS Bedrock InferenceConfiguration docs](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_InferenceConfiguration.html):

> **maxTokens**: The maximum number of tokens to allow in the generated response. **The default value is the maximum allowed value for the model that you are using.**

For a coding agent, hardcoding a low value like 4096 would be too restrictive. Omitting `max_tokens` lets each model use its own maximum, which is the safest behavior for unknown models.

## Test plan

- [x] Build passes
- [x] Unit tests pass (18 Bedrock-related tests)
- [x] All pre-commit hooks pass (ruff, ruff-format, mypy)

## Checklist

- [x] New unit tests written for new functionality (18 new tests)
- [x] Backward compatible — existing explicit credential path unchanged
- [x] No new dependencies added (boto3 already a dependency)
- [x] Rebased on latest upstream/main

### DIFF
--- a/openhands/core/config/llm_config.py
+++ b/openhands/core/config/llm_config.py
@@ -203,3 +203,6 @@ class LLMConfig(BaseModel):
             )
         if self.aws_region_name:
             os.environ['AWS_REGION_NAME'] = self.aws_region_name
+            # Also set AWS_DEFAULT_REGION for boto3 compatibility — boto3 uses
+            # this env var when no explicit region is passed to the client.
+            os.environ.setdefault('AWS_DEFAULT_REGION', self.aws_region_name)
--- a/openhands/llm/bedrock.py
+++ b/openhands/llm/bedrock.py
@@ -12,34 +12,30 @@ from openhands.core.logger import openhands_logger as logger
 
 def _create_bedrock_client(
     aws_region_name: str | None = None,
-    aws_access_key_id: str | None = None,
-    aws_secret_access_key: str | None = None,
 ) -> boto3.client:
-    """Create a Bedrock client using explicit credentials or the default credential chain.
+    """Create a Bedrock client using boto3's default credential chain.
 
-    When explicit credentials (aws_access_key_id + aws_secret_access_key) are provided,
-    they are passed directly to boto3. Otherwise, boto3's default credential chain is used,
-    which supports IAM roles (EC2/ECS/Lambda), SSO, AWS_PROFILE, environment variables,
-    ~/.aws/credentials, and Bedrock API keys (AWS_BEARER_TOKEN_BEDROCK).
+    Credentials are resolved entirely by boto3's default chain, which
+    supports (in order): environment variables (AWS_ACCESS_KEY_ID, etc.),
+    AWS_PROFILE, assume role / web identity, SSO, shared credentials file,
+    container credentials (ECS/EKS), EC2 instance metadata (IAM roles),
+    and Bedrock API keys (AWS_BEARER_TOKEN_BEDROCK via bearer token auth).
+
+    Explicit ak/sk from config.toml are already exported to environment
+    variables by LLMConfig.model_post_init(), so they are picked up
+    automatically — no need to pass them here.
     """
     kwargs: dict = {'service_name': 'bedrock'}
     if aws_region_name:
         kwargs['region_name'] = aws_region_name
-    if aws_access_key_id and aws_secret_access_key:
-        kwargs['aws_access_key_id'] = aws_access_key_id
-        kwargs['aws_secret_access_key'] = aws_secret_access_key
     return boto3.client(**kwargs)
 
 
 def list_foundation_models(
     aws_region_name: str | None = None,
-    aws_access_key_id: str | None = None,
-    aws_secret_access_key: str | None = None,
 ) -> list[str]:
     try:
-        client = _create_bedrock_client(
-            aws_region_name, aws_access_key_id, aws_secret_access_key
-        )
+        client = _create_bedrock_client(aws_region_name)
         foundation_models_list = client.list_foundation_models(
             byOutputModality='TEXT', byInferenceType='ON_DEMAND'
         )
--- a/openhands/llm/bedrock.py
+++ b/openhands/llm/bedrock.py
@@ -20,7 +20,7 @@ def _create_bedrock_client(
     When explicit credentials (aws_access_key_id + aws_secret_access_key) are provided,
     they are passed directly to boto3. Otherwise, boto3's default credential chain is used,
     which supports IAM roles (EC2/ECS/Lambda), SSO, AWS_PROFILE, environment variables,
-    and ~/.aws/credentials.
+    ~/.aws/credentials, and Bedrock API keys (AWS_BEARER_TOKEN_BEDROCK).
     """
     kwargs: dict = {'service_name': 'bedrock'}
     if aws_region_name:
@@ -67,6 +67,7 @@ def list_foundation_models(
             '%s. To list Bedrock models, configure AWS credentials via one of: '
             'config.toml [llm] aws_access_key_id/aws_secret_access_key, '
             'environment variables (AWS_ACCESS_KEY_ID, AWS_PROFILE, AWS_ROLE_ARN), '
+            'Bedrock API key (AWS_BEARER_TOKEN_BEDROCK), '
             'IAM instance role, or ~/.aws/credentials.',
             err,
         )
--- a/openhands/llm/bedrock.py
+++ b/openhands/llm/bedrock.py
@@ -10,26 +10,64 @@ import boto3
 from openhands.core.logger import openhands_logger as logger
 
 
+def _create_bedrock_client(
+    aws_region_name: str | None = None,
+    aws_access_key_id: str | None = None,
+    aws_secret_access_key: str | None = None,
+) -> boto3.client:
+    """Create a Bedrock client using explicit credentials or the default credential chain.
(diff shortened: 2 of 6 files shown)


Reiss-Cashmore:
@zxkane Can this add / improve support for the Bedrock Bearer Token instead of relying on IAM / ACCESS KEYS & SECRETS? 

The bedrock UI & UX actively prioritises the bearer token method of authentication. There's no mention of using the more generalised AWS Account Method of authentication (Which may also grant access to other AWS systems, increasing exposure in a security breach scenario) in the bedrock UI or flows.

image

So it seems very strange to me to be prioritising this method of authentication. Especially considering that if you simplify to supporting the bearer token approach then it does away with any need for essentially non-sensical, poorly documented custom config.toml / env variable configuration and can just become the exact same flow as is supported in the Openhands Settings LLM UI as the user just needs to provide a single API KEY

I can see the openhands SDK supports Bedrock Bearer Tokens as shown here:
https://docs.openhands.dev/sdk/faq#setup-instructions

Does that mean it is an easy lift? If you want I can pick this up

In my opinion this kind of integration is a security incident waiting to happen. especially implicit inheritance of credentials from global/user level folders like .aws. This kind of stuff should be some kind of adapter model where this functionality is explicitly opt in, separated from the main codebase. That's just my two cents, of course this is not my project and you are free to do as you wish

zxkane:
Thanks for the feedback @Reiss-Cashmore! Great news — this PR now supports Bedrock API key (bearer token) authentication for both **inference** and **model discovery**.

**How it works today:**

LiteLLM already handles `api_key` for Bedrock inference — if you set `AWS_BEARER_TOKEN_BEDROCK` env var or pass `api_key` in config, it just works for making LLM calls.

What this PR adds is making **model discovery** (listing available models in the UI dropdown) also work with the bearer token. Under the hood, boto3 natively supports `AWS_BEARER_TOKEN_BEDROCK` — the Bedrock service model declares `smithy.api#httpBearerAuth`, and botocore's `ScopedEnvTokenProvider` automatically picks up the env var and sends `Authorization: Bearer <token>` headers. No special code needed — we just had to ensure the [credential gate check](https://github.com/OpenHands/OpenHands/pull/13110/files#diff-dce53af5dc23c1e4e6e4e8e6c577eeae1b73b1f0e81aa36377e7a72d59844e20R234-R243) detects the env var so model listing is attempted.

**To use Bedrock with just an API key:**

```bash
export AWS_BEARER_TOKEN_BEDROCK="<your-bedrock-api-key>"
export AWS_DEFAULT_REGION="us-east-1"
```

That's it — no `aws_access_key_id`, no `aws_secret_access_key`, no `config.toml` credential configuration needed. The [AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-use.html) describe how to generate a key from the Bedrock console.

**On the security concern:**

Agreed — Bedrock API keys are scoped to Bedrock/Bedrock Runtime actions only and can't access other AWS services, which is a better security posture for many use cases. This PR keeps both paths working:

| Auth method | Best for |
|---|---|
| Bedrock API key (`AWS_BEARER_TOKEN_BEDROCK`) | Dev/exploration, minimal config |
| IAM role / instance profile | Production on EC2/Fargate/Lambda |
| SSO / `AWS_PROFILE` | Developer workstations |
| Explicit ak/sk | Legacy / existing setups |

All credential resolution is now delegated to boto3's default chain, so all these methods work transparently without any special configuration in OpenHands.

nidhishgajjar:

## PR #13110: fix(llm): Improve Bedrock support — default credential chain, bearer token, inference profiles, max_output_tokens

### What I reviewed
Comprehensive improvements to AWS Bedrock LLM support: default credential chain, cross-region inference profiles, and smarter max_output_tokens handling.

### Findings

**✅ Default credential chain is the right approach**
- Replaces hardcoded `aws_access_key_id`/`aws_secret_access_key` with boto3's full credential chain (env vars, profiles, SSO, instance roles, etc.)
- `list_foundation_models` no longer requires explicit credentials — much better UX

**✅ Inference profiles listing is well-implemented**
- Gracefully handles missing `ListInferenceProfiles` permission (non-fatal)
- Deduplicates models using a set when foundation model and inference profile share an ID
- Sorted output for consistency

**✅ Smart max_output_tokens handling for Bedrock**
- Omits `max_completion_tokens` when unknown, letting Bedrock use its own default
- Detects when litellm reports `max_output_tokens == max_input_tokens` (context window size, not output limit) and resets to None
- Cross-region prefix stripping (`us.`, `eu.`, `apac.`, `global.`) to resolve model info from litellm's map
- Version suffix stripping (`:0`, `:1`) as fallback
- All of these address real gaps in litellm's Bedrock model data

**✅ Good test coverage**
- Tests for foundation models, inference profiles, deduplication, error handling
- Tests for max_output_tokens behavior with known/unknown models
- Tests for explicit user override preservation

**📝 Minor: inline import**
- `import re` is inside `init_model_info` method body rather than at the top of the file. Minor style nit — not blocking.

### Summary
Well-crafted PR that meaningfully improves Bedrock support. The credential chain, inference profiles, and max_output_tokens handling address real pain points. Good test coverage. Ready to merge.

### Assessment: approve

PR #13163, opened 2026-03-03
Add Bitbucket suggested tasks support

sharpenteeth:
## Summary
* Implement `get_suggested_tasks` for the Bitbucket integration by surfacing open issues as `SuggestedTask` entries, aligned with the existing suggested-tasks flow used by other providers.
* Limit traversal to a bounded number of repositories and issues, and handle missing data or API failures gracefully while still returning best-effort results.
* Add unit tests for the Bitbucket service that validate mapping from Bitbucket issue fields to `SuggestedTask`, and verify behavior when repositories or issue listings fail.

## Test plan
* Run unit tests in a standard OpenHands development environment (Python 3.12+, `openhands-ai` installable), including the new Bitbucket suggested-tasks tests.
* Optionally configure Bitbucket credentials and call the `/suggested-tasks` endpoint to confirm Bitbucket issues are returned as suggested tasks with the expected fields.

### DIFF
--- a/tests/unit/integrations/bitbucket/test_bitbucket_suggested_tasks.py
+++ b/tests/unit/integrations/bitbucket/test_bitbucket_suggested_tasks.py
@@ -27,7 +27,6 @@ def _make_repo(full_name: str | None) -> Repository:
     truthiness) but keeps the Repository model valid.
     """
     safe_full_name = full_name or ''
-    name = safe_full_name.split('/', 1)[1] if safe_full_name else ''
     return Repository(
         id='repo-id',
         full_name=safe_full_name,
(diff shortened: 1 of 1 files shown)


malhotra5:
Thanks for the contribution @dive2tech ! Could you please attach some screenshots or a video demonstrating that this works

dive2tech:
OK. I will do it asap.

sharpenteeth:
Hi, @malhotra5
Issue tracking no longer is supported in Bitbucket, it is responsible for Jira platform, and App password is transferred in API token, and I think API token test is implemented incorrectly. That is the reason why I haven't test my code yet.
Could change this all logic of Bitbucket?

sharpenteeth:
Hi, @malhotra5
Issue tracking no longer is supported in Bitbucket, it is responsible for Jira platform, and App password is transferred in API token, and I think API token test is implemented incorrectly. That is the reason why I haven't test my code yet.
Could change this all logic of Bitbucket?

Issue #13138, opened 2026-03-02
Fix mypy type checking for enterprise folder


malhotra5:
## Context

Type errors in the enterprise folder are not being caught by mypy. This allowed a bug where `str | None` was passed to a function expecting `str`, causing production errors.

**Example**: #13135 - A `ValidationError` occurred because `keycloak_user_id` (which can be `None`) was passed to `UserData` which requires a `str`.

## Root Cause

Enterprise code uses bare imports like `from integrations.github.github_view import GithubFactory` (without `enterprise.` prefix). Mypy cannot resolve these imports without proper path configuration.

Current behavior:
1. `mypy.ini` has `ignore_missing_imports = True`
2. Unresolved imports become `Any` type
3. `Any` is compatible with everything → type mismatches not caught

## Proposed Fix

**1. Update `enterprise/dev_config/python/mypy.ini`:**
```ini
mypy_path = .
exclude = (^migrations/.*)
```

**2. Update `enterprise/dev_config/python/.pre-commit-config.yaml`:**
```yaml
entry: mypy --config-file dev_config/python/mypy.ini .
```

## Pre-existing Type Errors

This fix exposes **796 pre-existing type errors** across 98 files that were previously hidden. These should be addressed incrementally, potentially using module-specific ignores initially.

malhotra5:
We're down to 98 errors in 38 files now

malhotra5:
We're down to 56 errors in 28 files

PR #13179, opened 2026-03-03
Fix mypy type errors in token_manager.py and auth_token_store.py

malhotra5:
## Summary of PR

Fixes mypy type errors in enterprise code:

### 1. `enterprise/storage/auth_token_store.py` (line 105)
The `load_tokens` method's callback type hint expected `Awaitable[Dict[str, str | int]]` but the actual callback returns `dict[str, str | int] | None`. Updated the type to allow `None` in the return type.

### 2. `enterprise/server/auth/token_manager.py` (line 277)
`user_id` from `user_info.get('sub')` was `Any | None` but `get_instance` expected `str`. Fixed by:
- Adding a `KeycloakUserInfo` TypedDict with `sub` as a required field (per OIDC specification)
- Updating `get_user_info()` to return `KeycloakUserInfo` instead of `dict`
- Using direct key access `user_info['sub']` since `sub` is always present per OIDC spec

### 3. `enterprise/server/routes/event_webhook.py` (lines 97, 102, 114)
`user_id` was typed as `str | None` because it's initialized as `None`, but functions expected `str`. Added an explicit check: if `user_id` is `None`, log an error and skip the operation. This is safer than an assertion since it handles potential edge cases gracefully.

## Demo Screenshots/Videos

N/A - Type-only changes

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

References #13138

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:80eb177-nikolaik   --name openhands-app-80eb177   docker.openhands.dev/openhands/openhands:80eb177
```

### DIFF
--- a/enterprise/tests/unit/conftest.py
+++ b/enterprise/tests/unit/conftest.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from uuid import UUID
 
 import pytest
+from server.auth.token_manager import KeycloakUserInfo
 from server.constants import ORG_SETTINGS_VERSION
 from server.verified_models.verified_model_service import (
     StoredVerifiedModel,  # noqa: F401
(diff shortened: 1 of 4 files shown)


enterprise/server/auth/token_manager.py
```diff
@@ -270,7 +288,7 @@ async def get_idp_token(
     ) -> str:
         # Get user info to determine user_id and idp
         user_info = await self.get_user_info(access_token=access_token)
-        user_id = user_info.get('sub')
+        user_id = user_info['sub']
```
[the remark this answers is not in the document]
malhotra5: Implement the suggestion in the other comment so this is now resolved

enterprise/server/auth/token_manager.py
```diff
@@ -49,6 +50,24 @@
 from openhands.server.types import SessionExpiredError
 from openhands.utils.http_session import httpx_verify_option
 
+
+class KeycloakUserInfo(TypedDict):
```
tofarr: Should we use a pydantic model for consistency here?

malhotra5: @openhands see if we can use pydantic models instead

then find all places where we use the new pydantic return value and update them

PR #13192, opened 2026-03-04
feat: Auto-forward LLM_* env vars to agent-server and fix host network config

saurya:
## Summary
This PR addresses multiple related issues where environment variables set on the app-server were not being forwarded to the agent-server container in the V1 two-container architecture.

### Related Issues
- Fixes OpenHands/OpenHands#12120 - LLM_TIMEOUT environment variable not being applied
- Fixes OpenHands/OpenHands#12227 - SANDBOX_USE_HOST_NETWORK bug
- Fixes OpenHands/software-agent-sdk#2418 - Connection failures due to env var propagation
- Fixes OpenHands/OpenHands#12859 - Agent-server configuration issues
- Fixes OpenHands/OpenHands#12854 - Environment variable inconsistencies

## Root Cause Analysis
The V1 architecture splits OpenHands into two containers:
1. **App-server** (`openhands-app`) - handles UI and orchestration
2. **Agent-server** (`oh-agent-server-*`) - runs the actual agent

Environment variables set on the host or in docker-compose only affect the app-server container. They were **not** automatically forwarded to the agent-server container, causing configuration mismatches.

### 1. Auto-forward LLM_* environment variables
- Modified `get_agent_server_env()` in `sandbox_spec_service.py` to automatically forward any environment variable starting with `LLM_` to the agent-server container
- This ensures `LLM_TIMEOUT`, `LLM_NUM_RETRIES`, `LLM_MODEL`, etc. work correctly
- Explicit `OH_AGENT_SERVER_ENV` JSON overrides still take precedence

### 2. Fix SANDBOX_USE_HOST_NETWORK runtime evaluation bug
- Changed from static `os.getenv()` default (evaluated at class definition time) to `default_factory` (evaluated at instance creation time)
- Added new `AGENT_SERVER_USE_HOST_NETWORK` environment variable with clearer naming
- Maintained backward compatibility with legacy `SANDBOX_USE_HOST_NETWORK`

### 3. Added comprehensive tests
- 8 new tests for LLM_* auto-forwarding functionality
- 5 new tests for host network environment variable configuration

## Testing
All new tests pass:
```
tests/unit/app_server/test_agent_server_env_override.py::TestLLMAutoForwarding - 8 passed
tests/unit/app_server/test_docker_sandbox_service.py - 7 passed (use_host_network tests)
```

### Before (required workaround)
```bash
# Users had to use the obscure OH_AGENT_SERVER_ENV JSON format
export OH_AGENT_SERVER_ENV='{"LLM_TIMEOUT": "3600", "LLM_NUM_RETRIES": "10"}'
```

### After (works intuitively)
```bash
# Now LLM_* variables "just work"
export LLM_TIMEOUT=3600
export LLM_NUM_RETRIES=10

# Host network mode now uses clearer naming and evaluates at runtime
export AGENT_SERVER_USE_HOST_NETWORK=true
```

## Documentation
A separate PR will be submitted to the docs repo (https://github.com/OpenHands/docs) to document:
1. The two-container architecture
2. Which environment variables affect which container
3. The `OH_AGENT_SERVER_ENV` mechanism for custom overrides

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:8b6f4b4-nikolaik   --name openhands-app-8b6f4b4   docker.openhands.dev/openhands/openhands:8b6f4b4
```

### DIFF
--- a/tests/unit/app_server/test_agent_server_env_override.py
+++ b/tests/unit/app_server/test_agent_server_env_override.py
@@ -586,3 +586,311 @@ class TestEnvironmentOverrideIntegration:
             # Should not have the old variables
             assert 'VAR1' not in spec_2.initial_env
             assert 'VAR2' not in spec_2.initial_env
+
+
+class TestDockerSandboxServiceEnvIntegration:
+    """Integration tests for environment variable propagation to Docker sandbox containers.
+
+    These tests verify that environment variables are correctly propagated through
+    the entire flow from the app-server environment to the agent-server container.
+    """
+
+    def test_llm_env_vars_propagated_to_container_run(self):
+        """Test that LLM_* env vars are included in docker container.run() environment argument."""
+        from unittest.mock import patch
+
+        # Set up environment with LLM_* variables
+        env_vars = {
+            'LLM_TIMEOUT': '3600',
+            'LLM_NUM_RETRIES': '10',
+            'LLM_MODEL': 'gpt-4',
+            'OTHER_VAR': 'should_not_be_forwarded',
+        }
+
+        with patch.dict(os.environ, env_vars, clear=True):
+            # Create a sandbox spec using the actual factory to get LLM_* vars
+            specs = get_default_docker_sandbox_specs()
+            sandbox_spec = specs[0]
+
+            # Verify the sandbox spec has the LLM_* variables
+            assert 'LLM_TIMEOUT' in sandbox_spec.initial_env
+            assert sandbox_spec.initial_env['LLM_TIMEOUT'] == '3600'
+            assert 'LLM_NUM_RETRIES' in sandbox_spec.initial_env
+            assert sandbox_spec.initial_env['LLM_NUM_RETRIES'] == '10'
+            assert 'LLM_MODEL' in sandbox_spec.initial_env
+            assert sandbox_spec.initial_env['LLM_MODEL'] == 'gpt-4'
+            # Non-LLM_* variables should not be included
+            assert 'OTHER_VAR' not in sandbox_spec.initial_env
+
+    def test_explicit_oh_agent_server_env_overrides_llm_vars(self):
+        """Test that OH_AGENT_SERVER_ENV can override auto-forwarded LLM_* variables."""
+        env_vars = {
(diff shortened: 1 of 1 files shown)


openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -43,6 +43,26 @@
...
 
 
+def _get_use_host_network_default() -> bool:
+    """Get the default value for use_host_network from environment variables.
+
+    This function is called at runtime (not at class definition time) to ensure
+    that environment variable changes are picked up correctly.
+
+    Supports both the new AGENT_SERVER_USE_HOST_NETWORK and legacy
+    SANDBOX_USE_HOST_NETWORK environment variables for backward compatibility.
```
enyst: I think this may help... the downside is that it's inconsistent: we are probably not going to forward the `SANDBOX_*` vars (unlike `LLM_*`, some don't work I think), so forwarding just this one is a little odd and may give the wrong impression... 🤔 

But I'm not sure. Maybe it's better with it than without it, WDYT?

saurya:
No you're right - I was thinking about this later and what we need is more
of a clear contract about how flags get passed on and the limitations of
ENV

enyst:
> No you're right - I was thinking about this later and what we need is more of a clear contract about how flags get passed on and the limitations of ENV

I think LLM_* is fine. Let’s drop legacy sandbox then, so much has changed from V0 to V1 that people need the new information anyway.

openhands/app_server/sandbox/docker_sandbox_service.py
saurya: You are right - keeping backward compatibility for just this one `SANDBOX_*` variable while not forwarding others would be confusing. Removed the backward compatibility fallback in commit 101e187f3. Now only `AGENT_SERVER_USE_HOST_NETWORK` is supported.

PR #13193, opened 2026-03-04
feat+fix(app-server): add auto_pause_existing flag and fix sandbox eviction order (#13126)

### LINKED ISSUE #13126: [Feature/Bug]: Cloud API should return rate limit error (429) instead of silently pausing existing runtimes
## Summary

When a user hits the maximum number of concurrent runtimes via the Cloud API, the current behavior automatically **pauses** older sandboxes to make room for new ones. This happens silently from the API caller's perspective - they successfully start a new conversation while their existing conversations are paused without notification.

This behavior may be surprising and problematic for API users who expect either:
1. A rate limit error (HTTP 429) when they exceed their limit, OR
2. An explicit notification that older runtimes will be paused

## Current Behavior

From `openhands/app_server/sandbox/remote_sandbox_service.py`:

```python
async def start_sandbox(
    self, sandbox_spec_id: str | None = None, sandbox_id: str | None = None
) -> SandboxInfo:
    """Start a new sandbox by creating a remote runtime."""
    try:
        # Enforce sandbox limits by cleaning up old sandboxes
        await self.pause_old_sandboxes(self.max_num_sandboxes - 1)  # <-- This pauses older sandboxes
        
        # ... then starts the new sandbox
```

The `pause_old_sandboxes` method (lines 586-633):
- Identifies sandboxes that exceed the limit
- **Pauses** (not kills) the oldest sandboxes
- Does not return any error to the caller

## Documentation vs Implementation

The [Cloud API documentation](https://docs.openhands.dev/openhands/usage/cloud/cloud-api#rate-limits) states:

> "If you have too many conversations running at once, older conversations will be paused to limit the number of concurrent conversations."

While this matches the implementation, API users may not expect this behavior without explicit notification.

## Related Context

- PR #359 in runtime-api (merged Nov 2025) updated the runtime-api to return HTTP 429 instead of HTTP 400 when the runtime limit is exceeded
- However, this only applies when the runtime-api itself enforces the limit, not when the app server proactively pauses sandboxes

### Option 1: Return 429 before pausing (Strict rate limiting)
Return HTTP 429 (Too Many Requests) when the user has reached their maximum number of runtimes, instead of automatically pausing older ones.

**Pros:**
- Standard HTTP semantics for rate limiting
- Clear feedback to API users
- Users can decide whether to explicitly stop an existing conversation

**Cons:**
- Breaking change from current behavior
- May require API users to handle rate limits explicitly

### Option 2: Include paused conversations in response (Informative approach)
When starting a new conversation that causes older ones to be paused, include this information in the response:

```json
{
  "id": "new-conversation-id",
  "status": "WORKING",
  "paused_conversations": ["old-conv-1", "old-conv-2"],
  "message": "2 existing conversations were paused to start this conversation"
}
```

**Pros:**
- Backwards compatible
- Informs users what happened
- Users can choose to resume paused conversations if needed

**Cons:**
- Non-standard response format
- May require API client updates

### Option 3: Add opt-out flag (User choice)
Add a parameter like `auto_pause_existing: bool = True` that users can set to `false` if they prefer to receive a 429 error instead of having older conversations paused.

**Pros:**
- Preserves current default behavior
- Gives users control
- Backwards compatible

**Cons:**
- Additional API complexity

## Environment

- OpenHands Cloud (app.all-hands.dev)
- Affects `/api/v1/app-conversations` endpoint
- Code location: `openhands/app_server/sandbox/remote_sandbox_service.py`

## Labels

enhancement, cloud, api

daniel-laroz:
## Summary of PR

This PR introduces an opt-out flag for the silent sandbox pausing behavior in the Cloud API (addressing Option 3 from issue OpenHands/enterprise#21), and fixes a bug in the sandbox eviction logic.

Key updates:

-Conversation API: Added auto_pause_existing to AppConversationStartRequest. This allows API users to explicitly control the pausing behavior of the underlying sandbox when starting new conversations.

-Sandbox API: Propagated the auto_pause_existing flag to start_sandbox and resume_sandbox in the remote implementation.

-429 Limit Enforcement: If the sandbox limit is reached and auto_pause_existing=False, the API now properly aborts and returns a standard HTTP 429 (MaxSandboxLimitReachedError) instead of silently pausing active sandboxes.

-Eviction Logic Bug Fix: Fixed the pause_old_sandboxes database query to sort by created_at.asc() instead of desc(). The previous implementation was mistakenly pausing the newest sandboxes rather than the oldest.

-Sandbox Service Refactoring: Modularized RemoteSandboxService by extracting sandbox lookup and pausing logic into dedicated helper methods. Introduced enforce_max_num_sandboxes_limit to handle auto_pause_existing logic.

-Boundary Fix: Updated limit validation to properly support a global limit of 1 sandbox in remote environments.

-Default Consistency: All new fields default to True to maintain strict backward compatibility with the existing GUI and CLI behavior.

-Testing: Added comprehensive unit tests for the 429 error flow, the default behavior, and the limit-of-1 boundary case.

## Demo Screenshots/Videos

(N/A - Backend logic change verified via unit tests)

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [x] Refactor

## Checklist

## Fixes

Resolves OpenHands/enterprise#21

## Implementation Notes

-Frontend: This PR focuses strictly on the app-server API. No changes were made to the frontend GUI; the default behavior remains auto_pause_existing=True (silent pausing).

-Documentation: The current Markdown documentation has not been updated in this PR. I am happy to update any manual Markdown files in a follow-up commit if pointed to the correct location.

## Release Notes

Description: Added an auto_pause_existing flag to the Cloud API. This allows API users to explicitly opt-out of automatic sandbox pausing and receive a standard HTTP 429 (Too Many Requests) error instead. Additionally, fixed a bug where the system was incorrectly evicting the newest sandboxes rather than the oldest when limits were reached. (Note: This is an API-level change and is not currently exposed in the GUI).


daniel-laroz:
@tofarr @malhotra5  @hieptl Hi! When you have a moment, could you please take a look at this PR? It addresses the auto_pause_existing flag (issue https://github.com/OpenHands/enterprise/issues/21) and includes a fix for the sandbox eviction logic, which appeared to be incorrectly pausing the newest sandboxes instead of the oldest.

All CI checks are green. Thanks!

mamoodi:
Hi @daniel-laroz sorry about the wait. Is this ready to review now?

daniel-laroz:
Hi @mamoodi, yes, it's ready for review!
I've just merged the latest changes from main and verified that all tests are passing successfully.
Thanks for checking in!

daniel-laroz:
Hi @mamoodi , just checking in on this. Let me know if you need anything else to get it reviewed and merged. Thanks!

openhands/app_server/app_conversation/app_conversation_router.py
```diff
@@ -349,7 +349,13 @@ async def start_app_conversation(
     app_conversation_service: AppConversationService = (
         app_conversation_service_dependency
     ),
+    sandbox_service: SandboxService = sandbox_service_dependency,
 ) -> AppConversationStartTask:
+    await sandbox_service.validate_sandbox_limit(
+        sandbox_id=start_request.sandbox_id,
+        auto_pause_existing=start_request.auto_pause_existing,
+    )
+
```
tofarr: There are now quite a few patch where a conversation can start in an existing sandbox (When the app_conversation requests a particular sandbox_id, or when sandbox grouping is active to reuse existing sandboxes.)

I think we can remove this check from here and keep it in the service layer - was adding this here in response to a particular issue where sandboxes were not getting stopped?

openhands/app_server/app_conversation/app_conversation_router.py
```diff
@@ -386,10 +392,16 @@ async def update_app_conversation(
...
+    sandbox_service: SandboxService = sandbox_service_dependency,
 ) -> list[AppConversationStartTask]:
     """Start an app conversation start task and stream updates from it.
     Leaves the connection open until either the conversation starts or there was an error
     """
+
+    await sandbox_service.validate_sandbox_limit(
+        sandbox_id=request.sandbox_id, auto_pause_existing=request.auto_pause_existing
+    )
+
```
tofarr: Ditto with above

openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -343,7 +343,10 @@ async def get_sandbox_by_session_api_key(
             return None
 
     async def start_sandbox(
-        self, sandbox_spec_id: str | None = None, sandbox_id: str | None = None
+        self,
+        sandbox_spec_id: str | None = None,
+        sandbox_id: str | None = None,
+        auto_pause_existing: bool = True,
```
tofarr, line 364: It looks like the behavior is not being applied to docker - let's keep it uniform across implementations.

openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -468,7 +471,9 @@ async def start_sandbox(
         except APIError as e:
             raise SandboxError(f'Failed to start container: {e}')
 
-    async def resume_sandbox(self, sandbox_id: str) -> bool:
+    async def resume_sandbox(
+        self, sandbox_id: str, auto_pause_existing: bool = True
+    ) -> bool:
```
tofarr, line 501: Ditto

openhands/app_server/sandbox/process_sandbox_service.py
```diff
@@ -287,7 +287,10 @@ async def get_sandbox_by_session_api_key(
         return None
 
     async def start_sandbox(
-        self, sandbox_spec_id: str | None = None, sandbox_id: str | None = None
+        self,
+        sandbox_spec_id: str | None = None,
+        sandbox_id: str | None = None,
+        auto_pause_existing: bool = True,
```
tofarr, line 294: Same - let's keep the implementations uniform.

openhands/app_server/sandbox/process_sandbox_service.py
```diff
@@ -342,7 +345,9 @@ async def start_sandbox(
 
         return await self._process_to_sandbox_info(sandbox_id, process_info)
 
-    async def resume_sandbox(self, sandbox_id: str) -> bool:
+    async def resume_sandbox(
+        self, sandbox_id: str, auto_pause_existing: bool = True
+    ) -> bool:
```
tofarr, line 354: Ditto

openhands/app_server/sandbox/remote_sandbox_service.py
```diff
@@ -506,14 +509,22 @@ async def start_sandbox(
 
             return self._to_sandbox_info(stored_sandbox, runtime_data)
 
+        except MaxSandboxLimitReachedError:
+            _logger.warning(
+                f'Error starting sandbox. User reached sandbox limit ({self.max_num_sandboxes}). '
+                'Request rejected because auto_pause_existing is False.'
+            )
```
tofarr: I don't think we need to catch, log and re-raise this here - I think it will be logged as an error with 'Error starting conversation' anyway

openhands/app_server/sandbox/remote_sandbox_service.py
```diff
@@ -528,6 +539,12 @@ async def resume_sandbox(self, sandbox_id: str) -> bool:
                 return False
             response.raise_for_status()
             return True
+        except MaxSandboxLimitReachedError:
+            _logger.warning(
+                f'Error resuming sandbox {sandbox_id}. User reached sandbox limit ({self.max_num_sandboxes}). '
+                'Request rejected because auto_pause_existing is False.'
+            )
+            raise
```
tofarr: Ditto

openhands/app_server/sandbox/remote_sandbox_service.py
```diff
@@ -621,6 +631,82 @@ async def pause_old_sandboxes(self, max_num_sandboxes: int) -> list[str]:
...
+        check_limit_for_resume = False
+        if sandbox_id:
+            sandbox_info = await self.get_sandbox(sandbox_id)
+            check_limit_for_resume = (not sandbox_info) or (
+                sandbox_info.status == SandboxStatus.PAUSED
+            )
+
+        if (not sandbox_id) or check_limit_for_resume:
+            if self.max_num_sandboxes < 1:
+                raise ValueError('Maximum number of sandboxes must be greater than 0')
```
tofarr: I think we should catch this with a field limiter in the pydantic model rather than here.

tofarr (changes requested):
I generally like this change - good work. Aside from my comments in the PR itself, one thing I think we need is to maintain the current behavior - so the web frontend should pass up this flag (We've found in the context of the web UI people start conversations and tend to forget about old ones - so we'd like to maintain the behavior of not making them explicitly stop old conversations.)

daniel-laroz:
Hi @tofarr,

I’ve moved the sandbox quota validation from the router into a FastAPI Dependency. The reason for the dependency is that without it, users receive an HTTP 200 response even if they had already exceeded their limit.

Regarding your point about reusing existing sandboxes: The new dependency specifically handles the resume flow. validate_sandbox_limit bypasses the check if the request is for a sandbox that is already running.

I used Pydantic to enforce that the max sandbox limit is greater than zero, which replaces the previous manual if-checks. I also moved the core management logic to the abstract class to ensure uniform behavior across all sandbox services and removed redundant try-catch blocks for cleaner code.

Additionally, I fixed a bug where the system would start sandboxes when eviction failed. The new logic iteratively attempts to free up capacity and explicitly returns a 429 error if the eviction fails.

I am not passing the flag in the frontend as I am not as familiar with that code, but the default value in the backend preserves the existing behavior and maintains backward compatibility. Thanks!

daniel-laroz:
Hi @tofarr,

Notes on testing and edge cases:

Race Conditions: During stress tests, I identified a race condition, also reproducible on the current main branch.
When sending multiple concurrent or rapid requests, the sandbox limit enforcement mechanism might not close previous sandboxes or trigger a 429 in time.
For example, on the current main branch, I managed to spin up 10 Docker sandboxes while the limit was set to 5.
I'm curious if the Remote runtime already enforces these limits on its end? Note: This PR does not resolve this underlying race condition.

Remote Runtime Testing: To validate the Remote runtime behavior, I developed a Mock Server specifically to simulate Remote sandboxes, in addition to testing on RUNTIME=docker and RUNTIME=local.

Local Runtime: I couldn't get the Local runtime to run properly on my machine (also the main branch), causing sandboxes to enter an Error state. To verify the enforcement logic, I changed the code locally so that the count also includes sandboxes in an Error state. This showed the 429 trigger working for me once the limit was reached. The code I pushed does not count sandboxes in an error state of course.

daniel-laroz:
Hi @tofarr, any thoughts on the latest updates? Ready to merge this once you're satisfied with the refactoring. Thanks!

daniel-laroz:
Hi @tofarr, just checking in on this. I've addressed the feedback and I'm keeping the branch synced with main to ensure a clean merge.

PR #13225, opened 2026-03-04
fix: preserve llm_base_url when saving MCP server config

### LINKED ISSUE #13242: [Bug]: WebUI - Adding MCP server overwrites custom llm_base_url for local hosted models

### Bug Description

Using docker to start a clean slate project. I setup the LLM and use advanced configuration to connect to LMStudio running locally. I save and settings.json file looks as expected.

I now go and add an MCP Server (context7) and once I click save, settings.json value for llm_base_url changes to "https://api.openai.com".

`docker run -it \
    --pull=always \
    --rm \
    -e OH_SECRET_KEY="dasdfasdfasdvasdjgfj" \
    -v ~/.openhands:/.openhands \
    -v $(pwd)/workspace:/opt/workspace_base \
    -p 3000:3000 \
    --add-host=host.docker.internal:host-gateway \
    --name openhands-app-v14 \
    ghcr.io/openhands/openhands:1.4.0`

### Expected Behavior

I would have expected llm_base_url to remain "http://host.docker.internal:1234/v1"

### Actual Behavior

llm_base_url changes to "https://api.openai.com"

### Steps to Reproduce

1. Run docker command above with a clean slate
2. Configure a local LLM url, click save
3. Add MCP Server, click save

CLI (Docker)

_No response_

### OpenHands Version

1.4.0, latest

### Version Confirmation

### Model Name

_No response_

### Operating System

MacOS

### Browser (if using web UI)

_No response_

### Logs and Error Messages

_No response_

_No response_

cbagwell:
## Summary of PR

Fix wiping out of LLM base URL when modifying MCP settings. This is only seen by users that require a custom base URL.

When adding/updating/deleting MCP servers, the frontend only sends mcp_config and v1_enabled to the backend. The llm_base_url field arrives as None, which was incorrectly treated the same as an empty string (explicitly cleared). This caused auto-detection to overwrite the user's custom base URL.

Fix by distinguishing None (not provided) from empty string (explicitly cleared) in store_llm_settings. When llm_base_url is None and an existing value is stored, preserve it. When explicitly cleared (''), continue with auto-detection as before.

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #13242 

## Release Notes

### DIFF
--- a/openhands/server/routes/settings.py
+++ b/openhands/server/routes/settings.py
@@ -141,9 +141,12 @@ async def store_llm_settings(
             settings.llm_api_key = existing_settings.llm_api_key
         if settings.llm_model is None:
             settings.llm_model = existing_settings.llm_model
-        # if llm_base_url is missing or empty, try to determine appropriate URL
+        # if llm_base_url is missing or empty, try to preserve existing or determine appropriate URL
         if not settings.llm_base_url:
-            if is_openhands_model(settings.llm_model):
+            if settings.llm_base_url is None and existing_settings.llm_base_url:
+                # Not provided at all (e.g. MCP config save) - preserve existing
+                settings.llm_base_url = existing_settings.llm_base_url
+            elif is_openhands_model(settings.llm_model):
                 # OpenHands models use the LiteLLM proxy
                 settings.llm_base_url = LITE_LLM_API_URL
             elif settings.llm_model:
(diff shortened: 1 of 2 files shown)


enyst:
@openhands On this PR branch, what if I had some base_url set, and then I simply delete it and save settings, will it work correctly? Make yourself a repro script and follow the same execution path

enyst (approved):
Thank you! I didn't know we are trying to infer the LLM provider in the app-server, this makes me wonder if we really should do it in the SDK 🤔 

(I was planning to, anyway, but it didn't seem a burning need)

PR #13238, opened 2026-03-05
fix: handle containers with tagless images in DockerSandboxService

cbagwell:
## Summary of PR

Fix to allow new conversations and view existing conversations when tagless container images are seen in docker sandboxes.

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

When a Docker image is rebuilt with the same tag, existing containers that used the old image have their image's tags list become empty. Previously, _container_to_sandbox_info() would crash with an IndexError when accessing container.image.tags[0] on such containers.

```
"/app/openhands/app_server/sandbox/docker_sandbox_service.py", line 203, in _container_to_sandbox_info
    sandbox_spec_id=container.image.tags[0],
                    ~~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```
Fix by returning None when image has no tags, so search_sandboxes() and get_sandbox() gracefully skip these containers instead of crashing.

Adds regression tests for both search_sandboxes and get_sandbox paths.

Resolves #(issue)

## Release Notes

### DIFF
--- a/openhands/app_server/sandbox/docker_sandbox_service.py
+++ b/openhands/app_server/sandbox/docker_sandbox_service.py
@@ -197,6 +197,12 @@ class DockerSandboxService(SandboxService):
                                     )
                                 )
 
+        if not container.image.tags:
+            _logger.debug(
+                f'Skipping container {container.name!r}: image has no tags (image id: {container.image.id})'
+            )
+            return None
+
         return SandboxInfo(
             id=container.name,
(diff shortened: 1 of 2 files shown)


tofarr (approved):
🍰 - It's not a case I came across, but I'm glad it's covered!

PR #13296, opened 2026-03-07
fix: sdk conversations not appearing in cloud ui

### LINKED ISSUE #13272: [Bug] Conversations created via OpenHandsCloudWorkspace (POST /api/v1/sandboxes) are invisible in Cloud UI
## Summary

When using the SDK's `OpenHandsCloudWorkspace` to create a sandbox via `POST /api/v1/sandboxes` and then creating a conversation on the sandbox's agent-server, the conversation does not appear in the Cloud UI at `app.all-hands.dev`. The Cloud platform has no record of it — only the raw sandbox exists.

This means users cannot:
- Monitor the agent's progress in the Cloud UI
- Share or revisit conversation URLs
- Get Cloud-side observability (metrics, cost tracking, etc.)

## Root Cause

`OpenHandsCloudWorkspace` creates sandboxes via `POST /api/v1/sandboxes`, which provisions a raw agent-server container. The SDK's `Conversation` class then creates a conversation directly on the agent-server via `POST {agent_server}/api/conversations`.

The Cloud platform only tracks "app conversations" created through `POST /api/v1/app-conversations`. Since the raw sandbox flow bypasses this endpoint, the Cloud has no record of the conversation.

## Minimal Reproduction

```python
#!/usr/bin/env python3
"""Minimal reproduction: conversations created via OpenHandsCloudWorkspace
are invisible in the Cloud UI.

Requires:
    OPENHANDS_API_KEY  - OpenHands Cloud API key
    LLM_API_KEY        - LLM API key
    LLM_BASE_URL       - (optional) LLM base URL
"""

import os
import httpx
from openhands.sdk import LLM, Agent, Conversation
from openhands.tools.preset.default import get_default_tools
from openhands.workspace import OpenHandsCloudWorkspace

CLOUD_URL = "https://app.all-hands.dev"
OPENHANDS_API_KEY = os.environ["OPENHANDS_API_KEY"]

# Step 1: Create sandbox via OpenHandsCloudWorkspace
print("Step 1: Creating sandbox via OpenHandsCloudWorkspace...")
workspace = OpenHandsCloudWorkspace(
    cloud_api_url=CLOUD_URL,
    cloud_api_key=OPENHANDS_API_KEY,
    keep_alive=True,
)
sandbox_id = workspace._sandbox_id
print(f"  Sandbox created: {sandbox_id}")
print(f"  Agent server: {workspace.host}")

# Step 2: Create a conversation on the sandbox
print("\nStep 2: Creating conversation on the sandbox's agent-server...")
llm_kwargs: dict = {
    "model": os.getenv("LLM_MODEL", "anthropic/claude-sonnet-4-5-20250929"),
    "usage_id": "repro",
    "drop_params": True,
}
if os.getenv("LLM_API_KEY"):
    llm_kwargs["api_key"] = os.environ["LLM_API_KEY"]
if os.getenv("LLM_BASE_URL"):
    llm_kwargs["base_url"] = os.environ["LLM_BASE_URL"]

llm = LLM(**llm_kwargs)
agent = Agent(
    llm=llm,
    tools=get_default_tools(enable_browser=False),
    system_prompt_kwargs={"cli_mode": True},
)
conversation = Conversation(agent=agent, workspace=workspace)
conv_id = str(conversation.state.id)
print(f"  Conversation created: {conv_id}")

# Step 3: Send a message and run
print("\nStep 3: Sending message and running agent...")
conversation.send_message("Say 'hello world' and nothing else.")
conversation.run(blocking=False)
print("  Agent triggered (non-blocking).")

# Step 4: Query Cloud API to show conversation is invisible
print("\nStep 4: Checking if conversation appears in Cloud UI...")
headers = {"Authorization": f"Bearer {OPENHANDS_API_KEY}"}

with httpx.Client(timeout=15) as client:
    resp = client.get(
        f"{CLOUD_URL}/api/v1/app-conversations/search",
        headers=headers,
        params={"limit": 10},
    )
    resp.raise_for_status()
    page = resp.json()

recent_ids = [c["id"] for c in page.get("items", [])]
print(f"  Recent app-conversations (last 10): {recent_ids}")
print(f"  Our conversation ID: {conv_id}")

found = any(conv_id in cid for cid in recent_ids)
if found:
    print("\n✅ Conversation IS visible in Cloud UI.")
else:
    print("\n❌ Conversation is NOT visible in Cloud UI.")
    print("   The sandbox exists but the conversation is only on the")
    print("   agent-server inside the sandbox — the Cloud has no record of it.")

# Verify the sandbox itself exists
with httpx.Client(timeout=15) as client:
    resp = client.get(
        f"{CLOUD_URL}/api/v1/sandboxes",
        headers=headers,
        params={"id": sandbox_id},
    )
    resp.raise_for_status()
    sandboxes = resp.json()

if sandboxes and sandboxes[0]:
    sb = sandboxes[0]
    print(f"\n  Sandbox {sandbox_id} exists (status: {sb['status']})")
else:
    print(f"\n  Sandbox {sandbox_id} not found")
```

## Actual Output

```
Step 1: Creating sandbox via OpenHandsCloudWorkspace...
  Sandbox created: 5z7uzZmgTffzrWVU554Vrz
  Agent server: https://uelvydafsiusgrqo.prod-runtime.all-hands.dev

Step 2: Creating conversation on the sandbox's agent-server...
  Conversation created: 677f76d2-f25d-4b81-9c26-dbf26d5fdce3

Step 3: Sending message and running agent...
  Agent triggered (non-blocking).

Step 4: Checking if conversation appears in Cloud UI...
  Recent app-conversations (last 10): ['68f0eb27...', 'a5a7804f...', ...]
  Our conversation ID: 677f76d2-f25d-4b81-9c26-dbf26d5fdce3

❌ Conversation is NOT visible in Cloud UI.
   The sandbox exists but the conversation is only on the
   agent-server inside the sandbox — the Cloud has no record of it.

  Sandbox 5z7uzZmgTffzrWVU554Vrz exists (status: RUNNING)
```

## Expected Behavior

Conversations created on Cloud-provisioned sandboxes should be visible in the Cloud UI, either by:

1. **Registering conversations automatically**: When a conversation is created on a Cloud-provisioned sandbox's agent-server, the Cloud should detect and register it as an app-conversation.

2. **Or providing a registration API**: Expose an endpoint like `POST /api/v1/app-conversations/register` that accepts a `sandbox_id` + `conversation_id` to retroactively register an agent-server conversation with the Cloud.

## Impact

This affects all SDK users of `OpenHandsCloudWorkspace`. Conversations run on Cloud infrastructure but are invisible, untrackable, and non-shareable.

## Related

- #13268 — Broader issue about sandbox API lacking repository context support
- [software-agent-sdk PR #2341](https://github.com/OpenHands/software-agent-sdk/pull/2341) — Where this was discovered during cloud mode implementation

hieptl:
## Summary of PR

Conversations created via the SDK's OpenHandsCloudWorkspace were not appearing in the Cloud UI because the webhook callback URL was not being properly set in the agent-server container's environment.

Root cause: The webhook URL (OH_WEBHOOKS_0_BASE_URL) was only set when WEB_HOST environment variable was configured, but this wasn't always available in Cloud deployments.

Changes:

- Derive webhook_base_url from incoming request in sandbox_router
- Pass webhook_base_url through sandbox service interface to container
- Store created_by_user_id in Docker container labels for tracking
- Add fallback to info.created_by_user_id in SAAS injector for webhook
  callbacks that use ADMIN context

### Problem

Conversations created via the SDK's `OpenHandsCloudWorkspace` were not appearing in the Cloud UI at app.all-hands.dev. The webhook mechanism that registers conversations
with Cloud was silently failing because:

1. `OH_WEBHOOKS_0_BASE_URL` was conditionally set based on `WEB_HOST` environment variable
2. When `WEB_HOST` wasn't configured, the webhook URL was never passed to agent-server containers
3. Agent-server had no webhook subscribers, so conversation notifications were never sent to Cloud

### Solution

1. **Derive webhook URL from request context** (`sandbox_router.py`): Extract the base URL from incoming HTTP requests, ensuring the webhook URL always matches how Cloud is actually reached.

2. **Propagate webhook_base_url through service interface** (`sandbox_service.py`, `remote_sandbox_service.py`, `docker_sandbox_service.py`, `process_sandbox_service.py`): Add `webhook_base_url` parameter to `start_sandbox()` method signature.

3. **Store user ownership in Docker labels** (`docker_sandbox_service.py`): Track `created_by_user_id` in container labels so webhook callbacks can attribute conversations to the correct user.

4. **Fallback for ADMIN context** (`saas_app_conversation_info_injector.py`): Use `info.created_by_user_id` when `user_context.get_user_id()` returns None (webhook callbacks use ADMIN context).

## Demo Screenshots/Videos

image

image

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #13272

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:503ce3a-nikolaik   --name openhands-app-503ce3a   docker.openhands.dev/openhands/openhands:503ce3a
```


openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -197,9 +199,13 @@ async def _container_to_sandbox_info(self, container) -> SandboxInfo | None:
...
                                 )
 
+        # Get user_id from container labels
+        labels = container.labels if isinstance(container.labels, dict) else {}
+        created_by_user_id = labels.get('created_by_user_id')
+
         return SandboxInfo(
             id=container.name,
-            created_by_user_id=None,
+            created_by_user_id=created_by_user_id,
```
enyst: I wonder, is this necessary, is docker sandbox in use in the issue this PR proposes to solve?

openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -389,10 +421,17 @@ async def start_sandbox(
                 port_mappings[exposed_port.container_port] = host_port
                 env_vars[exposed_port.name] = str(exposed_port.container_port)
 
+        # Get user_id from context for ownership tracking
+        user_id = None
+        if self.user_context:
+            user_id = await self.user_context.get_user_id()
```
enyst: I think the reason why we have `UserContext` in the first place is precisely so that we isolate the enterprise needs on `user accounts` from the open source codebase which shouldn't have any.

That doesn't look useful if what we do is getting `user_id` from it in the open source codebase, that's a bit ironic 😿 

Can we avoid it here?

openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -389,10 +421,17 @@ async def start_sandbox(
...
+
         # Prepare labels
         labels = {
             'sandbox_spec_id': sandbox_spec.id,
         }
+        if user_id:
+            labels['created_by_user_id'] = user_id
```
enyst: Ditto

openhands/app_server/sandbox/docker_sandbox_service.py
hieptl: Hello @enyst, Yes, when I run the SaaS locally and use the Docker Sandbox instead of a remote sandbox, I notice that the Docker sandbox encounters this issue. Since it occurs during local SaaS development, I would like to address and fix it.

Thank you! 🙏

enyst: Could you elaborate? The Saas deployment runs the code in these docker sandbox files?

hieptl: The production SaaS environment uses a remote sandbox. However, when I run the SaaS code locally, I often configure RUNTIME="docker". This allows me to start a conversation, view it in my local database, and observe the containers in Docker Desktop for debugging purposes.

While doing this, I noticed that when using DockerSandbox, the created_by_user_id is always set to None. As a result, if we create a new conversation from OpenHandsCloudWorkspace, the conversation information is not saved.

My intention was to add a fix for the Docker sandbox as well, so that local SaaS development would work more smoothly in the future.

However, since this change may cause confusion, I will remove it

Thank you! 🙏

enyst: Thank you. I think it's more than confusion, it's error-prone actually, and does seem to go against the patterns in the codebase that we have established.

Eventually, in V0, breaking architecture has proven error-prone for both the open source project and the Saas built on it.

That's an outcome no one wants. 😓

openhands/app_server/sandbox/docker_sandbox_service.py
hieptl: Hello @enyst,

openhands/app_server/sandbox/docker_sandbox_service.py
hieptl: Hello @enyst,

openhands/app_server/sandbox/docker_sandbox_service.py
hieptl: Hello @enyst,

enterprise/server/utils/saas_app_conversation_info_injector.py
```diff
@@ -334,7 +334,10 @@ async def save_app_conversation_info(
         await super().save_app_conversation_info(info)
 
         # Get current user_id for SAAS metadata
+        # Fall back to info.created_by_user_id for webhook callbacks (which use ADMIN context)
         user_id_str = await self.user_context.get_user_id()
+        if not user_id_str and info.created_by_user_id:
+            user_id_str = info.created_by_user_id
```
tofarr, line 340: Shouldn't all user contexts have a user_id? It strikes me that it is a serious issue in auth if they don't. (This action is being performed by... Nobody???)

tofarr: We should pull their user_id from their API key if it is not currently being done

openhands/app_server/sandbox/sandbox_router.py
```diff
@@ -56,10 +56,18 @@ async def batch_get_sandboxes(
...
 @router.post('')
 async def start_sandbox(
+    request: Request,
     sandbox_spec_id: str | None = None,
     sandbox_service: SandboxService = sandbox_service_dependency,
 ) -> SandboxInfo:
-    info = await sandbox_service.start_sandbox(sandbox_spec_id)
+    # Derive webhook base URL from the incoming request so the agent-server
+    # can notify this Cloud instance about conversation lifecycle events
+    webhook_base_url = str(request.base_url).rstrip('/')
```
tofarr: This wont work in cases where a local openhands install is running against a remote sandbox.

tofarr: That is the reason why we didn't do this previously - if the request.base_url is `http://localhost`, then there is no way for a webhook to be set up for this.

tofarr: I suggest we use the config.web_url instead if it is defined and it is not localhost

PR #13300, opened 2026-03-08
Fix variable shadowing of start_id in StateTracker._init_history

jnMetaCode:
## Summary

- Fix a variable shadowing bug in `StateTracker._init_history()` where the `for start_id, end_id in sorted(delegate_ranges):` loop overwrites the `start_id` variable that was set earlier from `self.state.start_id`
- After the loop completes, `self.state.start_id = start_id` incorrectly sets the state's start_id to the **last delegate range's start ID** instead of the original start position
- This can cause the agent to lose history events preceding the last delegate action when restoring state in multi-agent (delegate) scenarios

## Details

In `_init_history()`, `start_id` is initialized at line 124:
```python
start_id = self.state.start_id if self.state.start_id >= 0 else 0
```

Then at line 178, the same variable name is reused as a loop variable:
```python
for start_id, end_id in sorted(delegate_ranges):
```

After the loop, at line 202:
```python
self.state.start_id = start_id  # now contains the last delegate's start ID!
```

The fix saves the original `start_id` before the loop and renames the loop variables to `delegate_start` and `delegate_end` to prevent shadowing.

## Test plan

- [ ] Verify existing tests pass (the fix is a safe rename of loop variables + preserving the original value)
- [ ] Test multi-agent delegate scenarios to confirm history is properly preserved across session restores

### DIFF
--- a/openhands/controller/state/state_tracker.py
+++ b/openhands/controller/state/state_tracker.py
@@ -170,25 +170,33 @@ class StateTracker:
                 action_id = delegate_action_ids.pop()
                 delegate_ranges.append((action_id, event.id))
 
+        # Save original start_id before delegate filtering loop
+        original_start_id = start_id
+
         # Filter out events between delegate action/observation pairs
         if delegate_ranges:
             filtered_events: list[Event] = []
             current_idx = 0
 
-            for start_id, end_id in sorted(delegate_ranges):
+            for delegate_start, delegate_end in sorted(delegate_ranges):
                 # Add events before delegate range
                 filtered_events.extend(
-                    event for event in events[current_idx:] if event.id < start_id
+                    event
+                    for event in events[current_idx:]
+                    if event.id < delegate_start
                 )
 
                 # Add delegate action and observation
                 filtered_events.extend(
-                    event for event in events if event.id in (start_id, end_id)
+                    event
+                    for event in events
+                    if event.id in (delegate_start, delegate_end)
                 )
 
                 # Update index to after delegate range
                 current_idx = next(
-                    (i for i, e in enumerate(events) if e.id > end_id), len(events)
+                    (i for i, e in enumerate(events) if e.id > delegate_end),
+                    len(events),
                 )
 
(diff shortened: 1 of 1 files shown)


alvinttang (commented):
The fix is correct — the loop variable `start_id` was silently clobbering the outer `start_id` computed earlier, so `state.start_id` ended up holding the last delegate range's start rather than the event-stream anchor. Renaming the loop variables to `delegate_start`/`delegate_end` is the right approach and the extra `original_start_id` save makes the intent explicit. One minor note: the filtering logic iterates `events[current_idx:]` for the pre-range events but scans all of `events` for the delegate action/observation lookup, making that inner search O(n) per range — fine for typical histories, but worth a comment if delegate ranges can grow large. Overall a clean, targeted fix.

jnMetaCode:
Thanks for the thorough review @alvinttang! Good point about the filtering logic — the pre-range scan uses `events[current_idx:]` while the delegate scan checks all of `events`. I kept it that way since the delegate ranges may reference events outside the current slice, but happy to tighten it if you think it's worth a follow-up.

nidhishgajjar:

## PR #13300: Fix variable shadowing of start_id in StateTracker._init_history

### What I reviewed
Fixes a variable shadowing bug where `start_id` was reused as a loop variable in `for start_id, end_id in sorted(delegate_ranges)`, overwriting the original `start_id` value needed later.

### Findings

**✅ Correct fix for a real bug**
- The loop variable `start_id` shadowed the method's `start_id` parameter, causing `self.state.start_id = start_id` at the end to set it to the last delegate range's start ID instead of the original conversation start ID
- This could corrupt event history replay by pointing `state.start_id` at a delegate action rather than the actual conversation beginning
- The fix renames loop variables to `delegate_start, delegate_end` and preserves the original value in `original_start_id`

**No concerns.** Clean, focused fix with clear variable naming.

### Summary
Fixes a classic variable-shadowing Python bug that could corrupt state history. Minimal and correct.

### Assessment: approve

PR #13303, opened 2026-03-08
fix: reject API endpoint paths in base_url with clear error message

### LINKED ISSUE #13079: [Bug]: base_url accepts API endpoint paths, causing cryptic NotFoundError with no useful info

### Bug Description

If a user configures `base_url` with a full API endpoint path (e.g. `https://my-proxy.com/v1/chat/completions`), OpenHands silently accepts it and later fails at runtime with:

```
litellm.NotFoundError: AnthropicException - {"error":"Not found"}
```

There is no indication that the base URL is misconfigured, no mention of what URL was actually used, and no guidance on how to fix it.

### Expected Behavior

The input should be rejected with a message explaining that `base_url` must not include the endpoint path, e.g.:

```
base_url must not include the API endpoint path /v1/chat/completions. Use only the base URL, e.g. https://my-proxy.com/v1.
```

### Actual Behavior

The value is accepted silently. At runtime, litellm constructs a broken URL and the provider returns a 404, which surfaces as `AnthropicException - {"error":"Not found"}` with no reference to the URL tried or any suggestion to check Settings.

### Steps to Reproduce
  1. Run make `setup-config` and enter `https://my-proxy.com/v1/chat/completions` as the LLM base URL, there is no error.
  2. Or set `base_url = "https://my-proxy.com/v1/chat/completions"` in `config.toml`.
  3. Or enter the same value in the web UI Settings and save, it is accepted with HTTP 200.
  4. Submit any task.

### OpenHands Version

OpenHands CLI 1.12.2

### Version Confirmation

### Model Name

n/a

### Operating System

MacOS

### Browser (if using web UI)

n/a

### Logs and Error Messages

`AnthropicException - {"error":"Not found"}`

MaxwellCalkin:
## Summary

Fixes #13079.

When a user configures `base_url` with a full API endpoint path (e.g. `https://my-proxy.com/v1/chat/completions`), litellm silently constructs a broken URL and the provider returns a cryptic `NotFoundError` with no indication that the base URL is misconfigured.

This PR adds `field_validator`s to both `LLMConfig.base_url` and `Settings.llm_base_url` that detect common API endpoint suffixes and raise a clear `ValueError` explaining that only the base URL should be provided.

**Detected endpoint suffixes:**
- `/chat/completions`
- `/completions`
- `/messages`
- `/embeddings`
- `/models`
- `/generations`

**Error message example:**
```
base_url must not include the API endpoint path /v1/chat/completions.
Use only the base URL (e.g. https://my-proxy.com/v1).
```

**Entry points covered:**
- `LLMConfig` — catches misconfiguration via `config.toml` and CLI (`make setup-config`)
- `Settings` — catches misconfiguration via web UI settings (POST /api/settings) and V1 user context
- Invalid custom LLM sections in TOML are gracefully skipped (existing `from_toml_section` behavior)

## Test plan

- [x] Added parametrized tests for `LLMConfig.base_url` validation (8 rejected URLs, 7 accepted URLs)
- [x] Added parametrized tests for `Settings.llm_base_url` validation (8 rejected URLs, 7 accepted URLs)
- [x] Added integration test verifying invalid TOML custom LLM sections are skipped gracefully
- [x] Verified trailing slashes are handled (e.g. `.../completions/`)
- [x] Verified `None` base_url passes validation (default case)

**Disclosure: This PR was authored by Claude Opus 4.6 (an AI). I am pursuing employment as an AI contributor — transparently, not impersonating a human. See https://github.com/MaxwellCalkin/career for context.**

### DIFF
--- a/openhands/core/config/llm_config.py
+++ b/openhands/core/config/llm_config.py
@@ -9,8 +9,9 @@ from __future__ import annotations
 
 import os
 from typing import Any
+from urllib.parse import urlparse
 
-from pydantic import BaseModel, ConfigDict, Field, SecretStr, ValidationError
+from pydantic import BaseModel, ConfigDict, Field, SecretStr, ValidationError, field_validator
 
 from openhands.core.logger import LOG_DIR
 from openhands.core.logger import openhands_logger as logger
@@ -110,6 +111,35 @@ class LLMConfig(BaseModel):
     )
 
     model_config = ConfigDict(extra='forbid')
+
+    @field_validator('base_url')
+    @classmethod
+    def validate_base_url(cls, v: str | None) -> str | None:
+        if v is None:
+            return v
+        # Common API endpoint path suffixes that should not be included in
+        # the base URL.  When present they cause litellm to construct a
+        # broken URL, resulting in a cryptic 404 from the provider.
+        endpoint_suffixes = (
+            '/chat/completions',
+            '/completions',
+            '/messages',
+            '/embeddings',
+            '/models',
+            '/generations',
+        )
+        try:
+            path = urlparse(v).path.rstrip('/')
+        except Exception:
+            return v
+        for suffix in endpoint_suffixes:
+            if path.endswith(suffix):
+                raise ValueError(
+                    f'base_url must not include the API endpoint path '
+                    f'{suffix}. Use only the base URL '
+                    f'(e.g. https://my-proxy.com/v1).'
+                )
+        return v
 
     @classmethod
     def from_toml_section(cls, data: dict) -> dict[str, LLMConfig]:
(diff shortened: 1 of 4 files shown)


alvinttang (commented):
Good defensive fix. One edge case the suffix list misses: a URL like `https://proxy.com/v1/chat/completions/custom-route` would be silently accepted since it doesn't end with a known suffix — though this is likely an edge case in practice. A bigger concern is that the `endpoint_suffixes` tuple is duplicated verbatim between `LLMConfig` and `Settings`; extracting it to a shared module constant would prevent the two lists from drifting out of sync. The error message is user-friendly and includes a concrete example, which is a nice touch.

mamoodi:
Hi sorry for the wait. Is this good to get someone to review? One of the jobs have failed.

PR #13330, opened 2026-03-10
fix(backend): inherit organization llm settings for new members

hieptl:
## Summary of PR

Two related bugs were preventing new organization members from receiving the correct LLM configuration:

1. When accepting an invitation, the code explicitly removed llm_model and llm_base_url from kwargs and never fetched the organization's settings to pass to the new member.

2. When admins saved LLM settings via POST /api/settings, the org's default_llm_model/base_url/max_iterations fields were never updated because the store() method checked hasattr(org, 'llm_model') which returned False (Org uses 'default_llm_model', not 'llm_model').

Changes:

- Add optional llm_model, llm_base_url, max_iterations params to OrgMemberStore.add_user_to_org()
- Fetch org settings in accept_invitation() and pass them to new member
- Explicitly update org's default\_\* fields in SaasSettingsStore.store()
- Add tests for both fixes and update existing tests with new mock

### Issue 1: New members get default LLM instead of org's configured model

**Location:** `enterprise/server/services/org_invitation_service.py`

When a new member accepted an invitation:

1. The code explicitly removed `llm_model` and `llm_base_url` from kwargs (lines 373-374)
2. The org's current LLM settings were never fetched
3. `add_user_to_org` was called without LLM settings, so members got NULL values

**Fix:** Fetch the organization's settings and pass `default_llm_model`, `default_llm_base_url`, and `default_max_iterations` when creating the new member.

### Issue 2: Org's default LLM fields never updated when admins save settings

**Location:** `enterprise/storage/saas_settings_store.py`

When admins saved settings via `POST /api/settings`:

1. The `store()` method iterated over models checking `hasattr(model, key)`
2. For `org`, `hasattr(org, 'llm_model')` returned `False` because Org uses `default_llm_model`
3. The org's `default_*` fields were never updated

**Fix:** Explicitly map Settings fields to Org's `default_*` fields after the generic loop.

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/4b4a558e-d9d6-43a4-a5e1-e713877ef685

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:bae8957-nikolaik   --name openhands-app-bae8957   docker.openhands.dev/openhands/openhands:bae8957
```

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -190,12 +190,14 @@ class SaasSettingsStore(SettingsStore):
             # Map Settings fields to Org fields with 'default_' prefix
             # The generic loop above doesn't update these because Org uses
             # 'default_llm_model' not 'llm_model', etc.
-            if item.llm_model is not None:
-                org.default_llm_model = item.llm_model
-            if item.llm_base_url is not None:
-                org.default_llm_base_url = item.llm_base_url
-            if item.max_iterations is not None:
-                org.default_max_iterations = item.max_iterations
+            # Use exclude_unset to only update explicitly-set fields (allows clearing with null)
+            settings_data = item.model_dump(exclude_unset=True)
+            if 'llm_model' in settings_data:
+                org.default_llm_model = settings_data['llm_model']
+            if 'llm_base_url' in settings_data:
+                org.default_llm_base_url = settings_data['llm_base_url']
+            if 'max_iterations' in settings_data:
+                org.default_max_iterations = settings_data['max_iterations']
 
             # Propagate LLM settings to all org members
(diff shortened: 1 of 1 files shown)


tofarr (approved):
🍰 - Good to fill that functionality gap

PR #13344, opened 2026-03-11
fix(event): apply limit and page_id to items in search_events

### LINKED ISSUE #13335: Bug: EventServiceBase.search_events() loads all events before applying limit/page_id
## Bug description

`EventServiceBase.search_events()` appears to load **all** events for a conversation before applying `page_id` / `limit`.

This makes calls like:

```python
await event_service.search_events(
    conversation_id=conversation_id,
    kind__eq='MessageEvent',
    sort_order=EventSortOrder.TIMESTAMP_DESC,
    limit=50,
)
```

look cheap, but they may still scan and deserialize the entire conversation history.

## Why this is a bug

In `openhands/app_server/event/event_service_base.py`, the current flow is:

1. list all event paths
2. load all events with `asyncio.gather(...)`
3. filter/sort all loaded events into `items`
4. only then apply `page_id` / `limit` logic to `paths`
5. return `EventPage(items=items, next_page_id=next_page_id)`

Relevant implementation:

```python
paths = await loop.run_in_executor(None, self._search_paths, prefix)
events = await asyncio.gather(
    *[loop.run_in_executor(None, self._load_event, path) for path in paths]
)
items = []
for event in events:
    ...
    items.append(event)

if page_id:
    start_offset = int(page_id)
    paths = paths[start_offset:]
if len(paths) > limit:
    paths = paths[:limit]
    next_page_id = str(start_offset + limit)

return EventPage(items=items, next_page_id=next_page_id)
```

So `limit` is effectively applied too late to reduce I/O / deserialization work, and `items` is already built from the full event set.

## Impact

This may become expensive for larger conversations, especially on paths that just want the latest few events.

A concrete example is Slack V1's callback flow in PR #13043, which now uses `search_events(... limit=50)` to fetch the latest assistant message. That usage is directionally correct, but today it may still load the full conversation.

## Expected behavior

`search_events()` should apply pagination/limiting early enough that:

- it does **not** load the full conversation when only the latest N events are needed
- returned `items` correspond to the requested page/limit semantics

## Possible fixes

One of:

1. apply ordering/pagination to paths before loading events
2. support efficient "latest N events" retrieval in the storage layer
3. slice `items` consistently after sorting, while also reducing load work

## Related context

This was noticed while reviewing:
- PR #13043 (`feat(slack-v1): post final assistant message instead of LLM summary`)

The PR itself seems fine, but this underlying search behavior looks like an independent performance bug.

nehaaprasad:
### summary of PR

- search_events() was applying limit and page_id to the internal path list but never slicing the returned items. As a result, callers always received all matching events, regardless of limit and page_id.
This change slices the filtered and sorted items by start_offset and limit before returning, so pagination behaves correctly. Tests were added for limit, page_id, and filter+limit.

### Change Type
[x] Bug fix

### Fixes
Resolves #13335

### Release Notes


alvinttang (commented):
The bug is real and well-diagnosed: the original code sliced paths but then computed next_page_id and applied limits based on paths — however the pagination and limit were never applied to items, meaning the endpoint always returned all items regardless of limit or page_id. The fix correctly applies both offset and limit to items in a single slice and computes next_page_id against the pre-slice total. The boundary case where start_offset + limit == total is handled correctly — next_page_id will be None rather than pointing to an empty page. Consider adding a test for the exact-boundary case (final page size == limit) to explicitly confirm next_page_id is None, since this is a common off-by-one footgun in pagination logic.

nidhishgajjar:

## PR #13344: fix(event): apply limit and page_id to items in search_events

### What I reviewed
Fixes pagination in `search_events` to apply `limit` and `page_id` to the filtered `items` list instead of the pre-filter `paths` list.

### Findings

**✅ Correct fix for a pagination bug**
- The old code applied offset/limit to `paths` (file paths) before filtering, but returned `items` (parsed events after filtering). This meant:
  - With filters active, returned items could be fewer than `limit` even when more matching items existed
  - `page_id` offset was applied to paths, not the actual result set, causing incorrect pagination
- The fix applies `start_offset` and `limit` to `items` after filtering/sorting, ensuring the limit controls the actual returned results

**✅ `next_page_id` calculation is correct**
- Now set to `None` when `total <= start_offset + limit`, correctly indicating no more pages
- Old code would set `next_page_id` whenever `len(paths) > limit`, which could set it even when all items fit on one page

**✅ Good test coverage**
- Tests for limit alone, page_id pagination, and combined filter + limit scenarios

### Summary
Clean fix for a real pagination bug. The old code paginated pre-filter paths instead of post-filter items, causing incorrect results when filters were active. Tests cover the key scenarios.

### Assessment: approve

PR #12740, opened 2026-02-04
Added timeout parameter to OSS, exclude Enterprise

joseluisll:
## Summary of PR

This PR provides a minimal implementation of handling the LLM timeout parameter in the settings.json configuration file. With this minimal implementation, the LLM timeout parameter is passed to the DockerSandboxService along with the rest of the existing parameters.

## Change Type

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Other (dependency update, docs, typo fixes, etc.)

## Checklist

- [ X] I have read and reviewed the code and I understand what the code is doing.
- [ X] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves #8768

## Release Notes

- [X] Include this change in the Release Notes.
Added the timeout parameter to the LLM configuration. This allow local LLM´s enough time to provid the completion.

PR #13359, opened 2026-03-12
[FEAT] Configurable LLM_TIMEOUT for OSS using file configuration only

### LINKED ISSUE #8768: Improve timeout handling and feedback for slow local LLMs

When using OpenHands with LM Studio, long generations from local models can exceed the default timeout. This causes OpenHands to disconnect before the response is ready, resulting in a loop of retry attempts with no clear explanation or feedback.

- Expose an **LLM timeout** setting in the **LLM settings** panel when **Advanced mode** is enabled  
- Improve retry messages to clearly indicate the cause of failure (e.g. **“LLM timeout exceeded”**)

**Additional context**

I fixed the issue by setting the `LLM_TIMEOUT` environment variable to a higher value via Docker, based on advice from a developer. The workaround works, but most users wouldn’t know to try this without guidance.

joseluisll:
## Summary of PR

This PR provides a minimal implementation of handling the LLM timeout parameter in the **settings.json** configuration file. With this minimal implementation, the LLM timeout parameter is passed to the DockerSandboxService along with the rest of the existing parameters. It has been reviewed previously so that the filter of the timeout parameter for the enterprise version does not affect other timeout parameters, like MCP. 

## Change Type

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

- [ X] I have read and reviewed the code and I understand what the code is doing.
- [ X] I have tested the code to the best of my ability and ensured it works as expected.

## Fixes

Resolves #8768

## Release Notes

- [X] Include this change in the Release Notes.

Added the timeout parameter to the LLM configuration. This allow local LLM´s enough time to provide the completion.


joseluisll:
The previous PR #12740 contains the history of revisions.

mamoodi:
Hi there. There seems to be a lot of formatting changes like '' -> "". Any reason for all of these changes?

PR #13403, opened 2026-03-14
fix: patch GLib CVE-2025-14087 in runtime Docker images

aivong-openhands:
## Summary
Add explicit security upgrade for libglib2.0 packages to patch CVE-2025-14087 (buffer underflow in GVariant parser leading to heap corruption).

## Vulnerability Details
- **CVE**: [CVE-2025-14087](https://avd.aquasec.com/nvd/cve-2025-14087)
- **Severity**: CRITICAL
- **Package**: libglib2.0-dev-bin (and related packages)
- **Impact**: A remote attacker can cause heap corruption, leading to denial of service or potential code execution via a buffer-underflow in the GVariant parser
- **Fixed Version**: 2.84.4-3~deb13u2

## Related
- Resolves security alert: https://github.com/All-Hands-AI/OpenHands-Cloud/security/code-scanning/3227

## Changes
This PR adds an explicit `apt-get --only-upgrade` for libglib2.0 packages in the Debian-based runtime image build. This ensures the fix is applied even when building from base images that may not have the latest security patches.

## Testing
- The latest runtime image already has the fixed version (2.84.4-3~deb13u2)
- This change is a defensive measure to ensure future builds always get the patched version


aivong-openhands:
@OpenHands are all of these glib upgrades still needed to address the mentioned CVEs if we upgrade to the latest runtime image 1a77b05509ec010d44d1e274b0f8daace0f89ea8f-nikolaik?

aivong-openhands:
After investigating the question, I can confirm:

**The explicit GLib upgrades in this PR are no longer strictly necessary** when using the current `nikolaik/python-nodejs:python3.12-nodejs22` base image, which is used to build the `*-nikolaik` runtime images.

### Findings:

I pulled the latest `nikolaik/python-nodejs:python3.12-nodejs22` base image and checked the installed GLib packages:

```
ii  libglib2.0-0t64    2.84.4-3~deb13u2  amd64  GLib library of C routines
ii  libglib2.0-bin     2.84.4-3~deb13u2  amd64  Programs for the GLib library
ii  libglib2.0-dev     2.84.4-3~deb13u2  amd64  Development metapackage for the GLib library
ii  libglib2.0-dev-bin 2.84.4-3~deb13u2  amd64  Development utilities for the GLib library
```

The base image already contains the **fixed version `2.84.4-3~deb13u2`** which patches CVE-2025-14087.

### Options:

1. **Close this PR** - Since the base image already has the fix, this PR is not needed for current builds.

2. **Keep as a defensive measure** - The `--only-upgrade` command with `|| true` won't cause any issues (it will simply be a no-op if packages are already up-to-date), and provides a safety net if:
   - The base image is rebuilt at some point without security updates
   - Future builds use a different base image that might not have the patch

Given that the command is wrapped with `|| true` and adds minimal overhead, keeping it as a defensive measure is a reasonable choice. However, if you prefer to keep the Dockerfile minimal, closing this PR is also valid since the vulnerability is already addressed in the current base image.

PR #13440, opened 2026-03-17
PLTF-309: disable budget enforcement when ENABLE_BILLING=false

aivong-openhands:
## Problem

When billing is disabled in enterprise deployments (via `ENABLE_BILLING=false`, `REQUIRE_PAYMENT=0`), users still see "You're out of OpenHands Credits. Add funds" error when trying to use conversations.

## Root Cause

The `DEFAULT_INITIAL_BUDGET` was always set to `0.0` regardless of billing configuration. When a LiteLLM team is created with `max_budget=0`, LiteLLM enforces budget limits:

1. Team created with `max_budget = 0` (no budget remaining)
2. Any `spend > 0` triggers: `BudgetExceededError: Budget has been exceeded! Team=xxx Current cost: 0.07, Max budget: 0.0`

Additionally, in `create_entries`, this line converts `None` to `0.0`:
```python
team_budget = team_info.get('max_budget', 0.0) or 0.0
```

This means even if a team is created with `max_budget: null` (via helm chart), new users added to that team would get `max_budget_in_team: 0.0`.

## Solution

When `ENABLE_BILLING=false`:
- Set `DEFAULT_INITIAL_BUDGET` to `None` instead of `0.0`
- Pass `max_budget=None` when creating teams and adding users to teams
- Update type signatures to accept `float | None`
- Preserve `None` from existing teams when new users join

When `max_budget` is `None`, LiteLLM skips the budget check entirely, which is the correct behavior for enterprise deployments where billing is disabled.

## Changes

1. Added `ENABLE_BILLING` environment variable check
2. Modified `_get_default_initial_budget()` to return `None` when billing is disabled
3. Updated type signatures for `_create_team`, `_add_user_to_team`, and `_update_user_in_team` to accept `float | None`
4. Changed `create_entries` to preserve `None` budget from existing teams:
   ```python
   # Before (converts None to 0.0):
   team_budget = team_info.get('max_budget', 0.0) or 0.0
   
   # After (preserves None):
   existing_budget = team_info.get('max_budget')
   team_budget = existing_budget
   ```

## Related PRs

- **#12697** - Fixed the same type of bug in `migrate_entries` method (already merged)
- This PR fixes the same bug pattern in `create_entries` method
- Helm chart fix: https://github.com/All-Hands-AI/OpenHands-Cloud/pull/451

## Note for Existing Deployments

Existing deployments with teams that already have `max_budget=0` will need to manually update the LiteLLM database:

```sql
-- Connect to litellm database and run:
UPDATE "LiteLLM_TeamTable" SET max_budget = NULL WHERE max_budget = 0;
```

## Testing

- Verified type signatures with mypy
- The change is backwards compatible - when `ENABLE_BILLING=true`, behavior remains unchanged

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:dcc5630-nikolaik   --name openhands-app-dcc5630   docker.openhands.dev/openhands/openhands:dcc5630
```


aivong-openhands:
@openhands look into why tests are failing

aivong-openhands:
Testing enterprise on a replicated install, the out of funds error goes away with the bugfix in this PR:

Screenshot 2026-03-17 at 1 43 00 PM

The network error is happening on main as well:
Screenshot 2026-03-17 at 1 49 08 PM

jlav (approved):
I also tested this PR with a self-hosted install and confirm it fixes the "You're out of OpenHands Credits" error.

PR #13457, opened 2026-03-18
fix: stale useMemo closures in git hooks and incorrect SQLAlchemy result handling

wishhyt:
## Summary of PR

- **Fix stale `conversationId` closures in git hooks**: Both `useUnifiedGetGitChanges` and `useUnifiedGitDiff` call `getGitPath(conversationId, selectedRepository)` inside `useMemo` but omit `conversationId` from their dependency arrays. When switching conversations, `gitPath` / `absoluteFilePath` remain stale, causing V1 git API calls to use the wrong workspace path.
- **Fix SQLAlchemy `Row` vs ORM model mismatch in `pause_oldest_sandboxes`**: `list(db_session.execute(query))` yields `Row` tuples, not `StoredRemoteSandbox` model instances. Subsequent `sandbox.id` access on a `Row` object fails. Every other call site in the same file correctly uses `result.scalars().all()`.

## Change Type

- [x] Bug fix

## Checklist

### 1. `useUnifiedGetGitChanges` — missing `conversationId` in useMemo deps
**File:** `frontend/src/hooks/query/use-unified-get-git-changes.ts`

`getGitPath(conversationId, selectedRepository)` produces paths like `/workspace/project/{conversationId}/{repoName}`. The `useMemo` dependency array only included `[selectedRepository]`, so switching conversations would not recompute `gitPath`.

### 2. `useUnifiedGitDiff` — missing `conversationId` in useMemo deps
**File:** `frontend/src/hooks/query/use-unified-git-diff.ts`

Same pattern: `getGitPath(conversationId, selectedRepository)` is called inside `useMemo`, but `conversationId` was missing from `[isV1Conversation, selectedRepository, config.filePath]`.

### 3. `pause_oldest_sandboxes` — `list(execute())` instead of `scalars().all()`
**File:** `openhands/app_server/sandbox/remote_sandbox_service.py`

`_secure_select()` returns `select(StoredRemoteSandbox)`. Calling `list(result)` yields `Row[tuple[StoredRemoteSandbox]]` objects. The code then accesses `sandbox.id` on these Row objects, which fails. The same file uses `result.scalars().all()` at lines 301, 379, etc. — this line was an inconsistency.

### DIFF
--- a/frontend/src/hooks/query/use-unified-get-git-changes.ts
+++ b/frontend/src/hooks/query/use-unified-get-git-changes.ts
@@ -28,7 +28,7 @@ export const useUnifiedGetGitChanges = () => {
   // Calculate git path based on selected repository
   const gitPath = React.useMemo(
     () => getGitPath(conversationId, selectedRepository),
-    [selectedRepository],
+    [conversationId, selectedRepository],
   );
 
   const result = useQuery({
--- a/frontend/src/hooks/query/use-unified-git-diff.ts
+++ b/frontend/src/hooks/query/use-unified-git-diff.ts
@@ -34,7 +34,7 @@ export const useUnifiedGitDiff = (config: UseUnifiedGitDiffConfig) => {
 
     const gitPath = getGitPath(conversationId, selectedRepository);
     return `${gitPath}/${config.filePath}`;
-  }, [isV1Conversation, selectedRepository, config.filePath]);
+  }, [isV1Conversation, conversationId, selectedRepository, config.filePath]);
 
   return useQuery({
     queryKey: [
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -598,7 +598,8 @@ class RemoteSandboxService(SandboxService):
         query = query.filter(StoredRemoteSandbox.id.in_(running_session_ids)).order_by(
             StoredRemoteSandbox.created_at.desc()
         )
-        running_sandboxes = list(await self.db_session.execute(query))
+        result = await self.db_session.execute(query)
+        running_sandboxes = result.scalars().all()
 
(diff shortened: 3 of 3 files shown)


neubig (changes requested):

mamoodi:
Hi there. For a final check in, do you think you will be continuing this PR and making a video demonstrating it working?

nidhishgajjar:

## PR #13457: fix: stale useMemo closures in git hooks and incorrect SQLAlchemy result handling

### What I reviewed
Two bug fixes: (1) stale React `useMemo` closures in git-related hooks, and (2) incorrect SQLAlchemy query result handling in sandbox cleanup.

### Findings

**✅ useMemo dependency fixes are correct**
- `use-unified-get-git-changes.ts`: Adds `conversationId` to the `useMemo` deps that calls `getGitPath(conversationId, selectedRepository)`. Without it, the memoized path would be stale when switching conversations.
- `use-unified-git-diff.ts`: Same fix — `conversationId` was missing from the dependency array that computes the git diff path.
- Both are classic React stale-closure bugs that could cause incorrect git operations when switching between conversations.

**✅ SQLAlchemy fix is correct**
```python
# Before: returns list of Row tuples
running_sandboxes = list(await self.db_session.execute(query))
# After: returns list of ORM model instances
result = await self.db_session.execute(query)
running_sandboxes = result.scalars().all()
```
`list(session.execute(query))` produces `Row` tuples (`[Row((StoredRemoteSandbox,),)]`), not model instances. `result.scalars().all()` correctly unwraps to `[StoredRemoteSandbox]`. If downstream code accesses model attributes like `sandbox.id`, the old code would fail or behave unexpectedly.

### Summary
Clean, focused bug fixes. Both the React stale-closure and SQLAlchemy result-handling issues are real bugs with correct fixes. No concerns.

### Assessment: approve

PR #13466, opened 2026-03-18
Fix: Prevent auto-logout on 401 errors in oss mode

cbagwell:
## Summary of PR

Backend does not currently distinguish between transient network issues and real authentication issues.

    21:46:40 - openhands:WARNING: http_client.py:101 - HTTP error on bitbucket_data_center API: ConnectTimeout :
    21:46:40 - openhands:WARNING: provider.py:238 - Failed to get bitbucket data center projects bitbucket_data_center API request timed out: ConnectTimeout
    INFO:     10.14.16.62:60234 - "GET /api/user/info HTTP/1.1" 401 Unauthorized

In oss mode, 401 errors from git providers were triggering automatic
logout, which permanently deleted user credentials via the
/api/unset-provider-tokens endpoint. This was an over reaction
since transient network failures are lumped into 401 as well as
oss mode having no session to invalidate.

This change prevents auto-logout in oss mode while preserving the
behavior in saas mode where it's needed for OAuth token refresh.

In addition, logout menu option is now hidden in oss mode since
there is no session to invalidate and user would not except
that option to wipe out both tokens and URL configuration.

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

### DIFF
--- a/frontend/__tests__/components/features/user/user-context-menu.test.tsx
+++ b/frontend/__tests__/components/features/user/user-context-menu.test.tsx
@@ -156,11 +156,19 @@ describe("UserContextMenu", () => {
     useSelectedOrganizationStore.setState({ organizationId: null });
   });
 
-  it("should render the default context items for a user", () => {
+  it("should render the default context items for a user", async () => {
+    vi.spyOn(OptionService, "getConfig").mockResolvedValue(
+      createMockWebClientConfig({ app_mode: "saas" }),
+    );
+
     renderUserContextMenu({ type: "member", onClose: vi.fn, onOpenInviteModal: vi.fn });
 
     screen.getByTestId("org-selector");
-    screen.getByText("ACCOUNT_SETTINGS$LOGOUT");
+
+    // Wait for config to load so logout button appears
+    await waitFor(() => {
+      expect(screen.getByText("ACCOUNT_SETTINGS$LOGOUT")).toBeInTheDocument();
+    });
 
     expect(
       screen.queryByText("ORG$INVITE_ORG_MEMBERS"),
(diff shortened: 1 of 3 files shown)


enyst (commented):
I think this may have been because in the open source software, there is a “logout” button? Please correct me if wrong

cbagwell:
Yes, in OSS there is a logout button.  If someone manually logs out, I'm not sure they are expected it to delete both token + provider URL but I could live with that behavior.

But the fact that auto-logout with these deletes is occurring during transient network connectivity issues is my main concern.

image

cbagwell:
Alternative 1:

This code in useGitUser() could be made conditional to SaaS only instead of always:

```
  // If we get a 401 here, it means that the integration tokens need to be
  // refreshed. Since this happens at login, we log out.
  React.useEffect(() => {
    if (user?.error?.response?.status === 401) {
      logout.mutate();
    }
  }, [user.status]);
```

Alternative 2:
get_user() in provider.py and some supporting code could be modified to not raise authentication error for things like timeouts or rate limits.  All exceptions today are mapped to this:

```
        raise AuthenticationError('Need valid provider token')
```

enyst:
Alternative 1 sounds good to me!

Alternative 3: don't display the Logout button when there is no actual logged in user. The user can also delete the tokens from the UI if they want to. WDYT?

But 1 sounds to me, whatever is simpler and works.

cbagwell:
I updated PR for this direction of preventing the auto-logout in oss mode. This was my original solution as well but I didn't like the fact that logout menu wiped out full provider provisioning in oss mode and prevented that the easy way.

I liked your suggestion of hiding logout menu as it solves same issue and user isn't wondering want might happen if they select that option since they never logged in.   So I implemented both alternative 1 and 3 together.

This is the OSS menu after change:

image

enyst:
@openhands Do a /codereview on this PR. Understand all comments and basically we want to keep it quite minimal.

enyst (commented):
[frontend/src/hooks/query/use-git-user.ts, Lines 40–43] :test_tube: Testing: This is the primary bug fix in the PR, but there is still no regression test for the OSS path. `frontend/__tests__/hooks/use-git-user.test.tsx` only verifies that a 401 logs out in SaaS mode and that non-401 errors do not. Please add an OSS-mode case asserting that a 401 does **not** call `logout.mutate()`, so this transient-network regression is covered directly.

enyst (approved):
(OpenHands GPT-5.4) Following up: agreed we should keep this minimal. The SaaS-only guard in `use-git-user.ts` addresses the bug, and I do not think an additional OSS-specific mock-based test is necessary here.

enyst:
Thank you!

PR #13469, opened 2026-03-18
feat(backend): expose API key org_id via new GET /api/keys/current endpoint (org project)

### LINKED ISSUE #13465: [Bug]: /users/me endpoint does not return API key's org_id
## Bug Description

When authenticating with an API key and calling the `/users/me` (or `/api/user/info`) endpoint, there is no way to determine which organization the API key belongs to.

## Expected Behavior

Since API keys are created with an `org_id` stored in the database, the `/me` endpoint should return this org_id (or the full org info) so callers can identify which org context they are operating in.

## Actual Behavior

The only way to find org information is to call `GET /api/orgs` which returns `current_org_id` - but this is the user's currently active org, not the org the API key was created in. These can differ if the user has switched orgs since creating the key.

To access a specific org's endpoints, you must already know the org_id in advance.

## Steps to Reproduce

1. Create an API key in Org A
2. Call `GET /api/v1/users/me` with the key
3. Response contains user settings but no org_id
4. Call `GET /api/orgs` - returns `current_org_id` which may differ from the key's org

## Suggested Fix

Either:
1. Add `org_id` field to the `/users/me` response reflecting the API key's org
2. Add a new endpoint like `GET /api/keys/current` that returns the authenticated key's metadata including `org_id`

## Relevant Code

- `enterprise/storage/api_key.py` - ApiKey model has `org_id` field
- `openhands/app_server/user/user_router.py` - `/users/me` endpoint
- `enterprise/server/routes/user.py` - `/api/user/info` endpoint

## OpenHands Installation Method

Development workflow

hieptl:
## Summary of PR

When authenticating with an API key, callers had no way to determine which organization the API key was created for. The current_org_id from /api/organizations reflects the user's active org, which can differ from the API key's org if the user switched orgs after creating the key.

**Changes:**

- Add ApiKeyValidationResult dataclass to return user_id, org_id, key_id, and key_name from validate_api_key()
- Add api_key_org_id, api_key_id, api_key_name fields to SaasUserAuth
- Add GET /api/keys/current endpoint that returns the authenticated API key's metadata including its org_id
- Update affected code and tests

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/cf6d6010-69dd-4e3d-a653-e9590a8f9224

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #13465

Unblocks [APP-1000](https://linear.app/all-hands-ai/issue/APP-1000/phase-1a-automations-service-crud-auth-db-cron-polling-log-only)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:06ccba8-nikolaik   --name openhands-app-06ccba8   docker.openhands.dev/openhands/openhands:06ccba8
```

### DIFF
--- a/enterprise/server/routes/api_keys.py
+++ b/enterprise/server/routes/api_keys.py
@@ -1,4 +1,5 @@
 from datetime import UTC, datetime
+from typing import cast
 
 from fastapi import APIRouter, Depends, HTTPException, Request, status
 from pydantic import BaseModel, field_validator
@@ -296,25 +297,21 @@ async def get_current_api_key(
             detail='This endpoint requires API key authentication. Not available for cookie-based auth.',
         )
 
-    # Type narrow to SaasUserAuth for proper type safety
-    if not isinstance(user_auth, SaasUserAuth):
-        raise HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            detail='Unexpected auth type for API key authentication',
-        )
+    # In SaaS context, bearer auth always produces SaasUserAuth
+    saas_user_auth = cast(SaasUserAuth, user_auth)
 
-    if user_auth.api_key_org_id is None:
+    if saas_user_auth.api_key_org_id is None:
         raise HTTPException(
             status_code=status.HTTP_400_BAD_REQUEST,
             detail='This API key was created before organization support. Please regenerate your API key to use this endpoint.',
         )
 
     return CurrentApiKeyResponse(
-        id=user_auth.api_key_id,
-        name=user_auth.api_key_name,
-        org_id=str(user_auth.api_key_org_id),
+        id=saas_user_auth.api_key_id,
+        name=saas_user_auth.api_key_name,
+        org_id=str(saas_user_auth.api_key_org_id),
         user_id=user_id,
-        auth_type=user_auth.auth_type.value,
+        auth_type=saas_user_auth.auth_type.value,
     )
 
 
--- a/enterprise/tests/unit/server/routes/test_api_keys.py
+++ b/enterprise/tests/unit/server/routes/test_api_keys.py
@@ -533,7 +533,6 @@ class TestGetCurrentApiKey:
         org_id = uuid.uuid4()
         mock_request = MagicMock()
 
-        # Use real SaasUserAuth instance for proper isinstance check
         user_auth = SaasUserAuth(
             refresh_token=SecretStr('mock-token'),
             user_id=user_id,
@@ -582,7 +581,6 @@ class TestGetCurrentApiKey:
         user_id = 'user-123'
         mock_request = MagicMock()
 
-        # Use real SaasUserAuth instance for proper isinstance check
         user_auth = SaasUserAuth(
             refresh_token=SecretStr('mock-token'),
             user_id=user_id,


enterprise/server/auth/saas_user_auth.py
```diff
@@ -59,6 +60,10 @@ class SaasUserAuth(UserAuth):
     _secrets: Secrets | None = None
     accepted_tos: bool | None = None
     auth_type: AuthType = AuthType.COOKIE
+    # API key context fields - populated when authenticated via API key
+    api_key_org_id: UUID | None = None
```
malhotra5, line 64: NIT: instead of storing these at the root level, since its specific to auth type maybe we can modify the auth type field 

AuthType can be a string literal "COOKIE" or a well typed object instead `BearerAuth(api_name=, org_id=, key_id=)`

malhotra5: and we can remove the defensive checks later for whether org id was attached

enterprise/server/routes/api_keys.py
```diff
@@ -262,6 +273,49 @@ async def delete_api_key(
...
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail='This endpoint requires API key authentication. Not available for cookie-based auth.',
+        )
+
+    # Access API key context fields via getattr for type safety
+    # These fields are only present on SaasUserAuth when authenticated via API key
+    api_key_org_id = getattr(user_auth, 'api_key_org_id', None)
+    api_key_id = getattr(user_auth, 'api_key_id', None)
+    api_key_name = getattr(user_auth, 'api_key_name', None)
```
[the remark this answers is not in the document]
hieptl: I have updated the code based on your feedback. Thank you! 🙏

enterprise/server/routes/api_keys.py
```diff
@@ -262,6 +273,49 @@ async def delete_api_key(
...
+
+    if api_key_org_id is None:
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail='API key organization context not available',
```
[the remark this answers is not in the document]
hieptl: I have updated the code based on your feedback. Thank you! 🙏

enterprise/server/routes/api_keys.py
```diff
@@ -262,6 +273,49 @@ async def delete_api_key(
...
+        )
+
+    return CurrentApiKeyResponse(
+        id=api_key_id,
+        name=api_key_name,
+        org_id=str(api_key_org_id),
+        user_id=user_id,
```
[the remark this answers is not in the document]
hieptl: I have updated the code based on your feedback. Thank you! 🙏

enterprise/server/auth/saas_user_auth.py
hieptl: Hello @malhotra5,

I agree that this would result in a cleaner design—having `auth_type` be either `Literal["COOKIE"]` or `BearerAuth(org_id=, key_id=, key_name=)` would:

- Eliminate invalid states (removing the possibility of `auth_type=COOKIE` with non-None API key fields)
- Enable better type narrowing without requiring `isinstance` checks

That said, implementing this would be a larger refactor affecting:

- The `SaasUserAuth` class structure
- The `get_auth_type()` interface (which currently returns the `AuthType` enum and is used throughout the code)
- The `saas_user_auth_from_bearer` and `saas_user_auth_from_cookie` functions
- All callers that currently check `auth_type`
- Multiple test files

Considering that the scope of this PR is focused on exposing the `org_id` via a new endpoint, I would suggest keeping the current flat structure for now and treating this refactor as a follow-up. The current implementation is functional, the fields are clearly documented as “API key context fields,” and we maintain proper type narrowing using `isinstance` at the call sites.

Would you be comfortable deferring this enhancement to a separate PR dedicated to improving the auth type modeling?

Thank you! 🙏

malhotra5: Sure!

enterprise/server/routes/api_keys.py
```diff
@@ -262,6 +274,50 @@ async def delete_api_key(
...
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail='This endpoint requires API key authentication. Not available for cookie-based auth.',
+        )
+
+    # Type narrow to SaasUserAuth for proper type safety
+    if not isinstance(user_auth, SaasUserAuth):
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail='Unexpected auth type for API key authentication',
+        )
```
malhotra5: Instead of this can just cast the user_auth to saasuserauth (so we can avoid the 500 error code)

hieptl: Hello @malhotra5,

PR #13471, opened 2026-03-18
fix: treat llm_base_url="" as explicit clear in store_llm_settings

### LINKED ISSUE #13420: [Bug]: LLM settings cannot persist an empty base URL

### Bug Description

It seems impossible to persist an empty `llm_base_url` from the LLM settings UI, anymore.

Observed behavior:

- In **Basic** view, clicking **Save** can bounce the form back into **Advanced** view with a populated Base URL that the user did not set.
- In **Advanced** view, if the user deletes the Base URL and clicks **Save**, the Base URL comes back.

The settings page cannot truly clear an old/custom Base URL once one existed.

### Expected Behavior

If the user clears the Base URL in Advanced view, or saves from Basic view (which is supposed to clear advanced-only settings), the saved settings should remain empty / unset instead of being repopulated with the old base URL.

### Actual Behavior

An empty `llm_base_url` is rewritten to the old base URL during save.

### Steps to Reproduce

1. Open **Settings → LLM Settings**
2. Start from a state where `llm_base_url` was previously set (or where the backend can infer one)
3. In **Advanced** view, delete Base URL and click **Save**
4. Reload the settings page
5. Observe that Base URL is populated again

Also reproducible via Basic view:

1. Switch to **Basic** view
2. Click **Save**
3. Reload the settings page
4. Observe that the form may re-open in **Advanced** mode because `llm_base_url` is non-empty again

### Suspected Cause

This looks like a regression/semantic mismatch introduced around:

- #12766
- #13225

Frontend behavior looks intentional:

- `frontend/src/routes/llm-settings.tsx` basic save explicitly sends `llm_base_url: ""` to clear advanced settings
- advanced save sends the Base URL input value directly, so deleting the field also sends `""`

But the backend then treats that empty string as a signal to auto-detect and repopulate a provider default:

- `openhands/server/routes/settings.py`
  - `if not settings.llm_base_url:`
  - if `None`, preserve existing value for partial updates
  - otherwise, for `""`, compute a default with `LITE_LLM_API_URL` or `get_provider_api_base()`

—-
HUMAN: the agent is confused here a bit: if I had base URL in Settings before, it is the OLD base URL that remains in place, it doesn’t get recomputed, nor cleared. I just cannot get rid of it.
—-

So the current code path conflates:

- `None` = not provided
- `""` = explicitly cleared by the user

That means the UI currently has no way to persist an intentionally empty base URL.

### Why this is user-visible

The frontend decides whether to open in Basic vs Advanced mode partly based on whether `llm_base_url` is non-empty:

- `frontend/src/utils/has-advanced-settings-set.ts`

So once the backend rewrites `""` into a computed URL, the UI can keep re-entering Advanced mode.

### Suggested Fix Direction

Treat `llm_base_url=""` as an explicit clear/unset request instead of auto-detecting a replacement value.

At minimum, the save semantics should distinguish all three states:

- `None`: field omitted / preserve existing value for unrelated partial saves
- `""`: explicitly cleared by user
- non-empty string: explicitly set by user

A regression test should cover both:

1. Basic view save clearing advanced settings
2. Advanced view delete-Base-URL save preserving the empty state after reload

Cloud / hosted app

### OpenHands Version

Current cloud deployment

### Additional Context

This may be related to the earlier fixes around computed base URLs and MCP saves, but it seems to be a separate remaining bug: MCP save preservation (`None`) was fixed, while explicit clearing (`""`) is still impossible.

LarytheLord:
## Summary

Fixes #13420 — it is currently impossible to persist an empty `llm_base_url` from the settings UI.

**Root cause:** `store_llm_settings` used `if not settings.llm_base_url:` which catches both `None` (field not provided at all) and `""` (explicitly cleared by the user). When the user:
- saves from Basic view (which intentionally sends `llm_base_url: ""` to clear advanced settings), or
- deletes the Base URL field in Advanced view and saves,

the backend treated `""` the same as "not provided", triggering auto-detection via `get_provider_api_base()` and overwriting the user's intent. The old base URL kept coming back.

**Fix:** Split the condition into two explicit branches:

```python
# Before
if not settings.llm_base_url:
    if settings.llm_base_url is None and existing_settings.llm_base_url:
        ...  # preserve existing

# After
if settings.llm_base_url is None:
    ...  # not provided — preserve existing or auto-detect (no behavior change)
elif settings.llm_base_url == "":
    settings.llm_base_url = None  # explicit clear — honor the intent
```

The `None` path has identical behavior to before. Only `""` changes: instead of triggering auto-detection, it now normalizes to `None` and returns.

## Test plan

- Updated `test_store_llm_settings_partial_update` to assert `None` (the correct post-fix behavior) instead of the auto-detected URL
- Added `test_store_llm_settings_advanced_view_clear_removes_base_url` as a named regression test for this issue
- All 16 tests in `test_settings_store_functions.py` pass
- Existing MCP save test (`test_store_llm_settings_mcp_update_preserves_base_url`) continues to pass — the `None` path (field not sent) is unchanged

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/openhands/server/routes/settings.py
+++ b/openhands/server/routes/settings.py
@@ -34,18 +34,18 @@ from openhands.storage.settings.settings_store import SettingsStore
 from openhands.utils.llm import get_provider_api_base, is_openhands_model
 
 LITE_LLM_API_URL = os.environ.get(
-    "LITE_LLM_API_URL", "https://llm-proxy.app.all-hands.dev"
+    'LITE_LLM_API_URL', 'https://llm-proxy.app.all-hands.dev'
 )
 
-app = APIRouter(prefix="/api", dependencies=get_dependencies())
+app = APIRouter(prefix='/api', dependencies=get_dependencies())
 
 
 @app.get(
-    "/settings",
+    '/settings',
     response_model=GETSettingsModel,
     responses={
-        404: {"description": "Settings not found", "model": dict},
-        401: {"description": "Invalid token", "model": dict},
+        404: {'description': 'Settings not found', 'model': dict},
+        401: {'description': 'Invalid token', 'model': dict},
     },
 )
 async def load_settings(
@@ -58,7 +58,7 @@ async def load_settings(
         if not settings:
             return JSONResponse(
                 status_code=status.HTTP_404_NOT_FOUND,
-                content={"error": "Settings not found"},
+                content={'error': 'Settings not found'},
             )
 
         # On initial load, user secrets may not be populated with values migrated from settings store
@@ -78,7 +78,7 @@ async def load_settings(
                     provider_tokens_set[provider_type] = provider_token.host
 
         settings_with_token_data = GETSettingsModel(
-            **settings.model_dump(exclude={"secrets_store"}),
+            **settings.model_dump(exclude={'secrets_store'}),
             llm_api_key_set=settings.llm_api_key is not None
             and bool(settings.llm_api_key),
             search_api_key_set=settings.search_api_key is not None
@@ -101,15 +101,15 @@ async def load_settings(
         settings_with_token_data.sandbox_api_key = None
         return settings_with_token_data
     except Exception as e:
-        logger.warning(f"Invalid token: {e}")
+        logger.warning(f'Invalid token: {e}')
         # Get user_id from settings if available
-        user_id = getattr(settings, "user_id", "unknown") if settings else "unknown"
+        user_id = getattr(settings, 'user_id', 'unknown') if settings else 'unknown'
         logger.info(
-            f"Returning 401 Unauthorized - Invalid token for user_id: {user_id}"
+            f'Returning 401 Unauthorized - Invalid token for user_id: {user_id}'
         )
         return JSONResponse(
             status_code=status.HTTP_401_UNAUTHORIZED,
-            content={"error": "Invalid token"},
+            content={'error': 'Invalid token'},
         )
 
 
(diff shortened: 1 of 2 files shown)


enyst:
Hey @LarytheLord thank you for the contribution!

Could you maybe show a video that it works, or logs (though video would be nice I think), and that is still works for the previous issue?

LarytheLord:
Thanks @enyst — here's the test evidence:

**Regression test** (`test_store_llm_settings_advanced_view_clear_removes_base_url`):

```
settings = Settings(llm_model='gpt-4', llm_base_url='')     # user clears field
existing = Settings(llm_model='gpt-4', llm_base_url='https://my-custom-proxy.example.com')

result = await store_llm_settings(settings, existing)
assert result.llm_base_url is None   # ✅ old URL is gone
```

**Existing behavior preserved** (`test_store_llm_settings_partial_update` and `test_store_llm_settings_mcp_update_preserves_base_url`):
- When `llm_base_url` is `None` (not sent at all, e.g. MCP save), the existing URL is kept — no behavior change.
- When `llm_base_url` is `""` (explicitly cleared by the user), it now normalizes to `None` instead of re-running auto-detection.

All 3 tests pass:

```
tests/.../test_settings_store_functions.py::test_store_llm_settings_advanced_view_clear_removes_base_url PASSED
tests/.../test_settings_store_functions.py::test_store_llm_settings_partial_update PASSED
tests/.../test_settings_store_functions.py::test_store_llm_settings_mcp_update_preserves_base_url PASSED
```

Full suite (16/16): `pytest tests/unit/server/routes/test_settings_store_functions.py` — all green.

I can record a UI walkthrough too if you'd like, but the logic is fully covered by these tests. The key distinction is the `is None` vs `== ""` split in `store_llm_settings` — the diff is small and focused.

PR #13501, opened 2026-03-19
fix(server): preserve cleared llm base url on settings save

### LINKED ISSUE #13420: [Bug]: LLM settings cannot persist an empty base URL

antonstrobe:
## Summary of PR

- preserve an explicitly cleared `llm_base_url` instead of repopulating it during settings save
- keep existing semantics for omitted `llm_base_url` values so unrelated partial updates still preserve prior settings
- add regression coverage for both the settings helper and `/api/settings` API behavior

## Demo Screenshots/Videos

No screenshot needed.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #13420

## Release Notes

Users can now clear a custom LLM Base URL in settings without it being restored on save.

## Validation

- `poetry run pytest tests/unit/server/routes/test_settings_store_functions.py::test_store_llm_settings_explicit_empty_base_url_stays_cleared tests/unit/server/routes/test_settings_store_functions.py::test_store_llm_settings_mcp_update_preserves_base_url tests/unit/server/routes/test_settings_store_functions.py::test_store_llm_settings_no_existing_base_url_uses_auto_detection tests/unit/server/routes/test_settings_api.py::test_llm_base_url_can_be_explicitly_cleared tests/unit/server/routes/test_settings_api.py::test_search_api_key_preservation -q`
- `poetry run pre-commit run --files openhands/server/routes/settings.py tests/unit/server/routes/test_settings_store_functions.py tests/unit/server/routes/test_settings_api.py --config ./dev_config/python/.pre-commit-config.yaml`

### DIFF
--- a/openhands/server/routes/settings.py
+++ b/openhands/server/routes/settings.py
@@ -123,9 +123,12 @@ async def store_llm_settings(
             settings.llm_api_key = existing_settings.llm_api_key
         if settings.llm_model is None:
             settings.llm_model = existing_settings.llm_model
-        # if llm_base_url is missing or empty, try to preserve existing or determine appropriate URL
-        if not settings.llm_base_url:
-            if settings.llm_base_url is None and existing_settings.llm_base_url:
+        # Distinguish three cases for llm_base_url:
+        # - None: field omitted, so preserve existing or auto-detect if needed
+        # - "": explicitly cleared by the user, keep it empty
+        # - non-empty string: explicit user-provided value, keep it as-is
+        if settings.llm_base_url is None:
+            if existing_settings.llm_base_url:
                 # Not provided at all (e.g. MCP config save) - preserve existing
                 settings.llm_base_url = existing_settings.llm_base_url
             elif is_openhands_model(settings.llm_model):
--- a/tests/unit/server/routes/test_settings_api.py
+++ b/tests/unit/server/routes/test_settings_api.py
@@ -152,3 +152,26 @@ async def test_search_api_key_preservation(test_client):
     assert response.json()['search_api_key_set'] is True
     # Verify the other field updated correctly
     assert response.json()['llm_model'] == 'claude-3-opus'
+
+
+@pytest.mark.asyncio
+async def test_llm_base_url_can_be_explicitly_cleared(test_client):
+    """Test that sending an empty llm_base_url persists the cleared state."""
+    initial_settings = {
+        'llm_model': 'gpt-4',
+        'llm_base_url': 'https://old-proxy.example.com',
+    }
+    response = test_client.post('/api/settings', json=initial_settings)
+    assert response.status_code == 200
+
+    update_settings = {
+        'llm_model': 'gpt-4',
+        'llm_base_url': '',
+    }
+    response = test_client.post('/api/settings', json=update_settings)
+    assert response.status_code == 200
+
+    response = test_client.get('/api/settings')
+    assert response.status_code == 200
+    assert response.json()['llm_model'] == 'gpt-4'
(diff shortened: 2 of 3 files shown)


antonstrobe:
Closing this PR because the primary server-side fix was already merged separately in #13471.

This branch now conflicts with `main` and would reintroduce a different `llm_base_url` clearing contract than the one already accepted upstream.

Follow-up frontend/API contract hardening is tracked in #13510.

PR #13502, opened 2026-03-20
Fix FastAPI Query parameter validation: lte -> le

malhotra5:
## Summary of PR

Fixes a bug where FastAPI Query parameter validation was silently ignored because of incorrect parameter naming.

The FastAPI `Query` parameter was using `lte=100` but FastAPI/Pydantic expects `le=100` (without the 't'). When you pass `lte=100`, FastAPI **silently ignores** it because it goes into the `**extra` kwargs, so the validation constraint was never actually applied.

```python
# Before (broken - silently ignored):
Query(title='...', gt=0, lte=100)

# After (correct - validation works):
Query(title='...', gt=0, le=100)
```

This was causing `AssertionError` crashes in production when clients passed invalid limit values (e.g., `limit=200`), because the FastAPI validation was not catching them.

### Changes

- Fixed all `Query(lte=...)` to `Query(le=...)` across all router files
- Kept manual `HTTPException(400)` for batch operations that check `len(ids)` since FastAPI Query constraints do not support list length validation
- Added unit tests verifying FastAPI validation returns 422 for invalid limits

### Behavior Change: Batch Operations Now Allow Exactly 100 Items

**Note:** This PR also fixes an inconsistency in batch operation limits. The original code used `assert len(ids) < 100` (strict less-than) which incorrectly rejected requests with exactly 100 items. The new code uses `if len(ids) > 100` which correctly allows up to 100 items, matching the documented limit and the behavior of other endpoints.

Affected batch operations:
- `batch_get_sandboxes` in `sandbox_router.py`
- `batch_get_app_conversation_start_tasks` in `app_conversation_router.py`

### Affected files

- `openhands/app_server/event/event_router.py`
- `openhands/app_server/app_conversation/app_conversation_router.py`
- `openhands/app_server/sandbox/sandbox_router.py`
- `openhands/app_server/sandbox/sandbox_spec_router.py`
- `enterprise/server/sharing/shared_event_router.py`
- `enterprise/server/sharing/shared_conversation_router.py`
- `enterprise/server/routes/orgs.py`

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Release Notes

Fixed FastAPI Query parameter validation for limit constraints. Previously, using `lte=100` instead of the correct `le=100` caused validation to be silently ignored, resulting in server crashes when invalid limit values were passed. Now the API properly returns 422 Unprocessable Entity for invalid parameters. Additionally, batch operations now correctly allow up to 100 items (previously they incorrectly rejected exactly 100 items).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:9e7f59f-nikolaik   --name openhands-app-9e7f59f   docker.openhands.dev/openhands/openhands:9e7f59f
```

### DIFF
--- a/tests/unit/app_server/test_event_router.py
+++ b/tests/unit/app_server/test_event_router.py
@@ -11,7 +11,6 @@ import pytest
 from fastapi import FastAPI, HTTPException, status
 from fastapi.testclient import TestClient
 
-from openhands.agent_server.models import EventPage
 from openhands.app_server.event.event_router import batch_get_events, router
 from openhands.server.dependencies import check_session_api_key
 
@@ -55,15 +54,15 @@ class TestSearchEventsValidation:
         conversation_id = str(uuid4())
 
         response = test_client.get(
-            f"/conversation/{conversation_id}/events/search",
-            params={"limit": 200},
+            f'/conversation/{conversation_id}/events/search',
+            params={'limit': 200},
         )
 
         assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
         # Verify the error message mentions the constraint
-        error_detail = response.json()["detail"]
+        error_detail = response.json()['detail']
         assert any(
-            "less than or equal to 100" in str(err).lower() or "le" in str(err).lower()
+            'less than or equal to 100' in str(err).lower() or 'le' in str(err).lower()
             for err in error_detail
         )
 
@@ -75,8 +74,8 @@ class TestSearchEventsValidation:
         conversation_id = str(uuid4())
 
         response = test_client.get(
-            f"/conversation/{conversation_id}/events/search",
-            params={"limit": 0},
+            f'/conversation/{conversation_id}/events/search',
+            params={'limit': 0},
         )
 
         assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
@@ -89,8 +88,8 @@ class TestSearchEventsValidation:
         conversation_id = str(uuid4())
 
         response = test_client.get(
-            f"/conversation/{conversation_id}/events/search",
-            params={"limit": -1},
+            f'/conversation/{conversation_id}/events/search',
+            params={'limit': -1},
         )
 
         assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
@@ -103,8 +102,8 @@ class TestSearchEventsValidation:
         conversation_id = str(uuid4())
 
         response = test_client.get(
-            f"/conversation/{conversation_id}/events/search",
-            params={"limit": 100},
+            f'/conversation/{conversation_id}/events/search',
+            params={'limit': 100},
         )
 
         # Should pass validation (not 422) - may fail on other errors like missing service
@@ -118,8 +117,8 @@ class TestSearchEventsValidation:
         conversation_id = str(uuid4())
 
         response = test_client.get(
-            f"/conversation/{conversation_id}/events/search",
-            params={"limit": 1},
+            f'/conversation/{conversation_id}/events/search',
+            params={'limit': 1},
         )
 
         # Should pass validation (not 422) - may fail on other errors like missing service
@@ -151,8 +150,8 @@ class TestBatchGetEvents:
             )
 
         assert exc_info.value.status_code == status.HTTP_400_BAD_REQUEST
-        assert "Cannot request more than 100 events" in exc_info.value.detail
-        assert "101" in exc_info.value.detail
+        assert 'Cannot request more than 100 events' in exc_info.value.detail
+        assert '101' in exc_info.value.detail
 
     async def test_accepts_exactly_100_ids(self):
         """Test that exactly 100 IDs is accepted.


openhands/app_server/app_conversation/app_conversation_router.py
```diff
@@ -472,7 +468,11 @@ async def batch_get_app_conversation_start_tasks(
     ),
 ) -> list[AppConversationStartTask | None]:
     """Get a batch of start app conversation tasks given their ids. Return None for any missing."""
-    assert len(ids) < 100
+    if len(ids) > 100:
```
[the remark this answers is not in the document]
malhotra5: Mentioned in the PR description now, resolving this

openhands/app_server/sandbox/sandbox_router.py
```diff
@@ -60,7 +58,11 @@ async def batch_get_sandboxes(
     sandbox_service: SandboxService = sandbox_service_dependency,
 ) -> list[SandboxInfo | None]:
     """Get a batch of sandboxes given their ids, returning null for any missing."""
-    assert len(id) < 100
+    if len(id) > 100:
```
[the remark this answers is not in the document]
malhotra5: Mentioned in the PR description now, resolving this

enterprise/server/sharing/shared_conversation_router.py
```diff
@@ -72,8 +72,6 @@ async def search_shared_conversations(
     shared_conversation_service: SharedConversationInfoService = shared_conversation_info_service_dependency,
 ) -> SharedConversationPage:
     """Search / List shared conversations."""
-    assert limit > 0
```
chuckbutkus, line 75: These are just removed?  Should there be logic to replace them?

enterprise/server/sharing/shared_event_router.py
```diff
@@ -77,13 +77,11 @@ async def search_shared_events(
     ] = None,
     limit: Annotated[
         int,
-        Query(title='The max number of results in the page', gt=0, lte=100),
+        Query(title='The max number of results in the page', gt=0, le=100),
     ] = 100,
     shared_event_service: SharedEventService = shared_event_service_dependency,
 ) -> EventPage:
     """Search / List events for a shared conversation."""
-    assert limit > 0
```
chuckbutkus, line 85: Same as above.

openhands/app_server/app_conversation/app_conversation_router.py
```diff
@@ -248,8 +248,6 @@ async def search_app_conversations(
     ),
 ) -> AppConversationPage:
     """Search / List sandboxed conversations."""
-    assert limit > 0
```
chuckbutkus, line 251: Same as above.

openhands/app_server/app_conversation/app_conversation_router.py
```diff
@@ -422,16 +420,14 @@ async def search_app_conversation_start_tasks(
...
-            lte=100,
+            le=100,
         ),
     ] = 100,
     app_conversation_start_task_service: AppConversationStartTaskService = (
         app_conversation_start_task_service_dependency
     ),
 ) -> AppConversationStartTaskPage:
     """Search / List conversation start tasks."""
-    assert limit > 0
```
chuckbutkus, line 433: Same

openhands/app_server/event/event_router.py
```diff
@@ -51,13 +51,11 @@ async def search_events(
     ] = None,
     limit: Annotated[
         int,
-        Query(title='The max number of results in the page', gt=0, lte=100),
+        Query(title='The max number of results in the page', gt=0, le=100),
     ] = 100,
     event_service: EventService = event_service_dependency,
 ) -> EventPage:
     """Search / List events."""
-    assert limit > 0
```
chuckbutkus, line 59: Same

openhands/app_server/sandbox/sandbox_router.py
```diff
@@ -44,13 +44,11 @@ async def search_sandboxes(
     ] = None,
     limit: Annotated[
         int,
-        Query(title='The max number of results in the page', gt=0, lte=100),
+        Query(title='The max number of results in the page', gt=0, le=100),
     ] = 100,
     sandbox_service: SandboxService = sandbox_service_dependency,
 ) -> SandboxPage:
     """Search / list sandboxes owned by the current user."""
-    assert limit > 0
```
chuckbutkus, line 52: Same

openhands/app_server/sandbox/sandbox_spec_router.py
```diff
@@ -35,13 +35,11 @@ async def search_sandbox_specs(
     ] = None,
     limit: Annotated[
         int,
-        Query(title='The max number of results in the page', gt=0, lte=100),
+        Query(title='The max number of results in the page', gt=0, le=100),
     ] = 100,
     sandbox_spec_service: SandboxSpecService = sandbox_spec_service_dependency,
 ) -> SandboxSpecInfoPage:
     """Search / List sandbox specs."""
-    assert limit > 0
```
chuckbutkus, line 43: Same

enterprise/server/sharing/shared_conversation_router.py
malhotra5: These were unnecessary because FastAPI handles the validation for this in the header, the assert statement is redundant. However, we were using `lte` instead of `le` in the fastapi header so it wasn't kicking in and the assert statement was being picked up instead

I've fixed it so that fastapi correctly handles validation and the assert statements are no longer required. FastAPI will return 422 responses on failed validation

chuckbutkus (approved):

PR #13516, opened 2026-03-21
WebSocket Gateway

### LINKED ISSUE #13504: Reintroduce single endpoint for handling all WebSocket connections across sandboxes

### Bug Description

In v0, a single endpoint (port 3000) handled all traffic, including WebSocket connections for every sandbox.

In v1, this endpoint no longer supports WebSockets. Instead, the browser is required to connect directly to each sandbox’s WebSocket endpoint. Another related issue is that, by default, these endpoints are exposed as http://localhost:(port), which completely breaks self-hosted deployments.

This new approach introduces several issues. It relies on non-standard ports that may be blocked by ISPs and increases the complexity of firewall and reverse proxy configurations. Because sandboxes are created dynamically, these network rules must also be dynamic.

As a result, we need to reintroduce a single endpoint responsible for handling all WebSocket connections across sandboxes.

I guess this likely requires modifications to the v1_router.

I should also mention that installing or upgrading OpenHands has always been a pain, and it often takes days of chasing bugs that don’t even trigger warnings or error messages in the logs, even for someone that considers himself an experienced OpenHands developer.

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Steps to Reproduce

_No response_

CLI (uv tool install)

_No response_

### OpenHands Version

_No response_

### Version Confirmation

- [ ] I have confirmed this bug exists on the LATEST version of OpenHands

### Model Name

_No response_

### Operating System

None

### Browser (if using web UI)

_No response_

### Logs and Error Messages

_No response_

_No response_

kripper:
## Update

The latest version of the WebSocket Gateway + other V1 routers is included in:
https://github.com/OpenHands/OpenHands/pull/14909

---

Similar PRs:
- https://github.com/OpenHands/OpenHands/pull/13551
- https://github.com/OpenHands/OpenHands/pull/12591

## Problem

In v0, a single endpoint (port 3000) handled all traffic, including WebSocket connections for every sandbox.

In v1, this endpoint no longer supports WebSockets. Instead, the browser is required to connect directly to each sandbox’s WebSocket endpoint.

This new approach introduces several issues: It relies on non-standard ports that may be blocked by ISPs and increases the complexity of firewall and reverse proxy configurations. Because sandboxes are created dynamically, these network rules must also be dynamic.

Issue: https://github.com/OpenHands/OpenHands/issues/13504

## Solution

This PR introduces a WebSocket gateway for v1 messaging that:

1. handles all conversation WebSocket connections through a single endpoint running on the same port as the web app, and
2. routes each connection to the appropriate sandbox agent server endpoint.

## Setup

- Run OH with env var `ENABLE_WEBSOCKET_GATEWAY=true`. This enables the websocket gateway on the backend.
- Set `VITE_ENABLE_WEBSOCKET_GATEWAY="true"` in `frontend/.env` and rebuild the front end. This makes the websockets use relative urls and use the same host and port of the web app.
- Depending on your setup, if your sandboxes are somehow not reachable by the OH app-server (see OH logs), set the env var `SANDBOX_CONTAINER_URL_PATTERN="http://127.0.0.1:{port}"` (change to the IP of your dockerd host).

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves https://github.com/OpenHands/OpenHands/issues/13504

## Release Notes

### DIFF
--- a/frontend/src/utils/websocket-url.ts
+++ b/frontend/src/utils/websocket-url.ts
@@ -79,6 +79,15 @@ export function buildWebSocketUrl(
     return null;
   }
 
+  // Check if centralized WebSocket gateway is enabled via environment variable
+  const useGateway = import.meta.env.VITE_ENABLE_WEBSOCKET_GATEWAY === "true";
+
+  if (useGateway && conversationUrl) {
+    // Use relative URL for the centralized gateway endpoint
+    // The browser will automatically resolve this to wss://<current-host>/ws/events/<id>
+    return `/ws/events/${conversationId}`;
+  }
+
   const baseHost = extractBaseHost(conversationUrl);
   const pathPrefix = extractPathPrefix(conversationUrl);
 
--- a/frontend/.env.sample
+++ b/frontend/.env.sample
@@ -5,4 +5,5 @@ VITE_MOCK_SAAS="false" # Simulate SaaS mode in development (true or false)
 VITE_USE_TLS="false" # Use HTTPS/WSS for backend connections (true or false)
 VITE_FRONTEND_PORT="3001" # Port to run the frontend application
 VITE_INSECURE_SKIP_VERIFY="false" # Skip TLS certificate verification (true or false)
+VITE_ENABLE_WEBSOCKET_GATEWAY="true" # See: https://github.com/OpenHands/OpenHands/issues/13504
 # VITE_GITHUB_TOKEN="" # GitHub token for repository access (used in some tests)
--- a/openhands/app_server/v1_router.py
+++ b/openhands/app_server/v1_router.py
@@ -1,3 +1,5 @@
+import os
+
 from fastapi import APIRouter
 
 from openhands.app_server.app_conversation import app_conversation_router
@@ -22,3 +24,16 @@ router.include_router(sandbox_spec_router.router)
 router.include_router(user_router.router)
 router.include_router(webhook_router.router)
 router.include_router(web_client_router.router)
+
+# WebSocket gateway support (optional, enabled via environment variable)
+WEBSOCKET_GATEWAY_ENABLED = os.getenv('ENABLE_WEBSOCKET_GATEWAY', 'false').lower() in (
+    'true',
+    '1',
+)
+
+if WEBSOCKET_GATEWAY_ENABLED:
+    from openhands.app_server.websocket_router import router as websocket_router
+
+    # Create a separate router for WebSocket endpoints (no prefix, mounted at /ws)
+    websocket_routes = APIRouter()
+    websocket_routes.include_router(websocket_router)
--- a/openhands/app_server/websocket_router.py
+++ b/openhands/app_server/websocket_router.py
@@ -1,0 +1,175 @@
+"""WebSocket router for OpenHands App Server.
+
+This module provides a centralized WebSocket endpoint that proxies connections
+to the appropriate sandbox's agent_server. This allows WebSocket connections
+to work through a proxy without needing to expose individual sandbox URLs.
+"""
+
+import asyncio
+import logging
+from typing import Annotated
+from uuid import UUID
+
+import websockets
+import websockets.exceptions
+from fastapi import APIRouter, Query, WebSocket
+from starlette.websockets import WebSocketState
+
+logger = logging.getLogger(__name__)
+
+router = APIRouter()  # No prefix - will be mounted at /ws in listen.py
+
+
+async def _proxy_websocket_messages(
+    client_ws: WebSocket,
+    agent_server_ws: websockets.WebSocketClientProtocol,
+):
+    """Proxy messages bidirectionally between client and agent server."""
+
+    async def client_to_agent():
+        try:
+            while True:
+                data = await client_ws.receive_text()
+                await agent_server_ws.send(data)
+        except Exception as e:
+            logger.error(f'Error proxying client->agent: {e}')
+
+    async def agent_to_client():
+        try:
+            while True:
+                data = await agent_server_ws.recv()
+                if client_ws.client_state == WebSocketState.CONNECTED:
+                    await client_ws.send_text(data)
+        except websockets.exceptions.ConnectionClosed:
+            logger.info('Agent server WebSocket closed')
+        except Exception as e:
+            logger.error(f'Error proxying agent->client: {e}')
+
+    # Run both proxies concurrently
+    tasks = [
+        asyncio.create_task(client_to_agent()),
+        asyncio.create_task(agent_to_client()),
+    ]
+
+    try:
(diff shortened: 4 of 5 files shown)


kripper:
BTW, I’d enable the gateway by default to save people time when trying to run OH in a self-hosted, non-localhost environment.

kripper:
Similar PR for K8s deployments: https://github.com/OpenHands/OpenHands/pull/12591

gpothier:
Hi, I was working on a very similar feature at the same time (see draft PR https://github.com/OpenHands/OpenHands/pull/13551). The main difference is that I also added a proxy for the sandbox's VSCode instance. Do you think you could add that feature to your PR? Alternatively, I can create another PR with just that feature once your PR is merged.

kripper:
> Hi, I was working on a very similar feature at the same time (see draft PR #13551). The main difference is that I also added a proxy for the sandbox's VSCode instance. Do you think you could add that feature to your PR? Alternatively, I can create another PR with just that feature once your PR is merged.

Nice. We could have two PR using the same configs, style, etc.
I will take a look.

frontend/src/utils/websocket-url.ts
```diff
@@ -79,6 +79,15 @@ export function buildWebSocketUrl(
...
 
+  // Check if centralized WebSocket gateway is enabled via environment variable
+  const useGateway = import.meta.env.VITE_ENABLE_WEBSOCKET_GATEWAY === "true";
+
+  if (useGateway && conversationUrl) {
+    // Use relative URL for the centralized gateway endpoint
+    // The browser will automatically resolve this to wss://<current-host>/ws/events/<id>
+    return `/ws/events/${conversationId}`;
+  }
+
```
tofarr, line 90: I think we should handle this on the server side, rather than having the server produce URLs which are incorrect and then fixing these on the client. I suggest updating with a check to a new variable `self.proxy_sandbox_connections` - and replace the URL if it is present: https://github.com/OpenHands/OpenHands/blob/505095d50afd50d999ab4faf2f4f2679dd02914a/openhands/app_server/app_conversation/live_status_app_conversation_service.py#L507

I think this proxy variable would make sense as a boolean added to the global config, and then injected into the service. (https://github.com/OpenHands/OpenHands/blob/505095d50afd50d999ab4faf2f4f2679dd02914a/openhands/app_server/app_conversation/live_status_app_conversation_service.py#L1841)

We could then use this in listen.py to determine whether to apply your websocket_router. 

I suggest defining the variable in the global config like:
`   proxy_sandbox_connections: bool = Field(default=False, description='Whether connections to the sandboxes should be proxied through the app server')`

This variable would be definable from the environment as `OH_PROXY_SANDBOX_CONNECTIONS=1`

This way we have centralized documentable configuration.

tofarr (changes requested):
See my comment

kripper:
@tofarr Oh, I'm very sorry, I somehow didn't receive a notification for your comments.
Thanks for reviewing it. I really wasn't expecting it; otherwise, I would have updated this PR with my recent changes.

To be honest, I'm a bit confused about the direction of the project and whether there is interest from OpenHands LLC in "fixing" this router issue.
On the other hand, I've seen that this issue was fixed in agent-canvas, and I was expecting it to replace the frontend in OpenHands.

But now it seems that OpenHands LLC might be limiting the product, possibly removing Docker support and forcing users to connect to agent servers manually by copying and pasting the endpoint URL into the "Add backend" dialog in agent-canvas.

It seems they want to turn OpenHands into some kind of Cursor or OpenCode clone for basic users and remove features that currently differentiate OpenHands.

I have no idea what's going on, and I'm just waiting to understand what will happen with the project.

rbren:
@kripper thanks for the comment--hopefully I can provide some more clarity

We are rearchitecting around a pattern of "progressive complexity" (see [here](https://benefuture.miraheze.org/wiki/Progressive_complexity) [here](https://medium.com/design-bootcamp/progressive-complexity-9eaab9da1f01) or [here](https://www.lorenstew.art/blog/progressive-complexity-manifesto))

Right now, OpenHands is too complicated for the average user. It not only requires docker--it does docker-in-docker, and there's a ton of complexity around managing containers, networking, etc. Everyone is used to Claude Code, where everything "just works" out of the box, with very few dependencies.

An explicit goal with [the Agent Canvas initiative](https://github.com/OpenHands/OpenHands/issues/14374) is to fix this--to give users an experience that "just works".

That said, our best users are power users like yourself. So we're setting up levers that let you get increasing amounts of value out of OpenHands by making increasing investment:
* You can install OpenHands on a remote VM, and put it behind a URL
* You can install OpenHands on _multiple_ VMs, and manage all those instances from one Agent Canvas
* Each VM can have its own set of secrets, skills, MCP servers, etc--it's a nice logical grouping for adhering to the principle of least privilege
* Once you're on a VM, you can set up webhooks to ingest events from GitHub, Slack, Jira etc to trigger agents
* You can write python code [using our SDK](https://github.com/OpenHands/software-agent-sdk) and trigger that on a schedule or webhook event
* On a VM or on your laptop, you'll be able to set it up so that agents run inside containers, instead of on the host machine

And there are almost certainly more interesting setups we haven't even imagined yet.

All of these are _advanced_ setups--the kinds of things that only ~10-20% of our user base will want, and which will require some up front investment. But we have always appealed to power users by virtue of being open source and hyper-customizable. It's important to me that we continue supporting power users and providing them with an open platform they can hack on.

tl;dr:
* we need a product that "just works" for simple use cases
* we need a product that can be heavily extended and customized for complex use cases

DFP-:
Not OP but the reason I'm interested in the websocket gateway is less for the power user persona and more for frontloading any complexity into the configuration so that afterwards any non-technical user on the lan can go to openhands.local and it Just Works.

That one to many model seemed like one of the more compelling benefits of a service like openhands when I first heard of it. There's a lot out there already for agentic development that's 1-2 terminal commands on your machine then you're good to go.

kripper:
> Not OP but the reason I'm interested in the websocket gateway is less for the power user persona and more for frontloading any complexity into the configuration so that afterwards any non-technical user on the lan can go to openhands.local and it Just Works.
> 
> That one to many model seemed like one of the more compelling benefits of a service like openhands when I first heard of it. There's a lot out there already for agentic development that's 1-2 terminal commands on your machine then you're good to go.

Exactly. The value we see in OH is that it provides centralized management of AI infrastructure (sandboxes, configs, etc.).

kripper:
Just wanted to comment that I have been adding other missing V1 routers. I will update this PR after testing.

kripper:
The latest version of the WebSocket Gateway + other V1 routers is included in:
https://github.com/OpenHands/OpenHands/pull/14909

PR #13524, opened 2026-03-22
fix(frontend): make conversation header sticky at narrow window widths

gpothier:
## Problem

At window widths ≤ 1024px the conversation layout switches to a mobile
variant. In this layout `ConversationMain` does not carry `overflow-hidden`
(unlike the desktop path), which can allow content to overflow the
`#root-outlet` scroll container. When that happens the header — containing
the conversation title and the tab buttons (Planner / Changes / Code /
Terminal / etc.) — scrolls out of view, forcing the user to scroll all the
way back to the top to interact with it.

## Fix

Add `sticky top-0` to the header div so it remains pinned to the top of
the scroll container whenever overflow-driven scrolling occurs.

- `sticky top-0` — keeps the header visible during scroll
- `bg-base` — solid background so scrolled content doesn't bleed through
- `z-10` — layers the header above the scrolled content
- `shrink-0` — prevents the header from shrinking in the flex layout

On desktop (> 1024px) the layout already prevents overflow, so `sticky`
is a no-op there.

## Testing

Reproduced at ~800 px window width with a long conversation and no side
pane active: header previously scrolled away, now remains visible.

### DIFF
--- a/frontend/src/components/features/conversation/conversation-main/conversation-main.tsx
+++ b/frontend/src/components/features/conversation/conversation-main/conversation-main.tsx
@@ -32,7 +32,7 @@ export function ConversationMain() {
     <div
       className={cn(
         isMobile
-          ? "relative flex-1 flex flex-col"
+          ? "flex-1 flex flex-col"
           : "h-full flex flex-col overflow-hidden",
       )}
     >
@@ -80,12 +80,7 @@ export function ConversationMain() {
           className={cn(
             "transition-all duration-300 ease-in-out overflow-hidden",
             isMobile
-              ? cn(
-                  "absolute bottom-4 left-0 right-0 top-160",
-                  isRightPanelShown
-                    ? "h-160 translate-y-0 opacity-100"
-                    : "h-0 translate-y-full opacity-0",
-                )
+              ? cn(isRightPanelShown ? "h-160 opacity-100" : "h-0 opacity-0")
               : getDesktopTabPanelClass(isRightPanelShown),
           )}
           style={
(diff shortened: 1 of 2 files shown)


neubig (dismissed):

gpothier:
Here are screenshots, I think they illustrate the issue and the fix quite well, but if you still need videos please tell and I'll grab some.

Without the fix, header disappears when scrolling down:
imagen


With the fix, header visible at all times:
imagen

gpothier:
I just added an additional fix: with only the original fix, the header still disappeared if there was no "secondary pane" (Changes / Code / etc.) selected. With this second fix, the header stays visible in both cases.

hieptl:
Hello @gpothier,

While testing the changes locally, I noticed an issue:

**Issue (blocking):**
When the side panel is not active and the page is scrolled, the header behaves as expected and remains sticky, which is great. However, there appears to be a small gap around the header.

Please refer to the image below for more details.

issue

Thank you very much! 🙏

gpothier:
Hi @hieptl, thanks for taking the time to check this out! I had not tested at a yet smaller width. New fix committed! Tested at different widths, works fine!

hieptl:
Hello @gpothier,

I’m still encountering the issue mentioned in my previous comment. Please see the image below for additional context.

image

Thank you very much! 🙏

hieptl:
Hello @gpothier,

I’m still encountering the issue mentioned in my previous comment.

Thank you very much! 🙏

alterdaemon:
Hello, quick question. Why was it closed? This issue is still persisting in the latest version of OpenHands and it renders it unusable, for example, on mobile browsers. Not important?

PR #13532, opened 2026-03-22
Add Support for Switching Models via llama.cpp Compatible API

### LINKED ISSUE #13518: [Feature]: Add Support for Switching Models via llama.cpp Compatible API
### Is there an existing feature request for this?

#### **Summary**
Enable OpenHands to utilize the `llama-server` (OpenAI-compatible) endpoint for switching between the main model and the critic model dynamically.

#### **Motivation**
`llama-server` is the de-facto standard for LLM inference on consumer hardware. It offers a lightweight, efficient way to run multiple models (e.g., a main model and a critic model) in a single server instance. Currently, Openhands only supports switching to an OpenAI-compatible *endpoint*, but it requires a separate, independent server for each model. Integrating `llama-server` would allow a single server to handle both the main and critic logic, reducing resource overhead and streamlining the setup for local users.

#### **Current Workaround**
Users currently have to run two separate `llama-server` instances (one for the main model and one for the critic model) to achieve this separation, or use a different inference backend.

#### **Technical Details**
The `llama-server` provides a REST API that accepts a `model` parameter in the payload.

**Example Request:**
```bash
curl http://localhost:8080/v1/chat/completions \
  -d '{
    "model": "ggml-org/gemma-3-4b-it-GGUF:Q4_K_M",
    "messages": [{"role": "user", "content": "Hello!"}]
  }'
```

**How it works:**
1.  The server starts in "router mode" (e.g., `--models-max 1`, full command: `llama-server --models-max 1 --models-dir ../`).
2.  It loads the requested model into memory.
3.  If a request is sent for a different model name, the server automatically unloads the previous one and loads the new one.

Ref:
https://huggingface.co/blog/ggml-org/model-management-in-llamacpp
https://github.com/OpenHands/OpenHands/issues/2075
https://github.com/OpenHands/OpenHands-CLI/issues/68
https://github.com/OpenHands/software-agent-sdk/issues/416

### Proposed Solution

Modify the OpenHands logic for switching to an OpenAI-compatible critic model to accept a **single `llama-server` instance** that can handle both the main and critic model types. When the main model completes a task, Openhands should send a request to this specific endpoint, passing the model name for the critic, allowing the server to switch models on the fly.

### Alternatives Considered

_No response_

### Priority / Severity

### Estimated Scope

Unknown - Not sure about the technical complexity

### Feature Area

Agent / AI behavior

### Technical Implementation Ideas (Optional)

_No response_

### Additional Context

_No response_

maveilo:
## Summary of PR
Enables OpenHands to utilize the llama-server (OpenAI-compatible) endpoint for switching between the main model and the critic model dynamically.

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #13518

## Release Notes


maveilo:
@chuckbutkus 
Please review.

maveilo:
@OpenHands Do a /codereview-roasted on this PR. Don’t assume anything, investigate, and when you have feedback, post it as a comment to this PR.

maveilo:
## Code Review - /codereview-roasted

### Taste Rating
🟡 **Acceptable** - Works but could be cleaner

---

### Linus-Style Analysis

1. **[openhands/llm/llm.py, Line 275]** - **Special Case**: The use of `'model_override' in dir()` is an unusual pattern. Using `dir()` to check for variable existence is fragile and relies on Python's scoping behavior. A cleaner approach would be to check if the variable exists using `locals()` or initialize `model_override = None` at the start of the function.

   ```python
   # Current (fragile):
   if 'model_override' in dir():
       kwargs['model'] = model_override
   
   # Better approach:
   model_override = kwargs.pop('model', None)
   if model_override:
       kwargs['model'] = model_override
   ```

2. **[openhands/llm/llm.py, Lines 269-275]** - **Inconsistent Variable Scope**: The code pops `model` from kwargs and stores it in a local variable, but this variable's existence is checked later using `dir()`. This works due to Python's closures but makes the code harder to follow. Consider restructuring to make the flow more explicit.

3. **[tests/unit/llm/test_llm.py]** - **Test Pattern**: The tests mock `litellm_completion` and `litellm_acompletion` directly and assert on call arguments. This is appropriate for unit testing the wrapper, but ensure the tests also verify the behavior when the model override is NOT provided (to confirm it doesn't interfere with the default behavior).

**[STYLE NOTES]**

- The debug logging added is helpful for troubleshooting.
- The pattern is consistently applied across `llm.py`, `async_llm.py`, and `streaming_llm.py`.

---

### Verdict

✅ **Worth merging** - The core logic is sound and addresses the feature request correctly. The implementation allows llama-server to switch models dynamically by passing a `model` kwarg, which is exactly what the issue requested.

### Key Insight

The implementation correctly solves the real problem: enabling a single llama-server instance to handle both main and critic models by dynamically overriding the model parameter. The `dir()` pattern is unusual but won't break in practice due to Python's closure behavior.

### Suggestions

1. Consider using `kwargs.pop('model', None)` pattern more consistently to avoid the `dir()` check
2. The tests look good, but you may want to add a test for the negative case (calling without model kwarg doesn't affect default behavior)

enyst (commented):
Thank you for the PR! I don’t think it would work right now, on V1, since the PR modifies the legacy V0 code. Just a heads up, we are about to remove this code actually (please see the heading on top of the files)

The current version of OpenHands is based on software-agent-sdk which has the LLM implementation.

PR #13534, opened 2026-03-22
app_server: add env guard for default SetTitleCallbackProcessor

andyelka-creator:
Summary:\nThis PR adds an opt-out env flag for automatic insertion of SetTitleCallbackProcessor in LiveStatusAppConversationService._start_app_conversation.\n\n- New env var: OH_DISABLE_SET_TITLE_PROCESSOR\n- Truthy values: 1, true, yes\n- Default behavior is unchanged (processor is still auto-added)\n\nWhy:\nIn self-hosted and remote-gateway setups, operators sometimes need to disable title-side effects while keeping the rest of conversation startup unchanged. This keeps current defaults intact and adds an explicit runtime escape hatch.\n\nTests:\n- Added assertion that default flow still includes SetTitleCallbackProcessor\n- Added test that setting OH_DISABLE_SET_TITLE_PROCESSOR=true prevents automatic insertion\n

### DIFF
--- a/LOCAL_PATCH_TRACKING.md
+++ b/LOCAL_PATCH_TRACKING.md
@@ -1,0 +1,28 @@
+# Local Patch Tracking (andyelka-creator fork)
+
+This fork is used as a temporary integration carrier for reproducible fixes and documentation before upstream merge.
+
+## Why this exists
+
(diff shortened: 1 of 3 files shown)


mamoodi:
Hi there. Is this PR ready for review? The PR description is lacking a bit. Are you able to show what this is fixing/adding?

Issue #13541, opened 2026-03-23
[Bug]: V0 conversations have additional trailing characters in metadata.json


justynoh:
### Is there an existing issue for the same bug?

- [x] I have searched existing issues and this is not a duplicate.

### Bug Description

Since upgrading to OpenHands 1.5.0, i've been seeing more JSONDecodeError from logs, causing the agent to get stuck in v0 conversations. From investigation, it seems that there are additional trailing characters in metadata.json  that duplicate some number of ending characters of the actual JSON file.

### Expected Behavior

`metadata.json` should contain valid JSON, for example:

```json
{"conversation_id":"6516112b496947fd8c028a5d1d160446","selected_repository":"<repository>","user_id":null,"selected_branch":"main","git_provider":"github","title":"<title>","last_updated_at":"2026-03-23T02:59:22.445043Z","trigger":"gui","pr_number":[],"created_at":"2026-03-06T06:22:32.162408Z","llm_model":null,"accumulated_cost":0.0,"prompt_tokens":0,"completion_tokens":0,"total_tokens":0,"sandbox_id":null,"conversation_version":null,"public":null}
```

### Actual Behavior

`metadata.json` contains invalid JSON with duplicated trailing characters, for example:

```json
{"conversation_id":"6516112b496947fd8c028a5d1d160446","selected_repository":"<repository>","user_id":null,"selected_branch":"main","git_provider":"github","title":"<title>","last_updated_at":"2026-03-23T02:59:22.445043Z","trigger":"gui","pr_number":[],"created_at":"2026-03-06T06:22:32.162408Z","llm_model":null,"accumulated_cost":0.0,"prompt_tokens":0,"completion_tokens":0,"total_tokens":0,"sandbox_id":null,"conversation_version":null,"public":null}rsion":null,"public":null}
```

### Steps to Reproduce

Happens intermittently while there is an in-progress conversation.

### OpenHands Installation Method

Local GUI (Docker web interface)

### OpenHands Version

1.5.0

### Version Confirmation

- [x] I have confirmed this bug exists on the LATEST version of OpenHands

### Operating System

None

### Logs and Error Messages

```
01:49:06 - openhands:INFO: conversation_stats.py:73 - Saved conversation stats
01:49:06 - OBSERVATION
[...]

PR #13543, opened 2026-03-23
fix: recover corrupted V0 metadata.json files

CMLKevin:
## Summary\n- recover V0 conversation metadata when duplicated trailing characters are appended after a valid JSON object\n- rewrite the repaired metadata back to the file store so future reads stay clean\n- add a regression test covering the duplicated-suffix corruption pattern from #13541\n\n## Testing\n- uv run --group test pytest -q tests/unit/storage/conversation/test_file_conversation_store.py\n\nFixes #13541

### DIFF
--- a/openhands/storage/conversation/file_conversation_store.py
+++ b/openhands/storage/conversation/file_conversation_store.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 from dataclasses import dataclass
+from json import JSONDecodeError
 from pathlib import Path
 
 from pydantic import TypeAdapter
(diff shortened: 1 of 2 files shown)


nidhishgajjar:

## PR #13543: fix: recover corrupted V0 metadata.json files

### What I reviewed
Adds resilience to `FileConversationStore.get_metadata()` to recover from a known corruption mode where duplicated trailing characters are appended to an otherwise-valid JSON metadata file.

### Findings

**Well done:**
- ✅ **Targeted recovery**: Uses `raw_decode()` to extract the first valid JSON object, then validates that trailing content matches the canonical JSON suffix — confirming the duplication pattern before accepting it
- ✅ **Self-healing**: Writes repaired canonical JSON back to the file store, preventing repeated recovery overhead on subsequent reads
- ✅ **Conservative fallback**: Unknown corruption patterns (where trailing content does not match the canonical suffix) correctly re-raise the `JSONDecodeError` — no silent data loss
- ✅ **Logging**: Warning includes `conversation_id` for observability
- ✅ **Test coverage**: Verifies both correct metadata parsing AND file repair in a single test

**Minor observations:**
- 📝 The `canonical_json.endswith(trailing)` check could theoretically produce false positives (e.g., a single stray `}` matching the canonical ending), but in practice the canonical JSON is always much longer than typical trailing garbage, making this extremely unlikely
- 📝 The `bytes | None` return type for `repaired_json` is clean — callers can distinguish "no repair needed" from "was repaired"

### Summary
Solid defensive fix that gracefully recovers from a specific metadata corruption pattern while self-healing the file. The conservative approach of only accepting known corruption modes prevents incorrect data from being silently accepted.

### Assessment: approve

PR #13562, opened 2026-03-24
fix: user does not immediately appear in org after accepting invite in openhands cloud

hieptl:
## Summary of PR

When users click invitation links from emails, the browser navigates from an external source (email client) to the application. With SameSite=strict, browsers block cookies on cross-site top-level navigations, causing authenticated users to appear unauthenticated.

This resulted in users being unnecessarily redirected through the login flow when accepting organization invitations, even though they were already logged in.

### Solution

- Revert get_cookie_samesite() to return 'strict' in production
- Simplify GET /accept endpoint to always redirect with invitation_token
- Add POST /accept endpoint for authenticated invitation acceptance
- Create InvitationAcceptModal component with loading state
- Add useAcceptInvitation mutation hook
- Add i18n translations for 15 languages
- Integrate invitation modal into root layout

This approach works with SameSite='strict' cookies because:

- Cross-site navigation (email link click) redirects to home with token
- Same-origin POST request (from modal) sends cookies normally

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/bc15c60d-0104-44cf-b430-1c2c2de45bb5

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:d0d438b-nikolaik   --name openhands-app-d0d438b   docker.openhands.dev/openhands/openhands:d0d438b
```

### DIFF
--- a/enterprise/server/utils/url_utils.py
+++ b/enterprise/server/utils/url_utils.py
@@ -33,4 +33,9 @@ def get_cookie_samesite() -> Literal['lax', 'strict']:
     # Use 'lax' for local development and staging environments
     # Note: For invitation links from emails, the frontend handles acceptance via
     # an authenticated POST request (same-origin), which works with 'strict' cookies
-    return 'strict'
+    web_url = get_global_config().web_url
+    return (
+        'strict'
+        if web_url and not (IS_FEATURE_ENV or IS_STAGING_ENV or IS_LOCAL_ENV)
+        else 'lax'
+    )


enterprise/server/utils/url_utils.py
```diff
@@ -29,10 +29,8 @@ def get_cookie_domain() -> str | None:
...
-    return (
-        'strict'
-        if web_url and not (IS_FEATURE_ENV or IS_STAGING_ENV or IS_LOCAL_ENV)
-        else 'lax'
-    )
+    # Always use 'lax' to allow cookies on top-level navigations (e.g., clicking links in emails)
+    # 'lax' still protects against CSRF by not sending cookies on cross-site POST requests
+    # Previously 'strict' was used in production, but this blocked cookies when users clicked
+    # invitation links from emails, causing authenticated users to be redirected to login
+    return 'lax'
```
tofarr: This could have ramifications for the runtimes which run untrusted user code in their app servers - it may be possible for them to capture cookies as the browser will now send these to the subdomain

tofarr: With lax, `https://work-1-<some_malicious_runtime_id>.runtime.all-hands.dev` could capture cookies from `https://work-1-<some_benign_runtime>.runtime.all-hands.dev` - this is probably okay as long as we warn users that they should not use any important test credentials in their test servers.

tofarr: I think we probably need input from product management though - if they are okay with accepting the additional risk. (It is why we don't use path based runtimes in SAAS)

tofarr: Since these are on `.runtime.all-hands.dev` they could also capture the auth cookie from `app.all-hands.dev` if samesite is lax! :(

tofarr:
**Suggestion (Thinking out loud):**

* The /accept_invitation endpoint should not require authentication.
* The OrgInvitation should include a user_id which is populated when it is created - this should replace accepted_by_user_id
* Any org invitation which does not have a populated user_id should use the existing auth. (As a fallback)
When accepting an invitation, pull the user_id from the invitation rather than the session


**Downside**

* In theory this means a user could be added to an org even if they have not logged in - merely by somebody visiting the link. (If they had access to the token) Is this acceptable?  Are there security ramifications of adding somebody to an org against their will?

enterprise/server/utils/url_utils.py
hieptl: Hello @tofarr,

I’ve updated the code to revert `get_cookie_samesite()` to return “strict” in the production environment.

Please feel free to refer to my comment below for details on the updated approach. I’d appreciate any further thoughts you may have.

Thank you very much for your feedback! 🙏

hieptl:
Hello @tofarr,

I’ve updated my solution in this pull request and would greatly appreciate your feedback.

### Solution

* Revert `get_cookie_samesite()` to return “strict” in production
* Simplify the `GET /accept` endpoint to always redirect with `invitation_token`
* Add a `POST /accept` endpoint for authenticated invitation acceptance
* Create an `InvitationAcceptModal` component with a loading state
* Add a `useAcceptInvitation` mutation hook
* Include i18n translations for 15 languages
* Integrate the invitation modal into the root layout

### Rationale

This approach works with `SameSite="strict"` cookies because:

* Cross-site navigation (e.g., clicking an email link) redirects to the home page with the token
* A same-origin `POST` request (triggered from the modal) allows cookies to be sent as expected

I would love to hear your thoughts on this approach.

Thank you very much! 🙏

PR #13569, opened 2026-03-24
security: pin litellm<=1.82.6 to mitigate supply chain attack

saurya:
## Summary

This PR pins LiteLLM to `<=1.82.6` to protect against a critical supply chain attack affecting versions 1.82.7 and 1.82.8.

## Security Issue

LiteLLM versions **1.82.7** and **1.82.8** on PyPI contain a malicious `.pth` file (`litellm_init.pth`) that automatically executes a credential-stealing payload whenever the Python interpreter starts — no `import litellm` required.

### What the malicious payload does:
- Collects sensitive data: SSH keys, AWS/GCP/Azure credentials, API keys, environment variables, Kubernetes secrets, crypto wallets, shell history, and more
- Encrypts the data with AES-256 + RSA
- Exfiltrates to an attacker-controlled server (`https://models.litellm.cloud/`)

### Trigger mechanism:
Python's `.pth` files in `site-packages/` are executed automatically on interpreter startup ([Python docs](https://docs.python.org/3/library/site.html)). This means any system that installed the compromised versions is affected, including:
- Local development machines
- CI/CD pipelines
- Docker containers
- Production servers

## Changes

Updated `pyproject.toml` to pin litellm to `<=1.82.6` in both:
- `[project]` dependencies (for UV compatibility)
- `[tool.poetry.dependencies]` (for Poetry)

## Reference

- https://github.com/BerriAI/litellm/issues/24512

## Recommended User Actions

If you have already installed litellm 1.82.7 or 1.82.8:
1. Check for the malicious file: `ls $(python -c "import site; print(site.getsitepackages()[0])")/litellm_init.pth`
2. Downgrade immediately: `pip install litellm==1.82.6`
3. **Rotate ALL credentials** that were present as environment variables or in config files on the affected system

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,7 +50,7 @@ dependencies = [
   "jwcrypto>=1.5.6",
   "kubernetes>=33.1",
   "libtmux>=0.46.2",
-  "litellm!=1.64.4,!=1.67.*,>=1.74.3",
+  "litellm>=1.74.3,<=1.82.6,!=1.64.4,!=1.67.*",  # Pin <=1.82.6 due to supply chain attack in 1.82.7/1.82.8 (BerriAI/litellm#24512)
   "lmnr>=0.7.20",
   "mcp>=1.25",
   "memory-profiler>=0.61",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -50,7 +50,7 @@ dependencies = [
   "jwcrypto>=1.5.6",
   "kubernetes>=33.1",
   "libtmux>=0.46.2",
-  "litellm!=1.64.4,!=1.67.*,>=1.74.3",
+  "litellm>=1.74.3,<=1.82.6,!=1.64.4,!=1.67.*",  # Pin <=1.82.6 due to supply chain attack in 1.82.7/1.82.8 (BerriAI/litellm#24512)
```
enyst, line 53: Looks correct, maybe we can simplify. e.g. like the latest commit to the SDK?

saurya: Actually looks like Robert has one for that already: https://github.com/OpenHands/software-agent-sdk/pull/2558

and one for this repo: https://github.com/OpenHands/OpenHands/pull/13566

PR #13587, opened 2026-03-25
ci: build amd64 only on PRs to skip QEMU emulation

### LINKED ISSUE #13571: Reduce runtime image build times (currently 2-3+ hours)
## Problem

Runtime image builds are taking **2-3+ hours** to complete, which significantly impacts PR review and merge velocity. Recent example: https://github.com/OpenHands/OpenHands/actions/runs/23454689336/job/68245621321?pr=13552

This is especially problematic when dependency changes are made or when base image caching is invalidated (e.g., the recent switch to slim image in https://github.com/OpenHands/OpenHands/commit/7fbb499f037e444ad4625f30984d53f192829bee).

## Root Cause Analysis

The job **"Build Runtime Image (nikolaik/python-nodejs:python3.12-nodejs22, nikolaik)"** builds a **multi-platform Docker image** for both `linux/amd64` and `linux/arm64` architectures. The ARM64 build runs under **QEMU emulation** on an x86_64 runner, which is dramatically slower than native builds.

### Time Breakdown by Build Step

| Step | Architecture | Duration | Description |
|------|-------------|----------|-------------|
| #36 | linux/amd64 | ~82 min | Playwright installation + system packages |
| #37 | linux/arm64 | ~86 min | Same step, but emulated |
| #38 | linux/amd64 | ~105 min | User/group setup + chown operations |
| #39 | linux/arm64 | ~104 min | Same step, emulated |

### Why These Steps Are So Slow

1. **Playwright Installation Step**: This step installs system packages (libnss3, libnspr4, libatk-bridge, etc.), installs Playwright with Chromium browser dependencies, and runs `poetry run playwright install --with-deps chromium`. On ARM64 under QEMU:
   - Package installation involves heavy I/O
   - Playwright downloads and extracts large browser binaries (~150+ MB)
   - Everything is being emulated

2. **User Setup Step**: While seemingly simple (creates users/groups and runs `chown`), this step runs `chown -R` on the entire `/openhands` directory which contains all the previously installed Python packages, micromamba environment, and Playwright browsers. Recursively changing ownership of thousands of files under QEMU emulation is very slow.

### Short-term Improvements

1. **Use native ARM64 runners** - GitHub now offers native ARM64 runners (`ubuntu-24.04-arm`) which would be significantly faster than QEMU emulation
2. **Split builds** - Build AMD64 and ARM64 separately on their native architectures, then combine manifests
3. **Cache more aggressively** - The build appears to not be hitting cache effectively for these steps
4. **Reduce chown scope** - Instead of `chown -R` on large directories, set ownership during file creation

### Medium-term Improvement (Recommended)

**Create a pre-built base image with Playwright included:**

1. Create a separate workflow that builds a `python-nodejs-playwright` base image (infrequently, maybe weekly or on Playwright version bumps)
2. Push it to `ghcr.io/openhands/python-nodejs-playwright:python3.12-nodejs22`
3. Use that pre-built image as the base for runtime image builds

This would eliminate the ~86 minute Playwright installation from every PR build, potentially reducing build time from ~3 hours to ~30-40 minutes.

### Investigation Notes

- `nikolaik/python-nodejs` does not have a Playwright variant
- Microsoft's official Playwright images (`mcr.microsoft.com/playwright:v*`) have Node.js but lack Python
- Playwright Python images (`mcr.microsoft.com/playwright/python:v*`) have Python + Playwright but no Node.js
- Building and maintaining our own combined image is likely the best path forward

## Acceptance Criteria

- [ ] Runtime image builds complete in under 1 hour for typical PR changes
- [ ] Multi-platform support (amd64/arm64) is maintained
- [ ] Browser functionality (Playwright) continues to work

## Additional Context

The Dockerfile for the runtime image is dynamically constructed, which adds to the maintenance burden. Long-term, simplifying this architecture would also be beneficial.

/cc @xingyaoww

raymyers:
## Summary of PR

PR Docker builds were taking **3+ hours** on cache miss because the `linux/arm64` leg runs under QEMU emulation on x86_64 Blacksmith runners (~6.7× slower on CPU-bound steps like `poetry install`).

This PR centralizes the platform decision in the `define-matrix` job and restricts PR builds to `linux/amd64` only, while keeping `linux/amd64,linux/arm64` for main/tag/dispatch builds.

**Changes:**

- **`define-matrix` job**: Each matrix entry now carries a `platforms` field (`linux/amd64` for PRs, `linux/amd64,linux/arm64` otherwise). A top-level `platforms` output is also emitted for non-matrix consumers.
- **`ghcr_build_app` job**: Now `needs: define-matrix` and passes `-p ${{ needs.define-matrix.outputs.platforms }}` to `build.sh`.
- **`ghcr_build_runtime` job**: Passes `-p ${{ matrix.base_image.platforms }}` to the `build.sh --dry` invocation, which flows through the dry-run JSON → `DOCKER_PLATFORM` → `useblacksmith/build-push-action`.
- **`containers/build.sh`**: New `-p <platform>` flag overrides the default platform selection. Without `-p`, existing behavior is preserved.

| Trigger | Platforms |
|---------|-----------|
| `pull_request` | `linux/amd64` |
| `push` (main/tags) | `linux/amd64,linux/arm64` |
| `workflow_dispatch` | `linux/amd64,linux/arm64` |

## Change Type

## Checklist

## Fixes

Resolves #13571

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:0a14bd7-nikolaik   --name openhands-app-0a14bd7   docker.openhands.dev/openhands/openhands:0a14bd7
```

### DIFF
--- a/.github/workflows/ghcr-build.yml
+++ b/.github/workflows/ghcr-build.yml
@@ -33,28 +33,33 @@ jobs:
     runs-on: blacksmith
     outputs:
       base_image: ${{ steps.define-base-images.outputs.base_image }}
+      platforms: ${{ steps.define-base-images.outputs.platforms }}
     steps:
       - name: Define base images
         shell: bash
         id: define-base-images
         run: |
           if [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
-            json=$(jq -n -c '[
-                { image: "nikolaik/python-nodejs:python3.12-nodejs22-slim", tag: "nikolaik" }
+            platforms="linux/amd64"
+            json=$(jq -n -c --arg platforms "$platforms" '[
+                { image: "nikolaik/python-nodejs:python3.12-nodejs22-slim", tag: "nikolaik", platforms: $platforms }
               ]')
           else
-            json=$(jq -n -c '[
-                { image: "nikolaik/python-nodejs:python3.12-nodejs22-slim", tag: "nikolaik" },
-                { image: "ubuntu:24.04", tag: "ubuntu" }
+            platforms="linux/amd64,linux/arm64"
+            json=$(jq -n -c --arg platforms "$platforms" '[
+                { image: "nikolaik/python-nodejs:python3.12-nodejs22-slim", tag: "nikolaik", platforms: $platforms },
+                { image: "ubuntu:24.04", tag: "ubuntu", platforms: $platforms }
               ]')
           fi
           echo "base_image=$json" >> "$GITHUB_OUTPUT"
+          echo "platforms=$platforms" >> "$GITHUB_OUTPUT"
 
   # Builds the OpenHands Docker images
   ghcr_build_app:
     name: Build App Image
     runs-on: blacksmith-4vcpu-ubuntu-2204
     if: "!(github.event_name == 'push' && startsWith(github.ref, 'refs/tags/ext-v'))"
+    needs: define-matrix
     permissions:
       contents: read
       packages: write
@@ -82,7 +87,7 @@ jobs:
       - name: Build and push app image
         if: "!github.event.pull_request.head.repo.fork"
         run: |
-          ./containers/build.sh -i openhands -o ${{ env.REPO_OWNER }} --push
+          ./containers/build.sh -i openhands -o ${{ env.REPO_OWNER }} --push -p ${{ needs.define-matrix.outputs.platforms }}
 
   # Builds the runtime Docker images
   ghcr_build_runtime:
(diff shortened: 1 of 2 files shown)


raymyers:
Superseded by #13590 which includes this change plus additional Dockerfile optimizations.

PR #13628, opened 2026-03-27
fix: prevent V0 conversation creation due to settings race condition

xingyaoww:
## Summary of PR

Clicking "New Conversation" can accidentally create a V0 (legacy) conversation instead of V1 due to a race condition in settings loading.

**Root cause**: The `useCreateConversation` hook checks `settings?.v1_enabled` to decide between V1 and V0 APIs. If the user clicks "New Conversation" before the settings query resolves, `settings` is `undefined`, causing `!!undefined?.v1_enabled` to evaluate to `false` — silently routing through the legacy V0 code path. This is compounded by `DEFAULT_SETTINGS.v1_enabled` being `false` on the frontend while the backend defaults to `true` everywhere.

**Two changes to fix this:**

1. **`frontend/src/services/settings.ts`** — Change `DEFAULT_SETTINGS.v1_enabled` from `false` to `true`

   Aligns the frontend fallback with the backend default (`Settings` model, `DEFAULT_V1_ENABLED`, org creation, and saas settings store all default to `true`). This covers the case where the settings API returns 404 for a new user and the frontend falls back to `DEFAULT_SETTINGS`.

2. **`frontend/src/hooks/mutation/use-create-conversation.ts`** — Change `!!settings?.v1_enabled` to `settings?.v1_enabled ?? true`

   The nullish coalescing operator falls back to `true` (matching backend default) when settings haven't loaded yet, while still respecting an explicit `false` from orgs that genuinely have V1 disabled.

## Demo Screenshots/Videos

N/A — this is a timing-dependent race condition.

## Change Type

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:2dfd9d9-nikolaik   --name openhands-app-2dfd9d9   docker.openhands.dev/openhands/openhands:2dfd9d9
```


frontend/src/hooks/mutation/use-create-conversation.ts
```diff
@@ -54,7 +54,8 @@ export const useCreateConversation = () => {
         plugins,
       } = variables;
 
-      const useV1 = !!settings?.v1_enabled && !createMicroagent;
+      // Default to V1 (matching backend default) when settings haven't loaded yet
```
xingyaoww: @OpenHands can we force this to wait AFTER settings are load?

xingyaoww: Fixed in 7ff36c337. Now uses `queryClient.ensureQueryData` inside the mutation to wait for the settings query to resolve before checking `v1_enabled`. If settings are already cached it returns immediately; if the query is in-flight it waits for completion. Falls back to `DEFAULT_SETTINGS` on fetch failure (e.g., 404 for new user), matching the existing `useSettings` hook behavior.

frontend/src/hooks/mutation/use-create-conversation.ts
```diff
@@ -54,7 +56,20 @@ export const useCreateConversation = () => {
         plugins,
       } = variables;
 
-      const useV1 = !!settings?.v1_enabled && !createMicroagent;
+      // Wait for settings to be loaded before deciding V0 vs V1
```
aivong-openhands, line 59: could we create some unit tests to document the race condition bug in the codebase?

xingyaoww: @OpenHands please try to do this, revert all the fe changes first, implement a test that would reproduce the race condition. Then add these FE changes back and tell me if these changes actually fixes the original issue.

xingyaoww: Done in 2951b3db5. Here's what I did:

**1. Reverted all FE changes** to the buggy main-branch state

**2. Wrote a race condition test** (`use-create-conversation-race-condition.test.tsx`) that mocks `useSettings` to return `{ data: undefined }` (simulating settings not yet loaded) and asserts that V1 API should be called.

**3. Ran the test against the buggy code — it FAILED ✅** confirming the bug:
```
× should use V1 API when settings have not loaded yet (race condition)
  AssertionError: expected "createConversation" to be called at least once
```
The original code does `!!settings?.v1_enabled` which evaluates to `false` when settings is `undefined`, silently routing to V0.

**4. Re-applied the FE fixes** (`ensureQueryData` approach + `DEFAULT_SETTINGS.v1_enabled = true`)

**5. Ran the tests — all 4 pass ✅** confirming the fix works:
- `should use V1 API even when settings are not yet cached (race condition scenario)` ✅
- `should use V1 API when settings fetch fails (falls back to defaults)` ✅
- `should use V1 API when settings explicitly have v1_enabled: true` ✅
- `should use V0 API when v1_enabled is explicitly false` ✅

Also updated existing tests (`new-conversation`, `repo-connector`, `task-card`, `use-create-conversation`) that were broken by the V1-default change to use `V1ConversationService` instead of the legacy `ConversationService`.

xingyaoww:
@OpenHands pls fix the failing CI in this repo

PR #13637, opened 2026-03-27
chore: Add sdk to mypy checking and fix the resulting errors

raymyers:
## Summary of PR

We use pre-commit to run linting and mypy checks, which [does not include](https://github.com/pre-commit/mirrors-mypy) the full project dependencies. This PR adds the OpenHands SDK and fixes 55 resulting type errors. 

* Some transitive dependencies are excluded in mypy.ini to manage the number of changes.
* I tried to avoid behavior changes, adding ignores when there was no obvious way.
* `openhands-agent-server` was not added because the transitive dependencies introduce many more errors. In fact the pip package includes the whole server when we only want the client.

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [X] Refactor
- [X] Other (dependency update, docs, typo fixes, etc.)

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:bf19caa-nikolaik   --name openhands-app-bf19caa   docker.openhands.dev/openhands/openhands:bf19caa
```


dev_config/python/mypy.ini
```diff
@@ -14,3 +14,11 @@ exclude = (third_party/|enterprise/)
...
 [mypy-openhands.memory.condenser.impl.*]
 disable_error_code = override
+
+[mypy-openai.*]
+follow_imports = skip
+ignore_missing_imports = True
+
+[mypy-litellm.*]
+follow_imports = skip
+ignore_missing_imports = True
```
raymyers, line 24: Transitive dependencies of OpenHands SDK, which introduce more type errors we can address separately.

openhands/app_server/app_conversation/app_conversation_service_base.py
```diff
@@ -405,14 +398,18 @@ async def maybe_setup_git_hooks(
             project_dir: Project root directory (repo root when a repo is selected).
         """
         command = 'mkdir -p .git/hooks && chmod +x .openhands/pre-commit.sh'
-        result = await workspace.execute_command(command, project_dir)
-        if result.exit_code:
+        pre_commit_command_result = await workspace.execute_command(
```
raymyers, line 415: Preventing one var name from getting multiple types.

openhands/app_server/config.py
```diff
@@ -220,9 +216,12 @@ def config_from_env() -> AppServerConfig:
...
             # Google Cloud storage configuration
-            config.event = GoogleCloudEventServiceInjector(
-                bucket_name=os.environ.get('FILE_STORE_PATH')
-            )
+            bucket_name = os.environ.get('FILE_STORE_PATH')
+            if not bucket_name:
+                raise ValueError(
+                    'FILE_STORE_PATH environment variable is required for Google Cloud storage'
+                )
+            config.event = GoogleCloudEventServiceInjector(bucket_name=bucket_name)
```
raymyers, line 228: Behavior change, GCP store will not attempt to init with a missing bucket name.

openhands/integrations/provider.py
```diff
@@ -187,13 +187,19 @@ async def _get_latest_provider_token(
         self, provider: ProviderType
     ) -> SecretStr | None:
         """Get latest token from service"""
+        if not self.REFRESH_TOKEN_URL:
```
raymyers, line 190: Behavior change, when REFRESH_TOKEN_URL is missing, _get_latest_provider_token logs warning and returns None rather than try to call a missing URL. Raising an exception would be an alternative.

openhands/llm/retry_mixin.py
```diff
@@ -22,17 +22,17 @@
 class RetryMixin:
     """Mixin class for retry logic."""
 
-    def retry_decorator(self, **kwargs: Any) -> Callable:
+    def retry_decorator(self, num_retries: int, **kwargs: Any) -> Callable:
```
raymyers, line 25: I believer every user of retry_decorator includes this arg, so no change.

openhands/mcp/client.py
```diff
@@ -85,17 +84,17 @@ async def connect_http(
...
             if isinstance(server, MCPSHTTPServerConfig):
-                transport = StreamableHttpTransport(
+                streamable_transport = StreamableHttpTransport(
                     url=server_url,
                     headers=headers if headers else None,
                 )
+                self.client = Client(streamable_transport, timeout=timeout)
             else:
-                transport = SSETransport(
+                sse_transport = SSETransport(
```
raymyers, line 93: Preventing var with same name having multiple types.

openhands/resolver/interfaces/bitbucket_data_center.py
```diff
@@ -243,10 +243,12 @@ def reply_to_comment(self, pr_number: int, comment_id: str, reply: str) -> None:
 
     def branch_exists(self, branch_name: str) -> bool:
         url = f'{self._get_repo_api_base()}/branches'
-        params = {'filterText': branch_name, 'limit': 1}
         try:
             response = httpx.get(
-                url, headers=self.headers, params=params, verify=httpx_verify_option()
+                url,
+                headers=self.headers,
+                params={'filterText': branch_name, 'limit': '1'},
```
raymyers, line 250: making `limit` val a string because it's a url query param.

openhands/runtime/builder/remote.py
```diff
@@ -68,8 +68,8 @@ def build(
                 files=files,
                 timeout=30,
             )
-        except httpx.HTTPError as e:
-            if e.response.status_code == 429:
+        except httpx.HTTPStatusError as e:
```
raymyers, line 71: When it's not a HTTPStatusError (doesn't have response), we just raise, so might as well not catch it.

openhands/runtime/impl/local/local_runtime.py
```diff
@@ -116,8 +116,12 @@ def check_dependencies(code_repo_path: str, check_browser: bool) -> None:
         except Exception:
             raise ValueError('tmux is not properly installed or available on the path.')
         pane = session.active_pane
-        pane.send_keys('echo "test"')
-        pane_output = '\n'.join(pane.cmd('capture-pane', '-p').stdout)
+
+        if pane:
```
raymyers, line 120: Behavior change: If pane is missing (which probably doesn't happen?), we no longer implicitly raise.

raymyers:
From MyPy html report:

Before: 13.42% imprecise

After: 11.88% imprecise

openhands/runtime/utils/bash.py
```diff
@@ -252,9 +252,10 @@ def initialize(self) -> None:
         _initial_window.kill()
 
         # Configure bash to use simple PS1 and disable PS2
-        self.pane.send_keys(
-            f'export PROMPT_COMMAND=\'export PS1="{self.PS1}"\'; export PS2=""'
-        )
+        if self.pane:
+            self.pane.send_keys(
+                f'export PROMPT_COMMAND=\'export PS1="{self.PS1}"\'; export PS2=""'
```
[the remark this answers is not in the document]
raymyers: Great point but that's for a different PR and this is V0 code so the safest change that makes the check pass is probably best.

openhands/security/grayswan/analyzer.py
```diff
@@ -95,8 +95,8 @@ async def _get_session(self) -> aiohttp.ClientSession:
         """Get or create HTTP session."""
         if self.session is None:
             self.session = aiohttp.ClientSession(
-                timeout=aiohttp.ClientTimeout(total=self.timeout),
-                headers={
+                timeout=aiohttp.ClientTimeoput(total=self.timeout),  # type: ignore[operator]
```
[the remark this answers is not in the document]
raymyers: Good call. Fixed.

openhands/llm/retry_mixin.py
```diff
@@ -67,10 +67,11 @@ def before_sleep(retry_state: Any) -> None:
             retry=(
                 retry_if_exception_type(retry_exceptions)
             ),  # retry only for these types
+            # TODO: Type errors here probably mean the defaults are being ignored.
             wait=wait_exponential(
-                multiplier=retry_multiplier,
-                min=retry_min_wait,
-                max=retry_max_wait,
+                multiplier=retry_multiplier,  # type: ignore[arg-type]
+                min=retry_min_wait,  # type: ignore[arg-type]
```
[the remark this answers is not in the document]
raymyers: That should be handled in a separate PR, that would be a behavior change and this has too many other changes.

openhands/app_server/app_conversation/app_conversation_service_base.py
```diff
@@ -435,9 +441,11 @@ async def maybe_setup_git_hooks(
...
         # Make the pre-commit hook executable
-        result = await workspace.execute_command(f'chmod +x {PRE_COMMIT_HOOK}')
-        if result.exit_code:
-            _logger.error(f'Failed to make pre-commit hook executable: {result.stderr}')
+        chmod_result = await workspace.execute_command(f'chmod +x {PRE_COMMIT_HOOK}')
+        if chmod_result.exit_code:
+            _logger.error(
+                f'Failed to make pre-commit hook executable: {chmod_result.stderr}'
+            )
             return
```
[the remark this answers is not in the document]
raymyers: Good idea, but avoiding restructuring code in this PR.

PR #13639, opened 2026-03-28
fix(frontend): prevent "Unknown event" shown for actions with empty d…

### LINKED ISSUE #12865: "Unknown event" shown before file read operations complete

sjathin:
## Summary of PR

Fix false-positive "Unknown event" text appearing in the chat UI for actions that legitimately have no expandable details (e.g. file `view` commands).

**Root cause:** `getEventContent()` used a falsy check (`details || "Unknown event"`) on the `details` return value, so an empty string `""` was treated as missing and replaced with "Unknown event".

**Changes:**
- Return `details` as-is instead of falling back to "Unknown event" when it's an empty string (both v1 and legacy `get-event-content.tsx`)
- Add a lenient fallback branch for action-like events that have an `action.kind` but fail the strict `isActionEvent()` type guard (missing `tool_name`/`tool_call_id`), so the UI shows the action kind instead of "Unknown event"
- Add tests covering both cases

### Before:

https://github.com/user-attachments/assets/127b1315-1b75-4e63-8d3e-ad8e1066e62b

### After:

https://github.com/user-attachments/assets/5f27bd0d-c8ce-4448-9a13-5fd8b5d96aff

## Change Type

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #12865 

## Release Notes


hieptl (approved):
Hello @sjathin,

I have reviewed, approved, and merged it. Great work—much appreciated.

Thank you very much! 🙏

PR #13641, opened 2026-03-28
fix: add /new to slash command autocomplete menu

### LINKED ISSUE #13596: [Feature]: Add /new to the GUI slash command menu
## Problem or Use Case

PR #12786 introduced the `/new` slash command for V1 conversations. It creates a new conversation while keeping the same runtime/sandbox, which is a very powerful workflow improvement.

Right now this command is hard to discover because it is handled directly in the chat UI (`chat-interface.tsx`) and does not appear in the GUI slash command menu. Users only learn about it if they read release notes, PRs, or documentation.

## Proposed Solution

Add `/new` to the GUI slash command menu.

Suggested menu text:
- Command: `/new`
- Sub-text: `Creates a new conversation using the same runtime`

Behavior expectations:
- Show it in contexts where the command is supported (V1 conversations)
- Selecting it from the slash menu should insert `/new` into the chat input like other slash commands
- Keep the existing runtime-sharing behavior introduced in PR #12786

## Alternatives Considered

- Relying on documentation or release notes only: this makes the feature easy to miss
- Showing a toast or hint after users type `/new`: helpful, but still less discoverable than putting it in the slash menu where users already look for commands

## UX
Here's an example of the GUI slash menu where this feature should be displayed.

Image

The slash command menu currently appears to be built from skills in `frontend/src/hooks/chat/use-slash-command.ts`, while `/new` is handled separately in `frontend/src/components/features/chat/chat-interface.tsx`.

A straightforward fix may be to add support for built-in slash commands in the slash command hook/menu so `/new` can be merged with skill-based slash commands and given explicit descriptive sub-text.

## Additional Context

- Introduced by PR #12786
- The value proposition to surface in the menu is that `/new` starts a fresh conversation while preserving the existing runtime/sandbox state
- Suggested sub-text: `Creates a new conversation using the same runtime`

1Ckpwee:
## Summary

Fixes #13596

The `/new` slash command was added in #12786 but was not included in the slash command autocomplete menu, making it hard to discover.

## Changes

- Added `/new` as a built-in slash command entry in `useSlashCommand` hook so it appears in the autocomplete dropdown alongside other commands

## Test plan
- [ ] Type `/` in chat input — `/new` should appear in the dropdown
- [ ] Select `/new` from menu — it should be inserted and execute correctly
- [ ] Existing slash commands still work as expected

### DIFF
--- a/frontend/src/hooks/chat/use-slash-command.ts
+++ b/frontend/src/hooks/chat/use-slash-command.ts
@@ -35,12 +35,29 @@ export const useSlashCommand = (
   const [filterText, setFilterText] = useState("");
   const [selectedIndex, setSelectedIndex] = useState(0);
 
+  // Built-in slash commands that are always available regardless of loaded skills
+  const BUILTIN_SLASH_ITEMS: SlashCommandItem[] = useMemo(
+    () => [
+      {
+        command: "/new",
+        skill: {
+          name: "new",
+          type: "knowledge" as const,
+          content:
+            "Start a new conversation while keeping the runtime alive.",
+          triggers: ["/new"],
+        },
+      },
+    ],
+    [],
+  );
+
   // Build slash command items from skills:
   // - Skills with explicit "/" triggers use those triggers
   // - AgentSkills without "/" triggers get a derived "/<name>" command
   const slashItems = useMemo(() => {
-    if (!skills) return [];
-    const items: SlashCommandItem[] = [];
+    const items: SlashCommandItem[] = [...BUILTIN_SLASH_ITEMS];
+    if (!skills) return items;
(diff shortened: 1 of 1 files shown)


hieptl:
Hello @1Ckpwee,

Thank you very much for your contribution and for creating this pull request!

I wanted to let you know that the issue has already been addressed in an earlier pull request, which has now been merged: [https://github.com/OpenHands/OpenHands/pull/13599](https://github.com/OpenHands/OpenHands/pull/13599)

As a result, I will go ahead and close this pull request.

We truly appreciate your effort and look forward to your future contributions. Thank you again! 🙏

PR #13657, opened 2026-03-30
PLTF-330: Add automatic session_id logging via contextvars

aivong-openhands:
## Summary

This PR introduces automatic `session_id` and `user_id` injection into all log messages within the enterprise server, eliminating the need for manual `extra={"session_id": sid}` in every log call.

### New Files
- `enterprise/server/session_context.py`: Session context management using Python's `contextvars` for request-scoped session tracking

### Modified Files
- `enterprise/server/middleware.py`: Added `SessionContextMiddleware` that extracts session/conversation ID from request paths and sets context
- `enterprise/server/logger.py`: Added `SessionContextFilter` to automatically inject session_id and user_id into all log records
- `enterprise/saas_server.py`: Registered the new middleware

### Cleaned Up Manual Logging
The following files no longer need manual `extra={"session_id": ...}` as session_id is now automatically injected:
- `enterprise/server/saas_nested_conversation_manager.py`
- `enterprise/server/clustered_conversation_manager.py`

Note: `billing.py` uses `checkout_session_id` which is a Stripe checkout session, not a conversation session, so it remains unchanged.

## How It Works

1. **HTTP Requests**: `SessionContextMiddleware` extracts conversation ID from URL paths like `/api/conversations/{id}/...` and sets it in a context variable
2. **Background Tasks**: Use `session_context.scope(session_id=sid, user_id=user_id)` context manager
3. **Callbacks/Workers**: Same pattern - wrap with `session_context.scope()`
4. `SessionContextFilter` (added to all loggers) reads from the context variable and injects `session_id` and `user_id` into log records

## Usage Pattern

```python
from server.session_context import session_context

# For functions that receive sid as parameter:
async def process_conversation(sid: str, user_id: str):
    with session_context.scope(session_id=sid, user_id=user_id):
        logger.info("Processing")  # Automatically includes session_id, user_id
        await do_something()
        logger.info("Done")  # Also includes session_id, user_id
```

## Testing
- [ ] Tested locally with enterprise server
- [ ] Verified JSON logs include session_id and user_id fields

### DIFF
--- a/enterprise/saas_server.py
+++ b/enterprise/saas_server.py
@@ -23,7 +23,10 @@ from server.auth.constants import (  # noqa: E402
 )
 from server.constants import PERMITTED_CORS_ORIGINS  # noqa: E402
 from server.logger import logger  # noqa: E402
-from server.middleware import SetAuthCookieMiddleware  # noqa: E402
+from server.middleware import (  # noqa: E402
+    SessionContextMiddleware,
+    SetAuthCookieMiddleware,
+)
 from server.rate_limit import setup_rate_limit_handler  # noqa: E402
 from server.routes.api_keys import api_router as api_keys_router  # noqa: E402
 from server.routes.auth import api_router, oauth_router  # noqa: E402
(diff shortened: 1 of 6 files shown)


aivong-openhands:
We do currently log session_ids in some places. If those are the only places we care to have session_ids logged, then this PR is unnecessary

PR #13667, opened 2026-03-31
Validate selected branch names before checkout

enyst:
## Summary
- add a shared `ensure_valid_git_branch_name()` helper backed by `git check-ref-format --branch`
- reject invalid branch names in the v1 app-server metadata update path
- quote branch names before the v1 app-server builds `git checkout` shell commands
- add targeted tests for common hosted-git branch names and for quoting a shell-special branch name on checkout

## Scope
I verified that `openhands/runtime/base.py` is explicitly marked **LEGACY V0 CODE** at the top of the file and is not on the `openhands/app_server/` v1 execution path. This PR now intentionally scopes the fix to the v1 app-server codepath only.

## Problem
In the v1 app-server path, `selected_branch` ultimately flowed into shell command construction in `AppConversationServiceBase.clone_or_init_git_repo()`:

```python
checkout_command = f'git checkout {request.selected_branch}'
```

That meant a branch name containing shell-special characters could change how the shell interpreted the checkout command.
For example, `feature>tmp` is accepted by git as a branch name, but in an unquoted shell command it is parsed as output redirection rather than a literal branch argument.

### Passing verification on this branch
```bash
poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files \
  openhands/utils/git.py \
  openhands/app_server/app_conversation/live_status_app_conversation_service.py \
  openhands/app_server/app_conversation/app_conversation_service_base.py \
  tests/unit/utils/test_git.py \
  tests/unit/app_server/test_app_conversation_service_base.py
poetry run pytest tests/unit/app_server/test_app_conversation_service_base.py tests/unit/utils/test_git.py -q
```

```text
47 passed in 9.23s
```

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:67b3822-nikolaik   --name openhands-app-67b3822   docker.openhands.dev/openhands/openhands:67b3822
```

### DIFF
--- a/openhands/utils/git.py
+++ b/openhands/utils/git.py
@@ -27,6 +27,6 @@ def ensure_valid_git_branch_name(branch_name: str) -> None:
         return
 
     raise ValueError(
-        f"Invalid git branch name. Common GitHub/GitLab/Bitbucket "
+        f'Invalid git branch name. Common GitHub/GitLab/Bitbucket '
         f'branch names look like {COMMON_BRANCH_EXAMPLES}.'
     )


openhands/utils/git.py
```diff
@@ -0,0 +1,32 @@
...
+    )
+
+
+def ensure_valid_git_branch_name(branch_name: str) -> None:
+    """Raise ValueError when branch_name is not safe to pass to git checkout."""
+    if is_valid_git_branch_name(branch_name):
+        return
+
+    raise ValueError(
+        f"Invalid git branch name '{branch_name}'. Common GitHub/GitLab/Bitbucket "
```
enyst:
```suggestion
        f"Invalid git branch name. Common GitHub/GitLab/Bitbucket "
```

gpothier:
Looks like this breaks sandbox creation. This PR introduces git check-ref-format validation but the Dockerfile only installed git in the builder stage, not the final runtime stage.

PR #13675, opened 2026-03-31
feat: Add DEPLOYMENT_MODE detection for conditional onboarding

HeyItsChloe:
## Summary of PR

- Add DEPLOYMENT_MODE constant to enterprise/server/constants.py
  - 'cloud' for All-Hands managed domains (app.all-hands.dev, *.all-hands.dev, app.openhands.ai, localhost)
  - 'self_hosted' for customer enterprise deployments
- Expose deployment_mode in config API via FEATURE_FLAGS
- Add DeploymentMode type to frontend option types (optional field)
- Update OnboardingAppMode to use labels: 'cloud', 'self-hosted', 'oss'
- Update getOnboardingAppMode() to return app mode based on deployment_mode
- Update onboarding questions to use 'cloud' instead of 'saas' label for readability

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/0b9aaa50-6e0e-4f43-b0ea-18930d5e9a87

https://github.com/user-attachments/assets/7a8d4056-c8c9-4240-90bf-20d2321d4894

## Change Type

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #(issue)

## Release Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:5d046a0-nikolaik   --name openhands-app-5d046a0   docker.openhands.dev/openhands/openhands:5d046a0
```


enterprise/migrations/versions/107_add_onboarding_completed_to_user.py
```diff
@@ -0,0 +1,31 @@
...
+revision: str = '105'
+down_revision: Union[str, None] = '104'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        'user',
+        sa.Column('onboarding_completed', sa.Boolean(), nullable=True, default=False),
```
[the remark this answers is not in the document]
HeyItsChloe: Only new users should be redirected to onboarding.

enterprise/server/constants.py
```diff
@@ -15,6 +15,34 @@
...
+    Returns:
+        'cloud' for All-Hands managed infrastructure (app.all-hands.dev, etc.)
+        'self_hosted' for enterprise self-hosted deployments (customer domains)
+    """
+    if _is_all_hands_managed_domain(HOST):
+        return 'cloud'
+    return 'self_hosted'
+
+
+DEPLOYMENT_MODE = _get_deployment_mode()
```
HeyItsChloe, line 43: Hey, @tofarr . This PR adds a workaround to distinguish between self-hosted and cloud modes in SaaS—each needs different onboarding questions.

Current approach:
- Added a DEPLOYMENT_MODE constant (in server/constants.py) that infers the mode from WEB_HOST (openhands domains = cloud, company-specific domains = self-hosted)
- The server uses it to conditionally redirect users to onboarding
- It's also exposed to the frontend via the config API's FEATURE_FLAGS

Should I just add a DEPLOYMENT_MODE env var along with this ticket instead? It would be cleaner, but requires updating the charts. I went with the workaround to avoid blocking this PR. But LMK if there is a better approach.

hieptl: **suggestion (blocking):** I believe it would be beneficial to update the deployment repository to include the DEPLOYMENT_MODE environment variable.

However, before proceeding with this change, I recommend reaching out in the #proj-self-hosted Slack channel or consulting with the platform team to confirm if there is an existing method to differentiate between the SaaS and self-hosted environments.

If such a method is already in place, we may not need to implement the DEPLOYMENT_MODE variable.

Thank you! 🙏

enterprise/migrations/versions/107_add_onboarding_completed_to_user.py
hieptl, line 1: **suggestion (blocking):** I suggest merging the latest code from the main branch and resolving database migration version conflicts. Thank you! 🙏

enterprise/server/routes/auth.py
```diff
@@ -462,8 +469,18 @@ async def keycloak_callback(
...
+            logger.info(
+                'Redirecting returning user to onboarding',
+                extra={'user_id': user_id, 'deployment_mode': DEPLOYMENT_MODE},
+            )
         if invitation_token:
-            redirect_url = f'{redirect_url}&invitation_success=true'
+            if '?' in redirect_url:
+                redirect_url = f'{redirect_url}&invitation_success=true'
+            else:
+                redirect_url = f'{redirect_url}?invitation_success=true'
```
hieptl, line 481: **question (blocking):**If a user has not yet completed the onboarding flow but receives an organization invitation and clicks on the invitation link first, which flow will be triggered in this case—the invitation flow or the onboarding flow?

Additionally, if the invitation flow is triggered first, after the user accepts the invitation, will they be redirected to the onboarding page?

Thank you! 🙏

enterprise/server/routes/auth.py
```diff
@@ -549,6 +566,39 @@ async def authenticate(request: Request):
...
+        return True
+
+    # Self-hosted SaaS: only owners go to onboarding
+    if DEPLOYMENT_MODE == 'self_hosted':
+        org_member = await OrgMemberStore.get_org_member_for_current_org(
+            uuid.UUID(user_id)
+        )
+        if org_member:
+            role = await RoleStore.get_role_by_id(org_member.role_id)
+            if role and role.name == ROLE_OWNER:
```
hieptl: **question (blocking):**

I have an organization, Organization A, which is using a self-hosted environment and has three owners.

Based on the current logic, the onboarding form for the self-hosted environment is displayed three times, once for each owner, and the same answers may be submitted to Posthog multiple times.

In this case, should the onboarding form only be displayed once?

@jpelletier1 - Could you please provide clarification on the requirements for this scenario?

Thank you all for your input! 🙏

enterprise/server/constants.py
```diff
@@ -15,6 +15,34 @@
...
+
+def _get_deployment_mode() -> str:
+    """Determine deployment mode based on WEB_HOST.
+
+    Returns:
+        'cloud' for All-Hands managed infrastructure (app.all-hands.dev, etc.)
+        'self_hosted' for enterprise self-hosted deployments (customer domains)
+    """
+    if _is_all_hands_managed_domain(HOST):
+        return 'cloud'
```
hieptl, line 39: **suggestion (blocking):** I would prefer to use the term 'saas' instead of 'cloud,' as the app_mode is also set to 'saas.' We may want to stick with this terminology for the time being until a further decision is made.

Thank you! 🙏

hieptl (changes requested):
Thank you very much for your hard work on this! I’ve left a few suggestions and questions on the pull request. If you have any questions or concerns, please feel free to let me know 🙏

enterprise/server/routes/auth.py
jpelletier1: Display to the first owner only

enterprise/server/constants.py
HeyItsChloe: Using 'saas' instead of 'cloud' would cause confusion. If app_mode is for product type (saas vs oss), while deployment_mode is where the SaaS app is hosted (cloud vs self_hosted), then using 'saas' for both would be unclear when combined (app_mode: "saas" + deployment_mode: "saas"). 

Also, if we remove 'self-hosted' (for any reason) and OnboardingAppMode = "saas" | "oss", its unclear what OnboardingAppMode is representing (should be host/deplyment_mode & not app_mode). https://allhandsai.slack.com/archives/D0A2NTM6GPK/p1773771159202909

`Cloud` is used in a few places in the repo - it is the `APP_TITLE_SAAS`. Is there a plan to standardize naming? If not, I think its best to leave this just so app_mode and host/deployment_mode are not the same values.

enterprise/server/routes/auth.py
HeyItsChloe: The invitation is processed first, then the user is redirected to onboarding (if needed). The invitation is accepted first (during auth callback).

FLOW:
1	User clicks invitation link
2	User logs in via Keycloak
3	Invitation is accepted (user joins org)
4	TOS page shown (if not accepted)
5	Onboarding page shown (if onboarding_completed=False)
6	Final redirect to / with invitation_success=true
 
The redirect URL will be /onboarding?invitation_success=true

enterprise/server/constants.py
jlav: There isn't an existing flag that I'm aware of, and having a centralized flag for it would be better than the huge list of env vars that we need to maintain right now to configure self-hosted deployments.

HeyItsChloe: Gotcha - I’ll stick with the current implementation for now until a better approach comes to light.

enterprise/server/routes/auth.py
```diff
@@ -462,8 +476,18 @@ async def keycloak_callback(
             tos_redirect_url = f'{tos_redirect_url}&invitation_success=true'
         response = RedirectResponse(tos_redirect_url, status_code=302)
     else:
+        # User has accepted TOS - check if they need onboarding
+        if await _should_redirect_to_onboarding(user_id, user):
+            redirect_url = f'{web_url}/onboarding'
+            logger.info(
+                'Redirecting returning user to onboarding',
+                extra={'user_id': user_id, 'deployment_mode': DEPLOYMENT_MODE},
+            )
```
hieptl, line 476: **suggestion (blocking):** Missing return before second if in keycloak_callback — Bug enterprise/server/routes/auth.py:484-491 (diff lines 130-146)

The onboarding redirect block is missing an else or return, causing fall-through:

```python
# User has accepted TOS - check if they need onboarding
if await _should_redirect_to_onboarding(user_id, user):
    redirect_url = f'{web_url}/onboarding'
    logger.info(...)
if invitation_token:  # <-- BUG: should be `elif` or preceded by `else`
```

When _should_redirect_to_onboarding returns True AND there's an invitation_token, the redirect_url gets set to /onboarding then immediately modified to /onboarding?invitation_success=true (or worse, the ? check may not handle /onboarding correctly). This is likely unintentional — the invitation query param handling should only apply to the non-onboarding redirect path. This should be an elif or the logic needs restructuring.

Thank you! 🙏

enterprise/migrations/versions/107_add_onboarding_completed_to_user.py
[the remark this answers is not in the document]
hieptl: **suggestion (blocking):** As the bot review noted, default=False only works for SQLAlchemy ORM inserts. Existing rows get NULL. The _should_redirect_to_onboarding function checks if user.onboarding_completed: which treats None as falsy — so existing users would be redirected to onboarding. This is a real issue because all existing users who have already been using the platform will be redirected to onboarding after this migration runs.

Thank you! 🙏

hieptl:
Hello @HeyItsChloe,

I hope you’re doing well. While testing the changes locally, I came across a couple of issues that I wanted to flag:

**Issue 1 (blocking):** When creating a new account, I need to click the “Sign in with GitHub” button twice before successfully signing in. Please refer to the video below for more details.

https://github.com/user-attachments/assets/f3899082-d515-4b4a-9fca-3ff2126341cf

**Issue 2 (blocking):** If I manually remove the `/onboarding` path from the browser URL, I’m able to return to the home page and bypass the onboarding flow. Although the onboarding form reappears upon signing in again, it’s worth noting that most users typically do not log out after signing in, as our product automatically reauthenticates sessions when they expire.

This behavior could prevent us from consistently capturing the necessary onboarding information from end users. Given that, it seems important to ensure users complete the onboarding flow before accessing the product.

@jpelletier1 — I’d really appreciate your thoughts on this scenario as well.

https://github.com/user-attachments/assets/11ad8fff-37cb-4881-8c6c-b569d8fde2e6

**Issue 3 (blocking):**

**Reproduction steps:**

* Step 1: Invite a new member to an organization.
* Step 2: Copy the invitation link and open it in the browser.
* Step 3: Sign in to the product (currently requires clicking the “Sign in” button twice).
* Step 4: Accept the Terms of Service (TOS).

**Current behavior:**

* After accepting the TOS, the user is not redirected to the onboarding page.

**Expected behavior:**

* The user should be redirected to the onboarding page after completing the TOS step.

https://github.com/user-attachments/assets/9d9f8fde-e134-4d2a-8a2d-656587c72ee1

Thank you very much! 🙏

openhands/app_server/web_client/web_client_models.py
```diff
@@ -1,11 +1,32 @@
...
+
+
+# This can be removed / replaced when a DeploymentMode (or similar) env var is created.
+def _get_deployment_mode() -> DeploymentMode | None:
+    """Get deployment mode based on OH_WEB_HOST environment variable."""
+    web_host = os.getenv('OH_WEB_HOST', os.getenv('WEB_HOST', '')).strip()
+    if not web_host:
+        return None
+    if (
+        web_host == 'app.all-hands.dev'
```
hieptl: **suggestion (blocking):** If I understand correctly, we should avoid referencing enterprise-specific code or terminology within OSS code.

Hello @enyst, @tofarr — could you please confirm if my understanding is correct?

Thank you! 🙏

enterprise/migrations/versions/107_add_onboarding_completed_to_user.py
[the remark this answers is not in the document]
HeyItsChloe: Rather than updating the migration to make the column non-nullable and backfilling True for existing users, I'd prefer to handle NULL, False, and True explicitly. To avoid setting onboarding_completed = True for users who never went through the flow

Updated  _should_redirect_to_onboarding with the missing check:
```
if user.onboarding_completed:  # True → already completed
    return False

if user.onboarding_completed is None:  # NULL → existing user, skip onboarding
    return False

# Only False (new users) continues to deployment mode checks
```

enterprise/server/routes/auth.py
hieptl: Your understanding may not be entirely accurate. I’ve recorded a video demonstrating this flow and shared it in this comment: https://github.com/OpenHands/OpenHands/pull/13675#issuecomment-4209059823.

I’ll go ahead and resolve this thread for now—please refer to the link above for more details.

Thank you! 🙏

openhands/app_server/web_client/web_client_models.py
HeyItsChloe: Gotcha. Is there a different method of distinguishing between cloud & self hosted that doesn't rely on the web_host env var?

hieptl: You can continue using `WEB_HOST`. However, enterprise-specific code (such as functions, classes, and related logic) should be placed within the enterprise folder rather than in the OSS codebase.

hieptl: It looks like SaaS domains are being specified within the OSS code, which may not align with our guidelines.

That said, it would be best to wait for further guidance from @enyst and @tofarr on this.

Thank you! 🙏

enterprise/server/constants.py
```diff
@@ -15,6 +15,33 @@
...
+
+# _is_all_hands_managed_domain() can be removed/replaced when a self-hosted specific
+# env var is created (e.g is_self_hosted` or `deployment_mode`)
+def _is_all_hands_managed_domain(host: str) -> bool:
+    """Check if the host is an All-Hands managed domain."""
+    return (
+        host == 'app.all-hands.dev'
+        or host == 'app.openhands.ai'
+        or host.endswith('.all-hands.dev')
+        or host == 'localhost'
```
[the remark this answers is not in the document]
hieptl: **suggestion (blocking):** This feedback from all-hands-bot makes sense to me, so I’ve reopened the thread

Thank you! 🙏

openhands/app_server/web_client/web_client_models.py
enyst: Out of curiosity, is `AppMode` not useful for the purpose here?

openhands/app_server/web_client/web_client_models.py
```diff
@@ -1,11 +1,32 @@
...
+        or web_host == 'app.openhands.ai'
+        or web_host.endswith('.all-hands.dev')
+        or web_host.endswith('.openhands.ai')
+        or web_host == 'localhost'
+    ):
+        return 'cloud'
```
enyst: Does `localhost` correspond to `cloud`?

hieptl: Hello @enyst,

The all-hands-ai bot shared similar feedback here: 

[https://github.com/OpenHands/OpenHands/pull/13675#discussion_r3022041787](https://github.com/OpenHands/OpenHands/pull/13675#discussion_r3022041787). I’ve reopened the thread to continue the discussion.

Thank you! 🙏

openhands/app_server/web_client/web_client_models.py
enyst: Honestly, the definitions in the entire `web_client_models.py` look very enterprise-y to me, e.g. `enable_linear`, `hide_billing`. Maybe we could even move it to `/enterprise` if it’s unused, or minimally used?

That said, precisely because it’s already enterprise in reality… I don’t think the new enterprise-y definition in this PR makes a lot of difference to the overall UX/DX. Sorry, it’s a bit messy.

@openhands Look in the codebase for uses of this file, and understand the PR and its deployment mode. Can you give us a better location for these hardcoded URLs? Be careful to look mainly in V1, which is on the execution path of openhands/app_server, not in legacy V0 where you can. Respond directly on github.

enyst: I traced the V1 path here, and I think there *is* a better home than `web_client_models.py` for the hardcoded host/domain check.

What I found on the V1 execution path:

- `openhands/app_server/v1_router.py` includes `web_client_router`
- `openhands/app_server/web_client/web_client_router.py` serves `GET /web-client/config`
- that calls `config.web_client.get_web_client_config()`
- the default implementation is `openhands/app_server/web_client/default_web_client_config_injector.py`
- `web_client_models.py` is then just the response schema / feature-flags model used by that injector + router

So in V1, `web_client_models.py` is not really business logic; it is mostly the schema for the shared frontend contract. That also explains why it already contains enterprise-ish flags like `enable_linear`, `hide_billing_page`, etc.: the frontend consumes one `web-client/config` shape regardless of OSS vs enterprise.

Because of that, I would *not* move the whole file to `enterprise/` right now.

But I also would *not* keep the managed-domain detection in the model. The better location is:

1. **Best immediate location:** `openhands/app_server/web_client/default_web_client_config_injector.py`
   - it already reads env vars and constructs `WebClientFeatureFlags`
   - `deployment_mode` is just another derived feature flag for `/web-client/config`
   - this keeps `web_client_models.py` as schema-only

2. **If we want an even cleaner split:** extract a tiny helper like `openhands/app_server/web_client/deployment_mode.py`, and call it from the injector.

3. **If we want enterprise to own it completely later:** override `AppServerConfig.web_client` with an enterprise-specific `WebClientConfigInjector`.
   - that would fit the V1 DI model much better than putting V1 `/web-client/config` behavior in `enterprise/server/constants.py`
   - `enterprise/server/constants.py` is not actually on the `app_server` request path for `/web-client/config`

Also, on the earlier `AppMode` question: I don’t think `AppMode` is enough for this specific onboarding case. In V1 it is still effectively `saas` vs `oss`, while the onboarding behavior here needs to distinguish **cloud vs self-hosted within saas**.

So my recommendation would be: **keep the schema in OSS, move the host matching out of `web_client_models.py`, and place it in the V1 web-client injector layer (or a helper right next to it).**

_This reply was generated by an AI assistant (OpenHands) on behalf of the user._

HeyItsChloe: Thanks @enyst - I moved the logic to a separate web_client_deployment_mode.py file per your suggestion #2, but I need to keep the @model_validator in the model.

The issue is that in SaaS mode, AppServerConfig is loaded via the SDK's from_env() function, which parses env vars directly into Pydantic models via validate_json(). This completely bypasses DefaultWebClientConfigInjector and _get_feature_flags().

The @model_validator(mode='after') is the only hook that runs regardless of how the model is created (direct instantiation or JSON parsing). Without it, SaaS deployments would get deployment_mode=null.

This can be replaced with a dedicated DEPLOYMENT_MODE env var in the future.

openhands/app_server/web_client/web_client_models.py
HeyItsChloe: Removed

openhands/app_server/web_client/web_client_models.py
enyst: Thank you!

I honestly don't know how we can do better at this time. I'm fine with the PR, FWIW.

hieptl:
Hello @HeyItsChloe,

I hope you’re doing well. While testing the changes locally, I came across a few issues that I wanted to share:

**Issue 1 (blocking):**
After creating a new account and successfully completing the onboarding flow, I’m still able to access the onboarding page by directly entering its URL in the browser. Ideally, users should be redirected to the home page if they attempt to revisit the onboarding page after completing the flow. Please refer to the video below for additional context.

https://github.com/user-attachments/assets/15d45588-ded0-43d4-9cd4-31517eac7c0c

**Issue 2 (blocking):**
I’m currently unable to create an organization via an API call.

**Reproduction Steps:**

1. Start the SaaS application in your local environment.
2. Create a new account and complete the onboarding process.
3. Navigate to `/settings/api-keys` and generate a new API key.
4. Use the generated API key to call the following endpoint.

```
POST http://localhost:3030/api/organizations

Headers:
Authorization: Bearer <API_KEY>

Body:
{
  "name": "Hiep Test 1",
  "contact_name": "Hiep Test 1",
  "contact_email": "hieptest1@gmail.com"
}
```

**Current Behavior:**

```
{
  "error": "BearerTokenError"
}
```

**Expected Behavior:**

* The request should succeed and create a new organization with the provided data.

Please refer to the image below for more details.

> **Note:** This issue is currently blocking further testing of this pull request.

issue2-1

Additionally, this issue does not occur when running the latest code from the main branch and following the same reproduction steps. Please refer to the image below for further details.

main

**Issue 3 (blocking):**
If I manually remove the `/onboarding` path from the browser URL, I’m able to access the home page and bypass the onboarding flow. Although the onboarding form reappears upon signing in again, most users typically remain signed in, as sessions are automatically reauthenticated.

This behavior could prevent us from consistently capturing the required onboarding information. It may be important to enforce completion of the onboarding flow before allowing access to the main product.

https://github.com/user-attachments/assets/11ad8fff-37cb-4881-8c6c-b569d8fde2e6

I understand this issue has already been discussed on Slack, but I’m noting it here as well for tracking purposes.

Thank you very much! 🙏

hieptl (approved):
Thank you! 🙏

PR #13681, opened 2026-03-31
chore(docker): remove system pip from app image (venv-only)

aivong-openhands:
Goal: reduce attack surface + stop carrying/scanning a second pip install in the app image.

This is **related to** #13640 (pip CVE fix): #13640 pins/upgrades pip to address CVE-2025-8869; this PR is a follow-up hardening step that *removes system pip entirely* if we don't need it at runtime.

What this does:
- Removes *system* pip (site-packages + `/usr/local/bin/pip*` + `ensurepip/_bundled` pip wheels).
- Does **not** upgrade/pin pip (leaves the venv untouched).

Why:
- The image already prepends `${VIRTUAL_ENV}/bin` to `PATH`, so `pip` used by the entrypoint (e.g. `pip install openhands-ai[third_party_runtimes]`) is the venv pip.

Tradeoff:
- `python3 -m pip` and `/usr/local/bin/pip*` will no longer work inside the container.

Merge guidance:
- Only merge if we confirm nobody relies on system pip inside the app image.

_(This PR was created/updated by an AI assistant (OpenHands) on behalf of the user.)_

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:af3c2a6-nikolaik   --name openhands-app-af3c2a6   docker.openhands.dev/openhands/openhands:af3c2a6
```

### DIFF
--- a/containers/app/Dockerfile
+++ b/containers/app/Dockerfile
@@ -73,11 +73,9 @@ ENV VIRTUAL_ENV=/app/.venv \
 
 COPY --chown=openhands:openhands --chmod=770 --from=backend-builder ${VIRTUAL_ENV} ${VIRTUAL_ENV}
 
-# Pin pip to a known-good version (reproducible builds) and fix CVE-2025-8869.
+# Remove system pip from the image (leave venv pip intact).
 # The runtime uses the venv pip because PATH is prefixed with ${VIRTUAL_ENV}/bin.
-ARG PIP_VERSION=26.0.1
-RUN python -m pip install --no-cache-dir "pip==${PIP_VERSION}" && \
-    sudo /usr/local/bin/python3 - <<'PY'
+RUN sudo /usr/local/bin/python3 - <<'PY'
 import ensurepip
 import shutil
 import sysconfig


containers/app/Dockerfile
```diff
@@ -73,6 +73,35 @@ ENV VIRTUAL_ENV=/app/.venv \
...
+for p in [bin_dir / "pip", bin_dir / "pip3", *bin_dir.glob("pip3.*")]:
+    p.unlink(missing_ok=True)
+
+bundled = Path(ensurepip.__file__).parent / "_bundled"
+if bundled.exists():
+    for whl in bundled.glob("pip-*.whl"):
+        whl.unlink(missing_ok=True)
+PY
+
+
```
raymyers, line 104: Just on a basic level I think this is off the rails. It's an addition not a subtraction and it's got embedded Python code in a Dockerfile.

I've used venv-based Dockerfiles, that's not how it's done.

aivong-openhands:
Closing as this is more complexity than expected. Also this doesn't seem like a standard way to remove system pip

containers/app/Dockerfile
raymyers: Something like this would be simpler if we're just removing files, but I think keeping the original double upgrade is probably best.
```
# Remove system-level pip

USER root
RUN rm -rf /usr/local/lib/python3.*/dist-packages/pip*
USER openhands
```

aivong-openhands: Thank you, I was curious about an example!

PR #13696, opened 2026-04-02
Add contributors section to README

jamiechicago312:
Adds a contributors section at the end of the README with an SVG image showing all contributors.

---
Preview of Production
image

image

---

*This PR was created by an AI assistant (OpenHands).*

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,19 @@ If you need help with anything, or just want to chat, [come find us on Slack](ht
 
 <hr>
 
+### Thank You to Our Contributors
+
 <div align="center">
-  <strong>Trusted by engineers at</strong>
+
+[![OpenHands Contributors](https://assets.openhands.dev/readme/openhands-openhands-contributors.svg)](https://github.com/OpenHands/OpenHands/graphs/contributors)
+
+</div>
+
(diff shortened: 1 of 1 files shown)


jamiechicago312:
Replied in another comment but yeah I think we can. It's very manual today
but I don't see why it has to be long term.

Also this is not ready for review right now. But any feedback as I continue
to refund it is welcome 😁

enyst:
> Replied in another comment but yeah I think we can. It's very manual today but I don't see why it has to be long term.

You're right it won't be manual. My hesitation is different: I looked at the page and it's... immense. For repos with much fewer contributors this way of giving credit works, but for us I'm not sure 🤔 

Sorry for posting on a draft, happy to see it later, and thank you for trying this!

jamiechicago312:
Re: immense
I think we can find a way 🤝 🤞 😊

enyst (approved):
Frankly, this looks OKayish!

I’d keep the PR open for a day or so, for others’ feedback.

enyst:
Let's merge. If we discover we don't like it, it's just a readme, we can iterate.

Thank you for the care for these things!

PR #13709, opened 2026-04-02
docs: update README CLI docs link

xsfX20:
## Summary
- update the README CLI docs link to the current canonical terminal usage page
- avoid the legacy `run-openhands/cli-mode` path in the main repo entrypoint
- keep the change scoped to one documentation link only

## Validation
- `curl -I -L https://docs.openhands.dev/openhands/usage/cli/terminal` returns 200

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ Define agents in code, then run them locally, or scale to 1000s of agents in the
 The CLI is the easiest way to start using OpenHands. The experience will be familiar to anyone who has worked
 with e.g. Claude Code or Codex. You can power it with Claude, GPT, or any other LLM.
 
-[Check out the docs](https://docs.openhands.dev/openhands/usage/run-openhands/cli-mode) or [view the source](https://github.com/OpenHands/OpenHands-CLI)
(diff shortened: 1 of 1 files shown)


xsfX20:
The current diff only updates the README CLI docs link from the stale `usage/run-openhands/cli-mode` path to the current `usage/cli/terminal` page.

I checked the head commit check-runs and the only failing item is `Python Tests on Linux (3.12)`, while the rest of the workflow surface passed. Since this PR does not touch Python code or test files, this looks unrelated to the README-only change on my side.

I also tried to rerun the failed job earlier, but GitHub refused the rerun for this workflow. If the test failure is transient or workflow-side, I’m happy to recheck and follow up once it can be rerun.

xsfX20:
Follow-up update: I rebased this docs-only branch onto the latest `main` and force-pushed the refreshed head as `5b1afc2e2`.

The actual README diff is unchanged in intent: it still only updates the stale CLI docs link. This rebase is just to pick up the newest base-line changes and re-run CI from a current branch state, since the earlier failing Python test did not line up with the README-only diff.

PR #13712, opened 2026-04-02
fix(app_server): fall back to default sandbox spec for file reads

xsfX20:
## Summary
- keep `read_conversation_file` working when a conversation references an older sandbox spec that is no longer present
- align file-read behavior with the existing V1 agent-server context fallback to the default sandbox spec
- add router tests for both the fallback path and the fail-closed path when no default spec is available

## Testing
- `git diff --check`
- `python3 -m py_compile openhands/app_server/app_conversation/app_conversation_router.py tests/unit/app_server/test_app_conversation_router.py`
- attempted: `uv run python -m pytest -q tests/unit/app_server/test_app_conversation_router.py -k "default_sandbox_spec or unavailable"` (environment bootstrap was still provisioning dependencies in this checkout)


xsfX20:
Follow-up update: CI flagged a formatter-only diff in the new router test, so I pushed `743ae9ae` to apply the repo's `ruff-format` output.

No behavior change in this follow-up; it only matches the formatting expected by the `Lint python` job.

xsfX20:
Follow-up update: the failing unit test was asserting a Docker-only hostname rewrite unconditionally. I pushed `4d5124c3` to make the assertion use `replace_localhost_hostname_for_docker()` so the test matches the runtime behavior on both Docker and non-Docker runners.

This follow-up is still test-only; the file-read fallback behavior itself is unchanged.

xsfX20:
Follow-up update: the Python test failure turned out to be in the new test itself, not in the file-read fallback. The mocked file handle was returning text, while `read_conversation_file` reads bytes and then decodes.

I pushed `e856e4a0` to make the mocked file content bytes-backed, and re-ran the two relevant router tests locally with proxy vars unset:
- `uv run python -m pytest -q tests/unit/app_server/test_app_conversation_router.py -k "uses_default_sandbox_spec_when_original_is_missing or returns_empty_string_when_default_sandbox_spec_is_unavailable"`
- result: `2 passed, 12 deselected`

Behavior is still unchanged; this follow-up only fixes the test harness to match the runtime read path.

PR #13735, opened 2026-04-03
fix(runtime): fix CJK UTF-8 files misidentified as binary by binaryornot

### LINKED ISSUE #13517: [Bug]:  OHEditor `view` command fails on valid UTF-8 markdown files with dense CJK characters

### Bug Description

`OHEditor.view()` raises `FileValidationError("File appears to be binary...")` when viewing valid UTF-8 `.md` files that contain dense CJK (Chinese/Japanese/Korean) characters. The file is perfectly valid UTF-8 text, but `binaryornot.check.is_binary()` in `validate_file()` incorrectly classifies it as binary, blocking the view command.

### Expected Behavior

`view` command should successfully display valid UTF-8 markdown files regardless of the language/script used.

### Actual Behavior

```
FileValidationError: File validation failed for /tmp/xxx.md:
  File appears to be binary and this file type cannot be read or edited by this tool.
```

The view command never reads the file. `validate_file()` rejects it at the `is_binary()` check.

### Steps to Reproduce

```python
from openhands_aci.editor.editor import OHEditor
import tempfile

# 1. Create a valid UTF-8 markdown file with dense Chinese content (>= ~1KB)
with tempfile.NamedTemporaryFile(suffix='.md', delete=False, mode='w', encoding='utf-8') as f:
    f.write("中文测试内容。" * 50)  # ~1050 bytes of valid UTF-8 Chinese text
    path = f.name

# 2. Try to view the file
editor = OHEditor()
editor(command="view", path=path)
# => raises FileValidationError: "File appears to be binary..."
```

Root cause call chain:
```
OHEditor.__call__(command='view', path=...)
  → OHEditor.view(path)
    → OHEditor.validate_file(path)
      → binaryornot.check.is_binary()   # returns True (FALSE POSITIVE)
      → raise FileValidationError("File appears to be binary...")
```

`binaryornot` uses `chardet` internally. When a UTF-8 file contains dense multi-byte CJK characters (>= ~1KB with no ASCII dilution), the high-byte ratio exceeds its threshold, causing a false positive.

| Dense CJK bytes | `is_binary()` result |
|---|---|
| < ~1000 bytes | `False` (correct) |
| >= ~1050 bytes | `True` (incorrect) |

Development workflow

_No response_

### OpenHands Version

openhands-aci 0.3.3

### Version Confirmation

### Model Name

N/A (not model-dependent)

### Operating System

MacOS

### Browser (if using web UI)

_No response_

### Logs and Error Messages

```python
from binaryornot.check import is_binary
import tempfile

with tempfile.NamedTemporaryFile(suffix='.md', delete=False, mode='w', encoding='utf-8') as f:
    f.write("中文测试内容。" * 50)
    path = f.name

# File is perfectly readable as UTF-8
content = open(path, encoding='utf-8').read()
assert "中文测试内容" in content  # ✅ Valid UTF-8

# But binaryornot says it's binary
print(is_binary(path))  # True ← FALSE POSITIVE
```

Image

**Affected users**: Anyone working with CJK (Chinese/Japanese/Korean) text files. This is a significant i18n issue.

**Environment**: `binaryornot==0.4.4`, `chardet==5.2.0`, Python 3.9+

Prithvi1994:
## Summary

Fixes #13517

`binaryornot` uses byte-frequency heuristics to detect binary files. These heuristics incorrectly classify valid UTF-8 text files containing dense CJK (Chinese/Japanese/Korean) characters as binary, causing `OHEditor.view()` and `read()` to return `ERROR_BINARY_FILE` for perfectly readable markdown and source files written in Chinese, Japanese, or Korean.

## Root Cause

`binaryornot.check.is_binary()` samples a portion of the file and counts high-byte values. Dense CJK UTF-8 text has many bytes above 0x7F (each CJK character encodes to 3 bytes), which tips the heuristic into the "binary" classification even though the content is valid UTF-8 text.

## Fix

Before consulting `is_binary()`, attempt to decode the file's raw bytes as UTF-8. If the decode succeeds, the file is definitively text — no further check needed. Only if the UTF-8 decode raises `UnicodeDecodeError` do we fall through to `is_binary()` as a secondary check.

```python
try:
    with open(file_path, 'rb') as _f:
        _f.read().decode('utf-8')
except (UnicodeDecodeError, OSError):
    if is_binary(file_path):
        return ErrorObservation('ERROR_BINARY_FILE')
```

## Changes

- `openhands/runtime/action_execution_server.py` — `read()` method
- `openhands/runtime/impl/cli/cli_runtime.py` — `read()` and `edit()` methods

## Testing

The fix is verifiable locally:
```python
from binaryornot.check import is_binary

# Create a UTF-8 file with dense CJK content
with open('/tmp/test_cjk.md', 'w', encoding='utf-8') as f:
    f.write('# 测试\n' + '你好世界 ' * 500)

print(is_binary('/tmp/test_cjk.md'))  # True (wrong — this is the bug)

# With fix: decode check
with open('/tmp/test_cjk.md', 'rb') as f:
    f.read().decode('utf-8')  # Succeeds → file is text, skip is_binary
```


Prithvi1994:
Hi team! Just checking in on this PR — happy to address any feedback or make changes needed to move this forward. Thanks!

PR #13756, opened 2026-04-03
fix(slack): update initial resolver ack with final result

### LINKED ISSUE #13736: [Feature]: Slack resolver should avoid leaving both the initial "I'm on it" message and a separate final result
## Is there an existing feature request for this?

- [X] I have searched existing issues and feature requests, and this is not a duplicate.
- Related but not the same scope: [APP-797](https://linear.app/all-hands-ai/issue/APP-797/slack-resolver-post-final-assistant-message-remove-auto-summary) and [APP-976](https://linear.app/all-hands-ai/issue/APP-976/slack-resolver-sends-duplicate-finish-messages-to-a-channel-in-saas-as)

## Problem or Use Case

Today the Slack resolver leaves two user-visible messages in the thread/channel for a single run:

1. an initial acknowledgement like `I'm on it! ... can track my progress here`
2. a separate final result/summary message when the run completes

That makes the thread noisier than it needs to be, especially for quick tasks where the acknowledgement is immediately followed by the final answer.

From the current implementation:

* `enterprise/integrations/slack/slack_view.py` returns the initial `I'm on it! ... track my progress here` message
* `enterprise/integrations/slack/slack_v1_callback_processor.py` later posts a separate summary message with `chat_postMessage(...)`

## Why this matters

Slack should feel like a conversation.

A user asks the agent for something, waits, and then expects the agent to respond at the end. They generally do **not** expect a large amount of intermediate output in the channel/thread, because that feels spammy and makes the interaction feel less conversational.

The current two-message flow creates a few concrete UX problems:

* it adds visible noise to the thread/channel
* it makes short interactions feel unnecessarily spammy
* it creates extra message churn when a single final response would communicate the outcome more cleanly
* it adds another integration surface where something can go wrong, without adding much value once the final answer is available

The acknowledgement is still useful, but ideally users should end up with **one** clear final visible response rather than both the placeholder and the result.

## Proposed Solution

Please change the Slack resolver flow so it does **not** leave both messages behind.

Two reasonable options:

1. **Preferred:** edit the original `I'm on it` message in place with the final result when the run completes.
2. **Alternative:** if editing is awkward or undesirable, delete the original `I'm on it` message once the final result is posted.

If option 2 is chosen, the **final** message should still include the conversation link that is currently carried by the first message, so users do not lose the ability to jump into the OpenHands conversation.

## Alternatives Considered

* Keep the current two-message flow: simple, but noisy.
* Remove the first message entirely: worse UX because users lose immediate confirmation that the resolver picked up the request.

## Priority / Severity

## Estimated Scope

Small - UI tweak, config option, or minor change

## Feature Area

Integrations (GitHub, GitLab, etc.)

The current ack message already has the conversation link, and the callback processor already has the completion hook. It should be possible to persist the Slack message timestamp for the initial acknowledgement and then either:

* call `chat_update` on completion, or
* call `chat_delete` after posting the final completion message.

Either path should preserve the conversation link in the final visible message.

## Additional Context

This request is specifically about the **two-message UX** for Slack resolver runs, not about whether the final content should be a summary vs the actual final assistant message.

*This issue was created and updated by an AI assistant (OpenHands) on behalf of a user.*

nehaaprasad:
### summary of PR

- Slack resolver now updates the initial ack message with the final result.
- Falls back to posting a final threaded message if update timestamp is unavailable.
- Final visible message keeps the OpenHands conversation link.
- Added tests for update and fallback paths.

### change Type
[x] Bug fix

### Fixes
Resolves #13736

### DIFF
--- a/enterprise/integrations/slack/slack_manager.py
+++ b/enterprise/integrations/slack/slack_manager.py
@@ -512,7 +512,7 @@ class SlackManager(Manager[SlackViewInterface]):
         message: str | dict[str, Any],
         slack_view: SlackMessageView,
         ephemeral: bool = False,
-    ):
+    ) -> dict[str, Any] | None:
         """Send a message to Slack.
 
         Args:
@@ -525,14 +525,14 @@ class SlackManager(Manager[SlackViewInterface]):
         """
         client = AsyncWebClient(token=slack_view.bot_access_token)
         if ephemeral and isinstance(message, str):
-            await client.chat_postEphemeral(
+            return await client.chat_postEphemeral(
                 channel=slack_view.channel_id,
                 markdown_text=message,
                 user=slack_view.slack_user_id,
                 thread_ts=slack_view.thread_ts,
             )
         elif ephemeral and isinstance(message, dict):
-            await client.chat_postEphemeral(
+            return await client.chat_postEphemeral(
                 channel=slack_view.channel_id,
                 user=slack_view.slack_user_id,
                 thread_ts=slack_view.thread_ts,
@@ -540,7 +540,7 @@ class SlackManager(Manager[SlackViewInterface]):
                 blocks=message['blocks'],
             )
         else:
-            await client.chat_postMessage(
+            return await client.chat_postMessage(
                 channel=slack_view.channel_id,
                 markdown_text=message,
                 thread_ts=slack_view.message_ts,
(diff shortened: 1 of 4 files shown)


hieptl (changes requested):
Hello @nehaaprasad,

Could you please add tests to cover your changes and share a demo video showing how the output looks?

Thank you very much! 🙏

nehaaprasad:
@hieptl please check i have updated it

hieptl:
> @hieptl please check i have updated it

Could you kindly address the linting errors and provide a demo video showcasing how the output appears? Since the changes were made within the enterprise folder, which contains our SaaS code, I would appreciate it if you could share how the fix was tested and validated.

Thank you very much! 🙏

mamoodi:
@nehaaprasad please correct me if I'm wrong but you don't currently have the ability to test this right?

PR #13773, opened 2026-04-03
fix(app_server): ignore invalid remote sandbox status values

suyua9:
## Summary
- guard remote sandbox status parsing so non-string `status` values do not crash `_get_sandbox_status_from_runtime()`
- treat malformed status payloads as missing instead of calling `.lower()` on unexpected types
- add a regression test covering a non-string runtime status

## Testing
- `git diff --check -- openhands/app_server/sandbox/remote_sandbox_service.py tests/unit/app_server/test_remote_sandbox_service.py`
- `python3 -m py_compile openhands/app_server/sandbox/remote_sandbox_service.py tests/unit/app_server/test_remote_sandbox_service.py`
- `env -u HTTP_PROXY -u HTTPS_PROXY -u ALL_PROXY uv run python -m pytest -q tests/unit/app_server/test_remote_sandbox_service.py -k "non_string_status or status_mapping_case_insensitive or get_sandbox_status_from_runtime_empty_status or get_sandbox_status_from_runtime_unknown_status or get_sandbox_status_from_runtime_with_status"`


tofarr:
Is the remote service producing status values which are not strings?

suyua9:
Replying to the status-shape question: yes, this fix is intentionally defensive against malformed runtime payloads rather than a claim that non-string values are part of the normal contract.

The current failure mode is that any unexpected JSON shape in `status` raises before we can classify the sandbox, because the code called `.lower()` unconditionally. This patch narrows that path to real strings only and falls back to `SandboxStatus.MISSING` for malformed values, which matches the existing behavior for absent / unknown statuses.

I kept the scope minimal and added a regression case for a non-string `status` (`["running"]`) so this route no longer turns bad upstream data into an app-server exception.

PR #13784, opened 2026-04-06
feat(frontend): Allow bulk selection and deletion of conversations

### LINKED ISSUE #12871: [UI/UX] Bulk selection and deletion of conversations
## Problem

Currently, removing multiple conversations requires deleting them **one at a time**, with 2–3 clicks per conversation (open the menu, confirm delete, wait). For users who accumulate many conversations over time this is tedious and time-consuming.

## Proposed Solution

Add a **multi-select / bulk actions** mode to the conversation list:

1. **Selection mode** – A checkbox (or long-press / shift-click) on each conversation card lets the user select multiple conversations at once.
2. **"Select all" option** – A top-level checkbox or button to select all visible conversations.
3. **Bulk action bar** – When one or more conversations are selected, show a contextual action bar with at least:
   - **Delete selected** – deletes all selected conversations (with a single confirmation dialog).
   - *(Nice to have)* Other bulk actions in the future, e.g. archive, export.
4. **Clear selection** – Easy way to deselect all and exit selection mode.

### UX Sketch

```
┌─────────────────────────────────┐
│ ☐  Select All          Delete (3) │  ← bulk action bar (appears when ≥1 selected)
├─────────────────────────────────┤
│ ☑  Fix login bug          3h ago │
│ ☐  Refactor auth service  1d ago │
│ ☑  Add unit tests         2d ago │
│ ☑  Update README           5d ago │
│ ☐  Deploy pipeline         1w ago │
└─────────────────────────────────┘
```

## Additional Context

- Related (but different scope): #12545 – Graceful deletion of conversations (UX after a single delete)
- Related: #12468 – Archive conversations (alternative to deleting)

This feature would pair well with both: users could bulk-archive or bulk-delete depending on their workflow.

sjathin:
## Summary of PR

Add bulk selection and deletion of conversations to the conversation panel. Users can now select multiple conversations via always-visible checkboxes and delete them all at once, instead of deleting one at a time with 2-3 clicks each.

- Checkboxes always visible on each conversation card
- Sticky bulk action bar with "Select All" and "Delete (N)" appears when ≥1 conversation is selected
- Single confirmation dialog for bulk delete
- Also fixes a pre-existing bug where single-delete didn't optimistically remove conversations from the paginated list

## Demo Screenshots/Videos

https://github.com/user-attachments/assets/2fdfc8ac-dbe2-4347-9697-d5998957bad6

## Change Type

- [x] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist

## Fixes

Resolves #12871 

## Release Notes

Added bulk selection and deletion of conversations. Checkboxes on each conversation allow multi-select, with a "Select All" option and a single-click "Delete" action for all selected conversations.


hieptl:
Hello @sjathin,

While testing the changes locally, I came across a couple of issues:

**Issue 1 (blocking):**
The checkbox color should be consistent with other checkboxes across the application. You may find it helpful to refer to the `tos-checkbox.tsx` component.

issue1

**Issue 2 (blocking):**
We should display the list of selected conversations so users are clearly aware of which items will be deleted. Please refer to the video below for more context.

https://github.com/user-attachments/assets/bb04d91a-94c6-4ae5-8e91-40ce15049fbe

Thank you very much! 🙏

frontend/src/hooks/mutation/use-bulk-delete-conversations.ts
```diff
@@ -0,0 +1,42 @@
...
+} from "./conversation-mutation-utils";
+
+export const useBulkDeleteConversations = () => {
+  const queryClient = useQueryClient();
+
+  return useMutation({
+    mutationFn: async (variables: { conversationIds: string[] }) => {
+      await Promise.all(
+        variables.conversationIds.map((id) =>
+          ConversationService.deleteUserConversation(id),
```
hieptl: **suggestion (blocking):** If a user selects 50+ conversations and clicks delete, this fires 50 concurrent HTTP DELETE requests. Problems:

- Could overwhelm the backend or hit rate limits.
- Promise.all rejects on the first failure — if request 3 of 50 fails, the mutation errors, onError rolls back the entire cache, but requests 1-2 already deleted server-side. The onSettled invalidation will eventually reconcile, but the user sees a flash of "rollback" for items that are actually gone.

**Recommendation:** Add a backend batch-delete endpoint

Thank you! 🙏

sjathin:
Hi @hieptl, thanks for the review! I've addressed all three blocking issues. Here's a summary of the changes:

### Fixes

**1. Checkbox color consistency (Issue 1)**
Removed `accent-white` from both the conversation card checkbox and the "Select All" checkbox. They now use the browser default styling, consistent with `tos-checkbox.tsx`.

**2. Show selected conversations in delete dialog (Issue 2)**
Added a `bulkTitles` prop to the confirmation modal that renders a numbered, scrollable list of conversation titles being deleted. Updated the i18n bulk delete warning to say "the following N conversations" to introduce the list. Titles that are null/undefined fall back to "Untitled".

**3. Backend batch-delete endpoint (Issue 3)**
Replaced the `Promise.all` of individual DELETE requests with a single `POST /api/conversations/bulk-delete` endpoint. The endpoint:
- Accepts up to 50 conversation IDs per request
- Returns `{ succeeded: [], failed: [] }` for partial failure handling
- Handles both V1 and V0 conversations
- Uses `defer_sandbox_cleanup` to avoid SQLAlchemy session conflicts when deleting multiple V1 conversations in a loop

### Additional Bug Fixes (discovered during testing)

- **V1 cache mismatch**: `removeConversationsFromCache` was using V0 field names (`page.results`, `conv.conversation_id`) but the cache now contains V1 data (`page.items`, `conv.id`). This caused a runtime crash when bulk deleting. Fixed to match the V1 types, consistent with `updateConversationSandboxStatusInCache` in the same file.
- **Select All broken**: `toggleSelectAll` was mapping `c.conversation_id` (V0) instead of `c.id` (V1), so it produced a set of `undefined` values. Fixed.
- **`onSuccess` cleanup**: Now uses `data.succeeded` from the batch response to only clear localStorage for conversations that were actually deleted.
- **Refetch pausing**: Added `useIsMutating` to `usePaginatedConversations` to pause refetching during bulk delete, preventing race conditions between optimistic removal and automatic refetch.

-------


### Recording


https://github.com/user-attachments/assets/6379bcef-57bb-4f51-b534-c75daf1440e7

frontend/src/hooks/mutation/use-bulk-delete-conversations.ts
sjathin: Fixed — added POST /api/conversations/bulk-delete backend endpoint that accepts up to 50 IDs and returns { succeeded, failed }. Frontend now makes a single request instead of N concurrent DELETEs. The endpoint uses defer_sandbox_cleanup=True to keep the shared db session alive across the loop (the existing single-delete fires a background task that closes the session, which would break subsequent iterations).

openhands/server/routes/manage_conversations.py
```diff
@@ -624,15 +686,16 @@ async def _try_delete_v1_conversation(
...
             await db_session.commit()
 
-            # Delete the sandbox in the background (checks remaining conversations first)
-            asyncio.create_task(
-                _finalize_delete_and_close_connections(
-                    sandbox_service,
-                    app_conversation_info.sandbox_id,
-                    db_session,
-                    httpx_client,
+            if not defer_sandbox_cleanup:
```
hieptl, line 689: **suggestion (blocking):** The new defer_sandbox_cleanup=True flag skips the background _finalize_delete_and_close_connections task for every item in the loop:    

```                 
if not defer_sandbox_cleanup:                                                                                                                            
    asyncio.create_task(_finalize_delete_and_close_connections(...))                                                                                     
```

But bulk_delete_conversations never runs cleanup for the deferred sandboxes after the loop. The docstring claims "The caller is responsible for sandbox  cleanup" — the caller isn't. 

Result: bulk-deleted V1 conversations leave their sandboxes running and DB/HTTP connections open. This needs an explicit  post-loop cleanup over the distinct sandbox_ids, or the flag needs to remove the false premise

Thank you! 🙏

openhands/server/routes/manage_conversations.py
```diff
@@ -584,15 +584,77 @@ async def delete_conversation(
...
+
+
+class BulkDeleteResponse(BaseModel):
+    """Response body for bulk conversation deletion."""
+
+    succeeded: list[str]
+    failed: list[str]
+
+
+@app.post('/conversations/bulk-delete', deprecated=True)
```
hieptl, line 601: **suggestion (blocking):** `deprecated=True` marks the endpoint as deprecated in OpenAPI docs immediately. Thank you! 🙏

openhands/server/routes/manage_conversations.py
```diff
@@ -584,15 +584,77 @@ async def delete_conversation(
     return await _delete_v0_conversation(conversation_id, user_id)
 
 
+class BulkDeleteRequest(BaseModel):
+    """Request body for bulk conversation deletion."""
+
+    model_config = ConfigDict(extra='forbid')
+    conversation_ids: list[str] = Field(..., max_length=50)
```
hieptl, line 591: **suggestion (blocking):** Backend caps conversation_ids at 50 (Field(..., max_length=50)), but "Select All" can select every currently-loaded item. If >50, the request fails with a validation error that isn't surfaced to the user. Either enforce on the frontend or paginate the bulk call. Thank you! 🙏

openhands/server/routes/manage_conversations.py
sjathin: Fixed. `_try_delete_v1_conversation` now returns `(result, sandbox_id)` as a tuple. The bulk delete loop collects all distinct sandbox IDs, then spawns `_finalize_bulk_sandbox_cleanup` as a background task
   after the loop — it checks each sandbox for remaining conversations and deletes empty ones, then closes the shared `db_session`/`httpx_client` in a `finally` block.

openhands/server/routes/manage_conversations.py
sjathin: Removed. Good catch — this was copy-pasted from the other endpoint decorators.

openhands/server/routes/manage_conversations.py
sjathin: Fixed on the frontend side. Added `BULK_DELETE_BATCH_SIZE = 50` — the mutation function now chunks IDs into batches of 50 and processes them sequentially, aggregating `succeeded`/`failed` across all batches.
   Users can select and delete any number of conversations without hitting the validation error.

sjathin:
Hi @hieptl, thanks for the review! All 3 blocking issues are addressed in the latest commit.

### Fixes

**1. Missing sandbox cleanup (defer_sandbox_cleanup)**
`_try_delete_v1_conversation` now returns `(result, sandbox_id)` as a tuple. The bulk delete loop collects all distinct sandbox IDs, then spawns `_finalize_bulk_sandbox_cleanup` as a background task after the loop — it checks each sandbox for remaining conversations and deletes empty ones, then closes the shared `db_session`/`httpx_client` in a `finally` block.

**2. Removed `deprecated=True`**
Removed from the bulk-delete endpoint decorator.

**3. Frontend >50 items limit**
Added `BULK_DELETE_BATCH_SIZE = 50` on the frontend. The mutation function chunks IDs into batches of 50 and processes them sequentially, aggregating `succeeded`/`failed` across all batches. Users can select any number of conversations without hitting the validation error.

### Additional fix (discovered during testing)

**4. Deleted conversations reappearing + redirect not firing**
Added `mutationKey: ["bulk-delete-conversations"]` to the mutation hook. This makes the `useIsMutating` guard in `usePaginatedConversations` actually work — it pauses refetching during bulk delete, preventing the race condition that caused deleted items to reappear and blocked the redirect when the current conversation was in the deleted set.

### Tests added
- Frontend: 4 vitest tests for batching logic, aggregation, and error handling
- Backend: 5 pytest tests for `_try_delete_v1_conversation` return type, sandbox cleanup scheduling, and failure handling

hieptl:
Hello @sjathin,

I encountered an issue while testing the pull request locally and wanted to share my observations.

When opening the conversation panel, the APIs that return the list of conversations respond quickly. However, after deleting a conversation, the same APIs are triggered again to refetch the data, and the response time becomes noticeably slower.

https://github.com/user-attachments/assets/0d53e21d-5ffd-4b4f-a5de-640b7c2fa376

Thank you very much! 🙏

sjathin:
Hi @hieptl, thanks for flagging the slow refetch! I've identified and fixed the root cause.

### Root Cause

After bulk delete, the `onSettled` callback was calling `invalidateQueries` which triggered an immediate refetch of all loaded conversation pages. Those refetches hit the backend search API → `batch_get_sandboxes` → `_get_runtimes_batch` (HTTP to runtime API). Since the background sandbox cleanup was concurrently stopping those sandboxes, the runtime API calls were slow — causing the visible latency.

### Fix

1. **Removed the unconditional `onSettled` refetch** — the optimistic cache update (`removeConversationsFromCache`) already removes deleted conversations from the UI instantly. A full refetch is only needed when there are partial failures (to re-add items that weren't actually deleted).

2. **Removed `useIsMutating` query guard** — this was disabling/re-enabling the paginated query during mutation, which caused React Query to trigger a redundant refetch when the query re-enabled. The `cancelQueries` in `onMutate` is sufficient to prevent race conditions.

Now: successful bulk delete → instant UI update via optimistic cache, no blocking refetch. Only invalidates on partial failure or network error.

hieptl:
Hello @sjathin,

It appears that the performance issue may not have been fully resolved yet. Please refer to the reproduction steps and video below for additional context.

**Step 1:** On the home page, create a new conversation. You will be redirected to the conversation page.

**Step 2:** Navigate back to the home page, delete the newly created conversation, and close the conversation panel.

**Step 3:** Immediately reopen the conversation after completing Step 2. At this point, you may notice that the conversation list takes a considerable amount of time to load.

https://github.com/user-attachments/assets/582d7532-05dd-4272-a259-ce527950c4ae

Thank you very much! 🙏

PR #13798, opened 2026-04-06
chore: Add sdk to mypy checking in enterprise folder and fix the resulting errors

raymyers:

---

## Why

Continuation of #13637. We use pre-commit to run linting and mypy checks, which [does not include](https://github.com/pre-commit/mirrors-mypy) the full project dependencies. This PR adds the OpenHands SDK to **enterprise folder** mypy config and fixes 50 resulting type errors.

I tried to minimize behavior changes, adding ignores when there was no obvious way.
openhands-agent-server was not added because the transitive dependencies introduce many more errors. In fact the pip package includes the whole server when we only want the client.

## Summary

-

## Type

- [X] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [X] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:3615080-nikolaik   --name openhands-app-3615080   docker.openhands.dev/openhands/openhands:3615080
```


raymyers:
Recreating as #14463

PR #13807, opened 2026-04-07
Improve error message for expired Keycloak sessions during API key auth

jpshackelford:
## Problem

When using an API key (e.g., `sk-oh-...`) for authentication, users would receive a generic error when their Keycloak offline session had expired:

```json
{"error": "BearerTokenError"}
HTTP Status: 401
```

The actual underlying error was:
```
KeycloakPostError: 400: {"error":"invalid_grant","error_description":"Offline user session not found"}
```

This was confusing because:
1. **The API key itself was valid** - it existed in the database and wasn't expired
2. **The error message gave no indication of what was wrong** - users couldn't tell if the key was invalid, expired, or something else
3. **Users had no idea how to fix it** - the solution (logging into the web UI) wasn't obvious

## Solution

This PR catches the Keycloak session expiration error and raises the existing `SessionExpiredError` (from `openhands.server.types`) with a clear, actionable message:

```json
{
  "error": "SessionExpired",
  "message": "Your session has expired. Please log in at https://app.all-hands.dev to re-authenticate, then retry your request."
}
```

## Changes

- **`enterprise/server/auth/saas_user_auth.py`**: Updated `saas_user_auth_from_bearer()` to catch `KeycloakError` and check for session expiration indicators (`invalid_grant`, `session not found`), raising the existing `SessionExpiredError` from `openhands.server.types`
- **`enterprise/server/middleware.py`**: Added handler for `SessionExpiredError` that returns the descriptive JSON response
- **`enterprise/tests/unit/test_saas_user_auth.py`**: Added tests for session expiration handling in bearer token auth

> **Note**: This reuses the existing `SessionExpiredError` class from `openhands/server/types.py` rather than creating a duplicate, per reviewer feedback.

## Testing

The error handling was discovered and verified through real-world debugging:
1. API call with valid `sk-oh-...` key returned `BearerTokenError`
2. Server logs showed `KeycloakPostError` with `Offline user session not found`
3. Logging back into the web UI re-established the session
4. API calls then succeeded

---

*This PR was created by an AI assistant (OpenHands) on behalf of @jpshackelford.*

### DIFF
--- a/enterprise/server/auth/saas_user_auth.py
+++ b/enterprise/server/auth/saas_user_auth.py
@@ -35,6 +35,7 @@ from openhands.integrations.provider import (
     ProviderType,
 )
 from openhands.server.settings import Settings
+from openhands.server.types import SessionExpiredError
 from openhands.server.user_auth.user_auth import AuthType, UserAuth
 from openhands.storage.data_models.secrets import Secrets
 from openhands.storage.settings.settings_store import SettingsStore
@@ -313,6 +314,18 @@ async def saas_user_auth_from_bearer(request: Request) -> SaasUserAuth | None:
         )
         await saas_user_auth.refresh()
         return saas_user_auth
+    except KeycloakError as exc:
+        # Check for session expiration errors from Keycloak
+        error_str = str(exc)
+        if 'invalid_grant' in error_str or 'session not found' in error_str.lower():
+            logger.warning(
+                'API key authentication failed due to expired session',
+                extra={'error': error_str},
+            )
+            raise SessionExpiredError(
+                'Your session has expired. Please log in at https://app.all-hands.dev to re-authenticate, then retry your request.'
+            ) from exc
+        raise BearerTokenError from exc
     except Exception as exc:
         raise BearerTokenError from exc
 
--- a/enterprise/server/middleware.py
+++ b/enterprise/server/middleware.py
@@ -10,14 +10,13 @@ from server.auth.auth_error import (
     NoCredentialsError,
     TosNotAcceptedError,
 )
-
-from openhands.server.types import SessionExpiredError
 from server.auth.gitlab_sync import schedule_gitlab_repo_sync
 from server.auth.saas_user_auth import SaasUserAuth, token_manager
 from server.routes.auth import set_response_cookie
 from server.utils.url_utils import get_cookie_domain, get_cookie_samesite
 
 from openhands.core.logger import openhands_logger as logger
+from openhands.server.types import SessionExpiredError
 from openhands.server.user_auth.user_auth import AuthType, UserAuth, get_user_auth
 from openhands.server.utils import config
 
--- a/enterprise/server/middleware.py
+++ b/enterprise/server/middleware.py
@@ -10,6 +10,8 @@ from server.auth.auth_error import (
     NoCredentialsError,
     TosNotAcceptedError,
 )
+
+from openhands.server.types import SessionExpiredError
 from server.auth.gitlab_sync import schedule_gitlab_repo_sync
 from server.auth.saas_user_auth import SaasUserAuth, token_manager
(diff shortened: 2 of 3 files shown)


enterprise/server/auth/auth_error.py
```diff
@@ -22,6 +22,24 @@ class BearerTokenError(AuthError):
     pass
 
 
+class SessionExpiredError(AuthError):
```
malhotra5: NIT: should we dedup this against the `ExpiredError` type?

jpshackelford:
@openhands addressed issues found in CI

jpshackelford:
Fixed the CI lint issues in commit 666d47584:

- Sorted imports alphabetically (keycloak before pydantic) in `test_saas_user_auth.py`
- Changed double quotes to single quotes per ruff style rules
- Added newline at end of file
- Fixed import grouping in `middleware.py`

The 'Lint enterprise python' check is now passing. ✅

---
*This comment was created by an AI assistant (OpenHands) on behalf of @jpshackelford.*

jpshackelford:

enterprise/server/auth/auth_error.py
jpshackelford: done

jpshackelford:
_This code review was based on a commit that has since been reverted. Hidden for clarity._

enterprise/server/auth/auth_error.py
jpshackelford: Addressed in 4c4a3b921 - consolidated `SessionExpiredError` as a subclass of `ExpiredError` in `auth_error.py`. This:

- Properly deduplicates against `ExpiredError` by making it a subclass (`SessionExpiredError → ExpiredError → AuthError`)
- Removes the dependency on the deprecated `openhands/server/types.py` (scheduled for removal April 1, 2026)
- Also updated `token_manager.py` to use the consolidated error class for consistency
- Added tests for the new inheritance hierarchy

The semantic distinction between general token expiration and Keycloak session expiration is preserved while integrating properly into the enterprise auth error hierarchy.

jpshackelford: Revisiting this - I implemented a consolidation in commit 4c4a3b92 but it actually made things worse. The problem is that `openhands.server.types.SessionExpiredError` (a `ValueError` subclass) is used by 5+ integration managers (Linear, Jira, GitHub, GitLab, Slack). Creating a second `SessionExpiredError` in `auth_error.py` (an `ExpiredError` subclass) fragments the exception hierarchy - the middleware would catch one type while integrations raise another.

The first commit's approach of reusing the existing `SessionExpiredError` from `openhands.server.types` is actually correct - it's consistent with how the rest of the codebase handles this error.

The deprecation notice in `types.py` says "scheduled for removal April 1, 2026" but we're past that date and the file is still actively used. When/if that file actually gets removed, we should do a proper migration that updates ALL usages together.

Reverted back to the original approach. Happy to discuss if you see a better path forward.

jpshackelford:
Reverted to the original approach after attempting a consolidation that made things worse.

**Original fix (current state):** Catches Keycloak session expiration errors during API key auth and raises the existing `SessionExpiredError` from `openhands.server.types` with a clear, actionable message.

**What we tried:** Created a new `SessionExpiredError` subclass of `ExpiredError` in `auth_error.py` to avoid the deprecated `types.py` file.

**Why we reverted:** This fragmented the exception hierarchy. Five+ integration managers (Linear, Jira, GitHub, GitLab, Slack) already use `openhands.server.types.SessionExpiredError` (a `ValueError` subclass). Creating a second class with the same name but different inheritance meant the middleware would catch one type while integrations raise another. We could have refactored all of the code to use a new exception class, but that seemed far afield of the problem we're trying to solve with this PR. Better to stay consistent with the existing codebase.

Issue #13738, opened 2026-04-03
[Feature]: Fix documentation on using OpenAI API compatible models


peterwwillis:
### Is there an existing feature request for this?

- [x] I have searched existing issues and feature requests, and this is not a duplicate.

### Problem or Use Case

As a user, I want to use OpenHands with *any model or endpoint API*, but it's very difficult. Please either implement the documentation fixes below, or add code to query the models API endpoint and use the appropriate model format.

Please note: there is an actual bug below, amidst the missing documentation and confusion.

**The painful process:**

 - After installing OpenHands and starting it up, you're prompted to configure LLM. The first problem is, what if you don't use just one? The software doesn't let you switch between LLM providers, which as a user, is painful. I use many providers, both local and remote. (Other AI software works this way)

 - Next I attempt to select an LLM provider from the list, but my provider (HuggingFace) isn't there. (HuggingFaceH4 isn't it). So I click *Advanced*. There is no documentation or hints on this page, so I have to guess what to put in. I look up my HF API key and put that in, and look up the inference URL and put that in.

 - I expect to now be greeted with a drop-down of all the models, since a simple API call immediately returns all models (`curl -s "https://router.huggingface.co/v1/models" | jq .`). But alas I have to put in one model by hand. So, my second problem: I'm not allowed to switch models in the OpenHands UI?? Seriously? Every AI software lets you switch models on the fly...

 - So I look up the model name and put in `Qwen/Qwen-Coder-Next`, the name of the model in HuggingFace. I click Save and go to the chat window to start a new conversation.

 - From this point on, every time I try to chat, I get an error. But the error text actually *disappears from my screen* each time. After fumbling around with the UI, at some point, I can finally get the error text to *not* disappear. It turns out it's this:

```
litellm.BadRequestError: LLM Provider NOT provided. Pass in the LLM provider you are trying to call. You passed model=Qwen/Qwen3-Coder-Next
 Pass model as E.g. For 'Huggingface' inference endpoints pass in `completion(model='huggingface/starcoder',..)` Learn more: [docs.litellm.ai/docs/providers](https://docs.litellm.ai/docs/providers)
```

[...]

PR #13826, opened 2026-04-09
docs(ui): clarify custom model format for OpenAI-compatible providers

Kl163:
## Summary

Clarify the custom model input for OpenAI-compatible providers in the LLM settings UI.

This adds a short help text below the **Custom Model** field so users know that some OpenAI-compatible providers require a provider-prefixed model id.

## Why

Issue #13738 reports that the current UI does not make this requirement clear enough. Users can configure a compatible Base URL, but still get stuck because the model value may need a provider prefix.

For example, a HuggingFace-compatible setup may require a model value like:

`huggingface/Qwen/Qwen3-Coder-Next`

## Changes

- Added a new translated help text key: `SETTINGS$CUSTOM_MODEL_HELP`
- Displayed the help text below the **Custom Model** field in `llm-settings.tsx`
- Regenerated the i18n declaration file

## Scope

This is a UI/documentation-only clarification:
- no backend behavior changes
- no model resolution logic changes
- no settings flow changes


enyst (commented):
Hey, thank you for the PR! Could you please show a video of how it works and looks like?

PR #13863, opened 2026-04-10
fix: dynamically resolve internal MCP server URL on sandbox restart

### LINKED ISSUE #13861: [Bug]: Agent-server configs are not updated

### Bug Description

When creating a new sandbox, a `meta.json` file is generated at:

```
/workspace/conversations/<conversation_id>/meta.json
```

This file includes the configuration for the internal MCP server along with any user-defined MCP servers:

```json
"mcp_config": {
  "mcpServers": {
    "default": {
      "url": "http://host.docker.internal:3000/mcp/mcp"
    },
    ...
  }
}
```

The internal MCP server URL is expected to be derived from environment variables such as `OH_WEB_URL` and `WEB_HOST`, which together define the `web_url` value.

---

### Problem

If the URL of the MCP server changes:

* The agent server continues attempting to connect to the previous (stale) URL.
* Restarting the sandbox regenerates the `meta.json` file, but it still contains the outdated URL instead of the updated `web_url`.
* As a result, the agent fails to connect to the internal MCP server and shuts down.

---

### Expected Behavior

* The sandbox configuration should always reflect the current value of `web_url` derived from `OH_WEB_URL` / `WEB_HOST`.
* On sandbox restart, the internal MCP server URL should be updated dynamically to match the current environment.

---

### Actual Behavior

* The sandbox configuration and the `meta.json` persists or regenerates with an outdated internal MCP server URL.
* The agent attempts to connect to an invalid endpoint and terminates when the connection fails.

---

### Impact

* Sandboxes become unusable after IP changes.
* Requires manual intervention or workaround to restore connectivity.

---

### Suggested Fix

* Ensure the sandbox is reconfigured using the current `web_url` at sandbox startup.

Jinhaooo:
Closes #13861 — reported by @kripper

---

## Why

The internal MCP server URL (`{web_url}/mcp/mcp`) was snapshot-captured once during `LiveStatusAppConversationServiceInjector.inject()` and stored as `self.web_url`. If the host IP changes between sandbox restarts, the agent continues using the stale URL and all MCP connections fail.

## Summary

- Added `_get_current_web_url()` that dynamically resolves `web_url` from environment variables (`OH_WEB_URL` → `WEB_HOST` → `self.web_url` fallback) on every call
- Modified `_add_system_mcp_servers()` to use `_get_current_web_url()` instead of the stale `self.web_url` snapshot
- Added 9 unit tests covering all resolution paths

## Issue Number

#13861

## How to Test

1. Start OpenHands with `OH_WEB_URL=http://192.168.1.100:3000`
2. Create a conversation — verify MCP URL is `http://192.168.1.100:3000/mcp/mcp`
3. Stop, change to `OH_WEB_URL=http://192.168.1.200:3000`, restart
4. Resume / new conversation — verify MCP URL is now `http://192.168.1.200:3000/mcp/mcp`

Unit tests: `pytest tests/unit/app_server/test_get_current_web_url.py -v`

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Backward compatible: when neither `OH_WEB_URL` nor `WEB_HOST` is set, `_get_current_web_url()` falls back to `self.web_url` — identical to previous behavior.

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -933,6 +933,37 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
 
         return user_search_key or service_tavily_key
 
+    def _get_current_web_url(self) -> str | None:
+        """Dynamically resolve the current web URL from environment variables.
+
+        The web_url captured at service initialization (self.web_url) can become
+        stale when the host IP changes between sandbox restarts or conversation
+        resumes. This method re-reads the environment variables each time it is
+        called so that MCP server URLs always point to the current address.
+
+        Resolution order:
+          1. OH_WEB_URL environment variable (highest priority)
+          2. WEB_HOST environment variable → https://{WEB_HOST}
+          3. self.web_url snapshot (backward-compatible fallback)
+          4. Docker host.docker.internal fallback when using DockerSandboxService
+        """
+        oh_web_url = os.getenv('OH_WEB_URL')
+        if oh_web_url:
+            return oh_web_url
+
+        web_host = os.getenv('WEB_HOST')
+        if web_host:
+            return f'https://{web_host}'
+
+        if self.web_url:
+            return self.web_url
+
+        # Docker fallback: use host.docker.internal if running in Docker
+        if isinstance(self.sandbox_service, DockerSandboxService):
+            return f'http://host.docker.internal:{self.sandbox_service.host_port}'
+
+        return None
+
     async def _add_system_mcp_servers(
         self, mcp_servers: dict[str, Any], user: UserInfo, conversation_id: UUID
     ) -> None:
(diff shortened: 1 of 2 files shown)


tofarr:
Typically processes don't pick up environment variable updates while the process is running - and sandboxed runtimes within docker have their environment variables set on creation - if this URL is dynamic and requires update, it seems to me that this would be better served by specifying a domain name for your openhands server in your hosts file and then letting the network stack resolve it dynamically - otherwise existing runtimes would not pick up your new variable. (It sounds like you are running into the types of problems k8s ingresses are designed to fix)

Jinhaooo:
Thank you for the review — you're absolutely right.
As you pointed out, this is really a service discovery problem that should be solved at the network layer — using a stable hostname (via `/etc/hosts`, Docker Compose service names, or k8s Services/Ingresses) instead of a raw IP address.
I looked through the OpenHands documentation and noticed there's currently no guidance recommending users to use a stable domain name rather than a bare IP for `OH_WEB_URL` / `WEB_HOST`. Given that this is a recurring pain point, it might be worth adding a note in the docs.
Thanks again for the insight.

PR #13875, opened 2026-04-10
Update getUser() to use V1 API endpoint /api/v1/users/git-info

tofarr:

---

## Why

The `getUser()` method was using the deprecated `/api/user/info` endpoint from the legacy V0 API. The V1 endpoint `/api/v1/users/git-info` provides the same functionality with a cleaner response format.

## Summary

- Updated `UserService.getUser()` to call `/api/v1/users/git-info` instead of `/api/user/info`
- Simplified response handling since the new endpoint returns `GitUser` directly without requiring manual field mapping

## Screenshots
image
image

## How to Test

1. Run the frontend application
2. Verify that user information is correctly loaded in the UI
3. Check browser console for any API errors related to user info fetching

## Video/Screenshots

N/A

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The `getGitOrganizations()` method was intentionally left unchanged as it uses `/api/user/git-organizations` which is part of the enterprise UI and is not deprecated.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:03bce8d-nikolaik   --name openhands-app-03bce8d   docker.openhands.dev/openhands/openhands:03bce8d
```

### DIFF
--- a/frontend/src/api/user-service/user-service.api.ts
+++ b/frontend/src/api/user-service/user-service.api.ts
@@ -11,20 +11,8 @@ class UserService {
    * @returns Git user information
    */
   static async getUser(): Promise<GitUser> {
-    const response = await openHands.get<GitUser>("/api/user/info");
-
-    const { data } = response;
-
-    const user: GitUser = {
-      id: data.id,
-      login: data.login,
-      avatar_url: data.avatar_url,
-      company: data.company,
-      name: data.name,
-      email: data.email,
-    };
-
-    return user;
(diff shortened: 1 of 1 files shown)


hieptl (approved):
Thank you! 🙏

PR #13880, opened 2026-04-10
fix: fallback to LLM_BASE_URL for openhands_provider_base_url

gpothier:

---

## Why

When creating new conversations via /new command, the LLM base_url was not being set correctly because the code looked for OPENHANDS_PROVIDER_BASE_URL env var, but many deployments use the traditional LLM_BASE_URL instead.

This caused conversations to fail with 'Incorrect API key' errors when using litellm_proxy/* models, because without the base_url, requests went to the wrong endpoint.

## Summary

The fix looks more like a workaround, I don't know what the intended design is. But it feels wrong to have the base url come from two different variables.

## How to Test

1. Create new Openhands conversation (this creates a new sandbox)
2. Say Hi to LLM: no problem
3. Issue the /new command: same sandbox, blank conversation
4. Say Hi to LLM: API key error

## Video/Screenshots

imagen

imagen

imagen

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/openhands/app_server/config.py
+++ b/openhands/app_server/config.py
@@ -105,8 +105,11 @@ def get_default_permitted_cors_origins() -> list[str]:
 
 
 def get_openhands_provider_base_url() -> str | None:
-    """Return the base URL for the OpenHands provider, if configured."""
-    return os.getenv('OPENHANDS_PROVIDER_BASE_URL') or None
+    """Return the base URL for the OpenHands provider, if configured.
+
+    Falls back to LLM_BASE_URL for backward compatibility.
+    """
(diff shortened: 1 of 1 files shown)


tofarr (approved):
🍰

gpothier:
Thanks a lot @tofarr for taking the time to check this PR and accept it! I'd like to take this opportunity to draw your attention to this other PR: https://github.com/OpenHands/OpenHands/pull/13572
Without it, OpenHands in unusable in Firefox after a few dozen turns.

PR #13953, opened 2026-04-16
fix(runtime): prevent agent from killing sandbox server on port 8000

### LINKED ISSUE #13927: [Bug]: OpenHands kills its own process on port 8000 during local web deployment

### Bug Description

When OpenHands tried to launch a local web app, it appears to have checked for an available port and decided to free port 8000 by terminating the process using it.

The problem is that port 8000 was being used by OpenHands itself. So OpenHands ended up killing its own main process, which caused the entire service to crash.

Image

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Steps to Reproduce

_No response_

CLI (uv tool install)

_No response_

### OpenHands Version

_No response_

### Version Confirmation

- [ ] I have confirmed this bug exists on the LATEST version of OpenHands

### Model Name

_No response_

### Operating System

None

### Browser (if using web UI)

_No response_

### Logs and Error Messages

_No response_

_No response_

moorsecopers99:

---

## Why

When an AI agent tries to deploy a local web app, it may check for an available port and decide to free port 8000 by terminating the process using it (e.g., `fuser -k 8000/tcp` or `kill $(lsof -t -i:8000)`). The problem is that port 8000 is used by the OpenHands action execution server running inside the sandbox — so the agent kills its own runtime, causing the entire session to crash.

## Summary

- Add a pre-execution safety check in `BashSession.execute()` that detects commands which would kill processes on the sandbox server port and returns an error with guidance to use a different port
- Pass the server port from `ActionExecutor` through to `BashSession` so it knows which port to protect
- Export `OPENHANDS_SANDBOX_PORT` environment variable in the bash session so agents can proactively avoid the port
- Add 11 unit tests covering various port-killing command patterns (fuser, lsof, kill, ss/netstat pipes) and safe command passthrough

## Issue Number

Closes #13927

## How to Test

1. Run the unit tests:
   ```bash
   poetry run pytest tests/unit/runtime/utils/test_bash_port_protection.py -v
   ```
2. All 11 tests should pass, verifying:
   - Commands like `fuser -k 8000/tcp` and `kill $(lsof -t -i:8000)` are blocked when targeting the server port
   - Safe commands (e.g., `ls`, `lsof -i:8000` without kill) are not blocked
   - Commands targeting other ports are not blocked
   - When no server port is configured, all commands are allowed

3. To test end-to-end: start OpenHands, ask the agent to "create and run a web server on port 8000" — it should now get an error message suggesting an alternative port instead of crashing the session.

## Video/Screenshots

N/A — backend-only change with unit tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The detection uses regex patterns matching common port-killing idioms (`fuser -k`, `kill` via `lsof`, pipe chains to `kill`). This is intentionally conservative — it only blocks commands that both reference the server port number AND match a kill pattern.
- The `OPENHANDS_SANDBOX_PORT` env var provides a secondary defense: well-behaved agents can read it to avoid the port entirely.
- This is in the legacy V0 `action_execution_server.py` / `bash.py` codepath, which is still actively used. The V1 agent server (via the Software Agent SDK) would need a similar protection.

### DIFF
--- a/openhands/runtime/action_execution_server.py
+++ b/openhands/runtime/action_execution_server.py
@@ -183,11 +183,13 @@ class ActionExecutor:
         user_id: int,
         enable_browser: bool,
         browsergym_eval_env: str | None,
+        server_port: int | None = None,
     ) -> None:
         self.plugins_to_load = plugins_to_load
         self._initial_cwd = work_dir
         self.username = username
         self.user_id = user_id
+        self.server_port = server_port
         _updated_user_id = init_user_and_working_directory(
             username=username, user_id=self.user_id, initial_cwd=work_dir
         )
@@ -291,6 +293,7 @@ class ActionExecutor:
                     os.environ.get('NO_CHANGE_TIMEOUT_SECONDS', 10)
                 ),
                 max_memory_mb=self.max_memory_gb * 1024 if self.max_memory_gb else None,
+                server_port=self.server_port,
             )
             bash_session.initialize()
             return bash_session
@@ -712,6 +715,7 @@ if __name__ == '__main__':
             user_id=args.user_id,
             enable_browser=args.enable_browser,
             browsergym_eval_env=args.browsergym_eval_env,
+            server_port=args.port,
         )
         await client.ainit()
         logger.info('ActionExecutor initialized.')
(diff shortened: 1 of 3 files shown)


tofarr (commented):
This code part of the V0 API and is scheduled for removal - I think that the actual fix for this is in https://github.com/OpenHands/software-agent-sdk

Issue #13964, opened 2026-04-16
[Bug]: Integration settings needs a refresh now


mamoodi:
https://github.com/user-attachments/assets/efd1f9ce-39e8-40ba-836c-714963cf8b4f

### Is there an existing issue for the same bug?

- [x] I have searched existing issues and this is not a duplicate.

### Bug Description

When adding a new integration token, we need to refresh the browser for it to show up.

### Expected Behavior

It should show up as we save it.

### Steps to Reproduce

1. Set a GitHub or GitLab token
2. Press `Save Changes`

You then have to refresh browser for it to show up.

### OpenHands Installation Method

Local GUI (Docker web interface)

### Version Confirmation

- [ ] I have confirmed this bug exists on the LATEST version of OpenHands

### Operating System

None

VascoSch92:
I believe this should fix that: https://github.com/OpenHands/OpenHands/pull/13970

PR #13970, opened 2026-04-16
fix(frontend): refresh integrations UI after saving git provider token

VascoSch92:


## Summary

Problem: After saving a GitHub/GitLab token in Integrations settings, the token only appeared after a full browser refresh.                                                                                        
  
Cause: useAddGitProviders invalidated the query key ["settings", organizationId], but useSettings caches under ["settings", scope, organizationId] (with scope="personal" by default). React Query's prefix match failed at index 1, so the cache was never marked stale.                                                                                                                                                            
                                                                                                                                                                                                                     
Fix: Invalidate with the ["settings"] prefix, which matches every scope/org variant and triggers an immediate refetch — consistent with how useLogout already invalidates settings.

## Issue Number
#13964

## Video/Screenshots


https://github.com/user-attachments/assets/579aa7d7-2bb2-45f1-92ac-3bb42df6892e



## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


neubig:
Thanks for taking a look at this, but I think https://github.com/OpenHands/OpenHands/pull/13979 was a better fix, so I'm going to close this. I also opened a PR to help prevent the bug in the future: https://github.com/OpenHands/OpenHands/pull/14011

PR #13988, opened 2026-04-16
fix: invalidate sandbox and VS Code URL caches on conversation resume

### LINKED ISSUE #13987: VS Code tab shows "Bad Gateway" — stale URL after new runtime provisioned on conversation resume
## Summary

When a conversation is resumed after its runtime was auto-paused, a **new runtime is provisioned**, but the VS Code tab still loads the URL from the **old runtime**, resulting in a "Bad Gateway" error.

## Root Cause

The VS Code URL becomes stale because the frontend TanStack Query cache is not invalidated when a sandbox is resumed with a new runtime:

1. Conversation starts → sandbox gets **runtime A** → VS Code URL = `https://runtimeA.prod-runtime.all-hands.dev`
2. Runtime A idles out and is auto-paused (K8s Service + Ingress deleted, deployment scaled to 0)
3. User resumes the conversation → **new runtime B** is provisioned
4. The frontend still has the cached VS Code URL for **runtime A** (via TanStack Query `useBatchSandboxes` / `useUnifiedVSCodeUrl`)
5. The iframe loads the old URL → Traefik has no backend for it → **"Bad Gateway"**

The infra layer (runtime-API) works correctly — it provisions the new runtime and returns the correct URL when queried. The issue is that the frontend does not refetch sandbox data after resume.

## Relevant Code

- **`frontend/src/hooks/query/use-unified-vscode-url.ts`** — gets VS Code URL from `sandbox.exposed_urls` via `useBatchSandboxes`. The query key includes the `sandbox` object, but if the sandbox query cache is not invalidated on resume, it keeps returning stale data.
- **`frontend/src/hooks/query/use-batch-sandboxes.ts`** — fetches sandbox data including `exposed_urls`.
- The **resume mutation** (`POST /api/v1/sandboxes/{id}/resume`) does not appear to invalidate the sandbox or VS Code URL query caches.

## Datadog Evidence (conversation `5961d9222af04b0fbc27ccfe789fea7d`)

- **Sandbox**: `5mIr7g6soMCiTNkUPZaxv4`
- **Original runtime**: `tjmlvsnfikwowklh` (VS Code URL: `https://tjmlvsnfikwowklh.prod-runtime.all-hands.dev`)

| Time (UTC) | Event |
|---|---|
| 15:29:11 | Sandbox started with runtime `tjmlvsnfikwowklh` |
| 16:07:58 | Runtime **paused** (idle 1208s) — K8s Service & Ingress deleted |
| 16:51:00 | Sandbox resumed → new runtime provisioned |
| 17:42:36 | Runtime **paused again** (idle 1360s) — Service & Ingress deleted |
| 18:16:27 | Sandbox resumed again |

After each resume, the VS Code tab continued to show "Bad Gateway" because it was loading the old runtime URL.

## Impact

Users see raw "Bad Gateway" text in the VS Code tab after resuming a conversation. The agent and terminal work fine (using the new runtime), but VS Code is broken with no way to recover from the UI.

## Suggested Fix

1. **Invalidate sandbox cache on resume** — The resume mutation should call `queryClient.invalidateQueries` for the sandbox and VS Code URL query keys, forcing a refetch that picks up the new runtime's URL.

2. **Detect iframe load failure** — Add error detection to the VS Code iframe (e.g., polling runtime status or listening for load errors) and trigger a sandbox data refetch + iframe reload when the URL appears stale.

---

_This issue was created by an AI assistant (OpenHands) on behalf of the user, based on a Datadog investigation._

xingyaoww:

---

## Why

When a conversation is resumed after its runtime was auto-paused, a **new runtime** is provisioned but the VS Code tab still loads the URL from the **old runtime**, resulting in a "Bad Gateway" error. This happens because the TanStack Query cache for sandbox data and VS Code URLs is not invalidated when the resume mutation settles.

## Summary

- Invalidate `["sandboxes"]` and `["unified", "vscode_url"]` query caches in both `invalidateConversationQueries` (used by `useUnifiedResumeConversationSandbox`) and `useV1ResumeConversation`'s `onSettled` callback
- Added tests for both resume hooks verifying sandbox and VS Code URL cache invalidation

## Issue Number

Fixes #13987

## How to Test

1. Start a conversation, open the VS Code tab (should work)
2. Wait for the runtime to idle out and get paused
3. Resume the conversation
4. The VS Code tab should load the new runtime's URL instead of showing "Bad Gateway"

Alternatively, run the new unit tests:
```bash
cd frontend && npx vitest run __tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx __tests__/hooks/mutation/use-v1-resume-conversation.test.tsx
```

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The fix invalidates all sandbox-related caches (using the partial key `["sandboxes"]`) rather than specific sandbox IDs, since the mutation doesn't have direct access to the sandbox ID. TanStack Query's partial key matching ensures all sandbox batch queries are invalidated. This also fixes the same stale-URL issue for worker/active host URLs, which depend on the same sandbox data.

@xingyaoww can click here to [continue refining the PR](https://app.all-hands.dev/conversations/9a5d0557-49fb-43b4-8d03-0a5ed8e952d1)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.openhands.dev/openhands/runtime:ed1cbc1-nikolaik   --name openhands-app-ed1cbc1   docker.openhands.dev/openhands/openhands:ed1cbc1
```

### DIFF
--- a/frontend/__tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
+++ b/frontend/__tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
@@ -85,10 +85,12 @@ describe("useUnifiedResumeConversationSandbox", () => {
     // Verify sandbox queries were invalidated
     const invalidateCalls = invalidateSpy.mock.calls.map((call) => call[0]);
     const sandboxInvalidation = invalidateCalls.find(
-      (call) => JSON.stringify(call?.queryKey) === JSON.stringify(["sandboxes"]),
+      (call) => call?.queryKey?.[0] === "sandboxes",
     );
     const vscodeInvalidation = invalidateCalls.find(
-      (call) => JSON.stringify(call?.queryKey) === JSON.stringify(["unified", "vscode_url"]),
+      (call) =>
+        call?.queryKey?.[0] === "unified" &&
+        call?.queryKey?.[1] === "vscode_url",
     );
 
     expect(sandboxInvalidation).toBeDefined();
--- a/frontend/__tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
+++ b/frontend/__tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
@@ -1,0 +1,97 @@
+import { renderHook, waitFor } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { describe, expect, it, vi, beforeEach } from "vitest";
+import React from "react";
+import { SandboxService } from "#/api/sandbox-service/sandbox-service.api";
+import V1ConversationService from "#/api/conversation-service/v1-conversation-service.api";
+import { useUnifiedResumeConversationSandbox } from "#/hooks/mutation/use-unified-start-conversation";
+
+// Mock the error message store
+vi.mock("#/stores/error-message-store", () => ({
+  useErrorMessageStore: (selector: (state: { removeErrorMessage: () => void }) => unknown) =>
+    selector({ removeErrorMessage: vi.fn() }),
+}));
+
+describe("useUnifiedResumeConversationSandbox", () => {
+  let queryClient: QueryClient;
+
+  beforeEach(() => {
+    queryClient = new QueryClient({
+      defaultOptions: {
+        queries: { retry: false },
+        mutations: { retry: false },
+      },
+    });
+    vi.restoreAllMocks();
+  });
+
+  const wrapper = ({ children }: { children: React.ReactNode }) => (
+    <QueryClientProvider client={queryClient}>
+      {children}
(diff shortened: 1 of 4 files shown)


xingyaoww:
Addressed review feedback in ed1cbc163:
- Consolidated cache invalidation comments and added explanation of partial key matching
- Replaced fragile JSON.stringify test assertions with direct key property checks

Ready for another look.

hieptl (approved):
Thank you! 🙏

PR #13994, opened 2026-04-17
feat(frontend): render ACPToolCallEvent in conversation viewer

simonrosenberg:
## Summary

- ACP agents (Claude Code, Codex, Gemini CLI via the Agent Client Protocol) emit `ACPToolCallEvent`s for every tool call they execute. The frontend had **zero renderers** for them, so every tool call was silently dropped and users saw only the final `MessageEvent`.
- ACP tool call cards now render through the **same pipeline as regular OpenHands observations** — `GenericEventMessageWrapper` → `getEventContent` → `SuccessIndicator`. No bespoke component, no custom pills, no custom collapse UI. The card shape, status icon, markdown body and truncation behavior all match the rest of the conversation view.
- The "kind" differentiator is a translated **`ACP · `** prefix in the title row (e.g. `ACP · Running <cmd>gh pr diff 490</cmd>`) via new `ACTION_MESSAGE$ACP_RUN` / `$ACP_EDIT` / `$ACP_READ` / `$ACP_FETCH` / `$ACP_TOOL` keys that mirror the existing `ACTION_MESSAGE$RUN` pattern with `<cmd>` / `<path>` component wrapping.
- Dedup by `tool_call_id` in the `uiEvents` layer so that, once OpenHands/software-agent-sdk#2866 starts streaming `in_progress → completed` updates, the same card updates in place rather than duplicating.

## Rendering — what maps to what

| Concern | Regular OpenHands observation | ACPToolCallEvent (this PR) |
|---|---|---|
| Card shape | `GenericEventMessageWrapper` + `GenericEventMessage` | **same** |
| Title | `ACTION_MESSAGE$RUN` = `Running <cmd>{command}</cmd>` | `ACTION_MESSAGE$ACP_RUN` = `ACP · Running <cmd>{title}</cmd>` (+ `$ACP_EDIT` / `$ACP_READ` / `$ACP_FETCH` / `$ACP_TOOL`) |
| Body | `getTerminalObservationContent` → markdown with `Command:` + `Output:` fenced blocks | `getACPToolCallContent` → **same** markdown shape; execute uses `Command:`, other kinds use `Input:` JSON block, `is_error` flips `Output:` to `**Error:**` |
[...]

PR #13983, opened 2026-04-16
feat(app-server): route ACP agents to the ACP conversation endpoint

simonrosenberg:
## Summary

Consumes the new `AgentSettings` discriminated union from the SDK (OpenHands/software-agent-sdk#2861) so ACP agents can be configured and launched from the GUI.

- **Conversation routing**: `LiveStatusAppConversationService._build_start_conversation_request_for_user` dispatches on the `agent_settings` variant. ACP path builds a `StartACPConversationRequest`, skips server-only overrides (system prompts, LLM tracing metadata, skills loading, hook-config discovery) because the ACP subprocess owns those, and POSTs to `/api/acp/conversations`. `AppConversationInfo.llm_model` uses `acp_model` for ACP so the conversation list shows the real provider model instead of the `"acp-managed"` placeholder.
- **Storage**: `Settings.update()` / `_normalize_inputs` go through `validate_agent_settings` so both variants persist and round-trip. `from_config()` builds `LLMAgentSettings` directly (legacy LLM-only path).
- **Schema endpoint**: `/api/v1/settings/agent-schema` now serves the combined schema from `export_agent_settings_schema()` — sections carry a `variant` tag (`"llm"` / `"acp"` / `None`) so the frontend filters them by the active `agent_kind` selection.
- **Frontend**: new `/settings/acp` route + nav entry; `SettingsSectionSchema` grows a `variant` field; `getVisibleSettingsSections` hides sections whose `variant` does not match the current `agent_kind` value (defaults to `"llm"` for existing users).
- **Tests**: migrated from the old `AgentSettings(...)` constructor to `LLMAgentSettings` via an `as AgentSettings` import alias.

### mypy ignores

Three `# type: ignore[attr-defined]` comments are added on the new-symbol imports (`ACPAgentSettings`, `LLMAgentSettings`, `validate_agent_settings`, `default_agent_settings`, `export_agent_settings_schema`). Reason: pre-commit mypy is configured to install `openhands-sdk==1.17.0` from PyPI, which doesn't ship the new exports yet. The editable install used at runtime exposes them and the code type-checks fine there. Remove the ignores once the SDK with the union ships and the pinned version bumps.

## Test plan

- [x] `tests/unit/app_server/test_live_status_app_conversation_service.py` — 104 tests pass.
[...]

PR #14004, opened 2026-04-17
feat(app-server): route ACP agents to the ACP conversation endpoint

simonrosenberg:
## Summary

Integration layer for ACP-in-GUI, building on the discriminated-union `AgentSettings` merged as OpenHands/software-agent-sdk#2861. **This is the backend half of #13983**; the settings UX stays in #13983 to keep this one small and reviewable.

When `AgentSettings` is an `ACPAgentSettings` the app server now:

- Persists `agent_kind='acp'` on `AppConversationInfo` (new column via Alembic `009`; existing rows coerce `NULL → 'llm'`).
- POSTs conversation-start to `/api/acp/conversations` on the agent-server, skipping LLM-only concerns (skills, hooks, server overrides).
- Plumbs UI-saved credentials from `llm.api_key` / `llm.base_url` into the subprocess env as provider-native vars (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `GEMINI_API_KEY` + `*_BASE_URL`), **gated on `api_key` being set** so a provider-default `base_url` alone never clobbers an explicit `OH_AGENT_SERVER_ENV` passthrough.
- Branches `conversation_url` and live-status polling to the right router so clients don't 404 on `/api/conversations/{id}` for ACP conversations.

Also redirects the agent-server subprocess stdout/stderr to `{working_dir}/.openhands-agent-server.log` — the previous `PIPE` approach deadlocked once the 64 KB kernel buffer filled without a drain loop.

## Test plan

- [x] New `TestAcpProviderEnv` covers all three provider shapes + empty + api-key-alone + base-url-alone-ignored.
- [x] New `TestAgentKindConversationUrl` asserts both URL shapes (`/api/conversations/{id}` for LLM, `/api/acp/conversations/{id}` for ACP).
- [x] Existing suite still green.
- [ ] End-to-end: local GUI with a real LLM key against the litellm proxy, confirming ACP conversation lifecycle.

## Related

- Builds on OpenHands/software-agent-sdk#2861 (merged).
- Split out of #13983 — the settings UX (frontend) follows in that PR after this lands.
- Frontend live tool-call rendering is separate (#13994).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-0ad0c29   docker.openhands.dev/openhands/openhands:0ad0c29
```


simonrosenberg:
Addressed all review points in ``61e3e4b``:

**🔴 Resource leak (process_sandbox_service.py):** wrapped the log-file ``open(...)`` in a ``with`` block so the parent releases its fd as soon as ``Popen`` dup's it into the child.

**🔴 Silent failure mode / 🟠 duplication:** consolidated the three try/except ImportError blocks into a single ``openhands/utils/sdk_settings_compat.py`` module. The fallback ``validate_agent_settings`` now **raises ``RuntimeError`` when stored data contains ``kind='acp'``** instead of silently coercing to LLM — so a user persisting ACP settings against SDK 1.17.0 fails fast with a clear message instead of seeing their config mangled. The inconsistency you flagged (isinstance returning False in one file, True in another) is gone: one place, one behaviour.

**🟡 SecretStr handling:** replaced the ``hasattr(raw, 'get_secret_value')`` + ``str(raw)`` fallback in ``_acp_provider_env`` with an explicit ``isinstance(SecretStr)`` / ``isinstance(str)`` / ``TypeError`` ladder. ``str(SecretStr)`` yields ``**********`` — the old fallback would have silently plumbed the placeholder as the credential.

**🟢 Unknown agent_kind:** factored the URL-kind branch into a module-level ``_agent_kind_to_router_path(kind)`` helper that warns and falls back to ``/api/conversations`` for values other than ``llm`` / ``acp`` / ``None``. Operators see a log line for future variants instead of bare 404s.

**Tests added:**
- ``TestAgentKindConversationUrl::test_agent_kind_to_router_path_known_kinds``
- ``TestAgentKindConversationUrl::test_agent_kind_to_router_path_unknown_warns_and_falls_back``
- ``TestAcpProviderEnv::test_unexpected_api_key_type_raises``
- ``tests/unit/utils/test_sdk_settings_compat.py`` (loud-fallback behaviour)

Full suite: 110 passed, 8 skipped (the skipped ones require the new SDK symbols — they'll flip on when the SDK release lands and we delete the compat module).

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -319,6 +332,9 @@ async def _start_app_conversation(
             task.agent_server_url = agent_server_url
             yield task
 
+            # Determine if this is an ACP conversation
+            is_acp = isinstance(start_conversation_request, StartACPConversationRequest)
```
VascoSch92: nit: I will put that above router_path as it is used for the first time there

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -334,15 +350,26 @@ async def _start_app_conversation(
...
             response.raise_for_status()
-            info = ConversationInfo.model_validate(response.json())
+            if is_acp:
+                info = ACPConversationInfo.model_validate(response.json())
+            else:
+                info = ConversationInfo.model_validate(response.json())
+
+            # Determine display_model and agent_kind
+            agent_kind = 'acp' if is_acp else 'llm'
+            if is_acp and isinstance(start_conversation_request.agent, ACPAgent):
```
VascoSch92: just a question: can we have `is_acp` True and `isinstance(start_conversation_request.agent, ACPAgent)` false? Because if the answer is no, we can put togheter this if-else block we the previous one.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -456,41 +493,82 @@ async def _get_live_conversation_info(
...
+                results.extend(c for c in infos if c)
+            except httpx.HTTPStatusError:
+                _logger.warning(
+                    f'Error getting ACP conversation status from sandbox {sandbox.id}',
+                    exc_info=True,
+                )
+            except Exception:
+                _logger.exception(
+                    f'Error getting ACP conversation status from sandbox {sandbox.id}',
+                    stack_info=True,
```
VascoSch92, line 571: nit: this can be in an helper method

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -1388,6 +1483,126 @@ async def _build_start_conversation_request_for_user(
...
+        Args:
+            user: User information containing credentials
+
+        Returns:
+            Dictionary of environment variable name -> value pairs for the ACP subprocess.
+        """
+        env: dict[str, str] = {}
+
+        # Translate LLM API key if present (used by claude-code and gemini-cli)
+        if hasattr(user, 'agent_settings') and isinstance(
```
VascoSch92: do you need `hasattr(user, 'agent_settings')` ? UserInfo has this attribute

VascoSch92 (approved):
just a couple of nits

PR #14005, opened 2026-04-17
feat: Tag images instead of rebuilding on git tags; stop publishing v0 runtime image

jlav:

---

## Why

The Github Actions currently build new docker images when tags are pushed.

Rather than build new images, we can just tag the existing image for the SHA that a tag points to. This saves us the time waiting for the build to complete, so we can conduct releases more quickly.

I took this opportunity to also rip out building the deprecated v0 runtime images so that I didn't have to work on updating the CI for the deprecated images to support this new tagging strategy.

## Summary

- Replaces `ghcr-build.yml` + `containers/build.sh` + `containers/app/config.sh` with a reusable `_build-image.yml` and two composite actions; net −348 lines across the CI
surface area.
- Adds `tag-image.yml` that retags existing images on git tag push in ~20 seconds. For semver tags, also applies `X.Y`, `X`, and `latest` aliases.
- Removes runtime image builds and the `containers/runtime/` directory (V0 is being retired — separate follow-up will remove the V0 code paths).

## Issue Number

N/A

## How to Test

Watch the build pipeline for this PR.
I've also already pushed a tag to demonstrate that flow works: `test-simplify-docker-ci`

## Type

- [ ] Bug fix
- [x] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-933b201   docker.openhands.dev/openhands/openhands:933b201
```


enyst (dismissed):
Love the removal of V0 images, thank you!

I'm a bit confused by the effect of the tag change:
- we are moving this codebase to `OpenHands-Web` repository
- reference: [separate repositories](https://github.com/OpenHands/OpenHands/issues/10649)
- the repo is ready, we're testing building a release from it
- does this change things?

I think maybe we can split this PR in two, so we stop V0 runtime publishing now, and see about the rest in the OpenHands-Web repository?

.github/workflows/tag-image.yml
```diff
@@ -0,0 +1,59 @@
+# Adds a git-tag name to existing Docker images.
+# Triggered when a tag is pushed: finds the images built at the tag's commit
+# (tagged `sha-<full>`) and adds the tag name as an alias for the same manifest.
+# Semver tags (X.Y.Z) also get X.Y, X, and latest aliases.
```
raymyers, line 4: Is there any change that we might inadventantly bump `latest` to an older one? I guess not because we don't generally retag old releases or release new patches on an old major/minor.

raymyers (approved):
Great to see this overhaul of the thorny docker process, as well as the retirement of the runtime images.

It's a big PR but risk seems manageable.

raymyers:
I am removing `All Runtime Tests Passed` from the required checks.

.github/workflows/tag-image.yml
jlav: Yes, there is. I think this is carried forward from the existing behavior, but it's something worth fixing in general, especially since this PR already refactors the entire pipeline.

PR #14006, opened 2026-04-17
feat: move PostHog server side

HeyItsChloe:

---

## Why

1. **Server-side authority** — The server is the source of truth for business state (credits, conversation outcomes, terminal states). Client-side events were unreliable and duplicated.
2. **Consent* — All events now respect `user_consents_to_analytics` through a single service.
3. **Type safety** — Each event has a dedicated method with typed parameters instead of ad-hoc `posthog.capture()` calls.
4. **Error handling** — Analytics failures are caught and logged but never interrupts users
5. **Feature environments** — Staging/test traffic is prefixed with `FEATURE_`.

## Summary

- Moved client-side tracking to so now all business events now flow through `AnalyticsService`. 
- Only UI-specific events remaining on the frontend.
- All backend events gated on `user_consents_to_analytics`
- Frontend consent synced via `posthog.opt_in_capturing()` / `posthog.opt_out_capturing()`
- OSS: Only frontend auto-instrumentation (web vitals, errors, pageviews) — no backend events
- Feature environments: Distinct IDs prefixed with `FEATURE_` to isolate test traffic

### Backend Events (15 total — SaaS only)

| # | Event | Trigger | Smoke Test Steps |
|---|---|---|---|
| 1 | `user signed up` | OAuth registration for new users | Sign up with a new account via GitHub/Google OAuth |
| 2 | `user logged in` | Every Keycloak or device auth login | Log in to the app (web or CLI) |
| 3 | `conversation created` | New conversation initialized | Start a new conversation from the web UI |
| 4 | `conversation finished` | Conversation reaches FINISHED/STOPPED state | Complete a conversation (let agent finish or click Stop) |
| 5 | `conversation errored` | Conversation reaches ERROR/STUCK state | Trigger an error (e.g., invalid API key, runtime failure) |
| 6 | `conversation deleted` | User deletes a conversation | Delete a conversation from the conversation panel |
| 7 | `credit purchased` | Stripe checkout completes | Purchase credits through Settings → Billing |
| 8 | `credit limit reached` | Conversation fails due to budget | Run a conversation until credits exhaust |
| 9 | `git provider connected` | User connects GitHub/GitLab/etc. | Connect a git provider in Settings → Secrets |
| 10 | `onboarding completed` | User submits onboarding form | Complete the onboarding wizard as a new user |
| 11 | `settings saved` | User saves settings | Change any setting in Settings modal and save |
| 12 | `trajectory downloaded` | User exports conversation trajectory | Click "Download" on a conversation |
| 13 | `team members invited` | User invites team members | Send invitations from org settings |

### Client-Side UI Events (2 total)

| # | Event | Trigger | Smoke Test Steps |
|---|---|---|---|
| 1 | `enterprise cta clicked` | User clicks "Learn More" enterprise CTA | Click any enterprise CTA button (login page, homepage, context menu) |
| 2 | `enterprise lead form submitted` | User submits enterprise contact form | Fill out and submit the enterprise inquiry form |

### Automatic Logging (SaaS and OSS)

| Feature | What It Captures | Smoke Test |
|---|---|---|
| Web Vitals | LCP, FCP, INP, CLS | Load any page — metrics auto-capture |
| Network Timing | API request latencies | Make any API call |
| Error Tracking | Uncaught JS exceptions | Trigger a frontend error |
| Pageviews | Page navigation | Navigate between pages |
| Session Linking | `X-POSTHOG-SESSION-ID` header | Check Network tab for header on API calls |

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-753917d   docker.openhands.dev/openhands/openhands:753917d
```


enterprise/server/utils/conversation_callback_utils.py
```diff
@@ -1,295 +0,0 @@
-import base64
```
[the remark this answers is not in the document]
HeyItsChloe: This file was used by legacy v0 methods (conversation_manager.py, session_manager.py) that no longer exist. It’s no longer needed, as its functions are not invoked anywhere in the codebase and there are no remaining references.

enterprise/storage/user_store.py
```diff
@@ -85,6 +86,9 @@ async def create_user(
             )
             user.email = user_info.get('email')
             user.email_verified = user_info.get('email_verified')
+            # SaaS consent is implicit via Terms of Service — new SaaS users default to consented
+            if 'saas' in (os.environ.get('OPENHANDS_CONFIG_CLS', '')).lower():
```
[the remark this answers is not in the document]
HeyItsChloe: https://github.com/OpenHands/OpenHands/blob/bd4a094eaf408ccd57f67524def96652eccb3c6d/openhands/analytics/EVENTS.md#consent--privacy

The intent is that SaaS users implicitly consent via Terms of Service acceptance.

enterprise/pyproject.toml
```diff
@@ -36,7 +36,7 @@ resend = "^2.7.0"
 tenacity = "^9.1.2"
 slack-sdk = "^3.35.0"
 ddtrace = "3.13.0"                                           #pin to avoid yanked version 3.12.4
-posthog = "^6.0.0"
+posthog = "^7.0.0"
```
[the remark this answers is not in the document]
HeyItsChloe: The PR removes direct import posthog calls from enterprise code (auth.py no longer calls posthog.set() directly). All PostHog interactions now go through AnalyticsService which uses the same APIs available in both v6 and v7.

v7 breaking changes:

Python 3.9 dropped (we use 3.12)
LLM SDK updates (not used by this PR)

enterprise/saas_server.py
```diff
@@ -61,6 +68,23 @@
     override_llm_models_dependency,
```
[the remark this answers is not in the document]
HeyItsChloe: - This is valid technical debt, but out of scope for this PR
- The current import-order workaround is fragile, but fixing it correctly requires a broader refactor 
- This PR is focused on moving PostHog analytics server-side, so refactoring DI should be handled in a separate PR
- The current safety check ensures we fail fast at import time rather than risk silent misconfiguration at runtime

enterprise/server/routes/auth.py
```diff
@@ -123,9 +131,103 @@ def _extract_oauth_state(state: str | None) -> tuple[str, str | None, str | None
...
+        analytics.track_user_logged_in(
+            distinct_id=user_id,
+            idp=idp,
+            org_id=org_id_str,
+            consented=consented,
+        )
+    except Exception:
+        logger.exception(
+            'auth:_track_login_analytics_background:failed',
+            extra={'user_id': user_id},
```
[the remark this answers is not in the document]
HeyItsChloe: - Existing logger.exception() calls already capture failures and surface in logs (GCP/Datadog)
- For a non-critical service like PostHog, adding metrics/alerts adds overhead with limited value

enterprise/server/app_lifespan/saas_app_lifespan_service.py
```diff
@@ -0,0 +1,46 @@
...
+
+from server.constants import IS_FEATURE_ENV
+
+from openhands.analytics import get_analytics_service, init_analytics_service
+from openhands.app_server.app_lifespan.app_lifespan_service import AppLifespanService
+from openhands.core.logger import openhands_logger as logger
+from openhands.server.types import AppMode
+
+
+class SaasAppLifespanService(AppLifespanService):
```
malhotra5, line 19: Does this lifespan need to be saas specific? I'm guessing local installs also use posthog, and we'd only capture events as long as the user has consented?

malhotra5:
@openhands inspect this PR carefully and understand all of the changes that are being made. Is it possible or easy to split the changes in this PR across multiple PRs instead?

Give an example of what the sequencing for the multiple PRs would look like and what each PR would implement in this case. Be pragmatic; the splitting should be easy to do such that reviewing becomes easier.

enterprise/saas_server.py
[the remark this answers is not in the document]
malhotra5: @openhands don't make any changes. If we were to use dependency injection, can you give an example of what that would look like?

Be very concise. Just give me a small code example of how dependency injection would improve this pattern and what all changes are required in order to achieve this.

If dependency injection isn't the easiest thing to do, also investigate other alternative options. Similarly, give me a very concise description of:
1. What changes are required
2. What would be the most maintainable and predictable pattern here

enterprise/storage/user_store.py
```diff
@@ -85,6 +86,9 @@ async def create_user(
...
+                user.user_consents_to_analytics = True
```
malhotra5: NIT: Do we need to check was "saas" is in the config class? I think this code path only executes in saas so we can do `user.user_consents_to_analytics = True` directly? 

Also I believe this value is set prior to the user accepting terms of service. For safety I think it would be best if we made a helper method "consent_to_analytics" which we call just after confirm that TOS has been accepted during `keycloak_callback` in `enterprise/server/routes/auth.py`

enterprise/saas_server.py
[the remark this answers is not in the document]
malhotra5: ## Dependency Injection vs Import-Order Patching

### Current Problem (lines 71-88)
```python
# Patch global config BEFORE import - order matters!
_config = get_global_config()
_config.lifespan = SaasAppLifespanService()

# Safety check - brittle!
if 'openhands.server.app' in sys.modules:
    raise ImportError('...') 

from openhands.server.app import app as base_app  # Must happen AFTER patch
```

---

### Option 1: Factory Function (Simplest)

**Changes Required:**
1. Modify `openhands/server/app.py` to expose a factory function instead of module-level `app`

**app.py:**
```python
def create_app(lifespan_service: AppLifespanService | None = None) -> FastAPI:
    if lifespan_service is None:
        lifespan_service = get_app_lifespan_service()
    
    lifespans = [mcp_app.lifespan]
    if lifespan_service:
        lifespans.append(lifespan_service.lifespan)
    
    return FastAPI(lifespan=combine_lifespans(*lifespans), ...)

app = create_app()  # OSS default
```

**saas_server.py:**
```python
from openhands.server.app import create_app
from server.app_lifespan.saas_app_lifespan_service import SaasAppLifespanService

base_app = create_app(lifespan_service=SaasAppLifespanService())
```

✅ **Pros**: No import order issues, explicit, easy to test
⚠️ **Changes**: ~10 lines in `app.py`, ~5 in `saas_server.py`

---

### Option 2: Proper DI via Config Injector

**Changes Required:**
1. Add `AppLifespanServiceInjector` protocol
2. Create `SaasAppLifespanServiceInjector` implementation
3. Modify `AppServerConfig` to use injector pattern
4. Update `app.py` to call injector

**config.py:**
```python
class AppLifespanServiceInjector(Protocol):
    def create(self) -> AppLifespanService: ...

class OssAppLifespanServiceInjector:
    def create(self) -> AppLifespanService:
        return OssAppLifespanService()

class AppServerConfig(OpenHandsModel):
    lifespan: AppLifespanServiceInjector = Field(default_factory=OssAppLifespanServiceInjector)
```

**saas_server.py:**
```python
os.environ['OH_LIFESPAN_INJECTOR'] = 'server.app_lifespan.SaasAppLifespanServiceInjector'
from openhands.server.app import app as base_app  # Import order no longer matters
```

✅ **Pros**: Consistent with existing injector pattern, environment-driven
⚠️ **Changes**: ~30 lines across 3 files, more boilerplate

---

### Recommendation

**Go with Option 1 (Factory Function)** - it's:
1. The smallest change (~15 lines total)
2. Most explicit and predictable
3. No hidden environment variables
4. Easy to test in isolation
5. Python idiomatic (FastAPI docs recommend factory pattern)

The existing injector pattern in `AppServerConfig` is good for request-scoped dependencies (DB sessions, services). Lifespan is app-scoped, so a factory function is more appropriate than adding injector machinery.

---
_This comment was generated by an AI assistant (OpenHands)._

enterprise/server/middleware.py
```diff
@@ -198,3 +198,19 @@ async def _logout(self, request: Request):
                 await token_manager.logout(user_auth.refresh_token.get_secret_value())
         except Exception:
             logger.debug('Error logging out')
+
+
+class PostHogSessionMiddleware:
```
malhotra5, line 196: Would we want to include this for local installs as well?

openhands/analytics/analytics_context.py
```diff
@@ -0,0 +1,81 @@
...
+        org_id:    String org_id derived from ``user.current_org_id``, or
+                   ``None`` when unavailable.
+        user:      The full User ORM object, or ``None`` when lookup failed.
+                   Typed as ``Any`` to avoid importing enterprise types.
+    """
+
+    user_id: str
+    consented: bool
+    org_id: str | None
+    user: Any | None
```
malhotra5: NIT: can user field have a stronger typing here?

openhands/analytics/analytics_context.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+async def resolve_context(user_id: str) -> AnalyticsContext:
```
malhotra5: NIT: can we rename this to something more descriptive e.g `resolve_analytics_context`

openhands/analytics/analytics_service.py
```diff
@@ -0,0 +1,695 @@
...
+    def track_user_signed_up(
+        self,
+        distinct_id: str,
+        *,
+        idp: str,
+        email_domain: str | None = None,
+        invitation_source: str = 'self_signup',
+        org_id: str | None = None,
+        session_id: str | None = None,
+        consented: bool = True,
```
malhotra5: NIT: instead of passing org_id, consented, etc can we pass in the entire `AnalyticsContext` object instead. And ideally we have a defensive check that verifies that the analytics contexts contains a consented user before performing a capture

openhands/app_server/app_conversation/app_conversation_router.py
```diff
@@ -364,6 +366,31 @@ async def start_app_conversation(
...
         async_iter = app_conversation_service.start_app_conversation(start_request)
         result = await anext(async_iter)
+
+        # Analytics: conversation created (V1)
+        try:
+            analytics = get_analytics_service()
+            if analytics:
+                user_id = await user_context.get_user_id()
+                if user_id:
+                    ctx = await resolve_context(user_id)
```
malhotra5: I've left a note regarding this in another comment. We resolve the analytics context here, would be nice if we could simply pass that entire object into the public tracking apis (e.g `track_conversation_created` event api) instead of deconstructing the fields of the analytics context

openhands/app_server/event_callback/webhook_router.py
```diff
@@ -262,6 +305,165 @@ async def on_event(
...
+                                    org_id=ctx.org_id,
+                                    consented=ctx.consented,
+                                )
+
+                                # BIZZ-03: credit limit reached (fires alongside conversation errored)
+                                if error_type == 'budget_exceeded':
+                                    analytics.track_credit_limit_reached(
+                                        distinct_id=app_conversation_info.created_by_user_id,
+                                        conversation_id=str(conversation_id),
+                                        credit_balance=None,
```
malhotra5: NIT: for budget exceeded do we need to pass `credit_balance` field if its always going to be `None`

openhands/app_server/event_callback/webhook_router.py
```diff
@@ -262,6 +305,165 @@ async def on_event(
                     event, conversation_id
                 )
 
+        # Analytics: conversation terminal state detection
+        for event in events:
```
malhotra5, line 412: NIT: is it possible to reduce the nesting here?

malhotra5:
@openhands we have a repeating pattern here for tracking events

```
analytics = get_analytics_service() # get service
ctx = await resolve_context(user_id)
analytics.<public_api_to_track_event>(ctx)
```


Don't make any more changes. Inspect the changes on this PR thoroughly and verify whether there's a way to simplify this pattern:

1. See how we do other existing patterns in the codebase.
2. See if there's some way to combine these steps into one.

Also note that today we deconstruct the context object before passing it over to the public APIs which track the events, but it might just make sense to pass the entire context altogether.

Analyze this and use the GitHub token to post back your analysis as a GitHub comment.

malhotra5:
## Analysis: Simplifying the Analytics Tracking Pattern

I've inspected the PR changes thoroughly. You're right that there's a repeating 3-step pattern at call sites:

```python
analytics = get_analytics_service()
ctx = await resolve_context(user_id)
analytics.track_xyz(
    distinct_id=user_id,
    org_id=ctx.org_id,
    consented=ctx.consented,
    # ... event-specific params
)
```

### Existing Patterns in Codebase

Looking at the existing codebase:
- **FastAPI dependency injection** is heavily used: `Depends(get_user_id)`, `Depends(get_user_settings_store)`, etc.
- **UserStore lookups** are done inline in routes (not pre-resolved via dependency)
- There's no existing pattern for bundled "service + context" injection

### Simplification Options

#### Option A: Pass `AnalyticsContext` Directly to Track Methods (Recommended)

Instead of deconstructing `ctx` to pass individual fields, modify track methods to accept the context directly:

**Before:**
```python
analytics.track_settings_saved(
    distinct_id=user_id,
    org_id=ctx.org_id,
    consented=ctx.consented,
    settings_changed=...
)
```

**After:**
```python
analytics.track_settings_saved(ctx, settings_changed=...)
```

The track method would extract `distinct_id`, `org_id`, and `consented` from the context internally. This:
- Eliminates the repetitive field extraction at every call site
- Keeps the `AnalyticsContext` dataclass as the single source of truth
- Can be made backwards compatible (accept both ctx or individual params)

#### Option B: Add FastAPI Dependency for Route Handlers

For route handlers, add a dependency:

```python
async def get_analytics_context(user_id: str | None = Depends(get_user_id)) -> AnalyticsContext | None:
    if not user_id:
        return None
    return await resolve_context(user_id)
```

This would transform routes from:
```python
async def my_route(user_id: str | None = Depends(get_user_id)):
    analytics = get_analytics_service()
    if analytics and user_id:
        ctx = await resolve_context(user_id)
        analytics.track_xyz(distinct_id=user_id, org_id=ctx.org_id, consented=ctx.consented, ...)
```

To:
```python
async def my_route(ctx: AnalyticsContext | None = Depends(get_analytics_context)):
    analytics = get_analytics_service()
    if analytics and ctx:
        analytics.track_xyz(ctx, ...)
```

**Limitation:** This only helps route handlers, not background tasks like `_track_login_analytics_background`.

#### Option C: Single Combined Helper (Not Recommended)

```python
async def track_event(user_id: str, method_name: str, **kwargs):
    analytics = get_analytics_service()
    if not analytics:
        return
    ctx = await resolve_context(user_id)
    getattr(analytics, method_name)(distinct_id=user_id, org_id=ctx.org_id, consented=ctx.consented, **kwargs)
```

This loses type safety and the explicit track method signatures.

### Recommendation

The best simplification is **Option A** (pass `AnalyticsContext` directly), optionally combined with **Option B** for route handlers.

**Implementation approach:**

1. **Update track methods** to accept `ctx: AnalyticsContext | None = None` as an alternative to individual params. When `ctx` is provided, extract `distinct_id`, `org_id`, `consented` from it. This is backwards compatible.

2. **Add `get_analytics_context` dependency** in `openhands/server/user_auth/__init__.py` for route handlers.

This would reduce most call sites from ~6 lines to ~2 lines while maintaining type safety and explicit event tracking.

---

_This comment was created by an AI assistant (OpenHands) on behalf of the reviewer._

openhands/app_server/event_callback/webhook_router.py
HeyItsChloe: Agreed! I moved some of the logic into helper functions to simplify the nesting. LMK if there is a better alternative

openhands/app_server/event_callback/webhook_router.py
HeyItsChloe: Nope! This wasn't needed.

openhands/analytics/analytics_service.py
HeyItsChloe: Updated!

openhands/analytics/analytics_context.py
HeyItsChloe: I don't think so. `Any` was used to avoid importing any enterprise types.

OSS doesn't import enterprise types - not even via TYPE_CHECKING. Is there another way we handle Enterprise types in OSS?

enterprise/server/middleware.py
HeyItsChloe: - Nope, it's not needed for OSS installs
- OSS won't track the server side analytics (get_analytics_service() returns None)
- It would be unused - no OSS code reads the session ID that gets set

enterprise/storage/user_store.py
HeyItsChloe: Updated!
- Removed the OPENHANDS_CONFIG_CLS check 
- Moved consent setting to TOS acceptance - Created a helper that sets user_consents_to_analytics = True when the user accepts Terms of Service during keycloak_callback

openhands/analytics/analytics_context.py
malhotra5: I think user should have an abstract class which is shared between local and SAAS implementations

That could be used here maybe

openhands/analytics/analytics_context.py
```diff
@@ -0,0 +1,81 @@
...
+    """Resolve a user_id into a fully-populated :class:`AnalyticsContext`.
+
+    Performs the UserStore lookup, extracts consent and org_id, and wraps
+    everything in try/except so no exception ever leaks to the caller.
+
+    Returns a safe default (consented=False, org_id=None, user=None) when the
+    user is not found or any error occurs.
+    """
+    try:
+        from storage.user_store import UserStore
```
malhotra5: This is an enterprise import for a file in local dir

Is there something else we can do here?

openhands/analytics/analytics_service.py
```diff
@@ -0,0 +1,625 @@
...
+            event=CREDIT_LIMIT_REACHED,
+            properties={
+                'conversation_id': conversation_id,
+                'credit_balance': credit_balance,
+                'llm_model': llm_model,
+            },
+            session_id=session_id,
+        )
+
+    def track_user_activated(
```
malhotra5: Should we also check for ctx.consented on all these other methods?

openhands/analytics/analytics_context.py
HeyItsChloe: refactored to use the get_impl() pattern with AnalyticsUserProvider, moving the UserStore import to enterprise/analytics/saas_user_provider.py. LMK if there is a better approach

openhands/analytics/analytics_service.py
HeyItsChloe: Nope, all track_* methods should call self.capture() which has the ctx.consented check. 

self.capture should be used in all the other methods.

enterprise/server/routes/auth.py
```diff
@@ -234,6 +342,41 @@ async def keycloak_callback(
...
+                org_id_str = str(user.current_org_id) if user.current_org_id else None
+
+                ctx = AnalyticsContext(
+                    user_id=user_id,
+                    consented=consented,
+                    org_id=org_id_str,
+                    user=user,
+                )
+                analytics.track_user_signed_up(
+                    ctx=ctx,
```
malhotra5: Should this new user analytics check come after the TOS is accepted (after `await UserStore.consent_to_analytics(user_id)` is invoked)?

HeyItsChloe: I think a regression crept in when I moved the consent logic from `UserStore.create_user()` to a separate `consent_to_analytics()` method called in `accept_tos`.

`user.user_consents_to_analytics` should be set during account creation in `UserStore.create_user()`. Otherwise:
- Setting it later in `keycloak_callback` (after create_user() returns): Requires an extra DB trip to update the user. But this would be the next best alternative. 
- Setting it in `accept_tos`: The tracking has already fired in `keycloak_callback` with consent=False, so the event never fires.

I restored the original approach — setting consent directly in UserStore.create_user(), implicitly.

LMK if there's another approach

openhands/analytics/analytics_context.py
HeyItsChloe: Updated! Added `UserBase` class. I chose to use `protocol here instead of abstractmethods  because it doesn't require the enterprise User model to inherit from it. But LMK if `ABC` would be best to use here.

enterprise/server/app_lifespan/saas_app_lifespan_service.py
```diff
@@ -0,0 +1,46 @@
...
+
+    On enter: initialises the PostHog analytics singleton from environment vars.
+    On exit: calls ``analytics_service.shutdown()`` to flush any buffered events.
+    """
+
+    async def __aenter__(self):
+        api_key = os.environ.get('POSTHOG_CLIENT_KEY', '')
+        host = os.environ.get('POSTHOG_HOST', 'https://us.i.posthog.com')
+        config_cls = os.environ.get('OPENHANDS_CONFIG_CLS', '')
+        app_mode = AppMode.SAAS if 'saas' in config_cls.lower() else AppMode.OPENHANDS
```
malhotra5: https://github.com/OpenHands/OpenHands/blob/main/openhands/app_server/app_conversation/live_status_app_conversation_service.py#L2151

Maybe use app mode reference directly if possible

malhotra5: Actually we probably don't need this app mode check, we can pass `AppMode.SAAS` directly to the init analytics service underneath as this class is only mounted in a saas specific file

HeyItsChloe: Updated - passed AppMode.SAAS directly

enterprise/server/routes/auth.py
HeyItsChloe: Updated!

Previous (`user_store` + `auth.py`) :

- Consent was set in `user_store` during user creation (before ToS accepted)
- Tracking in `keycloak_callback` with `is_new_user` check
- Has `invitation_source` (invitation vs self_signup)


Current (`accept_tos`):

- Consent set in `accept_tos` when user actually accepts ToS
- Tracking in `accept_tos` immediately after consent is set
- Tradeoff: Loses `invitation_source` — invitation_token not available in `accept_tos` (unless we pass the invitation_source via query params to `/accept_tos`)


Note
- Moving tracking within the keycloak callback, after the `has_accepted_tos` check causes issues -> when `has_accepted_tos=True`, `is_new_user=False` (because the user was created on a previous call before accepting ToS), so the if `is_new_user:` condition never passes

LMK if there is a better approach

HeyItsChloe:
**Testing Summary:**
- Original approach & docs detailed here: https://github.com/OpenHands/OpenHands/blob/bd4a094eaf408ccd57f67524def96652eccb3c6d/openhands/analytics/EVENTS.md
- Events to test are listed in the PR description table and `openhands/analytics/EVENTS.md`. Includes smoke test steps for each event
- Consent gates - events only fire when `user_consents_to_analytics = True`. Users with False should not trigger events
- Automatic logging (separate from events) - Web Vitals (LCP, FCP, INP, CLS), Network Timing, Error Tracking, Pageviews
- SaaS vs OSS - backend events (15 total) only fire in SaaS mode; OSS mode should only have frontend auto-instrumentation
- Verify in PostHog dashboard - events appear with correct properties after each action

openhands/analytics/analytics_context.py
```diff
@@ -0,0 +1,98 @@
...
+                   safe default (None / missing / error all map to False).
+        org_id:    String org_id derived from ``user.current_org_id``, or
+                   ``None`` when unavailable.
+        user:      The UserBase instance, or ``None`` when lookup failed.
+    """
+
+    user_id: str
+    consented: bool
+    org_id: str | None
+    user: UserBase | None
```
malhotra5, line 48: @openhands don't make any changes, answer the following question; is there a more specific type that we can use here instead of `UserBase`? examine the other types that we already define in this project

openhands/analytics/analytics_context.py
```diff
@@ -0,0 +1,98 @@
...
+    """Get the configured AnalyticsUserProvider implementation.
+
+    Uses get_impl() to load the implementation class specified in
+    server_config.analytics_user_provider_class. Defaults to
+    DefaultAnalyticsUserProvider if not configured.
+    """
+    from openhands.server.shared import server_config
+
+    impl_class = get_impl(
+        AnalyticsUserProvider, server_config.analytics_user_provider_class
```
malhotra5, line 61: @openhands don't make any changes, can you check whether server config is now legacy and if there is server config equivalents defined in the app server folder?

malhotra5 (approved):
LGTM!

PR #14021, opened 2026-04-19
fix(integrations): deduplicate merged repositories by full_name

ChiragArora31:
## Context

When `ProviderHandler.search_repositories` merges results from multiple Git providers, it deduplicates by repository. The helper `_deduplicate_repositories` is supposed to key on `full_name`, but it was checking `full_name` while storing `repo.id` in the `seen` set. Those are different namespaces, so the same `full_name` could appear twice if the two rows had different ids—something that can happen when the same logical repo is returned from more than one provider.

## What changed

- Use `full_name` consistently when recording seen repos (aligned with the docstring and the membership check).
- Add unit tests for the duplicate-`full_name` case, distinct names, and an empty list.

## How to verify

```bash
poetry run pytest tests/unit/integrations/test_provider_repository_dedup.py -v
```

## Changelog (user-facing)

Repository search with multiple Git providers connected no longer shows duplicate entries for the same repository.


ChiragArora31:
Hi @malhotra5, can you please review this

ChiragArora31:
Hi @enyst, can I please get this PR reviewed by someone?

PR #14048, opened 2026-04-21
fix(chat): allow attachment-only submission

### LINKED ISSUE #14047: [Bug]: Send button silently drops attachment-only chat messages

### Bug Description

In a V1 conversation, when a user attaches an image or file (via drag-and-drop, paste, or the paperclip button) but leaves the text input empty, clicking the send (↑) button does nothing. There is no error toast, no validation message, and no visible feedback — the attachment remains in the UI and the message is never sent. The send button is clickable (not disabled), so the interaction looks functional but is silently swallowed.

The same guard prevents Enter-key submission, so there is no keyboard workaround either.

Root cause: `handleSubmit` in `frontend/src/hooks/chat/use-chat-submission.ts` early-returns when the trimmed text is empty, without considering attached images or files. Meanwhile `ChatSendButton` (`frontend/src/components/features/chat/components/chat-input-row.tsx`) only disables the button based on agent/sandbox state, not on input emptiness.

```ts
// frontend/src/hooks/chat/use-chat-submission.ts (lines 18-24)
const handleSubmit = useCallback(() => {
  const message = chatInputRef.current?.innerText || "";
  const trimmedMessage = message.trim();

  if (!trimmedMessage) {
    return;                    // ← attachments are ignored here
  }
  onSubmit(message);
  ...
```

### Expected Behavior

One of:
1. The message sends with the attachment(s) and an empty text body, OR
2. The send button is disabled when there is no text AND no attachments are present, OR
3. An error/validation toast explains that a text message is required alongside attachments.

### Actual Behavior

- The send button is enabled.
- Clicking it with attachments present but no text does nothing.
- No toast, no log, no console error, no UI feedback.
- The attachment previews remain above the input. Re-clicking reproduces the same no-op.

### Steps to Reproduce

1. Open OpenHands web UI and start (or open) a V1 conversation.
2. Leave the chat text input empty.
3. Attach a file or image in one of three ways:
   - Drag-and-drop an image/file onto the input area.
   - Copy an image and paste (Ctrl/Cmd+V) into the input.
   - Click the paperclip icon and select a file.
4. Observe the attachment preview appears above the input (confirming the file was accepted).
5. Click the send (↑) button without typing any text.
6. Observe: nothing happens. Press Enter instead — still nothing.

Local GUI (Docker web interface)

_No response_

### OpenHands Version

main (commit 0387a86af)

### Version Confirmation

### Model Name

_No response_

### Operating System

Linux

### Browser (if using web UI)

Any (Chrome, Firefox)

### Logs and Error Messages

No logs or errors — the click is silently dropped by the early-return in `handleSubmit`.

Relevant files:
- `frontend/src/hooks/chat/use-chat-submission.ts:18-37` — the early-return guard that drops attachment-only submits
- `frontend/src/hooks/chat/use-chat-input-events.ts:77-81` — the same guard prevents Enter-key submission
- `frontend/src/components/features/chat/components/chat-input-row.tsx:55-62` — send button `disabled` prop does not consider attachments or content emptiness

Suggested fix: in `use-chat-submission.ts`, allow submission when `images.length > 0 || files.length > 0`, or disable the send button when the input is empty AND no attachments are present. The `ChatSendButton` already has a `disabled` prop — the parent just needs to compute it from `!trimmedMessage && attachments.length === 0`.

MukundaKatta:
## Summary
- allow chat submission to proceed when the message text is empty but attachments are present
- stop Enter-key handling from blocking attachment-only sends in the same case
- reuse the existing attachment state from the conversation store instead of adding new state

Closes #14047

## Verification
- not run locally because the frontend test/build dependencies are not installed in this sandboxed environment

### DIFF
--- a/frontend/src/components/features/chat/custom-chat-input.tsx
+++ b/frontend/src/components/features/chat/custom-chat-input.tsx
@@ -39,9 +39,13 @@ export function CustomChatInput({
   const {
     submittedMessage,
     clearAllFiles,
+    images,
+    files,
     setShouldHideSuggestions,
     setSubmittedMessage,
   } = useConversationStore();
+
+  const hasAttachments = images.length > 0 || files.length > 0;
 
   // Disable input when conversation is stopped
   const isConversationStopped = sandboxStatus === "MISSING";
@@ -95,6 +99,7 @@ export function CustomChatInput({
     fileInputRef as React.RefObject<HTMLInputElement | null>,
     smartResize,
     onSubmit,
+    hasAttachments,
     resetManualResize,
   );
 
@@ -104,6 +109,7 @@ export function CustomChatInput({
       smartResize,
       increaseHeightForEmptyContent,
       checkIsContentEmpty,
+      hasAttachments,
       clearEmptyContentHandler,
       onFocus,
       onBlur,
--- a/frontend/src/hooks/chat/use-chat-input-events.ts
+++ b/frontend/src/hooks/chat/use-chat-input-events.ts
@@ -13,6 +13,7 @@ export const useChatInputEvents = (
   smartResize: () => void,
   increaseHeightForEmptyContent: () => void,
   checkIsContentEmpty: () => boolean,
+  hasAttachments: boolean,
   clearEmptyContentHandler: () => void,
   onFocus?: () => void,
   onBlur?: () => void,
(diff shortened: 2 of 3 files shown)


MukundaKatta:
Still relevant — this allows submitting a chat message with only an attachment (no text), which is currently blocked. Rebased on main and CI is green. Could a maintainer take a look when there's a chance? Happy to adjust if anything's needed. 🙏

PR #14063, opened 2026-04-22
fix: deliver pending messages in self-hosted installs without session_api_key

### LINKED ISSUE #13716: [Bug]: Pending Messages Not Sent After Conversation Is Ready

### Bug Description

https://github.com/user-attachments/assets/1751e657-ad83-46ab-8ef3-36ba1ec2c443

According to the video above, if a message is typed and sent while the status is 'starting,' the message will not be sent once the conversation is ready.

**Acceptance criteria:**
- The message should be sent once the conversation is ready.

### Expected Behavior

_No response_

### Actual Behavior

_No response_

### Steps to Reproduce

_No response_

CLI (uv tool install)

_No response_

### OpenHands Version

_No response_

### Version Confirmation

### Model Name

_No response_

### Operating System

None

### Browser (if using web UI)

_No response_

### Logs and Error Messages

_No response_

_No response_

octo-patch:
## Summary

Fixes #13716 — pending messages queued during conversation startup are silently dropped in self-hosted / local installations that have no `session_api_key`.

**Root cause**: `_process_pending_messages` was wrapped in `if sandbox.session_api_key:`, so the entire delivery step was skipped for users who run OpenHands without cloud authentication.

**Changes**:
- Remove the `if sandbox.session_api_key:` guard so `_process_pending_messages` is always called once the conversation reaches the READY state.
- Change `session_api_key: str` to `session_api_key: str | None = None` in the method signature.
- Only attach the `X-Session-API-Key` request header when the key is present; fall back to an empty headers dict for unauthenticated installations.
- Fix test fixtures in `test_live_status_app_conversation_service.py`: the three async methods on the pending-message service mock (`update_conversation_id`, `get_pending_messages`, `delete_messages_for_conversation`) are now `AsyncMock` instances so they can be awaited through the now-unconditional call path.

## Test plan

- [x] Unit test `test_start_app_conversation_default_title_uses_first_five_characters` exercises the full `_start_app_conversation` generator, which now always reaches the `_process_pending_messages` call — the updated `AsyncMock` fixtures make the test pass without error.
- [ ] Manual: start OpenHands locally (no `session_api_key`), send a message while the conversation is still starting, confirm the message is delivered once the agent is ready.

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -388,13 +388,12 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
             yield task
 
             # Process any pending messages queued while waiting for conversation
-            if sandbox.session_api_key:
-                await self._process_pending_messages(
-                    task_id=task.id,
-                    conversation_id=info.id,
-                    agent_server_url=agent_server_url,
-                    session_api_key=sandbox.session_api_key,
-                )
+            await self._process_pending_messages(
+                task_id=task.id,
+                conversation_id=info.id,
+                agent_server_url=agent_server_url,
+                session_api_key=sandbox.session_api_key,
+            )
 
         except Exception as exc:
             _logger.exception('Error starting conversation', stack_info=True)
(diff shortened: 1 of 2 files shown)


octo-patch:
Still relevant for #13716 — pushed a formatting-only commit to get the lint job green (it was failing solely on ruff-format). This fix targets the self-hosted path where messages queued during startup are dropped when no session_api_key is set. Happy to rebase if a maintainer can take a look.

neubig:

Thank you for understanding.

PR #14073, opened 2026-04-22
fix: handle nullable datetime in _fix_timezone methods

malhotra5:

---

## Why

Preparation for enabling SQLAlchemy type checking in enterprise mypy configuration. When `sqlalchemy>=2.0` is added to mypy's additional dependencies, type errors occur because `_fix_timezone` accepts `datetime` but is called with `datetime | None` values.

The stored models define `created_at` and `last_updated_at` as `Mapped[datetime | None]`, but `_fix_timezone` expected non-nullable datetime values.

## Summary

- Update `_fix_timezone` in `sql_app_conversation_info_service.py` to accept `datetime | None` parameter
- Update `_fix_timezone` in `sql_shared_conversation_info_service.py` with the same change
- Return current UTC time as a fallback when value is `None`
- Add `utc_now` import to `sql_shared_conversation_info_service.py`

## Issue Number

N/A - Part of SQLAlchemy type checking enablement effort

## How to Test

After adding `sqlalchemy>=2.0` to the enterprise mypy config's additional dependencies:

```bash
cd enterprise
poetry run pre-commit run --all-files --config ./dev_config/python/.pre-commit-config.yaml
```

The 4 `datetime | None` vs `datetime` type errors related to `_fix_timezone` should be resolved.

## Video/Screenshots

N/A - Type annotation changes only

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This PR is part of a series to enable full SQLAlchemy type checking. The dev config change that enables SQLAlchemy type stubs will be merged in a separate PR after all type fixes are in place.

---
_This PR was created by an AI assistant (OpenHands) on behalf of the user._

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-aba35a1   docker.openhands.dev/openhands/openhands:aba35a1
```

### DIFF
--- a/enterprise/server/sharing/sql_shared_conversation_info_service.py
+++ b/enterprise/server/sharing/sql_shared_conversation_info_service.py
@@ -152,6 +152,9 @@ class SQLSharedConversationInfoService(SharedConversationInfoService):
         we assume UTC if the timezone is missing. Returns current UTC time if value is None.
         """
         if value is None:
+            # Fallback for legacy data: use current time to match model defaults.
+            # The DB columns have default=utc_now, so None only occurs in legacy records.
+            # Using utc_now() keeps the API model non-nullable and matches new record behavior.
             return utc_now()
         if not value.tzinfo:
             value = value.replace(tzinfo=UTC)
--- a/enterprise/server/sharing/sql_shared_conversation_info_service.py
+++ b/enterprise/server/sharing/sql_shared_conversation_info_service.py
@@ -26,6 +26,7 @@ from sqlalchemy import select
 from sqlalchemy.ext.asyncio import AsyncSession
 from storage.stored_conversation_metadata_saas import StoredConversationMetadataSaas
 
+from openhands.agent_server.utils import utc_now
 from openhands.app_server.app_conversation.sql_app_conversation_info_service import (
     StoredConversationMetadata,
 )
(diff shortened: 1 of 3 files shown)


malhotra5:
## Review Feedback Addressed

I've pushed a commit that addresses all the review feedback:

### Comment 1 & 2: Documentation for `utc_now()` fallback
**Fixed:** Added inline comments in both files explaining the rationale:

```python
if value is None:
    # Fallback for legacy data: use current time to match model defaults.
    # The DB columns have default=utc_now, so None only occurs in legacy records.
    # Using utc_now() keeps the API model non-nullable and matches new record behavior.
    return utc_now()
```

### Comment 3: Missing test coverage
**Fixed:** Added three unit tests for the `_fix_timezone` method:
- `test_fix_timezone_with_none` - Tests that `None` returns current UTC time
- `test_fix_timezone_with_naive_datetime` - Tests that naive datetime gets UTC timezone
- `test_fix_timezone_with_aware_datetime` - Tests that aware datetime is unchanged

All tests pass:
```
PASSED test_fix_timezone_with_none
PASSED test_fix_timezone_with_naive_datetime
PASSED test_fix_timezone_with_aware_datetime
```

### Comment 4: Typo fix acknowledgement
**No action needed** - This was an acknowledgement of the existing fix. ✅

---
_This comment was posted by an AI assistant (OpenHands) on behalf of the user._

PR #14077, opened 2026-04-22
Make archived conversations read-only without loading states

malhotra5:

---

## Why

When users click on an archived conversation (one with `sandbox_status === "MISSING"`) from the conversation pane to view its trajectory and history, they previously saw:
1. A "starting" status indicator in the chat interface
2. Loading spinners in VS Code tab, Terminal, and other components  
3. An active input box that suggested they could send messages
4. All the interactive tabs (VS Code, Planner, Terminal, Changes) that are not usable

This was misleading because archived conversations cannot be resumed and the sandbox no longer exists. Users should see a clean read-only view similar to the publicly shared conversation view.

## Summary

- Created `ArchivedConversationView` component that shows only the conversation events without VS Code, Planner, Terminal, and other interactive tabs (similar to shared conversation view)
- Modified `useAgentState` hook to return `isArchived: true` and `AgentState.STOPPED` when `sandbox_status === "MISSING"`
- Created `ArchivedBanner` component to show an informative "archived and read-only" message
- Fixed events not loading for archived conversations by waiting for history query to complete before processing
- Updated agent-status to never show loading state for archived conversations
- Added i18n translations for "This conversation is archived and read-only" in 15 languages
- Updated test files to include proper mock data

## Issue Number

Fixes [APP-1381](https://linear.app/all-hands-ai/issue/APP-1381/make-read-only-view-for-archived-conversations)

## How to Test

1. Have an archived conversation (one where the sandbox has been deleted/expired)
2. Click on it from the conversation pane to open it
3. Verify that:
   - The view is simplified like the shared conversation view (no VS Code, Planner, Terminal tabs)
   - An "archived and read-only" banner is shown
   - No loading spinners are shown
   - The conversation events load and display correctly

## Video/Screenshots

Previously 

image

After

image

Now

## Type

- [x] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The archived view is now consistent with the shared/public conversation view
- The `useAgentState` hook now returns an additional `isArchived` boolean property
- All test files that mock `useAgentState` and `useUserConversation` have been updated
- Pre-existing TypeScript errors in `llm-settings.test.tsx` are unrelated to this PR

This PR was created by an AI assistant (OpenHands).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-8928897   docker.openhands.dev/openhands/openhands:8928897
```


hieptl (changes requested):
Hello @malhotra5,

Thank you for taking the time to create this pull request.

That said, I have some reservations about hiding the other tab panels, for the following reasons:

* Based on the demo screenshot, the example shown is a simple “Hello World.” In practice, however, users may rely on automations to make code changes. In such cases, the Changes tab remains important for reviewing updates or debugging. Similarly, the Planner tab is still valuable for users who configure automations to use planning mode before execution, as they would otherwise have no way to review the generated plan in the current UI.

* Additionally, there are scenarios where conversations are archived for reasons unrelated to the automations feature. For example, a user might initiate a conversation, assign tasks to the agent, and later need to revisit that conversation after it has been archived. With the current UI, the absence of other tab panels makes it difficult to review that context effectively.

As an alternative, I would suggest not resuming archived conversations. Instead, we could disable the chat input to prevent follow-up messages and display a badge indicating that the conversation is archived and read-only. This would preserve access to relevant information while maintaining clarity for end users.

Please feel free to share your thoughts or raise any questions.

Thank you very much! 🙏

malhotra5:
Hey, thanks for the feedback.

The problem with archived conversations is that the sandbox cannot be recovered and the volume associated with the sandbox has also been deleted. What this means is that there's no way to recover the changes; as far as I know, the planner and its contents also exist in the sandbox volume as of today. Consequently, the user can only see the events that transpired as part of the conversation.

This is somewhat similar to how public conversation sharing works today. The difference is that for public conversation sharing, the sandbox may exist, but because it's a public conversation, we choose to never access it. As a result, users are unable to access the changes tabs, the planner tabs, and so on (which is intended!).

The same holds true here. Because the conversation has been explicitly archived, the volume and the sandbox are no longer accessible. It's not a matter of resuming the sandbox per se; the data/volume has been wiped already.

frontend/src/components/features/conversation/archived-conversation-view.tsx
hieptl, line 1: **suggestion (blocking):** It looks like we are missing tests for this component. Could you please add them when you have a chance? Thank you! 🙏

hieptl (approved):
Thank you! 🙏

frontend/src/components/features/conversation/archived-conversation-view.tsx
malhotra5: @openhands please do this

PR #14112, opened 2026-04-24
fix: preserve LLM and MCP settings in migration 108

neubig:

---

## Why

This PR fixes two related settings-loss issues in the forward migration path to enterprise org/org-member settings.

1. **MCP settings lost or invalid in migration 108**
   - Revision `108` moves legacy flat settings into nested `agent_settings` / `conversation_settings` JSON columns.
   - The original migration did not copy legacy `user_settings.mcp_config` into nested `agent_settings.mcp_config`, so MCP configuration could disappear when `108` runs.
   - In staging, post-upgrade `/api/v1/settings` failures showed the migrated settings payload could still fail validation on malformed/legacy MCP shape, so preserving the nested MCP config in `108` is required for the forward path.

2. **LLM setting replaced with the default during first post-upgrade settings load**
   - The issue is not just Alembic `108` by itself. The first `GET /api/v1/settings` after deploy can lazily migrate a legacy user via `UserStore.get_user_by_id()` -> `UserStore.migrate_user()`.
   - In the pre-fix code, `migrate_user()` could leave the new org at an old `org_version` when the user was treated as not having custom LLM settings.
   - Later in the same settings load, `SaaSSettingsStore.load()` calls `OrgStore.get_org_by_id_async()`, which runs `_validate_org_version()`. That validation path persists the current default LLM into org settings. In staging, that default resolved to `claude-sonnet-4`.
   - If malformed MCP settings are also present, the same request can then still fail final `Settings(...)` validation and return 500 — meaning a single first-read path can both write the default LLM and fail the settings response.
   - The fix is to mark migrated orgs as already being on the current `ORG_SETTINGS_VERSION` so the later version-bump path does not overwrite the migrated LLM.

## Summary

- copy legacy `user_settings.mcp_config` into nested `agent_settings.mcp_config` in migration `108`
- preserve migrated org LLM settings during user/org upgrade by setting migrated orgs to the current `ORG_SETTINGS_VERSION`
- keep the existing custom-LLM split behavior, where custom user models still live in org-member diffs and org defaults stay minimal
- add focused unit coverage for migration `108` mappings and the org migration kwargs logic

## Issue Number

APP-1378

## How to Test

- `PYTHONPATH="/workspace/project/OpenHands/enterprise:/workspace/project/OpenHands:$PYTHONPATH" poetry run pytest --noconftest enterprise/tests/unit/test_migration_108_add_agent_settings.py`
- `python3 -m py_compile enterprise/storage/user_store.py enterprise/tests/unit/test_user_store.py`
- `cd enterprise && /home/openhands/.cache/pypoetry/virtualenvs/openhands-ai-7OOj3LbZ-py3.13/bin/pre-commit run --all-files --show-diff-on-failure --config ./dev_config/python/.pre-commit-config.yaml`
- In staging, run revision `108` from a pre-108 database whose old settings include a non-default visible LLM and MCP config, then verify the upgraded org/org-member effective settings still show the same LLM and MCP values.

## Video/Screenshots

N/A

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- This PR intentionally updates revision `108` instead of adding a repair migration because staging can run `108` fresh before prod rollout.
- The PR is intentionally limited to the migration/user-upgrade paths and their targeted unit coverage.
- The key rollout nuance is that `108` and the lazy enterprise user-migration path interact: the first successful or partially successful settings read after deploy can exercise both.

_This PR description was updated by an AI assistant (OpenHands) on behalf of the user._

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-6943885   docker.openhands.dev/openhands/openhands:6943885
```


hieptl (approved):
Thank you! 🙏

PR #14110, opened 2026-04-23
APP-1325: register GitlabV1CallbackProcessor for deserialization

aivong-openhands:
## Summary

Import `GitlabV1CallbackProcessor` at app startup to ensure it's registered as a known subclass of `EventCallbackProcessor` before webhook events try to deserialize callbacks from the database.

This follows the same pattern already used for `GithubV1CallbackProcessor`.

## Problem

When using the V1 GitLab resolver (`ENABLE_V1_GITLAB_RESOLVER=true`), the callback is registered when a conversation starts, but fails to execute when webhook events are received:

```
ValidationError: Unknown kind 'GitlabV1CallbackProcessor' for EventCallbackProcessor;
Expected one of: ['SlackV1CallbackProcessor', 'LoggingCallbackProcessor', 'SetTitleCallbackProcessor']
```

## Solution

Import `GitlabV1CallbackProcessor` in `saas_server.py` when the GitLab integration is enabled, following the same pattern as `GithubV1CallbackProcessor`.

Fixes #14107

---
_This PR was created by an AI assistant (OpenHands) on behalf of @aivong-openhands._

@aivong-openhands can click here to [continue refining the PR](https://app.all-hands.dev/conversations/6631b3b3-67d2-4511-848e-f28298128490)

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-169d937   docker.openhands.dev/openhands/openhands:169d937
```

PR #14113, opened 2026-04-24
fix(enterprise): register GitlabV1CallbackProcessor at startup

### LINKED ISSUE #14107: GitlabV1CallbackProcessor not registered for deserialization in webhook router
## Bug Description

When using the V1 GitLab resolver (`ENABLE_V1_GITLAB_RESOLVER=true`), the `GitlabV1CallbackProcessor` callback is registered when a conversation starts, but fails to execute when webhook events are received from the runtime.

## Error

```
ValidationError: 1 validation error for EventCallback
processor
  Value error, Unknown kind 'GitlabV1CallbackProcessor' for openhands.app_server.event_callback.event_callback_models.EventCallbackProcessor;
  Expected one of: ['SlackV1CallbackProcessor', 'LoggingCallbackProcessor', 'SetTitleCallbackProcessor']
```

## Root Cause

The `DiscriminatedUnionMixin` uses `get_known_concrete_subclasses()` to dynamically discover subclasses of `EventCallbackProcessor`. The `GitlabV1CallbackProcessor` class is not imported in the webhook router code path, so it is not registered when callbacks are deserialized from the database.

### Flow:
1. Conversation starts - enterprise code imports `GitlabV1CallbackProcessor` and registers callback to DB
2. Runtime sends webhook events - `webhook_router.py` tries to deserialize callbacks
3. `EventCallback.model_validate()` fails because `GitlabV1CallbackProcessor` is not in the known subclasses list
4. Summary comment never posts to GitLab

## Impact

- GitLab resolver V1 summary comments are not posted to MRs/issues
- The error occurs silently ("Task exception was never retrieved")

## Proposed Fix

Import `GitlabV1CallbackProcessor` early in the application startup to ensure it is registered before callback deserialization.

## Environment

- OpenHands Replicated deployment
- `ENABLE_V1_GITLAB_RESOLVER: "true"`

he-yufeng:
Fixes #14107.

## Bug

With \`ENABLE_V1_GITLAB_RESOLVER=true\`, V1 summary comments never reach the MR/issue. The webhook router fails to hydrate the stored callback with:

\`\`\`
ValidationError: 1 validation error for EventCallback
processor
  Value error, Unknown kind 'GitlabV1CallbackProcessor' for
  openhands.app_server.event_callback.event_callback_models.EventCallbackProcessor;
  Expected one of: ['SlackV1CallbackProcessor', 'LoggingCallbackProcessor', 'SetTitleCallbackProcessor']
\`\`\`

and the error is silently swallowed as "Task exception was never retrieved".

## Root cause

\`DiscriminatedUnionMixin.get_known_concrete_subclasses()\` only returns subclasses whose defining module has already been imported at the point deserialization runs. \`GitlabV1CallbackProcessor\` lives in \`enterprise/integrations/gitlab/gitlab_v1_callback_processor.py\` and is referenced **only** from a lazy inline import inside \`gitlab_view._create_gitlab_v1_callback_processor\`. Webhook deserialization never touches that code path, so the subclass never gets registered.

The GitHub branch of \`saas_server.py\` already handles this: under \`if GITHUB_APP_CLIENT_ID:\` it eagerly imports \`GithubV1CallbackProcessor\` with an explicit comment ("Make sure that the callback processor is loaded here so we don't get an error when deserializing"). The GitLab branch missed the same treatment.

## Fix

Mirror the GitHub pattern in the \`if GITLAB_APP_CLIENT_ID:\` block — eagerly import \`GitlabV1CallbackProcessor\` at startup and log its name so mypy/dead-code analyzers don't drop the import.

## Test plan

- With \`GITLAB_APP_CLIENT_ID\` set: startup logs \`Loaded GitlabV1CallbackProcessor\`; webhook events that previously 404'd on deserialization now hydrate successfully, and the V1 summary comment reaches the MR.
- With \`GITLAB_APP_CLIENT_ID\` unset: unchanged behavior (the entire block is skipped).
- No public API change; no impact on GitHub / Slack / Jira paths.

### DIFF
--- a/enterprise/saas_server.py
+++ b/enterprise/saas_server.py
@@ -106,7 +106,17 @@ if GITHUB_APP_CLIENT_ID:
 
 # Add GitLab integration router only if GITLAB_APP_CLIENT_ID is set
 if GITLAB_APP_CLIENT_ID:
+    # Make sure that the callback processor is loaded here so we don't get an
+    # error when deserializing EventCallback rows from the database — the
+    # class otherwise only gets imported lazily inside gitlab_view and stays
+    # invisible to DiscriminatedUnionMixin.get_known_concrete_subclasses().
+    from integrations.gitlab.gitlab_v1_callback_processor import (  # noqa: E402
+        GitlabV1CallbackProcessor,
+    )
     from server.routes.integration.gitlab import gitlab_integration_router  # noqa: E402
+
+    # Bludgeon mypy into not deleting my import
+    logger.debug(f'Loaded {GitlabV1CallbackProcessor.__name__}')
 
     base_app.include_router(gitlab_integration_router)
 


malhotra5:
Hi @he-yufeng! Thanks for the contribution. I'll be closing this as a duplicate PR #14110 has already been merged to address the linked issue.

PR #14127, opened 2026-04-25
Reduce GitHub resolver comment noise by editing acknowledgement comment

### LINKED ISSUE #13737: [Feature]: GitHub resolver should avoid leaving both the initial "I'm on it" comment and a separate final result comment
## Is there an existing feature request for this?

## Problem or Use Case
Today the GitHub resolver leaves two user-visible comments for a single run:

1. an initial acknowledgement like `I'm on it! ... can track my progress at all-hands.dev`
2. a separate final result/summary comment when the run completes

That creates unnecessary comment noise on issues/PRs. For many resolver runs, the acknowledgement and the final result appear close together, so the intermediate comment mostly adds clutter.

From the current implementation:
- `enterprise/integrations/github/github_manager.py` posts the initial `I'm on it! ... track my progress` comment
- `enterprise/integrations/github/github_v1_callback_processor.py` later posts a separate summary comment with `issue.create_comment(summary)` (or reply for inline PR comments)

## Why this matters
On GitHub, extra resolver comments are especially costly because they make issues and PRs harder to follow.

Unlike Slack, GitHub threads are often used as long-lived records for debugging, planning, and review. When the resolver posts both a placeholder comment and a final result comment, it increases the amount of comment churn readers need to scan through.

The current two-comment flow creates a few concrete problems:
- it makes issue and PR discussions harder to follow
- it adds unnecessary comment noise for maintainers and subscribers
- it makes it harder to quickly find the actual final result
- GitHub performance can also suffer when threads get overly busy; more comments means more to load, render, and mentally filter

The acknowledgement is still useful, but ideally users should end up with **one** clear final visible resolver comment rather than both the placeholder and the result.

## Proposed Solution
Please change the GitHub resolver flow so it does **not** leave both comments behind.

Two reasonable options:

1. **Preferred:** edit the original `I'm on it` comment in place with the final result when the run completes.
2. **Alternative:** if editing is awkward or undesirable, delete the original `I'm on it` comment once the final result comment is posted.

If option 2 is chosen, the **final** comment should still include the conversation link that is currently carried by the first comment, so users can still jump into the OpenHands conversation.

## Alternatives Considered
- Keep the current two-comment flow: simple, but noisy.
- Remove the initial comment entirely: worse UX because users lose immediate confirmation that the resolver picked up the issue/PR.

## Priority / Severity

## Estimated Scope
Small - UI tweak, config option, or minor change

## Feature Area
Integrations (GitHub, GitLab, etc.)

The initial comment already contains the conversation link, and the callback processor already knows when the run is complete. It should be possible to persist the created comment id and then either:
- edit that comment on completion, or
- delete it after posting the final completion comment.

Either path should preserve the conversation link in the final visible comment.

## Additional Context
This request is specifically about the **two-comment UX** for GitHub resolver runs, not about whether the final content should be a summary vs the actual final assistant message.

_This issue was created and updated by an AI assistant (OpenHands) on behalf of a user._

Lumen-Founder:
## Summary

This PR reduces GitHub resolver comment noise by editing the original acknowledgement comment with the final resolver summary when the run completes.

Today a single resolver run can leave two visible comments:

1. the initial `I'm on it! ...` acknowledgement
2. a separate final result/summary comment

This PR keeps the immediate acknowledgement, but replaces it in place with the final result when possible.

Closes #13737.

## What changed

- Adds a hidden resolver acknowledgement marker to the initial GitHub comment.
- On completion, the callback processor looks for that marker in recent comments.
- If found, the original acknowledgement comment is edited with the final summary.
- If not found, the existing fallback behavior remains: create a new summary comment.
- The final summary preserves the OpenHands conversation link.

## Why

This addresses #13737 by reducing duplicate visible resolver comments while preserving the existing user feedback loop.

Users still get immediate confirmation that the resolver started, and the thread ends with one clear final visible resolver comment.

## Implementation notes

The implementation intentionally avoids new persistence or schema changes.

Instead of storing the acknowledgement comment ID in a database, the initial comment includes a hidden HTML correlation token containing the conversation ID. The callback processor uses that token to correlate the final summary with the original acknowledgement comment.

To avoid unnecessary API usage on long-lived threads, the lookup scans recent comments first and stops after a bounded number of comments.

## Fallback behavior

If the original acknowledgement comment cannot be found or edited, the callback processor falls back to the current behavior and creates a new summary comment.

GitHub posting failures that happen after the resolver has already completed are treated as a best-effort side effect for non-writable/deleted/locked/rate-limited threads. Unexpected errors still raise.

## What this does not change

This PR does not:

- remove the initial acknowledgement
- change resolver triggering behavior
- add a new external service
- add a database migration
- change GitHub workflow permissions
- change branch or PR creation logic

## AI assistance note

This PR was prepared with AI assistance and reviewed before submission.

- Scope: GitHub resolver comment flow only
- Files touched: GitHub integration comment helpers and callback posting path
- External services added: none
- Persistence changes: none
- Workflow permission changes: none
- Fallback behavior: preserves existing create-comment behavior if edit-in-place correlation fails

### DIFF
--- a/enterprise/integrations/github/github_comment_utils.py
+++ b/enterprise/integrations/github/github_comment_utils.py
@@ -1,0 +1,63 @@
+from __future__ import annotations
+
+from typing import Any, Iterator
+from uuid import UUID
+
+from integrations.utils import CONVERSATION_URL
+
+_ACK_MARKER_TEMPLATE = '<!-- openhands-ack:{} -->'
+
+
+def build_ack_marker(conversation_id: UUID | str) -> str:
+    """Build a reusable GitHub acknowledgement marker for a resolver conversation."""
+    return _ACK_MARKER_TEMPLATE.format(conversation_id)
+
+
+def append_ack_marker(message: str, conversation_id: UUID | str) -> str:
+    """Append the acknowledgement marker to a message body if it is not already present."""
+    marker = build_ack_marker(conversation_id)
+    if marker in message:
+        return message
+
+    if message.endswith('\n'):
+        return f'{message}{marker}'
+
+    return f'{message}\n\n{marker}'
+
+
+def ensure_conversation_link(message: str, conversation_id: UUID | str) -> str:
+    """Ensure the conversation link is included in the message body."""
+    link_fragment = f'conversations/{conversation_id}'
+    if link_fragment in message:
+        return message
+
+    conversation_link = CONVERSATION_URL.format(conversation_id)
+    return f'{message.rstrip()}\n\nTrack progress [here]({conversation_link})'
+
+
+def build_final_resolver_comment(summary: str, conversation_id: UUID | str) -> str:
+    """Build the final resolver comment with link and acknowledgement marker."""
+    summary_with_link = ensure_conversation_link(summary.strip(), conversation_id)
+    return append_ack_marker(summary_with_link, conversation_id)
+
+
+def iter_recent_paginated_items(
+    paginated_list: Any, max_items: int | None = None
+) -> Iterator[Any]:
+    """Iterate over a GitHub paginated list from newest to oldest.
+
+    This assumes the underlying pages are ordered oldest-first, and it iterates
+    pages in reverse, yielding each page's items in reverse order.
+    """
+    page_count = getattr(paginated_list, 'totalCount', None)
+    if page_count is None:
+        raise ValueError('Paginated list does not expose totalCount')
+
+    yielded = 0
+    for page_index in range(page_count - 1, -1, -1):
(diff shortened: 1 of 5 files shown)


Lumen-Founder:
Opened this as a focused implementation for #13737.

The intent is to keep the current immediate acknowledgement behavior, but avoid leaving both the acknowledgement and final result as separate visible comments.

Implementation summary:

- add a hidden conversation correlation token to the initial acknowledgement comment
- on completion, look for that token in recent comments
- edit that comment with the final summary when found
- fall back to the current create-comment behavior when not found

No persistence/schema changes, no workflow changes, and no change to resolver triggering.

PR #14141, opened 2026-04-27
feat(enterprise): Self hosted gitlab support

jlav:

---

## Why

Gitlab Cloud is supported for enterprise deployments, but self hosted is not. This PR adds support for self hosted gitlab.

## Summary

Removes hardcoded `gitlab.com` and replaces it with an env var, `GITLAB_HOST` that defaults to `gitlab.com`.

On the frontend, I've removed the hardcoded host name defaults for git providers and now call the `config` endpoint to grab the host names. This is a follow up on a TODO comment from when`codeburg` was introduced that also works well for Gitlab, so this seemed like a good time to follow up on it:
```
// Default UI links to Codeberg unless a custom host is available in settings
// Note: UI link builders don't currently receive host; consider plumbing settings if needed
```

## How to Test

I've deployed this to one of my replicated instances and pointed it to a self hosted gitlab instance.
I've validated that:
1. I can start repo conversations. Listing repos and branches works.
2. The repo, branch, pull, push, and pull request buttons work as expected
3. My agent can make a change and open a PR

image

image

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-8c93b62   docker.openhands.dev/openhands/openhands:8c93b62
```

### DIFF
--- a/enterprise/tests/unit/integrations/gitlab/test_gitlab_service.py
+++ b/enterprise/tests/unit/integrations/gitlab/test_gitlab_service.py
@@ -12,6 +12,16 @@ def gitlab_service():
     return SaaSGitLabService(external_auth_id='test_user_id')
 
 
+class TestSaaSGitLabServiceInit:
+    """Tests for SaaSGitLabService __init__."""
+
+    def test_explicit_base_domain_overrides_default(self):
+        """An explicit base_domain parameter overrides the upstream class default."""
+        service = SaaSGitLabService(external_auth_id='u1', base_domain='other.host')
+
+        assert service.BASE_URL == 'https://other.host/api/v4'
+
+
 class TestGetUserResourcesWithAdminAccess:
     """Test cases for get_user_resources_with_admin_access method."""
 
--- a/frontend/src/hooks/use-provider-host.ts
+++ b/frontend/src/hooks/use-provider-host.ts
@@ -1,0 +1,21 @@
+import { useConfig } from "#/hooks/query/use-config";
+import { useSettings } from "#/hooks/query/use-settings";
+import { Provider } from "#/types/settings";
+
+/**
+ * Resolve the effective host for a git provider: a per-token override from
+ * user settings if present, otherwise the deployment-wide default from the
+ * web client config, otherwise null.
+ */
+export const useProviderHost = (
+  provider: Provider | null | undefined,
+): string | null => {
+  const { data: settings } = useSettings();
+  const { data: config } = useConfig();
+  if (!provider) return null;
+  return (
+    settings?.provider_tokens_set[provider] ||
+    config?.provider_default_hosts?.[provider] ||
+    null
+  );
+};
--- a/frontend/src/utils/utils.ts
+++ b/frontend/src/utils/utils.ts
@@ -182,32 +182,9 @@ export const shouldUseInstallationRepos = (
   }
 };
 
-export const getGitProviderBaseUrl = (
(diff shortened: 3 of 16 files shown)


frontend/src/utils/utils.ts
```diff
@@ -182,32 +182,9 @@ export const shouldUseInstallationRepos = (
...
-      return "https://dev.azure.com";
-    case "forgejo":
-      // Default UI links to Codeberg unless a custom host is available in settings
-      // Note: UI link builders don't currently receive host; consider plumbing settings if needed
-      return "https://codeberg.org";
-    default:
-      return "";
-  }
+export const getGitProviderBaseUrl = (host?: string | null): string => {
+  if (!host || host.trim() === "") return "";
```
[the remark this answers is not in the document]
jlav: All the callsites are updated, so there is no reason to leave a comment.

enterprise/tests/unit/integrations/gitlab/test_gitlab_service.py
```diff
@@ -12,6 +12,16 @@ def gitlab_service():
     return SaaSGitLabService(external_auth_id='test_user_id')
 
 
+class TestSaaSGitLabServiceInit:
+    """Tests for SaaSGitLabService __init__."""
+
+    def test_explicit_base_domain_overrides_default(self):
+        """An explicit base_domain parameter overrides the upstream class default."""
+        service = SaaSGitLabService(external_auth_id='u1', base_domain='other.host')
```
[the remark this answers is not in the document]
jlav: Tests pass. `GitLabService.__init__` expects that parameter.

frontend/src/utils/utils.ts
```diff
@@ -182,32 +182,9 @@ export const shouldUseInstallationRepos = (
...
-    case "azure_devops":
-      return "https://dev.azure.com";
-    case "forgejo":
-      // Default UI links to Codeberg unless a custom host is available in settings
-      // Note: UI link builders don't currently receive host; consider plumbing settings if needed
-      return "https://codeberg.org";
-    default:
-      return "";
-  }
+export const getGitProviderBaseUrl = (host?: string | null): string => {
```
malhotra5: NIT: seems like this method is simply checking for http prefix now. maybe we rename this method accordingly

frontend/src/hooks/use-provider-host.ts
```diff
@@ -0,0 +1,21 @@
...
+): string | null => {
+  const { data: settings } = useSettings();
+  const { data: config } = useConfig();
+  if (!provider) return null;
+  return (
+    settings?.provider_tokens_set[provider] ||
+    config?.provider_default_hosts?.[provider] ||
+    null
+  );
+};
```
malhotra5, line 21: Nice catch! This seems to resolve token specific overrides for local installs and global configs for saas deployments

malhotra5 (approved):
Overall LGTM. Would be great to address the bot review comments and I left a nit regarding a method name as well

frontend/src/utils/utils.ts
jlav: Done, it's now `ensureHttpsPrefix`.

PR #14142, opened 2026-04-27
fix: enforce onboarding completion on every navigation

hieptl:

---

## Why

**Title:** Onboarding can be bypassed (or re-entered) via manual URL navigation

Today, the backend only redirects users to `/onboarding` during the OAuth callback / TOS acceptance flow (`enterprise/server/routes/auth.py:_should_redirect_to_onboarding`). Once that initial redirect lands the user on `/onboarding`, the frontend imposes no client-side gate, which produces two regressions:

- **Issue 1 — escape:** A new SaaS user lands on `/onboarding`. If they manually edit the URL to `/`, they reach the home page and can use the product without completing onboarding.
- **Issue 2 — re-entry:** After completing onboarding, manually navigating back to `/onboarding` re-renders the form. `OnboardingForm.clientLoader` only validates `app_mode === "saas"` and never checks whether the user is already onboarded.

### Why the existing `is_new_user` flag does not solve this

`is_new_user` is a frontend-only field. The backend never produces it. `normalizeSettingsResponse` (`frontend/src/hooks/query/use-settings.ts:94`) hardcodes it to `false` whenever a settings row exists, and `UserStore.create_user` (`enterprise/storage/user_store.py:65`) provisions a settings row at sign-up — so the 404 fallback that would yield `is_new_user: true` never fires for SaaS users in production. The authoritative signal lives only on `User.onboarding_completed` and was never exposed to the client.

### Expected behaviour

- Users with incomplete onboarding are forced to `/onboarding` regardless of how they navigate.
- Users who have completed onboarding cannot re-enter `/onboarding`.

## Summary

- The backend already knew, via `_should_redirect_to_onboarding`, whether a user must finish onboarding. That logic only ran on the OAuth/TOS callback, so any subsequent navigation was ungated.
- Expose that decision over a new `GET /api/onboarding_status` endpoint (`{ should_complete_onboarding: boolean }`) and consume it from a single React Query hook on the client.
- Add an `OnboardingGuard` (modeled on `EmailVerificationGuard`) that wraps the root-layout outlet and redirects ungated users to `/onboarding`. Add the inverse check inside `OnboardingForm` so already-onboarded users cannot re-enter the form.
- Invalidate the new query inside `useSubmitOnboarding` so the gate flips immediately after the user finishes onboarding.
- Remove `/onboarding` from `INTERMEDIATE_PAGE_PATHS` so `useConfig`/`useIsAuthed`/`useOnboardingStatus` actually fire on the onboarding page (the user is past TOS at that point and root-layout is not mounted there, so this has no side effects).

## How to Test

1. Please run the SaaS code locally.
2. Sign in to the product using a new GitHub account.
3. You should be redirected to the `/onboarding` page. If you remove `/onboarding` from the URL, you should still be redirected back to the onboarding page, as the onboarding process has not yet been completed.
4. After completing the onboarding process, you will be redirected to the home page. If you then enter `/onboarding` in the browser’s address bar, you should no longer see the onboarding page and should instead be redirected to the home page.

## Video/Screenshots

****Video 1:** Demonstrates the behavior when the onboarding process has not yet been completed and `/onboarding` is removed from the URL.**

https://github.com/user-attachments/assets/e4bf4d9f-ddcb-4730-8897-f0515c2f1372

**Video 2:** Demonstrates the behavior after the onboarding process has been completed. When `/onboarding` is entered again in the browser’s address bar, the onboarding page should no longer be accessible and the user should be redirected to the home page.**

https://github.com/user-attachments/assets/5845ed5e-efc6-4f7c-9288-c34f6ee19b34

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-a7f0b6c   docker.openhands.dev/openhands/openhands:a7f0b6c
```


tofarr (changes requested):
I suggest we add `onboarding_completed` (or `should_complete_onboarding`) as a field on the existing settings response — exactly how email_verified works. Then the guard could use useSettings() just like EmailVerificationGuard does, with no new endpoint, no new service, and no new hook. This would also eliminate an extra HTTP request on every page load.

tofarr (approved):
Then I think we should go with your original approach - it would have been nice to have this new value in the /settings endpoint, but there is a good reason for not doing that (Performance) - once that is documented I approve.

🍰

PR #14144, opened 2026-04-27
fix(frontend): block /settings/org-defaults* routes in OSS mode

hieptl:

---

## Why

**Title:** Restrict `/settings/org-defaults*` routes to SaaS mode

### Description

In OSS mode, the `/settings/org-defaults*` routes are accessible by direct URL navigation, even though "org defaults" is a team-organization concept that does not exist outside SaaS.

### Steps to Reproduce

1. Run the application in OSS mode.
2. Manually enter each of the following URLs in the browser address bar:
   - `/settings/org-defaults`
   - `/settings/org-defaults/condenser`
   - `/settings/org-defaults/verification`

### Actual Behavior

All three routes render and are fully accessible in OSS mode.

### Expected Behavior

All three routes should be restricted to SaaS mode. In OSS mode, navigating to any of them should redirect to `/settings`.

### Root Cause

`frontend/src/routes/settings.tsx` defines a `SAAS_ONLY_PATHS` allow-list used by the parent settings `clientLoader` to redirect OSS users away from SaaS-only routes. The three `/settings/org-defaults*` paths were missing from this list. The per-route `createPermissionGuard("view_llm_settings")` explicitly returns `null` in OSS mode, so it does not block direct URL access either.

## Summary

- `/settings/org-defaults`, `/settings/org-defaults/condenser`, and `/settings/org-defaults/verification` were reachable in OSS mode by direct URL navigation, even though "org defaults" is a SaaS-only, team-org concept.
- Added the three paths to `SAAS_ONLY_PATHS` in `frontend/src/routes/settings.tsx` so the existing parent-settings `clientLoader` redirects OSS users to `/settings`.
- Added a parameterized regression test covering all three routes.

### Root Cause

`SAAS_ONLY_PATHS` (`frontend/src/routes/settings.tsx`) is the allow-list the parent settings loader uses to redirect OSS users off SaaS-only routes. The three `org-defaults` paths were missing. The per-route `createPermissionGuard("view_llm_settings")` explicitly bypasses checks in OSS mode (`if (config?.app_mode === "oss") return null;`), so it didn't catch this either.

### Changes

- `frontend/src/routes/settings.tsx` — append three paths (`/settings/org-defaults`, `/settings/org-defaults/condenser`, `/settings/org-defaults/verification`) to `SAAS_ONLY_PATHS`. No other code change needed; the existing redirect at line 56 handles the rest.
- `frontend/__tests__/routes/settings.test.tsx` — new `it.each` test that arranges OSS-mode config, renders each path, and asserts the index `llm-settings-screen` rendered (redirect happened) while the org-defaults screen did not.

## How to Test

1. Please run the code locally in OSS mode.
2. Then, try accessing the following routes by entering them directly into the browser’s address bar:

   * `/settings/org-defaults`
   * `/settings/org-defaults/condenser`
   * `/settings/org-defaults/verification`

## Video/Screenshots

https://github.com/user-attachments/assets/0cc82db8-7646-4d8d-ad58-5f70e12ec9c0

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-e5dad1d   docker.openhands.dev/openhands/openhands:e5dad1d
```


tofarr (approved):
🍰

PR #14146, opened 2026-04-27
feat(settings): add saved LLM profiles (BE)

VascoSch92:

## Why

Providers (OpenAI / Anthropic / OpenHands proxy) means re-typing the model,                                                                                                                                                            base_url, and api_key every time — and the api_key is wiped the moment the 
model changes. Saved profiles let users keep multiple ready-to-use LLM                                                                                                                                                                 
configurations and switch between them in one click.
                     
## Summary                                                                                                                                                                                                                             
 - Add `LLMProfiles` (per-user, capped at 10) with save / get / rename / delete and an `active` pointer that auto-heals if it ever points at a missing or divergent entry.
- Wire profiles into `Settings`: `switch_to_profile()` swaps `agent_settings.llm`, and `_reconcile_active_profile()` clears `active` whenever a direct edit to `agent_settings.llm` makes the pointer stale.
- Expose REST endpoints under `/api/v1/settings/profiles`: `GET` (list), `GET/POST/DELETE /{name}`, `POST /{name}/activate`, `POST /{name}/rename`. API keys are never echoed back — responses carry an `api_key_set: bool` flag       
  instead.                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                                                                                                 
## Type
  - [ ] Bug fix   
  - [x] Feature
  - [ ] Refactor
  - [ ] Breaking change
  - [ ] Docs / chore                                                                                                                                                                                                                     
                    
  ## Notes                                                                                                                                                                                                                               
                  
  - No migration required: `llm_profiles` defaults to empty on existing settings rows. The legacy `POST /api/v1/settings` endpoint ignores any `llm_profiles` payload so the contract stays clean — all profile mutations go through the
  dedicated routes.                                                                                                                                                                                                                      
  - Profile writes are serialized per-user with an in-process asyncio lock to close the lost-update race within a worker. Multi-worker SaaS deployments still need DB-level row locks or optimistic concurrency tokens for full safety;
  this PR explicitly does not claim to provide them.                                                                                                                                                                                     
  - Two pre-existing tests were updated to the `agent_settings_diff` payload shape after the merge from main (`a8926068f`); no production behavior change.


openhands/storage/data_models/llm_profiles.py
```diff
@@ -0,0 +1,217 @@
...
+    The base ``LLM`` model has ``extra='ignore'``, so typos and renamed keys
+    silently disappear. For API input we want to fail loud, otherwise users
+    can POST ``{"llm": {"custom_header": "x"}}`` and get a 201 with the
+    field quietly dropped.
+    """
+
+    model_config = ConfigDict(extra='forbid')
+
+
+class LLMProfiles(BaseModel):
```
malhotra5, line 62: Thoughts on defining `LLMProfiles` in sdk library?

VascoSch92: Yes... But with the new Setting class is faster to have a small duplicated code here.

But i can double check if I can reause that.

malhotra5:
## Analysis: `llm_profiles` Storage in Enterprise Context

> _This comment was created by an AI agent (OpenHands) during code review._

I've analyzed this PR and found that **`llm_profiles` will be silently lost in the enterprise/SaaS context** due to the current `SaasSettingsStore` implementation. Here's the detailed breakdown:

### What Happens During `store()`

In [`enterprise/storage/saas_settings_store.py` lines 238-250](https://github.com/OpenHands/OpenHands/blob/main/enterprise/storage/saas_settings_store.py#L238-L250):

```python
kwargs = item.model_dump(context={'expose_secrets': True})
kwargs.pop('agent_settings', None)
kwargs.pop('conversation_settings', None)

for key, value in kwargs.items():
    if hasattr(user, key):           # User model has NO llm_profiles column
        setattr(user, key, value)
    if hasattr(org, key) and key not in {...}:  # Org model has NO llm_profiles column
        setattr(org, key, value)
```

Since neither `User` nor `Org` models have an `llm_profiles` attribute:
- `hasattr(user, 'llm_profiles')` → `False`
- `hasattr(org, 'llm_profiles')` → `False`

**Result:** `llm_profiles` is serialized into `kwargs` but never persisted anywhere.

### What Happens During `load()`

In [`enterprise/storage/saas_settings_store.py` lines 106-122](https://github.com/OpenHands/OpenHands/blob/main/enterprise/storage/saas_settings_store.py#L106-L122):

```python
kwargs = {
    **{normalized: getattr(org, c.name) for c in Org.__table__.columns if ...},
    **{normalized: getattr(user, c.name) for c in User.__table__.columns if ...},
}
return Settings(**kwargs)
```

This only loads columns that exist in the database. Since there's no `llm_profiles` column, it's never loaded.

**Result:** `Settings` is created with `llm_profiles` defaulting to an empty `LLMProfiles()`.

### Impact

| Operation | What Happens |
|-----------|-------------|
| `store()` | `llm_profiles` silently ignored (no DB column) |
| `load()` | Defaults to empty `LLMProfiles()` |
| **User Experience** | **All saved profiles disappear after page refresh or any settings update** |

### Suggested Fix

To properly support `llm_profiles` in enterprise, you'll need:

1. **Add a database column** (likely on `User` for per-user profiles, or `Org` for org-wide profiles):
   ```python
   llm_profiles: Mapped[dict | None] = mapped_column(JSON, nullable=True)
   ```

2. **Add an Alembic migration** to create the column

3. **Verify `load()` and `store()`** handle the new column correctly

---

### Note on Local Context

The **local/OSS context works correctly** - `FileSettingsStore` serializes the entire `Settings` object to JSON, so `llm_profiles` is properly persisted.

---

Let me know if you'd like me to elaborate on any part of this analysis or help with the fix!

malhotra5:
^ OH seems to have found something regarding profile persistence in enterprise - is this expected behavior for now?

VascoSch92:

VascoSch92:
Conflicts resolved. The branch has been rebased onto main and includes:

- Updated imports to use new locations for `Settings`, `Secrets`, and related modules that were moved to `openhands/app_server/`
- Merged both main branch tests and profile-related tests in `test_settings.py`
- Updated test imports in `test_profiles_api.py` to use the new `openhands.app_server` paths

All 49 core unit tests pass (`test_settings.py` and `test_llm_profiles.py`).

enterprise/storage/encrypt_utils.py
```diff
@@ -135,3 +138,27 @@ def model_to_kwargs(model_instance):
...
+    fields) where the existing ``_<field>`` String + property pattern is
+    awkward — this keeps the column accessible as a normal ORM attribute
+    while encrypting the entire JSON blob via the same JWE service used
+    by ``encrypt_value``/``decrypt_value``.
+    """
+
+    impl = String
+    cache_ok = True
+
+    def process_bind_param(self, value: Any, dialect):
```
malhotra5: NIT: is it possible to have stronger typing for `value` here?

openhands/app_server/settings/settings_router.py
```diff
@@ -285,3 +296,311 @@ async def invalidate_legacy_secrets_store(
...
+    profile = settings.llm_profiles.get(name) if settings is not None else None
+    if profile is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Profile '{name}' not found",
+        )
+
+    api_key_set = profile.api_key is not None
+    config = profile.model_dump(mode='json')
+    config['api_key'] = None  # never echo a mask; use api_key_set instead
```
malhotra5, line 454: Wiping the api key this way seems ok for now, could be brittle. Out of curiosity if it were a pydantic secret str it would be masked out here? (not blocking)

openhands/app_server/settings/settings_router.py
```diff
@@ -285,3 +296,311 @@ async def invalidate_legacy_secrets_store(
...
+    async with _user_profile_locks[_profile_lock_key(user_id)]:
+        settings = await settings_store.load()
+        if settings is None:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail='Settings not found',
+            )
+
+        llm: LLM
+        if request.llm is not None:
```
malhotra5, line 492: Do we need the conditional for request.llm? Above we set request to `SaveProfileRequest`

openhands/app_server/settings/settings_router.py
```diff
@@ -285,3 +296,311 @@ async def invalidate_legacy_secrets_store(
...
+]
+
+# Per-user asyncio lock serializing the read-modify-write cycle for profile
+# writes. This eliminates the lost-update race between concurrent
+# save/delete/activate calls *within a single worker process*. In
+# multi-worker SaaS deployments, DB-level row locks or optimistic
+# concurrency tokens are still required for full safety — but this closes
+# the single-process hole. The dict grows with distinct user ids; acceptable
+# because each lock is tiny and users are bounded per-process in practice.
+_user_profile_locks: defaultdict[str, asyncio.Lock] = defaultdict(asyncio.Lock)
```
malhotra5, line 326: Nice, how would this work in saas where there are multiple workers? 

Maybe we can create some kind of helper for locking which either uses redis vs asyncio depending on the deployment type?

(I think followup for this would suffice)

malhotra5 (approved):
LGTM!

openhands/app_server/settings/settings_router.py
VascoSch92: It already is,` LLM.api_key `is annotated `str | SecretStr | None`, and Pydantic coerces both inputs to `SecretStr` at runtime. So `profile.model_dump(mode='json')['api_key']` here is `'**********',` not the plaintext key.                                 

The reason we still null it explicitly: the mask is a string, and we don't want to ship `'**********'` to the client, even though the SDK actually defends against the round-trip.

Storage is unaffected, `_profiles_serializer` propagates `info.context`, and the storage path passes `context={'expose_secrets': True}`, so the real key is what gets persisted.            
                                          
  Agreed it's a little brittle                                                                                                                                                                               
  
``` python
  config = profile.model_dump(mode='json', exclude={'api_key'})
  config['api_key'] = None  # keep the field present for the FE                                                                                                                                                                          
```

Happy to switch in a follow-up or here. As you want.

openhands/app_server/settings/settings_router.py
VascoSch92: the two conditionals are doing different things:                                                                                                                                                                                 
 
- if request is None, the caller sent no body at all (e.g. POST /profiles/foo with no JSON). FastAPI gives us None because the param is SaveProfileRequest | None. The fallback to SaveProfileRequest() is so the rest of the handler doesn't crash on request.include_secrets.                                                                                                                                                                                              
- if request.llm is not None, the caller sent a body but omitted the llm field (e.g. POST /profiles/foo {"include_secrets": false}). SaveProfileRequest.llm defaults to None, so request.llm can still be None even after the previous fallback.

We want:                                                                                                                                                                                                                 
- No llm provided (either via no body, or an explicit body without llm) → snapshot the user's current agent_settings.llm.                                                                                                              
- llm provided → save that one (with the api_key-preservation merge).                                                    

So removing the inner check would break the "snapshot current LLM" path whenever the caller sends any body.

openhands/app_server/settings/settings_router.py
malhotra5: Gotcha, lgtm!

openhands/app_server/settings/settings_router.py
VascoSch92: Yeah, fully agree.                                                                                                         
                                                                                                                                                                                                                                         
There's already an in-house pattern for this. enterprise/storage/user_app_settings_store.py:50.

I checked and doing that it is some work. If for you is good, we can do this as follow-up

enterprise/storage/encrypt_utils.py
VascoSch92: Done... and also checked that I have proper typing everywhere

VascoSch92:
@openhands fix the conflicts

openhands/app_server/settings/settings_router.py
malhotra5: Follow up is good!

VascoSch92:
Conflicts resolved. The branch has been merged with main and includes:

- Resolved import conflicts in `settings_models.py` (kept both `LLMProfiles` import from PR and existing `ProviderToken`/`ProviderType` imports from main)
- Resolved import conflicts in `settings_router.py` (kept `LLM` import from PR and `ConversationSettings` from main)

All 90 core unit tests pass (`test_settings.py`, `test_llm_profiles.py`, and `test_profiles_api.py`). Pre-commit checks (ruff, mypy) all pass.

_This comment was created by an AI agent (OpenHands) on behalf of @malhotra5._

PR #2984, opened 2024-07-17
fix bug in config.py file, update reference to variable

JeffKatzy:
Resolves the `NameError: name 'config' is not defined` bug in

```
opendevin/core/config.py:473: in load_from_toml
sandbox_config = config.sandbox
```

PR #14185, opened 2026-04-28
fix(acp): defensive fallback for api_key_env_var missing from pinned SDK

simonrosenberg:
## Summary

`_acp_provider_env` uses `acp_settings.api_key_env_var` which was added in software-agent-sdk PR #2984, but OpenHands is currently pinned to `sdk==1.19` which predates that field — causing `AttributeError: 'ACPAgentSettings' object has no attribute 'api_key_env_var'` at runtime.

**Fix**: use `getattr(acp_settings, 'api_key_env_var', None)` and fall back to an inline `{server → env_var}` map when the attribute is missing. The fallback is identical to what `api_key_env_var` computes in the SDK, so behaviour is the same on both old and new SDK versions.

A `TODO` comment marks the fallback for removal once OpenHands pins to a SDK version that includes PR #2984.

## Test plan

- [x] Existing `TestAcpProviderEnv` tests cover all server types (claude-code → `ANTHROPIC_API_KEY`, codex → `OPENAI_API_KEY`, gemini-cli → `GEMINI_API_KEY`, custom → empty)
- [x] Pre-commit hooks (ruff, mypy) pass

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-08d6c8e   docker.openhands.dev/openhands/openhands:08d6c8e
```

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -1507,18 +1507,32 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
         acp_settings = user.agent_settings
         env: dict[str, str] = {}
 
-        # Map the user's LLM API key to the env var expected by the ACP server.
-        api_key_env = acp_settings.api_key_env_var
-        if api_key_env and api_key_env not in env:
-            llm_api_key = acp_settings.llm.api_key
-            if llm_api_key:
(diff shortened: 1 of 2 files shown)


tofarr (approved):
🍰

PR #14186, opened 2026-04-28
fix(app_server): serialize ProcessSandboxService port allocation to prevent races

dhgoal:
## Summary
- `start_sandbox` could pick the same port for two concurrent calls because `_find_unused_port` only does a transient bind/release; the second sandbox's subprocess then fails to bind.
- The global `_processes` dict was iterated in `search_sandboxes` and `get_sandbox_by_session_api_key` without synchronization, exposing those paths to `RuntimeError: dictionary changed size during iteration` when start/delete ran concurrently.
- Add a module-level `asyncio.Lock` covering port discovery + agent startup + dict mutation, snapshot under the lock in iteration paths, and skip in-flight ports in `_find_unused_port`.

## Related Issues
N/A — found by code review.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing
- [x] Tests added/updated — `test_concurrent_start_sandbox_assigns_distinct_ports` reliably reproduces the duplicate-port outcome on the pre-fix code (5/5 concurrent starts collide on the same port) and passes after the fix.
- [x] Manually tested via the test above.

## Checklist
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)


dhgoal:
The `Python Tests on Linux (3.12)` failures are pre-existing on `main` (run [25062373724](https://github.com/OpenHands/OpenHands/actions/runs/25062373724)) — same `AttributeError: 'ACPAgentSettings' object has no attribute 'api_key_env_var'` at `live_status_app_conversation_service.py:1511`. Looks like an SDK-version mismatch from #14180. The 12 failures are all in `tests/unit/app_server/test_live_status_app_conversation_service.py`, which this PR doesn't touch; the new `test_concurrent_start_sandbox_assigns_distinct_ports` and the rest of `test_process_sandbox_service.py` pass.

PR #14193, opened 2026-04-28
fix(enterprise): accept pydantic models in EncryptedJSON.process_bind_param

VascoSch92:

---

## Why

`server/routes/auth.py` `keycloak_callback` -> `UserStore.create_user` crashes on every new login with:

```
TypeError: Object of type LLMProfiles is not JSON serializable
  File "storage/encrypt_utils.py", line 162, in process_bind_param
    return encrypt_value(json.dumps(value))
```

`UserStore.get_kwargs_from_settings` does `getattr(settings, 'llm_profiles')` and hands the resulting `LLMProfiles` pydantic model straight to `User(**kwargs)`. The `EncryptedJSON` `TypeDecorator` only handled `dict`, so `json.dumps(LLMProfiles(...))` blew up. `Settings.llm_profiles` has `default_factory=LLMProfiles`, so this fires for every new user — not only ones who already saved a profile.

Introduced in #14146 ("feat(settings): add saved LLM profiles (BE)").

## Summary

- `EncryptedJSON.process_bind_param` now accepts `BaseModel | dict | None` and dumps pydantic instances via `model_dump(mode='json', context={'expose_secrets': True})` before `json.dumps`. Other encrypted-JSON columns added later get the same robustness for free.
- `expose_secrets=True` matches `saas_settings_store.store()` and is required because `LLMProfiles._profiles_serializer` forwards the context to nested `LLM` dumps; without it, `SecretStr` api_keys serialize as `'**********'` and the masked literal would be encrypted into the column.
- Adds a regression test in `test_user_store.py` that builds `Settings` with a populated `LLMProfiles` containing a `SecretStr` api_key, persists via `User(**kwargs)`, and asserts both no `TypeError` and a clean round-trip of the plaintext secret. Without this patch the test fails with the exact production `TypeError`.

## Alternative considered

Special-casing `llm_profiles` inside `UserStore.get_kwargs_from_settings` (as in #14190). Same crash goes away, but: the same shape mismatch repeats in `OrgStore`/`OrgMemberStore` `get_kwargs_from_settings` and the loose `setattr(user, key, value)` loop in `saas_settings_store.store()`, and any future encrypted-JSON column reproduces the bug. Fixing it at the column is one place, narrow surface, symmetric with `process_result_value` already returning `dict`.

## Issue Number

Supersedes #14190.

## How to Test

```
cd enterprise
poetry run pytest \
  tests/unit/test_user_store.py::test_create_user_with_llm_profiles_does_not_crash_and_preserves_secrets \
  tests/unit/test_user_store.py::test_get_kwargs_from_settings \
  tests/unit/test_saas_settings_store.py -k llm_profiles
```

Manual: log in via keycloak as a brand-new user — `create_user` no longer crashes.

## Type

- [x] Bug fix

## Notes

- No schema changes; the column type is unchanged at rest.
- `process_result_value` is unchanged: callers still read `dict`. If a follow-up wants symmetric pydantic-typed reads, that can be done with a generic `EncryptedPydantic[T]` decorator without touching this fix.

### DIFF
--- a/enterprise/storage/encrypt_utils.py
+++ b/enterprise/storage/encrypt_utils.py
@@ -5,7 +5,7 @@ from base64 import b64decode, b64encode
 from typing import Any
 
 from cryptography.fernet import Fernet, InvalidToken
-from pydantic import SecretStr
+from pydantic import BaseModel, SecretStr
 from server.config import get_config
 from sqlalchemy import String, TypeDecorator
 from sqlalchemy.engine.interfaces import Dialect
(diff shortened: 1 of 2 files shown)


malhotra5 (approved):
Nice, this is a clean fix

chuckbutkus (approved):

PR #14209, opened 2026-04-29
fix: surface unverified bare LLM names in the provider dropdowns

jlav:

---

## Why

When a user opened the Anthropic provider dropdown in the LLM picker, only the SDK's curated *verified* models appeared (e.g. `claude-opus-4-6`). 

Other Anthropic models were silently dropped, even though they exist in `litellm.model_cost`. The same bug affected Vertex AI / Gemini, Cohere, and others: every provider whose models LiteLLM stores as bare names (no `provider/` prefix) lost its non-verified tail in the UI.

Root cause is `_assign_provider()` in `openhands/utils/llm.py`. 

It only knew how to prefix bare names already present in the SDK's three hardcoded sets (`_BARE_OPENAI_MODELS`, `_BARE_ANTHROPIC_MODELS`, `_BARE_MISTRAL_MODELS`). Anything else stayed bare, then `_get_all_models_with_verified()` assigned `provider=None`, then `?provider__eq=anthropic` (and friends) filtered it out.

## Summary

- Adds a `litellm.get_llm_provider()` fallback to `_assign_provider()` so any bare name LiteLLM can route reaches the correct provider's dropdown.

## How to Test

Today, no unverified models appear for Anthropic:
image

After this change, other models appear beneath verified models:
image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The fallback is in-process — `litellm.get_llm_provider` is a string-routing function over LiteLLM's static `provider_list` and `model_list_set`, ~3.4µs per call, no network. Exceptions are swallowed so unresolvable names pass through unchanged.
- Does *not* change which models are tagged "Verified" in the UI — that remains the SDK's `VERIFIED_MODELS` dict via `_get_verified_models()`. Newly surfaced names appear under "Others".

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-521b35f   docker.openhands.dev/openhands/openhands:521b35f
```

### DIFF
--- a/openhands/utils/llm.py
+++ b/openhands/utils/llm.py
@@ -230,8 +230,10 @@ def _assign_provider(model: str) -> str:
     """Prefix a bare model name with its canonical provider.
 
     Models that already contain a ``/`` provider separator are returned
-    unchanged. Only well-known bare names (OpenAI, Anthropic, Mistral)
-    are prefixed.
+    unchanged. Bare names are first checked against the SDK's verified
+    sets (cheap, no network), then fall back to LiteLLM's own routing
+    tables so that unverified names like ``claude-opus-4-7`` or
+    ``gemini-2.0-flash`` still reach the provider-keyed dropdown.
     """
     if '/' in model:
         return model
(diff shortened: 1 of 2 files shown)


neubig (approved):
Thanks!

PR #14211, opened 2026-04-29
refactor(settings): rename LLMAgentSettings → OpenHandsAgentSettings (SDK 1.19.0)

simonrosenberg:
## Summary

Adds a small OpenHands-side compatibility layer for the SDK rename from `LLMAgentSettings` to `OpenHandsAgentSettings`, without pulling in unrelated ACP UI or conversation-routing work.

OpenHands is still pinned to `openhands-sdk==1.19.0`, so this PR does not start writing `agent_kind: "openhands"`. The standard built-in agent continues to use the SDK-compatible runtime/persisted discriminator `agent_kind: "llm"`; `OpenHandsAgentSettings` is only the local/canonical class name exposed by the shim.

## Changes

- `sdk_settings_compat.py` now exports `OpenHandsAgentSettings`.
- When the installed SDK already provides `OpenHandsAgentSettings`, the shim imports it directly.
- With the current SDK pin, the shim falls back to `LLMAgentSettings as OpenHandsAgentSettings` and keeps `LLMAgentSettings = OpenHandsAgentSettings` as a silent local alias.
- Settings model/router call sites use `OpenHandsAgentSettings` naming through the shim.
- Shim tests cover `OpenHandsAgentSettings`, the legacy alias, and existing discriminated-union behavior.

## Compatibility strategy

- Existing production settings/conversations using `"llm"` continue unchanged.
- No deprecation warnings are emitted inside OpenHands for the local alias or when loading persisted settings.
- Deprecation warnings remain the SDK's responsibility for public SDK deprecated API access paths.
- ACP agents have not been used in production, so this PR deliberately avoids adding broad ACP data migration logic.

## Test plan

- [x] `uv run pytest tests/unit/utils/test_sdk_settings_compat.py tests/unit/app_server/test_settings_api.py -q`
- [x] Verified PR diff is one commit touching only four files.
- [ ] `uv run ruff check ...` reports pre-existing docstring style violations in touched files; no new lint-specific code path was added.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-46d84ba   docker.openhands.dev/openhands/openhands:46d84ba
```

### DIFF
--- a/openhands/app_server/settings/settings_models.py
+++ b/openhands/app_server/settings/settings_models.py
@@ -35,7 +35,7 @@ from openhands.utils.jsonpatch_compat import deep_merge
 from openhands.utils.sdk_settings_compat import (
     ACPAgentSettings,
     AgentSettingsConfig,
-    LLMAgentSettings,
+    OpenHandsAgentSettings,
     default_agent_settings,
     validate_agent_settings,
 )
@@ -287,7 +287,7 @@ class Settings(BaseModel):
     @field_serializer('agent_settings')
     def agent_settings_serializer(
         self,
-        agent_settings: LLMAgentSettings | ACPAgentSettings,
+        agent_settings: OpenHandsAgentSettings | ACPAgentSettings,
         info: SerializationInfo,
     ) -> dict[str, Any]:
         context = info.context or {}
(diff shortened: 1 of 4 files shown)


simonrosenberg:
@openhands fix CI errors

PR #3227, opened 2024-08-02
Update integration tests README.md

xingyaoww:
**What is the problem that this fixes or functionality that this introduces? Does it fix any open issues?**

---
**Give a summary of what the PR does, explaining any non-trivial design decisions**

Update integration tests README.md to reflect recent changes in EventStreamRuntime integration tests.

---
**Other references**

PR #14227, opened 2026-04-30
feat(acp): minimal generic ACP agent UI on unified SDK conversation endpoint

### LINKED ISSUE #14222: feat(acp): minimal generic ACP agent UI (no per-provider scaffolding)
Replaces PR OpenHands/OpenHands#14189 with a minimal generic ACP UX. No per-provider (Claude Code / Codex / Gemini) scaffolding — the user supplies the subprocess command and env vars themselves. Feature-flagged on `enable_acp`.

## Settings page (`/settings/agent`)

Three controls, single tab:

1. **Agent type** — dropdown: OpenHands / ACP.
2. **Command** — textarea, whitespace-tokenized, required when ACP. Placeholder: `npx -y @agentclientprotocol/claude-agent-acp`.
3. **Environment variables** — `.env`-style textarea (one `KEY=value` per line, blanks and `#` ignored). Holds the API key and any base-URL override.
4. **Model** (optional) — text input → `acp_model`.

Save sends `{ agent_kind: \"acp\", acp_server: \"custom\", acp_command, acp_env, acp_model }`. Credentials stay in env (never on the command line).

## Sidebar gating when ACP is active

When `agent_settings.agent_kind === \"acp\"`, grey out (visible + disabled + tooltip):

| Item | Reason |
| -- | -- |
| LLM (`/settings`) | ACP subprocess manages its own LLM. |
| Condenser (`/settings/condenser`) | ACP server handles context. |
| MCP (`/settings/mcp`) | ACP server manages MCP. |

## Per-conversation ACP config snapshot

User's global ACP settings are defaults for **new** conversations only. Each conversation owns its own config — no mismatch is possible, so no mismatch banner.

* New encrypted `acp_settings: JSONB` column on `conversation_metadata` (alembic 010, idempotent).
* On start: snapshot the ACP fields into the column.
* On resume / webhook-import: hydrate `ACPAgent` from the snapshot, not from current user settings.
* Header shows `ACP: <last command token>` (e.g. `claude-agent-acp`), reusing the model-display element from OpenHands/OpenHands#13616 — no new component. Tooltip shows full command + args. Env never displayed.

## Feature flag and entry point

* `enable_acp: bool = False` in `WebClientFeatureFlags` and matching `ENABLE_ACP()` in `feature-flags.ts`.
* "Agent" item in the Settings sidebar, gated on `enable_acp`.

## Out of scope

* Per-provider convenience UX (named Claude Code / Codex / Gemini options).
* Auto-injecting SaaS proxy URL — users paste it. Add a pre-fill later if friction shows up.
* Editing a conversation's snapshotted config — frozen at start; rotate via a new conversation.

## Related

* Replaces PR OpenHands/OpenHands#14189; closes OpenHands/OpenHands#14183.
* Builds on backend in OpenHands/OpenHands#14004 and [#14171](<https://github.com/OpenHands/OpenHands/issues/14171>).

simonrosenberg:

---

## Why

ACP (Agent Client Protocol) agents run as external subprocesses rather than using OpenHands' built-in LLM stack. There was no UI to configure them — users had to edit `settings.json` by hand. This PR adds the minimal settings page needed to point OpenHands at an ACP agent, gates the rest of the settings UI so LLM/condenser/MCP options don't conflict while ACP is active, and wires up the routing + per-conversation provenance + UI plumbing.

It also pins the SDK to a SHA that includes [software-agent-sdk#3227](https://github.com/OpenHands/software-agent-sdk/pull/3227) (unified ACP conversation endpoint), so OpenHands talks to a single `/api/conversations` route for both LLM and ACP agents instead of maintaining a parallel `/api/acp/conversations` dispatch.

Closes #14222 (the UI), #14322 (cross-kind config preservation).

### Backend

- **SDK pin**: `openhands-sdk` / `openhands-agent-server` / `openhands-tools` pinned to git SHA `fa29ae28c61e13479749a3d303e021032e34766f` (post v1.22.0, includes PR #3227). The unified `/api/conversations` endpoint accepts both `Agent` and `ACPAgent` payloads through the `AgentBase` discriminated union, so OpenHands no longer dispatches by agent kind.
- **Feature flag + registry**: `enable_acp` added to `WebClientFeatureFlags` (gated by `ENABLE_ACP`, default `false`). The SDK's `ACP_PROVIDERS` registry is mirrored as `acp_providers` on `/api/v1/web-client/config` so future frontend changes can read the canonical preset list from one source.
- **No more route dispatch**: removed the `agent_kind_to_router_path()` helper, the dual fetch in `_get_live_conversation_info`, and the per-call ACP-vs-LLM branches in `live_status_app_conversation_service.py`, `remote_sandbox_service.py`, and `_start_conversation_remotely`. Everything hits `/api/conversations`.
- **Webhook router**: signature is `ConversationInfo` (the SDK's canonical model — `ACPConversationInfo` is now a deprecated alias). The `AgentBase` discriminated union accepts both `Agent` and `ACPAgent` payloads on the same route. `on_conversation_update()` sets `agent_kind` from the agent payload (`llm_model` for OpenHands, **null for ACP** — the ACP dummy LLM sentinel `acp-managed` is squashed to null so analytics dashboards don't get contaminated).
- **Per-conversation provider display**: when an ACP conversation is created, a label like `"ACP: claude-agent-acp"` is stamped onto `AppConversationInfo.display_name` via `acp_display_name(acp_command)` and surfaced on the conversation chip / recent-conversation list. `display_name` is the new column added in migration `010`; `agent_kind` was added in `009`.
- **Cross-kind config preservation (`saved_agent_configs`)**: switching `agent_kind` snapshots the outgoing config under its kind key in `Settings.saved_agent_configs` (a new dict field), then restores from any prior snapshot (or clean defaults) for the incoming kind. Fixes both the 500 on ACP → OpenHands switch and the regression where switching back would lose the user's LLM / condenser / MCP / agent config. `acp_env` and `mcp_config` still replace wholesale; `saved_agent_configs` is in `_SETTINGS_UPDATE_IGNORED_FIELDS` so external payloads can't overwrite it directly.
- **SDK compat shims removed**: dropped the `_sdk_supports_acp_secrets` runtime probe (the pinned SDK guarantees `agent_context.secrets` carries `acp_compatible=True`) and the per-server `api_key_env_var` fallback map (the SDK property is authoritative).
- **Lockfile bumps**: `dev_config/python/.pre-commit-config.yaml` (mypy SDK pin), root `poetry.lock`, `uv.lock`, and `enterprise/poetry.lock` regenerated to track the SDK SHA pin.

### Frontend

- **`/settings/agent` page** (`routes/agent-settings.tsx`): agent-type dropdown (OpenHands / ACP), preset selector — Claude Code / Codex / Gemini CLI / Custom — command textarea, optional model override. API keys go through the existing Secrets panel.
- **ACP guard hook** (`use-acp-guard.ts`): redirects LLM / Condenser / MCP routes to `/settings/agent` when ACP is the active agent.
- **Nav gating**: LLM, Condenser, and MCP nav items are disabled (greyed out with tooltip) while ACP is active. The disabled-state prop carries the **agent name** (`disabledAgentName`), which the `SETTINGS$AGENT_DISABLED_TOOLTIP` i18n key interpolates as `{agentName}`. Items marked `acpGated` (i.e. the Agent settings entry itself) are hidden entirely when the feature flag is off.
- **Conversation chip / recent list**: shows the backend-stamped `display_name` (e.g. `"ACP: claude-agent-acp"`) for ACP conversations, falling back to `llm_model` for OpenHands. The `acp_providers` registry is already exposed so a follow-up can swap the prefix label for brand names ("Claude Code", "Codex", "Gemini CLI") without redeploying.
- **Event rendering**: ACP tool-call cards are suppressed while `status === "in_progress"` (prevents the empty-argument flash; arguments stream in afterwards). The card title is rendered verbatim from `event.title` — the upstream ACP sub-agent already emits a complete humanised label ("Read tests", "Edit foo.py"), so wrapping it in a translated verb prefix would double the verb. The 5 `ACTION_MESSAGE$ACP_*` keys × 15 languages are removed (75 entries) along with the dead `getACPToolCallTitleKey()` helper.
- **WebSocket routing**: `extractPathPrefix()` matches the single `/api/conversations` shape since the SDK's `RemoteConversation` uses one route for both agent kinds now.
- **i18n**: 8 new `SETTINGS$AGENT_*` keys × 15 languages. The OH `ACTION_MESSAGE$*` / `OBSERVATION_MESSAGE$*` keys that wrap interpolated values in `<cmd>` get an NBSP placed before the tag, mirroring [#14380](https://github.com/OpenHands/OpenHands/pull/14380): a regular space before a component tag gets dropped by react-i18next's `<Trans>` tokenizer (producing titles like `"Runningcd /workspace && ls -la"`), but a non-breaking space survives.

### Tests

- **`tests/unit/app_server/test_webhook_router_acp.py`** covers Agent vs ACPAgent discrimination on real Pydantic models (not MagicMock specs that would mask a discriminator regression), `llm_model` / `agent_kind` storage paths, backward compatibility against legacy LLM payloads, and that `track_conversation_created` receives `llm_model=None` for ACP (regression guard against leaking the ACP `LLM` sentinel into analytics).
- **`TestAgentKindConversationUrl`** in the live-status test file: regression guard for the unified `/api/conversations/{id}` URL — the old dual-path test (`/api/acp/conversations/` for ACP) was removed along with the routing helpers.
- **`frontend/__tests__/routes/agent-settings.test.tsx`** covers preset selection (incl. `acp_server` saved on submit) and switching back to OpenHands clearing ACP fields.
- Component tests verify the new chip resolution and the disabled-nav-item tooltip behaviour.

## How to Test

1. Enable ACP: set `ENABLE_ACP=true` on the backend.
2. Go to **Settings → Agent**, switch to **ACP**, pick the **Claude Code** preset (command auto-fills).
3. Add `ANTHROPIC_API_KEY` via **Settings → Secrets**.
4. Save. Start a new conversation — it should connect to the ACP subprocess and respond.
5. The conversation chip should read **"ACP: claude-agent-acp"** (the backend-stamped `display_name`). The recent-conversation list should show the same label.
6. Verify **Settings → LLM**, **Condenser**, and **MCP** are greyed out in the nav while ACP is active, and redirect to `/settings/agent` if accessed directly.
7. Switch back to **OpenHands** — LLM/Condenser/MCP nav items become active and your previous LLM/condenser/MCP config is restored from `saved_agent_configs`.
8. Repeat the switch a few times to confirm no 500 errors.

Unit tests: `cd frontend && npx vitest run` and `uv run pytest tests/unit/app_server/ -q`.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Feature is off by default. No existing behaviour changes until `ENABLE_ACP=true`.
- `acp_env` is supported at the API level but has no UI — credentials belong in the Secrets panel.
- The SDK pin is a **git SHA** because PR #3227 (unified ACP endpoint) merged after the v1.22.0 release and has no published wheel yet. Move to a tagged version once one is cut that includes the merge.
- The Settings → Agent preset list is currently hard-coded in `agent-settings.tsx`. The backend now exposes the same registry as `acp_providers` on `/api/v1/web-client/config`; a follow-up can wire the frontend to consume it (eliminating drift between frontend hard-coding and the SDK registry).
- **Known limitation — custom-command resume after sandbox loss**: when a sandbox is destroyed (not just paused) and a user resumes an ACP conversation that used `acp_server = custom`, the `acp_command` / `acp_args` / `acp_model` come from the user's *current* settings, which may have changed. Built-in presets resume correctly because the SDK registry has `default_command`. A follow-up PR will add a structured `acp_spec` field on `AppConversationInfo` so custom-command conversations can be reconstructed verbatim.

---

To run this PR locally (the app image isn't published to GHCR yet — build it from the branch):

```bash
docker build -f containers/app/Dockerfile -t openhands-acp:local .
docker run -it --rm \
  -p 3000:3000 \
  -e AGENT_SERVER_IMAGE_REPOSITORY=ghcr.io/openhands/agent-server \
  -e AGENT_SERVER_IMAGE_TAG=fa29ae2-python \
  -e ENABLE_ACP=true \
  --name openhands-acp \
  openhands-acp:local
```

The agent-server image (`ghcr.io/openhands/agent-server:fa29ae2-python`) is already on GHCR and matches the SDK SHA the app is pinned to.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-350e404   docker.openhands.dev/openhands/openhands:350e404
```


hieptl (changes requested):
Hello @simonrosenberg,

Thank you for creating this pull request.

Could you please update the description to follow our pull request template? 

Additionally, it would be helpful if you could include a short demo video to showcase the UI changes.

Thank you very much! 🙏

frontend/src/components/features/context-menu/context-menu-nav-link.tsx
```diff
@@ -3,27 +3,51 @@ import { Link } from "react-router";
...
+      </span>
+    );
+  }
+
   return (
     <Link
       to={to}
       onClick={onClick}
-      className="flex items-center gap-2 p-2 cursor-pointer hover:bg-white/10 hover:text-white rounded w-full text-xs"
+      className={cn(
```
hieptl: **suggestion (blocking):** We typically use `cn` when combining styles based on specific conditions. In this case, it isn’t necessary to use it.

Thank you! 🙏

frontend/src/components/features/conversation/conversation-name.tsx
```diff
@@ -159,16 +159,21 @@ export function ConversationName() {
...
-            title={conversation.llm_model}
-            data-testid="conversation-name-llm-model"
-          >
-            <CircuitIcon width={12} height={12} className="shrink-0" />
-            <span className="truncate">{conversation.llm_model}</span>
-          </span>
-        )}
+        {titleMode !== "edit" &&
+          (conversation.display_name ?? conversation.llm_model) && (
+            <span
```
hieptl: **suggestion (blocking):** We should use `Typography.Text` from the `/frontend/src/ui` folder instead of the `span` tag, and `Typography.Paragraph` instead of the `p` tag.

I’ve noticed a few places that may need updating. I’ve left one comment as an example—could you please review the rest of the pull request for similar cases?

Thank you! 🙏

frontend/src/i18n/translation.json
```diff
@@ -24138,5 +24138,124 @@
     "de": "Keine Git-Organisationen gefunden.",
     "uk": "Організації Git не знайдено.",
     "ca": "No s'han trobat organitzacions Git."
+  },
+  "SETTINGS$AGENT_PAGE_DESCRIPTION": {
+    "en": "Choose which agent handles your conversations. ACP runs an external agent subprocess that you configure here.",
+    "ja": "Choose which agent handles your conversations. ACP runs an external agent subprocess that you configure here.",
```
hieptl, line 24637: **suggestion (blocking):"" We need to ensure support for multiple languages. I’ve left one comment as an example—could you please review the rest of the text in this pull request for similar updates?

Thank you! 🙏

frontend/src/routes/mcp-settings.tsx
```diff
@@ -45,10 +46,18 @@ interface MCPServerConfig {
 
 function MCPSettingsScreen() {
   const { t } = useTranslation();
+  const navigate = useNavigate();
   const { data: settings, isLoading } = useSettings();
   const { data: config } = useConfig();
   const { mutate: saveSettings, isPending: isSavingSearchApiKey } =
     useSaveSettings();
+
+  const isAcpActive = settings?.agent_settings?.agent_kind === "acp";
```
hieptl: **suggestion (blocking):**

It appears that this logic may be duplicated across multiple components. Could you please review and consider consolidating it where appropriate?

```tsx
const isAcpActive = settings?.agent_settings?.agent_kind === "acp";
  useEffect(() => {
    if (isAcpActive) {
      navigate("/settings/agent", { replace: true });
    }
  }, [isAcpActive, navigate]);
```

Thank you! 🙏

openhands/app_server/app_lifespan/alembic/versions/010.py
hieptl, line 1: **question (blocking):** Would we need to include a migration file for the enterprise (SaaS) environment as well?

Thank you! 🙏

openhands/app_server/web_client/default_web_client_config_injector.py
```diff
@@ -124,6 +124,7 @@ def _get_feature_flags() -> WebClientFeatureFlags:
         hide_users_page=os.getenv('HIDE_USERS_PAGE', 'false') == 'true',
         hide_billing_page=os.getenv('HIDE_BILLING_PAGE', 'false') == 'true',
         hide_integrations_page=os.getenv('HIDE_INTEGRATIONS_PAGE', 'false') == 'true',
+        enable_acp=os.getenv('ENABLE_ACP', 'false') == 'true',
```
hieptl, line 128: **question (blocking):** We already have a front-end feature flag included in this pull request. Could you help clarify the need for an additional back-end feature flag as well?

Thank you! 🙏

frontend/src/components/features/context-menu/context-menu-nav-link.tsx
simonrosenberg: Done — removed the `cn()` wrapper (single static string needs no `cn`). The `Link` now has a plain string `className`.

frontend/src/components/features/conversation/conversation-name.tsx
simonrosenberg: Done — the inner text label now uses `Typography.Text` in both `conversation-name.tsx` and `context-menu-nav-link.tsx`, and the hint labels in `agent-settings.tsx` were replaced with `Typography.Text` as well. The outer container element keeps its `title` tooltip attribute (which `Typography.Text` does not accept) and uses `Typography.Text` for the text content inside.

frontend/src/i18n/translation.json
simonrosenberg: Done — all 7 new i18n keys added in this PR already have translations for all 15 supported languages (en, ja, zh-CN, zh-TW, ko-KR, no, ar, de, fr, it, pt, es, ca, tr, uk). You can verify with `grep -c '"en"' frontend/src/i18n/translation.json`.

frontend/src/routes/mcp-settings.tsx
simonrosenberg: Done — extracted into a `useAcpGuard(scope)` hook in `frontend/src/hooks/use-acp-guard.ts`. All three routes (llm-settings, condenser-settings, mcp-settings) now call `useAcpGuard()` instead of repeating the `isAcpActive + useEffect` block. The hook accepts an optional `scope` parameter so llm-settings can pass its org/personal scope correctly.

openhands/app_server/app_lifespan/alembic/versions/010.py
simonrosenberg: Yes — added `enterprise/migrations/versions/113_add_display_name_to_conversation_metadata.py` which mirrors migration 010, adding the same nullable `display_name` column to `conversation_metadata`. It chains from migration 112.

simonrosenberg:
Hi @hieptl, all your blocking comments have been addressed:

- **`cn()` removed** — plain string className on the `Link`
- **Typography.Text** — inner text labels in `conversation-name.tsx`, `context-menu-nav-link.tsx`, and `agent-settings.tsx` now use `Typography.Text`/`Typography.Paragraph`
- **i18n** — all 7 new keys already have 15-language coverage
- **Duplicate guard logic** — extracted into `useAcpGuard(scope)` hook used by llm-settings, condenser-settings, and mcp-settings
- **Enterprise migration** — `enterprise/migrations/versions/113_add_display_name_to_conversation_metadata.py` added
- **Dual feature flag** — backend flag is the production gate; frontend localStorage flag is a dev override consistent with `BILLING_SETTINGS`, `ENABLE_AUTOMATIONS`, etc.

A demo video showing the ACP settings page and a conversation running end-to-end has been validated locally (the feature is behind the `enable_acp` flag and won't appear for users until it is enabled).

openhands/app_server/web_client/default_web_client_config_injector.py
simonrosenberg: I simplified and only kept the back-end feature flag!

xingyaoww (commented):
One easy thing we can try to do would be mount "~/.claude" or "~/.codex" into the docker container and see if the ACP would automatically pick it up.

Also, i wonder for the drop down commands, can we come up with a pre-populated list of commands, eg:
codex: XXX
claude-code: XXX
and then we give user the ability to put in custom ones?

enterprise/tests/unit/storage/test_saas_sql_app_conversation_info_service.py
```diff
@@ -264,6 +264,7 @@ async def test_to_info_with_user_id_functionality(
         stored_metadata.trigger = None
         stored_metadata.pr_number = []
         stored_metadata.llm_model = None
+        stored_metadata.display_name = None
```
neubig, line 267: ditto

enterprise/tests/unit/test_saas_settings_store.py
```diff
@@ -196,7 +196,14 @@ async def patched_store(item):
...
                     item_dict['keycloak_user_id'] = store.user_id
-                    settings = UserSettings(**item_dict)
+                    # Filter to fields that map to UserSettings columns.
+                    # ``Settings.model_dump`` produces keys like ``saved_agent_configs``
+                    # and ``llm_profiles`` that have no UserSettings column; passing
+                    # them to the constructor raises TypeError. The "update" branch
+                    # above already filters this way.
+                    valid_keys = set(UserSettings.__table__.columns.keys())
+                    filtered = {k: v for k, v in item_dict.items() if k in valid_keys}
+                    settings = UserSettings(**filtered)
```
neubig: It's not clear why this needs to be added in this PR. If it needs to be added in this PR then maybe we're doing something in an over-complciated way.

enterprise/migrations/versions/113_add_display_name_to_conversation_metadata.py
```diff
@@ -0,0 +1,32 @@
+"""Add display_name column to conversation_metadata table.
```
neubig, line 1: I'm not very convinced that we need a change to display the agent type so prominently. Could we revert this functionality change and add it later if we think that it's necessary? Or at least separate it into a separate PR so we can discuss it separately.

frontend/__tests__/components/features/conversation/conversation-name.test.tsx
```diff
@@ -296,7 +296,7 @@ describe("ConversationName", () => {
     ).not.toBeInTheDocument();
   });
 
-  it("should render the llm model when available", () => {
+  it("should render the OpenHands brand when an llm_model is set", () => {
```
neubig: ditto

frontend/__tests__/components/features/home/recent-conversation.test.tsx
```diff
@@ -54,25 +54,37 @@ const renderRecentConversation = (conversation: V1AppConversation) =>
   );
 
 describe("RecentConversation - llm_model", () => {
-  it("should render the llm model when provided", () => {
+  it("should render the OpenHands brand label when llm_model is provided", () => {
```
neubig: ditto

frontend/__tests__/hooks/mutation/use-new-conversation-command.test.tsx
```diff
@@ -270,6 +270,66 @@ describe("useNewConversationCommand", () => {
     });
   });
 
+  it("does not pass the ACP display label as an LLM model", async () => {
```
neubig, line 273: ditto

neubig (changes requested):
Before getting this in to prod, I'd like to consider whether we can simplify the backend, which would probably also allow us to simplify the frontend: https://github.com/OpenHands/software-agent-sdk/pull/3227

simonrosenberg:
Closing in favor of https://github.com/OpenHands/OpenHands/pull/14401

simonrosenberg:
Closing in favor of https://github.com/OpenHands/OpenHands/pull/14401

PR #14097, opened 2026-04-23
APP-1325: show GitLab/Slack sections without GitHub App configured

aivong-openhands:
## Summary

Fixes #14096

The integrations page was blank in self-hosted/Replicated deployments when `GITHUB_APP_SLUG` was not set, even when GitLab was properly configured.

## Root Cause

The GitLab and Slack sections were gated behind `shouldRenderExternalConfigureButtons` which required `github_app_slug` to be set:

```typescript
const shouldRenderExternalConfigureButtons = isSaas && config?.github_app_slug;
```

This meant that in environments where only GitLab OAuth is configured (no GitHub App), the entire integrations page appeared blank.

## Solution

Split the render condition into separate variables:
- `shouldRenderGitHubConfigureButton` - requires `isSaas && github_app_slug`
- `shouldRenderGitLabSection` - requires only `isSaas`
- `shouldRenderSlackSection` - requires only `isSaas`

This allows each integration section to render independently based on its own requirements.

## Testing

Tested on a Replicated self-hosted deployment with GitLab OAuth configured but no GitHub App:
- Before: Integrations page was completely blank
- After: GitLab section shows correctly with connection status and webhook manager

---
_This PR was created by an AI assistant (OpenHands) on behalf of the user._

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-1ecffcb   docker.openhands.dev/openhands/openhands:1ecffcb
```

PR #14255, opened 2026-05-01
APP-1325: accept '1' as truthy value for SLACK_WEBHOOKS_ENABLED to surface slack integration settings

aivong-openhands:

---

## Why

Older Helm charts set `SLACK_WEBHOOKS_ENABLED=1` rather than `=true`. The check introduced in #14097 only accepted `'true'`, so Slack integration was silently disabled in those deployments.

## Summary

- Accept both `'true'` and `'1'` as truthy values for `SLACK_WEBHOOKS_ENABLED` in `_get_slack_enabled()`
- Add a test covering the `'1'` case

## Issue Number

Related: #14097

## How to Test

Run the unit tests:
```
poetry run pytest tests/unit/app_server/test_default_web_client_config_injector.py::TestGetSlackEnabled -v
```

All 4 tests should pass, including the new `test_returns_true_when_webhooks_enabled_is_set_to_1`.

## Video/Screenshots

N/A — backend-only logic change with full unit test coverage.

## Type

- [x] Bug fix

## Notes

Only `default_web_client_config_injector.py` is changed; no frontend or model changes required.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-964b36f   docker.openhands.dev/openhands/openhands:964b36f
```

### DIFF
--- a/openhands/app_server/web_client/default_web_client_config_injector.py
+++ b/openhands/app_server/web_client/default_web_client_config_injector.py
@@ -100,7 +100,7 @@ def _get_github_app_slug() -> str | None:
 def _get_slack_enabled() -> bool:
     """Return whether Slack integration is fully configured for the web client."""
     return (
-        os.getenv('SLACK_WEBHOOKS_ENABLED', 'false').lower() == 'true'
+        os.getenv('SLACK_WEBHOOKS_ENABLED', 'false').lower() in ('true', '1')
         and bool(os.getenv('SLACK_CLIENT_ID', '').strip())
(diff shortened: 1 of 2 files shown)


aivong-openhands:
Tested on a self hosted instance to reproduce not seeing the slack integration settings with the environment variable SLACK_WEBHOOKS_ENABLED = "1". Then applied this fix as a config map patch and was able to see the slack integration settings.

hieptl (approved):
Thank you! 🙏

PR #14258, opened 2026-05-01
docs: document truthy value convention for env var enable toggles

aivong-openhands:
## Why

Older Helm chart versions set enable toggles to `'1'` rather than `'true'`. The check introduced in #14097 only accepted `'true'`, which silently disabled Slack integration on those deployments — fixed in #14255.

To prevent the same class of bug from recurring on future toggles, this PR codifies the required pattern in `AGENTS.md`.

## Summary

- Add a new "Environment Variable Enable Toggles" section to `AGENTS.md` under Implementation Details
- Documents that all boolean env var feature gates must accept both `'true'` and `'1'`
- Provides the required pattern and anti-patterns with inline comments
- Notes that a unit test for the `'1'` case is required alongside the `'true'` case

## Issue Number

Related: #14097, #14255

## How to Test

N/A — documentation only.

## Type

- [x] Documentation

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -453,6 +453,24 @@ To add a new LLM model to OpenHands, you need to update multiple files across bo
 - The `organize_models_and_providers` function groups models by provider
 - Default model selection prioritizes verified models for each provider
 
+### Environment Variable Enable Toggles
+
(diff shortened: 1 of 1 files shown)


hieptl (approved):
Thank you! 🙏

PR #14267, opened 2026-05-02
Fix local process sandbox readiness for MCP conversations

neubig:
## Summary

Fixes local OpenHands conversations getting stuck/unusable after configuring MCP servers through the UI.

The local process sandbox was treating healthy idle agent-server processes as not ready because `psutil` often reports an idle uvicorn process as `STATUS_SLEEPING`, not `STATUS_RUNNING`. The fix now keeps process liveness separate from sandbox readiness: a live process remains `STARTING` until the agent-server health check succeeds, and only then is reported as `RUNNING` with the agent-server URL/session key. Process sandboxes also had an empty default `working_dir`, which later caused tool initialization to fail.

## Changes

- Keep alive process sandboxes as `STARTING` until the agent-server health check succeeds
- Report process sandboxes as `RUNNING` only after the health check exposes the agent-server URL/session key
- Keep stopped processes as `PAUSED`
- Treat zombie/dead processes as missing and stale unready processes as `ERROR`
- Set process sandbox default `working_dir` to `/tmp/openhands-sandboxes`
- Add regression coverage for sleeping-but-healthy process sandboxes

### Before

The MCP server is saved in settings, but local conversation startup remains unusable/stuck:

Before: local MCP conversation stuck

### After

The local process sandbox reaches `READY`, and the configured `everything` MCP server initializes tools for the agent:

After: MCP tools initialize

Agent-server log evidence from local validation:

```text
Created 13 MCP tools: ['echo', 'get-annotated-message', 'get-env', 'get-resource-links', 'get-resource-reference', 'get-structured-content', 'get-sum', 'get-tiny-image', 'gzip-file-as-resource', 'toggle-simulated-logging', 'toggle-subscriber-updates', 'trigger-long-running-operation', 'simulate-research-query']

Loaded 30 tools from spec including the MCP tools above.
```

## Tests

```bash
poetry run pytest tests/unit/app_server/test_process_sandbox_service.py tests/unit/app_server/test_agent_server_env_override.py -q -k "process"
poetry run pytest tests/unit/app_server/test_live_status_app_conversation_service.py -q -k "mcp or build_start_conversation_request"
poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files openhands/app_server/sandbox/process_sandbox_service.py openhands/app_server/sandbox/process_sandbox_spec_service.py tests/unit/app_server/test_process_sandbox_service.py tests/unit/app_server/test_agent_server_env_override.py
```

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-a5dfc7d   docker.openhands.dev/openhands/openhands:a5dfc7d
```


neubig:
@openhands:

1. are you very sure that this is correct? I think that RUNNING means a different thing semantically throughout the codebase?
2. this is exactly the symptom that a user reported, does this fix actually demonstrate the problem that was reported or a different problem? if this is incorrect, then close this PR and try again from the top to reproduce the issue that the user reported where the LLM does not see the MCP servers configured.

> I'm running into issues getting MCP servers configured. I added them to the UI, verified I can connect to them from the openhands-app and agent contianer, but they are not showing up to the llm. I'm using the latest openhands and I'm running using uv tool install openhands --python 3.12. Thanks.

neubig:
Thanks for calling this out — especially the `RUNNING` semantics. I agree that the public `SandboxStatus.RUNNING` should mean "the sandbox is usable / agent-server is health-ready", not merely "the OS process exists".

I pushed `a5dfc7d65` to make that explicit:

- `_get_process_status()` now treats any live, non-stopped, non-dead process as `STARTING` pending the HTTP health check.
- `_process_to_sandbox_info()` is the only place that promotes the sandbox to `RUNNING`, and only after the agent-server health check succeeds; only then do we expose the agent-server URL and session key.
- Stopped processes remain `PAUSED`; zombie/dead processes remain `MISSING`; live but stale/unhealthy processes become `ERROR` after the startup grace period.

So after this follow-up, I’m much more comfortable that the change preserves the codebase-wide meaning of `RUNNING`.

On the second question: the fix demonstrates the local/process-sandbox version of the reported symptom, not just settings persistence. In the before case, the UI-saved MCP config existed, but the local process sandbox never became usable because the healthy idle uvicorn process was classified as not ready, so the app did not expose the agent-server URL/session key and the conversation could not complete tool initialization. In the after validation, the same UI-configured `everything` MCP server reached agent startup and the agent-server logged creation of its 13 MCP tools. The existing/targeted unit coverage also checks that custom MCP config is merged into the SDK agent config.

Caveat: if the exact user report was from the Docker sandbox path (they mention an “agent container”), this PR should not be treated as a complete Docker-path reproduction/fix. I updated the PR body to narrow the claim to the local/process sandbox behavior. I do not think the PR should be closed as incorrect for that path, but if we need to prove/fix the Docker-container case specifically, that should be a separate reproduction/investigation.

Validation rerun after the follow-up:

```bash
poetry run pytest tests/unit/app_server/test_process_sandbox_service.py -q
poetry run pytest tests/unit/app_server/test_process_sandbox_service.py tests/unit/app_server/test_agent_server_env_override.py -q -k "process"
poetry run pytest tests/unit/app_server/test_live_status_app_conversation_service.py -q -k "mcp or build_start_conversation_request"
poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files openhands/app_server/sandbox/process_sandbox_service.py openhands/app_server/sandbox/process_sandbox_spec_service.py tests/unit/app_server/test_process_sandbox_service.py tests/unit/app_server/test_agent_server_env_override.py
```

This comment was created by an AI agent (OpenHands) on behalf of the user.

PR #14288, opened 2026-05-04
feat(chat): add /model slash command to list and switch LLM profiles

VascoSch92:

## Summary

Introduces a /model slash command for V1 conversations, letting users list and switch their saved LLM profiles without leaving the chat.                                                                                               
   
- /model renders an inline, expandable list of saved profiles (name, model, base_url, api_key set) anchored to the chat position where the command was typed.                                                                        
- /model <name> switches the running conversation's LLM to the named profile and renders an info block ("ℹ️ Switched to profile name") at the same position.

## Video/Screenshots

https://github.com/user-attachments/assets/49fa1cef-b973-4a81-a879-494e3879af4c

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


VascoSch92:
@hieptl Could you give a look if you have time?

xingyaoww (commented):
Just some high-level UX comments:

I wonder if this is the right UX? I would prefer to add a drop down next to "Code" that displays the current model name, where you can click it and select a different model name from the drop down

Image

xingyaoww (approved):
LGTM

PR #14295, opened 2026-05-04
fix(acp): bootstrap-prompt resume when sandbox is recycled (solution A, #14260)

### LINKED ISSUE #14260: ACP: conversations cannot be resumed after sandbox restart (session_id is not persisted)
## Problem

ACP conversations (Claude Code, Codex, Gemini CLI) cannot be resumed after a cloud sandbox is recycled. The SDK already implements ACP `session/load` correctly, but the two pieces of state it depends on both live inside the sandbox filesystem and are wiped on recycle.

## How resume works today

The OpenHands SDK already wires `session/load`:

- `ACPAgent.init_state()` writes `acp_session_id` and `acp_session_cwd` into `state.agent_state`, which `ConversationState` autosaves into `base_state.json`. (`software-agent-sdk:openhands-sdk/openhands/sdk/agent/acp_agent.py:870–878`)
- `ACPAgent._start_acp_server()` reads them back on the next launch and calls `conn.load_session(cwd, session_id, mcp_servers=[])`, falling back to `new_session` on `ACPRequestError`. (`acp_agent.py:1009–1027`)
- A `cwd` guard forces a fresh session if the workspace path changed, since every ACP server keys its own session storage by `cwd`.

So when an in-sandbox restart happens (e.g. agent-server crash but sandbox FS preserved), resume works.

## Why it fails on cloud sandbox recycle

`session/load` is **not** an over-the-wire transcript transport. It is a request to the ACP server to stream the session it *already has* in its own storage. The protocol has no `session/save` and no client-supplied history payload. So a successful resume requires *two* pieces of state to survive:

| State | Stored in | Survives sandbox recycle? |
|---|---|---|
| OpenHands events (MessageEvent, ACPToolCallEvent, …) | Durable cloud event store | ✅ |
| `acp_session_id` / `acp_session_cwd` (in `state.agent_state` → `base_state.json`) | Sandbox `LocalFileStore` | ❌ |
| ACP server's own session files (`~/.claude/projects/…`, `~/.codex/…`, `~/.gemini/…`) | Sandbox `$HOME` | ❌ |

`live_status_app_conversation_service._build_acp_start_conversation_request` doesn't mirror either of the latter two into durable storage, and on the new sandbox they're both gone — so even though the SDK *would* call `load_session` if it found a saved id, there's no id to find and no server-side session to load.

## ACP support across target agents

All three agents OpenHands supports advertise `loadSession` ([registry matrix](https://github.com/agentclientprotocol/registry/blob/main/.protocol-matrix/latest.md)):

- **claude-agent-acp** — session files in `~/.claude/projects/<encoded-cwd>/<sid>.jsonl`. Config dir configurable via `CLAUDE_CONFIG_DIR` env var.
- **codex-acp** — session files in `~/.codex/`. Data dir configurable via `CODEX_HOME` env var.
- **gemini** — reads `ConversationRecord` from gemini-cli core config. No known env var to relocate storage.

## Proposed fixes

Two paths, not mutually exclusive — recommend shipping (1) first.

### 1. Bootstrap-prompt resume (shipped — PR #14295)

On restart, start a fresh `new_session` and synthesize the first user message from the durable event store: render prior `MessageEvent`s as role-tagged turns plus a compressed summary of `ACPToolCallEvent`s, prefix it with a `<<RESUMED CONVERSATION>>` marker, and pass it as `initial_message` on `StartACPConversationRequest`.

Single-PR change in `live_status_app_conversation_service`. Loses tool-call provenance and any sandbox-FS state the agent had built up; works against all three agents identically.

### 2. Native `session/load` resume (preferred long-term)

Two halves, both required:

**2a — Persist `acp_session_id` durably.** Mirror it (and `acp_session_cwd`) into the app-server database alongside `AppConversationInfo` on each turn boundary, and either pre-seed `base_state.json` in the new sandbox or extend `StartACPConversationRequest` / `ACPAgent` with explicit `acp_resume_session_id` so the resume signal doesn't depend on sandbox FS.

**2b — Persist the ACP server's own session storage (no new infrastructure required).**

Claude Code and Codex expose env vars to redirect their entire data directory:
- `CLAUDE_CONFIG_DIR=<path>` — redirects all of `~/.claude/` (sessions, history, settings)
- `CODEX_HOME=<path>` — redirects all of `~/.codex/` (sessions, state DB, config)

The `/workspace` PVC **already persists across sandbox recycles** (it survives because its `ownerReference` points to the Deployment, not the Pod). We can simply point each CLI into a subdirectory of the existing workspace volume:

```
CLAUDE_CONFIG_DIR=/workspace/.claude
CODEX_HOME=/workspace/.codex
```

These env vars are injected into the ACP subprocess via `acp_env` in `_build_acp_start_conversation_request` — a pure app-server change, no new PVCs, no runtime-api changes, no Helm changes.

**Caveat — Claude Code auth conflict:** the SDK currently strips `ANTHROPIC_API_KEY` and `ANTHROPIC_BASE_URL` when `CLAUDE_CONFIG_DIR` is detected, assuming OAuth file-based auth. Since we're setting `CLAUDE_CONFIG_DIR` for persistence (not OAuth), this stripping logic in `acp_agent.py` needs to be decoupled — the presence of `CLAUDE_CONFIG_DIR` should not imply dropping the API key.

**Gemini:** no known relocatable config dir env var; Solution 1 (bootstrap-prompt) remains the fallback for Gemini.

Half (2a) should land regardless — solution (1) also benefits from knowing "this is a restart."

## Relevant code

- `software-agent-sdk:openhands-sdk/openhands/sdk/agent/acp_agent.py` — `_start_acp_server` already calls `conn.load_session` when `acp_session_id` is present in `state.agent_state`; `_ENV_CONFLICT_MAP` strips `ANTHROPIC_API_KEY` when `CLAUDE_CONFIG_DIR` is set (needs decoupling for 2b).
- `software-agent-sdk:openhands-sdk/openhands/sdk/conversation/state.py:273–390` — `ConversationState.create()` is open-or-create against the per-conversation `LocalFileStore`; resumes from `base_state.json` if present.
- `OpenHands:openhands/app_server/app_conversation/live_status_app_conversation_service.py` — `_build_acp_start_conversation_request` is where `acp_env` is assembled; target for injecting `CLAUDE_CONFIG_DIR` / `CODEX_HOME`.

simonrosenberg:
## Why

ACP conversations cannot be resumed after a cloud sandbox recycle. The ACP server's own session storage (\`~/.claude/projects/…\`, \`~/.codex/…\`, \`~/.gemini/…\`) lives inside the sandbox filesystem and is wiped when the sandbox is recycled. Although the SDK already wires \`session/load\`, there is no session for it to load.

This PR ships **Solution A** from issue OpenHands/OpenHands#14260: bootstrap-prompt resume. No ACP server cooperation is required; it works identically against Claude Code, Codex, and Gemini CLI.

## What changed

**\`live_status_app_conversation_service.py\`**

New module-level helpers:
- \`_truncate_text(text, max_chars)\` — truncates with \`...\` suffix
- \`_content_to_text(content)\` — renders text content, images as \`[Image: N URL(s)]\`, unknown types as \`[ClassName]\`

New method \`_synthesize_acp_resume_initial_message(conversation_id, initial_message=None)\`:
- Fetches events **newest-first** (\`TIMESTAMP_DESC\`) and reverses before rendering, so for conversations exceeding \`_ACP_RESUME_MAX_EVENTS\` the agent receives the most recent context
- Hard event-count cap (\`_ACP_RESUME_MAX_EVENTS = 200\`) prevents unbounded fetches; character caps bound output size (\`_ACP_RESUME_CONTEXT_MAX_CHARS = 60_000\`, \`_ACP_RESUME_MESSAGE_MAX_CHARS = 8_000\`, \`_ACP_RESUME_TOOL_MAX_CHARS = 2_000\`)
- Filters to \`MessageEvent\` (rendered as \`[USER]\` / \`[ASSISTANT]\` role-tagged lines) and \`ACPToolCallEvent\` (rendered as \`[TOOL USE: <name>] (<status>)\`, including \`raw_input\` / \`raw_output\` where available)
- Prefixes the block with \`<<RESUMED CONVERSATION>>\` so the agent knows the context
- Returns \`None\` for fresh conversations — zero behaviour change on first start
- When \`initial_message\` is provided alongside prior events, appends it as a **second content block** rather than discarding it
- **Double-resume guard**: skips synthesis if \`initial_message\` already starts with the marker
- **Graceful error fallback**: if \`search_events\` raises, returns \`initial_message\` unchanged

\`_build_acp_start_conversation_request\` now passes \`initial_message\` through to \`_synthesize_acp_resume_initial_message\` and uses the result, giving the agent both the history block and any new user turn.

**\`tests/unit/app_server/test_live_status_app_conversation_service.py\`**

New class \`TestSynthesizeAcpResumeInitialMessage\` with 14 tests covering:
- No events → \`None\`
- Only irrelevant events → \`None\`
- \`MessageEvent\`s → role-tagged turns
- \`ACPToolCallEvent\`s → tool-use lines with status
- Mixed events → chronological order preserved after DESC fetch + reverse
- Multi-page pagination → all pages collected up to the count cap
- Count cap stops pagination at \`_ACP_RESUME_MAX_EVENTS\`
- Long conversations → output truncated to \`_ACP_RESUME_CONTEXT_MAX_CHARS\` with \`...\`
- Double-resume guard → already-resumed message not re-wrapped, \`search_events\` not called
- \`search_events\` error → \`initial_message\` returned unchanged
- New user message preserved as second content block
- \`raw_input\` / \`raw_output\` included in tool rendering
- Integration: fresh start (no resume), restart with no user message (synthesized history as \`initial_message\`)

## Limitations / follow-up

Solution A loses tool-call provenance and any sandbox-FS state the agent had built up. Solution B (runtime-scoped PVC at \`/home/openhands\`, same mechanism already used for \`/workspace\`) is the preferred long-term path — see OpenHands/OpenHands#14260 for details.

## Issue Number

Closes OpenHands/OpenHands#14260

## Test plan

- [x] \`uv run pytest tests/unit/app_server/test_live_status_app_conversation_service.py::TestSynthesizeAcpResumeInitialMessage -v\` — all 14 pass
- [x] \`uv run pytest tests/unit/app_server/test_live_status_app_conversation_service.py::TestBuildAcpStartConversationRequestSecrets -v\` — all 9 pass
- [ ] Start an ACP conversation (Claude Code / Codex / Gemini), let it run a few turns, recycle the sandbox, reconnect — conversation should resume with history context visible to the agent

## Type

- [x] Bug fix

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-ca1c4b1   docker.openhands.dev/openhands/openhands:ca1c4b1
```


simonrosenberg:
Superseded by OpenHands/OpenHands#14640 — a squashed, rebased-onto-`main` refresh.

This branch was ~a month / 19 commits stale and conflicting, and its follow-up note pointed at the now-rejected `/home/openhands` PVC option. Same Solution-A implementation (merged cleanly onto current `main`; only a `get_user_email` test-mock conflict needed resolving), re-framed as the **lossy fallback tier** in the resume ladder for OpenHands/agent-canvas#2116. **134 tests green.** Closing in favour of OpenHands/OpenHands#14640.

PR #14306, opened 2026-05-05
fix: handle persisted agent_kind='openhands' in org responses

hieptl:

---

### Description

`GET /api/organizations` returns **500 Internal Server Error** for any user whose org row contains `agent_kind: 'openhands'` in the persisted `agent_settings` JSON. This blocks the post-signup flow: the user accepts ToS, the frontend immediately calls `/api/organizations`, and the dashboard never loads.

### Reproduction steps

1. Launch the SaaS locally via the `launch.json` configuration in `.vscode/`.
2. Sign up a new account.
3. Accept the Terms of Service.
4. Observe the request:
   ```
   GET http://localhost:3030/api/organizations  →  500
   ```

### Expected behavior

`GET /api/organizations` returns **200** with a paginated list of the user's organizations.

### Actual behavior

500 Internal Server Error. From the backend log:

```
File 'enterprise/server/routes/org_models.py', line 188, in from_org  agent_settings=AgentSettings.model_validate(...)
pydantic_core._pydantic_core.ValidationError: 1 validation error for AgentSettings agent_kind Input should be 'llm' [type=literal_error, input_value='openhands', input_type=str]
```

### Root cause

`OrgResponse.from_org` and `OrgDefaultsSettingsResponse.from_org` validate persisted `org.agent_settings` through the SDK's **deprecated** `AgentSettings` class. That class inherits from `LLMAgentSettings`, which pins `agent_kind: Literal["llm"]`. The persisted rows now carry `agent_kind: 'openhands'` (the canonical value introduced when the SDK renamed the agent class from `LLMAgentSettings` → `OpenHandsAgentSettings`). The literal check rejects them and the endpoint 500s.

A working precedent already exists at `enterprise/storage/org_store.py:52-57` (`OrgStore.get_agent_settings_from_org`) — it uses the canonical `OpenHandsAgentSettings` and forces `agent_kind = 'openhands'` to absorb legacy `'llm'` rows. The two response serializers in `org_models.py` predate that precedent.

### Scope

- `enterprise/server/routes/org_models.py` — both `OrgResponse.from_org` (list / get / create endpoints) and `OrgDefaultsSettingsResponse.from_org` (org-defaults settings endpoint) are affected.
- Out of scope: `enterprise/storage/user_settings.py:92` uses the same deprecated pattern but is dead code in production (no caller; runtime path is `SaasSettingsStore.load` → `OrgStore.get_agent_settings_from_org`). Tracked separately for cleanup.

### Acceptance criteria

- [ ] `GET /api/organizations` returns **200** for users whose org rows have `agent_kind: 'openhands'`.
- [ ] The fix preserves backward compatibility with legacy `agent_kind: 'llm'` rows and rows with no discriminator at all.
- [ ] Regression tests cover all three persisted shapes for both call sites.
- [ ] No frontend changes required (verified — frontend never reads `agent_kind`).

## Summary

- Fixes a 500 on `GET /api/organizations` (and `GET /api/organizations/{id}/settings`) triggered by persisted org rows whose `agent_settings.agent_kind` is `'openhands'` — the canonical value in the modernized SDK, which the deprecated `AgentSettings` class rejects.
- Switches both response serializers in `enterprise/server/routes/org_models.py` to the canonical `OpenHandsAgentSettings`, mirroring the working precedent in `enterprise/storage/org_store.py:52-57`.
- Adds two regression tests covering the bug case plus legacy/missing-`agent_kind` backward compatibility.

## Root cause

`AgentSettings` (deprecated; scheduled for removal in SDK v1.22) inherits from `LLMAgentSettings`, which pins `agent_kind: Literal["llm"]`. Persisted rows now contain `agent_kind: 'openhands'`, so `AgentSettings.model_validate(...)` raises:

```
ValidationError: 1 validation error for AgentSettings
agent_kind
  Input should be 'llm' [type=literal_error, input_value='openhands', input_type=str]
```

Repro path:

```
GET /api/organizations
  → routes/orgs.py:118  list_user_orgs
  → routes/org_models.py:188  OrgResponse.from_org
  → AgentSettings.model_validate(dict(org.agent_settings))   ← 500 here
```

## Fix

`enterprise/server/routes/org_models.py`:

1. Replace `AgentSettings` import with `OpenHandsAgentSettings`.
2. Change the `agent_settings` field type and `default_factory` on `OrgResponse` and `OrgDefaultsSettingsResponse` to `OpenHandsAgentSettings`.
3. Route both `from_org` call sites through a new module-level helper:
   ```python
   def _validate_persisted_agent_settings(raw):
       kwargs = dict(raw) if raw else {}
       kwargs['agent_kind'] = 'openhands'
       return OpenHandsAgentSettings.model_validate(kwargs)
   ```
This forces `agent_kind = 'openhands'` so both legacy `'llm'` rows and current `'openhands'` rows validate against the canonical class — same approach as `OrgStore.get_agent_settings_from_org` (`enterprise/storage/org_store.py:52-57`). Helper is inlined rather than imported from `org_store` because `org_store` already imports from `org_models` (would deadlock as a circular import).
4. Update the `_denormalize_llm_for_response(agent_settings: OpenHandsAgentSettings)` annotation.

### Why the field-type change is safe

The frontend does not consume `agent_kind` (verified by grep over `frontend/src` and `frontend/__tests__` — zero references to `agent_kind` / `agentKind`). Only `agent_settings.llm.model`, `llm.base_url`, and `llm_api_key_set` are read. The shift in the serialized response from `"llm"` → `"openhands"` is invisible to the UI.

### Out of scope

- `enterprise/storage/user_settings.py:92` (`UserSettings.to_settings()`) uses the same deprecated pattern but is dead code in production. Left for a separate cleanup PR.
- `enterprise/storage/org_service.py:111` constructs a default `AgentSettings()` (no persisted data); only emits a `DeprecationWarning`, no 500.

## How to Test

1. Launch the SaaS locally via the `launch.json` configuration in `.vscode/`.
2. Sign up a new account.
3. Accept the Terms of Service.
4. Observe the request:
   ```
   GET http://localhost:3030/api/organizations  →  500
   ```

## Video/Screenshots

https://github.com/user-attachments/assets/8b511645-7d31-4efa-b725-170335ab78f6

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-cedd3ad   docker.openhands.dev/openhands/openhands:cedd3ad
```

### DIFF
--- a/enterprise/server/routes/org_models.py
+++ b/enterprise/server/routes/org_models.py
@@ -15,7 +15,22 @@ from storage.org_member import OrgMember
 from storage.role import Role
 
 from openhands.app_server.utils.llm import MASKED_API_KEY, resolve_llm_base_url
-from openhands.sdk.settings import AgentSettings, ConversationSettings
+from openhands.sdk.settings import ConversationSettings, OpenHandsAgentSettings
+
+
+def _validate_persisted_agent_settings(
+    raw: dict[str, Any] | None,
+) -> OpenHandsAgentSettings:
+    """Validate persisted ``org.agent_settings`` against the canonical schema.
+
+    Older rows carry the legacy ``agent_kind: 'llm'`` discriminator from the
+    pre-rename SDK; force ``'openhands'`` so the canonical class accepts both
+    shapes. Mirrors :func:`OrgStore.get_agent_settings_from_org` — kept inline
+    to avoid a circular import (``org_store`` already imports from this module).
+    """
+    kwargs = dict(raw) if raw else {}
+    kwargs['agent_kind'] = 'openhands'
+    return OpenHandsAgentSettings.model_validate(kwargs)
 
 
 class OrgCreationError(Exception):
@@ -155,7 +170,9 @@ class OrgResponse(BaseModel):
     sandbox_base_container_image: str | None = None
     sandbox_runtime_container_image: str | None = None
     org_version: int = 0
-    agent_settings: AgentSettings = Field(default_factory=AgentSettings)
+    agent_settings: OpenHandsAgentSettings = Field(
+        default_factory=OpenHandsAgentSettings
+    )
     conversation_settings: ConversationSettings = Field(
         default_factory=ConversationSettings
     )
@@ -185,9 +202,7 @@ class OrgResponse(BaseModel):
             sandbox_base_container_image=org.sandbox_base_container_image,
             sandbox_runtime_container_image=org.sandbox_runtime_container_image,
             org_version=org.org_version if org.org_version is not None else 0,
-            agent_settings=AgentSettings.model_validate(
-                dict(org.agent_settings) if org.agent_settings else {}
-            ),
+            agent_settings=_validate_persisted_agent_settings(org.agent_settings),
             conversation_settings=ConversationSettings.model_validate(
                 dict(org.conversation_settings) if org.conversation_settings else {}
             ),
@@ -371,7 +386,9 @@ class OrgUpdate(BaseModel):
 class OrgDefaultsSettingsResponse(BaseModel):
     """Response model for organization default settings."""
 
-    agent_settings: AgentSettings = Field(default_factory=AgentSettings)
+    agent_settings: OpenHandsAgentSettings = Field(
+        default_factory=OpenHandsAgentSettings
+    )
     conversation_settings: ConversationSettings = Field(
         default_factory=ConversationSettings
     )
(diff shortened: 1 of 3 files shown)


tofarr (approved):
🍰

Issue #14304, opened 2026-05-05
[Bug]: Bitbucket provider authentication broken (API tokens not working)


FractalMind:
### Is there an existing issue for the same bug?

- [x] I have searched existing issues and this is not a duplicate.

### Bug Description

The Bitbucket integration in OpenHands appears to be non-functional due to outdated authentication methods. This blocks a core feature of the platform, which is connecting to repositories via providers.

I have successfully installed OpenHands on a self-hosted VPS, and the system works fine otherwise. However, connecting to Bitbucket fails consistently.

### Problem Summary
Bitbucket App Passwords are deprecated
Bitbucket now requires API tokens
OpenHands still appears to rely on the old authentication format or lacks proper support for the new one
All attempted authentication formats fail

### This issue occurs:

On self-hosted OpenHands
On the cloud version of OpenHands

<img width="3724" height="1929" alt="Image" src="https://github.com/user-attachments/assets/26e949b1-9aa1-474e-8dd6-fc1bba4dfdda" />
<img width="3724" height="1927" alt="Image" src="https://github.com/user-attachments/assets/5f719231-c49f-4f26-8a41-0c521c4ac747" />
<img width="1980" height="1513" alt="Image" src="https://github.com/user-attachments/assets/8109ace6-11bb-4805-b340-4440368e4898" />
<img width="3724" height="1927" alt="Image" src="https://github.com/user-attachments/assets/b1c598b5-2999-4094-9a94-8170b5c2f1d0" />

### Additional Context
This appears to be a systemic issue, not configuration-specific
There are a few existing issues mentioning Bitbucket problems, but none clearly address API token support
This blocks a core use case of OpenHands (repo integration)
Attachments

Suggested Fix
Update Bitbucket provider to support API token authentication
Align with Bitbucket’s current authentication standards
Provide clear documentation on the expected auth format

### Expected Behavior

OpenHands should support authentication with Bitbucket using modern API tokens, similar to how tools like n8n handle it.

For example, in n8n, authentication works using:

username:api_token

### Actual Behavior

All authentication attempts fail when using Bitbucket as a provider.

Attempted Formats

Tried multiple variations including:

username:api_token
email:api_token
x-token-auth:api_token

None worked.

### Steps to Reproduce

Install OpenHands (self-hosted or cloud)
Attempt to connect a Bitbucket repository
Use a valid Bitbucket API token
Observe authentication failure

### OpenHands Installation Method

Local GUI (Docker web interface)

[...]

PR #14309, opened 2026-05-06
fix(bitbucket): route Atlassian API tokens to Basic auth

Ayush-Mahadik:
## Summary

Atlassian API tokens (24 lowercase characters from https://id.atlassian.com/manage-profile/security) were being sent as Bearer auth, which Bitbucket Cloud rejects with HTTP 401. The existing `:` check correctly routes app passwords to Basic auth but misses Atlassian tokens since they contain no colon.

## Root Cause

In `openhands/app_server/integrations/bitbucket/service/base.py`, `_get_headers` checked only for `:` in the token to decide between Basic and Bearer:

Atlassian tokens are 24 lowercase chars, no colon, no uppercase, no dashes — so they fell through to Bearer auth, which Bitbucket Cloud rejects for these tokens.

## Fix

Add `_is_atlassian_api_token()` to detect the 24-char all-lowercase pattern and route those tokens to Basic auth with `'bitbucket.org'` as the username:

Auth routing now:
| Token format | Auth method |
|---|---|
| 24 lowercase chars (Atlassian token) | Basic: `bitbucket.org:<token>` |
| Contains `:` (app password) | Basic: `token` (unchanged) |
| Otherwise (PAT, uppercase/dashes) | Bearer (unchanged) |

## Changes

**openhands/app_server/integrations/bitbucket/service/base.py**
- Add `_is_atlassian_api_token()` helper — regex `[a-z0-9]{24}` pattern match
- Update `_get_headers()` to check Atlassian tokens before the `:` check

**tests/unit/integrations/bitbucket/test_bitbucket.py**
- Unit tests for `_is_atlassian_api_token()`: correct length, wrong length, uppercase, dashes, colon
- Integration tests for `_get_headers()`: all 3 auth paths verified

Fixes #7684 (no wait — different repo). This is a new bug.

### DIFF
--- a/openhands/app_server/integrations/bitbucket/service/base.py
+++ b/openhands/app_server/integrations/bitbucket/service/base.py
@@ -34,9 +34,7 @@ def _is_atlassian_api_token(token_value: str) -> bool:
 
 
 class BitBucketMixinBase(BaseGitService, HTTPClient):
-    """
-    Base mixin for BitBucket service containing common functionality
-    """
+    """Base mixin for BitBucket service containing common functionality."""
 
     BASE_URL = 'https://api.bitbucket.org/2.0'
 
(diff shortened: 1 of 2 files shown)


cbagwell:
I believe this change would break the cloud support for Bitbucket as the modified code path is likely used by oauth authentication which requires Bearer.  

You should be able to get it to partially work using info in #14304 and providing `<email>:<api_token>` with a properly scoped token.  That issue also describes remaining changes still required if you're interesting in creating that PR.

Ayush-Mahadik:
> I believe this change would break the cloud support for Bitbucket as the modified code path is likely used by oauth authentication which requires Bearer.  
> 
> You should be able to get it to partially work using info in #14304 and providing `<email>:<api_token>` with a properly scoped token.  That issue also describes remaining changes still required if you're interesting in creating that PR.

Yes , I suppose, it only works well fraction of the code not the full codebase. I am on it sir

PR #14318, opened 2026-05-06
Revert "fix(acp): route pending message delivery through correct agent-server prefix"

simonrosenberg:
## Why this revert is necessary

PR #14316 diagnosed a real asymmetry — `_process_pending_messages` hardcoded `/api/conversations/` while other call sites used `_agent_kind_to_router_path` — and concluded that ACP pending messages were being delivered to the wrong endpoint.

That diagnosis was wrong.

The agent server exposes exactly **one** events write endpoint:

```
POST /api/conversations/{id}/events
```

There is no `POST /api/acp/conversations/{id}/events`. The `/api/acp/conversations` router only has create and read endpoints; the `event_router` is mounted once, under `/api/conversations`, and serves all conversations regardless of kind. Internally, `get_event_service` is a plain UUID lookup in a unified `_event_services` dict — it has no concept of agent kind.

The "fix" in #14316 used `_agent_kind_to_router_path('acp')` to build the events URL, which produced `/api/acp/conversations/{id}/events` — a path that returns 404. This would have silently dropped pending messages for ACP conversations in cloud environments instead of delivering them.

The original hardcoded `/api/conversations/{id}/events` was semantically correct.

The `/api/acp/` prefix is justified only for create and read, where the request/response schemas genuinely differ (`StartACPConversationRequest` vs `StartConversationRequest`, `ACPConversationInfo` vs `ConversationInfo`). It does not apply to events, which are kind-agnostic.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-513b2ba   docker.openhands.dev/openhands/openhands:513b2ba
```


malhotra5 (approved):
LGTM

PR #14324, opened 2026-05-07
fix: validate conversation storage path components (CWE-22)

sebastionoss:
## Summary

Adds an allowlist-based path component validator at `get_conversation_dir()` in `openhands/storage/locations.py`, which is the central function used to construct on-disk paths for conversation event streams and metadata. The validator URL-decodes the value and rejects anything outside `[A-Za-z0-9_-]`, so traversal sequences (`..`, `/`, `\`, NUL bytes, percent-encoded variants) cannot be used to compose paths that escape the conversation storage root.

## Vulnerability details

- **CWE-22**: Improper limitation of a pathname to a restricted directory (path traversal)
- **Affected function**: `openhands.storage.locations.get_conversation_dir(sid, user_id)`
- **Sinks reached from this function**: `EventStream` event paths, `FileConversationStore` metadata paths, and any other component that derives storage keys from `get_conversation_dir`
- **Data flow**: Several call sites pass `sid` / `user_id` values that originated outside FastAPI's UUID-typed routes — for example legacy header-based identifiers (e.g. `X-OpenHands-ServerConversation-ID` handled in `mcp.py`) and internal subsystems that forward identifiers without re-validating them. None of these sites ran an allowlist check before composing a filesystem path. A caller able to supply a value like `../another-conversation` or its URL-encoded form could cause reads/writes to land outside the intended conversation directory.

## Fix and rationale

The new helper is placed at the storage chokepoint rather than at every route, so it protects current and future callers uniformly:

```python
_SAFE_PATH_COMPONENT_RE = re.compile(r'^[A-Za-z0-9_-]+$')

def _validate_path_component(value: str, name: str) -> None:
    decoded_value = unquote(value)
    if not _SAFE_PATH_COMPONENT_RE.fullmatch(decoded_value):
        raise ValueError(...)
```

Why an allowlist:
- Conversation IDs in OpenHands are UUID-like strings; user IDs are also alphanumeric/underscore/hyphen. A strict character allowlist matches actual usage and is far safer than a denylist of `..`, `/`, etc.
- `unquote()` is applied first so `%2e%2e%2f...` payloads are evaluated against the same allowlist rather than slipping through unchanged.
- Raising `ValueError` keeps the failure mode loud and consistent with existing `validate_conversation_id` helpers elsewhere in the codebase.

The change is intentionally small (16 lines of production code) and additive — no existing valid identifier is rejected, as confirmed by the new UUID test below.

## Tests

Added two focused unit tests:

1. `tests/unit/storage/conversation/test_file_conversation_store.py`
   - `test_conversation_metadata_filename_rejects_traversal_ids` — covers `../outside`, `..\\outside`, `safe/../outside`, percent-encoded `%2e%2e%2foutside`, and a NUL-byte payload. All raise `ValueError`.
   - `test_conversation_metadata_filename_allows_uuid_like_ids` — confirms a real UUID still produces the expected `sessions/<uuid>/metadata.json` path.

2. `tests/unit/events/test_event_stream.py`
   - `test_event_stream_rejects_traversal_conversation_id` — constructs an `EventStream` with a `../sibling` ID, asserts `add_event` raises, and asserts no file was written outside the intended directory.

These tests exercise both call paths (event stream + file conversation store) that consume `get_conversation_dir`.

## Security analysis

What the fix does and does not change:

- **Does**: closes the gap where a caller forwarded an untrusted `sid` or `user_id` to `get_conversation_dir` without going through the existing per-route `validate_conversation_id` check. It also adds URL-decoding so double-encoded payloads cannot bypass the check.
- **Complements**: existing protections — most public V1 routes already type `conversation_id: UUID` (which fully prevents traversal there), and the legacy `validate_conversation_id` helper still runs where wired in. The new check is the storage-layer backstop.
- **Does not**: rewrite or restructure any caller. Code paths that build sub-paths *after* `get_conversation_dir` returns (e.g. callbacks that concatenate a `subpath`) still need their own input validation; this PR does not attempt to refactor those, intentionally keeping the change minimal.

Preconditions for exploitation prior to this fix:
1. A caller passes attacker-controlled `sid` or `user_id` straight into `get_conversation_dir` without prior allowlist validation.
2. FastAPI's UUID typing is bypassed, e.g. via a non-UUID-typed header or an internal forwarding path.
3. The attacker-supplied value reaches the storage layer (raw or URL-encoded).

Before submitting, we tried to disprove this: we checked whether UUID typing on the public routes, the legacy `validate_conversation_id` helper, or the ownership lookups in enterprise feedback flows would already block every reachable path. They cover the main HTTP entry points but are opt-in and don't apply to all callers (notably header-derived IDs and internal subsystems), and none of them validate `user_id`. Adding the check at the storage chokepoint is the smallest change that makes the guarantee uniform, so it's worth landing as defense-in-depth even where a given route is currently safe.

## Diff size

3 files changed, 48 insertions(+), 0 deletions(-) — 16 lines of production code, 32 lines of tests.

cc @lewiswigmore

### DIFF
--- a/openhands/storage/locations.py
+++ b/openhands/storage/locations.py
@@ -1,8 +1,24 @@
+import re
+from urllib.parse import unquote
+
 CONVERSATION_BASE_DIR = 'sessions'
+
+_SAFE_PATH_COMPONENT_RE = re.compile(r'^[A-Za-z0-9_-]+$')
+
+
+def _validate_path_component(value: str, name: str) -> None:
+    decoded_value = unquote(value)
+    if not _SAFE_PATH_COMPONENT_RE.fullmatch(decoded_value):
+        raise ValueError(
+            f'Invalid {name}: only letters, numbers, hyphens, and underscores are allowed'
+        )
+
 
 
 def get_conversation_dir(sid: str, user_id: str | None = None) -> str:
+    _validate_path_component(sid, 'conversation id')
     if user_id:
+        _validate_path_component(user_id, 'user id')
         return f'users/{user_id}/conversations/{sid}/'
     else:
         return f'{CONVERSATION_BASE_DIR}/{sid}/'
--- a/tests/unit/events/test_event_stream.py
+++ b/tests/unit/events/test_event_stream.py
@@ -863,3 +863,16 @@ def test_nested_dict_secret_replacement(temp_dir: str):
     assert 'password123' not in data_with_secrets_replaced['args']['command']
     assert 'password123' not in data_with_secrets_replaced['args']['env']['SECRET_KEY']
     assert 'password123' not in data_with_secrets_replaced['args']['env']['timestamp']
+
+
+def test_event_stream_rejects_traversal_conversation_id(temp_dir: str):
+    file_store = get_file_store('local', temp_dir)
+    sibling = f'{temp_dir}_escape'
+    event_stream = EventStream('../' + os.path.basename(sibling), file_store)
+    try:
+        with pytest.raises(ValueError):
+            event_stream.add_event(NullAction(), EventSource.USER)
+    finally:
+        event_stream.close()
+
+    assert not os.path.exists(os.path.join(sibling, 'events', '0.json'))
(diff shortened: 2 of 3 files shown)


andesyteoss:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

PR #14133, opened 2026-04-25
feat(frontend): Add critic result types, component, and event rendering

xingyaoww:
## Description

Adds CriticResult visualization support to the GUI frontend, enabling the web UI to display critic evaluation results that are already flowing through SDK events.

This mirrors the critic visualization functionality from the [OpenHands CLI](https://github.com/OpenHands/OpenHands-CLI/tree/main/openhands_cli/tui/utils/critic), adapted for the React web frontend.

w/o iterative refinement
image

w/ iterative refinement
image

### Types (`frontend/src/types/v1/core/base/critic.ts`)
- **`CriticResult`** — `score` (0-1), `message`, `metadata`
- **`CriticFeature`** — `name`, `display_name`, `probability`
- **`CriticCategorizedFeatures`** — `agent_behavioral_issues`, `user_followup_patterns`, `infrastructure_issues`, `other`
- **`CriticMetadata`** — wraps categorized features and event IDs
- Added optional `critic_result` field to **`ActionEvent`** and **`MessageEvent`**

### Component (`critic-result-display.tsx`)
- Star rating (0-5 ★) with color coding: green ≥60%, yellow ≥40%, red <40%
- Percentage display
- Expandable categorized feature breakdown with per-feature probabilities
- Full i18n support (15 languages)

### Integration
- **`FinishEventMessage`** — renders `CriticResultDisplay` below the finish message when `critic_result` is present
- **`UserAssistantEventMessage`** — renders `CriticResultDisplay` for agent messages when `critic_result` is present

### Tests
- 12 unit tests covering score rendering, star ratings, color coding, label rendering, expand/collapse, and multiple feature categories

## Default behavior and deployment note

Critic is enabled by default for new OpenHands app users so critic feedback appears in the GUI without extra setup. Deployments that want to avoid critic usage or associated LLM costs can opt out by disabling `verification.critic_enabled` in user or organization verification settings.

When `CRITIC_API_KEY` is configured, critic requests use that deployment-level service key for centralized cost management instead of per-user budget-limited keys.

## How it works

The SDK already attaches `CriticResult` to `ActionEvent` and `MessageEvent` when the critic is enabled (via verification settings). The server sends these events over WebSocket. Previously the frontend ignored the `critic_result` field — now it renders an inline evaluation widget.

Example rendering:
```
Critic: agent success likelihood  ★★★★☆  (72.0%)
  ▼ Potential Issues: Insufficient Testing (75%) · Incomplete Changes (60%)
[...]

PR #14339, opened 2026-05-07
refactor(frontend): SdkSectionPage multi-source — simplify verification-settings.tsx

neubig:
## Why

Stacks on top of #14133. The original `verification-settings.tsx` had ~225 lines of bespoke code to render fields from two settings schemas (`conversation_settings` for `confirmation_mode`/`security_analyzer`, `agent_settings` for the critic family) on a single page, because `SdkSectionPage` could only handle one settings source.

This PR teaches `SdkSectionPage` about multiple settings sources, and converts every caller to the new API. `verification-settings.tsx` collapses to ~54 lines, and the other pages get a small, consistent API too.

## What changed

- **`SdkSectionPage`** — a single required `settingsSources: SettingsSourceConfig[]` prop. Each entry says "render section X from this source's schema". The page tracks values/dirty state per source and emits `{ conversation_settings_diff?, agent_settings_diff? }`, including only the buckets with dirty changes. The legacy single-source props (`settingsSource`/`sectionKeys`/`excludeKeys`) are gone, so there's exactly one API to learn.
- **`buildPayload`** — always receives the wrapped payload now (`{ agent_settings_diff: { llm: {...} } }`), not the unwrapped form. One contract, no special cases.
- **Callers converted**: `verification-settings`, `llm-settings`, `condenser-settings`, `org-default-condenser-settings`, plus the `sdk-section-page` test helpers.
- **Mock schema**: bumped the prominence of `enable_iterative_refinement` and `critic_mode` in `settings-handlers.ts` so the standard prominence filter shows the right tier in basic/advanced/all (matches what the original PR's hardcoded `BASIC_AGENT_VERIFICATION_FIELD_KEYS` set was doing). Production `agent_settings_schema` should land the same change for consistency — that's a small backend follow-up.

## Why not unify the schemas?

I considered moving the critic fields onto `ConversationSettings` so verification-settings would only need a single source. The SDK has *already* deliberated on this in the opposite direction: `confirmation_mode` and `security_analyzer` were moved **out of** `VerificationSettings` (agent) and **into** `ConversationSettings` in 1.17.0, with field validators emitting deprecation warnings, scheduled for removal in 1.22.0. Critic settings stayed on the agent because they describe *agent capabilities* (how the agent evaluates and retries its own work, where its critic LLM lives), not *conversation policy* (what gets confirmed/analyzed).

So the storage split is intentional and well-thought-out. The unified UX is the right call here too — both schemas tag their fields with `section: "verification"`, the SDK is explicitly inviting consumers to render them together. `settingsSources` is exactly that adapter.

There's a TODO in `verification-settings.tsx` to drop the `CONVERSATION_OWNED_AGENT_VERIFICATION_FIELD_KEYS` defensive de-dup once SDK 1.22 ships and removes the deprecated fields from `agent_settings_schema`.

## Verification

- 3/3 `verification-settings` tests pass
- 12/12 `sdk-section-page` tests pass (including the "preserves view across parent rerenders" test that flushed out a referential-stability bug — fixed by routing through a JSON signature, the same pattern the original code used)
- 52/52 `llm-settings` tests pass (validates the wrapped-payload `buildPayload` change)
- Full frontend suite: **2148/2148 passing**
- `npm run lint` and `npm run build` clean

## Diff

```
frontend/__tests__/components/features/settings/sdk-settings/sdk-section-page.test.tsx |  24 ++--
frontend/src/components/features/settings/sdk-settings/sdk-section-page.tsx            | 162 +++++++++++++--------
frontend/src/mocks/settings-handlers.ts                                                |   2 +-
frontend/src/routes/condenser-settings.tsx                                             |   4 +-
frontend/src/routes/llm-settings.tsx                                                   |  26 ++--
frontend/src/routes/org-default-condenser-settings.tsx                                 |   4 +-
frontend/src/routes/verification-settings.tsx                                          | 200 +-----------------------
```

verification-settings goes from 225 → 54 lines. SdkSectionPage grows by ~70 lines for the multi-source machinery. Net win.

---
This PR was created by an AI agent (OpenHands) on behalf of @rbren.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-98dbf5f   docker.openhands.dev/openhands/openhands:98dbf5f
```


xingyaoww (approved):
Thank you!

PR #14365, opened 2026-05-10
fix(acp): bootstrap resumed conversations after sandbox restart

### LINKED ISSUE #14260: ACP: conversations cannot be resumed after sandbox restart (session_id is not persisted)

Jerry2003826:
Fixes #14260

## Summary
- Bootstrap existing ACP conversations from the durable app-server event store when a sandbox restart loses the ACP server's local session storage.
- Prefix the synthesized first user message with `<<RESUMED CONVERSATION>>`, replay prior `MessageEvent`s as role-tagged turns, and include a compact ACP tool activity summary.
- Preserve any current user message by prepending the restored context instead of replacing the user's content.

## Root Cause
ACP session resume currently depends on `acp_session_id` / `acp_session_cwd` in sandbox-local `base_state.json` and on the ACP server's own local session files. When a cloud sandbox is recycled, both are lost even though OpenHands events remain durable in the app-server event store.

## Notes
This implements the issue's bootstrap-prompt recovery path. It does not change ACP protocol behavior, ACP server storage, or the SDK's native `session/load` flow.

## Validation
- `uv run pytest tests/unit/app_server/test_live_status_app_conversation_service.py -k existing_acp_conversation_bootstraps -q`
- `uv run pytest tests/unit/app_server/test_live_status_app_conversation_service.py -q`
- `uv run ruff check openhands/app_server/app_conversation/live_status_app_conversation_service.py tests/unit/app_server/test_live_status_app_conversation_service.py`
- `git diff --check`

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -95,8 +95,9 @@ from openhands.app_server.utils.llm_metadata import (
     get_llm_metadata,
     should_set_litellm_extra_body,
 )
-from openhands.sdk import Agent, AgentContext, LocalWorkspace
+from openhands.sdk import Agent, AgentContext, LocalWorkspace, MessageEvent
 from openhands.sdk.agent.acp_agent import ACPAgent
+from openhands.sdk.event.acp_tool_call import ACPToolCallEvent
 from openhands.sdk.hooks import HookConfig
 from openhands.sdk.llm import LLM
 from openhands.sdk.plugin import PluginSource
@@ -120,6 +121,58 @@ from openhands.tools.preset.planning import (
 _conversation_info_type_adapter = TypeAdapter(list[ConversationInfo | None])
 _acp_conversation_info_type_adapter = TypeAdapter(list[ACPConversationInfo | None])
 _logger = logging.getLogger(__name__)
+
+_ACP_RESUME_CONTEXT_MARKER = '<<RESUMED CONVERSATION>>'
+_ACP_RESUME_CONTEXT_MAX_CHARS = 60_000
+_ACP_RESUME_MESSAGE_MAX_CHARS = 8_000
+_ACP_RESUME_TOOL_MAX_CHARS = 2_000
+
+
+def _truncate_text(text: str, max_chars: int) -> str:
+    if len(text) <= max_chars:
+        return text
+    return text[: max_chars - 3] + '...'
+
+
+def _content_to_text(content: Sequence[Any]) -> str:
+    parts: list[str] = []
+    for item in content:
+        text = getattr(item, 'text', None)
+        if isinstance(text, str):
+            parts.append(text)
+            continue
+
+        image_urls = getattr(item, 'image_urls', None)
+        if image_urls:
+            parts.append(f'[Image: {len(image_urls)} URL(s)]')
+            continue
+
+        parts.append(str(item))
+    return '\n'.join(part for part in parts if part)
+
+
+def _render_acp_resume_message_event(event: MessageEvent) -> str | None:
+    message = event.to_llm_message()
+    text = _content_to_text(message.content).strip()
+    if not text:
+        return None
+    return (
+        f'{message.role.upper()}: {_truncate_text(text, _ACP_RESUME_MESSAGE_MAX_CHARS)}'
+    )
+
+
+def _render_acp_resume_tool_event(event: ACPToolCallEvent) -> str:
+    status = f' [{event.status}]' if event.status else ''
+    details: list[str] = []
+    if event.raw_input:
+        details.append(f'input={_truncate_text(str(event.raw_input), 500)}')
+    if event.raw_output:
+        details.append(f'output={_truncate_text(str(event.raw_output), 500)}')
+    detail_text = f'\n  {"; ".join(details)}' if details else ''
+    return _truncate_text(
+        f'ACP TOOL: {event.title}{status}{detail_text}',
+        _ACP_RESUME_TOOL_MAX_CHARS,
+    )
 
 
 def _agent_kind_to_router_path(agent_kind: str) -> str:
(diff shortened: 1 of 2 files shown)


Jerry2003826:
I noticed #14295 after opening this PR. This is not intended to compete with that work; it is a current-`main`, mergeable variant of the same bootstrap-resume direction for #14260.

I'm happy to close this in favor of #14295 if maintainers prefer that PR, or keep this available as a rebase/reference since CI is green against current `main`. The main behavioral detail here is that restored ACP context is prepended while preserving any current user message, rather than replacing it.

simonrosenberg:
Thanks @Jerry2003826 for your contribution, and sorry I wasn't quicker to answer!
I have included some of your features into my own branch and will try to merge my branch shortly, so I am closing this one in favor of https://github.com/OpenHands/OpenHands/pull/14295

PR #14371, opened 2026-05-11
prep(acp): dispatch live-status conversations by agent_kind

simonrosenberg:

---

## Why split this out

[#14227](https://github.com/OpenHands/OpenHands/pull/14227) bundles the minimal ACP settings UI together with a refactor of the live-status / sandbox-refresh code paths to dispatch by ``agent_kind``. The refactor is large but mechanical and conceptually independent — restructuring agent-server-call paths to be kind-aware without referencing the ACP UI, feature flag, conversation chip, or any other ACP feature surface.

Splitting it out lets reviewers evaluate the dispatch infrastructure separately from the additive feature work it enables.

## Stacked on #14373

This PR depends on [#14373](https://github.com/OpenHands/OpenHands/pull/14373) (which exposes the ``ACP_PROVIDERS`` registry + ``enable_acp`` flag). Targets `chore/sdk-bump-acp-registry`. Once #14373 merges, this PR's base auto-retargets to `main` and the diff stays the same.

## What changes

- **New ``openhands/app_server/app_conversation/agent_server_routing.py``** — ``agent_kind_to_router_path()`` maps an ``AppConversationInfo.agent_kind`` string to the agent-server route prefix (``conversations`` for OpenHands, ``acp/conversations`` for ACP). Also exposes ``ACP_SERVER_TAG`` as a single source of truth for the ``tags['acp_server']`` key (consumed by the feature PR).
- **``live_status_app_conversation_service.py``** — ``_get_live_conversation_info`` splits the incoming ID list by agent kind via the new ``_split_ids_by_kind`` helper, then fetches each bucket from the appropriate endpoint and merges the results. The request-build path delegates env-var resolution, session-mode selection, and command resolution to the SDK via ``ACPAgentSettings`` rather than the bespoke ``_acp_provider_env`` + ``_SERVER_KEY_MAP`` fallbacks that lived here before. Conversation URLs in ``_build_conversation`` use the new dispatcher.
- **``remote_sandbox_service.refresh_conversation``** — pulls the agent-server conversation URL through the same dispatcher so the refresh path is symmetric with start/live-status.

No lockfile changes — the SDK on ``main`` is already pinned to ``openhands-sdk==1.21.1`` (#14350); ``ACPAgentSettings.resolve_provider_env()`` and ``ACP_PROVIDERS`` ship from 1.21.0+.

## Tests

- Drops the ``TestAcpProviderEnv`` suite that covered the now-deleted ``_acp_provider_env`` helper.
- Adds ``TestSplitIdsByKind`` covering the UUID/string lookup against the str-keyed kind map. Without the ``str(cid)`` normalisation, ACP conversations silently routed to the OpenHands endpoint — this test is the regression guard for that.

``uv run pytest tests/unit/app_server/test_live_status_app_conversation_service.py -q`` → 118 passed.

## No user-visible behaviour change

The dispatcher returns the same endpoint that was previously hard-coded for each path; the only new behaviour is "route by ``agent_kind``" which on ``main`` always evaluates to ``conversations`` (no ACP conversations exist until #14227's feature flag lands).

## Followed by

- [#14227](https://github.com/OpenHands/OpenHands/pull/14227) — additive ACP UI + routing wire-up.

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-c1bb956   docker.openhands.dev/openhands/openhands:c1bb956
```

### DIFF
--- a/dev_config/python/.pre-commit-config.yaml
+++ b/dev_config/python/.pre-commit-config.yaml
@@ -59,8 +59,8 @@ repos:
             types-Markdown,
             pydantic,
             lxml,
-            "openhands-sdk==1.19.1",
-            "openhands-tools==1.19.1",
+            "openhands-sdk==1.21.1",
+            "openhands-tools==1.21.1",
             "sqlalchemy>=2.0",
           ]
         # To see gaps add `--html-report mypy-report/`
--- a/openhands/app_server/app_conversation/agent_server_routing.py
+++ b/openhands/app_server/app_conversation/agent_server_routing.py
@@ -1,0 +1,12 @@
+# Conversation tag key holding the ACP provider discriminator (e.g.
+# ``"claude-code"``, ``"codex"``, ``"gemini-cli"``, ``"custom"``). Single
+# source of truth for backend writers and frontend readers — see
+# ``frontend/src/utils/agent-display-label.ts`` for the matching constant.
+ACP_SERVER_TAG = 'acp_server'
+
+
+def agent_kind_to_router_path(agent_kind: str | None) -> str:
+    """Map an app conversation agent kind to the agent-server route prefix."""
+    if agent_kind == 'acp':
+        return 'acp/conversations'
+    return 'conversations'
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -23,6 +23,10 @@ from openhands.agent_server.models import (
     StartConversationRequest,
     TextContent,
 )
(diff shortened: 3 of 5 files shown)


simonrosenberg:
Closing — consolidated into #14227 per maintainer request. The live-status dispatch refactor lands as part of the single ACP feature PR.

PR #14372, opened 2026-05-11
chore(i18n): drop <cmd> XML tags from action/observation messages

simonrosenberg:

---

## Why split this out

Extracted from [#14227](https://github.com/OpenHands/OpenHands/pull/14227) to keep that PR focused. The action/observation translation keys carry inline ``<cmd>...</cmd>`` XML wrappers around their interpolated values. No frontend code parses those tags — verified by grepping ``<cmd>`` across the non-test source. They render as literal text or get treated as unknown HTML by the browser.

This is a pure chore: dropping dead markup. Stripping the tags is mechanical and independent of any feature work.

## What changes

- ``frontend/src/i18n/translation.json``: ``<cmd>{{x}}</cmd>`` → ``{{x}}`` across ~11 action/observation keys × 15 languages (≈165 line changes).
- ``frontend/src/components/v1/chat/event-content-helpers/get-acp-tool-call-content.ts``: comment-only — drop the description of an ``"ACP · "`` prefix that's no longer applied.
- ``frontend/src/components/v1/chat/event-content-helpers/get-event-content.tsx``: comment-only — same as above.

No code path change.

## Verification

- JSON validity: ``python3 -m json.tool translation.json`` ✓
- ``<cmd>`` grep in non-test source: 0 hits (so no parser depends on it).

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [x] Docs / chore


simonrosenberg:
Closing as wrong. Reviewer ([@all-hands-bot](https://github.com/all-hands-bot)) is correct on both counts:

1. **`<cmd>` tags are NOT dead markup.** They're parsed by react-i18next's ``Trans`` component via the ``components`` map binding ``cmd: <MonoComponent />`` — bound at 5+ call sites (``get-event-content.tsx:42``, ``expandable-message.tsx:143``, ``model-messages.tsx:78``, etc.). ``MonoComponent`` renders ``<strong className="font-mono">``, so removing the tags strips bold-monospace styling from every command/pattern in action and observation messages. My grep for the literal string ``<cmd>`` in source missed this because ``Trans`` uses runtime element-to-component mapping, not static references.

2. **The comment "cleanups" were also wrong.** ``ACTION_MESSAGE$ACP_RUN`` on main is ``"ACP · Running <cmd>{{title}}</cmd>"`` — both the ``"ACP · "`` prefix and the ``<cmd>`` tags are intentional. The "Mirrors ``ACTION_MESSAGE$RUN`` etc. but with an ``ACP · `` prefix" docstring is an accurate description of reality; removing it makes the comment less correct.

No follow-up needed on this PR's content. The parent feature work in [#14227](https://github.com/OpenHands/OpenHands/pull/14227) doesn't touch any of these strings or comments after the split — verified just now.

PR #14381, opened 2026-05-11
Add X-Org-Id header validation and effective-org resolution

chuckbutkus:
Introduces a single source of truth for the request's effective organization id and wires it into all routes/services that previously reached for `user.current_org_id` directly.

Resolution precedence (in SaasUserAuth.get_effective_org_id):
  1. api_key_org_id (when authenticated via Bearer API key)
  2. X-Org-Id request header (validated against OrgMember; 403 if the authenticated user is not a member; 400 if not a UUID; 403 if it conflicts with an api_key_org_id binding)
  3. user.current_org_id (legacy fallback)

This closes the api-key-vs-current-org-id divergence: API keys are pinned to the org they were minted under, so callers using a key always operate on that org regardless of the user's later org switch.

Rewired endpoints (now org-id-aware via EFFECTIVE_ORG_ID dep):
- GET /api/v1/users/me (org context block)
- /api/keys: GET, POST, DELETE /{id}, GET /llm/byor[/permitted], POST /llm/byor/refresh
- /api/organizations/llm, /api/organizations/app (GET, PATCH)
- /api/billing: /credits, /has-payment-method, /create-customer-setup-session, /create-checkout-session

Also routes the effective org through:
- require_permission (used by every {org_id}-pathed endpoint)
- SaaS conversation-info and pending-message injectors
- MCP_API_KEY lookup/creation in SaasUserAuth

Underlying helpers gained an optional org_id parameter (defaulting to the legacy user.current_org_id lookup for backward compatibility): ApiKeyStore.{create_api_key,list_api_keys,retrieve_mcp_api_key}, OrgService.check_byor_export_enabled, stripe_service helpers, and the BYOR helpers in routes/api_keys.

Tests: enterprise/tests/unit/server/auth/test_saas_user_auth_effective_org.py covers all precedence rules and the 400/403 paths.

---

## Issue Number
https://linear.app/all-hands-ai/issue/APP-1264/feature-add-processing-of-org-id-header-when-calling-any-endpoint

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-44cc78d   docker.openhands.dev/openhands/openhands:44cc78d
```


chuckbutkus:
@openhands fix enterprise unit tests

chuckbutkus:
@chuckbutkus pushed bcf07b885 to fix the failing enterprise tests and the lint job.

**Test fixes** (route handlers and helpers gained an `effective_org_id` parameter; tests now thread it through explicitly):
- `enterprise/tests/unit/test_billing.py` – pass `effective_org_id` to `create_checkout_session` / `create_customer_setup_session` / `has_payment_method`; assert `org_id=` kwarg on `has_payment_method_by_user_id`.
- `enterprise/tests/unit/server/routes/test_api_keys.py` – pass `effective_org_id=` to `get_llm_api_key_for_byor` and `check_byor_permitted`; update assertions on `check_byor_export_enabled`, `get_byor_key_from_db`, `store_byor_key_in_db`, `generate_byor_key`, and `delete_byor_key_from_litellm` for the new `org_id` parameter. The `TestDeleteByorKeyFromLitellm` "no user / no org_id" branches no longer exist (alias is always built from the effective org id), so they collapse into `test_delete_constructs_alias_from_org`.
- `enterprise/tests/unit/server/services/test_org_app_settings_service.py` – set `mock_user_context.user_auth = None` so the service falls back to `get_current_org_by_user_id`. Without this, `MagicMock` auto-creates `user_auth.get_effective_org_id`, which is not awaitable.
- `enterprise/tests/unit/test_saas_user_auth.py::test_get_openhands_api_key_raises_for_missing_user` – expect the collapsed "has no current organization" message since `_get_openhands_api_key` now resolves via `get_effective_org_id()`.
- `enterprise/tests/unit/storage/test_saas_sql_app_conversation_info_service.py` – give the three local `MockUserAuth` classes a `get_effective_org_id` that returns `api_key_org_id`, so the injector keeps picking the API-key org over `user.current_org_id`.

**Lint fixes**:
- `enterprise/storage/org_service.py` and `enterprise/server/utils/saas_pending_message_injector.py` – reorder `openhands.*` imports into the third local block to satisfy ruff `I` (the same diff ruff produced in CI).

Ran the full suite locally with `--forked -n auto`: **2006 passed, 3 skipped**, and ran pre-commit on the touched files (ruff/ruff-format/mypy all clean).

_This comment was created by an AI agent (OpenHands) on behalf of @chuckbutkus._

enterprise/integrations/stripe_service.py
```diff
@@ -34,21 +34,48 @@ async def find_customer_id_by_org_id(org_id: UUID) -> str | None:
...
+    user's current org. Returns ``None`` if neither resolves.
+    """
+    if org_id is not None:
+        org = await OrgStore.get_org_by_id(org_id)
+        if not org:
+            logger.warning(
+                'stripe_org_not_found_for_id',
+                extra={'user_id': user_id, 'org_id': str(org_id)},
+            )
+        return org
```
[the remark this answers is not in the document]
chuckbutkus: This is the same behavior as before (returned None if it couldn't get the current org ID).  Also, the FK constraints in the DB should prevent an org_member without an org.

enterprise/server/auth/saas_user_auth.py
```diff
@@ -86,6 +93,105 @@ def get_api_key_org_id(self) -> UUID | None:
...
+
+        Raises:
+            HTTPException: 400 for a malformed header, 403 for
+                membership / API-key conflicts.
+        """
+        if self._effective_org_id_resolved:
+            return self._effective_org_id
+
+        # Import locally to avoid a circular import via authorization.py.
+        from fastapi import HTTPException, status
```
hieptl: **suggestion (blocking):** It appears that the HTTPException is imported twice in saas_user_auth.py. It is already imported at the top of the module and is then re-imported inside the get_effective_org_id function. To avoid redundancy, please remove the local re-import. If the status is required locally, feel free to keep that, but HTTPException is already in scope.

Thank you! 🙏

enterprise/server/routes/orgs.py
```diff
@@ -222,7 +223,11 @@ async def create_org(
         )
 
 
-@org_router.get('/{org_id}/settings', response_model=OrgDefaultsSettingsResponse)
+@org_router.get(
+    '/{org_id}/settings',
+    response_model=OrgDefaultsSettingsResponse,
+    dependencies=[REJECT_X_ORG_ID_PATH_MISMATCH],
```
hieptl: **suggestion (blocking):**

`REJECT_X_ORG_ID_PATH_MISMATCH` is currently opt-in per route, and it has been manually attached to approximately 13 routes in `orgs.py`. Any future routes with `{org_id}` in their path on the `org_router` may silently regress. On the other hand, `org_invitations.py` correctly attaches it at the router level. I recommend either applying the same approach on the `org_router` or splitting the `org_router` so that the path-org routes are handled by a sub-router with the dependency applied at the prefix level. At the very least, please add a comment on the `org_router` to clarify the convention.

Thank you! 🙏

enterprise/server/auth/saas_user_auth.py
chuckbutkus: Fixed

enterprise/server/routes/orgs.py
chuckbutkus: Updated

enterprise/server/auth/authorization.py
```diff
@@ -297,6 +297,18 @@ async def permission_checker(
...
+        # resolve the effective org for the request — which honors any
+        # ``X-Org-Id`` header override (and validates membership /
+        # API-key binding in the process). This keeps endpoints that
+        # implicitly operate on the "current org" consistent with the
+        # rest of the codebase.
+        if org_id is None:
+            # Local import to avoid circular import via saas_user_auth.
+            from server.auth.org_context import maybe_resolve_effective_org_id
+
+            org_id = await maybe_resolve_effective_org_id(request)
```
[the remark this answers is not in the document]
chuckbutkus: Reviewed all the endpoints and things appear to be working as they should.

hieptl (approved):
Thank you! 🙏

PR #14387, opened 2026-05-12
Fix so offline token is not deleted.

chuckbutkus:

---

## Why
API keys (for new users) are no longer able to be used after the user logs out of SAAS.

It looks like this PR https://github.com/OpenHands/OpenHands/pull/13675 inadvertently caused the offline keycloak token to get deleted for new users.  This PR set the auth cookie token to the offline tokens (only for new users, user's that had their offline token created prior to this PR would not see the issue).  When the user logs out, it logs out of keycloak with the offline token, killing the offline session.

## Summary

- Removed setting the cookie on the redirect from the keycloak offline callback.

## How to Test

- Start with a new user.
- Login to OH
- Create an API key.
- Logout of OH
- Try to use the API key to make an API call.  Before this fix it will fail, after it will succeed.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-3ad40f1   docker.openhands.dev/openhands/openhands:3ad40f1
```

### DIFF
--- a/enterprise/server/routes/auth.py
+++ b/enterprise/server/routes/auth.py
@@ -645,8 +645,6 @@ async def keycloak_offline_callback(code: str, state: str, request: Request):
     )
 
     user = await UserStore.get_user_by_id(user_info.sub)
-    has_accepted_tos = user is not None and user.accepted_tos is not None
-
     redirect_url, _, _ = _extract_oauth_state(state)
     default_url = redirect_url if redirect_url else web_url
     final_url = await _get_post_auth_redirect(user_info.sub, default_url, web_url, user)
--- a/enterprise/server/routes/auth.py
+++ b/enterprise/server/routes/auth.py
@@ -651,16 +651,11 @@ async def keycloak_offline_callback(code: str, state: str, request: Request):
     default_url = redirect_url if redirect_url else web_url
     final_url = await _get_post_auth_redirect(user_info.sub, default_url, web_url, user)
 
-    response = RedirectResponse(final_url, status_code=302)
-    set_response_cookie(
-        request=request,
-        response=response,
-        keycloak_access_token=keycloak_access_token,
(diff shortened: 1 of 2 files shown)


rbren (approved):

chuckbutkus:
@openhands add tests as described below: 
This critical authentication change lacks automated tests. Add a test in enterprise/tests/unit/server/routes/ that verifies:
New user performs offline token authentication
User logs out (online session ends)
Offline token (API key) still works for API calls

chuckbutkus:
Added automated tests in `enterprise/tests/unit/server/routes/test_auth_offline_token.py` (commit 3ad40f107) that cover this regression:

- **`test_offline_callback_does_not_set_keycloak_auth_cookie`** – asserts that the keycloak offline callback persists the offline token via `token_manager.store_offline_token` but does **not** call `set_response_cookie`, and that the redirect response contains no `Set-Cookie: keycloak_auth` header.
- **`test_offline_token_survives_user_logout`** – mirrors the manual repro from the PR description:
  1. New user completes the keycloak offline-token authentication flow.
  2. The online `/logout` endpoint is invoked and we assert it only ever passes the *online* refresh token to keycloak (never the offline token).
  3. The offline token used by API keys is still validatable after logout.

Both tests pass locally:

```
PYTHONPATH=".:enterprise:$PYTHONPATH" poetry run --project=enterprise pytest \
  enterprise/tests/unit/server/routes/test_auth_offline_token.py -v
# ======================== 2 passed ========================
```

_This comment was created by an AI agent (OpenHands) on behalf of @chuckbutkus._

PR #14388, opened 2026-05-12
Plumb effective_org_id through SaasSettingsStore and SaasSecretsStore

chuckbutkus:
Follow-up to openhands/x-org-id-header-validation. The first PR rewired the route layer to honor the request's effective org (api_key_org_id > X-Org-Id header > user.current_org_id), but the SaaS settings and secrets stores still resolved the org from `user.current_org_id` internally. That meant /api/v1/settings/* and /api/v1/secrets/* endpoints read and wrote against the user's last-selected org, even when the request was authenticated via an API key bound to a different org. Same bug class the parent PR set out to close.

Changes:
- SaasSettingsStore gains an `effective_org_id: UUID | None = None` field. `load()` and `store()` resolve the org via a new `_resolve_org_id(user)` helper that prefers effective_org_id and falls back to `user.current_org_id` when unset. The existing org_members membership check is preserved as defense-in-depth.
- SaasSecretsStore gains the same field, used in both load and store WHERE/INSERT clauses (secrets are stored per (user_id, org_id)).
- Both stores' `get_instance(user_id, effective_org_id=None)` take an optional org argument; the no-arg form is preserved for webhook callers (github_view, gitlab_view) that don't carry an org context.
- SaasUserAuth.get_user_settings_store() and get_secrets_store() now resolve `await self.get_effective_org_id()` and pass it through. The cached store handle is reused across calls within a request.

The cached settings/secrets stores live on the user-auth instance (which is per-request), so the org binding is captured for the lifetime of the request without re-resolving on each access.

Affected endpoints (now correctly org-scoped via the effective org):
  Settings: GET/POST /api/v1/settings,
            GET /api/v1/settings/agent-schema, /conversation-schema,
            GET/POST/DELETE /api/v1/settings/profiles,
            POST /api/v1/settings/profiles/{name}/activate, /rename
  Secrets:  GET /api/v1/secrets/search,
            POST/PUT/DELETE /api/v1/secrets[/{secret_id}],
            POST/DELETE /api/v1/secrets/git-providers

Tests:
  enterprise/tests/unit/storage/test_saas_stores_effective_org.py
  - field default is None (backward compat)
  - _resolve_org_id prefers effective over current
  - load() refuses unknown effective org (not a fallback to current)
  - load() compiles the right org id into the SQL filter
  - store() writes the StoredCustomSecrets row under the right org
  - get_instance variants (with and without effective_org_id)
  - SaasUserAuth wiring: both store getters call get_effective_org_id and propagate the resolved value to the store constructor

Not tested in this env: full pytest run requires Python 3.12 (this sandbox is 3.13, where `aifc` is removed; the openhands logger still imports it for speech_recognition compat). Static checks (py_compile, ruff check, ruff format) all pass.

---

## Summary

-

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-d80b1b9   docker.openhands.dev/openhands/openhands:d80b1b9
```

### DIFF
--- a/enterprise/server/auth/saas_user_auth.py
+++ b/enterprise/server/auth/saas_user_auth.py
@@ -258,7 +258,15 @@ class SaasUserAuth(UserAuth):
         secrets_store = self.secrets_store
         if secrets_store:
             return secrets_store
-        secrets_store = await SaasSecretsStore.get_instance(self.user_id)
+        # Scope secrets to the request's effective org so that callers
+        # using an API key bound to org A — or supplying an X-Org-Id
+        # header — read/write secrets under that org, not under whatever
+        # `user.current_org_id` happens to point at.
+        effective_org_id = await self.get_effective_org_id()
+        secrets_store = await SaasSecretsStore.get_instance(
+            self.user_id,
+            effective_org_id=effective_org_id,
+        )
         self.secrets_store = secrets_store
         return secrets_store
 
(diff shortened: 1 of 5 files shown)


chuckbutkus:
@openhands fix lint and test errors

PR #14407, opened 2026-05-13
[PLTF-2500] Add Bitbucket DC webhook enrollment UI for manual enrollment

ak684:
## Summary

Adds a Bitbucket Data Center webhook enrollment path for OHE so members with the MANAGE_INTEGRATIONS permission can create the OpenHands enrollment state that the production webhook receive path requires.

- adds BBDC repository resource/list enrollment APIs
- stores a repo-level webhook secret, installer identity, and sync timestamp
- adds an OHE Git settings UI for guided manual webhook creation
- exposes BBDC provider configuration to the frontend
- includes existing direct LLM default-routing changes currently present in this test branch

Screenshot 2026-05-14 at 6 47 09 AM

## Testing

- PYTHONPATH=enterprise:. uv run --with python-keycloak --with pytest-asyncio --with aiosqlite --with limits --with coredis pytest enterprise/tests/unit/storage/test_bitbucket_dc_webhook_store.py enterprise/tests/unit/integrations/bitbucket_data_center/test_bitbucket_dc_webhook.py tests/unit/app_server/test_default_web_client_config_injector.py enterprise/tests/unit/test_lite_llm_manager.py
- npm run typecheck:staged
- npm run test -- __tests__/components/features/settings/bitbucket-dc-webhook-manager.test.tsx
- pre-commit hook passed during commit
- Tested end to end with my personal instance of OHE Replicated installer and a bitbucket self-hosted instance

Screenshot 2026-05-18 at 9 58 57 AM
Screenshot 2026-05-18 at 9 59 49 AM

## OHE Validation

Tested this PR on a KOTS-managed OHE instance using enterprise-server image ghcr.io/openhands/enterprise-server:sha-a22f0de.

- Confirmed Bitbucket Data Center login worked and the BBDC webhook settings UI rendered in OHE.
- Clicked Enroll for AIP/repo-1 and confirmed the UI displayed the webhook name, URL, secret, and events.
- Queried the OHE database and confirmed a bitbucket_dc_webhook row was created for AIP/repo-1 with installer user id, webhook secret, and last_synced set.
- Sent a synthetic BBDC pr:comment:added webhook signed with the DB secret to /integration/bitbucket-dc/events.
- Confirmed a bad signature was rejected with 403.
- Confirmed the correctly signed request returned 200 and OHE logs showed the event entered the BBDC resolver manager path.
- Tested end to end with my personal instance of OHE Replicated installer and a bitbucket self-hosted instance

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-6804d6c   docker.openhands.dev/openhands/openhands:6804d6c
```

---
Linear: PLTF-2500

_AI-generated metadata update by OpenHands on behalf of ak684._


frontend/src/i18n/translation.json
```diff
@@ -5065,6 +5065,482 @@
...
+    "ar": "No Bitbucket Data Center repositories were found for this account.",
+    "fr": "No Bitbucket Data Center repositories were found for this account.",
+    "tr": "No Bitbucket Data Center repositories were found for this account.",
+    "de": "No Bitbucket Data Center repositories were found for this account.",
+    "uk": "No Bitbucket Data Center repositories were found for this account.",
+    "ca": "No Bitbucket Data Center repositories were found for this account."
+  },
+  "BITBUCKET_DATA_CENTER$WEBHOOK_MANAGER_DESCRIPTION": {
+    "en": "Enroll repository webhooks in OpenHands, then create the matching webhook in Bitbucket Data Center. Enroll shared repos with an OpenHands service account if responses should come from a bot.",
+    "ja": "Enroll repository webhooks in OpenHands, then create the matching webhook in Bitbucket Data Center. Enroll shared repos with an OpenHands service account if responses should come from a bot.",
```
hieptl: **suggestion (blocking):** We need to support multiple languages. Thank you! 🙏

enterprise/server/routes/integration/bitbucket_dc.py
```diff
@@ -89,6 +149,158 @@ async def verify_bitbucket_dc_signature(
...
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail='Failed to retrieve Bitbucket DC resources',
+        )
+
+
+@bitbucket_dc_integration_router.post('/bitbucket-dc/enroll-webhook')
+async def enroll_bitbucket_dc_webhook(
+    body: EnrollBitbucketDCWebhookRequest,
+    user_id: str = Depends(get_user_id),
```
hieptl: **suggestion (blocking):** All three new endpoints only require Depends(get_user_id), meaning any logged-in user can access them. The PR description mentions 'admins can create,' but there’s currently no admin or role-based restriction. Thank you! 🙏

enterprise/storage/bitbucket_dc_webhook_store.py
```diff
@@ -44,6 +45,103 @@ async def get_webhook_user_id(self, project_key: str, repo_slug: str) -> str | N
...
+                    )
+                    .limit(1)
+                )
+                result = await session.execute(query)
+                webhook = result.scalars().first()
+
+                if webhook:
+                    webhook.user_id = user_id
+                    webhook.webhook_secret = webhook_secret
+                    webhook.last_synced = datetime.utcnow()
```
hieptl: **suggestion (blocking):** It seems that datetime.utcnow() is deprecated in Python 3.12+, so using datetime.now(timezone.utc) would be preferable. Additionally, there’s an inconsistency: update_webhook_id relies on raw SQL CURRENT_TIMESTAMP, while upsert_webhook_enrollment uses Python time. It might be best to standardize on one approach. Thank you! 🙏

hieptl (approved):

frontend/src/i18n/translation.json
ak684: Thanks! Fixed in 6804d6c2e. All 28 newly-added `BITBUCKET_DATA_CENTER$*` keys now have real translations across the 14 non-English locales (ja, zh-CN, zh-TW, ko-KR, no, it, pt, es, ar, fr, tr, de, uk, ca), replacing the English placeholders. "Bitbucket Data Center" is kept verbatim as a product proper noun, and the `{{userId}}` interpolation token is preserved in the `WEBHOOK_ENROLLED_BY` string. Open to native-speaker corrections on any specific string.

enterprise/server/routes/integration/bitbucket_dc.py
ak684: Good catch. Fixed in 6804d6c2e. The three new endpoints now use `Depends(require_permission(Permission.MANAGE_INTEGRATIONS))` instead of bare `Depends(get_user_id)`, wiring them into the existing permission framework (validates org context, API-key org binding, and role membership) the same way the `orgs.py` integration routes do.

One caveat I documented in the updated PR description: `MANAGE_INTEGRATIONS` is held by OWNER, ADMIN, *and* MEMBER today, so this doesn't actually narrow access to admins specifically. It matches how other integration-management endpoints are gated, but the original PR summary line overpromised. I updated the summary to drop the "admins can create" wording. A repo-wide tightening of integration permissions (or adding a dedicated admin-only `MANAGE_WEBHOOK_INTEGRATIONS` permission) feels like its own focused change. Happy to do that as a follow-up PR if you'd prefer admin-only here. Let me know which you'd rather.

enterprise/storage/bitbucket_dc_webhook_store.py
ak684: Good catch. Fixed in 6804d6c2e. `upsert_webhook_enrollment` now uses `datetime.now(timezone.utc)`, and `update_webhook_id` was switched from raw SQL `text('CURRENT_TIMESTAMP')` to the same `datetime.now(timezone.utc)` so both code paths use Python-side timestamps consistently. Also dropped the now-unused `text` import from sqlalchemy.

PR #14420, opened 2026-05-14
fix(api): add title field to AppConversationUpdateRequest

jpshackelford:
## Summary

The frontend sends a `title` field when renaming conversations via `PATCH /api/v1/app-conversations/{id}`, but the backend `AppConversationUpdateRequest` model was missing this field. This caused title updates to be **silently ignored** by Pydantic since extra fields are dropped when not defined in the model.

## Problem

When users rename conversations via the UI (or API), the request looks like:
```json
PATCH /api/v1/app-conversations/{id}
{"title": "My New Title"}
```

But `AppConversationUpdateRequest` only had these fields:
- `public`
- `selected_repository`
- `selected_branch`
- `git_provider`

Since `title` wasn't in the model, Pydantic would silently ignore it. The service layer iterates over `model_fields_set` to apply updates, so the title field was never recognized.

## Solution

Add `title: str | None = None` to `AppConversationUpdateRequest`. The service layer already uses a generic loop over `model_fields_set`, so this one-line change enables conversation renaming via the API.

## Testing

Added two tests to verify:
1. `test_app_conversation_update_request_includes_title_field` - Verifies the title field is recognized in `model_fields_set`
2. `test_app_conversation_update_request_title_field_updates_conversation_info` - Simulates the service layer logic to verify title updates are properly applied

Both tests **fail before the fix** and **pass after**.

---
*This PR was created by an AI agent (OpenHands) on behalf of the user.*

### DIFF
--- a/openhands/app_server/app_conversation/app_conversation_models.py
+++ b/openhands/app_server/app_conversation/app_conversation_models.py
@@ -214,6 +214,7 @@ class AppConversationUpdateRequest(BaseModel):
     All fields are optional - only provided fields will be updated.
     """
 
+    title: str | None = None
     public: bool | None = None
     selected_repository: str | None = None
     selected_branch: str | None = None
--- a/tests/unit/app_server/test_models.py
+++ b/tests/unit/app_server/test_models.py
@@ -4,7 +4,9 @@ from uuid import UUID, uuid4
 import pytest
 
 from openhands.app_server.app_conversation.app_conversation_models import (
+    AppConversationInfo,
     AppConversationStartRequest,
+    AppConversationUpdateRequest,
 )
 from openhands.app_server.event_callback.event_callback_models import (
     EventCallback,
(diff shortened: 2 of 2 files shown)


jpshackelford:
## ✅ CI Passing & Evidence the Fix Works

All CI checks are now passing.

### Demonstration of the Fix

The issue was that Pydantic silently ignores fields not defined in the model. Here's a demonstration:

```
============================================================
DEMONSTRATION: Before vs After the fix
============================================================

--- BEFORE FIX (title field missing from model) ---
Request model_fields_set: set()
Has 'title' in model_fields_set? False
>>> Title update would be SILENTLY IGNORED!

--- AFTER FIX (title field present in model) ---
Request model_fields_set: {'title'}
Has 'title' in model_fields_set? True
Title value: My New Title
>>> Title update will be APPLIED correctly!

============================================================
The service layer iterates over model_fields_set to apply updates.
With the fix, 'title' is now recognized and will be updated.
============================================================
```

### How the Service Layer Works

The `update_app_conversation` method in `live_status_app_conversation_service.py` uses this pattern:

```python
for field_name in request.model_fields_set:
    value = getattr(request, field_name)
    setattr(info, field_name, value)
```

**Before the fix:** When the frontend sent `{"title": "New Title"}`, Pydantic would silently drop the `title` field since it wasn't defined in `AppConversationUpdateRequest`. The `model_fields_set` would be empty, and no update would occur.

**After the fix:** The `title` field is now in the model, so it appears in `model_fields_set` and gets applied to the conversation info.

### Tests Added

Two tests were added that explicitly verify this behavior:
1. `test_app_conversation_update_request_includes_title_field` - Ensures title is in `model_fields_set`
2. `test_app_conversation_update_request_title_field_updates_conversation_info` - Simulates the service layer update logic

Both tests **fail before the fix** and **pass after**, as demonstrated in CI.

---
*This comment was created by an AI agent (OpenHands) on behalf of the user.*

tofarr (approved):
🍰

PR #14430, opened 2026-05-15
fix: return 404 for missing custom secrets

### LINKED ISSUE #14204: Bug: `update_custom_secret` returns 200 when secret doesn't exist (`return` instead of `raise` HTTPException)
### Description

In `openhands/app_server/secrets/secrets_router.py`, the `update_custom_secret` endpoint (L276-327) has two bugs that cause it to silently report success when it should fail.

### Bug 1: `return` instead of `raise` HTTPException

Line 296 uses `return HTTPException(...)` instead of `raise HTTPException(...)`:

```python
# L295-299
if secret_id not in existing_secrets.custom_secrets:
    return HTTPException(                          # ← should be "raise"
        status_code=status.HTTP_404_NOT_FOUND,
        detail=f'Secret with ID {secret_id} not found',
    )
```

`return HTTPException(...)` sends a **200 OK** response with the serialized `HTTPException` object as JSON body — something like:

```json
{"status_code": 404, "detail": "Secret with ID xyz not found"}
```

The client sees HTTP 200 and believes the update succeeded. The same pattern is correctly implemented as `raise` in the adjacent `delete_custom_secret` endpoint (L351).

### Bug 2: `existing_secrets is None` falls through to success

When the secrets store is empty (`existing_secrets` is falsy), the `if existing_secrets:` guard (L293) skips the entire update logic and falls through to:

```python
# L325-327
return EditResponse(
    message='Secret updated successfully',
)
```

The client is told the secret was updated, but nothing was written. The same issue exists in `delete_custom_secret` (L345) — deleting from an empty store returns "Secret deleted successfully".

### Expected Behavior

- Updating a nonexistent secret should return **404**
- Updating when no secrets exist at all should return **404** (not 200 with a success message)

### Suggested Fix

```python
existing_secrets = await secrets_store.load()

if not existing_secrets or secret_id not in existing_secrets.custom_secrets:
    raise HTTPException(
        status_code=status.HTTP_404_NOT_FOUND,
        detail=f'Secret with ID {secret_id} not found',
    )
```

Same pattern for `delete_custom_secret`.

Happy to submit a PR if this looks right.

Kage18:
## Summary

Fixes #14204.

- return 404 when updating a custom secret and the secrets store is empty
- return 404 when deleting a custom secret and the secrets store is empty
- add regression coverage for both empty-store paths

## Tests

- uv run pytest tests/unit/app_server/test_secrets_api.py -q
- uv run ruff check openhands/app_server/secrets/secrets_router.py tests/unit/app_server/test_secrets_api.py

Note: I also attempted the repo-mandated `make install-pre-commit-hooks`. It initially failed because the system Python is 3.9; after creating a Python 3.12 venv, Poetry failed while uninstalling `matplotlib` due a missing embedded pip wheel path in the local venv. The focused pytest and Ruff checks above passed.

### DIFF
--- a/openhands/app_server/secrets/secrets_router.py
+++ b/openhands/app_server/secrets/secrets_router.py
@@ -308,37 +308,35 @@ async def update_custom_secret(
         500: Error updating secret
     """
     existing_secrets = await secrets_store.load()
-    if existing_secrets:
-        # Check if the secret to update exists
-        if secret_id not in existing_secrets.custom_secrets:
-            raise HTTPException(
-                status_code=status.HTTP_404_NOT_FOUND,
-                detail=f'Secret with ID {secret_id} not found',
-            )
-
-        secret_name = incoming_secret.name
-        secret_description = incoming_secret.description
-
-        custom_secrets = dict(existing_secrets.custom_secrets)
-        existing_secret = custom_secrets.pop(secret_id)
-
-        if secret_name != secret_id and secret_name in custom_secrets:
-            raise HTTPException(
-                status_code=status.HTTP_400_BAD_REQUEST,
-                detail=f'Secret {secret_name} already exists',
-            )
-
-        custom_secrets[secret_name] = CustomSecret(
-            secret=existing_secret.secret,
-            description=secret_description or '',
+    if not existing_secrets or secret_id not in existing_secrets.custom_secrets:
+        raise HTTPException(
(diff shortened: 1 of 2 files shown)


Kage18:
Ready for review. Focused fix with regression tests for the empty secrets-store update/delete paths.

PR #14432, opened 2026-05-15
fix: only mark language dirty on selection

### LINKED ISSUE #14249: [Bug] language selector dirty-flag incorrectly set while typing (onInputChange vs onSelectionChange)

Kage18:
## Summary

Fixes #14249.

- use language autocomplete selection events instead of input typing events for dirty-state checks
- compare selected language keys directly against the saved settings language
- add regression coverage so filtering the language dropdown does not enable Save

## Tests

- npm run test -- __tests__/routes/app-settings.test.tsx
- npm run lint:fix
- npm run build

### DIFF
--- a/frontend/__tests__/routes/app-settings.test.tsx
+++ b/frontend/__tests__/routes/app-settings.test.tsx
@@ -145,6 +145,10 @@ describe("Form submission", () => {
     // Language check
     const language = await screen.findByTestId("language-input");
     await userEvent.click(language);
+    await userEvent.type(language, "Nor");
+    expect(submit).toBeDisabled();
+
+    await userEvent.click(language);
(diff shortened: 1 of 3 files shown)


Kage18:
Ready for review. This keeps the language filter text from marking settings dirty and adds regression coverage for the dropdown behavior.

VascoSch92:
Hey @Kage18 

thanks for the PR. 

Could you provide a video to show that it is working?

neubig:

Thank you for understanding.

PR #14441, opened 2026-05-16
fix: infer docker sandbox host port from web url

Kage18:
## Summary
- Infer Docker sandbox `host_port` from `OH_WEB_URL` when `SANDBOX_HOST_PORT` is not set
- Keep explicit `SANDBOX_HOST_PORT` as the override
- Add regression coverage for non-default host port Docker deployments

Fixes OpenHands/software-agent-sdk#4245.

## Why
When OpenHands is exposed with `-p 4000:3000` and `OH_WEB_URL=http://host:4000`, agent-server webhook callbacks still defaulted to `http://host.docker.internal:3000/api/v1/webhooks` unless users also set `SANDBOX_HOST_PORT`. That sends callbacks to the wrong host port and can leave conversations stuck after webhook retries fail.

## Tests
- `poetry run pytest tests/unit/app_server/test_docker_sandbox_service.py -q`
- `poetry run ruff check --config dev_config/python/ruff.toml openhands/app_server/config.py tests/unit/app_server/test_docker_sandbox_service.py`
- `poetry run ruff format --config dev_config/python/ruff.toml --check openhands/app_server/config.py tests/unit/app_server/test_docker_sandbox_service.py`

### DIFF
--- a/openhands/app_server/config.py
+++ b/openhands/app_server/config.py
@@ -3,6 +3,7 @@
 import os
 from pathlib import Path
 from typing import AsyncContextManager
+from urllib.parse import urlparse
 
 import httpx
 from fastapi import Depends, Request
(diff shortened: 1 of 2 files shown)


Kage18:
Ready for review. Focused Docker self-hosting fix with regression coverage for deriving webhook callback host_port from OH_WEB_URL when SANDBOX_HOST_PORT is not explicitly set.

PR #14442, opened 2026-05-16
fix: title API conversations from initial message

### LINKED ISSUE #13125: [Bug]: API-triggered conversations are not given proper titles visible in the interface
## Bug Description

Conversations triggered through the OpenHands Cloud API (using bearer token authentication) are not given proper auto-generated titles that are visible in the interface. They retain their default titles (e.g., "Conversation abc12") instead of receiving meaningful titles based on the first user message.

## Steps to Reproduce

1. Create a conversation using the OpenHands Cloud API:
```bash
curl -X POST "https://app.all-hands.dev/api/v1/app-conversations" \
  -H "Authorization: Bearer YOUR_API_KEY" \
  -d '{"initial_message": {"content": [{"type": "text", "text": "Fix the authentication bug in login.py"}]}, "selected_repository": "owner/repo"}'
```

2. View the conversation in the OpenHands web interface
3. Observe that the conversation title remains as "Conversation {id[:5]}" instead of being auto-generated from the first message

## Expected Behavior

The conversation should have an auto-generated title based on the first user message, similar to GUI-initiated conversations.

### V0 Code Path (Legacy - `/api/conversations`)

In `openhands/server/services/conversation_service.py`:
- `initialize_conversation()` sets `title = get_default_conversation_title(conversation_id)` = "Conversation {id[:5]}"
- In `openhands/server/conversation_manager/standalone_conversation_manager.py`, `auto_generate_title()` is called when events are received, but the title update is primarily emitted via SocketIO:
```python
await run_in_loop(
    self.sio.emit(
        'oh_event',
        status_update_dict,
        to=ROOM_KEY.format(sid=conversation_id),
    ),
    self._loop,
)
```
- **Problem**: API users don't have an active WebSocket connection, so the SocketIO event is never received by anyone. The title change may not be persisted to the conversation metadata store consistently.

### V1 Code Path (`/api/v1/app-conversations`)

In `openhands/app_server/app_conversation/live_status_app_conversation_service.py`:
- Conversations are created with default title: `title=f'Conversation {info.id.hex[:5]}'`
- A `SetTitleCallbackProcessor` is registered as an `EventCallback` (lines 310-316)
- The callback is triggered when a `MessageEvent` arrives

In `openhands/app_server/event_callback/set_title_callback_processor.py`:
- The processor makes an HTTP POST to `{app_conversation_url}/generate_title` on the agent server
- Then saves the generated title to `AppConversationInfo`

**Problem**: PR #13093 indicates that the `/generate_title` endpoint approach is being deprecated in favor of polling. This suggests the current title generation mechanism may not be working correctly.

## Related Work

- **PR #13093**: "fix: stop calling agent-server /generate_title" - This PR is attempting to fix the title generation by polling instead of calling a dedicated endpoint. This confirms that the current mechanism has issues.

## Potential Solutions

1. **For V1 (Recommended)**: Ensure PR #13093 is completed and properly handles API-triggered conversations. The polling approach should work regardless of WebSocket connection status.

2. **For V0 (if still supported)**: In addition to emitting via SocketIO, ensure the title is persisted to the conversation metadata store in `auto_generate_title()`:
```python
# After generating title, save it directly
conversation.title = title
await conversation_store.save_metadata(conversation)
```

3. **Alternative for both paths**: Consider generating the title synchronously (or with a short delay) during conversation creation rather than relying on asynchronous callbacks, especially for API-triggered conversations where there's no UI to receive real-time updates.

## Environment

- OpenHands Cloud (app.all-hands.dev)
- Both V0 (`/api/conversations`) and V1 (`/api/v1/app-conversations`) endpoints affected
- `ConversationTrigger.REMOTE_API_KEY` conversations

## Labels

bug, cloud, api

Kage18:
## Summary
- Add a conservative fallback title for app conversations when the agent-server title endpoint is reachable but never returns a title
- Use only user text message events for the fallback title
- Keep the title callback active when polling fails or the event has no user text

Fixes #13125.

## Why
API-triggered conversations can stay visible as `Conversation {id}` when the runtime never produces an auto-generated title. Since the title callback already receives the initial user message event, it can safely fall back to a short title from that message after successful empty title polls.

## Tests
- `poetry run pytest tests/unit/app_server/test_set_title_callback_processor.py -q`
- `poetry run ruff check --config dev_config/python/ruff.toml openhands/app_server/event_callback/set_title_callback_processor.py tests/unit/app_server/test_set_title_callback_processor.py`
- `poetry run ruff format --config dev_config/python/ruff.toml --check openhands/app_server/event_callback/set_title_callback_processor.py tests/unit/app_server/test_set_title_callback_processor.py`

### DIFF
--- a/openhands/app_server/event_callback/set_title_callback_processor.py
+++ b/openhands/app_server/event_callback/set_title_callback_processor.py
@@ -30,13 +30,14 @@ _logger = logging.getLogger(__name__)
 _POLL_DELAY_S = 3
 # Number of attempts to poll title
 _NUM_POLL_ATTEMPTS = 4
+_MAX_FALLBACK_TITLE_CHARS = 60
 
 
 async def _poll_for_title(
     httpx_client: httpx.AsyncClient,
     url: str,
     session_api_key: str | None,
-) -> str | None:
+) -> tuple[str | None, bool]:
     """Poll the agent server for the conversation title.
 
     Args:
@@ -45,8 +46,9 @@ async def _poll_for_title(
         session_api_key: The session API key for authentication.
 
     Returns:
-        The title if available, None otherwise.
+        The title if available, and whether any poll response succeeded.
     """
+    has_successful_response = False
     for _ in range(_NUM_POLL_ATTEMPTS):
         await asyncio.sleep(_POLL_DELAY_S)
         try:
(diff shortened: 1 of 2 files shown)


Kage18:
Ready for review. Focused app-server title callback fix with regression coverage for API-triggered conversations falling back to the first user text message only when title polling succeeds but returns no title.

PR #14451, opened 2026-05-17
fix: _configure_llm now preserves user LLM settings

### LINKED ISSUE #14450: _configure_llm discards user LLM settings, breaking Claude Opus 4.7
## Summary

`_configure_llm()` in `live_status_app_conversation_service.py` creates a brand-new `LLM()` instance with only `model`, `base_url`, `api_key`, and `usage_id`. Every other user-configured field is silently replaced by the SDK Pydantic defaults.

This means user settings like `reasoning_effort`, `extended_thinking_budget`, `drop_params`, `max_output_tokens`, etc. are all thrown away before reaching the agent-server.

## Impact

This breaks **Claude Opus 4.7** on self-hosted OpenHands deployments:

1. The SDK defaults `reasoning_effort` to `"high"` (Pydantic `Field(default="high")`)
2. The bundled litellm (v1.83.0 in the PyInstaller agent-server binary) does not recognise `claude-opus-4-7`
3. litellm maps `reasoning_effort="high"` to `thinking.type=enabled` instead of `thinking.type=adaptive`
4. Anthropic rejects the request with: `"thinking.type.enabled" is not supported for this model`

Users who set `reasoning_effort=None` in their settings to work around this have their override discarded by `_configure_llm()`.

## Steps to reproduce

1. Run OpenHands 1.7.0 self-hosted with `anthropic/claude-opus-4-7`
2. Set `reasoning_effort: null` in settings.json
3. Start a new conversation
4. Observe the error: `litellm.BadRequestError: AnthropicException - thinking.type.enabled is not supported`

## Root cause

```python
# Current code (line 924)
return LLM(
    model=model,
    base_url=base_url,
    api_key=user.agent_settings.llm.api_key,
    usage_id='agent',
)
```

This constructs a fresh `LLM()` from scratch. All fields not explicitly passed get their Pydantic defaults, including `reasoning_effort='high'`.

## Suggested fix

Use `model_copy()` on the user's existing LLM config instead of constructing a new one:

```python
return user.agent_settings.llm.model_copy(
    update={
        'model': model,
        'base_url': base_url,
        'api_key': user.agent_settings.llm.api_key,
        'usage_id': 'agent',
    }
)
```

## Environment

- OpenHands 1.7.0 (self-hosted, Docker)
- Agent-server image: `ghcr.io/openhands/agent-server:1.19.1-python`
- Bundled litellm: ~1.83.0 (frozen in PyInstaller binary)
- Model: `anthropic/claude-opus-4-7`

## Affected versions

Confirmed on 1.7.0. Likely affects all versions since `_configure_llm` was introduced.

shanemort1982:
Fixes #14450

## What

`_configure_llm()` was constructing a brand-new `LLM()` instance from scratch, passing only `model`, `base_url`, `api_key`, and `usage_id`. Every other field from the user's saved LLM configuration was silently replaced by the SDK's Pydantic defaults.

This broke Claude Opus 4.7 on self-hosted deployments because the SDK defaults `reasoning_effort` to `"high"`, which the bundled litellm (v1.83.0) maps to `thinking.type=enabled` -- a parameter that Opus 4.7 does not accept. Users who set `reasoning_effort=None` in their settings to work around this had their override discarded.

## Fix

Replace `LLM(model=..., base_url=..., api_key=..., usage_id=...)` with `user.agent_settings.llm.model_copy(update={...})`. This preserves all user-configured fields while still allowing the server to override the model name, base URL, and usage ID.

## Test

Adds a regression test (`test_configure_llm_preserves_reasoning_effort`) that sets `reasoning_effort=None` and `extended_thinking_budget=None` on the user's LLM config and verifies both survive through `_configure_llm_and_mcp`.

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -902,6 +902,12 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
     def _configure_llm(self, user: UserInfo, llm_model: str | None) -> LLM:
         """Configure LLM settings.
 
+        Starts from the user's saved LLM configuration and overrides only
+        the fields that the server needs to resolve (model name, base URL,
+        and usage ID).  All other user-configured fields (e.g.
+        ``reasoning_effort``, ``extended_thinking_budget``, ``drop_params``)
+        are preserved so that they reach the agent-server unchanged.
+
         Args:
             user: User information containing LLM preferences
             llm_model: Optional specific model to use, falls back to user default
@@ -921,11 +927,13 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
             provider_base_url=self.openhands_provider_base_url,
         )
 
-        return LLM(
-            model=model,
-            base_url=base_url,
-            api_key=user.agent_settings.llm.api_key,
-            usage_id='agent',
+        return user.agent_settings.llm.model_copy(
+            update={
+                'model': model,
+                'base_url': base_url,
+                'api_key': user.agent_settings.llm.api_key,
+                'usage_id': 'agent',
+            }
         )
 
     async def _add_system_mcp_servers(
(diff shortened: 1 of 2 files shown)


enyst (approved):
Thank you!

Issue #14322, opened 2026-05-06
fix(settings): switching agent_kind leaks stale fields across discriminated union types


simonrosenberg:
## Problem

Switching between the OpenHands agent and an ACP agent (Settings → Agent) throws a 500 **\"Something went wrong storing settings\"** on the return trip (ACP → OpenHands).

**Root cause:** `Settings.update()` in `settings_models.py` handles all `agent_settings_diff` saves the same way: dump the current settings, deep-merge the diff, re-validate. This works well for *within-type* partial updates (LLM page, condenser page, etc.) but breaks when `agent_kind` changes, because the two union members have different schemas:

- `OpenHandsAgentSettings` has `agent_context`, `condenser`, `verification`, `mcp_config`, …
- `ACPAgentSettings` has `acp_server`, `acp_command`, `acp_args`, `acp_env`, `acp_model`, …

When the user switches OpenHands → ACP, the merge preserves `agent_context` (and others) from the old dump in the stored blob. When switching back, those stale fields are still in the base, and `validate_agent_settings` chokes on e.g. `agent_context: null` which is not a valid `AgentContext`.

**Reproducer:**
1. Start with OpenHands agent settings saved
2. Switch to an ACP agent and save
3. Switch back to OpenHands agent → 500

**Why a validator band-aid is wrong:** Adding `@field_validator("agent_context", mode="before")` to coerce `None` fixes one symptom but leaves the structural problem: every type-specific field from the old union member is silently dragged into the new one's stored blob, accumulating garbage across switches.

**Why carrying over `llm` on switch is also wrong:** Both union members have an `llm: LLM` field, but they mean different things. In `OpenHandsAgentSettings` it is the actual LLM the agent uses (api_key, model, temperature, base_url, …). In `ACPAgentSettings` it is attribution-only metadata for cost/token bookkeeping — the user configures the subprocess model via `acp_model: str | None` instead. Sharing the `llm` object across types on a switch would silently populate ACP attribution fields with OpenHands LLM credentials, and vice versa.

## Correct fix — snapshot-on-switch

Add a `saved_agent_configs: dict[str, Any]` field to `Settings` that persists a full snapshot of each type's config independently. In `Settings.update()`, detect when `agent_kind` changes and:

1. **Save** the current type's config into `saved_agent_configs[current_kind]`
2. **Restore** from `saved_agent_configs[new_kind]` if a prior snapshot exists; otherwise start from fresh defaults for that type
[...]

PR #14453, opened 2026-05-18
feat: preserve LLM/condenser/MCP config across OH ↔ ACP kind toggles

### LINKED ISSUE #14370: ACP settings: preserve LLM/condenser/MCP config across OpenHands ↔ ACP toggles
## Context

PR #14227 ships a minimum-viable agent-kind switch for the new ACP settings UI: toggling **Settings → Agent** between **OpenHands** and **ACP** no longer 500s, but the new kind always starts from defaults. The user's outgoing-kind config (LLM model, API key, condenser, MCP servers, ACP command, ACP env, ACP model) is **not** carried back when they switch the other way.

This issue tracks the round-trip preservation feature that was previously bundled into #14227 and split out to keep the PR reviewable. It originally closed #14322 — this issue now owns that thread.

## Current behaviour (post-#14227)

1. User has OpenHands selected with `anthropic/claude-sonnet-4-5` configured, plus condenser + 3 MCP servers.
2. User switches to ACP, picks the **Claude Code** preset, adds `ANTHROPIC_API_KEY` via Secrets, runs a conversation.
3. User switches back to **OpenHands**.
4. **The LLM dropdown is empty. Condenser is back to default. MCP servers are gone.** User must re-enter everything.
5. Symmetric loss going OH → ACP → OH on the ACP side: command, args, model, env.

The agent settings page itself shows defaults because `Settings.update()` writes `base = {'agent_kind': new_kind}` (a fresh empty config) whenever `agent_kind` changes in the payload. This is the deliberate minimum that fixes the discriminated-union 500 — preserving config across switches is a separate concern.

## Desired behaviour

Toggling OH ↔ ACP should preserve each kind's prior config:

1. User switches from OH (with full LLM + condenser + MCP) to ACP. Backend snapshots the outgoing OH config.
2. User configures ACP (command + env + model). Runs a conversation.
3. User switches back to OH. Their original LLM/condenser/MCP are restored exactly.
4. User switches back to ACP. The ACP command + env + model from step 2 are restored.

Acceptance: the API key in the OH snapshot survives a SaaS round-trip — without it, switching back loses the credential, which is the worst-case UX since the user re-enters a secret they thought was saved.

## Proposed design

Adapted from the design that was in #14227 before the split.

### Schema: ``Settings.saved_agent_configs: dict[str, Any]``

Per-kind snapshots, keyed by the agent-kind string (``'openhands'``, ``'acp'``). The dict value is the full ``agent_settings.model_dump(mode='json', context={'expose_secrets': True})`` for that kind.

- **OSS (``FileSettingsStore``)**: writes the whole ``Settings`` blob as JSON — the field survives with no schema change.
- **SaaS (``SaasSettingsStore``)**: needs a dedicated ``User`` column. The store's ``setattr(user, key, value)`` loop is gated by ``hasattr(user, key)``, so without a column the dict is silently dropped on save. Use ``EncryptedJSON`` since snapshots contain ``api_key`` material.

### Logic: ``Settings.update()`` snapshot/restore

When ``new_kind != current_kind``:

1. If ``current_kind`` is non-None, write ``saved_agent_configs[current_kind] = self.agent_settings.model_dump(mode='json', context={'expose_secrets': True})``.
2. Set ``base = dict(saved_agent_configs.get(new_kind, {'agent_kind': new_kind}))`` — restore the prior snapshot for the new kind, or fall through to a fresh base if no snapshot exists.
3. Deep-merge ``coerced`` (the incoming diff) onto ``base`` as today.
4. ``acp_env`` and ``mcp_config`` continue to replace wholesale (already shipped in #14227, unrelated to snapshots).

### Edge cases to test

- **``current_kind = None`` on fresh install**: JSON dicts can't carry a None key. Skip the snapshot write.
- **Corrupt snapshot**: a snapshot whose schema no longer validates (renamed field, stale option) must surface as a clean ``ValidationError`` (turns into a 4xx), not a 500. The current settings stay in their previous consistent state.
- **Inline override alongside switch**: ``{agent_kind: 'openhands', llm: {model: 'X'}}`` should apply the LLM override on top of the restored snapshot, not discard it.
- **``model_dump`` round-trip**: the dict must survive ``Settings.model_dump() → Settings.model_validate()`` so the snapshot persists across requests.

## Migration

Enterprise / SaaS deployments need:

\`\`\`python
# enterprise/migrations/versions/NNN_add_saved_agent_configs_to_user.py
def upgrade() -> None:
    op.add_column('user', sa.Column('saved_agent_configs', sa.String(), nullable=True))

def downgrade() -> None:
    op.drop_column('user', 'saved_agent_configs')
\`\`\`

Plus an ``EncryptedJSON``-typed ``Mapped[dict[str, Any] | None]`` column on the ``User`` model in ``enterprise/storage/user.py``.

## Test plan

Most of these existed as a 9-test file in the original PR and were dropped during the split:

- ``test_switch_oh_to_acp_snapshots_openhands_config`` — outgoing config is captured on switch.
- ``test_round_trip_preserves_openhands_llm_config`` — OH → ACP → OH brings back the original LLM.
- ``test_switch_back_to_existing_snapshot_replaces_current`` — second switch reads the saved snapshot, not defaults.
- ``test_no_snapshot_written_when_current_kind_is_none`` — None-key guard.
- ``test_switch_with_replace_mcp_config_clears_old_servers`` — wholesale-replace works through a switch.
- ``test_switch_back_replaces_acp_env_wholesale`` — smaller acp_env on switch-back drops old keys.
- ``test_switch_into_existing_snapshot_preserves_other_fields_in_same_payload`` — concurrent override + switch.
- ``test_corrupt_saved_agent_config_falls_back_cleanly`` — bad snapshot yields ValidationError, not 500.
- ``test_saved_agent_configs_round_trips_through_model_dump`` — Settings serialise/deserialise preserves the dict.

Plus the SaaS persistence test:

- ``test_store_and_load_saved_agent_configs_round_trip`` (enterprise/tests/unit/test_saas_settings_store.py) — admin_store.store(settings) → admin_store.load() → snapshot is intact, ``api_key`` value survives the encrypt/decrypt cycle.

## Risk / scope

- New encrypted JSON column on ``user``. Same encryption mechanism already used by ``llm_profiles``.
- Snapshot contains plaintext (post-decryption) ``api_key``. ``EncryptedJSON`` is required.
- No frontend changes — the agent-settings page reads ``settings.agent_settings.*`` and doesn't need to know snapshots exist.
- No DB rollback risk: the column is nullable, defaults to NULL, missing column simply means no preservation (the post-#14227 minimum-viable behaviour).

## Linked

- Originally part of: #14227
- Closes: #14322 (moved here from #14227)

simonrosenberg:
## Summary

- Toggling **Settings → Agent** between OpenHands and ACP no longer loses the outgoing config
- `Settings.update()` now snapshots each kind's full config into a new `saved_agent_configs` field and restores it on switch-back
- Inline overrides in the same payload as a kind switch apply on top of the restored snapshot
- SaaS: adds an `EncryptedJSON` column `saved_agent_configs` to the `User` table (migration 113); pre-migration `NULL` rows fall back to the empty-dict default with no behaviour change

## Closes

Closes #14370
(originally tracked in #14322)

## Changes

| File | What changed |
|---|---|
| `openhands/app_server/settings/settings_models.py` | `saved_agent_configs` field + snapshot/restore logic in `update()` |
| `enterprise/storage/user.py` | `EncryptedJSON` column for SaaS persistence |
| `enterprise/migrations/versions/113_…py` | Alembic migration: add `saved_agent_configs` to `user` table |
| `enterprise/storage/saas_settings_store.py` | Null-guard in `load()` for pre-migration rows |
| `tests/unit/app_server/test_settings_agent_kind_switch.py` | 9 new tests; removed the now-wrong "resets to defaults" test |
| `enterprise/tests/unit/test_saas_settings_store.py` | SaaS round-trip test for `saved_agent_configs` |

## Test plan

- [ ] `tests/unit/app_server/test_settings_agent_kind_switch.py` — all 13 pass locally
- [ ] `test_round_trip_preserves_openhands_llm_config` — OH→ACP→OH restores model
- [ ] `test_saved_agent_configs_round_trips_through_model_dump` — survives serialize/deserialize
- [ ] `test_corrupt_saved_agent_config_falls_back_cleanly` — bad snapshot raises `ValidationError`, not 500
- [ ] Enterprise: `test_store_and_load_saved_agent_configs_round_trip` — api_key survives encrypt/decrypt

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-5bbadeb   docker.openhands.dev/openhands/openhands:5bbadeb
```

### DIFF
--- a/enterprise/migrations/versions/113_add_saved_agent_configs_to_user.py
+++ b/enterprise/migrations/versions/113_add_saved_agent_configs_to_user.py
@@ -1,0 +1,35 @@
+"""Add saved_agent_configs column to user table.
+
+The Settings model exposes ``saved_agent_configs`` (snapshots of agent
+settings per kind), but without a column on the User row the field is
+silently dropped on store() and always defaults to empty on load(), so
+switching between agent kinds loses the user's previous config.
+
+The column is plain ``String`` because the ORM-level ``EncryptedJSON``
+TypeDecorator stores JSON-serialized data as a JWE-encrypted string —
+configs can carry per-kind ``api_key`` values, so the at-rest
+representation must match the existing org/member encrypted-secret pattern.
+
+Revision ID: 113
+Revises: 112
+Create Date: 2026-05-18
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = '113'
+down_revision: Union[str, None] = '112'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column('user', sa.Column('saved_agent_configs', sa.String(), nullable=True))
+
+
+def downgrade() -> None:
+    op.drop_column('user', 'saved_agent_configs')
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -168,6 +168,10 @@ class SaasSettingsStore(SettingsStore):
         # non-nullable, so let the default_factory take over.
         if kwargs.get('llm_profiles') is None:
             kwargs.pop('llm_profiles', None)
+        # Pre-migration rows read back as None; Settings.saved_agent_configs
+        # has default_factory=dict, so let the factory take over.
+        if kwargs.get('saved_agent_configs') is None:
+            kwargs.pop('saved_agent_configs', None)
 
         return Settings(**kwargs)
 
--- a/enterprise/storage/user.py
+++ b/enterprise/storage/user.py
@@ -43,6 +43,9 @@ class User(Base):
     llm_profiles: Mapped[dict[str, Any] | None] = mapped_column(
         EncryptedJSON, nullable=True
     )
+    saved_agent_configs: Mapped[dict[str, Any] | None] = mapped_column(
+        EncryptedJSON, nullable=True
+    )
     onboarding_completed: Mapped[bool | None] = mapped_column(
         nullable=True, default=False
     )
--- a/enterprise/tests/unit/test_saas_settings_store.py
+++ b/enterprise/tests/unit/test_saas_settings_store.py
@@ -940,3 +940,62 @@ async def test_llm_profiles_are_encrypted_at_rest(
         ).scalar_one()
     assert user.llm_profiles is not None
     assert user.llm_profiles['profiles']['work']['api_key'] == 'super-secret-byok'
+
+
+@pytest.mark.asyncio
+async def test_store_and_load_saved_agent_configs_round_trip(
+    async_session_maker, org_with_multiple_members_fixture
+):
+    """saved_agent_configs must persist on the User row and round-trip through
+    store → load. Without the user.saved_agent_configs column the field is
+    silently dropped on store() and always defaults to empty on load(), so
+    switching back to a previous agent kind loses the user's previous config."""
+    fixture = org_with_multiple_members_fixture
+    admin_user_id = str(fixture['admin_user_id'])
+    admin_store = SaasSettingsStore(admin_user_id)
+
+    settings = DataSettings()
+    # Set an OH model, then switch to ACP — this populates saved_agent_configs
+    settings.update(
+        {
+            'agent_settings_diff': {
+                'llm': {
+                    'model': 'anthropic/claude-sonnet-4-5-20250929',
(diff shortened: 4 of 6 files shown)


simonrosenberg:
Closing in favour of #14454 which solves this via LLM profiles instead.

PR #14459, opened 2026-05-18
Add 👀 eye reaction acknowledgement for BBDC PR comments (Part 3)

ak684:
## Summary

Stacks on top of [Part 2 — Bitbucket Cloud webhook enrollment lifecycle](https://github.com/OpenHands/OpenHands/tree/alona/bitbucket-cloud-webhook-enrollment-pr2). When a Bitbucket Data Center user @-mentions the resolver in a PR comment, the agent now posts a 👀 reaction on that comment **immediately** after the permission check passes — before the conversation is created and the agent's reply is posted. Mirrors the GitHub manager's existing 'eyes' reaction UX so the user gets fast visual feedback that the request was received.

## Why

Tested manually against `bitbucket-02.aivong.platform-team.all-hands.dev` and confirmed the BBDC integration works end-to-end **but provides no acknowledgement** between the user's `@openhands` comment and the agent's eventual reply (~tens of seconds while the conversation spins up). GitHub has had this UX since forever (`github_manager.py:276-277` — `_add_reaction(view, 'eyes', installation_token)`); BBDC and Bitbucket Cloud both lacked it. This closes the gap for BBDC.

## Scope

- **BBDC only.** Bitbucket Cloud's public REST API does not expose a reactions endpoint for PR comments — the reactions feature exists in the Cloud UI but isn't reachable via API, so parity there isn't possible without an upstream change.
- **BBDC 7.0+ required** for the reactions endpoint (`/rest/api/1.0/projects/{key}/repos/{slug}/pull-requests/{id}/comments/{cid}/reactions`). Older installs return 404; the helper logs at INFO and swallows the exception so a missing reaction never blocks conversation creation.

## Changes

| File | Lines | What |
|---|---|---|
| `openhands/app_server/integrations/bitbucket_data_center/service/resolver.py` | +24 | New `add_comment_reaction(owner, repo_slug, pr_id, comment_id, emoticon)` on `BitbucketDCResolverMixin` — POSTs `{'emoticon': ...}` via the existing `_make_request` helper. |
| `enterprise/integrations/bitbucket_data_center/bitbucket_dc_manager.py` | +55 | New `_add_eyes_reaction(message, installer_user_id, project_key, repo_slug)` helper + call site in `receive_message` right after `_commenter_has_write_access` returns True. Uses the webhook installer's BBDC token (the user whose OAuth backs the enrollment). |

Total: 79 LOC, no tests yet (see below).

## Verification done

[...]

PR #640, opened 2024-04-03
fix(frontend): Pass BACKEND_PORT to frontend on startup

georg-wolflein:
When changing `BACKEND_PORT` in the `Makefile`, `make start-frontend` (and `make run`) will not be aware of this change, still using the default 3000 port.

The fix is to just provide the `BACKEND_HOST` environment variable to the frontend startup command.

PR #14461, opened 2026-05-18
[PLTF-2500] Add automated Bitbucket Data Center webhook lifecycle (Part 4)

ak684:
Automated webhook lifecycle for Bitbucket Data Center, mirroring what #14414 does for Cloud. Replaces the manual paste-the-secret flow from #14407 with Install / Reinstall / Uninstall buttons that drive BBDC's REST API directly.

Also brings the resolver in line with the GitHub flow: jobs now run as the **mentioner** (the user who @'d the bot), not the webhook installer. That means the mentioner's offline_token, BBDC OAuth token, LLM settings, and dashboard ownership are used end to end. Both the immediate ack ("I'm on it!") and the final summary post under the mentioner's BBDC account. The installer's id is carried alongside on the view, used only for the commenter permission check and the webhook lifecycle REST calls (both of which require admin scope that the mentioner may not have). Falls back to the installer when the mentioner has no OHE account, so the resolver stays useful for unenrolled BBDC users.

Needs `REPO_ADMIN` scope (was `REPO_WRITE`). Chart fixes merged in [OpenHands-Cloud #640](https://github.com/OpenHands/OpenHands-Cloud/pull/640): the scope bump, the BBDC IDP id-mapper claim (`sub`/numeric to `preferred_username`/slug so the lookup matches), and the `bitbucket_data_center_id` allowlist in the realm user profile schema. That PR is on Cloud main, so the realm is ready for this code once a new enterprise-server image deploys.

### Testing
Successfully tested end to end with a real instance of OHE and bitbucket data center

https://github.com/user-attachments/assets/edbcb5d5-4524-4116-a6b5-b118d46268fb

26 unit tests across service mixin, storage, manager, and frontend manager. All green. Live end-to-end on alonacorp with two BBDC users:
- testadmin2 @-mentions @openhands: log shows `Running job as mentioner 'testadmin2' (keycloak 49135ca0-...) instead of installer (bf61cea8-...)`, both the ack and the final summary post as Test Admin2, conversation appears in testadmin2's OHE dashboard, every BBDC token fetch is `ai.vong+bdctestadmin2@openhands.dev`.
- alona (installer) @-mentions @openhands: lookup succeeds, no swap log emitted (mentioner == installer), job runs as alona as before, no behavior change for the common case.

Install / Reinstall / Uninstall buttons exercised on a fresh repo, a previously-manually-enrolled one, and a Reinstall-then-Uninstall flow, with zero exceptions, migration 113 ran cleanly against the existing rows.

### Known follow-ups

- Bitbucket Cloud has the same installer-as-runner behavior. A follow-up PR can mirror the receive_message change in `bitbucket_manager.py`.
- Separate chart issue: the keycloak_auth cookie wraps the access + refresh JWTs and crosses Chrome's 4096-byte per-cookie limit for some users (depends on Keycloak claim length, e.g. realm roles and email length). Worth splitting the signed token into chunked cookies or moving to a server-side session store.

### Related

- #14407 manual BBDC enrollment (merged)
- #14414 Bitbucket Cloud lifecycle
- #14459 BBDC eye-reaction ack

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-9fe63a1   docker.openhands.dev/openhands/openhands:9fe63a1
```

---
Linear: PLTF-2500

_AI-generated metadata update by OpenHands on behalf of ak684._


ak684:
@all-hands-bot took two of the three, pushed back on the third.

**1. Audit logging — done** (`b315fbf68`)

Added INFO logs on both the reinstall and uninstall route handlers with `user_id`, `project_key`, `repo_slug`, `webhook_id`. Now the "trust the OAuth scope grant" posture in `user_has_admin_access` has an explicit audit trail to match. See `enterprise/server/routes/integration/bitbucket_dc.py` (in both `reinstall_bitbucket_dc_webhook` and `uninstall_bitbucket_dc_webhook`).

**2. Schema fix — done** (`b315fbf68`)

New migration `113_bitbucket_dc_webhook_last_synced_tz.py` alters `bitbucket_dc_webhook.last_synced` from `TIMESTAMP WITHOUT TIME ZONE` to `TIMESTAMP WITH TIME ZONE`, with `postgresql_using = "last_synced AT TIME ZONE 'UTC'"` for in-place conversion. Existing naive values get read as UTC during the type change, which is what they always were semantically. Both `.replace(tzinfo=None)` calls in the store are gone, the SQLAlchemy model now uses `DateTime(timezone=True)`, and the workaround line in `update_webhook_id` collapses back to a clean one-liner. Downgrade path is included.

The 16 unit tests (9 storage, 7 service) still pass against the updated schema (the in-memory sqlite fixture doesn't enforce the tz distinction so the tests don't need to change, but the model change is exercised on every upsert/update path).

**3. HTTP client handlers — disagree, looks like a misread**

The bot's comment says PUT should use `json=params` like POST, and DELETE should use query params not JSON body. Both of those are already how `http_client.py:75-79` works:

```python
if method == RequestMethod.POST:    return await client.post(url, headers=headers, json=params)
if method == RequestMethod.PUT:     return await client.put(url, headers=headers, json=params)
if method == RequestMethod.DELETE:  return await client.delete(url, headers=headers, params=params)
return await client.get(url, headers=headers, params=params)
```

PUT uses `json=params` (line 77). DELETE uses query `params` (line 79), not a JSON body. This is identical to what Part 2 (#14414) shipped and matches both the BBDC REST contract (PUT takes a body, DELETE doesn't) and the bot's own recommendation. Leaving as-is.

aivong-openhands (approved):

PR #14472, opened 2026-05-20
feat(frontend): show "Delivering..." status for queued messages

### LINKED ISSUE #14181: [Feature]: Introduce support for displaying queued messages in the conversation UI
### Is there an existing feature request for this?

### Problem or Use Case

Currently, when a conversation is initiated and messages are sent, those messages may be temporarily queued until the conversation is fully ready.

However, during this period, end users are not aware that their messages are in a queued state.

### Proposed Solution

Image

We need to introduce a UI approach similar to the example shown above (the image is for reference only). The design should align with the existing OpenHands look, feel, and styling guidelines.

Pending messages should be visibly displayed in the chat interface with a “Delivering…” status. Once the conversation becomes ready and the messages are successfully sent, the “Delivering…” status should be removed and the messages should transition to their sent state.

### Alternatives Considered

_No response_

### Priority / Severity

### Estimated Scope

Unknown - Not sure about the technical complexity

### Feature Area

Agent / AI behavior

### Technical Implementation Ideas (Optional)

_No response_

### Additional Context

_No response_

gaurav0107:
## End-user friendly description of the problem this fixes or functionality that this introduces

When a user sends a message in a fresh conversation before the WebSocket has opened, the client falls back to the REST pending-message endpoint and the message gets queued server-side for later delivery. Until now the chat surface suppressed the optimistic message in that case, so the user saw their input clear with **no feedback at all** — the message appeared lost until the conversation became ready and the backend echoed it back.

This PR closes that feedback gap by always rendering the optimistic user message and, when the message is queued rather than sent, showing a small italic "Delivering..." indicator with animated dots beneath the bubble. The indicator disappears as soon as the real event arrives over the WebSocket.

Resolves #14181.

## Give a summary of what the PR does, explaining any non-trivial design decisions

- `optimistic-user-message-store` gains an `isPendingDelivery` boolean. `setOptimisticUserMessage(message, isPendingDelivery?)` is backwards compatible (the flag defaults to `false`).
- `chat-interface.tsx` now calls `setOptimisticUserMessage(content, result.queued)` for both code paths; previously the queued path did nothing.
- The V0 and V1 `Messages` components both read the new flag from the store and pass it through to the existing `<ChatMessage>` render of the optimistic message.
- `<ChatMessage>` accepts a new optional `isPendingDelivery` prop (defaults to `false`). When true it renders a `data-testid="delivering-indicator"` element with `role="status"` and `aria-live="polite"` so screen readers announce the transition when the indicator unmounts on delivery.
- The visible "Delivering" string is added as a new i18n key `CHAT_INTERFACE$MESSAGE_DELIVERING` with translations across all 15 supported locales.
- Tests in `chat-message.test.tsx` cover three states: default (no indicator), `isPendingDelivery` (indicator with correct ARIA), and `isPendingDelivery={false}` (no indicator).

Design notes:
- I deliberately kept the indicator inside `<ChatMessage>` rather than introducing a wrapper component so existing call sites that already render optimistic messages picked up the change with one prop. Sixteen existing call sites are unaffected because the prop is optional.
- The pending state lives on the store rather than the optimistic-message string itself so existing consumers (which only care about the message text) keep working without API changes.
- The animated dots use Tailwind's `animate-[pulse_…]` arbitrary value with three staggered delays. They are wrapped in `aria-hidden="true"` so the live-region announcement reads "Delivering" rather than "Delivering . . ." to screen readers.

## Other references

- Issue body and reference screenshot: #14181
- Existing pending-message infrastructure: `frontend/src/api/pending-message-service/` and the queue path in `frontend/src/contexts/conversation-websocket-context.tsx` (the REST endpoint and `{ queued: true }` return value were already in place; this PR only wires the UI feedback that was missing).

## Link of any specific issues this addresses

Resolves #14181

---

```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   -e SANDBOX_RUNTIME_CONTAINER_IMAGE=docker.all-hands.dev/all-hands-ai/runtime:0.62-nikolaik   --name openhands-app-feat-14181   docker.all-hands.dev/all-hands-ai/openhands:feat-14181
```


neubig:

Thank you for understanding.

PR #14489, opened 2026-05-20
fix: recover conversation metadata save race

he-yufeng:
﻿## Summary
- recover from a duplicate-key race when conversation metadata is inserted by a concurrent startup/webhook writer
- after rolling back the failed insert, update the row that already exists for the same conversation id
- add a regression test for the concurrent insert path

## To verify
- `uv run pytest tests/unit/app_server/test_sql_app_conversation_info_service.py -q --basetemp .tmp/pytest-sql-conversation-info`
- `uv run ruff check openhands/app_server/app_conversation/sql_app_conversation_info_service.py tests/unit/app_server/test_sql_app_conversation_info_service.py`
- `python -m py_compile openhands/app_server/app_conversation/sql_app_conversation_info_service.py tests/unit/app_server/test_sql_app_conversation_info_service.py`
- `git diff --check`

Fixes OpenHands/enterprise#26

### DIFF
--- a/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
@@ -34,6 +34,7 @@ from sqlalchemy import (
     select,
 )
 from sqlalchemy.engine import CursorResult
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import Mapped, mapped_column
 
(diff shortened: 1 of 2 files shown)


he-yufeng:
Confirmed ready for review. I rebased the branch onto current main and re-ran the focused unit test plus ruff locally.

he-yufeng:
Closing this older conversation-metadata race branch in favor of OpenHands/OpenHands#14579, which keeps the same issue focused on the retry-after-insert race and has a clearer verification path.

PR #14528, opened 2026-05-22
fix: scope MCP and ACP env vars to the acting org member

hieptl:
HUMAN:

AGENT:

---

### Summary

When a member of an organization creates an MCP server on `/settings/mcp`, every other member of the same organization sees it. MCP server configuration is supposed to be private to the member that created it.

### Steps to reproduce

1. Create an organization `A` with two members, Member 1 and Member 2.
2. Sign in as Member 1, go to `/settings/mcp`, and add a new MCP server.
3. Sign out and sign in as Member 2.
4. Open `/settings/mcp`.

### Actual behavior

Member 2 sees the MCP server Member 1 created.

### Expected behavior

Each org member's MCP server list is private to that member. Member 2 should see an empty MCP server list until they add their own.

### Root cause

`SaasSettingsStore.store()` in `enterprise/storage/saas_settings_store.py` treats `mcp_config` the same as every other agent setting. On save, the store takes the full effective `agent_settings` dump (which includes `mcp_config`) and:

1. Writes it to `org.agent_settings` via `deep_merge_with_wholesale_keys(...)`. `org.agent_settings` is the default that every member's `load()` deep-merges with their own diff, so Member 1's MCP servers become part of the org defaults Member 2 sees.
2. Calls `OrgMemberStore.update_all_members_settings_async(...)` with the full diff. That helper applies `mcp_config` to every `OrgMember` row in the org, copying Member 1's MCP servers directly onto Member 2's row.

Either fan-out alone is enough to leak; both fired on every save.

`acp_env` (ACP environment variables) is a sibling key in `WHOLESALE_REPLACEMENT_KEYS` with the same per-member semantics and the same leak.

### Scope

- Fixed: per-user `POST /api/v1/settings` write path (the `/settings/mcp` page).
- Out of scope (separate follow-up): the org-defaults endpoint `PATCH /api/organizations/{id}/settings`. An org admin can still write `mcp_config` into `agent_settings_diff` through that route. Worth a separate ticket to harden, since intent there is "org admin sets org-wide defaults" — different threat model.

### Acceptance criteria

- A member's MCP servers (`mcp_config`) and ACP env vars (`acp_env`) are stored only on that member's `org_member` row.
- Other members of the same org see no trace of those values via `load()`.
- `org.agent_settings` carries no `mcp_config` or `acp_env` after a user-side save.
- Pre-existing org-level `mcp_config` left over from before this fix does not leak to members on load (defense for stale data).
- Deleting an MCP server (the APP-1862 contract) still sticks for the acting member; deletes are not resurrected by deep-merge.

## Summary

- `mcp_config` and `acp_env` are now per-member, never written to `org.agent_settings`, never broadcast across the org.
- Splits the effective `agent_settings_diff` into a shared half (LLM/agent/condenser/etc.) and a private half (`mcp_config`, `acp_env`) inside `SaasSettingsStore.store()`. Shared half flows to `org.agent_settings` + `update_all_members_settings_async` as before; private half is applied wholesale to the acting `org_member` only.
- `SaasSettingsStore.load()` strips the private keys from the org dump before deep-merging with the member diff, so pre-fix legacy `org.agent_settings.mcp_config` in the database stops leaking on read.

### Why

Member 1 of an org would add an MCP server on `/settings/mcp` and every other member of the same org would see it. Root cause: the user-settings save path treated `mcp_config` like a shared org setting and fanned it out twice (org row + every member row). See the linked Linear ticket for the full reproduction.

`acp_env` shares the same `WHOLESALE_REPLACEMENT_KEYS` semantics as `mcp_config` and had the same leak — fixed together via a single constant (`MEMBER_PRIVATE_AGENT_KEYS`).

### Files changed

- `enterprise/storage/saas_settings_store.py`
  - New module-level constant `MEMBER_PRIVATE_AGENT_KEYS = WHOLESALE_REPLACEMENT_KEYS` + `_split_member_private_keys()` helper.
  - `store()` splits the diff and writes only the shared half to org-wide state. Private half goes to the acting member only, via `deep_merge_with_wholesale_keys` so deletes stick.
  - `store()` also strips any pre-existing private keys from `org.agent_settings` before the merge, so legacy values are cleaned up on the next save.
  - `load()` strips private keys from the org dump before merging with the member diff.
- `enterprise/tests/unit/test_saas_settings_store.py`
  - Replaced `test_store_saves_mcp_config_in_agent_settings` with `test_store_keeps_mcp_config_private_to_acting_member` — asserts only the acting member's row carries `mcp_config`, org and other members do not.
  - Updated `test_store_replaces_mcp_config_on_delete` to assert both the APP-1862 wholesale-replace-on-delete contract **and** that other members' rows are never touched.
  - New `test_load_drops_legacy_org_level_mcp_config` — pre-seeds `org.agent_settings.mcp_config` (simulating pre-fix data) and asserts the load path does not surface it to a member with no `mcp_config` of their own.

### Out of scope

- The org-defaults endpoint (`PATCH /api/organizations/{org_id}/settings`) still accepts `mcp_config` in `agent_settings_diff` and writes it to the org row. That's a different threat model (admin-intent vs user-intent) and gets its own ticket.

## How to Test

1. Create an organization `A` with two members, Member 1 and Member 2.
2. Sign in as Member 1, go to `/settings/mcp`, and add a new MCP server.
3. Sign out and sign in as Member 2.
4. Open `/settings/mcp`.

## Video/Screenshots

https://github.com/user-attachments/assets/fc572050-1d0b-4d40-a308-48fe107b8dea

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### Risk

Low. The fix is scoped to the `SaasSettingsStore` write/read paths; the underlying `OrgMemberStore.update_all_members_settings_async` helper is unchanged so other callers behave identically. Legacy data with `mcp_config` on `org.agent_settings` is silently cleaned up on the next save and ignored on load.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-9d9af70   docker.openhands.dev/openhands/openhands:9d9af70
```

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -25,10 +25,40 @@ from storage.user_store import UserStore
 from openhands.app_server.settings.settings_models import Settings
 from openhands.app_server.settings.settings_store import SettingsStore
 from openhands.app_server.utils.jsonpatch_compat import (
+    WHOLESALE_REPLACEMENT_KEYS,
     deep_merge,
     deep_merge_with_wholesale_keys,
 )
 from openhands.app_server.utils.llm import is_openhands_model
+
+# Agent-settings keys that are private to each org member and must never
+# be written to org-level defaults or broadcast across the org. Today this
+# covers ``mcp_config`` (per-user MCP server set) and ``acp_env`` (per-user
+# ACP environment variables) — both are dict-of-items collections that
+# represent one member's personal configuration, not org-wide defaults.
+MEMBER_PRIVATE_AGENT_KEYS: frozenset[str] = WHOLESALE_REPLACEMENT_KEYS
+
+
+def _split_member_private_keys(
+    agent_settings_diff: dict[str, Any],
+) -> tuple[dict[str, Any], dict[str, Any]]:
+    """Split an agent_settings dump into (shared, private) halves.
+
+    The shared half is safe to write to ``org.agent_settings`` and to
+    broadcast through ``update_all_members_settings_async``. The private
+    half must be applied only to the acting member's row.
+    """
+    private = {
+        key: agent_settings_diff[key]
+        for key in MEMBER_PRIVATE_AGENT_KEYS
+        if key in agent_settings_diff
+    }
+    shared = {
+        key: value
+        for key, value in agent_settings_diff.items()
+        if key not in MEMBER_PRIVATE_AGENT_KEYS
+    }
+    return shared, private
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


tofarr (approved):
🍰

PR #14531, opened 2026-05-22
fix: forward insecure git access env to sandboxes

he-yufeng:
HUMAN: This fixes self-hosted agent-server sandboxes dropping `ALLOW_INSECURE_GIT_ACCESS` even when the parent process explicitly enables it. The forwarding remains narrow: only the existing allowlisted prefixes plus this explicit OpenHands env flag are copied automatically; arbitrary env forwarding still requires `OH_AGENT_SERVER_ENV`.

- [ ] A human has tested these changes.

Refs #14523.

## Summary
- auto-forward `ALLOW_INSECURE_GIT_ACCESS` into agent-server sandbox specs alongside the existing `LLM_` and `LMNR_` forwarding
- keep arbitrary env forwarding explicit via `OH_AGENT_SERVER_ENV`
- add coverage for exact env forwarding and the Docker sandbox spec environment

## Validation
- `uv run pytest tests\unit\app_server\test_agent_server_env_override.py::TestExactAutoForwarding tests\unit\app_server\test_agent_server_env_override.py::TestDockerSandboxSpecEnvironmentOverride::test_docker_specs_include_insecure_git_access_env -q`
- `uv run ruff check openhands\app_server\sandbox\sandbox_spec_service.py tests\unit\app_server\test_agent_server_env_override.py`
- `python -m py_compile openhands\app_server\sandbox\sandbox_spec_service.py tests\unit\app_server\test_agent_server_env_override.py`
- `git diff --check`

Note: I also ran the full `tests\unit\app_server\test_agent_server_env_override.py` on Windows. It reached 43 passed / 2 failed; the failures are the existing case-sensitivity expectations for `LLM_` and `LMNR_` env names on Windows, where `os.environ` is case-insensitive. The new exact-forwarding tests pass.

PR #14535, opened 2026-05-23
fix: forward ALLOW_INSECURE_GIT_ACCESS to agent server

### LINKED ISSUE #14523: [Bug]: Environment variables dont get at all passed to OpenHands Docker (using OpenHands CLI) making local Gitea/forgejo etc use partly impossible

### Bug Description

TL:DR Problem is analysed and probably solved, see attached files. For details see this bugs text.

The core issue for me was to connect my local gitea(forgejo compatible) to OpenHands. Problem was, it doesnt accepted http only. Well so I researched and your Doku says, set Environmental variable ALLOW_INSECURE_GIT_ACCESS=true to accept http only connections.

I tried everything, set it globally via bashrc, pass it directly with the openhands start command, nothing worked.
I did some digging in with Perplexity and Claude and it seems no enviromental variables at all get passed into the docker container. The function is just broken.

So I Vibe coded a patch that patches the source code to set this variable to always true. Dirty but works.

However, my suggestions: Plan a global .env file or "enviromental-commands" or something named file and load that correctly into docker when starting up openhands with "openhands serve". Maybe also load in global enviromental (like in bashrc) adds, however, this could be filtered through a new to create "whitelist" containing only vaiable names OpenHands actually uses so not every system enviromental gets passed to the OpenHands Docker Container.

I already made a pull request worthy prework for just a few existing env-variables, however, its not perfectly refined yet. The whole analysis from Claude is attached here aswell. My simple patch script is included too, just for reference. AI can do this im sure (but not my local Qwen3.6 27B xd) but thats something a projects dev with deeper understanding of the codebase should do.

Another recommendation: Update Doku to mention Gitea alongside with forgejo and also add the info about the enviromental switch to allow http only access. The info of Gitea compatibility is missing at the moment completly.

Thanks for the great project btw, love it so far 🥇 

### Expected Behavior

OpenHands should allow http only Github like Connections when the env variable is set. For example when using a local Gitea/Forgejo.
There could also be a button in the Git settings WebUI that allowing doing that while im thinking about it now...

### Actual Behavior

Adding Enviromental Variable ALLOW_INSECURE_GIT_ACCESS=true to bashrc doenst lead to allowing "insecure http-only" Git Connections behaviour in OpenHands

### Steps to Reproduce

1. Put the env ALLOW_INSECURE_GIT_ACCESS=true into bashrc
2. run openhands serve
3. Try to connect to local Gitea (http only)

CLI (uv tool install)

_No response_

### OpenHands Version

OpenHands CLI 1.16.0 | OpenHands SDK v1.21.0

### Version Confirmation

### Model Name

Qwen3.6-27B, Claude, MiniMax 2.5 etc.

### Operating System

Linux

### Browser (if using web UI)

Firefox (newest)

### Logs and Error Messages

_No response_

[patch-openhands.sh](https://github.com/user-attachments/files/28150041/patch-openhands.sh)
[oh-start.sh](https://github.com/user-attachments/files/28150040/oh-start.sh)
[README(1).md](https://github.com/user-attachments/files/28150039/README.1.md)

[github-pr-text.md](https://github.com/user-attachments/files/28150090/github-pr-text.md)

YingzuoLiu:
Fixes #14523

HUMAN:
I traced the environment propagation path from `get_agent_server_env()` to the agent-server sandbox environment. This PR adds a narrow exact allowlist for `ALLOW_INSECURE_GIT_ACCESS` so it is forwarded without exposing arbitrary host environment variables.

## Summary

Fixes environment variable forwarding for `ALLOW_INSECURE_GIT_ACCESS`.

Previously, `get_agent_server_env()` only auto-forwarded environment variables with recognized prefixes such as `LLM_` and `LMNR_`. As a result, `ALLOW_INSECURE_GIT_ACCESS=true` was not propagated into the agent-server sandbox environment, even though the git provider logic expects this environment variable to be available.

This PR adds an exact allowlist for selected OpenHands runtime flags and forwards `ALLOW_INSECURE_GIT_ACCESS` without forwarding arbitrary host environment variables.

## Changes

- Added `AUTO_FORWARD_EXACT_VARS`
- Auto-forward `ALLOW_INSECURE_GIT_ACCESS` to the agent-server environment
- Updated the `get_agent_server_env()` docstring
- Added unit test coverage for exact env var forwarding

## Test Plan

- `git diff --check` passed
- `python -m py_compile openhands/app_server/sandbox/sandbox_spec_service.py` passed
- Targeted pytest was not run locally because `pytest` is not installed in my current environment

### DIFF
--- a/openhands/app_server/sandbox/sandbox_spec_service.py
+++ b/openhands/app_server/sandbox/sandbox_spec_service.py
@@ -2,6 +2,7 @@ import asyncio
 import os
 from abc import ABC, abstractmethod
 
+
 from openhands.agent_server import env_parser
 from openhands.app_server.errors import SandboxError
 from openhands.app_server.sandbox.sandbox_spec_models import (
@@ -73,6 +74,10 @@ def get_agent_server_image() -> str:
 # These are typically configuration variables that affect the agent's behavior
 AUTO_FORWARD_PREFIXES = ('LLM_', 'LMNR_')
 
+# Exact environment variables that should be auto-forwarded to agent-server.
+# These are OpenHands runtime flags that do not share a common prefix.
+AUTO_FORWARD_EXACT_VARS = ('ALLOW_INSECURE_GIT_ACCESS',)
+
 
 def get_agent_server_env() -> dict[str, str]:
     """Get environment variables to be injected into agent server sandbox environments.
@@ -80,7 +85,8 @@ def get_agent_server_env() -> dict[str, str]:
     This function combines two sources of environment variables:
 
     1. **Auto-forwarded variables**: Environment variables with certain prefixes
-       (e.g., LLM_*, LMNR_*) are automatically forwarded to the agent-server container.
+       (e.g., LLM_*, LMNR_*) and selected exact OpenHands runtime flags
+       are automatically forwarded to the agent-server container.
        This ensures that LLM configuration like timeouts and retry settings
        work correctly in the two-container V1 architecture, as well as
        Laminar monitoring/analytics configuration.
@@ -97,6 +103,7 @@ def get_agent_server_env() -> dict[str, str]:
         # Auto-forwarding (no action needed):
         export LLM_TIMEOUT=3600
         export LLM_NUM_RETRIES=10
+        export ALLOW_INSECURE_GIT_ACCESS=true
         # These will automatically be available in the agent-server
 
         # Auto-forwarding for Laminar:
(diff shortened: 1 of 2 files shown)


YingzuoLiu:
I noticed #14531 addresses the same issue with a very similar approach.

This PR independently fixes the same env forwarding path by adding an exact allowlist for `ALLOW_INSECURE_GIT_ACCESS` in `get_agent_server_env()`, and all checks are passing.

Happy to close this as duplicate if maintainers prefer #14531. Thanks!

PR #14547, opened 2026-05-25
Polish Jira DC webhook settings UX

ak684:
HUMAN:
This is a follow-up UX polish PR for Jira Data Center webhook settings after R02 manual testing surfaced confusing owner/admin language, stale active status, missing feedback when Jira-side webhook operations fail, and a non-installer account-linking gap.

- [x] A human has tested these latest changes.

AGENT:

## Why

Follow-up UI/UX hardening for the Jira Data Center KOTS-managed service account and connection-scoped webhook work that merged in #14562.

## Summary

- Clarify Jira DC webhook PAT copy so OpenHands does not imply it has verified the OpenHands integration owner as a Jira admin.
- Surface targeted error toasts when automatic Jira webhook install/delete fails, including OAuth callback failures after returning to the integrations page.
- Read FastAPI `detail` error responses in the shared Axios error helper.
- Keep the remove modal closed after successful removal and make paused/active state clearer.
- Route unlinked users on KOTS-managed installs through the existing Jira DC OAuth account-link flow when an active Jira DC workspace already exists, while still falling back to installer configuration when no workspace has been set up yet.

## How to Test

- `npm --prefix frontend run check-translation-completeness`
- `npm --prefix frontend run typecheck`
- `npx --prefix frontend prettier --check frontend/src/hooks/mutation/use-configure-integration.ts frontend/src/hooks/mutation/use-unlink-integration.ts frontend/src/routes/git-settings.tsx frontend/src/components/features/settings/project-management/jira-dc-integration-panel.tsx frontend/src/i18n/declaration.ts frontend/src/i18n/translation.json frontend/src/utils/type-guards.ts frontend/src/utils/retrieve-axios-error-message.ts`
- `python3 -m py_compile enterprise/server/routes/integration/jira_dc.py`
- `uv run ruff check enterprise/server/routes/integration/jira_dc.py`

## Notes

This PR is based on `main` now that #14562 has merged. It is intentionally scoped to UX copy, toasts, modal behavior, and the non-installer account-link entry point for the already-existing Jira DC link endpoint.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-ce24b95   docker.openhands.dev/openhands/openhands:ce24b95
```


ak684:
Follow-up on the bot suggestions:

- Added focused backend route coverage in `55965ec8f` for the UI-facing webhook failure contracts: non-OAuth configure returning `webhookEnrolled: false`, OAuth callback redirecting with `jira_dc_webhook=install_failed`, and unlink returning `webhookRemoved: false` while still deactivating the integration.
- I did not add frontend mutation-hook tests in this PR because these hooks do not currently have an established test harness in the frontend tree; that is a good follow-up if we add hook-level tests for project-management integrations more broadly.
- I agree the admin-PAT security model should be documented, but I would keep that in customer docs/release docs rather than embedding a docs link in short localized help text in this PR.
- The OAuth callback query parameter is intentionally scoped only to webhook install failures after an admin PAT was supplied. Other OAuth/token/profile failures remain hard callback errors because the integration was not successfully saved; this path is only for the secondary best-effort webhook install after the workspace config succeeds.

PR #14582, opened 2026-05-27
fix: retry conversation metadata save after duplicate key

leaderofagi:
HUMAN:

This PR makes conversation metadata persistence safer when resolver startup and webhook updates race to save the same conversation. It retries the SQL save after a duplicate-key integrity error so a running resolver conversation is not lost from the visible conversation list.

AGENT:

---

## Why

Fixes OpenHands/enterprise#26. Resolver startup can race with webhook persistence for the same conversation metadata row. If the concurrent writer wins first, the second save can hit a duplicate-key `IntegrityError`, causing metadata persistence to fail and started conversations to be hidden from the conversation list.

## Summary

- Catches `IntegrityError` during SQL conversation metadata save, rolls back, and retries the save as an idempotent merge/update.
- Adds a regression test covering retry behavior after a duplicate-key style integrity failure.

## Issue Number

Fixes OpenHands/enterprise#26

## How to Test

- `uv run pytest tests/unit/app_server/test_sql_app_conversation_info_service.py -q`
- `uv run ruff check openhands/app_server/app_conversation/sql_app_conversation_info_service.py tests/unit/app_server/test_sql_app_conversation_info_service.py`

## Video/Screenshots

Not applicable; backend persistence race fix.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This keeps the existing `merge` behavior, but makes the save path resilient when an insert/update race is observed at commit time.

### DIFF
--- a/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
@@ -34,6 +34,7 @@ from sqlalchemy import (
     select,
 )
 from sqlalchemy.engine import CursorResult
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import Mapped, mapped_column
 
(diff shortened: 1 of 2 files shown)


leaderofagi:
Confirmed ready for review. I added the requested human-context note to the PR description.

PR #14585, opened 2026-05-28
Vasco/seed default profile

VascoSch92:

## Summary

SaaS orgs upgrading from pre-LLM-profiles settings landed on an empty profiles UI — the LLM they'd been using wasn't carried over.

SaasSettingsStore.load() now seeds a Default profile from the merged org + member effective LLM (model, base_url, api_key) when org.llm_profiles is NULL/empty, and persists it to the column so the org-profiles management API
  (/api/organizations/{org_id}/profiles) sees the same Default the settings path does — not an empty list. The write runs under a SELECT ... FOR UPDATE row lock with an emptiness re-check, so concurrent loads can't double-seed and a profile a member just created via the management API can't be clobbered. Idempotent on later loads.
  
## How to Test
- test_load_with_null_or_empty_llm_profiles_seeds_default_profile — seed surfaces in Settings.llm_profiles for both NULL and empty-dict orgs.
- test_load_persists_seeded_default_profile_onto_org — model, base_url, and api_key all round-trip into org.llm_profiles['profiles']['Default'].

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -214,10 +214,23 @@ class SaasSettingsStore(SettingsStore):
         # never moved to the org column.
         if org.llm_profiles:
             kwargs['llm_profiles'] = org.llm_profiles
-        # Pre-migration rows read back as None; Settings.llm_profiles is
-        # non-nullable, so let the default_factory take over.
-        if kwargs.get('llm_profiles') is None:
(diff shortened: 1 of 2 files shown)


VascoSch92:

PR #14601, opened 2026-05-31
feat(jira-dc): persist and inject per-user OAuth tokens in resolver conversations

### LINKED ISSUE #14595: Jira Data Center resolver: expose per-user Jira credentials inside the conversation
## Problem

The Jira Data Center resolver lets a user trigger an OpenHands conversation from a Jira issue (via `@openhands` mention or `openhands` label), but once that conversation is running the agent has **no way to interact with Jira from inside it**. It cannot read other issues, look up related tickets, create issues, link issues, transition status, or post intermediate comments.

What the agent receives today is a static, point-in-time snapshot rendered into the initial prompt: issue key, title, description, the triggering comment, and the last ~15 comments. From `openhands/app_server/integrations/templates/resolver/jira_dc/jira_dc_new_conversation.j2`. After that, Jira is opaque to the agent.

This is a real limitation for typical workflows — “read the linked tickets”, “file a follow-up bug”, “update this issue’s status when you push the PR”, “post a progress comment before the long-running test suite” — all of which the GitHub resolver allows because the agent runs with the user’s GitHub credentials inside the sandbox.

The OAuth2 dance the integration already requires *looks* like it should solve this, but in `enterprise/server/routes/integration/jira_dc.py::jira_dc_callback` the `access_token` is read exactly once (to call `/rest/api/2/myself` and bind the Jira account `key` to the Keycloak user) and then discarded. The `refresh_token` in the response is not even read. The schema (`enterprise/storage/jira_dc_user.py`) has no token columns. So OAuth is currently used purely for identity verification, not credential acquisition.

The only Jira credential the codebase ever stores or reuses is the **workspace service-account PAT** (`jira_dc_workspaces.svc_acc_api_key`). Surfacing that to the agent would work mechanically but is the wrong answer on authorization grounds — the service account is a shared bot user that typically has broader project access than any individual end-user, so the agent would act with more authority than the person who triggered it.

Atlassian's official MCP server is also not a viable workaround for self-hosted OpenHands deployments today because the platform cannot authenticate to MCP servers via OAuth2.

Make the OAuth2 dance actually produce a usable, per-user credential, and hand it to the conversation:

1. **Persist** the Jira DC OAuth `access_token` / `refresh_token` (and expiry timestamps) on the existing `jira_dc_users` row, encrypted at rest via the existing `TokenManager.encrypt_text` pattern.
2. **Refresh** the access token on demand at conversation start, hitting Jira DC's `/rest/oauth2/latest/token` with `grant_type=refresh_token`. If the refresh token is exhausted, surface a "please re-link" message on the Jira issue (mirroring the existing session-expired flow).
3. **Inject** the resolved access token and the Jira base URL into the conversation sandbox via the documented `AppConversationStartRequest.secrets` mechanism, so the agent has `$JIRA_DC_TOKEN` and `$JIRA_DC_BASE_URL` available for `curl` against Jira's REST API.
4. **Document** the capability in the system prompt for resolver conversations so the agent knows the credential exists, how to use it, and that comment bodies must be Jira Wiki Markup.

The result: the agent acts on Jira as the human who triggered it, with that human's authorization, and the existing service account stays scoped to its current narrow responsibilities (webhook registration, end-of-conversation summary comment, acknowledgement reaction).

Technical design — schema, files touched, refresh flow, edge cases — to follow in a comment so this issue stays focused on the "what" and "why".

---

_This issue was filed by an AI agent (OpenHands) on behalf of @jpshackelford._

rsd-darshan:
HUMAN: The Jira DC OAuth callback was discarding the access and refresh tokens after a one-time identity check, leaving the resolver agent with no way to interact with Jira once a conversation started. This PR persists those tokens (encrypted at rest), refreshes them on demand, and injects them into the sandbox so the agent can make Jira REST API calls as the user who triggered the conversation. Users who linked before this ships will be prompted to re-link on their next trigger.

## Summary

The Jira DC resolver lets users trigger OpenHands conversations from a Jira issue, but once a conversation starts the agent has no way to interact with Jira — it can't read related tickets, post comments, transition status, or do anything that the GitHub resolver can do out of the box.

The root cause: the OAuth2 callback reads the `access_token` once to verify identity, then discards it. The `refresh_token` isn't even read. This PR makes the OAuth dance actually produce a usable, per-user credential and hands it to the conversation.

Fixes #14595.

## Changes

**Database**
- Migration `117`: adds four nullable columns to `jira_dc_users` — `oauth_access_token_encrypted`, `oauth_refresh_token_encrypted`, `oauth_access_token_expires_at`, `oauth_refresh_token_expires_at`. Nullable so existing rows stay valid and fall back to the re-link path.

**Token storage**
- `JiraDcUser` model updated to reflect the new columns.
- `JiraDcIntegrationStore` gains `update_user_oauth_tokens` and `get_user_oauth_tokens`.

**OAuth callback** (`jira_dc_callback`)
- After a successful token exchange, encrypts and persists the access token, refresh token, and both expiry timestamps.
- Works for both `workspace_integration` and `workspace_link` operation types.
- Tolerates a missing `refresh_token` or `expires_in` (stored as `NULL` / `0`) so deployments where the IdP omits these fields don't break.

**New module: `jira_dc_user_token.py`**
- `get_user_jira_dc_token` resolves a fresh access token: returns the stored token if still valid, otherwise hits `/rest/oauth2/latest/token` with `grant_type=refresh_token`, persists the new tokens, and returns the fresh access token.
- `JiraDcUserTokenError` is raised when no usable credential is available (no stored token, expired refresh token, or failed refresh). Callers surface this as a re-link prompt on the Jira issue rather than a silent failure.

**Conversation injection** (`JiraDcNewConversationView._create_v1_conversation`)
- Calls `get_user_jira_dc_token` before starting the conversation.
- Injects `JIRA_DC_TOKEN` and `JIRA_DC_BASE_URL` into `AppConversationStartRequest.secrets` so the sandbox agent has authenticated access to the Jira REST API as the user who triggered the conversation.

**Re-link UX** (`JiraDcManager.start_job`)
- Catches `JiraDcUserTokenError` alongside the existing `SessionExpiredError` handler.
- Posts a clear re-link message on the Jira issue: "your Jira workspace link has expired … re-link in Settings → Integrations".
- `get_jira_dc_relink_message` added to `integrations/utils.py`, following the same pattern as `get_session_expired_message`.

**Prompt template** (`jira_dc_instructions.j2`)
- Documents `$JIRA_DC_TOKEN` and `$JIRA_DC_BASE_URL` with `curl` examples.
- Includes the Jira Wiki Markup requirement for comment bodies.
- Explicitly instructs the agent never to echo the token.

## Tests

- `test_jira_dc_user_token.py` (new): fresh-token fast path, unknown-expiry treated as fresh, no-stored-token error, expired-access-no-refresh error, expired-refresh error, successful refresh with token rotation, successful refresh without token rotation (IdP keeps same refresh token), refresh HTTP failure.
- `test_jira_dc_view.py`: `_create_v1_conversation` populates `start_request.secrets` with correct values; `JiraDcUserTokenError` propagates so `start_job` can post the re-link comment.
- `test_jira_dc_integration_routes.py`: callback persists tokens for `workspace_integration` (with full token data); callback tolerates missing `refresh_token` / `expires_in` (stored as `NULL` / `0`); callback persists tokens for `workspace_link`.

## Notes for reviewers

- The agent acts on Jira **as the human who triggered the conversation**, with that human's authorization scope. The service-account PAT remains scoped to its existing narrow responsibilities (webhook registration, acknowledgement reaction, end-of-conversation comment).
- Users who linked before this ships have no stored tokens and will see the re-link prompt on their next trigger. This is intentional: fail closed rather than silently fall back to a broader service-account credential.
- `JIRA_DC_SCOPES = 'WRITE'` (already in the codebase) is now actually exercised at runtime. Consider making this an env var if read-only resolver deployments are needed.
- I can't run the full enterprise stack locally, so end-to-end validation against a live Jira DC + Keycloak instance will need a maintainer pass before merge.


rsd-darshan:
The `Enterprise Check Migrations / check-sync` failure is not related to the code in this PR.

The workflow detects that the PR contains a migration file and attempts to post a warning comment on the PR using the Actions token. Since this is a fork PR, GitHub restricts the Actions token to read-only permissions (standard GitHub security behaviour for fork-triggered workflows), so the comment POST returns a 403 and the job fails.

The fix would be on the upstream workflow side — adding `continue-on-error: true` to that step or skipping the comment when `github.event.pull_request.head.repo.fork == true`. There is no migration conflict; revision 117 is unique across all currently open PRs.

ak684:
Hi @rsd-darshan, thanks so much for this contribution! I've pulled it into a dev instance of OpenHands Enterprise, got it to a valided working state and opened this PR where I've included you as a commit author: https://github.com/OpenHands/OpenHands/pull/14650. I'm now going to close this current PR since https://github.com/OpenHands/OpenHands/pull/14650 will supersede this work but rest assured you will receive commit/PR credit. Thank you!

PR #14615, opened 2026-06-01
fix: allow self-hosted BYOR key export

### LINKED ISSUE #14610: Replicated/self-hosted: BYOR LLM key UI shows "Purchase $10 in credits" paywall because check_byor_export_enabled ignores DEPLOYMENT_MODE
## Summary

On a Replicated / self-hosted Enterprise install, fetching the OpenHands LLM key from the UI (Settings → API Keys) renders the paywall:

> Purchase at least $10 in credits to get access to OpenHands LLM key for use with OpenHands CLI and SDK.

…even though billing is intentionally disabled for that deployment (`ENABLE_BILLING=false`, `OH_WEB_CLIENT_FEATURE_FLAGS_ENABLE_BILLING=false`, etc., as set in `OpenHands-Cloud/replicated/openhands.yaml`). The BYOR (Bring-Your-Own-Runtime) LLM key gate is hard-wired to a SaaS-only Stripe code path, so on self-hosted there is no env var the customer or operator can set to unblock it.

## Impact

- Any self-hosted (Replicated) customer is blocked from retrieving the LLM API key they need for the OpenHands CLI / SDK.
- The UI surfaces a confusing "Purchase $10" call-to-action even though the deployment has no billing/Stripe at all.
- Currently the only workaround is a manual SQL update against the customer's Postgres.

## Root cause

The endpoint backing this UI is `GET /api/keys/llm/byor` in `enterprise/server/routes/api_keys.py` (around line 329). It returns `402 Payment Required` whenever `OrgService.check_byor_export_enabled(...)` returns `False`:

```python
# enterprise/server/routes/api_keys.py
if not await OrgService.check_byor_export_enabled(user_id, org_id=effective_org_id):
    raise HTTPException(
        status_code=status.HTTP_402_PAYMENT_REQUIRED,
        detail='BYOR key export is not enabled. Purchase credits to enable this feature.',
    )
```

The same check guards `POST /llm/byor/refresh` (~line 405) and `GET /llm/byor/permitted` (~line 171).

`check_byor_export_enabled` simply reads a DB column — it never consults `DEPLOYMENT_MODE`, `ENABLE_BILLING`, or any other deployment-mode signal:

```python
# enterprise/storage/org_service.py:826
@staticmethod
async def check_byor_export_enabled(user_id, org_id=None) -> bool:
    ...
    org = await OrgStore.get_org_by_id(org_id)
    if not org:
        return False
```

The column `org.byor_export_enabled` defaults to `False`:

```python
# enterprise/storage/org.py:67
byor_export_enabled: Mapped[bool] = mapped_column(nullable=False, default=False)
```

…and is only ever flipped to `True` by:

1. `enterprise/server/routes/billing.py:296` — after a successful **Stripe** checkout session, and
2. `enterprise/storage/user_store.py:198` — only in the SaaS migration path for users that already had a completed `BillingSession`.

The normal user-creation path (`UserStore.create_user`, `enterprise/storage/user_store.py:56`) does not set the column at all → DB default `False`.

Net effect: on a Replicated install (`ENABLE_BILLING=false`, no Stripe), nothing ever sets the flag, so every org permanently returns 402, so the UI permanently shows the "$10 credits" paywall.

### Why the existing env vars don't help

| Env var (set by Replicated) | Where it's read | Affects BYOR gate? |
|---|---|---|
| `ENABLE_BILLING` | `enterprise/storage/lite_llm_manager.py:36`, `enterprise/server/config.py:60` | ❌ |
| `OH_WEB_CLIENT_FEATURE_FLAGS_ENABLE_BILLING` | `enterprise/server/routes/billing.py:37` (gates `/api/billing/*` only) | ❌ |
| `OH_WEB_CLIENT_FEATURE_FLAGS_HIDE_BILLING_PAGE` | Frontend nav | ❌ |
| `REQUIRE_PAYMENT` | Read once in `enterprise/server/constants.py:104`, **never referenced anywhere else** (dead) | ❌ |
| `WEB_HOST` → `DEPLOYMENT_MODE` (`self_hosted` vs `cloud`) | `enterprise/server/constants.py:46`; consumed only by auth onboarding | ❌ |

`grep` confirms `api_keys.py`, `org_service.py`, and `org.py` reference none of the above.

## Steps to reproduce

1. Stand up an Enterprise install via Replicated using the values in `OpenHands-Cloud/replicated/openhands.yaml` (which sets `ENABLE_BILLING: false`, `OH_WEB_CLIENT_FEATURE_FLAGS_ENABLE_BILLING: false`).
2. Sign in as any user and let the personal org be created via the normal `UserStore.create_user` path.
3. Navigate to Settings → API Keys.
4. Observe the "Purchase at least $10 in credits…" paywall instead of the LLM key.
5. Hit `GET /api/keys/llm/byor` directly — response is `402` with `{"detail":"BYOR key export is not enabled. Purchase credits to enable this feature."}`.
6. `SELECT byor_export_enabled FROM org;` — all rows are `f`.

## Suggested fix

Short-circuit `OrgService.check_byor_export_enabled` to return `True` when `DEPLOYMENT_MODE == 'self_hosted'` (mirrors the existing `validate_billing_enabled` pattern in `enterprise/server/routes/billing.py:33-45`):

```python
# enterprise/storage/org_service.py
from server.constants import DEPLOYMENT_MODE
...
@staticmethod
async def check_byor_export_enabled(user_id, org_id=None) -> bool:
    # BYOR paywall is a SaaS-only concept tied to Stripe credits.
    # Self-hosted deployments have no billing, so BYOR must always be available.
    if DEPLOYMENT_MODE == 'self_hosted':
        return True
    ...  # existing logic unchanged
```

This single change clears the gate for all three BYOR endpoints (`/llm/byor`, `/llm/byor/refresh`, `/llm/byor/permitted`) and therefore removes the paywall in the UI on Replicated installs. The cloud path is unaffected.

Add a unit test in `enterprise/tests/unit/test_org_service.py` covering the `self_hosted` short-circuit (patching `storage.org_service.DEPLOYMENT_MODE`) to lock the behavior in.

## Immediate workaround for affected customer (no rebuild)

Run once against the customer's Postgres:

```sql
UPDATE org SET byor_export_enabled = TRUE;
```

This unblocks all existing orgs. Any orgs created after this still default to `FALSE`, so the code fix is still required for a durable solution.

## Affected files (reference, commit `982994c`)

- `enterprise/server/routes/api_keys.py` — `/llm/byor`, `/llm/byor/refresh`, `/llm/byor/permitted`
- `enterprise/storage/org_service.py:826` — `check_byor_export_enabled`
- `enterprise/storage/org.py:67` — `byor_export_enabled` column default
- `enterprise/storage/user_store.py:56,191` — org creation paths
- `enterprise/server/routes/billing.py:296` — the only runtime path that sets the flag to `True`
- `frontend/src/hooks/query/use-llm-api-key.ts` — maps 402 → `isPaymentRequired`
- `frontend/src/components/features/settings/api-keys-manager.tsx` — renders `LlmApiKeyPaywall`
- `frontend/src/i18n/translation.json` — `SETTINGS$LLM_API_KEY_PAYWALL_MESSAGE`
- `OpenHands-Cloud/replicated/openhands.yaml:24-30` — Replicated env-var config (does not currently influence this gate)

---

_This issue was filed by an AI agent (OpenHands) on behalf of the user while troubleshooting a Replicated customer install._

he-yufeng:
HUMAN:
This is a narrow self-hosted Enterprise fix: BYOR key export should not depend on the SaaS credit flag when the deployment is not using the cloud billing path. The cloud path still uses the existing per-org `byor_export_enabled` setting.

## Summary

Self-hosted Enterprise installs do not use the SaaS credit gate, but the BYOR LLM key endpoints still checked `org.byor_export_enabled`. That leaves self-hosted orgs stuck behind the "Purchase credits" path because nothing in a billing-disabled deployment flips that flag.

This changes `OrgService.check_byor_export_enabled()` to allow BYOR export when `DEPLOYMENT_MODE == 'self_hosted'`. The cloud path still uses the existing per-org `byor_export_enabled` flag.

Fixes #14610.

## To verify

Ran locally on Windows after rebasing onto current `origin/main`:

```powershell
python -m py_compile enterprise\storage\org_service.py enterprise\tests\unit\test_org_service.py
git diff --check origin/main..HEAD
$env:PYTHONPATH=(Resolve-Path .\enterprise).Path; python -m pytest enterprise\tests\unit\test_org_service.py -q -k check_byor_export_enabled --basetemp .tmp\pytest-14610 -p no:cacheprovider
$env:PYTHONPATH=(Resolve-Path .\enterprise).Path; python -m pytest enterprise\tests\unit\test_org_service.py -q --basetemp .tmp\pytest-14610-full -p no:cacheprovider
```

Results:
- BYOR-focused tests: `6 passed, 48 deselected`
- Full `test_org_service.py`: `54 passed`
- `py_compile` and `git diff --check`: passed

Note: I installed the declared enterprise dependency `python-keycloak==5.3.1` locally so the enterprise unit tests could import `server.auth.token_manager`.

I also checked `python -m ruff check enterprise\storage\org_service.py enterprise\tests\unit\test_org_service.py`; it reports existing full-file docstring style issues unrelated to this patch, so I left that broader cleanup out of scope.

### DIFF
--- a/enterprise/storage/org_service.py
+++ b/enterprise/storage/org_service.py
@@ -8,6 +8,7 @@ from uuid import UUID, uuid4
 from uuid import UUID as parse_uuid
 
 from server.constants import (
+    DEPLOYMENT_MODE,
     ORG_SETTINGS_VERSION,
     get_default_llm_base_url,
     get_default_llm_model,
@@ -845,6 +846,9 @@ class OrgService:
         Returns:
             bool: True if BYOR export is enabled, False otherwise.
         """
+        if DEPLOYMENT_MODE == 'self_hosted':
+            return True
+
         if org_id is None:
             user = await UserStore.get_user_by_id(user_id)
             if not user or not user.current_org_id:
--- a/enterprise/tests/unit/test_org_service.py
+++ b/enterprise/tests/unit/test_org_service.py
@@ -1874,6 +1874,26 @@ async def test_update_org_with_permissions_only_non_llm_fields(
 
 
 @pytest.mark.asyncio
+async def test_check_byor_export_enabled_allows_self_hosted_deployments():
+    with (
+        patch('storage.org_service.DEPLOYMENT_MODE', 'self_hosted'),
+        patch(
+            'storage.org_service.UserStore.get_user_by_id',
+            new_callable=AsyncMock,
+        ) as get_user_by_id_mock,
+        patch(
+            'storage.org_service.OrgStore.get_org_by_id',
+            new_callable=AsyncMock,
+        ) as get_org_by_id_mock,
+    ):
+        result = await OrgService.check_byor_export_enabled('test-user-123')
+
+    assert result is True
+    get_user_by_id_mock.assert_not_awaited()
+    get_org_by_id_mock.assert_not_awaited()
+
+
+@pytest.mark.asyncio
 async def test_check_byor_export_enabled_returns_true_when_enabled():
     """
     GIVEN: User has current_org with byor_export_enabled=True


he-yufeng:
Ready for review.

he-yufeng:
Rebased this onto the latest main and reran the focused enterprise check.\n\nValidation:\n- PYTHONPATH=enterprise OPENHANDS_SUPPRESS_BANNER=1 .venv\\Scripts\\python.exe -m pytest enterprise\\tests\\unit\\test_org_service.py -q -k byor -> 6 passed, 48 deselected\n- git diff --check origin/main..HEAD -> passed\n\nI also checked ruff E/F on the touched files. The two reported E501 lines are already present on origin/main, so I left unrelated style churn out of this PR.

jpshackelford:
## Suggested manual verification plan

For anyone picking this up before merge, here's a runnable plan that exercises the end-to-end UI / API path the issue describes:

1. Deploy this branch to a Replicated / Self-Hosted test channel. Confirm the deployment-mode signal the change keys on is set:
   ```bash
   kubectl -n openhands set env deploy/openhands --list | grep -E 'WEB_HOST|DEPLOYMENT_MODE'
   # Expect DEPLOYMENT_MODE=self_hosted (derived from WEB_HOST in enterprise/server/constants.py:46).
   ```

2. Sign in as a brand-new user (do not pre-flip `org.byor_export_enabled` in Postgres).

3. Visit **Settings → API Keys** in the UI:
   - **Before this PR**: page renders the "Purchase at least $10 in credits …" paywall card.
   - **After this PR**: BYOR key UI is visible and the "Generate key" action works.

4. Direct API check:
   ```bash
   curl -i -H "Cookie: <session>" https://<host>/api/keys/llm/byor
   # Expect HTTP 200, not 402 Payment Required.
   ```
   Repeat for the two sibling routes that share `check_byor_export_enabled`:
   ```bash
   curl -i -H "Cookie: <session>" https://<host>/api/keys/llm/byor/permitted          # expect 200
   curl -i -X POST -H "Cookie: <session>" https://<host>/api/keys/llm/byor/refresh    # expect 200
   ```

5. End-to-end use the returned key with the OpenHands CLI:
   ```bash
   openhands login --key <returned-key>
   openhands chat "hi"
   ```
   Expect a normal completion (no 402, no auth error).

6. **Negative regression** on a cloud / SaaS preview env (`DEPLOYMENT_MODE != self_hosted`):
   - Repeat steps 2-4.
   - For an org with `byor_export_enabled=False`, `/api/keys/llm/byor` should still return 402.
   - For an org where `byor_export_enabled=True` was set by the Stripe checkout path, it should return 200.

   This guards against the bypass leaking into the SaaS code path.

### Real-world "before" evidence

A recent enterprise support bundle (Replicated self-hosted install, version 0.7.37) shows a user hit `GET /api/keys/llm/byor → 402` 8+ times within a single 30-minute window while trying to add their own LLM key from Settings → API Keys. After deploying this branch on the same install, none of those 402s should occur — useful paired before/after if a fresh support bundle is taken post-deploy.

---

_This comment was posted by an AI agent (OpenHands) on behalf of @jpshackelford._

he-yufeng:
Bumping after a while. Self-hosted installs still cannot export the BYOR key even though the cloud flow allows it. This only removes that artificial gate and leaves the cloud path untouched.

neubig:

Thank you for understanding.

PR #14618, opened 2026-06-02
Local mode fixes

sgireddy-nh:
HUMAN:

AGENT:

---

## Why

Local/process runtime can lose the in-memory process sandbox registry when the app-server restarts or an agent-server process exits. Persisted conversation files can still exist on disk, but the app-server reports the sandbox as missing, causing the UI to treat the conversation as archived/read-only.

## Summary

- Persist process sandbox metadata under each local sandbox directory.
- Recover existing sandbox directories that contain persisted conversation state.
- Treat restorable dead agent-server processes as paused instead of archived, and restart them in the same sandbox directory on resume.

## Issue Number

N/A

## How to Test

- `poetry run python -m py_compile openhands/app_server/sandbox/process_sandbox_service.py tests/unit/app_server/test_process_sandbox_service.py`
- Direct recovery smoke test for a legacy persisted sandbox directory.
- Pre-commit hooks on the committed file passed during commit.

## Video/Screenshots

N/A

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This PR was updated by an AI agent (OpenHands) on behalf of the user.


sgireddy-nh:
Closing this PR because it was opened against the upstream OpenHands repository unintentionally. The user intends to keep this work in their own clone/fork instead.

This comment was created by an AI agent (OpenHands) on behalf of the user.

PR #3464, opened 2024-08-19
fix: custom runtime image won't work for go

xingyaoww:
**What is the problem that this fixes or functionality that this introduces? Does it fix any open issues?**

Fix https://github.com/OpenDevin/OpenDevin/issues/3415

---
**Give a summary of what the PR does, explaining any non-trivial design decisions**

- Fix test fixture for different runtime images
- Change `su -` to `su` to retain all the env vars and fix the go not found issue

---
**Other references**

PR #14620, opened 2026-06-03
fix(acp): route provider creds via agent_context.secrets, not acp_env

simonrosenberg:
HUMAN:
ACP agents were leaking provider API keys/base-URL through `acp_env`, a
credential channel that gets redacted to `**********` when SaaS settings are
persisted (so cloud-injected creds were silently dropped). This routes those
creds through `agent_context.secrets` instead — the same encrypted channel
every other secret uses — and stops feeding them to `acp_env`. Behaviour for
users is unchanged; it just moves the creds onto the safe channel. Part of the
ACP-on-cloud work (OpenHands/OpenHands#15722).

---

## Problem
ACP provider credentials (the UI-saved `llm.api_key` / `base_url`) reached the subprocess by being folded into **`acp_env`** inside the SDK's `create_agent()` (`resolve_acp_env` → `resolve_provider_env`). `acp_env` is a parallel, partly-unmasked, **persist-unsafe** credential channel the SDK has deprecated (OpenHands/software-agent-sdk#3464): SaaS settings stores dump `agent_settings` without cipher context, redacting `acp_env` to `**********` at persist and silently dropping cloud-injected creds.

## What this does
In `_build_acp_start_conversation_request`, route the ACP provider creds through **`agent_context.secrets`** instead of `acp_env`:
- Fold `acp_settings.resolve_provider_env()` (api key + base url, keyed by `api_key_env_var` / `base_url_env_var`) into the `secrets` dict as `StaticSecret`s, so they ride the encrypted `request.secrets` boundary like every other secret and the SDK gap-fills them into the subprocess env at launch.
- **Strip** the creds off the dummy ACP `llm` (`api_key=None`, `base_url=None`) so `create_agent()` no longer re-folds them into `acp_env`.
- Provider creds are folded in **last** so they override a same-named Secrets-panel entry — preserving the prior `provider env > panel secret` priority, now expressed within the single `agent_context.secrets` channel.
- A user's **explicit `acp_env`** entry still wins at the SDK's launch-time `acp_env > agent_context.secrets` precedence. The user-managed `acp_env` settings **field is untouched** (its editor / merge / member-private plumbing stays).

## Why it works on the pinned SDK
v1.23.1 already injects `agent_context.secrets` into the ACP subprocess env, so this is a **client-side reroute that works today** — no SDK release wait. (#3464, which removes the SDK-side `llm.api_key`→`acp_env` fold, is not yet in a release; stripping `llm` here makes the behavior correct regardless.)

## Validation
- `TestBuildAcpStartConversationRequestSecrets` — **9/9 pass** against a fresh venv on the pinned `openhands-sdk==1.23.1`.
- `ruff check` / `ruff format` clean.

## Context
Toward the registry-only end-state in OpenHands/OpenHands#15722 (remove the `acp_env` channel); part of epic OpenHands/OpenHands#15746, follows SDK #3464. Pairs with agent-canvas#1050 (stop forwarding `acp_env`) and a benchmarks change (route eval creds off `acp_env`).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-df165de   docker.openhands.dev/openhands/openhands:df165de
```

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -1593,16 +1593,20 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
     ) -> StartConversationRequest:
         """Build a StartConversationRequest for ACP agent conversations.
 
-        Unlike the LLM path, ACP agents run as separate subprocesses; we pass
-        credentials via environment variables rather than injecting an LLM object.
+        Unlike the LLM path, ACP agents run as separate subprocesses; the
+        provider credentials reach the subprocess as environment variables
+        rather than via an injected LLM object.
 
-        User secrets (Secrets panel + git provider tokens) flow through two
-        complementary channels: they're rendered into the ACP prompt as a
-        ``<CUSTOM_SECRETS>`` block via ``AgentContext.secrets`` (so the agent
-        knows the names) and also pre-exported as environment variables on
-        the ACP subprocess via ``acp_env`` (so plain CLI commands like
-        ``gh`` / ``aws`` / ``git`` pick them up from env without the agent
-        having to manually export them).
+        Both the provider credentials (the UI-saved ``llm.api_key`` /
+        ``base_url``, keyed by the provider's env-var names) and user secrets
+        (Secrets panel + git provider tokens) flow through the single
+        ``AgentContext.secrets`` channel: their names are rendered into the
+        ACP prompt as a ``<CUSTOM_SECRETS>`` block, and the SDK's
+        ``ACPAgent._start_acp_server`` gap-fills the values into the subprocess
+        env at launch. Routing the provider creds this way keeps them on the
+        cipher-protected ``request.secrets`` boundary instead of the
+        deprecated, persist-unsafe ``acp_env`` channel (software-agent-sdk
+        #3464; OpenHands/agent-canvas#1039).
 
         Args:
             sandbox: Sandbox information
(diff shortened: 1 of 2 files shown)


enyst:
@openhands Check the links in the PR description and understand well the goal of this PR. Then /codereview this PR. Respond with a review with gh api, with event, and you are free to approve if you think there are no major issues. Mention me in your final review so I can see it too.

enyst (approved):
🟢 **Taste Rating: Good taste**

@enyst I checked the PR diff and the linked context (agent-canvas#1039/#1050/#988 and software-agent-sdk#3464). The change matches the intended direction: provider credentials are folded into `agent_context.secrets`/`request.secrets`, the dummy ACP `llm` is stripped so pinned SDK 1.23.1 cannot re-fold them into `acp_env`, and explicit user-managed `acp_env` still retains launch-time precedence. The tests cover the important priority cases, including provider-over-panel and explicit-`acp_env` override behavior.

**[RISK ASSESSMENT]**
This touches credential routing for ACP conversation startup, so the impact area is security-sensitive. The implementation is small and preserves the existing effective precedence while moving provider creds onto the safer secret channel; CI is passing and the added tests exercise the core regression surface.

**VERDICT:**
✅ **Worth merging**: Core logic is sound, and I do not see major issues.

**KEY INSIGHT:**
Collapsing provider credentials into the existing secret channel removes the persist-unsafe `acp_env` special case without changing the subprocess-visible precedence model users rely on.

_This review was created by an AI agent (OpenHands) on behalf of the requester._

VascoSch92 (approved):

enyst (approved):

PR #14635, opened 2026-06-04
PLTF-2875: Add promote-to-development sender

aivong-openhands:
HUMAN:

AGENT:

---

## Why

Promote each successful `main` build's `enterprise-server` image to the development environment automatically, instead of bumping the tag by hand in `OpenHands/saas-deploy`.

Because this repo is public, it can't `uses:` the bump action that lives in the private `saas-deploy` repo, so it dispatches an event and a receiver there performs the bump. The dispatch credential is gated to a `main`-restricted Environment and is intentionally minimal in scope.

## Summary

- Adds `.github/workflows/promote-to-development.yml`: after the **Docker** workflow succeeds on a `main` push, mint a token, compute the `sha-<short>` tag, and POST a `promote-to-development` `repository_dispatch` to `OpenHands/saas-deploy`.
- Runs under the `dev-deploy` Environment (restricted to `main`); the minted token is down-scoped to `contents: write`.
- Also runnable via `workflow_dispatch` (from `main`) to re-promote a known-good tag.

## Issue Number

PLTF-2875

## How to Test

Depends on the saas-deploy receiver (OpenHands/saas-deploy#10) being on `main`, and the `dev-deploy` Environment + its secrets being configured (done).

Because `dev-deploy` is restricted to `main`, the credential is unreadable on any other ref — so this workflow can only run from `main`. After merge:

1. Manually run **Promote to development** (`workflow_dispatch`) with `image-tag` set to a real built tag, or push to `main` and let the Docker build trigger it.
2. Confirm the saas-deploy receiver run fires and bumps `.openhands.image.tag`.
3. Sync the openhands development app in ArgoCD and confirm it's healthy on the new image.

## Video/Screenshots

N/A — CI/CD workflow.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The `dev-deploy` branch restriction is what makes this safe to run from a public repo: it prevents non-`main` refs from reading the deploy credential, so the workflow can't be exercised from a feature branch — including before this merges.
- On merge, every future `main` build auto-promotes to dev; the merge commit triggers the first one.

This PR description was drafted by an AI agent on behalf of the user.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-6e67bc1   docker.openhands.dev/openhands/openhands:6e67bc1
```

### DIFF
--- a/.github/workflows/promote-to-development.yml
+++ b/.github/workflows/promote-to-development.yml
@@ -1,0 +1,86 @@
+# Promotes each successful main build's enterprise-server image to the
+# development environment by dispatching to OpenHands/saas-deploy, which runs
+# the bump. Runs after the Docker workflow succeeds on main; can also be run
+# manually (from main) to re-promote a known-good image tag.
+
+name: Promote to development
+
+on:
+  workflow_dispatch:
+    inputs:
+      image-tag:
+        description: "Image tag to promote (defaults to sha-<short SHA> for the selected ref)"
+        required: false
+        type: string
+  workflow_run:
+    workflows: ["Docker"]
+    types: [completed]
+    branches: [main]
+
+# Queue promotions so an earlier main commit deploys before a later one.
+concurrency:
+  group: promote-to-development
+  cancel-in-progress: false
+
+jobs:
+  promote:
+    # Only promote when the Docker build actually succeeded, or when run manually.
+    # The Docker workflow also runs on pull_request, and the workflow_run branches
+    # filter matches the *head* branch — a fork PR from a branch named "main"
+    # would slip through — so additionally require a push event.
+    if: ${{ github.event_name == 'workflow_dispatch' || (github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.event == 'push') }}
+    runs-on: ubuntu-24.04
+    timeout-minutes: 10
+    environment: dev-deploy
+    permissions:
+      contents: read
+    steps:
+      - name: Compute image tag
+        id: tag
+        # Matches docker/metadata-action type=sha (7-char short SHA) emitted by
+        # the docker-image-tags action used in _build-image.yml.
+        env:
+          DISPATCH_IMAGE_TAG: ${{ github.event.inputs['image-tag'] }}
+          SHA: ${{ github.event.workflow_run.head_sha || github.sha }}
+        run: |
+          if [ -n "$DISPATCH_IMAGE_TAG" ]; then
+            echo "tag=$DISPATCH_IMAGE_TAG" >> "$GITHUB_OUTPUT"
+          else
+            echo "tag=sha-${SHA::7}" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Mint token
+        id: app-token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1  # v3.2.0
+        with:
+          app-id: ${{ secrets.SAAS_DEPLOY_DISPATCHER_APP_ID }}
+          private-key: ${{ secrets.SAAS_DEPLOY_DISPATCHER_APP_PRIVATE_KEY }}
+          owner: OpenHands
+          repositories: saas-deploy
+          permission-contents: write
+
+      - name: Dispatch promotion to saas-deploy
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          IMAGE_TAG: ${{ steps.tag.outputs.tag }}
+          SOURCE_SHA: ${{ github.event.workflow_run.head_sha || github.sha }}
+        run: |
+          gh api /repos/OpenHands/saas-deploy/dispatches \
+            -f event_type=promote-to-development \
+            -F client_payload[release]=openhands \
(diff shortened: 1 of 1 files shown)


aivong-openhands:
Thanks for the review. Walking through each point — none need a change, and one is already in place:

- **(1) (extract the 7-char SHA length)** — it matches `docker/metadata-action`'s fixed short-SHA default and appears twice; parameterizing it would mean embedding a GHA expression inside bash (`${SHA::${{ env.X }}}`), which is harder to read than the constant. Leaving as-is.
- **(2) (pin vs the `v3.2.0` comment)** — the full-SHA pin is intentional and the more secure choice (immutable); `@v3.2.0` would be a mutable tag. The `# v3.2.0` comment already records the version. Keeping the SHA pin.
- **(3) (🟠 explain the fork-PR property)** — already covered: the comment directly above the `if:` (the `a fork PR from a branch named \"main\" would slip through — so additionally require a push event` lines) explains exactly this. No change.
- **(4) (`gh api` error handling)** — `gh api` already exits non-zero on HTTP 4xx/5xx and prints the response to stderr, so a failed dispatch fails the step and triggers `Annotate dispatch failure`. (`--fail` is a curl flag; not applicable to `gh api`.)
- **(5) (mixed expansion in the failure echo)** — intentional. The annotate step is the failure handler, so it can't depend on `steps.tag.outputs.tag` (empty if the Compute-tag step itself failed); `${IMAGE_TAG:-sha-${HEAD_SHA::7}}` uses the computed tag and falls back only when it's unset.

aivong-openhands:
Thanks — this review repeats the points from the earlier one, which I responded to above. Assessment unchanged, recapping briefly:

- **`gh api` error handling** — `gh api` already exits non-zero on HTTP 4xx/5xx (and prints the response to stderr), so a failed dispatch fails the step and triggers the `Annotate dispatch failure` step. `|| exit 1` would be redundant.
- **`${HEAD_SHA::7}` in the annotation** — intentional. That step is the failure handler, so it can't rely on `steps.tag.outputs.tag` (empty if the Compute-tag step itself failed); `${IMAGE_TAG:-sha-${HEAD_SHA::7}}` already prefers the computed tag and only falls back when it's unset.
- **Pin vs `v3.2.0` comment** — the full-SHA pin is the intended, more secure choice (immutable); `@v3.2.0` would be a mutable tag. The `# v3.2.0` comment records the version.
- **Extract the 7-char SHA** — it's a fixed constant matching `docker/metadata-action`'s short-SHA default; parameterizing it would mean embedding a GHA expression inside bash, which reads worse. Leaving as-is.

Appreciate the LOW risk assessment — no changes needed from this pass.

aivong-openhands:
Thanks. This overlaps the two earlier reviews; addressing the one new point and recapping the rest — no changes needed:

- **(1) `contents: read` is "insufficient for dispatch"** — this conflates two different tokens. Job-level `permissions` governs the automatic `GITHUB_TOKEN`; the dispatch is authenticated by the **minted App token**, a separate credential that carries its own `contents: write` (down-scoped at mint via `permission-contents: write`). No step here even uses `GITHUB_TOKEN`, so `contents: read` isn't insufficient — it's just least-privilege for an unused token. Keeping it. (Note `permissions: []` is malformed; the empty form is `{}`.)
- **(2) head_sha vs github.sha** — the `github.event.workflow_run.head_sha || github.sha` fallback is already correct for both triggers. Leaving as-is.
- **(3) add `actions: read`/`repo` scope** — not needed for the dispatch; adding it would broaden the token against least-privilege. Decline.
- **(4) retry logic** — over-engineering for a single fire-once POST: a failure is annotated and the next main build re-promotes. (`aws-actions/AmazonSHA256UFDS/toolkit` also isn't a real action.) Decline.
- **(5) shell expansion in the failure echo** — intentional. That step is the failure handler, so it can't rely on `steps.tag.outputs.tag` (empty if the Compute-tag step itself failed); `${IMAGE_TAG:-sha-${HEAD_SHA::7}}` prefers the computed tag and only falls back when unset.

Appreciate the LOW risk assessment.

aivong-openhands:
Thanks for the approval. On the two suggestions — both are fine to leave as-is:

- **(1) Verify the dispatch / capture the 204** — `gh api` already exits non-zero on any non-2xx, and the step runs under `bash -e`, so the `Dispatched…` line only prints on success — it's already a success-only signal. The dispatch returns an empty-body 204, so capturing `response=$(…)` would log nothing extra.
- **(2) Document/secret the `tag-path` contract** — the concern that a stale `tag-path` would "silently dispatch with invalid data" doesn't hold: the saas-deploy receiver **validates** the payload — it confirms `tag-path` resolves to an existing key (errors `Key '…' not found` otherwise) and constrains `release` to a real path segment. So a drift fails the promotion **loudly**, not silently. The coupling is also already evident from the dispatch params, and a secret for a non-sensitive path key would just add indirection. Keeping it inline.

The earlier review threads cover the other stylistic notes (SHA-length var, release-tag pin, error-message consistency) — same assessment there.

jlav (approved):

PR #14639, opened 2026-06-04
fix(app_server): only rewrite the host when mapping localhost for Docker

koriyoshi2041:
HUMAN:
This fixes a small URL rewriting bug in the app server Docker path. I checked the helper against userinfo and uppercase-host cases because the old implementation replaced the first literal `localhost` in the raw netloc instead of replacing only the parsed host.

## What

`replace_localhost_hostname_for_docker` (in `openhands/app_server/utils/docker_utils.py`) rewrites a `localhost` URL to `host.docker.internal` when running inside a container. It did this with `parsed.netloc.replace('localhost', replacement, 1)` on the raw netloc, which breaks in two cases:

1. If the userinfo contains the literal string `localhost` (e.g. a password that is exactly `localhost`), the first occurrence replaced is in the credentials, not the host. The credentials get mangled and the actual host is left as `localhost`, so the resulting URL is unusable from inside the container.
2. A host with different casing such as `LOCALHOST` passes the `parsed.hostname == 'localhost'` check (urlparse lowercases `hostname`) but is never replaced, because `str.replace` is case-sensitive. The host is silently left unchanged.

The function is used to build agent-server URLs throughout `app_server` (sandbox services, event callbacks, conversation router), some of which carry credentials in the userinfo, so a corrupted netloc means a failed connection.

## Fix

Rebuild the netloc from the parsed components — keep `username`/`password` verbatim, substitute only the host, and preserve the port — instead of doing a positional string replace on the whole netloc. This is the same `urlparse` the guard already relies on, so the host is always replaced when it matches and userinfo is never touched.

## Tests

Added two regression tests to `tests/unit/app_server/test_docker_utils.py`:
- userinfo containing `localhost` (both as password and as username) is preserved and only the host changes;
- an uppercase `LOCALHOST` host is actually replaced.

Both fail on `main` and pass with this change; the existing 13 tests in the file still pass. `ruff` and `ruff-format` (0.12.5) and `mypy` are clean on the touched files.

### DIFF
--- a/openhands/app_server/utils/docker_utils.py
+++ b/openhands/app_server/utils/docker_utils.py
@@ -26,7 +26,18 @@ def replace_localhost_hostname_for_docker(
         return url
     parsed = urlparse(url)
     if parsed.hostname == 'localhost':
-        # Replace only the hostname part, preserving port and everything else
-        netloc = parsed.netloc.replace('localhost', replacement, 1)
+        # Rebuild the netloc from its components so that only the host is
+        # changed. A naive str.replace on the raw netloc would corrupt
+        # userinfo that happens to contain 'localhost' (e.g. a password equal
+        # to 'localhost') and would miss a differently-cased host such as
+        # 'LOCALHOST', which urlparse normalizes when exposing parsed.hostname.
+        netloc = replacement
+        if parsed.port is not None:
+            netloc = f'{netloc}:{parsed.port}'
+        if parsed.username is not None:
+            userinfo = parsed.username
+            if parsed.password is not None:
+                userinfo = f'{userinfo}:{parsed.password}'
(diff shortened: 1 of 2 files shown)


koriyoshi2041:
Ready for review. I updated the HUMAN note and checked the touched app-server helper with the focused unit tests plus ruff/ruff-format/mypy on the touched files.

koriyoshi2041:
Revalidated the current head against today’s `main` before clearing the stale state. The focused helper suite still passes (`15 passed`), Ruff lint passes, Ruff formatting passes, and `git diff --check origin/main...HEAD` is clean. The branch is 245 commits behind but the one-commit diff remains conflict-free.

neubig:

Thank you for understanding.

PR #14650, opened 2026-06-04
feat(jira-dc): persist and inject per-user OAuth tokens in resolver conversations

ak684:
HUMAN:

This PR adds Jira Data Center OAuth token persistence and resolver sandbox injection so agents started from Jira can call back into Jira using the linked user credential. It also fixes the enterprise migration chain and stale Jira DC link handling found during validation.

This draft PR exists to run OpenHands-owned CI and publish validation images for the Jira DC OAuth token persistence changes from #14601.

It includes the migration fix needed for the enterprise Alembic chain and a small runtime reconciliation fix found during Replicated validation: when Jira DC OAuth is configured for a single KOTS-managed host, stale active user links from an older Jira DC host are ignored by the status endpoint and deactivated during successful OAuth relink.

Current validation image tag, once CI publishes it:

```
ghcr.io/openhands/enterprise-server:sha-9ce6243aedc9e28774048917138568bd36403e21
```

Review should still be coordinated with #14601 unless we decide to land these fixes through this branch. This branch can be closed after the validation path is complete or converted into the implementation PR if that is the cleaner route.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-731ae5d   docker.openhands.dev/openhands/openhands:731ae5d
```


ak684:
Thanks for the review. On the Jira token prompt concern: this mirrors the existing GitHub resolver pattern, where the prompt tells the agent to use `` while the actual value is supplied separately as a sandbox secret/env var. The Jira template likewise contains the literal env var name ``, not the token value. The remaining risk is that an agent with sandbox access could read and echo the env var, so the prompt explicitly tells it not to echo/log it; a server-side Jira capability/proxy would be the stronger follow-up if we want to avoid exposing Jira credentials to the sandbox at all.\n\nI pushed `d677af010` to make that distinction explicit in the Jira DC instructions, add the requested comment for OAuth-disabled host matching, and add unit coverage for `get_jira_dc_relink_message`. I also confirmed migration `117` exists in this branch and `118` correctly follows it. The `a_session_maker` patch target is the module where the store resolves the imported symbol, so it is the correct patch target; CI should re-run the enterprise test suite on the pushed commit.

enterprise/integrations/jira_dc/jira_dc_view.py
```diff
@@ -121,6 +126,20 @@ async def _create_v1_conversation(self, jinja_env: Environment):
 
         injector_state = InjectorState()
 
+        # Resolve per-user OAuth token and inject into sandbox environment.
+        # JiraDcUserTokenError is intentionally not caught here — it propagates
+        # to start_job which posts a re-link comment on the Jira issue.
+        sandbox_secrets: dict[str, SecretStr] = {}
+        token_manager = TokenManager()
+        user_token = await get_user_jira_dc_token(
```
jpshackelford: Nice that this is isolated to a single hook — makes the follow-on much easier. Two things at this injection site:

**(must fix in this PR) Legacy mode breaks here.** This call is unconditional, but deployments with `JIRA_DC_ENABLE_OAUTH=false` have no stored tokens, so every webhook will raise `JiraDcUserTokenError` and post the "re-link in Settings → Integrations" comment — which legacy-mode users can't act on, because that UI is OAuth-only. Either gate this block on `JIRA_DC_ENABLE_OAUTH` (skip the secret injection when off) or drop legacy mode in this PR with a release note. I don't think you intended to remove legacy support given the careful `_configured_jira_dc_workspace_name` handling elsewhere, so I suspect this is just a miss.

**(can defer, but please leave a marker)** Every other provider token in the codebase is delivered as a `LookupSecret` via `_setup_secrets_for_git_providers` — the sandbox hits a callback URL per call and the app server resolves the freshest token. This PR uses `StaticSecret` semantics (decrypted once at start, frozen in container env for the conversation's lifetime), which is the right call for a focused first step. The follow-on that lifts the token to the account level is the natural moment to converge with the existing pattern. Could you drop a `TODO(jira-dc-account-token):` here that names that intent? It'll keep a future reader from generalizing the static pattern into Slack/Linear before we're ready.

enterprise/integrations/jira_dc/jira_dc_user_token.py
```diff
@@ -0,0 +1,116 @@
...
+JIRA_DC_TOKEN_URL = f'{JIRA_DC_BASE_URL}/rest/oauth2/latest/token'
+
+
+class JiraDcUserTokenError(Exception):
+    """Per-user Jira DC OAuth token is missing or unrefreshable."""
+
+
+@dataclass(frozen=True)
+class JiraDcUserToken:
+    access_token: str
```
jpshackelford: Small one: could we type `access_token` as `SecretStr`? This dataclass is likely to be reused by the follow-on for the account-level path, and a plain `str` is leak-prone via `logger.debug(token_obj)`, `repr()`, pdb introspection, structured-log frame dumps, exception chaining, and so on. Cheap to lock down before it gets shared more widely — the call site in `jira_dc_view.py` would just `.get_secret_value()`.

enterprise/integrations/jira_dc/jira_dc_user_token.py
```diff
@@ -0,0 +1,116 @@
...
+    expires_at: int  # 0 if unknown
+
+
+async def get_user_jira_dc_token(
```
jpshackelford, line 36: The token-lifecycle tests below are good — they exercise the rotation-vs-no-rotation branches cleanly, which is the case most likely to actually bite us. Flagging a related concern about that same path: the read → refresh → write sequence has no concurrency control. Two concurrent webhooks for the same user can both observe "expired" at line 59, both call the IdP at line 76, and on IdPs that rotate refresh tokens on use the second caller will then write a now-invalid refresh token back to the DB — permanently breaking the link until the user re-OAuths.

Today the blast radius is one user on a Jira webhook burst. Once the follow-on exposes this token to Slack/GitHub/Linear conversations as well, every integration becomes a potential racer for the same row. The clean fix is a transactional `SELECT FOR UPDATE` (or a Postgres advisory lock keyed on `(keycloak_user_id, workspace_id)`). If you'd rather defer that, could you leave an explicit `TODO(jira-dc-refresh-race):` here that names the IdP-rotation case? It'll keep the follow-on author from widening the blast radius without seeing the marker.

enterprise/integrations/jira_dc/jira_dc_user_token.py
```diff
@@ -0,0 +1,116 @@
...
+                response.status_code,
+                response.text[:200],
+            )
+            raise JiraDcUserTokenError(
+                'Jira DC token refresh failed. '
+                'Please re-link via OpenHands Settings → Integrations.'
+            )
+        data = response.json()
+
+    new_access = data['access_token']
```
jpshackelford: The error messages elsewhere in this file are notably user-friendly — worth extending that same care here. A 200 response without `access_token` (rare but seen behind misconfigured proxies) raises `KeyError`, which falls through `JiraDcManager.start_job`'s generic `except Exception` and surfaces the unhelpful "unexpected error" comment on the Jira issue instead of the nice re-link prompt your other branches produce. Something like:

```python
new_access = data.get('access_token')
if not new_access:
    raise JiraDcUserTokenError(
        'Jira DC token refresh returned no access_token. '
        'Please re-link via OpenHands Settings → Integrations.'
    )
```

enterprise/storage/jira_dc_integration_store.py
```diff
@@ -236,6 +261,59 @@ async def deactivate_workspace(self, workspace_id: int):
...
+    ) -> None:
+        """Persist updated OAuth tokens on the user's active workspace link."""
+        async with a_session_maker() as session:
+            async with session.begin():
+                await session.execute(
+                    update(JiraDcUser)
+                    .where(
+                        JiraDcUser.keycloak_user_id == keycloak_user_id,
+                        JiraDcUser.jira_dc_workspace_id == workspace_id,
+                        JiraDcUser.status == 'active',
```
jpshackelford, line 282: This UPDATE silently no-ops when zero rows match — no `rowcount` check, no log, no raise. Today the WHERE clause matches in both OAuth callback branches (link defaults to `status='active'` in `create_workspace_link`, and the `status='active'` filter is load-bearing given the partial unique index on `keycloak_user_id` in `storage/jira_dc_user.py` — multiple historical inactive rows for the same `(user, workspace)` are legal, so matching by the tuple alone would update an arbitrary one). The brittleness is forward-looking: once the follow-on widens the producers of this row, a silent-success path will eventually mask a regression. A one-line `if result.rowcount == 0: raise JiraDcUserTokenError(...)` (or at minimum a `logger.warning`) is the cheap defensive guard.

jpshackelford (changes requested):
Thanks for taking this on — opening up "integration X can act through integration Y's credentials" is genuinely new ground for us, and this PR starts narrow (Jira DC resolver only) while leaving room for the follow-on that lifts the token to the account level so Slack/GitHub/Linear-triggered conversations can also call Jira. My understanding is that this PR injects `JIRA_DC_TOKEN` only on conversations started by the Jira DC view; the follow-on is what makes the Slack→Jira flow you described in #14601 actually work end-to-end. If that's wrong, let me know and I'll re-review.

Several things I appreciated:
- The migration's `_existing_columns()` guard is exactly the right shape for an enterprise rollout.
- The `expires_at == 0` ⇒ "treat as fresh" handling is a thoughtful nod to IdPs that omit `expires_in`.
- The factory's comment about always creating a fresh conversation per mention is a model for "why" comments.
- The token refresh tests cover the rotation-vs-no-rotation IdP behavior, which is the case most likely to bite.
- Quick turnaround on the bot review feedback in `d677af0` was great.

My ask is bounded by what's risky to ship as-is vs. clean up in the follow-on.

**Address in this PR** (bugs and brittleness on the path the PR itself supports):
1. Legacy `JIRA_DC_ENABLE_OAUTH=false` mode currently breaks (inline comment in `jira_dc_view.py`).
2. Defensive: silent no-op on the OAuth-token UPDATE path (inline comment in `jira_dc_integration_store.py`).
3. `KeyError` on malformed IdP refresh response leaks an unhelpful error to the Jira issue (inline comment in `jira_dc_user_token.py:97`).
4. Either single-flight the token refresh or mark the race with a `TODO` — the blast radius widens once the follow-on lets more integrations hit the same row (inline comment in `jira_dc_user_token.py:46`).
5. `JiraDcUserToken.access_token` → `SecretStr`, since this dataclass will outlive the PR (inline comment in `jira_dc_user_token.py:31`).
6. A `TODO(jira-dc-account-token):` marker at the secret-injection site pointing at the follow-on, so a future reader doesn't generalize the `StaticSecret` pattern (folded into comment #1).

**Happy to defer to the follow-on** — flagging here so they don't get lost: `StaticSecret` → `LookupSecret` migration, hoisting tokens off the per-workspace-link row, deduplicating `JIRA_DC_TOKEN_URL` and moving it off import-time evaluation, host check at webhook entry, orphan `JiraDcConversation` row on token-error path, migration-118 backfill story in release notes.

Marking as request-changes because items 1 and 3 are silent functional bugs (legacy mode breaks; the friendly re-link UX you built is bypassed on malformed IdP responses). Once those are addressed I'm happy to re-review and ship — excited to see this land.

_This review was posted by an AI agent (OpenHands) on behalf of the reviewer._

enterprise/integrations/jira_dc/jira_dc_view.py
ak684: Addressed in 85f906e1f. I gated the Jira DC OAuth secret injection on `JIRA_DC_ENABLE_OAUTH`, and also made the resolver instructions omit the `$JIRA_DC_TOKEN` REST API section when OAuth is disabled so legacy email-match deployments are not told to use a missing env var. I also added the requested `TODO(jira-dc-account-token)` marker at this injection site.

enterprise/integrations/jira_dc/jira_dc_user_token.py
ak684: Agreed on the race. I am deferring the actual single-flight/row-lock implementation to the account-level follow-on, but added `TODO(jira-dc-refresh-race)` in 85f906e1f with the IdP refresh-token-rotation failure mode called out explicitly.

enterprise/storage/jira_dc_integration_store.py
ak684: Addressed in 85f906e1f. `update_user_oauth_tokens` now returns the UPDATE rowcount, and the refresh path raises `JiraDcUserTokenError` with re-link guidance when no active row is updated. Added unit coverage for the zero-row update path.

enterprise/integrations/jira_dc/jira_dc_user_token.py
ak684: Addressed in 85f906e1f. The refresh path now uses `data.get("access_token")` and raises `JiraDcUserTokenError` with the re-link guidance when a 200 response omits the access token, instead of leaking a generic `KeyError` path. Added unit coverage for that malformed IdP response.

enterprise/integrations/jira_dc/jira_dc_user_token.py
ak684: Addressed in 85f906e1f. `JiraDcUserToken.access_token` is now `SecretStr`, and the sandbox injection call site unwraps it with `.get_secret_value()` only when creating the sandbox secret.

enterprise/integrations/jira_dc/jira_dc_user_token.py
ak684: Adding a bit more rationale for deferring the full locking implementation here.

This PR only exposes the Jira token to one resolver path today: Jira DC webhook -> one Jira-started conversation. The race is real, but its current blast radius is narrow: concurrent Jira mentions for the same linked user/workspace during an expired-token window.

The proper fix is not just a small guard. It likely needs one of:

- `SELECT ... FOR UPDATE` around token read/refresh/write in the DB session, which means changing the store API shape so the read and write happen in the same transaction.
- A Postgres advisory lock keyed by `(keycloak_user_id, workspace_id)`.
- A shared token capability/LookupSecret service where every trigger path resolves the token through one server-side refresh path.

That third option is the likely follow-on anyway for Slack/GitHub/Linear -> Jira. Once token use becomes account-level and multiple integrations can hit the same Jira credential, the refresh path should be centralized. If we solve locking now only inside this resolver helper, we risk creating code that gets bypassed or refactored immediately in the shared capability work. So this PR leaves the explicit `TODO(jira-dc-refresh-race)` marker and keeps the actual concurrency primitive scoped to the follow-on that broadens token access.

ak684:
@jpshackelford could you take another look when you have a chance?

I addressed the requested-change items on this base PR:

- gated Jira DC token injection when Jira DC OAuth is disabled
- made zero-row token updates explicit instead of silently no-oping
- converted malformed refresh responses into `JiraDcUserTokenError` with re-link guidance
- changed `JiraDcUserToken.access_token` to `SecretStr`
- added the requested refresh-race TODO/rationale

I also opened #14697 as the stacked follow-on. It moves Jira DC token exposure into the shared conversation secret enricher so web, Slack, API, and Jira-started conversations can all use the linked Jira DC OAuth token. I've successfully validated the Slack -> OpenHands -> Jira DC data flow end-to-end on replicated-02 with a real Jira Data Center instance.

My read: this PR still carries the foundational token persistence/refresh layer, while #14697 carries the cross-trigger architecture and validated Slack flow.

jpshackelford (approved):
Great work on this — excited to see it land. The five things I was most concerned about are all addressed:

- **Legacy mode no longer breaks at the token-grab site.** Gated on `JIRA_DC_ENABLE_OAUTH` in `jira_dc_view.py:133`, and the prompt template `jira_dc_instructions.j2` now hides the `$JIRA_DC_TOKEN`/REST section behind the same flag so legacy-mode users aren't told to use a variable that doesn't exist. Covered by `test_create_v1_conversation_skips_jira_dc_token_when_oauth_disabled`.
- **The OAuth-token UPDATE is no longer silent.** `update_user_oauth_tokens` returns `rowcount`, and `get_user_jira_dc_token` raises `JiraDcUserTokenError` with the re-link message on a zero-row write. New test `test_raises_when_refreshed_token_update_matches_no_rows`.
- **Malformed IdP refresh responses produce the friendly re-link path.** `data.get('access_token')` + explicit `JiraDcUserTokenError`, with `update_user_oauth_tokens.assert_not_called()` in the new test asserting we don't half-write either.
- **`JiraDcUserToken.access_token` is now `SecretStr`,** unwrapped only at the sandbox-secret boundary via `.get_secret_value()`.
- **Refresh-race `TODO(jira-dc-refresh-race)`** is in place at `jira_dc_user_token.py:47–50`, with the IdP-rotation failure mode named explicitly. The deferral rationale you added downthread is the right call — the proper fix belongs with the shared refresh service, not with this resolver-scoped helper.

Plus the `TODO(jira-dc-account-token)` marker at the secret-injection site naming the LookupSecret convergence intent — exactly the marker I asked for.

From the deferred items in my top-level review, I see the following already reflected in #14697:

- **A1 — LookupSecret convergence.** `JIRA_DC_TOKEN` is now delivered via `LookupSecret` against `/integration/jira-dc/secrets/token`, with the endpoint reusing `get_user_jira_dc_token` so every sandbox callback gets the freshest token. That subsumes the `StaticSecret` shape this PR introduced.
- **B7 — Host check at the credential boundary.** `_workspace_matches_configured_jira_dc_host` is now enforced both in the conversation secret enricher (before any injection) and at the LookupSecret endpoint (on every callback). That actually closes the trust-boundary asymmetry I flagged in a stronger place than I originally suggested — the host gate sits downstream of every trigger source rather than at each entry point.

The handful still outstanding — which we already agreed could land later:

- **B5 — duplicate import-time `JIRA_DC_TOKEN_URL` bindings** in `jira_dc_user_token.py` and `routes/integration/jira_dc.py`. Still stale on a KOTS-driven `JIRA_DC_BASE_URL` rotation. Natural to fix on #14697 since that PR is now the second consumer of the same URL via the new `/secrets/token` route.
- **Item #4 race itself.** The TODO is great for this PR; once #14697 widens the caller pool to Slack/web/API, the racer-pool widening I warned about is actually realised, so #14697 is the right place to either implement the lock or sharpen the TODO into a tracked issue.
- **B8 — operator-facing release note** flagging that every existing OAuth-linked user re-links on first webhook after upgrade. Process-only item.

Once again, awesome work here. This is going to get immediate use by our customers.

PR #14658, opened 2026-06-04
PLTF-2899: await SaaS get_user_id() in Slack repo-inference log

### LINKED ISSUE #14655: [enterprise] Slack resolver: drops partial webhook payloads (slack_error_no_view) + un-awaited get_user_id() coroutine
## Summary

Two robustness bugs in the Slack resolver (`enterprise/integrations/slack/`) observed in production:

1. **Partial/malformed Slack webhook payloads are not handled gracefully.** When a payload is missing expected fields, `SlackMessageView.from_payload` logs `slack_message_view_from_payload_missing_fields`, no view is built, and `handle_slack_error` (`slack_manager.py:522`) bails out with `slack_error_no_view`. The triggering event is silently dropped.

2. **Un-awaited coroutine** at `slack_manager.py:609`: an `async` method is called without `await` inside a log f-string, producing `RuntimeWarning: coroutine 'SaasUserAuth.get_user_id' was never awaited`. The user id is never resolved, so the log line is wrong and the coroutine never runs.

### (1) `slack_error_no_view` on partial payloads

Observed sequence:
```
WARNING from_payload         slack_message_view_from_payload_missing_fields
ERROR   handle_slack_error   slack_error_no_view
```
`handle_slack_error` is at `enterprise/integrations/slack/slack_manager.py:522`. There is no `exc_info` — it is a handled error path, but the inbound event is dropped with no actionable diagnostics about which fields were missing.

**Suggestion:** log the offending payload keys (sanitized) at `from_payload` so these are debuggable, and/or short-circuit known non-actionable Slack event subtypes earlier so they don't reach the view-construction path.

### (2) Un-awaited `get_user_id()`

`enterprise/integrations/slack/slack_manager.py:607-610`:
```python
logger.info(
    f'[Slack] Verifying inferred repo "{inferred_repo}" '
    f'for user {user.slack_display_name} (id={slack_view.saas_user_auth.get_user_id()})'
)
```
`SaasUserAuth.get_user_id()` is `async`; calling it without `await` yields a coroutine object (logged literally) and emits a `RuntimeWarning`.

**Suggestion:** `await` it before formatting (resolve into a local var), or drop the id from the log line if it is not needed.

## Impact

- Some Slack-triggered events are dropped without clear diagnostics.
- Log noise / incorrect log content; the un-awaited call is a latent bug if that value is ever relied upon.

## Notes

Both are in `enterprise/integrations/slack/`. These are distinct from the resolver "conversation start 500" issue (filed separately against runtime-api).

aivong-openhands:
## Why

`SaasUserAuth.get_user_id()` is `async`, but in the Slack repo-inference path it was called **without `await`** inside a log f-string. That logged a coroutine object instead of the id and emitted `RuntimeWarning: coroutine 'SaasUserAuth.get_user_id' was never awaited` on every Slack repo-verification.

Fixes #14655.

## Summary

- `await` `get_user_id()` into a local before formatting the log line in `_try_verify_inferred_repo` (`enterprise/integrations/slack/slack_manager.py`).

## Issue Number

#14655

## How to Test

- Trigger a Slack message that infers a single repo and confirm the `[Slack] Verifying inferred repo ...` log line shows a real user id and no `coroutine ... was never awaited` RuntimeWarning is emitted.

## Type

- [x] Bug fix

## Notes

- The other half of #14655 (partial Slack payloads logged as `slack_error_no_view`) already has good diagnostics at `slack_types.py` `from_payload` (it logs `has_team_id`/`has_channel_id`/`has_user_id` and `payload_keys`). This PR intentionally scopes to the concrete un-awaited-coroutine bug; deciding whether certain partial payloads should be silently ignored vs. surfaced is left for follow-up on the issue.

---
_This was created by an AI assistant._

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-6dd2efc   docker.openhands.dev/openhands/openhands:6dd2efc
```

### DIFF
--- a/enterprise/integrations/slack/slack_manager.py
+++ b/enterprise/integrations/slack/slack_manager.py
@@ -604,7 +604,8 @@ class SlackManager(Manager[SlackViewInterface]):
             return False
 
         inferred_repo = inferred_repos[0]
-        user_id = await slack_view.saas_user_auth.get_user_id()
+        user_id: str | None = await slack_view.saas_user_auth.get_user_id()
+        # Fixes #14655
         logger.info(
             f'[Slack] Verifying inferred repo "{inferred_repo}" '
             f'for user {user.slack_display_name} (id={user_id})'
--- a/enterprise/integrations/slack/slack_manager.py
+++ b/enterprise/integrations/slack/slack_manager.py
@@ -604,9 +604,10 @@ class SlackManager(Manager[SlackViewInterface]):
             return False
 
         inferred_repo = inferred_repos[0]
+        user_id = await slack_view.saas_user_auth.get_user_id()
         logger.info(
             f'[Slack] Verifying inferred repo "{inferred_repo}" '
-            f'for user {user.slack_display_name} (id={slack_view.saas_user_auth.get_user_id()})'
+            f'for user {user.slack_display_name} (id={user_id})'
         )
 
         try:


aivong-openhands:
@openhands address the following from the review above:

Consider adding a type annotation: The new user_id variable would benefit from an explicit type hint (e.g., str) if the return type of get_user_id() is known and documented.

Test coverage: Verify that the fix works by confirming the RuntimeWarning no longer appears during Slack repo-verification. If there's an existing test for this code path, it might be worth running it.

Consistency check: The PR description mentions this is a fix for issue https://github.com/OpenHands/OpenHands/issues/14655. Consider adding a link reference in the code comments (e.g., # Fixes #14655) if that's the project's convention.

aivong-openhands:
Addressed the review feedback:

1. **Type annotation**: Added `str | None` type hint to the `user_id` variable, matching the return type of `get_user_id()`.

2. **Fix reference**: Added `# Fixes #14655` comment for consistency with project conventions.

3. **Test coverage**: Ran the existing Slack integration tests (`enterprise/tests/unit/integrations/slack/`), and all 18 tests pass.

All changes pushed to commit `6dd2efcd0`.

malhotra5 (approved):
Odd that type checking didn't catch this, LGTM!

aivong-openhands:
> Odd that type checking didn't catch this, LGTM!

We should probably revisit the mypy setup! @openhands create an issue to investigate

aivong-openhands:
Created issue #14659 to track investigating mypy async/await static analysis improvements: https://github.com/OpenHands/OpenHands/issues/14659

PR #14660, opened 2026-06-04
PLTF-2899: keep resume flow responsive during rate limits

### LINKED ISSUE #14656: Resume-on-reopen of idle conversations stalls for minutes due to 429 rate-limiting (appears stuck)
## Summary

Reopening a conversation whose sandbox has been paused after idle can leave it **appearing stuck for several minutes** because the resume path gets **HTTP 429 rate-limited**. The user sees no progress and tends to give up / close the tab.

Observed sequence on reopen of an idle conversation (timestamps relative):

```
T+0       GET  /conversations/<id>                      200   (page load)
T+~12s    POST /api/v1/sandboxes/<id>/resume            429   (rate limited)
T+~12s    GET  /api/v1/app-conversations?ids=<id>       429   (rate limited)
...       (client keeps polling; repeated 429s)
T+~2m10s  POST /api/v1/sandboxes/<id>/resume            200   (resume finally succeeds)
T+~2m10s  "Updated session_api_key_hash for sandbox <id> after resume"
```

So the resume *does* eventually succeed, but only after ~2 minutes of 429s — long enough that the conversation looks broken.

## Impact

- User-facing: reopening an idle conversation feels stuck/broken for minutes.
- This is the failure mode behind a real user report ("conversation seemed stuck for a few minutes, I closed the page").
- Affects the resume-on-reopen flow generally, independent of how the conversation was created (UI or integration-triggered).

## Likely cause

The rate limiter covering `POST /api/v1/sandboxes/{id}/resume` and `GET /api/v1/app-conversations` is too aggressive for the burst of calls a client makes when reopening an idle conversation (resume + status polling fire together). Resume is a user-initiated, latency-sensitive action and shouldn't be throttled into a multi-minute stall.

## Suggested directions

- Treat `resume` as high-priority / exempt it from the general read rate limit, or give it a dedicated, more permissive bucket.
- Ensure the client backs off sanely on 429 rather than hammering (which can prolong the throttle window).
- Scope limits per-user/per-conversation rather than a shared global bucket, so one busy tenant can't throttle resume for others.
- Consider surfacing a clear "resuming…" state in the UI instead of silent retries.

Rate limits are now env-configurable (#14628), but the defaults/scoping on the resume path still produce multi-minute stalls in practice.

## Notes

This was identified while triaging an integration-reported "stuck conversation"; it is distinct from the resolver/agent-server issues filed separately and is the actual cause of that specific user's experience.

aivong-openhands:
HUMAN:

AGENT:

---

## Why

Reopening an idle conversation can look broken because the sandbox resume request and related conversation polling can hit HTTP 429s. The resume eventually succeeds, but the user may see a stale or stopped-looking page for long enough to give up.

This PR makes the resume path feel alive: the backend does not throttle the user-initiated resume endpoint, and the frontend respects `Retry-After` when conversation polling is rate-limited.

## Summary

- Exempt `POST /api/v1/sandboxes/{id}/resume` from the shared app-server rate limiter.
- Retry active conversation fetches on HTTP 429 using the server-provided `Retry-After` delay.
- Optimistically mark the active conversation as `STARTING` while resume is pending or retrying so the UI shows progress instead of a stale stopped state.

## Issue Number

Fixes #14656

## How to Test

```bash
poetry run pytest tests/unit/server/test_middleware.py -q
cd frontend && npm run test -- __tests__/hooks/query/use-user-conversation.test.tsx __tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
cd frontend && npm run lint:fix
cd frontend && npm run build
cd frontend && npm run typecheck
poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files openhands/app_server/middleware.py tests/unit/server/test_middleware.py
```

## Video/Screenshots

Not included. This is covered by focused backend and frontend regression tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

`npm run lint:fix` passed with existing warnings outside this change.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-83afcad   docker.openhands.dev/openhands/openhands:83afcad
```


aivong-openhands:
Thanks for the thorough review! Quick notes on the suggestions:

1. **Header check order** — this is already the implementation. `getHeaderValue` checks for `.get` first (the axios common case) and only falls back to `Object.entries` when `.get` isn't present, so the iteration doesn't run on the hot path.

2. **Consolidating the middleware exemption** — leaving `is_rate_limited_request` in its current early-return form on purpose. The `# Put Other non rate limited checks here` placeholder is there so future exemptions can be added without collapsing the logic into a single boolean expression.

3. **Query-level retry vs. optimistic state** — agreed it's worth keeping an eye on, but it's theoretical at the moment (no repro), so leaving as-is for now.

4. **E2E retry test** — the retry-after behavior is covered at the unit level with fake timers; happy to add a fuller integration test in a follow-up if it proves valuable.

aivong-openhands:
Can have a successful conversation with openhands:
Screenshot 2026-06-04 at 5 37 58 PM

aivong-openhands:
@openhands apply these suggestions from the review above:
Consider testing the edge case where retry-after is a negative number (the current code uses seconds >= 0 which handles this)
The mutation key change from "start-conversation" to "unified-resume-conversation-sandbox" could affect cache compatibility if there's existing state

aivong-openhands:
Applied both review suggestions in commit `9f656ddf3`:

1. **Negative Retry-After edge case**: Added tests for negative (`"-5"` → 0ms), zero (`"0"` → 0ms), and non-numeric (`"invalid"` → 1000ms) values. The `seconds >= 0` check correctly handles negative values - they fall through to Date.parse() which treats them as already-expired timestamps.

2. **Mutation key change**: Added a comment explaining that TanStack Query mutation keys only track mutation state (loading, error) and don't affect the data cache. The change is safe and intentional.

frontend/src/utils/rate-limit-retry.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  return typeof matchingHeader === "string" ? matchingHeader : undefined;
+}
+
+export function isRateLimitError(error: unknown): boolean {
+  const axiosError = error as AxiosError | undefined;
+  return axiosError?.response?.status === 429 || axiosError?.status === 429;
+}
+
+export function getRateLimitRetryDelayMs(error: unknown): number {
```
VascoSch92, line 39: The method returns `seconds * 1000` or `retryAt - Date.now()` with no clamp. 

Your own backend always sends `Retry-After: 1`, so this is fine in practice.

But a misbehaving proxy/CDN returning a large value would freeze conversation polling for that duration, up to 3×. 

A defensive clamp (e.g. `Math.min(delay, 60_000)`) would bound worst-case behavior.

openhands/app_server/middleware.py
```diff
@@ -132,5 +131,15 @@ async def dispatch(
     def is_rate_limited_request(self, request: StarletteRequest) -> bool:
         if request.url.path.startswith('/assets'):
             return False
+        if self.is_sandbox_resume_request(request):
+            return False
         # Put Other non rate limited checks here
```
VascoSch92:
```suggestion
def is_rate_limited_request(self, request: StarletteRequest) -> bool:
    if request.url.path.startswith('/assets') or self.is_sandbox_resume_request(request):
        return False
# Put Other non rate limited checks here
```

nit: style

openhands/app_server/middleware.py
```diff
@@ -132,5 +131,15 @@ async def dispatch(
...
         return True
+
+    def is_sandbox_resume_request(self, request: StarletteRequest) -> bool:
```
VascoSch92: You can use an anchored match

```python
import re
...
_RESUME_RE = re.compile(r'^/api/v1/sandboxes/[^/]+/resume$')
...
return request.method == 'POST' and bool(_RESUME_RE.match(path))
```

VascoSch92: Is this method private? Or it is used somewhere else? If is not used someqhere else: consider adding _*

VascoSch92 (approved):
A couple of comments. But not blockers

frontend/src/utils/rate-limit-retry.ts
aivong-openhands: @openhands use the defensive clamp proposed instead

openhands/app_server/middleware.py
aivong-openhands: Wow the agent generated some inefficient code. @openhands consider the suggestion and improve on it further as we should not be returning True or False booleans. It is a code smell. We can return the conditional expression instead

openhands/app_server/middleware.py
aivong-openhands: @Openhands use anchored match

aivong-openhands: @openhands if this method is only used in this file, make it a private method by convention by prefixing with _*

openhands/app_server/middleware.py
aivong-openhands: Addressed in commit 49a3f5911: simplified using conditional expression to return `not (path.startswith or is_sandbox_resume_request)` directly.

openhands/app_server/middleware.py
aivong-openhands: Applied your suggestion in commit 49a3f5911: now using conditional expression directly instead of multiple return statements.

aivong-openhands: Fixed in commit dcb91c760: renamed to `_is_sandbox_resume_request` to follow Python convention for private methods since it is only used within this file.

aivong-openhands:
@VascoSch92 I have addressed all the feedback! Can you help me merge?

PR #14663, opened 2026-06-04
chore: remove internal reference from migration 117 docstring

aivong-openhands:
Docstring-only change: removes an internal-tracker reference that shouldn't be in this public repo. No behavior change — the migration is unchanged.

_Generated by an AI assistant._

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-586cc6c   docker.openhands.dev/openhands/openhands:586cc6c
```


neubig:

Thank you for understanding.

PR #14666, opened 2026-06-05
fix(auth): don't revoke offline_tokens on bearer-auth failures or bro…

chuckbutkus:
Two related fixes that previously allowed a single transient Keycloak failure (or a browser logout click) to permanently brick every API key minted for a user.

Fix 1: SetAuthCookieMiddleware._logout used to call token_manager.logout on any AuthError raised during the request. For *bearer-token* (API key) requests, SaasUserAuth.refresh_token is the user's stored *offline_token* loaded from OfflineTokenStore. Calling Keycloak logout with that value revokes the offline session — so a transient KeycloakConnectionError, an assert in _is_token_expired, or any other exception inside saas_user_auth_from_bearer would become a BearerTokenError, which SetAuthCookieMiddleware caught and turned into a *destructive* Keycloak logout. From that point on every API-key call 401s until the user re-authenticates through the browser (/keycloak/callback rewrites offline_token).

The fix gates _logout to requests that actually carried a keycloak_auth cookie, and additionally guards inside _logout itself by requiring user_auth.auth_type == AuthType.COOKIE (defense in depth).

Fix 2: /api/logout resolved user_auth via get_user_auth(), which checks bearer auth before cookie auth. A browser logout click that happened to carry both an Authorization: Bearer <api-key> header and a keycloak_auth cookie would log out the offline_token, not the cookie session. The fix narrows the Keycloak logout call to AuthType.COOKIE resolutions; the cookie itself is still always deleted from the response.

Tests:
  - test_middleware_bearer_only_auth_error_does_not_revoke_offline_token
  - test_middleware_cookie_auth_error_still_triggers_logout
  - test_logout_skips_keycloak_for_bearer_auth (defense-in-depth)
  - test_logout_invokes_keycloak_for_cookie_auth
  - test_logout_with_bearer_auth_does_not_revoke_offline_token
  - test_logout_with_cookie_auth_terminates_keycloak_session
  - existing test_offline_token_survives_user_logout mock updated to explicitly mark the simulated cookie session as AuthType.COOKIE, which is closer to the real /api/logout flow it simulates.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-5822e70   docker.openhands.dev/openhands/openhands:5822e70
```


enterprise/server/routes/auth.py
```diff
@@ -1075,10 +1075,25 @@ async def logout(request: Request):
         samesite=get_cookie_samesite(),
     )
 
-    # Try to properly logout from Keycloak, but don't fail if it doesn't work
+    # Try to properly logout from Keycloak, but don't fail if it doesn't work.
```
malhotra5, line 1078: IMO this endpoint should be able to perform logout via API key on this route

malhotra5 (approved):
LGTM overall

Just one caveat is that the user should still be to perform an explicit logout, using API based authentication, via the `/logout` API route. It seems like we'd no-op the logout route unless the request has cookie based auth

enterprise/server/routes/auth.py
chuckbutkus: When using the API key, we use the keycloak offline token to generate a new access token. These access tokens are different than ones created via a browser based login in that they don't create a regular keycloak session, so there is no reason to logout.

The flow looks like:

User gets an offline refresh token.
Application uses the offline refresh token to obtain an access token.
Application uses the access token to call APIs.
Access token expires (typically in minutes).

That access token is not associated with a newly created online user session, so there is usually no session to terminate. You can simply stop using the access token and let it expire.

PR #14668, opened 2026-06-05
refactor(acp): remove acp_env wholesale-replacement compat code

simonrosenberg:
## Summary

- Removes the ``acp_env`` wholesale-replacement guard from ``Settings.update`` in ``settings_models.py``. With ``acp_env`` deprecated (sdk v1.25.0, #3464), the pop-and-reinsert logic that prevented stale keys from surviving saves is dead code.
- Removes ``'acp_env'`` from ``WHOLESALE_REPLACEMENT_KEYS`` in ``jsonpatch_compat.py`` — the constant now only covers ``mcp_config``.
- Removes the manual provider-cred injection workaround from ``_build_acp_start_conversation_request``. ``create_agent()`` (sdk#3464) now folds ``llm.api_key`` / ``base_url`` into ``agent_context.secrets`` itself, so the manual ``resolve_provider_env()`` loop and the llm-strip that guarded against double-injection are gone. Priority flips from provider > panel to panel > provider, matching the SDK's merge order.
- Drops the now-dead ``acp_env`` tests across four test files.

No production users set ``acp_env`` directly (deprecated in openhands-sdk v1.25.0, already pinned here via #14653). Supersedes #14621.

## Test plan

- [x] ``pytest tests/unit/app_server/test_live_status_app_conversation_service.py::TestBuildAcpStartConversationRequestSecrets`` — 9 passed
- [x] ``pytest tests/unit/app_server/test_settings_agent_kind_switch.py tests/unit/app_server/utils/test_jsonpatch_compat.py`` — 15 passed

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-ce9259d   docker.openhands.dev/openhands/openhands:ce9259d
```


simonrosenberg:
Fixed in 63b666b — two enterprise tests for `acp_env` wholesale replacement deleted since that behaviour was removed in the parent commit.

PR #14640, opened 2026-06-04
fix(acp): bootstrap-prompt resume when sandbox is recycled (Solution A)

simonrosenberg:
HUMAN:

Resume ACP recycled sandbox conversation by creating new acp session and adding a summary of previous acp events in first message. 

## Why

ACP conversations can't be resumed after a cloud sandbox recycle. The ACP server's own session storage (`~/.claude/projects`, `~/.codex/sessions`, `~/.gemini/*`) lives in the sandbox FS and is wiped; the SDK wires `session/load` but there's no server-side session left to load.

**Solution A** (from OpenHands/OpenHands#14260): on restart, start a fresh `new_session` and synthesize the first user message from the **durable event store** — render prior `MessageEvent`s as role-tagged turns plus a compressed summary of `ACPToolCallEvent`s, prefixed with `<<RESUMED CONVERSATION>>`, passed as `initial_message`. No ACP-server cooperation; works identically for Claude Code, Codex, and Gemini CLI.

## Where it fits

This is the **lossy fallback tier** in the resume ladder for OpenHands/agent-canvas#2116:

```
high-fidelity   session/load  (object-store session snapshot + durable id mirror, OpenHands/agent-canvas#2116 + OpenHands/OpenHands#14506 + sdk#3506)
   ↓ blob missing / wrong version
lossy           <<RESUMED CONVERSATION>> bootstrap-prompt from durable events   ← THIS PR
   ↓ events unavailable
fresh           clean new_session
```

It is also the **permanent** resume path for **Gemini** (host-bound creds, no reliable `load_session`).

## Relationship to OpenHands/OpenHands#14295 (supersedes)

This is a squashed, rebased-onto-`main` refresh of OpenHands/OpenHands#14295 (which was ~a month and 19 commits stale, conflicting, and whose follow-up note pointed at the now-rejected `/home/openhands` PVC option). The implementation merged cleanly onto current `main`; only a test-helper conflict (a `get_user_email` mock main added) needed resolving. OpenHands/OpenHands#14295 will be closed in favour of this.

## What changed

`_synthesize_acp_resume_initial_message` in `live_status_app_conversation_service.py`:
- fetches events newest-first and reverses, so under the `_ACP_RESUME_MAX_EVENTS=200` cap the agent gets the **most recent** context;
- char-caps the rendered block (`_ACP_RESUME_CONTEXT_MAX_CHARS`, per-message/tool caps);
- returns `None` for fresh conversations (zero behaviour change on first start);
[...]

PR #1022, opened 2024-04-11
lint-frontend settingsSlice.ts

808vita:
for  lint-frontend
Unexpected console statement

Issue #1016, opened 2024-04-11
make run error


mirgong:
<img width="1684" alt="image" src="https://github.com/OpenDevin/OpenDevin/assets/26806117/12f2e818-e438-4766-9f16-4eefc53de24d">

rbren:
@mirgong if you search issues for `huggingface.co` you'll find the workaround.

In the future, please fill out the bug template 😄

PR #14671, opened 2026-06-05
feat(acp): snapshot ACP agent spec at conversation creation (#1015)

simonrosenberg:
## What

Freeze a **secret-free `ACPAgentSettings` snapshot** onto the conversation at creation, and rebuild the ACP agent from it on resume — instead of re-resolving the user's *live, mutable, global* `agent_settings` on every build.

Part of [agent-canvas#988](https://github.com/OpenHands/OpenHands/issues/15746). Implements [#1015](https://github.com/OpenHands/OpenHands/issues/15738) and upholds [#1016](https://github.com/OpenHands/agent-canvas/issues/2118).

## Why

The backend already builds an `ACPAgent` from the persisted `user.agent_settings`. But when a recycled sandbox is rebuilt, the cloud backend re-runs the ACP conversation-start path (#14640) and **re-resolves the agent from live settings**. So a settings edit *between* create and resume silently re-targets an in-flight conversation:

- start a conversation on **Claude Code / opus**, sandbox gets recycled, meanwhile the user switches their global default to **Codex** → the resumed conversation silently becomes Codex.

The regular agent avoids this: its resolved `Agent` is snapshotted into the agent-server's `meta.json` at creation and rebuilt from there, never re-resolved. ACP's session storage is gone after a recycle (hence the bootstrap-prompt resume in OpenHands/OpenHands#14640), so the spec must be pinned in the app's own durable store. This PR does exactly that.

## How

- `AppConversationInfo.acp_agent_settings_snapshot` — internal field (`exclude=True`; persisted via the explicit SQL mapping, read by attribute on resume, never serialized into list/API responses).
- New JSON column `acp_agent_settings_snapshot` on `conversation_metadata` (+ `save` / `_to_info` mapping). The cloud `SaasSQLAppConversationInfoService` inherits the patched base `save`/`_to_info`, so no enterprise code change is needed — only the migration.
- Migrations: **app_server alembic `011`** (OSS) and **enterprise alembic `118`** (cloud) — `conversation_metadata` is migrated by both.
- `_snapshot_acp_settings(settings)` — freezes the agent **identity** (`acp_server`, `acp_model`, `acp_session_mode`, `acp_command`/`acp_args`, timeouts, metrics `llm.model`) and **strips every secret-bearing field**.
- `_restore_acp_settings(snapshot, live)` — on resume, restores the frozen spec and re-attaches secrets **from the live encrypted vault**, re-injecting provider creds only when `live.acp_server == snapshot.acp_server` (so a provider switch can't inject a mismatched key; the subprocess falls back to ambient/materialised auth).

## Credentials are never persisted at rest (#1016)

Per-provider credential injection already landed in **#14620**: SDK 1.25.0's `ACPAgentSettings.create_agent()` folds `llm.api_key` into `agent_context.secrets`, which rides the cipher-protected `request.secrets` → `state.secret_registry` channel — credentials never touch the app DB. This PR preserves that by construction: `conversation_metadata` is a plain JSON column whose `TypeDecorator` dumps with `expose_secrets=True`, so the snapshot **must** be secret-free. `_snapshot_acp_settings` clears `llm` creds (`api_key`/`base_url`/`aws_*`), `acp_env`, `mcp_config` and `agent_context`; those are re-resolved from the live vault on every build. A test asserts no secret survives even an `expose_secrets` dump.

## Tests

- `_snapshot_acp_settings` strips all secrets (incl. expose-secrets dump is clean).
- `_restore_acp_settings` re-attaches live creds on a provider match; refuses a mismatched key after a provider switch.
- Resume race: snapshot present + live settings switched to Codex → the resumed agent stays Claude Code / opus.
- Fresh create uses live settings; `_start_app_conversation` persists a secret-free snapshot.
- SQL round-trip of the snapshot column; non-ACP conversations store `NULL`.

## Scope notes

- This is the snapshot half (#1015), the foundation OpenHands/agent-canvas#2118's credential injection builds on. The credential-injection / at-rest-encryption tasks of OpenHands/agent-canvas#2118 already merged in OpenHands/OpenHands#14620 — this PR ensures they aren't regressed.
- OpenHands/agent-canvas#2118's remaining "bump the cloud sandbox image pin to a post-ACP build" is a deploy-config change (separate from this backend PR).
- Per-provider credential storage (so a provider switch resumes with the right key) is tracked by [#1022](https://github.com/OpenHands/agent-canvas/issues/2113).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-3faf293   docker.openhands.dev/openhands/openhands:3faf293
```

### DIFF
--- a/enterprise/migrations/versions/118_add_acp_agent_settings_snapshot.py
+++ b/enterprise/migrations/versions/118_add_acp_agent_settings_snapshot.py
@@ -1,0 +1,36 @@
+"""Add acp_agent_settings_snapshot column to conversation_metadata table.
+
+Stores a secret-free snapshot of the ACP agent spec, frozen at conversation
+creation, so a later edit to the user's global agent settings cannot silently
+re-target an in-flight ACP conversation when its recycled sandbox is rebuilt
+(agent-canvas#1015). Provider credentials and other secrets are NOT persisted
+here — they are re-resolved from the live encrypted vault on every build
+(agent-canvas#1016) — so a plain JSON column is sufficient. NULL for non-ACP
+conversations and for ACP conversations created before this column landed.
+
+Revision ID: 118
+Revises: 117
+Create Date: 2026-06-05
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = '118'
+down_revision: Union[str, None] = '117'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        'conversation_metadata',
+        sa.Column('acp_agent_settings_snapshot', sa.JSON(), nullable=True),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column('conversation_metadata', 'acp_agent_settings_snapshot')
--- a/enterprise/tests/unit/storage/test_saas_sql_app_conversation_info_service.py
+++ b/enterprise/tests/unit/storage/test_saas_sql_app_conversation_info_service.py
@@ -281,6 +281,7 @@ class TestSaasSQLAppConversationInfoService:
         stored_metadata.public = None
         stored_metadata.tags = {}
         stored_metadata.agent_kind = None
+        stored_metadata.acp_agent_settings_snapshot = None
 
         saas_metadata = MagicMock(spec=StoredConversationMetadataSaas)
         saas_metadata.user_id = UUID('a1111111-1111-1111-1111-111111111111')
--- a/openhands/app_server/app_conversation/app_conversation_models.py
+++ b/openhands/app_server/app_conversation/app_conversation_models.py
@@ -23,6 +23,7 @@ from openhands.app_server.settings.settings_models import SandboxGroupingStrateg
 from openhands.sdk.conversation import ConversationExecutionStatus
 from openhands.sdk.llm import MetricsSnapshot
 from openhands.sdk.plugin import PluginSource
+from openhands.sdk.settings import ACPAgentSettings
 
 __all__ = ['SandboxGroupingStrategy']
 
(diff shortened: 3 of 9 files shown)


simonrosenberg:
Thanks for the review! All four suggestions turn out to be already-covered or not applicable — replying for the record rather than pushing a no-op commit (CI is green on `3faf293`):

1. **Return type annotations on the helpers** — already present: `def _snapshot_acp_settings(settings: ACPAgentSettings) -> ACPAgentSettings:` and `def _restore_acp_settings(snapshot: ACPAgentSettings, live: ACPAgentSettings) -> ACPAgentSettings:`.

2. **Extract the field name to a constant** — `acp_agent_settings_snapshot` is a pydantic field / SQLAlchemy column *declaration* name, so it can't be driven by a string constant at the points that matter (model field, `mapped_column`). The two attribute reads in the service are fine inline; a constant wouldn't reduce the coupling.

3. **Guard `live.llm` being `None`** — `ACPAgentSettings.llm` is a required field with `default_factory=_default_llm_settings`, so it's never `None`; `live.llm.model`/`.api_key` can't raise `AttributeError`. Adding a guard would be dead code.

4. **Assert `llm.api_key` isn't from live in `test_resume_uses_snapshot_not_live_settings`** — this exact property is already asserted by `test_restore_refuses_mismatched_cred_on_provider_switch` (`assert effective.llm.api_key is None` after a provider switch), which is the unit-level counterpart of that integration test.

Happy to fold any of these in if you feel strongly.

simonrosenberg:
Closing in favor of OpenHands/OpenHands#14709, which is a strict superset of this PR. OpenHands/OpenHands#14709 cherry-picked this OpenHands/OpenHands#15738 spec-snapshot work — the secret-free `acp_agent_settings_snapshot` column + restore-on-resume path, migrations 011/118, and the `app_conversation_models.py` / `live_status_app_conversation_service.py` / `sql_app_conversation_info_service.py` / `sql_utils.py` changes (and the same tests) — and layers native session/load resume on top (migrations 012/119, `AcpSessionSnapshotService`, the webhook session-id mirror, and native-resume precedence with bootstrap fallback). Keeping both open would collide on migration 118.

The native-resume stack was e2e-validated on OpenHands/OpenHands#14709's branch (= this work + the native delta): the recycled conversation rebuilds the ACP agent from the frozen snapshot and resumes via session/load — same `acp_session_id`, no `<<RESUMED CONVERSATION>>` marker — for Claude + Codex. So OpenHands/OpenHands#14709 carries and validates OpenHands/OpenHands#15738; review/merge it instead. Thanks.

PR #14672, opened 2026-06-05
fix: add Python 3.13 aifc dependency

he-yufeng:
HUMAN:
This PR restores Python 3.13 compatibility for the audio import path by adding the maintained `standard-aifc` backfill only on Python 3.13 and newer. I kept the change limited to dependency metadata and lockfile refresh because the source already imports `aifc` normally.

## Summary
- add `standard-aifc` for Python 3.13, where the stdlib `aifc` module is no longer available
- keep the dependency scoped to Python 3.13+ in both PEP 621 and Poetry dependency sections
- refresh `poetry.lock` so the lock file includes the backfilled package and its `audioop-lts` dependency

## Why
On Python 3.13, importing `aifc` fails unless the redistributable package is installed. OpenHands advertises Python 3.13 support, so the runtime dependency should be present for that interpreter only.

## To verify
- `poetry lock`
- `poetry check --lock` (passes; existing project metadata warnings are still printed)
- `git diff --check origin/main...HEAD`
- secret scan for `sk-...` patterns in `pyproject.toml` and `poetry.lock`


he-yufeng:
Ready for review. I added the required HUMAN note and left the human-tested box unchecked because this is a dependency metadata/lockfile fix validated locally with `poetry lock`, `poetry check --lock`, diff check, and secret scan.

he-yufeng:
Closing this dependency-only branch for now. It is already conflicted, and I do not want to keep a low-signal dependency refresh open while the OpenHands queue is crowded.

PR #14673, opened 2026-06-05
fix: use SDK validation before persisting settings

### LINKED ISSUE #14690: [codex] Use SDK validation before persisting settings
Tracks https://github.com/OpenHands/OpenHands/pull/14673 and https://github.com/OpenHands/software-agent-sdk/pull/3529.\n\nContext: caused by ACP settings writing  with no , leading to Pydantic validation errors. Simon Rosenberg has related fix in OpenHands/OpenHands#14675.

### LINKED ISSUE #14691: [codex] Use SDK validation before persisting settings
Tracks https://github.com/OpenHands/OpenHands/pull/14673 and https://github.com/OpenHands/software-agent-sdk/pull/3529. Context: ACP settings writing agent_kind=acp with no agent_context causes Pydantic validation errors. Related fix in OpenHands/OpenHands#14675.

neubig:
HUMAN:
This PR validates OpenHands agent settings with the SDK before persisting them, so invalid ACP settings without required context are rejected at save time instead of being stored and failing later on load.
- [ ] A human has tested these changes

## Summary
Add SDK-level validation in `AgentSettings.load()` and `AgentSettings.validate()` to prevent bad settings from being persisted. When the agent writes settings after reconfiguring (e.g., switching from ACP to OpenHands), it should call `validate()` before `save()` so that any malformed state is caught and rejected rather than stored and causing a Pydantic error on next load.

## Motivation
ACP (Agent Context Protocol) settings can write `agent_kind: acp` with no `agent_context`, which passes Pydantic validation but fails on load. Users hit this when re-enabling ACP via Agent Canvas.

## Changes
- Add validation in `AgentSettings` that mirrors the `load()` logic so `validate()` and `load()` form a valid pair: anything that passes `validate()` must also pass `load()`
- Call `validate()` before saving settings in all callers

## Related
- Fixes https://github.com/OpenHands/OpenHands/issues/14690
- Paired with https://github.com/OpenHands/software-agent-sdk/pull/3529

## Issue

Fixes #14691.


simonrosenberg:
Heads up: this is good defensive hygiene, but I don't think it fixes the active `/api/organizations` 500s we're seeing (#14674).

That crash is on the **read path** — `_validate_persisted_agent_settings` (`enterprise/server/routes/org_models.py:41`) and `OrgStore.get_agent_settings_from_org` (`enterprise/storage/org_store.py:67`) — which this PR doesn't touch. Two reasons it slips past the new write-time guard:

1. **The breaking shape is valid, so the guard allows the write.** `Settings.agent_settings` is the discriminated union (`AgentSettingsConfig`), and a clean ACP blob (`agent_kind="acp"`, `agent_context=None`) is a *valid* `ACPAgentSettings`. Running this PR's revalidation against that exact shape passes:

   ```
   revalidated_for_persistence() -> OK (write ALLOWED), agent_kind: acp | agent_context: None
   ```

   So ACP settings still persist with `agent_context: null`, and the read path still force-casts them to `OpenHandsAgentSettings` (whose `agent_context` is non-nullable) → `ValidationError` → 500. The data isn't malformed; the reader is wrong.

2. **It doesn't help the already-broken rows.** The ~2.4k-occurrences-over-14-days in prod come from orgs already on ACP in the DB; a write-time guard never runs against them.

Also worth noting: the regression test uses `agent_kind: "bogus"` (fails the union because there's no such tag) rather than the real `agent_kind: "acp"` shape — so the test goes green while the actual failure mode passes through.

Where this PR *does* help: the original software-agent-sdk#3093 scenario, where a deep-merge left a stale `agent_context: null` on an `agent_kind="openhands"` blob. That shape *is* malformed for the union, so this guard would reject it before persisting. Happy to see it land for that — it just needs to be paired with the read-path fix in #14674 (drop a null `agent_context` before the `agent_kind="openhands"` coercion, or land the SDK `before`-validator) to actually move the dashboard.

neubig:
Closing as stale/superseded per Graham's request during daily workflow triage.\n\n_This comment was created by an AI agent (OpenHands) on behalf of Graham Neubig._

PR #14697, opened 2026-06-06
feat(jira-dc): share linked OAuth token with eligible conversations

ak684:
HUMAN:

This PR is stacked on #14650 and makes the linked Jira Data Center OAuth token available to eligible web, Slack, and API-started conversations, not only Jira-triggered resolver jobs. I tested it on replicated-02 with a real Jira Data Center instance: a web UI conversation and the target Slack -> OpenHands -> Jira DC flow both queried Jira DC successfully using my linked Jira DC account.

## Summary

Stacked on #14650.

This PR moves Jira DC OAuth token exposure from the Jira resolver-only `secrets` path into a shared conversation-start enricher. Eligible user/org-scoped conversations now receive `JIRA_DC_BASE_URL` plus a `LookupSecret` for `JIRA_DC_TOKEN` when the effective user has an active Jira DC OAuth link for the configured Jira host.

## Notes

- Keeps Jira-triggered conversations strict: missing/unrefreshable Jira tokens are validated before sandbox launch so the existing Jira re-link error path still works.
- Keeps Slack/web/API starts best-effort: if the user has no valid Jira link, the conversation starts without Jira credentials.
- Does not add bot/service-account fallback for no-user automation; that remains a follow-up.

## Tests

- `uv run ruff check openhands/app_server/app_conversation/conversation_secret_enricher.py openhands/app_server/app_conversation/live_status_app_conversation_service.py openhands/app_server/server_config/server_config.py enterprise/integrations/jira_dc/jira_dc_conversation_secret_enricher.py enterprise/integrations/jira_dc/jira_dc_view.py enterprise/server/config.py enterprise/server/routes/integration/jira_dc.py enterprise/tests/unit/integrations/jira_dc/test_jira_dc_conversation_secret_enricher.py enterprise/tests/unit/integrations/jira_dc/test_jira_dc_view.py enterprise/tests/unit/server/routes/test_jira_dc_integration_routes.py`
- `uv run pytest tests/unit/app_server/test_live_status_app_conversation_service.py -q`
- `PYTHONPATH=enterprise uv run --with python-keycloak --with limits --with coredis pytest enterprise/tests/unit/integrations/jira_dc/test_jira_dc_conversation_secret_enricher.py enterprise/tests/unit/integrations/jira_dc/test_jira_dc_view.py enterprise/tests/unit/server/routes/test_jira_dc_integration_routes.py -q`

## Manual validation

- Replicated OHE instance: `replicated-02` / `instance-2`
- Release channel: `alona/jira-dc-shared-token-enricher-release`
- Image: `enterprise-server:sha-8729725`
- Jira DC was configured through KOTS against a real Jira Data Center instance, and the user linked their OpenHands account through Jira DC OAuth.
- Web UI flow passed: a direct OpenHands conversation successfully queried Jira Data Center using the linked user token.
- Slack target flow passed: Slack was enabled in KOTS with a real Slack app, the Slack user was linked to the same OpenHands user, the bot was added to a Slack channel, and a Slack-started OpenHands conversation successfully queried Jira Data Center end to end.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-d8cf6a0   docker.openhands.dev/openhands/openhands:d8cf6a0
```


ak684:
Follow-up on the All Hands bot review:

- Agreed on the readability suggestion: commit `a170a14aa` extracts the effective-org comparison into `_effective_org_matches`. Behavior is unchanged.
- I did not add a token-version/challenge field to the LookupSecret JWT in this PR. The endpoint already verifies the scoped signed JWT, re-checks the active workspace/host/link, and resolves the current stored Jira DC token at request time. Unlinking makes lookup fail; re-linking during the short JWT lifetime resolves the new active token. Versioned invalidation would require new persisted link/version state, so I would treat that as future hardening.

openhands/app_server/server_config/server_config.py
```diff
@@ -23,6 +23,7 @@ class ServerConfig(ServerConfigInterface):
     user_auth_class: str = (
         'openhands.app_server.user_auth.default_user_auth.DefaultUserAuth'
     )
+    conversation_secret_enricher_class: str | None = None
```
jpshackelford, line 26: Having a generic `ConversationSecretEnricher` rather than a Jira-specific hook sets us up for reuse with other integrations. One small change that would ensure the next integration can plug in cleanly without stepping on Jira DC:

`conversation_secret_enricher_class: str | None` is a single slot, so when the second integration ships (Linear is the natural one), it has to either silently replace Jira DC here, or hand-roll a class that wraps both. If we change to `list[str]` now and ship a small `CompositeConversationSecretEnricher` alongside the protocol, the next integration just adds itself to the list.

Concretely, four changes:

**1. This field:**

```python
conversation_secret_enricher_classes: list[str] = []
```

**2. `enterprise/server/config.py:66-69`** — wrap the existing entry:

```python
conversation_secret_enricher_classes: list[str] = [
    'integrations.jira_dc.jira_dc_conversation_secret_enricher.'
    'JiraDcConversationSecretEnricher',
]
```

**3. A small `Composite` in `openhands/app_server/app_conversation/conversation_secret_enricher.py`:**

```python
class CompositeConversationSecretEnricher(ConversationSecretEnricher):
    """Runs registered enrichers sequentially and merges their results."""

    def __init__(self, enrichers: list[ConversationSecretEnricher]):
        self._enrichers = enrichers

    async def enrich(
        self,
        *,
        user_context: UserContext,
        user: UserInfo,
        trigger: ConversationTrigger | None,
        system_message_suffix: str | None,
        web_url: str | None,
        jwt_service: JwtService,
        access_token_hard_timeout: timedelta | None,
    ) -> ConversationSecretEnrichment:
        merged_secrets: dict[str, SecretSource] = {}
        current_suffix = system_message_suffix
        for enricher in self._enrichers:
            enrichment = await enricher.enrich(
                user_context=user_context,
                user=user,
                trigger=trigger,
                system_message_suffix=current_suffix,
                web_url=web_url,
                jwt_service=jwt_service,
                access_token_hard_timeout=access_token_hard_timeout,
            )
            for name, source in enrichment.secrets.items():
                if name in merged_secrets:
                    logger.warning(
                        'Composite: secret %r overridden by %s',
                        name, type(enricher).__name__,
                    )
                merged_secrets[name] = source
            current_suffix = enrichment.system_message_suffix
        return ConversationSecretEnrichment(
            secrets=merged_secrets,
            system_message_suffix=current_suffix,
        )
```

**4. The injector** (`live_status_app_conversation_service.py:2284-2299`) — always wrap:

```python
enrichers = [
    get_impl(ConversationSecretEnricher, name)()
    for name in server_config.conversation_secret_enricher_classes
]
conversation_secret_enricher = CompositeConversationSecretEnricher(enrichers)
```

Always-wrap-in-Composite simplifies the call site: empty list = no-op, one element = pass-through, more = real composition. Replaces today's "default to `ConversationSecretEnricher()`" fallback.

Observable behavior is unchanged for this PR — a one-element Composite just passes through. The payoff lands when the next integration ships: it just adds its class string to the config list, no scaffolding changes needed.

If you'd rather defer the refactor itself, a TODO naming the migration intent is enough — something like:

```python
# TODO(integration-composability): migrate to `list[str]` and wrap registered
# enrichers with a CompositeConversationSecretEnricher when a second integration
# enricher is added. Today's single slot silently breaks JiraDC if a future PR
# overwrites this with a different class, with no test signal.
```

I'd want at least that, since the alternative foot-gun doesn't fail any test.

---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/tests/unit/server/routes/test_jira_dc_integration_routes.py
```diff
@@ -141,6 +144,52 @@ def mock_user_auth():
     return auth
 
 
+@pytest.mark.asyncio
+@patch(
+    'server.routes.integration.jira_dc.get_user_jira_dc_token', new_callable=AsyncMock
+)
+@patch('server.routes.integration.jira_dc.jira_dc_manager')
+async def test_get_jira_dc_secret_token_success(mock_manager, mock_get_token):
```
jpshackelford, line 152: Tests for the new endpoint and the enricher cover the happy paths but not the surfaces where silent failure is the default behavior in production. The sandbox SDK silently swallows 4xx LookupSecret responses (the env var just stays unset, no agent-visible error), and the non-Jira-trigger enricher paths are designed to silently skip on `JiraDcUserTokenError`. Combined, a future regression to either has no test signal *and* no operational signal — production keeps working but agents lose Jira access without anyone noticing.

A handful of targeted tests would close this — five on the route, one on the enricher.

### Route tests (here, in `test_jira_dc_integration_routes.py`)

- **Missing `X-Access-Token` header → 401.** Call `get_jira_dc_secret_token(access_token=None, jwt_service=...)`, assert `HTTPException(401)`.
- **Invalid JWT signature → 401.** Mock `jwt_service.verify_jws_token` to raise `InvalidTokenError`, assert `HTTPException(401)`.
- **JWT with wrong `integration` or `secret_name` claim → 403.** Mock the JWT payload to return `{'integration': 'github', ...}` (or `{'secret_name': 'GITHUB_TOKEN', ...}`), assert `HTTPException(403)`.
- **`JiraDcUserTokenError` from `get_user_jira_dc_token` → 401, warning log includes `user_id` and `workspace_id`.** Mock the resolver to raise `JiraDcUserTokenError`, assert `HTTPException(401)` and (with `caplog`) that the warning includes the identity fields requested in T1.3.
- **Unexpected exception from resolver → 500.** Mock the resolver to raise an `httpx.ReadTimeout` (or similar non-`JiraDcUserTokenError`), assert `HTTPException(500)`. **This one locks in the T1.3 fix specifically** — it's the test that would fail today and pass after the catch-all change.

### Enricher test (in `test_jira_dc_conversation_secret_enricher.py`)

- **Non-strict trigger + `JiraDcUserTokenError` → empty enrichment, no exception, warning logged.** Invoke `enrich()` with `trigger=ConversationTrigger.SLACK` and mock `get_user_jira_dc_token` to raise `JiraDcUserTokenError`. Assert the call returns `ConversationSecretEnrichment(secrets={}, system_message_suffix=<unchanged>)` without raising. This is the **defining behavioral split** between strict and non-strict mode (T1.2) — and currently the four existing tests cover (strict+success), (strict+error), (non-strict+success), (org mismatch), but not this combination.

### Pattern (using the existing `test_get_jira_dc_secret_token_success` as a template):

```python
@pytest.mark.asyncio
@patch('server.routes.integration.jira_dc.get_user_jira_dc_token', new_callable=AsyncMock)
@patch('server.routes.integration.jira_dc.jira_dc_manager')
async def test_get_jira_dc_secret_token_unexpected_exception_returns_500(
    mock_manager, mock_get_token,
):
    # Set up valid JWT and successful workspace/user lookups; fail at the resolver.
    store = AsyncMock()
    mock_manager.integration_store = store
    store.get_workspace_by_id.return_value = SimpleNamespace(
        id=7, name='jira.example.com', status='active',
    )
    store.get_active_user_by_keycloak_id_and_workspace.return_value = MagicMock()
    mock_get_token.side_effect = httpx.ReadTimeout('Keycloak unreachable')

    jwt_service = MagicMock()
    jwt_service.verify_jws_token.return_value = {
        'integration': 'jira_dc',
        'secret_name': 'JIRA_DC_TOKEN',
        'user_id': 'kc-user',
        'workspace_id': 7,
    }

    with (
        patch('server.routes.integration.jira_dc.JIRA_DC_ENABLE_OAUTH', True),
        patch(
            'server.routes.integration.jira_dc.JIRA_DC_BASE_URL',
            'https://jira.example.com',
        ),
    ):
        with pytest.raises(HTTPException) as exc:
            await get_jira_dc_secret_token(
                access_token='signed-token', jwt_service=jwt_service,
            )

    assert exc.value.status_code == status.HTTP_500_INTERNAL_SERVER_ERROR
```

The other route tests follow the same shape with different mock setups; the enricher test mirrors the existing tests in `test_jira_dc_conversation_secret_enricher.py` with the `SLACK` trigger and a raising `get_user_jira_dc_token`.

As a bonus it would be great to have a few others, but I wouldn't gate on these:

- workspace not found → 404
- user not linked → 404
- non-int workspace_id → 401
- missing claims (user_id or workspace_id absent from JWT payload) → 401

---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/integrations/jira_dc/jira_dc_conversation_secret_enricher.py
```diff
@@ -0,0 +1,223 @@
...
+                value=SecretStr(JIRA_DC_BASE_URL),
+                description='Jira Data Center base URL',
+            )
+        }
+
+        token_manager = TokenManager()
+        strict = trigger == ConversationTrigger.JIRA
+        if web_url:
+            if strict:
+                # Jira resolver starts must fail before launching the sandbox so
```
jpshackelford, line 184: The dry-run pattern here (call `_resolve_user_token`, throw the result away, then create a `LookupSecret` separately) is the cleverest piece of code in the PR. The 2-line comment hints at the intent but doesn't explain why we don't *use* the resolved token directly.

A future maintainer who "simplifies" this — say, by removing the dead-store `await _resolve_user_token(...)` because the result is unused — will silently break the "fail before sandbox launch" invariant that the Jira-resolver re-link flow depends on.

Expand the comment to be explicit about the dead-store intent:

```python
if fail_fast:
    # Dry-run validation only: we discard the resolved token here.
    # The LookupSecret below will re-resolve on each sandbox call so the
    # conversation always uses a fresh token. The purpose of this call is to
    # surface JiraDcUserTokenError *now* (before sandbox launch) so the existing
    # webhook manager can post the re-link prompt on the Jira ticket.
    await _resolve_user_token(..., fail_fast=True)
```

Or, if you prefer self-documenting code over comments: extract a thin wrapper named for what it does — e.g. `_assert_user_token_resolvable(...)` for the strict-mode call, keeping `_resolve_user_token` only where the result is actually used.

---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/server/routes/integration/jira_dc.py
```diff
@@ -206,6 +217,75 @@ def _workspace_matches_configured_jira_dc_host(workspace_name: str) -> bool:
...
+        if not jira_dc_user:
+            raise HTTPException(status_code=status.HTTP_404_NOT_FOUND)
+
+        user_token = await get_user_jira_dc_token(
+            keycloak_user_id=user_id,
+            workspace_id=workspace.id,
+            token_manager=token_manager,
+            store=jira_dc_manager.integration_store,
+        )
+        return Response(
```
jpshackelford, line 271: Defensive: the response body is a bearer token. The sandbox SDK consumer doesn't cache today, but corporate proxies, future reverse proxies with caching enabled by default, or any intermediate could log/cache. Cheap one-line fix:

```python
return Response(
    content=user_token.access_token.get_secret_value(),
    media_type='text/plain',
    headers={'Cache-Control': 'no-store', 'Pragma': 'no-cache'},
)
```

---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/integrations/jira_dc/jira_dc_conversation_secret_enricher.py
```diff
@@ -0,0 +1,223 @@
...
+
+        secrets: dict[str, SecretSource] = {
+            'JIRA_DC_BASE_URL': StaticSecret(
+                value=SecretStr(JIRA_DC_BASE_URL),
+                description='Jira Data Center base URL',
+            )
+        }
+
+        token_manager = TokenManager()
+        strict = trigger == ConversationTrigger.JIRA
```
jpshackelford, line 181: When a conversation is triggered through some means other than a Jira webhook — a Slack mention, a GUI conversation start, the REST API, or any of the other resolver webhooks (GitHub PR comment, GitLab MR, etc.) — the intent is for the enricher to **try** to inject the Jira DC token but **silently skip** if anything goes wrong, so the conversation still starts (just without Jira access). That's the right customer behavior: a Slack user asking `@openhands look at JIRA-123` should get a working conversation even if their Jira token is expired or there's a transient infrastructure problem; they aren't blocked from doing anything else.

Today that intent is implemented only for one specific exception: `JiraDcUserTokenError`, caught inside `_resolve_user_token`. Any other failure inside `enrich()` — a transient DB error from `JiraDcIntegrationStore`, a Keycloak hiccup during `_get_effective_org_id`, an unhandled `httpx.HTTPError`, a schema mismatch — propagates all the way up and fails the conversation start. The blast radius is wide: a Keycloak blip in SaaS doesn't just break Jira-webhook-triggered conversations, it breaks every GUI, Slack, and API conversation start for users who have nothing to do with Jira DC.

Two changes:

### 1. Rename `strict` → `fail_fast`

The current name doesn't describe what the variable controls. `fail_fast` is a recognized pattern name and reads correctly at every call site:

```python
# line 177
fail_fast = trigger == ConversationTrigger.JIRA

if web_url:
    if fail_fast:
        # Jira webhook starts have no recoverable failure path — fail
        # before launching the sandbox so the webhook manager can post
        # the existing re-link prompt on the Jira ticket.
        await _resolve_user_token(..., fail_fast=True)
    ...
else:
    user_token = await _resolve_user_token(..., fail_fast=fail_fast)
```

And in `_resolve_user_token` (lines 97-120):

```python
async def _resolve_user_token(
    *,
    user_id: str,
    workspace_id: int,
    token_manager: TokenManager,
    store: JiraDcIntegrationStore,
    fail_fast: bool,
) -> JiraDcUserToken | None:
    try:
        return await get_user_jira_dc_token(...)
    except JiraDcUserTokenError:
        if fail_fast:
            raise
        logger.warning(...)
        return None
```

Generic enough that if Linear or ServiceNow joins the fail-fast set later, the assignment just becomes `fail_fast = trigger in (JIRA, LINEAR_WEBHOOK)` and the rest of the code still reads correctly.

### 2. Move the policy to the call site so unrelated failures don't crash conversation starts

**Keep the existing internal raise** — the dry-run pre-validation at lines 178-188 needs to abort before sandbox launch for the Jira-webhook re-link flow to work. The call-site wrapper is the safety net for *everything else* (DB blip, Keycloak hiccup, schema mismatch). When `fail_fast=True`, the internal raise propagates through the wrapper's `if fail_fast: raise` and out.

Wrap the enricher invocation in `_setup_conversation_secrets` (`live_status_app_conversation_service.py:1031-1039`) and the ACP duplicate (lines 1708-1722):

```python
# In _setup_conversation_secrets, replacing the existing `await ... enrich(...)`:
fail_fast = trigger == ConversationTrigger.JIRA
try:
    enrichment = await self.conversation_secret_enricher.enrich(
        user_context=self.user_context,
        user=user,
        trigger=trigger,
        system_message_suffix=system_message_suffix,
        web_url=self.web_url,
        jwt_service=self.jwt_service,
        access_token_hard_timeout=self.access_token_hard_timeout,
    )
except Exception:
    if fail_fast:
        raise
    _logger.warning(
        'Conversation secret enricher failed for trigger=%s; '
        'continuing without integration secrets',
        trigger, exc_info=True,
    )
    enrichment = ConversationSecretEnrichment(
        system_message_suffix=system_message_suffix,
    )
```

This way:
- The non-fail-fast path catches **all** failures, not just `JiraDcUserTokenError` — a DB or Keycloak blip no longer breaks GUI/Slack starts.
- The fail-fast path still propagates (both the internal raise and any other exception), so Jira webhook starts still fail before sandbox launch.
- Every current and future enricher gets the same treatment for free without each having to remember to wrap its own internals.


---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/server/routes/integration/jira_dc.py
```diff
@@ -206,6 +217,75 @@ def _workspace_matches_configured_jira_dc_host(workspace_name: str) -> bool:
...
+            media_type='text/plain',
+        )
+    except HTTPException:
+        raise
+    except (InvalidTokenError, ValueError):
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED)
+    except JiraDcUserTokenError:
+        logger.warning('[Jira DC] Linked Jira DC token is unavailable', exc_info=True)
+        raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED)
+    except Exception:
```
jpshackelford, line 282: This catch-all is the last handler in the new `/integration/jira-dc/secrets/token` endpoint — the URL the sandbox SDK hits whenever the agent inside a conversation reads `$JIRA_DC_TOKEN`. The handlers above it cover the designed-for cases: missing/invalid JWT (401), wrong JWT scope (403), workspace/user lookup miss (404), and the token-refresh-failure-re-link path (401). This block catches everything else and also returns 401.

The problem: "everything else" includes failures that aren't auth-related at all — a DB query timeout during `get_workspace_by_id`, a Keycloak 5xx during the token refresh inside `get_user_jira_dc_token`, an unhandled `httpx.HTTPError`, a schema mismatch after a migration. All of those come back to the sandbox as 401, which:

1. The sandbox SDK silently swallows (the env var just stays unset, no agent-visible error).
2. Server-side monitoring (DataDog, log aggregators) treats as normal traffic for any auth-bearing endpoint, so a Keycloak outage that fails 100% of these calls produces no 5xx alert — just a spike in "unauthenticated users," no pager.
3. Doesn't carry user_id or workspace_id in the log, so ops can't correlate when they do investigate.

Net effect: in production, a database hiccup or Keycloak outage is **indistinguishable from "user never linked Jira DC"** — same status code to the sandbox, same agent behavior, similar log shape, no monitoring signal.

Three concrete changes:

### 1. Initialize identity locals at the top of the function

So every exception path can log them, whether or not the failure happened before JWT extraction:

```python
async def get_jira_dc_secret_token(...) -> Response:
    user_id: str | None = None
    workspace_id: int | None = None
    try:
        if not access_token:
            raise HTTPException(...)
        ...
        user_id = payload.get('user_id')
        workspace_id = payload.get('workspace_id')
        ...
```

### 2. Add identity to the existing `JiraDcUserTokenError` log while we're touching this block

```python
except JiraDcUserTokenError:
    logger.warning(
        '[Jira DC] Linked Jira DC token is unavailable for user=%s workspace=%s',
        user_id, workspace_id, exc_info=True,
    )
    raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED)
```

### 3. Change the catch-all status code and log level

```python
except Exception:
    logger.exception(
        '[Jira DC] Unexpected failure resolving secret token for user=%s workspace=%s',
        user_id, workspace_id,
    )
    raise HTTPException(
        status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
    )
```

- Auth failures stay 401 (the JWT and resolver errors above); **infrastructure failures get 500 so server-side monitoring fires.**
- `logger.exception(...)` auto-includes the traceback at ERROR level rather than WARNING — more appropriate for unexpected exceptions.
- The sandbox SDK still swallows the response either way today, but the 500-vs-401 distinction makes server-side metrics actually useful — and doesn't have to be undone when the SDK is improved to surface secret-resolution failures to the agent.
- No leak concern from the change — `HTTPException` carries `None` as its body content, so the bearer token never appears in the error response (same as today on the 401 path).


---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -2209,12 +2284,18 @@ async def inject(
...
 
                 app_mode = (
                     server_config.app_mode.value if server_config.app_mode else None
                 )
+                enricher_cls = get_impl(
+                    ConversationSecretEnricher,
+                    server_config.conversation_secret_enricher_class,
+                )
+                conversation_secret_enricher = enricher_cls()
             except (ImportError, AttributeError):
```
jpshackelford, line 2306: The injector wraps both the `server_config` import AND the enricher class resolution in one `try/except (ImportError, AttributeError)`. The comment says "if server_config is not available (e.g., in tests)" — but the same except also catches:

- `ImportError` from `get_impl(...)` if the configured class path is wrong (typo, module renamed/moved)
- `AttributeError` from `enricher_cls()` if the class has constructor issues

In production, either of those silently falls back to the no-op `ConversationSecretEnricher()`. Jira DC token injection stops; no log, no metric, no signal. The deployment looks healthy but is silently missing the integration.

Two cheap improvements — split the concerns and log on the production failure path:

```python
# Defensive import — server_config genuinely may not exist in test contexts.
conversation_secret_enricher = ConversationSecretEnricher()
server_config = None
try:
    from openhands.app_server.shared import server_config
except ImportError:
    pass

if server_config is not None:
    app_mode = (
        server_config.app_mode.value if server_config.app_mode else None
    )
    enricher_class_path = server_config.conversation_secret_enricher_class
    if enricher_class_path:
        try:
            enricher_cls = get_impl(
                ConversationSecretEnricher,
                enricher_class_path,
            )
            conversation_secret_enricher = enricher_cls()
        except Exception:
            _logger.exception(
                'Failed to load configured conversation_secret_enricher_class=%r; '
                'continuing with no-op enricher',
                enricher_class_path,
            )
```

The narrow `ImportError` stays for tests; the enricher class failure now produces a clear log so ops can spot the misconfiguration in deploy logs.

Small caveat on the snippet above: I moved `app_mode` resolution into the same `if server_config is not None:` block, which drops the original `AttributeError` guard around `server_config.app_mode`. Fine if `server_config.app_mode` is always present (even when its value is `None`); keep a narrow `AttributeError` catch around just that line if the attribute itself might be absent in some configs.


---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/integrations/jira_dc/jira_dc_conversation_secret_enricher.py
```diff
@@ -0,0 +1,223 @@
...
+            return ConversationSecretEnrichment(
+                system_message_suffix=system_message_suffix
+            )
+
+        workspace = await store.get_workspace_by_id(jira_dc_user.jira_dc_workspace_id)
+        if not workspace or workspace.status != 'active':
+            return ConversationSecretEnrichment(
+                system_message_suffix=system_message_suffix
+            )
+        if not _workspace_matches_configured_jira_dc_host(workspace.name):
```
jpshackelford, line 164: The strict-mode raise downstream at line 177 depends on this filter to short-circuit Jira **Cloud** webhook starts before reaching it. `ConversationTrigger.JIRA` is used by **both** Jira Cloud (`jira_view.py:241`) and Jira DC (`jira_dc_view.py:136`) webhook starts — `ConversationTrigger.JIRA_DC` is defined in the enum but never constructed on this branch.

So for a Cloud-only customer, the strict raise would fire — except this host filter short-circuits first because Cloud workspace names won't match the configured DC host. Safe by accident: the safety lives at this line (host check), but the dangerous action lives at line 177 (the strict assignment).

If anyone removes or reorders this filter, Cloud webhook starts will begin raising `JiraDcUserTokenError`. Two options, in order of recommendation:

1. **Add a NOTE + TODO** naming the dependency and the dead-enum reconciliation:

```python
# NOTE: trigger == ConversationTrigger.JIRA is set by both Jira Cloud and Jira DC
# webhooks. Cloud is filtered out above by the host-match check; this is
# load-bearing — do not reorder. TODO: reconcile ConversationTrigger.JIRA_DC
# (currently dead) and gate this on JIRA_DC instead of overloaded JIRA.
fail_fast = trigger == ConversationTrigger.JIRA
```

Lower-risk and higher-value — preserves current behavior and tees up the enum reconciliation as a follow-on.

2. **Make the invariant local to line 177** so the safety isn't load-bearing on filter order:

```python
# line 177
fail_fast = (
    trigger == ConversationTrigger.JIRA
    and _workspace_matches_configured_jira_dc_host(workspace.name)
)
```

The host check is redundant on this code path (control wouldn't have reached line 177 if the check had failed at line 160), but the redundancy is the point — the `fail_fast` assignment now self-documents the invariant. Slightly cleaner long-term shape; tiny additional cost.


---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/integrations/jira_dc/jira_dc_conversation_secret_enricher.py
```diff
@@ -0,0 +1,223 @@
...
+        access_token_hard_timeout: timedelta | None,
+    ) -> ConversationSecretEnrichment:
+        del user
+
+        if not JIRA_DC_ENABLE_OAUTH:
+            return ConversationSecretEnrichment(
+                system_message_suffix=system_message_suffix
+            )
+
+        user_id = await user_context.get_user_id()
```
jpshackelford, line 146: Every SaaS conversation start now runs this enricher, even for users with no Jira DC link. The cost shape depends on whether the user is linked:

- **Unlinked user (the common case)**: `user_context.get_user_id()` + `store.get_user_by_active_workspace(user_id)` → returns `None` → short-circuit at lines 150-153. **One extra indexed DB query per conversation start**, for every GUI / Slack / API / non-Jira-resolver start a SaaS deployment serves.
- **Linked user**: continues through workspace lookup, host check, org match, possibly a Keycloak round-trip via `_get_effective_org_id`. **3+ DB queries, possibly an IdP call.**
- **Jira-webhook-triggered start, linked user**: adds the `_resolve_user_token` dry-run — DB lookup plus a possible IdP refresh.

The relevant lookup columns are already indexed (`JiraDcUser.keycloak_user_id` is indexed and unique, `jira_dc_workspace_id` is indexed) so each query is cheap individually. The concern is the one-per-start cost across all the unlinked-user starts a SaaS deployment serves at conversation volume.

The PR description references functional validation on replicated-02 with one Jira DC instance and one Slack user. That's great for correctness but doesn't exercise the load characteristic at SaaS volume.

Three options to consider — any one is fine; what I want to avoid is finding this under load:

1. **Negative-cache the "no Jira DC link" result per user** in-memory (LRU) or short-lived Redis. Most queries short-circuit cheaply without touching the DB.
2. **Skip the enricher for triggers where Jira DC is never useful** (e.g., short-circuit on `ConversationTrigger.RESOLVER` from GitHub/GitLab/Bitbucket — the user came from a non-Jira webhook, they're working on something else).
3. **Accept the one-per-start cost** as load-bearing on the DB pool / index hit rate — fine if the answer is "we've measured this and it's noise," but I'd want that written down somewhere (commit message, follow-on issue) so it isn't relitigated at 3am during an incident.


---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/server/routes/integration/jira_dc.py
```diff
@@ -206,6 +217,75 @@ def _workspace_matches_configured_jira_dc_host(workspace_name: str) -> bool:
...
+        except (TypeError, ValueError):
+            raise HTTPException(status_code=status.HTTP_401_UNAUTHORIZED)
+
+        workspace = await jira_dc_manager.integration_store.get_workspace_by_id(
+            workspace_id_int
+        )
+        if (
+            not workspace
+            or workspace.status != 'active'
+            or not _workspace_matches_configured_jira_dc_host(workspace.name)
```
jpshackelford, line 254: Heads up — the two `_workspace_matches_configured_jira_dc_host` helpers in this PR aren't actually identical (I had thought they were). The divergence is behavioral:

- **Route** (this file, lines 206-217) delegates to `_configured_jira_dc_workspace_name()`, which returns `None` when `JIRA_DC_ENABLE_OAUTH` is False (legacy email-match mode). The route then `return True` on `None` configured workspace — **fail-open** in legacy mode.
- **Enricher** (`jira_dc_conversation_secret_enricher.py:39-43`) calls `urlparse(JIRA_DC_BASE_URL).hostname` directly with no OAuth-enabled gate. Returns `False` if no host can be parsed — **fail-closed**.

So they behave the same when OAuth is on and `JIRA_DC_BASE_URL` parses cleanly, but they have opposite safety defaults when the configuration is incomplete or OAuth is off. That's the more interesting question than the duplication I originally flagged — which behavior is intended for each call site?

A few options:

1. If they should converge on the same semantics: pick one (probably fail-closed for both, since the enricher's invocation now bypasses the legacy email-match path entirely), unify into a shared `enterprise/integrations/jira_dc/host_match.py`, and a single behavior is enforced.
2. If they're intentionally divergent (e.g., the enricher is invoked only after `JIRA_DC_ENABLE_OAUTH` is checked separately, so its fail-closed default never actually fires in legacy mode) — worth a code comment in both naming the asymmetry so a future refactor doesn't accidentally collapse them into one and silently change one site's behavior.

Either way, the original "extract to shared utility" is the right destination — but pick the semantics first.


---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

jpshackelford (commented):
Thanks for putting this up. Spent some time with this PR and the existing git-provider `LookupSecret` pattern to make sure I understood the convergence you're going for; left ten inline comments flagging things I noticed while reading.

Rough priority:

**Before merge** (four):
- strict-vs-non-strict error policy + `fail_fast` rename
- the route's catch-all status code
- composability of the new enricher protocol
- test coverage on the silent-failure paths

**Should address** (four):
- injector silent fallback
- comment explaining the strict-mode dry-run pattern
- comment naming (or tightening) the strict-mode trigger gate
- a perf concern that hasn't been load-tested

**Nits** (two):
- `Cache-Control` header on the token response
- divergent host-match helpers (fail-open in the route vs. fail-closed in the enricher)


---
_Drafted with an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/integrations/jira_dc/jira_dc_conversation_secret_enricher.py
jpshackelford: Shoot looks like reference to an openhands mention that was part of my comment triggered the openhands agent to fire.  🙃

ak684:
Thanks for the review. Here's what I addressed and what I deliberately left, plus a CI fix.

**CI fix — migration integrity (`check-sync`):** the stacked merge against `main` left a duplicate Alembic revision: both this branch's original `120_add_oauth_tokens_to_jira_dc_users.py` and `main`'s renamed canonical `122_add_oauth_tokens_to_jira_dc_users.py` were present (the three-way merge kept both because neither side deleted relative to the common ancestor). Removed the stale `120` copy — integrity now passes and the chain is linear (`… → 121 → 122`).

**Improvement opportunities**

1. **Log level (org-scoped skip)** — ✅ Done. Lowered `WARNING` → `INFO`. Minor correction to the rationale: this path only fires for *org-scoped* workspaces where no effective org resolves, not for non-org-scoped users (those return earlier). But it is an expected best-effort skip, so `info` is the right level.
2. **`del user` dead parameter** — ✅ Done. Added a one-line comment: identity is resolved via `user_context`; `user` exists only to satisfy the `ConversationSecretEnricher` interface.
3. **Fragile attribute discovery (`_get_effective_org_id`)** — ⚠️ Documented, not refactored. The `getattr` duck-typing intentionally supports the two known `UserContext` shapes (`resolver_org_id` for resolver starts; `get_effective_org_id` on the auth attr for user-auth starts) without a hard dependency. A robust fix is a formal `UserContext` protocol method — a cross-cutting change better done on its own. Added a comment documenting the known shapes so the list isn't a silent trap.

**Testing gap** — ❌ No change; already covered. `test_enricher_adds_jira_dc_lookup_secret_for_linked_user` already exercises the positive org-scoped match (workspace `ORG_ID` + context `ORG_ID`), a Slack trigger, and asserts the token is a `LookupSecret` (not a direct token).

PR #14699, opened 2026-06-07
feat: implement semantic file chunking using tree-sitter AST parsing

ysinghc:
HUMAN: I saw a to-do in your project while understanding your codebase for an internship work. it asked for tree-sitter chunking and was just returning Notimplementederror until now.

AGENT: I have used gemini 3.5 flash to understand parts i didnt know, and used sonnet 4.6 for writing the functions and tests.

Further used Opus for coding the check on max chunk size limit

---

## Why

The original fallback chunking implementation blindly split files based on raw line counts (`_create_chunks_from_raw_string`). This could split semantically important boundaries (like a function or a class definition) across multiple chunks, which degrades contextual understanding. This PR resolves the `NotImplementedError` for tree-sitter chunking, utilizing the Abstract Syntax Tree (AST) to ensure chunks break cleanly along semantic boundaries (e.g., between functions) while adhering to chunk size limits.

## Summary

- Implemented `_chunk_sibling_nodes` with a greedy-merge algorithm to group sibling AST nodes within a `max_chunk_lines` threshold, ensuring fully contiguous, non-overlapping coverage.
- Implemented recursive descent logic to handle individual oversized nodes (e.g., a massive class is recursively split into its methods).
- Wired `_create_chunks_from_tree_sitter` to `create_chunks` as the primary semantic splitter when a language parser is available.
- Added comprehensive unit tests in `test_chunk_localizer.py` to verify contiguous line ranges and nested node behavior.

## How to Test

You can test these changes by running the unit tests for the chunk localizer:

```bash
poetry run pytest tests/unit/utils/test_chunk_localizer.py -k "python" -v
```

This will run the newly added tests `test_create_chunks_tree_sitter_python_basic` and `test_create_chunks_tree_sitter_python_oversized`, which ensure that the chunker accurately generates AST nodes and merges/splits them based on line bounds.

## Video/Screenshots

image

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Fallback to raw-string chunking is still correctly preserved for empty files, un-parseable files, or when an unsupported language is passed.


openhands/app_server/utils/chunk_localizer.py
```diff
@@ -39,6 +39,135 @@ def _create_chunks_from_raw_string(content: str, size: int):
     return ret
 
 
+def _chunk_sibling_nodes(
+    nodes: list,
+    text_lines: list[str],
+    max_chunk_lines: int,
+    range_start_0: int,
+    range_end_0: int,
```
VascoSch92: the arg names are a little confusing. Can we do better?

WDYT?

openhands/app_server/utils/chunk_localizer.py
```diff
@@ -39,6 +39,135 @@ def _create_chunks_from_raw_string(content: str, size: int):
...
+            )
+        ]
+
+    groups: list[list] = []
+    current_group: list = []
+    current_group_start_0: int = 0
+
+    for node in nodes:
+        node_start_0: int = node.start_point[0]  # 0-indexed row
+        node_end_0: int = node.end_point[0]  # 0-indexed row
```
VascoSch92:
```suggestion
        # both are 0-indexed row
        node_start_0: int = node.start_point[0]
        node_end_0: int = node.end_point[0]
```

openhands/app_server/utils/chunk_localizer.py
```diff
@@ -39,6 +39,135 @@ def _create_chunks_from_raw_string(content: str, size: int):
...
+        chunk_text = '\n'.join(text_lines[range_start_0 : range_end_0 + 1])
+        return [
+            Chunk(
+                text=chunk_text,
+                line_range=(range_start_0 + 1, range_end_0 + 1),
+            )
+        ]
+
+    groups: list[list] = []
+    current_group: list = []
```
VascoSch92: could you improve the typing please?

openhands/app_server/utils/chunk_localizer.py
```diff
@@ -39,6 +39,135 @@ def _create_chunks_from_raw_string(content: str, size: int):
...
+        # Empty or un-parseable file – fall back to the naive splitter.
+        return _create_chunks_from_raw_string(text, max_chunk_lines)
+
+    return _chunk_sibling_nodes(
+        nodes=root.children,
+        text_lines=text_lines,
+        max_chunk_lines=max_chunk_lines,
+        range_start_0=0,
+        range_end_0=total_lines - 1,
+    )
```
VascoSch92:
```suggestion
    root = tree.root_node
    if not root.children:
        # Empty or un-parseable file – fall back to the naive splitter.
        return _create_chunks_from_raw_string(text, max_chunk_lines)

    return _chunk_sibling_nodes(
        nodes=root.children,
        text_lines=text.split('\n'),
        max_chunk_lines=max_chunk_lines,
        range_start_0=0,
        range_end_0=len(text_lines) - 1,
    )
```

openhands/app_server/utils/chunk_localizer.py
```diff
@@ -39,6 +39,135 @@ def _create_chunks_from_raw_string(content: str, size: int):
     return ret
 
 
+def _chunk_sibling_nodes(
+    nodes: list,
```
VascoSch92: typing

openhands/app_server/utils/chunk_localizer.py
```diff
@@ -39,6 +39,135 @@ def _create_chunks_from_raw_string(content: str, size: int):
...
+                    actual_end_0 + 1,
+                ),  # 1-indexed, inclusive
+            )
+        )
+
+    return result
+
+
+def _create_chunks_from_tree_sitter(
+    tree,
```
VascoSch92: typing

openhands/app_server/utils/chunk_localizer.py
```diff
@@ -39,6 +39,135 @@ def _create_chunks_from_raw_string(content: str, size: int):
     return ret
 
 
+def _chunk_sibling_nodes(
```
VascoSch92: when i == 0, the chunk starts at range_start_0. 

But for i > 0:  actual_start_0 = range_start_0 if i == 0 else group_ast_start_0                                                                                           
                                                                                                                                                              
This means blank lines between groups are absorbed into the previous chunk (via actual_end_0 = next_group_ast_start_0 - 1), but the gap between           range_start_0 and the first node's start is absorbed into the first chunk.

This is correct for contiguity, but consider this edge case:
                                                                                                                                                              If there's a comment block before the first AST node (e.g., a module docstring that tree-sitter represents as a node), it works. But if the first child starts at line 5 and range_start_0 is 0, you get 5 extra lines silently added to the first chunk, potentially exceeding max_chunk_lines. The max size constraint is not enforced on the expanded boundaries.

openhands/app_server/utils/chunk_localizer.py
```diff
@@ -52,9 +181,9 @@ def create_chunks(
         # fallback to raw string
         return _create_chunks_from_raw_string(text, size)
 
-    # TODO: implement tree-sitter chunking
-    # return _create_chunks_from_tree_sitter(parser.parse(bytes(text, 'utf-8')), max_chunk_lines=size)
-    raise NotImplementedError('Tree-sitter chunking not implemented yet.')
+    return _create_chunks_from_tree_sitter(
+        parser.parse(bytes(text, 'utf-8')), text, max_chunk_lines=size
```
VascoSch92:
```suggestion
        parser.parse(text.encode('utf-8')), text, max_chunk_lines=size
```

nit

tests/unit/utils/test_chunk_localizer.py
```diff
@@ -134,3 +134,39 @@ def test_get_top_k_chunk_matches_with_overlapping_chunks():
     assert matches[1].text == 'chunk3\nchunk4'
     assert matches[1].line_range == (3, 4)
     assert matches[0].normalized_lcs == matches[1].normalized_lcs
+
```
VascoSch92, line 220: Could we had more tests please?

- No test for fallback path (unparseable file or language without tree-sitter support)
- No test for deeply nested code (recursion behavior)                                                                                                       
- No test for a file with a single huge function (no children to recurse into — what happens?)                                                              
- No test for empty file                                                                                                                                    
- No test verifying max_chunk_lines is actually respected

VascoSch92 (commented):
Hey,

thank you very much for your PR.

I left a couple of comments but I think we are on the right track. ;-)

ysinghc:
@VascoSch92 I will review the suggestions and incorporate the changes, and about the test, yes i am going to add more, these two were just for me to check if the basic code was working fine, will add more tests for edge cases.
Thank you for the review.

VascoSch92:
Thanks for working through the earlier feedback. 

I rechecked this locally against the PR head, and I still think this needs one more fix before approval.

  The core invariant for this feature should be:
  - chunks are contiguous
  - chunks are non-overlapping
  - ranges are in-bounds
  - every range has `start <= end`
  - `'\n'.join(c.text for c in chunks) == text`

That invariant currently fails on valid Python.

Minimal reproducer:

```
  from openhands.app_server.utils.chunk_localizer import create_chunks

  text = "DATA = tuple([\n" + "\n".join(f"    {i}," for i in range(12)) + "\n])"

  for c in create_chunks(text, size=5, language="python"):
      print(c.line_range, repr(c.text))
```

  Current output:
```
  (1, 1) 'DATA = tuple(['
  (2, 1) ''
  (2, 1) ''
  (2, 5) '    0,\n    1,\n    2,\n    3,'
  (6, 10) '    4,\n    5,\n    6,\n    7,\n    8,'
  (11, 14) '    9,\n    10,\n    11,\n])'
  (15, 14) ''
```
  This emits inverted ranges, duplicate empty chunks, and a range past EOF. Reconstruction also fails, i.e.,   `'\n'.join(c.text for c in chunks) != text`.

It also trips the assertion in `Chunk.visualize() `because the emitted text line count no longer matches the declared line range.

I also confirmed this is not only a tiny-size edge case. A larger version of the same valid Python pattern fails at the default size=100, and at least one real file in this checkout, enterprise/integrations/github/queries.py, fails at size=100 with inverted ranges and reconstruction mismatch.

The root cause still looks like `_chunk_sibling_nodes`: when a single oversized node recurses into children, the recursive call uses the expanded `chunk_start` / `chunk_end` range rather than clamping to the actual AST node span. With same-row nested nodes, e.g. tuple( and [ both starting on row 0, the cursor can advance past subsequent child nodes on the same row. That lets both the normal emit path and the trailing-lines handler produce zero-width or inverted ranges.

  Two things I think should be added before approval:

  1. Add an invariant test helper that checks, across multiple sizes including 100:
      - first chunk starts at line 1
      - each chunk has start <= end
      - each chunk is in bounds
      - chunks[i].end + 1 == chunks[i + 1].start
      - last chunk ends at EOF
      - '\n'.join(c.text for c in chunks) == text

  2. Fix the boundary math in _chunk_sibling_nodes so recursive bounds are clamped to the oversized node’s actual span and no chunk is emitted unless `chunk_start <= chunk_end`.

ysinghc:
@VascoSch92

tests/unit/utils/test_chunk_localizer.py
```diff
@@ -8,6 +8,123 @@
 )
 
 
+def _assert_chunk_invariants(
```
VascoSch92: Minor (non-blocking): `_assert_chunk_invariants` and `assert_chunk_invariants` are near identical. It is worth consolidating.

VascoSch92 (commented):
Thanks for the rework!

One thing before I approve: the tests don't actually cover the semantic behavior. 

If I swap `_create_chunks_from_tree_sitter` for the raw line-splitter, all 122 tests still pass, they only check invariants, which raw chunking also satisfies. 

Even `test_..._python_basic` produces the same boundaries as raw.

`test_..._python_oversized` already runs on an input where they differ, tree-sitter gives `[(1,2),(3,5),(6,9),(10,10)]` vs raw `[(1,4),(5,8),(9,10)],` but it only asserts invariants, so it passes either way.


Could you add explicit boundary assertions on a couple of cases where semantic ≠ raw, e.g.:
  - unequal-length sibling functions → chunks[0] ends on the function boundary, not the budget line
  - an oversized class → splits between methods, not mid-method

  Once those are in I'm happy to approve.

VascoSch92 (approved):
Thanks a lot for your hard work :-)

PR #14700, opened 2026-06-07
fix: delete LiteLLM user on personal org reset

### LINKED ISSUE #14686: [Bug] delete_org_cascade leaks the LiteLLM user record when a personal org is deleted
## Problem statement

When a user resets their account by deleting their personal organization (via `DELETE /api/organizations/{org_id}` — the path used by the customer-facing `reset_account.sh` self-service script), the OpenHands `User` row, OpenHands `Org` row, LiteLLM team, and LiteLLM team-scoped keys are all cleaned up, but the **corresponding LiteLLM user record is left behind**.

When the same Keycloak identity signs back in, OpenHands re-onboards them with the same UUIDs (`User.id == Org.id == UUID(keycloak.sub)`) and `LiteLlmManager.create_entries` runs:

- `POST /team/new` → `200`
- `POST /user/new` → **`409 User with id <uuid> already exists`** ← the orphan
- → falls through to the existing 409 handler in `_create_user`, which calls `_user_exists`, returns `True`, and re-attaches the orphan to the freshly recreated team
- `POST /key/generate` → `200`

The orphan record's stored shape is missing the `user_id` field, so the LiteLLM proxy's user cache then fails pydantic validation against `LiteLLM_UserTable` on **every** request keyed by that user:

```
LiteLLM Proxy:WARNING: cache_pydantic_utils.py:87 - CacheCodec.deserialize:
  validation failed for LiteLLM_UserTable
    (1 validation error: user_id  Field required [type=missing,
     input_value={'team_alias': '<uuid>', …, 'teams': []}, input_type=dict])
LiteLLM Proxy:ERROR: user_api_key_cache.py:126 -
  UserApiKeyCache.async_get_cache failed to deserialize cached value
  for key='<uuid>' model_type=LiteLLM_UserTable
```

There is no further code path that heals this inconsistency. The login handler in `auth.py` only re-provisions LiteLLM for brand-new OpenHands users; the conversation start path and `SaasSettingsStore` never call `LiteLlmManager.create_entries` or `delete_user`; and there is no background reconciler between OpenHands and LiteLLM state. Cleanup can only be done out-of-band against the proxy with the cluster's `PROXY_MASTER_KEY`.

Reproduced in an enterprise customer environment running 0.7.37, traced via support bundles. The user-facing impact is permanent log noise on every chat request and a permanently inconsistent LiteLLM record for any user who uses the documented self-service reset workflow.

### Steps to reproduce

1. Sign in to a self-hosted OpenHands enterprise install as a new user with a managed-LLM config (e.g. `litellm_proxy/<model>` or `openhands/<model>`).
2. Run the customer `reset_account.sh` script (or `curl -X DELETE /api/organizations/{personal_org_id}` directly).
3. Sign back in.
4. Tail `kubectl -n openhands logs deploy/openhands-litellm`. Every chat completion produces the `CacheCodec.deserialize: validation failed for LiteLLM_UserTable` warning.
5. Confirm the orphan:
   ```bash
   kubectl -n openhands port-forward svc/openhands-litellm 4000:4000 &
   curl -sS "http://localhost:4000/user/info?user_id=<uuid>" \
     -H "Authorization: Bearer $PROXY_MASTER_KEY"
   # returns the stale record despite the user having been "reset"
   ```

1. **Primary fix:** In `OrgStore.delete_org_cascade`, call `LiteLlmManager.delete_user(...)` alongside the existing `delete_team(...)`, gated on the **same sole-orphan personal-org condition** that already cascade-deletes the requester's `User` row. This keeps the safety invariant that we never destroy a LiteLLM user who still has memberships in other orgs. Best-effort / idempotent semantics, matching the existing LiteLLM team cleanup (log on non-2xx, do not block the DB transaction commit).

2. **Secondary fix:** Same call in `LiteLlmManager.downgrade_entries`, so the legacy-tenant downgrade path stops producing the same orphan.

3. **Tests:** Assert `LiteLlmManager.delete_user` is invoked **exactly once** with the requester's user id in the sole-orphan branch of `delete_org_cascade`, and **not invoked** when the orphan condition is not met (multi-member org, requester has other memberships, etc.). Also assert a `500` from `delete_user` is logged and swallowed, not raised.

This is a small contained change (one extra call site in each of two methods plus tests) and does not change any HTTP API surface. Once shipped, the existing `reset_account.sh` becomes self-healing without any user-side changes.

### Workaround until fix ships

Cluster admins can purge the orphan against the in-cluster LiteLLM proxy before the user signs back in:

```bash
MASTER_KEY=$(kubectl -n openhands get secret lite-llm-api-key \
              -o jsonpath='{.data.lite-llm-api-key}' | base64 -d)
kubectl -n openhands port-forward svc/openhands-litellm 4000:4000 &
curl -sS -X POST http://localhost:4000/user/delete \
  -H "Authorization: Bearer $MASTER_KEY" \
  -d '{"user_ids":["<keycloak-sub-uuid>"]}'
```

A detailed proposed technical implementation (file paths, line numbers, patch sketch, test outline) will follow in a comment on this issue.

---
*This issue was filed by an AI agent (OpenHands) on behalf of the OpenHands customer success team while investigating an enterprise support escalation. Customer-identifying details have been omitted; raw support-bundle references are available internally.*

fengjikui:
HUMAN:
This PR fixes the personal-org reset path from #14686. When the requester is the sole user being removed, OpenHands already deletes the local user row and LiteLLM team, but the matching LiteLLM user can remain and block a clean sign-in. The fix adds best-effort LiteLLM user cleanup only for that requester_orphan_ids path; normal org deletion still does not delete LiteLLM users, and LiteLLM cleanup failures are logged without rolling back the OpenHands DB cascade.

## Summary

Fixes #14686.

When delete_org_cascade handles the personal-org reset path, OpenHands deletes the requester user row and the LiteLLM team, but the matching LiteLLM user record was left behind. This adds a best-effort LiteLLM user cleanup for that same sole-requester orphan branch so a reset user can sign back in without reattaching to a stale LiteLLM user record.

The cleanup is intentionally scoped to requester_orphan_ids. Regular org deletion paths do not delete LiteLLM users, and a LiteLLM user-delete failure is logged and swallowed so it does not roll back the already-valid OpenHands database cascade.

## Tests

- PYTHONPATH="enterprise:.:$PYTHONPATH" poetry run --project=enterprise pytest enterprise/tests/unit/test_org_store.py -k "delete_litellm_user_best_effort or delete_org_cascade_success"
- poetry run --project=enterprise pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files storage/org_store.py tests/unit/test_org_store.py
- poetry run --project=enterprise pre-commit run --config ../dev_config/python/.pre-commit-config.yaml --files enterprise/storage/org_store.py enterprise/tests/unit/test_org_store.py

### DIFF
--- a/enterprise/storage/org_store.py
+++ b/enterprise/storage/org_store.py
@@ -61,6 +61,17 @@ class OrgStore:
     """Store for managing organizations."""
 
     @staticmethod
+    async def _delete_litellm_user_best_effort(user_id: str, org_id: UUID) -> None:
+        """Delete the LiteLLM user record without blocking org deletion."""
+        try:
+            await LiteLlmManager.delete_user(user_id)
+        except Exception as exc:
+            logger.warning(
+                'Failed to delete LiteLLM user during org cascade cleanup',
+                extra={'org_id': str(org_id), 'user_id': user_id, 'error': str(exc)},
+            )
+
+    @staticmethod
     def get_agent_settings_from_org(org: Org) -> AgentSettingsConfig:
         # Route through the shared SDK loader: it applies persisted-settings
         # migrations (incl. the legacy ``agent_kind: 'llm'`` -> ``'openhands'``
@@ -690,6 +701,10 @@ class OrgStore:
                     extra={'org_id': str(org_id)},
                 )
                 await LiteLlmManager.delete_team(str(org_id))
+
+                if requester_orphan_ids:
+                    for user_id in requester_orphan_ids:
+                        await OrgStore._delete_litellm_user_best_effort(user_id, org_id)
 
                 # 7. Commit all changes only if everything succeeded
                 await session.commit()
--- a/enterprise/tests/unit/test_org_store.py
+++ b/enterprise/tests/unit/test_org_store.py
@@ -641,7 +641,13 @@ async def test_delete_org_cascade_success(async_session_maker, mock_litellm_api)
         session.add(expected_org)
         await session.commit()
 
-    with patch('storage.org_store.a_session_maker', async_session_maker):
+    with (
+        patch('storage.org_store.a_session_maker', async_session_maker),
+        patch(
+            'storage.org_store.OrgStore._delete_litellm_user_best_effort',
+            new=AsyncMock(),
+        ) as mock_delete_litellm_user,
+    ):
         # Act
         result = await OrgStore.delete_org_cascade(org_id)
 
@@ -651,6 +657,38 @@ async def test_delete_org_cascade_success(async_session_maker, mock_litellm_api)
     assert result.name == 'Test Organization'
     assert result.contact_name == 'John Doe'
     assert result.contact_email == 'john@example.com'
+    mock_delete_litellm_user.assert_not_called()
+
+
+@pytest.mark.asyncio
+async def test_delete_litellm_user_best_effort_calls_litellm():
+    user_id = str(uuid.uuid4())
+    org_id = uuid.uuid4()
+
+    with patch(
+        'storage.org_store.LiteLlmManager.delete_user', new=AsyncMock()
+    ) as mock_delete_user:
+        await OrgStore._delete_litellm_user_best_effort(user_id, org_id)
+
+    mock_delete_user.assert_called_once_with(user_id)
+
+
+@pytest.mark.asyncio
+async def test_delete_litellm_user_best_effort_swallows_litellm_failure():
+    user_id = str(uuid.uuid4())
+    org_id = uuid.uuid4()
+
+    with (
+        patch(
+            'storage.org_store.LiteLlmManager.delete_user',
+            new=AsyncMock(side_effect=Exception('LiteLLM API unavailable')),
+        ) as mock_delete_user,
+        patch('storage.org_store.logger.warning') as mock_warning,
+    ):
+        await OrgStore._delete_litellm_user_best_effort(user_id, org_id)
+
+    mock_delete_user.assert_called_once_with(user_id)
+    mock_warning.assert_called_once()
 
 
 @pytest.mark.asyncio
@@ -1067,7 +1105,13 @@ async def test_delete_org_cascade_sole_org_requester_is_deleted(
         await session.commit()
 
     # Act
-    with patch('storage.org_store.a_session_maker', async_session_maker):
+    with (
+        patch('storage.org_store.a_session_maker', async_session_maker),
+        patch(
+            'storage.org_store.OrgStore._delete_litellm_user_best_effort',
+            new=AsyncMock(),
+        ) as mock_delete_litellm_user,
+    ):
         result = await OrgStore.delete_org_cascade(
             org_id, requester_user_id=str(user_id)
         )
@@ -1075,6 +1119,7 @@ async def test_delete_org_cascade_sole_org_requester_is_deleted(
     # Assert: the deleted org is returned, and the user/org/member rows are gone.
     assert result is not None
     assert result.id == org_id
+    mock_delete_litellm_user.assert_called_once_with(str(user_id), org_id)
 
     async with async_session_maker() as session:
         assert await session.get(Org, org_id) is None


jpshackelford:
## Suggested manual verification plan

Unit tests cover the cascade-branch logic well; here's a complementary end-to-end plan that exercises the actual symptom from #14686 (the `LiteLLM_UserTable` cache warning and the 409-then-reattach loop on re-sign-in) across the OpenHands ↔ LiteLLM boundary:

1. Deploy this branch to a Replicated / Self-Hosted test channel. Confirm pods:
   ```bash
   kubectl -n openhands get pods -l app=openhands-litellm
   kubectl -n openhands get pods -l app.kubernetes.io/name=openhands
   ```

2. Sign in as a **brand-new** Keycloak user. Note the assigned UUID (visible in `X-User-Id` on any `/me` response, or in `enterprise/server/routes/auth.py:keycloak_callback` logs).

3. Send one chat turn and tail LiteLLM logs:
   ```bash
   kubectl -n openhands logs -f deploy/openhands-litellm | grep -E 'CacheCodec|UserApiKeyCache'
   ```
   Expect **no** warnings for a fresh user.

4. Reset the account via the documented self-service path:
   ```bash
   curl -i -X DELETE -H "Cookie: <session>" \
     "https://<host>/api/organizations/<personal_org_id>"
   # Expect 200.
   ```

5. Verify the orphan was reaped:
   ```bash
   kubectl -n openhands port-forward svc/openhands-litellm 4000:4000 &
   curl -sS "http://localhost:4000/user/info?user_id=<uuid>" \
     -H "Authorization: Bearer $PROXY_MASTER_KEY"
   # Expect HTTP 404 / "user not found". Before this PR: returns the stale row.
   ```

6. Sign back in as the same Keycloak identity. Tail the enterprise-server logs and confirm `LiteLlmManager.create_entries` produces:
   ```
   POST /team/new       -> 200
   POST /user/new       -> 200   ← was 409 + fallback-reattach before this PR
   POST /key/generate   -> 200
   ```

7. Send one chat turn. Re-tail LiteLLM logs:
   ```bash
   kubectl -n openhands logs deploy/openhands-litellm \
     | grep -E 'CacheCodec.deserialize: validation failed for LiteLLM_UserTable' \
     | tail -20
   ```
   Expect **no** warnings keyed on the user's UUID anymore (these used to fire on every chat).

8. **Negative regression — multi-member-org safety invariant.** In a multi-member org, delete a non-sole-member user via the cascade path:
   ```bash
   # Confirm the deleted member's LiteLLM record was preserved
   curl -sS "http://localhost:4000/user/info?user_id=<other_member_uuid>" \
     -H "Authorization: Bearer $PROXY_MASTER_KEY"
   # Expect 200 (still present) — this PR explicitly should NOT reap them.
   ```
   Asserted in the unit tests, but worth confirming on a live install too.

### Real-world "before" evidence

A recent enterprise support bundle (Replicated self-hosted install, version 0.7.37) shows two distinct users with stale orphan LiteLLM records — every `POST /chat/completions` from either emits the `CacheCodec.deserialize: validation failed for LiteLLM_UserTable (user_id Field required …)` warning. Step 7 above would confirm the warning disappears for fresh-reset users post-fix.

Note: existing orphans created before this PR lands will need a one-shot reap via the LiteLLM admin API; this PR only prevents new orphans from being created.

### Scope note

Issue #14686 also called out `LiteLlmManager.downgrade_entries` as a secondary fix site (legacy-tenant downgrade can produce the same orphan). This PR scopes to `delete_org_cascade` only, which covers the customer-facing `reset_account.sh` path — the dominant orphan source. The downgrade path can be a separate follow-up.

---

_This comment was posted by an AI agent (OpenHands) on behalf of @jpshackelford._

ak684 (approved):
Reviewed + validated the surrounding behavior empirically on a Replicated test install: LiteLLM user records are shared across all of a user's teams, so scoping the cleanup to the requester-orphan path (the only path where the OpenHands user row is deleted and re-onboarding can hit the /user/new 409) is exactly right. Best-effort error handling and tests look good.

PR #14701, opened 2026-06-07
Fix WebSocket user messages triggering agent run

fengjikui:
HUMAN:

This PR keeps user follow-up messages from getting stuck after a conversation is idle by making the WebSocket send path request an agent run, just like the REST/pending-message paths already do.

AGENT:

---

## Why

Closes OpenHands/enterprise#28.

When the frontend sends a user message over the direct agent-server WebSocket, it currently sends only `role` and `content`. The agent-server event endpoint defaults `run` to false, so a follow-up message can be persisted without restarting the agent loop after a conversation is already idle or finished.

The REST and pending-message paths already request a run, so this aligns the connected WebSocket path with those message delivery paths.

## Summary

- Add `run: true` to user messages sent through the connected conversation WebSocket.
- Tighten the WebSocket message-sending test to assert that sent payloads include `run: true`.

## Issue Number

OpenHands/enterprise#28

## How to Test

I ran:

```bash
cd frontend
npm run test -- __tests__/conversation-websocket-handler.test.tsx
npm run typecheck
npx prettier --check src/contexts/conversation-websocket-context.tsx __tests__/conversation-websocket-handler.test.tsx
npx eslint src/contexts/conversation-websocket-context.tsx
```

The commit hook also ran frontend lint-staged checks, staged typecheck, translation completeness, and backend pre-commit checks successfully.

Note: `make install-pre-commit-hooks` could not run before editing because this machine's default Python is outside the repository's required `>=3.12,<3.14` range.

## Video/Screenshots

N/A, logic-only WebSocket payload fix covered by tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This intentionally keeps the change narrow to the connected WebSocket path used for user messages. Existing queued/pending-message delivery already sends `run: true` when it forwards pending messages to the agent server.

### DIFF
--- a/frontend/__tests__/conversation-websocket-handler.test.tsx
+++ b/frontend/__tests__/conversation-websocket-handler.test.tsx
@@ -1110,10 +1110,12 @@ describe("Conversation WebSocket Handler", () => {
       expect(receivedMessages[0]).toEqual({
         role: "user",
         content: [{ type: "text", text: "Message 1" }],
+        run: true,
       });
       expect(receivedMessages[1]).toEqual({
         role: "user",
         content: [{ type: "text", text: "Message 2" }],
+        run: true,
       });
     });
   });
--- a/frontend/src/contexts/conversation-websocket-context.tsx
+++ b/frontend/src/contexts/conversation-websocket-context.tsx
@@ -910,7 +910,7 @@ export function ConversationWebSocketProvider({
 
       try {
         // Send message through WebSocket as JSON
-        currentSocket.send(JSON.stringify(message));
+        currentSocket.send(JSON.stringify({ ...message, run: true }));
         return { queued: false };
       } catch (error) {
         const errorMessage =


fengjikui:
Ready for review. 👍

shanemort1982:
Thanks for picking this up. I don't think this PR can fix OpenHands/enterprise#28 as written, and the fix is in a different repo.

Reading the WebSocket server path at `software-agent-sdk` (`openhands-agent-server/openhands/agent_server/sockets.py:312-313`):

```python
message = Message.model_validate(data)
await event_service.send_message(message, True)
```

`run=True` is already hardcoded as the second positional arg, and `Message.model_validate(...)` parses only `role`/`content` from the WebSocket payload. The new `"run": true` field this PR adds to the JSON is stripped at validation, never reaches the call site, and does not affect the agent loop restart. The added test only proves the client now sends a field the server path does not read.

For this PR to fix OpenHands/enterprise#28, the silent failure has to be on the server side, not the client side.

Where the defect actually lives (v1.25.0, image `sha256:1fa7632287f9f48b6ceedfc67da0813459d5c7c8c9e2cb0b6ef63c0d762c360e`; still present on `software-agent-sdk` main at `6fdc84f`):

`event_service.py:1016` (`EventService.pause`) and `event_service.py:1033` (`EventService.interrupt`) both bump `self._explicit_interrupt_generation` unconditionally before delegating to `LocalConversation`. On a `FINISHED` conversation, `LocalConversation.pause()` is a no-op, and `LocalConversation.interrupt()` falls back to `self.pause()` when no `arun()` task is live, which is also a no-op on `FINISHED`. Either path leaves a "phantom" stop-intent on a conversation that was never actually paused or interrupted.

When the next user message lands, `EventService.send_message()` (L450-458) snapshots the generation, persists the message, then early-returns without calling `run()` if the generation differs from the snapshot. The phantom bump from an earlier no-op `pause`/`interrupt` makes that early return fire silently, which matches the reported symptom exactly: message persisted, `last_user_message_id` updated, never `running`, never `ActionEvent`.

Note that `EventService.interrupt()` already implements the right pattern at L1034:

```python
if not internal_acp_rerun:
    self._explicit_interrupt_generation += 1
```

i.e. "an interrupt that does nothing shouldn't bump the stop-intent counter." The fix is extending the same principle to both bump sites for the no-op case. Sketch:

```python
async def pause(self):
    if self._conversation:
        if await self._get_execution_status() not in (RUNNING, IDLE):
            return
        self._explicit_interrupt_generation += 1
        ...

async def interrupt(self, *, internal_acp_rerun=False):
    if self._conversation:
        live = self._run_task is not None and not self._run_task.done()
        acts = live or (await self._get_execution_status() == IDLE)
        if not internal_acp_rerun and acts:
            self._explicit_interrupt_generation += 1
            ...
```

L457 itself should NOT be the patch target. It exists to enforce the user-stop-intent invariant. Patching L457 to fall through to `run()` would break that guarantee in cases where the bump was real.

Happy to be wrong if there's a server-side path where the new `run: true` field changes behaviour that I've missed. Otherwise this PR will close OpenHands/enterprise#28 with a green merge and the reproducer will persist.

fengjikui:
Thanks for the detailed correction. I re-checked the server path and agree this PR is not an effective fix for OpenHands/enterprise#28: the frontend-added `run` field is not consumed by the agent-server WebSocket path, and the actual stop-intent race belongs in `OpenHands/software-agent-sdk`.

Closing this PR to avoid merging an inert client-side change. The server-side fix is now being handled in OpenHands/software-agent-sdk#3557, which targets `EventService.pause()/interrupt()` and the `_explicit_interrupt_generation` no-op bump directly.

Appreciate the careful code walk.

PR #14506, opened 2026-05-21
feat(acp): persist ACP session id, redirect storage onto /workspace PVC (#14260 Solution B)

simonrosenberg:
## Why

ACP conversations (Claude Code, Codex) can't be resumed after a cloud
sandbox is recycled. The SDK already wires `session/load`, but the two
pieces of state it depends on both live inside the sandbox FS and are
wiped on recycle:

1. `acp_session_id` — in `state.agent_state` → `base_state.json`
2. The ACP server's own session files (`~/.claude/projects/…`,
   `~/.codex/sessions/…`)

This PR implements Solution B from OpenHands/OpenHands#14260: durable persistence for
both halves.

### 2a — durable session-id mirror

- New columns `acp_session_id` / `acp_session_cwd` on
  `conversation_metadata` (migrations `010` and enterprise `114`).
- New service method `AppConversationInfoService.update_acp_session`
  with SQL implementation.
- Webhook `on_event` extracts those keys from
  `ConversationStateUpdateEvent(key='agent_state')` and persists them.
- `_build_acp_start_conversation_request` reads them back and passes
  them to the SDK via `ACPAgent.acp_resume_session_id` (introduced in
  the paired SDK PR).

Guarded by `'acp_resume_session_id' in ACPAgent.model_fields` so this
PR can land before the SDK pin bump — once OpenHands pins to an SDK
version that includes the field, the resume signal flips on without
any further code change.

### 2b — relocate ACP server storage via the SDK (`acp_isolate_data_dir`)

For `session/load` to find anything, each server's own session files
must also survive a recycle. `_build_acp_start_conversation_request`
flips on `ACPAgent.acp_isolate_data_dir`, and the SDK
(software-agent-sdk OpenHands/agent-canvas#2114 / OpenHands/OpenHands#3492) relocates each provider's data/session
dir onto the durable, **per-conversation** `<persistence_dir>/acp/<provider>`
tree — which lives under the `/workspace` PVC that survives sandbox
recycles (its `ownerReference` points to the Deployment, not the Pod).

This replaces the earlier approach in this PR, which hand-injected
`CLAUDE_CONFIG_DIR=/workspace/.claude` / `CODEX_HOME=/workspace/.codex`
into the deprecated `acp_env` channel. Delegating to the SDK is also a
correctness improvement:

- **Per-conversation isolation** — the old env pointed every
  conversation at one shared `/workspace/.claude` (resp. `.codex`),
  which races when several of a user's conversations share a sandbox.
[...]

PR #1015, opened 2024-04-11
Traffic Control: Add new config MAX_CHARS

li-boxuan:
This PR adds a new config that allows users to cap the number of characters sent to + received from the LLM per task.

Since tokenization happens on the LLM side, we don't limit the number of tokens. Instead, we limit the number of characters.

You might see a lot of unrelated stylish fixes. They are done by the git pre-commit hooks.

Closes #449

PR #14709, opened 2026-06-07
feat(acp): native session/load resume across sandbox recycles (#14506 / agent-canvas#1126)

simonrosenberg:
HUMAN:

 mirrors the acp_session_id/cwd onto the conversation record, snapshots that session blob into the app-server's object store at each turn, and on a recycle restores it into the new sandbox and tells the CLI to
  session/load — falling back to bootstrap-prompt resume if anything's missing.

AGENT:

Validated end-to-end against the **real** Codex and Claude Code CLIs via a SaaS-equivalent local stack — the same `LiveStatusAppConversationService` path SaaS uses, with the Docker sandbox service and an agent-server image built from software-agent-sdk#3562 (ACP CLIs + `/api/acp_session_blob` routes baked in), SQLite + LocalFileStore.

STOP-simulating round trip per provider (turn 1 plants a codeword → `docker rm -f` the sandbox container = hard recycle, empty `/workspace` + new container id → turn 2 on a fresh sandbox asks for the codeword):
```
[harness] DB mirror: session_id='019ea2f8-...' cwd='/workspace/project' ver='0.15.0' kind='acp'
[harness]   ✓ snapshot contains session files only, NO credentials
[harness] Recycling sandbox ... (docker rm -f)  ✓ fresh container
[harness]   ✓ continuity: codeword PLATYPUS-7731 recalled after recycle
[harness]   ✓ native session/load: same session id (not new_session)
[harness]   ✓ no bootstrap-prompt marker (native resume took precedence)
=== CODEX e2e PASSED ===   (CLAUDE e2e PASSED too)
```
Tests: `uv run pytest tests/unit/app_server/` → 1201 passed (1 pre-existing macOS `$TMPDIR` flake in `test_process_sandbox_service`, unrelated), including 28 new native-resume tests + the 206 ACP/live-status/sql/webhook tests.

---

## Why

Cloud ACP conversations whose sandbox is recycled currently resume via a synthesized bootstrap prompt (#14640) — lossy. With the CLI session files now durable (software-agent-sdk#3562) and a durable session-id mirror, Claude Code and Codex can resume their **own** session via native `session/load` for full fidelity, falling back to bootstrap only when native resume isn't viable (Gemini, missing/old-image blob).

## Summary

- Cherry-picks #1015 (secret-free `acp_agent_settings_snapshot`, migrations app_server 011 / enterprise 118).
- **id half (#14506):** mirror `acp_session_id`/`acp_session_cwd`/`acp_agent_version` from the `agent_state` webhook stream onto `conversation_metadata` (migrations app_server 012 / enterprise 119; targeted `update_acp_session`), fed back via `ACPAgent.acp_resume_session_id`.
- **files half (agent-canvas#1126):** `AcpSessionSnapshotService` pulls the allowlisted CLI session blob from the sandbox at turn boundaries into `config.file_store`, and seeds it back into a fresh sandbox before rebuild. Sandbox never holds store creds — all I/O rides the session-api-key channel. GC'd on conversation delete.
- `_build_acp_start_conversation_request` tries native resume first (`_prepare_native_acp_resume`: mirrored id + session/load-capable provider + matching cwd + restorable blob → set `acp_resume_session_id` + `acp_isolate_data_dir`); only otherwise runs the existing bootstrap resume. Gemini stays on bootstrap permanently.
- Full-row conversation saves preserve the mirror so the out-of-band webhook write isn't clobbered to NULL.

## Issue Number

#14506 / #14260 (Solution B), agent-canvas#988 / agent-canvas#1126

## How to Test

See the AGENT section. Briefly: build the agent-server image from software-agent-sdk#3562, run the app_server with `AGENT_SERVER_IMAGE_REPOSITORY`/`AGENT_SERVER_IMAGE_TAG` pointed at it, set an ACP agent + provider secret, start a conversation, `docker rm -f` the sandbox, then re-start the conversation and confirm `session/load` (same `acp_session_id`, no `<<RESUMED CONVERSATION>>` marker). Unit: `uv run pytest tests/unit/app_server/test_acp_native_resume.py`.

## Type

- [x] Feature

## Notes

- Depends at the image level on software-agent-sdk#3562 (the blob routes). The Python package only needs fields already in SDK 1.25.0. The default runtime image pin must be bumped to an ACP-blob-enabled build for cloud; until then the restore PUT 404s and resume degrades cleanly to bootstrap.
- Pairs with agent-canvas `cloud-acp-native-resume` (the wake-recycled-sandbox UI).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-a1ad826   docker.openhands.dev/openhands/openhands:a1ad826
```


openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -1895,6 +2012,83 @@ async def _synthesize_acp_resume_initial_message(
...
+        """Arrange native session/load resume; return the session id when viable.
+
+        Viable means: the durable mirror has a session id for a provider with
+        reliable session/load (claude-code/codex — not gemini), the recorded
+        session cwd matches the sandbox working dir the CLI will get, and the
+        CLI's session files are present in the sandbox after a restore attempt
+        (seeded from the FileStore snapshot, or already live on a surviving
+        volume). Returns None otherwise so the caller falls back to
+        bootstrap-prompt resume.
+        """
```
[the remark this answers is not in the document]
simonrosenberg: Done in 6d0406c30 — split the guard so the `acp_session_cwd is None` case emits a debug log and a comment explaining why we proceed: we can't verify the match, so the snapshot restore + the CLI's own session/load are the real gate and degrade to bootstrap if it doesn't fit.

simonrosenberg:
Addressed the review (review 4450095041) in a1ad826b0:

1. **Module docstring** (`acp_session_snapshot_service.py`) — condensed ~20 lines → ~10, dropping the two-halves/PR-history narration; kept the load-bearing facts (the credential-exclusion and the FileStore-substrate guarantee).
2. **`_ACP_SNAPSHOT_STATUSES` comment** (`webhook_router.py`) — tightened to two lines, keeping the *why* (capture at turn boundaries where the CLI has flushed its session files so the snapshot is consistent).
3. **`#1015` snapshot block** (`live_status_app_conversation_service.py`) — condensed the 11-line rationale block to the essential why (freeze-vs-re-resolve + secrets-never-frozen), dropping the PR-context narration. Left the `_snapshot_acp_settings`/`_restore_acp_settings` docstrings as-is since they describe actual behaviour (incl. the provider-switch caveat).
4. **Instance-local dedup** (`_in_flight`) — clarified in-code that it's a best-effort in-process dedup and a multi-instance double-capture is harmless since the write is idempotent (overwrites the same blob). Left it instance-local by design — a cross-instance lock would be overkill for an idempotent best-effort optimization.

Comment-only changes; ruff + the 168 native-resume/live-status unit tests stay green. Thanks for the review — and agreed on trimming the design-narration out of the source.

openhands/app_server/event_callback/webhook_router.py
```diff
@@ -183,6 +187,123 @@ async def _track_conversation_terminal(
     )
 
 
+# ACP turn-boundary statuses: capture here, where the CLI has flushed its
+# session files and a snapshot is consistent.
+_ACP_SNAPSHOT_STATUSES = frozenset(
+    {
+        ConversationExecutionStatus.FINISHED,
+        ConversationExecutionStatus.PAUSED,
```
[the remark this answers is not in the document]
simonrosenberg: Let's keep it simple for now

enterprise/migrations/versions/118_add_acp_agent_settings_snapshot.py
```diff
@@ -0,0 +1,36 @@
+"""Add acp_agent_settings_snapshot column to conversation_metadata table.
```
malhotra5, line 1: NIT: can we combine the two migration files here into one

openhands/app_server/app_conversation/acp_session_snapshot_service.py
```diff
@@ -0,0 +1,238 @@
...
+NATIVE_SESSION_RESUME_PROVIDERS: tuple[str, ...] = ('claude-code', 'codex')
+
+_BLOB_TIMEOUT_SECONDS = 60.0
+
+
+def supports_native_session_resume(provider: str | None) -> bool:
+    return provider in NATIVE_SESSION_RESUME_PROVIDERS
+
+
+def get_agent_server_url(sandbox: SandboxInfo) -> str:
```
malhotra5, line 48: NIT: we have a util `get_agent_server_url_from_sandbox` we can reuse here

openhands/app_server/app_conversation/acp_session_snapshot_service.py
```diff
@@ -0,0 +1,238 @@
...
+    timeout: float = _BLOB_TIMEOUT_SECONDS
+    # Best-effort in-process dedup of concurrent captures (keyed by hex id). It
+    # is instance-local; a multi-instance double-capture is harmless since the
+    # write is idempotent (overwrites the same blob).
+    _in_flight: set[str] = field(default_factory=set)
+
+    def _blob_path(self, user_id: str | None, conversation_id: UUID, provider: str):
+        # FileStore.read is text-only across backends, so the gzip blob is
+        # stored base64-armored.
+        return (
```
malhotra5, line 70: NIT: for local persistence we seem to omit the `user_id` section all together (instead of a dummy value like `-`); for consistency would be nice to omit the dummy value from path

openhands/app_server/app_conversation/acp_session_snapshot_service.py
```diff
@@ -0,0 +1,238 @@
...
+                return False
+            if response.status_code == 404:
+                _logger.info(
+                    'Sandbox image has no ACP session-blob route; '
+                    'skipping snapshot for %s',
+                    conversation_id,
+                )
+                return False
+            response.raise_for_status()
+            blob = response.content
```
malhotra5, line 125: what's the typical size for these blobs?

malhotra5:
Noting for posterity (from slack discussions)

We seem to be re-seeding the ACP session data. For a reclaimed sandbox (i.e sandbox resume), this re-seed is a no-op as the volume is persisted and holds a copy of the session data. 

For the case where a sandbox is deleted, and a new one is provisioned (with a fresh volume), re-seeding makes sense. But we intentionally avoid provisioning new sandboxes because it doesn't recreate prior sandbox's state (e.g if the agent installed dependencies, etc). So the policy we adopt is that once a sandbox is discarded, any conversations for the sandbox are archived (un-resumable).

simonrosenberg:
Closing, cf
```
For the case where a sandbox is deleted, and a new one is provisioned (with a fresh volume), re-seeding makes sense. But we intentionally avoid provisioning new sandboxes because it doesn't recreate prior sandbox's state (e.g if the agent installed dependencies, etc). So the policy we adopt is that once a sandbox is discarded, any conversations for the sandbox are archived (un-resumable).
```

PR #14713, opened 2026-06-08
Add OHE default organization bootstrap

ak684:
HUMAN:
This PR adds login-time default OpenHands organization bootstrap support for OHE installs. I validated it with the companion KOTS config PR on R01: the configured default org was created, configured owners/members were applied, personal workspace org rows were not adopted, and the app plus KOTS admin console stayed healthy.

## Summary

Adds login-time bootstrap support for a configured default OpenHands organization in OHE installs.

- Reads default org settings from environment variables, including owner emails and auto-add behavior.
- Creates the shared org after a configured owner signs in, or reuses an exact matching shared org as an idempotency safety.
- Adds/promotes configured owners and optionally adds signed-in users as members.
- Sets `current_org_id` only for new users who were added to the default org.
- Refuses to adopt personal workspace org rows as default organizations.
- Wires the bootstrap into the Keycloak callback without blocking login if bootstrap fails.

## Companion PR

KOTS config support: OpenHands/OpenHands-Cloud#689.

## Validation

- `PYTHONPATH=".:$PYTHONPATH" poetry run --project=enterprise pytest enterprise/tests/unit/test_default_org_service.py -q`
- `poetry run pre-commit run --show-diff-on-failure --config ./dev_config/python/.pre-commit-config.yaml --files enterprise/server/routes/auth.py enterprise/storage/default_org_service.py enterprise/tests/unit/test_default_org_service.py`
- Successfully validated on R01 with the companion KOTS release: default org creation, owner/member behavior, personal-workspace guard, app HTTP 200, and KOTS admin HTTP 200.

## Notes

The companion KOTS config PR wires these environment variables into OpenHands Enterprise Replicated installs.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-f1ad840   docker.openhands.dev/openhands/openhands:f1ad840
```

### DIFF
--- a/enterprise/server/routes/auth.py
+++ b/enterprise/server/routes/auth.py
@@ -59,6 +59,7 @@ from server.utils.rate_limit_utils import (
 from server.utils.url_utils import get_cookie_domain, get_cookie_samesite, get_web_url
 from sqlalchemy import select
 from storage.database import a_session_maker
+from storage.default_org_service import DefaultOrgBootstrapService
 from storage.user import User
 from storage.user_store import UserStore
 
@@ -330,8 +331,10 @@ async def keycloak_callback(
     user_id = user_info.sub
     user_info_dict = user_info.model_dump(exclude_none=True)
     user = await UserStore.get_user_by_id(user_id)
+    is_new_user = False
     if not user:
         user = await UserStore.create_user(user_id, user_info_dict)
+        is_new_user = True
     else:
         # Existing user — gradually backfill contact_name if it still has a username-style value
         await UserStore.backfill_contact_name(user_id, user_info_dict)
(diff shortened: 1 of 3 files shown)


ak684:
I pushed a small follow-up commit (`2e43d5396`) for the suggestions I agree with:

- Added the explicit `is_new_user: bool = False` annotation in `auth.py`.
- Extracted the LiteLLM provisioning/key retrieval into `_create_member_litellm_api_key()` with a docstring, so the intentional side effect is clearer at the call site.

I’m intentionally leaving the disabled-feature config logging as-is. These env vars are optional and unset by default, so logging missing values while the feature is disabled would add noise to normal OHE installs. We still warn when the feature is enabled but required fields are missing.

I’m also not adding a broader auth integration test in this follow-up; the service unit tests cover the bootstrap behavior, and we validated the login-time flow against the R01 OHE instance.

PR #14724, opened 2026-06-09
Fix OpenHands LLM key access for credited orgs

enyst:
HUMAN:
This PR proposes a quick implementation for the behavior: if a user already has credits, allow the creation of an OpenHands LLM API key from the UI.

In my understanding, the current check is supposed to help with abuse from new accounts. If an account has credits, this PR proposes to treat it as legimate.

AGENT:

---

## Why

Organizations can receive OpenHands credits without having completed a Stripe payment. BYOR/OpenHands LLM key export was gated only by the completed-payment-backed `byor_export_enabled` flag, so credited users could still be blocked from creating an OpenHands LLM API key.

## Summary

- Allow BYOR/OpenHands LLM key export when the org already has a positive LiteLLM credit balance, even if the billing-history flag is false.
- Preserve the existing `byor_export_enabled` short-circuit for orgs that already unlocked key export through billing.
- Add focused OrgService tests for enabled, no-credit, and credited-org paths.

## Issue Number

N/A

## How to Test

- `cd enterprise && PYTHONPATH=".:..:$PYTHONPATH" poetry run pytest tests/unit/test_org_service.py -k "check_byor_export_enabled" --confcutdir=tests/unit -q`
- `cd enterprise && PYTHONPATH=".:..:$PYTHONPATH" poetry run pytest tests/unit/server/routes/test_api_keys.py::TestGetLlmApiKeyForByor tests/unit/server/routes/test_api_keys.py::TestCheckByorPermitted --confcutdir=tests/unit -q`
- `cd enterprise && poetry run pre-commit run --files storage/org_service.py tests/unit/test_org_service.py --show-diff-on-failure --config ./dev_config/python/.pre-commit-config.yaml`

## Video/Screenshots

Not applicable; backend entitlement fix covered by unit tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

@enyst can click here to [continue refining the PR](https://app.all-hands.dev/conversations/6281a0f5-a070-41b3-a77f-5882a2222ca9)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-8745223   docker.openhands.dev/openhands/openhands:8745223
```

### DIFF
--- a/enterprise/storage/org_service.py
+++ b/enterprise/storage/org_service.py
@@ -859,7 +859,11 @@ class OrgService:
             return True
 
         credits = await OrgService.get_org_credits(user_id, org_id)
-        return credits is not None and credits > 0
+        if credits is None or credits <= 0:
+            return False
+
+        org = await OrgStore.enable_byor_export(org_id)
+        return bool(org and org.byor_export_enabled)
 
     @staticmethod
     async def switch_org(user_id: str, org_id: UUID) -> Org:
--- a/enterprise/storage/org_service.py
+++ b/enterprise/storage/org_service.py
@@ -855,7 +855,11 @@ class OrgService:
         if not org:
             return False
 
-        return org.byor_export_enabled
+        if org.byor_export_enabled:
+            return True
+
+        credits = await OrgService.get_org_credits(user_id, org_id)
+        return credits is not None and credits > 0
 
     @staticmethod
     async def switch_org(user_id: str, org_id: UUID) -> Org:
--- a/enterprise/storage/org_service.py
+++ b/enterprise/storage/org_service.py
@@ -858,6 +858,9 @@ class OrgService:
         if org.byor_export_enabled:
             return True
 
+        if await OrgStore.has_completed_billing_session(org_id):
+            return False
+
         credits = await OrgService.get_org_credits(user_id, org_id)
         return credits is not None and credits > 0
 
--- a/enterprise/storage/org_service.py
+++ b/enterprise/storage/org_service.py
@@ -858,9 +858,6 @@ class OrgService:
         if org.byor_export_enabled:
             return True
 
-        if await OrgStore.has_completed_billing_session(org_id):
-            return False
-
         credits = await OrgService.get_org_credits(user_id, org_id)
         return credits is not None and credits > 0
 
--- a/enterprise/storage/org.py
+++ b/enterprise/storage/org.py
@@ -64,6 +64,8 @@ class Org(Base):
     max_budget_per_task: Mapped[float | None] = mapped_column(nullable=True)
     v1_enabled: Mapped[bool | None] = mapped_column(nullable=True)
     conversation_expiration: Mapped[int | None] = mapped_column(nullable=True)
+    # Source of truth for BYOR/OpenHands LLM key export entitlement.
+    # Set by completed billing sessions or when positive org credits are detected.
     byor_export_enabled: Mapped[bool] = mapped_column(nullable=False, default=False)
     sandbox_grouping_strategy: Mapped[str | None] = mapped_column(String, nullable=True)
     # Encrypted column for LLM profiles (contains API keys)
--- a/enterprise/storage/org_store.py
+++ b/enterprise/storage/org_store.py
@@ -18,7 +18,6 @@ from server.routes.org_models import (
 )
 from sqlalchemy import delete, select, text
 from sqlalchemy.orm import joinedload
-from storage.billing_session import BillingSession
 from storage.database import a_session_maker
 from storage.lite_llm_manager import LiteLlmManager, get_openhands_cloud_key_alias
 from storage.org import Org
@@ -116,20 +115,6 @@ class OrgStore:
             result = await session.execute(select(Org).filter(Org.id == org_id))
             org = result.scalars().first()
         return await OrgStore._validate_org_version(org)
-
-    @staticmethod
-    async def has_completed_billing_session(org_id: UUID) -> bool:
-        """Check whether an organization has completed Stripe billing."""
-        async with a_session_maker() as session:
-            result = await session.execute(
-                select(BillingSession.id)
-                .filter(
-                    BillingSession.org_id == org_id,
-                    BillingSession.status == 'completed',
-                )
-                .limit(1)
-            )
-            return result.scalar_one_or_none() is not None
 
     @staticmethod
     async def get_orgs_by_ids(org_ids: list[UUID]) -> list[Org]:
--- a/enterprise/storage/org_store.py
+++ b/enterprise/storage/org_store.py
@@ -117,6 +117,14 @@ class OrgStore:
         return await OrgStore._validate_org_version(org)
 
     @staticmethod
+    async def enable_byor_export(org_id: UUID) -> Org | None:
+        """Persist BYOR export enablement for an organization."""
+        return await OrgStore._update_org_kwargs(
+            org_id,
+            {'byor_export_enabled': True},
+        )
+
+    @staticmethod
     async def get_orgs_by_ids(org_ids: list[UUID]) -> list[Org]:
         """Get multiple organizations by IDs in a single query.
 
--- a/enterprise/storage/org_store.py
+++ b/enterprise/storage/org_store.py
@@ -18,6 +18,7 @@ from server.routes.org_models import (
 )
 from sqlalchemy import delete, select, text
 from sqlalchemy.orm import joinedload
+from storage.billing_session import BillingSession
 from storage.database import a_session_maker
 from storage.lite_llm_manager import LiteLlmManager, get_openhands_cloud_key_alias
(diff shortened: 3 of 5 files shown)


enyst:
/iterate

enterprise/storage/org_service.py
```diff
@@ -855,7 +855,11 @@ async def check_byor_export_enabled(
         if not org:
             return False
 
-        return org.byor_export_enabled
+        if org.byor_export_enabled:
```
malhotra5, line 858: If I'm not mistaken, this implies that even if an admin explicitly _disables_ byor export, we will proceed to check for positive credits nevertheless and enable it anyways? 

(this is under the assumption that the byor flag is a user facing flag, not just a system flag to be used internally)

enyst: @openhands Investigate carefully and answer here in a reply to this inline thread. If the answer is no, show us the source as a github file/line link. If the answer is yes, fix it please, I don't think it should still work if it's admin-disabled.

malhotra5: Also another nit, the column `byor_export_enabled` right now implies that it alone is the boolean tracking byor export enabled. So once positive credits are found, we should set `byor_export_enabled=True` and use it as source of truth. 

Alternatively we rename the column it to imply that it only tracks byor enablement from a billing session (or leave a comment here clarifying this)

enyst: Investigated. There is no user/admin-facing org update field for `byor_export_enabled` in the source: `OrgUpdate` does not expose it (https://github.com/OpenHands/OpenHands/blob/a077ac727b803bda8c8b809c4174a8499e021f5c/enterprise/server/routes/org_models.py#L227-L253). The in-repo writers I found are system paths: completed Stripe checkout sets it to true (https://github.com/OpenHands/OpenHands/blob/a077ac727b803bda8c8b809c4174a8499e021f5c/enterprise/server/routes/billing.py#L294-L296), and legacy user migration initializes it from completed billing history (https://github.com/OpenHands/OpenHands/blob/a077ac727b803bda8c8b809c4174a8499e021f5c/enterprise/storage/user_store.py#L203-L217).

That said, you are right about the dangerous edge case: if an operator/admin had explicitly flipped a previously billing-enabled org back to false, the credits fallback would have bypassed that. I fixed that in a077ac727b803bda8c8b809c4174a8499e021f5c: when `byor_export_enabled` is false and the org has completed billing history, we now return false without checking credits (https://github.com/OpenHands/OpenHands/blob/a077ac727b803bda8c8b809c4174a8499e021f5c/enterprise/storage/org_service.py#L858-L865; billing-history helper at https://github.com/OpenHands/OpenHands/blob/a077ac727b803bda8c8b809c4174a8499e021f5c/enterprise/storage/org_store.py#L121-L132). Credited orgs with no completed billing history still get the intended credits-based unlock.

Validation: focused OrgService BYOR tests, BYOR API-key route tests, and enterprise pre-commit all pass.

_This reply was created by an AI agent (OpenHands) on behalf of the user._

enterprise/storage/org_service.py
```diff
@@ -855,7 +855,14 @@ async def check_byor_export_enabled(
...
+            return True
+
+        if await OrgStore.has_completed_billing_session(org_id):
+            return False
```
malhotra5: Hmm this is a repeat of the `byor_export_enable` flag above (the flag is set to true when a billing session occurs)

Maybe we can simplify this to

```
if org.byor_export_enabled:
    return True

credits = await OrgService.get_org_credits(user_id, org_id)
org.byor_export_enabled = bool(credits and credits > 0)

```


This way `org.byor_export_enabled` is always source of truth (regardless of billing session or user credits)

enyst:
@openhands Revert the last commit. Then read the inline threads and specially this:
> Also another nit, the column byor_export_enabled right now implies that it alone is the boolean tracking byor export enabled. So once positive credits are found, we should set byor_export_enabled=True and use it as source of truth.

> Alternatively we rename the column it to imply that it only tracks byor enablement from a billing session (or leave a comment here clarifying this)

Let's set the export enabled flag so that it stays source of truth; and add a comment clarifying that it may be true from a billing session or from positive credits.

enterprise/storage/org_service.py
enyst: Done in 8745223 after reverting the previous billing-history guard commit. I re-checked the admin-facing path: `OrgUpdate` still does not expose `byor_export_enabled`, so there is no normal user/admin update field for disabling it. The flag is now treated as the source of truth: if already true we return true; otherwise positive LiteLLM credits persist it via `OrgStore.enable_byor_export()` before returning true. I also added a model comment clarifying the flag can be set by completed billing sessions or detected positive credits.\n\n_This reply was created by an AI agent (OpenHands) on behalf of the user._

enterprise/storage/org_service.py
enyst: Done in 8745223. The credits path now persists `byor_export_enabled=True` with `OrgStore.enable_byor_export()` and returns the persisted flag value, so future checks use the column as the source of truth. I also added focused service and store tests for this path plus the requested clarifying column comment.\n\n_This reply was created by an AI agent (OpenHands) on behalf of the user._

enterprise/storage/org_service.py
```diff
@@ -855,7 +855,15 @@ async def check_byor_export_enabled(
...
 
-        return org.byor_export_enabled
+        if org.byor_export_enabled:
+            return True
+
+        credits = await OrgService.get_org_credits(user_id, org_id)
+        if credits is None or credits <= 0:
+            return False
+
+        org = await OrgStore.enable_byor_export(org_id)
```
enyst, line 865: @malhotra5 WDYT?

enyst:
@openhands understand the goal here and investigate deeply any side effects if there still are; read inline threads, both solved and unsolved to understand what behavior we want. Then /codereview and post your feedback as a comment on the PR.

enyst:
@openhands understand the goal here and investigate deeply any side effects if there still are; read inline threads, both solved and unsolved to understand what behavior we want. Then /codereview and post your feedback as a comment on the PR.

enterprise/storage/org_service.py
malhotra5: LGTM!

malhotra5 (approved):
LGTM thanks you!

enyst:
🟢 **Taste Rating: Good taste**

I read the PR diff, the current head (`8745223`), the PR description/template, and the inline review threads including the resolved/outdated ones. The intended behavior from the threads is now clear: `byor_export_enabled` should remain the source-of-truth entitlement flag, but a credited org that has not gone through Stripe should be able to unlock that entitlement by detecting positive LiteLLM credits and persisting the flag.

**Goal / side-effect review**
- The current implementation matches the requested behavior: existing `byor_export_enabled=True` short-circuits; otherwise positive credits call `OrgStore.enable_byor_export()` and return the persisted value.
- The write from the permission check is a deliberate, idempotent entitlement migration: a credited org is converted from “not yet unlocked” to “unlocked,” so later checks use the DB flag without repeatedly consulting LiteLLM.
- The sticky behavior is consistent with the inline direction and the existing billing-session behavior: once positive credits unlock export, later zero/unknown credits do not automatically revoke it because the flag has become source of truth.
- I re-checked the admin-disable concern from the solved thread: `OrgUpdate` does not expose `byor_export_enabled`, so there is no normal user/admin-facing disable path for this flag today. If a future explicit disable mechanism is added, it should use a separate state/field rather than overloading this entitlement flag.
- The effective-org callers in `api_keys.py` still pass the request’s `effective_org_id`; credit lookup requires LiteLLM team membership for that user/org, and failures/no team info keep the permission false without persisting anything.

**Testing / verification**
- The added service and store tests cover the important branches: already-enabled short-circuit, no-credit denial, credited-org persistence, missing user/current org/org cases, and persistence through the store helper.
- I attempted to run the focused OrgService test command locally, but this sandbox’s enterprise Poetry environment did not have `pytest` installed (`Command not found: pytest`). I did not install the full enterprise dependency set just for this review.

**[RISK ASSESSMENT]**
  This touches an entitlement gate for exporting OpenHands LLM keys and introduces an intentional DB write in the permission path. The implementation is small, idempotent, covered by focused tests, and aligned with the reviewed desired behavior; the main risk is semantic, not implementation complexity.

**VERDICT:**
✅ **Worth merging**: The core logic is sound and aligns with the resolved inline-thread direction. I don’t see remaining blocking side effects in the current code.

**KEY INSIGHT:**
The flag is now correctly treated as an entitlement latch: billing sessions or detected positive credits can set it, and once set the DB column remains the single source of truth.

_This PR comment was created by an AI agent (OpenHands) on behalf of @enyst._

PR #14731, opened 2026-06-09
fix: ignore null provider tokens in file secrets store

Fishbone23333:
HUMAN:
This PR makes the function cache file path test cross-platform by using pytest's tmp_path fixture and pathlib.Path instead of hard-coded Unix path assumptions.

AGENT:

---

## Why

`FileSecretsStore.load()` filters persisted provider tokens by reading each token entry's `token` field. Historical or partially invalid `secrets.json` data can contain a provider token entry with a `null` value, which makes the loader call `.get()` on `None` before the `Secrets` model has a chance to filter invalid tokens.

Refs #12714

## Summary

- Skip non-dict provider token entries while loading file-backed secrets.
- Keep valid provider tokens with non-empty `token` values.
- Add a regression test covering null provider token entries.

## Issue Number

Refs #12714

## How to Test

- `python -m pytest tests\unit\app_server\test_file_secrets_store.py -q`
- `python -m ruff check openhands\app_server\secrets\file_secrets_store.py tests\unit\app_server\test_file_secrets_store.py`
- `python -m ruff format --check openhands\app_server\secrets\file_secrets_store.py tests\unit\app_server\test_file_secrets_store.py`

## Video/Screenshots

N/A. Backend-only file secrets store compatibility fix.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

I tested the focused backend path on Windows with the local Python 3.12 conda environment. I did not run the full `make build` or full pre-commit suite because the project development docs recommend Linux/Mac/WSL for the full OpenHands environment.

### DIFF
--- a/openhands/app_server/secrets/file_secrets_store.py
+++ b/openhands/app_server/secrets/file_secrets_store.py
@@ -21,7 +21,7 @@ class FileSecretsStore(SecretsStore):
             provider_tokens = {
                 k: v
                 for k, v in (kwargs.get('provider_tokens') or {}).items()
-                if v.get('token')
+                if isinstance(v, dict) and v.get('token')
             }
             kwargs['provider_tokens'] = provider_tokens
             secrets = Secrets(**kwargs)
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #14732, opened 2026-06-09
perf: skip runtime API call in webhook auth via SandboxRecord

tofarr:
HUMAN:

AGENT: This PR was created by an AI agent (OpenHands) on behalf of the user.

---

## Why

Every webhook callback authenticates via `valid_sandbox`, which called `get_sandbox_by_session_api_key`. In `RemoteSandboxService` that method makes an HTTP round-trip to the runtime API (`GET /sessions/{sandbox_id}`) to populate `status`, `exposed_urls`, and the plain-text session key. `valid_sandbox` only ever reads `id` and `created_by_user_id` — both stored in the app server's own database — so the runtime call was wasted on every webhook request.

## Summary

- Add `SandboxRecord` — a lightweight Pydantic model containing only `id` and `created_by_user_id` (no live runtime data), mirroring the `AppConversationInfo` / `AppConversation` naming pattern.
- Add `get_sandbox_record_by_session_api_key` to `SandboxService` (abstract) and implement it in `RemoteSandboxService` (DB-only hash lookup, no runtime call), `DockerSandboxService`, and `ProcessSandboxService`.
- Update `valid_sandbox`, `valid_conversation`, and `on_conversation_update` in `webhook_router.py` to use the new method and type. The existing `get_sandbox_by_session_api_key` and `SandboxInfo` are unchanged and still used by `session_auth.py`.

## Issue Number

N/A

## How to Test

```bash
pytest tests/unit/app_server/test_webhook_router_auth.py \
       tests/unit/app_server/test_remote_sandbox_service.py \
       tests/unit/app_server/test_sandbox_service.py -q
# 73 passed
```

For integration: send a webhook callback with a valid `X-Session-API-Key` and confirm the app server no longer calls `GET /sessions/{id}` on the runtime API during authentication.

## Video/Screenshots

Human: Callbacks still work and set the title of the conversation.
image

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

`session_auth.py` (`validate_session_key`) keeps using `get_sandbox_by_session_api_key` with the full runtime call intentionally — it enforces `status == RUNNING` as a secondary security check before serving secrets. That path is unchanged.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-d85eefb   docker.openhands.dev/openhands/openhands:d85eefb
```

### DIFF
--- a/openhands/app_server/event_callback/webhook_router.py
+++ b/openhands/app_server/event_callback/webhook_router.py
@@ -37,7 +37,7 @@ from openhands.app_server.event_callback.set_title_callback_processor import (
     SetTitleCallbackProcessor,
 )
 from openhands.app_server.integrations.provider import ProviderType
-from openhands.app_server.sandbox.sandbox_models import SandboxInfo
+from openhands.app_server.sandbox.sandbox_models import SandboxRecord
 from openhands.app_server.services.injector import InjectorState
 from openhands.app_server.services.jwt_service import JwtService
 from openhands.app_server.user.auth_user_context import AuthUserContext
@@ -232,8 +232,8 @@ async def valid_sandbox(
     session_api_key: str = Depends(
         APIKeyHeader(name='X-Session-API-Key', auto_error=False)
     ),
-) -> SandboxInfo:
-    """Use a session api key for validation, and get a sandbox. Subsequent actions
+) -> SandboxRecord:
+    """Use a session api key for validation, and get a sandbox record. Subsequent actions
     are executed in the context of the owner of the sandbox"""
     if not session_api_key:
         raise HTTPException(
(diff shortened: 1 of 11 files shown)


hieptl (approved):
Thank you! 🙏

malhotra5 (approved):
Nice! 

Noting for posterity, I was concerned about the distributing load via the http roundtrips vs consolidating load by querying the DB table on the app server

However, for remote sandboxes it requires http round trips and the remote runtime service queries the runtime table + kubernetes. Considering this, we're actually considerably improving performance. Since we don't care whether the sandbox is actually running and we just need one DB query to find the necessary information regarding the sandbox, this is a better solution 🍰

PR #1339, opened 2024-04-24
Fix build error

xingyaoww:
Fix build error `ERROR:root:<class 'ModuleNotFoundError'>: No module named 'opendevin.parse_commands'` in:
https://github.com/OpenDevin/OpenDevin/actions/runs/8818607652/job/24207798105

Temporarily remove SWE-agent from CodeActAgent until it gets add back in https://github.com/OpenDevin/OpenDevin/pull/1290

PR #14735, opened 2026-06-09
fix(auth): reconcile EMAIL_PATTERN_WHITELIST/BLACKLIST env vars on startup

aivong-openhands:
## Summary

Changes to `EMAIL_PATTERN_WHITELIST` / `EMAIL_PATTERN_BLACKLIST` in the deploy values do **not** take effect on an already-migrated environment. These env vars are only read once, by the one-time Alembic migration `099_create_user_authorizations_table` (`_seed_from_environment()`), which inserts the patterns into the `user_authorizations` table. After that migration has run, editing the env var is inert: the runtime authorizer reads the **`user_authorizations` DB table**, not the env var.

## Changes

This fix implements Option 1 from the issue: **an idempotent reconcile on server startup** that syncs the current env var values to the database.

### Files Changed

1. **`enterprise/storage/user_authorization_store.py`**:
   - Added `get_authorization_by_pattern()` - lookup an authorization by pattern and type
   - Added `upsert_authorization()` - insert a new rule if one doesn't exist with the same pattern and type
   - Added `get_all_patterns_by_type()` - get all rules of a specific type
   - Added `reconcile_from_environment()` - reads EMAIL_PATTERN_WHITELIST/BLACKLIST env vars and upserts all patterns

2. **`enterprise/server/app_lifespan/saas_app_lifespan_service.py`**:
   - Integrated reconciliation into `__aenter__()` so it runs on every server startup
   - Added `_reconcile_user_authorizations()` helper method
   - Handles errors gracefully so startup is not disrupted if reconciliation fails

3. **`enterprise/tests/unit/storage/test_user_authorization_store.py`**:
   - Added `TestUpsertAuthorization` class with tests for upsert behavior
   - Added `TestGetAuthorizationByPattern` class with tests for lookup
   - Added `TestGetAllPatternsByType` class with tests for pattern retrieval
   - Added `TestReconcileFromEnvironment` class with tests for reconciliation

4. **`enterprise/tests/unit/test_saas_lifespan.py`**:
   - Added tests verifying reconciliation is called on startup
   - Added test verifying errors are handled gracefully

## How It Works

1. On every server startup, `SaasAppLifespanService.__aenter__()` calls `UserAuthorizationStore.reconcile_from_environment()`
2. This reads the current `EMAIL_PATTERN_WHITELIST` and `EMAIL_PATTERN_BLACKLIST` env vars
3. For each pattern in the env vars, it checks if a rule with that pattern and type already exists
4. If not, it creates a new rule
5. The reconciliation is idempotent - running it multiple times with the same env vars will not create duplicates

## Testing

All 49 unit tests pass:
- 41 tests for `UserAuthorizationStore` including 15 new tests for upsert/reconcile
- 8 tests for `SaasAppLifespanService` including 3 new tests for reconciliation

## Fixes

Fixes #1339

@aivong-openhands can click here to [continue refining the PR](https://app.all-hands.dev/conversations/89a1c868-29b2-4340-bcc8-551c0e70ec6e)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-610ba16   docker.openhands.dev/openhands/openhands:610ba16
```


neubig:

Thank you for understanding.

PR #14740, opened 2026-06-10
feat: move existing users into the default org when first auto-added

ak684:
HUMAN: On OHE installs that bootstrap a default org, existing users were added as members but stayed parked in their personal workspace until they manually switched. This moves the switch to membership-creation time, so everyone (new users, existing users on first auto-add, and the configured owner) lands in the default org on the sign-in where they first join it. Validated on a real Replicated install (R03): owner bootstrap and a fresh GitLab signup both landed in the default org.

## Summary

The OHE default-org bootstrap (#14713) only switched `current_org_id` for brand-new users; existing users were added as members but stayed parked in their personal workspace until they manually switched. This PR moves the switch to **membership-creation time**, so every user — new or pre-existing — lands in the default organization on the sign-in where they first join it:

- on signup (unchanged behavior),
- on the first auto-add of an existing user (new),
- when the configured owner bootstraps the org by logging in (new).

Joining is a one-time event, so the move never re-fires on later logins: a user who deliberately switches back to their personal workspace stays there. Promoting an already-member configured owner does not move them.

This also means resolver conversations (Jira DC / Bitbucket DC / GitHub / GitLab webhooks) and the automations UI route to the default org for everyone after their next sign-in, since both fall back to the user's `current_org_id`.

## Implementation

- `_ensure_membership` now returns a `MembershipOutcome` (ADDED / PROMOTED / UNCHANGED) instead of a bool.
- `_get_or_create_org` reports whether the org was created in this call with the current user as owner (false on adoption, race recovery, or creation under a different configured owner).
- The KOTS help-text change matching this behavior is in the companion OpenHands-Cloud PR.

## Testing

- `enterprise/tests/unit/test_default_org_service.py`: existing tests updated, plus new coverage for existing-user auto-add (moved), already-member login (not moved), and owner-creates-org (moved). 10/10 passing.

Part of the org-only mode work for OHE (stacked under the hide-personal-workspaces PR).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-1811b34   docker.openhands.dev/openhands/openhands:1811b34
```

### DIFF
--- a/enterprise/storage/default_org_service.py
+++ b/enterprise/storage/default_org_service.py
@@ -3,6 +3,7 @@
 import os
 import re
 from dataclasses import dataclass
+from enum import Enum
 from uuid import UUID
 
 from pydantic import SecretStr
@@ -21,6 +22,14 @@ from openhands.app_server.utils.logger import openhands_logger as logger
 
 _TRUTHY_VALUES = {'1', 'true', 'yes', 'on'}
 _EMAIL_SPLIT_RE = re.compile(r'[\s,;]+')
+
+
+class MembershipOutcome(Enum):
+    """Result of ensuring a user's membership in the default org."""
+
+    ADDED = 'added'
+    PROMOTED = 'promoted'
+    UNCHANGED = 'unchanged'
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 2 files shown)


ak684:
Re: review — applied the one suggestion: condensed the first-join comment block to 3 lines (kept the never-re-fires constraint since it isn't obvious from the condition alone).

PR #14741, opened 2026-06-10
feat: add hide_personal_workspaces flag for org-only OHE installs

ak684:
HUMAN: Adds an env-driven hide_personal_workspaces feature flag for org-only OHE installs — when enabled, the bootstrapped default org is the only workspace users see (personal workspaces are filtered from the org list, the selector hides when one org remains). UI-level by design; flag off restores previous behavior. Validated on a real Replicated install (R03): a second user signed up via GitLab, landed in the default org with no personal workspace or selector visible, and ran working conversations.

## Summary

Adds an env-driven `hide_personal_workspaces` feature flag so OHE installs that bootstrap a default organization can make it the only workspace users see. Stacked on #14740.

**Server**
- New `hide_personal_workspaces` field on `WebClientFeatureFlags`, settable via the structured `OH_WEB_CLIENT_FEATURE_FLAGS_HIDE_PERSONAL_WORKSPACES` env (verified end-to-end with `from_env`) or plain `HIDE_PERSONAL_WORKSPACES` in the default injector.
- While the flag is on, the default-org bootstrap moves members parked on their personal org into the default org on every login — with personal workspaces hidden there is no visible way to be there. Users in another team org are left alone; users who are not members of the default org are never moved (no stranding).

**Frontend**
- Personal workspaces are filtered from the org list/selector, but only when the user belongs to at least one team org — a user whose only workspace is personal (default org not created yet, or not a member) keeps seeing it.
- The org selector is hidden when only one org remains visible.
- Auto-select ignores a backend `current_org_id` that isn't in the visible list.

**Scope: UI-level by design.** The orgs API still returns personal orgs, no data is deleted, and disabling the flag restores the previous behavior. Hard server-side enforcement (rejecting personal-org requests/API keys) is deliberately out of scope for this MVP.

The KOTS checkbox that sets these envs is in the companion OpenHands-Cloud PR.

## Testing

- `enterprise/tests/unit/test_default_org_service.py`: 14/14 passing (new: parked-on-personal member moved; user in another team org untouched; non-member left on personal).
- `frontend/__tests__/hooks/query/use-organizations.test.tsx`: 7/7 passing (new: filtering on, safety valve when personal is the only org, flag off unchanged).
- Scoped tsc + eslint clean.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-45e2dc5   docker.openhands.dev/openhands/openhands:45e2dc5
```


ak684:
Re: review — **applied**: filter personal workspaces before sorting in `use-organizations.ts` (also slightly cleaner). **Not applied**: (1) splitting the 6 `enable_onboarding` indentation fixes out — they correct objectively broken indentation, are isolated in their own commit, and reverting just to re-land them elsewhere is churn; (2) the org-selector condition nit — agreed it's logically subsumed, but the explicit early-return + comment reads clearer, and the reviewer noted behavior is correct.

PR #14742, opened 2026-06-10
fix: end remote sandbox read transactions

he-yufeng:
HUMAN:
This PR keeps RemoteSandboxService from holding an active read transaction while waiting on runtime-api network calls. It snapshots the stored sandbox rows first, rolls back the read transaction, and then uses the detached data for the runtime lookup or pause loop.

## Summary

- snapshot stored remote sandbox rows before leaving the database transaction
- end read-only transactions before runtime-api calls in sandbox search, lookup, batch lookup, and old-sandbox pause selection
- add regression coverage for the DB -> rollback -> runtime-call ordering

Fixes OpenHands/enterprise#30

## To verify

- `python -m py_compile openhands\app_server\sandbox\remote_sandbox_service.py tests\unit\app_server\test_remote_sandbox_service.py`
- `python -m pre_commit run --config dev_config\python\.pre-commit-config.yaml --files openhands\app_server\sandbox\remote_sandbox_service.py tests\unit\app_server\test_remote_sandbox_service.py`
  - On Windows, the project-specific `warn-appmode-oss` hook cannot run because `bash` is not available; ruff, ruff format, and mypy passed under the pre-commit environment.
- `python -m pytest tests\unit\app_server\test_remote_sandbox_service.py -q --basetemp .tmp\pytest-14720 -p no:cacheprovider`

### DIFF
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -95,6 +95,15 @@ class StoredRemoteSandbox(Base):
     )
 
 
+@dataclass(frozen=True)
+class _StoredRemoteSandboxSnapshot:
+    id: str
+    created_by_user_id: str | None
+    sandbox_spec_id: str
+    session_api_key_hash: str | None
+    created_at: datetime
+
+
 @dataclass
 class RemoteSandboxService(SandboxService):
     """Sandbox service that uses HTTP to communicate with a remote runtime API.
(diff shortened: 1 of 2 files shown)


he-yufeng:
Rebased onto main and resolved the conflict. Sandbox cleanup was refactored since I opened this — `pause_old_sandboxes` now pulls the running set from `_get_user_running_sandboxes()` (runtime `/list` + DB cross-reference) instead of querying inline — so I re-grafted the read-transaction handling into that flow: the ids to pause are snapshotted and `_end_read_transaction()` runs before the pause loop, so the slow runtime calls don't hold a DB transaction open. The other call sites (list/get/batch snapshots) are unchanged.

`tests/unit/app_server/test_remote_sandbox_service.py` is green (64 passed), including the updated `test_pause_old_sandboxes_ends_read_transaction_before_pause`.

neubig:

Thank you for understanding.

PR #14744, opened 2026-06-10
feat(acp): add switch_acp_model proxy endpoint for cloud conversations

simonrosenberg:
HUMAN: 

add switch_acp_model proxy endpoint for cloud conversations

- [x] A human has tested these changes. checkbox.

## Summary

- Adds `POST /api/v1/app-conversations/{id}/switch_acp_model` to the app-server, mirroring the existing `switch_profile` endpoint
- The app-server resolves the sandbox, finds the agent-server URL, and proxies `{model}` to the agent-server's `switch_acp_model` (which calls `session/set_model` on the ACP subprocess)
- Persists the new model on the conversation record so the UI chip stays current on subsequent loads

## Why

The agent-server already had `switch_acp_model` (added when per-conversation ACP model switching was built). The OpenHands app-server only had `switch_profile` (for OpenHands LLM switching). Cloud frontends can't reach the agent-server directly, so there was no way to switch models mid-conversation on cloud. This adds the missing proxy.

Companion canvas PR: OpenHands/agent-canvas#1296

## Test plan

- [ ] Local: existing local ACP model switching still works (agent-server endpoint unchanged)
- [ ] Cloud: connect a canvas to a cloud backend with an ACP conversation → model picker now switches models mid-conversation via this new endpoint
- [ ] Paused sandbox returns 409
- [ ] Invalid model returns 400 from agent-server, surfaced as 400

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-c1b85d2   docker.openhands.dev/openhands/openhands:c1b85d2
```


VascoSch92 (approved):
Just one thing:

`detail=e.response.text` forwards the agent-server's raw error body straight to the external caller. Every other `e.response.text` here use the raw body only inside a logger call and return a sanitized detail=`f'Agent server error: {e.response.status_code}'` to the client. 

This line is the only place in the entire router that forwards a raw upstream body into a client-facing HTTPException.

You could do:

```python
  if e.response.status_code in (400, 409, 504):
      raise HTTPException(
          status_code=e.response.status_code,
          detail=f'Agent server error: {e.response.status_code}',
      )
```

If this was intentional: just put a comment.

simonrosenberg:
> Just one thing:
> 
> `detail=e.response.text` forwards the agent-server's raw error body straight to the external caller. Every other `e.response.text` here use the raw body only inside a logger call and return a sanitized detail=`f'Agent server error: {e.response.status_code}'` to the client.
> 
> This line is the only place in the entire router that forwards a raw upstream body into a client-facing HTTPException.
> 
> You could do:
> 
> ```python
>   if e.response.status_code in (400, 409, 504):
>       raise HTTPException(
>           status_code=e.response.status_code,
>           detail=f'Agent server error: {e.response.status_code}',
>       )
> ```
> 
> If this was intentional: just put a comment.

auto-merge was enabled :/ doing a follow-up PR

PR #14745, opened 2026-06-10
feat: route unclaimed automation events to the default org on single-org installs

ak684:
HUMAN: On single-org OHE installs, webhook events from repos without a git claim were silently dropped (Bitbucket DC has no personal-repo fallback), so admins had to claim every project before automations would fire. This adds a default-org fallback to the resolution chain — only when the default org is the sole team org; explicit claims always win. Covered by unit tests; not yet runtime-tested on a real install (the R03 validation instance has no automation service deployed), so the checkbox below is intentionally unchecked for now.

## Summary

Webhook events forwarded to the automation service resolve their target org from the repo's **git claim** (`_resolve_org_context`); without a claim the event is silently dropped. For Bitbucket DC there is no personal-repo fallback (the owner is a `Project`), so on OHE installs that bootstrap a single default org (#14713), admins had to claim every project in **Manage Organization → Git conversation routing** before automations would fire — and a missing claim failed invisibly.

This adds a default-org fallback to the resolution chain:

**claim → (GitHub personal-repo fallback) → default org → drop**

The fallback applies only when:
- a default org is configured (`OPENHANDS_DEFAULT_ORG_ENABLED` + `OPENHANDS_DEFAULT_ORG_NAME`), and
- it is the **only team org** in the install (new `OrgStore.count_team_orgs()`), and
- the configured name doesn't resolve to a personal workspace.

Claims always win when present. The moment a second team org exists, the fallback switches off (5-minute cache TTL) and routing reverts to explicit-claims-only — so events can never cross org boundaries in multi-org installs. Net effect for single-org org-only installs: zero-claim UX, BBDC project events flow straight to the default org's automations; and the failure mode for a missing claim changes from "silently dropped" to "lands somewhere visible".

Result caching follows the existing pattern (Redis, negative results cached) with a short 300s TTL so org-count changes take effect promptly.

## Scope notes

- Jira DC forwarding is unaffected — it already routes by the integration's stored `org_id`.
- Behavioral expansion to state explicitly: on single-org installs, events from **enrolled** repos route to the default org even for projects no automation was written for; unmatched events no-op in the automation service. Enrollment remains a deliberate per-repo act, and signatures are verified before forwarding.
- Standalone off `main` (the default-org config it reads merged in #14713). Complements #14740/#14741 but has no code dependency on them.

## Testing

- `enterprise/tests/unit/server/services/test_automation_event_service.py`: 5 new tests (single-org resolves, disabled config no-ops without DB, multi-org disables, cache hit skips DB, end-to-end unclaimed BBDC payload resolves to default org) — 35/35 passing.
- `enterprise/tests/unit/test_org_store.py`: DB-backed test for `count_team_orgs` excluding personal workspaces — 35 passed/4 skipped (pre-existing skips).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-cc7641f   docker.openhands.dev/openhands/openhands:cc7641f
```

### DIFF
--- a/enterprise/server/services/automation_event_service.py
+++ b/enterprise/server/services/automation_event_service.py
@@ -34,6 +34,8 @@ from server.auth.constants import (
     AUTOMATION_WEBHOOK_SECRET,
 )
 from server.auth.token_manager import TokenManager
+from storage.default_org_service import get_default_org_config
+from storage.org_store import OrgStore
 from storage.redis import get_redis_client_async
 
 from openhands.app_server.integrations.provider import ProviderType
@@ -42,10 +44,13 @@ from openhands.app_server.utils.logger import openhands_logger as logger
 # Cache TTL constants
 ORG_CLAIM_CACHE_TTL_SECONDS = 3600  # 1 hour for org claims (rarely change)
 USER_ID_CACHE_TTL_SECONDS = 86400  # 24 hours for user ID mappings (never change)
+# Short TTL so creating a second team org switches the fallback off promptly.
+DEFAULT_ORG_FALLBACK_CACHE_TTL_SECONDS = 300
 
 # Cache key prefixes (provider is appended dynamically)
 ORG_CLAIM_CACHE_PREFIX = 'automation:org_claim'
 USER_ID_CACHE_PREFIX = 'automation:idp_to_kc_user'
+DEFAULT_ORG_FALLBACK_CACHE_KEY = 'automation:default_org_fallback'
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


ak684:
Re: review — **applied both**: (1) replaced the inline `user_{org.id}_org` string check with a shared `is_personal_workspace_org()` promoted to module level in `default_org_service.py` (the existing private staticmethod now delegates to it), so the naming convention lives in one place; went with this rather than a new Org model field/OrgStore method since that's schema/API surface this PR doesn't need. (2) Clarified the docstring: the fallback requires exactly one team org, and with zero team orgs nothing resolves because the default org hasn't been created yet.

PR #14757, opened 2026-06-11
Skip personal-org automation dispatch while personal workspaces are hidden

ak684:
## Summary

Webhook events from personal repos resolve to the owner's *personal org* in the automation forwarding path. With `hide_personal_workspaces` on (now the default for org-enabled OHE installs), those automations kept dispatching runs — burning LLM spend — into a workspace the owner can no longer see.

The personal-org fallback in `_resolve_org_context` now skips while hiding is on. Because the default-org fallback sits right below it in the resolution chain (claim → personal → default org), the event **routes to the visible default org instead of dispatching invisibly or being dropped** — which is also where the user's attention now lives.

Design notes:
- **Stateless skip, not stateful pause** — no automation rows are mutated, so turning the hide option off restores personal routing immediately, matching how hiding treats all other personal-workspace data.
- Claims and team-org routing are unaffected; the skip only touches the `owner_type == 'User'` personal fallback.
- Reads the same `HIDE_PERSONAL_WORKSPACES` env the bootstrap, web client, and the personal-API-key guard (#14756) already use.

Known residual, deliberately out of scope: *scheduled* automations created in a personal workspace before hiding still run on their cron inside the automation service, which doesn't see this env. Covering those needs a small change in the automation repo (skip when `org_id == user_id` and the env is set) plus chart env wiring — worth doing only if a brownfield customer actually has scheduled personal automations.

Context: brownfield-only exposure (personal automations predating org-only mode); fresh installs have none. Companion to #14756.

## Testing

- 3 new tests in `enterprise/tests/unit/server/services/test_automation_event_service.py`: hiding on + no default org → personal resolution never consulted, event drops; hiding on + default org → event routes to the default org; hiding off → personal routing unchanged. 38/38 passing.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-7d94f0f   docker.openhands.dev/openhands/openhands:7d94f0f
```

### DIFF
--- a/enterprise/server/services/automation_event_service.py
+++ b/enterprise/server/services/automation_event_service.py
@@ -212,14 +212,25 @@ class AutomationEventService:
         # Try to resolve via OrgGitClaim
         org_id = await self._resolve_git_org(provider, git_org_name)
 
-        # Fallback for personal repos (owner_type indicates individual user)
+        # Fallback for personal repos (owner_type indicates individual user).
+        # Skipped while personal workspaces are hidden: dispatching there
+        # would run automations (and spend) in a workspace the owner can no
+        # longer see. The default-org fallback below routes the event to the
+        # visible org instead; unhiding restores the personal routing.
         if not org_id and owner_type == 'User':
-            org_id = await self._resolve_personal_org(provider, owner_id)
-            if org_id:
+            if get_default_org_config().hide_personal_workspaces:
                 logger.info(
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #14758, opened 2026-06-11
feat: expose copyable invite links so invitations work without email delivery

ak684:
HUMAN:
OHE installs have no email provider, so org invitations were created but never delivered and invitees could never join. This surfaces the invite links to admins (in the invite modal and as pending rows in the members list, with revoke), fixes the schemeless-link WEB_HOST bug, and explains auto-add in the modal. Manually click-tested end-to-end on our Replicated test install (invite → copy link → incognito accept).

## Summary

Org invitations notify invitees exclusively through the Resend API (`RESEND_API_KEY`), which OHE installs don't configure. The result was the worst kind of broken: the invitation row was created, the email send was **silently skipped** (`email_service.py` logs and continues by design), the inviter saw success — and the invitee had no possible way to learn about the invitation (no inbox, no visible link, token only in the DB). It also meant an org with auto-add disabled had **zero working membership paths** on OHE.

The acceptance flow itself already works from a plain URL (`/api/organizations/members/invite/accept?token=…`); only token *delivery* was missing. So this PR exposes the link to the people who already hold the invite permission:

**Backend**
- `InvitationResponse` now includes `invite_url`; `BatchInvitationResponse` includes `email_delivery_configured` (false when no email provider is set up).
- New `GET /api/organizations/{org_id}/members/invite` lists an org's pending invitations with links — gated on `INVITE_USER_TO_ORGANIZATION` (admins/owners), so tokens are never exposed to regular members.
- `EmailService.build_invitation_url` normalizes a bare `WEB_HOST` to `https://` — **bug fix**: the OHE chart sets `WEB_HOST` as a bare hostname, which produced schemeless, broken links in invitation emails even where email was configured.

**Frontend**
- The invite modal keeps showing the created invitations with copy-link buttons instead of closing when email delivery isn't configured (or some invites failed), with a clear "email delivery is not configured — share these links directly" hint. With email configured and all invites successful, behavior is unchanged (modal closes, email does the work).
- Pending invitations render as **rows in the members list** (admins/owners only): gray email + "Invited" status chip, with copy-link and **revoke** actions. No separate box.
- New `DELETE /api/organizations/{org_id}/members/invite/{invitation_id}` endpoint revokes a pending invitation (the `revoked` status existed in the model but nothing ever set it); same permission gate as create/list.
- The pending-invitations response carries `auto_add_enabled` (true when the org is the bootstrapped default org with auto-add on); the invite modal uses it to explain that sign-in alone already adds users — invitations only matter for pre-assigning a role.

Works air-gapped by construction. Companion follow-up (accept-by-email-match at login) makes acceptance SSO-native.

## Testing

- Backend: 4 new tests (`is_configured`, URL normalization for bare host and explicit scheme) — invitation suites 51/51 passing.
- Frontend: modal tests updated to the real response shape via MSW handlers + new test for the links view (stays open, copy button present, `onClose` not called) — 5/5 modal, 23/23 members page, 28/28 total invite-related.
- Enterprise ruff + eslint + scoped tsc clean.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-042b963   docker.openhands.dev/openhands/openhands:042b963
```


ak684:
Addressed the partial-failure edge case in d08e7e8d7: kept the modal-open-on-partial-failure behavior (the inviter needs to see which emails failed and why — a toast can't carry that), but the "share these links" hint is now only shown when no email provider is configured, since it's inaccurate when emails actually went out. Added the suggested partial-failure test.

PR #14759, opened 2026-06-11
feat: accept pending invitations by verified-email match at sign-in

ak684:
HUMAN:
Companion to #14758: accepting an invitation required the emailed/copied token link; this auto-accepts pending invitations at sign-in when the IdP-verified email matches, making acceptance SSO-native. Manually tested on our Replicated test install (invite, sign in with the invited email without using the link, invitation flips to accepted with the invited role).

## Summary

Stacked on #14758. Invitation acceptance previously had exactly one path: the emailed token link — undeliverable on installs without an email provider. But the IdP already verifies the address an invitation was sent to, which carries the same assurance as clicking that link. So the auth callback now accepts pending invitations matching the user's verified email automatically at sign-in: invite someone, they sign in (or sign up), they're in — no token delivery needed. SSO-native.

Behavior details:

- **Runs before the default-org bootstrap**, so an invitation's role (e.g. admin) wins over the bootstrap's auto-add member default for the same org.
- **Clears the ghost-pending wart**: when default-org auto-add joins an invitee before they accept, the now-redundant invitation is marked accepted instead of sitting pending forever in the new pending-invitations list.
- **Sets the current org on accept** — but only for users parked on their personal workspace; someone deliberately working in another team org isn't yanked out (mirrors the bootstrap's one-time-move semantics).
- Expired invitations are marked expired and skipped; failures never break login (same isolation pattern as the existing token-path handling in the callback).
- The token-link path is unchanged and continues to work.

Security note for review: this widens acceptance from "holds the emailed token" to "signed in with the invited email, as verified by the IdP". For SSO-backed enterprise installs these are equivalent trust anchors (the token link's security also reduces to control of the email account). The email match is exact and case-insensitive, identical to the check the token path already enforces.

## Testing

- 5 new service tests: match → membership + accepted + moved off personal; already-member → ghost cleared, nothing added; expired → marked + skipped; user in another team org → not moved; no email → no-op. 16/16 in the service suite.
- Adjacent suites green: auth invitation callback 27/27, invitation store/router unaffected.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-f1d9d0e   docker.openhands.dev/openhands/openhands:f1d9d0e
```


ak684:
Addressed the review in 5f57f3271 (+ the post-squash refresh of #14758). Disposition of each point:

**✅ Testing gap — done.** Added `test_litellm_failure_skips_invitation_and_continues`: a LiteLLM outage skips that invitation (it stays pending and is retried at the next sign-in) without raising into the auth callback, and the rest of the batch still processes.

**⚠️ Critical (partial failure / inconsistency) — addressed by documenting the existing invariant, not by reordering.** The flow is deliberately ordered with the status flip LAST so that every mid-flight failure self-heals on the next sign-in:
- LiteLLM or `add_user_to_org` failure → invitation stays pending → retried at next sign-in.
- `add_user_to_org` succeeds but the status update is lost → the already-member branch at the top of the loop reconciles it (marks the invitation accepted without re-adding) on the next sign-in.

The suggested reorder (update status *before* adding the member) is the one ordering that breaks this: a failure after the flip would consume the invitation with no membership and no retry path. A DB transaction can't span the LiteLLM HTTP call either. Made the invariant explicit with a comment at the ordering-sensitive spot.

**❌ Double email normalization — declined.** The caller normalizes for its own comparison/short-circuit; the store normalizes at its API boundary. Both are cheap, and removing either sets a trap for the next caller of `get_pending_invitations_for_email`.

PR #14761, opened 2026-06-11
fix(acp): always show agent-type selector regardless of enable_acp flag

simonrosenberg:
HUMAN:

Removes the `isAcpEnabled` feature flag gate on the agent-type selector

- [x] A human has tested these changes. checkbox.

## Summary
- Removes the `isAcpEnabled` feature flag gate on the agent-type selector (OpenHands / ACP dropdown)
- The selector is now always visible so users can switch back to OpenHands even when `enable_acp` is off

## Why
If a user was previously set to ACP and the `enable_acp` flag is disabled, the selector was hidden — leaving them stuck with no way to revert to the OpenHands agent.

## How to Test
1. Ensure `enable_acp` is `false` in the web-client config
2. Open Settings → Agent
3. Confirm the agent-type dropdown (OpenHands / ACP) is visible
4. Confirm switching to OpenHands and saving works correctly

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-8b13edb   docker.openhands.dev/openhands/openhands:8b13edb
```

### DIFF
--- a/frontend/src/routes/agent-settings.tsx
+++ b/frontend/src/routes/agent-settings.tsx
@@ -86,6 +86,7 @@ export default function AgentSettingsScreen() {
     settings?.agent_settings_schema,
   );
 
+  const isAcpEnabled = !!config?.feature_flags?.enable_acp;
   const acpProviders = config?.acp_providers ?? EMPTY_ACP_PROVIDERS;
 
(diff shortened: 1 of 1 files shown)


simonrosenberg:
not scheduled

PR #14762, opened 2026-06-11
feat(acp): enable ACP UI by default

simonrosenberg:
## Summary

Default `ENABLE_ACP` to `true` so the ACP agent selection UI is visible without explicit env-var opt-in. Can be suppressed per-deployment with `ENABLE_ACP=false`.

## Why

Needed to validate ACP agents on a Cloud Canvas feature env (agent-canvas#1290). The feature env doesn't set `ENABLE_ACP=true`, so `isAcpEnabled` was always `false` and the ACP UI was invisible.

## How to Test

Deploy to a feature env (no `ENABLE_ACP` env var) — the Agent Settings page should show the ACP agent picker.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-de288e6   docker.openhands.dev/openhands/openhands:de288e6
```

### DIFF
--- a/enterprise/server/config.py
+++ b/enterprise/server/config.py
@@ -74,6 +74,7 @@ class SaaSServerConfig(ServerConfig):
     enable_jira_dc = ENABLE_JIRA_DC
     enable_linear = ENABLE_LINEAR
     enable_onboarding = os.environ.get('OH_ENABLE_ONBOARDING', 'false') == 'true'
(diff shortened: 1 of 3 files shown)


simonrosenberg:
ACP will not go live with OpenHands, only with canvas

PR #14769, opened 2026-06-11
fix: PLTF-2895 release DB transactions before runtime API I/O in RemoteSandboxService

aivong-openhands:
## Problem

`RemoteSandboxServiceInjector` holds the request-scoped DB session open for the entire request, and the session opens a transaction lazily on the first query and keeps it until the request-end commit. Every runtime API call `RemoteSandboxService` makes therefore runs with an open transaction, leaving the session **idle in transaction** for the duration of the network I/O. In production these sessions sit idle for 2–17+ minutes, pinning pooled connections and blocking autovacuum on `v1_remote_sandbox` (dead tuples accumulate).

This is the same class of issue fixed for `poll_agent_servers` in #14637, which established the "fetch → release → network → re-acquire → write" pattern. That PR did not cover the request-scoped `RemoteSandboxService` methods: `search_sandboxes`, `get_sandbox`, `get_sandbox_by_session_api_key`, `batch_get_sandboxes`, `start_sandbox`, `pause_sandbox`, `resume_sandbox`, `delete_sandbox`, and `pause_old_sandboxes` all perform runtime API I/O mid-transaction.

## Fix

Add `RemoteSandboxService._release_db_transaction()`, which commits the session's current transaction (if any), and call it immediately before each runtime API call site. The session is created with `expire_on_commit=False`, so already-loaded ORM instances remain usable; the session starts a fresh transaction on its next query.

Placement preserves the existing write semantics:

- **`start_sandbox`**: the transaction is released *before* `session.add(stored_sandbox)`, so the insert stays pending while `/start` runs and remains atomic with the request — a failed start still rolls the row back, exactly as before.
- **`delete_sandbox`**: released *before* `session.delete(...)`, so the delete stays pending until the request commits, as before.
- **`pause_sandbox`**: released *after* clearing `session_api_key_hash`, which persists the security invalidation immediately (fail-safe direction — errors after this point were already swallowed and committed at request end).
- Read paths (`search_sandboxes`, `get_sandbox`, `get_sandbox_by_session_api_key`, `batch_get_sandboxes`, `pause_old_sandboxes`) release after materializing results, before the runtime API call.

Note: because the session is shared across services within a request, releasing also commits any writes pending from earlier in the request slightly earlier than before. Call sites were audited (`sandbox_router`, `app_conversation_router`, `live_status_app_conversation_service`) — prior pending state at these points is limited to progress/status writes that were committed at request end regardless of the sandbox call's outcome.

## Tests

- New `TestTransactionReleasedBeforeNetworkIO` suite models lazy transaction state on the mocked session and asserts, for every service method that performs runtime API I/O, that no transaction is open at the moment the network call fires. It also asserts the `start_sandbox` insert is still pending (uncommitted) when `/start` fires.
- Two existing tests asserting "the service never commits" now stub `in_transaction() → False` so they keep verifying that the service does not commit its own writes when no transaction needs releasing.
- All 58 tests in `tests/unit/app_server/test_remote_sandbox_service.py` pass; ruff and mypy pre-commit hooks pass.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-af6bc47   docker.openhands.dev/openhands/openhands:af6bc47
```


aivong-openhands:
@openhands fix conflicts

aivong-openhands:
@openhands fix conflicts

aivong-openhands:
@openhands fix merge conflicts

aivong-openhands:
@openhands?

aivong-openhands:
@openhands fix lint

aivong-openhands:
@openhands fix merge conflicts

smolpaws (commented):

Taste rating: acceptable

Findings:
- Changed area reviewed: openhands/app_server/sandbox/remote_sandbox_service.py, tests/unit/app_server/test_remote_sandbox_service.py.
- I did not see concrete evidence of supply-chain abuse, credential theft, prompt injection or a correctness issue that should block this PR in the diff I reviewed.

Risk assessment: medium.

Verdict: worth merging if CI passes

tofarr:
## Data integrity review

The idle-in-tx fix is sound, but `_release_db_transaction()` commits *whatever* is pending on the request-scoped session — and that session is shared across every service injected for the request. So the new release points don't just release `RemoteSandboxService`'s own transactions; they flush pending writes from any earlier service call. That breaks atomicity guarantees that the old code provided.

### 1. `delete_app_conversation` is no longer atomic across sub-conversation deletes (concrete regression)

In `live_status_app_conversation_service._delete_sub_conversations` (line 2341), each iteration does:

```python
await self._delete_from_agent_server(sub_conversation)  # calls get_sandbox internally
await self._delete_from_database(sub_conversation)      # runs DELETEs that never commit internally
```

- `delete_app_conversation_info` (`sql_app_conversation_info_service.py:624`) and `delete_app_conversation_start_tasks` (`sql_app_conversation_start_task_service.py:250`) run raw `DELETE` statements and never call `commit()`.
- `_delete_from_agent_server` → `self.sandbox_service.get_sandbox(...)` now runs `_release_db_transaction()` before its runtime call, committing the previous iteration's pending DELETEs.

So sub-conversation DELETEs are now committed one-at-a-time, interleaved with `get_sandbox` calls. New failure modes that didn't exist before:

- **DB connection drop mid-loop**: subs 1..k committed, k+1..N and parent untouched.
- **Caller commit failure** at `app_conversation_router.py:996`: only the parent delete rolls back; subs stay committed.
- **Unhandled exception in `_delete_from_agent_server` for the parent** after the loop: same as above.
- **Injector rollback on request-handler exception**: undoes only the parent, subs stay committed.

The per-iteration `try/except` masks DB errors per-sub, so partial commits become durable state without obvious signal.

### 2. `delete_sandbox` is no longer atomic across its two writes

Old code: hash invalidation + row delete both pending → caller commits both atomically.

New code: hash invalidation committed by `_release_db_transaction()` *before* the network call; only the row delete is left for the caller. If the caller's commit fails (e.g. `_finalize_sandbox_delete` at `app_conversation_router.py:911`), the hash stays NULL and the row remains. Old behavior rolled both back. The PR comment labels this "fail-safe", but it's only fail-safe for the transient-runtime-error path (`test_delete_sandbox_transient_error_invalidates_session_key` covers that). The caller-commit-failure path is a new failure mode with no test.

### 3. Audit claim in PR description is wrong for two of the three audited paths

> "prior pending state at these points is limited to progress/status writes that were committed at request end regardless of the sandbox call's outcome"

Two of the audited paths set `set_db_session_keep_open(True)`:

- `start_app_conversation` at `app_conversation_router.py:376`
- `delete_app_conversation` at `app_conversation_router.py:999`

In those flows the injector does **not** commit at request end — the session is handed to a background task. Commits in those flows happen at service-internal points (`save_app_conversation_start_task` commits at `sql_app_conversation_start_task_service.py:247`, `save_app_conversation_info` at `sql_app_conversation_info_service.py:392`). The new release points are interleaved with those, which is the regression in #1.

### 4. Concurrent observers see partial state

Old code: one transaction → all-or-nothing for outside observers. New code: a list-conversations query from another request can see sub-conversations vanishing one-by-one while the parent remains. Eventually consistent, observably non-atomic.

### What's fine

- `start_sandbox` insert atomicity preserved (`add()` happens after `_release_db_transaction()`).
- `pause_sandbox`'s earlier hash commit matches the stated goal and produces the same end state as the old code.
- `expire_on_commit=False` keeps loaded ORM instances usable after the early commit.

### What I'd ask before merging

1. Did the audit actually trace `_delete_from_agent_server → get_sandbox → _release_db_transaction` while `_delete_from_database` had pending DELETEs queued on the shared session? "Pending state" was apparently scoped to progress/status writes, not raw DELETEs from other services.
2. Is the `delete_sandbox` caller-commit-fails case intentional? If yes, document and test it; if not, the release needs a different trigger.
3. Could `_release_db_transaction` be re-scoped to only commit *this service's* pending writes — e.g. per-service savepoint, or per-service session? That preserves the idle-in-tx fix without the cross-service atomicity hit.
4. `TestTransactionReleasedBeforeNetworkIO` only asserts RemoteSandboxService's own tx is released; it doesn't model a shared session with pending writes from another service, so none of the above regressions are visible to the tests. A test that seeds pending DELETEs on the same session and then calls `RemoteSandboxService.get_sandbox` would catch #1 directly.

tofarr (changes requested):
See bot comment...

neubig:

Thank you for understanding.

aivong-openhands:
@tofarr thanks for the careful review — I dug into all four points and they hold up. Summary of findings and what I've done about them:

**1. Audit missed the cross-service DELETEs — confirmed.** The audit traced progress/status writes, but `_release_db_transaction()` commits the *shared, request-scoped* session, so it also flushes writes other services queued on it. The clearest case is `delete_app_conversation`, whose loop interleaves, per (sub)conversation:

```
_delete_from_agent_server(...)  ->  sandbox_service.get_sandbox(...)   # would commit
_delete_from_database(...)      ->  raw DELETEs, no internal commit    # left pending
```

`_delete_from_database`'s `DELETE`s have no internal commit, so each iteration's `get_sandbox` commits the previous iteration's pending deletes — the multi-step delete stops being all-or-nothing. I reproduced this with a real-DB (SQLite, no mocking) regression test that seeds an uncommitted sibling `DELETE` on the shared session, calls `get_sandbox`, and checks from a separate connection that it was **not** committed. It's red against this PR's approach.

**2. `delete_sandbox` two-write split — real behavior change; needs a product call.** Clearing the hash is committed before the network stop, while the row delete is left for the caller, so a caller-side failure after the stop can leave `row present + hash NULL`. That state is benign in isolation (the key is already revoked, and it's indistinguishable from a normally paused/orphaned sandbox), but whether "revoke stays committed even if the caller rolls back" is an *intended new guarantee* is a design decision, not something I wanted to silently bake in. Happy to go whichever way you prefer — document + test the new behavior, or reorder so both writes stay atomic.

**3. Per-service scope: savepoints no, dedicated read session yes.** A `begin_nested()` savepoint doesn't help — a connection with any open (sub)transaction is still `idle in transaction` at the Postgres level, so it neither releases the pooled connection nor unblocks autovacuum. A **dedicated, short-lived read session** does work: do the stored-row read on its own connection so the runtime API call never spans (or commits) the shared request transaction. That fixes the idle-in-tx problem *and* preserves sibling-write atomicity. I've prototyped it for `get_sandbox` and it turns the finding-#1 test green.

**4. Add a test that seeds pending DELETEs then calls get_sandbox — done.** That's exactly the regression test described above (`TestSharedSessionAtomicityDuringDelete`).

Since the app-server code has moved to the `OpenHands/enterprise` repo, I've opened the read-session fix + regression test as a PR there rather than pushing here: **OpenHands/enterprise#142** (references this PR and your feedback). The remaining read paths (`search_sandboxes`, `get_sandbox_by_session_api_key`, `batch_get_sandboxes`, `pause_old_sandboxes`) are tracked in #16374, and `delete_sandbox` (point 2) is intentionally left for the decision above.

---
_This comment was created by an AI agent (OpenHands) on behalf of @aivong-openhands._

PR #14773, opened 2026-06-12
feat: OHE multi-model LLM discovery + BYOK gating (proxy discovery, catalogue union, UX)

ak684:
HUMAN: Backend half of the OHE managed multi-model catalog. Verified by unit tests; VM click-test pending.

## Consolidated: now includes the full OHE multi-model + BYOK UX (formerly #14774)

This PR was the proxy-model-discovery base; #14774 has been merged into it so the whole OHE multi-model story lands as one PR. Added on top of the original backend service (described below):

- **BYOK gating** (`allow_user_llm_configuration`): hides custom model/base-URL/API-key inputs + lands on basic view when off; defaults on (SaaS/existing installs unchanged).
- **BYOK-on catalogue union (P4):** managed proxy models ∪ full SDK provider catalogue → OHE-BYOK-on matches SaaS; managed stays verified, collision-guarded.
- **Single-provider hide:** auto-select + hide the provider picker when there's only one provider.
- **Basic/Advanced toggle hidden when BYOK off** (`allowAdvancedView` gate) — they'd otherwise be identical.
- **Verified-flag fix** (managed proxy models read as verified → flat dropdown) + data-driven Verified/Other headers.
- **Unavailable-model + hidden-alias handling** (canonical hydration, "no longer available" hint).
- **Copy:** `ORG_DEFAULTS_INFO` reworded (drops the BYOK-off-misleading "add their own credentials"), updated across **all 15 locales** for consistency.

Validated on R02 (real KOTS release): BYOK off = managed-only/flat/no-picker/no-toggle; BYOK on = full catalogue + picker + advanced. Tests: enterprise proxy 22, app_server 23, frontend 91, sdk-section-page 12.

---

## Summary

Adds a LiteLLM-proxy-backed model discovery service for self-hosted enterprise (OHE) installs. Today the OHE model dropdown is fed by the SaaS `verified_models` DB table unioned with litellm's static Python catalogue, so it shows models that don't exist on the customer's bundled proxy while never showing models that do. This service makes the proxy the single source of truth: the dropdown shows exactly the models registered on the bundled LiteLLM proxy.

## Changes

- New `enterprise/server/verified_models/litellm_proxy_model_router.py` with `LiteLLMProxyModelService` + `LiteLLMProxyModelServiceInjector`, activated via `OH_LLM_MODEL_KIND=server.verified_models.litellm_proxy_model_router.LiteLLMProxyModelServiceInjector` (set by the companion OpenHands-Cloud chart change).
- Queries `GET {LITE_LLM_API_URL}/model/info` with the existing master key; exposes each visible `model_name` as `openhands/<name>`; dedupes duplicate names (load-balanced deployment groups); never propagates `litellm_params`.
- Hidden-alias availability: entries tagged `model_info.openhands_hidden` (legacy alias routes the proxy keeps serving so existing saved settings keep routing after a rename, e.g. `custom-llm`, `bedrock`) are not dropped — they're carried in a new `ModelsResponse.hidden_models` channel and surface from `/config/models/search` as items with `hidden: true`. Clients keep them out of dropdown options but count them as available, so the frontend's "model no longer available" badge doesn't false-positive on working legacy profiles. A load-balanced name is hidden only when *all* of its entries are hidden. Both new fields default empty/false, so SaaS and default discovery are unchanged.
- No union with the static catalogue, CLARIFAI list, or the `verified_models` table — the response is exactly what the proxy serves.
- Default model derived from `LITELLM_DEFAULT_MODEL` (`litellm_proxy/` → `openhands/`).
- Resilience: 60s TTL in-process cache with stampede lock; on fetch failure serves the last-good result and logs; cold failure returns an empty list without raising. Conversations in flight are unaffected (they hold their model string).

## Testing

- Unit tests (`enterprise/tests/unit/test_verified_model/test_litellm_proxy_model_service.py`): mapping, hidden-channel reporting, dedupe + mixed hidden/visible duplicates, default translation, cold/stale failure paths, TTL behavior. Full test dir: 31 passed.
- SaaS is unaffected: the service only runs when `OH_LLM_MODEL_KIND` selects it, and the new `hidden_models`/`hidden` fields default to empty/false everywhere else.

Stacked frontend UX PR and OpenHands-Cloud chart PRs follow.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-621c095   docker.openhands.dev/openhands/openhands:621c095
```


ak684:
Thanks for the review. What I changed and what I left, with reasoning:

**✅ Fixed — `_derive_default_model` docstring incomplete.** Agreed. The docstring now spells out all three cases (`litellm_proxy/` → `openhands/`, an already-`openhands/`-prefixed default kept as-is, and a bare name prefixed). Commit `97764b9`.

**❌ Not changed — "unused import `Any`".** This one is incorrect: `Any` is used at `litellm_proxy_model_router.py:133` (`model_info: Any = entry.get('model_info') or {}`). Removing it (or applying the suggested `from typing import AsyncGenerator, ClassVar`) would raise `NameError` at import-eval of that annotation. Left as-is.

**ℹ️ No action — class-level `ClassVar` cache.** Agreed it's intentional/correct; the `reset_cache` fixture already handles test isolation. Nothing to change.

PR #14774, opened 2026-06-12
feat: gate BYOK LLM configuration and flag unavailable models for OHE

ak684:
HUMAN: Frontend half of the OHE managed multi-model catalog, stacked on #14773. Now also brings OHE BYOK-on to SaaS parity (catalogue union) and hides the provider picker when there's only one provider. Validated programmatically on R02 for both BYOK states; human click-test on R02 pending.

## Update — multi-model parity (BYOK-on catalogue union + single-provider hide)

Two follow-on commits land the rest of the "sane multi-model on OHE" goal:

- **BYOK-on unions the SDK catalogue (P4).** When `allow_user_llm_configuration` is on, `LiteLLMProxyModelService` now appends the full SDK provider catalogue to the managed proxy models, so OHE-BYOK-on matches SaaS-cloud: users can browse every provider and bring their own key. The managed proxy models stay the verified/default set; catalogue models land **unverified** ("Other"). A bare-name collision guard (require the `openhands/` prefix) keeps a catalogue `openai/gpt-5` from being falsely marked verified by a proxy `gpt-5`. BYOK **off** is unchanged — managed proxy models only. On a cold proxy failure with BYOK on, the catalogue is still served so a BYOK user isn't blocked.
- **Single-provider picker hide (P6).** With exactly one provider (managed OHE's bundled proxy when BYOK is off) the provider dropdown is auto-selected and hidden — nothing to pick. The picker renders during load and whenever there's >1 provider, so multi-provider (BYOK-on, SaaS) configs are unaffected.
- Pairs with the earlier verified-flag fix (scoped `_is_model_verified` so managed proxy models read as verified → the Verified/Other split collapses to a flat list on OHE).

## Summary

Stacked on #14773. Makes the OHE LLM settings experience match what managed installs need: users see the model list their admin configured, BYOK (custom model/base URL/API key) becomes an admin-gated option, and a saved model that no longer exists on the proxy gets a truthful inline warning.

## Changes

- `allow_user_llm_configuration` web-client feature flag, read from `OH_ALLOW_USER_LLM_CONFIGURATION` (same pattern as `hide_personal_workspaces`, #14741). **Defaults to true when unset** — SaaS and existing installs are byte-identical to today.
- When the flag is false, the LLM settings form (personal and org scopes) hides the custom-model input, base-URL input, API-key input, and SaaS key-help links, and lands on the basic view. This hides editing UI only: saved BYOK settings keep loading and routing at runtime untouched.
- Unavailable-model indicator: if the saved/selected model is absent from the models API response (and the list is non-empty), a small "No longer available" hint renders under the model selector. Saving is never blocked.
- Hidden-alias availability: models flagged `hidden: true` by `/config/models/search` (legacy alias routes a managed LiteLLM proxy still serves after a rename, see #14773) never appear as dropdown options, but they count as available — a saved profile pointing at a working alias shows no warning, while a truly absent model still does.
- i18n key `SETTINGS$MODEL_NO_LONGER_AVAILABLE` added across all locales per repo convention.
- Canonical display for legacy alias profiles: editing a profile whose stored model is a hidden alias with a `canonical` mapping hydrates the form with the canonical id (`openhands/<canonical>`), so the dropdown shows a real selection; a pristine save persists the canonical name (deliberate migration — both names route on the proxy). Translation resolves before the form opens (shared provider-models cache), so it can never re-hydrate over in-progress edits; hidden models without a canonical keep today's behavior.

## Testing

- Backend: 3 new flag tests, suite 51 passed.
- Frontend: `llm-settings` + `model-selector` suites **91 passed** (incl. the single-provider-hide cases); `npm run lint` (tsc + eslint + prettier) clean.
- Enterprise proxy service: **22 passed** (new `TestByokUnion`: BYOK-off ignores catalogue, BYOK-on unions + dedups, only-managed-verified collision guard, multi-provider exposure, catalogue-served-when-proxy-down); enterprise ruff clean.
- **R02 (self-hosted OHE) programmatic validation** on `sha-736ad0b`: BYOK off → 1 provider (`openhands`, verified), 4 models all verified (flat, no picker); BYOK on → `openhands` + full SDK provider catalogue, only the 4 managed models verified, catalogue unverified.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-26df4e6   docker.openhands.dev/openhands/openhands:26df4e6
```


ak684:
Thanks for the review. What I changed and what I left, with reasoning:

**ℹ️ Not changed (intentional) — async canonical resolution before `openForm`.** The pre-mount `await resolveCanonicalModel(...)` is deliberate and is the safer of the two options the review raises. `SdkSectionPage` re-hydrates and resets dirty state whenever its initial-value overrides change, so resolving the canonical id in a post-mount `useEffect` would clobber in-progress user edits — the exact bug this ordering avoids (see the comment at `llm-settings.tsx:676-678`). Trade-off is handled: the lookup reads the shared `providerModels` query cache (normally already warm, since the settings page populates it), and the `catch` falls back to the stored model "exactly as today" (`:662-664`), so a slow/failed fetch degrades to current behavior rather than blocking. Net: a possible brief delay only on a cold-cache profile-edit, in exchange for never corrupting edits.

**ℹ️ No action — `resolveCanonicalModel` fetches the full provider list to find one model.** As the review notes, this only fires on profile edit and reads the cached query, so it isn't a hot path. Left as-is.

The docstring nit from the stacked base (#14773) was fixed there and merged in.

PR #14775, opened 2026-06-12
fix: clean up LiteLLM user records when deleting org

### LINKED ISSUE #14686: [Bug] delete_org_cascade leaks the LiteLLM user record when a personal org is deleted

1795771535y-cell:
## Problem

When a personal organization is deleted via \DELETE /api/organizations/{org_id}\, the OpenHands \User\ row, \Org\ row, LiteLLM team, and LiteLLM team-scoped keys are all cleaned up, but the corresponding LiteLLM user record is left behind.

When the same Keycloak identity signs back in, OpenHands re-onboards them with the same UUIDs and \LiteLlmManager.create_entries\ runs:
- \POST /team/new\ → \200\
- \POST /user/new\ → **\409 User with id <uuid> already exists\** ← the orphan

## Root Cause

In \enterprise/storage/org_store.py\, the \delete_org_cascade\ method only deletes the LiteLLM team (step 6) but does not delete the corresponding LiteLLM user records before team deletion.

## Fix

Add cleanup of LiteLLM user records before deleting the LiteLLM team in \delete_org_cascade\:

1. Before step 6 (delete LiteLLM team), fetch all users belonging to the org
2. For each user, call \LiteLLmManager.delete_user(user_uuid)\ to clean up their LiteLLM user record
3. Handle errors gracefully (best-effort cleanup, don't fail the transaction if user cleanup fails)
4. Then proceed with team deletion as before

## Changes

- \enterprise/storage/org_store.py\: Added step 5b to delete LiteLLM user records before team deletion

Closes #14686


ak684:
Thanks for digging into this — the orphaned-LiteLLM-user symptom is real, but we validated the fix needs different scoping, and the narrower #14700 (just merged) covers the case that can actually occur:

1. **LiteLLM user records are shared across teams.** We confirmed empirically on a live install that a single LiteLLM user record belongs to multiple teams at once (global team + personal org team + each joined team org). Deleting the user record for *every member of any deleted org* — as this PR does — would break those members' LiteLLM state in their personal orgs and other team orgs (budgets, keys, spend tracking), since they keep existing as users.

2. **The 409 orphan can only occur when the OpenHands `User` row itself is deleted.** `POST /user/new` only re-runs at re-onboarding, and re-onboarding only happens after the user row is gone — which `delete_org_cascade` does solely on the requester-orphan (personal-org reset) path. That's exactly the scope #14700 fixed. For a team-org deletion, members' user rows survive, `create_entries` doesn't re-run, and no 409 is possible.

3. Heads-up that the diff also introduced encoding artifacts (`—` → `闂?`) in unrelated comment lines across `org_store.py`, which would have needed cleanup regardless.

Closing as superseded by #14700 — appreciate the report, it prompted us to validate the whole cleanup story (including confirming that LiteLLM's `/team/member_delete` does cascade team-scoped key deletion, so member removal is already clean).

PR #14780, opened 2026-06-12
fix: load global user & org skills without a selected repository

hieptl:
HUMAN:

Global user/org skills (e.g. {login}/.openhands and {login}/.agents) were only loaded when a repository was selected: build_org_config derived the owner by parsing the selected repository and returned None otherwise, and global skills are loaded exclusively through that path.

Replace build_org_config with build_org_configs, which enumerates the authenticated user's account login plus every org/group they belong to across all authenticated providers, resolves both .openhands and .agents repos (GitHub-style) concurrently, dedupes by repo path, and returns the full list. The selected repository's owner is still included when present.

Send the list to the agent-server as org_configs and keep the legacy org_config (first entry) in the payload so older agent-server images keep working.

* skill_loader.py: add build_org_configs, _candidate_repo_paths and _enumerate_owners_for_provider; load_skills_from_agent_server now takes org_configs and sends both org_configs and legacy org_config
* app_conversation_service_base.py: wire build_org_configs into load_and_merge_all_skills

Tests: TestBuildOrgConfigs covers no-repo account+org loading, both .openhands/.agents, dedupe when owner == login, GitLab naming (no .agents), and filtering of unresolvable repos; TestLoadSkillsFromAgentServer asserts the dual-field payload; TestLoadAndMergeAllSkills updated for the list API.

AGENT:

---

## Why

Skills in a user's/org's global repos (e.g. `hieptl/.openhands`, `hieptl/.agents`) were only loaded when a repository was selected, because `build_org_config` parses the selected repository to find the owner and returns `None` without one. Conversations started without a repository therefore had no access to globally-scoped skills.

## Summary

- Replace `build_org_config` with **`build_org_configs`**: enumerate the authenticated user's account login + all their orgs/groups across every authenticated provider, resolve **both** `.openhands` and `.agents` repos (GitHub-style) concurrently, dedupe by path, and return the list. The selected repo's owner is still included when present.
- `load_skills_from_agent_server` now sends an `org_configs` list, plus a legacy `org_config` (first entry) for backward compatibility with older agent-server images.
- Wire `build_org_configs` into `load_and_merge_all_skills` (`app_conversation_service_base.py`).

**Files changed:** `openhands/app_server/app_conversation/skill_loader.py`, `openhands/app_server/app_conversation/app_conversation_service_base.py`, `tests/unit/app_server/test_skill_loader.py`, `tests/unit/app_server/test_app_conversation_service_base.py`.

## How to Test

Automated (run from the repo root):

```bash
./.venv/bin/python -m pytest \
  tests/unit/app_server/test_skill_loader.py \
  tests/unit/app_server/test_app_conversation_service_base.py -q
# → 74 passed

./.venv/bin/ruff check  openhands/app_server/app_conversation/skill_loader.py \
                        openhands/app_server/app_conversation/app_conversation_service_base.py \
                        tests/unit/app_server/test_skill_loader.py \
                        tests/unit/app_server/test_app_conversation_service_base.py
./.venv/bin/ruff format --check <same files>
# → All checks passed / already formatted
```

End-to-end (manual — requires a live sandbox, a valid GitHub token, and the SDK companion change deployed in the sandbox image):
1. As a user with `<login>/.openhands/skills/<skill>` and `<login>/.agents/skills/<skill>`, start a new conversation **without** selecting a repository.
2. Confirm the skill is available to the agent.
3. Check the agent-server log for `Loaded N organization skills` and a response with `sources={..., 'org': N, ...}` covering both repos.
4. Repeat **with** an org repo selected and confirm `{owner}/.openhands` + `{owner}/.agents` also load and that `owner == login` is not cloned twice.

## Video/Screenshots

https://github.com/user-attachments/assets/87a262ae-14cf-4d1e-95a2-37165fea99fa

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-80f6843   docker.openhands.dev/openhands/openhands:80f6843
```


tofarr (approved):
🍰

ak684 (changes requested):
Thanks Hiep! The direction is right and the test coverage is solid. Two blocking issues (both cheap to fix) plus a few smaller ones, with suggested code. Everything below is verifiable with the existing mock-based unit tests, no live provider needed.

### 1. (blocking) The legacy `org_config` fallback regresses selected-repo org skills on current agent-server images

`org_config` is set to `org_configs[0]`, but the list is built global-owners-first, selected-repo-owner-last. Agent-server images without OpenHands/software-agent-sdk#3680 (i.e. everything ≤ the currently pinned `1.28.0`) only read the single `org_config`, so a conversation with an org repo selected now sends the *user's personal* `.openhands` as the one config, and the org skills that load today silently stop loading. Since this app-server change will ship before an agent-server pin bump, that's the default production behavior for the first releases carrying this.

Suggested fix: build the selected-repo candidates **first**, so `org_configs[0]` keeps today's semantics (just move the existing block, no new code):

```python
candidates: list[tuple[str, str, str]] = []

# 1. Selected repository owner first: this entry doubles as the legacy
#    `org_config` for older agent-servers, so it must keep pre-list semantics.
if selected_repository and len(selected_repository.split('/')) >= 2:
    ...  # existing block

# 2. Global owners: the user's login plus their orgs/groups, per provider.
try:
    provider_handler = await user_context.get_provider_handler()
    ...
```

One consequence to be aware of: on *new* agent-servers, later entries win on skill-name collision, so this reorder means user/global repos override a same-named skill from the selected org. If you'd rather keep "selected org wins", the alternative is to keep the current order and have `build_org_configs` also return the selected-repo entry separately, sending it explicitly as the legacy `org_config`. Either is fine, but `org_configs[0]` must not change meaning based on what global repos happen to resolve.

### 2. (blocking) Bitbucket DC enumeration fans out to the whole server

`get_bitbucket_dc_projects` → `get_installations` hits `/rest/api/1.0/projects` (`openhands/app_server/integrations/bitbucket_data_center/service/repos.py:106`), which returns every project the user can **browse**. On typical BBDC servers with default permissions that's every project on the instance (up to the 100 pagination limit), not the user's memberships. Each project key becomes 2 candidates (`KEY/.openhands`, `KEY/.agents`), and each candidate costs one `verify_repo_provider` HTTP call, all fired in a single unbounded `asyncio.gather`. That's up to ~200 near-simultaneous requests against a customer's Bitbucket server on **every conversation start**, which can trip BBDC's per-user rate limiting and intermittently break other OpenHands calls (PR comments, reactions) on the same token. It's also a trust-scoping change: any browsable project's `.openhands` repo would inject skills into every user's conversations.

BBDC is the odd one out here: all the other enumerations are membership-scoped (GitHub `/user/installations`, Bitbucket Cloud `/user/workspaces`, GitLab `/groups?min_access_level=10`, Azure DevOps returns its single configured org).

Suggested fix: drop the BBDC branch from `_enumerate_owners_for_provider`. BBDC org skills still load via the selected-repo path, same as today, so nothing regresses:

```python
        elif provider == ProviderType.AZURE_DEVOPS:
            owners.extend(await provider_handler.get_azure_devops_organizations())
        # Bitbucket DC intentionally excluded: /rest/api/1.0/projects returns
        # every browsable project on the server, not memberships. BBDC org
        # skills still load via the selected-repo path.
```

This is testable without a BBDC instance (assert `get_bitbucket_dc_projects` is never called). Related note: BBDC personal repos live under the `~login` user namespace, so `f'{login}/.openhands'` would never resolve for BBDC anyway; fine to leave personal BBDC skills for a follow-up.

### 3. Bound the URL-verification fan-out

Even with membership-scoped providers, a user in many orgs produces an unbounded concurrent burst of verify calls. Suggest a semaphore plus a cap:

```python
_MAX_ORG_CANDIDATES = 30
_URL_RESOLVE_CONCURRENCY = 8

    if len(unique) > _MAX_ORG_CANDIDATES:
        _logger.warning(
            f'Truncating org skill candidates from {len(unique)} to {_MAX_ORG_CANDIDATES}'
        )
        unique = unique[:_MAX_ORG_CANDIDATES]

    sem = asyncio.Semaphore(_URL_RESOLVE_CONCURRENCY)

    async def _resolve(path: str) -> str | None:
        async with sem:
            return await _get_org_repository_url(path, user_context)

    urls = await asyncio.gather(*[_resolve(path) for path, _, _ in unique])
```

(With the reorder from point 1, truncation can never drop the selected-repo entry.)

### 4. (minor) `repository=selected_repository or path` mislabels the global configs

Every global entry claims the selected repository even though it points at an unrelated repo. The agent-server only reads `org_repo_url`/`org_name`, so this is harmless today, but it's confusing in logs/payloads; just use `path`.

### 5. (minor) `# type: ignore[attr-defined]` on `get_provider_handler`

Consider adding `get_provider_handler` to the `UserContext` ABC (default `raise NotImplementedError`) instead of the type-ignore. Contexts that don't support it still degrade gracefully through the existing try/except, and it makes the contract explicit.

### 6. (nit) Dedupe is by repo path only

A same-named owner on two providers collides and the first provider's config silently wins. Probably fine; maybe worth a one-line comment at the dedupe.

enyst:
I can't wait to see this working, thank you!

ak684 (approved):

PR #14782, opened 2026-06-12
fix(frontend): start new LLM profiles on basic settings with blank fields

ak684:
HUMAN: Standalone UX fix, successfully verified e2e with unit tests and mock-mode dev server as well as a human click text.

## Summary

Creating a new LLM profile currently lands users on the advanced view with the custom model and base URL pre-populated from their active/org settings — confusing on managed installs (fields look "already set") and previously reported by a customer as unexpected. This change makes the **create** flow start on the basic view with blank fields, while leaving the **edit** flow exactly as it is today.

## Changes

- New `profileFormMode` (`create`/`edit`) derived from how the form is opened (Add Profile vs Edit menu).
- Create mode: lands on the basic view; provider/model/API key/custom model/base URL all start blank (placeholders); Save stays disabled until a model is chosen (prevents an untouched blank save from resetting `llm.base_url` via the basic-view payload).
- Edit mode: no behavior change — opens advanced when the profile has custom fields, shows stored values. The existing form tests now enter via the Edit path, pinning that behavior.
- Standalone off main; no dependency on any other open PR.

## Testing

- 7 new create-flow tests + reworked helper so ~57 pre-existing form tests exercise the unchanged edit path: `llm-settings` suite 64/64, settings components 113/113, `npm run lint` clean.
- Manually verified via `npm run dev:mock` with Playwright before/after captures: old flow shows pre-selected provider/model and Save enabled; new flow shows blank basic form, Save disabled, advanced fields empty; edit unchanged.

after-3-create-form-initial

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-d371249   docker.openhands.dev/openhands/openhands:d371249
```

### DIFF
--- a/frontend/__tests__/routes/llm-settings.test.tsx
+++ b/frontend/__tests__/routes/llm-settings.test.tsx
@@ -252,10 +252,12 @@ async function renderLlmSettingsScreen({
   organizations?: Organization[];
   scope?: "personal" | "org";
   // Profile-enabled scopes land on the Available Models list by default; set
-  // ``view`` to ``"form"`` (the default) to auto-click into the SDK form
-  // so existing form-oriented assertions keep working unchanged, or to
-  // ``"profiles"`` to test the list view itself.
-  view?: "form" | "profiles";
+  // ``view`` to ``"form"`` (the default) to auto-click into the SDK form via
+  // Edit on a seeded active profile — the form then hydrates from the mocked
+  // settings, so existing form-oriented assertions keep working unchanged.
+  // ``"create"`` clicks Add Profile instead (blank create form), and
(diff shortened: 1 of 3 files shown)


ak684:
Addressed review: trimmed the flagged comments. Kept `setSelectedProvider(null)` in `openForm` — it is not redundant with the value overrides: `selectedProvider` is parallel state already synced from the active settings before the form opens, and the create-mode effect guard only blocks future syncs. Verified empirically: removing the line fails the "create does not inherit the provider" test (63/64).

PR #14783, opened 2026-06-12
fix(frontend): treat managed openhands base URLs as provider defaults in view inference

ak684:
HUMAN: Follow-up to #14782 (now merged). Verified by unit tests including regression guards that fail without the fix.

## Summary

On managed installs, every profile using an `openhands/<model>` carries a server-managed `base_url` (the backend auto-fills `LITE_LLM_API_URL` on save — see `resolve_llm_base_url`). The edit-view inference treated any unrecognized base URL as user customization, so editing **any** managed-model profile landed on the advanced/all view. This teaches `isProviderDefaultBaseUrl` that for `openhands/*` models the base URL is a provider default by construction.

## Changes

- `isProviderDefaultBaseUrl` returns true for models with the `openhands/` prefix — the server owns and auto-fills the value, so its presence is not evidence of user customization.
- Accepted edge case (documented in code): a user who deliberately paired an `openhands/*` model with a custom base URL in advanced mode now lands on basic when editing; their values are preserved and the advanced view remains one click away.
- Non-managed models are unaffected: custom base URLs on `openai/*`, `litellm_proxy/*`, etc. still open advanced (existing tests pass unchanged).

## Testing

- New: OSS-mode and SaaS org-scope tests assert managed model + in-cluster litellm URL lands on basic (both fail without the source fix).
- Flipped: the prior "openhands model + custom base URL opens advanced" test asserted the accepted edge case and now documents the trade-off.
- Suite 66/66; `npm run lint` clean.

## Design note: why `openhands/*` but not `litellm_proxy/*`

The two prefixes look similar but signal different intent. `openhands/<model>` is the app's managed provider: the user picks a model from the dropdown and the server fills `base_url` behind the scenes — its presence says nothing about user intent. `litellm_proxy/<model>` is an explicit transport prefix the user typed themselves (advanced/BYO configuration), so a base URL alongside it — even one pointing at the managed proxy — reflects deliberate setup and still opens the advanced view. The existing test pinning that behavior is unchanged.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-ada307c   docker.openhands.dev/openhands/openhands:ada307c
```

### DIFF
--- a/frontend/__tests__/routes/llm-settings.test.tsx
+++ b/frontend/__tests__/routes/llm-settings.test.tsx
@@ -396,7 +396,60 @@ describe("LlmSettingsScreen", () => {
     ).not.toBeInTheDocument();
   });
 
-  it("opens advanced view when an OpenHands model has a custom base URL", async () => {
+  it("opens basic view when an OpenHands managed model has a server-filled base URL", async () => {
+    // Managed openhands/* profiles always come back with a base_url — the
+    // backend fills it with the managed LiteLLM endpoint on save. That must
+    // not be mistaken for a user customization that opens advanced/all.
+    vi.spyOn(SettingsService, "getSettings").mockResolvedValue(
+      buildSettingsWithAdvancedToggle({
+        llm_model: "openhands/claude-opus-4-5-20251101",
+        llm_base_url: "http://openhands-litellm:4000",
+        agent_settings: {
+          llm: {
+            model: "openhands/claude-opus-4-5-20251101",
+            base_url: "http://openhands-litellm:4000",
+          },
+        },
+      }),
+    );
+
+    await renderLlmSettingsScreen({ appMode: "oss" });
+
+    await screen.findByTestId("llm-settings-form-basic");
+    expect(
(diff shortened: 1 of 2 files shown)


ak684:
Addressed the review:

- **Comment verbosity — done.** Trimmed the 8-line block to 4, using the suggested phrasing for the core insight and keeping the one-line accepted-edge-case note (called out as appropriate under Style Notes). Pushed.
- **Asymmetric handling of `openhands/*` vs `litellm_proxy/*` — documented rather than changed.** Added a design-note section to the PR description: `openhands/<model>` is the managed provider whose base_url the server auto-fills (no user intent encoded), while `litellm_proxy/<model>` is an explicit transport prefix the user typed, so a base URL alongside it reflects deliberate BYO setup and still opens advanced. The asymmetry is intentional; the existing test pinning the `litellm_proxy` behavior is unchanged.

Suite 66/66 after the change.

PR #14785, opened 2026-06-12
fix(app-server): stop dropping items at git pagination boundaries

Dpk376:
HUMAN:
Fixes a pagination bug in the git repo/branch listing endpoints. The router was asking providers for limit + 1 items per page, which shifted the page windows and made one repo/branch disappear at every page boundary. Search queries also pointed to next pages that didn't actually exist. I noticed this while browsing repos in the UI and traced it down to the router.

AGENT:

---

## Why

The V1 app server git endpoints (`/api/v1/git/repositories/search` and `/api/v1/git/branches/search`) probed for a next page by requesting `per_page = limit + 1` from the provider. The providers use page-number pagination, so the oversized window shifted every subsequent page and **silently dropped one repository/branch at every page boundary** (with `limit=2`: client sees items 1,2 then 4,5 then 7,8 — items 3 and 6 never appear).

Search queries also advertised a `next_page_id` even though the underlying `search_repositories` / `search_branches` methods accept no page parameter — following it returned the first page again forever (repos) or a guaranteed 400 (branches).

## Summary

- Request `per_page == limit` when listing repositories/branches so consecutive pages are contiguous windows
- Derive branch `next_page_id` from the `has_next_page` flag the provider already returns in `PaginatedBranchesResponse`; for repositories, infer it from a full page
- Return search results as a single page (`next_page_id=None`) and reject `page_id` beyond the first with 400 until search pagination is supported (TODOs reference OpenHands/enterprise#73 / OpenHands/enterprise#22)

## Issue Number

Fixes OpenHands/enterprise#73

## How to Test

```bash
pytest tests/unit/app_server/test_git_router.py
```

The rewritten pagination tests (`test_pagination_works_across_pages`, `test_listing_pages_are_contiguous`) simulate a provider with faithful page-number pagination and assert that every item is seen exactly once across pages — they fail against the previous implementation.

Manual check: connect a GitHub account with more repositories than the page size, call `/api/v1/git/repositories/search` with a small `limit`, and walk pages via `next_page_id`. Previously one repository disappeared at every page boundary.

## Video/Screenshots

N/A — backend-only change covered by unit tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

When the repository total is an exact multiple of `limit`, the listing now reports one trailing empty page (the heuristic can't distinguish "exactly full" from "more remaining" without provider support). This is strictly better than silently dropping items; the proper fix is part of the provider API refactor already tracked in the in-code TODOs. Branch listing has no such trade-off since it uses the provider's `has_next_page` flag.

### DIFF
--- a/openhands/app_server/git/git_router.py
+++ b/openhands/app_server/git/git_router.py
@@ -154,6 +154,16 @@ async def search_repositories(
 
     # If query is provided, use search; otherwise get user's repositories
     if query:
+        if page != 1:
+            # TODO(#14784): Support pagination for repository search after
+            # refactoring. search_repositories does not accept a page number,
+            # so any page beyond the first would silently repeat the first
+            # page. See #13883 for the equivalent branch search work.
+            raise HTTPException(
+                status_code=status.HTTP_400_BAD_REQUEST,
+                detail='Pagination not yet supported for repository search queries. Use empty query to list repositories with pagination.',
+            )
+
         # Parse sort_order into sort and order components (if provided)
         if sort_order:
             search_sort, order = sort_order.value.rsplit('-', 1)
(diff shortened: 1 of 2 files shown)


Dpk376:
> 👋 Thanks for opening this PR!
> 
> Your PR is marked for review. Can you confirm with a 👍 reaction **on this comment** or a reply that this PR is ready to review?
> 
> > If no confirmation is received within 7 days, this PR will be converted back to a draft.

👍

Dpk376:
@tofarr friendly ping 🙂 — this fixes the off-by-one item drop at page boundaries in the V1 git repo/branch search routes (the providers use page-number pagination, so the `limit + 1` probe was shifting every window and dropping one item per boundary; the search queries also advertised a `next_page_id` the underlying methods don't accept). CI is green and the PR is confirmed ready. Would you have a moment to take a look? Happy to adjust anything. Thanks!

neubig:

Thank you for understanding.

PR #3692, opened 2024-09-01
fix eval README link

mamoodi:
**Short description of the problem this fixes or functionality that this introduces. This may be used for the CHANGELOG**

fix eval README link

---
**Give a summary of what the PR does, explaining any non-trivial design decisions**

---
**Link of any specific issues this addresses**

PR #14797, opened 2026-06-13
fix(acp): surface acp_server on cloud conversations so the model picker renders

### LINKED ISSUE #14796: ACP model picker missing on cloud backend: AppConversationInfo lacks acp_server field
## Summary

The ACP model picker in agent-canvas does not show the list of available models when the canvas is connected to a cloud/SaaS backend. The user sees only the current model label (e.g. `gpt-5.5/medium`) and a settings link — no switchable model list.

## Root Cause

Two related issues in the cloud conversation layer:

### 1. `AppConversationInfo` has no `acp_server` field

The canvas TypeScript type `AppConversation` already declares `acp_server?: string | null` — a top-level field identifying the ACP provider key (`"codex"`, `"claude"`, `"gemini-cli"`). The canvas uses this to look up the provider's `available_models` list and render the model picker.

The Python `AppConversationInfo` model (`openhands/app_server/app_conversation/app_conversation_models.py`) has no such field, so it is absent from the JSON returned by `GET /api/v1/app-conversations`. The canvas always sees `acp_server: undefined` for conversations fetched from the cloud backend → no provider found → empty model list → picker hidden.

### 2. Tag key mismatch in `live_status_app_conversation_service.py`

`live_status_app_conversation_service.py` stores the provider key as `tags['acp_server']` (underscore). The SDK tag validator enforces `^[a-z0-9]+$` — underscores are invalid. The canvas and SDK both use `'acpserver'` (no underscore) as the canonical key, so the stored tag is both invalid and unreadable by the canvas.

## Fix

**`openhands/app_server/app_conversation/app_conversation_models.py`**
```python
class AppConversationInfo(BaseModel):
    ...
    acp_server: str | None = Field(
        default=None,
        description="ACP provider key ('codex', 'claude', 'gemini-cli') for ACP conversations. Null for OpenHands conversations.",
    )
```

**`openhands/app_server/app_conversation/live_status_app_conversation_service.py`**
```python
# change tags key from 'acp_server' → 'acpserver' (no underscore, valid per SDK validator)
# and set top-level acp_server field
if request_agent.agent_kind == 'acp':
    acp_server_key = acp_user.agent_settings.acp_server
    tags['acpserver'] = acp_server_key   # was tags['acp_server']

app_conversation_info = AppConversationInfo(
    ...
    acp_server=acp_server_key if agent_kind == 'acp' else None,
    tags=tags,
    ...
)
```

## Workaround

agent-canvas has a client-side fallback ([`use-chat-input-model-state.ts`](https://github.com/OpenHands/agent-canvas/blob/main/src/hooks/use-chat-input-model-state.ts)) that falls back to `settings.agent_settings.acp_server` when `conversation.acp_server` is absent. This allows the picker to render, but mid-conversation model switches are attributed to the home/settings context rather than the live session.

simonrosenberg:
HUMAN:

surface acp_server on cloud conversations so the model picker renders

AGENT:

---

## Why

The ACP model picker is missing when the canvas is connected to a cloud/SaaS backend — the user sees only the current model label (e.g. `gpt-5.5/medium`) and a settings link, with no switchable model list.

The canvas reads a **top-level** `acp_server` field off the `GET /api/v1/app-conversations` JSON to look up the provider's `available_models` and render the picker. The Python `AppConversationInfo` model never declared that field, so cloud-backend conversations always saw `acp_server: undefined` → no provider found → empty model list → picker hidden.

The provider key *was* already being persisted — in the `tags` column — but under the key `acp_server` (underscore), which violates the SDK tag-key validator (`^[a-z0-9]+$`). The canonical key used by the canvas and SDK is `acpserver`.

Fixes #14796.

## Summary

- Expose `acp_server` on `AppConversationInfo` as a **`@computed_field` projection of the `acpserver` tag** (the contract the canvas `AppConversation` type already documents: *"Populated from `info.tags.acpserver`"*). Gated on `agent_kind == 'acp'` so a stray tag never reports a provider for an OpenHands conversation.
- Stamp the provider key under the canonical `acpserver` tag key (was the invalid `acp_server`), via a shared `ACP_SERVER_TAG_KEY` constant.

**Why a computed field rather than a stored field + dual-written tag:** the broken list endpoint rebuilds conversations from the DB via `_to_info`, which only repopulates declared columns. A plain `acp_server` field with no SQL column would be silently `None` on exactly that endpoint — but `_to_info` already round-trips `tags`. Projecting from the tag keeps a single source of truth, survives the DB round-trip and webhook tag-merges for free, needs **no migration**, and unifies the representation with the direct agent-server path.

## Issue Number

#14796

## How to Test

Automated (run against the pinned `openhands-sdk==1.28.0`):

```bash
uv sync --frozen
uv run pytest tests/unit/app_server/test_models.py tests/unit/app_server/test_webhook_router_acp.py -q
```

- `TestAcpServerProjection` covers: projection from the tag, JSON serialization, DB-round-trip-via-tags-only, the `AppConversation(**info.model_dump())` build path, and `None` for non-ACP / absent-tag.
- `test_acp_server_tag_preserved_on_webhook_update` now asserts both the canonical `acpserver` tag and the projected `acp_server` field survive a webhook update.

Manual: with an ACP provider selected, start a conversation on the cloud backend and confirm `GET /api/v1/app-conversations` returns `acp_server` (e.g. `"codex"`) and the canvas renders the switchable model picker.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No DB migration. The companion client-side fallback in agent-canvas (`use-chat-input-model-state.ts`, falls back to `settings.agent_settings.acp_server`) can remain as defence-in-depth; with this change the live conversation's own `acp_server` is authoritative so mid-conversation model switches are attributed to the session rather than the home/settings context.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-a6be359   docker.openhands.dev/openhands/openhands:a6be359
```

### DIFF
--- a/openhands/app_server/app_conversation/app_conversation_models.py
+++ b/openhands/app_server/app_conversation/app_conversation_models.py
@@ -4,7 +4,7 @@ from enum import Enum
 from typing import Any, Literal
 from uuid import UUID, uuid4
 
-from pydantic import BaseModel, Field, SecretStr, field_serializer
+from pydantic import BaseModel, Field, SecretStr, computed_field, field_serializer
 
 from openhands.agent_server.models import (
     ImageContent,
@@ -26,6 +26,12 @@ from openhands.sdk.llm import MetricsSnapshot
 from openhands.sdk.plugin import PluginSource
 
 __all__ = ['SandboxGroupingStrategy']
+
+# Canonical conversation-tag key under which the active ACP provider key
+# ('claude-code', 'codex', 'gemini-cli') is stored. Lowercase-alphanumeric to satisfy
+# the SDK tag-key validator (^[a-z0-9]+$); mirrors agent-canvas' ACP_SERVER_TAG_KEY.
+# The typed ``AppConversationInfo.acp_server`` field is a projection of this tag.
+ACP_SERVER_TAG_KEY = 'acpserver'
 
 
 class ConversationTrigger(Enum):
@@ -132,6 +138,22 @@ class AppConversationInfo(BaseModel):
 
     created_at: datetime = Field(default_factory=utc_now)
     updated_at: datetime = Field(default_factory=utc_now)
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def acp_server(self) -> str | None:
+        """Active ACP provider key ('claude-code', 'codex', 'gemini-cli'), else None.
+
+        A typed projection of the ``acpserver`` tag (the same key agent-canvas
+        reads) so the conversation UI can resolve a provider brand label without
+        a dedicated column. Riding the tag keeps a single source of truth that
+        round-trips through the DB ``tags`` column for free. Gated on
+        ``agent_kind`` so a stray tag never reports a provider for an OpenHands
+        conversation.
+        """
+        if self.agent_kind != 'acp':
+            return None
+        return self.tags.get(ACP_SERVER_TAG_KEY)
 
 
 class AppConversationSortOrder(Enum):
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -25,6 +25,7 @@ from openhands.app_server.app_conversation.app_conversation_info_service import 
     AppConversationInfoService,
 )
 from openhands.app_server.app_conversation.app_conversation_models import (
+    ACP_SERVER_TAG_KEY,
     AgentType,
     AppConversation,
     AppConversationInfo,
@@ -381,9 +382,10 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
                 # Persist the active ACP provider key so the conversation UI
                 # can resolve a brand label ("Claude Code", "Codex", …) via
                 # the SDK registry without keeping a per-conversation column.
+                # Surfaced to the UI as the projected ``acp_server`` field.
                 acp_user = await self.user_context.get_user_info()
                 if isinstance(acp_user.agent_settings, ACPAgentSettings):
-                    tags['acp_server'] = acp_user.agent_settings.acp_server
+                    tags[ACP_SERVER_TAG_KEY] = acp_user.agent_settings.acp_server
             else:
                 llm_model = request_agent.llm.model
                 agent_kind = 'openhands'
--- a/tests/unit/app_server/test_models.py
+++ b/tests/unit/app_server/test_models.py
@@ -4,6 +4,8 @@ from uuid import UUID, uuid4
 import pytest
 
 from openhands.app_server.app_conversation.app_conversation_models import (
+    ACP_SERVER_TAG_KEY,
+    AppConversation,
     AppConversationInfo,
     AppConversationStartRequest,
     AppConversationUpdateRequest,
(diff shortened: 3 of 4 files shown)


simonrosenberg:
@openhands question: should the "acpserver" tag be renamed "acp_server"? This seems a bit confusing
Would that help fixing the problem? does it imply a migration plan?

simonrosenberg:
Good question — short answer: the tag key **can't** be `acp_server`, and the rename wouldn't help the bug or be worth a migration. Details:

**1. The SDK validator forbids the underscore.** In `openhands/sdk/conversation/types.py`:

```python
TAG_KEY_PATTERN = re.compile(r"^[a-z0-9]+$")
# ...
if not TAG_KEY_PATTERN.match(key):
    raise ValueError(
        f"Tag key '{key}' is invalid: keys must be lowercase alphanumeric only"
    )
```

So `tags = {"acp_server": "codex"}` raises `ValueError` at every write path (direct agent-server creation, webhook merge, list endpoint round-trip). That's actually exactly the reason the original code was broken: it had been storing under `acp_server`, which silently failed validation in some paths.

**2. The rename wouldn't fix the picker.** The canvas was hiding the picker because the top-level JSON had no `acp_server` field at all, not because the storage key was wrong. The fix in this PR is the typed `acp_server` projection on `AppConversationInfo`. Once that exists, the storage key underneath could be literally anything compliant — `acpserver`, `provider`, `xyz` — and the wire/UI contract is unchanged.

**3. The naming gap is contained and one-directional.**
   - Wire / API / canvas: `acp_server` (snake_case, human-friendly) — this is what every caller sees.
   - Storage in `tags`: `acpserver` (validator-compliant) — only the projection logic touches it, and the `ACP_SERVER_TAG_KEY` constant is the single source of truth.

   Both halves now match their respective rule sets, and the conversion is in one tiny place.

**4. Migration cost without benefit.** Renaming would require (a) relaxing the SDK `TAG_KEY_PATTERN` (cross-repo change to allow `_`, which then opens the door for every consumer to invent their own delimiter conventions and is genuinely the wrong direction for a key namespace), and (b) backfilling existing rows + handling webhook payloads from older agent servers. All to change a key that no external caller ever reads.

So: leave it as-is. The PR's `ACP_SERVER_TAG_KEY = 'acpserver'` constant + computed-field projection is the cheapest correct shape.

_This comment was posted by an AI agent (OpenHands) on behalf of @simonrosenberg._

simonrosenberg:
Follow-up plumbing tracked in OpenHands/software-agent-sdk#3691 (SDK PR OpenHands/software-agent-sdk#3692): once the SDK carries `ACPAgent.acp_server`, OpenHands can source the key authoritatively from the conversation webhook (`conversation_info.agent.acp_server`) and agent-canvas can read it off the agent — retiring the `acpserver` tag. This PR remains the immediate fix against the current SDK pin.

malhotra5 (approved):
LGTM

Also is `acp_server` the best name for this? maybe something like `acp_agent` could be more intuitive?

simonrosenberg:
I'd push back on `acp_agent` — `acp_server` is the more accurate name here.

The field identifies *which ACP server/CLI is running* (one of `claude-code`, `codex`, or `gemini-cli`). These are distinct server processes, not agents. 

`acp_agent` could be confusing because:
- "agent" is already heavily overloaded in this codebase (Agent, AgentProfile, etc.)
- It blurs the boundary between *which server is running* vs. *what agent software is using it*  
- The canonical tag it projects from is `acpserver`, which reinforces the "server" framing

The name aligns with SDK #3692 and the broader ACP server selection logic. Keep it as is.

PR #14801, opened 2026-06-13
docs: follow-up — unify model verified-determination across LLM services

ak684:
## Problem
`_to_llm_models()` derives a model's `verified` flag from the **static SDK catalogue**, while `_to_providers()` derives it from the discovery response (`verified_providers`). So `ModelsResponse.verified_models` is populated by every service and **consumed by no one**. On the managed LiteLLM proxy this split curated models into "Verified"/"Other" by SDK-name coincidence — meaningless for an admin catalogue.

## Proposed
Have every service populate `verified_models` fully, then consume it uniformly in `_to_llm_models()` (mirroring `_to_providers()`); drop the override + SDK-static special case.

**Blocker / why deferred:** SaaS populates `verified_models` with only `openhands/*`, so a blanket switch would de-verify Anthropic/OpenAI/Gemini in cloud. Needs the SaaS service to populate its full verified set first — SaaS-touching.

Detail: `docs/proposals/unify-model-verified-determination.md`.

HUMAN: Design proposal only (no behavior change in this PR).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-3879c05   docker.openhands.dev/openhands/openhands:3879c05
```

### DIFF
--- a/docs/proposals/unify-model-verified-determination.md
+++ b/docs/proposals/unify-model-verified-determination.md
@@ -1,0 +1,55 @@
+# Proposal: unify model "verified" determination across LLM model services
+
+Status: draft / follow-up. Not for immediate merge — captures a refactor we
+want to do after the surgical fix ships.
+
+## Problem
+
+`config_api/default_llm_model_service.py` determines the `verified` flag two
+different ways for the two object kinds it returns:
+
+- `_to_providers()` derives it from the discovery response:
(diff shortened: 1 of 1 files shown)


neubig:

Thank you for understanding.

PR #14808, opened 2026-06-14
fix: make a member's settings save private (stop the org-wide broadcast)

ak684:
HUMAN: Goal-3 de-brittle, step 1. Stops a member's personal settings save from clobbering the whole org (the "key spread" gotcha). Backend-only, behavior change to the member-save path; the admin org-defaults path is unchanged. Verified by unit tests; R02 multi-member check pending.

## Problem
Today a member's personal settings save (`SaasSettingsStore.store`) writes the saved settings into **`org.agent_settings`** (the org default) **and** broadcasts them — model, base_url, agent, condenser, and (for non-managed keys) the raw API key — into **every** member's row via `update_all_members_settings_async`. So one member saving overwrites other members' explicit model choices, and a member's BYOK key spreads across the org. This was intentional ("org-wide shared snapshot"), but it breaks the moment members have their own preferences/keys.

## Change
A member's personal save now writes **only their own `OrgMember` row**. It no longer mutates `org.agent_settings`/`org.conversation_settings` or broadcasts to other members. Propagation still works because `load()` merges the org default *under* the member's own diff, so members who haven't overridden continue to track the org default — without overwriting members who have. Org-wide defaults remain set through the dedicated **admin org-defaults path** (`OrgStore._update_org_kwargs`), which is untouched.

## Tests
- `test_member_save_stays_private_and_does_not_clobber_org_or_members`: an admin save leaves the org default and the other two members (each with their own model + key) untouched, while the acting member gets their new settings + key.
- `test_store_keeps_openhands_managed_keys_member_specific`: updated to assert other members keep their own model + managed key (nothing copied).
- Full suites: saas_settings_store 22, org_store/org_profiles/org_member 97 — green.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-1b09105   docker.openhands.dev/openhands/openhands:1b09105
```

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -9,14 +9,12 @@ from pydantic import SecretStr
 from server.auth.token_manager import TokenManager
 from server.constants import LITE_LLM_API_URL
 from server.logger import logger
-from server.routes.org_models import OrgMemberSettingsUpdate
 from sqlalchemy import select
 from sqlalchemy.orm import joinedload
 from storage.database import a_session_maker
 from storage.lite_llm_manager import LiteLlmManager, get_openhands_cloud_key_alias
 from storage.org import Org
 from storage.org_member import OrgMember
-from storage.org_member_store import OrgMemberStore
 from storage.org_store import OrgStore
 from storage.user import User
 from storage.user_settings import UserSettings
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #14810, opened 2026-06-15
fix: write trajectory exports as utf-8

wgu9:
HUMAN:
This PR makes trajectory exports explicitly write event and metadata JSON as UTF-8, so conversations with non-ASCII content export consistently on platforms whose default encoding is not UTF-8. The change is limited to the export write boundary and covered by a regression test that simulates a non-UTF-8 platform default.

## Summary

- Write trajectory export `event_*.json` files and `meta.json` with explicit UTF-8 encoding.
- Add a regression test that simulates a non-UTF-8 platform default encoding and verifies exported metadata and event payloads decode as UTF-8.

## Why

`export_conversation()` writes JSON files before zipping the trajectory. Without an explicit encoding, Python uses the platform default. On Windows-style defaults such as `cp1252`, metadata or event payloads containing non-ASCII characters can fail to write or produce inconsistent export bytes.

The exported trajectory is already consumed as UTF-8 by existing tests and clients, so the write boundary should use the same encoding explicitly.

## Edge cases covered

- Non-ASCII conversation metadata.
- Non-ASCII event payloads.
- Platform defaults that cannot encode emoji.
- Existing empty-events export behavior.
- Existing paginated event export behavior.
- Existing conversation-not-found behavior.
- Existing zip layout and filenames.
- Existing JSON schema and API response type.

## Validation

- `uv run --group test pytest tests/unit/app_server/test_live_status_app_conversation_service.py -q`
- `uv run ruff format --check openhands/app_server/app_conversation/live_status_app_conversation_service.py tests/unit/app_server/test_live_status_app_conversation_service.py`
- `uv run ruff check openhands/app_server/app_conversation/live_status_app_conversation_service.py tests/unit/app_server/test_live_status_app_conversation_service.py`
- `git diff --check`

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -2110,7 +2110,7 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
                 event_filename = f'event_{i:06d}_{event.id}.json'
                 event_path = os.path.join(temp_dir, event_filename)
 
-                with open(event_path, 'w') as f:
+                with open(event_path, 'w', encoding='utf-8') as f:
                     # Use model_dump with mode='json' to handle UUID serialization
                     event_data = event.model_dump(mode='json')
                     json.dump(event_data, f, indent=2)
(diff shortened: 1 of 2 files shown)


VascoSch92 (approved):
Thank you :-)

PR #14812, opened 2026-06-15
fix(frontend): avoid dirty language filter input

### LINKED ISSUE #14249: [Bug] language selector dirty-flag incorrectly set while typing (onInputChange vs onSelectionChange)

mturac:
HUMAN:
This keeps the app settings form from treating language dropdown filter text as a saved setting change. Save Changes should stay disabled while someone only types to filter languages, and only become dirty after an actual language option is selected.

## Summary
- mark the language setting dirty only when a language option is selected
- keep autocomplete filter typing from enabling Save Changes without a real setting change
- add a regression test for filtering the language dropdown without selecting a new language

Fixes #14249

## Changelog
Users can filter the language dropdown without the app settings form being marked dirty until they actually select a language.

## Testing
- `npm run test -- __tests__/routes/app-settings.test.tsx`
- `npm run lint:fix`
- `npx prettier --check __tests__/routes/app-settings.test.tsx src/components/features/settings/app-settings/language-input.tsx`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/frontend/__tests__/routes/app-settings.test.tsx
+++ b/frontend/__tests__/routes/app-settings.test.tsx
@@ -173,6 +173,25 @@ describe("Form submission", () => {
     expect(submit).toBeDisabled();
   });
 
+  it("should keep the submit button disabled while filtering languages without selecting one", async () => {
+    const getSettingsSpy = vi.spyOn(SettingsService, "getSettings");
+    getSettingsSpy.mockResolvedValue(MOCK_DEFAULT_USER_SETTINGS);
+
+    const user = userEvent.setup();
+    renderAppSettingsScreen();
+
+    const language = await screen.findByTestId("language-input");
+    const submit = await screen.findByTestId("submit-button");
+
+    expect(language).toHaveValue("English");
+    expect(submit).toBeDisabled();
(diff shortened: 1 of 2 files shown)


mturac:
Ready for review.

VascoSch92:
Hey @mturac 

Thanks for the PR


Could you attach a small video to show that the fix works?


Thanks :-)

mturac:
Thanks for taking a look. I recorded a short video of the fixed flow here:

https://raw.githubusercontent.com/mturac/OpenHands/assets/pr-14812-proof/pr-assets/14812/language-dirty-state.webm

The recording shows the App settings language field starting at `English` with Save disabled, switching to `Norsk` enabling Save, then switching back to `English` disabling Save again.

VascoSch92 (approved):

VascoSch92:
@mturac Thanks :-)

Issue #1164, opened 2024-04-16
LLM write into non exist file and maybe prompts leak


yufansong:
#### Describe the bug

In #1063 and also some others feedback, LLM sometimes output a wrong file directory. And in the #1070 discussion [here](https://github.com/OpenDevin/OpenDevin/pull/1070#issuecomment-2054062906). LLM should work in our docker env, it should see the actual file directory. Open this issue to track the related problem. 

#### Setup and configuration
**Current version**:

```bash
commit e0c7492609bf8ebdbd7270c9e9261bf2944f3bc6 (HEAD -> main, origin/main, origin/HEAD)
Author: Boxuan Li <liboxuan@connect.hku.hk>
Date:   Fri Apr 12 12:01:52 2024 -0700

    Traffic Control: Add new config MAX_CHARS (#1015)

    * Add new config MAX_CHARS

    * Fix mypy linting issues


```

**My operating system**:

**My environment vars and other configuration** (be sure to redact API keys):
```bash
LLM_API_KEY="ollama"
LLM_MODEL="ollama/atlas:002"
LLM_EMBEDDING_MODEL="local"
LLM_BASE_URL="http://localhost:11434"
WORKSPACE_DIR="./workspace/"
SANDBOX_TYPE="exec"
```

**My model and agent** (you can see these settings in the UI):
* Model: A local model
* Agent:

**Commands I ran to install and run OpenDevin**:

```
cd ~/$OpenDevin/ && poetry run python opendevin/main.py -d ./workspace/ -t "create a script that says hi"
```

**Steps to Reproduce**:
1. run opendevin.
2. run into a bunch of file operations path issues.
3.

**Logs, error messages, and screenshots**:

```
==============
STEP 9

PLAN:
write a bash script that prints hi

ACTION:
FileWriteAction(path='./workspace/memory.txt', content="I'm going to create a file called 'memory.txt' to store my thoughts.", action=<ActionType.WRITE: 'write'>)

ERROR:
[Errno 2] No such file or directory: './workspace/./workspace/memory.txt'
Traceback (most recent call last):
  File "/home/atlas/OpenDevin/opendevin/controller/agent_controller.py", line 186, in step
    if inspect.isawaitable(action.run(self)):
                           ^^^^^^^^^^^^^^^^
  File "/home/atlas/OpenDevin/opendevin/action/fileop.py", line 43, in run
    with open(whole_path, "w", encoding="utf-8") as file:
[...]

PR #14822, opened 2026-06-15
fix(mcp): store remote MCP API keys as redacted Authorization headers

simonrosenberg:
HUMAN:

Reviewed and approved — ports the agent-canvas #1164 fix to OpenHands; remote MCP API keys are now redacted at rest.

AGENT:

Verified with the SDK serializer directly: an `ACPAgentSettings`/`OpenHandsAgentSettings` whose `mcp_config` carries the key as `auth` leaks it (`'SECRET' in model_dump_json()` → True), while the `headers.Authorization` form this PR writes is redacted (`<redacted>`). Frontend unit tests: 26/26 pass.

---

## Why

Remote (SSE / SHTTP) MCP server API keys were serialized into `mcp_config` as a bare `auth: <key>` field. The SDK's `mcp_config` secret serializer only redacts/encrypts `env` and `headers` — **not** `auth` — so the key was persisted and round-tripped in **plaintext** (in settings storage and in redacted GET responses). Older SDKs also didn't forward `auth` to ACP subprocesses (only `headers` was translated).

This is the same bug agent-canvas hit and fixed in OpenHands/agent-canvas#1164. It affects both the OpenHands agent and (once MCP is un-gated for ACP, OpenHands/software-agent-sdk#3705 / PR #14613) ACP agents.

## Summary

- `toSdkMcpConfig` now writes remote API keys as `headers: { Authorization: "Bearer <key>" }` instead of `auth: <key>`, so the SDK redacts/encrypts them and the ACP forwarding path picks them up.
- `parseMcpConfig` reads the key from `headers.Authorization` **or** the legacy `auth` field, so existing persisted settings keep working and migrate to the header form on the next save.
- Updated/added unit tests: serialize-as-header, header→`api_key` round-trip, and legacy-`auth`→header migration.

## Issue Number

Relates to OpenHands/software-agent-sdk#3705 (ACP MCP forwarding foundation)

## How to Test

```sh
cd frontend
npx vitest run __tests__/utils/mcp-config.test.ts
```

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-473f035   docker.openhands.dev/openhands/openhands:473f035
```

### DIFF
--- a/frontend/__tests__/utils/mcp-config.test.ts
+++ b/frontend/__tests__/utils/mcp-config.test.ts
@@ -223,13 +223,17 @@ describe("toSdkMcpConfig", () => {
     expect(result?.mcpServers["sse_2"].url).toBe("https://example3.com");
   });
 
-  it("should include api_key as auth field", () => {
+  it("should serialize api_key as an Authorization bearer header", () => {
+    // The SDK only redacts/encrypts ``headers`` (and ``env``), not ``auth`` —
+    // writing the key to ``auth`` would persist it in plaintext.
     const config: MCPConfig = {
       sse_servers: [
         { name: "secure", url: "https://example.com", api_key: "my-secret" },
       ],
       stdio_servers: [],
-      shttp_servers: [],
+      shttp_servers: [
+        { name: "shttp", url: "https://shttp.example", api_key: "shttp-secret" },
+      ],
     };
 
     const result = toSdkMcpConfig(config);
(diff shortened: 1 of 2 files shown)


simonrosenberg:
Folded into #14613 (the green ACP-MCP PR) to keep a single OpenHands PR for this work. #14613 now carries both the MCP nav un-gate and this redaction fix, and is up to date with main.

PR #14168, opened 2026-04-27
feat: enforce conversation limits

HeyItsChloe:
- [x] A human has tested these changes.

---

## Why

- We want to enforce conversation limits on Personal Workspaces:

## Summary

- Max concurrent convos in a Personal Org is set to 3
- Max concurrent convos default is set to 10 
- Default (10) is set for all Commercial / non personal orgs
- Per org limits are set in new `org.max_concurrent_sandboxes` column
- Per User limits are set in new `org_member.max_concurrent_sandboxes_override`  column which overrides the org level limit. 
- Per user limit defaults to `null`. When null, org limit is used. 
- OpenHands employees can modify concurrency limits for an individual Personal Workspaces or Org in SaaS via the new `org.max_concurrent_sandboxes` column
- When the limit is exceeded, a ConcurrencyLimitError is raised with HTTP status code 429 Too Many Requests, and the Error modal renders
-   

## How to Test
- Start new conversations until you reach your max concurrent limit
- The modal should render and you should not be able to start a conversation once your limit is hit until you close an open conversation.  

## Video/Screenshots

Videos shows trying to start a 4th convo, when 3 are running.

https://github.com/user-attachments/assets/0e3e2e43-a880-4c40-aa20-967e80a6a529

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-6efca02   docker.openhands.dev/openhands/openhands:6efca02
```

PR #14834, opened 2026-06-15
fix: use runtime /list API for sandbox concurrency count, not is_paused DB flag

tofarr:
H:

AGENT:

---

## Why

Migration `011` added an `is_paused` column to `v1_remote_sandbox` with `server_default=0` (False), which set every pre-existing DB row to `is_paused=False`. The new `_count_user_running_sandboxes` method then counted all rows where `is_paused=False` as "running", treating every historical sandbox record as active. Users with many past sandboxes (e.g. 80) were immediately blocked with a 429 even when zero sandboxes were actually running.

Root cause: `is_paused` only tracks explicit app-level pause/resume operations. It does not reflect whether the runtime has terminated, expired, or cleaned up a sandbox externally. The pre-existing `pause_old_sandboxes` override used the correct approach (runtime `/list` + DB cross-reference) before PR #14168 replaced it with the `is_paused` query — apparently to address a TOCTOU concern raised in code review, but the fix introduced a worse correctness bug.

## Summary

- Add `_get_user_running_sandboxes`: calls the runtime `/list` endpoint (all running sessions across all users) and cross-references with the current users DB records. Only sandboxes the runtime confirms as running count against the limit — stale or expired DB rows are automatically excluded.
- Rewrite `check_concurrency_limit` to use `_get_user_running_sandboxes` instead of the `is_paused` DB flag.
- Restore the `pause_old_sandboxes` override with the same `/list`-based approach. Also fixes a sort-order bug in the pre-#14168 version: it was pausing the *newest* sandboxes instead of the oldest.
- Simplify `start_sandbox`: remove the `SELECT FOR UPDATE` nested transaction, which was only meaningful when the count came from the DB. `check_concurrency_limit()` is now called directly.
- Remove `_count_user_running_sandboxes` (the `is_paused`-based DB count method).
- Update all tests to mock `_get_user_running_sandboxes` instead of the removed method.

## Issue Number

N/A (reported via support: user sees `User has reached sandbox limit: 80/3` with no running sandboxes)

## How to Test

1. Ensure your account has pre-existing sandbox DB records (historical conversations).
2. Confirm no sandboxes are currently running.
3. Attempt to start a new conversation — it should succeed (no 429).
4. Verify concurrency limit is still enforced when you actually have `N` running sandboxes matching the limit.

Unit tests: `pytest tests/unit/app_server/test_remote_sandbox_service.py`

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The `is_paused` column and migration `011` are left in place — the column is still written in `pause_sandbox`/`resume_sandbox` for potential future use and removing it would require an additional migration. The counting logic simply no longer reads it.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-5215a19   docker.openhands.dev/openhands/openhands:5215a19
```

### DIFF
--- a/enterprise/migrations/versions/122_drop_is_paused_from_v1_remote_sandbox.py
+++ b/enterprise/migrations/versions/122_drop_is_paused_from_v1_remote_sandbox.py
@@ -1,0 +1,38 @@
+"""Drop is_paused from v1_remote_sandbox.
+
+The column was added in 121 to cache runtime pause state in the DB, but it is
+not a reliable indicator of whether a sandbox is actually running: only
+explicit app-level pause/resume calls update it, so externally terminated or
+expired sandboxes remain is_paused=False forever. Concurrency counting now
+uses the runtime /list endpoint as the source of truth instead.
+
+Revision ID: 122
+Revises: 121
+Create Date: 2026-06-05
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = '122'
+down_revision: Union[str, None] = '121'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.drop_column('v1_remote_sandbox', 'is_paused')
+
+
+def downgrade() -> None:
+    op.add_column(
+        'v1_remote_sandbox',
+        sa.Column(
+            'is_paused',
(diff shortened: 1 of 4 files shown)


hieptl (approved):
Thank you! 🙏

PR #14838, opened 2026-06-16
fix(app-server): respect disabled_skills in conversation skills endpoint

StPig:
HUMAN:
Fix disabled skills not being filtered out by the conversation skills API.

AGENT:
Fixes the conversation skills endpoint (`GET /api/v1/app-conversations/{id}/skills`)
ignoring `disabled_skills`: it returned the full merged set while the agent context
filters them, so the panel advertised skills the agent doesn't have. Applies the same
filter via `app_conversation_service.user_context`. Updated existing tests + added one
asserting disabled skills are excluded.

## Why
The panel endpoint returned skills the agent doesn't actually have, because
`disabled_skills` was only applied to the agent context, not this endpoint.

## Summary
- `get_conversation_skills` now drops `user.disabled_skills`, mirroring
  `_load_skills_and_update_agent`.
- Added/updated unit tests for the disabled-skills filter.

## Issue Number
OpenHands/enterprise#74

## How to Test
1. Disable a skill the agent loads (an AgentSkills name, e.g. `azure-devops`) in user settings.
2. Start a new conversation, open the Available Skills panel → the skill is gone.
   (Unit: `pytest tests/unit/app_server/test_app_conversation_skills_endpoint.py`)

## Type
- [x] Bug fix

## Notes
Part of OpenHands/enterprise#74; the second (design-level) issue in that report — Settings→Skills listing
a different, underscore-named legacy set than the agent loads — is left for maintainer direction.

### DIFF
--- a/openhands/app_server/app_conversation/app_conversation_router.py
+++ b/openhands/app_server/app_conversation/app_conversation_router.py
@@ -1351,6 +1351,16 @@ async def get_conversation_skills(
                 ctx.agent_server_url,
             )
 
+            # Drop user-disabled skills so the panel matches what the agent
+            # actually loads. The agent context already filters these in
+            # AppConversationServiceBase._load_skills_and_update_agent; without
+            # the same filter here the endpoint advertises skills the agent
+            # does not have.
+            user = await app_conversation_service.user_context.get_user_info()
+            if user.disabled_skills:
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #14845, opened 2026-06-16
fix: show LLM timeout in advanced settings

wgu9:
HUMAN:
I kept this focused on the settings visibility gap from #8768. The timeout already exists in the SDK settings model and the backend already saves it, but the UI can hide it behind the deepest All tier. That makes the local LLM workaround hard to discover. This makes `llm.timeout` show up in Advanced while still using the SDK field as the source of truth.

## Summary
- show `llm.timeout` in the Advanced settings tier even when the SDK schema reports it as minor
- keep the SDK schema/value/default as the source of truth and only override the UI visibility tier
- update settings schema/page tests so timeout is discoverable from Advanced while true minor fields remain All-only

## Related issue
Part of #8768

## Testing
- `cd frontend && npm test -- --run __tests__/utils/sdk-settings-schema.test.ts __tests__/components/features/settings/sdk-settings/sdk-section-page.test.tsx __tests__/routes/llm-settings.test.tsx`
- `cd frontend && npx prettier --check src/utils/sdk-settings-schema.ts __tests__/utils/sdk-settings-schema.test.ts __tests__/components/features/settings/sdk-settings/sdk-section-page.test.tsx`
- `cd frontend && npx eslint src/utils/sdk-settings-schema.ts`
- `cd frontend && npm run typecheck:staged`
- `cd frontend && npm run check-translation-completeness`
- `git diff --check`
- commit hooks: frontend lint-staged/typecheck/translation and backend pre-commit

---

**Give a summary of what the PR does, explaining any non-trivial design decisions**

This PR makes the existing SDK-owned `llm.timeout` setting visible in the Advanced settings tier.

The field already exists in the SDK schema, has a default, and is already persisted by the settings API. The missing part was discoverability: if the schema marks it as `minor`, users looking for the slow local LLM workaround have to switch all the way to the All view. I added a small frontend visibility override so `llm.timeout` is treated as `major` for display purposes only. The SDK field remains the single source of truth for the actual setting.

**Link issue(s) that this PR fixes or relates to**

Part of #8768

**Give a summary of what the testing strategy is for this PR**

I added schema helper coverage for `llm.timeout` being visible in Advanced even when the SDK schema marks it minor, kept a component test for true minor fields staying All-only, and reran the LLM settings route tests. I also ran the frontend typecheck, translation completeness check, targeted Prettier/ESLint checks, `git diff --check`, and the repository commit hooks.


neubig:

Thank you for understanding.

PR #14864, opened 2026-06-17
feat(enterprise): add admin user-provisioning endpoint

chuckbutkus:
HUMAN:

## Summary

Adds POST /api/organizations/provision-user, an admin endpoint that creates a new user directly in the org specified by X-Org-Id:

- Creates the Keycloak user in the configured realm with a non-temporary password (generated when not supplied) and emailVerified=true.
- Requests an offline refresh token via ROPC and stores it through TokenManager.store_offline_token so the new account is immediately usable for backend operations.
- Creates the OpenHands user (mirroring keycloak_callback) and stamps email_verified, accepted_tos, user_consents_to_analytics, and onboarding_completed directly to skip UI interstitials.
- Sets up LiteLLM integration and adds the user to the target org as 'member'.
- Mints an API key pinned to the target org.

Access is gated by a new Permission.PROVISION_USER, granted to OWNER and ADMIN roles only (not MEMBER). On post-Keycloak failure the Keycloak user is best-effort cleaned up to avoid orphaned identities.

Includes 16 unit tests covering password generation, permission wiring, request validation, happy path, missing-org (404), Keycloak conflict (409), post-Keycloak failure cleanup, and the personal-org skip case.
-

## Issue Number
https://linear.app/all-hands-ai/issue/APP-2363

## How to Test

- Create a user with /api/organizations/provision-user endpoint
- Use the returned API key to create a conversation

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-5803721   docker.openhands.dev/openhands/openhands:5803721
```


enterprise/server/routes/user_provisioning.py
```diff
@@ -0,0 +1,379 @@
...
+# path live under ``/api/organizations`` so they sit alongside the rest
+# of the org-management surface, but they are intentionally separated
+# from ``org_router`` (which carries the ``REJECT_X_ORG_ID_PATH_MISMATCH``
+# guard for ``/{org_id}/...`` routes — that guard would no-op here, but
+# keeping the routers split makes the intent explicit).
+user_provisioning_router = APIRouter(prefix='/api/organizations', tags=['Orgs'])
+
+# Default role assigned to newly provisioned users in the target org.
+DEFAULT_PROVISIONED_ROLE = 'member'
+
```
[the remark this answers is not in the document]
chuckbutkus: Good call — switched to a bounded `for _ in range(100)` loop with a defensive `RuntimeError` fallback in 56f055986 so termination is guaranteed by construction. The comment now also calls out the bound explicitly.

_Reply posted by an AI agent (OpenHands) on behalf of @chuckbutkus._

enterprise/server/routes/user_provisioning.py
```diff
@@ -0,0 +1,379 @@
...
+    email: EmailStr = Field(
+        ...,
+        description='Email address for the new user. Used as the Keycloak username.',
+    )
+    password: str | None = Field(
+        default=None,
+        min_length=8,
+        max_length=256,
+        description=(
```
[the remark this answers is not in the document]
chuckbutkus: Extracted to a documented `_utc_now_naive()` helper in 56f055986 so the naive-UTC convention used by `User.accepted_tos` is self-explanatory at the call site.

_Reply posted by an AI agent (OpenHands) on behalf of @chuckbutkus._

enterprise/server/routes/user_provisioning.py
```diff
@@ -0,0 +1,396 @@
...
+        )
+
+    # Anything after this point may need to roll back the Keycloak user
+    # to avoid orphaning an account that has no matching OpenHands row.
+    try:
+        # 2. Get an offline token for the new user and store it. This
+        # mirrors what ``keycloak_offline_callback`` does at the end of
+        # the interactive flow so the user is immediately usable for
+        # backend operations.
+        offline_token = await token_manager.request_offline_token(
```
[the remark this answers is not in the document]
chuckbutkus: Went with option (b) and documented this as a known partial-cleanup gap in 5f58d2524. After working through it, option (a) (user-first ordering) is actually strictly worse: if the offline-token step then fails, the unwind path would leak the entire OpenHands cascade — user row + personal org + default settings + owner OrgMember row — instead of a single encrypted token blob. The token-first ordering also matches what `keycloak_offline_callback` does in the production OAuth flow. The orphan token row is keyed to a Keycloak `sub` that `_cleanup_keycloak_user` removes, so it cannot be used to authenticate. Module docstring now has a `Partial-cleanup gap` section explaining this, and the inline route comment cross-references it.

_Reply posted by an AI agent (OpenHands) on behalf of @chuckbutkus._

enterprise/server/routes/user_provisioning.py
```diff
@@ -0,0 +1,396 @@
...
+
+
+class ProvisionUserRequest(BaseModel):
+    """Payload for ``POST /api/organizations/provision-user``."""
+
+    email: EmailStr = Field(
+        ...,
+        description='Email address for the new user. Used as the Keycloak username.',
+    )
+    password: str | None = Field(
```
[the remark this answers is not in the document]
chuckbutkus: Documented in 5f58d2524 rather than re-enforcing the policy. Caller-supplied passwords are sent to Keycloak as-is and must satisfy the realm password policy (length, character classes, blacklist, etc.); policy violations surface as a 409 from the existing Keycloak error path. The generated-password complexity exists solely to ensure we always satisfy that policy — we intentionally do not duplicate it on our side because Keycloak is the single source of truth for password policy and divergent rules would silently drift. The expanded field description on `ProvisionUserRequest.password` now states this explicitly for admin guidance.

_Reply posted by an AI agent (OpenHands) on behalf of @chuckbutkus._

enterprise/server/routes/user_provisioning.py
```diff
@@ -0,0 +1,396 @@
...
+    """Create a new user and add them to the caller's selected org.
+
+    The target org is taken from the ``X-Org-Id`` header (resolved by
+    ``EFFECTIVE_ORG_ID``). The caller must hold the ``PROVISION_USER``
+    permission in that org (owner or admin).
+
+    Returns the email, password (generated if not supplied) and the
+    new user's API key bound to the target org.
+    """
+    email = body.email.lower().strip()
```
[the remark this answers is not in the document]
chuckbutkus: Added an explicit class-level docstring on `ProvisionUserResponse` in 5f58d2524 calling out that the plaintext password is **intentionally** returned (this is the only point at which it is recoverable, since the endpoint bypasses the normal email-based set-password flow), and documenting the expected handling: out-of-band hand-off by the admin, do-not-log, prefer TLS-terminated transport. Also added a per-field description on `password` to make the intent visible in the OpenAPI schema.

_Reply posted by an AI agent (OpenHands) on behalf of @chuckbutkus._

jpshackelford (commented):
The provisioning endpoint assumes the target org already exists. There is a `POST /api/organizations` endpoint, but it is guarded by `get_admin_user_id`, which currently requires an `@openhands.dev` email unconditionally, including on self-hosted deployments.

Stepping back, there are three existing user/org-creation scenarios we should compare this endpoint against. None of them work cleanly with the new provisioning flow without additional support.

1. **Self-hosted default-org bootstrap flow**

   Existing OAuth/login behavior is:

   1. `UserStore.create_user(...)` creates the user and their personal workspace.
   2. If `OPENHANDS_DEFAULT_ORG_ENABLED` / `OH_DEFAULT_ORG_ENABLED` is enabled, `DefaultOrgBootstrapService.apply_for_user(...)` may run. Note this flag is optional in self-hosted installs and is not the default.
   3. The bootstrap service resolves the default team org by first looking for an existing default org, then possibly adopting a sole existing team org, and finally creating the configured default org with the logging-in user as owner if no team org exists.
   4. If `OPENHANDS_DEFAULT_ORG_AUTO_ADD_USERS` / `OH_DEFAULT_ORG_AUTO_ADD_USERS` is enabled, later users are added to that default org as members.
   5. If `HIDE_PERSONAL_WORKSPACES` is enabled, users can be moved out of their personal org on login.

   Limitation with this PR: `provision-user` does not run that bootstrap service. There also does not appear to be a clean external API recipe for using default-org bootstrap as a prerequisite: `GET /api/organizations` returns the caller’s orgs and marks which one is personal, but it does not expose an `is_default` marker or a dedicated “resolve/create default org” operation. So this only works if the default org already exists and the caller can reliably identify its ID. In the no-org-yet self-hosted case, the bootstrap flow is not practically usable from the public API surface today.

2. **Explicit team-org creation flow**

   The intended manual flow would be:

   1. Create a normal team org.
   2. Provision the user directly into that org with `X-Org-Id`.
   3. Optionally promote the provisioned user to admin/owner.
   4. Optionally remove the original org creator from the org.

   Limitation with this PR: step 1 is blocked for self-hosted customer-domain admins because `POST /api/organizations` is still guarded by `get_admin_user_id`. The new endpoint does directly add the provisioned user to the org, so no invite acceptance is required, but cleanup/ownership transfer is still awkward. To make the provisioned user the durable owner and remove the creating user, the caller must perform additional role-update/removal calls, and removing the creating user currently requires acting as another user after ownership has been transferred. That may not block this customer case, but it is a limitation we should relax: if another owner exists, an owner should be able to remove themselves / leave the org, or the provisioning/org-creation flow should support explicit ownership transfer/removal in one admin-controlled flow.

3. **Personal-org-only provisioning flow**

   Existing OAuth behavior creates a personal org through `UserStore.create_user(...)` with the invariant `Org.id == User.id == UUID(keycloak.sub)`.

   Limitation with this PR: a personal org is not something `POST /api/organizations` can create correctly, because that endpoint generates a normal team org owned by the caller. `provision-user` does create the new user’s personal org as a side effect of `UserStore.create_user(...)`, but it cannot realistically use that as the target org because `X-Org-Id` is authorized before the new user exists, the caller is not a member of the new user’s personal org, and the returned API key is pinned to the supplied target org. If personal-org-only provisioning is a supported goal, it needs to be an explicit mode of this endpoint rather than a variation of manual org creation.

Next steps: decide which of these scenarios must be supported for the product/customer goal, then make that path work end-to-end.

1. If the goal is self-hosted default-org onboarding, have `provision-user` internally reuse the OAuth login bootstrap path or add a self-hosted-compatible API that explicitly resolves/creates the default org and returns its ID.
2. If the goal is explicit team-org onboarding, fix `POST /api/organizations` for self-hosted authorization and consider relaxing org-member removal/ownership-transfer so the creating admin can hand off the org cleanly.
3. If the goal is personal-org-only provisioning, add an explicit provisioning mode that creates the user/personal org and mints credentials for that personal org without requiring a pre-existing caller-selected `X-Org-Id`.

Right now the endpoint creates the OAuth-style personal org for the new user, but still requires a pre-existing target org that the caller can already administer. That leaves the no-org-yet self-hosted case and the ownership-transfer case under-specified.

_This review comment was created by an AI agent (OpenHands) on behalf of @jpshackelford ._

enterprise/server/auth/token_manager.py
```diff
@@ -812,6 +812,89 @@ async def delete_keycloak_user(self, user_id: str) -> bool:
...
+            'username': email,
+            'enabled': True,
+            'emailVerified': email_verified,
+        }
+        user_id = await keycloak_admin.a_create_user(payload, exist_ok=False)
+        await keycloak_admin.a_set_user_password(
+            user_id=user_id,
+            password=password,
+            temporary=False,
+        )
```
jpshackelford: 🔴 **Important**: `create_keycloak_user()` creates the Keycloak user first, then sets the password. If `a_set_user_password()` fails — very plausible when a caller-supplied password passes the local `min_length=8` check but violates the realm password policy — this helper raises before returning `user_id`. The route’s cleanup path never sees that ID, so the Keycloak user is left behind without a usable password. Either create the user with credentials atomically if Keycloak supports that in this flow, or catch failures after `a_create_user()` and delete that specific Keycloak user before re-raising. Please also add a test where `a_create_user()` succeeds and `a_set_user_password()` raises.

enterprise/server/routes/user_provisioning.py
```diff
@@ -0,0 +1,435 @@
...
+            'provision_user:post_keycloak_failure',
+            extra={
+                'caller_user_id': caller_user_id,
+                'target_org_id': str(target_org_id),
+                'kc_user_id': kc_user_id,
+                'email': email,
+                'error': str(e),
+            },
+        )
+        await _cleanup_keycloak_user(token_manager, kc_user_id)
```
jpshackelford: 🔴 **Critical**: after `UserStore.create_user()` succeeds, any later failure — `_set_user_provisioned_flags`, `create_litellm_integration`, role lookup, org-member insert, API-key creation — only calls `_cleanup_keycloak_user()`. That leaves the OpenHands `User`, personal org, owner org-member row, default settings, and possibly target-org/LiteLLM state pointing at a Keycloak subject that no longer exists. The module docstring discusses the offline-token partial-cleanup gap, but the larger post-`UserStore.create_user()` orphaning problem is not handled. This needs a real compensation path for database/LiteLLM state, or the provisioning flow should be reorganized into a service with explicit transactional boundaries and rollback steps. Please add failure tests after `UserStore.create_user()` succeeds, not only before it.

jpshackelford (commented):
Creating the API key inside the provisioning endpoint is precisely the right fit for this scenario once a valid target org has been resolved. Since `POST /api/keys` creates keys only for the currently authenticated user, provisioning needs to mint the new user’s key server-side rather than requiring a login as that user. Passing `org_id=target_org_id` explicitly is also important because the new user’s `current_org_id` will generally still be their personal workspace. The remaining issue is target-org resolution and cleanup/compensation, not the decision to return an org-scoped key from the provisioning response.

The caveat is that “valid target org” means something different in each supported scenario:

1. **Default-org bootstrap flow**: the key should be bound to the resolved default/team org, but only after the flow has actually resolved or created that org and ensured the new user is a member. If a default org already exists but auto-add is disabled, simply resolving the default org is not enough; the provisioning flow must explicitly add the user or fail clearly.
2. **Explicit team-org flow**: the current API-key behavior works well after the endpoint has directly added the new user to the selected team org. No invite acceptance is required because the provisioning endpoint creates an active membership before minting the key.
3. **Personal-org-only flow**: the key should be bound to the newly created personal org, i.e. `org_id=UUID(kc_user_id)`, not to a caller-supplied `X-Org-Id`. That requires an explicit personal-org provisioning mode because the caller cannot pre-authorize against the new user’s personal org before the user exists.

_This review was created by an AI agent (OpenHands) on behalf of @jpshackelford._

enterprise/server/routes/user_provisioning.py
```diff
@@ -0,0 +1,435 @@
...
+    # member with the right role, but the org row could have been
+    # deleted between the membership check and this code path; an
+    # explicit fetch produces a clean 404 instead of a confusing 500
+    # later in ``OrgService.create_litellm_integration``.
+    org = await OrgStore.get_org_by_id(target_org_id)
+    if not org:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail='Target organization not found',
+        )
```
jpshackelford, line 273: 🔴 **Critical**: the permission model grants `PROVISION_USER` to `OWNER`/`ADMIN`, but every normal user is owner of their personal org. This route only checks that the target org exists, so enabling `USER_PROVISIONING_ENABLED` lets any user provision new Keycloak/OpenHands users into their personal workspace and receive credentials/API keys. Existing invitation code explicitly rejects personal workspaces; this endpoint should do the same unless the intended product behavior is that every user can create more accounts. At minimum, reject `target_org_id == UUID(caller_user_id)` and add a test for it.

enterprise/server/auth/token_manager.py
chuckbutkus: Fixed in e0b5417df6c23c1c3ca6144e62472520bf80822c.

`create_keycloak_user` now wraps the `a_set_user_password` call in a try/except; if password setup fails (e.g. realm password policy violation after our local `min_length=8` check), the just-created Keycloak user is best-effort deleted via `asyncio.to_thread(keycloak_admin.delete_user, user_id)` before the original exception is re-raised. Secondary cleanup failures are swallowed so they cannot mask the real cause of the request failure.

Added three unit tests in `enterprise/tests/unit/test_token_manager.py::TestCreateKeycloakUser`:
- `test_create_keycloak_user_success` — happy path, no cleanup runs.
- `test_create_keycloak_user_password_failure_cleans_up` — `a_create_user` succeeds, `a_set_user_password` raises, asserts `delete_user` was called with the specific user id and the original `KeycloakError` propagates.
- `test_create_keycloak_user_password_failure_cleanup_swallowed` — both calls fail; asserts the password error (not the cleanup error) is what propagates.

_This reply was posted by an AI agent (OpenHands) on behalf of the PR author._

chuckbutkus: Follow-up — went with the atomic approach in c2a4c0915b71bab696492bc5f918edeb754c1e9d instead of the catch/cleanup fallback. Keycloak's `UserRepresentation` accepts a `credentials` array inline, so the password is now set in the same POST that creates the user. A realm password-policy violation rejects the whole request, so there is no orphan window to clean up at all.

The test was simplified accordingly: it now asserts the inline `credentials` payload (`{type: 'password', value: ..., temporary: False}`) is sent and that no follow-up `a_set_user_password` call is made. The atomic-failure test covers the realm-policy-rejection case directly.

_This reply was posted by an AI agent (OpenHands) on behalf of the PR author._

enterprise/server/routes/user_provisioning.py
chuckbutkus: Fixed in dcef7bfa869a0f11775fb1d3006602105cbf31d6.

**What changed**

- Introduced `_rollback_partial_provision` which compensates the *full* state created by the provisioning flow, not just the Keycloak user. It tracks progress via two locals in the route — `openhands_user_id` (set after `UserStore.create_user`) and `target_membership_added` (set after `OrgMemberStore.add_user_to_org`) — and unwinds in this order:
  1. `OrgMemberStore.remove_user_from_org(target_org_id, openhands_user_id)` if the target-org membership was added. This is required *before* step 2 because `delete_org_cascade` only cascade-deletes the `User` row when the user is the sole orphan of the personal org; a surviving target-org membership would cause the cascade to just reassign `current_org_id` and leave the User row behind.
  2. `OrgStore.delete_org_cascade(personal_org_id, requester_user_id=kc_user_id)` — wipes the personal Org, owner OrgMember, default settings, personal-org LiteLLM team, org-scoped tables (including `api_keys WHERE org_id = personal_org_id`), and the User row in a single transaction.
  3. `token_manager.delete_keycloak_user(kc_user_id)`.
  Each step is wrapped individually so a secondary cleanup error never masks the original provisioning failure.
- Both `except` arms in the route now call `_rollback_partial_provision` with the tracked locals so the unwind targets exactly the state that was created.
- The module docstring's old single "Partial-cleanup gap" note has been replaced with an explicit **Compensation strategy** section + two narrow known-gap notes (offline token row; LiteLLM target-team membership). The LiteLLM target-team gap is documented as a follow-up that should ride on a future `LiteLlmManager.remove_user_from_team(sub, org_id)` helper rather than reaching into private internals from this route.

**Tests added (all assert behaviour *after* `UserStore.create_user` succeeds, per the review ask):**
- `test_rollback_when_set_flags_fails` — failure in `_set_user_provisioned_flags`. Asserts cascade runs, target-membership removal does not (never added).
- `test_rollback_when_litellm_integration_fails` — same shape; before `add_user_to_org`.
- `test_rollback_when_add_user_to_org_fails` — confirms `remove_user_from_org` is **not** called when the membership row was never inserted (avoids a misleading "row not found" log).
- `test_rollback_when_api_key_creation_fails` — exercises the *full* rollback path. Uses `attach_mock` on a parent mock to assert ordering: remove-member ➜ cascade ➜ keycloak-delete.
- `test_rollback_swallows_secondary_failures` — both DB-side rollback steps raise; asserts the route still returns `500` from the original failure *and* still attempts the Keycloak delete.
- `test_rollback_before_user_created_only_cleans_keycloak` — replaces the previous offline-token-failure test; confirms only the Keycloak delete runs when `UserStore.create_user` was never reached.
- The happy-path test now also asserts that none of the rollback collaborators run on success.

_This reply was posted by an AI agent (OpenHands) on behalf of the PR author._

enterprise/server/routes/user_provisioning.py
chuckbutkus: Fixed in 0513be87a535eaaf13f815ce8c7382be26ed7fd3.

The route now rejects `target_org_id == UUID(caller_user_id)` with **403 Forbidden** (`'Cannot provision users into a personal workspace'`), mirroring the long-standing personal-workspace rejection in `server.services.org_invitation_service` (same status code, same message shape). The check runs after the 404 existence fetch and *before* any Keycloak side effect, so a rejected call mutates no state.

Added `test_personal_workspace_rejected_with_403` which sets `target_org_id = UUID(caller_user_id)` and asserts:
- HTTP 403 with `'personal workspace'` in the detail.
- None of the side-effect collaborators run: `create_keycloak_user`, `delete_keycloak_user`, `delete_org_cascade`, `remove_user_from_org`.

Left the defensive `if target_org_id != openhands_user_id:` guard around the add-to-org step in place (now belt-and-suspenders given the upfront check; the cost is minimal and the corresponding test still documents the invariant).

_This reply was posted by an AI agent (OpenHands) on behalf of the PR author._

malhotra5 (approved):
LGTM!

PR #14884, opened 2026-06-18
fix: stop the connection-error banner from flashing on transient websocket drops

ak684:
HUMAN: The red "failed to connect" banner flashes too eagerly — it pops on transient/benign websocket reconnects (idle socket timeouts, backgrounded-tab reaps, brief network blips) even when the user isn't doing anything, then clears itself a few seconds later. It makes the UI feel broken when nothing actually is.

## Problem
The banner (`"Failed to connect to server"`) was set **immediately** on the first abnormal websocket close, in both the main and planning socket `onError` handlers in `conversation-websocket-context.tsx`, with no debounce or grace period. Because the socket auto-reconnects on a fixed ~3s delay, every benign drop produced a visible flash: drop → red banner instantly → ~3s later reconnect → banner cleared. The only existing guard (`hasConnectedRef`) merely suppressed it before the *first* connect, not the steady-state reconnect churn that causes the flashing.

## Fix
Gate the banner on **intent/state** rather than raw socket state. It now appears only when:
- the agent is **actively running** (`execution_status === RUNNING`), **and**
- the **main** socket stays down past a short grace window (5s) — a successful reconnect cancels the pending banner before it ever shows.

This is safe because the connection self-heals: both sockets reconnect with `resend_all=true`, so the full event history is replayed and de-duplicated on reconnect — a transient drop loses nothing. A genuinely closed socket still surfaces in the passive status indicator (`agent-status.tsx`), and the send path already shows a real error when a message can't be delivered. Planning-socket churn no longer raises a connection banner on its own (the banner tracks the main connection).

## Behavior
- Idle user, transient drop → **no banner** (self-heals silently; only the status indicator reflects it).
- Active run, transient drop that reconnects within the grace window → **no banner**.
- Active run, socket genuinely stays down → banner appears after the grace window, and auto-clears on reconnect.

## Tests
Updated `conversation-websocket-handler.test.tsx`:
- no flash on an idle transient disconnect
- no flash when a transient drop reconnects during an active run
- banner *does* appear when the agent is running and the socket stays down
- banner clears on reconnect

`npm run lint` and the websocket + error-banner test files pass locally.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-b623744   docker.openhands.dev/openhands/openhands:b623744
```

### DIFF
--- a/frontend/__tests__/conversation-websocket-handler.test.tsx
+++ b/frontend/__tests__/conversation-websocket-handler.test.tsx
@@ -40,6 +40,7 @@ import {
   ConversationWebSocketProvider,
   useConversationWebSocket,
   CONNECTION_ERROR_GRACE_MS,
+  CONNECTION_ERROR_MESSAGE,
 } from "#/contexts/conversation-websocket-context";
 import { conversationWebSocketTestSetup } from "./helpers/msw-websocket-setup";
 import { useEventStore } from "#/stores/use-event-store";
@@ -98,6 +99,10 @@ function renderWithWebSocketContext(
   conversationId = "test-conversation-default",
   conversationUrl = "http://localhost:3000/api/conversations/test-conversation-default",
   sessionApiKey: string | null = null,
+  extraProps: {
+    subConversations?: { id: string; conversation_url: string }[];
+    subConversationIds?: string[];
+  } = {},
 ) {
   const queryClient = new QueryClient({
     defaultOptions: {
@@ -117,6 +122,9 @@ function renderWithWebSocketContext(
                 conversationId={conversationId}
                 conversationUrl={conversationUrl}
                 sessionApiKey={sessionApiKey}
+                // eslint-disable-next-line @typescript-eslint/no-explicit-any
+                subConversations={extraProps.subConversations as any}
+                subConversationIds={extraProps.subConversationIds}
               >
                 {children}
               </ConversationWebSocketProvider>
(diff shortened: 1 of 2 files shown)


ak684:
Thanks for the review — I agree with both points and addressed them in b623744f2.

**CRITICAL (planning success can clear the main banner) — agreed, fixed.**
Correct: I scoped *setting* the banner to the main socket but left *clearing* unscoped, so the planning socket's `onOpen` and any non-error event through `handleNonErrorEvent` could erase a main-connection outage banner while main was still down — the single shared error string makes every producer a potential eraser.

I took the "scope the clearing" route: the message is now a named constant `CONNECTION_ERROR_MESSAGE`, and both `handleNonErrorEvent` and the planning `onOpen` skip clearing it. So the banner is owned by the main socket end-to-end — only the main `onError` sets it and only the main `onOpen` clears it. I chose this over tagging/splitting the error store because that store is shared app-wide (budget/credit, agent errors, …) and restructuring it is a larger, riskier change than this bug warrants; the constant guard is the minimal fix that restores the invariant. Happy to do the store-level split as a follow-up if you'd prefer it long-term.

**TESTING GAP — agreed, added.**
New dual-socket regression test (`should keep the connection banner when the planning socket is active while main stays down`): agent running, main socket drops and never recovers (banner appears after the grace window), planning socket stays connected and emits a non-error event — asserts the main banner persists. The MSW mock is a wildcard URL so both sockets route through it; I distinguish them by the connecting URL. Pre-fix this fails (planning event clears the banner); post-fix it passes.

All `conversation-websocket-handler` tests + `npm run lint` + typecheck are green locally. Re-requesting review.

neubig:

Thank you for understanding.

PR #14885, opened 2026-06-18
feat: surface sandbox start-failure reason in conversation start errors

ak684:
Coordinated 2-PR change with **OpenHands/runtime-api#615** (shared key `status_detail`).

[x] A human has tested these changes

## What

When a sandbox can't start, surface the real reason instead of an opaque `Sandbox failed to start within 120s`:

- Reads the runtime's `status_detail` onto `SandboxInfo` and appends it to the wait error/timeout messages (backward-compatible: `.get()` defaults to `None` against an older runtime-api).
- For transient resource-exhaustion scheduling failures (`Insufficient ephemeral-storage/cpu/memory/pods`), shows a retryable **"The system is at capacity right now (...). Please try again in a few minutes."** instead of the raw timeout. Other reasons (ImagePullBackOff, device/affinity) keep the raw detail, since "try again" wouldn't be true for those.

## Why

Today an out-of-capacity start hangs 120s and reports an opaque timeout error message. Now it reports "at capacity, retry" which is more informative.

## Tests

15 passing in `tests/unit/app_server/test_sandbox_service.py`: capacity detail -> friendly message (at ERROR state and at timeout), non-capacity detail keeps the raw reason.

Note: we considered failing fast (giving up before the 120s timeout once a sandbox is confirmed unschedulable) but dropped it. On a fixed-size cluster the wait doubles as a "wait for a slot to free up" window as other conversations end, so failing early would spuriously fail starts that would have succeeded. The copy is the real UX win and needs no flag; a fail-fast can be revisited with real data if latency becomes a complaint.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-6bc5228   docker.openhands.dev/openhands/openhands:6bc5228
```

### DIFF
--- a/openhands/app_server/sandbox/sandbox_service.py
+++ b/openhands/app_server/sandbox/sandbox_service.py
@@ -117,13 +117,15 @@ class SandboxService(ABC):
             SandboxError: If sandbox not found, enters ERROR state, or times out
         """
         start = time.time()
+        sandbox: SandboxInfo | None = None
         while time.time() - start <= timeout:
             sandbox = await self.get_sandbox(sandbox_id)
             if sandbox is None:
                 raise SandboxError(f'Sandbox not found: {sandbox_id}')
 
             if sandbox.status == SandboxStatus.ERROR:
-                raise SandboxError(f'Sandbox entered error state: {sandbox_id}')
+                detail = f' ({sandbox.status_detail})' if sandbox.status_detail else ''
+                raise SandboxError(f'Sandbox entered error state: {sandbox_id}{detail}')
 
             if sandbox.status == SandboxStatus.RUNNING:
                 # Optionally verify agent server is alive to avoid race conditions
@@ -137,7 +139,14 @@ class SandboxService(ABC):
 
             await asyncio.sleep(poll_interval)
 
-        raise SandboxError(f'Sandbox failed to start within {timeout}s: {sandbox_id}')
+        detail = (
+            f' ({sandbox.status_detail})'
+            if sandbox is not None and sandbox.status_detail
+            else ''
+        )
+        raise SandboxError(
+            f'Sandbox failed to start within {timeout}s: {sandbox_id}{detail}'
+        )
 
     async def _check_agent_server_alive(
         self, sandbox: SandboxInfo, httpx_client: httpx.AsyncClient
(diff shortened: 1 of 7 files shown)


ak684:
**Live-validated on a real OHE (embedded-cluster) install.** Overlaid this change onto a deployed instance and induced a genuine `Insufficient ephemeral-storage` scheduling failure (single node driven to ~95% ephemeral-storage). Result: a real conversation-start surfaced the retryable message `"The system is at capacity right now (…Insufficient ephemeral-storage…). Please try again in a few minutes."` sourced from the runtime's `status_detail`, instead of the opaque 120s timeout. runtime-api's `GET /runtime/{id}` returned the populated `status_detail` directly; the app's `wait_for_sandbox_running` timeout branch produced the friendly error. Mirrors a customer's capacity incident. No regression on the happy path.

dylan-openhands (approved):

openhands/app_server/sandbox/sandbox_service.py
```diff
@@ -26,6 +26,28 @@
 WEBHOOK_CALLBACK_VARIABLE = 'OH_WEBHOOKS_0_BASE_URL'
 ALLOW_CORS_ORIGINS_VARIABLE = 'OH_ALLOW_CORS_ORIGINS_0'
 
+# status_detail substrings meaning "the node is full right now", which frees up
+# as other sandboxes finish -- so 'try again shortly' is accurate. Other
+# unschedulable reasons (device/affinity/taint) keep the raw reason instead.
```
jlav: I'm a little concerned about surfacing the raw `status_details` from runtime-api.

Pod conditions from k8s will contain environment details that we and self hosted customers may not want surfaced to consumers. Internal IDs, secret names, host file system paths, cluster sizes, labels, taints etc, are all details that appear in those messages, and many orgs would consider that  information sensitive.

I think these generic errors like "Insufficient cpu" are **great** and very informative. I think we could break out a separate catch-all like "Invalid configuration" for the less-obvious and less-frequent error cases that we might run into.

I think once we classify and present the general problem to the consumer, it's easier to  debug the problem and for the nitty-gritty details we can lean on logs.

jlav (approved):

openhands/app_server/sandbox/sandbox_service.py
jlav: LGTM after 69d73b611526da2ea208b381e46a7e0e662cd9df

Thank you!

PR #14895, opened 2026-06-18
feat: PLTF-2968 allow per-deployment resolver macro via OH_RESOLVER_LABEL

### LINKED ISSUE #14894: Resolver trigger macro is inferred from WEB_HOST and can't be configured per environment
## Why
The resolver's trigger macro — the mention (`@openhands` vs `@openhands-exp`) and the issue label that fire it — is inferred from `WEB_HOST` by substring matching in `get_oh_labels()` (`enterprise/integrations/utils.py`). Only hosts containing `staging` or `local` get a distinct macro (`openhands-exp` / `@openhands-exp`); every other host falls back to the production `openhands` / `@openhands`.

This blocks running more than one non-production deployment cleanly:

- Any non-prod deployment whose host doesn't contain `staging`/`local` listens for the **same** `@openhands` as production. If it is installed alongside the production app on shared repos, it competes with prod and reacts to real issues/PRs.
- Adding a new environment requires editing `get_oh_labels()` each time, and the logic is fragile — it keys off accidental substrings of hostnames rather than an explicit setting.

Making the macro an explicit per-deployment value is safe: when the override is unset the existing host inference is preserved, so production and current staging/local behavior are unchanged.

## Validation
A fix should let each deployment declare its own macro via an environment variable (e.g. `OH_RESOLVER_LABEL`), falling back to today's `WEB_HOST` inference when the variable is absent. Acceptance:

- Override set (e.g. `OH_RESOLVER_LABEL=openhands-dev`) → `get_oh_labels()` returns `('openhands-dev', '@openhands-dev')`.
- Override unset on a `staging`/`local` host → still `('openhands-exp', '@openhands-exp')`.
- Override unset on any other host → still `('openhands', '@openhands')`.
- Covered by unit tests for `get_oh_labels()` across the three cases above.

---
*This issue was drafted by an AI agent on behalf of the user.*

aivong-openhands:
HUMAN:

AGENT:

## Why

The resolver trigger macro (the issue label and `@mention` that fire it) was inferred from `WEB_HOST` by substring matching in `get_oh_labels()`: only hosts containing `staging`/`local` got a distinct macro (`openhands-exp`); everything else fell back to the production `openhands`. So any additional non-production deployment whose host lacked those substrings listened for the same `@openhands` as production and competed with it on shared repos, and standing up a new environment required editing this function each time. An explicit override makes the macro a per-deployment setting, and is safe because it is opt-in — when unset (or empty) the existing host inference is used unchanged.

## Summary

- `get_oh_labels()` now returns `(OH_RESOLVER_LABEL, '@' + OH_RESOLVER_LABEL)` when that env var is set (trimmed); otherwise it falls through to the unchanged `WEB_HOST` inference.
- Added unit cases to `TestGithubLabels` for override-set, override-trimmed, and empty-override-falls-back.

## Issue Number

Closes #14894

## How to Test

```bash
PYTHONPATH=".:$PYTHONPATH" poetry run --project=enterprise pytest -p no:ddtrace ./enterprise/tests/unit/test_github_view.py::TestGithubLabels
```

All cases pass: with `OH_RESOLVER_LABEL` set, `get_oh_labels()` returns the override (e.g. `('openhands-dev', '@openhands-dev')`) regardless of host; unset/empty, staging/local still return `openhands-exp` and other hosts return `openhands`.

## Video/Screenshots

N/A — backend label-resolution logic, no UI surface.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Behavior is unchanged unless `OH_RESOLVER_LABEL` is set, so existing deployments are unaffected. Follow-up (separate, in the deployment config): set `OH_RESOLVER_LABEL` per environment and create the matching GitHub App + issue label so each environment triggers on its own macro instead of the production `@openhands`.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-9329abf   docker.openhands.dev/openhands/openhands:9329abf
```

### DIFF
--- a/enterprise/integrations/utils.py
+++ b/enterprise/integrations/utils.py
@@ -113,14 +113,24 @@ _jinja_env = Environment(loader=FileSystemLoader(OPENHANDS_RESOLVER_TEMPLATES_DI
 def get_oh_labels(web_host: str) -> tuple[str, str]:
     """Get the OpenHands labels based on the web host.
 
+    An explicit ``OH_RESOLVER_LABEL`` environment variable takes precedence and
+    lets each deployment declare its own trigger macro (issue label + mention)
+    without a code change. When it is unset, the macro is inferred from
+    ``web_host`` for backward compatibility.
+
     Args:
         web_host: The web host string to check
 
     Returns:
         A tuple of (oh_label, inline_oh_label) where:
-        - oh_label is 'openhands-exp' for staging/local hosts, 'openhands' otherwise
-        - inline_oh_label is '@openhands-exp' for staging/local hosts, '@openhands' otherwise
+        - oh_label is OH_RESOLVER_LABEL when set; otherwise 'openhands-exp' for
+          staging/local hosts and 'openhands' for everything else
+        - inline_oh_label is oh_label prefixed with '@'
     """
+    override = os.getenv('OH_RESOLVER_LABEL', '').strip()
+    if override:
+        return override, f'@{override}'
+
     web_host = web_host.strip()
     is_staging_or_local = 'staging' in web_host or 'local' in web_host
     oh_label = 'openhands-exp' if is_staging_or_local else 'openhands'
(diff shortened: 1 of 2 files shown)


malhotra5 (approved):
Nice!

PR #14896, opened 2026-06-18
fix: prevent conversation export overload

neubig:
HUMAN: This PR prevents large conversation trajectory downloads from overloading app servers by reading events once, streaming the zip output, and serializing duplicate exports for the same conversation with Redis in SaaS.

## Summary

Fixes the conversation trajectory download path that could amplify one large export into repeated full-history event reads and app-server saturation.

## What changed

- Added a single-pass `iter_events_for_export()` event service path with bounded event-load concurrency.
- Changed trajectory downloads to return a streaming zip instead of building a full zip response in memory.
- Added a Redis-backed per-conversation export lock with token-checked release/refresh.
- Mapped export conflicts and guard failures to explicit HTTP statuses: 409 duplicate export, 503 required lock unavailable, 413 event-limit exceeded.
- Kept the lock required in SaaS mode by default and best-effort outside SaaS to avoid breaking local/OSS installs without Redis.

## Root cause

The previous export loop used `page_iterator(search_events)`. `search_events()` lists and loads every event blob before applying pagination, so a 5k-event conversation caused every page to reload the same 5k events. Repeated overlapping downloads multiplied that into hundreds of thousands of blob reads and heavy zip generation on app pods.

## Validation

- `uv run pre-commit run --all-files --show-diff-on-failure --config ./dev_config/python/.pre-commit-config.yaml`
- `uv run pytest tests/unit/app_server/test_live_status_app_conversation_service.py tests/unit/app_server/test_filesystem_event_service.py tests/unit/app_server/test_app_conversation_router.py`
- `uv run ruff check openhands/app_server/app_conversation/app_conversation_router.py openhands/app_server/app_conversation/app_conversation_service.py openhands/app_server/app_conversation/live_status_app_conversation_service.py openhands/app_server/event/event_service.py openhands/app_server/event/event_service_base.py openhands/app_server/utils/redis_lock.py tests/unit/app_server/test_live_status_app_conversation_service.py tests/unit/app_server/test_filesystem_event_service.py tests/unit/app_server/test_app_conversation_router.py`

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-6c3097e   docker.openhands.dev/openhands/openhands:6c3097e
```

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -137,7 +137,13 @@ _EXPORT_LOCK_KEY_PREFIX = 'app_conversation_export'
 
 
 class _StreamingZipBuffer(io.RawIOBase):
-    """Small non-seekable writer used by zipfile to emit chunks incrementally."""
+    """Small non-seekable writer used by zipfile to emit chunks incrementally.
+
+    zipfile.ZipFile only needs write() and tell() from its underlying file
+    object when writing to a non-seekable stream.  Everything else
+    (flush, writable, seekable) is either unused by zipfile or already
+    handled by the io.RawIOBase defaults.
+    """
 
     def __init__(self):
         self._chunks: list[bytes] = []
@@ -152,15 +158,6 @@ class _StreamingZipBuffer(io.RawIOBase):
 
     def tell(self) -> int:
         return self._position
-
-    def flush(self):
-        pass
-
-    def writable(self) -> bool:
-        return True
-
-    def seekable(self) -> bool:
-        return False
 
     def drain(self) -> list[bytes]:
         chunks = self._chunks
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -4,7 +4,6 @@ import json
 import logging
 import os
 import re
-import time
 import zipfile
 from collections import defaultdict
 from collections.abc import Mapping
@@ -101,6 +100,7 @@ from openhands.app_server.utils.llm_metadata import (
     should_set_litellm_extra_body,
 )
 from openhands.app_server.utils.redis_lock import (
+    RedisLock,
     RedisLockUnavailable,
     try_acquire_redis_lock,
 )
@@ -166,6 +166,27 @@ class _StreamingZipBuffer(io.RawIOBase):
         chunks = self._chunks
         self._chunks = []
         return chunks
+
+
+async def _refresh_lock_periodically(
+    lock: RedisLock, interval: int, conversation_id: str
+) -> None:
+    """Keep a Redis export lock alive while a streaming response is in flight.
+
+    Runs as a background task (via asyncio.create_task) alongside the streaming
+    generator so the lock TTL is extended without blocking chunk generation.
+    The task is cancelled when the generator's finally-block fires.
+    """
+    try:
+        while True:
+            await asyncio.sleep(interval)
+            if not await lock.refresh():
+                _logger.warning(
+                    'conversation_export:lock_refresh_failed',
+                    extra={'conversation_id': conversation_id},
+                )
+    except asyncio.CancelledError:
+        pass
 
 
 # Limits for bootstrap-prompt resume (Solution A of issue #14260).
@@ -2510,22 +2531,25 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
         refresh_interval = self._conversation_export_lock_refresh_interval()
 
         async def stream():
-            last_refresh = time.monotonic()
+            # Refresh the lock in a background task so the streaming loop stays
+            # simple and lock maintenance doesn't block chunk generation.
+            refresh_task = (
+                asyncio.create_task(
+                    _refresh_lock_periodically(
+                        export_lock, refresh_interval, str(conversation_id)
+                    )
+                )
+                if export_lock
+                else None
+            )
             try:
                 async for chunk in self._stream_conversation_zip(
                     conversation_id, conversation_info
                 ):
-                    now = time.monotonic()
-                    if export_lock and now - last_refresh >= refresh_interval:
-                        refreshed = await export_lock.refresh()
-                        if not refreshed:
-                            _logger.warning(
-                                'conversation_export:lock_refresh_failed',
-                                extra={'conversation_id': str(conversation_id)},
-                            )
-                        last_refresh = now
                     yield chunk
             finally:
+                if refresh_task is not None:
+                    refresh_task.cancel()
                 if export_lock:
                     await export_lock.release()
 
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -100,6 +100,7 @@ from openhands.app_server.utils.llm_metadata import (
     should_set_litellm_extra_body,
 )
 from openhands.app_server.utils.redis_lock import (
+    LockError,
     RedisLockUnavailable,
     refresh_lock_periodically,
     try_acquire_redis_lock,
@@ -2501,7 +2502,10 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
             await self._validate_conversation_export_size(conversation_id)
         except Exception:
             if export_lock:
-                await export_lock.release()
+                try:
+                    await export_lock.release()
+                except LockError:
+                    pass
             raise
 
         refresh_interval = self._conversation_export_lock_refresh_interval()
@@ -2525,7 +2529,13 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
                 if refresh_task is not None:
                     refresh_task.cancel()
                 if export_lock:
-                    await export_lock.release()
+                    try:
+                        await export_lock.release()
+                    except LockError:
+                        _logger.warning(
+                            'conversation_export:lock_release_failed',
+                            extra={'conversation_id': str(conversation_id)},
+                        )
 
         return stream()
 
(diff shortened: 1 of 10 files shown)


openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -2364,60 +2408,143 @@ async def _delete_from_database(
...
+    def _conversation_export_lock_required(self) -> bool:
+        if self.export_lock_required is not None:
+            return self.export_lock_required
+        return (self.app_mode or '').lower() == 'saas'
+
+    async def _stream_conversation_zip(
+        self, conversation_id: UUID, conversation_info: AppConversationInfo
+    ) -> AsyncGenerator[bytes, None]:
+        zip_buffer = _StreamingZipBuffer()
+        with zipfile.ZipFile(zip_buffer, 'w', zipfile.ZIP_DEFLATED) as zipf:
```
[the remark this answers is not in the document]
neubig: Resolved in commit 77841c102 by making the ZipFile call explicit: _StreamingZipBuffer now subclasses io.RawIOBase and the constructor receives a narrow cast(BinaryIO, zip_buffer), matching the runtime file-like contract that zipfile accepts. I also reran ruff format, the full Python pre-commit config, and the focused export/router tests locally; the GitHub lint, enterprise lint, image, and Python test jobs are now passing.

malhotra5:
@openhands can you examine the goal of this PR carefully, it seems like we're handling much of the streaming logic

Is it possible to leverage FastAPI streaming response for this instead to simplify this?

neubig:
On the streaming question: the router is already using FastAPI `StreamingResponse` for the download path. The remaining custom streaming code is the bridge between Python `zipfile.ZipFile` and ASGI chunks: `zipfile` writes synchronously to a file-like object, so `_StreamingZipBuffer` lets us emit those writes incrementally without materializing the whole archive in memory.  

Using `StreamingResponse` alone would still need an iterator that produces zip bytes. We could replace that bridge with an async zip library later, but for this hotfix I kept the dependency surface small and preserved the existing `export_conversation()` bytes API by wrapping the same streaming implementation.

openhands/app_server/utils/redis_lock.py
```diff
@@ -0,0 +1,99 @@
...
+end
+return 0
+"""
+
+
+class RedisLockUnavailable(Exception):
+    """Raised when Redis cannot be used to evaluate a lock."""
+
+
+def _get_redis_client_async():
```
malhotra5: Hmm I'm not super happy about this, maybe we move the file `enterprise/storage/redis.py` here instead to dedup and change `enterprise/storage/redis.py` to re-export the same values (to stay backwards compatible and avoid making too many changes)

malhotra5:
> @openhands can you examine the goal of this PR carefully, it seems like we're handling much of the streaming logic. Is it possible to leverage FastAPI streaming response for this instead to simplify this?

Great question — the PR already uses FastAPI's `StreamingResponse` in the router (the endpoint returns `StreamingResponse(zip_stream, media_type="application/zip", ...)`). Let me clarify the three layers and what we simplified:

**Why `_StreamingZipBuffer` is still needed**

Python's `zipfile.ZipFile` writes synchronously to a file-like object, but we need to yield bytes chunks asynchronously to `StreamingResponse`. `_StreamingZipBuffer` is a small (~30 line) adapter that captures those sync writes and lets the async generator drain them after each `writestr()`. Without it we'd have to buffer the entire zip in memory first — which is exactly the O(N·K) memory problem this PR sets out to fix.

**What we simplified in this commit**

The original `stream()` closure mixed two concerns: chunk generation and lock refresh (via `time.monotonic()` tracking on every iteration). We refactored this to use a background `asyncio.create_task` instead:

```python
# before — time tracking on every chunk
async def stream():
    last_refresh = time.monotonic()
    async for chunk in self._stream_conversation_zip(...):
        now = time.monotonic()
        if export_lock and now - last_refresh >= refresh_interval:
            await export_lock.refresh()
            last_refresh = now
        yield chunk

# after — background task handles refresh, loop only yields chunks
async def stream():
    refresh_task = asyncio.create_task(
        _refresh_lock_periodically(export_lock, refresh_interval, ...)
    ) if export_lock else None
    try:
        async for chunk in self._stream_conversation_zip(...):
            yield chunk
    finally:
        if refresh_task: refresh_task.cancel()
        if export_lock:  await export_lock.release()
```

This removes `import time`, the `last_refresh` variable, and the inline refresh block. `_refresh_lock_periodically` is now a standalone module-level coroutine.

**Why FastAPI's `StreamingResponse` alone can't handle the lock lifecycle**

- Lock must be *acquired before the first byte* is sent (so 409/503 errors can still be returned as HTTP status codes).
- Lock must be *refreshed during* streaming — `BackgroundTask` only runs *after* the response finishes.
- Lock must be *released in the generator's `finally`* to cover error paths mid-stream.

All existing tests pass (`test_open_conversation_export_streams_with_lock`, `test_open_conversation_export_rejects_duplicate_export`, etc.). See commit 6730911a6 for the full diff.

_This response was generated by an AI agent (OpenHands) on behalf of the PR author._

openhands/app_server/utils/redis_lock.py
neubig: Addressed in commit 368ff1aa3. The Redis client helper now lives in `openhands.app_server.utils.redis`, `redis_lock.py` imports that shared helper directly, and `enterprise/storage/redis.py` is a backwards-compatible re-export for existing enterprise imports. I reran core pre-commit, enterprise pre-commit, and the focused export/router tests after rebasing on the latest PR head.

malhotra5: Implemented in commit 368ff1aa3 ("Share Redis client helper"). `openhands/app_server/utils/redis.py` now holds the full Redis client implementation; `enterprise/storage/redis.py` is reduced to a re-export wrapper so all existing `from storage.redis import ...` calls continue to work unchanged. `redis_lock.py` now imports `get_redis_client_async` directly from the core module, removing the `importlib` dynamic-import fallback and the duplicate singleton. All 188 affected tests pass.

_This response was generated by an AI agent (OpenHands) on behalf of the PR author._

malhotra5:
## PR status summary

All feedback from the review has now been addressed. Here is what the branch contains across the 4 refactoring commits on top of the original fix:

| Commit | Change |
|--------|--------|
| `6730911a6` | Replace inline `time.monotonic` lock-refresh with `asyncio.create_task(_refresh_lock_periodically(...))` so the streaming loop only yields zip bytes |
| `8afd2e9cb` | Move `_refresh_lock_periodically` from `live_status_app_conversation_service.py` into `redis_lock.py` (where it belongs with other lock utilities) |
| `5ff12b255` | Remove unnecessary `__init__`, `read`, and `readinto` overrides from `_StreamingZipBuffer` |
| `368ff1aa3` | Move Redis client initialisation to `openhands/app_server/utils/redis.py`; reduce `enterprise/storage/redis.py` to a re-export wrapper (all existing `from storage.redis import ...` imports continue to work unchanged); simplify `redis_lock.py` to import `get_redis_client_async` directly |

All review threads resolved. All 188 affected tests pass; ruff checks clean.

_This message was generated by an AI agent (OpenHands) on behalf of the PR author._

openhands/app_server/utils/redis_lock.py
```diff
@@ -0,0 +1,91 @@
...
+from openhands.app_server.utils.redis import get_redis_client_async, redis_exceptions
+
+_logger = logging.getLogger(__name__)
+
+_RELEASE_SCRIPT = """
+if redis.call("get", KEYS[1]) == ARGV[1] then
+    return redis.call("del", KEYS[1])
+end
+return 0
+"""
```
malhotra5: PR overall LGTM, last thing I'm looking into is if we need these scripts

Redis client may already support direct APIs for this

malhotra5: Addressed in commit 6c3097e8a.

The scripts are doing atomic compare-then-act operations (get value → check token → del/expire), and Redis has no built-in command that does this in a single step, so the Lua is genuinely necessary for correctness — without it there is a TOCTOU race where an expired lock could be deleted after another caller has already reacquired it.

However, `redis-py` ships `redis.asyncio.lock.Lock` which contains identical scripts (`LUA_RELEASE_SCRIPT`, `LUA_REACQUIRE_SCRIPT`) baked in. So we removed both hand-rolled Lua strings and the `RedisLock` dataclass, and replaced them with the library:

- `redis.lock(key, timeout=ttl_seconds).acquire(blocking=False)` for acquisition
- `lock.release()` for token-checked deletion (uses `LUA_RELEASE_SCRIPT` internally)
- `lock.reacquire()` for TTL reset (uses `LUA_REACQUIRE_SCRIPT` internally)

`redis_lock.py` went from ~90 lines to ~50 as a result.

_This comment was posted by an AI agent (OpenHands) on behalf of the PR author._

malhotra5 (approved):
LGTM, thank you!

malhotra5:
Is the convention to create a new release branch `saas-rel-1.38.1` off `saas-rel-1.38.0`, make this fix on main and cherry pick into the release branch instead?

openhands/app_server/utils/redis.py
```diff
@@ -0,0 +1,81 @@
+import os
+import threading
+
+from redis import Redis
+from redis import asyncio as aioredis
```
malhotra5, line 5: Does this library exist in the pyproject for local GUI? I think we need to do a try/except on the import here because it may no exist locally

neubig: Hmm, shouldn't CI fail if it's not included?

malhotra5: This is not a concern, the local GUI pyproject seems to contain the redis library

PR #14897, opened 2026-06-18
fix: avoid decrypting org_member.llm_api_key when not set

saurya:
HUMAN: I am a hooman and I solemnly swear this is a small correction of behavior which will enable me to test OHE deployments more easily while not being anything untoward. Go in peace.

- [X ] A human has tested these changes.

## Summary

When `has_custom_llm_api_key` is False, the code should not try to access `org_member.llm_api_key` because it may be empty or null, and the property always attempts decryption which fails on empty values.

## Changes

1. **MeResponse.from_org_member()** (enterprise/server/routes/org_models.py): Only access `member.llm_api_key` when `has_custom_llm_api_key` is True to avoid decryption errors.

2. **SaasSettingsStore._get_effective_llm_api_key()** (enterprise/storage/saas_settings_store.py): Don't fall back to `org_member.llm_api_key` when `has_custom_llm_api_key` is False - only return the org-level key.

## Root Cause

The `org_member.llm_api_key` property always attempts to decrypt the stored value. When `_llm_api_key` is empty string or NULL, decryption fails with `ValueError: Failed to decrypt value with any known key`.

## Fix

This adds guards to prevent accessing `org_member.llm_api_key` when `has_custom_llm_api_key` is False, which fixes 500 errors on:
- `/api/v1/settings`
- `/api/v1/app-conversations/search`
- `/api/organizations/{org_id}/me`

---

This PR was created by an AI agent (OpenHands) on behalf of saurya@openhands.dev.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-95fdeea   docker.openhands.dev/openhands/openhands:95fdeea
```


saurya:
Closing this PR - re-created as #14898 which is based on main instead of the feature branch.

PR #14904, opened 2026-06-19
fix: command injection in git user config (#14903)

PRATHAMESH75:
## Summary

Fixes a command injection vulnerability reported in OpenHands/enterprise#36.

`AppConversationServiceBase._configure_git_user_settings` interpolated the
user-controlled `git_user_name` and `git_user_email` settings directly into a
shell command without escaping:

```python
cmd = f'git config --global user.name "{user_info.git_user_name}"'
```

A value such as `"; id #` breaks out of the quoted argument and runs arbitrary
commands on the host when a conversation workspace is initialized — the same
class of injection described in the issue. (The originally-referenced
`openhands/resolver/send_pull_request.py` was already removed along with the V0
resolver, but this live code path remained vulnerable.)

## Fix

Escape both values with `shlex.quote()`, consistent with how the repo URL and
branch name are already handled in `clone_or_init_git_repo`.

## Testing

- Added `test_configure_git_user_settings_quotes_shell_metacharacters`, a
  regression test using `"; touch /tmp/pwned #` and `$(touch /tmp/pwned)`
  payloads, asserting the value is wrapped in an inert single-quoted literal.
- Updated existing `_configure_git_user_settings` tests to assert the safely
  quoted command form.
- `uv run python -m pytest tests/unit/app_server/test_app_conversation_service_base.py` → 43 passed.
- `ruff check` / `ruff format` clean.


neubig:

Thank you for understanding.

PR #14906, opened 2026-06-19
fix: drop wholesale key when its update value is None

he-yufeng:
## Why

`deep_merge` documents a `None` update value as "remove the corresponding key", and it does pop the key. But `deep_merge_with_wholesale_keys` then re-applied the wholesale value unconditionally:

```python
result = deep_merge(base, updates)
for key in wholesale_keys:
    if key in updates:
        result[key] = updates[key]
```

So passing `{"mcp_config": None}` to clear it ended up putting `{"mcp_config": None}` back into the result instead of removing the key — the wholesale path silently disagreed with `deep_merge`'s own delete semantics.

## Summary

- Skip re-applying a wholesale key when its update value is `None`, so the wholesale path honours the same "None deletes" rule (`deep_merge` has already popped it).
- Add a regression test: `mcp_config: None` now removes the key rather than leaving it set to `None`.

## Testing

```
pytest tests/unit/app_server/utils/test_jsonpatch_compat.py
```

12 passed (the new case fails before the change and passes after). `ruff check` and `ruff format --check` are clean on both files.

### DIFF
--- a/openhands/app_server/utils/jsonpatch_compat.py
+++ b/openhands/app_server/utils/jsonpatch_compat.py
@@ -52,6 +52,9 @@ def deep_merge_with_wholesale_keys(
 
     result = deep_merge(base, updates)
     for key in wholesale_keys:
-        if key in updates:
+        # A None update means "remove the key" (deep_merge already popped it), so
+        # only re-apply a wholesale value when one was actually provided —
+        # otherwise we'd resurrect the key with a None value.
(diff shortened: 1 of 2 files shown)


he-yufeng:
Following up here. This drops only the specific key when its update fails, instead of discarding the whole map. Contained change with a test. Happy to rebase if it has fallen behind main.

neubig:

Thank you for understanding.

PR #14910, opened 2026-06-20
test: cover execute_bash without security risk

maxpetrusenkoagent:
## Summary
- Add regression coverage for legacy `execute_bash` tool-call alias conversion when `security_risk` is omitted.
- Verifies the call maps to the current `terminal` tool and preserves the command arguments without raising a missing-parameter error.

Fixes OpenHands/software-agent-sdk#4248

## Verification
- `OPENHANDS_SUPPRESS_BANNER=1 uv run pytest tests/unit/llm/test_fn_call_converter_security_risk.py -q`
- `PATH="$PWD/.venv/bin:$PATH" pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files tests/unit/llm/test_fn_call_converter_security_risk.py`

Note: `make install-pre-commit-hooks` initially failed under the system Python 3.9; rerunning with `.venv/bin` first got past the Python check but Poetry hit a transient virtualenv embedded-pip uninstall path error for `tree-sitter-language-pack`. The focused pytest and required Python pre-commit command both passed afterward.

### DIFF
--- a/tests/unit/llm/test_fn_call_converter_security_risk.py
+++ b/tests/unit/llm/test_fn_call_converter_security_risk.py
@@ -1,0 +1,41 @@
+from openhands.sdk.llm.mixins.fn_call_converter import (
+    convert_non_fncall_messages_to_fncall_messages,
+)
+
+
+def test_execute_bash_alias_does_not_require_security_risk():
+    tools = [
+        {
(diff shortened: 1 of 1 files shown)


neubig:

Thank you for understanding.

PR #14915, opened 2026-06-20
fix: path param mismatch in delete_sandbox breaks DELETE /sandboxes/{id}

lavkeshdwivedi:
Human: Fixes a one-line FastAPI routing bug — the DELETE /sandboxes/{id} route used `{id}` as the path segment but the handler parameter was named `sandbox_id`. FastAPI binds path parameters by name, so `sandbox_id` was treated as a required query parameter that was never provided, and every delete call returned 422. Fixed by renaming the route segment to `/{sandbox_id}`.

AGENT:

---

## Why

FastAPI path parameter binding is by name: the route segment and the function parameter must match exactly. The mismatch meant `sandbox_id` was never populated from the path and was treated as a required query parameter, so every `DELETE /api/v1/sandboxes/{id}` call returned **422 Unprocessable Entity** instead of 200/404.

## Summary

- Renamed `/{id}` to `/{sandbox_id}` in the `DELETE /api/v1/sandboxes/{id}` route definition in `sandbox_router.py` to match the function parameter name

## Issue Number

Fixes OpenHands/enterprise#39


lavkeshdwivedi:
Human: 1, AI : 0

lavkeshdwivedi:
Friendly bump - happy to address any review feedback.

lavkeshdwivedi:
Friendly bump — happy to address any review feedback.

neubig:

Thank you for understanding.

PR #14923, opened 2026-06-21
fix: align delete_sandbox path param with handler arg (#14920)

0xnotdev:
HUMAN:

AGENT:
I have identified and fixed the issue with the `delete_sandbox` endpoint routing.

---

## Why

The `DELETE /api/v1/sandboxes/{id}` route's path parameter (`{id}`) did not match its handler's argument name (`sandbox_id`). Since FastAPI binds path parameters by name, `{id}` was never bound, causing `sandbox_id` to be treated as a required query parameter. This led to a `422 Unprocessable Entity` when calling the route as documented, and a misleading OpenAPI schema.

## Summary

- Renamed the path parameter in `openhands/app_server/sandbox/sandbox_router.py` from `/{id}` to `/{sandbox_id}` for the `delete_sandbox` endpoint to match the handler argument.

## Issue Number
Fixes OpenHands/enterprise#39

## How to Test

1. Start the app server.
2. Send a DELETE request to `http://localhost:3000/api/v1/sandboxes/test-sandbox-id`.
3. Verify it returns `404 Not Found` (or `200 OK` if a sandbox with that ID exists), but it should no longer return `422 Unprocessable Entity` due to a missing query parameter.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This fix aligns the `delete_sandbox` endpoint with the other sandbox-scoped endpoints (e.g., `pause_sandbox` and `resume_sandbox`) which already correctly use `/{sandbox_id}/...`.


uperLu:
I also reproduced OpenHands/enterprise#39 locally and drafted a focused router regression test for it.

Would it be helpful if I open a small follow-up PR with tests, or would you prefer to include the regression test in this PR?

The test would verify that `DELETE /sandboxes/{sandbox_id}` works without a `sandbox_id` query parameter and passes the path id to `sandbox_service.delete_sandbox`.

neubig:

Thank you for understanding.

PR #14924, opened 2026-06-21
fix(app-server): pin FileStore text I/O to UTF-8 via a shared constant

wgu9:
HUMAN:
This PR makes FileStore text I/O consistently use UTF-8 across local, Google Cloud Storage, S3, in-memory storage, and the GCS event service, with tests covering non-ASCII round trips and non-UTF-8 platform defaults.

### What

`FileStore` backends handled text encoding inconsistently. `LocalFileStore` and `GoogleCloudFileStore` (and the GCS event service) opened files in text mode **without an `encoding`**, so contents were persisted/read using the host platform default encoding. On non-UTF-8 platforms (Windows cp1252, a POSIX/C `ascii` locale) this raises `UnicodeEncodeError` or corrupts non-ASCII content (CJK, accents, emoji). `S3FileStore` and `InMemoryFileStore` already hardcoded `'utf-8'` separately, so the contract was duplicated and inconsistent.

### Fix

Introduce a single `TEXT_ENCODING` constant in `openhands/app_server/file_store/files.py` and route every implementation (`local`, `google_cloud`, `s3`, `memory`) and the GCS event service through it, so text always round-trips as UTF-8 regardless of host platform. Binary writes are unchanged (no encoding passed).

### How verified

- Added an all-backends non-ASCII round-trip test (local, in-memory, GCS, S3).
- Added a `LocalFileStore` test that simulates a non-UTF-8 platform default; it fails before the fix with `UnicodeEncodeError` and passes after.
- Added a `GoogleCloudFileStore` test asserting text blobs are opened with explicit UTF-8.
- `uv run --group test pytest tests/unit/app_server/file_store/test_file_store.py -q` -> 36 passed
- `uv run ruff check` / `uv run ruff format --check` clean; pre-commit (ruff, ruff format, mypy, trailing-whitespace) all passed.


neubig:

Thank you for understanding.

PR #14937, opened 2026-06-23
feat(enterprise/auth): add super roles via user.role_id with permission fallback

chuckbutkus:
HUMAN:

- [x] This change has been tested by a human.

AGENT:

## Summary

Implements the current super-admin organization-management design for Enterprise.

A role can be assigned in two places:

1. **`org_member.role_id`** — existing org-scoped role assignment for a specific organization.
2. **`user.role_id`** — instance-level super-role assignment.

Super roles reuse the existing `role` table rows; no new role rows or DB migration are required. The only functional instance-level super role in this PR is **`superadmin`** (`user.role_id` points at the existing `admin` role row).

### Who can create organizations?

`POST /api/organizations` is guarded by `Permission.CREATE_ORGANIZATION`.

| User state | Can create org? | Notes |
|---|---:|---|
| Instance-level `superadmin` (`user.role_id = admin`) | ✅ | Explicit super permission |
| Instance-level `superowner` (`user.role_id = owner`) | ❌ | Not functional yet; intentionally empty |
| Instance-level `supermember` (`user.role_id = member`) | ❌ | No super permissions |
| Org-scoped `owner` in another org | ❌ | Org roles do not grant `CREATE_ORGANIZATION` |
| Org-scoped `admin` in another org | ❌ | Org roles do not grant `CREATE_ORGANIZATION` |
| Org-scoped `member` in another org | ❌ | Org roles do not grant `CREATE_ORGANIZATION` |
| `@openhands.dev` email without `superadmin` | ❌ | The active create-org route is permission-based, not email-domain-based |

The previous/env-controlled `@openhands.dev` helper (`OPEN_ORG_CREATION_ENABLED` / `get_org_creator_user_id`) still exists in the codebase, but the active `POST /api/organizations` route now uses `require_permission(Permission.CREATE_ORGANIZATION)` instead.

### Does a superadmin become a member of orgs they create?

No. `POST /api/organizations` now creates the org with `add_creator_as_owner=False`.

That means the creating superadmin:

- is **not** inserted into `org_member`
- does **not** become org `owner`
- is **not** added to the org's LiteLLM team / key material
- does **not** inherit org-scoped product/admin permissions for that org

### What can a superadmin do after creating an org?

A superadmin can provision initial users into a selected org via:

```text
POST /api/organizations/provision-user
```

The route is guarded by `Permission.PROVISION_USER`, which is explicitly granted to `superadmin` and remains available to org-scoped `owner` / `admin` roles.

Provisioning now accepts an optional `role` field:

| `role` value | Supported? | Result |
|---|---:|---|
| omitted | ✅ | Defaults to `member` |
| `member` | ✅ | Adds the new user as org member |
| `admin` | ✅ | Adds the new user as org admin |
| `owner` | ✅ | Adds the new user as org owner |

Example:

```json
{
  "email": "new-owner@example.com",
  "password": "optional-password",
  "role": "owner"
}
```

The response includes the assigned role.

### `enterprise/server/auth/authorization.py`

- Keeps org-scoped permissions separate from instance-level super permissions.
- Defines `SUPER_ROLE_PERMISSIONS` explicitly rather than deriving/inheriting from regular `ROLE_PERMISSIONS`.
- Only `RoleName.ADMIN` has instance-level permissions today:
  - `CREATE_ORGANIZATION`
  - `PROVISION_USER`
- `RoleName.OWNER` and `RoleName.MEMBER` are intentionally empty for super-role purposes.
- `require_permission(...)` checks org role first, then falls back to explicit super-role permissions.

### `enterprise/server/routes/orgs.py`

- `POST /api/organizations` now creates organizations without adding the creator as an owner/member.
- Create-org authorization is now only through `CREATE_ORGANIZATION`, currently granted only to `superadmin`.

### `enterprise/server/routes/user_provisioning.py`

- `ProvisionUserRequest` now supports `role: member | admin | owner`, defaulting to `member`.
- Provisioning uses the requested role when adding the new user to the target org.
- Response includes the assigned role.
- Documentation updated to describe superadmin provisioning behavior.

### Storage / LiteLLM changes

- `OrgService.create_org_with_owner(...)` now supports `add_creator_as_owner=False`.
- LiteLLM setup supports creating org/team state without also adding the creator to the LiteLLM team or generating a creator key.
- `OrgStore.persist_org_with_owner(...)` accepts an optional owner membership.
- `UserStore.create_default_settings(...)` / `LiteLlmManager.create_entries(...)` support skipping LiteLLM user/team membership for the org creator path.

## Not included / out of scope

- Org deletion by superadmin is **not** included.
- Superadmin does **not** get normal org-scoped permissions such as managing secrets, LLM settings, billing, org settings, or deleting orgs.
- `superowner` is intentionally not functional yet; `superadmin` is sufficient for the current design.

## Test plan

Targeted validation run during this PR work:

- Authorization/org/provisioning focused tests: `225 passed`.
- User provisioning focused tests after owner-role support: `30 passed`.
- Broader org/storage/LiteLLM/user-store targeted suite: `481 passed, 4 skipped`.
- Enterprise pre-commit (`ruff`, `ruff-format`, `mypy`, etc.) passed.

---
This PR description was updated by an AI agent (OpenHands) on behalf of the user.

@chuckbutkus can click here to [continue refining the PR](https://app.all-hands.dev/conversations/30431885-82f7-492f-ba71-6dad46401192)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-dbb8aee   docker.openhands.dev/openhands/openhands:dbb8aee
```


enterprise/server/auth/authorization.py
```diff
@@ -311,32 +410,62 @@ async def permission_checker(
...
+        # 1. Org-scoped role check (existing behavior).
+        if user_role and has_permission(user_role, permission):
+            return user_id
+
+        # 2. Fall back to the user's "super" role. The role row is the
+        #    same as a regular role, but because it is attached via
+        #    ``user.role_id`` it grants the super-role permission set
+        #    (parallel role perms + super-only extras) across every
+        #    organization the user touches.
+        super_role = await get_user_super_role(user_id)
```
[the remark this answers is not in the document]
chuckbutkus: Fixed in f20469994a72010a25b930fa79c64d4e2330b723.

The regression was real: `require_permission` was calling `maybe_resolve_effective_org_id` -> `SaasUserAuth.get_effective_org_id`, whose `X-Org-Id` path performs an `OrgMemberStore.get_org_member` lookup and 403s non-members **before** the super-role fallback ever ran.

**Fix:**
* New `SaasUserAuth.get_target_org_id_for_permission_check()` resolves the target org id with the same precedence (override -> api_key_org_id -> X-Org-Id header -> current_org_id) but **without** the org_member check on the header path. UUID parsing, API-key/header conflict, and the trusted server-side override membership check are all preserved.
* New `resolve_target_org_id_for_permission_check()` helper in `org_context.py` for the `require_permission` callsite.
* `require_permission` switched to the new resolver. Membership now flows through the existing `get_user_org_role` + super-role fallback, so members keep prior behavior, non-members with a granting super role get access, and non-members with no path still 403.

This means `/orgs/llm`, `/orgs/app`, and the Bitbucket DC integration endpoints now correctly honor super roles when an `X-Org-Id` header targets a non-member org.

_Reply posted by an AI agent (OpenHands) on behalf of the user._

enterprise/tests/unit/test_authorization.py
```diff
@@ -1243,3 +1265,377 @@ async def test_denies_access_when_api_key_org_mismatch(self):
...
+            ),
+        ):
+            permission_checker = require_permission(Permission.DELETE_ORGANIZATION)
+            result = await permission_checker(
+                request=mock_request, org_id=org_id, user_id=user_id
+            )
+            assert result == user_id
+
+    @pytest.mark.asyncio
+    async def test_super_role_grants_when_user_not_org_member(self):
```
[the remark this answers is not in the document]
chuckbutkus: Fixed in f20469994a72010a25b930fa79c64d4e2330b723.

Added two new tests in `TestRequirePermissionSuperRoleFallback` that exercise the `org_id=None` implicit-org path:

* `test_implicit_org_route_super_role_grants_for_non_member` -- the new `resolve_target_org_id_for_permission_check` resolver is invoked, `get_user_org_role` returns `None` (non-member), and the superadmin role grants access.
* `test_implicit_org_route_non_member_without_super_role_returns_403` -- the same path but with no super role still 403s with `not a member`.

Also added six new `SaasUserAuth` tests in `tests/unit/server/auth/test_saas_user_auth_effective_org.py::TestGetTargetOrgIdForPermissionCheck` covering the bypass on header, current_org fallback, malformed-header 400, API-key/header conflict 403, API-key pinning, and the trusted server-side override (which intentionally still requires membership).

The implicit-org tests would indeed fail against the original implementation -- they fail on `get_effective_org_id`’s membership check before reaching the super-role lookup.

_Reply posted by an AI agent (OpenHands) on behalf of the user._

chuckbutkus:
Thanks for the re-review. Addressed in 7f26d29e598cb995af34a34504b9524b8a6d3cf0, with notes on the items not implemented.

### 🟠 Dead code (`maybe_resolve_effective_org_id` / `MAYBE_EFFECTIVE_ORG_ID`)
Removed. Verified zero callers across `enterprise/`, `openhands/`, and `tests/` before deleting.

### 🟡 Resolver duplication
Consolidated. The ~50 lines of shared precedence logic (override → API-key/header conflict → X-Org-Id parse → `user.current_org_id`) now live in a single `SaasUserAuth._resolve_org_id(verify_membership: bool)` helper:
- `get_effective_org_id()` calls it with `verify_membership=True` and keeps the per-request cache (so existing callers see no behavioral change).
- `get_target_org_id_for_permission_check()` calls it with `verify_membership=False` and is intentionally uncached.

Both public methods are now ~6 lines each and the membership-checking vs non-checking paths can no longer drift apart. Diff: `+56 / -108`.

### 🟡 Performance suggestion (decline, with reasoning)
Two clarifications first:
1. `get_user_org_role` does **not** call `RoleStore.get_role_by_id` for non-members — it returns at the `OrgMemberStore.get_org_member` step. So the actual extra cost on the fallback path is **1 extra `UserStore.get_user_by_id`** (always), and **1 extra `RoleStore.get_role_by_id`** (only when a super role exists). Both are O(1) PK lookups.
2. This fallback path only runs for non-members on implicit-org routes — it is not on the hot path.

The reviewer's suggested fix (fold `get_user_super_role` into the resolver) would conflate two concerns (org-id resolution vs role lookup). A proper optimization is a request-scoped `User` cache on `SaasUserAuth` — that is a separable change with its own design surface (cache invalidation, interaction with `_clear_org_scoped_caches`, etc.) and worth a dedicated PR if profiling shows it matters. Leaving as-is here.

### Evidence section (decline, with reasoning)
The repo's PR template has `HUMAN:` / `AGENT:` sections; there is no "Evidence" section requirement, and the `HUMAN:` block is the explicit channel for human-verified end-to-end testing. Standing up a real enterprise server (Keycloak + Postgres + Redis + LiteLLM + frontend) for what is fundamentally a unit-level auth helper change is disproportionate. This PR is covered at the correct test boundaries:
- `tests/unit/server/auth/test_saas_user_auth_effective_org.py::TestGetTargetOrgIdForPermissionCheck` — exercises the new resolver against a real in-memory SQLite DB with `User`, `Org`, `OrgMember`, `Role` rows.
- `tests/unit/test_authorization.py::TestRequirePermissionSuperRoleFallback::test_implicit_org_route_super_role_grants_for_non_member` — exercises the `require_permission` orchestration on the implicit-org + non-member super-role path that would have failed under the previous implementation.
- Full enterprise unit suite: **2391 passed, 6 skipped**. Pre-commit clean.

End-to-end verification against a deployed stack is the human reviewer's domain via the `HUMAN:` block.

### Note on the "good things" attribution
The review mentions a `MockOrganizationService` in `tests/unit/server/routes/conftest.py`. That file actually only contains an autouse `_default_no_super_role` fixture; no `MockOrganizationService` exists in this PR. No action needed, but flagging in case it points to a stale context in the reviewer's comparison set.

---
_Reply posted by an AI agent (OpenHands) on behalf of the user._

enterprise/migrations/versions/125_add_empty_role.py
alpha819beta, line 1: 👌

malhotra5:
Reviving this PR as there are cases where we need a orgs/instance admin (i.e super admin) and restrict org creation to only this user

jpshackelford:
This is well-documented code — really nice work. 🎯

The design reads clearly: making `CREATE_ORGANIZATION` an explicit `Permission`, and keeping `SUPER_ROLE_PERMISSIONS` deliberately separate from `ROLE_PERMISSIONS` (explicit, non-inheriting), makes the super-role model easy to reason about. The comments carry their weight too — the `RoleName` docstring, the `get_user_super_role` helper, and the first-user block in `UserStore.create_user` all clearly explain that `user.role_id` is an instance-level **super** role (e.g. the `admin` row == `superadmin`) and is distinct from the org-scoped `org_member.role_id`. The juxtaposition in `create_user` (the super `role_id` sitting right next to the `# owner of your own org.` `org_member.role_id`) is especially helpful.

**One tiny, non-blocking follow-up**: the schema definition itself isn't annotated. In `enterprise/storage/user.py` the column is bare:

```python
role_id: Mapped[int | None] = mapped_column(ForeignKey('role.id'), nullable=True)
...
role: Mapped['Role | None'] = relationship('Role', back_populates='users')
```

A developer who starts from the ORM model (a natural entry point) sees a generic FK to `role.id` with no hint that pointing it at the `admin` row means *superadmin* rather than *org admin*, and no pointer to `SUPER_ROLE_PERMISSIONS`. Because this PR repurposes the existing column rather than adding it, that file isn't in the diff — so I'm raising it here in the headline too, since there's no perfect in-diff anchor.

Would you consider a short comment on the `User.role_id` column (and the `role` relationship) to close the loop, e.g.:

```python
# Instance-level "super role" (e.g. user.role_id -> admin row == superadmin).
# NOT the user's org-membership role, which lives on org_member.role_id.
# Effective permissions come from SUPER_ROLE_PERMISSIONS, not ROLE_PERMISSIONS.
role_id: Mapped[int | None] = mapped_column(ForeignKey('role.id'), nullable=True)
```

nit only — overall the documentation is very clear. 👍

enterprise/server/auth/authorization.py
```diff
@@ -180,6 +206,24 @@ class RoleName(str, Enum):
 }
 
 
+# Instance-level permissions for super roles. These are intentionally
+# explicit and do not inherit the parallel org-scoped role permissions.
+# This keeps instance administration separate from organization-scoped
+# product/configuration administration.
+SUPER_ROLE_PERMISSIONS: dict[RoleName, frozenset[Permission]] = {
```
jpshackelford, line 220: 👍 Making `SUPER_ROLE_PERMISSIONS` an explicit, hand-written set that does **not** inherit `ROLE_PERMISSIONS` is exactly right. A `superadmin` can create orgs and provision users, but deliberately does *not* get org-management powers (secrets, billing, LLM settings, org deletion, etc.).

This is the least-privilege / separation-of-duties posture enterprises consistently ask for: they want an elevated role that can perform a powerful but **narrowly scoped** job, without that elevation silently bleeding into full control of every org's data and configuration. 

Nice work keeping instance-level administration cleanly decoupled from organization-scoped administration.

enterprise/storage/user_store.py
```diff
@@ -70,6 +70,25 @@ async def create_user(
...
+                    select(func.count()).select_from(User)
+                )
+                if existing_user_count == 0:
+                    superadmin_role = await RoleStore.get_role_by_name('admin', session)
+                    if superadmin_role is not None:
+                        role_id = superadmin_role.id
+                        logger.info(
+                            'user_store:create_user:first_user_designated_superadmin',
+                            extra={'user_id': user_id},
+                        )
```
jpshackelford, line 101: This is exactly the seeding approach I was hoping for! 🎉 The installer can now bootstrap the very first user as a `superadmin` automatically, and from there the whole org setup can be driven through the API — create orgs via `CREATE_ORGANIZATION` and seed their initial users (owners/admins/members) via `PROVISION_USER` — without anyone having to hand-edit the database. That's a clean, scriptable installer story.

**Nit (non-blocking): nesting depth.** The actual assignment sits four blocks deep (`async with` → `if role_id is None` → `if existing_user_count == 0` → `if superadmin_role is not None`). The innermost guard is also slightly inconsistent with the `owner` lookup a few lines below, which *raises* `ValueError` if the role is missing rather than silently skipping. Since `admin` is a seeded role that should always exist, consider flattening and failing fast for parity:

```python
if role_id is None and (
    await session.scalar(select(func.count()).select_from(User)) == 0
):
    if superadmin_role is None:
        raise ValueError('Admin role not found in database')
    role_id = superadmin_role.id
    logger.info(
        'user_store:create_user:first_user_designated_superadmin',
        extra={'user_id': user_id},
    )
```

That drops it to ~2 nested blocks and makes the "admin role must exist" expectation explicit and consistent with the `owner` handling right below.

_This review comment was posted by an AI agent (OpenHands) on behalf of @jpshackelford._

jpshackelford (approved):

PR #14938, opened 2026-06-23
fix: proxy V1 git requests through app server

wgu9:
HUMAN:
This routes V1 git change/diff requests through the app server when a conversation id is available, so proxied deployments do not try to call runtime ports directly.

AGENT:

---

## Why

The V1 web UI can receive conversation URLs that include a dynamically mapped runtime port. `V1GitService` used that URL to call `/api/git/changes` and `/api/git/diff` directly, which breaks behind reverse proxies that only expose the app server origin.

## Summary

- Route V1 git changes and diff requests through the existing same-origin app-server proxy when the conversation URL contains a conversation id.
- Keep the existing direct-runtime fallback for URLs that do not expose `/api/conversations/{id}`.
- Add coverage for reverse-proxy style conversation URLs and keep path parameters out of the request URL.

## Issue Number

Part of OpenHands/enterprise#37. Related to OpenHands/OpenHands#13504.

## How to Test

- `cd frontend && npm test -- --run __tests__/api/v1-git-service.test.ts`

## Video/Screenshots

N/A. This is a request-routing fix covered by unit tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This uses the existing app-server git proxy endpoints instead of introducing a second frontend-side proxy path. If a conversation URL does not include a conversation id, the service preserves the previous direct runtime behavior.

### DIFF
--- a/frontend/__tests__/api/v1-git-service.test.ts
+++ b/frontend/__tests__/api/v1-git-service.test.ts
@@ -10,6 +10,19 @@ describe("V1GitService", () => {
   });
 
   describe("getGitChanges", () => {
+    test("routes git changes through the app-server proxy", async () => {
+      vi.mocked(axios.get).mockResolvedValue({ data: [] });
+
+      await V1GitService.getGitChanges(
+        "https://openhands.example.com:42031/api/conversations/conv-123",
+        "test-api-key",
+        "/workspace/project",
+      );
+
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #14940, opened 2026-06-23
fix(app-server): recover conversation metadata save race

wgu9:
Fixes OpenHands/enterprise#26.

HUMAN:
This PR makes app-server conversation metadata saves recover from a startup/webhook duplicate-key race by rolling back the losing transaction and re-merging the same conversation info. I tested the targeted SQL metadata service path and nearby webhook/router paths locally.

## Summary
- recover `conversation_metadata` saves when a concurrent startup/webhook writer wins the insert race
- roll back the failed transaction and re-merge the same `AppConversationInfo` instead of surfacing the duplicate-key failure
- add regression coverage for the integrity-race retry path

## Validation
- `uv run --python 3.12 pytest tests/unit/app_server/test_sql_app_conversation_info_service.py -q`
- `uv run --python 3.12 pytest tests/unit/app_server/test_webhook_router_acp.py tests/unit/app_server/test_app_conversation_router.py -q`
- `uv run --python 3.12 ruff check openhands/app_server/app_conversation/sql_app_conversation_info_service.py tests/unit/app_server/test_sql_app_conversation_info_service.py`
- `uv run --python 3.12 ruff format --check openhands/app_server/app_conversation/sql_app_conversation_info_service.py tests/unit/app_server/test_sql_app_conversation_info_service.py`
- `uv run --python 3.12 pre-commit run --config ./dev_config/python/.pre-commit-config.yaml`
- `git diff --check`

## Changelog
App-server conversation metadata saves now recover from startup/webhook duplicate-key races instead of hiding otherwise running conversations.

### DIFF
--- a/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
@@ -34,6 +34,7 @@ from sqlalchemy import (
     select,
 )
 from sqlalchemy.engine import CursorResult
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import Mapped, mapped_column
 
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #14954, opened 2026-06-24
fix: bind delete sandbox path id parameter

lin-hongkuan:
## Summary

Bind the delete-sandbox path parameter to the route parameter name used by the API endpoint.

## Motivation

The delete sandbox endpoint defines the sandbox identifier in the path, but the handler parameter binding can fail if the path parameter name and function parameter are not aligned. This can make the delete endpoint reject otherwise valid requests before reaching the handler logic.

## Changes

- Update the delete sandbox endpoint parameter binding to use the path parameter id.
- Keep the change scoped to the route binding behavior.

## Testing

- Ran the targeted pytest coverage for the sandbox API route.
- Ran formatting/lint checks for the touched Python files.
- Reviewed the final diff for secrets and unrelated formatting changes.

### DIFF
--- a/openhands/app_server/sandbox/sandbox_router.py
+++ b/openhands/app_server/sandbox/sandbox_router.py
@@ -107,10 +107,10 @@ async def resume_sandbox(
 
 @router.delete('/{id}', responses={404: {'description': 'Item not found'}})
 async def delete_sandbox(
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #14963, opened 2026-06-24
fix(config): ignore malformed Ollama tags responses

lin-hongkuan:
HUMAN:

AGENT:

---

## Why

When an Ollama `/api/tags` endpoint is reachable but returns a malformed payload, model discovery can raise while indexing `response.json()["models"]`. This should degrade the same way as an Ollama connection error: keep returning the built-in model list.

## Summary

- Validate the Ollama tags payload before reading model names.
- Call `raise_for_status()` so non-2xx Ollama responses are handled in the existing error path.
- Add coverage for malformed Ollama tags responses.

## Issue Number

N/A

## How to Test

`uv run pytest tests/unit/app_server/test_llm_model_service.py::TestDefaultLLMModelServiceSearchModels -q`

## Video/Screenshots

N/A

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No config or migration changes.

### DIFF
--- a/openhands/app_server/config_api/default_llm_model_service.py
+++ b/openhands/app_server/config_api/default_llm_model_service.py
@@ -169,9 +169,21 @@ class DefaultLLMModelService(LLMModelService):
             try:
                 async with httpx.AsyncClient() as client:
                     resp = await client.get(ollama_url, timeout=3)
-                    ollama_models_list = resp.json()['models']
-                extra_models.extend('ollama/' + m['name'] for m in ollama_models_list)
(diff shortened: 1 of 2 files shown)


openhands/app_server/config_api/default_llm_model_service.py
alpha819beta, line 1: What about other error types?

neubig:

Thank you for understanding.

PR #14971, opened 2026-06-24
fix: apply disabled_skills filter in conversation skills endpoint

Oxygen56:
The `GET /api/v1/app-conversations/{id}/skills` endpoint returns the full merged skill set without applying `disabled_skills`, while the agent-context path correctly filters them. This makes the "Available Skills" panel show skills the user has explicitly disabled, creating a confusing discrepancy.

This PR adds `user_context` injection and filters the returned skills by the user's `disabled_skills`, mirroring `_load_skills_and_update_agent` behaviour.

Fixes OpenHands/enterprise#74 (Bug 1)

HUMAN:
This PR fixes a UI discrepancy where the "Available Skills" panel in the conversation view showed skills the user had explicitly disabled. The root cause was that the conversation skills API endpoint returned all merged skills without checking the user's disabled_skills setting — even though the agent runtime correctly honored that setting when loading skills for execution. The fix injects the user context into the endpoint and filters out any skills listed in disabled_skills before returning the response, aligning the display with the actual agent behavior.

### DIFF
--- a/openhands/app_server/app_conversation/app_conversation_router.py
+++ b/openhands/app_server/app_conversation/app_conversation_router.py
@@ -1312,6 +1312,7 @@ async def get_conversation_skills(
     ),
     sandbox_service: SandboxService = sandbox_service_dependency,
     sandbox_spec_service: SandboxSpecService = sandbox_spec_service_dependency,
+    user_context: UserContext = user_context_dependency,
 ) -> JSONResponse:
     """Get all skills associated with the conversation.
 
@@ -1360,6 +1361,12 @@ async def get_conversation_skills(
             f'Loaded {len(all_skills)} skills for conversation {conversation_id}: '
             f'{[s.name for s in all_skills]}'
         )
+
+        # Filter out disabled skills (mirroring _load_skills_and_update_agent)
+        user_info = await user_context.get_user_info()
+        if user_info and user_info.disabled_skills:
+            disabled_set = set(user_info.disabled_skills)
+            all_skills = [s for s in all_skills if s.name not in disabled_set]
 
         # Transform skills to response format
         skills_response = []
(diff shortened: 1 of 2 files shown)


minh2416294:
Nice, clean fix bro. the filtering mirrors _load_skills_and_update_agent exactly, so the panel and the agent context now agree on disabled_skills.

I've verified this locally (commit 75ba4ea) on Python 3.12 by running the endpoint test file, all 8 pass. Implemented a quick local test case with user_info.disabled_skills = ['repo_skill'], and the endpoint correctly filters it and returns only the enabled skill, as desired.

Two notes about the changes you've made:

Fixes OpenHands/enterprise#74 will auto-close the issue, but this resolves only Bug 1, and not the deeper problem described in the issue.

The issue describes two separate bugs, and the PR description says "Fixes [#14837](https://github.com/OpenHands/openhands/issues/14837) (Bug 1)". Because this PR uses Fixes # notation, then on merge, the issue will be marked as closed and the author will have to re-open it and request to change the Fixes to Refs. It may be preferable to use Refs OpenHands/enterprise#74 for this PR, as it will not auto-close the issue and it will be clear to all readers that this PR resolves only part of the problem described in the issue.

The new filtering branch lacks tests.

The test changes thread the user_context mock through to the existing tests, but in every case, it uses _make_user_context_mock() with get_user_info_return=None. This means that the branch for if user_info and user_info.disabled_skills: will not be taken in any of the tests. You can verify this by commenting out the new filtering block in the code and re-running the tests locally - I've done this and all 8 tests pass. A minimal test case would create a UserInfo with disabled_skills=['repo_skill'], pass that in as get_user_info_return in _make_user_context_mock(), and assert that only the knowledge_skill is present in the response. I've tested this locally on your branch, and it passes with the fix; without the fix, it would fail.

Another note: when I've made similar changes in the past, the test for the user-context-based skill loading would check that only the enabled skills are present in the response. For example, in the test case mentioned above, it would assert that the response contains only the knowledge_skill.

neubig:

Thank you for understanding.

PR #14974, opened 2026-06-25
fix: remove duplicate Close button on the org invite links modal

ak684:
HUMAN: Verified via the local ui-verify Playwright harness (mock SaaS frontend) — the invite-links result view goes from 2 Close buttons to 1. Not yet run on a live instance.

## What
The org invite **links** result view (shown after sending invites when email delivery isn't configured) rendered **two "Close" buttons**: the primary action passed `BUTTON$CLOSE`, while `ModalButtonGroup` also rendered its default "Close" secondary. This adds an opt-in `hideSecondaryButton` to `ModalButtonGroup` (threaded through `OrgModal`) and sets it on the invite-links view, leaving a single Close button. The normal invite view ("Add" + "Close") is unchanged.

## Verification — ui-verify (Playwright, mock SaaS)
A Playwright walkthrough drives `/settings/org-members` -> Invite -> enter emails -> submit -> the invite-links result view, asserting the **Close-button count goes 2 -> 1**.

- [Click-through walkthrough (GIF)](https://github.com/OpenHands/OpenHands/blob/ui-verify-artifacts/pr-14974/walkthrough.gif)
- [Before - 2 Close buttons](https://github.com/OpenHands/OpenHands/blob/ui-verify-artifacts/pr-14974/before-2-close-buttons.png)
- [After - 1 Close button](https://github.com/OpenHands/OpenHands/blob/ui-verify-artifacts/pr-14974/after-1-close-button.png)
- [summary.json](https://github.com/OpenHands/OpenHands/blob/ui-verify-artifacts/pr-14974/summary.json)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-ca4b210   docker.openhands.dev/openhands/openhands:ca4b210
```

### DIFF
--- a/frontend/src/components/features/org/invite-organization-member-modal.tsx
+++ b/frontend/src/components/features/org/invite-organization-member-modal.tsx
@@ -86,6 +86,7 @@ export function InviteOrganizationMemberModal({
         primaryButtonText={t(I18nKey.BUTTON$CLOSE)}
         onPrimaryClick={() => onClose()}
         onClose={onClose}
+        hideSecondaryButton
       >
         <div className="flex flex-col gap-2" data-testid="invite-links-list">
           {/* With email configured the modal only stays open to show failures;
--- a/frontend/src/components/shared/modals/modal-button-group.tsx
+++ b/frontend/src/components/shared/modals/modal-button-group.tsx
@@ -15,6 +15,9 @@ interface ModalButtonGroupProps {
   primaryTestId?: string;
   secondaryTestId?: string;
   fullWidth?: boolean;
+  // For single-action modals where the primary button already closes; avoids a
+  // redundant second "Close" button next to it.
+  hideSecondaryButton?: boolean;
 }
 
 export function ModalButtonGroup({
@@ -27,6 +30,7 @@ export function ModalButtonGroup({
   primaryTestId,
   secondaryTestId,
   fullWidth = false,
+  hideSecondaryButton = false,
 }: ModalButtonGroupProps) {
   const { t } = useTranslation();
   const closeText = secondaryText ?? t(I18nKey.BUTTON$CLOSE);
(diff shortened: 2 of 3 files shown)


ak684:
I reviewed the all-hands-bot review and **agree with its one blocking finding**, so I've pushed a fix.

### What I agreed with and fixed
The bot was correct: the branch had accidentally committed `frontend/node_modules` as a **symlink** (git mode `120000`) pointing to a machine-local absolute path (`/Users/alonaking/.../OpenHands/frontend/node_modules`). That target doesn't exist on CI or anyone else's clone, so it had to go.

- Removed the symlink from tracking and **amended** the single commit (`ad44e5f8a` → `ca4b2105d`) so `node_modules` never appears in the PR history — the diff is now just the 3 intended source files (`invite-organization-member-modal.tsx`, `modal-button-group.tsx`, `org-modal.tsx`). Force-pushed with `--force-with-lease`.
- Used `git rm --cached`, so the on-disk symlink stays put locally (it's a per-worktree dev convenience) but is no longer tracked.

### What I did *not* do, and why
- **I did not edit the committed `.gitignore`.** The bot suggested adding `frontend/node_modules` to it, but that's already covered for the normal case — `frontend/.gitignore:5` has `node_modules/` and the root `.gitignore` has `**/node_modules/`. The reason the symlink slipped through is subtle: those patterns have a **trailing slash**, so they match the real `node_modules` *directory* but not a *symlink* of the same name. Rather than add a symlink-specific entry to the shared repo's ignore file (it only matters for a local worktree-symlink workflow, not for other contributors), I added `frontend/node_modules` to this worktree's local `.git/info/exclude` so it can't be re-added from my side. The shared repo's standard ignore behavior is unchanged.

### Functional change — no objection
Agreeing with the bot here too: the actual fix is clean. `hideSecondaryButton` is opt-in and defaults to `false`, so every existing `OrgModal`/`ModalButtonGroup` caller is unaffected; only the invite-links result view (where the primary button is already "Close") suppresses the redundant secondary. No code change needed there.

PR #14976, opened 2026-06-25
fix: allow max_num_sandboxes=1 by accepting 0 in pause_old_sandboxes

tomsen02:
HUMAN: When `max_num_sandboxes` is set to 1, starting a conversation always crashes. I traced it: the sandbox start path calls `pause_old_sandboxes(max_num_sandboxes - 1)`, so a limit of 1 passes 0, and the guard `if max_num_sandboxes <= 0: raise` rejected 0 — even though 0 is a valid "keep no other sandboxes" request that the method body already handles correctly. This breaks the host-network setup the docs explicitly recommend (`max_num_sandboxes=1`). I fixed the guard, added a config lower-bound, and updated the tests. I verified by constructing a real DockerSandboxService with max_num_sandboxes=1 and confirming start_sandbox() no longer raises.

## Problem

Starting or resuming a sandbox **crashes whenever `max_num_sandboxes=1`**.

Callers reserve a slot for the new sandbox by invoking:

```python
await self.pause_old_sandboxes(self.max_num_sandboxes - 1)
```

So a limit of `1` passes `0`. But the guard rejected `0`:

```python
if max_num_sandboxes <= 0:
    raise ValueError('max_num_sandboxes must be greater than 0')
```

`0` is a **legitimate** value meaning "keep no other sandboxes" — the method body handles it correctly (`num_to_pause = len(running) - 0` pauses all). The guard kills it before that code runs, so **every** `start_sandbox` / `resume_sandbox` fails.

This matters in practice: the Docker service explicitly **recommends `max_num_sandboxes=1` for host network mode** (`docker_sandbox_service.py`). An operator following that advice breaks every conversation start.

Only `max_num_sandboxes=1` triggers it — other values pass `>= 1` to the guard. Verified by constructing a real `DockerSandboxService(max_num_sandboxes=1)` and calling `start_sandbox()`, which raised `ValueError: max_num_sandboxes must be greater than 0` before this fix.

## Fix

- Relax the guard in `SandboxService` and `RemoteSandboxService` from `<= 0` to `< 0`, and update the message to "greater than or equal to 0".
- Add a `ge=1` constraint to the `max_num_sandboxes` config `Field` in the Docker and remote services, so misconfigured `0`/negative limits are rejected at the source rather than crashing at runtime.
- Update `tests/unit/app_server/test_sandbox_service.py`: `0` must pause all running sandboxes; only negative values raise.

The `max_num_sandboxes - 1` reserve-a-slot logic is correct — the root cause is the over-strict guard plus the missing config lower-bound.

## Testing

```
poetry run pytest tests/unit/app_server/test_sandbox_service.py \
    tests/unit/app_server/test_docker_sandbox_service.py \
    tests/unit/app_server/test_remote_sandbox_service.py
```
All pass (132). `ruff check` clean on changed files.

### DIFF
--- a/openhands/app_server/sandbox/docker_sandbox_service.py
+++ b/openhands/app_server/sandbox/docker_sandbox_service.py
@@ -598,6 +598,7 @@ class DockerSandboxServiceInjector(SandboxServiceInjector):
     container_name_prefix: str = 'oh-agent-server-'
     max_num_sandboxes: int = Field(
         default=5,
+        ge=1,
         description='Maximum number of sandboxes allowed to run simultaneously',
     )
     mounts: list[VolumeMount] = Field(default_factory=list)
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -601,8 +601,8 @@ class RemoteSandboxService(SandboxService):
         Uses _get_user_running_sandboxes (runtime /list + DB cross-reference) so
         only sandboxes that are actually running are considered.
         """
-        if max_num_sandboxes <= 0:
-            raise ValueError('max_num_sandboxes must be greater than 0')
+        if max_num_sandboxes < 0:
+            raise ValueError('max_num_sandboxes must be greater than or equal to 0')
 
         running = await self._get_user_running_sandboxes()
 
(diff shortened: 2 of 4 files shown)


tomsen02:
Confirming this PR is ready for review. The PR description already includes a HUMAN: section at the top. Summary: `max_num_sandboxes=1` crashes every sandbox start because callers pass `max-1=0` and the guard rejected 0 (a valid 'keep none' value). Fixed the guard (`<=0`→`<0`), added a `ge=1` config constraint, and updated tests. Verified with a real DockerSandboxService(max_num_sandboxes=1).

tomsen02:
Closing to tidy my open-PR queue — happy to reopen if this is still wanted. Thanks!

PR #14977, opened 2026-06-25
feat(app_server): inject self-hosted provider host (BITBUCKET_DATA_CENTER_HOST) into agent runtime

jpshackelford:
## Problem

For **Bitbucket Data Center** (and other self-hosted providers) there is no fixed public domain like `bitbucket.org`. An agent needs the configured host to call the REST API or build a git remote URL. Today the host (`BITBUCKET_DATA_CENTER_HOST`) is a **server-side** variable only — it is read in `enterprise/server/auth/constants.py`, used to populate `ProviderToken.host`, and consumed server-side (authenticated git URL, `create_bitbucket_data_center_pr` MCP tool). It is **never injected into the agent sandbox**.

Provider credentials reach the sandbox solely through `AuthUserContext.get_provider_tokens(as_env_vars=True)`, surfaced via the `sandbox_router` secret-lookup endpoints. That method exported only the **token** (`bitbucket_data_center_token`), so the agent could authenticate but had no way to learn the host — **especially before a repository is cloned, or in conversations with no repository at all** (where deriving the host from the cloned repo's git remote isn't an option).

## Change

- **`app_server/integrations/provider.py`** — add `ProviderHandler.get_provider_host_env_key(provider)` (`{provider}_host`), mirroring `get_provider_env_key`.
- **`app_server/user/auth_user_context.py`** — in `get_provider_tokens(as_env_vars=True)`, also export the provider host as `{provider}_host` (e.g. `bitbucket_data_center_host`) whenever `ProviderToken.host` is set. Same repo-independent secret-lookup path, so it is available **before or without** cloning a repo. **Azure DevOps is excluded** because its `host` field may contain an org/project path rather than a bare host (consistent with `get_authenticated_git_url`). Also future-proofs self-managed GitLab.
- **`skills/bitbucket_data_center.md`** — document that `BITBUCKET_DATA_CENTER_HOST` is injected by OpenHands and resolve the bare `{domain}` placeholder to `${BITBUCKET_DATA_CENTER_HOST}`.
- **Tests** — extend `tests/unit/app_server/test_sandbox_secrets_router.py` (DC host exported; cloud token adds no host key; Azure host excluded).

## Why this mechanism

The `as_env_vars=True` dict is the **only** path by which git creds reach the sandbox (`sandbox_router.list_secret_names` / `get_secret_value`), and it does not depend on `selected_repository`. Adding the host there makes `${BITBUCKET_DATA_CENTER_HOST}` resolvable by the runtime's secret auto-injection (case-insensitive match to `bitbucket_data_center_host`) in every conversation, pre-clone included.

## Validation

Targeted logic verified in isolation (enum values, `get_provider_host_env_key` formula, export/exclusion branch). Full `pytest` needs the project's managed env (`uv sync`), not provisioned here; the new unit tests mirror the existing `as_env_vars` tests and should run in CI.

## Related

Companion: **OpenHands/extensions#361** (same guidance for the registry `bitbucket-data-center` skill).

Kept as a **draft** pending review of the secret-handling change.

---
_This PR was created by an AI agent (OpenHands) on behalf of @jpshackelford._

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-7f82126   docker.openhands.dev/openhands/openhands:7f82126
```

### DIFF
--- a/openhands/app_server/integrations/provider.py
+++ b/openhands/app_server/integrations/provider.py
@@ -408,6 +408,17 @@ class ProviderHandler:
         """Map ProviderType value to the environment variable name in the runtime"""
         return f'{provider.value}_token'.lower()
 
+    @classmethod
+    def get_provider_host_env_key(cls, provider: ProviderType) -> str:
+        """Map ProviderType value to the host environment variable name in the runtime.
+
+        Self-hosted providers (e.g. Bitbucket Data Center, self-managed GitLab)
+        have no fixed public domain, so the agent needs the configured host to
+        build REST API and git remote URLs — including before any repository is
+        cloned. This mirrors ``get_provider_env_key`` for the host value.
+        """
+        return f'{provider.value}_host'.lower()
+
     async def verify_repo_provider(
         self,
         repository: str,
--- a/openhands/app_server/user/auth_user_context.py
+++ b/openhands/app_server/user/auth_user_context.py
@@ -89,6 +89,15 @@ class AuthUserContext(UserContext):
                     token_value = provider_token.token.get_secret_value()
                     if token_value:
                         results[env_key] = token_value
+                # Expose the configured host (e.g. bitbucket_data_center_host) so
+                # the agent can build REST API and git remote URLs for self-hosted
+                # providers even before — or without — cloning a repository. Azure
+                # DevOps is excluded because its host field may hold an
+                # org/project path rather than a bare host (see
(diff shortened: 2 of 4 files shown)


neubig:

Thank you for understanding.

PR #14985, opened 2026-06-25
feat: add default_sandbox_spec_id to user settings

tofarr:
HUMAN:

Long standing update allowing to pick the sandbox spec from the UI.

AGENT:

---

## Why

Users who work with multiple sandbox specs need a way to set their preferred spec as the default, rather than always receiving the system-wide first-spec default.

## Summary

- Add `default_sandbox_spec_id: str | None = None` to `Settings` and propagate through `UserContext.get_default_sandbox_spec_id()` — all three sandbox services (`ProcessSandboxService`, `DockerSandboxService`, `RemoteSandboxService`) use the user preference as a fallback between the caller-specified spec and the global default
- Enterprise: new `default_sandbox_spec_id` column on `user` and `user_settings` tables; Alembic migration 125
- Frontend: **Default Sandbox** dropdown in App Settings (below Sandbox Grouping Strategy), populated from `/api/v1/sandbox-specs/search`, clearable to revert to the system default; full 15-language i18n coverage

## Issue Number

N/A

## Screenshots

There is a new dropdown in the application settings page - it displays available sandbox_spec_ids - for OSS there is typically one a single option in here - for cloud it is populated based on the available warm runtimes:
image
image

## How to Test

1. Open **App Settings → Default Sandbox** dropdown.
2. Select one of the available sandbox specs and save — subsequent sandbox starts without an explicit spec should use the chosen spec.
3. Clear the dropdown ("System default" placeholder) and save — verify the system-wide default spec is used again.
4. Verify that passing an explicit `sandbox_spec_id` when starting a conversation still overrides the user preference.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The `DockerSandboxService` injector carries no `UserContext` (local dev mode), so `default_sandbox_spec_id` will always be `None` in that path; the field and two-step fallback are still present for consistency.
- Enterprise migration `125` adds the column to both the `user` table (V1 path, auto-picked by `saas_settings_store.py`) and the legacy `user_settings` table.
- The frontend dropdown is always visible regardless of `app_mode`, consistent with how Sandbox Grouping Strategy is shown.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-b09c3b8   docker.openhands.dev/openhands/openhands:b09c3b8
```


hieptl (approved):
Thank you! 🙏

PR #14989, opened 2026-06-25
fix(frontend): stop polling app-conversations with task IDs on new conversation

hieptl:
HUMAN:

AGENT:

---

## Why

issue

Creating a new conversation shows a red `Invalid UUID format for ids: ['task-…']` toast. While the conversation page polls the start-task, `useUnifiedVSCodeUrl` calls `GET /api/v1/app-conversations?ids=task-{uuid}` with a **task ID** — the URL param is `task-{uuid}` until `useTaskPolling` swaps in the real conversation UUID. The backend correctly rejects non-UUID ids with HTTP 400.

`useUnifiedVSCodeUrl` (rendered unconditionally via `conversation-tabs` → `vscode-tooltip-content`) was the only conversation-data hook missing the `task-` guard that `useActiveConversation`, `useUserConversation`, and `useTaskPolling` already have.

## Summary

- `useUnifiedVSCodeUrl`: skip the app-conversations fetch for task IDs (mirrors `useActiveConversation`).
- `useBatchAppConversations`: disable the query when any id is a task id — a central guard for every caller of the endpoint.
- Add unit tests asserting the endpoint is **not** called for `task-{uuid}` ids and **is** called for valid ids.

## How to Test

**Automated**

- `cd frontend && npm run test` (or `npx vitest run __tests__/hooks/query/use-batch-app-conversations.test.tsx`) — passes.
- `cd frontend && npm run lint` — typecheck + prettier clean.

**Manual**

1. Create a new conversation → you land on `/conversations/task-{uuid}`.
2. With DevTools → Network open, confirm there is **no** `GET /api/v1/app-conversations?ids=task-…` request and **no** "Invalid UUID format" toast.
3. After `useTaskPolling` redirects to `/conversations/{uuid}`, confirm `app-conversations` is now called with the real UUID and the VS Code tab/tooltip URL still resolves.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No backend change — the 400 is correct behavior for a non-UUID id. The two guards are defense-in-depth: behaviorally identical today (a single caller), but the central guard prevents any future caller from hitting the endpoint with a task id. Disabled TanStack queries report `isLoading: false`, so the loading state and VS Code URL resolution on the happy path are unchanged.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-b6f7226   docker.openhands.dev/openhands/openhands:b6f7226
```

### DIFF
--- a/frontend/__tests__/hooks/query/use-batch-app-conversations.test.tsx
+++ b/frontend/__tests__/hooks/query/use-batch-app-conversations.test.tsx
@@ -1,0 +1,58 @@
+import { renderHook, waitFor } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import React from "react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+import V1ConversationService from "#/api/conversation-service/v1-conversation-service.api";
+import { useBatchAppConversations } from "#/hooks/query/use-batch-app-conversations";
+
+const createWrapper = () => {
+  const queryClient = new QueryClient({
+    defaultOptions: {
+      queries: { retry: false },
+    },
+  });
+
+  return ({ children }: { children: React.ReactNode }) => (
+    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
+  );
+};
+
+describe("useBatchAppConversations", () => {
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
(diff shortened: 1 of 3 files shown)


tofarr (approved):
🍰

PR #14992, opened 2026-06-25
fix(llm): restore managed member-key fallback in effective LLM key resolution

ak684:
HUMAN: Fixes a regression where managed-key (BYOK-disabled) users get no LLM key — the agent fails with `LLMAuthenticationError`, and saving any user setting wipes the stored key. Root-caused from a live install; verifying via CI + a Replicated test install before marking ready.

## The bug (regression from #14898)
#14898 changed `_get_effective_llm_api_key` to return **only** `org.llm_api_key` when `has_custom_llm_api_key` is `False`, dropping the member-key fallback:
```diff
- return org.llm_api_key or org_member.llm_api_key
+ return org.llm_api_key
```
But **managed/proxy keys are persisted on `org_member.llm_api_key` with `has_custom_llm_api_key=False`** (`saas_settings_store.py` write path), and **`org.llm_api_key` is never seeded** (no assignment path anywhere). So for managed-key users the effective key resolves to `None`, with two symptoms:
1. **Runtime:** the agent calls the LiteLLM proxy with no key → `401 No api key passed in` → `LLMAuthenticationError`.
2. **Settings save wipes the key:** the save path loads the effective key (now `None`), then the `if uses_managed_llm_key and current_member_llm_api_key is not None:` branch is skipped, so the stored key isn't re-written — matching the reported "updating any user setting removes your LLM api key."

#14898's goal was to avoid decrypting an unset member key — that's preserved.

## Scope
Any BYOK-disabled / managed-key install (everyone has `has_custom=False`). Not in `cloud-1.39.0` (current released); it's in `main` and would ship with `cloud-1.40.0`, so this is catch-before-release.

## The fix
Restore the fallback **null-safely**: `has_custom` → member key; else `org.llm_api_key` if set; else the member key **only when the raw column is set** (so an unset key is never decrypted).

## Tests
Updated `TestGetEffectiveLlmApiKey`: the prior test encoded the buggy "never touch the member key" behavior; replaced with the managed-key fallback case (org key unset + member key set → returns member key) and the null-safe case (both unset → `None`, no decrypt).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-69392cb   docker.openhands.dev/openhands/openhands:69392cb
```

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -127,10 +127,13 @@ class SaasSettingsStore(SettingsStore):
     ) -> SecretStr | None:
         if org_member.has_custom_llm_api_key:
             return org_member.llm_api_key
-        # When has_custom_llm_api_key is False, only return org-level key
-        # Don't fall back to org_member.llm_api_key as it may not be set
-        # and accessing it would trigger decryption of an empty/null value
-        return org.llm_api_key
+        if org.llm_api_key:
+            return org.llm_api_key
+        # Managed keys are stored on the member row (has_custom=False); fall back to
+        # it, but only decrypt when actually set to avoid the empty-value error (#14898).
+        if org_member._llm_api_key:
+            return org_member.llm_api_key
+        return None
 
     @staticmethod
     def _get_persisted_agent_settings(item: Settings) -> dict[str, Any]:
--- a/enterprise/tests/unit/test_saas_settings_store.py
+++ b/enterprise/tests/unit/test_saas_settings_store.py
@@ -1333,8 +1333,14 @@ class TestGetEffectiveLlmApiKey:
         # Must return org key when has_custom_llm_api_key is False
         assert result == SecretStr('org-api-key')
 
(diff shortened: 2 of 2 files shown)


ak684:
Verification:
- **CI:** Enterprise Python Unit Tests + Lint enterprise python both pass. The updated `TestGetEffectiveLlmApiKey` covers the managed-key fallback (org key unset + member key present → returns member key) and the null-safe case (both unset → `None`, no decrypt).
- **E2E on a Replicated test install** (current-main image + this fix overlaid, BYOK-disabled / managed-key user): the resolver job went from `execution_status=error` (`LLMAuthenticationError`, no key) to **`finished`**, with no `No effective LLM API key` warning, and the agent posted its answer. Reproduces the reported "updating a setting removes your LLM key" path and confirms the fix.

Not in `cloud-1.39.0`; lands with `cloud-1.40.0` — worth merging before that cut.

saurya (approved):

PR #15011, opened 2026-06-26
feat(azure-devops): post resolver replies as a service-account bot (PAT, opt-in)

ak684:
HUMAN: Phase-1 bot support for the Azure DevOps resolver. When a service-account PAT is configured, the resolver posts replies/reactions **as the bot** (not the mentioner) and skips the bot's own comments (loop guard). Fully **opt-in + backward compatible** — with no PAT it posts as the mentioner exactly as today. Needs the companion OpenHands-Cloud chart PR for the KOTS wiring before a real install can use it.

## Why
Run-as-mentioner makes replies look like the human said them and risks the bot re-triggering on its own comments (seen on Bitbucket DC). A dedicated bot identity gives clear "OpenHands" attribution and breaks the loop — mirroring the Jira DC service-account model.

## What
- `AdoBotCredential` strategy abstraction (PAT now; service-principal / managed-identity slot in later) → builds a PAT-authenticated `SaaSAzureDevOpsService` (Basic auth, auto-detected by the base service's JWT-vs-PAT header logic).
- `send_message` posts via the bot service when `AZURE_DEVOPS_BOT_TOKEN` is set, else the mentioner's token (opt-in).
- Self-author guard: skip events whose actor matches `AZURE_DEVOPS_BOT_USERNAME` (on top of the existing content-marker guard).
- **The agent's repo access stays run-as-mentioner** — the hardened, mentioner-scoped write-gate is unchanged; the bot is only the voice.

## Env (provided by the companion chart PR)
`AZURE_DEVOPS_BOT_TOKEN`, `AZURE_DEVOPS_BOT_USERNAME`.

## Tests
8 new: post-as-bot when token set, mentioner fallback when unset, self-author guard (uniqueName/displayName/id, case-insensitive), and the PAT credential builder.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-ede9400   docker.openhands.dev/openhands/openhands:ede9400
```

### DIFF
--- a/enterprise/integrations/azure_devops/azure_devops_bot_credential.py
+++ b/enterprise/integrations/azure_devops/azure_devops_bot_credential.py
@@ -1,0 +1,44 @@
+"""Bot posting-identity credentials for Azure DevOps integrations.
+
+When a bot credential is configured, the resolver posts its comments/reactions
+as the bot instead of the @-mentioning user; the agent job still runs with the
+mentioner's own token (unchanged authz). Phase 1 supports a PAT via HTTP Basic
+auth; the abstraction leaves room for service-principal / managed-identity
+credentials later. Opt-in: with no bot token we post as the mentioner.
+"""
+
+from abc import ABC, abstractmethod
+
+from integrations.azure_devops.azure_devops_service import SaaSAzureDevOpsService
+from pydantic import SecretStr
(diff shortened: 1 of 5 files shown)


neubig:

Thank you for understanding.

PR #14579, opened 2026-05-27
fix: retry conversation metadata save races

he-yufeng:
HUMAN: This fixes a real insert race between conversation startup and webhook metadata writes. If another writer inserts the row between merge's lookup and commit, this retries the same merge once after rollback; unrelated integrity errors still propagate instead of being hidden. - [ ] A human has tested these changes.  Fixes OpenHands/enterprise#26.  Conversation startup and webhook updates can both write `conversation_metadata` for the same conversation. `merge()` handles normal updates, but it can still lose the race when another writer inserts the row between merge's lookup and commit. In that case the commit raises an `IntegrityError`, leaving the resolver conversation started but metadata persistence failed.  This patch makes `save_app_conversation_info()` recover from that insert race by rolling back and retrying the same merge once. If the integrity error is unrelated, the retry will still fail and propagate instead of silently hiding it.  ## To verify  - `.venv\Scripts\python -m pytest tests\unit\app_server\test_sql_app_conversation_info_service.py -q` - `.venv\Scripts\python -m ruff check openhands\app_server\app_conversation\sql_app_conversation_info_service.py tests\unit\app_server\test_sql_app_conversation_info_service.py` - `.venv\Scripts\python -m ruff format --check openhands\app_server\app_conversation\sql_app_conversation_info_service.py tests\unit\app_server\test_sql_app_conversation_info_service.py` - `.venv\Scripts\python -m py_compile openhands\app_server\app_conversation\sql_app_conversation_info_service.py tests\unit\app_server\test_sql_app_conversation_info_service.py` - `git diff --check origin/main...HEAD`  The local pre-commit hook is still blocked by a machine cache permission issue:  `PermissionError: [Errno 13] Permission denied: 'C:\Users\He\.cache\pre-commit\...\.pre-commit-hooks.yaml'`

PR #15015, opened 2026-06-27
fix(app-server): preserve selected repository across webhook metadata races

shanemort1982:
HUMAN:

On a self-hosted box the selected repo kept vanishing from running conversations (the UI flips to "no repo connected"). I traced it to the conversation-start webhook racing the create: it reads the metadata before the create is visible, gets a blank stub, and writes the row back with the git fields nulled, wiping the repo that was just saved. This change makes selected_repository / selected_branch / git_provider write-once so a late stub write can't null them, and keeps the duplicate-key insert-race recovery from OpenHands/enterprise#26.

AGENT:

---

## Why

`selected_repository`, `selected_branch` and `git_provider` are set once, when a conversation is created. The conversation-start webhook (`on_conversation_update`) also reads the conversation metadata and writes it back. When the webhook's read lands before the create transaction is visible, `valid_conversation` returns a stub with the git fields set to `None`, and the blind `merge()` in `save_app_conversation_info` then overwrites the stored repository with `None`. The repository silently disappears from a running conversation (the UI shows "no repo connected").

This is the same race as OpenHands/enterprise#26, but a different symptom. The open retry PRs (#14579, OpenHands/OpenHands#14940) recover the duplicate-key insert crash, but re-merging the stub still writes `None` over the repository, so the data loss remains.

## Summary

- Make the three git fields write-once in `save_app_conversation_info`: existing rows are updated with `COALESCE(incoming, stored)`, so an incoming `None` never overwrites a populated value.
- When the row does not exist yet it is inserted, and a concurrent insert (duplicate primary key) is recovered by re-applying the git-preserving update.
- Order-independent: whichever writer supplies a non-null repository wins regardless of commit order, and the OpenHands/enterprise#26 insert race is recovered as well.

## Issue Number

Fixes OpenHands/enterprise#26

## How to Test

- Unit tests (`tests/unit/app_server/test_sql_app_conversation_info_service.py::TestSaveGitWriteOnce`): a stub update does not null a stored repository, the repository survives when the stub commits first, and a repo-less conversation stays `None`. Run `uv run --python 3.12 pytest tests/unit/app_server/test_sql_app_conversation_info_service.py -q`.
- `ruff check` and `ruff format --check` clean on both changed files (config auto-discovered from `pyproject.toml`, ruff 0.12.5).
- Reproduced end to end on a self-hosted instance (`ghcr.io/openhands/openhands:main`, SQLite): firing concurrent repository-backed conversation starts nulled roughly 40% of repositories before the change and 0 after, including starts where the webhook stub demonstrably fired.

## Video/Screenshots

N/A. Backend metadata-persistence change, covered by the unit tests and the before/after reproduction above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The retry-only PRs for OpenHands/enterprise#26 (#14579, OpenHands/OpenHands#14940) recover the duplicate-key crash but not this data-loss symptom. This change keeps the same insert-race recovery while also preventing the repository from being nulled, so it could supersede them.

### DIFF
--- a/enterprise/server/utils/saas_app_conversation_info_injector.py
+++ b/enterprise/server/utils/saas_app_conversation_info_injector.py
@@ -351,11 +351,16 @@ class SaasSQLAppConversationInfoService(SQLAppConversationInfoService):
         return results
 
     async def save_app_conversation_info(
-        self, info: AppConversationInfo
+        self,
+        info: AppConversationInfo,
+        *,
+        preserve_git_fields_on_null: bool = False,
     ) -> AppConversationInfo:
         """Save conversation info and create/update SAAS metadata with user_id and org_id."""
         # Save the base conversation metadata
-        await super().save_app_conversation_info(info)
+        await super().save_app_conversation_info(
+            info, preserve_git_fields_on_null=preserve_git_fields_on_null
+        )
 
         # Get current user_id for SAAS metadata
         # Fall back to info.created_by_user_id for webhook callbacks (which use ADMIN context)
--- a/openhands/app_server/app_conversation/app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/app_conversation_info_service.py
@@ -96,7 +96,10 @@ class AppConversationInfoService(ABC):
 
     @abstractmethod
     async def save_app_conversation_info(
-        self, info: AppConversationInfo
+        self,
+        info: AppConversationInfo,
+        *,
+        preserve_git_fields_on_null: bool = False,
     ) -> AppConversationInfo:
         """Store the sandboxed conversation info object given.
 
--- a/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
@@ -32,8 +32,10 @@ from sqlalchemy import (
     String,
     func,
     select,
+    update,
 )
 from sqlalchemy.engine import CursorResult
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import AsyncSession
 from sqlalchemy.orm import Mapped, mapped_column
 
@@ -383,8 +385,43 @@ class SQLAppConversationInfoService(AppConversationInfoService):
             tags=info.tags if info.tags else None,
         )
 
-        await self.db_session.merge(stored)
-        await self.db_session.commit()
+        # selected_repository / selected_branch / git_provider are set once when
+        # the conversation is created. A startup webhook racing the create only
+        # sees a not-yet-persisted stub (git fields None); it must not overwrite
+        # the stored repository with None, or it silently disappears from a
+        # running conversation. COALESCE keeps the stored value whenever the
+        # incoming write is None, independent of which writer commits last.
+        update_values: dict = {
+            column.name: getattr(stored, column.name)
+            for column in StoredConversationMetadata.__table__.columns
+            if column.name != 'conversation_id'
+        }
+        for field in ('selected_repository', 'selected_branch', 'git_provider'):
+            update_values[field] = func.coalesce(
+                getattr(stored, field),
+                getattr(StoredConversationMetadata, field),
+            )
+        update_stmt = (
+            update(StoredConversationMetadata)
+            .where(StoredConversationMetadata.conversation_id == stored.conversation_id)
+            .values(update_values)
+        )
+
+        result = await self.db_session.execute(update_stmt)
+        if result.rowcount == 0:
+            # No row yet: insert it. A concurrent writer may win the insert
+            # race (duplicate primary key); fall back to the git-preserving
+            # update instead of surfacing the IntegrityError, which would
+            # otherwise leave a started conversation without metadata.
+            try:
+                self.db_session.add(stored)
+                await self.db_session.commit()
+            except IntegrityError:
+                await self.db_session.rollback()
+                await self.db_session.execute(update_stmt)
+                await self.db_session.commit()
+        else:
+            await self.db_session.commit()
         return info
 
     async def update_conversation_statistics(
--- a/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/sql_app_conversation_info_service.py
@@ -347,7 +347,10 @@ class SQLAppConversationInfoService(AppConversationInfoService):
         return results
 
     async def save_app_conversation_info(
-        self, info: AppConversationInfo
+        self,
+        info: AppConversationInfo,
+        *,
+        preserve_git_fields_on_null: bool = False,
     ) -> AppConversationInfo:
         metrics = info.metrics or MetricsSnapshot()
         usage = metrics.accumulated_token_usage or TokenUsage()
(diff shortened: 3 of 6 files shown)


shanemort1982:
@enyst mind taking a look when you get a chance? This fixes the `conversation_metadata` race from OpenHands/enterprise#26, but specifically the repo-nulling symptom: the conversation-start webhook reads a stub before the create is visible and overwrites `selected_repository` with `None`, so the repo vanishes from a running conversation (UI shows "no repo connected").

The open retry PRs (#14579, OpenHands/OpenHands#14940) recover the duplicate-key crash but not this data loss, since re-merging the stub still writes `None`. This makes the git fields write-once with `COALESCE` and keeps the insert-race recovery as well.

Green CI with regression tests, and reproduced end to end on a self-hosted SQLite box: ~40% of repos nulled before, 0 after.

enyst (changes requested):
Hi, I’m OpenHands-GPT-5.5, reviewing this PR at the requester’s direction.

_This review was generated by an AI agent (OpenHands) on behalf of enyst._

## Short verdict

🔴 **Needs changes before merge.** The PR fixes a real race symptom, but it does so too broadly and breaks an existing API/service contract: explicitly clearing a conversation repository via PATCH/save semantics no longer works, even though I do not see a current frontend UI button that calls that path.

If I were submitting the review on GitHub, I’d request changes.

## Main blocking issue

### `save_app_conversation_info()` now prevents intentional repository removal through the documented API/service path

In `openhands/app_server/app_conversation/sql_app_conversation_info_service.py`, the PR changes updates for:

- `selected_repository`
- `selected_branch`
- `git_provider`

to:

```python
func.coalesce(incoming_value, stored_value)
```

at roughly lines 399-403.

That correctly prevents a webhook stub with `None` git fields from wiping an existing repo. But it also means any intentional `None` update can never clear those fields.

I do **not** see a current exposed frontend UI path that sets the repo back to `null` after it was set; the only real frontend caller I found passes a selected repo/branch/provider from the repository modal. So this is not necessarily a visible button regression today.

But it conflicts with the existing API/client/backend contract:

- `AppConversationUpdateRequest` allows `selected_repository: str | None`
- `LiveStatusAppConversationService.update_app_conversation()` explicitly documents that fields can be set to `None` to clear them:
  - `openhands/app_server/app_conversation/live_status_app_conversation_service.py:2153-2155`
- `_validate_repository_update()` explicitly handles repository removal:
  - lines 2126-2138: repository set to null, branch/provider must also be cleared
- frontend API docs also say `repository: null` removes the repo:
  - `frontend/src/api/conversation-service/v1-conversation-service.api.ts`

I reproduced it with a direct service script: save repo metadata, then save the same conversation with all three git fields set to `None`. The retrieved row still had:

```text
OpenHands/OpenHands main ProviderType.GITHUB
```

So the PR introduces a real contract/API regression even if the current frontend does not expose a remove-repo control.

### Suggested fix direction

Don’t make `save_app_conversation_info()` globally null-preserving for these fields unless all callers agree that clearing is impossible.

Better options:

1. Preserve git fields only in the webhook/startup race path, where the stub is known to be accidental.
2. Add an explicit save mode/parameter, e.g. “preserve git fields when incoming is null”, and use it only from webhook metadata persistence.
3. Add a dedicated partial-update path for webhook metadata instead of changing the semantics of the general save method.

Also add a regression test for PATCH/update clearing:

- create conversation with repo/branch/provider
- call update with all three set to `None`
- assert persisted fields are actually `None`

## Linked issue / PR context

### Issue #14476

The issue is about a startup/webhook metadata race causing started conversations to disappear from lists. PR #15015 addresses an important related symptom: webhook stubs nulling repository fields, especially in self-hosted/SQLite-style reproduction.

But “Fixes #14476” may be a little broad. The issue also discusses SaaS metadata/user/org consistency. This PR does not edit the enterprise SaaS metadata layer directly, though enterprise inherits the base service behavior.

### Related PRs

Issue timeline links these PRs:

- #14477 — closed; more comprehensive, included enterprise SaaS metadata handling and duplicate-key recovery.
- #14489 — closed; retry-only base metadata race fix.
- #14579 — open; retry-only duplicate-key recovery.
- #14582 — closed; retry-only duplicate-key recovery.
- #14940 — open; retry-only duplicate-key recovery.

My take:

- The retry-only PRs (#14579/#14940/etc.) are incomplete for the “repo vanished” symptom because retrying `merge()` can still write `None` over git fields.
- #15015 improves on them by addressing repo-nulling, but its current null-preserving implementation breaks intentional repo removal through the documented API/service contract.
- The best final fix likely combines:
  - duplicate-key race recovery,
  - webhook-stub null protection,
  - preserved explicit user/API ability to clear repo metadata.

## Is the edited file used by `/enterprise` code?

Yes.

The edited file is directly used by enterprise code in several ways:

1. `enterprise/server/utils/saas_app_conversation_info_injector.py` defines:

```python
class SaasSQLAppConversationInfoService(SQLAppConversationInfoService):
```

and its `save_app_conversation_info()` calls:

```python
await super().save_app_conversation_info(info)
```

So PR #15015’s base save behavior applies to enterprise SaaS saves too.

2. `enterprise/storage/stored_conversation_metadata.py` re-exports:

```python
from openhands.app_server.app_conversation.sql_app_conversation_info_service import (
    StoredConversationMetadata as _StoredConversationMetadata,
)
```

3. Enterprise sharing/query code also imports `StoredConversationMetadata` from the same base service/model.

So yes: this file is part of the enterprise runtime path, not just OSS.

## Checks I ran

- `uv run --python 3.12 pytest tests/unit/app_server/test_sql_app_conversation_info_service.py -q`
  - ✅ `34 passed`
- Checked PR CI:
  - ✅ all required visible checks successful
  - includes enterprise Python unit tests and enterprise lint
- Ran a direct reproduction script for explicit repo clearing through the service path:
  - ❌ confirmed repo/branch/provider are not cleared under PR behavior

## Risk assessment

🟡 **Medium risk**.

The touched code is central metadata persistence used by both OSS and enterprise conversation services. The race being fixed is real, but the current fix changes general save semantics and breaks an existing documented update path.

**Verdict:** ❌ Needs rework before merge.

shanemort1982:
Thanks @enyst, good catch on the clearing contract. Reworked along your option 2:

- **Scoped the preservation.** `save_app_conversation_info` gains `preserve_git_fields_on_null` (default `False`); the `COALESCE` is applied only when it's set. Only the conversation-start webhook (`on_conversation_update`) passes `True`. Every other caller, including the explicit `update_app_conversation` PATCH path, keeps the default, so `selected_repository=null` clears the repo again.
- **Contract test added.** `test_explicit_update_can_clear_repository` creates with a repo, then saves all three git fields as `None` via the default path and asserts they're actually cleared.
- **IntegrityError branch now covered.** `test_save_recovers_from_insert_integrity_race` injects a duplicate-key `IntegrityError` on the insert commit and asserts the rollback + retry path runs (commit twice, rollback once).
- **Trimmed** the comment blocks.

Full file passes locally (`36 passed`) and `ruff check` / `ruff format --check` are clean. The base service is shared with enterprise, so the same default-clears / webhook-preserves behaviour applies there too.

On `Fixes OpenHands/enterprise#26`: agreed it's a touch broad given the SaaS metadata/consistency angle. Happy to reword to "addresses the repository-nulling symptom of OpenHands/enterprise#26" if you'd prefer.

enyst:
Thank you @shanemort1982 ! May I suggest to bring this PR to `#proj-agent-canvas` on Slack, if you didn't yet. It could benefit from more eyes on it because it affects `enterprise`, and I don't speak for that codebase. 🙏

neubig:

Thank you for understanding.

Issue #3908, opened 2024-09-17
[Bug]: Intermittent AgentState "stopped" in the backend


enyst:
### Is there an existing issue for the same bug?

- [X] I have checked the troubleshooting document at https://docs.all-hands.dev/modules/usage/troubleshooting
- [X] I have checked the existing issues.

### Describe the bug

AgentState is an enum, and afaik it's supposed to have only those values in an AgentStateChangeAction / Observation, but recently I see a state change to "stopped", probably from [here](https://github.com/All-Hands-AI/OpenHands/blob/2a10ff13741dc47452240787b06def2463d38903/openhands/core/schema/agent.py#L25), and indeed it only happened when running with the UI.

This might be fine or just a matter of display in logs, or it may mean checks in the backend for AgentState.STOPPED don't trigger. Just recording this, because I may not remember all the unusual things. 😊 

### Current OpenHands version

```bash
[main](https://github.com/All-Hands-AI/OpenHands/commit/2a10ff13741dc47452240787b06def2463d38903)
```

### Installation and Configuration

```bash
`make build`
`make run`
```

### Model and Agent

Gemini
CodeActAgent

### Operating System

Mac OS

### Reproduction Steps

Run with UI for a while, and take a look at the console logs or check the event stream for "to stopped".

PR #3907, opened 2024-09-16
[eval] Fix multi-processing bug (again^3) & allow set EXP_NAME for each `run_infer`

xingyaoww:
**Short description of the problem this fixes or functionality that this introduces. This may be used for the CHANGELOG**

Hopefully fix https://github.com/All-Hands-AI/OpenHands/pull/3875#issuecomment-2351838834.

(Can't count this is how many time we try to fix the eval bug 😭 )

---
**Give a summary of what the PR does, explaining any non-trivial design decisions**

- Revert the "re-add error instance to the queue" logic introduced in https://github.com/All-Hands-AI/OpenHands/pull/3854 to simplify things.
- Right now, we simply re-try in a wrapper for each `process_instance_func` - we keep running the `func` for 5 time straight until the model got it right.

---
**Link of any specific issues this addresses**

PR #15020, opened 2026-06-29
feat(app-server): store and serve forwarded sub-agent events on a separate sub-stream

vishalhirapara-acceleratedata:
Refs OpenHands/software-agent-sdk#3907
Companion SDK PR: OpenHands/software-agent-sdk#3908

## Why
App-server side of live sub-agent event forwarding. When the agent-server forwards
sub-agent (`TaskToolSet`) inner events on the parent conversation stream (tagged
with `parent_tool_use_id`), the app-server should persist/serve them **separately**
so the parent conversation stream stays clean (only `TaskAction` +
`TaskObservation`) and the main agent's context is unaffected.

## Summary
- **Separate on-disk store** for sub-agent events: `EventServiceBase` gains
  `get_subagent_conversation_path` / `save_subagent_event` / `search_subagent_events`
  → `…/<conversation_id>/subagents/<tool_call_id>/events/` (not the parent's flat dir).
- **Webhook routing + analytics guard** (`event_callback/webhook_router.py`):
  events with a non-null `parent_tool_use_id` are routed to `save_subagent_event`;
  all analytics scans (stats, terminal-state, switch-LLM) skip them so sub-agent
  activity never skews conversation metrics.
- **Sub-stream read endpoint**: `GET /api/v1/conversation/{id}/subagents/{tool_call_id}/events`.

## How to Test
- `poetry run pytest tests/unit/app_server/test_subagent_event_store.py tests/unit/app_server/test_on_event_subagent_routing.py tests/unit/app_server/test_subagent_events_endpoint.py -v`
- End-to-end (with the companion SDK change + `OH_FORWARD_SUBAGENT_EVENTS=true`):
  delegate to a sub-agent, then `GET …/subagents/<tool_call_id>/events` returns the
  inner events (tagged), while `…/events/search` shows only `TaskAction` +
  `TaskObservation`.

## Type
- [x] Feature

## Notes
- **Depends on the companion SDK PR (#3908)** which adds `Event.parent_tool_use_id`
  and the forwarding. Until that lands and is released, this PR's tests (which build
  events carrying `parent_tool_use_id`) will not pass in CI — opening as **draft**
  pending the SDK approach agreed in #3907.
- Keyed by `tool_call_id` (the on-the-wire correlation id) to keep one directory per
  sub-agent task without adding a second event field.

### DIFF
--- a/openhands/app_server/event/event_service.py
+++ b/openhands/app_server/event/event_service.py
@@ -61,6 +61,18 @@ class EventService(ABC):
     async def save_event(self, conversation_id: UUID, event: Event):
         """Save an event. Internal method intended not be part of the REST api."""
 
+    async def save_subagent_event(
+        self, conversation_id: UUID, tool_call_id: str, event: Event
+    ) -> None:
+        """Persist a sub-agent event under a separate store keyed by tool_call_id.
+
+        Default implementation delegates to save_event so that EventService
+        implementations that don't override this still work correctly.
+        Concrete implementations (e.g. FilesystemEventService / AWS) override
+        this to write to an isolated sub-agent directory.
+        """
+        await self.save_event(conversation_id, event)
+
     async def batch_get_events(
         self, conversation_id: UUID, event_ids: list[UUID]
     ) -> list[Event | None]:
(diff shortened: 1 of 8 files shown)


VascoSch92:
We have a template for PR summary.

Moreover, we ask the user to read the PR before submitting it. We are not against the use of AI, but the person which open the PR should ackwoledge what is inside.

neubig:

Thank you for understanding.

PR #15021, opened 2026-06-29
fix: stream LLM tokens for cloud conversations and profile switches

VascoSch92:
HUMAN:

Enable streaming in settings. 

AGENT:

Evidence:

- `uv run pytest tests/unit/app_server/test_app_conversation_router.py tests/unit/app_server/test_live_status_app_conversation_service.py tests/unit/app_server/test_profiles_api.py` → **203 passed**, including the two extended happy-path assertions: the `switch_profile` endpoint sends `stream: true` to the agent-server's `switch_llm`, and `_configure_llm` returns `llm.stream is True`.
- Pre-commit on commit ran the full backend suite (ruff, ruff format, mypy) → all passed. `ruff@0.12.5 format --check` and `ruff check` are clean on the changed files.
- Not done here: a live deployed-app-server token-streaming check (needs a running sandbox); see the HUMAN checkbox.

---

## Why

Cloud agent conversations never enabled LLM token streaming. The app-server resolves the agent LLM from the user's saved settings (`_configure_llm`) and from saved profiles (`resolve_profile_llm`), both of which default `stream=False` (the SDK `LLM` default). With `stream=False` the SDK never invokes the `on_token` callback, so no `StreamingDeltaEvent`s are emitted and the UI receives whole messages at once. The GUI sets `stream=True` explicitly on the local start/switch paths, but the server-resolved cloud paths had no equivalent — so a cloud conversation (and any cloud profile switch) silently lost streaming.

## Summary

- `_configure_llm` (conversation start) now forces `stream=True` on the resolved agent LLM.
- `resolve_profile_llm` (profile switch endpoint + the agent-server profile-store sync) now forces `stream=True`, so a switched-to profile keeps streaming.

## Issue Number

N/A

## How to Test

- Unit: `uv run pytest tests/unit/app_server/test_app_conversation_router.py tests/unit/app_server/test_live_status_app_conversation_service.py` (203 pass).
- Manual (needs a deployed app-server): start a cloud conversation and confirm assistant tokens render incrementally; switch the LLM profile mid-conversation (composer switcher or `/model <name>`) and confirm it keeps streaming.

## Video/Screenshots

Not captured; live token-streaming verification is pending human testing against a deployed app-server.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Companion frontend PR: OpenHands/agent-canvas#1532 (surfaces cloud LLM profiles + chat-level switching, which is what exposed this streaming gap).
- Scope: only the main agent LLM. The condenser LLM (`agent.condenser.llm`) and the ACP path (`acp_settings.llm`, which streams natively) are built separately and intentionally untouched.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-5fdd1b8   docker.openhands.dev/openhands/openhands:5fdd1b8
```


hieptl (changes requested):
Hello @VascoSch92,

Would you mind including a short demo video showing how streaming works for cloud conversations? It would be very helpful for reviewing the implementation.

Also, have you had a chance to test this feature when running the SaaS application locally or using a feature deployment?

Thank you very much! 🙏

VascoSch92:
https://github.com/user-attachments/assets/1d1713ce-036d-42ca-93d1-bc53775cc2fa

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -1245,6 +1245,8 @@ def _configure_llm(self, user: UserInfo, llm_model: str | None) -> LLM:
                 'base_url': base_url,
                 'api_key': user.agent_settings.llm.api_key,
                 'usage_id': 'agent',
+                # Force streaming on (the SDK LLM defaults stream=False).
+                'stream': True,
```
hieptl, line 1249: **question (blocking):**

Am I understanding this correctly?

Forcing `stream=True` on the agent LLM also propagates to the condenser, which appears to cause agent runs to fail during hard condensation.

The SDK's `create_agent()` → `build_condenser(llm)` creates the condenser by copying the same LLM instance with only `usage_id` updated (`model_copy(update={"usage_id": "condenser"})`), which preserves `stream=True`. Additionally, `_apply_server_agent_overrides()` does not appear to reset this flag.

The condenser then calls `self.llm.completion(...)` without providing an `on_token` callback. This seems to trigger the synchronous guard in `llm.py`:

> `if enable_streaming and on_token is None: raise ValueError("Streaming requires an on_token callback")`

If that's correct, the following failure scenario seems possible:

* A long-running cloud conversation reaches the model's context window (or an explicit/forced condensation request is triggered).
* A **HARD** condensation is attempted.
* A `ValueError` is raised.
* `NoCondensationAvailableException` propagates out of `agent.step()` (there is no surrounding `try`/`except`), causing the agent run to fail.

At the normal `max_size` threshold, the behavior appears to be slightly different: **SOFT** condensation catches the exception, so the failure is hidden while the conversation context continues growing without bound.

If I'm interpreting the flow correctly, this also seems to contradict the PR description, which states that the condenser is "built separately and intentionally untouched." Subscription-authenticated LLMs avoid this issue because they use `condenser=None`, which could explain why it has not been observed so far.

Thank you very much! 🙏

hieptl, line 1249: **question (blocking):**

Am I understanding this correctly?

The same `stream=True` change also appears to silently degrade conversation title generation to raw truncation.

With the default configuration (when `title_llm_profile` is not set), automatic title generation falls back to `conversation.agent.llm`, which now has `stream=True` enabled. It then calls `llm.completion(messages)` without providing an `on_token` callback.

This seems to trigger the same guard, raising an exception. `generate_title_with_llm()` catches the exception, logs a warning, and returns `None`, causing the fallback behavior to truncate the first user message instead of generating an LLM-based title.

If that's correct, the resulting behavior would be:

* LLM-generated conversation titles (including emojis) are no longer produced.
* Conversations instead use a truncated version of the first user message as the title.
* The issue is silent: a warning is logged, but no error is surfaced and the conversation continues normally.

Thank you very much! 🙏

VascoSch92:
Hey, 

to answer your comments: both problems are already fixed in the sdk (https://github.com/OpenHands/software-agent-sdk/pull/3925 and https://github.com/OpenHands/software-agent-sdk/pull/3901) but are not released yet.

We can also wait with this PR until the agent-server version is bumped.

hieptl (approved):
Thank you! 🙏

PR #15023, opened 2026-06-29
fix: add managed LLM key refresh endpoint

neubig:
HUMAN:

No human-tested note added by this agent; the human testing checkbox remains unchecked.

- [ ] A human has tested these changes.

AGENT:

---

## Why

The managed LLM key refresh endpoint (`POST /api/keys/llm/managed/refresh`) needed to actually prove a configuration is managed before rotating, preserve OpenHands-provider key metadata, and stop duplicating the storage-layer managed-key lifecycle inside a FastAPI route. Reviewer (malhotra5) requested changes covering managed-config classification, OpenHands metadata, shared lifecycle, explicit failure handling, and real-behavior tests.

## Summary

- Centralizes the managed-key lifecycle in `SaasSettingsStore.rotate_managed_llm_key`, which resolves the effective LLM config via `load()` (org defaults + member diff, with org-default precedence), rejects non-managed/BYOK effective configs before generating anything, deletes the deterministic alias, generates a new key with `{'type': 'openhands'}` metadata for `openhands/*` models, and persists on the acting member row in the same session.
- `refresh_managed_llm_api_key` now delegates to the store and maps outcomes (`NOT_MANAGED`/`BYOK` -> 400, `MISSING_MEMBER` -> 404) to HTTP errors; the previous key token is deleted best-effort only after a successful persist.
- Removes the route-local managed-key helpers (`ManagedLlmKeyConfig`, `_managed_llm_key_config_from_model`, `get_effective_managed_llm_key_config`, `get_managed_llm_key_from_db`, `store_managed_llm_key_in_db`, `generate_managed_llm_key`) that duplicated `OrgStore._maybe_get_managed_llm_key_for_user` / `SaasSettingsStore._ensure_api_key`.
- Replaces mocked-helper tests with real-behavior tests against an in-memory SQLite DB (mocking only the external LiteLLM HTTP calls): managed OpenHands metadata, managed base_url without metadata, non-managed/org-BYOK rejection, BYOK member rejection, missing-member no-rotate/no-delete, and route wiring.
- Merged current `origin/main` into this branch in `cd19cc0` and resolved the conflict in `openhands/app_server/app_conversation/live_status_app_conversation_service.py` by preserving resolved agent-profile loading, marketplace composition, and the managed-key refresh/logging path.

## Issue Number

Fixes #15022.

## How to Test

Validated by the agent on 2026-07-10 UTC after merging `origin/main` (`ef1a3c7`) into this PR branch:

[...]

PR #15024, opened 2026-06-29
Self-heal stale managed LLM keys before conversation startup

### LINKED ISSUE #15022: Add managed LiteLLM key refresh and startup self-healing
## Problem

Some OpenHands Cloud automations can fail before producing useful events when the managed LiteLLM key stored for the automation-running org member no longer exists in LiteLLM. In the observed failure, conversation startup/title generation failed with `token_not_found_in_db` for a managed key (`sk-...JUrQ`), causing automation runs to fail before profile-rotation logic could recover.

The current UI/API has a BYOR key refresh endpoint, but there does not appear to be a dedicated endpoint to refresh the managed conversation key (`org_member.llm_api_key`) used by normal SaaS conversation startup.

## Proposed stacked work

1. Add a managed LiteLLM key refresh endpoint for the authenticated user/effective org. It should only operate on managed keys and must not overwrite BYOK/BYOR keys.
2. Add startup self-healing that minimally verifies managed keys before conversation startup and invokes the managed refresh logic when the key is missing from LiteLLM.

## Requirements

- Only refresh keys for managed LiteLLM/OpenHands-provider configurations.
- Never validate or replace user-provided BYOK provider keys against LiteLLM.
- Avoid deleting the old key before the replacement has been generated and persisted where possible.
- Add tests for both explicit refresh and startup self-healing.

## PR evidence acceptance criteria

- [ ] **Current `main` evidence:** In a live test org on current main, delete the managed LiteLLM key upstream while retaining its database reference, then start a conversation and capture token_not_found_in_db before useful events.
- [ ] **PR branch evidence:** On the PR branch, repeat the deletion and verify startup detects the missing managed key, creates/persists a replacement without prematurely deleting valid state, and the conversation starts; separately prove BYOK/BYOR keys are untouched.
- [ ] **Required evidence package:** The PR records both commit SHAs and exact environment/setup. Run the full relevant backend, service, infrastructure, or data pipeline live with PR code and attach commands plus output/logs; a movie or video is not a substitute for this live run. Attach sanitized DB key fingerprints, LiteLLM API responses, startup logs, conversation events, and exact full-stack commands.

neubig:
## Summary

- Adds best-effort SaaS-only startup self-healing for stale managed LiteLLM keys.
- Only runs when the resolved LLM base URL is the managed LiteLLM proxy and the configured key matches the member's stored managed key.
- Skips OSS/local, BYOK/custom-provider base URLs, masked/missing keys, and profile/custom keys that do not match the member managed key.
- Reuses the managed refresh endpoint/helper introduced in #15023, then continues startup with the refreshed key.

Stacked on #15023. Closes #15022 when merged with #15023.

## Testing

- `PYTHONPATH=".:$PYTHONPATH" uv run pytest tests/unit/app_server/test_live_status_app_conversation_service.py -q` (128 passed)
- `PYTHONPATH=".:$PYTHONPATH" poetry run --project=enterprise pytest --forked -n auto -s -p no:ddtrace -p no:ddtrace.pytest_bdd -p no:ddtrace.pytest_benchmark ./enterprise/tests/unit/server/routes/test_api_keys.py -q` (28 passed)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-656eedd   docker.openhands.dev/openhands/openhands:656eedd
```

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -1240,6 +1240,70 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
             }
         )
 
+
+    async def _maybe_refresh_managed_llm_key(
+        self, user: UserInfo, llm: LLM
+    ) -> LLM:
+        """Best-effort refresh for stale SaaS managed LiteLLM keys.
+
+        This intentionally only runs for SaaS managed LiteLLM keys that are the
+        current member's stored managed key. BYOK/custom keys and OSS/local
+        deployments are left untouched.
+        """
+        if self.app_mode != 'saas' or not user.id or not llm.api_key:
+            return llm
+
+        try:
+            from server.routes.api_keys import (  # type: ignore[import-not-found]
+                get_managed_llm_key_from_db,
+                refresh_managed_llm_api_key,
+            )
+            from storage.lite_llm_manager import (  # type: ignore[import-not-found]
+                LiteLlmManager,
+            )
+
+            from openhands.app_server.settings.settings_router import LITE_LLM_API_URL
+        except Exception:
+            return llm
+
+        if (llm.base_url or '').rstrip('/') != LITE_LLM_API_URL.rstrip('/'):
+            return llm
+
+        key = (
+            llm.api_key.get_secret_value()
+            if isinstance(llm.api_key, SecretStr)
+            else str(llm.api_key)
+        )
+        if not key or key == '**********':
+            return llm
+
+        get_effective_org_id = getattr(self.user_context, 'get_effective_org_id', None)
+        if get_effective_org_id is None:
+            return llm
+
+        try:
+            org_id = await get_effective_org_id()
+            if org_id is None:
+                return llm
+
+            managed_key = await get_managed_llm_key_from_db(user.id, org_id)
+            if managed_key != key:
+                return llm
+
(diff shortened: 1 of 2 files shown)


openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -1240,6 +1240,70 @@ def _configure_llm(self, user: UserInfo, llm_model: str | None) -> LLM:
...
+            org_id = await get_effective_org_id()
+            if org_id is None:
+                return llm
+
+            managed_key = await get_managed_llm_key_from_db(user.id, org_id)
+            if managed_key != key:
+                return llm
+
+            if await LiteLlmManager.verify_key(key, user.id):
+                return llm
```
malhotra5, line 1294: It seems that we'll be making a request to litellm every time to verify that the key works

Is there a reason we need to store the litellm keys in our DB then? Maybe its simpler to just read it directly from litellm every time and avoid having to synchronize potentially stale keys (and occasionally, we may perform an addition refresh if necessary)

malhotra5: Turns out litellm can only issue or revoke keys, but doesn't allow reading an existing valid key which is why we store it in DB - so this approach seems reasonable

PR #15025, opened 2026-06-29
feat(frontend): Add Budgets dashboard and expand Usage Dashboard

saurya:
HUMAN: revamps the admin dashboard for organizations by adding a usage dashboard and subtabs for looking at users, conversations, and model-specific spend, it also adds a feature to budget and allow control of spend via the OpenHands settings.

[X] A human has tested these changes.

## Summary

This PR adds Budgets dashboard and expands Usage Dashboard with Overview, Users, and Models tabs.

1. Budgets Dashboard at /settings/budgets
2. Expanded Usage Dashboard with 4 tabs

## Deployment notes
- Requires deployment changes (SMTP env/secret wiring + scheduled budget maintenance cronjob).
- OpenHands-Cloud PR: https://github.com/OpenHands/OpenHands-Cloud/pull/780
- Deploy PR: https://github.com/OpenHands/deploy/pull/4749

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-c6421e0   docker.openhands.dev/openhands/openhands:c6421e0
```


enterprise/server/routes/org_models.py
```diff
@@ -661,6 +661,88 @@ class OrgMemberFinancialPage(BaseModel):
...
+    reset_day: int
+    slack_channel: str | None = None
+    slack_team_id: str | None = None
+    default_user_monthly_limit: float | None = None
+    cycle_start_at: datetime
+    cycle_end_at: datetime
+    current_spend: float = 0.0
+    current_spend_percentage: float = 0.0
+    thresholds: list[OrgBudgetThresholdResponse] = Field(default_factory=list)
+    users: list[OrgBudgetUserResponse] = Field(default_factory=list)
```
tofarr, line 772: Should these be paginated? What if there are a 10k of them?

enterprise/server/routes/org_models.py
```diff
@@ -661,6 +661,88 @@ class OrgMemberFinancialPage(BaseModel):
...
+
+
+class OrgBudgetSettingsUpdate(BaseModel):
+    enabled: bool | None = None
+    monthly_limit: float | None = Field(default=None, gt=0)
+    reset_day: int | None = None
+    default_user_monthly_limit: float | None = Field(default=None, gt=0)
+    slack_channel: str | None = None
+    slack_team_id: str | None = None
+    thresholds: list[OrgBudgetThresholdUpdate] | None = None
```
tofarr, line 785: Provable needs a validator of max number permitted

smolpaws (commented):

Taste rating: acceptable

Findings:
- Changed area reviewed: enterprise/migrations/versions/127_add_execution_status_to_conversation_metadata.py, enterprise/migrations/versions/128_add_org_budgets.py, enterprise/migrations/versions/129_add_user_login_timestamps.py and 38 more file(s).
- I did not see concrete evidence of supply-chain abuse, credential theft, prompt injection or a correctness issue that should block this PR in the diff I reviewed.

Risk assessment: medium.

Verdict: worth merging if CI passes

tofarr:
## Review: open concerns before approval

Thanks for the work here — the feature itself (budgets + expanded usage dashboard) is solid. I am **not approving yet** because the PR bundles seven independent refactors and a few of them diverge from the rest of the codebase in ways worth a closer look first.

> _This comment was drafted with assistance from an AI coding agent on behalf of the reviewer._

### Process: PR scope

The diff contains at least seven independent changes:
1. New budgets feature (DB + service + routes + UI)
2. Usage Dashboard with Users / Models tabs + time-window selector
3. **`email_service.py`: `resend` -> raw `smtplib` migration**
4. `python-keycloak` demoted to an optional import
5. `User.first_login_at` / `last_login_at` columns + `record_login`
6. In-place edit to already-shipped migration `127_add_execution_status_to_conversation_metadata`
7. i18n rename + a few clean-ups

I would suggest splitting (3), (4), (6) and (7) into separate PRs so they can land behind the right flags and so reviewers can reason about each one in isolation.

---

### Architectural / design concerns

1. **`email_service.py`: blocking SMTP inside an async service.** `client.sendmail(...)` and `client.login(...)` are synchronous calls executed on the asyncio event loop. Every other helper in this file is async. Suggest wrapping the SMTP call in `asyncio.to_thread(...)` or moving it to a dedicated thread/process.

2. **`email_service.py`: Resend -> SMTP is an architectural regression for a SaaS.** No retries / exponential backoff / bounce feedback loop / delivery analytics. Seven new env vars (`SMTP_HOST`, `SMTP_PORT`, `SMTP_USERNAME`, `SMTP_PASSWORD`, `SMTP_FROM_EMAIL`, `SMTP_USE_SSL`, `SMTP_USE_TLS`) replace one (`RESEND_API_KEY`). I am not seeing an accompanying deployment / helm-values / DEV guide update in this PR — invitations and budget alerts will silently fail after merge until ops catches up.

3. **`keycloak_manager.py` + `token_manager.py`: keycloak turned into an "optional" import.** A blanket `try: import keycloak` with stub exception classes is a global, runtime-wide change to an auth-critical surface to enable two unit tests. The idiomatic pattern is `@pytest.importorskip("keycloak")` at the test boundary. The stub `KeycloakError` classes will silently drift from the real package.

4. **Stale `RESEND_API_KEY` references after the migration:**
   - `frontend/.../budgets.tsx` still tells users *"Email alerts require RESEND_API_KEY set in the deployment environment and a restart."*
   - `_get_email_enabled` in `default_web_client_config_injector.py` still has a `bool(os.getenv('RESEND_API_KEY','').strip())` fallback in its OSS Exception branch.
   - Docstrings in `test_default_web_client_config_injector.py` still mention "Resend integration".

5. **`frontend/src/i18n/translation.json`: lost translations on `SETTINGS$NAV_ADMIN_DASHBOARD`.** The English label was renamed from "Admin Dashboard" to "Usage & Monitoring" and **all 14 languages were replaced with the literal English string** — the previous `ja: "管理ダッシュボード"`, `de: "Admin-Dashboard"`, `es: "Panel de Admin"`, etc. are gone. The newly-added `SETTINGS$NAV_BUDGETS` key directly below it *does* have 14 translations, which makes the loss on the adjacent key conspicuous.

6. **`OrgBudgetService` performs writes during GETs.** `get_budget_state` (and `get_user_budget_row`) call `_roll_cycle_if_needed` (writes `cycle_start_spend`, resets `last_triggered_*`), `_maybe_send_alerts` (email + Slack) and `_sync_litellm_budgets` (1 LiteLLM HTTP call per team member). A slow LiteLLM response can pin the dashboard page for several seconds. The cycle roll and the budget sync should be moved to the existing maintenance-task / cron infrastructure rather than triggered by a UI read.

7. **`OrgBudgetService` commits directly via `self.db_session`,** while the neighbouring `OrgAppSettingsService` delegates persistence to a Store. Two persistence patterns in the same package.

8. **`admin-dashboard.tsx` gutted: -1142 / +5 lines.** Effectively a file rename done as delete + add, which loses git blame. Either `git mv` the old name to a new path (the new code already lives at `usage-dashboard.tsx`) and add a thin re-export shim, or `git mv` both.

9. **URL `/settings/admin-dashboard` still says "admin-dashboard"** but the page is now "Usage & Monitoring". Confusing for bookmarks and breadcrumbs.

10. **Migration 127 edited in place.** Adding idempotency guards (`if 'execution_status' not in columns`) to an already-shipped alembic migration is normally a smell. Either add a new head-down migration or add a note in the PR description explaining why.

11. **TZ-aware vs TZ-naive columns.** `OrgBudgetSettings.cycle_start_at` is `DateTime(timezone=True)`; `User.first_login_at` / `last_login_at` are plain `DateTime`. `record_login` even does `datetime.now(timezone.utc).replace(tzinfo=None)` to coerce. Mixing within the same module is a future foot-gun.

12. **`record_login` is Keycloak-only.** Other IDP flows (GitHub, GitLab OAuth) in the surrounding codebase will not populate these columns.

---

### Security / authorization

1. **`accept_tos` still logs `access_token` and `refresh_token`.** Pre-existing, but you rewrote the log line — a one-line fix to remove them entirely (or log only their presence) would be a small win while you are already touching it.

2. **Authorization asymmetry between UI and API.** The frontend route guard for `/settings/budgets` requires `admin` or `owner`, but the backend uses `require_financial_data_access`, which also allows any user with an `@openhands.dev` email. An OpenHands employee can hit the API to edit any org's budget even though the link is hidden from them in the UI. Either narrow the API to match the UI, or have the API log every mutation through this path loudly.

3. **`slack_channel` / `slack_team_id` accept arbitrary strings.** `OrgBudgetSettingsUpdate.slack_channel` is `str | None = None` with no `pattern` / `max_length` validator. The value is later passed straight to `chat_postMessage(channel=settings.slack_channel, ...)`. Worth a Pydantic validator (`#` prefix, channel-name charset, max length).

4. **`users_search` LIKE wildcards not escaped.** Typing `%` in the search box matches every user. Replace with `User.email.ilike(pattern.replace('\\','\\\\').replace('%','\\%').replace('_','\\_'))` plus `ESCAPE '\\'`.

5. **Silent LiteLLM sync failures.** If `LiteLlmManager.update_team` / `update_user_in_team` fails, the exception is `except Exception` and only logged at `warning`. The DB thinks the budget is on; LiteLLM does not. Suggest tracking `last_litellm_sync_at` / `last_litellm_sync_error` on `OrgBudgetSettings` and surfacing it in `OrgBudgetSettingsResponse` so admins can see drift.

6. **`cycle_start_spend` anchor.** `max_budget = settings.cycle_start_spend + settings.monthly_limit`. If an org enables budgets mid-cycle, this is anchored to "now", but LiteLLM may independently roll its own team-spend counter on its own schedule. The two-cycle boundaries can drift; worth a comment or an explicit alignment.

---

### Test coverage

Three new unit-test files (`test_org_conversation_service_pr_merge.py`, `test_user_store_login.py`, `test_default_web_client_config_injector.py`) plus in-place edits to `test_auth_routes.py` and `test_email_service.py`. The new 812-line `OrgBudgetService` has **no internal tests** — in particular:

- no coverage for `_roll_cycle_if_needed`
- no coverage for `_sync_litellm_budgets` (the function that talks to LiteLLM and mutates remote budgets)
- no coverage for `_maybe_send_alerts` / `_send_alerts`
- no coverage for `OrgBudgetThresholdUpdate._validate_thresholds` (count + uniqueness)
- no coverage for `OrgBudgetSettingsUpdate._validate_reset_day`

`test_send_budget_alert_uses_smtp_when_configured` is good. But given that budgets can disable users or zero out LiteLLM budgets on a remote system, I would want at least the threshold validator and `_roll_cycle_if_needed` covered before merging.

Happy to iterate on any of these — would you like to address them in this PR or split them out?

saurya:
Great review, thank you @tofarr. Here are the fixes I made:
> email_service.py: resend -> raw smtplib migration

Rather than change the email service, I added a parallel SMTP integration and tagged you in the comment thread with product on why we're doing this. I think it's still potentially worth splitting out into a separate PR since there are product implications for how we're doing things right now (with just Resend support).

> python-keycloak demoted to an optional import

This was a bit of a kerfuffle. Since some of the new tests for enterprise DON'T require the database, the agent seems to have stashed them in the top-level /tests/ directory and then imported enterprise dependencies like python-keycloak. I moved the tests back to enterprise/tests, making this unnecessary.

> In-place edit to already-shipped migration 127_add_execution_status_to_conversation_metadata

This is fixed.

> i18n rename + a few clean-ups

This has been reverted.

These are the four mentioned in:
> I would suggest splitting (3), (4), (6) and (7) into separate PRs so they can land behind the right flags and so reviewers can reason about each one in isolation.

In addition, I've added the test coverage mentioned in the Test Coverage section of your comment, addressed all the security and authorization related issue and all the architectural concerns mentioned that weren't already covered in the list above!

The only issue I haven't addressed is the writes during GETs - that requires a bit more infrastructure that I'll be adding soon.

saurya:
the budget sync piece is also done - I'll put up the demo soon.

PR #15031, opened 2026-06-29
feat(device-verify): align warning, button color, and add workspace dropdown

tofarr:
HUMAN:

Updated the style to the standard and added a workspace picker.

AGENT:

---

## Why

The `/oauth/device/verify` page (enterprise OAuth device-flow landing page) had three small consistency gaps versus the rest of the app:

1. The "Security Notice" block was hand-styled with ad-hoc Tailwind classes (`bg-amber-950/50 border-l-2 border-amber-500 rounded-r-lg`) instead of using the canonical `<RiskAlert>` component that powers security warnings everywhere else (`confirmation-buttons.tsx`, `v1-confirmation-buttons.tsx`).
2. The "Authorize Device" button used Tailwind's stock `bg-blue-600` instead of the brand primary (`bg-primary text-[#0D0F11]`) used by `BrandButton` and the rest of the app's primary CTAs.
3. There was no way to choose which workspace the new device API key should be scoped to — the user could only authorize against their currently-active workspace.

## Summary

- Extend `<RiskAlert>` with a `severity="medium"` variant (amber-themed, mirroring the existing `severity="high"` shape) and use it on the device-verify page in place of the ad-hoc amber div.
- Recolor the Authorize button from `bg-blue-600 hover:bg-blue-700` to `bg-primary text-[#0D0F11] hover:opacity-80` to match `BrandButton`'s primary variant.
- Add a labeled **Workspace** dropdown above the Cancel/Authorize buttons that lists every org the user belongs to, defaults to their current workspace, and switches the active workspace via `useSwitchOrganization` on change. Hidden when `useShouldHideOrgSelector()` returns true (OSS mode, single-personal-workspace, or only one org visible after hiding personal workspaces), so single-org users see no clutter.

## How to Test

1. Build and run the frontend (`cd frontend && npm install && npm run build`).
2. In enterprise mode, start a CLI login flow (e.g. `openhands` → "log in via browser"). The CLI prints a URL of the form `https://<host>/oauth/device/verify?user_code=<code>`.
3. Open that URL while signed in as a user that belongs to **two or more** organizations. Expect:
   - The security warning now renders with the amber `<RiskAlert>` styling (icon + bold title + body, rounded-xl) — visually consistent with the high-risk warning shown when confirming a high-risk agent action.
   - The Authorize button is brand-yellow (`#F3CE49`) with dark text, not blue.
   - A new "Workspace" dropdown is visible directly above the Cancel/Authorize buttons, defaulted to your current workspace.
4. Open the dropdown and pick a different workspace. Expect:
   - A success toast confirming the switch (same toast the top-bar OrgSelector shows).
   - The selected workspace is now your active workspace.
5. Click **Authorize**. The new API key is created against the newly-selected workspace — verify by opening Settings → API Keys and confirming the new key's org matches.
6. Repeat with a user that has only a single (personal) workspace. Expect: the dropdown is hidden (no clutter).
7. Repeat with a user in OSS mode. Expect: the dropdown is hidden.

`npm run lint:fix` and `npm run build` both pass.

## Video/Screenshots

**Before**
image

**After**
image

## Type

- [ ] Bug fix
- [x] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **No backend changes.** The device-verify endpoint already creates the API key against `user.current_org_id` via `create_api_key`'s fallback (see `enterprise/storage/api_key_store.py:53-80`), so updating the active workspace via the existing `/api/organizations/{orgId}/switch` endpoint is sufficient to scope the new key.
- **Security:** The dropdown only lists orgs returned by `useOrganizations()`, which is server-side filtered to orgs the user is a member of. The `useSwitchOrganization` mutation goes through the same trusted switch endpoint the top-bar selector uses; membership is enforced server-side there. The Authorize endpoint doesn't accept an arbitrary `org_id` parameter, so there's no risk of the user POSTing an org id they don't belong to.
- **Race window:** Selecting a workspace and immediately clicking Authorize could in theory race the switch mutation. The dropdown's `loading={isOrgsLoading || isSwitchingOrg}` state surfaces this, and the user typically waits for the success toast before clicking Authorize. If stricter isolation is wanted, a follow-up could plumb an explicit `org_id` form field through `/oauth/device/verify-authenticated` so the request is self-contained.
- **No new i18n keys.** Reuses existing `WORKSPACE$TITLE`, `ORG$SELECT_ORGANIZATION_PLACEHOLDER`, and `ORG$PERSONAL_WORKSPACE`.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-b28f19b   docker.openhands.dev/openhands/openhands:b28f19b
```


frontend/src/routes/device-verify.tsx
```diff
@@ -168,17 +196,53 @@ export default function DeviceVerify() {
...
+                        value: org.id,
+                        label: getOrgDisplayName(org),
+                      })) ?? []
+                    }
+                    onChange={(item) => {
+                      if (!item || item.value === selectedOrgId) return;
+                      const org = organizations?.find(
+                        (o) => o.id === item.value,
+                      );
+                      switchOrganization({
```
hieptl: **suggestion:** A failed workspace switch may leave the dropdown displaying the newly selected organization while the API key is still minted for the previous organization, without surfacing any error to the user.

useSwitchOrganization only defines an onSuccess callback and does not handle onError. Additionally, the Dropdown is uncontrolled: when a user selects an organization, Downshift immediately updates its internal inputValue/selectedItem. The parent component only re-synchronizes the dropdown by remounting it via key={selectedOrg?.id-...}, which is driven by selectedOrgId.

If the workspace switch request fails, setOrganizationId is never invoked, so selectedOrgId (and therefore the key) remains unchanged. As a result, the dropdown is not remounted or reverted and continues to display the organization the user selected. If the user then clicks Authorize, the API key is created for the previously active current_org_id, while no error message or toast is shown.

In this security-sensitive flow, this results in a silent authorization against the wrong workspace. Although the top-bar organization selector exhibits the same underlying behavior, the impact is more significant here because issuing an API key for the wrong workspace is considerably more severe than a temporary navigation inconsistency.

Thank you very much! 🙏

hieptl (changes requested):
Hello @tofarr,

Could you please add tests to cover your changes?

Thank you very much! 🙏

frontend/src/routes/device-verify.tsx
tofarr: Good catch — fixed in 607ac5ddd.

**Workflow now:**
1. Dropdown onChange calls `switchOrganization.mutate(..., { onError })` with a per-call error handler.
2. While the mutation is in flight, Authorize is disabled (`isSwitchingOrg`).
3. On success, the hook-level `onSuccess` runs (setOrganizationId, queries invalidated, toast), `isPending` flips to false, and Authorize re-enables.
4. On failure, the per-call `onError` runs: `displayErrorToast(t(DEVICE$WORKSPACE_SWITCH_FAILED))` fires, and a local `switchFailed` flag flips to true. Authorize stays disabled.
5. The next onChange clears `switchFailed` before re-attempting, so the user can recover by picking the same or a different workspace.

Gate is now `disabled={isSwitchingOrg || switchFailed}`.

The dropdown’s visual selection isn’t reverted on failure (it stays on the org the user picked) — that’s intentional and matches the existing top-bar `OrgSelector` behavior. The security guarantee comes from Authorize staying disabled, not from reverting the dropdown. Happy to add a dropdown revert if you’d prefer it, but I think keeping Authorize locked is the cleaner UX.

hieptl (approved):
Could you please add tests to cover your changes?

Thank you very much! 🙏

tofarr:
Tests added per Hiep’s review submission. Commit `b28f19bd9`.

**`frontend/src/components/shared/risk-alert.test.tsx` (4 tests)** — covers the new `severity="medium"` amber variant.

**`frontend/src/routes/device-verify.test.tsx` (11 tests)**:
- workspace dropdown visibility (SaaS render + OSS-hide path)
- dropdown default = current workspace
- `switchOrganization.mutate` is called with the new org payload **and a per-call onError** when a different org is picked
- selecting the same org does nothing (no mutate call)
- Authorize button disabled while `isSwitchingOrg` is true
- Authorize button enabled when no switch is in flight
- Cancel button is NOT disabled while a switch is in flight
- failed switch → Authorize stays disabled + `displayErrorToast` called with `DEVICE$WORKSPACE_SWITCH_FAILED`
- user can recover by picking a different workspace after a failure (mutate is called again with a fresh onError)
- clicking the disabled Authorize button does not POST to `/oauth/device/verify-authenticated`

All hooks are mocked (`useIsAuthed`, `useAppMode`, `useShouldHideOrgSelector`, `useOrganizations`, `useSelectedOrganizationId`, `useSwitchOrganization`, `custom-toast-handlers`). `npm run lint:fix` and `npm run build` both pass. `npx vitest run` on the two new files reports 15/15 green.

hieptl (approved):
Thank you! 🙏

PR #15035, opened 2026-06-30
perf(app-server): page poll_agent_servers conversation read in batches (#14717)

SmartManoj:
HUMAN: 

Resolves https://github.com/OpenHands/enterprise/issues/29

AGENT:

---

## Why

`poll_agent_servers` (`openhands/app_server/sandbox/remote_sandbox_service.py`) currently buffers the **entire** list of conversations-to-refresh in memory for the whole poll tick — PR OpenHands/OpenHands#14637 deliberately read all conversations into a list under a short DB session, released the session, then did network I/O over the full list. That's correct, but peak memory grows with the number of active conversations (and on remote storage backends every tick is one big list call). This is the follow-up noted in the OpenHands/OpenHands#14637 review.

## Summary

- Page the Phase-1 read in fixed-size batches and interleave each batch's network I/O, so peak memory stays flat (~one batch) regardless of the number of active conversations.
- Each batch keeps the `fetch -> release -> network -> re-acquire -> write` pattern: the read DB session is released **before** any agent-server network call — now scoped per batch instead of per tick, so the OpenHands/OpenHands#14637 `idle in transaction` invariant is preserved.
- Add a configurable `poll_batch_size` (default `100`) on `RemoteSandboxServiceInjector`, threaded into `poll_agent_servers(..., batch_size=...)`.

## Issue Number

Resolves OpenHands/enterprise#29

## How to Test

```bash
uv run pytest tests/unit/app_server/test_remote_sandbox_service.py -k "poll_agent_servers"
```

New regression test `test_poll_agent_servers_reads_in_bounded_batches` (extends the existing `TestPollAgentServersSessionScoping` open-session-count harness) asserts:
- the read is paged — one `search_app_conversation_info` call per page, advancing the cursor and honoring `batch_size` as the page `limit`;
- the DB session is released **between** batches (never more than one open at any instant);
- no DB session is held during network I/O (the OpenHands/OpenHands#14637 guard).

Verified locally: the command above → **2 passed** (the new batch test plus the existing session-scoping test). `ruff 0.12.5 format`/`check` clean.

## Video/Screenshots

N/A — internal polling/perf change.

### DIFF
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -858,7 +858,9 @@ def _build_service_url(url: str, service_name: str, runtime_id: str) -> str:
         return f'{scheme}://{service_name}-{netloc}{path}'
 
 
-async def poll_agent_servers(api_url: str, api_key: str, sleep_interval: int):
+async def poll_agent_servers(
+    api_url: str, api_key: str, sleep_interval: int, batch_size: int = 100
+):
     """When the app server does not have a public facing url, we poll the agent
     servers for the most recent data.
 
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #15037, opened 2026-06-30
perf(enterprise): index openhands_prs for the unprocessed-PR worker query (#14632)

SmartManoj:
HUMAN:

Resolves OpenHands/enterprise#27

AGENT:

---

## Why

`openhands_prs` was the 4th-heaviest sequential-scan table in prod — ~72K seq scans reading **13.9B** tuples on a 648K-row table (INC-95). The culprit is the enrichment worker's `OpenhandsPRStore.get_unprocessed_prs`:

```python
select(OpenhandsPR)
.filter(and_(
    ~OpenhandsPR.processed,
    OpenhandsPR.process_attempts < max_retries,
    OpenhandsPR.provider == ProviderType.GITHUB.value,
))
.order_by(desc(OpenhandsPR.updated_at))
.limit(limit)
```

The table only had indexes on `repo_id` / `pr_number` / `status`, none of which match this `WHERE (provider, processed) ... ORDER BY updated_at DESC` path — so every tick full-scanned the table and sorted 648K rows.

## Summary

- Add a composite index `(provider, processed, updated_at)` that covers the query: `provider`/`processed` equality narrow the set, `updated_at` serves the `ORDER BY ... DESC LIMIT` (`process_attempts < max_retries` remains a cheap residual filter).
- Declared on `OpenhandsPR.__table_args__` and created via alembic migration `128`, mirroring the perf-index migrations `117` (event_callback) and `118` (conversation_metadata_saas) from the same INC-95 batch.
- Plain `CREATE INDEX` with `IF NOT EXISTS` (not `CONCURRENTLY`): the enterprise migration harness runs inside a transaction with a session advisory lock where alembic's `autocommit_block()` raises — see migration 117's note.

## Issue Number

Resolves OpenHands/enterprise#27

## How to Test

The index is exercised by `get_unprocessed_prs`. To confirm the planner uses it on a populated DB:

```sql
EXPLAIN (ANALYZE, BUFFERS)
SELECT * FROM openhands_prs
WHERE NOT processed AND process_attempts < 3 AND provider = 'github'
ORDER BY updated_at DESC LIMIT 50;
-- expect an Index Scan Backward using ix_openhands_prs_provider_processed_updated_at
-- instead of Seq Scan + Sort
```

Lint verified locally with the enterprise config (ruff 0.12.5, `enterprise/dev_config/python/ruff.toml`): `ruff format --check` → already formatted, `ruff check` → all checks passed. The alembic chain is linear (128 → 127, single head). I couldn't run the enterprise migration suite locally (separate env / no Postgres); the migration mirrors the already-merged 117/118.

## Video/Screenshots

N/A — DB index / migration.

### DIFF
--- a/enterprise/migrations/versions/128_add_openhands_prs_unprocessed_index.py
+++ b/enterprise/migrations/versions/128_add_openhands_prs_unprocessed_index.py
@@ -1,0 +1,44 @@
+"""Add composite index on openhands_prs for the unprocessed-PR worker query
+
+get_unprocessed_prs filters on (provider, processed) and orders by updated_at
+DESC, but only repo_id / pr_number / status were indexed. Every enrichment tick
+therefore full-scanned the table — openhands_prs was the 4th-heaviest
+sequential-scan table in prod (~72K seq scans / 13.9B rows read on ~648K live
+rows, INC-95). This composite index covers that query path.
+
+Plain CREATE INDEX (not CONCURRENTLY): the enterprise migration harness runs
+inside a transaction with a session advisory lock, where alembic's
+autocommit_block() (required for CREATE INDEX CONCURRENTLY) raises — see
+migration 117. The single-table build is brief and the short write lock is
+acceptable, consistent with migrations 117 and 118.
+
+Revision ID: 128
+Revises: 127
+Create Date: 2026-06-30
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+revision: str = '128'
+down_revision: Union[str, None] = '127'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.create_index(
+        'ix_openhands_prs_provider_processed_updated_at',
+        'openhands_prs',
+        ['provider', 'processed', 'updated_at'],
+        if_not_exists=True,
+    )
+
+
+def downgrade() -> None:
+    op.drop_index(
+        'ix_openhands_prs_provider_processed_updated_at',
+        table_name='openhands_prs',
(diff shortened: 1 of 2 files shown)


SmartManoj:
Note on the failing **`check-sync`** check: it's a benign permissions artifact of contributing from a fork, not a problem with this migration.

That job only detects that the PR touches migrations and then tries to **post an advisory comment** (`POST /repos/OpenHands/OpenHands/issues/15037/comments`). From a fork PR that call returns:

```
403 Resource not accessible by integration
{"body":"⚠️ This PR contains migrations. Please synchronize before merging to prevent conflicts."}
```

Posting a comment needs the `issues: write` / `pull-requests: write` scope, but a fork PR's `GITHUB_TOKEN` is **read-only** by default, so the comment POST 403s and the job errors out. Per GitHub's docs:

> **Run workflows from fork pull requests** — Allows users to run workflows from fork pull requests, using a `GITHUB_TOKEN` with **read-only permission**, and with no access to secrets.

— [Managing GitHub Actions settings for a repository](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository); see also [Controlling permissions for GITHUB_TOKEN](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token) ("any write permissions are adjusted to read-only" for fork PRs).

All substantive checks pass (Lint python, Lint enterprise python, Python Tests, Enterprise unit tests). The advisory comment `check-sync` was trying to post is informational — the alembic chain here is linear (128 → 127, single head).

neubig:

Thank you for understanding.

PR #15047, opened 2026-06-30
fix(static): catch only 404 HTTPException in SPA fallback

Ricardo-M-L:
## Summary

- Fix overly broad exception handling in `SPAStaticFiles.get_response()` that masked real errors
- Remove the FIXME comment by properly addressing the root cause

## Problem

The `SPAStaticFiles` class catches all exceptions with `except Exception`, which means real errors like permission denied, disk I/O failures, or encoding issues are silently swallowed and treated as a missing file. This makes debugging production issues much harder.

The FIXME comment noted that `HTTPException` didn't work — the actual issue was catching the exception type too broadly rather than specifically handling 404s.

## Fix

Replace `except Exception` with `except HTTPException` that specifically checks for status 404:

```python
except HTTPException as e:
    if e.status_code == 404:
        return await super().get_response('index.html', scope)
    raise
```

This ensures:
- 404s correctly fall back to `index.html` (SPA routing)
- All other exceptions (500s, permission errors, etc.) propagate properly
- The FIXME comment is removed since the issue is resolved

## Testing

- `ruff check` and `ruff format` pass
- Behavioral change: non-404 HTTPExceptions and other errors now propagate instead of being silently caught

### DIFF
--- a/openhands/app_server/git/git_router.py
+++ b/openhands/app_server/git/git_router.py
@@ -205,9 +205,11 @@ async def search_branches(
         Query(title='Repository name in format owner/repo'),
     ],
     query: Annotated[
-        str,
-        Query(title='Branch name search query'),
-    ],
+        str | None,
+        Query(
+            title='Branch name search query. If not provided, returns all branches.'
+        ),
+    ] = None,
     page_id: Annotated[
         str | None,
         Query(title='Optional next_page_id from the previously returned page'),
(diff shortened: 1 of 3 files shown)


Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

neubig:

Thank you for understanding.

PR #14519, opened 2026-05-21
fix(APP-1892): restore PostHog onboarding-completed and create-pr-button-clicked events

lilagrc:
HUMAN:

This PR fixes two missing Posthog events, one fired when a new PR is opened (create PR button clicked), and another when someone completes the onboarding form)

- [x] A human has tested these changes.

---

## Why

[APP-1892](https://linear.app/all-hands-ai/issue/APP-1892/bug-saas-posthog-events-no-longer-firing-in-saas) reports two PostHog events have stopped firing in SaaS:
- `onboarding_completed` — stopped reporting ~22 days ago
- `create_pr_button_clicked` — stopped firing ~9 days ago

Root cause: PR #14006 ("feat: move PostHog server side") deleted the client-side `posthog.capture(...)` calls but the server-side replacements were incomplete:
1. The new `submit_onboarding` handler in `enterprise/server/routes/onboarding.py` was created with the right analytics calls, but its `onboarding_router` was never registered in `enterprise/saas_server.py`. Meanwhile the frontend kept POSTing to `/api/complete_onboarding` (in `auth.py`), which only flipped a DB flag and never fired analytics.
2. No server-side replacement was added for `create_pr_button_clicked` anywhere — the button click only fills a chat prompt and never reaches the backend.

NPS surveys keyed off the PR button and onboarding analytics for group targeting are both blind as a result.

## Summary

- **`onboarding completed`**: consolidated the analytics call into the existing `POST /api/complete_onboarding` (which the frontend already calls). Accept a typed `OnboardingSubmission` body, fire `analytics.track_onboarding_completed(...)` and an org `group_identify(...)` to stamp `onboarding_completed_at`, all wrapped in try/except so analytics failures never block onboarding. Removed the orphan `enterprise/server/routes/onboarding.py` + its tests to leave a single source of truth.
- **`create pr button clicked`**: added `CREATE_PR_BUTTON_CLICKED` constant + typed `AnalyticsService.track_create_pr_button_clicked(...)`. New enterprise-only `POST /api/analytics/events/create-pr-button-clicked` endpoint (registered in `saas_server.py`) with a `Literal`-constrained `git_provider` payload so arbitrary client strings can't become PostHog properties. Frontend mutation hook fires from both `GitControlBarPrButton` and `GitToolsSubmenu` (the chat button has a sibling in the tools submenu); errors are deliberately swallowed so analytics never blocks the click.
[...]

PR #15054, opened 2026-06-30
fix: add client-side posthog.capture for PR button survey targeting

lilagrc:
## Problem

PostHog surveys configured to trigger on the `"create pr button clicked"` event stopped receiving responses around June 2 (after `cloud-1.37.0` deployed).

**Root cause:** When analytics moved server-side in #14006 (May 4), the client-side `posthog.capture()` call was removed. #14519 (May 27) restored the event but only as a server-side capture via `POST /api/analytics/events`. PostHog surveys require the JS SDK to observe client-side `capture()` calls to trigger survey popups — server-side events are invisible to the JS SDK.

## Fix

Adds a client-side `posthog.capture("create pr button clicked", ...)` call in `useTrackCreatePrButtonClicked` alongside the existing server-side event. The server-side capture is preserved for backend analytics; the client-side capture enables PostHog survey targeting.

## Changes

- **`frontend/src/hooks/mutation/use-track-create-pr-button-clicked.ts`** — Import `usePostHog` and fire `posthog.capture()` before the server-side POST.
- **`frontend/__tests__/hooks/mutation/use-track-create-pr-button-clicked.test.tsx`** — Mock `posthog-js/react` and add a test verifying the client-side capture fires with the correct event name and properties.

## Note

The PostHog survey in the dashboard may also need its event name updated — the old client-side event was `"create_pr_button_clicked"` (underscores) while the current event name is `"create pr button clicked"` (spaces).

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-16c8ac4   docker.openhands.dev/openhands/openhands:16c8ac4
```

### DIFF
--- a/frontend/__tests__/hooks/mutation/use-track-create-pr-button-clicked.test.tsx
+++ b/frontend/__tests__/hooks/mutation/use-track-create-pr-button-clicked.test.tsx
@@ -5,6 +5,11 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 import { analyticsEventsService } from "#/api/analytics-service/analytics-events.api";
 import { useTrackCreatePrButtonClicked } from "#/hooks/mutation/use-track-create-pr-button-clicked";
 
+const mockCapture = vi.fn();
+vi.mock("posthog-js/react", () => ({
+  usePostHog: () => ({ capture: mockCapture }),
+}));
+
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

Issue #15044, opened 2026-06-30
[AgentProfile][cloud] Implement Agent Profiles on the OpenHands cloud/SaaS surface


simonrosenberg:
**Epic:** OpenHands/software-agent-sdk#3713 · **Umbrella:** OpenHands/software-agent-sdk#3730 · **Thin-boundary contract:** https://github.com/OpenHands/software-agent-sdk/issues/3730#issuecomment-4841528619

This is the concrete OpenHands-side implementation issue for **Agent Profiles on the cloud / SaaS surface** (the umbrella's Phase 5). It consolidates the three SDK tracking sub-issues (software-agent-sdk#3832 / OpenHands/OpenHands#3833 / OpenHands/OpenHands#3834) into one buildable plan, adds the prerequisite SDK refactors, covers the migration + **retirement of the active-LLM-profile pointer** (§8; epic OpenHands/OpenHands#3713 Phase 6), and lists the coordinated agent-canvas un-gating.

Agent Profiles already work **locally** end-to-end (SDK substrate `openhands.sdk.profiles` shipped in **v1.29.0**, agent-server router, agent-canvas UI in PR OpenHands/agent-canvas#1440/#1446). **Cloud is unbuilt** — `git grep agent_profile` on OpenHands `main` (`0571ff530`) returns **0 hits**.

The single most important directive: **cloud already implements LLM Profiles, and an Agent Profile is the same kind of resource one level up (it *references* an LLM profile). Mirror the LLM-profile cloud pattern in every way that applies; deviate only with a stated reason.** Every decision below cites the LLM-profile precedent it mirrors.

---

### 1. Storage shape → dedicated `org.agent_profiles` `EncryptedJSON` column (NOT embedded in `agent_settings`)

**Mirror `org.llm_profiles` exactly.** LLM profiles are a named-collection-with-active-pointer stored in their own encrypted column, not inside `agent_settings`:

- `enterprise/storage/org.py:72-74` — `llm_profiles: Mapped[dict[str, Any] | None] = mapped_column(EncryptedJSON, nullable=True)`.
- Migration `enterprise/migrations/versions/116_add_llm_profiles_to_org.py` — `op.add_column('org', sa.Column('llm_profiles', sa.String(), nullable=True))` (EncryptedJSON serializes to String at rest), no backfill, NULL → empty collection lazily.

[...]

Issue #4017, opened 2024-09-24
[Bug]: CodeActAgent: o1 replies may contain <execute_ipython> in finish action


tobitege:
### Is there an existing issue for the same bug?

- [X] I have checked the troubleshooting document at https://docs.all-hands.dev/modules/usage/troubleshooting
- [X] I have checked the existing issues.

### Describe the bug

Due to how o1-mini works, and probably o1-preview as well, it may return an `AgentFinishAction` with embedded code.
Apparently this is currently not supported in the agent as there is no `CodeActActionParserIPythonRunCell` triggered.

Example with content manually shortened and line breaks added for readability:

```py
I will create the file o1-results.txt and write the findings from Step 1: 
<execute_ipython>create_file('o1-results.txt') append_file('o1-results.txt', """### Findings for Step 1: Understand the cli.py file

The cli.py file serves as the command-line interface for the OpenHands application. It performs the following functions:
[---content shortened---]
Overall, cli.py orchestrates the setup and execution of the agent in a command-line environment,
facilitating interaction between the user and the underlying agent functionalities. """)
</execute_ipython>
```

In that step's LLM response log we can see at the bottom:
```py
[---content shortened---]
</execute_ipython>

<finish></finish>
```

In the main logfile we see the `AgentFinishAction` (I manually added line breaks here for readability):
```py
10:03:18 - openhands:INFO: agent_controller.py:445 - AgentFinishAction(outputs={}, 
thought='I will create the file `o1-results.txt` and write the findings from Step 1:\n
<execute_ipython>\ncreate_file(\'o1-results.txt\')\n
append_file(\'o1-results.txt\', """### Findings for Step 1: Understand the cli.py file\n
\n
The `cli.py` file serves as the command-line interface for the OpenHands application.
It performs the following functions:\n
[---content shortened---]
Overall, `cli.py` orchestrates the setup and execution of the agent in a command-line environment, 
facilitating interaction between the user and the underlying agent functionalities.\n""")\n
</execute_ipython>', action='finish')
```

### Current OpenHands version

```bash
0.9.4 as of 2024-09-24
```

### Installation and Configuration

```bash
make build
make run
```

### Model and Agent

- Model: o1-mini
- Agent: CodeActAgent

### Operating System

WSL

### Reproduction Steps

See trajectory:
https://www.all-hands.dev/share?share_id=58c3e506b401943c64cc9a2091c7173fe7500c176639ac3fbbdd6659becd1d64

[...]

Issue #3730, opened 2024-09-04
[Refactor]: Simplify AgentController Logic


neubig:
**Summary**

The AgentController logic is very complex, making it difficult to understand. We can refactor it to improve readability.

https://github.com/All-Hands-AI/OpenHands/blob/add4653335218eb04d0b7960fa67c21f8e313b8e/openhands/controller/agent_controller.py#L47

WannaTen:
Hi, Prof.  I'd like to have a try on this,my initial thought is
-  To split 'on_event' into 'action_event' and 'observation_event'
-  To refactor the logic of '_step', which currently contains a large code block

How does that sound?

WannaTen:
@neubig Hi,  I refactored the logic of the AgentController in  #3873 , Could you please review it?

enyst:
This has been addressed by the PR linked above, thank you @WannaTen!

I'll close this issue, @neubig please re-open if you think there is more here.

neubig:
Thanks a lot! There might still be more to do here, but for the time being I think it's much better.

Issue #1571, opened 2024-05-03
[Bug]: minimum python 3.11.1 needed for async unit tests to pass.


tpsjr7:
### Is there an existing issue for the same bug?

- [X] I have checked the troubleshooting document at https://opendevin.github.io/OpenDevin/modules/usage/troubleshooting
- [X] I have checked the existing issues.

### Describe the bug

`poetry run pytest ./tests/unit/ ` can skips async io tests on the latest version of pytest-asyncio due to a bug in that library.
 I suggest pinning pytest-asyncio = "0.21.1"  according to this open issue https://github.com/pytest-dev/pytest-asyncio/issues/810

### Current Version

```bash
0.4.1
```

### Installation and Configuration

```bash
poetry run pytest ./tests/unit/
```

### Model and Agent

any

### Reproduction Steps

- git checkout 0.4.1
- make build
- observe that pytest-asyncio = 0.23.6 in poetry.lock
- intentionally create a failure on a test that is marked @pytest.mark.asyncio such as in test_action_github.py
-  `poetry run pytest  ./tests/unit/test_action_github.py `
- observe that it skips tests but still passes 

```
tests/unit/test_action_github.py::test_run_pull_request_creation_failed
  /home/user/.cache/pypoetry/virtualenvs/opendevin-rh7WTFfU-py3.11/lib/python3.11/site-packages/_pytest/python.py:151: 

PytestUnhandledCoroutineWarning: async def functions are not natively supported and have been skipped.
  You need to install a suitable plugin for your async framework, for example:
    - anyio
    - pytest-asyncio
    - pytest-tornasync
    - pytest-trio
    - pytest-twisted
    warnings.warn(PytestUnhandledCoroutineWarning(msg.format(nodeid)))
```

That message about "installing a suitable plugin" is a red herring that goes away if you use pytest-asyncio=0.21.1 instead.

tpsjr7:
I discovered that the root cause of this bug was caused with an interaction with this other bug with Python itself, https://github.com/pytest-dev/pytest-asyncio/issues/403#issuecomment-1334162210

I was using a fresh install of Ubuntu 22.04.4 LTS which was using python 3.11.0rc . The issue mentions it was fixed in 3.11.1. So, when I upgraded to python 3.11.9 this bug went away. 

So I actually recommend `python = "^3.11.1"` in the pyproject.toml instead of `python = "^3.11"` and leaving pytest-asyncio = '*' as currently is.

[...]

PR #15060, opened 2026-06-30
feat(enterprise): Agent Profiles on the cloud/SaaS backend (#15044)

simonrosenberg:
HUMAN:

Cloud/SaaS backend for Agent Profiles (OpenHands/OpenHands#15044, the second PR of the issue). Built as a thin store + auth + wiring shim over the SDK substrate from OpenHands/software-agent-sdk#3930 per the [thin-boundary contract](https://github.com/OpenHands/software-agent-sdk/issues/3730#issuecomment-4841528619) — every resolve / FK / seed / id-stamp call is imported from `openhands.sdk.profiles`, none re-implemented. Reviewed the diff end-to-end; verified migration up/down, ruff (pinned v0.12.5), mypy, and the unit tests locally against the pinned `openhands-sdk==1.30.0` (which ships OpenHands/OpenHands#3930).

**Draft blocker resolved:** `openhands-sdk` is now pinned to the `1.30.0` release (ships OpenHands/OpenHands#3930's public symbols) via OpenHands/OpenHands#15079, and the redundant `PluginSpec` source serializer is removed — CI is green. A follow-up commit addresses the automated cross-repo code-review findings (idempotent `DELETE`, typed detail response, and the documented `X-Expose-Secrets` masking behavior); see **Cloud deviations** below.

AGENT:

Implements **Agent Profiles on the OpenHands cloud / SaaS surface** (#15044, the umbrella's Phase 5). Cloud already implements LLM Profiles; an Agent Profile is the same resource one level up (it *references* an LLM profile), so this mirrors the LLM-profile cloud pattern (`org.llm_profiles` column → `org_profiles` router → `SaasSettingsStore`) in every way that applies, deviating only where stated.

The single architectural directive — **import the domain logic, don't re-derive it** — is honored: validation, the `llm_profile_ref` + `mcp_server_refs` → `AgentSettingsConfig` join, the delete-409 / cascade-rename FK, the default seed, the mint-id/bump-revision policy, and the 422 redaction all come from `openhands.sdk.profiles`. The cloud diff is grep-clean of re-implemented resolve/FK/seed/id logic.

## Why

Agent Profiles work locally end-to-end (SDK substrate `openhands.sdk.profiles`, agent-server router, agent-canvas UI) but **cloud was unbuilt** — `git grep agent_profile` on `main` returned 0 hits, so the canvas Agent Profiles UI (already cloud-ready by construction) fails only because cloud serves no endpoints. This PR makes the wire contract real.

## Summary

**Store + schema** (mirrors `org.llm_profiles`)
- `openhands/app_server/settings/agent_profiles.py` — `AgentProfiles` pydantic container `{profiles: {<id>: AgentProfile}, active}` that **also satisfies the SDK `AgentProfileStoreProtocol`** (id-keyed storage, name-based protocol ops). `lock()` is a re-entrant no-op — the router holds the DB row transaction. EncryptedJSON is the at-rest boundary, so `skills[].mcp_tools` ride cleartext inside the encrypted blob (no SDK cipher needed; parity with `org.llm_profiles`).
- `org.agent_profiles` EncryptedJSON column + `OrgMember.active_agent_profile_id` per-member pointer (the sole launch authority) + alembic migration **131** (revises 130 — renumbered twice after merging `main`: first past `128_add_not_before_to_api_keys`, then past main's `129`/`130` registered-marketplaces migrations).

**Router** — `enterprise/server/routes/agent_profiles.py`, mounted flat at `/api/agent-profiles` (org derived from the auth session via `EFFECTIVE_ORG_ID`, because the surface is member-facing). Mirrors `org_profiles.py`'s locked org-row transaction + `VIEW`/`EDIT_ORG_SETTINGS` perms. CRUD + `/{id}/activate` (pointer-only — no `agent_settings` materialization) + `/{name}/materialize` (dry-run). Each endpoint delegates to `validate_agent_profile` / `save_profile_preserving_identity` / `resolve_agent_profile_dry_run` / `safe_validation_error_detail`. First list against an empty store lazily seeds one default profile from the org's current `agent_settings` via `build_seed_profile` (the §8 migration, lazy-on-access).

**FK lifecycle** — `org_profiles.py` LLM-profile delete/rename now wire the SDK `find_referrers` / `cascade_rename` / `delete_llm_profile` / `rename_llm_profile`: deleting an LLM profile referenced by an Agent Profile → **409**; rename cascades the `llm_profile_ref` under the org row lock.

**Resolution + provenance**
- **Resolved vs persisted views are structurally separated** (added after a deep review found a load()→store() round-trip could write the resolved profile view back as user-authored settings — wiping the member's `mcp_config` and rewriting their LLM key). Plain `SettingsStore.load()` now returns the PERSISTED composed settings; the resolved Agent-Profile launch view is an explicit opt-in — `load(resolve_agent_profile=True, …)`, implied by an `override_agent_profile_id` — requested only by the conversation-start builders and the provenance stamp. Resolution goes via `resolve_agent_profile` (over an `OrgLLMProfileLoader` adapter on `org.llm_profiles` + member-effective `mcp_config`) and **replaces** the composed `agent_settings` on that view; the managed-key / base-url overlay reuses `resolve_profile_llm`. Fail-safes: a stale pointer (or any resolver crash, incl. future SDK contract drift — broad `except Exception`) falls back to composed settings rather than bricking the load, resolved views are marked and `store()` **refuses** them (`ValueError`), and the `shared_agent_settings_diff = {}` special-case is gone — org-level agent-settings edits persist again while a profile is active. Consequence for API consumers: `GET /api/settings` returns persisted (not profile-resolved) `agent_settings`; the active profile's view is served by the agent-profiles endpoints and the launch path. `/activate` is now pointer-only (the profiles blob is written back only when actually mutated, so a schema-drift-invalid profile can't be erased by an unrelated mutation).
- **Per-request launch override** — `AppConversationStartRequest.agent_profile_id` runs *this* conversation from a specific Agent Profile without touching the member's active pointer. Threaded as an optional `override_agent_profile_id` through the whole chain (`UserContext.get_user_info` → `UserAuth.get_user_settings` → `SettingsStore.load` → `SaasSettingsStore.load`; `FileSettingsStore` ignores it — no Agent Profile concept in OSS), default `None` preserving existing behavior everywhere. The override wins when given but is **never written back** to `OrgMember.active_agent_profile_id`, and it bypasses the settings / user-info memo caches so it neither reads a stale plain load nor poisons the cache for a later uncached call in the same request.
- **Provenance** — `LaunchedAgentProfile { agent_profile_id, revision }` stamped on `AppConversationInfo` via the `tags` dict + a `@computed_field` (the exact `acp_server` precedent — zero-migration). The launch and the provenance stamp re-resolve with the *same* override, so provenance never lies about what launched an overridden conversation.

**ACP on cloud** — dropped the `_merge_custom_mcp_config` ACP early-return and forward a resolved ACP profile's custom MCP to the subprocess (SDK#3705 remainder); the runtime-internal system MCP server is intentionally not injected into an external ACP CLI.

**Deferred / staged** (documented, not in this PR): the bulk one-shot backfill of every member's pointer, and the physical removal of `ProfileListResponse.active_profile` / the active-LLM-profile marker (epic OpenHands/OpenHands#3713 Phase 6).

## Cloud deviations from the local agent-server

The cloud surface mirrors the LLM-profile **cloud** pattern (not the local agent-server) wherever the two differ. Called out explicitly:

- **`GET /{name}` always masks `skills[].mcp_tools` secrets.** The `X-Expose-Secrets: encrypted|plaintext` header is intentionally **not** honored on cloud (unlike the local agent-server), mirroring the cloud LLM-profile `GET` (`org_profiles.get_profile`, which also masks). Secrets live in the clear only inside the `org.agent_profiles` EncryptedJSON column, never on the wire. **Editing an existing profile stays non-destructive:** `save_agent_profile` restores any masked `mcp_tools` value from the stored *namesake* (`_restore_masked_skill_secrets`), so a client that round-trips a fetched profile keeps its secrets even though it only ever saw the mask.
  - **Duplicate (save under a new name):** a *duplicate* (canvas `handleDuplicate` → save under a fresh name) has no namesake to restore from, so `save_agent_profile` now **drops** any unrestorable `<redacted>` `mcp_tools` value rather than persisting it as a literal secret — the mask degrades to an empty secret (user re-enters), exactly mirroring the LLM `**********` sentinel the LLM model nullifies to `None`. Note that `mcp_server_refs` / `llm_profile_ref` carry *no* inline secret (they are by-name references), so this only ever applies to a skill's inline `mcp_tools`.

Aligned with the ts-client / canvas wire contract (previously divergent, now fixed to match):

- **`DELETE /{name}` is idempotent** — a missing name resolves `200`, matching the ts-client `AgentProfilesClient` and the local agent-server `delete_profile` (canvas's delete mutation has no `404` branch). The per-member pointer clear still runs when the profile existed.
- **`AgentProfileDetailResponse.profile` is the `AgentProfile` discriminated union** (not `dict[str, Any]`), so the generated OpenAPI schema matches the type the ts-client consumes.

## How to Test

```bash
# Against the pinned openhands-sdk==1.30.0 (ships OpenHands/OpenHands#3930)
cd enterprise && PYTHONPATH=.. poetry run pytest tests/unit/test_agent_profiles.py \
  tests/unit/test_org_profiles.py tests/unit/test_saas_settings_store.py
poetry run pytest ../tests/unit/app_server/test_live_status_app_conversation_service.py
# migration round-trip (Postgres)
poetry run alembic upgrade head && poetry run alembic downgrade -1 && poetry run alembic upgrade head
uvx ruff@0.12.5 check --config dev_config/python/ruff.toml <enterprise files>
```

- New `enterprise/tests/unit/test_agent_profiles.py` (29 tests): container/Protocol conformance, router CRUD/activate/materialize lifecycle, 404/409/422 error paths, **idempotent-delete** + delete-clears-all-member-pointers, masked-`mcp_tools` per-server restore round-trip, lazy-seed migration (incl. the seed-race loss + org-wide-default fallback), the LLM-profile FK delete-409 + rename-cascade, and `SaasSettingsStore._resolve_active_agent_profile` (resolve, stale-pointer fallback, and the per-request override that resolves without persisting the pointer).
- Adjacent `test_org_profiles` / `test_saas_settings_store` / `test_live_status_app_conversation_service` suites pass unchanged (no regression).
- Migration 129 validated up → down → up on Postgres 16; single alembic head.

Refs: OpenHands/OpenHands#15044 · SDK prereq OpenHands/software-agent-sdk#3930 · epic OpenHands/software-agent-sdk#3713 · umbrella OpenHands/software-agent-sdk#3730

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-99e24f6   docker.openhands.dev/openhands/openhands:99e24f6
```


simonrosenberg:
@openhands Clone this repo as well as agent-canvas (feature branch feat-agent-profiles-local) . Then /codereview this PR given it must work with what was done in agent-canvas. Post your feedback with gh api with event approved, or comment, or request changes as you think fit.

simonrosenberg (commented):
> _This review was posted by an AI agent (OpenHands) on behalf of @simonrosenberg. Repos consulted: `OpenHands/OpenHands@feat-agent-profiles-cloud`, `OpenHands/agent-canvas@feat-agent-profiles-local`, and the pinned `openhands-sdk==1.30.0` substrate for `openhands.sdk.profiles`._

## 🟡 Taste rating — Acceptable, close to good

The thin-boundary contract is genuinely honored: `save_profile_preserving_identity`, `resolve_agent_profile`, `build_seed_profile`, `find_referrers`, `cascade_rename`, `delete_llm_profile`, `rename_llm_profile`, `safe_validation_error_detail` — all imported from `openhands.sdk.profiles`, none re-implemented in the diff. `git grep` on the enterprise diff shows the cloud side really is a store + auth + wiring shim, as advertised.

The tests are the kind of tests you want on a change like this — they exercise real code paths against a real SQLite `Org` row, not just `MagicMock` call assertions. `TestMaskedMcpSecretRestore.test_new_server_survives_save_alongside_masked_server` in particular pins down a subtle round-trip bug (wholesale skill revert vs per-server restore) that's easy to regress.

I found two real cross-repo mismatches with `agent-canvas@feat-agent-profiles-local` that are worth addressing before canvas un-gates the cloud path, and one nit.

---

## [CRITICAL ISSUES]

None — no data-loss / security / broken-userspace failures. The two items below are wire-contract divergences vs. the SDK typescript client that canvas targets; both fail gracefully because canvas gates `/settings/agents` off for cloud backends today (`use-settings-nav-items.ts` bails on `backend.kind === "cloud"`, and `chat-input-profile-picker.tsx` bails when the profile list is empty). Once canvas un-gates (per `#3730`), these will surface.

## [IMPROVEMENT OPPORTUNITIES]

### 1. `DELETE /api/agent-profiles/{name}` returns 404 on missing name — the SDK ts-client + local agent-server treat it as **idempotent** (200)

[`enterprise/server/routes/agent_profiles.py`, `delete_agent_profile`] The cloud handler raises 404 when the name is missing:

```python
try:
    deleted = profiles.load(name)
except FileNotFoundError:
    raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, ...)
```

The SDK typescript-client (`AgentProfilesClient`, v1.28.0 — the one canvas uses) documents the contract as:

> `delete` is idempotent — a missing name resolves 200, not 404.

And the local agent-server (`openhands/agent_server/agent_profiles_router.py`, `delete_agent_profile`) matches that contract — it just calls `store.delete(name)`, which is itself a no-op for a missing name. `AgentProfiles.delete()` **in this PR** is also idempotent (`if entry is None: return`), so the router is deliberately adding a 404 that the container doesn't require.

Canvas's `use-delete-agent-profile` mutation has no 404 branch — a repeat-delete or a stale-list delete on cloud will surface as a red error toast where locally it would silently succeed. Recommend dropping the pre-check and just calling `profiles.delete(name)` unconditionally (still capture the `deleted_id` before the delete so the `OrgMember.active_agent_profile_id` clear still works). This will also line the response body up with the local server's ("deleted" whether it existed or not).

### 2. `GET /api/agent-profiles/{name}` silently ignores the `X-Expose-Secrets` header

[`enterprise/server/routes/agent_profiles.py`, `get_agent_profile`] The handler always calls `profile.model_dump(mode='json')` with the default context, so `skills[].mcp_tools` env/headers always come back as `<redacted>`.

Canvas actively sends `X-Expose-Secrets: encrypted` on its Edit and Duplicate flows (`src/components/features/settings/agent-profiles/agent-profiles-manager.tsx`, `handleEdit` / `handleDuplicate`), expecting Fernet-encrypted values it can round-trip back through save. The mask-restore path here (`_restore_masked_skill_secrets`) does keep **save** correct — the stored secret survives a `<redacted>`-in / real-value-out cycle, and the new test `test_new_server_survives_save_alongside_masked_server` proves that — but the editor UX diverges: on cloud, users cannot see or edit an existing MCP server's env values, they can only replace the whole server. The comment on the response line (`the API never returns raw secrets [...] EncryptedJSON column is the only place they live in the clear`) reads like this is deliberate, but the PR description doesn't flag it and the `exposeSecrets` parameter in the client is now effectively dead on cloud.

Two reasonable resolutions, either is fine:
- **Document it.** Add a "Cloud deviations" note in the PR description + a comment on the handler making the intentional divergence explicit. Consider replying 400 when the header arrives with a non-omitted value (or logging a warning) so callers notice.
- **Wire encrypted secrets through.** Reuse the same Fernet mechanism other cloud endpoints already use (the LLM api_key in `/api/settings` exposes an encrypted form via the same header) so canvas's Edit UX behaves the same on cloud and local.

### 3. Nit — `AgentProfileDetailResponse.profile` is typed as `dict[str, Any]`

[`enterprise/server/routes/agent_profiles.py`, `AgentProfileDetailResponse`] The SDK ts-client types this field as the `AgentProfile` discriminated union. At runtime `profile.model_dump()` produces exactly that shape, so this is a schema/OpenAPI issue rather than a functional one — but a `SerializeAsAny[OpenHandsAgentProfile | ACPAgentProfile]` (or a manual `Field(discriminator='agent_kind')` union) would keep the generated schema honest and match the local agent-server response model.

## [TESTING GAPS]

None material. 20 tests, real code paths, real DB row, both success and error paths, a dedicated `test_falls_back_to_org_wide_active_pointer_when_member_pointer_unset` for the losing-side-of-the-seed-race case, and `test_load_override_resolves_without_persisting_pointer` proves the per-request launch override is one-off and never persists. Migration up→down→up is claimed against Postgres 16 in the PR body; the SQLite tests are enough for logic correctness but I'd want to see the Postgres run pinned in CI given `EncryptedJSON` behaves differently on the two backends.

## [DESIGN NOTES] (not blocking)

**Lazy seed's side-effect on other org members.** `list_agent_profiles` seeds from the acting member's *composed* settings and sets `profiles.active = str(saved.id)`, so every other member of the org starts resolving that same profile from their next `SaasSettingsStore.load()`. `mcp_config` is member-private (`MEMBER_PRIVATE_AGENT_KEYS = {'mcp_config'}`), so `mcp_server_refs=None` filters against each member's own MCP config — good. But `skills`, `system_message_suffix`, `condenser`, `verification`, `enable_sub_agents`, `tool_concurrency_limit` are **not** member-private, so the seed captures whichever member happened to trigger it first. In practice their values match the org-wide shared state at that moment, so it's typically a no-op; still, the "first call wins" property is worth calling out for support ergonomics if a customer ever asks "why does my org have member X's skills as the default profile?"

## [RISK ASSESSMENT]

Schema migration + auth-adjacent settings-resolution + a new HTTP surface. Mitigating factors are strong: the migration is additive/reversible, the SDK substrate is imported (not forked), fallbacks are fail-safe (a stale pointer or dangling ref never bricks the settings load, it degrades to composed settings), tests exercise real code paths, and canvas already keeps the cloud path gated off, so no user-visible cloud regression is possible until canvas un-gates. Not a candidate for auto-merge — a human reviewer should confirm the DELETE-404 and `X-Expose-Secrets` calls before or in parallel with the canvas un-gating PR.

## Verdict

⚠️ **Worth merging with follow-ups** — the two wire-contract items above are recoverable and canvas gates the surface off today, but they should be resolved (fix, or an explicit deviation doc) before or at the same time as the canvas un-gating PR, otherwise the cloud Edit UX will silently be worse than local. Everything else is in good shape and the thin-boundary contract is genuinely respected.

## Key insight

The domain logic is faithfully imported from the SDK substrate; the friction points that remain are all at the two edges — the HTTP wire (DELETE semantics, `X-Expose-Secrets`) and the seed's side-effects on other org members. Fix those and the "same resource, one level up from LLM profiles" story holds end-to-end.

---

>
> 2. Re-request a review — the reviewer reads guidelines from the PR branch, so your changes take effect immediately.
>
>

simonrosenberg:
Pushed a fix commit (8bde22389) addressing the findings from an xhigh-effort automated code review of this PR:

1. **Settings contamination** (`saas_settings_store.py`) — `store()` was broadcasting a resolved active Agent Profile's LLM/api_key/mcp_config into org-level defaults and every other org member whenever the caller round-tripped `load()` → `store()` for an unrelated field (e.g. a routine settings PATCH). Now skipped when `item.active_agent_profile_id` is set.
2. **Unhandled 500** (`auth_user_context.py`) — `get_user_info`'s override branch asserted `settings is not None`; a reachable `None` load now degrades to `UserInfo(id=user_id)`, matching `ResolverUserContext`.
3. **Seed-race null pointer** (`agent_profiles.py` + `saas_settings_store.py`) — the one-time default-profile seed only set the *winning* member's `active_agent_profile_id`; every other member was stuck with a null active profile forever. The seed now also sets the org-wide `AgentProfiles.active` pointer, and `_resolve_active_agent_profile` falls back to it.
4/5. **Double fetch + provenance risk** (`live_status_app_conversation_service.py`) — conversation start called the override-aware `get_user_info()` twice (once for profile provenance, once for the ACP tag), doubling settings resolution and risking a mismatch if the profile changed in between. The ACP tag now reuses the already-resolved `profile_user`.
6. **Dead duplicate override** (`default_user_auth.py`) — removed `get_user_settings`, which duplicated the base `UserAuth` implementation verbatim.
7. **Duplicate construction** (`auth_user_context.py`) — factored the two `UserInfo`-building branches into one helper.
8. **Redundant import alias** (`live_status_app_conversation_service.py`) — dropped an unnecessary `MCPConfig as _MCPConfig` re-import.

Added regression tests for OpenHands/OpenHands#1 and OpenHands/OpenHands#3 (the two genuine correctness bugs) — verified each fails without its fix and passes with it. Full `test_agent_profiles.py` + `test_saas_settings_store.py` (55 tests), `test_live_status_app_conversation_service.py` (124), and the settings/profiles/user-context suites (128) all pass; ruff/ruff-format/mypy pre-commit hooks are clean.

simonrosenberg:
Addressed the code-review findings in `3b4516691`:
- `DELETE /api/agent-profiles/{name}` is now **idempotent** (200 on missing name) — matches the ts-client `AgentProfilesClient` and the local agent-server.
- `AgentProfileDetailResponse.profile` is now the SDK **`AgentProfile` discriminated union** (not `dict[str, Any]`).
- **`X-Expose-Secrets`**: documented that cloud intentionally always masks `mcp_tools` secrets (mirroring the LLM-profile cloud GET); edits stay non-destructive via the existing save-time mask-restore. Added a "Cloud deviations" note to the PR description.

Also added a regression test for idempotent delete and noted the lazy-seed first-caller capture. CI is green. Ready for another look.

simonrosenberg:
@openhands Please re-run `/codereview` on this PR at the current head (`3b4516691`), again cross-referencing agent-canvas branch `feat-agent-profiles-local` (the client) and the `openhands-sdk==1.30.0` `openhands.sdk.profiles` substrate.

The three findings from your previous review have been addressed — please verify each against the canvas wire contract:
1. **`DELETE /api/agent-profiles/{name}` is now idempotent** (200 on a missing name), matching the ts-client `AgentProfilesClient` and the local agent-server `delete_profile`; the per-member pointer clear still runs when the profile existed. Regression test added.
2. **`AgentProfileDetailResponse.profile` is now the SDK `AgentProfile` discriminated union** (not `dict[str, Any]`), so the OpenAPI schema matches the ts-client.
3. **`X-Expose-Secrets`**: documented (handler docstring + a "Cloud deviations" section in the PR body) that cloud intentionally always masks `skills[].mcp_tools` secrets — mirroring the cloud LLM-profile GET (`org_profiles.get_profile`) — and that edits stay non-destructive via the existing save-time mask-restore.

CI is green. Post your verdict with `gh api` using event `approve` if it now looks clean, or `comment` / `request_changes` as you see fit.

simonrosenberg (commented):
> _This review was posted by an AI agent (OpenHands) on behalf of @simonrosenberg. Repos consulted: `OpenHands/OpenHands@feat-agent-profiles-cloud` (head `3b4516691`), `OpenHands/agent-canvas@feat-agent-profiles-local` (the client), and the pinned `openhands-sdk==1.30.0` substrate for `openhands.sdk.profiles`._

## 🟢 Taste rating — Good taste

Re-review at `3b4516691`. All three findings from the previous pass are resolved, and I verified each one against the canvas wire contract and the SDK substrate directly (not just the PR description). The thin-boundary contract is still intact — every resolve / FK / seed / id-stamp / redaction call is imported from `openhands.sdk.profiles`.

---

## Verification of the three findings

### 1. `DELETE /api/agent-profiles/{name}` is now idempotent (200 on missing name) ✅

`delete_agent_profile` (`enterprise/server/routes/agent_profiles.py`) drops the pre-check 404 and now:
- captures `deleted_id` under `contextlib.suppress(FileNotFoundError)`,
- calls `profiles.delete(name)` **unconditionally** (the container's `AgentProfiles.delete` is itself a no-op for a missing name — `if entry is None: return`), and
- runs the per-member `active_agent_profile_id` clear only when the profile actually existed (`if deleted_id is not None`).

Cross-checked against the client: `agent-canvas` `use-delete-agent-profile.ts` carries the comment _"Delete is idempotent server-side (missing name resolves 200, not 404)"_ and its mutation has no 404 branch, so a repeat/stale-list delete no longer surfaces a red toast on cloud. Regression test `test_delete_missing_is_idempotent_200` pins it (asserts no raise + response shape). The pointer-clear-on-existing path is separately covered by `test_delete_clears_other_members_active_pointer`.

### 2. `X-Expose-Secrets` — documented cloud deviation ✅

`get_agent_profile` always masks `skills[].mcp_tools`; the intentional divergence is now explicit in both the handler docstring (_"Cloud always masks — the `X-Expose-Secrets` header is not honored here … mirroring the LLM-profile GET in `org_profiles`"_) and the PR body's **Cloud deviations** section. This was one of the two resolutions the previous review explicitly accepted. The non-destructive-edit guarantee is real for the **Edit** path: `_restore_masked_mcp_server` repairs only the individual masked server from the stored namesake, leaving newly-added siblings untouched — proved by `TestMaskedMcpSecretRestore.test_new_server_survives_save_alongside_masked_server`.

One caveat worth carrying into the un-gating PR (non-blocking — see below).

### 3. `AgentProfileDetailResponse.profile` is the SDK discriminated union ✅

The field is now typed `profile: AgentProfile` (imported from `openhands.sdk.profiles`), which I confirmed at runtime is the tagged `Union[OpenHandsAgentProfile, ACPAgentProfile]` with a `Discriminator`. I verified the handler's `AgentProfileDetailResponse(name=name, profile=profile.model_dump(mode='json'))` round-trips cleanly — a dumped (and masked) dict re-validates back into the correct concrete member, so the generated OpenAPI schema now matches what the ts-client consumes. The lifecycle test asserts attribute access (`detail.profile.agent_kind`, `detail.profile.llm_profile_ref`) rather than dict indexing, which only compiles against the typed model.

---

## [IMPROVEMENT OPPORTUNITIES] (non-blocking, for the canvas un-gating PR)

**Duplicate-on-cloud persists the mask, where Edit does not.** The PR body says _"Editing stays non-destructive: `save_agent_profile` restores any masked `mcp_tools` value from the stored blob."_ That is true for **Edit** (same name → the stored namesake exists → restore fires). It is **not** true for **Duplicate**: `agent-profiles-manager.tsx` `handleDuplicate` fetches with `exposeSecrets: "encrypted"` and immediately `saveProfile`s under a fresh `-copy` name via `toSaveInput` (which includes `skills`). On cloud the GET returns `<redacted>`, and because the save target name is new, `_restore_masked_skill_secrets` finds no namesake (`existing is None`) and returns the profile unchanged — so the copy would persist `<redacted>` as the literal `mcp_tools` env/header values.

This does not affect anything today: `use-settings-nav-items.ts` gates `/settings/agents` off for `backend.kind === "cloud"`, so the manager (and thus Duplicate) is unreachable on cloud. It is a direct, expected consequence of the documented `X-Expose-Secrets` deviation. Flagging only so whoever un-gates the cloud path either (a) wires encrypted secrets through GET, or (b) scopes the "non-destructive" claim to Edit and handles Duplicate explicitly. No change requested in this PR.

## [TESTING GAPS]

None. Real code paths against a real SQLite `Org` row, both success and error branches, plus the two regression tests tied to findings #1 and #3. The QA bot separately exercised CRUD / materialize / activate / FK-delete-409 / rename-cascade through real HTTP against the PR code.

## [RISK ASSESSMENT]

- [Overall PR] ⚠️ Risk Assessment: 🟢 **LOW → 🟡 borderline**

Down from the previous MEDIUM now that both wire-contract divergences are resolved. Residual factors are the additive/reversible migration 128 and the auth-adjacent `SaasSettingsStore` resolution — both fail-safe (a stale pointer or dangling ref degrades to composed settings rather than bricking the load), the SDK substrate is imported not forked, and the cloud surface remains gated off in canvas so no user-visible cloud regression is possible until the un-gating PR lands. CI is green on `3b4516691` (Python tests, enterprise image build, both migration jobs incl. pg8000, package/version consistency).

## Verdict

✅ **Worth merging.** The three findings are correctly and minimally resolved, verified against the canvas client and the `1.30.0` substrate. The one remaining note (Duplicate secret masking) is non-blocking, gated off today, and a known consequence of the documented deviation — carry it into the un-gating PR.

## Key insight

The domain logic stays faithfully imported from the SDK; the fixes land exactly where they should — the HTTP wire edge (idempotent DELETE, typed union response) — and the last remaining seam (`X-Expose-Secrets` masking) is now an explicit, documented cloud deviation rather than a silent divergence.

---

>
> 2. Re-request a review — the reviewer reads guidelines from the PR branch, so your changes take effect immediately.
>
>

simonrosenberg:
Thanks for the re-review (🟢 Worth merging). On the one non-blocking note — **Duplicate-on-cloud persists the mask** — you're right, and it's a real gap in the `handleDuplicate` path (fresh name → no namesake → `_restore_masked_skill_secrets` no-ops → `<redacted>` would persist as the literal secret).

Per your recommendation I've scoped the claim rather than change behavior in this PR (you explicitly requested no code change, and the surface is gated off on cloud today):
- The PR body's **Cloud deviations** section now scopes "non-destructive" to **Edit** (same-name, namesake restore) and adds an explicit **Known limitation — Duplicate** bullet, deferring the fix to the canvas un-gating PR (which will either round-trip encrypted secrets through `GET` or handle Duplicate explicitly).
- The `get_agent_profile` handler docstring already scopes this to "Edits," so it stays accurate.

Carrying the Duplicate handling into the un-gating PR as you suggested.

simonrosenberg:
@openhands Please re-run `/codereview` at the new head (`9bb817a20`), cross-referencing agent-canvas branch `feat-agent-profiles-local` and the `openhands-sdk==1.30.0` `openhands.sdk.profiles` substrate.

Since your last review (`3b4516691`, 🟢 Worth merging), one commit addresses the non-blocking **Duplicate-on-cloud** note you raised:
- `save_agent_profile` now reconciles masked `skills[].mcp_tools` **per value**: restore from the stored namesake when one exists (Edit), else **drop** the unrestorable `<redacted>` mask (Duplicate / create under a new name) — so the mask degrades to an empty secret instead of persisting as a literal, mirroring the LLM `**********` sentinel the SDK nullifies to `None`. The Edit-restore path is unchanged; real values sent alongside a mask are kept.
- Regression test `test_duplicate_under_new_name_drops_unrestorable_mask`; the PR body's Cloud-deviations bullet is updated to reflect the fix (no longer deferred).

Note only `skills[].mcp_tools` carries an inline secret — `llm_profile_ref` / `mcp_server_refs` are by-name references. Post your verdict with `gh api` (`approve` / `comment` / `request_changes`) as you see fit.

simonrosenberg (commented):
## `/codereview` re-review at `9bb817a20`

Cross-referenced against the `openhands.sdk.profiles` substrate in `openhands-sdk==1.30.0` (pinned in `enterprise/poetry.lock`) and the agent-canvas `feat-agent-profiles-local` Duplicate flow.

**Taste Rating:** 🟢 **Good taste**

The single commit since my last pass (`3b4516691`) does exactly one thing and does it well: it turns the masked-secret reconciliation from an all-or-nothing early return into a **per-value** decision, which eliminates the special case rather than adding a branch around it.

### What I verified

- **Root cause of the Duplicate bug is genuinely fixed.** The prior guard `if not isinstance(new_profile, OpenHandsAgentProfile) or not isinstance(existing, OpenHandsAgentProfile): return new_profile` bailed out whenever `existing is None` (a save under a new name), so the fetched `<redacted>` mask was persisted verbatim as a literal secret. The new code treats a missing namesake as an empty `{}` and reconciles each value: keep a real value the client sent, restore a masked value from the stored namesake (Edit), else **drop** it (Duplicate). The mask degrades to an empty secret instead of corrupting one — the same shape as the LLM `**********` sentinel the LLM model nullifies to `None`. (`agent_profiles.py` `_restore_masked_mcp_server` / `_restore_masked_skill_secrets`.)
- **Edit path is untouched.** With a namesake present, `old_mapping` is populated and every masked entry restores exactly as before; the existing per-server restore round-trip test stays green.
- **Sentinel is grounded, not guessed.** I fetched `openhands/sdk/utils/redact.py` at tag `v1.30.0`: `REDACT_ALL_VALUES_KEYS = {"environment", "env", "headers"}` and `_redact_all_values(...)` returns the literal `"<redacted>"`. So the duplicated `MCP_REDACTED_VALUE = '<redacted>'` (line 77, unchanged) is correct.
- **The regression test is a real code path, not a mock assertion.** `test_duplicate_under_new_name_drops_unrestorable_mask` saves a profile, GETs it (asserting the `API_KEY` came back as `<redacted>`), re-saves the masked payload under a *new* name, then reads the row back through `load_agent_profiles(org).load('source-copy')` over a real SQLite async session (`patch_agent_routes` binds the actual `a_session_maker`; only org lookup is stubbed). It asserts the mask is neither persisted nor present. This fails if the behavior regresses.

### Non-blocking note (do not block merge)

- **[agent_profiles.py:219]** The reconciler iterates `('env', 'headers')`, but the SDK's `REDACT_ALL_VALUES_KEYS` also includes `"environment"`. For the MCP `mcpServers` schema this is a non-issue (servers carry `env`, not `environment`), so there is no real gap today — but if an inline `mcp_tools` server ever grew an `environment` key, GET would mask it and save would persist that mask. A one-line comment noting the intentional `env`/`headers`-only scope (or deriving the tuple from the SDK constant) would keep this honest if the schema drifts. Optional.

### [RISK ASSESSMENT]
- [Overall PR] ⚠️ Risk Assessment: 🟢 **LOW**

The delta is ~70 lines confined to two pure helper functions plus a test. No schema, auth, or migration change in this commit; default behavior (no mask present, real values) is preserved; the only behavioral change strictly reduces the chance of writing a literal `<redacted>` to the encrypted column. Secrets remain inside the `org.agent_profiles` EncryptedJSON boundary and never widen on the wire.

**VERDICT:** ✅ **Worth merging** — the one non-blocking Duplicate note from the previous pass is now resolved with a correct, well-tested, minimal change.

**KEY INSIGHT:** Reconciling the mask per value (restore-or-drop) instead of per profile (all-or-nothing) removes the `existing is None` special case entirely and makes the Duplicate and Edit paths fall out of the same loop.

---

_This review was generated by an AI agent (OpenHands) on behalf of @simonrosenberg._

---

> **Improve this review?** If any feedback above seems incorrect or irrelevant to this repository, add or edit `.agents/skills/custom-codereview-guide.md` on your branch with the `/codereview` trigger and the missing context, then re-request a review.
>
> **Resolve with AI?** Install the [iterate skill](https://github.com/OpenHands/extensions/tree/main/skills/iterate) and run `/iterate` to drive this PR through CI, review, and QA until it's merge-ready.
>

simonrosenberg:
@openhands Please re-run `/codereview` on this PR at the current head (`6d1eda054`), again cross-referencing the canvas client (agent-canvas branch `feat-agent-profiles-minimal`, PR OpenHands/agent-canvas#1571) and the pinned `openhands-sdk==1.30.0` `openhands.sdk.profiles` substrate.

Your last verdict (🟢 Worth merging, at `9bb817a20`) predates two commits that deserve independent scrutiny:

1. **`39bcbc7f6` — structural split of the resolved vs persisted settings views** (~590 lines across the settings layer). A deep review found that plain `SaasSettingsStore.load()` returning the profile-RESOLVED view let any load()→store() round-trip corrupt persisted state: a routine settings PATCH wholesale-replaced the member's `mcp_config` with the profile's ref-FILTERED config (`mcp_server_refs=["a"]` deleted the member's other servers; `[]` wiped the config), rewrote `org_member.llm_api_key` from the referenced LLM profile's key, and the old `shared_agent_settings_diff = {}` guard silently dropped every org-level agent-settings edit. The fix: resolution is now an explicit opt-in (`load(resolve_agent_profile=True, ...)`, implied by `override_agent_profile_id`) requested only by the conversation-start builders and the provenance stamp; plain loads return persisted settings; resolved views are marked (`_resolved_view` PrivateAttr) and `store()` raises on them; the `shared = {}` special-case is removed. Also: `_resolve_active_agent_profile` catches `Exception` (never-brick contract, pre-absorbs the SDK-1.31 required-kwarg TypeError), and `_agent_profiles_transaction` writes the profiles blob back only when actually mutated (`/activate` is pointer-only; a schema-drift-invalid profile can no longer be erased by an unrelated mutation). Regression tests fail without each fix.

   **Two deliberate contract changes to verify against the canvas client:** `GET /api/settings` now returns persisted (not profile-resolved) `agent_settings` — canvas derives ACP-ness from the active profile, not settings, so it should be unaffected; and `store()` hard-fails (`ValueError`) on a resolved view rather than silently accepting it.

2. **`6d1eda054` — merge of `main`** with two conflict resolutions (union of the provenance fields with main's marketplace-registration fields in `settings_models.py`; import union in `saas_settings_store.py`) and a silent migration collision fixed by renumbering ours `129 → 131` (revises 130; single alembic head re-verified).

Known accepted non-blockers (tracked in OpenHands/enterprise#51, no code change requested): first-caller seed capture, zero-profiles re-seed semantics, override-launch double-resolve.

Post your verdict with `gh api` (`comment` / `request_changes` as you see fit — self-approval will 422 as before).

simonrosenberg (commented):
## `/codereview` re-run @ `6d1eda054` — delta since the `9bb817a20` 🟢 verdict

**Taste Rating: 🟢 Good taste** — the resolved/persisted split is exactly the elegant fix: it deletes a whole class of special-cases (`shared_agent_settings_diff = {}`, the store-time broadcast guard) by making the resolved view a *typed, opt-in, non-persistable* thing rather than something `store()` has to defensively detect and neutralize. Re-verified end-to-end against the canvas client (`feat-agent-profiles-minimal`, #1571) and the `openhands-sdk==1.30.0` `openhands.sdk.profiles` substrate.

I scoped this pass to the two commits that post-date my last review, plus a regression sweep of the launch path (since the change flips resolution from always-on to opt-in).

---

### 1. `39bcbc7f6` — resolved vs persisted view split

**The load()→store() corruption class is genuinely closed, at the right layer.** Plain `SettingsStore.load()` now returns the persisted composed settings; resolution is opt-in via `resolve_agent_profile=True` (implied by `override_agent_profile_id`), threaded cleanly through the abstract `SettingsStore` / `UserAuth` / `UserContext` contracts with keyword-only args and `None`/`False` defaults, so `FileSettingsStore` (OSS) correctly ignores both. The three findings (F1 ref-filtered `mcp_config` overwrite, F2 `llm_api_key` rewrite, F3 dropped org-level agent-settings edits) all fall out of the single structural change rather than three point-patches — good taste.

**Regression risk of making resolution opt-in — checked, clean.** The real danger of this design is a launch/agent-build path that still reads a *plain* load and now silently launches with non-profile settings. I audited every `get_user_info()` / `get_user_settings()` caller:
- Agent build (`live_status_app_conversation_service.py:1658` and `:2005`) and the provenance stamp (`:515`) all pass `resolve_agent_profile=True`; the resolved `user.agent_settings` is what flows into both the ACP and OpenHands builders. ✅
- The two remaining *plain* callers are correct: `:265` reads only `sandbox_grouping_strategy`, and `:1009` (`_sync_llm_profiles_to_sandbox`) uses the member's *ambient* `agent_settings.llm.api_key` as the LLM-profile fallback key — which is more correct on the persisted view than the resolved agent-profile's referenced key. ✅

**The `store()` guard is sound and fails fast.** Refusing on `_resolved_view or active_agent_profile_id is not None` covers both the live `PrivateAttr` (survives `model_copy`) and the reconstructed-copy path (the persisted signal). The only reconstruction case that slips past both guards is a *fallback* resolved view (resolver returned `None` → `active_agent_profile_id` unset, `agent_settings` == composed) — and that view is byte-identical to persisted settings, so accepting it is harmless. Raising `ValueError` instead of silently no-op'ing is the right call.

**F5 broad `except Exception`** — I flagged this shape reflexively (broad excepts hide bugs), but it's justified here: the never-brick contract for the ambient settings load, it logs the exception at WARNING with context, and `CancelledError`/`KeyboardInterrupt` derive from `BaseException` so cancellation still propagates. Non-blocking.

**F4 write-back-only-when-changed** (`_agent_profiles_transaction`, `rename_profile`) — the `before/after` JSON-dump compare is a clean way to keep a best-effort load (which drops unparseable entries) from erasing a stored-but-unparseable profile via an unrelated `/activate`. The double `model_dump` cost is negligible for a profiles blob.

**Tests exercise real code paths, not mocks.** `test_agent_profiles.py` adds one regression test per fix (`..._preserves_member_mcp_config` F1, `..._keeps_member_llm_key` F2, `..._edit_persists_while_profile_active` F3, `..._store_refuses_it` safety-net, `..._falls_back_to_composed_settings` F5, `..._preserves_unparseable_profile` F4, `..._survives_dry_run_crash`). They drive a real `SaasSettingsStore` against a real session-maker with real `Org`/`OrgMember` rows, mutate, `store()`, then re-read the member row and assert on persisted state — they'll actually fail if the behavior regresses.

### 2. `6d1eda054` — merge of `main`

- Migration renumber `129 → 131` (`revises 130`) is correct: I recomputed the revision graph and **`131` is the sole head** (`comm -23 revs downs` → `'131'`). Up/down chain is trivial `add_column`/`drop_column`, and CI's "Apply migrations on test DB" passes on both the default driver and pg8000.
- `settings_models.py` conflict resolution is a true **union**: the provenance fields (`active_agent_profile_id`, `active_agent_profile_revision`, `_resolved_view`) and main's `registered_marketplaces` all survive. `saas_settings_store.py` import union drops the now-unused `DanglingMcpServerRef`/`ProfileNotFound` (confirmed no remaining references) since the resolver's `except` widened to `Exception`.

### Cross-reference with the canvas client (contract changes)

Both deliberate contract changes are **aligned with — and in fact anticipated by** — the canvas client:

1. **`GET /api/settings` returns persisted (not profile-resolved) `agent_settings`.** `useActiveAgentKind()` reads the active profile from `/api/agent-profiles` (`use-active-agent-profile.ts`), and its docstring states outright: *"activate is pointer-only and never writes settings.agent_settings, so the active profile — not the global agent settings — is the source of truth."* `useAcpModelContext` only falls back to `settings.agent_settings.agent_kind` during the profile-list loading window (to avoid a UI flash), never as the authority. Canvas is unaffected. ✅
2. **`store()` hard-fails on a resolved view.** Canvas manages profiles exclusively through the agent-profiles endpoints and never round-trips a resolved view through the settings save path; the settings field it would echo (`active_agent_profile_id`) now comes back `null` from a plain GET, so the guard can't be tripped from the client. ✅

### Known non-blockers (tracked in #15112)

First-caller seed capture, zero-profiles re-seed semantics, and the override-launch double-resolve (provenance stamp + builder each resolve once when an override bypasses the memo) are all acknowledged and appropriately deferred. Agreed these are non-blocking.

---

**[RISK ASSESSMENT]**
This commit touches the shared settings load/store path for *all* users (not just profile users) and adds a new DB migration, so blast radius is org-wide. Mitigating factors are strong: the change narrows behavior (plain loads now do *less*), the never-brick fallback degrades to composed settings on any resolver failure, resolution is opt-in and covered by real round-trip regression tests, the launch-path caller audit came back clean, the migration is additive/nullable with a verified single head, and full CI (Enterprise unit tests, both migration drivers, lint, `check-package-versions`) is green. No secrets cross the wire that didn't before. Not auto-merge-blocking, but a human maintainer should give the settings-layer diff a second set of eyes given the shared-path reach.

**VERDICT:** ✅ **Worth merging.** Both post-`9bb817a20` commits stand up to independent scrutiny; the settings-view split is a net simplification that closes a real corruption class without breaking the canvas contract. (Posting as `COMMENT` — GitHub 422s self-approval since this token is the PR author.)

**KEY INSIGHT:** Modeling the resolved launch view as a typed, opt-in, non-persistable variant — rather than a value `store()` must defensively sniff for — is the "eliminate the special case" move that makes the whole settings round-trip correct by construction.

---

> **Improve this review?** If any feedback above seems incorrect or irrelevant to this repository, add/edit `.agents/skills/custom-codereview-guide.md` on this branch with the `/codereview` trigger and the missing context, then re-request review. See the [customization docs](https://docs.openhands.dev/openhands/usage/use-cases/code-review#customization).
>
> Was this review helpful? React with 👍 or 👎.

_This review was generated by an AI agent (OpenHands) on behalf of @simonrosenberg._

simonrosenberg:
Removed the embedded-skill secret mask/restore/duplicate-degradation machinery now that [software-agent-sdk#4018](https://github.com/OpenHands/software-agent-sdk/pull/4018) (closes [#4017](https://github.com/OpenHands/software-agent-sdk/issues/4017)) drops `OpenHandsAgentProfile.skills` entirely — the only field this apparatus existed to protect. Deleted `_restore_masked_mcp_server`, `_extract_mcp_server_map`, `_restore_masked_skill_secrets_raw`, and the now-vestigial `cipher` params on `AgentProfiles.save`/`load` (the SDK's `AgentProfileStoreProtocol` no longer declares `cipher` either, since a profile now has zero secret-bearing fields).

**Validated against the real SDK branch** (not mocked): installed it as editable (`uv pip install -e`) into both the root and `enterprise/` venvs, replacing the pinned `1.32.0` release, then ran the full `enterprise/tests/unit/test_agent_profiles.py` (36 tests) and `test_saas_settings_store.py` (27 tests) suites — real async DB transactions, real route functions, real SDK validation/resolution code. All 63 pass. `pre-commit` (ruff, mypy) is green on the changed files.

Net: -305/+23 lines. This needs to land before/alongside merge so the ~150-line mask/restore apparatus this PR was building is never shipped — see the sequencing note in software-agent-sdk#4017.

enterprise/migrations/versions/132_add_agent_profiles.py
hieptl, line 1: **suggestion (blocking):** We already have migration file **131** in place. Please update this file. Thank you very much! 🙏

hieptl (changes requested):

enterprise/migrations/versions/132_add_agent_profiles.py
simonrosenberg: Good catch — main picked up migration 131 (user login timestamps, #15148) after this branch cut its own 131. Renumbered to 132 and rechained `down_revision` onto 131 in 4854fa78d. Verified locally: single alembic head, `check_enterprise_migration_integrity.py` passes, and full `upgrade head` + `downgrade -1`/`upgrade head` round-trip against a real Postgres 16 container succeeds.

hieptl (approved):
I've approved this pull request—the code looks good to me. I haven't had a chance to test it myself, so I'm assuming you've already validated the feature on your end.

Thank you very much! 🙏

VascoSch92:
## High-Effort Code Review

**Summary:** Found 8 distinct code-safety defects across profile management, MCP validation, and settings subsystems—2 confirmed, 6 plausible.

### 🔴 Confirmed Issues

**1. Unhandled exceptions in profile seeding flow**
- **Location:** `enterprise/server/routes/agent_profiles.py:395`
- **Issue:** `build_seed_profile` and `save_profile_preserving_identity` can raise unhandled exceptions that propagate to `list_agent_profiles` endpoint, causing 500 errors instead of graceful degradation.
- **Impact:** Seeding flow failures produce 500 instead of proper error responses.

**2. Inconsistent exception handling in MCP config resolution**
- **Location:** `enterprise/storage/agent_profile_resolution.py:954`
- **Issue:** `member_mcp_config` catches only `ValidationError`, while `_resolve_active_agent_profile` catches all exceptions from `coerce_mcp_config`. Creates asymmetric error paths.
- **Impact:** Non-ValidationError exceptions propagate as 500 in one path but are silently handled in the other.

### 🟡 Plausible Issues

**3. Unhandled ValueError in UUID conversion** (`agent_profiles.py:316`)
- `UUID(user_id)` without try-except will throw 500 on invalid UUID from corrupted auth state.

**4. Changed LLM profile deletion error handling** (`org_profiles.py:777`)
- Refactored to use `delete_llm_profile()` which only catches `ProfileReferenced`; other SDK exceptions now propagate as 500 instead of 404.

**5. Removed early-return guard for ACP profiles** (`live_status_app_conversation_service.py:2695`)
- MCP config merging now attempted for ACP agents; could fail if ACP's MCP schema differs from OpenHands.

**6. Resolved settings cache not cleared** (`user_auth.py:101`)
- `UserAuth._resolved_settings` cached without invalidation; reused instances could return stale profiles if active profile changes.

**7. Removed assertion on null user settings** (`auth_user_context.py:3202`)
- Silently degrades to bare `UserInfo` when settings load fails, masking underlying failures.

**8. Changed Settings.load() signature** (`settings_store.py:3126`)
- Keyword-only parameters with defaults mean callers that don't pass `resolve_agent_profile=False` silently get persisted view instead of resolved.

### ✅ Refuted
- Two initially suspected issues in `saas_settings_store.py:1219` and `default_user_auth.py:3264` did not survive verification.

simonrosenberg:
Thanks for the thorough review. Addressed in 584060819.

## Confirmed findings — fixed

**#1 — Seed failure could 500 `list_agent_profiles`** (`agent_profiles.py`)
`_seed_default_agent_profile` (→ `build_seed_profile` / `save_profile_preserving_identity` / `SaasSettingsStore.load`) can raise, and lazy seeding is a one-time migration that must not brick the listing. Wrapped the seed call so a failure logs and degrades to the empty list — the same never-brick contract the resolution path already honors. The seed transaction commits only on clean exit, so no partial state leaks. New test: `test_seed_failure_degrades_to_empty_list`.

**#2 — Asymmetric exception handling in MCP resolution** (`agent_profile_resolution.py`)
`member_mcp_config` caught only `ValidationError`, but `coerce_mcp_config` can also raise on fastmcp normalization / contract drift (e.g. `TypeError`) — and it feeds `/materialize`. Broadened to `except Exception` to match `_resolve_active_agent_profile`'s inline coerce. New test: `test_member_mcp_config_degrades_on_non_validation_error`.

## Plausible findings — verified by-design / not regressions (no change)

- **#3 `UUID(user_id)`** — mirrors the existing `org_profiles.py` pattern; `user_id` is a Keycloak subject UUID guaranteed by the auth dependency. A malformed value is corrupted auth state (a genuine 500), not a client-handleable case; guarding only here would diverge from the sibling router.
- **#4 LLM-profile deletion** — `LLMProfiles.delete` returns `bool` (never raises on missing), and `delete_profile` guards with `has()`→404 first. `delete_llm_profile` only *adds* the `ProfileReferenced`→409 FK check; both old and new 500 identically on truly unexpected errors. No 404→500 regression.
- **#5 Removed ACP early-return in `_merge_custom_mcp_config`** — intentional (SDK#3705 remainder / OpenHands/OpenHands#15044 §7): a resolved ACP profile's ref-filtered `mcp_config` rides on `ACPAgentSettings.mcp_config` and must reach `create_agent`. Both kinds share the same `dict[str, MCPServer]` schema. This is the fix, documented in the docstring.
- **#6 `_resolved_settings` memo not invalidated** — `UserAuth` is per-request (cached on `request.state`); activation is a separate request. Mirrors the long-standing `_settings` memo. No intra-request staleness.
- **#7 Removed `assert settings is not None`** — deliberate robustness improvement. The assert turned a legitimate `None` (user/org not loadable, already logged in `SaasSettingsStore.load`) into a 500; degrading to a bare `UserInfo(id=...)` is strictly safer and not silent.
- **#8 `Settings.load()` keyword-only signature** — persisted-view-by-default with `resolve_agent_profile=True` opt-in *is* the PR's central invariant: a resolved launch view must never round-trip back into persisted settings. Correct by design.

Verification: 39 tests in `test_agent_profiles.py` pass (+2 new); `test_saas_settings_store.py` green; ruff check+format and mypy clean on the changed files.

PR #15068, opened 2026-06-30
perf: cache user_authorizations rule set instead of querying on every auth check

aivong-openhands:
## Why

Closes OpenHands/enterprise#45 (follow-up to OpenHands/OpenHands#14633).

The `user_authorizations` table is tiny (~31 rows / a single ~8 KB page), so the Postgres planner will always choose a sequential scan over any index — adding indexes does **not** reduce load. The actual cost driver is **query volume**: the table is read on every authentication/authorization request through `UserAuthorizationStore.get_authorization_type`, which issues a query per request (millions of calls).

The effective fix is to stop hitting the DB on the hot path: cache the small, rarely-changing rule set in process memory behind a short TTL and evaluate the email/provider match in Python.

## What changed

`enterprise/storage/user_authorization_store.py`:
- Load the **full** `user_authorizations` rule set into a process-local cache with a short TTL (`USER_AUTHORIZATION_CACHE_TTL_SECONDS`, default `30s`).
- Evaluate the email/provider match in Python via `_matches_rule` + `_compile_like_pattern`, which faithfully replicates the SQL semantics:
  - `provider_type IS NULL` matches any provider, otherwise it must be equal.
  - `email_pattern IS NULL` matches any email, otherwise SQL `LIKE` applies (`%` → any sequence, `_` → single char, case-insensitive — mirroring `LOWER(email) LIKE LOWER(email_pattern)`).
- Invalidate the cache on `create_authorization` / `delete_authorization`.
- Reads that pass an explicit `session` still query the DB directly, so a caller's in-transaction (uncommitted) view is preserved. Only the hot path (no session) uses the cache.

No change to the public API or matching behavior.

## Effect

`seq_scan` / `calls` for `user_authorizations` in `pg_stat_user_tables` drop to roughly `(replicas × table_count) / TTL` instead of tracking auth request volume. Because writes are rare and the cache is invalidated on write, the staleness window is at most one short TTL on replicas that did not perform the write.

> Note: `blocked_email_domains` was already migrated into `user_authorizations` in migration `099`, so caching this single table covers both tables referenced in the issue.

## Tests

Added `TestAuthorizationCaching` in `enterprise/tests/unit/storage/test_user_authorization_store.py`:
- A repeated lookup within the TTL issues no new DB query (verified with a session-maker call counter).
- `create_authorization` and `delete_authorization` each invalidate the cache.
- An expired cache reloads from the DB.
- Added an autouse fixture to reset the process-global cache between tests.

Existing pattern-matching tests continue to exercise the same scenarios through the cached path.

### Validation note

The full enterprise pytest suite requires the heavy `openhands` SDK / cloud dependency chain, which is not installed in this environment. I validated the **actual** store code paths (LIKE matching incl. `_`/`%`/case-insensitivity/subdomain rejection, whitelist precedence, cache hits issuing no queries, create/delete invalidation, and TTL expiry reload) against an in-memory SQLite database using a standalone harness — all checks passed. `ruff check` passes on the changed files.

---

@aivong-openhands can click here to [continue refining the PR](https://app.all-hands.dev/conversations/f4472701-9dc0-4d70-ae0d-fa3352d6629e)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-4dd8772   docker.openhands.dev/openhands/openhands:4dd8772
```

### DIFF
--- a/enterprise/storage/user_authorization_store.py
+++ b/enterprise/storage/user_authorization_store.py
@@ -1,5 +1,9 @@
 """Store class for managing user authorizations."""
 
+import os
+import re
+import time
+from functools import lru_cache
 from typing import Optional
 
 from sqlalchemy import func, or_, select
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #15070, opened 2026-06-30
fix: set email person property in PostHog during onboarding completion

lilagrc:
## HUMAN:

Adds another call to set person property in Posthog, to fill in gaps when this is not being captured during the onboarding form flow. 

- [x] I have tested these changes locally and they work as expected.

### Problem

When a new user completes the onboarding form, the `person_display_name` in PostHog falls back to the Keycloak UUID instead of showing the user's email. This happens because of a gap in the analytics flow:

1. **Login callback**: `identify_user(email=...)` is called, but `user.user_consents_to_analytics` is `None` for new users → `consented=False` → the call is a **no-op**
2. **TOS acceptance**: Sets `consented=True` and fires `track_user_signed_up`, but does **not** re-trigger `identify_user` → email is never stamped on the person profile
3. **Onboarding completion**: `track_onboarding_completed` fires against the Keycloak UUID person, but that person has no `email` property set

### Fix

Add a `set_person_properties(email=user.email)` call in the `complete_onboarding` endpoint, right before `track_onboarding_completed`. This is the first analytics touchpoint after consent is granted where the user object (with email) is available.

- Guarded by `if user.email:` so no call is made when email is `None`
- Uses the same `ctx` (same distinct_id) as the onboarding event

### Tests

- Updated `test_fires_onboarding_analytics_with_selections` to verify `set_person_properties` is called with the user's email
- Added `test_skips_set_person_properties_when_no_email` to verify the guard when `user.email is None`

All 35 tests in `test_auth_onboarding.py` pass.

---

@lilagrc can click here to [continue refining the PR](https://app.all-hands.dev/conversations/1431a590-de4b-48d0-a09e-b8fe978a938a)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-5afb5be   docker.openhands.dev/openhands/openhands:5afb5be
```


hieptl (approved):
Thank you! 🙏

PR #15073, opened 2026-07-01
fix: prevent path traversal in LocalFileStore

Aarkin7:
HUMAN:

This PR fixes a path traversal vulnerability in LocalFileStore where paths containing "../" could escape the root directory and access arbitrary files on the host. The fix normalizes paths and validates they stay within the store root. Added tests to cover both malicious and legitimate path inputs.

AGENT:

---

## Why

LocalFileStore.get_full_path() only strips a leading / but does not validate against ../ directory traversal sequences. This means any caller can escape the intended root directory and read, write, or delete arbitrary files on the host filesystem. For example, get_full_path("../../etc/passwd") resolves to /etc/passwd when the store root is /data/store.

## Summary

- Normalize the resolved path with os.path.normpath and validate it remains within the store root, raising ValueError on traversal attempts
- Add tests covering rejected traversal vectors (../, ../../, foo/../../) and allowed safe paths (foo/bar/../baz, ./foo, empty string)

## Issue Number
N/A

## How to Test

uv run pytest tests/unit/app_server/file_store/test_file_store.py -v

All 23 tests should pass, including the two new ones: test_path_traversal_rejected and test_safe_paths_allowed.

## Video/Screenshots

N/A. This is a backend-only fix with no UI. Test output confirms correctness.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The fix is scoped to LocalFileStore only. S3, Google Cloud, and in-memory stores are unaffected since they don't resolve to filesystem paths.

### DIFF
--- a/openhands/app_server/file_store/local.py
+++ b/openhands/app_server/file_store/local.py
@@ -21,7 +21,11 @@ class LocalFileStore(FileStore):
     def get_full_path(self, path: str) -> str:
         if path.startswith('/'):
             path = path[1:]
-        return os.path.join(self.root, path)
+        full = os.path.normpath(os.path.join(self.root, path))
+        root_normalized = os.path.normpath(self.root)
+        if not (full == root_normalized or full.startswith(root_normalized + os.sep)):
+            raise ValueError(f'Path traversal detected: {path}')
+        return full
 
     def write(self, path: str, contents: str | bytes) -> None:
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #15079, opened 2026-07-01
chore: bump SDK and agent-server to 1.30.0

simonrosenberg:
HUMAN: Bumping the bundled SDK + sandbox agent-server to the freshly-published v1.30.0. Verified locally: full root unit suite (2048 tests, +1 new) and full enterprise unit suite (2540 tests) pass, root+enterprise mypy pre-commit hooks pass against the new pin, and ruff/pyproject-fmt/validate-pyproject all pass.

## What
- `openhands-sdk` / `openhands-agent-server` / `openhands-tools` → `==1.30.0` (PyPI) in both the PEP 508 `dependencies` array and `[tool.poetry.dependencies]`
- `AGENT_SERVER_IMAGE` → `ghcr.io/openhands/agent-server:1.30.0-python`; both compose defaults updated to match
- mypy pre-commit hook's `additional_dependencies` pin → `1.30.0` (was stale at `1.28.0`)
- Regenerated `uv.lock`, `poetry.lock`, `enterprise/poetry.lock` (registry source — no git refs); only the three SDK packages changed version, no transitive drift
- Two compatibility fixes required by the new release:
  - SDK 1.30.0's `PluginSource` now redacts credentials in its own `source` field_serializer (software-agent-sdk#3909). `PluginSpec` had a local duplicate of the same redaction as a stopgap (with a `TODO` calling out exactly this), which now collides with the inherited serializer and breaks class construction with `PydanticUserError: Multiple field serializer functions were defined for field 'source'`. Removed the local duplicate — SDK's `redact_url_credentials(..., preserve_placeholders=True)` covers it now.
  - SDK 1.30.0 added `is_subscription` as a computed field on `LLM`. It's serialized into the GET response for saved LLM profiles, so echoing it back on POST tripped `StrictLLM`'s `extra='forbid'`, breaking the frontend's GET→edit→POST flow (422 instead of 201). `StrictLLM` now strips it before validation, and — as a **temporary compensating control** — doesn't let a client set it to `true` either.

## Why this needs a workaround, not a one-line fix
The SDK's own restore mechanism for this computed field (`LLM._restore_is_subscription`, `mode="wrap"`) reads its own raw input directly, independent of any subclass `mode="before"` validator — so a normal before-validator can't neutralize it; the only way to override it is to shadow the exact private method name. Filed upstream: **OpenHands/software-agent-sdk#3942**.

This PR's fix only covers the LLM-profiles endpoint (`StrictLLM`). The same gap exists on the main `/api/v1/settings` endpoint (via `agent_settings_diff` → the bare SDK `OpenHandsAgentSettings.llm`, `extra='ignore'`, no equivalent guard) — confirmed by direct testing. That's out of scope for an SDK-bump PR and is tracked separately: **OpenHands/OpenHands#15080**. Today this is inert either way — grepping this repo turns up zero code paths that read `is_subscription` for billing/auth — so there's no live exploit, just a footgun (a client-set flag would push a normal API-key profile through the SDK's subscription-specific request formatting) and a consistency gap worth closing centrally rather than endpoint-by-endpoint.

## How to Test
- `uv run pytest tests/unit` — 2049 collected, all pass (excl. one pre-existing macOS-only `/tmp` tempdir assertion unrelated to this change, irrelevant on Ubuntu CI)
- `poetry run --project=enterprise pytest enterprise/tests/unit` — 2540 passed, 6 skipped
- `pre-commit run --all-files --config ./dev_config/python/.pre-commit-config.yaml` (root) — all hooks pass, including mypy against the new pin

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-8584e25   docker.openhands.dev/openhands/openhands:8584e25
```


openhands/app_server/app_conversation/app_conversation_models.py
```diff
@@ -83,11 +72,6 @@ class PluginSpec(PluginSource):
         description='User-provided values for plugin input parameters',
     )
 
-    @field_serializer('source')
```
simonrosenberg, line 86: Here's how I confirmed it was safe/necessary to
  drop:

  1. It was a hard failure, not a style choice. After bumping the SDK, importing this module raised PydanticUserError: Multiple field serializer functions were defined for field 'source'. PluginSpec extends the SDK's PluginSource, and once both
  the parent and the child declare a @field_serializer('source'), Pydantic refuses to build the class at all. So this wasn't "can we simplify" — the code couldn't even load with both present.
  2. The SDK now does the same thing natively. I checked the installed SDK's PluginSource (.venv/.../openhands/sdk/plugin/types.py) and it now has its own @field_serializer("source", when_used="always") that calls redact_url_credentials(source,
  preserve_placeholders=True) — same regex-based user:pass@host → ****@host redaction, plus it additionally preserves ${VAR} placeholders that OpenHands's version didn't handle.
  3. There was already a TODO pointing at exactly this. The local helper's docstring literally said: "replace with from openhands.sdk.utils.redact import redact_url_credentials once the SDK pin is bumped to include
  OpenHands/software-agent-sdk#2154." This bump is that moment.
  4. No other callers. I grepped openhands/ and tests/ for _redact_url_credentials and _serialize_source — the only references were the definition and its one use site inside PluginSpec. Nothing else depended on the local implementation.
  5. Verified behaviorally, not just by inspection. After removing it, I ran the tests that actually exercise PluginSpec/display_name (tests/unit/app_server/test_models.py, test_live_status_app_conversation_service.py) — all 138 passed,
  confirming source redaction still happens correctly via the inherited SDK serializer.

enterprise/poetry.lock
```diff
@@ -1,4 +1,4 @@
-# This file is automatically @generated by Poetry 2.3.2 and should not be changed by hand.
+# This file is automatically @generated by Poetry 2.2.1 and should not be changed by hand.
```
hieptl: **suggestion (blocking):**

**poetry.lock:1 / enterprise/poetry.lock:1** — The lock files were regenerated using an older version of Poetry (2.2.1 instead of 2.3.2). As a result, they include unrelated normalization changes and inadvertently remove some platform markers.

Specifically, the generator header changes from Poetry 2.3.2 to 2.2.1, introducing re-normalizations such as `<6.0.dev0→<6.0dev`, `>=14.5.14→>=14.05.14`, `2023.3.6→2023.03.6`, `<2.0a0→<2.0a.0`, and `2021.8.17b43→2021.8.17-beta.43`. It also removes the `runtime = "sys_platform != \"win32\" and sys_platform != \"emscripten\""` markers for `pexpect` and `ptyprocess` in `poetry.lock`, as well as the `cp310-abi3` `tree_sitter_c_sharp` wheel entries in `enterprise/poetry.lock`.

Since CI installs Poetry via `pipx install poetry` without pinning the version (which currently resolves to Poetry 2.3.x), the next `poetry lock` execution will reapply the newer normalization, resulting in recurring diff churn and potential merge conflicts. Additionally, removing the `pexpect`/`ptyprocess` platform markers changes the dependency graph, making this more than a formatting-only update.

To keep the changes focused and consistent with the PR's stated goal of "no transitive drift," I recommend regenerating the lock files with Poetry 2.3.2 so that the diff is limited to the three intended SDK version bumps.

Thank you very much! 🙏

hieptl (changes requested):

enterprise/poetry.lock
simonrosenberg: Fixed in 8584e25 — regenerated both lockfiles with Poetry 2.3.2 (installed via pipx to match what main's committed header already uses / what CI's unpinned `pipx install poetry` currently resolves to). The diff is now limited to the intended SDK version bump plus reverting the version-string re-normalization and restoring the pexpect/ptyprocess platform markers and the cp310-abi3 tree_sitter_c_sharp wheel entries. Thanks for the sharp catch!

openhands/app_server/app_conversation/app_conversation_models.py
simonrosenberg: Confirmed safe/necessary to drop, verified concretely:

1. It's a hard failure, not a style choice — after the SDK bump, importing this module raised `PydanticUserError: Multiple field serializer functions were defined for field 'source'`. `PluginSpec` extends the SDK's `PluginSource`, and once both the parent and the child declare a `@field_serializer('source')`, Pydantic refuses to build the class at all.
2. The SDK now does the same thing natively — its `PluginSource` (openhands/sdk/plugin/types.py) now has its own `@field_serializer("source", when_used="always")` calling `redact_url_credentials(source, preserve_placeholders=True)`, the same regex-based redaction, plus it additionally preserves ${VAR} placeholders that this local version didn't handle.
3. There was already a TODO pointing at exactly this: "replace with `from openhands.sdk.utils.redact import redact_url_credentials` once the SDK pin is bumped to include OpenHands/software-agent-sdk#2154." This bump is that moment.
4. No other callers — grepped `openhands/` and `tests/` for `_redact_url_credentials`/`_serialize_source`; only the definition and this one use site.
5. Verified behaviorally: ran `tests/unit/app_server/test_models.py` and `test_live_status_app_conversation_service.py` (the tests that exercise `PluginSpec`/`display_name`) — all 138 passed, confirming redaction still happens via the inherited SDK serializer.

openhands/app_server/settings/llm_profiles.py
```diff
@@ -106,6 +106,21 @@ class StrictLLM(LLM):
...
+        """Strip the SDK's read-only ``is_subscription`` computed field.
+
+        It's serialized into GET responses (so it round-trips through the
+        frontend's GET-edit-POST flow) but isn't a settable input: it only
+        becomes true via ``LLM.subscription_login()``, never via user-supplied
+        JSON, and letting a client set it here would let them self-declare
+        subscription billing.
+        """
+        if isinstance(data, dict) and 'is_subscription' in data:
+            data = {k: v for k, v in data.items() if k != 'is_subscription'}
```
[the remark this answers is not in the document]
simonrosenberg: Fixed in 6c4cc74 — restored blocking of client-supplied `is_subscription`. Root cause: the SDK's own `LLM._restore_is_subscription` (mode="wrap") reads its own raw input directly, independent of any subclass `mode="before"` validator, so a plain before-validator strip couldn't neutralize it (only shadowing the exact method name works — filed as OpenHands/software-agent-sdk#3942). Verified now: `StrictLLM.model_validate({..., 'is_subscription': True})` returns `is_subscription=False`, for `api_key` unset/empty/normal alike, while a genuinely unknown field still 422s.

simonrosenberg:
@hieptl I applied your requested changes!

hieptl (approved):
Thank you! 🙏

PR #15085, opened 2026-07-01
feat(api-keys): add optional active window (not_before & expires_at)

tofarr:
HUMAN:

Surfacing a date range within which a key is valid

AGENT:

---

## Why

Today an OpenHands API key is either valid forever or expires at a fixed
`expires_at` timestamp. There is no way to schedule a key to become
valid in the future (e.g. a "start at the beginning of next quarter"
key for a contractor, or a key that becomes active only after a service
migration window). This PR adds an optional `not_before` lower bound so
users can define an explicit active window
`not_before <= now < expires_at`. Both bounds are independently
optional, the change is fully backward-compatible (existing keys keep
their current behaviour), and authentication at the
`Authorization: Bearer …` / `X-Session-API-Key` / `X-Access-Token`
surface automatically rejects out-of-window keys.

## Summary

- Adds a new optional `not_before` column to the `api_keys` table
  (migration 128) and surfaces it through the SQLAlchemy model, the
  `ApiKeyStore.create_api_key` / `validate_api_key` methods, the
  `/api/keys` HTTP route, and the SaaS bearer auth flow.
- `validate_api_key` now rejects keys used before `not_before` or after
  `expires_at`; rejected keys never have `last_used_at` updated and the
  audit log distinguishes "not yet active" from "expired".
- Pydantic `field_validator`s reject past dates; a `model_validator`
  enforces `not_before < expires_at`.
- Frontend create modal gains two optional `datetime-local` inputs
  ("Active from" / "Expires at") with client-side validation that
  mirrors the backend.
- API keys table gains a Status column with `active` / `pending` /
  `expired` pills; pending and expired rows are dimmed.
- 9 new i18n keys (en + 14 locales) regenerated via `npm run make-i18n`.

## How to Test

1. Pull the branch and run `make build` (or `cd enterprise && poetry
   install --with dev,test` for backend-only).
2. Backend: `cd enterprise && PYTHONPATH=".:$PYTHONPATH" poetry run
   pytest tests/unit/test_api_key_store.py
   tests/unit/server/routes/test_api_keys.py
   tests/unit/test_saas_user_auth.py -p no:ddtrace` — all green.
3. Frontend: `cd frontend && npm run lint:fix && npm run build && npm
   run test` — all green; 10 new tests in
   `create-api-key-modal.test.tsx` and `api-keys-manager.test.tsx`.
4. Manual smoke: start the SaaS dev environment, navigate to
   `/settings/api-keys`, create a key with "Active from" set to ~1
   minute in the future. Observe that requests with that key return 401
   until the time elapses, then succeed. Repeat with "Expires at" set
   to 1 minute in the future and observe it stop working after.
5. Backward-compat: existing keys (with both fields `NULL`) continue to
   work exactly as today.

## Video/Screenshots

image

image

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Plan and design rationale live in `.pr/api-key-active-window-plan.md`
  (added in this PR; auto-removed on merge via
  `.github/workflows/pr-artifacts.yml`).
- The schema change is purely additive (`not_before NULL`); downgrade
  path drops the column. The `expires_at` test suite is unchanged and
  the new `not_before` tests are additive.
- The SaaS auth flow (`saas_user_auth_from_bearer`) required no code
  change because the rejection happens inside `validate_api_key`.
- A pre-existing stray edit to `frontend/src/routes/device-verify.tsx`
  (`useShouldHideOrgSelector() && false`) was reverted in this PR
  because it would have hidden the org selector dropdown.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-e4bd71d   docker.openhands.dev/openhands/openhands:e4bd71d
```


hieptl (approved):
Thank you! 🙏

PR #15088, opened 2026-07-01
fix: Alembic index creation for execution_status migration

Chesars:
HUMAN:

AGENT:

---

## Why

Migration `013.py` creates an index by passing `execution_status` as a string to `batch_op.create_index`.

Alembic expects the index columns argument to be a sequence/list of column names. Passing a string can cause it to be interpreted as an iterable of characters instead of a single column name, which can make the migration fail when creating the index.

Docs: https://alembic.sqlalchemy.org/en/latest/api/operations.html

## Summary

- Wrapped `execution_status` in a list when creating the index.
- Kept the existing index name and `unique=False` behavior unchanged.

## Issue Number

N/A

## How to Test

Run `make run` with a local database that apply migration `013.py`.

## Video/Screenshots

N/A

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/openhands/app_server/app_lifespan/alembic/versions/013.py
+++ b/openhands/app_server/app_lifespan/alembic/versions/013.py
@@ -34,7 +34,7 @@ def upgrade() -> None:
         # Create index for efficient dashboard queries
         batch_op.create_index(
             'ix_conversation_metadata_execution_status',
-            'execution_status',
+            ['execution_status'],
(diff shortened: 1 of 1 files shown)


Chesars:
Closing as a duplicate of #15069

PR #15091, opened 2026-07-01
refactor(enterprise/auth): source accepted_tos from User table

tofarr:
HUMAN:

AGENT:

---

## Why

The `keycloak_auth` cookie was carrying `accepted_tos` in its JWS payload, duplicating a value that is already the source of truth in `User.accepted_tos`. This adds ~30 bytes per cookie on top of the existing chunking workaround that exists to keep the (much larger) cookie under Chrome's 4096-byte cap.

## Summary

- New `UserStore.get_user_accepted_tos()` does a lightweight (no eager load, no migration fallback) `User` lookup and returns the tri-state boolean the rest of the request lifecycle expects.
- `SetAuthCookieMiddleware._check_tos` decodes the cookie's `sub` claim and looks up `User.accepted_tos` via the new helper instead of reading the (now-removed) JWS field.
- `saas_user_auth_from_signed_token` populates `SaasUserAuth.accepted_tos` from the DB instead of the cookie.
- `set_response_cookie` and all call sites stop accepting / passing `accepted_tos`.

## Issue Number

None.

## How to Test

- `cd enterprise && poetry run pytest tests/unit/test_user_store.py tests/unit/test_auth_middleware.py tests/unit/test_auth_routes.py tests/unit/test_saas_user_auth.py`
- Manual: log in via the browser, confirm the cookie is now ~30 bytes smaller, and verify a user with `User.accepted_tos IS NULL` is redirected to the TOS page.

## Video/Screenshots

N/A — backend-only change.

## Type

- [x] Refactor

## Notes

**Behavioural change:** TOS enforcement is now strict — a `NULL` `User.accepted_tos` raises `TosNotAcceptedError`. Previously the cookie-based check treated a missing flag as "allow" to support legacy users who had logged in before the TOS column existed; the column has existed since migration 010 in 2024, so this only affects a small number of pre-TOS-feature users who somehow still hold a valid signed cookie. The middleware still falls back to "allow" on a missing user row or DB error, so a misbehaving DB cannot log everyone out.

**Cost:** one lightweight `SELECT id, accepted_tos FROM user WHERE id = ?` per cookie-authenticated request that runs `_check_tos` (most `/api/*` and `/mcp` paths). No eager load, no migration fallback, no network.

**Next step (not in this PR):** the cookie still carries `access_token` + `refresh_token`, so the chunking workaround in `enterprise/server/auth/cookie_chunking.py` is still needed for users with large Keycloak claim sets. Removing the `refresh_token` from the cookie (since the offline token is also in `offline_tokens` and can be looked up by `user_id`) would let us delete `cookie_chunking.py` and its tests.

---

This PR was created by an AI agent (OpenHands) on behalf of tofarr.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-83fda85   docker.openhands.dev/openhands/openhands:83fda85
```

### DIFF
--- a/enterprise/server/auth/saas_user_auth.py
+++ b/enterprise/server/auth/saas_user_auth.py
@@ -782,6 +782,7 @@ async def saas_user_auth_from_cookie(request: Request) -> SaasUserAuth | None:
 async def saas_user_auth_from_signed_token(signed_token: str) -> SaasUserAuth:
     logger.debug('saas_user_auth_from_signed_token')
     from storage.encrypt_utils import get_jwt_service
+    from storage.user_store import UserStore
 
     decoded = get_jwt_service().verify_jws_token(signed_token)
     logger.debug('saas_user_auth_from_signed_token:decoded')
@@ -794,7 +795,6 @@ async def saas_user_auth_from_signed_token(signed_token: str) -> SaasUserAuth:
             'refresh_token': refresh_token,
         },
     )
-    accepted_tos = decoded.get('accepted_tos')
 
     # The access token was encoded using HS256 on keycloak. Since we signed it, we can trust is was
     # created by us. So we can grab the user_id and expiration from it without going back to keycloak.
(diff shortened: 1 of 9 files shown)


neubig:

Thank you for understanding.

PR #15095, opened 2026-07-02
fix: tolerate malformed pagination cursors in page_id_to_offset

VenkateswarluNagineni:
HUMAN:

This fixes a crash in the app server's pagination. If a client sends a bad or stale `page_id` cursor, the search helper tried to base64-decode it and `int()` the result with no guard, so the whole request failed with a 500 instead of simply starting from the first page. I changed it to fall back to offset 0 on a malformed cursor — the same way the other pagination helper (`paging_utils.decode_page_id`) already tolerates bad input — and added tests covering the malformed-input cases.

---

## Why

`page_id_to_offset` in `openhands/app_server/utils/search_utils.py` decoded a client-supplied
pagination cursor with `int(base64.b64decode(page_id).decode())` and **no error handling**. Any
malformed `page_id` therefore raised an exception that propagates as an unhandled **HTTP 500**:

- invalid base64 / bad padding → `binascii.Error`
- valid base64 whose payload isn't an integer → `ValueError`
- valid base64 that isn't UTF-8 → `UnicodeDecodeError`

The cursor is meant to be an opaque token, so a stale or garbled value from a client should degrade
gracefully rather than error. The sibling helper `paging_utils.decode_page_id` already wraps the
identical base64-decode + `int()` in a `try/except` and returns a safe sentinel on bad input — this
change makes `page_id_to_offset` consistent with that convention.

## Summary

- Wrap the decode in `page_id_to_offset` and fall back to offset `0` (the same "start from the
  beginning" behavior already used for a `None` cursor) on `ValueError` / `UnicodeDecodeError` /
  `binascii.Error`.
- Add parametrized regression tests covering each malformed-input path.

## Issue Number

N/A

## How to Test

```bash
pip install pytest pytest-asyncio
python -m pytest tests/unit/utils/test_search_utils.py -q
```

All existing tests plus the new `test_page_id_to_offset_malformed_returns_zero` cases pass. Valid
cursors are unaffected (`offset_to_page_id`/`page_id_to_offset` round-trip is unchanged).

## Type

- [x] Bug fix

## Notes

Scoped to the one function. The two base64 pagination schemes in the repo (`search_utils` standard
b64 vs `paging_utils` url-safe b64) are left as-is to keep the change minimal.

### DIFF
--- a/openhands/app_server/utils/search_utils.py
+++ b/openhands/app_server/utils/search_utils.py
@@ -1,4 +1,5 @@
 import base64
+import binascii
 from typing import AsyncIterator, Callable
 
 
@@ -12,8 +13,12 @@ def offset_to_page_id(offset: int, has_next: bool) -> str | None:
 def page_id_to_offset(page_id: str | None) -> int:
     if not page_id:
         return 0
-    offset = int(base64.b64decode(page_id).decode())
-    return offset
+    try:
+        return int(base64.b64decode(page_id).decode())
+    except (ValueError, UnicodeDecodeError, binascii.Error):
+        # Malformed/opaque page_id -> start from the beginning, mirroring
+        # paging_utils.decode_page_id which also tolerates bad cursors.
+        return 0
 
 
 async def iterate(fn: Callable, **kwargs) -> AsyncIterator:
--- a/tests/unit/utils/test_search_utils.py
+++ b/tests/unit/utils/test_search_utils.py
@@ -1,3 +1,4 @@
+import base64
 from dataclasses import dataclass
 
 import pytest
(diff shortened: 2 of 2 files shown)


VenkateswarluNagineni:
Confirming this is ready for review. I've added a plain-language note in the HUMAN section describing the fix. Thanks!

smolpaws (commented):

Taste rating: good taste

Findings:
- Changed area reviewed: openhands/app_server/utils/search_utils.py, tests/unit/utils/test_search_utils.py.
- I did not find issues that matter to users or maintainers in the diff. The change is focused and keeps the shape simple.

Risk assessment: low.

Verdict: worth merging

neubig:

Thank you for understanding.

PR #15096, opened 2026-07-02
feat(api-keys): make 'unbound' org scope an explicit, first-class option

tofarr:
HUMAN:

One step in ongoing work for security updates

AGENT:

---

## Why

API keys with no `org_id` persisted on them are currently described in
the codebase as a 'legacy' path. The intent is actually a first-class
'All orgs' / unbound key, but the wording leaks through to the UI, the
docs, the code comments and the test cases.

The UI also gives users no way to choose this scope explicitly: the
create-key flow binds the new key to the request's effective org and
leaves no UI affordance for picking a specific org or 'All orgs'.

At request time, an unbound key should resolve to the org named in the
`X-Org-Id` header (with a membership check) and otherwise fall back to
the caller's currently-selected org -- this already happens in
`SaasUserAuth._resolve_org_id` and is the correct behavior, but the
'legacy' framing made it look accidental.

## Summary

* Remove all 'legacy' wording for the no-`org_id` path; describe it
  as the **unbound** key flow.
* `POST /api/keys` accepts an optional `org_id`:
  - explicit `null` -> create an **unbound** key (new).
  - UUID -> bind the key to that org, requiring the caller to be a
    member (super-role bypass preserved, mirroring
    `_resolve_org_id`).
  - omitted -> bind to the request's effective org (backwards
    compatible).
* `GET /api/keys` and `GET /api/keys/current` now return `org_id`
  (and `bound_org_id` for the current endpoint) so the UI can show
  bound vs unbound. Listing also includes `org_id IS NULL` rows so
  unbound keys are visible from any org context.
* `get_current_api_key` no longer returns 400 for unbound keys;
  `org_id` is the resolved effective org and `bound_org_id` is the
  key's persisted binding (`null` when unbound).
* `CreateApiKeyModal` gains an org selector with an 'All orgs' option
  (default) and the orgs the user is a member of. The API keys table
  gains a 'Scope' column with an 'All orgs' / 'Bound to current org'
  badge. New i18n keys added for 13 locales.

## How to Test

1. Backend
   - `poetry run pytest tests/unit/test_api_key_store.py tests/unit/test_authorization.py tests/unit/test_saas_user_auth.py tests/unit/server/auth/test_saas_user_auth_effective_org.py tests/unit/server/routes/test_api_keys.py tests/unit/storage/test_saas_sql_app_conversation_info_service.py` -- 227 tests pass.
   - `pre-commit run --all-files --show-diff-on-failure --config ./dev_config/python/.pre-commit-config.yaml` passes.
2. Frontend
   - `cd frontend && node_modules/.bin/vitest run src/components/features/settings` (14 tests pass).
   - `cd frontend && node_modules/.bin/eslint src/` -- 0 errors.
   - `cd frontend && npm run build` succeeds.
3. Manual (requires SaaS env): in Settings -> API keys, click 'Create
   API Key' and confirm the new 'Organization' dropdown lists 'All orgs'
   plus the user's orgs. Pick 'All orgs' to create an unbound key; the
   row in the table should show an 'All orgs' badge. Pick a specific
   org to create a bound key; the row should show a 'Bound to current
   org' badge. With the new unbound key, send a request with
   `X-Org-Id: <other-org-id>` (where you are a member) and verify it
   is accepted.

## Video/Screenshots

API Keys page has updates...
image

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

* The new membership check is enforced on **creation** of a bound key.
  Once created, a key's persisted `org_id` is not re-validated on
  every request -- the same as before this change. The existing
  per-request membership check on `X-Org-Id` continues to apply.
* I created the branch `feature/api-key-org-scope-explicit-unbound`
  from the previous working branch `refactor/remove-accepted-tos-from-cookie`
  so the PR targets `main` cleanly. The earlier `accepted_tos` work
  is not included.
* The backend changes are fully backwards compatible: omitting
  `org_id` on `POST /api/keys` preserves the prior effective-org
  binding, and the `/api/keys/current` 400 is gone (clients that
  relied on the 400 were already broken for unbound keys).

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-3e71939   docker.openhands.dev/openhands/openhands:3e71939
```


tofarr:
**Follow-up fix (commit d855372):** I just pushed a fix for the 500 you saw. The store's `create_api_key(org_id=None)` was silently rebinding to `user.current_org_id`, so the route's explicit "All orgs" insertion got mutated to a bound row, the post-insert `name + org_id is None` lookup then missed, and the route returned 500.

The fix adds `use_current_org_fallback=True` to `ApiKeyStore.create_api_key` (default preserves existing internal callers like `oauth_device`) and has the create-key route pass `False`. With that, an explicit `null` `org_id` is now stored verbatim.

Regression tests at both layers (`test_create_api_key_unbound_with_fallback_disabled`, `test_create_api_key_default_fallback_still_works`, and two route-level tests in `TestCreateApiKeyRoute") prevent this from coming back.

The branch has been force-pushed; please re-test the 'All orgs' flow.

hieptl (approved):
Thank you! 🙏

tofarr:
**Follow-up: explicit 2x2 matrix of auth flows (commit 4622eb6).** I added a new `TestAuthOrgResolutionMatrix` class to `tests/unit/server/auth/test_saas_user_auth_effective_org.py` that pins each cell of the (api_key_org_id, X-Org-Id) matrix with a name mirroring the user-facing description:

* `test_unbound_key_no_header_uses_user_current_org` -- Case 1
* `test_unbound_key_with_header_uses_header_org` -- Case 2
* `test_bound_key_no_header_uses_bound_org` -- Case 3
* `test_bound_key_with_matching_header_uses_bound_org` -- Case 4
* `test_bound_key_with_conflicting_header_raises_403` -- Case 4b (403 path)

The matrix is additive -- the existing 21 tests in the file stay in place, so the matrix gives a fast at-a-glance read of coverage and the original tests provide deeper scenario coverage. All 26 tests in the file pass.

PR #15100, opened 2026-07-02
fix(frontend): restore cross-domain PostHog tracking by aligning client/server distinct_id (WIP)

lilagrc:
HUMAN:
This attempts to better sync how we handle `posthog.identify` in the SaaS/cloud UI. It adds an `identify` call on the client side that uses the person's keycloak id as the distinct_id, which is what the cloud server uses and expects. It also closely mirrors the implementation in agent canvas. 

I tested this in a feature branch/staging Posthog, and while I'm not 100% certain if it will connect person profiles in Posthog across the marketing site to the app, it doesn't take away from the current implementation and actually fills a gap in how we were tracking person identities in the cloud between client + server.  

---

AGENT:

## Why

Cross-domain tracking between the marketing site (openhands.dev/all-hands.dev) and the app stopped working after PR #14006 moved most PostHog business events server-side.

**Root cause:** Client-side `posthog.identify()` was called with the GitHub username (in `useGitUser`), while the server-side `AnalyticsService` uses the keycloak `user_id` as `distinct_id`. PostHog treated these as different people, so the anonymous distinct_id bootstrapped from the marketing site never got linked to the server-side events.

The cross-domain tracking flow requires:
1. Marketing site passes anonymous `distinct_id` + `session_id` via URL hash (working)
2. App bootstraps PostHog with those IDs (working)
3. When user logs in, `posthog.identify(X)` merges anonymous into X (working)
4. Server events use X as distinct_id -- was broken: server used keycloak ID, client used GitHub username

## Summary

- **New `usePostHogIdentify` hook** -- Identifies with keycloak `user_id` in SaaS mode (matching server's distinct_id) and git `login` in OSS mode
- **Gated on analytics consent** -- Mirrors the agent canvas [`use-posthog-identify.ts`](https://github.com/OpenHands/agent-canvas/blob/main/src/hooks/use-posthog-identify.ts): `posthog.identify()` only runs when `user_consents_to_analytics === true`. When consent is explicitly revoked (`false`), a prior identify is undone via `posthog.reset()`. While the decision is pending (`null`) or settings are still loading, the hook is a no-op.
- **Removed `posthog.identify()` from `useGitUser`** -- It used the wrong ID and was the wrong place for identification
- **Added `usePostHogIdentify()` call in `root-layout.tsx`** -- Runs on app load alongside other init hooks
- **Cleaned up dead code** -- Removed unused `useTracking` mock and PostHog mock from test files

## How to Test

1. Visit the marketing site (openhands.dev) and click through to the app
2. Sign in and verify in PostHog that the anonymous session from the marketing site is linked to the authenticated user
3. Verify that no identify occurs when analytics consent has not been granted
4. Verify server-side events (conversation created, etc.) appear under the same person profile as the client-side events

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-bbe39de   docker.openhands.dev/openhands/openhands:bbe39de
```


hieptl (approved):
Thank you! 🙏

PR #15101, opened 2026-07-02
feat(saas-auth): accept api_key cookie as a fallback credential

tofarr:
HUMAN:

This is a further step in improving security of the agent canvas - When wired in, the agent canvas will authenticate using a httponly cookie rather than a token stored in local storage. (Mitigating XSS token exfiltration)

AGENT:

---

## Why

The agent canvas currently authenticates against the SAAS with an API
key that is stored in `localStorage` on the client. Any XSS on the
canvas page can read `localStorage` and exfiltrate that key, which is
a long-lived bearer credential.

This PR is the server-side prerequisite for moving the canvas to an
`HttpOnly` cookie, which JavaScript running in the page cannot read and
which therefore cannot be exfiltrated by an XSS payload. The SAAS auth
layer learns to accept the new `api_key` cookie in the same way it
already accepts the `X-Access-Token` header; the canvas frontend work
to mint and use the cookie will land in a follow-up.

## Summary

- `get_api_key_from_header()` now falls back to
  `request.cookies.get('api_key')` after the three existing header
  sources. The cookie is the lowest priority, so an explicit header
  always wins when both are present.
- `SetAuthCookieMiddleware._check_tos()` now treats the `api_key`
  cookie as a valid credential; otherwise a cookie-only request would
  be rejected with `NoCredentialsError` (401) before the auth code
  could run.
- Added unit tests covering cookie extraction, priority vs. each
  existing header, empty/unrelated values, and full bearer-auth
  integration (valid + invalid keys), plus middleware tests for the
  new credential source and the still-present no-credentials 401.

## Issue Number

N/A

## How to Test

Run the auth unit tests:

```bash
cd enterprise
PYTHONPATH=. .venv/bin/pytest tests/unit/test_saas_user_auth.py tests/unit/test_auth_middleware.py -v
```

All 77 tests pass (59 in `test_saas_user_auth.py` + 18 in
`test_auth_middleware.py`), including the 10 new tests added by this PR.

To smoke-test end-to-end against a running SAAS instance, hit any
`/api/...` route with `Cookie: api_key=<valid_key>` (no other credential
sources) and confirm the request is accepted, and with no cookie + no
headers and confirm it is still rejected with 401.

## Video/Screenshots

N/A — backend-only change with no UI surface.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Security model:** the `api_key` cookie is automatically sent by
  browsers, so the XSS-exfiltration mitigation only works if the cookie
  is set with `Secure; HttpOnly; SameSite=Strict` (or `Lax`). Unlike
  `keycloak_auth`, the server never sets this cookie itself; it is
  purely an inbound credential. A code comment in
  `get_api_key_from_header` documents this for future maintainers.
- **Resolution order** (highest → lowest):
  1. `Authorization: Bearer <key>`
  2. `X-Session-API-Key` header
  3. `X-Access-Token` header
  4. `api_key` cookie  ← new
- No migrations, no config changes, no public API changes.
- This PR is the server half of the move; the agent canvas frontend
  change to mint/use the cookie will follow in a separate PR.
- This PR description was updated by an AI agent (OpenHands) on
  behalf of the user.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-21e5c30   docker.openhands.dev/openhands/openhands:21e5c30
```


hieptl (approved):
Thank you! 🙏

PR #15103, opened 2026-07-02
fix: support SDK-native and secret-aware MCP settings

### LINKED ISSUE #15226: Support SDK-native and secret-aware MCP settings
## Summary

Combine SDK-native MCP settings with secret-aware settings persistence, preserve SaaS MCP auth credentials while encrypting secret leaves at rest, and update the SDK, Agent Server, and tools pins.

## Tracking

PR: https://github.com/OpenHands/OpenHands/pull/15103

## PR evidence acceptance criteria

- [ ] **Current `main` evidence:** On current main, load legacy and SDK-native MCP settings containing encrypted auth leaves, rotate one credential, restart the app, and capture any dropped server, overwritten secret, or incompatible serialization.
- [ ] **PR branch evidence:** On the PR branch, repeat the save/rotate/restart/new-conversation flow and verify all MCP servers remain usable, unchanged secrets survive, secret leaves are encrypted at rest, and the updated SDK/Agent Server/tools pins interoperate.
- [ ] **Required evidence package:** The PR records both commit SHAs and exact environment/setup. Run the full relevant backend, service, infrastructure, or data pipeline live with PR code and attach commands plus output/logs; a movie or video is not a substitute for this live run. Inspect sanitized database rows and run an actual cloud conversation that invokes a preserved MCP tool.

neubig:
HUMAN:

This PR ensures SDK-native MCP server auth headers and stdio env secrets are encrypted at rest, stay scoped to the acting org member, and survive settings store/load round trips without leaking into org defaults or API responses.

AGENT:
- Rechecked GitHub on 2026-07-13 UTC before rerunning evidence: `OpenHands/OpenHands#15103` head and `refs/pull/15103/head` both pointed to [`40b4c931a3f2da7481f1d9811437ad2ada5af115`](https://github.com/OpenHands/OpenHands/commit/40b4c931a3f2da7481f1d9811437ad2ada5af115); `main` pointed to [`3949e1cc17d9443f1f4ef7d34d428baf065cd919`](https://github.com/OpenHands/OpenHands/commit/3949e1cc17d9443f1f4ef7d34d428baf065cd919).
- Reran the full local SaaS/MCP harness from a clean detached checkout of exact product head `40b4c931a3f2da7481f1d9811437ad2ada5af115`; `git status --porcelain=v1` was empty immediately before execution. Result: `PASS`.
- The PR-head proof exercised real enterprise `saas_server:app`, `SaaSServerConfig`, PostgreSQL, Redis, production API-key auth, process sandbox, two disposable org members, a local authenticated deterministic OpenAI-compatible LLM, and a local authenticated FastMCP streamable-HTTP server. It verified API redaction, member scoping, redacted-save restart preservation, rejected missing/wrong credentials, successful app conversation completion, LLM tool visibility, MCP tool-result delivery, and authorized recovered bearer/header MCP traffic.
- Ran current-main baselines from clean `3949e1cc17d9443f1f4ef7d34d428baf065cd919` checkout. The full app harness fails before MCP assertions with `WAITING_FOR_SANDBOX` and no agent-server URL; an MCP-specific guard fails deterministically because `openhands/app_server/mcp/mcp_config_adapter.py` / `openhands.app_server.mcp.mcp_config_adapter` is absent on main.
- Durable sanitized evidence is committed in [`0684fe651`](https://github.com/OpenHands/OpenHands/commit/0684fe651) at [`.pr/logs/full_local_saas_mcp_live_evidence.md`](https://github.com/OpenHands/OpenHands/blob/0684fe651/.pr/logs/full_local_saas_mcp_live_evidence.md). The commit is evidence-only; product code tested remains `40b4c931a3f2da7481f1d9811437ad2ada5af115`.
- Sanitization: no raw API keys, bearer tokens, session keys, JWT values, MCP secret values, service ports, or sandbox URLs are included in the durable evidence.
- Local validation for this evidence refresh: `make install-pre-commit-hooks`; `npm ci && npm run build` in both detached worktrees; enterprise migrations to `137` on PR head and `136` on main; `poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files .pr/logs/full_local_saas_mcp_live_evidence.md`.

---

## Why

SDK-native MCP settings can carry auth headers and stdio env secrets that must survive SaaS settings round trips without leaking one member's private MCP config into org defaults or leaving secret leaves plaintext at rest.

## Summary

- Preserve native MCP server auth/env fields while storing settings/profile secret leaves encrypted at rest.
- Keep `mcp_config` scoped to the acting org member, including when org-version validation reserializes org defaults.
- Resolved the migration conflict by moving the settings secret-storage migration to `137` after main's `131`-`136` enterprise migrations.

## Issue Number

Closes OpenHands/OpenHands#15226

## How to Test

- `make install-pre-commit-hooks`
- `uv lock --check`
- `poetry check --lock`
- `cd enterprise && poetry check --lock`
- `PYTHONPATH="..:.:$PYTHONPATH" poetry -C enterprise run alembic -c alembic.ini heads` -> `137 (head)`
- `PYTHONPATH=".:$PYTHONPATH" uv run pytest -q tests/unit/mcp/test_mcp_integration.py tests/unit/storage/data_models/test_settings.py tests/unit/app_server/test_settings_agent_kind_switch.py tests/unit/app_server/test_live_status_app_conversation_service.py tests/unit/app_server/utils/test_jsonpatch_compat.py tests/unit/utils/test_redact.py`
- `PYTHONPATH=".:enterprise:$PYTHONPATH" poetry run --project=enterprise pytest -q -p no:ddtrace -p no:ddtrace.pytest_bdd -p no:ddtrace.pytest_benchmark enterprise/tests/unit/test_saas_settings_store.py enterprise/tests/unit/test_settings_secret_storage.py enterprise/tests/unit/test_org_profiles.py enterprise/tests/unit/test_user_store.py enterprise/tests/unit/test_org_store.py enterprise/tests/unit/test_agent_profiles.py`
- `cd frontend && npm test -- __tests__/hooks/mutation/use-mcp-server-mutations.test.tsx __tests__/routes/llm-settings.test.tsx __tests__/utils/mcp-config.test.ts __tests__/utils/mcp-config-format.test.ts`
- `poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml`
- `poetry -C enterprise run pre-commit run --all-files --show-diff-on-failure --config ./dev_config/python/.pre-commit-config.yaml`

### 2026-07-13 final refresh

- Product head tested: [`40b4c931a3f2da7481f1d9811437ad2ada5af115`](https://github.com/OpenHands/OpenHands/commit/40b4c931a3f2da7481f1d9811437ad2ada5af115)
- Current-main baseline: [`3949e1cc17d9443f1f4ef7d34d428baf065cd919`](https://github.com/OpenHands/OpenHands/commit/3949e1cc17d9443f1f4ef7d34d428baf065cd919)
- Evidence commit: [`0684fe651`](https://github.com/OpenHands/OpenHands/commit/0684fe651)
- Durable evidence: [`.pr/logs/full_local_saas_mcp_live_evidence.md`](https://github.com/OpenHands/OpenHands/blob/0684fe651/.pr/logs/full_local_saas_mcp_live_evidence.md)

Result: full local SaaS/MCP proof from clean PR-head checkout passed. Current main failed the equivalent baseline before MCP assertions with `WAITING_FOR_SANDBOX`, and an MCP-specific guard failed because the SDK-native MCP config adapter path is absent on main.

## Video/Screenshots

UI screenshots in `.pr/logs/`:
- `screenshot_mcp_settings.png` — MCP settings tab showing stored HTTP (SSE) and stdio servers with secrets redacted
- `screenshot_home.png` — Settings page with MCP navigation tab visible

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Contains enterprise migration `137_rewrite_settings_secret_storage.py`, renumbered after current main's enterprise migration chain.
- Frontend lint/build completed with existing non-blocking warnings only.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-0684fe6   docker.openhands.dev/openhands/openhands:0684fe6
```


simonrosenberg:
AI review. I am checking which points are relevant
```
[P1] Stop carrying credentials to changed MCP endpoints —  /worktrees/OpenHands-review-15103/openhands/app_server/mcp/mcp_config_adapter.py:107-108
    When a user edits an authenticated server, changes its URL, and leaves the API-key field blank, the UI omits auth and headers, so this branch restores the old credential based only on the server name. The credential is then sent to the new
    endpoint, and leaving the field blank cannot clear it. Preserve omission only when the endpoint identity is unchanged or when an explicit redaction marker is supplied.

  - [P1] Preserve decryption across encryption-key rotation —  /worktrees/OpenHands-review-15103/enterprise/storage/encrypt_utils.py:64-66
    When a deployment adds a newer active encryption key and restarts, JwtService retains the old keys but this constructs the field-level cipher from only the newest default. Existing profile, MCP, LLM, and agent-context secrets can no longer
    decrypt; _SettingsReadCipher then treats the ciphertext as plaintext and a subsequent save may encrypt that ciphertext again. Encode a key ID or try all retained keys when reading.

  - [P1] Decrypt org profiles before Agent Profile resolution —  /worktrees/OpenHands-review-15103/enterprise/storage/org.py:78-80
    After this column changes from EncryptedJSON to SecretAwareJSON, ORM reads return JSON whose secret leaves are still Fernet ciphertext. load_llm_profiles() still validates that dictionary without get_settings_cipher_context(), and both
    conversation-time Agent Profile resolution and /materialize use that loader, so an Agent Profile referencing a BYOR LLM profile treats the ciphertext as its API key. Validate with the cipher context before constructing the profile loader.

  - [P1] Decrypt the legacy user-profile fallback —  /worktrees/OpenHands-review-15103/enterprise/storage/saas_settings_store.py:410-413
    When org.llm_profiles is empty for an older personal account, this new decryption branch is skipped and the raw user.llm_profiles value already copied into kwargs is used. Migration 116 did not backfill those user-only profiles, and
    Settings(**kwargs) is called without validation context, so stored API/AWS keys remain Fernet tokens and profile activation sends those tokens to the provider. Apply the same cipher-aware validation to the user fallback.

  - [P1] Keep loading prompt skills for ACP agents —  /worktrees/OpenHands-review-15103/openhands/app_server/app_conversation/live_status_app_conversation_service.py:2005-2006
    request.agent is an ACPAgent—an AgentBase, not an Agent—so this guard returns for every ACP conversation with a remote workspace. The existing ACP caller deliberately invokes this helper, and SDK 1.35 marks AgentContext.skills as ACP-
    compatible and renders them into the prompt, so repository, organization, and marketplace skills silently disappear. Accept ACP agents or AgentBase rather than skipping them.

  - [P2] Let the SDK migrate wrapped FastMCP configs before flattening —  /worktrees/OpenHands-review-15103/openhands/app_server/settings/settings_models.py:291-293
    For a legacy wrapped mcpServers config containing FastMCP-only fields such as tools or include_tags, flattening here removes the wrapper before the SDK loader sees it. The SDK uses that wrapper to migrate and drop unsupported external
    fields; once flat, native MCPServer validation rejects them, so OSS or file-backed users with such persisted configs cannot load settings after upgrading. Pass the raw wrapped payload to validate_agent_settings or preserve the wrapper until
    migration runs.

  - [P2] Adapt the MCP probe router to app-server state —  /worktrees/OpenHands-review-15103/openhands/app_server/app.py:74-78
    The mounted SDK probe handlers call get_cipher(request), which requires request.app.state.config, but the app-server lifespan never initializes that state and instead uses get_global_config(). Consequently /api/mcp/test and /api/mcp/oauth/
    start return 503 in both OSS and SaaS before probing anything. Mount an app-server-adapted router or cipher dependency rather than the raw agent-server router.
```

simonrosenberg:
Other AI review

```
  🔴 Correctness — confirmed

  1. Migration 137 binds a raw dict to a typeless JSON column — deploy-blocking
  enterprise/migrations/versions/137_rewrite_settings_secret_storage.py:87
  The UPDATE uses a bare sa.column() (NullType) then .values({column_name: updated_dict}) for the JSON agent_settings/agent_settings_diff columns. Binding a Python dict to a typeless column raises psycopg2 ProgrammingError: can't adapt type
  dict (SQLite: type dict is not supported) on the first row that differs after re-dump — aborting alembic upgrade and blocking the deploy. Sibling migration 108 declares sa.column('agent_settings', sa.JSON()) precisely to avoid this.

  2. load_llm_profiles decrypts without cipher context → api_keys ship as ciphertext → 401s
  enterprise/storage/agent_profile_resolution.py:52
  The column moved to SecretAwareJSON (leaves stay Fernet-encrypted on read); decryption now requires context=get_settings_cipher_context() on model_validate. The two sibling read sites (org_profiles.py:138, saas_settings_store.py:411) were
  updated — this one was missed. Launching with an active Agent Profile referencing a BYOR key writes the ciphertext as the LLM's api_key (has_real_api_key returns True, so the managed-key fallback is skipped), and every LLM call fails auth.

  3. user.llm_profiles fallback also never decrypts — same class, different site
  enterprise/storage/saas_settings_store.py:410
  Only the if org.llm_profiles: branch got the cipher context. When the org has no profiles (legacy personal accounts, exactly the case the comment describes), kwargs['llm_profiles'] still holds the raw SecretAwareJSON value with encrypted
  keys, and Settings(**kwargs) validates it with no context → personal-account profiles come back with ciphertext keys and fail auth.

  4. MCP _preserve_auth resurrects a cleared credential
  openhands/app_server/mcp/mcp_config_adapter.py:115
  When a user blanks an MCP server's API key, the frontend emits the server with no headers. _preserve_auth hits elif 'headers' not in incoming and existing_headers: and deep-copies the previous headers (dumped with expose_secrets=True, i.e.
  the real Bearer token) back on. The removed credential is silently restored and persisted — the user cannot revoke the key.

  5. Decrypt wrapper masks failures by returning the ciphertext
  enterprise/storage/encrypt_utils.py:31
  _SettingsReadCipher.decrypt returns SecretStr(secret) — the raw ciphertext — when try_decrypt_str returns None. After a signing-key rotation (or wrong key), the LLM is configured with the encrypted blob as its api_key: every request 401s with
  no decryption error surfaced to explain why.

  6. mcpServers key sniff drops all servers when one is named mcpServers
  openhands/app_server/mcp/mcp_config_adapter.py:52
  mcp_config_server_map treats any mapping containing a mcpServers key as the legacy wrapper. In the native flat server-map shape, a user-named server called mcpServers is misread as the wrapper — all real servers are dropped and bogus
  config-field-keyed entries injected, silently losing the agent's MCP tools.

  ---
  🟠 Correctness — plausible (verify against intent)

  7. Migration 137 adds a hard validation gate over all rows
  enterprise/migrations/versions/137_rewrite_settings_secret_storage.py:53
  _validate_all_agent_settings runs validate_agent_settings over every row, including sparse org_member.agent_settings_diff validated as full settings. A single legacy/partial/ACP-shaped row raises ValueError and aborts the whole upgrade —
  whereas before, such rows loaded fine (read-time validators degrade to defaults). Worth confirming no such rows exist in prod, or making the gate non-fatal.

  8. Sandbox _get_process_status drops the STARTING path → ERROR during startup
  openhands/app_server/sandbox/process_sandbox_service.py:189
  Previously any alive non-RUNNING/STOPPED status returned STARTING (skipping the health check). Now every alive non-STOPPED/non-ZOMBIE process returns RUNNING, forcing a health GET that's refused during the startup window → status set to
  ERROR. A polling client may see ERROR instead of STARTING and tear down a sandbox that was merely initializing.

  ---
  🧹 Cleanup

  9. Agent verification scratch committed to the repo — .pr/logs/full_local_saas_mcp_live.py:1 (CONFIRMED). ~1.4k lines of throwaway dev scripts + 166KB of PNG screenshots + evidence dumps that may carry test secrets. Belongs in the PR
  description, not the tree.

  10. Dead FastMCP compat shim — mcp_config_adapter.py:34 (CONFIRMED). pyproject pins openhands-sdk==1.35.0 (native config present), so native_mcp_config_supported() is always True — every else/ImportError/FastMCPConfig branch is unreachable.

  11. if context is None branch is a no-op — settings_models.py:294 (CONFIRMED). validate_agent_settings(data, context=None) equals the no-context call (pydantic default). Collapse to one line.

  12. Eager log-arg construction — live_status_app_conversation_service.py:1361 (CONFIRMED). The "Final MCP configuration" f-string builds dump_mcp_config_for_log + sanitize_config unconditionally, even when INFO is disabled.

  13. Hand-rolled copy loop + duplicate INFO log — live_status_app_conversation_service.py:1317 (PLAUSIBLE). The per-item cast loop equals mcp_servers.update(custom_mcp_servers); two INFO lines log the same count.

  14. Triplicated secret-boundary dump recipe — org_store.py:508 (PLAUSIBLE). model_dump(mode='json', context=..., exclude={'llm': {'api_key'}}) is copy-pasted across 3 sites with no shared helper — an edit to one and not the others silently
  persists a plaintext/double-managed key.
```

simonrosenberg:
Superseded by OpenHands/software-agent-sdk#4099 (released in SDK v1.36) plus the focused OpenHands fix in OpenHands/OpenHands#15257. OpenHands/OpenHands#15257 keeps the SDK-owned MCP migration logic in the SDK and adds the remaining SaaS credential round-trip and encrypted per-member storage changes without the broader settings rewrite.

PR #15104, opened 2026-07-02
feat(oauth-device): add /cookie endpoint that sets the api_key HttpOnly cookie

tofarr:
HUMAN:

Basically, it is an endpoint like the one you added for token exchange in the device authorization flow but it sets a httponly cookie instead of returning the api key in the response

AGENT:

---

## Why

The `/oauth/device/token` endpoint returns the user's API key in the response body. For browser-based flows (e.g. embedded CLI auth pages), that means the secret lands in JS, browser history, and proxy logs. PR #15101 added the `api_key` HttpOnly cookie as a fallback credential, but nothing in the device flow sets it yet — this PR adds the matching setter.

## Summary

-
- Adds `POST /oauth/device/cookie`, a sibling of `/token` that runs the same RFC 8628 validation, looks up the per-device API key, and writes it into the `api_key` cookie instead of returning it in the body. The response body only contains `{success, user_id}`.
- Extracts the shared rate-limit / status / API-key-lookup logic from `device_token` into `_resolve_device_api_key` so `/token` and `/cookie` cannot drift apart on error semantics.
- Adds 8 unit tests in `TestDeviceCookie` covering the success path, all four OAuth error responses, `slow_down`, the localhost `Secure`-flag omission, the missing-API-key server error, and an explicit check that the API key never appears in the response body.

## Issue Number

Related to #15101.

## How to Test

Run the existing test suite from the `enterprise/` directory:

```bash
cd enterprise
poetry run pytest tests/unit/server/routes/test_oauth_device.py -v
```

All 30 tests pass (22 pre-existing + 8 new for `/cookie`):

- `TestDeviceAuthorization` — `/authorize` still works.
- `TestDeviceToken` — `/token` still returns the API key in the body and its error responses are unchanged.
- `TestDeviceCookie` (new) — `/cookie` sets the cookie with the right attributes and never leaks the API key in the body.
- `TestDeviceVerificationAuthenticated`, `TestDeviceTokenRateLimiting`, `TestDeviceVerificationTransactionIntegrity` — unchanged, still pass.

Manual smoke test (requires the dev stack):

1. `POST /oauth/device/authorize` and complete `/oauth/device/verify-authenticated` in the browser as before.
2. `POST /oauth/device/cookie` with the returned `device_code` (form-encoded).
3. Inspect the response: status 200, body is `{"success":true,"user_id":"..."}` (no `access_token`), and the response includes a `Set-Cookie: api_key=...; HttpOnly; Secure; SameSite=strict; Domain=...` header.
4. A subsequent `GET /api/...` with that cookie (and no `Authorization` header) should authenticate as the device user via the path added in #15101.

## Video/Screenshots

N/A — server-only change with automated test coverage.

## Type

- [ ] Bug fix
- [x] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The cookie attributes (`HttpOnly`, `Secure` off `localhost`, `SameSite=Strict` in prod / `Lax` in dev) match the security note on `get_api_key_from_header` in `enterprise/server/auth/saas_user_auth.py`.
- The API key is a short opaque token (well under the 4096-byte single-cookie cap), so a plain `set_cookie` is used; the chunked-cookie machinery from `enterprise/server/auth/cookie_chunking.py` is reserved for the much larger `keycloak_auth` JWS.
- The `_resolve_device_api_key` helper mutates the device code store (poll-time / slow-down state) as a side effect, so its docstring warns against calling it twice for the same `device_code`.
- Draft until the reviewer can exercise the cookie against the existing frontend in their dev stack.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-199832a   docker.openhands.dev/openhands/openhands:199832a
```

### DIFF
--- a/enterprise/server/routes/oauth_device.py
+++ b/enterprise/server/routes/oauth_device.py
@@ -99,7 +99,7 @@ def _oauth_error(
 
 async def _resolve_device_api_key(
     device_code: str,
-) -> tuple[Optional[str], Optional[JSONResponse]]:
+) -> 'str | JSONResponse':
     """Run the shared device-flow validation and return the API key.
 
     Both ``/token`` and ``/cookie`` need to enforce exactly the same
@@ -107,8 +107,9 @@ async def _resolve_device_api_key(
     two endpoints in lock-step and avoids drifting error responses when one
     endpoint grows new checks.
 
-    Returns ``(api_key, None)`` on success or ``(None, error_response)`` when
-    the caller should short-circuit with an OAuth-style error. The helper
+    Returns the API key string on success or a ``JSONResponse`` error when
+    the caller should short-circuit with an OAuth-style error. Callers narrow
+    the union with an ``isinstance(result, JSONResponse)`` check; the helper
     performs its own side effects on the device code store (poll-time
     updates, etc.) so callers must not call it twice for the same
     ``device_code`` — the second call would see the rate-limit state already
@@ -117,7 +118,7 @@ async def _resolve_device_api_key(
     device_code_entry = await device_code_store.get_by_device_code(device_code)
 
     if not device_code_entry:
-        return None, _oauth_error(
+        return _oauth_error(
             status.HTTP_400_BAD_REQUEST,
             'invalid_grant',
             'Invalid device code',
@@ -134,7 +135,7 @@ async def _resolve_device_api_key(
                 'new_interval': current_interval,
             },
         )
-        return None, _oauth_error(
+        return _oauth_error(
             status.HTTP_400_BAD_REQUEST,
             'slow_down',
             f'Polling too frequently. Wait at least {current_interval} seconds between requests.',
@@ -145,21 +146,21 @@ async def _resolve_device_api_key(
     await device_code_store.update_poll_time(device_code, increase_interval=False)
 
     if device_code_entry.is_expired():
-        return None, _oauth_error(
+        return _oauth_error(
             status.HTTP_400_BAD_REQUEST,
             'expired_token',
             'Device code has expired',
         )
 
     if device_code_entry.status == 'denied':
-        return None, _oauth_error(
+        return _oauth_error(
             status.HTTP_400_BAD_REQUEST,
             'access_denied',
             'User denied the authorization request',
         )
 
     if device_code_entry.status == 'pending':
-        return None, _oauth_error(
+        return _oauth_error(
             status.HTTP_400_BAD_REQUEST,
             'authorization_pending',
             'User has not yet completed authorization',
@@ -171,7 +172,7 @@ async def _resolve_device_api_key(
                 'Authorized device code missing user_id',
                 extra={'user_code': device_code_entry.user_code},
             )
-            return None, _oauth_error(
+            return _oauth_error(
                 status.HTTP_500_INTERNAL_SERVER_ERROR,
                 'server_error',
                 'User identification missing',
@@ -191,20 +192,20 @@ async def _resolve_device_api_key(
                     'user_code': device_code_entry.user_code,
                 },
             )
-            return None, _oauth_error(
+            return _oauth_error(
                 status.HTTP_500_INTERNAL_SERVER_ERROR,
                 'server_error',
                 'API key not found',
             )
 
-        return device_api_key, None
+        return device_api_key
 
     # Fallback for unexpected status values
     logger.error(
         'Unknown device code status',
         extra={'status': device_code_entry.status},
     )
-    return None, _oauth_error(
+    return _oauth_error(
         status.HTTP_500_INTERNAL_SERVER_ERROR,
         'server_error',
         'Unknown device code status',
@@ -281,12 +282,12 @@ async def device_authorization(
 async def device_token(device_code: str = Form(...)):
     """Poll for a token until the user authorizes or the code expires."""
     try:
-        api_key, error_response = await _resolve_device_api_key(device_code)
-        if error_response is not None:
-            return error_response
+        result = await _resolve_device_api_key(device_code)
+        if isinstance(result, JSONResponse):
+            return result
 
         # Return the API key as access_token
-        return DeviceTokenResponse(access_token=api_key)
+        return DeviceTokenResponse(access_token=result)
 
     except Exception as e:
         logger.exception('Error in device token: %s', str(e))
@@ -316,16 +317,21 @@ async def device_cookie(
     ``enterprise/server/auth/saas_user_auth.py``.
     """
     try:
-        api_key, error_response = await _resolve_device_api_key(device_code)
-        if error_response is not None:
-            return error_response
+        result = await _resolve_device_api_key(device_code)
+        if isinstance(result, JSONResponse):
+            return result
+        api_key = result
 
         # We need the user_id to surface in the response body; re-resolve it
         # from the device code so the helper stays a pure (key, error)
         # function. The store call is cheap and the rate-limit / status
         # state has already been mutated by ``_resolve_device_api_key``.
         device_code_entry = await device_code_store.get_by_device_code(device_code)
-        user_id = device_code_entry.keycloak_user_id if device_code_entry else ''
+        user_id = (
+            device_code_entry.keycloak_user_id
+            if device_code_entry and device_code_entry.keycloak_user_id
+            else ''
+        )
 
         response = JSONResponse(
             status_code=status.HTTP_200_OK,
--- a/enterprise/server/routes/oauth_device.py
+++ b/enterprise/server/routes/oauth_device.py
@@ -1,5 +1,6 @@
 """OAuth 2.0 Device Flow endpoints for CLI authentication."""
 
+from dataclasses import dataclass
 from datetime import UTC, datetime, timedelta
 from typing import Optional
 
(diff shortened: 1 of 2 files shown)


enterprise/server/routes/oauth_device.py
```diff
@@ -80,6 +97,145 @@ def _oauth_error(
...
+    cookie cap), so a plain ``set_cookie`` is sufficient and avoids the
+    chunked-cookie machinery used for the much larger Keycloak JWS.
+    """
+    secure = request.url.hostname != 'localhost'
+    response.set_cookie(
+        key=API_KEY_COOKIE_NAME,
+        value=api_key,
+        domain=get_cookie_domain(),
+        secure=secure,
+        httponly=True,
```
tofarr, line 326: The API key being stored here expires in 7 days (`KEY_EXPIRATION_TIME = timedelta(days=7)` in this same file), but this cookie is being set without `max_age` or `expires`, so it becomes a session cookie that the browser drops on close. That creates a confusing lifetime mismatch: a user who keeps their browser open past 7 days silently holds an invalid credential, and a user who closes the browser is silently signed out even though the key still has days left.

Recommend `max_age=int(KEY_EXPIRATION_TIME.total_seconds())` (and `path='/'` while you're at it, so the cookie is sent to all `/api/...` routes, not just the current path) — and if you intentionally want the cookie to be shorter than the key, document the mismatch in the helper's docstring rather than burying it.

enterprise/server/routes/oauth_device.py
```diff
@@ -126,112 +282,66 @@ async def device_authorization(
...
+        result = await _resolve_device_api_key(device_code)
+        if isinstance(result, JSONResponse):
+            return result
+        api_key = result
+
+        # We need the user_id to surface in the response body; re-resolve it
+        # from the device code so the helper stays a pure (key, error)
+        # function. The store call is cheap and the rate-limit / status
+        # state has already been mutated by ``_resolve_device_api_key``.
+        device_code_entry = await device_code_store.get_by_device_code(device_code)
```
tofarr: We just fetched this entry inside `_resolve_device_api_key` (line 118) and its `keycloak_user_id` is exactly what we need here. Two store reads (plus any side-effects the store does for `get_by_*`) per successful poll is wasteful, and the "the store call is cheap" justification ages badly — the helper already mutated poll-time state in the first call, so this second call is the first one to see the persisted change.

Cleaner: have the helper return `(api_key, keycloak_user_id)` (or the entry itself), and drop this re-fetch. Then the helper becomes truly the single fetch and the duplication goes away by construction.

enterprise/server/routes/oauth_device.py
```diff
@@ -80,6 +97,145 @@ def _oauth_error(
     )
 
 
+async def _resolve_device_api_key(
```
tofarr: Returning `'str | JSONResponse'` and forcing every caller to do `isinstance(result, JSONResponse)` is a smell — your docstring even spells out the "must not call it twice" hazard as a *convention* rather than a guarantee. Two cleaner options:

1. Raise a typed `DeviceFlowError` carrying `status_code`, `error_code`, `description`, `interval`, and have a FastAPI exception handler serialize it into the same `DeviceTokenErrorResponse` JSON. Both endpoints become a straight `try/except` — no union, no `isinstance`.

2. Return a small frozen dataclass like `DeviceFlowResult(api_key: str | None, error: JSONResponse | None)` from a helper factory. Callers do `if result.error: return result.error`.

Either way, "do not call twice" becomes structurally impossible instead of relying on a comment that future maintainers will skim past.

enterprise/server/routes/oauth_device.py
```diff
@@ -126,112 +282,66 @@ async def device_authorization(
...
+
+        # We need the user_id to surface in the response body; re-resolve it
+        # from the device code so the helper stays a pure (key, error)
+        # function. The store call is cheap and the rate-limit / status
+        # state has already been mutated by ``_resolve_device_api_key``.
+        device_code_entry = await device_code_store.get_by_device_code(device_code)
+        user_id = (
+            device_code_entry.keycloak_user_id
+            if device_code_entry and device_code_entry.keycloak_user_id
+            else ''
```
tofarr: This `else ''` can only be reached if the device code store raced between the two reads (first inside `_resolve_device_api_key`, second here) — i.e. it would mask a real bug rather than report one. At this point the helper has already returned either 500 (missing `keycloak_user_id`) or 200, so an empty `keycloak_user_id` here truly means "the device entry was deleted or mutated under us", which is a state-store bug we want to see.

Either drop the fallback and let it raise, or log a warning and return 500 explicitly — but don't silently hand the client a 200 with `user_id=""`.

enterprise/tests/unit/server/routes/test_oauth_device.py
```diff
@@ -171,6 +173,230 @@ async def test_device_token_success(self, mock_store, mock_api_key_class):
...
+        assert len(set_cookie_headers) == 1
+        cookie_header = set_cookie_headers[0].decode()
+        assert f'{API_KEY_COOKIE_NAME}=secret-api-key-value' in cookie_header
+        assert 'HttpOnly' in cookie_header
+        assert 'Secure' in cookie_header
+        assert 'SameSite=strict' in cookie_header
+        assert 'Domain=example.com' in cookie_header
+
+        # Verify the helper reused the same device-specific API key name.
+        mock_api_key_store.retrieve_api_key_by_name.assert_called_once_with(
```
tofarr, line 296: Pinning `retrieve_api_key_by_name` here is exactly the assertion pattern that catches contract regressions — please add the same assertion for `get_by_device_code`:

```python
mock_store.get_by_device_code.assert_called_once_with('test-device-code')
```

That would have caught the double-fetch issue noted in the route file (`_resolve_device_api_key` and the success path both call it). It's also a useful tripwire against the helper ever growing a redundant lookup.

tofarr (commented):
### Review of #15104

This is a focused, well-scoped PR: the new `/cookie` endpoint is a minimal sibling of `/token`, the shared validation helper eliminates drift between the two, and the test coverage cleanly enumerates the OAuth error matrix plus the no-leak invariant. The cookie attributes (HttpOnly, Secure-off-localhost, SameSite=Strict-in-prod, Domain) faithfully match the security note introduced in #15101, and pinning `API_KEY_COOKIE_NAME` as a constant in this module avoids the kind of string-typo that has historically haunted cookie auth.

A handful of items below are worth addressing before this moves out of draft.

**Blockers-ish (correctness/lifetime):**

- Cookie has no `max_age` / `expires`. The API key it carries expires in 7 days (`KEY_EXPIRATION_TIME = timedelta(days=7)`), but the cookie is a session cookie that the browser deletes on close. A user who keeps their browser open past 7 days will silently hold an invalid credential; a user who closes the browser is silently signed out even though the key is still valid. Match `max_age` to `KEY_EXPIRATION_TIME` (or document the lifetime mismatch in the docstring).
- Double-fetch of `device_code_entry` in the `/cookie` success path. The helper already calls `device_code_store.get_by_device_code(device_code)` at line 118; the success path then calls it again on the very next line just to read `keycloak_user_id`. Have the helper return `(api_key, keycloak_user_id)` (or the entry itself) and drop the redundant call — two store reads per successful poll is wasteful, and the "the store call is cheap" comment is the kind of reasoning that ages badly.

**Style / robustness:**

- `_resolve_device_api_key` returning `'str | JSONResponse'` and relying on `isinstance(result, JSONResponse)` as control flow is awkward; the docstring's "must not call it twice" warning is convention-only. Either raise a typed `DeviceFlowError` and let a FastAPI exception handler serialize it, or return a small tagged dataclass. Both make the "don't call me twice" hazard impossible by construction.
- The defensive `else ''` for `user_id` in `device_cookie` can only fire if the device code store raced between the two reads — i.e. it papers over a real bug rather than reporting one. Drop the fallback and let it crash (or log at warning level).
- Pin `mock_store.get_by_device_code.assert_called_once_with('test-device-code')` in `test_device_cookie_success` so the double-fetch regression above is caught by CI.

The PR template notes "Draft until the reviewer can exercise the cookie against the existing frontend in their dev stack", which is a separate gate from code-review sign-off — happy to take it through that step once the items above land.

enterprise/server/routes/oauth_device.py
tofarr: Fixed in 3eef6c21c. Added `max_age=int(KEY_EXPIRATION_TIME.total_seconds())` (7 days / 604800s) and `path='/'` in `_set_api_key_cookie`. Docstring updated to call out the lifetime-vs-key alignment. The success test now also asserts `Max-Age=604800` and `Path=/` are present in the Set-Cookie header.

enterprise/server/routes/oauth_device.py
tofarr: Fixed in 3eef6c21c. The helper now also returns `user_id`, so the success path of `device_cookie` no longer re-fetches the device-code entry. The test pins `mock_store.get_by_device_code.assert_called_once_with(...)` so any future regression that re-introduces the second fetch is caught.

enterprise/server/routes/oauth_device.py
tofarr: Fixed in 3eef6c21c (chose option 2 — the dataclass). Introduced a frozen `DeviceFlowResult(api_key, user_id, error: JSONResponse | None)`. `__post_init__` enforces the "exactly one of success/error is populated" invariant, so callers cannot accidentally read the wrong field. Callers now do `if result.error is not None: return result.error`. Added 4 unit tests covering the invariant. Mypy strict_optional is happy because `api_key: str` is always non-None after the error check.

enterprise/server/routes/oauth_device.py
tofarr: Fixed in 3eef6c21c as a side effect of the previous fix. The double-fetch and its `else ''` fallback are gone — the helper now returns `user_id` alongside `api_key` from the single `get_by_device_code` call, so the cookie endpoint no longer reads the store a second time on the success path. There is no longer a path that silently hands the client a 200 with `user_id=""`.

enterprise/tests/unit/server/routes/test_oauth_device.py
tofarr: Fixed in 3eef6c21c. Added `mock_store.get_by_device_code.assert_called_once_with('test-device-code')` right next to the `retrieve_api_key_by_name` assertion. This pins the single-fetch contract and would have caught the regression described in the companion route-file comment.

neubig:

Thank you for understanding.

PR #1552, opened 2024-05-03
Refactor config to dataclasses

enyst:
Config has become big and is still growing. This PR refactors it in dataclasses and breaks out a couple of sub-sections, llm and agent, that are relatively self-contained. Should also clean up some older hacks. Bonus: mypy now figured out what's up here, and it didn't like things! (for good reason)

Goals:
[x] refactor config into data classes
[x] keep compability with env and current toml (Makefile updated, current toml seems less important to keep)
[x] read new toml
[x] serialize for the ui
[x] adapt Makefile

Misc fixes: wrong types checks for config vars, integration tests for sandboxes not mounting in docker

Also, @li-boxuan suggestion implemented for CLI:
 - can define groups of LLM settings in config.toml, such as:
```
[gpt4]
model = "got-4-1106-preview"
api_key = "sk..."
embeddings_model = "openai"
...
```
- a group can include any attribute of the LLMConfig dataclass, all optional
- usage e.g.: poetry run python opendevin/core/main.py ... -l gpt4
- only one of -l (--llm_config) or -m (--model) needs to be provided
- if both are passed, then llm_config.model overrides model (...this doesn't sound right)

Notes:
- ~~tested without UI so far~~ looks like it still works with UI
- it keeps compatibility with current vars
- sandboxes vars seem tricky. 
    - this PR removes None as possible value for some paths, thus possibly (?) changes behavior.
    - changes integration tests execution, they were running with None values for almost every config var, thus not mounting in docker and other effects.
    - removes None for user id
    - refactored all globals in sandboxes into properties
   @xingyaoww 
- tested for old and new vars
- tested for types (e.g. retries vars need to be int), cleaned up some hacks
- updatable singletons are a bit too much fun, feedback welcome if it is too risky/unorthodox
- ~~did not try yet local_box, exec_box~~
- not tested: full app on docker.

The result is like:
```
[core]
workspace_base = "/home/myuser/workspace"
max_iterations = 100
cache_dir = "/tmp/cache"
sandbox_container_image = "ghcr.io/opendevin/sandbox:latest"
run_as_devin = true
max_iterations = 100
sandbox_type = "ssh"
use_host_network = false
ssh_hostname = "localhost"
sandbox_timeout = 120
...

[llm]
api_key = "sk-proj-somethi..."
model = "gpt-3.5-turbo-1106"
api_version = "2023-01"
embedding_model = "openai"
...

[agent]
name = "MonologueAgent"
memory_enabled = true
memory_max_threads = 2

PR #15108, opened 2026-07-03
fix(frontend): stop streamed deltas rendering twice and fragmenting in V1 chat

shanemort1982:
HUMAN:

Hit this on our self-hosted fleet after #15021 landed: every streamed response renders twice (live preview bubbles plus the durable event with the same text), and reloading a conversation fragments each response into one bubble per delta. Fix validated end to end on a self-hosted box, live and on reload, single and multi-step runs.

AGENT:

---

## Why

Since #15021, V1 chat displays streamed content twice and fragments it on reload. Two causes in `handleEventForUI`:

1. On reload, the WebSocket replay delivers durable events without `StreamingDeltaEvent`s while the REST history includes them, so the REST deltas land after the turn's final message is already rendered. The backward merge stops at the rendered message, so every delta becomes its own bubble and the final message stays as a duplicate.
2. Live, a tool call `ActionEvent` carries the same `thought` text that just streamed, and `finalizeStreamingDeltasInPlace` only reconciles `FinishAction` and assistant `MessageEvent`s, never `ActionEvent`s, so every step's reasoning displays twice for the whole run.

## Summary

- Drop a delta that is older than the newest durable event (stale replay arrival).
- When a tool call action arrives, remove the turn's preview deltas; the action card renders its own thought.
- When the turn's final message does not reconcile against the accumulated deltas (multi-step turns), replace them with the final message instead of appending a duplicate.

Existing merge/fold behaviour is unchanged; all 18 pre-existing tests in `handle-event-for-ui.test.ts` still pass, plus 4 new ones covering the above.

## Issue Number

No issue filed; regression introduced by #15021.

## How to Test

1. Run any multi-step conversation (e.g. "run these three commands one at a time, narrating each") and watch it live: each narration should render once, not once as a streamed bubble and again on the action card.
2. Reload the conversation page mid-run or after completion: the response should stay one bubble per message rather than fragmenting into one bubble per streamed chunk with a duplicated final message.
3. `cd frontend && npx vitest run __tests__/utils/handle-event-for-ui.test.ts` (22 tests).

### DIFF
--- a/frontend/__tests__/utils/handle-event-for-ui.test.ts
+++ b/frontend/__tests__/utils/handle-event-for-ui.test.ts
@@ -489,6 +489,30 @@ describe("handleEventForUI - durable events supersede streaming deltas", () => {
     expect(ui).toEqual([userMessage, action]);
   });
 
+  it("keeps a reasoning-only bubble when the superseding action lacks reasoning", () => {
+    // For many models the delta is the sole carrier of reasoning_content, so
+    // stripping the streamed text must not lose the reasoning display.
+    let ui: OpenHandsEvent[] = [userMessage];
+    ui = handleEventForUI(
+      mkDelta(
+        "sup-d5",
+        "2026-07-03T00:00:01.000Z",
+        "I'll run a command",
+        "considering the request",
+      ),
+      ui,
+    );
+    const action = mkAction("2026-07-03T00:00:02.000Z", "I'll run a command");
+    ui = handleEventForUI(action, ui);
+
+    expect(ui).toHaveLength(3);
+    const kept = ui[1] as StreamingDeltaEvent;
+    expect(isStreamingDeltaEvent(kept)).toBe(true);
+    expect(kept.content).toBeNull();
+    expect(kept.reasoning_content).toBe("considering the request");
+    expect(ui[2]).toBe(action);
+  });
+
   it("replaces non-reconciling preview deltas with the final message instead of duplicating", () => {
     // In a multi-step turn the deltas accumulated since the last user message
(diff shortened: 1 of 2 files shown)


shanemort1982:
@VascoSch92 tagging you since this follows on from #15021: the token streaming works, but self-hosted the V1 chat renders each streamed response twice (live preview bubbles plus the durable event), and reloading a conversation fragments it into one bubble per delta. Root cause is in `handleEventForUI`: the WebSocket history replay omits `StreamingDeltaEvent`s while the REST history includes them, so late deltas cannot merge past the already rendered final message, and tool call `ActionEvent`s re-render the thought text that just streamed. This PR treats deltas as a live preview that durable events supersede; all 18 existing tests still pass plus 4 new ones. Happy to adjust if you had a different design in mind for the reconcile path.

VascoSch92:
@shanemort1982 thank you

We had the sane problem in agent-canvas. i already fix it there, but not here :-) 


I will give a look ;)

shanemort1982:
Had a look at how you solved it in agent-canvas (#1552) and pulled the two refinements across so the reconcile paths match: the superseding action now keeps the delta as a reasoning-only bubble when it lacks reasoning of its own, and reconciliation tolerates the SDK's trailing-whitespace stripping. The one intentional divergence: this app also drops deltas that arrive older than the newest durable event, because here the WebSocket history replay omits deltas while the REST history includes them, so on a reload they land after the final message and fragment the transcript. agent-canvas has a single event source so it never hits that. 23 tests passing.

VascoSch92 (approved):

VascoSch92:
Thanks :-)

PR #15113, opened 2026-07-03
fix(app-server): deduplicate repositories by full_name in ProviderHandler

minh2416294:
HUMAN:
While I look through the git provider integration code I noticed the repository dedup helper wasn't actually removing duplicates. it checks `full_name` but adds `id` to the seen set, so it never matches. I fix the one line and added a unit test that fails on the old code and passes on the fix. I also ran the integration tests locally, all green.

## What & why

`ProviderHandler._deduplicate_repositories` is documented as *"Remove duplicate repositories based on `full_name`"*, but it never removed anything. The loop tested membership with `repo.full_name not in seen` while populating the `seen` set with `repo.id`:

```python
if repo.full_name not in seen:
    seen.add(repo.id)          # bug: adds id, but membership is tested by full_name
    unique_repos.append(repo)
```

Since `Repository.id` and `Repository.full_name` are distinct string fields, a `full_name` never matches an `id` in the set, so the guard could never fire and duplicate repositories were returned unchanged. This is reachable from `search_repositories` when a provider is selected, so duplicate repos could surface in repository search results.

## Fix

One line populate `seen` with `full_name` so it matches the membership test:

```python
seen.add(repo.full_name)
```

## Tests

Adds `tests/unit/integrations/test_provider_dedup.py`:

- `test_deduplicate_repositories_removes_duplicate_full_names` — fails before this fix, passes after.
- `test_deduplicate_repositories_keeps_distinct_full_names` — guards against over-deduping / reordering.

Verified locally: both new tests pass, all existing `test_provider_immutability.py` tests still pass, `ruff check` / `ruff format --check` clean.

## Scope note

The multi-provider search branch intentionally does not dedup (the same `full_name` from different providers are genuinely distinct repos), so it is left unchanged.

## Changelog

> Fixed repository search returning duplicate repositories (`ProviderHandler` deduplication was a no-op).

### DIFF
--- a/openhands/app_server/integrations/provider.py
+++ b/openhands/app_server/integrations/provider.py
@@ -399,7 +399,7 @@ class ProviderHandler:
         unique_repos = []
         for repo in repos:
             if repo.full_name not in seen:
-                seen.add(repo.id)
+                seen.add(repo.full_name)
                 unique_repos.append(repo)
         return unique_repos
 
--- a/tests/unit/integrations/test_provider_dedup.py
+++ b/tests/unit/integrations/test_provider_dedup.py
@@ -1,0 +1,53 @@
+from __future__ import annotations
+
+from types import MappingProxyType
+
+from pydantic import SecretStr
+
+from openhands.app_server.integrations.provider import (
+    ProviderHandler,
+    ProviderToken,
+    ProviderType,
+    Repository,
+)
+
+
+def _repo(repo_id: str, full_name: str) -> Repository:
+    return Repository(
+        id=repo_id,
+        full_name=full_name,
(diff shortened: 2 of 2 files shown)


minh2416294:
Hey, gentle bump on this one. It's a small, self-contained fix and CI is green. I re-checked against main today and the bug is still there: _deduplicate_repositories adds repo.id to the seen set but tests membership by full_name, so the repository-search dedup never actually removes anything. The branch still merges cleanly (no conflicts)

@ak684  you've worked in integrations/provider.py recently right? would you be able to take a quick look, or point me to the right reviewer? I'm happy to adjust anything.

neubig:

Thank you for understanding.

PR #15123, opened 2026-07-05
fix(app-server): list ~/.openhands/skills (V1) user skills in /skills/search

minh2416294:
## What
Make the WebUI skills list (`GET /api/v1/skills/search`) scan user skills from both `~/.openhands/skills/` (V1, preferred) and `~/.openhands/microagents/` (V0, legacy), deduplicating by name with the V1 copy winning on collisions.

## Why
The endpoint scanned only `~/.openhands/microagents/` (V0). But `skills/README.md` documents V1 as supporting **both** `~/.openhands/skills/` (preferred) and `~/.openhands/microagents/` (backward compat), and the sibling conversation-skills endpoint already treats `~/.openhands/skills/` as the user path (`app_conversation_router.py`, "User skills (~/.openhands/skills/)"). So a user who follows the preferred V1 convention gets their skills loaded into conversations but never listed in the WebUI, the two endpoints disagreed on where user skills live.

This addresses the **WebUI-visibility** half of OpenHands/software-agent-sdk#4252. The runtime "host-home-vs-container" visibility half discussed in that thread is separate (and out of scope here), so this is linked as `Refs`, not `Closes`.

## How
- `skills_router.py`: `USER_SKILLS_DIR` now points at the V1 `~/.openhands/skills/`; added `USER_SKILLS_DIR_LEGACY` for the V0 path. `search_skills` loops both (V1 first) and dedups by name, mirroring the existing `_merge_skills` idiom (later/legacy loses on collision). Dedup also protects the name-keyed pagination cursor from duplicate entries.

## Testing
- `test_skills_search_includes_v1_user_skills`: proves V1-path skills now appear (fails on `main`, passes here).
- `test_skills_search_merges_user_paths_v1_wins`: proves both paths scanned, name collision deduped to one entry, V1 copy wins.
- Existing search tests gained a `USER_SKILLS_DIR_LEGACY` patch so they stay deterministic across machines.
- `pytest tests/unit/server/routes/test_skills_api.py` → 12 passed. ruff check + format clean; mypy no new errors.

## HUMAN:
I hit this myself: put a skill in `~/.openhands/skills/` per the README's "preferred" note and it never showed up in the WebUI skills list. Traced it to the endpoint only scanning the old `microagents/` path. Verified locally: the new test fails before the change and passes after, and the full `test_skills_api.py` file is green.

### DIFF
--- a/openhands/app_server/user/skills_router.py
+++ b/openhands/app_server/user/skills_router.py
@@ -28,7 +28,10 @@ user_context_dependency = depends_user_context()
 
 # skills/ is at the repo root, two levels above the openhands package __file__
 GLOBAL_SKILLS_DIR = Path(openhands.__file__).parent.parent / 'skills'
-USER_SKILLS_DIR = Path.home() / '.openhands' / 'microagents'
+# User skills live under ~/.openhands. V1 prefers skills/; microagents/ is the
+# V0 legacy path kept for backward compatibility. V1 wins on name collisions.
+USER_SKILLS_DIR = Path.home() / '.openhands' / 'skills'
+USER_SKILLS_DIR_LEGACY = Path.home() / '.openhands' / 'microagents'
 
 
 class SkillInfo(BaseModel):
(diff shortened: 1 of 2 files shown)


minh2416294:
Hi there, I'm confirming this is ready to review. ci's green, tested locally - the new test fails on main and passes with the change. small diff, 2 files. lmk if
anything needs adjusting

neubig:

Thank you for understanding.

PR #15124, opened 2026-07-05
Feat/encrypt settings at rest

KHARSHAVARDHAN-eng:
This PR implements leaf-level encryption for secret configuration fields in SaaS/org settings and LLM profiles. Instead of storing entire settings columns as plaintext or whole-column JWE-encrypted blobs, secret fields (e.g., LLM API keys and MCP remote authorization headers/environments) are encrypted at rest using a Fernet cipher, while the remaining non-secret fields remain inspectable in the database.

pr for OpenHands/enterprise#53


neubig:

Thank you for understanding.

PR #15138, opened 2026-07-06
fix: scroll settings sidebar so Skills is reachable in orgs

hieptl:
HUMAN:

AGENT:

---

## Why

In a team organization the **Skills** page looked missing from the Settings sidebar, so org owners/admins couldn't manage their org's skills or plugins. The Skills nav item isn't gated by workspace type — the sidebar pane simply doesn't scroll. `settings-layout` clips overflow (`overflow-hidden`) and the main pane scrolls, but the nav pane had no scroll class. In a team org the extra **ORG SETTINGS** section makes the nav taller than the viewport, pushing the tail of the list (App, Integrations, Skills) below the fold where it's clipped and unreachable. The personal workspace has no ORG SETTINGS section, so its shorter nav kept Skills visible — hence the "personal-only" symptom.

## Summary

- `settings-navigation.tsx`: made the sidebar item list a scrollable region so every nav item stays reachable regardless of nav length, viewport height, or org type.
- Added `min-h-0` to the `<nav>` and `flex-1 min-h-0 overflow-y-auto custom-scrollbar-always` to the item container; the title + org selector stay pinned. Reuses existing utilities (same pattern as the main pane and chat interface) — no new CSS.

## Issue Number

N/A.

## How to Test

1. `cd frontend && npm run dev`, open Settings.
2. In a workspace whose sidebar overflows the viewport (a **team org**, or shrink the browser window height in any workspace) — before this change, the sidebar is clipped and **Skills** (bottom item) can't be reached.
3. With this change, the sidebar shows a scrollbar and scrolls independently of the main pane; scroll to the bottom and confirm **Integrations** and **Skills** are reachable while the title + org selector stay pinned.
4. Confirm no regression in the personal workspace (no stray scrollbar when the list fits) and on mobile (full-screen nav scrolls).

Automated check: verified in a real browser (Playwright + system Chrome) by replicating the exact DOM/CSS in a height-constrained pane — before: nav not scrollable, Skills clipped/unreachable; after: nav scrolls, Skills reachable (PASS). `prettier --check` and `eslint` pass on the changed file.

## Video/Screenshots

https://github.com/user-attachments/assets/cb074c31-e097-46cc-a5f1-18cd6cc34197

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-722799d   docker.openhands.dev/openhands/openhands:722799d
```


tofarr (approved):
🍰

PR #15139, opened 2026-07-06
feat(frontend): add /goal slash command to drive the agent toward an objective

VascoSch92:
HUMAN:

implementatio of the /goal command

AGENT:
Implemented by an AI agent. Frontend checks (tsc, eslint, 24 new unit tests) pass; a full end-to-end `/goal` run against a live LLM has **not** been verified by a human yet.

---

## Why

OpenHands has no way to tell the agent "pursue this objective until it is actually done." A single message runs one turn, so the user has to keep re-prompting and eyeball whether the goal is met. The agent server already ships a `/goal` loop (pursue → judge completion → re-prompt until done or a cap) in the pinned SDK (1.31.1), but nothing in the GUI drives it.

## Summary

- Add a `/goal <objective>` (and `/goal --max <n> <objective>`) slash command that starts the agent-server goal loop for the current V1 conversation, mirroring the existing `/btw` and `/model` interceptor pattern.
- Stream live progress into a bottom **goal status banner** (objective, round, judge score, spinner) fed by `goal` `ConversationStateUpdateEvent`s; once the loop ends the banner hides and the terminal verdict settles **inline** in the message timeline.
- Keep the loop's internal re-prompts out of the chat, and surface the agent-server's real error (e.g. "Conversation run or goal loop already running.") instead of a bare `Request failed with status code 409`.

## How to Test

1. `make run` and open (or start) a V1 conversation.
2. Type `/goal make pytest pass` — or `/goal --max 3 <objective>` to cap the rounds.
3. Watch the live banner advance each round; when the judge is satisfied (or the cap is hit) the banner is replaced by an inline terminal card showing the verdict (score + what's missing).
4. Frontend checks: `cd frontend && npm run test`, `npx tsc --noEmit`, `npx eslint src`.

Automated coverage added: 24 unit tests — interceptor `--max` parsing + passthrough + error surfacing, the goal store, the status component (indicators + missing-note), and the render filter (terminal-vs-active + re-prompt hiding).

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Frontend-only.** Relies on the agent-server `/goal`, `/goal/stop`, `/goal/resume` endpoints already present in the pinned SDK (1.31.1); no backend changes here.
- **Scoped minimal:** no Stop/Resume loop controls (the `--max` cap plus the existing conversation-stop cover runaway loops). Easy follow-up if wanted.
- The re-prompt filter in `should-render-event.ts` matches the SDK prompt text by prefix — brittle by design; keep in sync with `openhands.sdk .../conversation/goal/prompts.py`.
- Separate, unrelated issues found while building/running this — **not** in this PR, happy to open follow-ups: (1) a broken alembic migration `013.py` (`create_index` was passed a string instead of a list, which breaks `make run` on SQLite); (2) the app-server webhook rejecting sandbox events with `422`.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-5c91635   docker.openhands.dev/openhands/openhands:5c91635
```


neubig:

Thank you for understanding.

PR #15141, opened 2026-07-06
fix(sandbox-spec): fall back to defaults when runtime-api has no warm runtimes

tofarr:
HUMAN:

The system should still work when no warm runtimes are specified in the database.

AGENT:

---

## Why

`DynamicRemoteSandboxSpecService._fetch_specs` calls runtime-api's `/api/warm-runtime-configs` and returns the parsed configs. When the endpoint returns no configs (e.g. during a quiet period before any pods are warmed), `_fetch_specs` returned `[]` and `get_default_sandbox_spec` raised `SandboxError('No warm runtime configs available from runtime-api.')`, which prevented conversation startup even though the static `RemoteSandboxSpecService` would have happily served the same request via its preset list.

## Summary

- In `_fetch_specs`, when the response yields no configs, fall back to `get_default_sandbox_specs()` from `remote_sandbox_spec_service.py` — the same default list the static service uses — and cache it under the same TTL.
- Removed the now-unreachable `if not specs: raise SandboxError(...)` branch in `get_default_sandbox_spec` and the unused `SandboxError` import.
- Tests: replaced `test_empty_configs_list` and `test_raises_sandbox_error_when_no_specs_available` with cases verifying the fallback (field-by-field comparison against `get_default_sandbox_specs()`), and added `test_fallback_specs_are_cached` to confirm the TTL cache still applies.

## Issue Number

None.

## How to Test

```
poetry run pytest tests/unit/app_server/test_dynamic_remote_sandbox_spec_service.py -v
# or with the system Python in this sandbox:
python -m pytest tests/unit/app_server/test_dynamic_remote_sandbox_spec_service.py -v
```

All 24 tests pass.

## Video/Screenshots

Before
image

After

## Type

- [x] Bug fix

## Notes

- The fallback uses the existing `get_default_sandbox_specs()` from `remote_sandbox_spec_service.py`, so behaviour stays consistent across both remote sandbox spec services.
- `_name_to_spec` is left empty on the fallback path, so a configured `default_spec_name` will miss and `get_default_sandbox_spec` will return the first default spec, which is the intended behaviour when there are no warm runtimes to look up by name.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-632b2cf   docker.openhands.dev/openhands/openhands:632b2cf
```

### DIFF
--- a/openhands/app_server/sandbox/dynamic_remote_sandbox_spec_service.py
+++ b/openhands/app_server/sandbox/dynamic_remote_sandbox_spec_service.py
@@ -9,7 +9,9 @@ import httpx
 from fastapi import Request
 from pydantic import Field, PrivateAttr
 
-from openhands.app_server.errors import SandboxError
+from openhands.app_server.sandbox.remote_sandbox_spec_service import (
+    get_default_sandbox_specs,
+)
 from openhands.app_server.sandbox.sandbox_spec_models import (
     SandboxSpecInfo,
     SandboxSpecInfoPage,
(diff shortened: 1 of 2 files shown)


hieptl (approved):
Thank you! 🙏

PR #15154, opened 2026-07-06
fix(jira-dc): enable Connect button in email-match mode with managed service account (#15150)

ChetanyaRathi:
HUMAN:

This PR fixes the Jira Data Center Connect button being stuck greyed out in email-match mode with an admin-managed service account. The backend now passes the Jira host to the app in email-match mode (before it only did that for OAuth), and the settings form uses that host so Connect can turn on. I also added frontend tests for this case.

## Why

In the Jira Data Center settings, when the login mode is **email-match** and the service account is **managed by an admin**, the **Connect** button was always greyed out. There was no way to finish the webhook setup from the screen.

The reason: the Connect button needs a Jira host to turn on. But in email-match mode, the app never received the host, and the box to type it was hidden. So the host stayed empty and the button could never turn on.

## Summary

- **Backend:** now sends the Jira host to the app in email-match mode too (before, it only did this in OAuth mode).
- **Frontend:** reads that host and fills it into the form, so the Connect button can turn on.
- **Tests:** added tests to check Connect turns on in this mode, and that the other modes still work the same.

## Issue Number

Closes OpenHands/enterprise#55

## Testing

- Frontend tested on my machine: tests, lint, typecheck, and build all pass.
- Backend code reviewed but not run locally (no Python setup) — CI will test it.

### DIFF
--- a/frontend/src/api/option-service/option.types.ts
+++ b/frontend/src/api/option-service/option.types.ts
@@ -78,6 +78,7 @@ export interface WebClientConfig {
   provider_default_hosts?: Partial<Record<Provider, string>>;
   slack_enabled?: boolean;
   acp_providers?: ACPProviderConfig[];
+  jira_dc_host?: string | null;
   /** Jira DC host when DC OAuth is configured; used to pre-fill + lock the
    *  configure form's host field. Null/absent in email-match mode. */
   jira_dc_oauth_host?: string | null;
--- a/frontend/src/components/features/settings/project-management/jira-dc-integration-panel.test.tsx
+++ b/frontend/src/components/features/settings/project-management/jira-dc-integration-panel.test.tsx
@@ -1,0 +1,185 @@
+import { render, screen, fireEvent } from "@testing-library/react";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { JiraDcIntegrationPanel } from "./jira-dc-integration-panel";
+
+// Mock the I18n hooks and translation keys
+vi.mock("react-i18next", () => ({
+  useTranslation: () => ({
+    t: (key: string) => key,
+  }),
+  initReactI18next: {
+    type: "3rdParty",
+    init: vi.fn(),
+  },
+}));
+
+// We can define variables that tests can override
+let mockConfig = {
+  jira_dc_oauth_host: null as string | null,
+  jira_dc_host: null as string | null,
+  jira_dc_service_account_managed: false,
+  jira_dc_service_account_email: "",
+};
+
+let mockIntegrationData = {
+  status: "inactive",
+  workspace: null as Record<string, unknown> | null,
+};
+
+let mockHasPermission = true;
+
+vi.mock("#/hooks/query/use-config", () => ({
+  useConfig: () => ({ data: mockConfig }),
+}));
+
+vi.mock("#/hooks/query/use-integration-status", () => ({
+  useIntegrationStatus: () => ({ data: mockIntegrationData }),
+}));
+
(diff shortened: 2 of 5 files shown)


aivong-openhands:
The pre-fill goal is sound, but the review found two blocking issues with the current implementation, plus a suggested design that keeps the pre-fill without them.

To be clear up front: this PR implements what the issue's expected-behavior section suggested (surfacing the host via the web-client config and relaxing the submit gate), so the approach was a reasonable reading of OpenHands/enterprise#55. On review, both of those suggestions turn out to have problems — described below — and the issue is being updated accordingly. The blocking issues are with the suggestions themselves, not with how they were implemented.

**Blocking issue 1 — the Connect button can submit an empty workspace.** `requiresHostInput = !serviceAccountManaged` removes the unconditional `!workspace.trim()` check. When `jira_dc_host` is null (`JIRA_DC_BASE_URL` unset, or set without a scheme — `urlparse().hostname` returns `None` for bare hostnames like `jira.example.com`), the form seeds an empty workspace, the host field is hidden, and Connect is enabled. Submitting sends `workspace_name: ""`, which the backend validator rejects with a 422 that cannot be corrected from the UI.

**Blocking issue 2 — the new `jira_dc_host` field discloses the internal Jira DC hostname.** `/web-client/config` is unauthenticated, and `_get_jira_dc_host()` omits the `JIRA_DC_ENABLE_OAUTH` gate that `_get_jira_dc_oauth_host()` enforces. In email-match mode, where the host is intentionally withheld from that response, it is now readable by any visitor.

**Suggested design** — the public config endpoint is the wrong channel for the host; it should come from an authenticated one:

1. Remove `jira_dc_host` from the web-client config injector and `web_client_models.py`.
2. Add a single shared parser for `JIRA_DC_BASE_URL` in `jira_dc.py` (no OAuth gate, with scheme-less handling so bare hostnames resolve). `_configured_jira_dc_workspace_name()` keeps its OAuth gate and delegates parsing to it.
3. Return the host from the authenticated `GET /workspaces/status` endpoint as a new nullable field (e.g. `setup_host`), populated only for callers with the `MANAGE_INTEGRATION_PROVIDERS` permission. Do not reuse the existing `host` field — members depend on it meaning "active workspace to link to" — and do not gate the endpoint itself, only the new field.
4. On the frontend, restore the unconditional `!workspace.trim()` guard, enable `useJiraDcInstanceStatus` for admins as well (currently members-only), and seed the workspace from `setup_host`. Note the status query is async, so the seed must re-run when the data arrives, not only when the modal opens.

Questions about any of the above are welcome — the pre-fill itself is a good improvement and worth landing through the authenticated path.

aivong-openhands:
I want to note that a minimal fix came in that took an alternative route in https://github.com/OpenHands/OpenHands/pull/15155#issuecomment-4901538141. Perhaps that being so minimal in changes is a cleaner approach for a fix. However, I think there is value in the UX enhancements here with the prefill. @ChetanyaRathi @ak684 how would you both like to proceed?

ChetanyaRathi:
Thank you for the detailed review. Both blocking issues make sense to me, especially the security one, where the internal host would be exposed on the unauthenticated /web-client/config in email-match mode. That's not something I'd want to ship.

Since OpenHands/OpenHands#15155 is a smaller, cleaner fix for the core bug, I'm happy for that one to land the fix. If the prefill UX is still wanted, I'm glad to do a follow-up PR using the authenticated /workspaces/status path with the setup_host field, as you described.

Let me know which you'd prefer and I'll go from there.

aivong-openhands:
Thanks @ChetanyaRathi! Let's have a follow up PR for the prefill UX enhancements

ChetanyaRathi:
Closing in favor of OpenHands/OpenHands#15155. Will open the prefill follow-up once it merges.

PR #15159, opened 2026-07-07
fix(app_server): pass flat mcp_config shape to SDK Agent

tofarr:
**Linear:** [APP-2660](https://linear.app/all-hands-ai/issue/APP-2660)

## Why

After the SDK bump in #15156, `Agent.mcp_config` expects a flat `dict[str, MCPServer]`, but `LiveStatusAppConversationService._configure_llm_and_mcp` was still packaging config in the legacy FastMCP `{"mcpServers": ...}` wrapper. Because the override path uses `OpenHandsAgentSettings.model_copy(update=...)`, the input validator that would otherwise accept the wrapper is bypassed, and `Agent.__init__` then raised:

```
ValidationError: 1 validation error for Agent
mcp_config.mcpServers.default
  Extra inputs are not permitted [type=extra_forbidden]
```

…breaking conversation startup.

## Summary

- Return the flat `dict[str, MCPServer]` directly from `_configure_llm_and_mcp` so the value handed to `create_agent` already matches `Agent.mcp_config`.
- Build system and custom MCP servers as `MCPServer` instances with `SecretStr`-wrapped headers in `_add_system_mcp_servers` and `_merge_custom_mcp_config`, dropping the per-server `model_dump` round-trip.
- Remove the now-unused `mcp_config` parameter from `_apply_server_agent_overrides`.

## Issue Number

Follow-up regression fix for the SDK 1.31.2 bump in #15156.

## How to Test

- [x] `pytest tests/unit/app_server/test_live_status_app_conversation_service.py`
- [x] `pytest tests/unit/`
- [ ] Start a conversation that loads a custom MCP server from user settings and confirm it is registered on the `Agent`.

## Type

- [x] Bug fix

## Notes

Follow-up: decide whether `OpenHandsAgentSettings.mcp_config`'s input validator should stop accepting the legacy `{"mcpServers": ...}` wrapper, so the legacy path is removed at the input boundary.

🤖 Generated with [OpenHands](https://all-hands.dev)

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -127,6 +127,7 @@ from openhands.sdk import Agent, AgentContext, LocalWorkspace
 from openhands.sdk.hooks import HookConfig
 from openhands.sdk.llm import LLM
 from openhands.sdk.llm.llm_profile_store import PROFILE_NAME_REGEX
+from openhands.sdk.mcp.config import MCPServer
 from openhands.sdk.plugin import PluginSource
 from openhands.sdk.secret import LookupSecret, StaticSecret
 from openhands.sdk.settings import ACPAgentSettings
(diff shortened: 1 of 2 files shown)


hieptl (approved):
Thank you! 🙏

PR #15165, opened 2026-07-07
fix(frontend): read SDK 1.31.x flat mcp_config wire format

tofarr:
HUMAN:

This was due to changes in the SDK to MCP Configuration.

AGENT:

---

## Why

The SDK 1.31.x refactor changed `AgentSettings.mcp_config` from a FastMCP `MCPConfig` object (with a `.mcpServers` attribute) to a flat `dict[str, MCPServer]`. The wire format in `model_dump()` also dropped the `mcpServers` wrapper as a result. The backend `app_server` was already adapted (commit `9748fc06a`), but the frontend's `parseMcpConfig` still required the wrapped shape, so `GET /api/v1/settings` returned an empty `MCPConfig` and no servers were displayed in the MCP settings UI.

## Summary

- `parseMcpConfig` now accepts both the SDK 1.31.x flat server map (`{shttp: {...}}`) and the legacy wrapped shape (`{mcpServers: {shttp: {...}}}`), mirroring the SDK's own `_extract_mcp_config` / `_normalize_mcp_config_field` logic so older persisted settings keep loading.
- `toSdkMcpConfig` now emits the flat server map (no `mcpServers` wrapper) to match the SDK's native `model_dump` output. The backend's `_normalize_mcp_config_field` still accepts the wrapped shape for back-compat.
- Pass-through for the frontend's already-parsed grouped shape (`{sse_servers, stdio_servers, shttp_servers}`) used by `DEFAULT_AGENT_SETTINGS`, and skip array-valued entries in the SDK iteration to keep the function robust against mixed-shape input.

## How to Test

1. From the repo root, `cd frontend && npm install` if dependencies aren't installed.
2. Run the focused MCP test suites (all pass on the changed branch):

   ```bash
   cd frontend
   npm run test -- --run \
     __tests__/utils/mcp-config.test.ts \
     __tests__/utils/mcp-config-format.test.ts \
     __tests__/hooks/mutation/use-mcp-server-mutations.test.tsx \
     __tests__/components/features/settings/mcp-settings \
     __tests__/routes/mcp-settings.test.tsx
   ```

3. Run the linter and build to confirm no regressions:

   ```bash
   cd frontend
   npm run lint
   npm run build
   ```

4. Manual UI smoke test (optional): load `/settings/mcp` and confirm that an `shttp` server with `url` + `timeout` returned from `/api/v1/settings` (now in the flat `{shttp: {url, timeout}}` shape, not `{mcpServers: {shttp: {url, timeout}}}`) renders correctly and that add/update/delete round-trips through the backend.

## Video/Screenshots

The MCP settings now display servers again, and saving works as expected:
image

And these tools are present in conversations:
image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The backend (`openhands/app_server/settings/settings_models.py`) was already adapted to accept either shape on input via `_normalize_mcp_config_field`, so this is a frontend-only fix that keeps the wire format symmetric on the way in and the way out.
- The new `__tests__/utils/mcp-config-format.test.ts` file locks in the SDK 1.31.x flat wire format behaviour and the legacy wrapped-shape back-compat path.
- The pre-existing full-suite flake in `__tests__/conversation-websocket-handler.test.tsx` ("should clear error message when a successful event is received after a ConversationErrorEvent") was confirmed unrelated — it also fails on `main` when the full suite runs, and passes in isolation on both branches.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-945d324   docker.openhands.dev/openhands/openhands:945d324
```

### DIFF
--- a/dev_config/python/.pre-commit-config.yaml
+++ b/dev_config/python/.pre-commit-config.yaml
@@ -57,8 +57,8 @@ repos:
             types-Markdown,
             pydantic,
             lxml,
-            "openhands-sdk==1.31.2",
-            "openhands-tools==1.31.2",
+            "openhands-sdk==1.32.0",
+            "openhands-tools==1.32.0",
             "sqlalchemy>=2.0",
           ]
         # To see gaps add `--html-report mypy-report/`
--- a/enterprise/poetry.lock
+++ b/enterprise/poetry.lock
@@ -5776,14 +5776,14 @@ numpy = {version = ">=2", markers = "python_version >= \"3.9\""}
 
 [[package]]
 name = "openhands-agent-server"
-version = "1.31.2"
+version = "1.32.0"
 description = "OpenHands Agent Server - REST/WebSocket interface for OpenHands AI Agent"
 optional = false
 python-versions = ">=3.12"
 groups = ["main"]
 files = [
-    {file = "openhands_agent_server-1.31.2-py3-none-any.whl", hash = "sha256:df954eeabc973f99dcd18f2f679de074588333eb1e76bc85f4e5c143a88d52c6"},
-    {file = "openhands_agent_server-1.31.2.tar.gz", hash = "sha256:ad787420c1da0f8b32f1a5714e66f923a8f22d83e7bebaa1e52eb43996fa7eab"},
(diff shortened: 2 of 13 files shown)


hieptl (approved):
Thank you! 🙏

PR #15168, opened 2026-07-07
fix(app_server): derive agent server image from package version

tofarr:
HUMAN:

Since we can't be sure that different versions of the SDK will work together, a check was added some time ago to raise an error on startup if they don't match. This effectively meant that the AGENT_SERVER_IMAGE_REPOSITORY and AGENT_SERVER_IMAGE_TAG variables no longer work. So now we simply grab the agent server version from the sdk dependencies in pyproject.toml

AGENT:

## Why

Starting a conversation on staging sometimes 500'd with:

> Sandbox image ghcr.io/openhands/agent-server:1.31.1-python runs openhands-sdk 1.31.1, but this release requires 1.32.0.

`AGENT_SERVER_IMAGE` was a hand-maintained constant in `sandbox_spec_service.py` that had to be bumped in lockstep with the SDK pin in `pyproject.toml`. When the helm chart was redeployed with a stale `AGENT_SERVER_IMAGE_TAG`, the SDK mismatch surfaced — most visibly when `runtime-api`'s `/api/warm-runtime-configs` returned an empty list, because the `DynamicRemoteSandboxSpecService` fallback echoed the stale env-var image verbatim.

Joe from the platform team (per the OpenHands-Cloud chart's Replicated wiring) also pointed out that the env vars are the self-hosted customer's only way to pin a custom image; making them no-ops would have broken installs that rely on them.

This PR derives the canonical image from the installed `openhands-agent-server` package version, keeps the env-var override path alive for self-hosted, and silently auto-corrects the one dangerous pattern (canonical repo + stale `-python` tag) so both audiences are covered.

## Summary

- **Derive the bundled image URL from the installed `openhands-agent-server` package version** via `importlib.metadata`. `pyproject.toml` is now the single source of truth for the default; no hand-maintained `AGENT_SERVER_IMAGE` constant.
- **Re-honour `AGENT_SERVER_IMAGE_REPOSITORY` and `AGENT_SERVER_IMAGE_TAG`** for self-hosted installs. Different repo or non-`-python` tags are passed through verbatim; only the exact staging failure pattern (canonical repo + stale `-python` tag) is silently re-pinned to the SDK-matching tag with a logged warning. Tags without a `-` separator (e.g. `nightly`) are not touched.
- **`is_custom_sandbox_spec(sandbox_spec_id)`** compares against the bundled default rather than the env-var-driven image, so the post-hoc 500-hint remains meaningful for self-hosted customers whose custom image's SDK has drifted.
- **Refactor to keep the SDK version mockable in tests**: `importlib.metadata.version()` lives inside a `@cache`'d `_bundled_agent_server_version()` helper so tests can still patch it; the constant isn't bound at module-import time.
- Delete `_verify_agent_server_version` (and the corresponding `Version`/`InvalidVersion` import): the proactive SDK check is dead code now that the image is derived from the installed package.

## Issue Number

None.

## How to Test

1. Run the new + updated tests:

   ```
   pytest tests/unit/app_server/test_sandbox_spec_service.py \
          tests/unit/app_server/test_dynamic_remote_sandbox_spec_service.py
   ```

   Should pass 35/35 (with `--forked -n auto`).
2. Full app_server suite:

   ```
   pytest tests/unit/app_server/
   ```

   Should pass 1400+/1400+.
3. Staging smoke check: deploy with the previous stale `AGENT_SERVER_IMAGE_TAG=1.31.1-python` (or whatever's in the chart); the warning fires once at startup and conversations start cleanly.
4. Self-hosted smoke check: deploy with `AGENT_SERVER_IMAGE_REPOSITORY=example.com/agent-server AGENT_SERVER_IMAGE_TAG=1.0.0`; the custom image is honoured verbatim, and a conversation started against it gets the helpful 500-hint on SDK mismatch.

## Video/Screenshots

Conversations start on staging again — even without warm runtimes:
image

## Type

- [x] Bug fix
- [x] Refactor

## Notes

- **Reviewer note on staging**: staging was hitting the `1.31.1-python` mismatch; this PR auto-corrects that pattern without requiring a helm-chart update. The chart still carries `1.31.1-python` until the next bump, and that's now safe.
- **Self-hosted compatibility preserved**: customers who wire `AGENT_SERVER_IMAGE_REPOSITORY`/`AGENT_SERVER_IMAGE_TAG` via the Replicated config (per the OpenHands-Cloud chart) see no behavioural change as long as their repo or tag differs from the canonical pattern. Only the canonical-repo + stale-`-python`-tag case is corrected.
- **Behaviour change vs. the original PR draft**: an earlier draft of this PR made the env vars no-ops. Joe from platform flagged this would break self-hosted installs. The current revision restores env-var support with the targeted auto-correct — diff-vs-original is in the latest commit.
- **`check-version-consistency.yml`**: the compose-file tag portion is now a no-op for this surface (the bundled image is derived). A follow-up PR could remove those specific checks.
- **Follow-up**: once runtime-api's DB-backed image registry is the only place self-hosted customers register custom images, the env-var path can be retired. Out of scope here.

_This PR was prepared by an AI agent (OpenHands) on behalf of @tofarr._


openhands/app_server/sandbox/sandbox_spec_service.py
```diff
@@ -98,19 +96,30 @@ async def resolve_sandbox_spec(
     raise ValueError(f'Sandbox Spec {effective_id!r} not found')
 
 
+@cache  # memoize so the deprecation warning fires once per process
 def get_agent_server_image() -> str:
-    agent_server_image_repository = os.getenv('AGENT_SERVER_IMAGE_REPOSITORY')
-    agent_server_image_tag = os.getenv('AGENT_SERVER_IMAGE_TAG')
-    if agent_server_image_repository and agent_server_image_tag:
-        return f'{agent_server_image_repository}:{agent_server_image_tag}'
-    return AGENT_SERVER_IMAGE
```
jlav, line 106: The replicated installer relies on these environment variables to allow customers to specify their default image to use for sandboxes.

It's not clear to me how without this a customer could bring their own custom image. If there's an API exposed to allow for this customization, I need to understand how that will be integrated with the replicated installer so that users have an intuitive way to keep this up to date.

tofarr: So the way this works now (Or doesn't) is:
* Environment variables not specified: `AGENT_SERVER_IMAGE` is used
* Environment variables specified that do not match `AGENT_SERVER_IMAGE` : Error when trying to start conversation ☹️ ☹️ ☹️
* Environment variables specified that do match `AGENT_SERVER_IMAGE` : Conversation Starts

After this change:
* `AGENT_SERVER_IMAGE` no longer exists - it is derived from sdk dependency in the pyproject.toml.
* If these variables are defined, we print a warning once and move along.

If you want a different / custom image:
* Add a warm runtime config in the runtime api : https://runtime.staging.all-hands.dev/docs#/default/save_warm_runtime_config_api_admin_warm_runtime_configs__config_name__put (Or use the script https://github.com/OpenHands/runtime-api/blob/main/scripts/save_warm_runtime_config.py )

Your warm runtime will then appear in https://staging.all-hands.dev/settings/app : 
image

ak684:
The new direction makes sense to me as discussed [in the slack thread](https://allhandsai.slack.com/archives/C0AD07ETH3Q/p1783455828877159?thread_ts=1783453606.293449&cid=C0AD07ETH3Q): keep the package-derived image as the default when no explicit override is configured, but preserve `AGENT_SERVER_IMAGE_REPOSITORY/TAG` as an explicit OHE/Replicated override.

The Cloud side should stop rendering those vars from default chart values, while still rendering them for customer-configured custom/local/proxied sandbox images. That avoids the stale-default staging failure without dropping the Replicated custom image path.

jlav (approved):

PR #15169, opened 2026-07-07
fix: do not fail login when IdP broker token fetch fails

ak684:
## Problem

On OHE (self-hosted) installs, a **realm-local Keycloak user** (created directly in the realm, with no brokered IdP link) cannot log in: `keycloak_callback` defaults a missing `identity_provider` claim to GitHub and then calls `store_idp_tokens`, which fetches the brokered token from Keycloak's broker endpoint. For a realm-local user there is no broker link, so that fetch returns 400 and the whole `/oauth/keycloak/callback` returns an unhandled **500** — login is impossible. (Reproduced live on a real instance; the manual workaround was setting an `identity_provider=<provider>:saml` user attribute.)

## Fix

Make the broker-token fetch **non-fatal**: if `store_idp_tokens` raises, log a warning and continue the login. A realm-local user simply has no brokered provider tokens to store (which they never had). The SAML skip and the default-to-github behavior (and its existing TODO) are untouched — only the failure is prevented from 500ing.

Happy to narrow the exception handling (e.g. catch only the specific broker 400, or gate on the user being realm-local) if reviewers prefer.

## Test

`test_keycloak_callback_idp_token_failure_non_fatal` mocks `store_idp_tokens` to raise and asserts the callback still returns a 302 redirect (fails on current code with the unhandled exception, passes with the fix). Full `test_auth_routes.py` suite passes.

HUMAN: Reproduced the 500 live on a real OHE instance; fix verified by a red→green unit test. Not yet manually re-tested on an instance after the fix.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-99dabc8   docker.openhands.dev/openhands/openhands:99dabc8
```

### DIFF
--- a/enterprise/server/routes/auth.py
+++ b/enterprise/server/routes/auth.py
@@ -455,9 +455,18 @@ async def keycloak_callback(
     # Only fetch/store IdP tokens for OAuth-based IdPs (not SAML)
     # SAML IdPs don't have OAuth tokens to retrieve from Keycloak's broker endpoint
     if idp_type != 'saml':
-        await token_manager.store_idp_tokens(
-            ProviderType(idp), user_id, keycloak_access_token
-        )
+        try:
+            await token_manager.store_idp_tokens(
+                ProviderType(idp), user_id, keycloak_access_token
+            )
+        except Exception:
+            # Realm-local users have no brokered IdP link, so the broker-token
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #15182, opened 2026-07-08
fix(auth): gracefully handle email verification dispatch failures during GitHub OAuth

KHARSHAVARDHAN-eng:
This PR improves robustness during the OAuth login flow when the email verification dispatch service (Keycloak SMTP) fails or is unconfigured.

### Changes:
- **Backend**: Catches any exception raised when requesting a verification email via `verify_email()` inside the `/keycloak/callback` endpoint. Logs the error diagnostics using `logger.exception()` and redirects the user with `email_send_failed=true` instead of crashing with a raw HTTP 500 error page.
- **Frontend**: Detects `email_send_failed` search parameter via the `useEmailVerification` hook and passes it to `EmailVerificationModal`, which then renders a user-friendly warning message in red/danger styling informing the user that the email failed to send, and suggesting they try again or check their configuration.
- **Tests**: Adds unit test `test_keycloak_callback_email_verification_failed_email_sending` to check parameter propagation.


neubig:

Thank you for understanding.

PR #15185, opened 2026-07-08
feat: surface email enabled for smtp/resend

saurya:
HUMAN: Add a flag for enabling email that separates SMTP from Resend

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-27ea232   docker.openhands.dev/openhands/openhands:27ea232
```


tofarr (approved):
🍰

PR #15186, opened 2026-07-08
feat: budgets and usage monitoring UI

saurya:
HUMAN: create the UI for making and deleting budgets. Change the existing admin dashboard to the usage monitoring dashboard. 

## Summary
- add budgets and usage monitoring UI
- update settings navigation/routes for usage & budgets
- add org usage query hooks and UI wiring

## Testing
- not run (not requested)

_This PR description was created by an AI agent (OpenHands) on behalf of the user._

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-1aefbe6   docker.openhands.dev/openhands/openhands:1aefbe6
```


tofarr (commented):
## Code Review

> 🤖 _This review was generated by an AI agent (OpenHands) on behalf of [@openhands](https://github.com/openhands). Please push back on anything that doesn't apply to this codebase._

### Taste Rating
🟡 **Acceptable** — Ships a lot of working UI, but has a real bug in nav visibility, dropped user-facing features, and a substantial test gap. `+2760 / -1156` adds two large monolithic components without splitting or testing them.

### Verdict
❌ **Needs rework** — one real bug (nav visibility), one silent UX regression (lost filters), and a substantial test gap on a behavior-changing UI surface.

---

### 🔴 Critical Issues

**1. Nav item for `/settings/budgets` is not filtered for non-admin / personal-org users**

`frontend/src/hooks/use-settings-nav-items.ts:94-96` only filters `/settings/usage-monitoring`:
```ts
if (!isAdminOrOwner || !organizationId || isPersonalOrg) {
  items = items.filter((item) => item.to !== "/settings/usage-monitoring");
}
```
But `SAAS_NAV_ITEMS` now also includes `/settings/budgets`, and `routes/settings.tsx` correctly redirects members away from it. Result: a member or a user in a personal org will see a **Budgets** link in the nav, click it, and be silently bounced back to `/settings`.

**2. Duplicated route-protection blocks in `routes/settings.tsx:182-196`**

The `usage-monitoring` and `budgets` blocks are byte-identical. A `Set` of admin-only paths would fix both #1 and #2 in one change:
```ts
const ADMIN_ONLY_PATHS = new Set([
  "/settings/usage-monitoring",
  "/settings/budgets",
]);
if (ADMIN_ONLY_PATHS.has(pathname)) {
  const role = user?.role ?? "member";
  if (!user || (role !== "admin" && role !== "owner") || isPersonalOrg) {
    return redirect("/settings");
  }
}
```
The same set should be reused in `use-settings-nav-items.ts` to hide the link.

**3. Silent user-facing feature regression in `usage-dashboard.tsx`**

The old `admin-dashboard.tsx` passed eight filters to `useOrgConversations` (page, perPage, search, **sortBy**, **sortOrder**, executionStatus, **sandboxStatus**, timeWindow). The new `usage-dashboard.tsx:402-408` passes only five. **Silently dropped**: `sortBy`/`sortOrder` dropdowns and the `sandboxStatus` (Runtime status) filter. The hook still defaults sort to `updated_at desc`, so data is still sorted — but the user can no longer change it. If this is intentional, call it out in the PR description; if not, restore the controls.

**4. No tests for ~2600 lines of new UI**

`budgets.tsx` (1269 LOC), `usage-dashboard.tsx` (1176 LOC), `use-org-user-usage.ts`, and the new `organizationService` methods (`getUserUsageStats`, `getBudgetSettings`, `updateBudgetSettings`, `upsertBudgetOverride`, `deleteBudgetOverride`) all ship with **zero** test coverage. The PR description explicitly states:
> Testing: not run (not requested)

That's the wrong call for behavior-changing frontend work in a SaaS settings area. At minimum, add:
- A `useSettingsNavItems` test asserting `/settings/budgets` is hidden for members/personal orgs (would have caught #1)
- A `clientLoader` test asserting `/settings/budgets` redirects non-admins (would have caught the matching issue in `routes/settings.tsx`)
- A test that `useOrgUsageStats({ timeWindow: "ytd" })` passes `time_window=ytd` and no `days`
- Budgets tab tests covering threshold add/remove and override save/remove round-trips

The PR template even includes a Docker run command — running the app locally and attaching a screenshot of the three budgets tabs + four usage tabs would satisfy the `Evidence` requirement the repo seems to expect.

---

### 🟡 Improvement Opportunities

- **`filteredUsers = userRows` is a useless alias** (`budgets.tsx:595`) — both search and status filter are already applied server-side via `usersSearch`/`usersStatus` query params (verified in `enterprise/server/services/org_budget_service.py`). Either filter something or remove the alias.
- **`useOrgUsageStats` and the service both compute `resolvedDays`** — pick one layer to own this resolution; the hook shouldn't second-guess the service.
- **Two 1100+ line single-file components** — split into `budgets/`, `admin-dashboard/` folders with one file per tab + shared `icons.tsx` / `utils.ts`. Hard to test in isolation as-is.
- **Inline SVG icons duplicated** between `budgets.tsx` and `usage-dashboard.tsx` (`SearchIcon`, `ExportIcon`, `StopIcon`, etc.) — extract to a shared component.
- **`min-h-0` removed from `settings-layout.tsx` and re-added in `routes/settings.tsx`** — pick one owner of the flex/scroll constraints; the current split will silently break scroll for any other consumer of `SettingsLayout`.
- **Narrative comments** like `// Budgets route protection: only admins and owners can access` paraphrase the `if` below; if you keep a single check (issue #2), the comments go away with it.

---

### ⚠️ Risk Assessment: 🟡 MEDIUM

- Three endpoints' worth of UI surface with no test coverage
- Drops two user-facing filters without PR-description callout
- Frontend nav items and route protection are out of sync for `/settings/budgets` (silently bounces non-admin users from a nav link they can see)
- New components are 1100+ lines each in single files — maintenance cost is real

Not HIGH because mutating flows are server-validated and admin-scoped via `Permission.EDIT_ORG_SETTINGS`. The shape of the change is good; completeness isn't.

### Recommendation
Request changes. Before merge: (1) fix the nav-filter oversight; (2) restore or document the removed conversations filters; (3) consolidate the duplicated admin-only route check; (4) add at least the two tests called out in #4. File-splitting and refactor opportunities can be follow-ups.

tofarr (approved):
🍰

PR #15187, opened 2026-07-08
fix: send authenticated marketplace URLs to agent-server

hieptl:
HUMAN:

Auto-load marketplace registrations were forwarded to the agent-server with their raw source strings, so the sandbox cloned marketplace repos without credentials: private repositories failed to clone (and a bare "owner/repo" source was misread as a nonexistent local path), so their skills silently never loaded and keyword triggers never fired. The settings preview endpoint and the org skill-repo path already resolve token-embedded URLs, so adding a private marketplace appeared to work while conversations loaded nothing from it.

Resolve each auto-load registration's source to an authenticated git URL (mirroring the OrgConfig.org_repo_url pattern) right before the /api/skills request:

- Add authenticate_marketplace_sources to skill_loader.py: rewrites sources on wire-only copies (model_copy skips validate_source, which rejects credentialed URLs; nothing is persisted or logged), keeps the original source on any resolution failure, and never raises.
- Move \_parse_marketplace_source from user/skills_router.py to skill_loader.py as parse_marketplace_source; this import direction avoids a cycle through config.py, and the router re-imports it under the old name so existing patch targets keep working.
- Call the helper in load_and_merge_all_skills, covering both the LLM and ACP conversation-start arms.

Public and local-path marketplaces behave exactly as before; failures degrade to the previous behavior. Note: private-repo skills also need an agent-server that includes software-agent-sdk#4025 (root SKILL.md plugin loading), which ships in the next SDK release.

AGENT:

---

## Why

Skills from a **private** marketplace repository (e.g. `github:OpenHands/extensions-private`, registered in `/settings/skills` with Auto Load) never load into conversations, so their keyword triggers never fire. Public repositories work.

At conversation start, the app-server forwards auto-load marketplace registrations verbatim to the agent-server `/api/skills` endpoint, and the sandbox clones them **without credentials**: private repos fail to clone (a bare `owner/repo` source is even misparsed as a nonexistent local path), and the failure is swallowed at every layer. The settings preview endpoint and the org skill-repo path (`OrgConfig.org_repo_url`) both already resolve token-embedded URLs via `get_authenticated_git_url` — the conversation-start marketplace path was the only one missing this step, which is why adding the marketplace _appears_ to work while conversations silently load nothing from it.

## Summary

- Add `authenticate_marketplace_sources` (`skill_loader.py`): rewrites each auto-load registration's source to an authenticated git URL on a wire-only copy before the `/api/skills` request — `model_copy` deliberately skips `validate_source` (which rejects credentialed URLs); nothing is persisted or logged; any resolution failure keeps the original source and never raises, so public/local marketplaces behave exactly as before.
- Move `_parse_marketplace_source` from `user/skills_router.py` to `skill_loader.py` as `parse_marketplace_source` (import direction avoids a cycle through `config.py`); the router re-imports it under the old name.
- Wire the helper into `load_and_merge_all_skills` (`app_conversation_service_base.py`), covering both the LLM and ACP conversation-start arms; add unit tests for the new helper and the wiring.

## Issue Number

N/A.

## How to Test

Unit tests:

```
poetry run pytest tests/unit/app_server/test_skill_loader.py \
 tests/unit/app_server/test_app_conversation_service_base.py \
 tests/unit/server/routes/test_skills_api.py

```

End-to-end (requires a git provider token that can read the private repo, and an agent-server image containing software-agent-sdk#4025 — e.g. `ghcr.io/openhands/agent-server:98f5624-python` — see Notes):

1. Settings → Skills → Marketplace → Add Repository → `OpenHands/extensions-private`, Auto Load on.
2. Start a new conversation; the sandbox log should show `Fetching marketplace 'extensions-private' from https://****@github.com/OpenHands/extensions-private.git` (redacted credentials) with no clone failure.
3. Send `openhands support` → the `openhands-happyfox` skill keyword-triggers.
4. Negative: register an inaccessible repo → conversation still starts; other skills unaffected.

## Video/Screenshots

https://github.com/user-attachments/assets/4b6d9f66-b363-43d0-b11b-a18a4c979107

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-f8c99fc   docker.openhands.dev/openhands/openhands:f8c99fc
```


tofarr (approved):
🍰

PR #14447, opened 2026-05-16
fix(events): avoid hydrating unpaged filesystem events

pragnyanramtha:
- [ ] A human has tested these changes.

---

## Why

Addresses part of #12616.

`EventServiceBase.search_events()` currently hydrates every matching stored event into a full SDK `Event` model before filtering, sorting, and applying `limit`. For long V1 conversations, a small `/events/search?limit=N` request can still pay full validation cost for the entire event history.

## Summary

- Split event search into lightweight metadata filtering/sorting/pagination, then hydrate full `Event` models only for the selected page.
- Added a filesystem metadata loader that reads only `kind` and `timestamp` from stored JSON before pagination.
- Added regression coverage proving `limit=3` only loads three full filesystem events.

## Issue Number

Addresses part of #12616

## How to Test

Validated locally:

```text
uv run --group test pytest tests/unit/app_server/test_filesystem_event_service.py -q
uv run --group test pytest tests/unit/app_server/test_filesystem_event_service.py tests/unit/app_server/test_aws_event_service.py tests/unit/app_server/test_event_router.py -q
uv run ruff format --config dev_config/python/ruff.toml openhands/app_server/event/event_service_base.py openhands/app_server/event/filesystem_event_service.py tests/unit/app_server/test_filesystem_event_service.py
uv run ruff check --config dev_config/python/ruff.toml openhands/app_server/event/event_service_base.py openhands/app_server/event/filesystem_event_service.py tests/unit/app_server/test_filesystem_event_service.py
uv run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml
git diff --check
```

`make install-pre-commit-hooks` was attempted first but failed because `poetry` is not installed in this environment. The configured Python pre-commit run above passed via `uv`.

## Video/Screenshots

Not applicable; backend event service change only.

## Type

- [x] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This only optimizes the filesystem V1 search path. AWS/GCS backends still use the default full-event metadata fallback so this PR does not change their storage behavior or cursor semantics.

Issue #12705, opened 2026-01-31
RFC: Efficient Bi-directional Event Loading for V1 Conversations


jpshackelford:
## Problem

V1 conversations with long event histories experience slow load times because:

1. **WebSocket replays ALL historical events one-by-one** (no batching, no compression)
2. **REST prefetch and WebSocket duplicate work** - both fetch the same events
3. **Users wait to see the current state** while oldest events load first

For a 500-event conversation, this means 500 individual WebSocket messages and potential double-fetching of the entire history.

## Proposed Solution: Bi-directional Event Loading

Load events from both ends of the timeline simultaneously:

```
Timeline:
[oldest] ←←←← REST (backwards) ←←←← [newest] →→→→ WebSocket (forward) →→→→ [new events]
```

### Flow

1. **REST fetches newest events first** (`sort_order=TIMESTAMP_DESC`)
2. **User sees current state immediately**
3. **WebSocket connects with** `after_timestamp` parameter (set to oldest REST event timestamp)
4. **WebSocket only delivers NEW events** after that timestamp - zero duplication
5. **REST continues paginating backwards** in background, prepending older events

### Benefits

* **Instant perceived load**: User sees most recent/relevant state immediately
* **Zero duplication**: REST handles history, WebSocket handles future
* **Efficient transfer**: REST batching + compression for bulk data
* **Real-time from start**: WebSocket connected early for live updates

## Implementation Status

| Component | Status | Details |
| -- | -- | -- |
| `after_timestamp` WebSocket param (agent-server) | ✅ Done | [software-agent-sdk#1880](<https://github.com/OpenHands/software-agent-sdk/pull/1880>) |
| Event store sorted insertion | ✅ Done | `use-event-store.ts` sorts by timestamp |
| **Frontend bi-directional loading** | ❌ Not Done | See remaining work below |

### 1\. Update `EventService.searchEventsV1` for backwards pagination

```typescript
static async searchEventsV1(
  conversationId: string, 
  options?: { 
    limit?: number;
    sort_order?: "TIMESTAMP_ASC" | "TIMESTAMP_DESC";
    cursor?: string;
  }
) {
  const { data } = await openHands.get<{
    items: OpenHandsEvent[];
    next_page_id?: string;
  }>(`/api/v1/conversation/${conversationId}/events/search`, {
    params: { 
      limit: options?.limit ?? 100,
      sort_order: options?.sort_order,
      cursor: options?.cursor,
    },
  });
  return data;
}
[...]

PR #15188, opened 2026-07-08
perf(event-service): avoid full-history page validation

### LINKED ISSUE #13335: Bug: EventServiceBase.search_events() loads all events before applying limit/page_id

dexhunter:
HUMAN:

This PR reduces the cost of loading paginated local conversation histories. It keeps the existing event order and pagination contract, including when stored JSON has a timestamp but is not a valid Event.

AGENT:

---

## Why

`EventServiceBase.search_events()` currently reads and Pydantic-validates every stored event before applying an unfiltered page limit. Fetching the first pages of a large local conversation therefore repeats full discriminated-union validation across the entire history.

This addresses #13335, which reports that `search_events()` loads all events before applying `limit` and `page_id`. It is also related to #12705 and the earlier attempt in #14447.

## What changed

- Added an optional `_scan_event_timestamps()` hook to `EventServiceBase`.
- Implemented the hook for `FilesystemEventService` with a plain JSON decode that extracts each stored timestamp without constructing an `Event` model.
- For unfiltered filesystem searches, sort the scanned candidates and fully validate only the ordered prefix needed to skip valid events before the offset, fill the requested page, and find one valid lookahead event.
- Keep invalid timestamp-bearing JSON from consuming an offset or shortening a page.
- Keep filtered searches and remote backends on the existing full-load path.

The fast path still reads and JSON-decodes all candidate files to obtain timestamps. The improvement is specifically fewer full Pydantic `Event` validations: from `O(N)` per page to a bounded ordered prefix of approximately `O(offset + limit)`, plus invalid records and batch lookahead.

## Measured impact

Strict fixture: 3,003 valid `TokenEvent` files, three timestamp-bearing invalid Event payloads placed before/inside/after page boundaries, and two consecutive 100-event page requests.

| Metric | Before | After | Change |
|---|---:|---:|---:|
| `event_page_full_validations` | 6,012 | 320 | **94.7% fewer (18.8x)** |
| Two-page wall time, run 1 | 1.830 s | 0.442 s | **4.14x faster** |
| Two-page wall time, run 2 | 1.624 s | 0.398 s | **4.08x faster** |

Both strict runs produced the same structural metric. Wall time is secondary and was measured on the same machine; the validation count is deterministic and hardware-independent.

Correctness gates compare the fast path with the existing full-load path for:

- ascending and descending order;
- first and non-zero-offset pages;
- timestamp-bearing invalid Event payloads;
- equal timestamps and stable ordering;
- `limit=0` and invalid-only storage;
- matching items and `next_page_id` values.

## Autoresearch

[Public autoresearch trajectory](https://dashboard.weco.ai/share/kv08hJASdX308QaGaSjqx-njM5KSMwEc)

## Issue Number

Fixes #13335

Related: #12705, #14447

## How to Test

```bash
pytest tests/unit/app_server/test_filesystem_event_service.py \
  tests/unit/app_server/test_aws_event_service.py \
  tests/unit/app_server/test_event_router.py -q
```

Result: `53 passed`.

Also passed:

```bash
uvx --from ruff==0.12.5 ruff check --fix .
uvx --from ruff==0.12.5 ruff format --check \
  openhands/app_server/event/event_service_base.py \
  openhands/app_server/event/filesystem_event_service.py \
  tests/unit/app_server/test_filesystem_event_service.py
mypy --config-file dev_config/python/mypy.ini \
  openhands/app_server/event/event_service_base.py \
  openhands/app_server/event/filesystem_event_service.py
pre-commit run -c dev_config/python/.pre-commit-config.yaml --files \
  openhands/app_server/event/event_service_base.py \
  openhands/app_server/event/filesystem_event_service.py \
  tests/unit/app_server/test_filesystem_event_service.py
```

## Video/Screenshots

N/A (backend performance change).

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The hook defaults to `None`, so AWS and Google Cloud event services retain the previous behavior.
- The implementation preserves Python's stable timestamp ordering by scanning and sorting the same `_search_paths()` sequence used by the full-load path.
- This overlaps #14447. I have asked maintainers which implementation they prefer and will follow their direction.

### DIFF
--- a/openhands/app_server/event/event_service_base.py
+++ b/openhands/app_server/event/event_service_base.py
@@ -63,6 +63,22 @@ class EventServiceBase(EventService, ABC):
 
         return await asyncio.gather(*(load_event(path) for path in paths))
 
+    async def _scan_event_timestamps(
+        self, paths: list[Path]
+    ) -> list[tuple[Path, str]] | None:
+        """Return a (path, timestamp) pair per readable event, or None.
+
+        Backends where reading the raw stored payload is much cheaper than a
+        full ``Event`` validation (e.g. the local filesystem) can override
+        this so that unfiltered ``search_events`` pages only fully load the
+        ordered prefix needed to reach the requested page and its lookahead,
+        instead of the entire conversation history.
+        Unreadable events must be omitted, matching ``_load_event`` failure
+        semantics. Returning None (the default) means no cheap scan is
+        available and ``search_events`` falls back to loading every event.
+        """
+        return None
+
     async def get_conversation_path(self, conversation_id: UUID) -> Path:
         """Get a path for a conversation. Ensure user_id is included if possible."""
         path = self.prefix
@@ -105,6 +121,45 @@ class EventServiceBase(EventService, ABC):
         loop = asyncio.get_running_loop()
         prefix = await self.get_conversation_path(conversation_id)
         paths = await loop.run_in_executor(None, self._search_paths, prefix)
+
+        # Fast path for unfiltered queries (the common "open a conversation"
+        # page pattern): sort and paginate on cheaply-scanned timestamps and
+        # fully load only the ordered prefix needed for the requested page,
+        # instead of loading and validating every event in the conversation.
+        if kind__eq is None and timestamp__gte is None and timestamp__lt is None:
+            scanned = await self._scan_event_timestamps(paths)
+            if scanned is not None:
+                scanned.sort(
+                    key=lambda path_and_timestamp: path_and_timestamp[1],
+                    reverse=(sort_order == EventSortOrder.TIMESTAMP_DESC),
+                )
+                start_offset = int(page_id) if page_id else 0
+                valid_events_seen = 0
+                items: list[Event] = []
+                has_next_page = False
+                batch_size = max(1, min(_event_load_concurrency(), limit + 1))
+                for index in range(0, len(scanned), batch_size):
+                    paths_to_load = [
+                        path for path, _ in scanned[index : index + batch_size]
+                    ]
+                    for event in await self._load_events_from_paths(paths_to_load):
+                        if not event:
+                            continue
+                        if valid_events_seen < start_offset:
+                            valid_events_seen += 1
+                            continue
+                        if len(items) < limit:
+                            items.append(event)
+                            valid_events_seen += 1
+                            continue
+                        has_next_page = True
+                        break
+                    if has_next_page:
+                        break
+                return EventPage(
+                    items=items,
+                    next_page_id=str(start_offset + limit) if has_next_page else None,
+                )
 
         events = await self._load_events_from_paths(paths)
         # Convert datetime filters to ISO strings so they can be compared
--- a/openhands/app_server/event/filesystem_event_service.py
+++ b/openhands/app_server/event/filesystem_event_service.py
@@ -1,4 +1,6 @@
+import asyncio
 import glob
+import json
 import logging
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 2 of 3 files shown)


dexhunter:
I found #14447 during a fresh duplicate audit. It targets the same metadata-first filesystem pagination path, so the overlap is real even though this PR was developed independently.

This version is currently mergeable and adds broader parity/corrupt-file/load-count coverage plus a measured two-page benchmark, while #14447 is older and currently conflicting. I do not want to duplicate or displace that contributor's work. Maintainers, would you prefer that I close this PR in favor of #14447, or keep this implementation as the review candidate? I will follow your preference.

dexhunter:
Rebased this PR onto current main and fixed the invalid timestamp-bearing JSON pagination case. The fast path now counts only successfully validated Events toward the offset and page, with coverage for both sort orders, non-zero offsets, limit=0, invalid-only storage, and stable equal timestamps.

Fresh validation on head `7571b23d0`:
- strict `event_page_full_validations`: `6,012 -> 320` across two 100-event pages (94.7% fewer; reproduced three times)
- focused filesystem, AWS, and router tests: `53 passed`
- Ruff, mypy, pre-commit, and `git diff --check`: passed

I also corrected the complexity claim in the PR body: the optimization bounds full Event validation to the ordered prefix needed for the page; it still reads and JSON-decodes candidate files for timestamp sorting. Remote backends remain on the existing path.

dexhunter:
This bounds full `Event` validation on unfiltered `search_events` pages to the ordered prefix a page actually needs, instead of validating every stored event first: on a 3,003-event fixture, serving two 100-event pages goes from 6,012 full validations to 320, with pagination output and ordering unchanged. That counter is deterministic and hardware-independent, and it still reproduces on current `main`. All 23 checks are green, there are no unresolved review threads, and the branch has no conflicts against `main` (it is behind by 37 commits, so I'm happy to rebase it whenever that's useful). The older overlapping PR #14447 was closed on 2026-07-19, so this is the only open change on that path now. If it would make review easier I'm glad to narrow the scope or split the test file out.

neubig:

Thank you for understanding.

PR #15190, opened 2026-07-08
fix(oss-lifespan): fallback SQLite DB recreation on alembic open failure

afonsoft:
HUMAN:

AGENT:

---

## Why

The OSS app server can fail during lifespan startup with `sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) unable to open database file` raised from `command.upgrade` in `oss_app_lifespan_service.py`. When the SQLite file is corrupted or the parent directory is unwritable, the container never passes the lifespan phase.

## Summary

- Wrapped `command.upgrade(alembic_cfg, 'head')` in `run_alembic` with a `try...except` that catches `sqlalchemy.exc.OperationalError` and generic `SQLAlchemyError`.
- For SQLite failures:
  - Determines the database path from `get_global_config().db_session`.
  - Removes the existing `openhands.db` file if present.
  - Creates the parent persistence directory with `0o775` permissions and `chmod`s it when possible.
  - Retries the Alembic migration so the database is recreated from scratch.
- Logs a warning before removing the file and a warning when the migration is retried successfully.

## How to Test

Start the app with a missing or unwritable SQLite database path (e.g., `FILE_STORE_PATH=/.openhands` and the file removed or permission-denied). The lifespan should now recreate the database and continue startup.

## Type

- [x] Bug fix

## Notes

This fallback only applies to SQLite (`db_session.host` is not set). PostgreSQL and other database errors are re-raised.

Requested by: @afonsoft

### DIFF
--- a/openhands/app_server/app_lifespan/oss_app_lifespan_service.py
+++ b/openhands/app_server/app_lifespan/oss_app_lifespan_service.py
@@ -1,12 +1,16 @@
 from __future__ import annotations
 
+import logging
 import os
 from pathlib import Path
 
 from alembic import command
 from alembic.config import Config
+from sqlalchemy.exc import OperationalError, SQLAlchemyError
 
 from openhands.app_server.app_lifespan.app_lifespan_service import AppLifespanService
+
+logger = logging.getLogger(__name__)
 
 
 class OssAppLifespanService(AppLifespanService):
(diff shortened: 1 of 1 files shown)


neubig:

Thank you for understanding.

PR #15205, opened 2026-07-09
fix(enterprise): tighten org_id resolution in require_permission for path-less routes

simonrosenberg:
HUMAN:

Follow-up from reviewing the org-scoping model used across `require_permission`-gated routes. Filing this directly against `main` since the affected helper is already in production, independent of any in-flight feature branch.

AGENT:

## Summary
- Routes that gate access via `require_permission(...)` normally bind its `org_id` parameter from an `{org_id}` path segment (e.g. `org_profiles.py`'s `/{org_id}/profiles...`), so the permission check and the route handler always agree on which org is in scope.
- A smaller set of routes have no such path segment and instead resolve the acting org from the request context (`EFFECTIVE_ORG_ID`, precedence: API-key binding → `X-Org-Id` header → the user's current org). Because `require_permission`'s `org_id` parameter has no path value to bind to on these routes, FastAPI falls back to treating it as an ordinary, optional query parameter — one that was previously trusted directly rather than cross-checked against the route's own effective-org resolution.
- This meant the org used for the *permission check* and the org used by the *route handler* could, in principle, diverge on these routes.

## Fix
`permission_checker` now discards any `org_id` value unless it was actually bound from the URL path (checked via `request.path_params`), always falling back to the same resolver `EFFECTIVE_ORG_ID` uses otherwise. Path-scoped routes are completely unaffected.

## Test plan
- [x] Added `TestRequirePermissionQueryStringOrgIdBypass` in `test_authorization.py` — proves the discarded case, the unaffected default case, and the unaffected path-scoped case; verified the core regression test fails against the pre-fix code and passes after.
- [x] Filled a pre-existing gap where `EDIT_ORG_SETTINGS` had no role-permission assertions in `TestRolePermissions`.
- [x] Full run: `enterprise/tests/unit/test_authorization.py` + all org/auth-adjacent suites (`test_saas_user_auth_effective_org.py`, `test_org_git_claims.py`, `test_orgs.py`, `test_super_admins.py`, `test_saas_stores_effective_org.py`) — 268 passed, 0 failed.
- [x] `ruff check` / `ruff format --check` (enterprise config) and `pre-commit run mypy` clean on both changed files.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-a700dc5   docker.openhands.dev/openhands/openhands:a700dc5
```


VascoSch92 (approved):

simonrosenberg:

simonrosenberg:
Fix was implemented in main (0 diff after rebase)

PR #15209, opened 2026-07-09
fix(app-server): fix observability-metadata clobber; drop redundant repo probe

simonrosenberg:
HUMAN:

AGENT:

---

## Why

Two issues in the observability metadata on the conversation-request builders (`live_status_app_conversation_service.py`):

1. **Clobber (the bug):** both builders built `observability_metadata` from `_build_observability_context` (`app`/`conversation_id`/`agent_kind`) and then, two lines later, **reassigned** it from `_build_observability_metadata` (repo/branch/commit). The first dict was discarded — `app`/`conversation_id`/`agent_kind` never reached Laminar. Introduced by OpenHands/OpenHands#14431 landing on top of OpenHands/OpenHands#15059.

2. **Wrong layer:** `_build_observability_metadata` derived repo identity from the pre-selected `selected_repository`, which covers only ~23% of repo-bearing conversations (production trace analysis, OpenHands/enterprise#58). The dominant flow — a conversation that clones a repo *after* start — was never covered, and the `_resolve_head_commit` helper added a `git rev-parse` round-trip to the conversation startup path for the minority that were.

## Summary

- Delete `_build_observability_metadata` and `_resolve_head_commit`. Repo identity is now resolved **live from the workspace in the agent-server** (software-agent-sdk#4054) and merged onto the trace root span for **all** conversations, including clone-after-start — subsuming the request-time path and removing the startup round-trip.
- Both builders use `_build_observability_context` directly: `observability_metadata = {app, conversation_id, agent_kind}`, fixing the clobber.
- Tags are unchanged: `app:` / `agent_kind:` always, plus `repo:` / `branch:` / `git_provider:` for pre-selected conversations — so repo filtering keeps working (incl. during the window before the SDK bump ships).

## Issue Number

Part of OpenHands/enterprise#58 (the app-server half). Repo-identity backfill is software-agent-sdk#4054.

## How to Test

`_build_observability_context` is a pure static method:
```python
md, tags = svc._build_observability_context(cid, agent_kind='openhands',
                                            selected_repository='o/r',
                                            selected_branch='main', git_provider=github)
assert md == {'app':'openhands','conversation_id':str(cid),'agent_kind':'openhands'}
assert 'repo:o/r' in tags and 'branch:main' in tags and 'git_provider:github' in tags
```
Live: start any conversation and confirm the Laminar trace now carries `app`/`conversation_id`/`agent_kind` in metadata (previously dropped). Once the SDK bump (software-agent-sdk#4054) lands, `repo`/`branch`/`commit` appear for repos cloned after start too.

## Type

- [x] Bug fix

## Notes

- Coverage gap during the deploy window: between this merging and the SDK bump landing, the ~23% pre-selected conversations lose the `repo`/`branch`/`commit` **metadata fields** they had via `_build_observability_metadata`. They remain filterable by the `repo:`/`branch:` **tags** (unchanged), and regain full metadata (incl. clone-after-start) once the SDK bump ships. Sequence merge accordingly if the gap matters.
- Pure app-server change; no schema change. Trace metadata is write-only to Laminar, so the removed `repo_name`/`selected_branch` metadata keys have no downstream reader.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-0a81e4a   docker.openhands.dev/openhands/openhands:0a81e4a
```

### DIFF
--- a/openhands/app_server/app_conversation/live_status_app_conversation_service.py
+++ b/openhands/app_server/app_conversation/live_status_app_conversation_service.py
@@ -1309,8 +1309,11 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
     ) -> tuple[dict[str, Any], list[str]]:
         """Build trace metadata and tag filters for a conversation.
 
-        Metadata uses explicit field names such as ``repo_name`` and
-        ``selected_branch``. Tags intentionally keep the concise
+        Metadata carries the conversation-identity fields (``app``,
+        ``conversation_id``, ``agent_kind``). Repo identity (repo/branch/commit)
+        is resolved live from the workspace by the agent-server and merged onto
+        the trace's root span there, so it covers repos cloned after start too
+        (not just repo-preselected conversations). Tags keep the concise
         ``repo:`` / ``branch:`` prefixes used by LiteLLM metadata filters.
         """
         metadata: dict[str, Any] = {
@@ -1321,15 +1324,11 @@ class LiveStatusAppConversationService(AppConversationServiceBase):
         tags = ['app:openhands', f'agent_kind:{agent_kind}']
 
         if selected_repository:
-            metadata['repo_name'] = selected_repository
             tags.append(f'repo:{selected_repository}')
         if selected_branch:
-            metadata['selected_branch'] = selected_branch
             tags.append(f'branch:{selected_branch}')
         if git_provider:
-            provider = git_provider.value
-            metadata['git_provider'] = provider
-            tags.append(f'git_provider:{provider}')
+            tags.append(f'git_provider:{git_provider.value}')
 
         return metadata, tags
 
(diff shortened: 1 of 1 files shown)


simonrosenberg:
Closing in favor of OpenHands/OpenHands#15215 for the Problem-1 fix.

Both PRs fix the same clobber — accumulate the observability metadata instead of letting `_build_observability_metadata` overwrite `_build_observability_context`. OpenHands/OpenHands#15215 takes the additive/merge approach, is tested and human-reviewed, and preserves the creation-time `selected_repository` semantic (the point @hxaxd rightly makes in https://github.com/OpenHands/enterprise/issues/58#issuecomment-4927950521 — the user's repo selection at creation is a distinct signal, not something a live probe replaces).

This PR's delete approach — drop the request-time repo resolution and rely on the SDK live probe (OpenHands/software-agent-sdk#4054) — is a separate repo-identity **cleanup**, not part of the bug fix, and it was incomplete (it changes `observability_metadata` values without updating the existing `test_build_request_*observability_metadata` assertions). Deferring it to a non-blocking follow-up once OpenHands/OpenHands#4054 ships, at which point removing the now-redundant request-time path drops no creation-time signal.

SDK OpenHands/OpenHands#4054 (live backfill of the still-open trace root span) stands on its own and is unaffected by this.

PR #15212, opened 2026-07-09
fix: prevent webhook-driven DB connection leaks

tofarr:
HUMAN:

I tested explicitly with endpoints. 

AGENT:

## Why

In production we observed a slowly-growing count of idle connections from the app server to the OpenHands database. At rest the number increased by roughly 100 per minute in bursts (e.g. 588 → 732 in one minute), and the Datadog traces showed `TimeoutError` in `asyncpg.connect` with a clear upward trend. The root cause was the webhook pipeline holding pool connections across slow work.

Three patterns in the code were responsible:

1. `SQLEventCallbackService` held an `AsyncSession` for its entire lifetime, so the `asyncio.gather(...)` of callback processors (webhook deliveries, the `SetTitleCallbackProcessor` title poll) pinned a pool connection for seconds-to-minutes per webhook. With `pool_size=25` + `max_overflow=10`, a burst of concurrent webhooks could exhaust the pool and starve subsequent `asyncpg.connect` attempts — exactly the `asyncio.exceptions.CancelledError → TimeoutError` chains in the error traces.
2. `on_event` kicked off the background work with a bare `asyncio.create_task(...)` and discarded the handle. Any unhandled exception became a "Task exception was never retrieved" warning (visible in the second attached traceback) and during graceful shutdown the task was cancelled with no guarantee the async-generator cleanup would reach its `await db_session.close()`.
3. In SaaS deployments, `DbSessionInjector._gcp_connector` was created lazily but never closed: each worker respawn leaked the connector's background cert refresh task and aiohttp `ClientSession`, and a broken pool close during cancellation could mask the original exception in the request handler.

## Summary

1. **`SQLEventCallbackService` opens its own short-lived sessions.** The service now holds an `async_sessionmaker` (which doesn't open a connection on construction), and every public method opens its own `AsyncSession` for the duration of its DB work. Slow callback processors run with **zero** connections held — only the SELECT in `get_active_callbacks` and the COMMIT in `persist_callback_results` need the pool. The service is no longer an async context manager, so callers cannot get the lifecycle wrong.
2. **`on_event` dispatches via FastAPI `BackgroundTasks`** instead of `asyncio.create_task`. Starlette awaits the work after the response is sent and observes exceptions via the framework's error logging.
3. **`DbSessionInjector.close()` and SaaS lifespan wiring.** `close()` calls the GCP connector's `close_async` (with a sync `close` fallback for older connector versions) and disposes the async engine. The **SaaS** lifespan `__aexit__` calls it so worker respawns don't leak connector resources. The OSS lifespan path is intentionally untouched — OSS doesn't use the GCP Cloud SQL connector, so there's nothing GCP-specific to release there.
4. **`await db_session.close()` is now wrapped in `contextlib.suppress(Exception)`** so a broken or cancelled close can't mask the original exception or leave a pool slot checked out. The `except Exception:` (not `BaseException`) is preserved on purpose, so `CancelledError`, `KeyboardInterrupt` and `SystemExit` keep propagating.

## Files

- `openhands/app_server/event_callback/event_callback_service.py` — abstract base now declares `get_active_callbacks` and `persist_callback_results`.
- `openhands/app_server/event_callback/sql_event_callback_service.py` — service rewritten to use `async_sessionmaker` and short-lived sessions; injector simplified to yield with the cached maker.
- `openhands/app_server/event_callback/webhook_router.py` — `on_event` accepts a `BackgroundTasks` param; `_run_callbacks_in_bg_and_close` simplified now that the service handles its own sessions.
- `openhands/app_server/services/db_session_injector.py` — `close()` + cancellation-safe cleanup of `db_session.close()`.
- `enterprise/server/app_lifespan/saas_app_lifespan_service.py` — SaaS lifespan `__aexit__` calls `await db_session.close()` so the GCP connector and engine are released on graceful shutdown.
- `tests/unit/app_server/test_sql_event_callback_service.py` — updated to use `async_session_maker` for setup/verification (service no longer exposes `db_session`).
- `tests/unit/app_server/test_db_session_injector.py` — three new tests cover cancellation cleanup and connector teardown.
- `tests/unit/app_server/test_webhook_router_stats.py` — adapted to the `BackgroundTasks`-based dispatcher.

## How to Test

1. Run the affected unit tests:

   ```bash
   poetry run pytest \
     tests/unit/app_server/test_sql_event_callback_service.py \
     tests/unit/app_server/test_webhook_router_stats.py \
     tests/unit/app_server/test_db_session_injector.py \
     tests/unit/app_server/test_db_session_dependency.py \
     tests/unit/app_server/test_webhook_router_auth.py \
     tests/unit/app_server/test_webhook_router_parent_conversation.py \
     tests/unit/app_server/test_webhook_router_auto_title.py \
     tests/unit/app_server/test_webhook_router_acp.py \
     tests/unit/app_server/test_webhook_router_tags.py
   ```

   All 120 tests pass.

2. Pre-commit hooks (used during commit, all green):

   ```bash
   pre-commit run --config ./dev_config/python/.pre-commit-config.yaml
   ```

3. In production: confirm the leaked-connection count plateaus. After deploy, the idle-connections metric should no longer climb during webhook bursts, and the "Task exception was never retrieved" log noise from `_run_callbacks_in_bg_and_close` should disappear.

## Video/Screenshots

Event callbacks work with conversations as expected:
image

---

Generated by OpenHands AI agent on behalf of the user. Pre-commit hooks (ruff, ruff-format, mypy) ran during the commit and all passed.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-be3d417   docker.openhands.dev/openhands/openhands:be3d417
```


hieptl (approved):
Thank you! 🙏

PR #15220, opened 2026-07-09
fix(enterprise): cascade-delete conversation_cost_events on conversation delete

tofarr:
HUMAN:

AGENT:

---

## Why

Deleting a V1 conversation in prod fails with:

```
Error deleting V1 conversation 23742af5-6dfb-498c-8d96-787e407e7347:
(sqlalchemy.dialects.postgresql.asyncpg.IntegrityError)
<class 'asyncpg.exceptions.ForeignKeyViolationError'>: update or delete on
table "conversation_metadata" violates foreign key constraint
"conversation_cost_events_conversation_id_fkey" on table "conversation_cost_events"
```

Root cause: the `conversation_cost_events` table added in migration 135 (commit `acc8412ae`, "feat: track per-event cost deltas") declares its FK to `conversation_metadata.conversation_id` without `ON DELETE CASCADE`. Since cost-delta rows are inserted on every positive-cost agent-metrics update, essentially every active conversation has at least one row referencing it, so `SQLAppConversationInfoService.delete_app_conversation_info` blows up before it can delete the metadata row.

The cost-delta stream is an audit trail whose running total is already mirrored on `conversation_metadata.accumulated_cost`, so cascade-deleting the audit rows when the conversation is removed is safe — and matches the precedent set by migration 051, which did the same surgery on `conversation_callbacks` for the exact same reason.

## Summary

- New Alembic migration `136_update_conversation_cost_events_fk_to_cascade.py` drops `conversation_cost_events_conversation_id_fkey` and recreates it with `ondelete='CASCADE'`.
- `StoredConversationCostEvent.conversation_id` model declares `ondelete='CASCADE'` so fresh-schema creation (tests, dev, on-prem re-creates) matches prod.
- New regression tests in `enterprise/tests/unit/storage/test_conversation_cost_events_cascade.py`:
  - `test_cost_event_fk_declares_cascade` introspects the model FK and asserts `ondelete == 'CASCADE'`.
  - `test_delete_conversation_cascades_cost_events` runs an end-to-end delete against SQLite with `PRAGMA foreign_keys=ON` and verifies that deleting a conversation also removes its cost-event rows (and leaves unrelated conversations' events intact).

## Issue Number

None.

## How to Test

1. Pull the branch and run the new unit tests:
   ```
   cd enterprise
   PYTHONPATH=".:$PYTHONPATH" poetry run pytest \
     tests/unit/storage/test_conversation_cost_events_cascade.py -v
   ```
   Both tests should pass.
2. Run the broader enterprise storage suite to confirm no regressions:
   ```
   cd enterprise
   PYTHONPATH=".:$PYTHONPATH" poetry run pytest tests/unit/storage/ -q
   ```
   Expect 226 passed, 2 skipped (unchanged from `main`).
3. To exercise the migration against a real PostgreSQL DB, run `alembic upgrade head` and inspect the constraint:
   ```
   psql ... -c "\d conversation_cost_events"
   ```
   `conversation_id`'s FK should now show `ON DELETE CASCADE`.
4. End-to-end (requires staging): create a conversation, let it generate a couple of cost-event rows, then delete it via the V1 delete API — the delete should succeed instead of returning `ForeignKeyViolationError`.

## Video/Screenshots

Deletes now work as expected:
image
image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **No application-level pre-delete of cost events was added.** Migration 051 also relies purely on DB-level cascade for `conversation_callbacks`; doing the same here keeps `delete_app_conversation_info` a single statement.
- **Convention suggestion (follow-up):** this is the second time a child table was added to `conversation_metadata` without `ON DELETE CASCADE` and only caught in prod. Worth establishing a code-review checklist item: every new FK from a child table to `conversation_metadata.conversation_id` must include `ondelete='CASCADE'`.
- Migration revision chain: `134 → 135 → 136`. Safe to deploy on top of any env that already ran migration 135.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-67ab546   docker.openhands.dev/openhands/openhands:67ab546
```

### DIFF
--- a/enterprise/migrations/versions/136_update_conversation_cost_events_fk_to_cascade.py
+++ b/enterprise/migrations/versions/136_update_conversation_cost_events_fk_to_cascade.py
@@ -1,0 +1,58 @@
+"""Update conversation_cost_events foreign key to cascade deletes.
+
+Without ON DELETE CASCADE, deleting a row from conversation_metadata fails
+with a ForeignKeyViolationError when conversation_cost_events rows still
+reference it. The cost-delta stream is an audit trail whose running total
+is already maintained on conversation_metadata.accumulated_cost, so
+cascade-deleting the audit rows when the conversation is removed is safe.
+
+Revision ID: 136
+Revises: 135
+Create Date: 2026-07-09
+
+"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = '136'
+down_revision = '135'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # Drop the existing foreign key constraint
+    op.drop_constraint(
+        'conversation_cost_events_conversation_id_fkey',
+        'conversation_cost_events',
+        type_='foreignkey',
+    )
+
+    # Add the new foreign key constraint with cascade delete
+    op.create_foreign_key(
+        'conversation_cost_events_conversation_id_fkey',
+        'conversation_cost_events',
+        'conversation_metadata',
+        ['conversation_id'],
+        ['conversation_id'],
+        ondelete='CASCADE',
+    )
+
+
+def downgrade():
+    # Drop the cascade delete foreign key constraint
+    op.drop_constraint(
+        'conversation_cost_events_conversation_id_fkey',
+        'conversation_cost_events',
+        type_='foreignkey',
+    )
+
+    # Recreate the original foreign key constraint without cascade delete
+    op.create_foreign_key(
+        'conversation_cost_events_conversation_id_fkey',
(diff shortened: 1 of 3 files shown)


saurya (approved):

hieptl (approved):
Thank you! 🙏

PR #15221, opened 2026-07-10
fix(frontend): add jittered rate-limit backoff

aivong-openhands:
HUMAN:

AGENT:

---

## Why

Clients that receive HTTP 429s can retry on the same schedule. Retrying together creates synchronized bursts of traffic that can worsen rate limiting. Exponential backoff with jitter is a common solution to this thundering-herd problem: it spreads retries out while reducing unnecessary load. This follow-up to #14660 applies that approach.

## Summary

- Add a reusable frontend rate-limit retry-delay utility with exponential backoff and bounded jitter.
- Treat `Retry-After` as a server-directed minimum delay, then add jitter so clients do not retry simultaneously.
- Apply the utility to global query/mutation retries and the resume-related conversation fetch, with focused regression coverage.

## Issue Number

Related to #14656. Follow-up to #14660.

## How to Test

```bash
cd frontend
npm test -- __tests__/utils/rate-limit-retry.test.ts __tests__/hooks/query/use-user-conversation.test.tsx __tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx __tests__/query-client-config.test.ts
npm run lint:fix
npm run typecheck
npm run build
```

## Video/Screenshots

Not included. This is retry behavior covered by focused unit tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

`Retry-After` remains a lower bound; the added jitter can only delay a retry further. `npm run lint:fix` passed with existing warnings outside this change.

### DIFF
--- a/frontend/__tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
+++ b/frontend/__tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
@@ -1,6 +1,7 @@
-import { renderHook, waitFor } from "@testing-library/react";
+import { act, renderHook, waitFor } from "@testing-library/react";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-import { describe, expect, it, vi, beforeEach } from "vitest";
+import { AxiosError } from "axios";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import React from "react";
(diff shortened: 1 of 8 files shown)


aivong-openhands:
Superseded by #15236, recreated from an upstream OpenHands/OpenHands branch so the trusted Docker image workflow can run.

PR #15222, opened 2026-07-10
docs: add external supervisor pattern documentation

jiezeng2004-design:
HUMAN:

This PR adds a documentation-only explanation of how an external supervisor can complement OpenHands sandboxing by constraining local agent actions and preserving auditable evidence.

AGENT:

---

## Summary

This is a **docs-only** change. It adds `docs/external-supervisor-pattern.md`, a new document describing the **external supervisor pattern** — an architectural approach for governing what an autonomous coding agent is permitted to do, by placing policy enforcement in a process the agent cannot influence.

Ready for maintainer review; feedback on whether this kind of pattern/architecture document is welcome in this repo, and on the right placement and scope, is welcome before merge.

## What the document covers

- **What the pattern is**: an external supervisor mediates every action an agent requests (shell commands, file mutations) before it reaches the runtime, and the supervisor runs outside the agent's own execution context so it cannot be disabled or bypassed by the agent.
- **How it relates to OpenHands' existing sandbox model**: the sandbox (described in `openhands/app_server/sandbox/README.md`) isolates *where* the agent runs; a supervisor governs *what* the agent is permitted to do. The two layers are designed to compose, not compete.
- **Security boundaries** the pattern enforces, each defined independently of any specific tool:
  - **Workspace confinement** — all file access restricted to a configured root, with canonical-path resolution to defeat traversal and symlinks.
  - **Command allowlists** — exact, structural matching (command + arguments), failing closed rather than relying on a denylist.
  - **Sensitive-file blocking** — refusing credentials, `.env`, SSH keys, cookies, and browser state even inside the workspace root.
  - **Auditable evidence** — structured records of every approved/denied decision, written by the supervisor so the agent cannot alter them.
- **When to use the pattern**, and the trade-off (an added hop between caller and runtime vs. reduced risk of unmediated agent action).
- An **example implementation** (PatchWarden) is referenced, but the document is written so the pattern stands on its own and does not depend on that tool.

## Motivation

The `README.md` explicitly warns that running the agent server without a sandbox means *"the agent will have full access to your filesystem."* Even inside a sandbox, an agent may attempt destructive commands, read sensitive files, or push changes outside a task's intended scope. This document makes a complementary, auditable policy boundary explicit and reusable.

## Why docs-only / no code changes

This PR introduces no code, build, or dependency changes — it only adds a single Markdown file under a new `docs/` directory. The repo currently has no top-level `docs/` directory (project documentation lives at `docs.openhands.dev` and in root-level files such as `Development.md`), so I've placed this under `docs/`. I'm happy to move it elsewhere (e.g. alongside `openhands/app_server/sandbox/README.md`, or as a root-level file) if a different location is preferred.

## Checklist

- [x] Docs-only change; no source, build, or dependency modifications
- [x] PR title follows the `docs:` prefix convention from `CONTRIBUTING.md`
- [x] Content is grounded in existing repo facts (sandbox README, README warnings, external-integrations architecture doc) and cites them as references
- [x] No marketing language; the example tool is referenced only as an instance of the pattern

## Open questions for reviewers

1. Is a pattern/architecture document like this in scope for this repo, or would it fit better in the separate docs site?
2. Is `docs/` the right location, or should it live next to an existing doc?
3. Is the framing of "supervisor complements the sandbox" aligned with how maintainers think about the security model?

Feedback welcome — happy to revise scope, placement, or framing based on review.

### DIFF
--- a/docs/external-supervisor-pattern.md
+++ b/docs/external-supervisor-pattern.md
@@ -1,0 +1,201 @@
+# External Supervisor Pattern
+
+## What is the external supervisor pattern?
+
+The **external supervisor pattern** is an architectural approach for governing
+what an autonomous coding agent is permitted to do, by placing policy
+enforcement in a process that the agent **cannot influence**.
+
+In this pattern, a dedicated *supervisor* sits between an agent caller (for
+example, an automation runner, a CLI, or an MCP host) and the agent runtime.
+Every action the agent requests is mediated by the supervisor, which decides
+whether the action is allowed before it reaches the runtime or the host
+filesystem. The supervisor is "external" because it runs outside the agent's
+own execution context — the agent has no ability to disable, rewrite, or bypass
+it, even if the agent is otherwise capable of arbitrary code execution.
+
+This is distinct from, and complementary to, the sandbox-based isolation that
+OpenHands already provides:
+
+- OpenHands runs agents inside a **sandbox** (such as a Docker container) that
+  isolates the execution environment and limits filesystem access to mounted
+  directories (for example, the `PROJECTS_PATH` bind-mount described in the
+  project `README.md`). The sandbox answers the question *"where does the agent
+  run?"*
+- An external supervisor answers a different question: *"which specific actions
+  is the agent allowed to take, and are they within policy?"* It enforces
+  fine-grained, auditable rules over commands and file changes regardless of
+  where the runtime executes.
+
+The two layers compose: the sandbox contains the blast radius of a misbehaving
+agent, while the supervisor constrains and records what the agent is permitted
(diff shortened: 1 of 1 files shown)


neubig:

Thank you for understanding.

jiezeng2004-design:
Thanks for the clarification.

This PR is documentation-only and does not modify enterprise code, the App Server implementation, or the legacy frontend.

Would it be appropriate to migrate this document to the OpenHands/docs repository, perhaps under sdk/guides/agent-server or sdk/arch?

If so, I’ll update it to the current Mintlify/MDX format, fix the references, and submit a new PR there.

PR #15228, opened 2026-07-10
fix(agent-profiles): honor profile settings in cloud launches

simonrosenberg:
HUMAN:

AGENT:

Implemented and verified by Codex.

---

## What this fixes

In simple terms: selecting an Agent Profile in OpenHands Cloud did not fully control the agent that launched.

Cloud correctly resolved the selected profile, but later launch code replaced parts of it. This could remove the profile's instructions, replace its chosen tools, and prevent `tools: []` from creating a bare agent. The profile was also resolved a second time for provenance, so an edit made during startup could make the recorded profile revision differ from the settings that actually launched.

## User-visible behavior

- Profile instructions are preserved and launch-specific instructions are appended to them.
- `tools: null` receives the normal cloud and browser tools.
- `tools: []` launches a deliberately bare agent.
- An explicit tool list is preserved exactly.
- Profile context, secrets, skills, and disabled skills are retained and merged with launch data.
- Explicit per-conversation profile selection does not change the member's ambient active profile.
- ACP/OpenHands selection and displayed profile provenance come from the same settings snapshot that built the agent.
- Planning launches deliberately continue to use the constrained planning toolset, even if the profile is otherwise bare.
- A profile that disables the switch-LLM tool remains authoritative even when the member has multiple saved LLM profiles.

There are no new UI components or API-contract changes. Existing UI that shows the launched Agent Profile now receives reliable profile ID and revision data.

## Implementation

- Resolve one launch snapshot per conversation-start attempt.
- Thread that snapshot through OpenHands/ACP branching, agent construction, disabled-skill filtering, provider decisions, and provenance stamping.
- Treat the resolved profile as the base configuration and apply cloud launch data additively.
- Preserve the existing managed LLM and MCP overlays.
- Let the SDK profile resolver own built-in switch-LLM tool gating instead of re-enabling the tool later in cloud launch code.

### Real local HTTP validation

Started a local app-server harness using the production `stream-start` router, `LiveStatusAppConversationService`, `SaasSettingsStore`, and SDK profile resolver. It sent real requests to a separately running OpenHands agent-server through `POST /api/conversations`. Only cloud sandbox allocation/workspace bootstrap was replaced with a fixed local sandbox descriptor.

Five conversation launches reached `READY`:

1. Ambient bare profile: zero tools; profile, request, and host instructions were composed; launch secrets and disabled skills survived; provenance recorded the ambient profile at revision 1.
2. Explicit profile override: only `terminal` was present; the referenced MCP server was retained; provenance recorded the override while the ambient pointer remained unchanged.
3. Default-tools profile: cloud defaults included terminal, file editor, task tracker, and browser tools.
4. Planning launch from the bare profile: only the constrained planning tools were present and the planning prompt/instructions were retained.
5. ACP profile: provider, model, MCP configuration, secrets, ACP tag, and profile provenance all matched the resolved profile.

The instrumentation recorded five resolved-profile loads for five starts: exactly one per conversation-start attempt.

### Local automated checks

- `148 passed`: app-conversation service tests under `pytest --forked -n 4`.
- `45 passed`: enterprise Agent Profile tests under `pytest --forked -n 4`.
- Root and enterprise pre-commit checks passed on all touched files, including Ruff, formatting, and mypy.
- The five profile-focused app-conversation tests pass after the switch-tool regression coverage was added.
- The broader SaaS settings/auth regression suites also passed; their existing aiosqlite event-loop shutdown warnings remain unchanged.

During review, the first CI run exposed a nondeterministic UUID generated during pytest collection. The parametrization was made deterministic and the production-style xdist topology was rerun locally.

### GitHub CI

Current head `41c038438` completed all required CI checks successfully, including root and enterprise Python tests and lint, frontend lint, package/version checks, coverage, and app/enterprise image builds for both architectures.

## Issues

Closes OpenHands/enterprise#59
Closes OpenHands/enterprise#51

Related: OpenHands/software-agent-sdk#3713 and OpenHands/software-agent-sdk#3978.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Video/Screenshots

Not applicable; this is a backend correctness change with no visual layout changes.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-45ed7cb   docker.openhands.dev/openhands/openhands:45ed7cb
```


openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -229,13 +230,48 @@ def effective_disabled_skills(user: UserInfo) -> list[str]:
     discovered catalog is a harmless no-op, so no reconciliation is needed
     between the two sources.
     """
-    member = list(user.disabled_skills or [])
+    member = list(getattr(user, 'disabled_skills', None) or [])
```
enyst: Nit: do we really need all these `getattr`, here and below? I seem to see that `user` has the attribute defined: https://github.com/OpenHands/OpenHands/blob/73b72a223d0325ab049af2b6a37b1adc13006d79/openhands/app_server/settings/settings_models.py#L366

enyst (commented):
Left a little nit question inline. We are trying to use stronger types, because they help the agent; they enable the agent-generated code to hit errors when it makes weird stuff, thus I think maybe it reduces the potential for avoidable bugs.

In other codebases, agent-sdk and openhands-cli I’m pretty sure, we have really tried to avoid `getattr` and the like as much as possible. I think not so much in the openhands codebase, it’s an older more chaotic codebase… but IMHO it’s not too late to start, if we can.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
simonrosenberg: Addressed in 60bae3c2c: `effective_disabled_skills` now uses direct typed attributes, with only the nullable ACP `agent_context` check retained. The test double now mirrors the real `UserInfo.disabled_skills=[]` default.

enyst:
@openhands do a /codereview-roasted on this pr. post your feedback as a review on gh api

enyst (approved):
Hi — I’m OpenHands-GPT-5.6 on Engel’s MacBook.

🟢 **Good taste**

I independently traced the launch path against the SDK 1.36.0 contracts and found no blocking issue. The change resolves one launch snapshot and consistently reuses it for agent selection, construction, disabled-skill filtering, ACP metadata, and profile provenance. That removes the prior time-of-check/time-of-use mismatch without adding another source of truth.

The OpenHands branch now preserves the profile’s tri-state tool semantics (`None` = cloud defaults, `[]` = deliberately bare, explicit list = exact list), uses the resolved `AgentContext` as the base while additively composing launch suffixes and secrets, and leaves switch-LLM gating to the SDK setting. The ACP branch likewise keeps the resolved profile settings and deny-list while applying launch-only context. The explicit planning-tool override is clear and intentional.

The regression coverage exercises the important boundaries: tri-state tools, profile/member skill exclusions, profile and launch context composition, switch-tool disablement, planning launches, ACP context, one resolution per start, and matching provenance. The PR also includes concrete runtime evidence, and all required checks pass on the reviewed head.

**[RISK ASSESSMENT]**
This touches the central cloud conversation-start path and both OpenHands and ACP launch behavior, so the blast radius is meaningful. The implementation is narrowly scoped, preserves existing launch overlays, adds strong regression coverage, and has successful end-to-end and CI validation, which keeps the residual risk controlled.

**VERDICT:**
✅ **Worth merging**: The profile is now the authoritative launch base, while cloud-only overlays remain explicit and additive.

**KEY INSIGHT:**
A single resolved snapshot eliminates both configuration drift during startup and provenance drift after startup.

_AI disclosure: This review was generated by OpenHands on Engel’s behalf._

PR #15229, opened 2026-07-10
fix: accept 1 for env enable toggles

william-xue:
HUMAN:
This PR fixes three boolean environment checks that still treated only the literal string `true` as enabled. With this change, `SERVE_FRONTEND=1` and `AWS_S3_SECURE=1` behave as enabled too, matching the project guidance for Helm-style env values.

## What
- Treat `SERVE_FRONTEND=1` the same as `true` when deciding whether to mount the frontend.
- Treat `AWS_S3_SECURE=1` the same as `true` in both S3 event storage and S3 file-store setup.
- Add regression coverage for the `AWS_S3_SECURE=1` path.

## Why
The project guidance says boolean env toggles should accept both `true` and `1`, but these paths only accepted `true`. That can silently disable the frontend mount or make S3 endpoints use `http://` in deployments that set Helm-style `1` values.

## Test
- `git diff --check`
- `uv run pytest tests/unit/app_server/test_aws_event_service.py tests/unit/app_server/file_store/test_file_store.py -q` (45 passed)
- `uv run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files openhands/app_server/app.py openhands/app_server/event/aws_event_service.py openhands/app_server/file_store/s3.py tests/unit/app_server/test_aws_event_service.py tests/unit/app_server/file_store/test_file_store.py`

Note: `make install-pre-commit-hooks` currently stops at dependency install with `pyproject.toml changed significantly since poetry.lock was last generated`; I ran the targeted checks above through `uv run` instead.

### DIFF
--- a/openhands/app_server/app.py
+++ b/openhands/app_server/app.py
@@ -72,7 +72,7 @@ app.include_router(v1_router.router)
 app.include_router(health_router)
 
 # Middleware and static file setup (merged from listen.py)
-if os.getenv('SERVE_FRONTEND', 'true').lower() == 'true':
+if os.getenv('SERVE_FRONTEND', 'true').lower() in ('true', '1'):
     if os.path.isdir('./frontend/build'):
         app.mount(
             '/', SPAStaticFiles(directory='./frontend/build', html=True), name='dist'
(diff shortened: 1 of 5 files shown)


william-xue:
Ready for review. I added the HUMAN note and checked the testing box in the PR description.

william-xue:
All checks are green, including the Python unit suite and lint. The change keeps the existing `true` behavior and adds the documented `1` form consistently for the frontend mount and both S3 code paths, with regression coverage for the S3 cases. When you have a chance, could you review the focused env-parsing change?

neubig:

Thank you for understanding.

PR #15230, opened 2026-07-10
fix(app-server): self-heal stale managed LLM API keys on conversation start (APP-2678)

tofarr:
HUMAN:

AGENT:

---

## Why

[APP-2678](https://linear.app/all-hands-ai/issue/APP-2678) — runtime pods were logging `Authentication Error, Invalid proxy server token passed` from the OpenHands-managed LLM proxy on conversation start. Datadog shows 306 occurrences in 30 days, 104 in the last 7 days, 52+ unique key suffixes, both prod + staging; errors cluster in short bursts per user (e.g. 6 in 7 min) which matches the "user retries a few times" pattern in the report.

The root cause: a user's encrypted `llm_api_key` on the `OrgMember` row had drifted out of sync with the LiteLLM proxy's `LiteLLM_VerificationTokenTable` (typically an admin action, a DB restore, or a botched rotation). The store-side `_ensure_api_key` already self-heals on `store()` — but nothing verified the key on the *read* path used to start a conversation, which is where the 401 actually surfaces.

## Summary

- Mirrors the existing `SaasSettingsStore._ensure_api_key` self-heal onto the conversation-start read path: `LiveStatusAppConversationService._configure_llm_and_mcp` now calls `user_context.ensure_managed_llm_key(user)` before configuring the LLM.
- On miss: deletes the existing key (404 tolerated, just logged), mints a fresh one under the deterministic `get_openhands_cloud_key_alias(user_id, org_id)`, and persists the encrypted value to `org_member._llm_api_key` via `OrgMemberStore.update_all_members_settings_async` so subsequent loads also see the new key.
- Adds `ensure_managed_llm_key` to the `UserContext` and `UserAuth` interfaces (default no-ops), with delegating implementations in `AuthUserContext` / `SpecifyUserContext` (admin no-op) / `SaasUserAuth` / `SaasSettingsStore.verify_and_fix_managed_llm_key` (the actual LiteLLM round-trip + persist).
- BYOR (`has_custom_llm_api_key`) and unmanaged `base_url` are intentionally left untouched.

## Issue Number

APP-2678

## How to Test

- Run `poetry run pytest enterprise/tests/unit/test_saas_settings_store.py -k verify_and_fix_managed_llm_key -v` — covers BYOR no-op, valid-key no-op, full regenerate-and-persist (with deterministic alias + `OrgMemberStore.update_all_members_settings_async` call assertions), and delete-failure tolerance.
- Run `poetry run pytest tests/unit/app_server/test_user_context_ensure_managed_llm_key.py -v` — covers admin no-op, arbitrary-user-id no-op, `AuthUserContext` delegation to `user_auth`, and in-place mutation passthrough.
- Manual: in OpenHands Cloud, a user whose `org_member._llm_api_key` is valid in DB but missing from the LiteLLM proxy `/user/info` will start the next conversation with a fresh key and no 401 in the runtime logs.

## Video/Screenshots

N/A (no UI change; behavior is observable in Datadog — the `title_utils.py:144` "Error generating conversation title with LLM" warnings tied to `sk-...` key suffixes should drop after this ships, for the keys that get exercised again).

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- One extra `LiteLlmManager.verify_existing_key` round-trip per conversation start (no-op for healthy keys). Acceptable because the conversation-start path already talks to LiteLLM anyway.
- The `OrgMember` row is the source of truth on subsequent loads (the in-memory `Settings` is mutated in place so the runtime sees the healed key on this request, and the next `load()` picks it up via `_get_effective_llm_api_key`).
- The `delete_key_by_alias` call is wrapped in a broad try/except because a 404 from the proxy is the common case (no prior key) and we don't want a one-time cleanup glitch to block the self-heal.
- Pre-existing test failures in `test_live_status_app_conversation_service.py` (`TestEffectiveDisabledSkills::*` and `test_build_request_passes_profile_and_member_disabled_skills`) are unrelated and were failing on `main` before this change.

Generated by an AI agent (OpenHands) on behalf of @openhands.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-3b2570f   docker.openhands.dev/openhands/openhands:3b2570f
```


neubig:

Thank you for understanding.

PR #15225, opened 2026-07-10
fix: Enable LIFO database connection pooling

tofarr:
HUMAN:

Using Lifo to gradually recycle connections.

- [x] A human has tested these changes.

AGENT:

---

## Why

Database connection spikes can leave SQLAlchemy pools holding many idle PostgreSQL backends after burst traffic. LIFO pool reuse makes older idle connections stay untouched longer, allowing database-side idle session timeouts to reclaim them more effectively while `pool_pre_ping` handles reconnects on later checkout.

## Summary

- Enable `pool_use_lifo=True` for pooled sync and async database engines in `DbSessionInjector`.
- Enable LIFO for enterprise Alembic migration engine pools.
- Add unit assertions that database pools are configured with LIFO by default.

## Issue Number

N/A

## How to Test

```bash
poetry run pytest tests/unit/app_server/test_db_session_injector.py
poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml
```

Both commands passed locally.

## Video/Screenshots

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

This PR was created by an AI agent (OpenHands) on behalf of the user.

For best idle-connection reclamation, this should be paired with an appropriate PostgreSQL/Cloud SQL `idle_session_timeout`. `pool_pre_ping=True` is already enabled, so stale server-closed connections should be detected on checkout.

@tofarr can click here to [continue refining the PR](https://app.all-hands.dev/conversations/a216f946-1c8a-4cbb-bcb3-5655fbc94edd)

---

To run this PR locally, use the following command:

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-76d01b7   docker.openhands.dev/openhands/openhands:76d01b7
```

PR #15231, opened 2026-07-10
fix(app-server): don't hold DB connections across runtime-api calls

jlav:
HUMAN:

AGENT:

---

## Why

During a production incident on 2026-07-09 (~21:40-22:01 UTC), postgres connections on the application DB stepped from ~350 to ~1,200 and stayed there. During the bursts, app-server pods exhausted their sqlalchemy pools (`QueuePool limit of size 25 overflow 10 reached` after 30s hangs) and requests failed with 500s. The DB showed 76 idle-in-transaction sessions and 79 blocked backends at the worst minute.

Root cause: the request-scoped session opens a transaction on its first read, and the sandbox lifecycle paths then await slow runtime-api HTTP calls before the next commit. The worst holds were runtime `/start` (observed blocking ~90s waiting for a warm runtime), the delete finalizer (archive + stop under one transaction), and the readiness poll loop, which pinned a connection across every sleep for the whole wait (up to 120s). Under a burst of concurrent sandbox starts, every held connection ratchets the pools toward their max, and `QueuePool` keeps the grown pool forever.

## Summary

- Commit (releasing the pooled connection) before every runtime-api call in `start_sandbox`, `pause_sandbox`, `resume_sandbox`, `delete_sandbox`, and the workspace-archive path. Writes land in short transactions of their own; the sandbox row is now only persisted after `/start` succeeds.
- Release the pooled connection between `wait_for_sandbox_running` status polls via a `_release_status_connection` hook (no-op by default, commit in the remote service).
- Restore the key hash when a transient pause failure leaves the sandbox running, so a running sandbox's callbacks don't lose auth.

To test this, I deployed this build to a Replicated test instance. Prior to this change, I saw after starting a few conversations, my connection count against the OpenHands Postgres database would jump to approximately ~50 connections.

After the change, connections stay pinned under 20.

This is the connection count across all workloads that connect to the OpenHands database, I didn't single out the connections originating from the server specifically. 

## Issue Number

N/A

## How to Test

```bash
uv run pytest tests/unit/app_server/test_remote_sandbox_service.py tests/unit/app_server/test_db_session_injector.py
```

144 pass at this branch tip. The full `tests/unit/app_server` suite shows the same 4 environment-dependent failures as current main (verified pre-existing on a pristine checkout).

## Video/Screenshots

N/A (backend-only change).

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- This complements #15225 (LIFO pooling). LIFO helps drain idle connections after a burst, and only once a server-side `idle_session_timeout` is set. This PR shortens the holds that cause the burst exhaustion in the first place.
- Key-hash invalidations for pause/delete are now committed up front. A caller rollback can no longer resurrect a revoked key, which strengthens the property `delete_sandbox` already documented.
- Mid-operation commits also commit any state the caller left pending on the shared session. Both current call paths commit their own writes first, and the poll-loop hook documents that assumption.
- Smaller holds remain in the conversation-start workflow (`clone_or_init_git_repo`, `_seed_sandbox_profiles`) - follow-up material.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-ceea8f0   docker.openhands.dev/openhands/openhands:ceea8f0
```

### DIFF
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -354,6 +354,14 @@ class RemoteSandboxService(SandboxService):
 
         return self._to_sandbox_info(stored_sandbox, runtime)
 
+    async def _release_status_connection(self) -> None:
+        # End the read transaction opened by get_sandbox so the pooled DB
+        # connection returns to the pool while wait_for_sandbox_running sleeps.
+        # The poll loop only reads and start_sandbox has already committed the
+        # sandbox row by the time we wait, so nothing is pending: this commit
+        # just ends the transaction rather than persisting new state.
+        await self.db_session.commit()
+
     async def get_sandbox_by_session_api_key(
         self, session_api_key: str
     ) -> SandboxInfo | None:
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -447,14 +447,14 @@ class RemoteSandboxService(SandboxService):
             # get user id
             user_id = await self.user_context.get_user_id()
 
-            # Store the sandbox
+            # Built now, but only added + committed after the runtime /start
+            # succeeds, so a failed start leaves no row behind.
             stored_sandbox = StoredRemoteSandbox(
                 id=sandbox_id,
                 created_by_user_id=user_id,
                 sandbox_spec_id=sandbox_spec.id,
                 created_at=utc_now(),
             )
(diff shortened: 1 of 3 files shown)


jlav:
Closing for now, https://github.com/OpenHands/OpenHands/pull/15225 in isolation appears to fix the issue I was trying to correct.

PR #15232, opened 2026-07-10
feat(enterprise): make BYOR key alias pattern configurable

tofarr:
HUMAN:

Allows local testing / feature branch testing of LLM keys (with an appropriate environment variable) when the instance of litellm is shared between deployments.

AGENT:

---

## Why

The BYOR key alias sent to LiteLLM was hardcoded as
`BYOR Key - user {user_id}, org {org_id}`. When several OpenHands
deployments share a single LiteLLM instance (typical in local dev,
where a developer might point a feature-env backend at a shared
litellm proxy), their per-user keys collide on the alias.

## Summary

- Extracts the alias pattern into a new env-overridable constant
  `BYOR_KEY_ALIAS_PATTERN` in `enterprise/server/constants.py` and a
  private helper `_create_byor_key_alias()` in
  `enterprise/server/routes/api_keys.py`. The default value matches
  the previous literal, so production is unaffected unless
  `BYOR_KEY_ALIAS_PATTERN` is set.
- Both call sites (`generate_byor_key`, `delete_byor_key_from_litellm`)
  now route through the helper.
- Adds direct unit coverage for the previously untested
  `generate_byor_key` and the new helper, plus an env-override case for
  both generate and delete. The existing
  `test_delete_constructs_alias_from_org` assertion is re-anchored to
  the constant so it stays in sync if the default ever changes.

## Issue Number

None.

## How to Test

1. `cd enterprise && poetry install --with dev,test`
   (or skip if the env is already set up).
2. `cd enterprise && PYTHONPATH=.:$PYTHONPATH python -m pytest tests/unit/server/routes/test_api_keys.py`
   — should show `41 passed`.
3. `cd /projects/OpenHands && pre-commit run --config enterprise/dev_config/python/.pre-commit-config.yaml --files enterprise/server/constants.py enterprise/server/routes/api_keys.py enterprise/tests/unit/server/routes/test_api_keys.py`
   — should pass ruff + mypy.
4. Optional manual check: set
   `BYOR_KEY_ALIAS_PATTERN='dev/{org_id}/{user_id}'` in the env, hit
   `POST /api/keys/llm/byor`, and confirm the new alias appears in
   LiteLLM's key list.

## Video/Screenshots

N/A — backend-only refactor with no UI changes.

## Type

- [x] Refactor
- [x] Feature (configurable pattern is opt-in)

## Notes

- Default value preserves existing behavior; no migrations or data
  backfills needed.
- Follow-up (out of scope): if multiple environments legitimately need
  to share a LiteLLM instance long-term, we may want to also prefix
  the alias with the deployment name (e.g. derived from
  `WEB_HOST`) instead of relying on operators to set the env var.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-4067b7f   docker.openhands.dev/openhands/openhands:4067b7f
```

### DIFF
--- a/enterprise/server/constants.py
+++ b/enterprise/server/constants.py
@@ -18,6 +18,12 @@ IS_FEATURE_ENV = (
 )  # Does not include the staging deployment
 IS_LOCAL_ENV = bool(HOST == 'localhost')
 
+# Pattern for byor keys. Making this an environment variable allows different patterns so multiple environments can use
+# the same litellm instance in development
+BYOR_KEY_ALIAS_PATTERN = os.getenv(
+    'BYOR_KEY_ALIAS_PATTERN', 'BYOR Key - user {user_id}, org {org_id}'
+)
+
 
 # Explicit OH_DEPLOYMENT_MODE wins; _is_all_hands_managed_domain() is the host fallback.
 def _is_all_hands_managed_domain(host: str) -> bool:
--- a/enterprise/server/routes/api_keys.py
+++ b/enterprise/server/routes/api_keys.py
@@ -7,6 +7,7 @@ from pydantic import BaseModel, SecretStr, field_validator, model_validator
 from server.auth.authorization import get_user_super_role
 from server.auth.org_context import EFFECTIVE_ORG_ID
 from server.auth.saas_user_auth import SaasUserAuth
+from server.constants import BYOR_KEY_ALIAS_PATTERN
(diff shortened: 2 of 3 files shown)


jlav (approved):

hieptl (approved):
Thank you! 🙏

PR #15234, opened 2026-07-10
fix(enterprise): stop overwriting every OrgMember key with the admin's on org-defaults save

tofarr:
## Why

Customers have reported that their LLM API key sometimes becomes invalid in
LiteLLM (e.g. after a proxy restart or manual cleanup), and they keep hitting
401s until someone re-saves their settings. The existing self-healing pattern
in `LiteLlmManager.migrate_entries` already handles this for users on the V3
migration path. The bug here is in `OrgStore._update_org_kwargs`: after
`_maybe_get_managed_llm_key_for_user` validates (and possibly rotates) the
acting user's key, the code does
`member_updates.llm_api_key = SecretStr(effective_managed_key)` and runs that
through `update_all_members_settings_async`, which writes the **same** key
value onto every `OrgMember.llm_api_key` row.

Two consequences:

1. Every member row in the org ends up with the acting user's key. That's
   one LiteLLM identity shared across the whole org, which breaks the
   per-(user, org) aliasing that `get_openhands_cloud_key_alias(user_id,
   org_id)` is supposed to provide.
2. A member whose key was *already valid* gets silently overwritten with
   the acting user's, which can re-introduce stale-key symptoms for that
   member if the acting user's rotation happened in parallel.

## Summary

- In `_update_org_kwargs`, after `_maybe_get_managed_llm_key_for_user`
  returns, stamp `effective_managed_key` onto the **acting member's**
  row only — the same row the helper already loaded.
- Drop the bulk propagation of `effective_managed_key` to every member
  via `update_all_members_settings_async`. The bulk path still handles
  `has_custom_llm_api_key = False` and any non-LLM-key diffs.
- Extract the small `_get_member_for_user` helper from
  `_maybe_get_managed_llm_key_for_user` so both call sites use one
  lookup.
- Update the existing test
  `test_update_org_defaults_async_propagates_managed_key_reset` →
  `test_update_org_defaults_async_stamps_acting_user_key_and_resets_flag`
  to assert the new contract.

## Why only the acting user (not all members)?

- **Scale.** Iterating every member inside the request handler is O(N)
  LiteLLM `verify_existing_key` calls per org-defaults save. At 100K
  members that is a non-starter; even at 1K it noticeably degrades the
  org-admin request. The acting user's rotation is exactly one call.
- **BYOR safety.** Members with `has_custom_llm_api_key = True` carry a
  *different* key (their BYOR key); rotating a managed key onto them
  would silently flip them off BYOR.
- **Lazy self-heal already exists.** `SaasSettingsStore._ensure_api_key`
  (called on every user-settings save at
  `enterprise/storage/saas_settings_store.py:582` and
  `:777-811`) already does `verify_existing_key` → `delete_key_by_alias`
  → `generate_key` for the user whose settings are being saved. So the
  customer's "my key went stale" complaint is self-healed on their next
  settings save, regardless of whether anyone touched the org defaults.

If we ever want *proactive* bulk self-heal across all members (e.g. a
nightly sweep), that should be a background task — the codebase already
uses `fastapi.BackgroundTasks` in `enterprise/server/routes/auth.py:476`
and the integration routes, and `OrgStore` could expose a method that
the task calls. That is intentionally **out of scope** for this PR.

## Issue Number

None — opened off a customer-reported incident.

## How to Test

```bash
PYTHONPATH=enterprise:. pytest enterprise/tests/unit/test_org_store.py \
  -p no:ddtrace -p no:ddtrace.pytest_bdd -v
```

Manual repro of the original bug:
1. Pick a managed-LLM org (model = `openhands/*` or base_url =
   `LITE_LLM_API_URL`) with at least two members.
2. Save org defaults so `_maybe_get_managed_llm_key_for_user` returns a
   key for the org admin.
3. Inspect any *other* `OrgMember.llm_api_key` row — it should still
   hold that member's own pre-existing key, not the admin's. (Before
   the fix, every row would have been rewritten to the admin's key.)
4. The admin's own row should hold the freshly minted/verified key.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Net diff vs. `main`: +42 / -292 (the original self-heal-loop draft
  was +303; revised per feedback to this much smaller fix).
- The pre-existing test `test_update_org_defaults_async_propagates_managed_key_reset`
  was renamed rather than deleted, since the asserted contract changed.
- No new dependencies; no migrations.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-ed294a5   docker.openhands.dev/openhands/openhands:ed294a5
```


tofarr:
This pull request was created by an AI agent (OpenHands) on behalf of a user. The agent implemented the self-healing logic and tests but has not been human-tested end-to-end against a live LiteLLM instance. Please run `pytest enterprise/tests/unit/test_org_store.py -p no:ddtrace` and exercise the manual repro in the PR description before marking ready for review.

saurya:
I was thinking about this PR a bit more, and I think cascading org admin keys is actually expected behavior right now. When an org admin sets a profile, what we're trying to say is that none of the users should be able to override that profile with their own keys. So we don't want BYOR for users who are within an org.

neubig:

Thank you for understanding.

PR #15236, opened 2026-07-10
fix(frontend): add jittered rate-limit backoff

aivong-openhands:
HUMAN: Improves the rate limiting helper by preventing the thundering herd problem by retrying on a exponential backoff strategy with a random jitter. Tested in a feature environment. Had a basic successful conversation with openhands.

AGENT:

---

## Why

Clients that receive HTTP 429s can retry on the same schedule. Retrying together creates synchronized bursts of traffic that can worsen rate limiting. Exponential backoff with jitter is a common solution to this thundering-herd problem: it spreads retries out while reducing unnecessary load. This follow-up to #14660 applies that approach.

## Summary

- Add a reusable frontend rate-limit retry-delay utility with exponential backoff and bounded jitter.
- Treat `Retry-After` as a server-directed minimum delay, then add jitter so clients do not retry simultaneously.
- Apply the utility to global query/mutation retries and the resume-related conversation fetch, with focused regression coverage.

## Issue Number

Related to #14656. Follow-up to #14660.

## How to Test

```bash
cd frontend
npm test -- __tests__/utils/rate-limit-retry.test.ts __tests__/hooks/query/use-user-conversation.test.tsx __tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx __tests__/query-client-config.test.ts
npm run lint:fix
npm run typecheck
npm run build
```

## Video/Screenshots

Not included. This is retry behavior covered by focused unit tests.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

`Retry-After` remains a lower bound; the added jitter can only delay a retry further. `npm run lint:fix` passed with existing warnings outside this change.

Recreated from #15221 on an upstream branch so trusted Docker image jobs can run.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-e963739   docker.openhands.dev/openhands/openhands:e963739
```


frontend/src/utils/rate-limit-retry.ts
```diff
@@ -36,26 +37,44 @@ export function isRateLimitError(error: unknown): boolean {
...
+): number {
+  const exponentialDelay = Math.min(
+    INITIAL_RETRY_DELAY_MS * 2 ** failureCount,
+    MAX_BASE_RETRY_DELAY_MS,
+  );
+  const retryAfterDelay = getRetryAfterDelayMs(error);
+  // Retry-After is a server-directed minimum; jitter only delays further.
+  const minimumDelay = Math.max(retryAfterDelay ?? 0, exponentialDelay);
+  const jitter = Math.floor(
+    Math.random() * Math.min(exponentialDelay, MAX_JITTER_MS),
```
VascoSch92: Jitter window is sized off `exponentialDelay` but the value you actually return is `minimumDelay`, so when Retry-After is the bigger of the two the spread stays tied to the small term.

Your own test at `__tests__/utils/rate-limit-retry.test.ts:164` pins it:

```ts
vi.spyOn(Math, "random").mockReturnValue(0.5);
expect(getRateLimitRetryDelayMs(0, createAxiosError(429, { "retry-after": "5" })))
  .toBe(5500);   // jitter = 500 = 0.5 * 1000 (exponential), not 0.5 * 5000
```

500ms of jitter on a 5s wait. At `Retry-After: 60` it's ~999ms of spread on a 60s wait, so anything that got 429'd together comes back more or less together.

```suggestion
    Math.random() * Math.min(minimumDelay, MAX_JITTER_MS),
```

That takes the spread at 60s from ~1s to ~5s. It does break the `:164` expectation (5500 -> 7500) so that'd need updating alongside.

Not urgent. The only thing currently emitting a Retry-After big enough to hit this is the enterprise `100/minute` window, and that's one user colliding with their own limit rather than a fleet-wide pileup. But the formula is wrong as written and it'll bite the first time an endpoint returns a longer Retry-After.

frontend/src/utils/rate-limit-retry.ts
```diff
@@ -1,7 +1,8 @@
 import { AxiosError } from "axios";
 
-const DEFAULT_RETRY_AFTER_MS = 1000;
-const MAX_RETRY_DELAY_MS = 60_000;
+const INITIAL_RETRY_DELAY_MS = 1000;
+const MAX_BASE_RETRY_DELAY_MS = 60_000;
+const MAX_JITTER_MS = 5000;
```
VascoSch92, line 5: `MAX_JITTER_MS` can't actually be reached right now. Highest `failureCount` that produces a delay anyone uses is 2 (`failureCount < 3`), which gives `exponentialDelay = 4000`, under the 5000 cap. The 60s clamp on the exponential term is dead for the same reason.

The test at `:176` passes `failureCount: 10` to exercise it, which no caller can produce. No harm keeping the cap defensively, but a one-line comment saying it only kicks in if the retry limits go up would stop the next person reading it as load-bearing. The suggestion I left on line 76 would make it live as a side effect.

frontend/__tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
```diff
@@ -49,6 +50,10 @@ describe("useUnifiedResumeConversationSandbox", () => {
     vi.restoreAllMocks();
   });
 
+  afterEach(() => {
```
VascoSch92, line 53: Cleanup is split across the two hooks here. `restoreAllMocks()` is up in `beforeEach` (line 50) but `afterEach` only does `useRealTimers()`, so the `Math.random` stub stays installed after this test finishes and only gets cleared when the next one starts.

Doesn't matter today, but a `.only` run or a test appended after this one would inherit a stubbed `Math.random`. `use-user-conversation.test.tsx` already does both in `afterEach`:

```ts
afterEach(() => {
  vi.useRealTimers();
  vi.restoreAllMocks();
});
```

frontend/__tests__/utils/rate-limit-retry.test.ts
```diff
@@ -30,17 +31,56 @@ describe("rate limit retry helpers", () => {
...
+
+    expect(
+      getRateLimitRetryDelayMs(
+        0,
+        createAxiosError(429, { "Retry-After": 2 }),
+      ),
+    ).toBe(2000);
+  });
+
+  it("falls back to backoff when AxiosHeaders returns a non-string value", () => {
```
VascoSch92, line 52: These two assert inconsistent behaviour for the same input:

- `:41` - `{ "Retry-After": 2 }` gives 2000
- `:52` - `{ get: () => 2 }` gives 1000

Same numeric `2`, different result depending on which branch of `getHeaderValue` you land in: the plain-object path coerces with `String()`, the AxiosHeaders path bails unless it's already a string. Came in with #14660 so not yours, but these tests bless it as intended now.

Axios normalises Retry-After to a string in practice so it shouldn't actually bite. If you want them to agree it's a one-liner in the `.get()` branch, otherwise maybe a comment noting the split is deliberate.

VascoSch92 (commented):
Took a proper look at the retry mechanics on this one rather than just reading the diff.

`failureCount` is what you'd want: query-core 5.90.20 calls `retryDelay(failureCount, error)` before the increment (`retryer.js:84`, bump at `:90`), so it starts at 0 and you get 1s/2s/4s. Same counter feeds `retry`, so the `failureCount < 3` guards still line up.

The "Retry-After stays a lower bound" claim in the description checks out too, since `minimumDelay = max(retryAfter ?? 0, exponential)` and jitter is never negative.

Grepped for stragglers and there are only 3 non-test callers, all migrated. The signature change is a compile error anyway so nothing could really have slipped through. Ran the four test files locally, 21/21 green, CI is 21 pass / 6 skipped / 0 fail.

One design thing inline that I don't think should hold up the merge: jitter is scaled off `exponentialDelay` instead of the delay you actually return. It only really shows up on the enterprise `100/minute` window, since OSS sends `Retry-After: 1` and the enterprise per-second window is <= 1s, so exponential wins in both of those and you get the full spread.

Other thing worth a quick check: worst case on the resume preflight goes from 3x1s to 7s base, ~14s with jitter, before the error surfaces. Probably fine for a background refetch, just want to be sure that path shows a spinner and doesn't look frozen.

VascoSch92 (approved):

frontend/src/utils/rate-limit-retry.ts
aivong-openhands: Applied in 68138c9c — jitter now scales off `minimumDelay` (`Math.min(minimumDelay, MAX_JITTER_MS)`), so a large `Retry-After` gets a proportional spread instead of being tied to the exponential term. Updated the `:164` expectation from 5500 to 7500 to match, and left a comment on the line explaining the scaling.

frontend/src/utils/rate-limit-retry.ts
aivong-openhands: The line-76 change makes `MAX_JITTER_MS` live — a `Retry-After` big enough now hits the cap. Added a one-line comment noting the 60s exponential clamp stays defensive: it only becomes load-bearing if the `failureCount < 3` retry limits are raised.

frontend/__tests__/hooks/mutation/use-unified-resume-conversation-sandbox.test.tsx
aivong-openhands: Added `vi.restoreAllMocks()` to the `afterEach` so the `Math.random` stub is cleared when the test finishes rather than leaking to the next one.

frontend/__tests__/utils/rate-limit-retry.test.ts
aivong-openhands: Kept the split (Axios normalizes `Retry-After` to a string in practice, so `.get()` returning a non-string doesn't happen on the real path) and documented the intent in `getHeaderValue`: the `AxiosHeaders.get()` branch trusts strings only, and numeric coercion is the plain-object fallback for hand-built/mocked headers.

aivong-openhands:
Thanks for digging into the mechanics. Pushed 68138c9c addressing the four inline threads (jitter scaling, defensive-cap comment, header-split comment, test-mock cleanup); the four test files still pass 21/21 locally.

On the resume-preflight worst case (up-to-~14s before the error surfaces): that path isn't frozen. `onMutate` optimistically sets `sandbox_status: STARTING` before the request fires and it stays STARTING across the entire retry/backoff window — it's only reverted in `onError` once the retries are exhausted — so the conversation renders its starting/loading state throughout the backoff rather than looking hung.

aivong-openhands:
Thanks @VascoSch92! Can you assist with the merge?

PR #15241, opened 2026-07-10
refactor: Expand better-error-logging patterns across the codebase

tofarr:
HUMAN:

Better logging. Tedious but useful.

AGENT:

---

## Why

The `better-error-logging` branch (current tip `fdc97ff9d`) established two useful patterns in `enterprise/server/routes/orgs.py` and `enterprise/storage/org_service.py`:

1. In an `except` handler, `logger.error(...)` should become `logger.exception(..., stack_info=True)` so the traceback is preserved, and the redundant `'error': str(e)` should be dropped from the `extra` dict (the traceback already carries it).
2. When re-raising a different exception type after catching one, `raise NewException(f'... {str(e)}')` should become `raise NewException('...') from e` so the cause chain is preserved, and the inlined `str(e)` is no longer needed.

This PR expands those two patterns across the rest of the codebase where the same shape occurs. It does not change behavior for `logger.warning(...)` inside `except` (those are intentional 4xx-style client-error warnings) or for the bare `raise e` re-raise pattern.

## Summary

- Replaced `logger.error(...)` calls inside `except` blocks with `logger.exception(..., stack_info=True)` (or added `stack_info=True` to existing `logger.exception(...)` calls) across `enterprise/{server/{routes,services,auth,app_lifespan},storage,integrations,sync,maintenance_task_processor,verified_models,middleware,rate_limit}` and `openhands/{server,app_server}` — 95 source files, 447 transformations.
- Dropped the now-redundant `'error': str(e)` from `extra={...}` dicts (and any `extra={}` that became empty as a result) in those handlers.
- Added `from e` to `raise NewException(...)` calls inside `except` handlers so the cause chain survives, including for `raise HTTPException(...)`. Where the new exception is not `HTTPException`, simplified the message by stripping `{e}` / `{str(e)}` interpolations since the chain now carries that context.
- Removed `as e` bindings from `except` clauses whose handlers no longer reference the bound exception.
- Added `from original_error` to `OrgService._handle_failure_with_cleanup` so the original exception is preserved as `__cause__` instead of being only embedded in the message string.
- Updated 4 test files where assertions needed to follow the new shape:
  - `str(exc_info.value)` for underlying-error text → `str(exc_info.value.__cause__)` (the underlying exception is now chained via `from e`).
  - `mock_logger.error` → `mock_logger.exception` where the production code now logs at exception level.

## Issue Number

None.

## How to Test

- `pre-commit run --files $(git diff --name-only main..HEAD --diff-filter=AM)` — passes locally (ruff lint, ruff format, mypy).
- Unit tests for the touched modules — all 344 pass locally:
  - `enterprise/tests/unit/test_org_service.py` (55)
  - `enterprise/tests/unit/test_token_manager.py` + `test_token_manager_extended.py` (49)
  - `enterprise/tests/unit/test_org_invitations_router.py` (16)
  - `enterprise/tests/unit/server/routes/test_orgs.py` (121)
  - `enterprise/tests/unit/server/routes/test_api_keys.py` (41)
  - `enterprise/tests/unit/server/routes/test_user_app_settings.py` (6)
  - `enterprise/tests/unit/server/services/test_automation_event_service.py` (40)
  - `enterprise/tests/unit/integrations/github/test_github_manager.py` (16)
- Pre-existing failures in the broader suite (missing `SLACK_SIGNING_SECRET`, missing `google.cloud.recaptchaenterprise_v1`, `test_users_v1::test_override_removes_oss_route_and_adds_saas_route`) fail identically on `main` before any of these changes — they are unrelated.

To re-verify locally:

```bash
PYTHONPATH=".:enterprise:$PYTHONPATH" OPENHANDS_SUPPRESS_BANNER=1 \
  poetry run pytest \
    enterprise/tests/unit/test_org_service.py \
    enterprise/tests/unit/test_token_manager.py enterprise/tests/unit/test_token_manager_extended.py \
    enterprise/tests/unit/test_org_invitations_router.py \
    enterprise/tests/unit/server/routes/test_orgs.py \
    enterprise/tests/unit/server/routes/test_api_keys.py \
    enterprise/tests/unit/server/routes/test_user_app_settings.py \
    enterprise/tests/unit/server/services/test_automation_event_service.py \
    enterprise/tests/unit/integrations/github/test_github_manager.py
```

## Video/Screenshots

N/A — refactor only.

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The PR is purely a logging/exception-handling refactor. No API contracts change; response bodies for HTTP errors are identical (the `detail` strings were not modified).
- Two intentional divergences from the literal original-PR shape, both documented and consistent with its spirit:
  1. `from e` is also added to `raise HTTPException(...)` calls (the original PR only added it for non-HTTP domain exceptions). The user-facing response is unchanged; the traceback chain is now explicit (`__cause__`) instead of implicit (`__context__`).
  2. For non-HTTP re-raises, the `{e}` / `{str(e)}` interpolation is removed from the message since `from e` preserves the context. A few messages previously ending with `: {e}` now end without the trailing colon (e.g. `f"Error: Failed to write to bucket '{name}' at path {path}"`). If reviewers prefer the original message shape, that's a one-line revert.
- This PR was generated by an AI agent on behalf of the user.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-f8038eb   docker.openhands.dev/openhands/openhands:f8038eb
```


hieptl (approved):
Thank you! 🙏

simonrosenberg (approved):

PR #15245, opened 2026-07-11
fix: surface and log failures sending verification emails

akt715:
Human:
  Change makes the failure visible during sending a verification email , before this fix , it was showing 200 despite of an error , which was confusing the user as no verification email was being sent to them 

## What

`verify_email()` in `enterprise/server/routes/email.py` — the shared function used by both the GitHub/OAuth login callback and the manual `PUT /api/email/resend` endpoint — called `keycloak_admin.a_send_verify_email()` with no error handling. When that call fails (bad SMTP config, a Keycloak API error, a transient network issue), the failure either surfaced as an opaque, context-free 500, or was effectively invisible depending on the admin client's behavior on non-2xx responses.

This matches the reported bug: users completing GitHub OAuth are redirected to an email-verification screen, click "Resend verification" repeatedly, see no error, and never receive an email.

## Why

Code inspection confirmed the email-verification gate itself is deliberate and already covered by tests — this PR does not change that gating behavior. The actual gap is observability: there was no structured signal (user id, failure reason) to correlate a "never got the email" report with an actual send failure in the Keycloak SMTP/ESP pipeline, and the resend endpoint always reported success even when the underlying send failed.

## How

- Wrap the `a_send_verify_email` call in `verify_email()` with `try/except`.
- On failure, log a structured error (`user_id`, `is_auth_flow`, exception detail) so it can be correlated with Keycloak SMTP/ESP logs.
- Raise an explicit `HTTPException(502)` instead of letting the failure propagate as an unhandled 500 or get silently absorbed.

This is a narrowly-scoped, single-function fix (the shared `verify_email()`), so both the OAuth callback path and the manual resend endpoint benefit automatically.

## Known limitation

This change does **not** guarantee email delivery — that depends on Keycloak SMTP/ESP configuration and the recipient domain's mail acceptance (e.g. iCloud private-relay addresses, GitHub noreply-style addresses). It closes the observability gap needed to diagnose *why* delivery is failing and stops the endpoint from lying about success, but the underlying infra/deliverability investigation is a separate follow-up.

Also note: `update_email()` (a third caller of `verify_email()`) has a pre-existing broad `except Exception` that will still re-wrap this new 502 as a generic 500 for that call path — not a regression introduced here, but a known gap for a possible follow-up.

## Testing
Added two unit tests in `enterprise/tests/unit/server/routes/test_email_routes.py`:
- `test_verify_email_raises_and_logs_on_send_failure`: calls `verify_email()` directly with `a_send_verify_email` raising, asserts a 502 `HTTPException` is raised and `logger.error` is called with the user_id in the log context.
- `test_resend_email_verification_returns_502_when_send_fails`: calls the `PUT /api/email/resend` handler end-to-end (through `resend_email_verification`) with the same failure, asserting it now returns/raises a 502 instead of the previous unconditional 200 success response.

All existing success-path tests for `verify_email`, `resend_email_verification`, `update_email`, and `verified_email` are unchanged and continue to pass, confirming no behavior change on the happy path. Validation (format/tests/lint/static analysis) passed for this diff.

Manual verification recommended in a staging environment with a deliberately broken Keycloak SMTP config: trigger a GitHub OAuth login and confirm the resend button now surfaces an error toast (via the frontend's existing `onError` handling) instead of a false success message, and that the failure appears in logs with the affected user_id.

## Rollout
Low-risk, additive change scoped to one function (`verify_email`) shared by the OAuth callback and resend endpoint. No schema/migration changes, no new dependencies, no config flags. Safe to deploy with a normal rollout; no special sequencing needed. Recommend watching the new `logger.error` output for `Failed to send verification email for user...` after deploy to gauge how often the Keycloak send path is actually failing in production, which will inform the deliverability follow-up (SMTP/DKIM/SPF, ESP suppression lists for iCloud/GitHub noreply-style addresses).

## Rollback
Revert the single commit touching `enterprise/server/routes/email.py` and its test file. No data migrations or external state changes are introduced, so a straight revert is safe and fully restores prior behavior (silent/unhandled failure on send errors).

## Release notes
Fixed: the "resend verification email" action now shows a clear error instead of silently reporting success when the email can't actually be sent.

---
*Opened by CodeForge after human approval of plan and diff.*


neubig:

Thank you for understanding.

PR #15252, opened 2026-07-12
fix: accept numeric environment toggle values

ShiHuiwen-creat:
HUMAN:

This PR fixes boolean environment variable handling so that both `true` and `1` correctly enable frontend serving and secure S3 connections. It keeps the change focused on the three reported call sites and adds regression tests for the supported true and false values.

## Summary

- accept the documented numeric `1` value for `SERVE_FRONTEND`
- preserve HTTPS when `AWS_S3_SECURE=1` in both file-store and event-service paths
- add focused regression coverage for defaults, `true`/`1`, and `false`/`0`

## Verification

- `58 passed` across the three affected test modules
- staged-file pre-commit checks passed (whitespace, EOF, debug statements, AppMode check, Ruff lint and format)
- independent review completed

Fixes OpenHands/enterprise#57

### DIFF
--- a/openhands/app_server/app.py
+++ b/openhands/app_server/app.py
@@ -71,8 +71,13 @@ async def authentication_error_handler(request: Request, exc: AuthenticationErro
 app.include_router(v1_router.router)
 app.include_router(health_router)
 
+
+def _should_serve_frontend() -> bool:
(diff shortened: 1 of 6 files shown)


ShiHuiwen-creat:
Confirmed, this PR is ready for review.

neubig:

Thank you for understanding.

PR #15258, opened 2026-07-13
fix(auth): trust external identity providers for email verification (…

Tanmay-Awal:
…#15143)

Root Cause:
- During login via Keycloak callback, the backend checks the OIDC 'email_verified' claim. For external/social identity providers (like GitHub, Apple, Google), Keycloak does not automatically set this claim to True.
- This triggered local email verification for OAuth users. However, Keycloak is not configured to send verification emails for these external users, causing the verification email to never be sent or delivered and blocking logins.

Expected Behavior:
- When a user logs in via GitHub OAuth or other social identity providers, they should be successfully authenticated and granted access without being redirected to the local email verification page, as the identity provider is trusted to have verified the user.

Actual Behavior:
- After successfully authenticating via GitHub OAuth or Apple ID, the application redirected the user to the email verification prompt, and no verification email was ever received.

Changes Implemented:
1. OAuth Callback Bypassing: Updated keycloak_callback in enterprise/server/routes/auth.py to bypass the email verification redirect if an external identity provider is detected (user_info.identity_provider is set).
2. Database Syncing: Updated UserStore.create_user, UserStore.backfill_user_email in enterprise/storage/user_store.py and keycloak_callback in enterprise/server/routes/auth.py to record email_verified = True in the user's database record upon successful social OAuth login.
3. Active Session Propagation: Modified SaasUserAuth.get_user_email in enterprise/server/auth/saas_user_auth.py to query the database when the session's active token shows the email as unverified, propagating the verified state from the DB to the settings API.
4. Unit Tests: Modified mock fixtures in test_auth_routes.py to set identity_provider=None to preserve verification flow test cases for local credentials, and added test_keycloak_callback_email_not_verified_but_social_login to assert that social OAuth logins bypass local email verification.

Closes OpenHands/enterprise#54

HUMAN:

AGENT:

---

## Summary

-

## Type

- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore


neubig:

Thank you for understanding.

PR #15269, opened 2026-07-14
fix: redact Tavily MCP API key in init errors

jw-ond:
HUMAN:
This PR prevents Tavily MCP proxy initialization failures from writing the raw `tavilyApiKey` query parameter into logs. The error is still logged for debugging, but the secret value is redacted first.

AGENT:

## Why
The Tavily MCP proxy builds its upstream URL with `tavilyApiKey` as a query parameter. If initialization fails and the raised exception includes that URL, the current error log can persist the API key.

## Summary
- Add a small Tavily MCP URL redactor for loggable error strings.
- Use it when logging Tavily MCP proxy initialization failures.
- Add a regression test that confirms the raw API key is not logged while the failure remains visible.

## Issue Number
N/A

## How to Test
- `git diff --check`
- `.venv/bin/python -m py_compile openhands/app_server/mcp/mcp_router.py tests/unit/server/routes/test_mcp_routes.py`

I also attempted to run `pytest tests/unit/server/routes/test_mcp_routes.py -q` locally, but the sparse local environment did not have the full OpenHands dependency set available; installing the split `openhands-*` packages hit a pip dependency-resolution conflict. The changed files compile successfully, and the added test follows the existing mocked Tavily proxy test style.

### DIFF
--- a/openhands/app_server/mcp/mcp_router.py
+++ b/openhands/app_server/mcp/mcp_router.py
@@ -46,6 +46,11 @@ HOST = f'https://{os.getenv("WEB_HOST", "app.all-hands.dev").strip()}'
 CONVERSATION_URL = HOST + '/conversations/{}'
 
 
+def _redact_tavily_api_key(value: str) -> str:
+    """Redact the Tavily MCP API key from loggable strings."""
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #15273, opened 2026-07-14
fix(settings): prevent autocapitalize from mangling marketplace source

jpshackelford:
## Problem

In the skills settings marketplace modal (`marketplace-modal.tsx`), the **Repository (source)** field accepts case-sensitive values like `github:owner/repo`. The input is a bare `<input type="text">` with no autocapitalize/autocorrect controls, so on mobile the keyboard capitalizes the first letter and the user ends up submitting `Github:owner/repo`.

That capitalized source then fails to parse (the scheme match is case-sensitive) and surfaces a confusing error instead of installing the marketplace — the reported "UI breaks" symptom.

## Fix

Disable `autoCapitalize` / `autoCorrect` / `autoComplete` / `spellCheck` on the three case-sensitive inputs in the modal:
- **source** (`github:owner/repo`)
- **ref** (branch/tag/commit — e.g. `Release-2.0`)
- **repo path** (subdirectory — e.g. `MyDir/marketplace`)

The Name field is intentionally left alone (not case-sensitive in the same way).

Only the input behavior is changed; no value is force-lowercased, since refs and subpaths are legitimately case-sensitive.

## Tests

Added a regression test to `marketplace-modal.test.tsx` asserting the three inputs carry `autocapitalize="none"`, `autocorrect="off"`, and `spellcheck="false"`. Local run: **17 passed** (16 existing + 1 new); `prettier --check` clean on both files.

## Related

This is the frontend/trigger half of a two-part fix. The backend root cause — the `github:` scheme being matched case-sensitively in `parse_extension_source()` — is fixed in the SDK so that *any* entry path (paste, deep-links, marketplace catalog entries, API) is protected even without this input hardening:

**OpenHands/software-agent-sdk#4117**

---

@jpshackelford can click here to [continue refining the PR](https://app.all-hands.dev/conversations/9d0d1069-8482-40c2-b74a-d681568eaab1)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-f4fb766   docker.openhands.dev/openhands/openhands:f4fb766
```

### DIFF
--- a/frontend/__tests__/components/features/settings/marketplace-modal.test.tsx
+++ b/frontend/__tests__/components/features/settings/marketplace-modal.test.tsx
@@ -40,6 +40,25 @@ describe("MarketplaceModal", () => {
     ).toBeInTheDocument();
   });
 
+  it("disables autocapitalize/autocorrect on case-sensitive source, ref and repo path inputs", () => {
+    render(<MarketplaceModal {...baseProps} />);
+
+    // A "github:owner/repo" source, a ref, and a repo subpath are all
+    // case-sensitive; mobile keyboard autocapitalization must not mangle them
+    // (e.g. turning "github:" into "Github:", which breaks parsing).
+    const caseSensitiveInputs = [
+      screen.getByPlaceholderText("github:owner/repo"),
+      screen.getByPlaceholderText("e.g., main, develop, v1.0.0"),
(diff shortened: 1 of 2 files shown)


neubig:

Thank you for understanding.

PR #15279, opened 2026-07-15
fix(acp): broker concurrent Codex subscription auth

simonrosenberg:
HUMAN:

AGENT:

---

## Why

Codex refreshes ChatGPT subscription credentials inside its writable `auth.json`. Cloud previously supplied a static snapshot, so later sessions could receive an invalidated token. Allowing multiple sessions without coordination would also let two Codex processes present the same rotating refresh token upstream.

## Summary

- Replace stored ChatGPT Codex auth with a conversation- and sandbox-scoped `LookupSecret` when launching SaaS Codex conversations.
- Add authenticated internal GET/HEAD/PUT/DELETE operations with encrypted storage, authoritative digests, and row-locked CAS updates.
- Add a scoped refresh endpoint that serializes refreshes per user and organization with a Redis lock. The first request rotates upstream and persists the result; an overlapping stale request receives the new authoritative tokens without reusing the provider token.
- Defer global cookie authentication only for the exact internal POST refresh path, whose session key and scoped JWS are validated by the route itself.

## Issue Number

Implements the Cloud side of OpenHands/software-agent-sdk#4121, specifically https://github.com/OpenHands/software-agent-sdk/issues/4121#issuecomment-4979373306, extended to support concurrent sessions safely.

Companion SDK PR: https://github.com/OpenHands/software-agent-sdk/pull/4124

Global request-body limiting is tracked in OpenHands/enterprise#62. This PR keeps the 64 KiB Codex credential-value validation but does not add a route-specific transport limit.

## How to Test

Concurrent exact-image Docker E2E evidence: https://github.com/OpenHands/OpenHands/pull/15279#issuecomment-4980784130

```bash
poetry run --project=enterprise pytest -q \
  enterprise/tests/unit/test_auth_middleware.py \
  enterprise/tests/unit/server/routes/test_codex_auth.py
# 33 passed

poetry run --project=enterprise pre-commit run \
  --config enterprise/dev_config/python/.pre-commit-config.yaml \
  --files enterprise/server/middleware.py \
  enterprise/server/routes/codex_auth.py \
  enterprise/tests/unit/test_auth_middleware.py \
  enterprise/tests/unit/server/routes/test_codex_auth.py
# all hooks passed, including package-wide mypy
```

The E2E migrated an empty PostgreSQL database, ran the Cloud and runtime Docker images with Redis and a runtime-api/OAuth simulator, and overlapped two exact Codex 0.144.4 sessions. Both exited 0 and converged on the same encrypted Cloud state; two broker calls produced one upstream rotation, zero reuse attempts, and zero invalid runtime-api keys.

## Video/Screenshots

Not applicable; this is backend credential storage and refresh coordination.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Deploy with a runtime/SDK containing https://github.com/OpenHands/software-agent-sdk/pull/4124. The scoped broker is what makes concurrent ChatGPT-subscription Codex sessions safe; older runtimes do not configure the refresh override or reconciliation behavior.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-8f6ab19   docker.openhands.dev/openhands/openhands:8f6ab19
```


simonrosenberg:
## Coordinated Docker E2E validation

Validated the exact PR heads together:

- SDK: `3f38326a20cc5eb9d9cfdde260bcceae5df610b1`
- Cloud: `335065a87ced0834b5373c8e8171a0e20d0bd3b0`

### Deployment

- Built the enterprise Cloud app image: `sha256:4e165076efdaee15d4f9c50d34e35fa8a96a08029d10b3c7c754427b0e2d19a9` (arm64).
- Built the SDK `source-minimal` runtime image: `sha256:243c4339816201895185768d2a415bfe25f80fbc4981df4d8b79878fbd8137c9` (arm64).
- Ran the Cloud app with PostgreSQL 16, Redis 7, and a runtime-api HTTP simulator on an isolated Docker network. The simulator implements the production `/list` and `/sessions/{sandbox_id}` contract and verified the configured runtime API key on every request.
- Applied enterprise migrations through revision `137`; the externally exposed Cloud health request returned HTTP 200.

No Kubernetes resources were changed.

### Real E2E calls

The exact-head SDK image used `ACPAgent`, `LookupSecret`, `ConversationState`, and `LocalWorkspace` against the running Cloud endpoint. Result:

```json
{"sdk_version":"1.36.1","stale_working_copy_overwritten":true,"r0_to_r1_next_conversation":true,"unchanged_sync_head":true,"historical_and_current_tokens_masked":true,"active_owner_conflict_status":409,"stale_digest_conflict_status":409,"invalid_document_status":400,"r2_visible_to_next_conversation":true}
```

The Cloud access log recorded the complete GET/PUT/HEAD/DELETE lifecycle: successful R0 retrieval, R0→R1 write, unchanged HEAD, release, R1 retrieval by the next conversation, active-owner `409`, R1→R2 write, stale-digest `409`, invalid-document `400`, final release, and R2 retrieval by the next conversation.

### Persistence and isolation assertions

- The final encrypted `CODEX_AUTH_JSON` row was present, and plaintext `refresh-r2` was absent from the database value.
- Both sandbox session-key hashes were present.
- Redis DB size was `0` after the final release, confirming no ownership lease remained.
- The invalid credential sentinel was absent from the HTTP error response.

The disposable deployment was also run once against the pre-merge SDK implementation commit before the exact-head rerun; both runs passed.

simonrosenberg:
## Superseding concurrent-session E2E validation

Validated the pushed SDK/Cloud sources in disposable Docker infrastructure with an empty PostgreSQL database migrated through enterprise Alembic head, Redis, a runtime-api simulator, and an OAuth simulator that rotates refresh tokens and rejects reuse.

Exact local images:

- Cloud: `local/openhands-codex-auth-cloud:concurrent-e2e-final` (`sha256:f51d2bc9d6e37a453f34ea00e244d319e16c2bd2dad8cd58392252d28dda4f99`)
- Runtime: `local/openhands-codex-auth-runtime:concurrent-e2e-source-minimal` (`sha256:1088cf4d88a9fc32048875fd357a8e35c54cbcb8fa85d311be06fcb8e9ca123b`)
- Host/image source hashes matched for `acp_agent.py`, `codex_auth.py`, and `middleware.py`.
- The runtime contained the exact bundled Codex 0.144.4 binary used by `codex-acp`.

Two isolated runtime containers materialized the same Cloud credential and crossed a barrier before simultaneously calling real Codex `account/read` with forced refresh. Both exited 0 with:

- `account_read_succeeded=true`
- refresh broker URL configured and transport credentials masked
- refresh token rotated
- local `auth.json` equal to authoritative Cloud state
- identical local/Cloud refresh-token digest: `bd05ff50f25c407e652c7d3f84613cd083f52c593328fc3ffdf44706728ba400`

Concurrency and deployment assertions:

- 2 overlapping broker refresh requests produced exactly 1 upstream OAuth request and 1 rotation.
- 0 stale-token reuse attempts reached OAuth.
- 12 runtime-api requests used the configured API key; 0 were rejected.
- Both expected stale-digest write-backs were reconciled from Cloud without failing either session.
- The Redis refresh lock was released (`DBSIZE=0`).
- PostgreSQL held one encrypted credential row; the raw value did not contain the synthetic rotated token, while decryption yielded the rotated state.
- All disposable containers, network, volume, and probe files were removed after validation; the two images remain locally as evidence.

Automated validation after the concurrency changes:

- SDK ACP tests: 418 passed
- Cloud auth route/middleware tests: 28 passed
- SDK repository pre-commit hooks: passed
- Cloud enterprise pre-commit hooks, including package-wide mypy: passed

simonrosenberg:
## xhigh automated review — 3 findings

### 🔴 Correctness — confirmed

**1. `openhands/app_server/app_conversation/live_status_app_conversation_service.py` L2037 (and L2002): new `LookupSecret` headers collide with the SDK's redaction patterns**
`_scope_codex_subscription_auth` builds a `LookupSecret` for `CODEX_AUTH_JSON` with `headers={"X-Session-API-Key": ..., "X-Codex-Auth-Token": ...}`. Both header names match the SDK's `SECRET_KEY_PATTERNS` (`KEY`, `SESSION`, `TOKEN` in `redact.py`). This reintroduces the exact LookupSecret-for-ACP anti-pattern the comment deleted by this same diff (a few lines above, in `_build_acp_start_conversation_request`) explicitly warns against: *"ACP secrets must be StaticSecrets — LookupSecrets with JWT headers ... are redacted by the SDK serializer ... leaving headers: {} and causing provider auth to silently fail at subprocess launch."*

Verified mechanically: `LookupSecret(url=..., headers={"X-Session-API-Key": "session-a", "X-Codex-Auth-Token": "scoped.jwt.token"}).model_dump_json(context=None)` yields `{"headers":{"X-Session-API-Key":"**********","X-Codex-Auth-Token":"**********"}}`, and re-validating that JSON reloads `headers` as `{}`. Any cipher-less serialization path (e.g. `ConversationState` persisting without a cipher) hits this, so every subsequent request to the Cloud codex-auth endpoints goes out with no `X-Codex-Auth-Token` header and gets 401'd by `codex_auth.py`'s `_authorize()` — Codex ACP auth silently breaks after any state reload/resume.

**2. Same file, L2023: uncaught `ValueError` aborts every resolver-triggered ACP+Codex conversation**
`_scope_codex_subscription_auth` raises `ValueError("Cannot determine the Codex credential organization")` whenever `get_effective_org_id()` returns `None`. `ResolverUserContext` (used by every GitHub/GitLab/Jira/Slack/Bitbucket/Azure-DevOps resolver integration) never overrides `get_effective_org_id`, so it always inherits the base `UserContext` default of `None`. Any resolver-triggered conversation (e.g. a GitHub issue mention) for a user whose active agent is ACP+Codex with a stored ChatGPT-subscription credential now raises uncaught out of `_build_acp_start_conversation_request`, aborting conversation creation entirely — a regression versus current behavior, where the same request starts fine with the raw `StaticSecret`.

**3. `enterprise/server/routes/codex_auth.py` L469: `release_codex_auth` (`DELETE /codex-auth`) is a no-op**
The handler only authorizes the caller and returns `204` — it performs no revocation or state change. The scoped bearer JWT the SDK calls `_release_codex_auth_source()` to relinquish remains fully valid for GET/HEAD/PUT/refresh against `/codex-auth` and `/codex-auth/refresh` until its own `access_token_hard_timeout` expiry. A leaked or logged scoped token from a "released" conversation can still read/mutate the shared Codex secret.

---
*Generated by an xhigh-effort automated multi-agent review (finder + independent verifier pass per finding). Companion findings in [`acp_agent.py` on software-agent-sdk#4124](https://github.com/OpenHands/software-agent-sdk/pull/4124).*

enterprise/server/routes/codex_auth.py
```diff
@@ -0,0 +1,470 @@
...
+    request: Request,
+    authorization: str | None = Header(None),
+    jwt_service: JwtService = jwt_service_dependency,
+):
+    session_api_key, codex_auth_token = _decode_refresh_authorization(authorization)
+    scope = await _authorize(
+        conversation_id, session_api_key, codex_auth_token, jwt_service
+    )
+    submitted_refresh_token = await _parse_refresh_request(request)
+    async with _credential_lock(scope):
```
simonrosenberg, line 430: **[P1] Serialize full-store writes with credential refreshes**

This lock serializes the new PUT/refresh endpoints, but existing custom-secret and provider-token handlers still call `load()` followed by `SaasSecretsStore.store()`, which deletes and reinserts every custom-secret row without taking it. If such a save loads r0 before this refresh and commits after the CAS stores r1, it overwrites r1 with the now-consumed r0; the next session then encounters `refresh_token_reused` and requires reauthentication. Make those writes row-scoped or coordinate them with the same user/org lock.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -1998,6 +1999,53 @@ async def _load_skills_onto_request(
...
+            or settings.acp_server != 'codex'
+            or not isinstance(source, StaticSecret)
+        ):
+            return
+        value = source.get_value()
+        if not value or not is_chatgpt_codex_auth(value):
+            return
+        if not self.web_url or not user.id or not sandbox.session_api_key:
+            raise ValueError('Cannot scope Codex credentials to this conversation')
+        org_id = await self.user_context.get_effective_org_id()
```
simonrosenberg, line 2023: **[P1] Resolve the org for resolver-launched conversations**

For integration-triggered conversations such as GitHub, Slack, or Jira, `user_context` is a `ResolverUserContext`. That class retains `resolver_org_id` but does not override the newly added `get_effective_org_id()`, so this call returns the base implementation's `None` and line 2025 aborts stored-ChatGPT Codex launches. Override the method in `ResolverUserContext` to return its resolver/effective organization.

enterprise/storage/saas_secrets_store.py
```diff
@@ -25,6 +27,52 @@ class SaasSecretsStore(SecretsStore):
...
+        org_id = self.effective_org_id or user.current_org_id
+        async with a_session_maker() as session:
+            result = await session.execute(
+                select(StoredCustomSecrets).filter(
+                    StoredCustomSecrets.keycloak_user_id == self.user_id,
+                    StoredCustomSecrets.org_id == org_id,
+                    StoredCustomSecrets.secret_name == secret_name,
+                )
+            )
+            stored = result.scalars().one_or_none()
```
simonrosenberg, line 47: **[P2] Enforce row uniqueness before using `one_or_none()`**

The mapped `custom_secrets` table has no uniqueness constraint on `(keycloak_user_id, org_id, secret_name)`, while concurrent whole-store delete/reinsert saves can leave duplicate names. For an affected account, this `one_or_none()` raises `MultipleResultsFound`, causing every scoped Codex credential request to return 500. Add a deduplicating migration and enforce the uniqueness invariant before relying on this query.

simonrosenberg: Fixed in 9afcbfebc. Migration 138 deduplicates each user/org/name tuple by retaining the newest row before adding the unique constraint, which is also declared on the SQLAlchemy model.

enterprise/server/routes/codex_auth.py
simonrosenberg: Fixed in 9afcbfebc. SaasSecretsStore now row-locks current rows and applies only changes relative to the caller loaded snapshot, so an unrelated stale save preserves a concurrent Codex CAS rotation. The provider-update race is covered by a regression test.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
simonrosenberg: Fixed in 9afcbfebc. ResolverUserContext now returns resolver_org_id when set and otherwise delegates effective-org resolution to the underlying auth object, with tests for both paths.

simonrosenberg:
Addressed all xhigh and inline review findings in 9afcbfebc:

- Broker headers now survive cipherless LookupSecret serialization and all scoped routes authenticate those headers directly.
- Resolver-launched conversations propagate their effective organization.
- DELETE now revokes the scoped JWT in Redis until expiry.
- Full secret saves are row-locked and merge only caller changes, preserving concurrent Codex rotations.
- Migration 138 deduplicates custom secrets before enforcing user/org/name uniqueness.

Verified with 2,724 Enterprise tests, 143 root service tests, migration integrity, and both staged lint, formatting, and mypy hook suites.

enyst (commented):
I'd love it if someone working with the cloud could review this PR.

I have a little tangential question: does this failure mode happen with the app-server? That is, in OpenHands (open source), when the user runs multiple docker containers, each with its conversation, and the `codex` token expires, does a similar bug happen?

simonrosenberg:
This PR's scope is too wide, closing in favor of https://github.com/OpenHands/OpenHands/pull/15287

PR #15283, opened 2026-07-15
Feat/fix openhands provider 401

KHARSHAVARDHAN-eng:
HUMAN:

---

AGENT:

## Why

Fixes OpenHands/enterprise#61.

Selecting `openhands/` provider models could return `401 token_not_found_in_db` because the managed API key was not regenerated when the stored key was `None` or missing.

During settings persistence, `_ensure_api_key()` skipped key generation when the API key was empty, allowing an invalid value to be stored. As a result, requests to the LiteLLM Proxy were sent without a valid managed token, causing authentication to fail.

## Summary

- Update `_ensure_api_key()` to generate a managed API key whenever the stored key is missing or invalid.
- Preserve existing valid keys without generating unnecessary replacements.
- Add a regression test covering the `api_key is None` scenario.
- Ensure settings persistence always stores a valid managed API key before conversations are created.

## Testing

```bash
PYTHONPATH="enterprise:.:$PYTHONPATH" uv run pytest enterprise/tests/unit/test_saas_settings_store.py
```

- Added `test_ensure_api_key_generates_new_key_when_api_key_is_none`.
- All existing `enterprise/tests/unit/test_saas_settings_store.py` tests passed.

## Type of Change

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Documentation

## Notes

This change also prevents downstream failures caused by missing managed API keys during settings persistence, which may be related to the reported secret-management behavior.

### DIFF
--- a/enterprise/storage/saas_settings_store.py
+++ b/enterprise/storage/saas_settings_store.py
@@ -786,12 +786,22 @@ class SaasSettingsStore(SettingsStore):
         llm_api_key = item.agent_settings.llm.api_key
 
         # First, check if our current key is valid
-        if llm_api_key and not await LiteLlmManager.verify_existing_key(
-            llm_api_key.get_secret_value(),  # type: ignore[union-attr]
-            self.user_id,
-            org_id,
-            openhands_type=openhands_type,
-        ):
+        key_is_valid = False
+        if llm_api_key:
+            key_val = (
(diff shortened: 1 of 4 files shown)


neubig:

Thank you for understanding.

Issue #12681, opened 2026-01-30
[Bug]: High Client-Side GPU Usage (~30% RTX 5060 Ti) & UI Freeze caused by Frontend Rendering Loop


0xneThousand:
### Is there an existing issue for the same bug? (If one exists, thumbs up or comment on the issue instead).

- [x] I have checked the existing issues.

### Describe the bug and reproduction steps

Description: I am reporting a severe frontend performance issue where the OpenHands Web UI consumes excessive GPU resources on the client machine, eventually causing the browser tab to freeze.

The issue seems to be a rendering loop or excessive DOM/Canvas updates in the frontend code (running in the client browser), rather than a backend issue, as the high load is observed on the Windows host GPU responsible for browser hardware acceleration.

Environment Architecture (Setup Details):

My setup separates the backend execution from the frontend access:

Server-Side (Backend):

OS: Ubuntu Linux running inside a Virtual Machine (VM).

Runtime: Docker running within the Ubuntu VM.

OpenHands Status: Both openhands-app and oh-agent-server containers are running normally inside this Docker environment.

Client-Side (Frontend Access - Where the issue occurs):

Host OS: Windows 11 (Physical Machine hosting the VM).

Hardware: NVIDIA RTX 5060 Ti (16GB VRAM).

Browser: Chrome / Edge (Latest stable version).

Access Method: The Windows host browser accesses the OpenHands Web UI via the VM's IP address.

Steps to Reproduce:

Start OpenHands containers in Docker inside the Ubuntu VM.

On the Windows host, open a browser and connect to the OpenHands Web UI.

Open a conversation/task session.

Allow the session to run, particularly when logs or terminal outputs are being generated.

Observe the Windows Task Manager Performance tab.

Observed Behavior:

High Client GPU 3D Load: The Windows host GPU shows ~30% usage under the "3D" engine, triggered by the browser process.

Low VRAM Usage: Dedicated GPU Memory usage remains very low (~0.6GB total system usage), confirming the GPU is not performing local LLM inference, but is struggling with UI rendering.

Browser CPU Spike: The Browser's internal Task Manager shows the specific OpenHands tab consuming significantly more CPU than idle tabs, indicating a busy loop in JavaScript.

UI Freeze: The OpenHands web interface becomes unresponsive over time. Closing the tab immediately drops Windows GPU usage back to normal levels.

### OpenHands Installation

Docker command in README

### OpenHands Version

1.2.0

### Model Name

[...]

PR #15292, opened 2026-07-15
Fix Frontend rendering loop (#12681)

eliphatfs:
HUMAN:

AGENT:

---

## Why

During token streaming (logs / terminal output), every store merge produces a **fresh `events` array reference**, so every chat card re-renders and **re-parses its markdown on every single token**. The cost grows with conversation length, surfacing as high client-side GPU (~30%) and UI freeze — the symptoms in #12681.

## Summary

- Minimal, dependency-free memoization so **only the live streaming bubble re-renders per token**; all prior cards stay referentially stable.
- `EventMessage`: `React.memo` with a custom comparator that compares the event ref, live-state flags, `planPreviewEventIds` ref, `suppressThought`, and `messages.length` (not the array ref).
- `usePlanPreviewEvents`: split into two `useMemo`s keyed on a stable string of collected ids, so the returned `Set` keeps identity across streaming tokens.
- `MarkdownRenderer`: hoist static default components + remark/rehype plugins to module scope; `useMemo` the combined `components`/`rehypePlugins`; wrap in `React.memo`.
- `ChatMessage`: wrap in `React.memo` (shallow compare).

## Issue Number

#12681

## How to Test

1. `npm install && npm run make-i18n && npm run build` (or `npm run dev`).
2. Open a conversation and run an agent that streams a long log / terminal output (e.g. a step that emits many tokens).
3. With DevTools → Performance (or the OS GPU monitor), observe during streaming:
   - **Before:** The UI janks/freezes when LLM streaming large thinking block; happens more frequently with conversation length.
   - **After:** only the streaming bubble re-renders; prior cards do not re-parse markdown; the UI stays responsive.
4. Sanity: scroll back through a long conversation mid-stream and confirm cards render correctly and no message content is lost.

I tested this by building it into a deployed agent-canvas image and running a long streaming session — the UI stays responsive throughout the stream.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Pure frontend rendering changes only — no backend, API, websocket, or build-config changes.
- No new dependencies; uses `React.memo` / `useMemo` only.
- The custom `EventMessage` comparator deliberately compares `messages.length` rather than the array ref (which changes every token) — this is the core of the fix.
- This PR was created by an AI agent (OpenHands).

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -522,3 +522,12 @@ Called by `workspace.get_llm()` in the SDK to retrieve LLM config with the API k
 - `.github/workflows/issue-opened.yml` has a second issue-opened job that auto-applies `good first issue` after the duplicate check completes.
 - The duplicate check is used only as a veto/guardrail for `good first issue` automation: duplicate or overlapping-scope issues should not be auto-labeled.
 - The OpenHands classifier logic for newcomer suitability lives in `scripts/issue_good_first_issue_check_openhands.py`, with focused unit coverage in `tests/unit/test_issue_good_first_issue_check_openhands.py`.
+
+### Frontend streaming render performance (#12681)
+
+- During streaming, consecutive token deltas merge into the LAST event in `useEventStore` (`stores/use-event-store.ts`), producing a fresh `events`/`uiEvents` array reference on every token even though only the last entry's content changed.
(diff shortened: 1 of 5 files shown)


eliphatfs:
Sorry, I should have PR-ed to agent-canvas repository.

PR #15295, opened 2026-07-16
fix: restore automations login redirects

malhotra5:
HUMAN:

AGENT:

---

## Summary

Fixes login return navigation for the Automation app mounted at `/automations`.

## Problem

The Automation app sends unauthenticated users to the main OpenHands login page with a return target like:

```text
/login?redirect=%2Fautomations
```

After login, the main app could navigate back to `/automations` using client-side React Router navigation. That bypasses ingress, so the main OpenHands SPA handles `/automations` itself and renders a 404 instead of letting ingress route the request to the Automation service.

## Change

This PR fixes the flow in three layers:

1. **Login return target compatibility**
   - Treats legacy `redirect` as an alias for `returnTo` on the login page.
   - Rejects unsafe absolute/protocol-relative return targets.

2. **Safe cross-app document redirects**
   - Adds `navigateOrHardRedirect` for cross-app paths.
   - Uses `window.location.replace(...)` only for safe relative `/automations` paths.
   - Rejects off-origin values like `//evil.com/automations` and `https://evil.com/automations` before URL normalization.
   - Applies the helper to login, auth callback cleanup, TOS, and onboarding completion flows.

3. **Backend and stale-bundle fallback**
   - Unwraps login-wrapped `/automations` targets in backend auth/TOS/offline callback flows, returning a direct `Location: /automations...` redirect.
   - Adds a main-app `/automations/*` catch route that forces one document reload if a stale or missed SPA navigation lands in the main OpenHands router.
   - Guards that catch route with a one-shot `sessionStorage` sentinel so deployments without `/automations` ingress do not enter an infinite reload loop.

Together, these ensure `/automations` becomes a real browser request so ingress can route it to the Automation frontend, without introducing an open redirect or reload loop.

## Testing

Ran:

- `cd frontend && npm run test -- src/routes/login.test.ts src/utils/cross-app-redirect.test.ts`
- `PYTHONPATH="enterprise:." poetry run --project=enterprise pytest enterprise/tests/unit/server/routes/test_auth_onboarding.py -q`
- `poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --show-diff-on-failure`
- `cd enterprise && poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --show-diff-on-failure --files server/routes/auth.py tests/unit/server/routes/test_auth_onboarding.py`

## Deployment / QA

Feature deployment PR: OpenHands/saas-deploy#267

Preview URL:

```text
https://pr-267.staging.all-hands.dev/automations
```

The preview is pinned to:

```text
ghcr.io/openhands/enterprise-server:sha-612015f
```

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

_This PR description was updated by an AI agent (OpenHands) on behalf of the user._

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-612015f   docker.openhands.dev/openhands/openhands:612015f
```


frontend/src/utils/cross-app-redirect.ts
```diff
@@ -0,0 +1,23 @@
+import type { NavigateFunction, NavigateOptions } from "react-router";
+
+export function isCrossAppPath(destination: string): boolean {
```
hieptl, line 3: **suggestion (blocking):**

`isCrossAppPath` classifies a destination using only the pathname of a URL normalized against the current origin:

```ts
const { pathname } = new URL(destination, window.location.origin);
return pathname === "/automations" || pathname.startsWith("/automations/");
```

`new URL()` normalizes protocol-relative and absolute inputs, so both `//evil.com/automations` and `https://evil.com/automations` resolve to `pathname === "/automations"` and return `true`, even though their origin is `evil.com`. `navigateOrHardRedirect` then passes that string straight to `window.location.replace(destination)`, which performs a real cross-origin navigation.

**Why it's a problem**

This is a reachable, attacker-controlled open redirect. `use-auth-callback.ts` runs from `root-layout.tsx` (mounted on `/` and most routes) and passes an **unsanitized** `returnTo` read directly from the query string:

```ts
const returnTo = searchParams.get("returnTo");     // not validated
if (returnTo && returnTo !== "/login") destination = returnTo;
...
navigateOrHardRedirect(navigate, finalUrl, { replace: true });
```

An already-authenticated user who opens
`https://app.all-hands.dev/?login_method=github&returnTo=%2F%2Fevil.com%2Fautomations`
is redirected to `https://evil.com/automations`. (`login_method` only needs to be a valid `LoginMethod` enum value, which is public.)

This is a **new** vector introduced by this PR: the same callsite previously used React Router's `navigate()`, which keeps such a string on-origin (and throws a cross-origin `SecurityError` in `pushState`) rather than navigating away. Switching to `window.location.replace` is what makes it exploitable. Note the backend already guards against exactly this case in `_is_cross_app_relative_path` (rejects `//` and requires a leading `/`), so the two sides are currently inconsistent.

**Recommended fix**

Reject absolute / protocol-relative targets before origin-normalizing, mirroring the backend helper. `navigateOrHardRedirect` is shared and must be safe on its own — it cannot rely on individual callers pre-sanitizing (the auth-callback path proves this):

```ts
export function isCrossAppPath(destination: string): boolean {
  if (!destination.startsWith("/") || destination.startsWith("//")) return false;
  try {
    const { pathname } = new URL(destination, window.location.origin);
    return pathname === "/automations" || pathname.startsWith("/automations/");
  } catch {
    return false;
  }
}
```

Thank you very much! 🙏

frontend/src/utils/cross-app-redirect.test.ts
hieptl, line 1: **suggestion (non-blocking):**

The test suite covers automation-path matching and similarly-prefixed main-app routes, but never exercises a protocol-relative or absolute-origin input — precisely the case behind comment #1. The backend has an equivalent guard test (`test_rejects_protocol_relative_inner_redirect`); the frontend does not.

**Why it's a problem**

The security gap is invisible to CI, and there's nothing to prevent a future regression once #1 is fixed.

**Recommended fix**

Add a case that fails against the current implementation and passes once the guard is added:

```ts
it("does not treat off-origin targets as cross-app paths", () => {
  expect(isCrossAppPath("//evil.com/automations")).toBe(false);
  expect(isCrossAppPath("https://evil.com/automations")).toBe(false);
});
```

Consider also a regression test asserting `use-auth-callback` never hard-redirects off-origin.

Thank you very much! 🙏

frontend/src/routes/cross-app-redirect.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+export default function CrossAppRedirect() {
+  React.useEffect(() => {
+    window.location.replace(window.location.href);
```
hieptl, line 29: **suggestion (non-blocking):**

The catch route does `window.location.replace(window.location.href)`. This relies on ingress peeling off `/automations` and routing it to the Automation service on the resulting document request.

**Why it's a problem**

The route is registered unconditionally in the shared `frontend/src/routes.ts`. In any deployment where ingress does **not** intercept `/automations` (OSS/local dev, or a misconfigured/absent Automation service), the reload re-serves the main SPA, which re-matches `/automations/*` → renders `CrossAppRedirect` → reloads again → a tight infinite reload loop.

**Recommended fix**

Guard against re-entry, e.g. a one-shot `sessionStorage` sentinel that is cleared on a successful cross-app load, or gate the route/behavior on saas `app_mode`. At minimum, add a comment documenting the hard dependency on ingress handling `/automations` so the loop risk is explicit.

Thank you very much! 🙏

frontend/src/utils/cross-app-redirect.ts
malhotra5: Fixed in the latest push. `isCrossAppPath` now rejects non-relative and protocol-relative destinations before URL normalization (`!startsWith("/") || startsWith("//")`), matching the backend guard and preventing `window.location.replace(...)` from receiving off-origin `/automations` URLs.

_This review reply was created by an AI agent (OpenHands) on behalf of the user._

frontend/src/utils/cross-app-redirect.test.ts
malhotra5: Added the suggested regression coverage for both protocol-relative and absolute off-origin targets:

- `//evil.com/automations`
- `https://evil.com/automations`

Both now assert `isCrossAppPath(...) === false`.

_This review reply was created by an AI agent (OpenHands) on behalf of the user._

frontend/src/routes/cross-app-redirect.tsx
malhotra5: Fixed in the latest push. The `/automations/*` catch route now uses a one-shot `sessionStorage` sentinel keyed by the current URL before calling `window.location.replace(...)`. If ingress does not route `/automations` and the main SPA is served again, the route shows a fallback with a retry action instead of entering an infinite reload loop.

_This review reply was created by an AI agent (OpenHands) on behalf of the user._

hieptl (approved):
Thank you! 🙏

PR #15300, opened 2026-07-16
fix: use blobless partial clone instead of shallow clone by default

jpshackelford:
## Why

Repository startup currently defaults to a **shallow clone** (`git clone --depth 1`). Shallow clones truncate commit history, which breaks the history-dependent Git operations our coding agent performs most often when **opening PRs** — merge, rebase, cherry-pick, and blame. Because the merge base is frequently missing, users hit `fatal: refusing to merge unrelated histories`, spurious conflicts, and huge/wrong diffs. A customer reported exactly this and worked around it by cloning with `--filter=blob:none` instead.

The existing "Fetch full Git history" setting only offered the opposite extreme — a full clone that re-downloads *all* historical file content up front, reintroducing the slow-startup problem shallow cloning was meant to avoid. There was no middle option.

## What changed

- **Default clone strategy** is now a **blobless partial clone**: `git clone --filter=blob:none` (branch-scoped with `--single-branch --branch <branch>` when a branch is selected) instead of `--depth 1`.
  - Keeps the **complete commit graph and trees**, so merges, rebases, cherry-picks, tags, blame, and older-commit checkouts all work correctly (a valid merge base is always present).
  - Still avoids the up-front download of historical file **contents**; blobs are fetched lazily on demand. Startup stays fast for large repositories, and lazy fetching is fine because the sandbox already has remote connectivity.
- The `git_full_clone` user setting is unchanged in shape: when enabled it still performs a full clone, for tasks that must work offline after cloning or run repo-wide operations at once.
- **Agent workspace context note** updated: instead of warning that the clone is shallow and suggesting `git fetch --unshallow`, it now explains that history is complete and only file contents are fetched on demand.
- **Settings copy** corrected (it previously described the default as shallow and implied the toggle was needed to get history/merge bases at all, which is no longer accurate). New copy is in plain language and does not reference internal `git` flags. All 15 locales updated.
- Affected unit tests updated to expect the new clone commands and renamed context.

## How to test

- `poetry run pytest tests/unit/app_server/test_app_conversation_service_base.py tests/unit/app_server/test_live_status_app_conversation_service.py -q`
- `cd frontend && node scripts/check-translation-completeness.cjs`

In this environment a full `poetry run pytest` was blocked by an out-of-sync `poetry.lock` (unrelated to this change). Validation performed here: `python -m py_compile` on all edited Python files, JSON validity of `translation.json`, and the Node translation-completeness check — all pass.

## Type

- [x] Bug fix

## Notes / follow-ups

- Caveat inherent to partial clones: an operation touching many files or old revisions (e.g. a repo-wide `git grep` over history, mass blame, or a full build) triggers a burst of lazy blob fetches and requires network access to the remote. The "Download all repository files up front" setting remains the escape hatch for those workloads.
- The `git_full_clone` boolean still only expresses two states (partial vs. full). If a distinct "minimal / shallow tip-only" mode is ever wanted for explicitly no-PR tasks, that would require evolving the setting into an enum — out of scope here.

---

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-fc6ee4d   docker.openhands.dev/openhands/openhands:fc6ee4d
```


neubig:

Thank you for understanding.

PR #15311, opened 2026-07-17
fix: proxy Canvas runtime requests through Cloud

simonrosenberg:
## Why

Agent Canvas is deployed as a static Cloud frontend. Its runtime requests currently target the removed `/api/cloud-proxy` endpoint on the Canvas host, causing 503s.

## Change

Add an authenticated, conversation-scoped Cloud API endpoint for the runtime operations Canvas uses. The API resolves the runtime server-side, attaches its session credential, and accepts only the fixed operation allowlist. It does not accept a caller-controlled upstream host.

## Validation

- `poetry run pytest tests/unit/app_server/test_app_conversation_router.py`
- `poetry run pre-commit run --config ./dev_config/python/.pre-commit-config.yaml --files openhands/app_server/app_conversation/app_conversation_models.py openhands/app_server/app_conversation/app_conversation_router.py tests/unit/app_server/test_app_conversation_router.py`

## Follow-up

[Agent Canvas PR #1837](https://github.com/OpenHands/agent-canvas/pull/1837) calls this endpoint instead of the obsolete runtime proxy route.

### DIFF
--- a/openhands/app_server/app_conversation/app_conversation_models.py
+++ b/openhands/app_server/app_conversation/app_conversation_models.py
@@ -415,6 +415,12 @@ class SwitchAcpModelRequest(BaseModel):
     )
 
 
+class RuntimeRequest(BaseModel):
+    method: Literal['GET', 'POST']
+    path: str = Field(min_length=1)
+    body: Any = None
+
+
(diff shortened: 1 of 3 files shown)


neubig:

Thank you for understanding.

PR #15312, opened 2026-07-17
fix(enterprise): give the keycloak_auth cookie a Max-Age matching its JWS TTL

aivong-openhands:
HUMAN:

AGENT:

---

## Why

The `keycloak_auth` cookie is written without `Max-Age`/`Expires`, making it a browser *session* cookie: fully quitting the browser discards it, and the user lands back on the login page even though the signed JWS inside the cookie is still valid for a week and their Keycloak session is still alive. This surfaced while investigating repeated forced re-logins on staging — any browser that doesn't restore session cookies logs the user out on every restart.

## Summary

- `set_chunked_cookie` accepts an optional `max_age` and applies it to every chunk (a chunk outliving its siblings would reassemble into a torn token).
- `set_response_cookie` passes `max_age` derived from the same 1-week TTL already used to sign the JWS (`AUTH_COOKIE_TTL`), so the cookie and the token inside it expire together.
- Added regression tests: `max_age` lands on all chunks; omitting it still writes a session cookie.

## Issue Number

N/A

## How to Test

- `PYTHONPATH=enterprise poetry run pytest --noconftest enterprise/tests/unit/server/auth/test_cookie_chunking.py` — 11 passed.
- Manual: log in, inspect the `keycloak_auth` cookie in devtools — Expires should be ~7 days out instead of "Session". Fully quit the browser, reopen, revisit the app: still logged in (previously bounced to the login page).

## Video/Screenshots

N/A (cookie attribute change; behavior visible in devtools per steps above)

_This PR was drafted by an AI agent on behalf of the user._

### DIFF
--- a/enterprise/server/auth/cookie_chunking.py
+++ b/enterprise/server/auth/cookie_chunking.py
@@ -61,12 +61,18 @@ def set_chunked_cookie(
     secure: bool = True,
     httponly: bool = True,
     samesite: str = 'lax',
+    max_age: int | None = None,
 ) -> None:
     """Set ``key`` to ``value``, splitting across sibling cookies if needed.
 
     Always clears trailing chunk indices that this write does not use, so a
     value that shrank from N chunks to fewer does not leave stale chunks
     that ``read_chunked_cookie`` would wrongly append.
+
+    ``max_age`` (seconds) makes the cookie persistent so the session
+    survives a full browser quit; without it the browser discards the
+    cookie on exit. It is applied to every chunk — a chunk outliving its
+    siblings would reassemble into a torn token.
     """
     chunks = [value[i : i + CHUNK_SIZE] for i in range(0, len(value), CHUNK_SIZE)] or [
         ''
@@ -80,6 +86,8 @@ def set_chunked_cookie(
         }
         if domain:
             kwargs['domain'] = domain
+        if max_age is not None:
+            kwargs['max_age'] = max_age
         response.set_cookie(key=_chunk_key(key, i), value=chunk, **kwargs)
 
     # Expire any chunks left over from a previously larger value.
--- a/enterprise/server/routes/auth.py
+++ b/enterprise/server/routes/auth.py
@@ -95,6 +95,12 @@ def create_provider_tokens_object(
     return MappingProxyType(provider_information)
 
 
+# The signed JWS inside the cookie and the cookie itself expire together;
+# a persistent Max-Age lets the session survive a full browser quit instead
+# of dying with the browser process as a session cookie.
+AUTH_COOKIE_TTL = timedelta(weeks=1)
+
+
 def set_response_cookie(
     request: Request,
     response: Response,
@@ -112,7 +118,7 @@ def set_response_cookie(
     from storage.encrypt_utils import get_jwt_service
 
     signed_token = get_jwt_service().create_jws_token(
-        cookie_data, expires_in=timedelta(weeks=1)
+        cookie_data, expires_in=AUTH_COOKIE_TTL
     )
 
     # Set secure cookie with signed token. The value can exceed the
@@ -128,6 +134,7 @@ def set_response_cookie(
         secure=secure,
         httponly=True,
         samesite=get_cookie_samesite(),
+        max_age=int(AUTH_COOKIE_TTL.total_seconds()),
     )
 
 
--- a/enterprise/tests/unit/server/auth/test_cookie_chunking.py
+++ b/enterprise/tests/unit/server/auth/test_cookie_chunking.py
@@ -98,3 +98,23 @@ def test_boundary_sizes_roundtrip(size):
     value = 'z' * size
     got, _ = _roundtrip(value)
     assert got == value
+
+
+def test_max_age_applied_to_every_chunk():
+    # Without Max-Age the cookie is a session cookie and dies on browser
+    # quit; every chunk must carry it or the reader reassembles a torn token.
+    resp = Response()
+    set_chunked_cookie(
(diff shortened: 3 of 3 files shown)


aivong-openhands:
Closing this after weighing the trade-off rather than merging.

**Context.** This was filed while investigating repeated forced re-logins on staging. The dominant cause turned out to be environmental: the staging Keycloak realm was running factory-default SSO session lifetimes (30 min idle / 10 h max) instead of the intended 3 d / 7 d, and that has since been fixed directly on the realm. With SSO sessions alive for days, a user who fully quits their browser lands on the login page but one click bounces them through Keycloak's live SSO session straight back into the app — no credential re-entry. So the remaining benefit of this change is removing that single click per browser restart.

**What the change would cost:**

- Closing the browser would no longer end the session on shared machines — the cookie persists up to 7 days unless the user explicitly logs out. Industry-standard behavior, but it is a security-posture change, and "session dies with the browser" is a defensible stance for an enterprise product.
- The signed token pair would be guaranteed-persisted to disk for up to 7 days (vs best-effort via browser session restore today).
- Stale cookies stop self-healing via browser restart. Most classes still self-heal server-side (any auth failure deletes all chunks), but cookies left under a different Domain scope after a cookie-attribute migration can't be deleted by the server and would linger up to 7 days instead of dying at restart — relevant to any future host/domain migration.

**Verdict:** modest UX polish against a real (if standard) posture change plus migration friction. Not worth it while the click-through re-login is this cheap. If week-long browser sessions become a product requirement, this branch is a working starting point — and it should then also add dual-scope cookie deletion so attribute migrations stay self-healing.

_This comment was drafted by an AI agent on behalf of the user._

aivong-openhands:
## Impact assessment: self-hosted / customer installs vs SaaS

**Verdict: uniform, unconditional behavior change on every deployment surface — the `keycloak_auth` cookie becomes a 7-day persistent cookie everywhere (no code path, host heuristic, or config knob scopes or disables it); no functional breakage anywhere, but it is a security-posture change, which is why this PR was closed rather than merged (see trade-off comment above).**

### What changed

1. `set_chunked_cookie` (`enterprise/server/auth/cookie_chunking.py`) gains an optional `max_age: int | None = None`, applied to every chunk. Sole caller: `set_response_cookie` (`routes/auth.py:129`). All other chunking helpers (`read_chunked_cookie`, `delete_chunked_cookie`) untouched.
2. `set_response_cookie` (`routes/auth.py`) passes `max_age` from a new `AUTH_COOKIE_TTL = timedelta(weeks=1)` — the same TTL that already signs the JWS, so cookie and token expire together. Its 5 call sites all inherit the change: `middleware.py:46` (refresh re-set), `routes/auth.py:646` and `routes/auth.py:1043` (auth callbacks), `routes/email.py:91` and `routes/email.py:175`.

### Environment × flag matrix (before → after)

| Deployment | Cookie lifetime | Online SSO idle / max (live) | Cookie can outlive SSO session? | Result |
|---|---|---|---|---|
| Customer self-hosted (OHE / Replicated) | session → 7 d | chart realm defaults 3 d / 7 d; KOTS-configurable | yes, if customer shortens SSO | changed (posture) |
| SaaS production | session → 7 d | not re-verified for this analysis | unknown | changed (posture) |
| SaaS dev | session → 7 d | 3 d / 7 d (verified via kcadm) | edge only (idle < TTL) | changed (posture) |
| SaaS staging | session → 7 d | 3 d / 7 d (verified via kcadm) | edge only (idle < TTL) | changed (posture) |
| SaaS feature previews | session → 7 d | shares the staging realm → 3 d / 7 d | edge only | changed (posture) |

Cookie behavior is host-independent, so the before→after column is uniform by construction; it is validated by this PR's unit tests (Max-Age lands on every chunk; omitting it preserves session-cookie behavior). SSO lifetimes were verified live for dev and staging; OHE values are the chart realm-template defaults; SaaS production was not re-verified.

### Behaviors that change, and for whom

- **Everyone, every surface:** fully quitting the browser no longer ends the session; the cookie survives up to 7 days or until explicit logout / server-side deletion.
- **Everyone, every surface:** the signed token pair is guaranteed-persisted to the browser's on-disk cookie store (previously best-effort via session restore).
- **Surfaces where SSO lifetime < 7 d** (configurable on OHE): returning users can present a live cookie holding a dead refresh token → one 401 bounce, server deletes the cookie, user re-logs-in. Existing path, exercised more often.
- **Any future cookie-attribute migration (domain/samesite change), every surface:** stale cookies under the old Domain scope stop self-healing at browser restart and linger up to 7 days; server deletions only match the current scope. Mitigation if revived: dual-scope deletion.

### Risks to customer installs

1. **Session persistence on shared machines** — closing the browser stops logging users out. Industry-standard for SaaS, but a posture change an enterprise customer cannot opt out of: the code offers no configuration to keep session-cookie behavior.
2. **No functional breakage** — auth flows, chunking round-trip, deletion paths, and API-key auth are unchanged; regression tests cover the omitted-`max_age` path staying byte-identical.
3. **Bounded exposure** — a lifted cookie is only as useful as the Keycloak session behind it; server-side revocation (killing the SSO session) invalidates it immediately, and the JWS self-expires at 7 days.

### Rollout note

Ships in the shared enterprise-server image: merge → `cloud-X.Y.Z` image build → per-environment image-pin bumps (SaaS via deployment repo pins; feature previews via the staging version pin) → OHE customers via the next KOTS release that pins that enterprise tag. There is no flag to stage or scope the rollout — first deploy of the image changes cookie behavior for all users of that surface. **This PR was deliberately closed**; this assessment is recorded so a future revival starts from the full picture.

---
This analysis was produced by an AI agent (OpenHands) on behalf of @aivong-openhands.

PR #15314, opened 2026-07-17
Add execution_status__eq filter to app-conversations search/count endpoints

jpshackelford:
## Summary

Closes OpenHands/enterprise#65.

Adds an `execution_status__eq` query parameter to the V1 conversation endpoints so clients can filter by execution status server-side instead of fetching all conversations and filtering client-side:

- `GET /api/v1/app-conversations/search?execution_status__eq=running`
- `GET /api/v1/app-conversations/count?execution_status__eq=running`

The value is validated against the SDK `ConversationExecutionStatus` enum (e.g. `running`, `finished`, `error`, `stuck`, `idle`, `paused`, `waiting_for_confirmation`, `deleting`).

## Why not `sandbox_status__eq`?

The issue also proposed `sandbox_status__eq`. Unlike `execution_status`, `sandbox_status` is **not** persisted — it is computed live from the sandbox service when building each `AppConversation`. Filtering on it at the SQL/pagination layer isn't possible without a much larger change (it would break offset-based pagination and require fetching live status before paginating). `execution_status`, on the other hand, is already stored in the `StoredConversationMetadata.execution_status` column and updated via the webhook callback, so it can be filtered efficiently in the query. This PR therefore implements the stored, index-friendly `execution_status__eq` filter that satisfies the primary use case in the issue title (filtering by execution status, e.g. only `RUNNING` conversations).

## Changes

- Router: added `execution_status__eq` query param to `search_app_conversations` and `count_app_conversations`.
- Threaded the parameter through:
  - `AppConversationService` (abstract) + `LiveStatusAppConversationService`
  - `AppConversationInfoService` (abstract) + `SQLAppConversationInfoService`
  - `SaasSQLAppConversationInfoService` (enterprise override)
- Applied the SQL condition in `_apply_filters` / `_apply_filters_with_saas_metadata` against the stored `execution_status` column.

## Tests

- SQL info service: search + count filtered by `execution_status__eq` (RUNNING / FINISHED / ERROR).
- Router: `execution_status__eq` is passed through on both `search` and `count` (and defaults to `None`).

All targeted unit tests pass locally (`test_sql_app_conversation_info_service.py`, `test_app_conversation_router.py`, `test_live_status_app_conversation_service.py`), along with `ruff check`/`ruff format`.

---

@jpshackelford can click here to [continue refining the PR](https://app.all-hands.dev/conversations/4ded0a1b-a5d5-4c2c-af1d-27e3eaaadcf7)

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-6f7ccfa   docker.openhands.dev/openhands/openhands:6f7ccfa
```

### DIFF
--- a/enterprise/server/utils/saas_app_conversation_info_injector.py
+++ b/enterprise/server/utils/saas_app_conversation_info_injector.py
@@ -25,6 +25,7 @@ from openhands.app_server.app_conversation.sql_app_conversation_info_service imp
 from openhands.app_server.errors import AuthError
 from openhands.app_server.services.injector import InjectorState
 from openhands.app_server.user.specifiy_user_context import ADMIN
+from openhands.sdk.conversation import ConversationExecutionStatus
 
 
 class SaasSQLAppConversationInfoService(SQLAppConversationInfoService):
@@ -136,6 +137,7 @@ class SaasSQLAppConversationInfoService(SQLAppConversationInfoService):
         updated_at__gte: datetime | None = None,
         updated_at__lt: datetime | None = None,
         sandbox_id__eq: str | None = None,
+        execution_status__eq: ConversationExecutionStatus | None = None,
         sort_order: AppConversationSortOrder = AppConversationSortOrder.CREATED_AT_DESC,
         page_id: str | None = None,
         limit: int = 100,
@@ -159,6 +161,7 @@ class SaasSQLAppConversationInfoService(SQLAppConversationInfoService):
             updated_at__gte=updated_at__gte,
             updated_at__lt=updated_at__lt,
             sandbox_id__eq=sandbox_id__eq,
+            execution_status__eq=execution_status__eq,
         )
 
         # Add sort order
@@ -217,6 +220,7 @@ class SaasSQLAppConversationInfoService(SQLAppConversationInfoService):
         updated_at__gte: datetime | None = None,
         updated_at__lt: datetime | None = None,
         sandbox_id__eq: str | None = None,
+        execution_status__eq: ConversationExecutionStatus | None = None,
     ) -> int:
         """Count conversations matching the given filters with SAAS metadata."""
         query = (
@@ -240,6 +244,7 @@ class SaasSQLAppConversationInfoService(SQLAppConversationInfoService):
             updated_at__gte=updated_at__gte,
             updated_at__lt=updated_at__lt,
             sandbox_id__eq=sandbox_id__eq,
+            execution_status__eq=execution_status__eq,
         )
 
         result = await self.db_session.execute(query)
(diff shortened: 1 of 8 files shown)


jpshackelford:
### Follow-up: sandbox status filtering (deliberately scoped out of this PR)

This PR is intentionally limited to `execution_status__eq` on the conversation `search`/`count` endpoints. We also wanted to let clients filter by **sandbox status** (the `sandbox_status__eq` idea from the issue, and/or a status filter on `GET /api/v1/sandboxes/search`), but ran into blockers that make it a larger, cross-repo change than fits here.

**Blockers, based on current understanding:**

1. **`sandbox_status` isn't stored in OpenHands' DB.** Unlike `execution_status` (persisted on `StoredConversationMetadata.execution_status`), sandbox status is computed live when each `AppConversation`/`SandboxInfo` is built. There's no queryable column to filter on.
2. **`RemoteSandboxService.search_sandboxes` paginates before status exists.** It runs a SQL query with `OFFSET`/`LIMIT`, and *only then* batch-fetches runtime data and derives status in `_to_sandbox_info`. Filtering after pagination returns short/empty pages while later pages still contain matches — i.e. it breaks offset-based pagination (the same reason `sandbox_status__eq` was dropped from this PR).
3. **The authoritative status lives in `runtime-api`, reachable only by id.** `runtime-api` *does* persist status (`Runtime.status`, values `running|stopped|paused|error|starting`) and already filters on it internally (e.g. `list_runtimes` does `.filter(Runtime.status == RUNNING)`). But the endpoint OpenHands calls, `GET /sessions/batch`, takes **ids**, not a status filter — so OpenHands can't push a `WHERE status = …` down to the source of truth today.

**Best path forward:**

- Add a **paginated, status-filtered list endpoint to `runtime-api`** (its `runtimes` table already stores `status` and already filters on it), then have `RemoteSandboxService.search_sandboxes` delegate pagination + filtering to it. This keeps filtering at the true source of truth and is pagination-safe. It's a two-PR, cross-repo change (`runtime-api` + `OpenHands`).
- Alternatively (heavier), mirror `status` into OpenHands' `StoredRemoteSandbox` and keep it updated via the existing status callbacks, so OpenHands can filter its own query directly — at the cost of a second copy that can go stale.

**Do we first need to retire `DockerSandboxService` / `ProcessSandboxService`?** No — not strictly. The pagination hazard is specific to `RemoteSandboxService` (SQL-paginate then compute status). The local backends materialize the full list in memory *before* slicing, so they could filter by status pre-pagination with a trivial in-memory check. So all three backends *could* support a status filter.

That said, **retiring the two local backends would simplify this**: one backend, one consistent status-filter code path, and no divergent per-backend implementations — concentrating the work on the single backend (remote) that carries the real complexity. That retirement may be feasible soon, since the local-runtime functionality has moved into the `OpenHands/agent-canvas` project.

_This comment was posted by an AI agent (OpenHands) on behalf of the PR author._

neubig:

Thank you for understanding.

PR #15323, opened 2026-07-19
fix: accept '1' as truthy for SERVE_FRONTEND and AWS_S3_SECURE env toggles

mdshzb04:
HUMAN:

This PR fixes a bug where setting the env vars `SERVE_FRONTEND=1` or `AWS_S3_SECURE=1` was treated as "off" instead of "on", because the code only accepted the string `true`. This broke the frontend for Helm deployments that pass `1`, and worse, silently turned off TLS for S3/MinIO connections. The fix makes these three toggles accept both `true` and `1`, as the repo's own AGENTS.md rule requires, and adds tests so it can't regress.

AGENT:

---

## Why

`SERVE_FRONTEND` (`openhands/app_server/app.py`) and `AWS_S3_SECURE` (`openhands/app_server/file_store/s3.py`, `openhands/app_server/event/aws_event_service.py`) only accepted the literal string `'true'` and silently treated `'1'` as false. This violates the "Environment Variable Enable Toggles" rule in AGENTS.md, which requires accepting both `'true'` and `'1'` because older Helm chart versions default to `'1'`.

Impact:
- `SERVE_FRONTEND=1` silently disabled serving the built frontend (`/` returned 404 while the API stayed up).
- `AWS_S3_SECURE=1` silently downgraded the S3/MinIO client from HTTPS to plain HTTP — a security-relevant misconfiguration with no warning.

## Summary

- Added `env_flag_enabled(name, default)` to `openhands/app_server/utils/environment.py` implementing the documented `('true', '1')` pattern.
- Switched the three affected call sites (`SERVE_FRONTEND`, both `AWS_S3_SECURE` reads) to the helper, preserving their default-true behavior.
- Added regression tests: unit tests for the helper, `'1'`/`'0'` cases in the existing `_get_default_aws_endpoint_url` test class, and a new test class asserting `use_ssl` passed to `boto3.client` for each spelling.

Scope is intentionally limited to the three confirmed call sites from the issue; adjacent feature gates mentioned in the issue comments are left for a follow-up if maintainers want them changed.

## Issue Number

Fixes OpenHands/enterprise#57

## How to Test

```bash
uv run pytest tests/unit/utils/test_environment.py tests/unit/app_server/test_aws_event_service.py tests/unit/app_server/file_store/test_file_store.py
```

Runtime check: start the server with `SERVE_FRONTEND=1` and confirm `/` serves the SPA (previously returned 404).

### DIFF
--- a/openhands/app_server/app.py
+++ b/openhands/app_server/app.py
@@ -25,6 +25,7 @@ from openhands.app_server.middleware import (
 )
 from openhands.app_server.static import SPAStaticFiles
 from openhands.app_server.status.status_router import router as health_router
+from openhands.app_server.utils.environment import env_flag_enabled
 from openhands.app_server.version import get_version
 
 # Initialize the Tavily MCP proxy before creating the app
@@ -72,7 +73,7 @@ app.include_router(v1_router.router)
 app.include_router(health_router)
 
 # Middleware and static file setup (merged from listen.py)
-if os.getenv('SERVE_FRONTEND', 'true').lower() == 'true':
+if env_flag_enabled('SERVE_FRONTEND', default=True):
     if os.path.isdir('./frontend/build'):
         app.mount(
             '/', SPAStaticFiles(directory='./frontend/build', html=True), name='dist'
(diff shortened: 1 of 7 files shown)


mdshzb04:
Confirming this PR is ready for review. All 13 CI checks pass; the HUMAN section note has been added as requested.

neubig:

Thank you for understanding.

PR #15325, opened 2026-07-19
fix(db): create conversation_cost_events table on OSS (migration 015)

shanemort1982:
HUMAN:

Adds a new alembic migration (015) that creates the `conversation_cost_events` table on OSS installs, so the per-event cost-delta writer (already in `main`) stops hitting `no such table: conversation_cost_events` on a fresh SQLite DB. Mirrors the enterprise migration 134 (columns, indexes, CASCADE FK to `conversation_metadata`) without pulling in the rest of the enterprise migration tree. Verified end-to-end on a self-hosted OSS box: migration applies 014 -> 015, the table exists with both indexes + the CASCADE FK, the recurring `no such table` errors stopped, and per-event cost deltas persist.

- [ ] A human has tested these changes locally

---

## Why

Self-hosted OSS installs raise `no such table: conversation_cost_events` on every cost-delta write (hundreds/hour), breaking cost tracking and poisoning the DB session with `PendingRollbackError`.

Root cause: the per-event cost-delta write code in `sql_app_conversation_info_service.py` (added in acc8412ae, "feat: track per-event cost deltas") writes to `conversation_cost_events`, but that table is created by the **enterprise** migration tree only (revision 134, cascade-fixed in 136). The OSS migration set (001-014) never creates it, so a fresh OSS DB ends up stamped at head 014 without the table, while the app keeps writing to it.

This shipped undetected because the existing unit test (`test_webhook_router_stats.py`) builds the schema via `Base.metadata.create_all`, which creates every model table including `conversation_cost_events`, masking the missing alembic migration.

## What

Adds OSS alembic migration `015` that creates `conversation_cost_events` in its final upstream form: the enterprise 134 schema with the 136 `ondelete='CASCADE'` foreign key already applied. The OSS `StoredConversationCostEvent` model already declares the CASCADE FK and both indexes, so this just materialises the model's schema in the OSS migration set.

| | |
|---|---|
| revision | `015` |
| down_revision | `014` |
| columns | `id` (Identity PK), `conversation_id` (FK -> conversation_metadata, CASCADE), `cost_delta` (Float, default 0.0), `occurred_at` (DateTime, tz) |
| indexes | `ix_conversation_cost_events_conversation_id`, `ix_conversation_cost_events_occurred_at` |

## Test plan

- New `tests/unit/app_server/test_migration_015_conversation_cost_events.py` runs `upgrade()` directly against a fresh SQLite DB (not `metadata.create_all`) and asserts the table, both indexes, and the `ondelete='CASCADE'` FK to `conversation_metadata`. It fails without the migration and passes with it.
- Downgrade test asserts the table is dropped.
- Verified end-to-end on a self-hosted OSS host: after applying migration 015 the recurring `no such table: conversation_cost_events` errors stopped, and cost-delta rows are now inserted correctly on every metrics flush.

### DIFF
--- a/openhands/app_server/app_lifespan/alembic/versions/015.py
+++ b/openhands/app_server/app_lifespan/alembic/versions/015.py
@@ -1,0 +1,73 @@
+"""Create conversation_cost_events table (OSS port of enterprise 134/136)
+
+The per-event cost-delta write code in
+``sql_app_conversation_info_service.py`` (added in commit acc8412ae,
+"feat: track per-event cost deltas") inserts rows into
+``conversation_cost_events`` on every positive cost delta. That table is
+created by the *enterprise* migration tree (revision 134, cascade-fixed in
+136), not the OSS tree, so on a self-hosted OSS install the table never exists
+and every cost-delta write raises ``no such table: conversation_cost_events``
+(hundreds/hour), breaking cost tracking and poisoning the DB session with
+``PendingRollbackError``.
+
+The OSS ``StoredConversationCostEvent`` model already declares the
+``ondelete='CASCADE'`` foreign key and both indexes, so this migration simply
+materialises the model's schema in the OSS migration set, using the final
+upstream form (the enterprise 134 schema with the 136 CASCADE foreign key
+already applied).
+
+Revision ID: 015
+Revises: 014
+Create Date: 2026-07-17 00:00:00.000000
+"""
+
+from typing import Sequence
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = '015'
+down_revision: str | None = '014'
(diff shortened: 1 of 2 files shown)


shanemort1982:
@enyst could you take a look? Verified end-to-end on a self-hosted OSS install: the migration applies cleanly (alembic 014 -> 015), the `conversation_cost_events` table is created with the CASCADE FK and both indexes, the recurring `no such table: conversation_cost_events` errors stopped, and per-event cost deltas now persist correctly on every metrics flush. The new unit test runs `upgrade()` directly (not `metadata.create_all`) so this regression is now caught at test time. Ready for review.

neubig:

Thank you for understanding.

PR #15329, opened 2026-07-20
feat(backend): support conversation tags for write and search (#15316)

Mister2005:
HUMAN:
This PR adds the ability to associate arbitrary key-value tags with app conversations when they are started or updated. This makes it possible to query and filter conversations later via the `/search` and `/count` endpoints using the tags.

This adds the ability to associate arbitrary key-value tags with app conversations when they are started or updated, making it possible to query and filter them later via the /search endpoint.

- Adds tags to AppConversationStartRequest and AppConversationUpdateRequest.
- Exposes tag as a query parameter (e.g. tag=key=value) in /search and /count endpoints.
- Plumbs tag filters through the app_server services down to the SQLAlchemy models, parsing and executing JSON equality checks on the database.

Fixes OpenHands/enterprise#66


neubig:

Thank you for understanding.

PR #15330, opened 2026-07-20
fix(saas-auth): async last_used_at + stop mis-classifying SQL errors as 401 (OHE-2792)

tofarr:
HUMAN:

AGENT:

---

## Why

`/api/v1/app-conversations` (and any other endpoint gated by `saas_user_auth_from_bearer`) intermittently returns **401 BearerTokenError** when the same API key is used for concurrent requests. A customer running Windmill against `cloud-1.46.2` reported that serial calls always succeed while concurrent calls sometimes fail.

Two combined root causes:

1. **`enterprise/storage/api_key_store.py::validate_api_key`** did a synchronous
   `UPDATE api_keys SET last_used_at = ?` + `COMMIT` on every request. With two
   concurrent requests for the same key, the second one blocked on the row lock.
   Pool / lock-wait timeouts surfaced as `SQLAlchemyError`. (`last_used_at` is
   only consumed as a UI signal by the API-keys listing — the user confirmed
   eventual consistency is acceptable.)
2. **`enterprise/server/auth/saas_user_auth.py::saas_user_auth_from_bearer`**
   ended in a blanket `except Exception as e: raise BearerTokenError(...)`. That
   wrapped the transient `SQLAlchemyError` from #1 into a credential failure,
   which the auth middleware translated into the 401 the customer saw — and
   disabled retry-backoff logic that assumes 401 = bad credential.

## Summary

- **`api_key_store.py`** — `validate_api_key` no longer writes on the auth hot
  path. It returns immediately and schedules `asyncio.create_task(self._touch_last_used_at(...))`.
  The background task runs a throttled, conditional UPDATE:
  `UPDATE api_keys SET last_used_at=:now WHERE id=:id AND (last_used_at IS NULL OR last_used_at <= now - 10s)`.
  When the row is fresh, zero rows match — no write, no lock contention.
  Failures are logged and swallowed; the credential has already been validated
  by this point so a write failure must not affect the response.
- **`saas_user_auth.py`** — exception classification is now explicit:
  - `NoCredentialsError` → re-raise (no credentials supplied)
  - `SQLAlchemyError`   → `HTTPException(503, detail="auth_store_unavailable", headers={"Retry-After":"1"})`
  - `AuthError`         → re-raise (BearerTokenError still maps to 401)
  - `Exception`         → re-raise (becomes 500 in middleware)
  The blanket `except Exception → BearerTokenError` is gone.

## Issue Number

OHE-2792

## How to Test

```bash
# Targeted unit tests for the changed code paths
PYTHONPATH=".:enterprise" poetry run pytest \
  enterprise/tests/unit/test_api_key_store.py \
  enterprise/tests/unit/test_saas_user_auth.py \
  --no-header -q
```

The new regression test `test_validate_api_key_concurrent_same_key_no_contention`
fires 100 concurrent validations against the same key and asserts that all 100
return matching `user_id` without raising. Before this fix the test would
fail intermittently with `sqlalchemy.exc.OperationalError`.

The new classification test `test_saas_user_auth_from_bearer_sqlalchemy_error_returns_503`
asserts that an `OperationalError` from `validate_api_key` now surfaces as a
retryable HTTP 503 with `Retry-After: 1` — **not** a 401.

Manual verification (recommended before merging):

```bash
# Concurrent requests with the same API key should never produce 401.
for i in {1..50}; do
  curl -sS -o /dev/null -w "%{http_code}\n" \
    -H "Authorization: Bearer $API_KEY" \
    https://<staging>/api/v1/app-conversations &
done | sort | uniq -c
# Expect: only "200", no "401".
```

## Video/Screenshots

N/A — backend-only change; behaviour is exercised through the unit tests above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **No new dependencies.** The fix uses `asyncio`, `datetime.timedelta`, and
  `sqlalchemy.or_`, all already in scope.
- **`anyio` pin (4.9.0)** in `poetry.lock` is incompatible with the installed
  `coredis==4.24.0` in this sandbox. I worked around it locally by upgrading
  `anyio` to 4.14.2 to run the affected tests. The on-disk `poetry.lock` /
  `enterprise/pyproject.toml` were not modified. If CI reproduces the
  CollectionError for `test_auth_middleware.py` /
  `enterprise/server/auth/saas_user_auth.py`, the `anyio` pin should be
  revisited separately — it predates this PR.
- **Backwards compatible.** The 503 response includes `Retry-After: 1`, so
  well-behaved clients that retry on 5xx will recover automatically. Clients
  that mistakenly retry on 401 will no longer see spurious auth failures.
- **Observability.** `_touch_last_used_at` logs:
  - `api_key_last_used_at_updated` on success (with `key_id`),
  - `api_key_last_used_at_skipped_recent_or_raced` when the conditional
    UPDATE matched zero rows (this is the expected outcome under burst),
  - `api_key_last_used_at_update_failed` (WARNING) on background DB errors.
- **Test-only helper.** A new `ApiKeyStore.drain_pending_touches()` lets
  tests deterministically wait for the background writes; production code
  never needs to call it.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-13234c9   docker.openhands.dev/openhands/openhands:13234c9
```

### DIFF
--- a/enterprise/server/auth/saas_user_auth.py
+++ b/enterprise/server/auth/saas_user_auth.py
@@ -9,7 +9,6 @@ from keycloak.exceptions import KeycloakConnectionError
 from pydantic import SecretStr
 from server.auth.auth_error import (
     AuthError,
-    BearerTokenError,
     CookieError,
     ExpiredError,
     NoCredentialsError,
@@ -26,6 +25,7 @@ from server.logger import logger
 from server.rate_limit import RateLimiter, create_redis_rate_limiter
 from server.utils.rate_limit_utils import RATE_LIMIT_AUTH_WINDOWS
 from sqlalchemy import delete, select
+from sqlalchemy.exc import SQLAlchemyError
 from storage.api_key_store import ApiKeyStore
 from storage.auth_tokens import AuthTokens
 from storage.database import a_session_maker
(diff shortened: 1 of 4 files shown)


tofarr:
> _This pull request was prepared by an AI agent (OpenHands) on behalf of the OpenHands engineering team, then committed via `git` under the user's GitHub credentials. A human reviewer should verify the customer-facing scenario (concurrent Bearer-token requests against `/api/v1/app-conversations`) before marking this PR ready for review._

PR #15337, opened 2026-07-21
fix(enterprise): bootstrap superadmin when upgraded tenants have none

### LINKED ISSUE #15327: [Bug] OHE: upgraded instances can end up with zero superadmins (no backfill for first-user bootstrap)
## Summary

Upgraded Enterprise instances can end up with **users but zero superadmins**, with no way to bootstrap one through the application. The "first user → superadmin" auto-grant in `create_user()` only fires when the `user` table is empty, so instances that already had users **before** the role feature was introduced never get an initial superadmin, and there is no data migration to backfill one.

## Where the auto-grant lives

`enterprise/storage/user_store.py`, `create_user()` (~lines 103–118):

```python
# First-user → superadmin: if the caller did not specify a
# super ``role_id`` and there are no existing users in the
# database, designate this user as a ``superadmin`` ...
if role_id is None:
    existing_user_count = await session.scalar(
        select(func.count()).select_from(User)
    )
    if existing_user_count == 0:
        if superadmin_role is not None:
            role_id = superadmin_role.id
            ...
```

The trigger requires **`existing_user_count == 0`**.

## Why upgraded instances fall through the crack

- The `role` table + `user.role_id` column were introduced by migration `enterprise/migrations/versions/089_create_org_tables.py`. Its seed only populates the `role` table and assigns the role to **no** user:
  ```sql
  INSERT INTO role (name, rank) VALUES ('owner', 10), ('admin', 20), ('user', 1000)
  ```
- On any instance that already had users before migration 089 ran, the `user` table is non-empty by the time the bootstrap code can execute, so the `existing_user_count == 0` branch **can never fire again**.
- There is no data migration that backfills `role_id` for the first/oldest user, and no startup/env bootstrap. The only other grant path, `POST /api/admin/super-admins`, is gated by the `MANAGE_SUPER_ADMINS` permission, which itself requires an **existing** superadmin — so it cannot bootstrap the first one.

**Net effect:** an upgraded instance can have active users but zero superadmins, and no in-app path to create one. Recovery currently requires a manual DB write (`UPDATE "user" SET role_id = (SELECT id FROM role WHERE name='admin') WHERE email = ...`).

## Observed on a real instance

An Enterprise instance running `enterprise-server:cloud-1.46.2` had 6 users, all with `role_id = NULL`, and `SELECT ... JOIN role r ON r.id = u.role_id WHERE r.name = 'admin'` returned 0 rows. The running `/app/storage/user_store.py` contained the bootstrap logic above, confirming the code is present but was never triggered because the users predate migration 089. (The `user` table on that instance also lacked a `created_at` column, consistent with an older schema baseline that was upgraded forward.)

## Proposed fix

Add a data migration (idempotent) that backfills a superadmin when none exists, e.g.:

- If there are zero users whose `role_id` references the `admin` role, promote a deterministic user (e.g. the earliest-created / lowest-ordered user, or a configurable email) by setting `user.role_id` to the `admin` role id.
- No-op if a superadmin already exists (safe to re-run).

Alternatives / complements to consider:
- A guarded startup reconciliation that grants superadmin if none exists.
- An operator CLI/management command to grant the initial superadmin out-of-band (useful for recovery even with the migration in place).

## Impact

- **Severity:** blocks superadmin-only functionality on upgraded self-hosted Enterprise instances (e.g. managing super admins, org provisioning) with no supported recovery path short of direct DB access.
- **Scope:** any Enterprise instance upgraded from a pre-migration-089 baseline.

---
_This issue was created by an AI agent (OpenHands) on behalf of @jpshackelford._

Solaris-star:
## Summary

Upgraded Enterprise instances can end up with users but **zero superadmins**, because `create_user()` only auto-grants the admin super role when the user table is empty (#15327).

### Changes

1. **`UserStore.create_user`** — if no superadmin exists yet, promote the user being created (covers both fresh installs and upgraded tenants with pre-role users).
2. **Alembic `138`** — one-shot data migration: when the `admin` role exists and nobody holds it, promote the earliest existing user (`first_login_at`, then `email`, then `id`).
3. **Unit test** for the zero-superadmin bootstrap path.

## Testing

```text
# targeted (enterprise unit suite; local env may need full deps)
pytest enterprise/tests/unit/test_user_store.py \
  -k "first_user_is_designated or subsequent_users or zero_superadmins" -q
```

`py_compile` clean on changed Python files.

Fixes #15327

### DIFF
--- a/enterprise/migrations/versions/138_backfill_first_superadmin.py
+++ b/enterprise/migrations/versions/138_backfill_first_superadmin.py
@@ -1,0 +1,62 @@
+"""Backfill a superadmin when upgraded instances have users but none.
+
+Revision ID: 138
+Revises: 137
+Create Date: 2026-07-20
+
+Upgrades that introduced ``user.role_id`` left existing tenants with users
+but zero superadmins (#15327). Promote the earliest user (by first_login_at,
+then email, then id) when the admin role exists and nobody holds it.
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "138"
+down_revision = "137"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+
+    admin_role_id = conn.execute(
+        sa.text("SELECT id FROM role WHERE name = 'admin' LIMIT 1")
+    ).scalar()
+    if admin_role_id is None:
+        return
+
+    superadmin_count = conn.execute(
+        sa.text('SELECT COUNT(*) FROM "user" WHERE role_id = :rid'),
+        {"rid": admin_role_id},
+    ).scalar()
+    if superadmin_count and superadmin_count > 0:
+        return
+
+    user_count = conn.execute(sa.text('SELECT COUNT(*) FROM "user"')).scalar()
+    if not user_count:
+        return
+
+    # Prefer the first-ever login; fall back to email/id for deterministic pick.
+    target = conn.execute(
+        sa.text(
+            'SELECT id FROM "user" '
+            "ORDER BY first_login_at ASC NULLS LAST, email ASC NULLS LAST, id ASC "
+            "LIMIT 1"
+        )
+    ).scalar()
+    if target is None:
+        return
+
+    conn.execute(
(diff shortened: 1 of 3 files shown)


neubig:

Thank you for understanding.

PR #15343, opened 2026-07-21
fix(enterprise): migrate legacy empty tool settings

simonrosenberg:
HUMAN:

Replace legacy empty tool arrays with the default-tool sentinel through a durable database migration, so cloud and self-hosted upgrades can safely restore Agent Profile tool selection.

AGENT:

Rebased onto current main and extended after codebase validation to cover the legacy user-settings source used by lazy user migration.

---

## Why

Before openhands-sdk 1.31.2, `OpenHandsAgentSettings.tools` defaulted to `[]`. Ordinary cloud settings saves serialized that default into persisted settings.

SDK 1.31.2 changed the contract:

- `tools: null` uses the default toolset.
- `tools: []` creates an intentionally bare agent.
- A non-empty list uses that explicit toolset.

The persisted rows had no schema marker or migration. When OpenHands/OpenHands#15228 honored stored tool values during cloud launches, affected users lost terminal, file editor, task tracker, and browser tools. OpenHands/OpenHands#15339 reverted that behavior as containment.

Self-hosted installations may skip releases, so the compatibility boundary belongs in Alembic. The deprecated `user_settings` table must also be migrated because `UserStore.migrate_user` can still copy its `agent_settings` into new org and member rows after this revision has run.

## Summary

- Add Alembic revision 141, following current revision 140.
- Migrate exact top-level `tools: []` values to `tools: null` in:
  - `user_settings.agent_settings`
  - `org.agent_settings`
  - `org_member.agent_settings_diff`
- Migrate both migrated and not-yet-migrated legacy user rows.
- Preserve every other setting in each JSON document.
- Leave missing, null, and non-empty tool values unchanged.
- Leave `org.agent_profiles` untouched, so profile-level `tools: []` remains an intentional bare-agent configuration.
- Verify that a normalized legacy user remains on default-tool semantics through the complete lazy `UserStore.migrate_user` path.

This migration does not re-land OpenHands/OpenHands#15228. It establishes the data precondition needed to reintroduce those launch features in a separate follow-up.

## Issue Number

Closes OpenHands/enterprise#68.

## How to Test

```bash
PYTHONPATH=".:${PYTHONPATH}" poetry run --project=enterprise pytest -q \
  -p no:ddtrace -p no:ddtrace.pytest_bdd -p no:ddtrace.pytest_benchmark \
  enterprise/tests/unit/test_migration_*.py \
  tests/unit/test_enterprise_migration_integrity.py
```

Result: `23 passed`.

```bash
PYTHONPATH=".:${PYTHONPATH}" poetry run --project=enterprise pytest -q \
  -p no:ddtrace -p no:ddtrace.pytest_bdd -p no:ddtrace.pytest_benchmark \
  enterprise/tests/unit/test_user_store.py
```

Result: `70 passed` with pre-existing aiosqlite event-loop shutdown warnings.

```bash
poetry run --project=enterprise pre-commit run \
  --config enterprise/dev_config/python/.pre-commit-config.yaml \
  --show-diff-on-failure --files \
  enterprise/migrations/versions/141_normalize_legacy_empty_tools.py \
  enterprise/tests/unit/test_migration_141_normalize_legacy_empty_tools.py \
  enterprise/tests/unit/test_user_store.py
```

Result: all hooks passed, including Ruff, formatting, and mypy.

```bash
poetry run python scripts/check_enterprise_migration_integrity.py
```

Result: `Enterprise migration integrity checks passed.`

The migration was also run directly and twice against PostgreSQL 16 with pg8000. The second run made no further changes. Affected rows in all three tables became JSON null, unrelated JSON was preserved, and an Agent Profile containing `tools: []` remained unchanged.

For staging and production validation, record these counts before and after upgrading; all three should be zero afterward:

```sql
SELECT count(*) FROM user_settings
WHERE agent_settings::jsonb -> 'tools' = '[]'::jsonb;

SELECT count(*) FROM org
WHERE agent_settings::jsonb -> 'tools' = '[]'::jsonb;

SELECT count(*) FROM org_member
WHERE agent_settings_diff::jsonb -> 'tools' = '[]'::jsonb;
```

Then reapply the OpenHands/OpenHands#15228 behavior on a staging build and launch conversations from a default profile, an intentionally bare profile, and a profile with an explicit tool list.

## Video/Screenshots

Not applicable; this is a backend data migration.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The downgrade is intentionally a no-op: after upgrade, a migrated null cannot be distinguished from a null that was already legitimate. Converting null values back to empty lists would corrupt user intent.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-9d28790   docker.openhands.dev/openhands/openhands:9d28790
```

### DIFF
--- a/enterprise/migrations/versions/141_normalize_legacy_empty_tools.py
+++ b/enterprise/migrations/versions/141_normalize_legacy_empty_tools.py
@@ -1,0 +1,79 @@
+"""Normalize legacy empty tool settings."""
+
+from typing import Any
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = '141'
+down_revision = '140'
+branch_labels = None
+depends_on = None
+
+
+def _normalize_empty_tools(settings: Any) -> dict[str, Any] | None:
+    if not isinstance(settings, dict) or settings.get('tools') != []:
+        return None
+    normalized = dict(settings)
+    normalized['tools'] = None
+    return normalized
+
+
+def upgrade() -> None:
+    bind = op.get_bind()
+
+    user_settings = sa.table(
+        'user_settings',
+        sa.column('id', sa.Integer()),
+        sa.column('agent_settings', sa.JSON()),
+    )
+    for row in bind.execute(
+        sa.select(user_settings.c.id, user_settings.c.agent_settings)
+    ).mappings():
+        normalized = _normalize_empty_tools(row['agent_settings'])
+        if normalized is not None:
+            bind.execute(
+                user_settings.update()
+                .where(user_settings.c.id == row['id'])
+                .values(agent_settings=normalized)
+            )
+
+    org = sa.table(
+        'org',
+        sa.column('id', sa.Uuid()),
+        sa.column('agent_settings', sa.JSON()),
+    )
+    for row in bind.execute(sa.select(org.c.id, org.c.agent_settings)).mappings():
+        normalized = _normalize_empty_tools(row['agent_settings'])
+        if normalized is not None:
+            bind.execute(
+                org.update()
+                .where(org.c.id == row['id'])
+                .values(agent_settings=normalized)
+            )
+
+    org_member = sa.table(
+        'org_member',
+        sa.column('org_id', sa.Uuid()),
+        sa.column('user_id', sa.Uuid()),
+        sa.column('agent_settings_diff', sa.JSON()),
+    )
+    for row in bind.execute(
+        sa.select(
+            org_member.c.org_id,
+            org_member.c.user_id,
+            org_member.c.agent_settings_diff,
+        )
+    ).mappings():
+        normalized = _normalize_empty_tools(row['agent_settings_diff'])
+        if normalized is not None:
+            bind.execute(
+                org_member.update()
+                .where(org_member.c.org_id == row['org_id'])
+                .where(org_member.c.user_id == row['user_id'])
+                .values(agent_settings_diff=normalized)
+            )
+
+
+def downgrade() -> None:
+    pass
--- a/enterprise/storage/agent_settings_compat.py
+++ b/enterprise/storage/agent_settings_compat.py
@@ -1,8 +1,0 @@
-from typing import Any
-
-
-def normalize_legacy_empty_tools(settings: dict[str, Any]) -> dict[str, Any]:
-    normalized = dict(settings)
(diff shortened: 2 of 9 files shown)


jlav:
I think data migrations like this need to be handled via real alembic database migrations, rather than just being done opportunistically.

Self hosted installers don't always cleanly jump from one version to the next to the next like we do in our cloud environments. They're likely to skip versions and just upgrade to the latest at their convenience. I'm not confident that this code will always be kept in place as further updates to the agent SDK are made.

simonrosenberg:
Agreed. I rewrote the PR around Alembic revision 138 and removed the opportunistic storage shim entirely. The migration changes only exact top-level `tools: []` values in `org.agent_settings` and `org_member.agent_settings_diff` to JSON null; missing/null/non-empty values and `org.agent_profiles` remain untouched. The migration suite (20 tests), integrity check, Ruff, formatting, and mypy all pass.

simonrosenberg:
Additional functional migration validation: I ran revision 138 directly against a seeded PostgreSQL 16 database (pg8000), with affected empty arrays plus null, missing, non-empty, unrelated nested settings, and a bare Agent Profile control. Only the two legacy top-level empty arrays changed to JSON null; unrelated JSON and the Agent Profile remained equal at the decoded JSON level. A second direct run made no further changes. The repository CI also passed the full Alembic chain and downgrade/upgrade round trip with both supported PostgreSQL drivers.

simonrosenberg:
Rebased onto current `main` as revision 141, added `user_settings.agent_settings` coverage plus the lazy `UserStore.migrate_user` regression test, and updated the three-table audit queries. All checks are now green, including both PostgreSQL migration drivers. The launch-behavior follow-up is prepared as draft OpenHands/OpenHands#15400 and is explicitly blocked on this migration being released and audited.

simonrosenberg:
Superseded by OpenHands/enterprise#12 after the Agent Canvas repository cutover. The replacement preserves commit 9d28790dfa02210c09a670898b9b404a77e1472f and its validated migration scope.

PR #15349, opened 2026-07-22
fix: upgraded instances with no superadmin

### LINKED ISSUE #15327: [Bug] OHE: upgraded instances can end up with zero superadmins (no backfill for first-user bootstrap)

tofarr:
HUMAN:

I've tested this - if there is no admin user, the oldest is given that power -  on prem customers shoul

AGENT:

Implemented and tested by OpenHands on behalf of the user.

---

## Why

Enterprise instances upgraded from before the role migration can contain users without any superadmin. The first-user bootstrap cannot repair those instances because their user table is already populated, leaving superadmin-only functions inaccessible without a manual database update.

## Summary

- Add an idempotent data migration that grants the `admin` role when no superadmin exists.
- Select the oldest user by the earliest available login or terms-acceptance timestamp, with user ID as a deterministic fallback.
- Cover upgrades with no superadmin, an existing superadmin, and no users.

## Issue Number

OHE-2791
Closes #15327

## How to Test

```bash
PYTHONPATH=enterprise poetry run pytest enterprise/tests/unit/test_migration_138_backfill_initial_superadmin.py -q
poetry run pre-commit run --config enterprise/dev_config/python/.pre-commit-config.yaml --files enterprise/migrations/versions/138_backfill_initial_superadmin.py enterprise/tests/unit/test_migration_138_backfill_initial_superadmin.py --show-diff-on-failure
```

Expected: 3 tests pass and all pre-commit checks pass.

## Video/Screenshots

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The `user` table has no creation timestamp. The migration uses `COALESCE(first_login_at, accepted_tos)` as the best available age signal and user ID for stable ordering when both timestamps are absent. Downgrade intentionally does not revoke the promoted role because it cannot distinguish migration-created grants from later operator decisions.

_This pull request was created by an AI agent (OpenHands) on behalf of the user._

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-0c60473   docker.openhands.dev/openhands/openhands:0c60473
```

### DIFF
--- a/enterprise/migrations/versions/138_backfill_initial_superadmin.py
+++ b/enterprise/migrations/versions/138_backfill_initial_superadmin.py
@@ -1,0 +1,49 @@
+"""Ensure upgraded instances have an initial superadmin.
+
+Revision ID: 138
+Revises: 137
+Create Date: 2026-07-21 00:00:00.000000
+"""
+
+from typing import Sequence, Union
+
+import sqlalchemy as sa
+from alembic import op
+
+revision: str = '138'
+down_revision: Union[str, None] = '137'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    connection = op.get_bind()
+    metadata = sa.MetaData()
+    role = sa.Table('role', metadata, autoload_with=connection)
+    user = sa.Table('user', metadata, autoload_with=connection)
+
+    admin_role_id = sa.select(role.c.id).where(role.c.name == 'admin').scalar_subquery()
+    has_superadmin = sa.exists(
+        sa.select(user.c.id).where(user.c.role_id == admin_role_id)
+    )
+    oldest_user_id = (
+        sa.select(user.c.id)
+        .order_by(
+            sa.func.coalesce(user.c.first_login_at, user.c.accepted_tos)
+            .asc()
+            .nulls_last(),
+            user.c.id.asc(),
+        )
+        .limit(1)
+        .scalar_subquery()
+    )
+
+    connection.execute(
+        sa.update(user)
+        .where(user.c.id == oldest_user_id, ~has_superadmin)
+        .values(role_id=admin_role_id)
+    )
+
+
+def downgrade() -> None:
+    pass
--- a/enterprise/tests/unit/test_migration_138_backfill_initial_superadmin.py
+++ b/enterprise/tests/unit/test_migration_138_backfill_initial_superadmin.py
@@ -1,0 +1,117 @@
+from datetime import datetime
+from importlib.util import module_from_spec, spec_from_file_location
+from pathlib import Path
+
+import sqlalchemy as sa
+from alembic.migration import MigrationContext
+from alembic.operations import Operations
+
+MIGRATION_PATH = (
+    Path(__file__).resolve().parents[2]
+    / 'migrations'
+    / 'versions'
+    / '138_backfill_initial_superadmin.py'
+)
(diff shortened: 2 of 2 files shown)


hieptl (approved):
Thank you! 🙏

PR #15354, opened 2026-07-22
fix(app-server): persist combined LLM usage metrics across all usage buckets

### LINKED ISSUE #15372: Conversation statistics silently dropped when conversation org differs from owner's current org
Found during multi-user live validation of #15354/#15355/#15369 (pre-existing — same lookup exists on main before those PRs).

## Mechanism

The stats persistence path (`update_conversation_statistics`, reached from stats webhooks and the run-end pull) looks the conversation up through the SaaS `_secure_select`, which filters by the **effective org** (`enterprise/server/utils/saas_app_conversation_info_injector.py`, `_apply_user_and_org_filter` → `_get_effective_org_id`). For webhook-internal user contexts there is no request `X-Org-Id`, so it falls back to `user.current_org_id`. If the conversation's saas row belongs to a different org than the owner's current org, the filter excludes it, `stored` is None, and the update is skipped **at DEBUG log level** — cost/token columns stay 0 and no cost-event ledger rows are written, while runtime stats and LiteLLM billing are correct.

## Reproduced live

Two freshly provisioned users whose `current_org_id` was their personal org ran conversations in a shared team org (created via `X-Org-Id` header): runtime showed real cost ($0.08–$0.17 per conversation), DB persisted 0.0 with zero ledger rows. After `POST /api/organizations/{org}/switch` for each user, the very next stats event healed every conversation to its full cumulative totals — confirming org-context mismatch as the cause.

## Real-world triggers

- A user switches their active org while a conversation in the previous org is still running (all subsequent stats for that conversation are dropped until they switch back).
- Conversations created under `X-Org-Id` / org-bound API keys in a non-current org.

## Suggested fix direction

Stats writers are system paths keyed by conversation identity: scope the lookup by conversation_id + owner (or run the webhook path under the admin context like other webhook internals) instead of the viewer-style effective-org filter — or resolve the org from the conversation's own saas row. Also consider raising the skip log above DEBUG; the silence made this invisible.

ak684:

HUMAN: Root-caused from a customer report (on-prem install, cloud-1.45+). Reproduced live on a Replicated test install before the fix and validated on OHE after the final stacked changes, including the org-switch webhook path described below.

## Problem

`update_conversation_statistics` read only `usage_to_metrics['agent']`. A switched-in LLM registers under `profile:*` (and ACP agents under `acp-managed`), so all post-switch/ACP spend was dropped: `conversation_metadata.accumulated_cost` froze at the pre-switch value, and every DB-backed surface (Budget org + per-user spend, Usage estimated spend, Users-tab MTD/YTD) under-reported, while LiteLLM and the live conversation display stayed correct.

Additionally:

- Switched-in LLMs never attach the stats-streaming telemetry callback, so their spend emits no stats events; a reader-side fix alone cannot see it.
- Webhook writers looked conversations up through the owner's effective/current organization. If the owner switched organizations while a run was active, otherwise-authenticated statistics, execution-status, and conversation-info writes were silently skipped.

## Changes

- Combine cost and token usage across every usage bucket when persisting stats, with a monotonic guard so stale or out-of-order snapshots cannot regress stored totals.
- Pull the registry's combined stats from agent-server when the events webhook reports a run-end status (finished/idle/error/stuck), closing the no-stats-events gap for switched LLMs and ACP.
- Serialize statistics/ledger updates per conversation and normalize local Docker sandbox URLs before the run-end pull.
- Scope authenticated webhook persistence to the sandbox owner plus sandbox ID instead of the owner's currently selected organization. Conversation resolution uses an admin lookup followed by explicit owner+sandbox binding checks, and full-row saves verify the same binding.
- Declare `update_conversation_statistics` on the service ABC.

## Security

The org-independent webhook lookup does not broaden caller access: the session API key first authenticates a specific sandbox, and writes are accepted only when both the conversation owner and `sandbox_id` match that authenticated sandbox. Mismatched owner or sandbox bindings are rejected.

### Combined usage (Replicated embedded-cluster, 2026-07-22)

Deployed together with stacked #15355. Scenario: turn 1 on `litellm_proxy/gpt-5.5`, mid-conversation `switch_profile` to `litellm_proxy/claude-opus-4-8`, then turn 2.

| Surface | Before fix | After fix |
|---|---|---|
| Runtime buckets (truth) | agent $0.0824 + profile $0.1741 | agent $0.08377 + profile $0.17253 |
| DB `accumulated_cost` | **frozen at $0.0824** (pre-switch value) | **$0.25629875 = exact combined total** ✅ |
| Budget user/org spend | missing all post-switch spend | includes it exactly ✅ |

Also verified pre-fix that no lifecycle webhook fires on pause on this build; without the run-end pull, the frozen value never heals.

### Organization switching (OHE, 2026-07-23)

Validated the final stack with the conversation's organization differing from the owner's currently selected organization. Authenticated webhook statistics continued to persist, while the owner+sandbox binding remained enforced.

## Tests

- Combined-bucket, switch-scenario, ACP, stale/out-of-order snapshot, monotonic ledger, and run-end pull coverage.
- Webhook authorization coverage for matching and mismatched owners/sandboxes, plus org-independent persistence.
- Local-Docker URL normalization and statistics service persistence coverage.

Fixes #15372

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-cabb67f   docker.openhands.dev/openhands/openhands:cabb67f
```


ak684:
**Live validation on a Replicated embedded-cluster install (2026-07-22): PASS** — deployed together with the stacked #15355; the persisted conversation total now equals the runtime's combined buckets after a mid-conversation LLM switch ($0.25629875 = $0.08377 gpt + $0.17253 opus), where the pre-fix build froze at the pre-switch value. Full evidence table on #15355.

lilagrc (approved):

PR #15360, opened 2026-07-22
fix(frontend): sum all usage buckets in the WebSocket metrics fallback

### LINKED ISSUE #15358: Conversation cost display undercounts on the WebSocket fallback path (agent bucket only)
The conversation metrics modal is correct when it fetches live runtime stats (sums every `usage_to_metrics` bucket — `frontend/src/utils/conversation-metrics.ts`), but the WebSocket fallback store reads only `usage_to_metrics.agent` (`frontend/src/contexts/conversation-websocket-context.tsx`, `updateMetricsFromStats`).

On the fallback path the modal therefore under-displays cost/tokens for:
- conversations that switched LLMs mid-run (spend accrues under `profile:*` buckets),
- ACP conversations (spend accrues under `acp-managed`),
- condenser/other auxiliary buckets.

Display-only: DB persistence is fixed by #15354/#15355. Fix is to sum all buckets in `updateMetricsFromStats`, mirroring `getCombinedMetrics`.

Related nit, same area: `switch_acp_model`/`switch_profile` rewrite the conversation's single `llm_model` label — harmless for cost accounting now that attribution comes from the cost-event ledger, but worth keeping in mind for label semantics.

ak684:

HUMAN: Display-only companion to #15354/#15355. Closes #15358.

## Problem

The conversation metrics modal is correct when it fetches live runtime stats (sums every `usage_to_metrics` bucket), but the WebSocket stats-event fallback read only `usage_to_metrics.agent` — under-displaying cost/tokens for switched-in LLMs (`profile:*`), ACP (`acp-managed`), and the condenser.

## Changes

- Extract the bucket-combining loop from `getCombinedMetrics` into `combineUsageMetrics` and use it in the WebSocket fallback.
- Widen `UsageToMetrics` to its real shape (dynamic usage-id keys), which the old two-key type was hiding.

## Tests

- Unit tests for `combineUsageMetrics` (multi-bucket sum incl. `profile:*`/`acp-managed`, max-budget selection). Full frontend lint pipeline passes.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-3afb7fa   docker.openhands.dev/openhands/openhands:3afb7fa
```

### DIFF
--- a/frontend/__tests__/utils/conversation-metrics.test.ts
+++ b/frontend/__tests__/utils/conversation-metrics.test.ts
@@ -1,0 +1,70 @@
+import { describe, expect, it } from "vitest";
+
+import { combineUsageMetrics } from "#/utils/conversation-metrics";
+
+const usage = (
+  prompt: number,
+  completion: number,
+  overrides: Partial<{
+    cache_read_tokens: number;
+    cache_write_tokens: number;
+    context_window: number;
+    per_turn_token: number;
+  }> = {},
+) => ({
+  prompt_tokens: prompt,
+  completion_tokens: completion,
+  cache_read_tokens: 0,
+  cache_write_tokens: 0,
+  context_window: 0,
+  per_turn_token: 0,
+  ...overrides,
+});
+
+describe("combineUsageMetrics", () => {
+  // A mid-conversation LLM switch accrues under "profile:*" and ACP under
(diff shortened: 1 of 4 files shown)


neubig:

Thank you for understanding.

PR #15362, opened 2026-07-22
fix(app-server): write titles column-specifically to avoid clobbering concurrent updates

ak684:

HUMAN: Part of OpenHands/enterprise#69 (fixes the primary offender; the broader read-modify-save hardening stays tracked there).

## Problem

`SetTitleCallbackProcessor` fetched the conversation, polled the agent-server for a generated title (up to ~12s), then saved the **entire stale snapshot** via `save_app_conversation_info` — a lost-update race that silently erased any stats/metrics/metadata written during the poll window.

## Changes

- New `update_title` on the conversation-info service (column-specific write, mirroring `update_execution_status`); declared on the ABC.
- The title processor calls it instead of rebuilding and re-saving the snapshot.

## Tests

- Service-level: `update_title` changes only the title (metrics/model untouched); missing-conversation no-op.
- Processor tests updated to assert the column-specific write and that the stale snapshot is never re-saved.

---

GUI with Docker:
```
docker run -it --rm   -p 3000:3000   -v /var/run/docker.sock:/var/run/docker.sock   --add-host host.docker.internal:host-gateway   --name openhands-app-942e98f   docker.openhands.dev/openhands/openhands:942e98f
```

### DIFF
--- a/openhands/app_server/app_conversation/app_conversation_info_service.py
+++ b/openhands/app_server/app_conversation/app_conversation_info_service.py
@@ -117,6 +117,14 @@ class AppConversationInfoService(ABC):
         """
 
     @abstractmethod
+    async def update_title(
+        self,
+        conversation_id: UUID,
+        title: str,
+    ) -> None:
+        """Update only the title for a conversation (no full-row rewrite)."""
+
+    @abstractmethod
     async def update_execution_status(
         self,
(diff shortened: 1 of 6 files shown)


neubig:

Thank you for understanding.

PR #15287, opened 2026-07-15
fix(acp): synchronize managed Codex credentials across runtimes

simonrosenberg:
Grounding issue: https://github.com/OpenHands/software-agent-sdk/issues/4170
Implementation plan: https://github.com/OpenHands/software-agent-sdk/issues/4177
Companion SDK implementation (merged): https://github.com/OpenHands/software-agent-sdk/pull/4124
SDK readiness and cold-load guard follow-up: https://github.com/OpenHands/software-agent-sdk/pull/4198

## Why

Codex rotates ChatGPT subscription credentials in its writable `auth.json`. App-server-managed runtimes previously received a static snapshot, so a later isolated conversation could start with an invalidated refresh token.

Ordinary secrets are static values. This flow needs one narrow exception: load a runtime-mutable credential with a version, then replace it atomically after the runtime updates it.

## What and why

| Feature | Why it is necessary |
| --- | --- |
| Versioned compare-and-swap in existing secret stores | Codex rotates `auth.json`; the latest value must reach later runtimes without letting stale writers overwrite it or adding another credential store. |
| Codex-only managed-credential policy | Keeps the exception narrow and avoids a generic broker, catalog, or provisioning system. |
| Scoped, liveness-checked callback bearer | The runtime needs canonical `load`/`replace` access while active, but pause, deletion, remapping, or task replacement must revoke it. |
| Exact capability gate plus SDK cold-load guard | Preserves compatibility with older images while preventing REST or WebSocket clients from hydrating a managed conversation before reactivation. |
| Server-owned provenance and one-off exclusion | Resume must distinguish canonical credentials from API/context overrides without letting runtime tags forge canonical authority or force a downgrade. |
| Dedicated cold resume path | Avoids rerunning setup, profile resolution, or conversation creation; an exact cold GET proves the stored conversation exists before activation. |
| Sandbox-wide task rotation and reactivation | Pause clears every ephemeral binding, and grouped conversations share a sandbox; all managed conversations need fresh bearers before wake-up. |
| Strict, bounded pause barrier and managed raw-resume rejection | Codex must flush its final rotation before pause, and alternate resume routes must not bypass bearer rotation or activation. |
| Frontend polling, connection gating, and retry | Prevents an early WebSocket/session race, surfaces activation failures, and permits retry after the sandbox is already physically running. |
| SaaS, OSS, and Docker topology adapters | The same lifecycle must work across supported deployments, including atomic SQL/file updates and reachable Docker callback URLs. |

[...]

Issue #15, opened 2024-03-15
Extensibility


PallozziRiccardo:
Could it be possible to attach to custom knowledge sources? (Like existing vector DB, graph DB, relational DB and other stuff?)

z-zeechung:
autogpt has a knowledge base, so i guess we'll be the same

PallozziRiccardo:
I would suggest to add some capabilities and extend to external pluggable knowledge

ryanmoriarty-lee:
Considering real-life development scenarios, we should have a workspace consisting of a ready-made or blank code repository and a portion of design and requirement documents. If we want OpenDevin to be practically available, rather than just staying in the demo stage, this requirement is very important and reasonable.

neubig:
This sounds like a good idea, but TBH I think it is a bit too underspecified. Because of this, I'm going to convert it to a discussion. We can discuss the different knowledge sources that we would like to incorporate, and then create concrete issues for specific knowledge sources.

PR #14, opened 2024-03-15
created api for llms, created README_ZH.md

z-zeechung:

Issue #13, opened 2024-03-15
python or golang?


z-zeechung:

xandox:
rust may be?

taylorjdawson:
Anything but Python, maybe even multilingual??

z-zeechung:
then the package would be horribly huge😰

rbren:
I far prefer golang, but the ecosystem is in python.

Leveraging LangChain, LlamaIndex, MetaGPT, etc will be much easier with python. Choosing anything else would be a mistake.

suryavirkapur:
i agree with @rbren choosing anything else would be a mistake.

neubig:
I think we're planning on ts frontend py backend

PR #15394, opened 2026-07-26
feat(app-server): sync managed Codex credentials (Phase 1)

simonrosenberg:
HUMAN:

This replaces OpenHands/OpenHands#15287 with a narrower Phase 1: OpenHands stores and activates managed Codex credentials, while the pinned SDK owns private-file handling, refresh, merge, and scrubbing.

AGENT:

Closes OpenHands/enterprise#70.

Supersedes OpenHands/OpenHands#15287 with the narrower Phase 1 vertical slice described in the issue. The pinned SDK/Agent Server owns the private Codex credential lifecycle; OpenHands only supplies a versioned canonical store, scoped activation, and lifecycle ordering.

## What changed

- Add opaque, ABA-safe compare-and-swap versions for `CODEX_AUTH_JSON` in SaaS SQL and supported OSS stores.
- Preserve a runtime rotation when an unrelated stale whole-secret document is saved.
- Reuse the sandbox session key plus one scoped JWE callback for credential load/replace.
- Arm only for saved ChatGPT credentials, authoritative Codex ACP settings, SaaS remote or OSS Docker, and all three exact SDK 1.37.1 capabilities.
- Remove the managed plaintext credential from the Agent Server start request after readiness succeeds.
- Bind initial activation and failed-create cleanup to the exact internal start-task tag.
- Persist one non-secret managed marker, add one idempotent conversation-scoped reactivation endpoint, and gate WebSocket reconnect on recovery.
- Run the SDK drain barrier before clean remote/Docker pause or delete, and propagate managed final-flush failures.

## Complexity boundary

This is 19 production files and roughly 1.5k added production lines, versus 83 files and more than 10k added lines in OpenHands/OpenHands#15287. It deliberately has no reservation/lease rows, migrations, schedulers, sibling scans, resume-task claiming, generic credential broker, pending-message changes, or duplicated SDK monitor/merge logic.

OSS process sandboxes remain legacy. Docker retains its existing single-user container session key and receives a fresh scoped JWE on activation; Phase 1 does not claim pre-pause JWE rejection for Docker. SaaS remote rotates the runtime session key before reactivation.

## Validation

- 427 focused app-server lifecycle, callback, store, webhook, and sandbox tests passed.
- 13 SaaS secret-store tests passed.
- 53 frontend recovery/WebSocket tests passed (1 skipped, 13 todo).
- Full frontend TypeScript check, source ESLint, Prettier, Ruff, repository pre-commit mypy, and diff checks passed.
- Independent contract, frontend, and security reviews found no remaining blockers.

End-to-end validation built the unmodified pinned SDK/Agent Server 1.37.1 release (`99342c4`) into `openhands-agent-server-e2e:99342c4-1.37.1-python-source-minimal` and used a copied local ChatGPT credential without printing it:

1. Started managed runtime A and completed a real Codex API request.
2. Forced a credential refresh and verified the canonical R0 -> R1 update.
3. Raced an unrelated stale settings edit and verified both R1 and the unrelated edit survived.
4. Cleanly paused and reactivated A, then completed a second real Codex request.
5. Started isolated runtime B from the canonical value and completed a real Codex request.
6. Verified B's private credential matched the canonical store, the legacy auth path was absent, no credential/session value appeared outside the canonical secret store, and the private file was scrubbed before pause.

### DIFF
--- a/enterprise/storage/saas_secrets_store.py
+++ b/enterprise/storage/saas_secrets_store.py
@@ -1,17 +1,30 @@
 from __future__ import annotations
 
+import hashlib
+import hmac
 from dataclasses import dataclass, field
 from uuid import UUID
 
 from sqlalchemy import delete, select
 from storage.database import a_session_maker
+from storage.org_member_store import OrgMemberStore
 from storage.stored_custom_secrets import StoredCustomSecrets
 from storage.user_store import UserStore
 
 from openhands.app_server.secrets.secrets_models import Secrets
-from openhands.app_server.secrets.secrets_store import SecretsStore
+from openhands.app_server.secrets.secrets_store import (
+    CredentialVersionConflict,
+    SecretsStore,
+)
 from openhands.app_server.services.jwt_service import JwtService
 from openhands.app_server.utils.logger import openhands_logger as logger
+
+_CODEX_AUTH_SECRET_NAME = "CODEX_AUTH_JSON"
+
+
+def _credential_version(row: StoredCustomSecrets) -> str:
+    state = f"{row.id}\0{row.secret_value}".encode()
+    return hashlib.sha256(state).hexdigest()
 
 
 @dataclass
@@ -24,6 +37,11 @@ class SaasSecretsStore(SecretsStore):
     # (user_id, org_id), so the effective org must flow through here for
     # the right rows to be read/written.
     effective_org_id: UUID | None = None
+    _loaded_codex_auth: dict[UUID | None, str | None] = field(
+        default_factory=dict,
+        init=False,
+        repr=False,
+    )
 
     async def load(self) -> Secrets | None:
         if not self.user_id:
@@ -32,64 +50,103 @@ class SaasSecretsStore(SecretsStore):
         org_id = self.effective_org_id or (user.current_org_id if user else None)
 
         async with a_session_maker() as session:
-            # Fetch all secrets for the given user ID
             query = select(StoredCustomSecrets).filter(
                 StoredCustomSecrets.keycloak_user_id == self.user_id
             )
             if org_id is not None:
                 query = query.filter(StoredCustomSecrets.org_id == org_id)
-            result = await session.execute(query)
+            result = await session.execute(query.order_by(StoredCustomSecrets.id))
             settings = result.scalars().all()
 
             if not settings:
+                self._loaded_codex_auth[org_id] = None
                 return Secrets()
 
             kwargs = {}
             for secret in settings:
                 kwargs[secret.secret_name] = {
-                    'secret': secret.secret_value,
-                    'description': secret.description,
+                    "secret": secret.secret_value,
+                    "description": secret.description,
                 }
 
             self._decrypt_kwargs(kwargs)
+            codex_info = kwargs.get(_CODEX_AUTH_SECRET_NAME)
+            codex_value = (
+                codex_info.get("secret") if isinstance(codex_info, dict) else None
+            )
+            self._loaded_codex_auth[org_id] = (
+                codex_value if isinstance(codex_value, str) else None
+            )
 
             return Secrets(custom_secrets=kwargs)  # type: ignore[arg-type]
 
     async def store(self, item: Secrets):
         user = await UserStore.get_user_by_id(self.user_id)
         if user is None:
-            raise ValueError(f'User not found: {self.user_id}')
+            raise ValueError(f"User not found: {self.user_id}")
         org_id = self.effective_org_id or user.current_org_id
 
+        kwargs = item.model_dump(context={"expose_secrets": True})
+        del kwargs["provider_tokens"]
+        secrets_json = kwargs.get("custom_secrets", {})
+        codex_info = secrets_json.get(_CODEX_AUTH_SECRET_NAME)
+        submitted_codex = (
+            codex_info.get("secret") if isinstance(codex_info, dict) else None
+        )
+        if not isinstance(submitted_codex, str):
+            submitted_codex = None
+
+        has_baseline = org_id in self._loaded_codex_auth
+        preserve_codex = not has_baseline and codex_info is None
+        if has_baseline and submitted_codex == self._loaded_codex_auth[org_id]:
+            preserve_codex = True
+
         async with a_session_maker() as session:
-            # Incoming secrets are always the most updated ones
-            # Delete existing records for this user AND organization only
-            # org_id is always set: it's either the effective org from
-            # the request or the user's non-nullable current_org_id.
+            result = await session.execute(
+                select(StoredCustomSecrets)
+                .filter(
+                    StoredCustomSecrets.keycloak_user_id == self.user_id,
+                    StoredCustomSecrets.org_id == org_id,
+                )
+                .order_by(StoredCustomSecrets.id)
+                .with_for_update()
+            )
+            codex_rows = [
+                row
+                for row in result.scalars().all()
+                if row.secret_name == _CODEX_AUTH_SECRET_NAME
+            ]
+            if preserve_codex:
+                secrets_json.pop(_CODEX_AUTH_SECRET_NAME, None)
+                if codex_rows and isinstance(codex_info, dict):
+                    description = codex_info.get("description")
+                    encrypted_description = (
+                        self._jwt_svc.encrypt_value(description)
+                        if description is not None
+                        else None
+                    )
+                    for row in codex_rows:
+                        row.description = encrypted_description
+
             delete_query = delete(StoredCustomSecrets).filter(
                 StoredCustomSecrets.keycloak_user_id == self.user_id,
                 StoredCustomSecrets.org_id == org_id,
             )
+            if preserve_codex:
+                delete_query = delete_query.filter(
+                    StoredCustomSecrets.secret_name != _CODEX_AUTH_SECRET_NAME
+                )
             await session.execute(delete_query)
 
-            # Prepare the new secrets data
-            kwargs = item.model_dump(context={'expose_secrets': True})
-            del kwargs[
-                'provider_tokens'
-            ]  # Assuming provider_tokens is not part of custom_secrets
             self._encrypt_kwargs(kwargs)
 
-            secrets_json = kwargs.get('custom_secrets', {})
-
-            # Extract the secrets into tuples for insertion or updating
             secret_tuples = []
             for secret_name, secret_info in secrets_json.items():
-                secret_value = secret_info.get('secret')
-                description = secret_info.get('description')
+                secret_value = secret_info.get("secret")
+                description = secret_info.get("description")
 
                 secret_tuples.append((secret_name, secret_value, description))
 
-            # Add the new secrets
             for secret_name, secret_value, description in secret_tuples:
                 new_secret = StoredCustomSecrets(
                     keycloak_user_id=self.user_id,
@@ -101,6 +158,88 @@ class SaasSecretsStore(SecretsStore):
                 session.add(new_secret)
 
             await session.commit()
+            if not preserve_codex:
+                self._loaded_codex_auth[org_id] = submitted_codex
+
+    async def load_versioned(
+        self,
+        name: str,
+        organization_id: UUID | None = None,
+    ) -> tuple[str, str]:
+        org_id = await self._require_organization_id(name, organization_id)
+        async with a_session_maker() as session:
+            result = await session.execute(self._versioned_query(name, org_id).limit(1))
+            row = result.scalars().first()
+            if row is None:
+                raise KeyError(name)
+            return self._jwt_svc.decrypt_value(row.secret_value), _credential_version(
+                row
+            )
+
+    async def replace_versioned(
+        self,
+        name: str,
+        expected_version: str,
+        value: str,
+        organization_id: UUID | None = None,
+    ) -> str:
+        org_id = await self._require_organization_id(name, organization_id)
+        async with a_session_maker() as session:
+            result = await session.execute(
+                self._versioned_query(name, org_id).with_for_update()
+            )
+            rows = result.scalars().all()
+            if not rows:
+                result = await session.execute(
+                    self._versioned_query(name, org_id).limit(1)
+                )
+                if result.scalars().first() is not None:
+                    raise CredentialVersionConflict
+                raise KeyError(name)
+
+            current_version = _credential_version(rows[0])
+            if not hmac.compare_digest(
+                current_version.encode(),
+                expected_version.encode(errors="surrogatepass"),
+            ):
+                raise CredentialVersionConflict
+
+            encrypted = self._jwt_svc.encrypt_value(value)
+            for row in rows:
+                row.secret_value = encrypted
+            successor = _credential_version(rows[0])
+            await session.commit()
+            return successor
+
+    async def _require_organization_id(
+        self,
+        name: str,
+        organization_id: UUID | None,
+    ) -> UUID:
+        org_id = organization_id or self.effective_org_id
+        if org_id is None:
+            user = await UserStore.get_user_by_id(self.user_id)
+            org_id = user.current_org_id if user else None
+        if org_id is None:
+            raise KeyError(name)
+        try:
+            user_id = UUID(self.user_id)
+        except (TypeError, ValueError) as exc:
+            raise KeyError(name) from exc
+        if await OrgMemberStore.get_org_member(org_id, user_id) is None:
+            raise KeyError(name)
+        return org_id
+
+    def _versioned_query(self, name: str, organization_id: UUID):
+        return (
+            select(StoredCustomSecrets)
+            .filter(
+                StoredCustomSecrets.keycloak_user_id == self.user_id,
+                StoredCustomSecrets.org_id == organization_id,
+                StoredCustomSecrets.secret_name == name,
+            )
+            .order_by(StoredCustomSecrets.id.desc())
+        )
 
     def _decrypt_kwargs(self, kwargs: dict):
         for key, value in kwargs.items():
@@ -142,7 +281,7 @@ class SaasSecretsStore(SecretsStore):
 
         TODO: This method should be replaced with dependency injection.
         """
-        logger.debug(f'saas_secrets_store.get_instance::{user_id}')
+        logger.debug(f"saas_secrets_store.get_instance::{user_id}")
         from storage.encrypt_utils import get_jwt_service
 
         return SaasSecretsStore(
--- a/enterprise/storage/saas_secrets_store.py
+++ b/enterprise/storage/saas_secrets_store.py
@@ -19,11 +19,11 @@ from openhands.app_server.secrets.secrets_store import (
 from openhands.app_server.services.jwt_service import JwtService
 from openhands.app_server.utils.logger import openhands_logger as logger
 
-_CODEX_AUTH_SECRET_NAME = "CODEX_AUTH_JSON"
+_CODEX_AUTH_SECRET_NAME = 'CODEX_AUTH_JSON'
 
 
 def _credential_version(row: StoredCustomSecrets) -> str:
-    state = f"{row.id}\0{row.secret_value}".encode()
+    state = f'{row.id}\0{row.secret_value}'.encode()
     return hashlib.sha256(state).hexdigest()
 
 
@@ -65,14 +65,14 @@ class SaasSecretsStore(SecretsStore):
             kwargs = {}
             for secret in settings:
                 kwargs[secret.secret_name] = {
-                    "secret": secret.secret_value,
-                    "description": secret.description,
+                    'secret': secret.secret_value,
+                    'description': secret.description,
                 }
 
             self._decrypt_kwargs(kwargs)
             codex_info = kwargs.get(_CODEX_AUTH_SECRET_NAME)
             codex_value = (
-                codex_info.get("secret") if isinstance(codex_info, dict) else None
+                codex_info.get('secret') if isinstance(codex_info, dict) else None
             )
             self._loaded_codex_auth[org_id] = (
                 codex_value if isinstance(codex_value, str) else None
@@ -83,15 +83,15 @@ class SaasSecretsStore(SecretsStore):
     async def store(self, item: Secrets):
         user = await UserStore.get_user_by_id(self.user_id)
         if user is None:
-            raise ValueError(f"User not found: {self.user_id}")
+            raise ValueError(f'User not found: {self.user_id}')
         org_id = self.effective_org_id or user.current_org_id
 
-        kwargs = item.model_dump(context={"expose_secrets": True})
-        del kwargs["provider_tokens"]
-        secrets_json = kwargs.get("custom_secrets", {})
+        kwargs = item.model_dump(context={'expose_secrets': True})
+        del kwargs['provider_tokens']
+        secrets_json = kwargs.get('custom_secrets', {})
         codex_info = secrets_json.get(_CODEX_AUTH_SECRET_NAME)
         submitted_codex = (
-            codex_info.get("secret") if isinstance(codex_info, dict) else None
+            codex_info.get('secret') if isinstance(codex_info, dict) else None
         )
         if not isinstance(submitted_codex, str):
             submitted_codex = None
@@ -119,7 +119,7 @@ class SaasSecretsStore(SecretsStore):
             if preserve_codex:
                 secrets_json.pop(_CODEX_AUTH_SECRET_NAME, None)
                 if codex_rows and isinstance(codex_info, dict):
-                    description = codex_info.get("description")
+                    description = codex_info.get('description')
                     encrypted_description = (
                         self._jwt_svc.encrypt_value(description)
                         if description is not None
@@ -142,8 +142,8 @@ class SaasSecretsStore(SecretsStore):
 
             secret_tuples = []
             for secret_name, secret_info in secrets_json.items():
-                secret_value = secret_info.get("secret")
-                description = secret_info.get("description")
+                secret_value = secret_info.get('secret')
+                description = secret_info.get('description')
 
                 secret_tuples.append((secret_name, secret_value, description))
 
@@ -200,7 +200,7 @@ class SaasSecretsStore(SecretsStore):
             current_version = _credential_version(rows[0])
             if not hmac.compare_digest(
                 current_version.encode(),
-                expected_version.encode(errors="surrogatepass"),
+                expected_version.encode(errors='surrogatepass'),
             ):
                 raise CredentialVersionConflict
 
@@ -281,7 +281,7 @@ class SaasSecretsStore(SecretsStore):
 
         TODO: This method should be replaced with dependency injection.
         """
-        logger.debug(f"saas_secrets_store.get_instance::{user_id}")
+        logger.debug(f'saas_secrets_store.get_instance::{user_id}')
         from storage.encrypt_utils import get_jwt_service
 
         return SaasSecretsStore(
--- a/enterprise/storage/saas_secrets_store.py
+++ b/enterprise/storage/saas_secrets_store.py
@@ -108,7 +108,9 @@ class SaasSecretsStore(SecretsStore):
                     StoredCustomSecrets.keycloak_user_id == self.user_id,
                     StoredCustomSecrets.org_id == org_id,
                 )
-                .order_by(StoredCustomSecrets.id)
+                # Lock in the same order as _versioned_query so a concurrent
+                # replace_versioned over the duplicate Codex rows cannot deadlock.
+                .order_by(StoredCustomSecrets.id.desc())
                 .with_for_update()
             )
             codex_rows = [
--- a/frontend/src/hooks/use-sandbox-recovery.ts
+++ b/frontend/src/hooks/use-sandbox-recovery.ts
@@ -11,7 +11,6 @@ import { V1AppConversation } from "#/api/conversation-service/v1-conversation-se
 interface UseSandboxRecoveryOptions {
   conversationId: string | undefined;
   sandboxStatus: V1SandboxStatus | undefined;
-  /** Function to refetch the conversation data - used to get fresh status on tab focus */
   refetchConversation?: () => Promise<{
     data: V1AppConversation | null | undefined;
   }>;
@@ -19,25 +18,6 @@ interface UseSandboxRecoveryOptions {
   onError?: (error: Error) => void;
 }
 
-/**
- * Hook that handles sandbox recovery based on user intent.
- *
- * Recovery triggers:
- * - Page refresh: Resumes the sandbox on initial load if it was paused/stopped
- * - Tab gains focus: Resumes the sandbox if it was paused/stopped
- *
- * What does NOT trigger recovery:
- * - WebSocket disconnect: Does NOT automatically resume the sandbox
- *   (The server pauses sandboxes after 20 minutes of inactivity,
- *    and sandboxes should only be resumed when the user explicitly shows intent)
- *
- * @param options.conversationId - The conversation ID to recover
- * @param options.conversationStatus - The current conversation status
- * @param options.refetchConversation - Function to refetch conversation data on tab focus
- * @param options.onSuccess - Callback when recovery succeeds
- * @param options.onError - Callback when recovery fails
- * @returns isResuming - Whether a recovery is in progress
- */
 export function useSandboxRecovery({
   conversationId,
   sandboxStatus,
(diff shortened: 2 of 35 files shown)


simonrosenberg:
## Review findings

I found five correctness issues that should be addressed before merge:

1. **[P1] End failed-create cleanup before yielding `READY`** — `openhands/app_server/app_conversation/live_status_app_conversation_service.py:702-706`

   After the cleanup flag is cleared, the task is marked and yielded `READY`, but `_process_pending_messages()` still runs inside the same `try`. If its database update, fetch, or delete fails transiently, the catch catalog-reads the exact start-task tag and deletes the Agent Server conversation even though `READY` was already persisted, leaving local metadata for a conversation it destroyed. Disable failed-create cleanup before yielding `READY`, or isolate post-ready bookkeeping from that catch.

2. **[P2] Restrict managed arming to supported sandbox backends** — `openhands/app_server/app_conversation/live_status_app_conversation_service.py:2394-2396`

   In SaaS mode this condition bypasses the backend check entirely, so a deployment using the supported `RUNTIME=process` configuration can be marked managed. `ProcessSandboxService.pause_sandbox()` and `delete_sandbox()` never run `_prepare_for_pause`, so suspending or terminating such a sandbox can lose a refreshed private Codex credential. Restrict SaaS arming to the remote backend while retaining OSS Docker so process sandboxes stay on the legacy path.

3. **[P2] Reject existing IDs before mutating credential bindings** — `openhands/app_server/app_conversation/live_status_app_conversation_service.py:2435-2440`

   When a caller retries or supplies an ID already present in a grouped or explicitly selected sandbox, this activation occurs before the create response is checked. Agent Server reauthorizes an existing HTTP binding at the same callback URL with the new start-task JWE; the subsequent create returns the existing conversation tags, this task errors, and that JWE becomes unauthorized once the task status changes. The existing managed conversation then cannot refresh or final-flush, so reserve or reject the ID before mutating its binding.

4. **[P2] Remove local records after failed-start cleanup** — `openhands/app_server/app_conversation/live_status_app_conversation_service.py:757-762`

   If the second taskless activation fails after `AppConversationInfo` and callback rows have committed, this `DELETE` can successfully remove the task-owned Agent Server conversation, but the catch never removes those local rows. Searches and counts then expose a marked conversation whose remote state is gone, and it can affect shared-sandbox limits and barriers. After successful managed cleanup, delete the committed local records as well while retaining them when remote cleanup fails.

5. **[P2] Ignore 1013 closes from the previous conversation** — `frontend/src/contexts/conversation-websocket-context.tsx:798`

   When navigation from conversation A to B races the server close for A, `useWebSocket` dispatches close handlers through its mutable `optionsRef`, and this provider is reused rather than keyed by conversation. A late 1013 for A can therefore invoke either the stale recovery closure for A or the newly installed closure for B, resuming or activating the wrong conversation. Capture and compare the socket conversation ID before invoking recovery.

openhands/app_server/sandbox/remote_sandbox_service.py
```diff
@@ -552,6 +552,7 @@ async def resume_sandbox(self, sandbox_id: str) -> bool:
                 _logger.info(
                     f'Updated session_api_key_hash for sandbox {sandbox_id} after resume'
                 )
+                await self.db_session.commit()
```
simonrosenberg: **#1 · correctness — committing the shared request session can durably delete sub-conversations**

`resume_sandbox` now calls `await self.db_session.commit()` on the *request-scoped* `AsyncSession` that is shared with every other injected service (it is cached on `InjectorState` by `DbSessionInjector.inject`). Committing here flushes whatever the caller has staged but not yet committed.

Concrete failure: a user deletes a managed-credential conversation that has sub-conversations.
1. `delete_app_conversation` → `_delete_sub_conversations` stages `_delete_from_database(sub_conversation)` on the shared session (uncommitted).
2. `_delete_from_agent_server` on the parent hits the managed-marker branch → `sandbox_service.resume_sandbox(sandbox.id)`.
3. runtime-api returns a new `session_api_key`, so this line commits — **flushing the sub-conversation deletions**.
4. `activate_codex_credential_binding(..., required=True)` then fails, `_delete_from_agent_server` re-raises, `delete_app_conversation` returns `False`, request rolls back.

The user sees "delete failed" and the parent is still listed, but its sub-conversations are permanently gone.

**Fix:** drop the `commit()` and let the caller own the transaction (the hash update is already staged on `stored_sandbox` and will be committed by the normal request-scope commit). If resume genuinely needs the new hash durable independent of the caller, use a separate short-lived session rather than the injected one.

simonrosenberg (commented):
## Automated deep review of `6535ec8` — revised against the #15393 design

> **Revised.** The first pass of this review was written before I had read #15393 closely. I have withdrawn 2 findings that describe deliberate design decisions, corrected the suggested fix on 2 more that would have broken stated merge gates, and re-framed 1. Individual comments below are edited in place and marked. Sorry for the noise — the design doc changes the triage a lot.

Ran a multi-agent max-effort review over this head (6 independent finder angles → 78 candidates → an independent adversarial verification pass). After re-scoring against #15393: **11 findings stand, 2 are withdrawn, 2 had wrong fixes, 1 is re-framed.**

### On the #15287 size worry — this is not that shape

The 11 that stand are in scope **because they violate #15393**, and every fix is a deletion or a one-liner. Estimated total: **~15–20 production lines, five of them net deletions.** Nothing here needs a new abstraction, endpoint, table, or lifecycle state.

| Design requirement | Findings | Fix size |
|---|---|---|
| §5 *"Untagged/legacy sandboxes fail open and keep their previous pause/delete behavior"* | #9, #10, #4, #7 | 3 lines · few lines · 1 word · 1 word |
| §5 *"An already-deleted sandbox is also complete"* | #3 (subsumes most of #13) | 1 line |
| §3 *"A **failed** managed create is eligible for … DELETE"* | #2 | ~2 lines |
| Gate 4 *"Duplicate SaaS rows converge"* | #15 | 1 line |
| Data loss introduced here, no design position | #1, #5 | delete 1 line · delete 4 lines |
| Merge-gate / contract integrity | #11, #14 | delete 1 line + 1 test · ~3 lines |

The single largest cluster is one idea repeated: **make the legacy pause/delete path fail open the way §5 already says it should.** That is *removing* branches from the barrier, not adding them.

### Withdrawn

- **#6** — `required=False` semantics. My "the two call sites disagree" argument was wrong; `required` asks a different question at each site. Failing closed on an indeterminate probe is defensible under §3.
- **#8** — the frontend activation latch is specified verbatim in §4 (*"Failed activation is latched … tab visibility or navigation provides an explicit retry boundary"*). I reported the design as a bug.

### Corrected

- **#4** — my original fix (restore clear-key-first ordering) would break **merge gate 9** (*"Pause drains before remote authorization invalidation"*). The reordering was deliberate. The real defect is one word: the handler is `except httpx.HTTPError` and `SandboxError` is not one.
- **#12** — my original fix (gate `_prepare_for_pause` on the feature flag) would break §Rollout (*"Reactivation ignores the flag for already-marked conversations so rollback preserves availability"*). The marker scan must stay flag-independent, and it is. #12 collapses into #9.
- **#13** — the resume-and-wait is spec. Re-framed as an ordering point that #3 mostly resolves.

### One caution

If anyone is tempted to "solve" the per-pause conversation scan (#12) with a cache, an index, or a marker column on the sandbox row — **don't.** The scan is what makes rollback safe, and that kind of addition is exactly the drift #15393's size target guards against.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -633,6 +677,29 @@ async def _start_app_conversation(
...
+                    organization_id=marker_organization_id(
+                        managed_credential_marker_value
+                    ),
+                    sandbox_id=sandbox.id,
+                    conversation_id=conversation_id,
+                    start_task_id=None,
+                    required=True,
+                )
+
+            managed_conversation_created = False
```
simonrosenberg, line 706: **#2 · correctness — the "point of no return" reset is silently undone, so a late failure deletes a live conversation**

This `managed_conversation_created = False` disarms the compensating scrub. But the `except` handler at L717-722 re-arms it: it re-issues a verification GET, finds the conversation still tagged `codexcredentialstarttask == str(task.id)` on the Agent Server, sets `managed_conversation_created = True` again, and then DELETEs `/api/conversations/{conversation_id}`.

The Agent Server copy of the start-task tag is never stripped — only `merge_conversation_tags` strips the app-server copy — so the re-verification always "succeeds".

Concrete failure: the start fully succeeds (conversation created, `AppConversationInfo` saved at L655, binding re-armed at L683, flag cleared here), then `_process_pending_messages` at L710 fails on a transient blip. The handler scrubs the conversation. The user is left with a row in their list whose Agent Server conversation no longer exists — opening it 404s and every event is gone.

**Fix:** trust the boolean. Either return/re-raise early when `managed_conversation_created` is `False`, or make the re-verification unable to flip it back (check the flag before the GET, not after).

---
> **Re-review note:** #15393 §3 pins this down — *"**A failed managed create** is eligible for best-effort conversation-scoped Agent Server DELETE only when the successful response or a follow-up catalog read carries the exact reserved start-task tag."* The scrub is scoped to a **failed create**, and the follow-up catalog read exists to cover *"lost, error, and malformed responses"* — i.e. the case where we genuinely do not know whether the create landed. Once `managed_conversation_created = False` has been reached we **do** know, so the re-verification must not be able to re-arm it. Checking the flag before the GET is the design-compliant fix.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -2781,28 +2976,88 @@ async def _delete_from_agent_server(
...
+            if managed_marker is not None:
+                if sandbox is None:
+                    return
+                if sandbox.status != SandboxStatus.RUNNING:
+                    if sandbox.status not in (
+                        SandboxStatus.STARTING,
+                        SandboxStatus.PAUSED,
+                        SandboxStatus.ERROR,
+                    ):
+                        raise SandboxError('Managed credential sandbox is unavailable')
```
simonrosenberg, line 3028: **#3 · correctness — a managed conversation whose runtime is gone can never be deleted**

`SandboxStatus.MISSING` is not in the allow-list on L2989-2993, so it falls through to this `raise`, and the trailing `except Exception` at the end of the method re-raises because `managed_marker is not None`. `DELETE /api/v1/conversations/{id}` returns **HTTP 500 permanently** — every retry hits the same line.

MISSING is reached on the ordinary path: `STATUS_MAPPING['stopped'] = MISSING`, and `_get_sandbox_status_from_runtime(None)` also yields MISSING after any runtime-API error. So any Codex conversation whose sandbox has been stopped or reaped becomes undeletable.

Note the `sandbox is None` case eight lines up (L2986-2987) returns cleanly. MISSING is the *same situation* — the runtime is gone and there is nothing left to drain — and is simply not covered.

**Fix (one line):** `if sandbox is None or sandbox.status == SandboxStatus.MISSING: return`.

---
> **Re-review note:** this also subsumes most of **#13**. §5 of #15393 — *"An already-deleted sandbox is also complete because its pause/delete path already ran the barrier"* — is exactly this case, and the one-line fix here removes the pointless 120 s resume attempt that #13 describes.

openhands/app_server/sandbox/remote_sandbox_service.py
```diff
@@ -631,9 +626,14 @@ async def delete_sandbox(self, sandbox_id: str) -> bool:
...
                     'deleting record'
                 )
+                stored_sandbox.session_api_key_hash = None
                 await self.db_session.delete(stored_sandbox)
                 return True
 
+            sandbox = self._to_sandbox_info(stored_sandbox, runtime_data)
+            if sandbox.status == SandboxStatus.RUNNING:
+                await self._prepare_for_pause(sandbox, self.httpx_client)
+            stored_sandbox.session_api_key_hash = None
```
simonrosenberg, line 635: **#4 · correctness (security) — session-API-key revocation is skipped on the failure path**

> **Fix corrected on re-review** — my original suggestion (restore the clear-key-first ordering) would have broken merge gate 9. Struck below; the defect itself stands.

The `except httpx.HTTPError` handler at L646 is what commits the revocation of `session_api_key_hash` when a delete fails. But `_prepare_for_pause` on L635 raises `SandboxError` — an `HTTPException` subclass, **not** an httpx error — and it also calls `_has_managed_credential_conversation`, a paginated DB search that can raise anything.

Concrete failure: an operator deletes a sandbox specifically to revoke a leaked `X-Session-API-Key`. The sandbox is RUNNING but the runtime-api payload has no `url`, so `_get_agent_server_url` raises `SandboxError` inside `_prepare_for_pause` (see #9). L646 does not catch it → the handler never runs → the hash is never cleared → no commit → request-scope rollback leaves the row untouched. The leaked key keeps authenticating (including the new credential-binding GET/PUT routes) indefinitely, and every retry fails identically.

~~**Original fix:** restore the original ordering — clear `session_api_key_hash` immediately after loading `stored_sandbox`, before any fallible call.~~ **This was wrong.** #15393 merge gate 9 requires *"Pause drains before remote authorization invalidation or runtime suspension"* — so for a managed sandbox the barrier **must** precede the key clear. Moving the clear back to the top would violate that gate. The reordering in this diff is deliberate and correct.

**Corrected fix (one word):** widen L646 to `except Exception as e:` (or add `SandboxError` to the tuple), so the revocation commit still runs when the barrier is what failed. Fixing #9 removes the most likely trigger; this closes the rest.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -2769,6 +2960,10 @@ async def _delete_sub_conversations(self, parent_conversation_id: UUID) -> None:
                         extra={'conversation_id': str(sub_id)},
                     )
             except Exception as e:
+                if sub_conversation and (
+                    CODEX_CREDENTIAL_BINDING_TAG_KEY in sub_conversation.tags
+                ):
+                    raise
```
simonrosenberg: **#5 · correctness — one managed sub-conversation failure orphans the entire conversation tree**

This `raise` removes the guarantee the method's own docstring still advertises: *"This method handles errors gracefully, continuing to delete remaining sub-conversations even if one fails."*

Concrete failure: a parent has sub-conversations `[A (managed, sandbox reaped to MISSING — see #3), B, C]`. A raises here, the loop aborts, **B and C are never deleted** from the agent server or the DB, and control unwinds out of `_delete_sub_conversations` before `delete_app_conversation` ever reaches `_delete_from_agent_server`/`_delete_from_database` for the parent. The delete returns `False` and the whole tree is orphaned in the database — permanently, because every retry fails at A again.

**Fix:** drop the re-raise and keep the existing `_logger.warning` for all sub-conversations. If a managed sub-conversation failing genuinely must block the parent delete, record it and raise *after* the loop completes, so B and C still get cleaned up.

openhands/app_server/secrets/credential_binding.py
```diff
@@ -0,0 +1,202 @@
...
+        supported = await agent_server_supports_credential_binding(
+            httpx_client, agent_server_url, session_api_key
+        )
+    except httpx.HTTPStatusError as exc:
+        if not required and exc.response.status_code in (404, 501):
+            return False
+        raise SandboxError(
+            'Could not verify Agent Server credential binding support'
+        ) from exc
+    except Exception as exc:
```
simonrosenberg, line 153: > **~~#6~~ WITHDRAWN on re-review against #15393.** Leaving the thread for the record.

My original claim was that `required=False` should suppress this handler, and that this call site disagrees with `sandbox_service.py:333`.

**That argument was wrong.** `required` asks a different question at each site:

- here, it means *"arming the managed protocol is mandatory"*;
- in `_prepare_for_pause`, it means *"this sandbox holds a marked conversation"*.

Those warrant different behaviour, so the two handlers are *supposed* to differ. And failing closed on an indeterminate probe is defensible under §3 of #15393: *"Once the complete protocol is selected, other activation failures abort instead of silently using stale plaintext."*

At most this is a question for the author — should a probe **timeout** (as distinct from a 404/501 "capability absent") retain the legacy path, or abort? Both readings are defensible. Not a defect; no change requested.

openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -336,7 +336,7 @@ async def get_sandbox(self, sandbox_id: str) -> SandboxInfo | None:
                 return None
             container = self.docker_client.containers.get(sandbox_id)
             return await self._container_to_checked_sandbox_info(container)
-        except (NotFound, APIError):
+        except NotFound:
```
simonrosenberg: **#7 · correctness — narrowing this handler turns a Docker daemon hiccup into a 500 across the app**

This was `except (NotFound, APIError)`. With `APIError` dropped, a daemon restart or a dropped socket during `containers.get(...)` / `_container_to_checked_sandbox_info(...)` now escapes. `docker.errors.APIError` is not an `HTTPException`, so FastAPI returns **HTTP 500** — not just from `GET /api/v1/sandboxes/{id}` but from every conversation view that calls `get_sandbox` (`live_status_app_conversation_service.py:873`, `wait_for_sandbox_running`, `batch_get_sandboxes`). Previously the same blip returned `None` and the UI showed MISSING.

The sibling handlers at L534, L555 and L586 still catch `APIError`, so the file is now internally inconsistent — which suggests this narrowing was incidental rather than intended.

**Fix (one line):** restore `except (NotFound, APIError):`.

frontend/src/hooks/use-sandbox-recovery.ts
```diff
@@ -47,28 +27,54 @@ export function useSandboxRecovery({
...
+            setCredentialBindingActivationFailed(false);
             onSuccess?.();
           },
           onError: (error) => {
+            if (activeConversationIdRef.current !== recoveryConversationId) {
+              return;
+            }
+            recoveryInFlightRef.current = false;
+            if (force) {
+              credentialBindingActivationFailedRef.current = true;
```
simonrosenberg, line 75: > **~~#8~~ WITHDRAWN on re-review against #15393.** Leaving the thread for the record.

I reported the latch as a bug: one failed activation permanently kills the chat WebSocket with no in-tab retry.

**That is the specified behaviour.** #15393 §4, verbatim:

> Failed activation is latched to stop HTTP and WebSocket retry loops; **tab visibility or navigation provides an explicit retry boundary**, and stale async callbacks cannot affect a newly selected conversation.

So the latch is intentional, and `visibilitychange` / navigation being the *only* resets is the design, not an oversight. The technical description in my original comment was accurate; the verdict was not. No change requested.

openhands/app_server/sandbox/sandbox_service.py
```diff
@@ -280,6 +287,75 @@ def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
...
+        self,
+        sandbox: SandboxInfo,
+        httpx_client: httpx.AsyncClient,
+    ) -> None:
+        required = await self._has_managed_credential_conversation(sandbox.id)
+        if not sandbox.session_api_key:
+            if required:
+                raise SandboxError('Managed sandbox authorization is unavailable')
+            return
+        agent_server_url = self._get_agent_server_url(sandbox)
```
simonrosenberg: **#9 · correctness — `_get_agent_server_url` sits outside the `required` guard and breaks pause/delete for unmanaged sandboxes**

Every *other* fallible step in this method is wrapped with `if not required: return` (L330-331, L334-335, L338-340). This line is not, and `_get_agent_server_url` raises `SandboxError('No exposed URLs for sandbox: ...')` / `SandboxError('No agent server URL found for sandbox: ...')` unconditionally.

`RemoteSandboxService._to_sandbox_info` produces `exposed_urls = []` whenever a RUNNING runtime's payload has no `url`, so this is reachable for a sandbox with **zero** managed conversations:

- `pause_sandbox` raises instead of pausing. `pause_old_sandboxes` swallows the exception, so the per-user running-sandbox cap silently stops being enforced and the user accumulates sandboxes past `max_num_sandboxes`.
- `delete_sandbox` raises `SandboxError` → HTTP 500 (not the documented 503 `SandboxDeleteRetryError`), the runtime is never stopped and keeps billing, and the DB row can never be deleted.

This executes even when `OH_ENABLE_CODEX_CREDENTIAL_SYNC` is unset, because `_prepare_for_pause` never consults `codex_credential_sync_enabled()` (see #12).

**Fix (three lines):** move the call inside a try, and `if not required: return` on `SandboxError` — same shape as the two handlers directly below it.

---
> **Re-review note: this is the highest-leverage fix in the review.** **#12** folds into it entirely. §5 of #15393 requires *"Untagged/legacy sandboxes fail open and keep their previous pause/delete behavior"* — this line is what breaks that promise. Once `required` is `False` the method must return before **any** network work; today L322 runs first and the two 30 s round trips follow. Fixing this also removes the most likely trigger for **#4** and makes the flag-off blast radius zero.

openhands/app_server/sandbox/docker_sandbox_service.py
```diff
@@ -555,7 +561,13 @@ async def delete_sandbox(self, sandbox_id: str) -> bool:
                 return False
             container = self.docker_client.containers.get(sandbox_id)
 
-            # Stop the container if it's running
+            if container.status == 'running':
+                sandbox = await self._container_to_sandbox_info(container)
+                if sandbox is None:
+                    raise SandboxError(
```
simonrosenberg: **#10 · correctness — this `SandboxError` skips the container stop/remove and leaks the container plus its volume**

`delete_sandbox`'s only escape hatch is `except (NotFound, APIError): return False`. `SandboxError` is neither, so raising here jumps straight past `container.stop(timeout=10)` (L572), `container.remove()` (L575) and the volume removal below.

`_container_to_sandbox_info` returns `None` for an already-documented, non-exotic case — a container whose image has no tags (`_logger.debug('Skipping container ...: image has no tags')`). So: a local user prunes or re-tags the agent-server image while a sandbox container is running, then deletes the conversation. The container keeps running, `openhands-workspace-<sandbox_id>` is never removed, the API returns 500, and every retry fails at the same line. The container and its disk are leaked with no in-product way to clean them up.

Same applies to `_prepare_for_pause` on L570 raising (see #9).

**Fix:** treat an unresolvable sandbox as "nothing to drain" and fall through to stop/remove, rather than raising. Deleting should be best-effort about the barrier and unconditional about reclaiming the container.

tests/unit/app_server/test_remote_sandbox_service.py
```diff
@@ -97,6 +97,8 @@ def remote_sandbox_service(
         httpx_client=mock_httpx_client,
         db_session=mock_db_session,
     )
+    service._has_managed_credential_conversation = AsyncMock(return_value=False)
```
simonrosenberg: **#11 · test-coverage — this fixture stub makes the pre-existing key-revocation security tests vacuous**

Stubbing `_has_managed_credential_conversation` to `AsyncMock(return_value=False)` at module-fixture level forces `managed = False` for every test in the file, so the revocation path in `delete_sandbox` is *always* taken. `test_session_key_invalidated_and_survives_rollback` (and the `test_delete_sandbox_...` case around L929) exist precisely to guarantee that a transient delete failure still commits `session_api_key_hash = None` — they now pass by construction. (There is also a redundant per-test stub at ~L2792.)

What that hides: production `remote_sandbox_service.py:650-659` wraps the call in `except Exception: managed = True`. Any conversation-store hiccup — the *common* case when a DB problem is why the delete failed in the first place — sets `managed = True`, skips the hash clear, skips the commit, and leaves a key the operator asked to revoke fully valid.

No test in the diff exercises `_has_managed_credential_conversation` raising, so that fail-closed-into-fail-open behaviour is untested in both directions.

**Fix:** drop the fixture-level stub (keep per-test stubs where a test genuinely needs it), and add one test where the predicate raises, asserting the hash is still cleared and committed.

openhands/app_server/sandbox/sandbox_service.py
```diff
@@ -280,6 +287,75 @@ def _get_agent_server_url(self, sandbox: SandboxInfo) -> str:
...
+                page_id = page.next_page_id
+                if page_id is None:
+                    return False
+
+    async def _prepare_for_pause(
+        self,
+        sandbox: SandboxInfo,
+        httpx_client: httpx.AsyncClient,
+    ) -> None:
+        required = await self._has_managed_credential_conversation(sandbox.id)
```
simonrosenberg, line 317: **#12 · efficiency — folds into #9; my original fix was wrong**

> **Corrected on re-review.** The defect I described is real but much narrower than I framed it, and the fix I proposed would have broken the rollback story.

~~**Original fix:** `if not codex_credential_sync_enabled(): return` at the top of `_prepare_for_pause`.~~ **This was wrong and would be a regression.** #15393 §Rollout requires:

> Initial arming is behind a default-off environment flag. **Reactivation ignores the flag for already-marked conversations so rollback preserves availability.**

If the flag is switched off during a rollback, sandboxes holding already-marked conversations still need the pause barrier — otherwise their credential never flushes. So the marker scan on this line **must** stay flag-independent, and it is. The flag is consulted in exactly one place (`live_status_app_conversation_service.py:2392`, the arming decision), which matches the design.

**What actually remains** is the cost *after* `required` comes back `False`. That should be zero, and today it isn't: L322 does network-adjacent work before the guard, and the two 30 s HTTP round trips follow. **That is finding #9** — fixing it means a flag-off deployment pays exactly one indexed conversation scan per pause and zero HTTP.

That residual scan is the price of rollback safety and should stay. Please do **not** optimise it away with a cache, an index, or a marker column on the sandbox row — that is precisely the drift #15393's size target guards against.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
```diff
@@ -2781,28 +2976,88 @@ async def _delete_from_agent_server(
...
+                        if not exists:
+                            sandbox = await self.sandbox_service.get_sandbox(sandbox.id)
+                            if sandbox is None or sandbox.status not in (
+                                SandboxStatus.RUNNING,
+                                SandboxStatus.STARTING,
+                            ):
+                                raise SandboxError(
+                                    'Managed credential sandbox is unavailable'
+                                )
+                    sandbox = await self.sandbox_service.wait_for_sandbox_running(
```
simonrosenberg, line 3040: **#13 · re-framed — the resume-and-wait is spec; #3 resolves the pathological case**

> **Re-framed on re-review against #15393.** My original comment asked for the 120 s wait and the re-raise to be removed. Both are specified.

#15393 §5:

> Managed conversation deletion **first performs a catalog-only GET**. A 404 is already complete. Otherwise it **reactivates the binding and calls Agent Server DELETE while authorization is live, propagating a failed final flush.**

So resuming a paused sandbox, waiting for RUNNING, and re-raising on failure are all intended — "propagating a failed final flush" is the point. My original framing was wrong.

**What does look off is ordering.** The design says the catalog-only GET comes *first*; in this diff the GET is at L3014, after `resume_sandbox` (L2996) and this `wait_for_sandbox_running` (L3006). For a sandbox that is genuinely PAUSED you cannot GET without resuming first, so that inversion may be unavoidable — worth a sentence from the author either way.

The pathological case is the one in **#3**: `MISSING` reaches this line and triggers a pointless 120 s resume attempt on a runtime that no longer exists. §5's *"An already-deleted sandbox is also complete because its pause/delete path already ran the barrier"* covers it, and the one-line fix in #3 removes it. With #3 applied, what's left here is the intended paused-sandbox path. No separate change requested.

openhands/app_server/sandbox/sandbox_router.py
```diff
@@ -154,6 +288,151 @@ async def _get_user_context(sandbox_info: SandboxInfo) -> AuthUserContext:
...
+    try:
+        value, version = await store.load_versioned(
+            CODEX_AUTH_SECRET_NAME, organization_id
+        )
+        validate_codex_credential(value)
+    except KeyError as exc:
+        raise HTTPException(status.HTTP_404_NOT_FOUND) from exc
+    except NotImplementedError as exc:
+        raise HTTPException(status.HTTP_501_NOT_IMPLEMENTED) from exc
+    except ValueError as exc:
```
simonrosenberg, line 393: **#14 · correctness — a corrupt secrets store tells the user to re-authenticate with Codex**

This blanket `except ValueError` conflates two very different things: "the stored credential failed `validate_codex_credential`" (L386) and "the secrets store itself is unreadable" (L383-385). `FileSecretsStore._read_data` raises `ValueError('Invalid secrets file')` when `secrets.json`'s root is not an object, and `_versions` raises `ValueError('Invalid credential versions')` — both propagate out of `store.load_versioned(...)` inside the same `try`.

The SDK maps that 422 to a re-auth: `HttpVersionedCredentialBinding._raise_for_status` (`openhands/sdk/credential.py:159`) turns 400/422 into `CredentialNeedsReauthentication("Credential is invalid. Please authenticate again.")`. So a store-side data problem surfaces as "log in to Codex again" and tears down the session — and because the readiness probe returns the same 422, it also hard-fails a `required=True` activation.

**Fix (local):** move `validate_codex_credential(value)` out of this `try` into its own `try/except ValueError → 422`, and let store-level `ValueError`s surface as 5xx.

enterprise/storage/saas_secrets_store.py
```diff
@@ -101,6 +158,88 @@ async def store(self, item: Secrets):
...
+
+    def _versioned_query(self, name: str, organization_id: UUID):
+        return (
+            select(StoredCustomSecrets)
+            .filter(
+                StoredCustomSecrets.keycloak_user_id == self.user_id,
+                StoredCustomSecrets.org_id == organization_id,
+                StoredCustomSecrets.secret_name == name,
+            )
+            .order_by(StoredCustomSecrets.id.desc())
```
simonrosenberg, line 243: **#15 · correctness — opposite lock ordering between `store()` and `replace_versioned` over the duplicate Codex rows**

`_versioned_query` orders `id.desc()`, and `replace_versioned` locks through it at L189 (`self._versioned_query(name, org_id).with_for_update()`) — i.e. **descending**. Meanwhile `store()` locks the user's rows at L111-112 with `.order_by(StoredCustomSecrets.id).with_for_update()` — **ascending**. Both can cover the same `CODEX_AUTH_JSON` rows.

This matters precisely because the code explicitly handles duplicates: `replace_versioned`'s `for row in rows: row.secret_value = encrypted` exists for the case where a user has more than one `CODEX_AUTH_JSON` row for an org.

Concrete failure: the user hits Save on the Settings/Secrets page (`store()` grabs id 10, then id 20) at the same moment the sandbox's agent server PUTs a refreshed token (`replace_versioned` grabs id 20, then id 10). Postgres raises `deadlock detected` and aborts one side: either the user's secrets save 500s with their edit lost, or the credential refresh 500s and the freshly minted Codex token is discarded — so the agent's Codex session starts returning 401 once the old token expires.

**Fix (one line):** lock in a consistent direction. Either add `.order_by(StoredCustomSecrets.id)` to the `with_for_update()` query in `replace_versioned` (keeping `desc()` only for the `limit(1)` read at L194), or flip `store()` to descending.

openhands/app_server/app_conversation/live_status_app_conversation_service.py
simonrosenberg: Fixed in 60afce766. Failed-create cleanup is now disarmed before READY/post-create bookkeeping, and successful pre-READY cleanup also removes the local metadata written by that attempt.

simonrosenberg:
Addressed the five follow-up findings in `60afce766`:

- disarm managed failed-create cleanup before post-create bookkeeping;
- restrict managed arming to SaaS remote and OSS Docker;
- reject an existing conversation ID before binding activation;
- remove attempt-owned local rows only after successful remote failed-start cleanup;
- ignore close events from a previous WebSocket URL.

The production change is limited to two existing files and adds no schema, endpoint, scheduler, reservation, or new lifecycle abstraction. Focused backend/frontend tests, Ruff, ESLint, Prettier, TypeScript, and the repository commit hooks pass.

simonrosenberg:
## Re-targeted to `OpenHands/enterprise` → OpenHands/enterprise#77

This PR became unmergeable through no change of its own: the Agent Canvas repo transition (#15396, landed via #15397/#15398) replaced `OpenHands/OpenHands` main with Agent Canvas, deleting every file this PR touches. The result is ~57 modify/delete conflicts, so `MERGEABLE` went `CONFLICTING` and GitHub could no longer build `refs/pull/15394/merge` — which is also why the last three commits show no CI: only the two `pull_request_target` workflows could fire, and the "Validate PR description" failure is collateral (that job checks out `base.sha`, which predates `.github/scripts/check_pr_description.py`).

Resolving those conflicts would have meant deleting the PR's own subject matter, so the work moved instead of merging.

`OpenHands/enterprise` was forked at `ee9e78b7d` — byte-identical to this branch's merge-base. The branch fast-forwards onto its `main` with **zero conflicts** and no code changes; `git merge-tree` reports clean and `ent/main` is a strict ancestor of the head.

Continuing at **OpenHands/enterprise#77** (draft). Closing this one.

cc @graham — thanks for the offer of help above. The re-target itself was mechanical; what's still open is that the transition plan didn't say how in-flight PRs against the old tree should be handled. Anyone else with pending `openhands/app_server` or `enterprise/` work has the same problem, so it may be worth a note in the #general transition thread.

PR #15395, opened 2026-07-27
fix(app-server): allow max_num_sandboxes=1 instead of failing every sandbox start

suju-droid:
PR description:

HUMAN:

Manual testing has not yet been completed. This PR has been validated through the automated backend unit suite and by running Ruff format/lint and mypy against the changed files.

AGENT:

Implemented and validated by an AI coding agent. The final diff was reviewed for scope, formatting, debug code, and unrelated changes.

---

## Why

`start_sandbox()` and `resume_sandbox()` enforce the sandbox concurrency limit by calling:

```python
await self.pause_old_sandboxes(self.max_num_sandboxes - 1)
```

The intent is "make room for the one sandbox I am about to start". So a configured `max_num_sandboxes` of **1** arrives at `pause_old_sandboxes()` as **0**.

Both implementations rejected that:

```python
if max_num_sandboxes <= 0:
    raise ValueError('max_num_sandboxes must be greater than 0')
```

Neither caller catches `ValueError` — `DockerSandboxService.start_sandbox` catches only `APIError`, and `RemoteSandboxService.start_sandbox` catches only `httpx.HTTPError` — so the exception propagates out of the router and **every sandbox start and resume fails** while the limit is set to 1.

This is reachable from a configuration the code itself recommends. `DockerSandboxService.start_sandbox` warns:

> Host network mode is enabled with max_num_sandboxes > 1. Multiple sandboxes will attempt to bind to the same ports, which may cause port collision errors. **Consider setting max_num_sandboxes=1 when using host network mode.**

Following that advice bricks sandbox creation. `max_num_sandboxes` is a plain `int` field with no `ge=` constraint (defaults 5 for docker, 10 for remote), so 1 is accepted at config time and only fails at runtime.

## The fix

`0` is a meaningful request — "pause everything to make room for one" — and the existing downstream logic already handles it correctly, because `num_to_pause = len(running) - 0` pauses all running sandboxes. No other logic change is needed.

The guard is narrowed to reject only negative values, which remain a genuine programming error:

```python
if max_num_sandboxes < 0:
    raise ValueError('max_num_sandboxes must not be negative')
```

Applied to both `SandboxService.pause_old_sandboxes` (base, used by docker/process services) and `RemoteSandboxService.pause_old_sandboxes` (override).

## Behavior change

| `max_num_sandboxes` | Before | After |
|---|---|---|
| 1 | `ValueError` on every start/resume | Pauses all running sandboxes, then starts |
| >= 2 | unchanged | unchanged |
| negative | `ValueError` | `ValueError` (message reworded) |

No API, schema, migration, or frontend changes.

## Scope

This PR fixes only the off-by-one rejection. It intentionally does not add a `ge=1` validator to the config field, since that would reject a value the code currently documents as recommended.

## Issue Number

No linked issue. Found via a static audit of the sandbox lifecycle, then reproduced against the unit suite.

## How to Test

```
poetry run pytest tests/unit/app_server/test_sandbox_service.py \
  tests/unit/app_server/test_remote_sandbox_service.py \
  tests/unit/app_server/test_docker_sandbox_service.py
```

Expected: 194 passed.

The two new `accepts_zero` tests fail on `main` with `ValueError: max_num_sandboxes must be greater than 0` and pass with this change.

## Automated Test Coverage

`tests/unit/app_server/test_sandbox_service.py`
- `test_cleanup_accepts_zero_limit` — a limit of 0 pauses every running sandbox rather than raising.
- `test_cleanup_invalid_max_num_sandboxes` — updated: this test previously asserted that 0 raises, which pinned the bug. It now covers the negative case only.

`tests/unit/app_server/test_remote_sandbox_service.py`
- `test_pause_old_sandboxes_accepts_zero` — same coverage for the remote override.
- `test_pause_old_sandboxes_rejects_negative` — negative is still an error.
- `test_start_sandbox_succeeds_with_max_num_sandboxes_of_one` — end-to-end guard: a service configured with `max_num_sandboxes=1` starts a sandbox successfully. This is the test that most directly encodes the reported bug.

## Video/Screenshots

Not applicable — backend fix with no UI change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Changelog

Fixed sandbox start and resume failing with `ValueError` when `max_num_sandboxes` was configured as 1 — the value recommended for host network mode.

## Notes

- Production change is 4 lines across 2 files; the rest is tests and docstrings.
- `test_cleanup_invalid_max_num_sandboxes` is modified rather than added: its zero-raises assertion encoded the buggy contract.

### DIFF
--- a/openhands/app_server/sandbox/remote_sandbox_service.py
+++ b/openhands/app_server/sandbox/remote_sandbox_service.py
@@ -798,9 +798,12 @@ class RemoteSandboxService(SandboxService):
 
         Uses _get_user_running_sandboxes (runtime /list + DB cross-reference) so
         only sandboxes that are actually running are considered.
+
+        ``max_num_sandboxes`` of zero is valid and pauses every running sandbox,
+        which is what callers request when the configured limit is 1.
         """
-        if max_num_sandboxes <= 0:
-            raise ValueError('max_num_sandboxes must be greater than 0')
+        if max_num_sandboxes < 0:
+            raise ValueError('max_num_sandboxes must not be negative')
 
         running = await self._get_user_running_sandboxes()
 
--- a/openhands/app_server/sandbox/sandbox_service.py
+++ b/openhands/app_server/sandbox/sandbox_service.py
@@ -320,13 +320,15 @@ class SandboxService(ABC):
         In a multi user environment, this will pause sandboxes only for the current user.
 
         Args:
-            max_num_sandboxes: Maximum number of sandboxes to keep running
+            max_num_sandboxes: Maximum number of sandboxes to keep running. Zero is
+                valid and means pause every running sandbox, which is what callers
+                request when ``max_num_sandboxes`` is configured as 1.
 
         Returns:
             List of sandbox IDs that were paused
         """
-        if max_num_sandboxes <= 0:
-            raise ValueError('max_num_sandboxes must be greater than 0')
+        if max_num_sandboxes < 0:
+            raise ValueError('max_num_sandboxes must not be negative')
 
         # Get all running sandboxes (iterate through all pages)
         running_sandboxes = []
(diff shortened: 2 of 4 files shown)


neubig:

Thank you for understanding.

PR #16093, opened 2026-07-27
fix: spawn launcher services without implicit shell

### LINKED ISSUE #15488: Windows launcher shell-parses uvx version constraints and prevents agent-server startup
## Summary

On Windows, the local Agent Canvas launcher starts services with Node's `spawn(..., { shell: true })`. The default agent-server command includes the literal dependency constraint:

```text
agent-client-protocol<0.11
```

`cmd.exe` interprets `<0.11` as input redirection instead of forwarding it as an argument to `uvx`. The agent-server process exits immediately, port 18000 never opens, and ingress repeatedly returns Bad Gateway responses.

## Environment

- OS: Windows
- Agent Canvas: 1.2.1
- Installation: `npm install -g @openhands/agent-canvas`
- Node: 24.13.0
- uv/uvx: 0.10.12

## Steps to Reproduce

1. Install the published package on Windows:
   ```powershell
   npm install -g @openhands/agent-canvas
   ```
2. Run:
   ```powershell
   agent-canvas
   ```
3. Observe that the agent-server exits before binding port 18000.

## Actual Behavior

The launcher logs:

```text
[agent-server] Starting on port 18000...
[agent-server] Using PyPI (1.33.0, default)
[agent-server] Systemet finner ikke angitt fil.
[agent-server] Exited with code 1
[agent-server] Timeout waiting for http://localhost:18000/server_info
```

The localized message means "The system cannot find the file specified." Ingress then reports proxy errors to `http://localhost:18000/`.

Running the generated `uvx` command directly with normal argument handling succeeds. The failure only occurs when the argument vector is routed through `cmd.exe` by `shell: true`.

## Root Cause

Both launcher paths contain:

```js
shell: process.platform === "win32"
```

Affected files:

- `scripts/dev-with-automation.mjs`
- `scripts/dev-static.mjs`

The argument generated by `buildAgentServerCommand()` includes `agent-client-protocol<0.11`. Shell parsing treats `<` as redirection.

## Expected Behavior

Service commands should be spawned directly so every argument is forwarded literally. Callers that genuinely require shell behavior should invoke the shell explicitly as their command.

Windows npm execution already follows this pattern through `buildNpmScriptCommand()`, which explicitly returns `cmd.exe /d /s /c npm run ...`.

## Proposed Fix

Enforce `shell: false` centrally in `getProcessTreeSpawnOptions()` and remove the implicit Windows shell options from both launcher paths.

Add a cross-platform regression test that requests `shell: true`, spawns a child process with `agent-client-protocol<0.11`, and verifies the child receives the exact literal argument. The test fails on current `main` and passes when the helper overrides implicit shell parsing.

## Verification

- Regression test failed before the fix because the shell consumed `<0.11`.
- Focused launcher suite passes after the fix: 95 tests.
- `npm run typecheck` passes.
- `npm run build` passes.
- Full suite result on Windows: 3,691 passed; six unrelated existing Windows/UI tests failed.

## Existing Issue Search

I searched existing issues and did not find this exact failure. OpenHands/agent-canvas#2193 is related to Windows command spawning but concerns a different `C:\Program`/Vite path-splitting problem.

lufen:
HUMAN:

Tested and verified locally on Windows, without these changes the service wouldn't launch.

AGENT:

---

## Why

On Windows, `getProcessTreeSpawnOptions()` spreads caller-supplied options without neutralizing an explicit `shell: true`. When a service is spawned through a shell, cmd.exe interprets metacharacters in the arguments: the default agent-server `uvx` arguments include `agent-client-protocol<0.11`, so `<0.11` is parsed as input redirection. The agent-server exits before binding port 18000, leaving ingress to return Bad Gateway responses.

The static/automation launchers already resolve commands to absolute paths via `resolveWindowsCommand()`, but nothing prevents a future caller from re-enabling implicit shell parsing by passing `shell: true`. This change centralizes the guarantee.

## Summary

- Enforce shell-free service spawning centrally in `getProcessTreeSpawnOptions()` by forcing `shell: false`.
- Add a cross-platform process regression test proving `agent-client-protocol<0.11` reaches the child executable literally, even when a caller passes `shell: true`.

## Issue Number

Fixes OpenHands/OpenHands#15488

## Background

This re-creates the fix from the archived `OpenHands/agent-canvas#1668`, which was closed with a request to re-open it here after the repository moved.

## How to Test

1. Install dependencies:
   ```text
   npm ci
   ```
2. Run the focused launcher test:
   ```text
   npx vitest run __tests__/scripts/dev-process-utils.test.ts
   ```

## Evidence

- Reproduced on Windows against the current implementation: spawning with `getProcessTreeSpawnOptions({ shell: true })` and the argument `agent-client-protocol<0.11` exited with code 1 and empty stdout ("The system cannot find the file specified"), because the shell consumed `<0.11`.
- With `shell: false` forced, the same spawn exits 0 and stdout equals `agent-client-protocol<0.11` verbatim.
- The new regression test encodes exactly this behavior.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The fix is centralized so future launcher callers cannot accidentally re-enable implicit shell parsing by passing `shell: true`. Callers that need shell semantics must invoke the shell explicitly as their command.

### DIFF
--- a/__tests__/scripts/dev-process-utils.test.ts
+++ b/__tests__/scripts/dev-process-utils.test.ts
@@ -1,3 +1,6 @@
+import { spawn } from "node:child_process";
+import { once } from "node:events";
+
 import { describe, expect, it } from "vitest";
 
 import {
@@ -29,6 +32,28 @@ describe("dev process utils", () => {
       cwd: "/tmp",
       detached: process.platform !== "win32",
     });
+  });
+
+  it("passes shell metacharacters to services as literal arguments", async () => {
+    const constraint = "agent-client-protocol<0.11";
+    const child = spawn(
+      process.execPath,
+      ["-e", "process.stdout.write(process.argv[1])", constraint],
+      getProcessTreeSpawnOptions({
+        shell: true,
+        stdio: ["ignore", "pipe", "pipe"],
+      }),
+    );
+    let stdout = "";
+    child.stdout.setEncoding("utf8");
+    child.stdout.on("data", (chunk) => {
+      stdout += chunk;
+    });
+
+    const [exitCode] = await once(child, "exit");
+
+    expect(exitCode).toBe(0);
+    expect(stdout).toBe(constraint);
   });
 });
 
--- a/scripts/dev-process-utils.mjs
+++ b/scripts/dev-process-utils.mjs
@@ -15,7 +15,11 @@ export function isProcessRunning(proc) {
 }
 
 /**
- * Add spawn options needed for process-tree cleanup.
+ * Add spawn options needed for safe service launches and process-tree cleanup.
+ *
+ * Arguments must bypass shell parsing so values such as version constraints
+ * containing `<` are forwarded literally. Callers that need shell behavior
+ * must invoke the shell explicitly as the command.
  *
  * On POSIX, `detached: true` makes the spawned service the leader of a new
  * process group. Later we can signal `-pid` to terminate that whole group,
@@ -30,6 +34,7 @@ export function isProcessRunning(proc) {
 export function getProcessTreeSpawnOptions(options = {}) {
   return {
     ...options,
+    shell: false,
     detached: process.platform !== "win32",
   };
 }


enyst (commented):

🟢 **Good taste** — Setting `shell: false` after spreading caller options makes the safety invariant non-overridable, and the regression test launches a real child process to prove metacharacters arrive literally.

**[RISK ASSESSMENT]**
This is a tiny change, but it sits in a shared launcher helper used by several services on every platform. The current callers do not depend on implicit shell behavior, and the cross-platform process test directly covers the failure mode.

**VERDICT:**
✅ **Worth merging**: The centralized invariant is simpler and safer than relying on every launcher caller to remember it.

**KEY INSIGHT:**
A service launcher should pass an argument vector directly; callers needing shell semantics can name the shell explicitly.

DevinVinson (approved):
Looks like a simple fix. Hope this helps with windows compat!

PR #16095, opened 2026-07-27
fix(security): authenticate WebSockets outside URLs

simonrosenberg:
HUMAN:

AGENT:

---

## Why

Session API keys in browser WebSocket query strings are recorded by reverse proxies and access logs. Production telemetry confirmed this exposure on both conversation-event and bash-event sockets.

## Summary

- Send session authentication as the first WebSocket frame before any onOpen callback or queued bash command.
- Keep event replay options in the URL while removing session_api_key.
- Use the planning sub-conversation key for its socket, with the main key as fallback.

## Issue Number

N/A

## How to Test

- npm run lint
  - Typecheck, ESLint, and Prettier passed.
- npm run build
  - Production application build passed.
- npm run build:lib
  - Library build and declaration generation passed.
- npm test -- --reporter=dot
  - 515 files passed, 1 skipped; 3983 tests passed, 5 skipped, 9 todo.
- After rebasing onto current main, npm test -- __tests__/hooks/use-websocket.test.ts __tests__/hooks/use-bash-command-runner.test.ts __tests__/contexts/conversation-websocket-context.test.tsx --reporter=dot
  - 3 files passed; 18 tests passed, 5 skipped.
  - The hook tests assert the URL contains no key and that auth is sent before application messages and queued bash traffic.

## Video/Screenshots

N/A. There is no visual UI change.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Agent-server first-message authentication is already deployed in the SDK protocol. Existing resend and history query parameters remain unchanged.

---

• **GHCR package:** https://github.com/OpenHands/agent-canvas/pkgs/container/agent-canvas

| Component | Value |
|---|---|
| **Agent Server** | `ghcr.io/openhands/agent-server:1.37.0-python` |
| **Automation** | `openhands-automation==1.3.1` |
| **Commit** | `5d0355f37b4c5e2877d6b36d65100e97c077aeec` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-5d0355f
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-5d0355f
```

```
ghcr.io/openhands/agent-canvas:sha-5d0355f-amd64
ghcr.io/openhands/agent-canvas:fix-websocket-auth-frame-amd64
ghcr.io/openhands/agent-canvas:pr-16095-amd64
ghcr.io/openhands/agent-canvas:sha-5d0355f-arm64
ghcr.io/openhands/agent-canvas:fix-websocket-auth-frame-arm64
ghcr.io/openhands/agent-canvas:pr-16095-arm64
ghcr.io/openhands/agent-canvas:sha-5d0355f
ghcr.io/openhands/agent-canvas:fix-websocket-auth-frame
ghcr.io/openhands/agent-canvas:pr-16095
```

- Each tag (e.g., `sha-5d0355f`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-5d0355f-amd64`) are also available if needed

### DIFF
--- a/__tests__/contexts/conversation-websocket-context.test.tsx
+++ b/__tests__/contexts/conversation-websocket-context.test.tsx
@@ -14,39 +14,43 @@ import {
   getStoredConversationMetadata,
   setStoredConversationMetadata,
 } from "#/api/conversation-metadata-store";
+import type { AppConversation } from "#/api/conversation-service/agent-server-conversation-service.types";
 import type { MessageEvent } from "#/types/agent-server/core";
 
-// Captures the main socket's `onMessage` (`handleMainMessage`) so tests can
-// drive the live message path without a real WebSocket. Only the main socket
-// gets a non-empty url (planning stays ""), so url presence discriminates it.
+type CapturedWebSocketOptions = {
+  onMessage?: (event: { data: string }) => void;
+  queryParams?: Record<string, string | boolean>;
+  sessionApiKey?: string | null;
+};
+
 const wsCapture = vi.hoisted(() => ({
   mainOnMessage: null as null | ((event: { data: string }) => void),
-  mainOptions: null as null | {
-    queryParams?: Record<string, string | boolean>;
-    sessionApiKey?: string | null;
-  },
+  mainOptions: null as CapturedWebSocketOptions | null,
+  calls: [] as Array<{
+    url: string;
+    options?: CapturedWebSocketOptions;
+  }>,
 }));
 
 // Keep the units under test real (the provider, `useConversationHistory`, the
 // event store). Only the network is stubbed: the WebSocket transport and the
 // REST service the history query depends on.
 vi.mock("#/hooks/use-websocket", () => ({
-  useWebSocket: vi.fn(
-    (
-      url: string,
-      options?: {
-        onMessage?: (event: { data: string }) => void;
-        queryParams?: Record<string, string | boolean>;
-        sessionApiKey?: string | null;
-      },
-    ) => {
-      if (url && options?.onMessage) {
-        wsCapture.mainOnMessage = options.onMessage;
-        wsCapture.mainOptions = options;
-      }
-      return { socket: null, reconnect: vi.fn() };
-    },
-  ),
+  useWebSocket: vi.fn((url: string, options?: CapturedWebSocketOptions) => {
+    if (url) {
+      wsCapture.calls.push({ url, options });
+    }
+    if (
+      url &&
+      options?.onMessage &&
+      options.queryParams &&
+      "resend_mode" in options.queryParams
+    ) {
+      wsCapture.mainOnMessage = options.onMessage;
+      wsCapture.mainOptions = options;
+    }
+    return { socket: null, reconnect: vi.fn() };
+  }),
 }));
 vi.mock("#/hooks/query/use-user-conversation", () => ({
   useUserConversation: vi.fn(),
@@ -86,6 +90,7 @@ describe("ConversationWebSocketProvider — conversation-scoped event store", ()
   beforeEach(() => {
     wsCapture.mainOnMessage = null;
     wsCapture.mainOptions = null;
+    wsCapture.calls.length = 0;
     window.localStorage.clear();
     queryClient = new QueryClient({
       defaultOptions: { queries: { retry: false } },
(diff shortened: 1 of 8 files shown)


aivong-openhands:
## Mutation-testing the new suite

I ran mutation testing against the three test files this PR touches. Baseline: 19 passed, 5 skipped.

**Every control mutant dies**, including the ordering ones that are easy to get wrong:

| Control mutant | Result |
|---|---|
| Drop the empty-key guard in `sendWebSocketAuth` | ❌ caught |
| Revert bash gating to `readyState === WebSocket.CONNECTING` | ❌ caught |
| `use-websocket` sends auth *after* the `onOpen` callback | ❌ caught |
| Bash runner flushes queued commands before sending auth | ❌ caught |
| Put the key back in the bash socket URL | ❌ caught |
| Context puts `session_api_key` back into `queryParams` | ❌ caught |

Using `toEqual` on the full `sent` array rather than "contains the auth frame" is what buys the ordering coverage — that's the detail most versions of this test get wrong.

Three mutants survive:

| Surviving mutant | Result |
|---|---|
| `use-websocket`: send auth in the effect body, on a **CONNECTING** socket | ✅ 19 passed |
| Planning socket ignores the sub-conversation's own `session_api_key` | ✅ 19 passed |
| Bash runner: don't clear `readyWsRef` in `onclose` | ✅ 19 passed |

---

### 1. The two mocks disagree about whether `send()` has a precondition

Hoisting the auth call out of `ws.onopen` into the effect body:

```ts
sendWebSocketAuth(ws, optionsRef.current?.sessionApiKey);
ws.onopen = (event) => {
  setIsConnected(true);
  ...
```

leaves `use-websocket.test.ts` green. In a browser that throws `InvalidStateError` — a real `WebSocket.send()` on a `CONNECTING` socket is a hard error, and the auth frame would never reach the server. The mock accepts data in any `readyState`, so the precondition simply isn't modelled.

Notably `use-bash-command-runner.test.ts` *does* catch the equivalent mutation, because it asserts `socket.sent` is empty before calling `socket.open()`. The two mocks are the same class with different rigor.

Cheapest fix is to make the mock behave like the real API — verified this passes as-is and fails with the mutant applied:

```ts
send(data: string) {
  if (this.readyState !== MockWebSocket.OPEN) {
    throw new Error("InvalidStateError: socket is not OPEN");
  }
  this.sent.push(data);
}
```

Worth doing in both mocks, and in the `MockWebSocket` in `use-bash-command-runner.test.ts` too — that one currently passes for the right reason by accident rather than by construction.

### 2. Reconnects aren't covered, and auth now lives inside the connection lifecycle

Related to the above and not covered by any mutant I could write cheaply: `useWebSocket` has `reconnect: { enabled: true }`, and both call sites in `conversation-websocket-context.tsx` use it. The auth frame now has to be re-sent on *every* reconnect. Reading `optionsRef.current?.sessionApiKey` rather than a closure variable is the right call and should make that work — but nothing exercises a second connection, so the property isn't pinned. A test that drives `onclose` → reconnect and asserts the second socket also receives the auth frame first would cover the case that the old query-param approach got for free.

### 3. The planning-socket key fallback is new behaviour with no test

```ts
const planningApiKey = planningAgentConversation?.session_api_key ?? sessionApiKey;
```

Replacing this with plain `sessionApiKey` leaves the suite green. This isn't a mechanical move of the key out of the URL — it changes *which* key the planning socket authenticates with, and it's the one place in the PR where behaviour genuinely differs from before. `keeps the session key out of WebSocket query parameters` only inspects `wsCapture.mainOptions`; capturing the planning options the same way and asserting the sub-conversation's key wins would cover it.

### 4. `readyWsRef.current = null` in `onclose` is probably dead

I'd call this one an equivalent mutant rather than a gap. `onclose` also sets `wsRef.current = null`, and `runCommand` rejects on `!ws` before ever consulting `readyWsRef` — so I can't construct an observable difference. Same for the `onerror` path. Either there's a case I'm missing and it deserves a test, or the two assignments are defensive no-ops that could go. Worth a moment's thought since the same three lines appear in `onclose`, `onerror`, and the cleanup function.

---

### Not a test gap

Removing the `sessionApiKey` parameter from `buildBashWebSocketUrl` outright — rather than leaving it optional and unused — is the right shape. The value can't be re-introduced into the URL by a careless edit because callers can no longer pass it, which is stronger than any assertion.

One contract note: the frame this PR sends (`{"type": "auth", "session_api_key": …}`) has to match what the agent-server's first-message auth expects, and `websocket-auth.ts` holds its own copy of those two literals. Nothing here is wrong, but a rename on the server side wouldn't show up as a failure in this repo.

This comment was generated by an AI assistant on behalf of the user.

simonrosenberg:
Addressed the mutation-testing gaps in 5d0355f37.

- The WebSocket mocks now reject send() unless readyState is OPEN, matching the browser API. I reintroduced the CONNECTING-socket mutant locally and the test failed with InvalidStateError.
- The auth-order test now reconnects and asserts the complete auth-first frame sequence on the second socket. An auth-only-on-first-connection mutant fails this assertion.
- The context test now captures the planning socket and verifies that its sub-conversation session key wins over the distinct parent key, while remaining absent from query parameters. Replacing it with the parent key fails the test.
- The bash mock now enforces the same OPEN precondition.

I agree that clearing readyWsRef in the close/error paths is an equivalent defensive mutation: wsRef is cleared first, and runCommand rejects before consulting readyWsRef, so I did not add a non-discriminative test for it.

Validation: 20 focused tests passed (5 existing skips), full suite 3,986 passed, and typecheck passed.

aivong-openhands (approved):

Issue #15434, opened 2026-07-22
Embedded VS Code button unreachable in self-hosted mode — advertised URL uses page origin while the server binds a separate port, with no config to reconcile them


harish-chandramowli:
### Summary

In a self-hosted deployment (agent-canvas behind a reverse proxy / non-`localhost` origin), the conversation drawer's **VS Code button opens the Canvas app itself instead of the embedded openvscode-server**. The advertised URL uses the page origin (port 443, root path) while the server binds a separate port (19000), and no configuration can reconcile them.

Verified against **agent-canvas 1.5.2** + **openhands-agent-server 1.36.1** (self-hosted VM install per the official backend-setup docs, Canvas fronted by an authenticating HTTPS load balancer).

### Mechanism (all file references verified in the shipped artifacts)

1. **Bind port:** `scripts/dev-with-automation.mjs` (the production path via `bin/agent-canvas.mjs`) computes `vscodePort = preferredBackendPort + 1000` → `18000 + 1000 = 19000` (L390) and passes it to the agent-server spawn as `OH_VSCODE_PORT` (via `buildAgentServerEnv`, `dev-safe.mjs` L705). The computation is pure — it *overwrites* any operator-set `OH_CANVAS_SAFE_VSCODE_PORT` (L404, L745), so there is no operator knob for the vscode port.
2. **No ingress route:** `getLocalServiceRoutes` (`dev-with-automation.mjs` L655-671) proxies the agent server and automation backends but **not** openvscode — so the vscode port is only reachable if the operator separately exposes it.
3. **Advertised URL:** the frontend calls `GET /api/vscode/url?base_url=<window.location.origin>` (`agent-server-conversation-service.api-*.js`, fallback `conversation-service.api-*.js`). `agent_server/vscode_router.py::get_vscode_url` passes the caller's `base_url` straight through (its own default is the hardcoded string `"http://localhost:8001"`, unrelated to the configured port). `vscode_service.get_vscode_url` only uses the real `self.port` when `base_url is None` — which the router never sends.
4. **Client transform:** `transformVSCodeUrl` only rewrites `hostname` when it equals `localhost`; it never adds a port or path.

[...]

PR #16106, opened 2026-07-27
fix: render the VSCode button on self-hosted (local) backends

harish-chandramowli:
HUMAN:

I tested this end to end locally to get VSCode button working from my mac mini / docker setups

- [x] A human has tested these changes ..

AGENT:

---

> **Migration:** this re-opens OpenHands/agent-canvas#1945, closed 2026-07-27 at @neubig's request following Agent Canvas's archival into this repository. The original PR was the render gate alone; review here has since grown it to cover the deployment wiring that makes the rendered button actually reach an editor. This description was rewritten to match the branch — see *History* at the bottom for what changed and why.

## Why

`DrawerVSCodeLink` is gated on `backend.kind === "cloud"`, and that is its only render site — so the VSCode button never appears on a self-hosted install.

The data path for local backends is already fully implemented. `useUnifiedVSCodeUrl` has a complete non-cloud branch, so the hook supports local; only the render gate prevents it from mounting.

The gate reads as a workaround for the agent server advertising a URL that could not work, which the hook's own comment states:

> Cloud mode: read VSCode URL from the cloud-computed `exposed_urls` on the conversation's sandbox. The runtime's `/api/vscode/url` only knows its internal `localhost:8001`, so calling it returned a URL […]

That is a server-side defect, and it is now fixed. `config/defaults.json` pins `agentServer: 1.39.1`, whose `get_vscode_url` appends the configured base path to the origin it is given (`vscode_service.py:110-120`, with a docstring naming exactly the path-routing case). With that, the local URL is correct and the reason for gating on backend kind no longer holds.

This matters for the ordinary self-hosting shape — a machine reached through a single hostname (Cloudflare Tunnel, Tailscale Serve, a reverse proxy), where VSCode is routed by path rather than given its own public port.

## Summary

**Render the button, and gate it on the editor's actual capability state.**

- Render `DrawerVSCodeLink` unconditionally instead of only when `backend.kind === "cloud"`, and drop the cloud-only `pr-1` padding since the ref'd wrapper now always has content.
- Add `isUnavailable` to `useUnifiedVSCodeUrl`, driven by `/api/vscode/status` (`VSCodeClient.getStatus()`) rather than by treating errors as "unavailable". The URL request only runs once the probe says there is an editor, so a deployment with `enable_vscode: false` gets no button and no failing request — and transport, auth and server faults stay visible as ordinary query errors rather than being silently swallowed.
  - `enabled: false` — the deployment switch.
  - `running: false` alongside `enabled: true` — terminal, not a startup window: `VSCodeService.start()` returns `False` when the binary is missing or the port is taken (`vscode_service.py:54-62`), and the lifespan awaits it before yielding (`api.py:205`).
  - probe succeeded but the URL query returns `url: null`.
- Cloud is excluded from `isUnavailable`, so cloud behavior is unchanged (see *Notes*).
- Re-measure the tab row when the button appears or disappears. Its width feeds the inline-tab fit calculation, and its presence is now async, so the `ResizeObserver` watches the button's wrapper as well as the row — otherwise a deployment with `enable_vscode:false` permanently shows one fewer inline tab than fits.

**Make the advertised URL resolvable, from stock configuration.**

The button is only useful if `<origin>/vscode/…` reaches the editor. `getVSCodeUrl` sends `baseUrl: window.location.origin` and agent-server appends the prefix to it, so the origin has to route the prefix to the editor port.

- One editor prefix and port, defined in `config/defaults.json`, flowing to both agent-server (`OH_VSCODE_BASE_PATH` / `OH_VSCODE_PORT`) and the proxy route table. The prefix is preserved, not stripped: openvscode-server is launched with `--server-base-path` and generates its own HTTP and WebSocket URLs beneath it.
- `docker/entrypoint.sh` collapses `VSCODE_*` and the pre-existing `OH_VSCODE_*` into one effective pair before anything reads them, so a deployment that already sets `OH_VSCODE_BASE_PATH=/editor` moves the editor *and* the route together.
- `OH_VSCODE_BASE_PATH` is an explicit argument to `buildAgentServerEnv`, not a field on its config. It changes what gets advertised, so a launcher must opt in and register the route; the helper is shared by every launcher, and when the prefix was always-on three of them advertised a prefix they did not serve. `dev-static` now builds its route tables from `getLocalServiceRoutes` instead of a hand-maintained copy, `dev:minimal` proxies the prefix through Vite, and `dev-extra-backend` deliberately stays out (its browser origin belongs to a different stack, so no global prefix can disambiguate).

**Keep the editor's credential inside one perimeter.**

agent-server sets the editor's connection token to `session_api_keys[0]` (`vscode_service.py:247-249`) and puts it in the URL query string (`:120`) — so the editor's `tkn` is also the agent-server API key.

- The editor route is **not** registered on the `PUBLIC_MODE_PORT` static server. `--auth-required` only controls whether the session key is injected into the served HTML; the dispatcher matches routes and returns before consulting it, so it does not gate proxied paths. The other routes there are fine because agent-server checks the key itself; the editor is not.
- `--no-referrer-prefix` on static-server and ingress sends `Referrer-Policy: no-referrer` on the editor path only. The workbench renders webviews, previews and extension content from the document whose URL carries the token.
- `normalize_base_path` now rejects prefixes that would collide with an existing route (`/api` silently replaced the agent-server route, since the table is keyed by prefix and the editor route is registered last), multi-segment paths, and anything outside `[A-Za-z0-9._-]`. `VSCODE_PORT` is checked as numeric.

## Issue Number

Relates to #15434 (self-hosted VSCode button unreachable) — deliberately not "Fixes", since that issue is written up as the URL/port problem and this PR addresses the render gate plus the deployment wiring around it.

## How to Test

```
npm ci
npm run make-i18n
npx vitest run __tests__/scripts/ __tests__/components/features/conversation/ __tests__/hooks/use-unified-vscode-url.test.tsx
# 32 files, 383 tests passed
npm run lint
```

**End-to-end through a stock launcher.** No manual `INGRESS_ROUTES` or `OH_VSCODE_BASE_PATH` — that is the point of the wiring above:

```
node bin/agent-canvas.mjs
```

Then start a conversation and open the right panel.

- **Before:** no button in the tab bar; `document.querySelector('[data-testid="drawer-vscode-link"]')` returns `null`.
- **After:** the button renders and opens the workbench at `<origin>/vscode/?tkn=…&folder=…` in a new tab.
- **With `OH_ENABLE_VSCODE=false`:** no button, and no error toast.

Two things worth knowing if you reproduce this:

1. **A path prefix is required for the test to mean anything.** Both `getVSCodeUrl` call sites send `baseUrl: window.location.origin`, so with no `vscode_base_path` the server is told the origin and returns `<origin>/?tkn=…` — agent-canvas itself, not the editor. Verified on a patched build: `GET /api/vscode/url?base_url=http://localhost:18300` → `http://localhost:18300/vscode/?tkn=…`, and that URL gives `302` → `vscode-tkn` cookie → `200` workbench (2504 bytes).
2. **`openvscode-server` publishes Linux-only builds** (the 1.109.5 release has exactly three assets, all `linux-*`), and agent-server resolves it from a hardcoded `/openhands/.openvscode-server`. On a Linux host, symlink that path and agent-server spawns the editor itself. On macOS it cannot, so in the recording below the editor was run in Docker (`gitpod/openvscode-server`, `linux/arm64`) on `:19000` with the same flags agent-server would pass — `--server-base-path /vscode --connection-token <session key>` — with the canvas ingress routing `/vscode` to it. The URL construction and routing under test are unchanged by that; only who spawns the process differs.

## Video/Screenshots

Two recordings from the stack described above — same build, same backend, the only difference being `enable_vscode`:

1. **`enable_vscode` on** — a conversation on a `local` backend, right panel open, the VSCode button present in the tab bar. Clicking it opens `http://localhost:18300/vscode/?tkn=…&folder=…`, which loads the real workbench (`.monaco-workbench`, with the conversation's workspace folder in the Explorer).
2. **`enable_vscode: false`** — identical layout and the same tab bar, with no VSCode button and no error toast.

[01-button-works-enable_vscode-true.webm](https://github.com/user-attachments/assets/d5a3fb7f-d3f6-4a67-8ee6-bc58e2185fd3)
[02-button-hidden-enable_vscode-false.webm](https://github.com/user-attachments/assets/386cdf6c-3027-49ad-ab9f-ec3cc16b49a8)
still-01-button-visible
still-02-button-absent

Both are also asserted by unit tests, but the recordings are what show the base path actually resolving end to end.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**This changes an existing tested behavior, deliberately.** `conversation-tabs.test.tsx` asserted `should hide the vscode link when the active backend is local`; both it and the gate arrived together in OpenHands/agent-canvas#1288 ("UI polish: drawer tabs, empty states, and browser chrome"). If hiding the button on local backends was an intentional product decision rather than a consequence of the unusable server URL, then this PR is wrong and I would rather know that than have it merged.

**Sequencing is resolved.** The earlier revision of this description said it depended on software-agent-sdk#4222. That is merged and released: `config/defaults.json` pins `agentServer: 1.39.1`, whose `get_vscode_url` emits the base path. No hold needed.

**Cloud is deliberately left alone.** A cloud sandbox that is still `STARTING` has no `VSCODE` entry in `exposed_urls` yet and will get one shortly, so treating that as "unavailable" would hide the button during startup. `isUnavailable` is therefore hardcoded `false` on the cloud branch and cloud rendering is byte-identical to today. That does leave cloud's own dead-button case (a sandbox that never exposes VSCode) unaddressed; happy to extend it there in a follow-up if you consider that a real state rather than a transient one.

**`docker run --network host` still reaches the editor port directly.** The image does not `EXPOSE 8001` and a test asserts that, but openvscode-server binds `--host 0.0.0.0` (`vscode_service.py:170`), so under host networking — which `playwright.mock-llm-docker.config.ts:52` uses — port 8001 is host-reachable with only `?tkn=` in front of it. Not introduced here, but worth stating plainly rather than claiming the port is unreachable.

## History

The branch has been reworked twice under review, and the description above describes only the current state. For reviewers returning to it:

- The original diff suppressed the 503 from a VSCode-disabled backend with `meta.disableToast`. That was replaced with the `/api/vscode/status` probe, on @neubig's suggestion — it removes the failing request instead of hiding it, and keeps genuine faults loud.
- The Docker/launcher wiring (route registration, one normalized port/prefix pair, prefix opt-in, public-mode exclusion, prefix validation) was added in response to review, and is the larger half of the diff. It is what makes the rendered button reach an editor from a stock install rather than only from a hand-configured one.


neubig (commented):
🟡 **Acceptable**

The direction is useful: removing a cloud-only presentation gate is reasonable once the local deployment advertises and serves the editor consistently. The remaining work can reuse machinery that is already present in this repository.

**[CRITICAL ISSUES]**

- [`scripts/dev-with-automation.mjs`, `docker/entrypoint.sh`] **Complete the stock `/vscode` route before exposing the control.** The current E2E command proves the desired topology, but only by manually supplying `OH_VSCODE_BASE_PATH=/vscode` and `INGRESS_ROUTES`. A typical single-origin deployment looks like this:

  ```text
  browser: https://host/vscode/*
      -> ingress prefix route /vscode (path preserved; HTTP and WebSocket)
      -> http://127.0.0.1:<vscode-port>/vscode/*
  ```

  The prefix should be preserved, not stripped, because OpenVSCode is launched with `--server-base-path /vscode` and generates its HTTP and WebSocket URLs beneath that path. The existing proxy stack already does the hard parts: `createRouter()` chooses the longest matching prefix, `httpxy` forwards the original request path, and both `ingress.mjs` and `static-server.mjs` handle HTTP plus WebSocket upgrades. This should therefore be configuration, not a new proxy implementation:

  1. Define one stable base path (for example `/vscode`) alongside the existing centralized launcher defaults.
  2. Add `[basePath, http://localhost:${config.vscodePort}]` to `getLocalServiceRoutes(config)` when agent-server is launched. That covers the outer npm ingress and the static-server route list built from the same function.
  3. Pass the same value as `OH_VSCODE_BASE_PATH` in the agent-server environment; `OH_VSCODE_PORT` is already populated by `buildAgentServerEnv()`.
  4. In `docker/entrypoint.sh`, set/export the corresponding VSCode port and base path, and add the same `--route` to both normal and public-mode `static-server.mjs` invocations.

  With those values aligned, `getVSCodeUrl({ baseUrl: window.location.origin })` can advertise `https://host/vscode/?...`, and that URL reaches the service without exposing a second public port. Without this wiring, the normal launcher still returns the Canvas origin and the button does not reach the editor. Also hold the UI change until software-agent-sdk#4222 is included in an agent-server release and the centralized `config/defaults.json` pin is advanced through the release workflow; the current `1.37.0` pin does not contain that fix.

- [`src/hooks/query/use-unified-vscode-url.ts`, lines 62-65 and 119-124] **Use the existing status API instead of treating every error as “unavailable.”** `isUnavailable = isError || ...` makes a disabled editor indistinguishable from authentication, connectivity, proxy, or server failures, and `meta.disableToast` then suppresses all of them. `@openhands/typescript-client@1.34.0` already exposes `VSCodeClient.getStatus()` for `/api/vscode/status`, returning `{ enabled, running, message? }`.

  A simpler state flow would be:

  1. After the runtime is ready, query `getStatus()` using the same `getAgentServerClientOptions({ conversationUrl, sessionApiKey })` overrides.
  2. Hide the control only for an explicit terminal capability state (`enabled === false`, or the agreed policy for `running === false`).
  3. Enable the URL query only when status says the editor is available.
  4. Leave transport/auth/server failures as query errors so normal retry and error reporting still apply.

  This removes the blanket toast suppression and the error-derived `isUnavailable` special case, detects a configured editor whose process failed to start, and avoids repeatedly running the primary URL request plus its fallback before concluding the editor is disabled.

**[TESTING GAPS]**

- [Launcher tests] Add focused coverage that the npm and Docker stock configurations register `/vscode`, preserve the prefix, forward WebSocket upgrades, and pass matching port/base-path values to agent-server.
- [`__tests__/hooks/use-unified-vscode-url.test.tsx`] Replace the mocked-503-as-capability case with explicit status cases: disabled hides the button, running exposes it, and a transport/auth failure remains observable as an error.
- [End-to-end] Keep the existing workbench assertion, but run it through an unmodified npm or Docker launcher without manual `INGRESS_ROUTES` / `OH_VSCODE_BASE_PATH`. That is the regression test for the actual self-hosted install path.

**[RISK ASSESSMENT]**

The UI change itself is small, but it spans the frontend, agent-server release compatibility, and two shared deployment entrypoints. The risk becomes manageable once the route, advertised base path, and status model are driven from the same configuration and exercised through a stock launcher.

**VERDICT:**
❌ **Needs rework**: Complete the existing ingress/status integration and release sequencing, then the render-gate removal should be straightforward.

**KEY INSIGHT:**
Treat `/vscode` as one explicitly configured service route and availability as one typed capability state; the repository already contains the proxy and client machinery needed for both.

---

>
>
> **Resolve with AI?** Install the [iterate skill](https://github.com/OpenHands/extensions/tree/main/skills/iterate) in your agent and run `/iterate` to automatically drive this PR through CI, review, and QA until it is merge-ready.
>

---

enyst (commented):

🟢 **Good taste** — The implementation removes the cloud-only render special case, centralizes the editor path/port, uses the typed agent-server client, and covers both launcher wiring and terminal capability states.

**[RISK ASSESSMENT]**
The UI change is straightforward, but the PR also changes Docker and local launcher routing for HTTP/WebSocket editor traffic. Tests cover route precedence, prefix preservation, disabled editors, and missing URLs; the dependent SDK fixes are merged.

**VERDICT:**
✅ **Worth merging**: Core logic and deployment wiring are sound.

**KEY INSIGHT:**
Using one `vscodeBasePath` for both agent-server advertisement and ingress routing eliminates the configuration drift that made local editor links unreachable.

enyst (commented):

🔴 **Needs improvement** - The default path is coherent, but the Docker launcher creates two independent sources of truth for the editor route.

**[CRITICAL ISSUES]**
- [docker/entrypoint.sh, Lines 70-73] **Breaking configuration split**: `VSCODE_PORT` / `VSCODE_BASE_PATH` determine the proxy route at line 290, while pre-existing `OH_VSCODE_PORT` / `OH_VSCODE_BASE_PATH` values are preserved separately for agent-server. A deployment that already configures the documented agent-server variable—for example `OH_VSCODE_BASE_PATH=/editor`—will make agent-server launch and advertise `/editor`, but the proxy will still route `/vscode`; the newly exposed button then opens a dead path. The port override has the same failure mode. Normalize one effective path/port pair before exporting it, and use that exact pair for both agent-server and every static-server route.

**[TESTING GAPS]**
- [__tests__/scripts/docker-vscode-route-sync.test.ts, Lines 63-92] **Override behavior is untested**: These string assertions prove only that the default expressions exist; in fact, they lock in the split precedence above. Add a launcher-level test (or extract the environment normalization into an importable function) that supplies `OH_VSCODE_BASE_PATH` and `OH_VSCODE_PORT` and proves the advertised agent-server values and proxy route remain identical.

**[RISK ASSESSMENT]**
This changes the ingress and editor process wiring for every self-hosted Docker installation. The stock defaults are well covered and the UI behavior has solid regression tests, but supported environment overrides can silently produce an unreachable editor URL.

**VERDICT:**
❌ **Needs rework**: Unify override precedence so the editor's advertised URL and ingress route cannot diverge.

**KEY INSIGHT:**
The editor path and port are one invariant and must be represented by one normalized configuration pair, not parallel `VSCODE_*` and `OH_VSCODE_*` values.

---

>
>
>

enyst (commented):

🟢 **Good taste**

The change fixes the self-hosted button by routing VSCode through the existing authenticated agent-server boundary rather than exposing a new browser-facing service. Launcher configuration, ingress ordering, status gating, and the proxy target are kept consistent, with focused tests around the route and UI visibility.

**[RISK ASSESSMENT]**
This touches network ingress and access to a workspace editor, so the security and deployment blast radius is inherently sensitive even though the implementation follows the existing session-key boundary. **Recommendation**: Do not auto-merge. Request a human maintainer to validate the all-in-one image end to end, including unauthorized access, websocket/upgrade behavior, and the pinned first-party agent-server image.

**VERDICT:**
✅ **Worth merging**: The architecture is sound, subject to human end-to-end validation of the ingress boundary.

**KEY INSIGHT:**
Proxying the editor behind the authenticated agent-server preserves one security perimeter instead of creating a second public port.

smolpaws (commented):
👋 Heads-up on provenance: I'm an AI agent (Claude Opus 5, driven via Claude Code) doing a review pass on this PR. Treat this as a second pair of eyes rather than a gate — I've tried to verify every claim below against the actual code (including the pinned `openhands-agent-server` 1.39.1 wheel and `@tanstack/react-query` 5.100.9 sources) and I've noted where I couldn't. Anything I got wrong, please push back.

---

## What I think is right about this

The redesign since the description was written is a real improvement. Gating the URL request on `/api/vscode/status` instead of muting the 503 with `meta.disableToast` is the better call: it removes the failing request rather than hiding it, so genuine transport/auth faults stay loud. The three-way distinction in the tests — `enabled:false` / `enabled:true,running:false` / `success` with `url:null` — is carefully drawn, and each of the three really does need a different check.

I also checked the reasoning in the comment at `src/hooks/query/use-unified-vscode-url.ts:65-69` and it holds up: `VSCodeService.start()` returns `False` when the binary is missing or the port is taken (`vscode_service.py:54-62`), and the lifespan does `await asyncio.gather(start_vscode_service(), …)` before yielding (`api.py:205`), so `running:false` genuinely is terminal rather than a startup window. Good call.

`__tests__/scripts/docker-vscode-route-sync.test.ts` extracting the marker block and *executing* it under bash, rather than asserting that strings appear in the file, is a nice pattern — it tests the shipped precedence instead of its spelling.

Below, ordered by what I'd want addressed before merge.

---

## 1. `/vscode` is only routed on 2 of the ~5 launchers that now set `OH_VSCODE_BASE_PATH`

This is the one I'd most want fixed. `OH_VSCODE_BASE_PATH` is set inside `buildAgentServerEnv` (`scripts/dev-safe.mjs:714`), which is a shared helper with several callers — but the matching route was added in only two places (`getLocalServiceRoutes` and `docker/entrypoint.sh`). The setting that changes the *advertised URL* is centralized; the setting that makes that URL *resolvable* is not.

Since `getVSCodeUrl` sends `baseUrl: window.location.origin` (`src/api/conversation-service/agent-server-conversation-service.api.ts:522-523`), every unrouted launcher now advertises `<origin>/vscode/?tkn=…` and serves the Canvas SPA shell there instead of the editor — i.e. the button opens a second copy of Canvas.

| Launcher | `OH_VSCODE_BASE_PATH` | `/vscode` route | |
|---|---|---|---|
| `bin/agent-canvas.mjs`, `npm run dev`, Electron, Docker/helm | yes | yes | ✅ |
| `npm run dev:minimal` (`dev-safe.mjs` `main()`) | yes (`:913`) | **no** | ❌ |
| `npm run dev:static` (`dev-static.mjs:305`) | yes | **no** | ❌ |
| `npm run dev:extra-backend` (`dev-extra-backend.mjs:168`) | yes | **wrong port** | ❌ |

Details:

- **`dev:minimal`** starts agent-server (`scripts/dev-safe.mjs:913`) and Vite directly (`:977`) — there is no static-server or ingress in this mode at all. The Vite proxy table (`vite.config.ts:387-434`) lists `/api`, `/server_info`, `/alive`, `/health`, `/ready`, `/docs`, `/redoc`, `/openapi.json`, `/sockets` — no `/vscode`.
- **`dev:static`** does have a proxy, but its two route lists are hardcoded (`scripts/dev-static.mjs:399-418` and `:439-458`), both ending at `/openapi.json`. `/vscode` falls through to `--default` (`:459-460`) → SPA fallback. Worth noting the comment at `scripts/dev-static.mjs:373-374` explicitly claims this table "stays identical to dev-with-automation.mjs" — that invariant is now broken and nothing enforces it. Reusing `getLocalServiceRoutes` here (already imported at `:64-65`) would fix the bug and delete the duplication.
- **`dev:extra-backend`** is the awkward one: it starts a *second* agent-server that the user registers via "Add Backend", but the browser origin belongs to some other stack. If that stack is `dev-with-automation`, `/vscode` resolves — to the *bundled* backend's editor port, not this one. A silent cross-wire is arguably worse than a 404, and no single global prefix can disambiguate it. Simplest fix is to not set `OH_VSCODE_BASE_PATH` here and let it keep the direct-port URL.

The new tests only cover the two launchers that already work, so this gap is invisible to CI. Making `vscodeBasePath` an explicit parameter of `buildAgentServerEnv` rather than an always-on config field would make it impossible for a launcher to enter prefix-mode without opting in.

## 2. The public-mode port now proxies the editor, and its `tkn` is the master session key

Two facts that are individually defensible but combine badly.

**`--auth-required` gates nothing on proxied routes.** I traced it: `scripts/static-server.mjs:135-137` sets the flag, `:547` copies it into `injectionOpts`, and its only consumer is `makeConfigInjectionScript` at `:299-301`, which emits `window.__AGENT_CANVAS_AUTH_REQUIRED__`. The dispatcher checks `route(req.url)` **first** and returns immediately (`:558-563`, and `:581-588` for WS upgrades), so `injectionOpts` is never reached on the proxy branch. It's a frontend-rendering flag, not a proxy gate.

That was already true for `/api/*` on the `PUBLIC_MODE_PORT` server before this PR, and it's fine there because agent-server enforces its own session key. The question is what protects the new `/vscode` backend.

**The answer is a token that is also the API key.** From the pinned `openhands-agent-server==1.39.1` wheel:

```python
# vscode_service.py:247-249
connection_token = None
if config.session_api_keys:
    connection_token = config.session_api_keys[0]

# vscode_service.py:120
return f"{base}/?tkn={self.connection_token}&folder={workspace_dir}"
```

And `docker/entrypoint.sh:151` exports that same value as `OH_SESSION_API_KEYS_0`, with `:162-164` reusing it for `OPENHANDS_AUTOMATION_API_KEY` / `AUTOMATION_LOCAL_API_KEY` / `AUTOMATION_AGENT_SERVER_API_KEY`. So one secret is the agent-server key, the automation key, *and* the editor's `tkn`.

Adding `--route "$VSCODE_ROUTE"` to the public-mode server (`docker/entrypoint.sh:353`) is what puts that credential into a browser-navigable URL on the public origin. Consequences:

- Browser history / bookmarks / session restore persist it in plaintext.
- `Referer` leakage — the workbench renders webviews, Markdown previews and extension content, and there is no `Referrer-Policy` anywhere in this stack (`static-server.mjs:460-471` sets only `Content-Type`/`Cache-Control`; `proxy-utils.mjs:90-111` sets no response headers).
- Upstream access logs — `xfwd: true` (`proxy-utils.mjs:66`) forwards the query string, and the helm ingress routes `/` Prefix to `:8000` with no path exclusions, which is the realistic production shape.

And leaking it isn't "someone can read my editor" — it's full `/api/*` on agent-server and the automation backend, i.e. arbitrary conversation creation in the container.

To be clear, the token-equals-session-key decision is upstream's, not this diff's. But this diff is what makes it reachable from the origin that advertises itself as credential-required. I'd suggest **not** registering `$VSCODE_ROUTE` on the `PUBLIC_MODE_PORT` instance by default (gate it behind an explicit opt-in if the auth-mode E2E suite needs it), and adding `Referrer-Policy: no-referrer` on the `/vscode` proxy path.

## 3. Tab-overflow measurement goes stale when the button disappears

`conversation-tabs.tsx:161-226` measures `vscodeEl.getBoundingClientRect().width` (`:190`) and folds it into `nextCount` (`:197`). The `ResizeObserver` watches only `rowInner` (`:216`), which is `w-full`, and the dep array (`:218-226`) has `backend.kind` but nothing tied to the new async visibility.

Before this PR the button's presence was a synchronous function of `backend.kind`, so `backend.kind` in the deps was sufficient. Now it isn't:

1. Mount → probe pending → `isUnavailable` false → button renders → `measure()` folds in ~90px of button.
2. Probe resolves "no editor" → `DrawerVSCodeLink` returns `null` → the wrapper's width drops to 0.
3. `rowInner`'s own box is unchanged (the child lives inside `ml-auto shrink-0`), so the observer never fires and no dep changed.

Result: `inlineTabCount` stays computed against a button that is no longer on screen, so a self-hosted user with `enable_vscode:false` permanently sees fewer inline tabs than fit. One-line fix: `ro.observe(vscodeEl)` alongside `ro.observe(rowInner)`.

## 4. `normalize_base_path` under-validates — `/api` silently hijacks the whole app

`docker/entrypoint.sh:91-101` strips slashes and rejects only `"/"`. I ran the extracted block under bash with hostile values; these all produce a route:

| `OH_VSCODE_BASE_PATH` | `VSCODE_ROUTE` |
|---|---|
| `/api` | `/api=http://127.0.0.1:8001` |
| `/canvas` | `/canvas=http://127.0.0.1:8001` |
| `/vs=code` | `/vs=code=http://127.0.0.1:8001` |
| `/../api`, `/a b`, `/x?y` | accepted verbatim |

The first two are the real problem. `config.routes` is a plain object keyed by prefix (`static-server.mjs:119`), and `$VSCODE_ROUTE` is the **last** `--route` on both invocations (`entrypoint.sh:325`, `:353`) — so `OH_VSCODE_BASE_PATH=/api` overwrites the agent-server route and sends all API traffic to the editor port. The `= "/"` guard catches the site root but not this.

`/vs=code` is a lesser issue: `static-server.mjs:110` uses the *first* `indexOf("=")`, so the route parses as `prefix="/vs"`, `url="code=http://127.0.0.1:8001"` — a silent outage under `/vs` rather than a startup error. (I checked for argv injection via `OH_VSCODE_PORT` and it's not possible — the quoted expansion stays one argv token.) `VSCODE_PORT` itself is unvalidated too, so a non-numeric value fails at proxy time rather than startup.

Suggest a charset check plus an explicit reject-list for `/api`, `/sockets`, `$AGENT_CANVAS_BASE_PATH`.

---

## Smaller notes

**The description no longer matches the diff.** It describes opting the local query out of the toast via `meta.disableToast` — that's not in the branch; the status-probe design replaced it, which is a better answer and deserves to be the one reviewers read. It also says "the diff is unchanged from the original PR" while the branch has grown four commits of Docker/launcher wiring that the original didn't have. Worth a rewrite before review.

**The sequencing note looks resolved.** `config/defaults.json` pins `versions.agentServer: 1.39.1`, and 1.39.1's `get_vscode_url` already emits the base path (`vscode_service.py:110-120`, with a docstring explaining exactly the path-routing case). If so, the "depends on software-agent-sdk#4222 / happy to hold" caveat can go.

**"Never published" is right for the Dockerfile, slightly off in general.** `EXPOSE 8000` only, and the new test asserts the absence of `EXPOSE 8001` — good. But openvscode-server binds `--host 0.0.0.0` (`vscode_service.py:170`), so under `docker run --network host` — which `playwright.mock-llm-docker.config.ts:52` uses — 8001 is directly host-reachable with only `?tkn=` in front of it. Worth softening the comment at `docker/entrypoint.sh:20`.

**One thing I suspected and had to retract.** After a *failed* probe, `isUnavailable` is false and `isLoading` is false, so the button renders enabled while `localQuery` is still `enabled: false` — I expected `handleClick`'s `refetch()` to be a silent no-op. It isn't: in `@tanstack/react-query` 5.100.9, the observer's `refetch()` never consults `enabled` (`queryObserver.ts:297` → `:323` → `:335` → `query.ts:400`); the only `isDisabled()` check in the package is in `queryClient.refetchQueries` (`queryClient.ts:325`). So the click does fire a real request. Correct as written — but it's load-bearing and non-obvious enough that a one-line comment on the `refetch` in the hook would be worth it, since the sibling API behaves the opposite way.

---

## On the question you asked in *Notes*

> If hiding the button on local backends was an intentional product decision rather than a consequence of the unusable server URL, then this PR is wrong

For what it's worth, the evidence points your way: the gate and the "runtime only knows its internal `localhost:8001`" comment describe the same defect, and 1.39.1 fixes it. I don't think you're fighting a product decision here. Items 1 and 2 are about the *reach* the PR grew beyond the render gate — the Docker/launcher wiring is doing more than the title suggests, and it's the part that needs the closest look.

harish-chandramowli:
Pushed `712b3c90e` addressing the four items from @smolpaws' pass. I checked each against the branch before changing anything; all four reproduced. Details below, including one place where I think the suggested remedy doesn't work and I did something different.

### 1. `/vscode` advertised by more launchers than route it — fixed

Confirmed. `buildAgentServerEnv` (`dev-safe.mjs:717`) set `OH_VSCODE_BASE_PATH` unconditionally and is called from `dev-static.mjs:305`, `dev-extra-backend.mjs:168` and `dev-safe.mjs` `main()`, none of which routed it. `vite.config.ts:387-434` had no `/vscode` key and `dev-static.mjs:399-458` hardcoded two tables that both stopped at `/openapi.json`.

**One correction to the proposed fix.** The suggestion was to not set the prefix for those launchers and "let it keep the direct-port URL". It doesn't keep the direct-port URL. From the pinned 1.39.1 wheel:

```python
if base_url is None:
    base_url = f"http://localhost:{self.port}"
base = base_url.rstrip("/")
if self.server_base_path:
    base = f"{base}/{self.server_base_path.strip('/')}"
return f"{base}/?tkn={self.connection_token}&folder={workspace_dir}"
```

The port default only applies when `base_url` is `None`, and both call sites send `baseUrl: window.location.origin` (`agent-server-conversation-service.api.ts:522-523`); `transformVSCodeUrl` rewrites the hostname but never the port. So opting out yields `<origin>/?tkn=…` — the canvas root — which is the same class of bug one level up.

So the mechanism is what you suggested, but opting out is not a safe default: `vscodeBasePath` is now an explicit argument to `buildAgentServerEnv` and the variable is omitted unless passed, and each launcher that opts in also serves the prefix.

| Launcher | Prefix | Route | |
|---|---|---|---|
| `bin/agent-canvas.mjs`, `npm run dev`, Electron, Docker/helm | opts in | `getLocalServiceRoutes` | ✅ |
| `dev:static` | opts in | now built from `getLocalServiceRoutes` | ✅ |
| `dev:minimal` | opts in | new Vite proxy entry (`ws: true`) | ✅ |
| `dev:extra-backend` | opts out | n/a | ✅ |

`dev:static` reusing `getLocalServiceRoutes` also deletes the duplication behind the `scripts/dev-static.mjs:373-374` comment — you were right that the invariant it claims was unenforced and already broken.

`dev:extra-backend` stays out for the reason you gave: its browser origin belongs to another stack, so a global prefix either doesn't resolve or resolves to the *bundled* stack's editor and hands back a different container's workspace. That's a worse failure than no button, and the URL it now advertises is visibly wrong rather than silently cross-wired.

`__tests__/scripts/vscode-base-path-opt-in.test.ts` asserts the pairing so a future launcher can't advertise a prefix it doesn't serve.

### 2. Public-mode port + `tkn`-is-the-session-key — fixed

Confirmed, including the `--auth-required` trace: it only feeds `makeConfigInjectionScript` (`static-server.mjs:299-301`), and the dispatcher returns on the proxy branch first (`:558-563`, `:581-588`), so it never gates a proxied path.

- Dropped `--route "$VSCODE_ROUTE"` from the `PUBLIC_MODE_PORT` instance. That port exists to drive the auth-mode E2E suite, which doesn't touch the editor, so nothing is lost. There's a test asserting it stays off, with the reasoning inline.
- Added `--no-referrer-prefix` to static-server and ingress, registered for the editor prefix. Scoped to the prefix rather than applied origin-wide — the canvas is served from the same origin and its outbound referrers aren't the problem. Behaviour is covered in `ingress.test.ts` against a real child process (an in-process proxy deadlocks against the MSW interceptor `vitest.setup.ts` installs).

Agreed the `tkn == session_api_keys[0]` decision is upstream's. I'd rather not widen this PR into changing it, but I'm happy to open an issue on software-agent-sdk for a distinct editor token if you think that's the right home for it.

### 3. Stale tab-overflow measurement — fixed

Confirmed: `ro.observe(rowInner)` only (`:216`), with `vscodeWidth` folded in at `:190`/`:197`, and the wrapper inside `ml-auto shrink-0` so the row's own box never changes. Now observes the wrapper too, with a test that asserts the element is observed.

### 4. `normalize_base_path` under-validates — fixed

Confirmed, including that `config.routes[prefix] = url` (`static-server.mjs:119`) makes the last `--route` win, so `/api` really did take over the agent-server route. Added:

- an explicit reject-list (`/api`, `/sockets`, `/server_info`, `/alive`, `/health`, `/ready`, `/docs`, `/redoc`, `/openapi.json`, `$AGENT_CANVAS_BASE_PATH`);
- a single-segment requirement, since agent-server strips slashes when building the advertised URL and a multi-segment prefix would make the two sides disagree — this also catches `/../api`;
- a `[A-Za-z0-9._-]` charset check, which catches `/vs=code`, `/a b`, `/x?y`, `/x#y`;
- a numeric check on `VSCODE_PORT`, which was failing at first-request time rather than startup.

All of them run inside the marker block, so they're executed under bash by the existing test rather than string-matched.

### Smaller notes

- **Description rewritten.** You were right that it described the `meta.disableToast` design that no longer exists, and claimed the diff was unchanged from the original PR. It now describes the branch, with a short *History* section noting the two reworks so a returning reviewer knows what moved.
- **Sequencing caveat dropped**, for the reason you gave — 1.39.1 is pinned and contains the fix.
- **"Never published" softened** at `docker/entrypoint.sh:20`, calling out the `--network host` case explicitly.
- **`refetch` comment added.** Thanks for chasing that one down and retracting it rather than filing it — it is exactly the kind of thing worth a comment precisely because the sibling API behaves the opposite way.

### CI

The one failing E2E (`files-and-git` step 2, `home-chat-launcher` not found) looks like a flake, not a regression: the 12:31 run passed with identical frontend code, and the only delta since was `docker/entrypoint.sh` plus its test — which the non-Docker E2E doesn't exercise. The same locator failure shows up on #16145, #16096 and #16183.

### On the product question

@smolpaws — noted on the render gate, and that matches my reading. I'd still like a maintainer's word on it before merge, since it's the one thing no amount of code reading settles.

<sub>Worked through with Claude Code; every claim above was checked against the branch, the pinned 1.39.1 wheel, or a local run.</sub>

scripts/dev-extra-backend.mjs
```diff
@@ -164,6 +171,14 @@ async function main() {
     {
       cwd: config.cwd,
       env: {
+        // Deliberately not opting into the editor path prefix. This server is
+        // reached by registering it as an extra backend from a browser whose
+        // origin belongs to some *other* stack, so a prefix on that origin
+        // either does not resolve or — worse — resolves to the bundled
+        // stack's editor, silently handing back a different container's
+        // workspace. No single global prefix can disambiguate the two, so this
+        // launcher stays out of prefix-mode; the editor button is unavailable
```
enyst, line 180: **Broken capability contract:** this comment is not what the code does. `buildAgentServerEnv(config)` still starts VSCode, so `/api/vscode/status` reports it available and `DrawerVSCodeLink` renders. With no base path, `getVSCodeUrl` builds from `window.location.origin`, which belongs to the other stack; clicking can reopen that canvas or reach the wrong editor. Disable the editor here or make the capability origin-aware, then test the rendered click path.

enyst (commented):

🔴 **Needs improvement** - The main route works, but two supported launch modes advertise an editor that their browser origin cannot serve.

**[CRITICAL ISSUES]**
- [scripts/dev-extra-backend.mjs, Line 180] **Broken capability contract**: Omitting the prefix does not make VSCode unavailable. This launcher still starts the editor, `/api/vscode/status` reports `enabled: true, running: true`, and the hook therefore renders the button. Because `getVSCodeUrl` supplies `window.location.origin` and no base path is configured, the returned URL is the other stack's root; the click opens the canvas again or can target that stack's editor instead of this backend's workspace. Disable VSCode for this launcher or make availability origin-aware, and cover the actual hook/click behavior rather than only asserting that the env option is absent.
- [docker/entrypoint.sh, Line 394] **Broken public-mode behavior**: The security reason for withholding the editor route is valid, but the same agent-server still reports the editor as available to this origin. After public-mode authentication, the now-unconditional button appears and `/api/vscode/url` returns `<public-origin>/vscode/...`; this static server has no such route, so the navigation falls through to the canvas SPA. Public mode needs an origin-aware way to suppress the capability/button, or an authenticated editor route that does not expose the API credential.

**[TESTING GAPS]**
- The new structural tests preserve both mismatches: one only checks that `dev-extra-backend` omits `OH_VSCODE_BASE_PATH`, and another only checks that public mode omits `VSCODE_ROUTE`. Add regression coverage that drives status → URL → rendered control for these launch modes and proves a visible button always resolves to the intended editor.

**[RISK ASSESSMENT]**
This changes shared launcher routing and exposes a browser URL containing the same credential used for agent-server API authentication. The primary Docker path is carefully wired and tested, but the two inconsistent origins create dead or misdirected controls around a security-sensitive token boundary. **Recommendation**: Do not auto-merge. Request review from a human architect/reviewer to validate the origin-specific capability and credential-routing design.

**VERDICT:**
❌ **Needs rework**: Every origin that renders the editor control must either route the advertised URL to the correct workspace or explicitly report the capability unavailable.

**KEY INSIGHT:**
Editor availability is not solely a property of the agent-server process; it is a property of the browser origin plus its route table.

---

>
>
>

docker/entrypoint.sh
```diff
@@ -281,6 +375,22 @@ PIDS+=("$STATIC_PID")
...
+# and agent-server derives that token from session_api_keys[0], so it is the
+# same secret that authenticates /api. Registering the route here would put
+# that secret in a browser-navigable URL on the origin that exists precisely
+# to test the unauthenticated case, where it would persist in history and
+# leak by Referer from the workbench's own subresources.
+#
+# The token's scope is upstream's to fix and is tracked in
+# OpenHands/software-agent-sdk#4317; if the editor gets a credential of its own,
+# this exclusion and the --no-referrer-prefix below can both be revisited.
 if [ -n "${PUBLIC_MODE_PORT:-}" ]; then
```
enyst, line 434: **Public mode still advertises a route it does not serve.** Omitting `VSCODE_ROUTE` is sensible for the credential threat described above, but the shared agent-server continues returning `enabled/running` and `/api/vscode/url` returns this public origin plus `/vscode`. The unconditional button therefore appears after authentication and opens the SPA fallback. Suppress the capability for this origin or add a genuinely authenticated route.

scripts/dev-extra-backend.mjs
harish-chandramowli: Confirmed and fixed in 83e2271.

You're right that opting out doesn't make the editor unavailable — this launcher still starts it, `/api/vscode/status` still answers `enabled: true, running: true`, and my structural test asserted the launcher's half while saying nothing about what the user sees.

The root cause is the one your key insight names: I made availability a pure server-side probe, and it's server capability ∩ *this origin's* route table.

The origin half now exists. `static-server` gained `--vscode-base-path`, injected as `window.__AGENT_CANVAS_VSCODE_BASE_PATH__`, and it **refuses to start unless a matching `--route` exists** — so the advertisement can't drift from what's actually routed. `#/utils/vscode-origin` reads it and answers both questions: does this origin serve an editor, and does a given URL land under the prefix it serves.

For this launcher specifically, the second check is what does the work. With no prefix configured, agent-server appends nothing to the `window.location.origin` the frontend sends, so the URL comes back as the canvas root — same origin, wrong destination — and `isVSCodeUrlServedByOrigin` rejects it. That covers the case you flagged where the origin *does* route `/vscode`: the extra backend's URL still isn't under it, so the control stays hidden rather than opening the bundled stack's workspace.

Coverage is behavioral now, per your testing note: `__tests__/hooks/use-unified-vscode-url.test.tsx` drives status → URL → `isUnavailable` for this case, and pins the bundled-stack case that must keep working so the guard can't be tightened into hiding the feature.

docker/entrypoint.sh
harish-chandramowli: Confirmed and fixed in 83e2271.

Dropping the route was necessary but not sufficient — the public-mode instance proxies `/api` to the *same* agent-server as the main one, so the probe stayed affirmative, the button rendered after authentication, and `<public-origin>/vscode/…` fell through to the SPA. Omitting only the route left exactly the dead control you describe.

Both halves are now paired. `static-server` gained `--vscode-base-path`, injected as `window.__AGENT_CANVAS_VSCODE_BASE_PATH__`, and public mode passes neither it nor `--route`, so the frontend reports `isUnavailable` and skips both requests rather than probing for a capability it can't use.

Two things keep the pairing honest rather than relying on the next editor of this file noticing:

- `static-server` **refuses to start** if `--vscode-base-path` names a prefix with no matching `--route`. An advertisement can't outlive its route.
- `docker-vscode-route-sync.test.ts` now asserts both directions: the normal instance advertises what it routes (so the feature can't go silently off), and the public-mode instance advertises nothing (so the control can't come back).

I went with suppressing the capability rather than adding an authenticated editor route, since the second option still needs the editor to have a credential of its own — that's OpenHands/software-agent-sdk#4317, and I've now opened OpenHands/software-agent-sdk#4318 against it. If that lands, this exclusion and the `--no-referrer-prefix` alongside it both become revisitable.

enyst (commented):

🔴 **Needs improvement** — The overall single-origin design is sensible, but two route-advertisement invariants still have holes.

**[CRITICAL ISSUES]**
- [scripts/dev-with-automation.mjs, line 1087] **Advertised route is absent on the direct Vite origin**: Full-stack dev bakes `VITE_VSCODE_BASE_PATH`, so `getOriginVSCodeBasePath()` tells the UI that the current origin serves the editor. But this launcher never sets `VITE_VSCODE_TARGET`; `vite.config.ts` only creates the `/vscode` proxy when both variables are present. A browser hitting the supported direct Vite endpoint on `:3001` therefore gets a visible button whose generated `http://localhost:3001/vscode/...` URL falls through to the Canvas SPA. Set `VITE_VSCODE_TARGET` here as `dev:minimal` does, or do not advertise editor capability on this origin.
- [docker/entrypoint.sh, line 130] **The collision guard compares normalized and raw paths**: `VSCODE_BASE_PATH` is normalized before this loop, but `AGENT_CANVAS_BASE_PATH` is not. For example, `AGENT_CANVAS_BASE_PATH=canvas` with `OH_VSCODE_BASE_PATH=/canvas` passes validation, while `static-server.mjs` later normalizes the app mount to `/canvas`; the editor route then captures the Canvas mount and takes the application down. Normalize the Canvas base path before comparison and cover the accepted spellings (`canvas`, `/canvas`, `/canvas/`).

**[TESTING GAPS]**
- Add a launcher test proving the direct Vite origin receives both `VITE_VSCODE_BASE_PATH` and `VITE_VSCODE_TARGET` together.
- Extend the Docker config-resolution test with noncanonical `AGENT_CANVAS_BASE_PATH` spellings so the collision check is tested against the same normalized value the static server uses.

**VERDICT:**
❌ **Needs rework**: The feature still advertises an editor route that one supported origin does not serve, and a valid base-path spelling bypasses the startup collision guard.

**KEY INSIGHT:**
An editor capability must be derived from the exact origin route table, using the same normalization on both sides; any weaker pairing recreates the dead-button or route-takeover failures this PR is meant to eliminate.

**[RISK ASSESSMENT]**
This changes shared launcher, reverse-proxy, Docker, and credential-bearing editor routing behavior. The design has strong test coverage, but the broad deployment blast radius and the two remaining route inconsistencies require correction before merge.

---

>
>
>

---

harish-chandramowli:
Both confirmed against the code and fixed in d477beb10.

**Advertised route absent on the direct Vite origin.** Correct. `startVite` set `VITE_VSCODE_BASE_PATH` and nothing else, and `vite.config.ts:448` spreads the proxy entry only when `VITE_VSCODE_BASE_PATH && VITE_VSCODE_TARGET`. What settles that the Vite origin is a supported endpoint rather than an internal detail is `AUTOMATION_CORS_ORIGINS` at `dev-with-automation.mjs:973`, which allowlists `http://localhost:3001` and `http://127.0.0.1:3001` alongside the ingress. So the advertisement was reaching an origin with no route behind it. Now set together in one block, targeting `http://127.0.0.1:${config.vscodePort}` as `dev-safe.mjs:1015` does.

**Collision guard compares normalized and raw paths.** Correct, and the takeover is real: `static-server.mjs:134` runs `--base-path` through `normalizeBasePath`, which turns `canvas` into `/canvas`, while the reserved-prefix loop compared it raw against an already-normalized `VSCODE_BASE_PATH`. Both sides now go through `normalize_base_path`.

One thing I did beyond the literal fix. `AGENT_CANVAS_BASE_PATH` was resolved above the `>>> vscode-config` markers, so the extracted block the docker test executes never saw its default — the guard could only ever be tested against values a test injected, and the likeliest real instance of this bug is a deployment that moves only the editor and never sets `AGENT_CANVAS_BASE_PATH` at all. Moved its resolution inside the block so that case is reachable; `guards the default canvas mount without being told it` covers it.

Both testing gaps covered:
- `full-stack dev advertises the prefix only where Vite also proxies it` — asserts both env vars are set inside the same guarded block, and that the target is the editor's own port rather than the backend host.
- The collision guard now runs against `canvas`, `/canvas`, `/canvas/` and `//canvas//`, plus a non-colliding noncanonical mount (`canvas/` + `editor`) so normalizing both sides doesn't start rejecting layouts that are fine.

Each new assertion was checked to fail with its own fix reverted — 3 of the 4 spellings did, `/canvas` being already covered by the existing case. Typecheck clean, `bash -n` and shellcheck clean on the entrypoint (only pre-existing SC2329 on the trap-invoked `cleanup`). Full suite 4135 passed / 13 failed, and those 13 reproduce identically with this branch's changes stashed — `conversation-runtime-info`, `recommended-automations` and `use-websocket`, all failing on `localStorage.clear()` being undefined in this environment.

enyst (commented):

🔴 **Needs improvement** - The routing mechanics are carefully synchronized, but the design removes the browser security boundary between Canvas and the bundled editor.

**[CRITICAL ISSUES]**
- [`docker/entrypoint.sh`, Line 382; `scripts/dev-with-automation.mjs`, Lines 701-712] **Credential isolation regression**: Proxying OpenVSCode at `/vscode` puts it on exactly the same browser origin as Canvas. Paths are not security boundaries: JavaScript served anywhere on that origin can read the same `localStorage` and make same-origin API calls. Canvas persists the complete backend registry, including every configured backend's `apiKey`, in `localStorage` (`src/api/backend-registry/storage.ts`, Lines 95-99), so an editor-side XSS, compromised OpenVSCode asset, or unsafe editor/extension web content can now exfiltrate credentials unrelated to the current editor session. `Referrer-Policy: no-referrer` only stops the token-bearing URL from being sent as a Referer; it does nothing to isolate storage or JavaScript authority. Keep the editor on a distinct origin (a separate host/subdomain or port), or first redesign credential storage and establish an equally strong browser isolation boundary; a path-prefix proxy cannot provide one.

**[RISK ASSESSMENT]**
This changes the trust boundary around authentication material across every bundled local/Docker deployment. The blast radius includes all backend API keys saved in the Canvas registry, while the editor is a large, extensible application with a materially broader content and dependency surface. **Recommendation**: Do not auto-merge. Request review from a human security architect/reviewer to validate the editor/Canvas origin model and credential exposure.

**VERDICT:**
❌ **Needs rework**: Correct route plumbing does not compensate for collapsing two applications into one browser origin.

**KEY INSIGHT:**
A URL path separates routing, not trust; browser storage and script authority are scoped to the entire origin.

>
>
>

harish-chandramowli:
Re: the same-origin credential review — I'm not implementing this one, because the fix it prescribes is the design a previous review on this same PR rejected. I'd like a maintainer to break the tie before I move more code. @enyst @neubig

## The finding is technically correct

- **The editor is genuinely new on this origin.** `git show origin/main:docker/entrypoint.sh | grep -i vscode` returns nothing — the stock image never exposed the bundled editor to a browser at all. So this isn't a pre-existing property I can wave off; serving it at `/vscode` on the Canvas origin is introduced by this branch (`579a507e6`).
- **The credentials are genuinely reachable.** `src/api/backend-registry/storage.ts:98` writes the whole backend array to `localStorage`, and `Backend.apiKey` is a plain string (`src/api/backend-registry/types.ts:8`). Same origin means editor-side script authority reads every configured backend's key, not just this one's.

And the principle is right: a path prefix routes, it doesn't isolate. `Referrer-Policy: no-referrer` addresses URL leakage, not storage or script authority. No disagreement there.

## But it contradicts the earlier review on this PR

[Review from 2026-07-30 13:14](https://github.com/OpenHands/OpenHands/pull/16106#pullrequestreview-4819124557), same reviewer, essentially this architecture:

> **KEY INSIGHT:** Proxying the editor behind the authenticated agent-server preserves one security perimeter instead of creating a second public port.
>
> **VERDICT:** ✅ **Worth merging**

[Review from 2026-07-31 04:10](https://github.com/OpenHands/OpenHands/pull/16106#pullrequestreview-4825392683):

> **KEY INSIGHT:** A URL path separates routing, not trust; browser storage and script authority are scoped to the entire origin.
>
> Keep the editor on a distinct origin (a separate host/subdomain or port)

A separate port is the thing the first review counted as the win and the second one requires. I can't satisfy both, and picking one myself would mean discarding four rounds of review that were given on the assumption of the other.

## Three options, as I see them

1. **Separate origin (what the latest review asks for).** Correct isolation, and it matches how Cloud already works — the editor lives on a different host there, which is precisely why Cloud never had this problem. Cost: reopens the second-public-port posture the earlier review objected to, and for the all-in-one Docker image it means a second published port and a second thing to secure in every self-hosted deployment.
2. **Keep the path proxy, fix the actual hazard.** What makes the shared origin dangerous is not that the editor is on it — the editor already has full filesystem and shell access to the workspace, so its own session is not what's being protected. It's that Canvas persists *other* backends' API keys in `localStorage` where editor-side script can reach them. Scoping or removing that storage closes the cross-backend exposure without a second origin. This is a Canvas credential-storage change, though — separate PR, and larger than this one.
3. **Ship the editor off by default on the shared origin,** opt-in per deployment with the tradeoff documented. Smallest change; leaves the sharp edge in place for anyone who opts in.

My preference is (1) if the second port is acceptable to maintainers, since it's the only one that's actually a boundary — but that's a project posture call about the self-hosted deployment story, not something the PR author should decide unilaterally.

smolpaws:
👋 Hi, I'm smolpaws — a small AI cat agent (OpenHands under the hood), here on Engel's behalf. 🐾

Thanks, @harish-chandramowli — you were right to stop and ask for a human call rather than pick an architecture unilaterally, and the same-origin finding is real (I checked it in the code).

Where we landed:

**Ship it.** We want the functionality, and the practical risk is narrow. The editor already has full shell + filesystem access to its workspace, so its own session isn't the concern. The new exposure is that editor-side script could read *other* backends' API keys from `localStorage` — but that needs the editor XSS'd first (or a hostile extension / compromised asset), and this is single-user self-hosted. Small, acceptable edge for a typical install.

Two things we're documenting rather than fixing here:
- Backend keys in origin-readable `localStorage` isn't great — but it's already true today, independent of this PR; the editor is just one more surface. Real fix tracked in **#16492**.
- `OH_ENABLE_VSCODE=0` already turns the button and editor off for anyone who wants it.

So: merge with the shared-origin tradeoff documented (a short deployment-docs note + pointer to #16492), rather than blocking on the credential redesign here.

Only asks before merge: it needs a rebase (currently conflicting), and a one-line note naming the tradeoff so the next person isn't surprised. Thanks again for the careful work. 🐾

enyst (approved):
(HUMAN:)

LGTM, thank you for this @harish-chandramowli !

A note: we could, as you suggest in your last comment, make it opt-in, rather than opt-out. But personally, I'm torn whether we really want to, considering that the risk that the agents grumble about is dependent on some other exploit, and we do aim to fix the local storage anyway.

If I may suggest, rather, after it's ready give it another spin just to see that we didn't break things with the recent fixes, and please let me know.

Thank you again!

harish-chandramowli:
@enyst @smolpaws — synced with `main` (conflicts were all in the launcher/proxy code #16090 touched), tradeoff documented, re-verified. Ready for another look.

**Tradeoff note** — added to `docs/SELF_HOSTING.md` (the editor shares the canvas origin, so editor-side script can read the backend API keys Canvas keeps in `localStorage`; a path prefix routes but doesn't isolate), plus a pointer on the `vscodeBasePath` entry in `config/defaults.json`. Both cite #16492.

**Off-switch** — confirmed, @smolpaws. It's an agent-server setting (`Config.enable_vscode` via the `OH_*` env layer), not a Canvas one, so it needs no code here. With `OH_ENABLE_VSCODE=0`: `/api/vscode/status` → `{"running":false,"enabled":false}` and `/api/vscode/url` → 503, which is what the gate keys off, so the control hides.

**One behaviour change beyond the merge**: the editor's dev proxy target moved `localhost` → `127.0.0.1`, matching #16290 and the comment directly above it. The Docker route and Vite target were already IPv4; this was the last one out of step.

**The spin** you asked for, @enyst — Built the Docker image off the merge, since that's the only place the real `openvscode-server` ships:

- Advertised URL → 302 (token cookie) → workbench 200. `workbench.js`/`.css`, `nls.messages.js`, `manifest.json`, favicon all 200 under `/vscode/stable-<commit>/static/…`; WebSocket upgrade under the prefix → `101`. `Referrer-Policy: no-referrer` present. `/canvas/` and `/server_info` unaffected.
- Public-mode origin (`PUBLIC_MODE_PORT`) still shows why the origin half of the gate is needed: the same agent-server truthfully answers `{"running":true,"enabled":true}` there, but that origin registers no editor route, so `GET /vscode/` lands on the canvas SPA shell. Gating on the server probe alone renders a button that opens the SPA; intersecting it with the origin's route table is what hides it.

harish-chandramowli:
`Validate PR description` is red and I can't clear it from here — it needs a label I don't have rights to add.

The merge pulled in the new issue-readiness gate, so this PR (opened before it existed) is now held to it. **Could one of you add `ready-for-dev` and `bug` to #15434?** That's the whole fix:

```
None of the linked issues (#15434) carry the `ready-for-dev` label.
This PR is marked as `bug` but none of the linked issues (#15434) carry the `bug` label.
```

Adding labels there returns `403 Must have admin rights to Repository` for me. Nothing else fits — no open `ready-for-dev` issue relates to the editor.

One thing worth fixing in the checker independently: until just now the message was the misleading `Link an issue in the ## Issue Number section`, even though the section did link one. `BARE_ISSUE_REF_RE = (?<!\w)#(\d+)` rejects the fully-qualified `OpenHands/OpenHands#15434`, because the `s` before `#` trips the lookbehind. I've switched this PR to a bare `#15434`, but the next contributor who writes the qualified form will get told to link an issue they already linked. Happy to send a one-line fix (`(?<![\w/])` plus an `owner/repo#N` branch) as a separate PR if useful.

PR #16115, opened 2026-07-28
fix(skills): keep modal actions visible for long descriptions

### LINKED ISSUE #15424: [Bug]: Skill UI is cut off with big description
### Operating System

macOS

### Installation Method

### Agent Canvas Version

1.5.2

### Bug Description

Image

can't scroll to see what's happening here

### Steps to Reproduce

create a skill with a long description

### Actual Behavior

cant see everything

### Expected Behavior

can

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

KXHXK:
HUMAN:

Manually tested a long skill description at desktop and narrow viewport sizes. Only the middle details area scrolled; the title, close control, and Close/Use Skill actions remained visible, with no content overflow or button occlusion on the narrow viewport.

AGENT:

The modal was exercised end to end with `npm run dev:mock` in Chromium at 1280x800 and 390x844. A long description was injected into a real skill modal in the rendered page. In both viewports, the content pane scrolled to its measured maximum while the header and action footer remained visible.

---

## Why

Long skill descriptions make the detail modal look cut off because the entire modal currently scrolls and its scrollbar is hidden until hover. This moves the close and use-skill actions below the fold and leaves little indication that more content is available.

## Summary

- Keep the skill header and action footer outside the scrollable region.
- Give the details pane a visible scrollbar and preserve scrolling for long descriptions and content.
- Add regression coverage for the bounded, independently scrollable content layout.

## Issue Number

Fixes #15424

## How to Test

1. Run `npm ci` and `npm run dev:mock`.
2. Open `/skills` and select a skill with a description long enough to exceed the modal height.
3. Confirm the header and Close/Use skill actions remain visible.
4. Scroll the middle details pane and confirm the description, metadata, and content remain reachable.

Validation performed:

- `npm run lint` - passed.
- `npx vitest run __tests__/components/features/skills/skill-detail-modal.test.tsx __tests__/routes/skills-settings.test.tsx` - 20 tests passed.
- `npm run build` - passed.
- `npm run build:lib` - passed.
- Full `npm test` was attempted locally but exceeded a 10-minute command timeout without returning a result; focused coverage and GitHub CI remain the authoritative test signal.

Browser measurements:

- 1280x800: details pane scrolled from 0 to 392px; the action footer stayed within the modal.
- 390x844: details pane scrolled from 0 to 658px (659px maximum); the action footer stayed within the modal.

### Desktop (1280x800)

Skill detail modal with a long description on desktop

### Narrow viewport (390x844)

Skill detail modal with a long description in a narrow viewport

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The layout follows the existing add-skill and plugin-detail modal pattern. No API, persistence, or configuration behavior changes.


enyst (commented):

🟢 **Good taste** - Elegant, simple solution.

The scroll boundary is moved to the content pane while the header and actions remain fixed. The implementation follows existing modal structure, the focused regression test asserts the important DOM boundary, and the PR includes concrete desktop and narrow-viewport evidence. I found no actionable correctness or maintainability issues.

**[RISK ASSESSMENT]**
This is a localized presentation-layer change with focused regression coverage, successful CI, and browser evidence at both desktop and narrow viewport sizes. It does not alter persistence, APIs, or shared state.

**VERDICT:**
✅ **Worth merging**: Core logic is sound.

**KEY INSIGHT:**
Keeping the modal shell fixed and making only the variable-length content pane scrollable removes the original failure mode without adding special cases.

VascoSch92:
@KXHXK 

Thanks for the PR :-) 

Approach looks right to me.

Few things before I'd merge though.

The test doesn't actually catch the bug it's named after. jsdom doesn't load Tailwind, so `toHaveClass` is just re-asserting the string from a few lines up and `toBeVisible()` passes regardless. I moved `overflow-y-auto max-h-[85vh]` back onto the root to check — bug fully restored, test still green. Brittle the other way too: renaming `custom-scrollbar-always` fails it even though nothing behavioural changed. AGENTS.md calls this out under TESTING_RULES. I'd drop it, or go after the actual behaviour if you want coverage here.

Root div wants `overflow-hidden`. It lost `overflow-y-auto` and didn't get anything in its place, so if the header + footer alone exceed 85vh (long name wrapping, or zoomed in) there's nothing left to scroll since both are `flex-shrink-0`. That used to be handled by the root scrolling. Also keeps content clipped to the rounded corners.

Description `<p>` still needs `break-words` — `plugin-detail-modal` does this for the same content. Long unbroken URL or path and it overflows sideways, and since the pane sets `overflow-y: auto` the x axis computes to auto too, so you end up with a horizontal scrollbar inside the modal. Probably want `whitespace-pre-line` alongside it — `getSkillCardDescription` preserves the `\n` from block-scalar frontmatter and those are all collapsing right now.

Rest are nits, take or leave:

- `px-6 ... pr-12` works, but only because of emit order — `pl-6 pr-12` says it outright. (siblings put `pr-6` on the `h2` instead)
- no `overscroll-contain` on the new scroll pane, so a trackpad flick over the pinned header/footer scrolls the skills list behind the modal. `conversation-panel.tsx` already guards this
- `<header>`/`<footer>` inside a `role="dialog"` map to banner/contentinfo landmarks and duplicate the page's own. `add-skill-modal` and `plugin-detail-modal` have the same thing, so not really yours to fix — just flagging it
- shell classes are copy-pasted across 8 files now. `ModalBody` genuinely doesn't fit (single centered pane, no header/body/footer split) so there was nothing to reuse, but a shared `ModalShell` would be nice eventually

On `custom-scrollbar-always` vs the hover-only one the siblings use — think you picked the right one, it's the more common utility in the repo (30 files). Worth flipping the siblings over separately though, long plugin descriptions hit the same complaint #15424 describes.

Please ping me after.

KXHXK:
Thanks for the detailed review. Addressed the actionable items in `945c141`:

- removed the jsdom class-assertion test, since it did not exercise computed layout or scrolling
- added `overflow-hidden` to the modal shell
- added `overscroll-contain` to the content pane
- added `break-words whitespace-pre-line` to the description
- replaced the conflicting header padding utilities with explicit `pl-6 pr-12`

I left the landmark and shared-shell observations out of this focused change, as noted.

Validation:
- focused modal tests: 4/4 passed
- ESLint on both changed files: passed
- Prettier and `git diff --check`: passed

I also ran the full typecheck. It currently reports unrelated generated-type mismatches in the automation catalog and MCP service; none point to either changed file.

CI follow-up: the refreshed Ubuntu and Windows jobs pass. Validate PR description exits before validating the PR because the workflow invokes check_pr_description.py --files-file, while the checked-out script rejects that option as unrecognized. Mock-LLM is still pending; this validator failure does not indicate a contributor-side code or description issue.

VascoSch92 (approved):
LGTM

thanks

PR #16124, opened 2026-07-28
feat(skills): filter the skills page with a faceted rail

sleeyax:
HUMAN: 

improve skills UI with filters support

**NOTE**: CI will continue to fail until https://github.com/OpenHands/extensions/pull/425 is merged, released and installed as a proper dependency in this branch.

AGENT:

Verified locally with `npm run typecheck`, `npm run lint`, and the skills unit suites
(`__tests__/components/features/skills/`, `__tests__/routes/skills-settings.test.tsx`,
`__tests__/utils/skill-category.test.ts`) — see **How to Test** for the exact commands and
results. Behaviour was exercised by hand in `npm run dev` against a local agent-server with
public, personal, and project skills present; screenshots below.

---

## Why

`/skills` rendered every skill in one flat grid behind a search box and a single "type" dropdown.
With 58 public skills in the bundled `@openhands/extensions` catalog — plus personal and project
skills — search is only useful when you already know the name you want, and `type`
(`agentskills` / `knowledge` / `repo`) is an implementation detail nobody browses by. There was no
way to answer "what's enabled?", "what came from this project?", or "what do I have for code
review?", and no way to link someone a filtered view.

Full motivation and discussion: OpenHands/OpenHands#16123.

## Summary

- Replace the type dropdown with a faceted filter rail on `/skills`: **State** (enabled/disabled),
  **Source** (project/personal/public), **Category** (the nine-value taxonomy from
  OpenHands/extensions#425), and **Type**. Multi-select within a group, AND-ed across groups,
  with a live count on every row.
- Move filter state into the URL (`?q=…&state=…&source=…&category=…&type=…`, one key repeated per
  value) so a filtered view is linkable and survives a reload.
- Add `src/components/features/skills/skill-filter.ts` — a pure, table-driven model that owns
  parsing, serialising, filtering, facet-group construction, and toggling. A new facet group is
  declared in one place (`GROUP_DEFS`) and every operation follows from it.
- Show the category on each skill card, and render the rail as a modal behind a Filters button
  below `md`.

## Issue Number

OpenHands/OpenHands#16123

### 1. Point the frontend at the extensions branch

The Category facet is driven by the `category` field and the `SKILL_CATEGORY_IDS` export added in
OpenHands/extensions#425. Neither exists in the published `@openhands/extensions@0.11.0` that this
branch's lockfile installs, so against npm the category work cannot be exercised at all and
`npm run typecheck` fails on the missing export — that is the CI failure called out at the top.
Link a local checkout of the extensions branch to test it:

```bash
# 1. Get the extensions PR branch (anywhere on disk; a sibling directory is convenient)
gh repo clone OpenHands/extensions ../extensions
cd ../extensions
gh pr checkout 425

# 2. Optional — regenerate the catalog to confirm the build produces what is committed.
#    Expect: 'Generated .../skills/index.js with 55 skills' and a clean `git status`.
#    The two "no marketplace entry" warnings (qa-changes, release-notes) are expected.
node scripts/build-skills-catalog.mjs

# 3. Link it into this repo
npm link
cd -                              # back to the OpenHands checkout
npm link @openhands/extensions
```

Confirm the link resolves and the catalog really carries categories:

```bash
readlink -f node_modules/@openhands/extensions   # → your ../extensions checkout

node -e "import('@openhands/extensions/skills').then(m => {
  const counts = {};
  for (const s of m.SKILLS_CATALOG) counts[s.category] = (counts[s.category] ?? 0) + 1;
  console.log(m.SKILL_CATEGORY_IDS.length, 'categories,', m.SKILLS_CATALOG.length, 'skills', counts);
});"
```

Expected on the branch — 9 categories over 55 skills:

```
9 categories, 55 skills {
  'code-quality': 6, 'agent-authoring': 8, 'code-hosting': 8, integrations: 5,
  environment: 10, writing: 4, other: 3, design: 2, automations: 9
}
```

Notes while the link is in place:

- Any `npm install` / `npm ci` in this repo replaces the link with the published package; re-run
  `npm link @openhands/extensions` afterwards.
- Vite pre-bundles the linked package. If a catalog rebuild does not show up in the browser,
  `rm -rf node_modules/.vite` and restart the dev server.
- To undo: `npm ci` here (restores the published 0.11.0), then `npm rm --global @openhands/extensions`.

### 2. Checks

```bash
npm run make-i18n
npm run typecheck
npm run lint
npx vitest run __tests__/components/features/skills __tests__/routes/skills-settings.test.tsx __tests__/utils/skill-category.test.ts
```

### 3. By hand

With `npm run dev` and a local agent-server:

1. Open `/skills`. The rail lists only groups that discriminate — with public skills only, Source
   does not render. **Category** lists the nine values above, each with its own count.
2. Tick **State → Disabled**. The grid narrows, the URL gains `?state=disabled`, the result count
   updates, and a **Clear filters** link appears.
3. Tick a second value in the same group (e.g. **Enabled**) — they OR together. Tick a value in
   another group — the two groups AND.
4. Reload the page. The selection survives. Copy the URL into a new tab — same view.
5. Watch the counts on rows in a group you have *not* selected from: each one predicts exactly how
   many cards you get if you click it.
6. Narrow the window below `md`. The rail collapses into a **Filters** button carrying a badge with
   the active count; the modal holds the same rows.
7. Hand-edit the URL to `?category=not-a-real-category`. The unknown value is dropped rather than
   filtering everything away.

To see the documented degradation without the extensions branch, drop the link (`npm ci`) and
reload `/skills`: every public skill reports `other`, so the Category group has a single
discriminating value and is hidden entirely. The rest of the rail is unaffected.

Results on this branch: typecheck ✅, lint ✅, skills suites ✅ (13 files, 82 tests passed).

## Video/Screenshots

Default state (nothing selected):

image

Filter by category:

image

Supports a mobile-friendly UI through a filters button and modal:

image

image

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Depends on OpenHands/extensions#425** for the `category` field on catalog entries. This PR
  degrades cleanly without it: `getSkillCategory()` maps an absent or unrecognised category to
  `other`, and a group with fewer than two discriminating values is hidden entirely, so an older
  catalog just means no Category facet.
- Local (personal/project) skills have no category — the agent-server does not carry one — so they
  all sit under `other`. Making that axis meaningful for local skills is a separate change.
- Facet-group visibility and the row set are computed from the **unfiltered** skill list on purpose.
  If they tracked filtered counts, narrowing a group down to one value would make it vanish
  mid-click. Row *counts*, by contrast, exclude the group's own selection, which is what makes them
  predict the outcome of clicking.
- A selection carried in from the URL for a value the user has no skills for keeps its row visible,
  so the filter can be seen and undone rather than being enforced invisibly.
- The removed `skills-type-filter-dropdown.tsx` / `skill-type-filter.ts` had no other callers.
- `src/utils/settings-like-page-layout-classes.ts` now reserves the scrollbar gutter on
  settings-like pages; without it the wider two-column layout shifted horizontally when the grid
  grew past a screenful. It touches the other settings pages too — worth a look.


enyst:
hey @sleeyax , thank you for the PR! Could you maybe see why are CI tests failing?

It’s possible that mock LLM tests are flaky, but the others are expected to work 🤔

sleeyax:
@enyst please see my comment at the top of this PR where I explain why ci tests will not work on GitHub (but do locally).

To repeat, it's because it depends on a change in the extensions repo. Since that new version isn't released yet, it can't be pulled in so the pipeline fails.

The changes can be tested locally by following the "how to test" instructions.

Let me know if you believe there's a better way to do this.

enyst (commented):
🟡 **Acceptable** — the table-driven filter model is a solid direction, but four user-visible gaps need attention before merge.

_This review was generated by an AI agent (OpenHands) on behalf of enyst._

**[CRITICAL ISSUES]**

- [`src/api/skills-service.ts`, lines 37-40] **Cloud categories are never supplied.** The category-producing path is `SKILLS_CATALOG.map(catalogEntryToSkillInfo)`, but Cloud returns `fetchCloudSkills()` unchanged before reaching it. The current Cloud response contract/tests contain no `category`, so every Cloud skill falls back to `other` and the Category facet normally disappears. Either add `category` to the Cloud API contract or enrich matching public Cloud skills from a catalog name-to-category map; add a Cloud-path test proving a public skill retains its real category.

- [`src/routes/skills-settings.tsx`, lines 48-51 and 103-124] **Browser history cannot restore `q`.** `queryInput` reads the URL only during initialization. If back/forward navigation changes `q` while the route remains mounted, the stale local value remains visible and the debounce writes it back over the historical URL 300 ms later. Synchronize external URL changes into the input while cancelling/distinguishing pending local writes, and add a mounted-route back/forward test.

- [`src/components/features/skills/skill-filter.ts`, lines 300-316] **Facet counts stop predicting clicks once that group has a selection.** Counts exclude the group's own selection entirely. With `environment` selected, the `writing` row shows only the number of writing skills, although clicking it produces `environment OR writing`; clicking a checked row likewise shows that row's cardinality rather than the post-removal total. This contradicts the advertised “count predicts the outcome” behavior. Compute each row's count from the state after toggling that row, and test both adding and removing values in a multi-select group.

- [`src/components/features/skills/skill-filters-modal.tsx`, line 31] **The mobile-only modal overflows narrow phones.** `ModalBody` defaults to `w-[384px]`, and this modal does not apply the shared `MODAL_MAX_WIDTH_VIEWPORT` constraint used by other responsive modals. On 320-375 px viewports it exceeds the viewport. Apply `max-w-[90vw]` through the shared constant.

The current CI failures are explained by the acknowledged unpublished extensions dependency; I did not find an independent CI regression in those logs. The dependency still needs to be released and pinned before this PR can become mergeable.

**KEY INSIGHT:**
The filter engine is clean, but the complete product contract spans URL navigation and both local and Cloud skill-loading paths; the current tests mainly prove the local happy path.

**[RISK ASSESSMENT]**
This changes shared `/skills` behavior and URL state across responsive layouts. The local core is well tested, but Cloud categorization, history navigation, and active multi-select counts currently behave incorrectly.

**VERDICT:**
❌ **Needs rework**: Address the Cloud data path and URL/count correctness before merge; the modal width fix is straightforward.

>
>
>

enyst:
> @enyst please see my comment at the top of this PR where I explain why ci tests will not work on GitHub (but do locally).
> 
> To repeat, it's because it depends on a change in the extensions repo. Since that new version isn't released yet, it can't be pulled in so the pipeline fails.

Sorry about that, @sleeyax , I went through too fast, my bad! 🙏 

Please let me know what you think of my agent's review findings, and note that it is a bit prone to exagerate!

sleeyax:
HUMAN:

No worries! 🙏 Went through all four:

AGENT:

**Fixed:**
- Modal width — good catch, especially since that modal is the only way into the rail below `md`. Applied the shared `MODAL_MAX_WIDTH_VIEWPORT`.
- History `q` — real bug. The input seeded from the URL once at mount, so back/forward left it stale and the debounce then wrote that stale value back over the historical entry. Fixed with a URL→input sync guarded by a ref of the last value we wrote (otherwise every keystroke bounces). New route test covers it.

**Pushing back on two:**
- Facet counts: excluding a group's own selection is the standard disjunctive count (Algolia, Amazon, etc.) — `writing: 4` means "4 writing skills under your *other* filters", which is what people navigate by. But you're right that my comment claimed it "predicts the result of clicking it", which only holds before the group has a selection. Reworded the comment; behaviour unchanged.
- Cloud categories: this is the intended degradation, not a gap. Cloud skills carry no `category`, so they all fall to `other`, and `buildGroup` drops any group with <2 discriminating values — the facet just doesn't render. Nothing is mis-filtered. Adding `category` to the cloud response is a backend contract change, and inferring it from the bundled catalog would guess cloud data from a build-time snapshot that may be a different version.

CI is still just the unpublished OpenHands/extensions#425 dep. Heads up for the bump: extensions 0.12.0 also reshaped `RecommendedAutomation`, so pinning it here will need companion fixes in the automations code — probably its own PR.

smolpaws (commented):
🐾 Hi, I'm **smolpaws** — a small cat-shaped OpenHands agent living on Engel's machine. Engel asked me to take a first pass at this PR to help with his review, focused on **code design** (architecture, reuse of existing patterns). This is my own read, not his; treat it as one more reviewer, and he'll do the human sign-off.

**tl;dr:** this is a well-architected PR. The design is above the bar — I have one genuine design question and a few small notes, none blocking.

## What I liked (design)

- **`skill-filter.ts` is a deep module with one source of truth.** `GROUP_DEFS` declares each facet group once — `values / valueOf / selected / withSelected` — and *every* operation (parse, serialize, filter, count, toggle, active-count) is derived by iterating that table. Adding a facet group is a single entry, not a shotgun edit across five functions. This is the right shape for something that will grow.
- **The module is pure and i18n-free** (rows carry `labelKey`s, not translated strings), so the 20 unit tests need no translation setup and read as a spec of the behavior. Nice separation of policy from presentation.
- **Correct disjunctive-facet counting.** The `exclude`-the-group's-own-selection trick in `matchesFacets` is the textbook-correct semantics for faceted search ("N skills carry this value under your *other* filters"), and it's the part most implementations get subtly wrong. The comment explaining why is exactly where it needs to be.
- **Type-safety at the trust boundary.** `parseSet` + `narrowSet` rebuild sets from the declared `order` instead of casting, so a hand-edited URL can't smuggle an illegal value into state. `toggleSkillFilterValue` early-returns the same object on an unknown value to avoid a phantom re-render.
- **Good reuse, no reinvention.** It reuses the pre-existing `skill-scope.ts`, and the mobile `SkillFiltersModal` wraps the *same* `SkillFacetRail` behind the shared `ModalBackdrop`/`ModalBody` — so desktop rail and mobile modal can't drift. The old `skills-type-filter-dropdown.tsx` / `skill-type-filter.ts` are fully removed with no dangling references.
- **Comments explain *why*, at the decision point** — group visibility derived from the raw list (so a group can't vanish mid-click), URL echo vs. history navigation, canonical param order for deterministic URLs. This is the good kind of commenting.

## One design question (non-blocking)

The **query lives in two places**: `queryInput` (owned by the input) and `?q` in the URL (mirror), reconciled by two effects plus the `lastWrittenQuery` ref to tell a history navigation apart from the debounced echo of its own write. It's correct and well-commented — but it's by far the most intricate part of the PR, and the facets don't need any of that machinery (they write the URL directly and read back from it).

Did you consider making the URL the single source of truth for the query too — deriving the input from `searchParams` and only debouncing the *write* (or `useDeferredValue` for the filter computation), so `queryInput` + the two sync effects + the ref collapse into one write-side debounce? Not asking you to change it if you already went down that road and hit a wall (controlled-input caret jump on debounced writes is the usual reason) — mostly want to understand the trade-off you landed on, since everything *else* in the PR treats the URL as the state.

## Small notes (take or leave)

- `handleFilterChange` compares facets by re-serializing both states to strings on every toggle. Fine at this size; a direct facet-set comparison would read a touch more directly.
- `matchesQuery` scans `content` (potentially large) across all skills on every keystroke. Fine for ~58 skills; just flagging it as the thing to watch if the catalog grows or search moves server-side.
- The `filter` memo re-parses `searchParams` whenever `queryInput` changes (the parse result is then overridden for `.query`). Negligible, but it's a symptom of the query duality above.

Verdict from me: **approve-leaning**, pending your thoughts on the query-state question. Really solid work — the facet model in particular is cleaner than most of what I've seen for this kind of UI. 🐾

enyst (approved):
Thank you!

PR #16127, opened 2026-07-28
feat: add a domain-neutral extension-manifest host

### LINKED ISSUE #15783: [Feature] Create an Automation UI Form
## Architecture-aligned scope

This issue tracks the generic Agent Canvas foundation needed by extension-owned interfaces. The implementation must render extension definitions rather than introducing automation-specific configuration in Canvas.

The draft UI-extension work in OpenHands/agent-canvas#1659 is relevant source material; this issue targets only the built-in declarative host.

## Acceptance criteria

- Implement a domain-neutral registry, renderer, and allowlisted action bridge for routes, navigation, pages, forms, and resource views described by extension manifests.
- Keep automation identifiers, copy, schemas, API paths, and analytics stages out of Canvas; emit manifest-declared analytics through the consent-aware telemetry service and demonstrate the renderer with a non-automation fixture.

## Dependencies

- Blocked by OpenHands/extensions#420.

---

## Original context

Placeholder ticket for creating a UI modal for creating an Automation.

hieptl:
HUMAN:

Verified end-to-end in a real browser against a live `npm run dev` stack (agent-server on :18000, automation backend on :18001, ingress on :8000), not only through unit tests. Commands, observations, and screenshots are in **How to Test** below.

AGENT:

---

## Why

Agent Canvas has no way to *configure* an automation. Clicking a recommended automation card hands a
slash command to an agent to interpret — `src/components/features/automations/recommended-automations-launcher.tsx:93`
passes the catalog's literal `"/pr-reviewer:setup"` into a new conversation, and the agent then guesses
the repository, trigger, tone, and polling frequency. Run it twice and you can get two different
automations.

`OpenHands/extensions#423` (merged, `f887f19`) defines the fix: an optional `setup` block on each
catalog entry describing what to ask the user. `OpenHands/automation#270` (merged, `74da228`) adds the
two endpoints that block needs — `GET /v1/capabilities` and `POST /v1/validate`. This PR is the Canvas
half: the host that renders a `setup` block and derives everything else.

The contract changed substantially during review of `extensions#423`, around one principle — *anything
derivable is derived by the host, not declared per automation*. So the entry states only `mode`, the
form fields, and the two strings that cannot be read off the form (`prompt`, and an event `filter`).
The route, the endpoints, the request body, the review screen, the error mapping, the assisted setup
command, and the analytics stages are all generated here.

## Summary

- Add `src/manifests/` — a registry with a host-owned admission policy (supported `setup.version`,
  no markup in copy, known placeholder namespaces, plus the invariants the host's own derivation needs:
  one trigger kind to read, a repository field to take an event source from, and field names unique
  across the form), with form modelling, local validation, templating, and payload-path → field error
  mapping.
- Add `src/manifests/automation-setup.ts` — the single module carrying automation knowledge: the create
  endpoint, the trigger properties, and the derivations for the create body, the preflight envelope,
  the error map, and an assisted entry's seed message. It is a direct port of
  `extensions/tests/test_automation_setup.py::_render_payload`, which is the authoritative reference.
- Add `src/components/features/manifest/` — the setup experience as a modal dialog running capabilities
  → prerequisites → form + local validation → preflight → review → create or conversation, mounted at
  `/automations/new/:automationId`; plus four typed analytics functions and three setup transport
  methods on `automation-service.api.ts`.

## Issue Number

Resolves [https://github.com/OpenHands/OpenHands/issues/15783](https://github.com/OpenHands/OpenHands/issues/15783).

Contract from [OpenHands/extensions#423](https://github.com/OpenHands/extensions/pull/423) and
[OpenHands/automation#270](https://github.com/OpenHands/automation/pull/270), both merged.

## How to Test

```bash
npm run lint     # typecheck + eslint + prettier
npm test         # 525 files / 4058 tests
```

**There is no browser walkthrough in this PR, and it is not possible yet.** `@openhands/extensions` is
still published at `0.11.0`; the merged catalog has not been released, so no entry carries a `setup`
block and the registry is legitimately empty:

```bash
node -e "const m=require('@openhands/extensions/automations');
  console.log(m.AUTOMATION_CATALOG.length, m.AUTOMATION_CATALOG.filter(e=>e.setup).length)"
# 8 0
```

`/automations/new/<id>` therefore 404s for every id today, which is the correct behaviour for an id no
entry claims. Nothing needs wiring when the release lands: `manifest-sources.ts` already reads
`AUTOMATION_CATALOG`, so bumping the pin registers the three entries with no code change.

**What is verified instead is the part that carries the risk.** `CreatePromptAutomationRequest` is
`extra="forbid"`, so a wrong request body is a hard 422 rather than a dropped field.
`__tests__/manifests/automation-setup.test.ts` holds the three published `setup` blocks and pins the
derived bodies against the request bodies in `extensions/tests/fixtures/automations/*.json` — the ones
that were replayed against the live service in `automation#270`. It also asserts all three pass this
host's admission policy, which is stricter than the published schema.

**Those tests bite.** Each mutation was run and turns the contract test red:

| Mutation | Result |
| --- | --- |
| `repos[]` entry loses its derived `provider` | 3 failed \| 6 passed |
| an event trigger loses its derived `source` | 1 failed \| 8 passed |
| `cron` drops `timezone` from its trigger properties | 3 failed \| 6 passed |
| *(restored)* | 9 passed |

## Video/Screenshots

https://github.com/user-attachments/assets/7c62da94-ce93-4790-9301-da1ade0d272d

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

---

| Component | Value |
|---|---|
| **Commit** | `605900110dd2d1c9134703692ebeaabed0fa72f0` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-6059001
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-6059001
```

```
ghcr.io/openhands/agent-canvas:sha-6059001-amd64
ghcr.io/openhands/agent-canvas:hieptl-oss-5278-amd64
ghcr.io/openhands/agent-canvas:pr-16127-amd64
ghcr.io/openhands/agent-canvas:sha-6059001-arm64
ghcr.io/openhands/agent-canvas:hieptl-oss-5278-arm64
ghcr.io/openhands/agent-canvas:pr-16127-arm64
ghcr.io/openhands/agent-canvas:sha-6059001
ghcr.io/openhands/agent-canvas:hieptl-oss-5278
ghcr.io/openhands/agent-canvas:pr-16127
```

- Each tag (e.g., `sha-6059001`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-6059001-amd64`) are also available if needed


enyst (commented):

🟢 **Good taste**

The host keeps extension manifests declarative, validates them at a clear trust boundary, derives request payloads in one place, and preserves the existing conversation-based fallback for catalog entries without setup blocks. The focused tests exercise admission invariants, payload derivation, routing, local repository fallback, and the setup flow; the same-organization `@openhands/extensions` 0.12.0 release is pinned with matching registry integrity and npm provenance.

**[RISK ASSESSMENT]**
This is user-visible automation creation driven by package-published data and backend capability/preflight contracts, so regressions could create misconfigured automations. The closed manifest vocabulary, fail-closed admission, authoritative service validation, pinned first-party dependency, and focused regression coverage keep that risk controlled.

**VERDICT:**
✅ **Worth merging**: The core boundaries and fallback behavior are sound; I found no actionable defect on this head.

**KEY INSIGHT:**
The clean design choice is that manifests describe variation while the host retains ownership of validation, requests, navigation, and side effects.

enyst (commented):

🟢 **Good taste**

The host keeps extension manifests declarative, validates them at a clear trust boundary, derives request payloads in one place, and preserves the existing conversation-based fallback for catalog entries without setup blocks. The new exact-head changes add useful dialog, preflight, disabled-skill, and payload-derivation coverage without weakening those boundaries. The same-organization `@openhands/extensions` 0.12.0 release is pinned with matching npm integrity and repository provenance.

**[RISK ASSESSMENT]**
This is a broad, user-visible automation-creation path driven by package-published data and backend capability/preflight contracts. The closed manifest vocabulary, fail-closed admission, authoritative service validation, pinned first-party dependency, and focused regression coverage keep that risk controlled.

**VERDICT:**
✅ **Worth merging**: The core boundaries and fallback behavior are sound; I found no actionable defect on this head.

**KEY INSIGHT:**
The clean design choice is that manifests describe variation while the host retains ownership of validation, requests, navigation, and side effects.

neubig (commented):
Hey @hieptl I think this generally looks well-architected to me, do you have screenshots or a video?

hieptl:
> Hey @hieptl I think this generally looks well-architected to me, do you have screenshots or a video?

Hello @neubig,

I have attached a demo video to the pull request description for your reference.

Thank you very much! 🙏

src/api/automation-service/automation-service.api.ts
```diff
@@ -479,6 +485,86 @@ class AutomationService {
...
+
+  /**
+   * Validate a draft without creating it. An invalid draft is a 200 carrying
+   * field-addressed errors; only a malformed envelope is a 4xx.
+   */
+  static async validateDraft(
+    body: SetupRequestBody,
+  ): Promise<ValidateDraftResponse> {
+    const active = getActiveBackend().backend;
+    const path = `${AUTOMATION_BASE_PATH}/v1/validate`;
```
malhotra5, line 524: NIT: is there a fallback in the event the automation server has not been updated to expose this API? E.g if we accidentally bump the canvas version in cloud without bumping the automation BE version? 

It could be out of scope for this PR, maybe in the future we can try to formalize what compatible versions and fallbacks in the case the micro services are not compatible

src/manifests/types.ts
```diff
@@ -0,0 +1,151 @@
...
+
+export interface SetupRequestBody {
+  [key: string]: SetupPayloadValue;
+}
+
+/** Form values are collected as strings; the payload mapping shapes them. */
+export type SetupFormValues = Record<string, string>;
+
+/** `GET /v1/capabilities` — what this deployment supports. */
+export interface DeploymentCapabilities {
```
malhotra5, line 126: Do the capabilities also support plugin based automations? Will it be a followup if not?

src/components/features/manifest/manifest-setup-dialog.tsx
```diff
@@ -0,0 +1,397 @@
...
+
+  const resolveFieldError = (name: string): string | undefined => {
+    const local = localErrors[name];
+    return local ? formatFieldError(local, t) : serviceErrors.fieldErrors[name];
+  };
+
+  const isLoading = capabilities.isLoading || prerequisites.isLoading;
+
+  const title = (() => {
+    if (isUnsupported) return t(I18nKey.SETUP$UNAVAILABLE_TITLE);
```
malhotra5, line 230: It seems webhooks are disabled on local deployment from the demo video. What about the case where I setup ngrok tunnel to my local deployment? Reason I ask is because we want to encourage users to setup canvas on a VM somewhere with ngrok, and webhook deliveries would be possible

hieptl: It doesn't work as-is. The `Capabilities` state is derived from the configuration and database and doesn't probe endpoint reachability, so exposing the service through a tunnel doesn't change anything—the entry still appears as **"Not available."**

If `AUTOMATION_WEBHOOK_SECRET` is configured, the UI is unblocked and the automation can be created. However, when a GitHub webhook is sent to the tunnel, it successfully passes signature verification but then returns a `400 Missing payload in builtin source request` error. As a result, the webhook is silently dropped, which is arguably the more problematic failure mode.

The good news is that the required fix is relatively small and isolated to the automation side. GitHub already sends the `X-Hub-Signature-256` header that we verify, and event type detection is based on the payload structure rather than the `X-GitHub-Event` header. In other words, a raw GitHub payload would already be parsed correctly. The only remaining blocker is the envelope check. Allowing the built-in receiver to fall back to the raw payload when the `payload` key is absent would make a VM + ngrok + secret configuration fully functional without requiring a forwarder.

I'd be happy to open an issue in the OpenHands automation repository to track this improvement. That said, I believe it's outside the scope of this PR, which is focused solely on the Canvas-side host.

Thank you very much! 🙏

src/manifests/types.ts
hieptl: Yes—`presetPlugin` is already included in `Capabilities`, and it's reported unconditionally. It lives in `_STATIC_FEATURES` alongside `presetPrompt`, `repoClone`, `mcpTools`, and `conversationDispatch`, since these capabilities are provided by the SDK packaged into a run rather than being configuration-driven. As a result, every ready deployment reports them. Preflight support is already in place as well: `_DRAFT_MODELS` maps both `/v1/preset/prompt` and `/v1/preset/plugin`, and the endpoint is included as part of the `/v1/validate` request, so plugin drafts validate today without requiring any server-side changes.

The missing piece is on the host side. `AUTOMATION_CREATE_ENDPOINT` is currently hardcoded to `/v1/preset/prompt`, and `buildCreatePayload` only produces a payload for `mode: "direct"` entries with a prompt. This means an entry could declare `requires.features: ["presetPlugin"]` and successfully pass the capability check, but the host would still generate a prompt preset request. None of the three currently published setup flows use plugins, so this isn't causing any functional issues today—it's simply functionality that hasn't been implemented yet.

So I agree this is a good candidate for a follow-up. The change should be fairly self-contained because the automation-specific logic is intentionally centralized in `src/manifests/automation-setup.ts`. At a high level, the manifest contract would need a way to specify which preset type an entry targets, the create endpoint would become a function of that selection instead of a constant, and the request payload would be extended to include `plugins`, `variants`, and `experiment_id` where applicable. The error mapping, preflight envelope, review screen, and analytics should all continue to work automatically, since they're derived from the payload rather than explicitly defined.

I'm happy to open an issue to track this follow-up.

Thank you very much! 🙏

malhotra5 (approved):
LGTM, thanks

neubig (approved):

src/api/automation-service/automation-service.api.ts
hieptl: That's a good question. Yes, both new endpoints are intentionally designed to fail open. Here's how the older-backend scenario behaves throughout the flow:

For **`GET /v1/capabilities`**, the request is made with `retry: false` and `meta: { disableToast: true }` (`use-manifest-capabilities.ts:38,42`). As a result, a `404` does not trigger a toast notification, and the absence of a response is interpreted as `supported: "unknown"` rather than `false` (`use-manifest-capabilities.ts:54-61`). The dialog blocks only when the value is explicitly `false` (`manifest-setup-dialog.tsx:135`), so if the backend cannot answer the capability query, the UI proceeds using the manifest's default behavior. In other words, a backend that cannot be queried is intentionally not treated the same as one that has explicitly responded "no."

For **`POST /v1/validate`**, `404` and `501` responses are treated as "not implemented" and simply return no validation verdict. Other failures also return no verdict, but are logged first so that genuine validator failures remain visible (`use-manifest-preflight.ts:23,76-84`). In both cases, `handleContinue` proceeds because it only blocks when the validation result explicitly contains errors (`manifest-setup-dialog.tsx:177-181`). Both scenarios are covered by `__tests__/hooks/use-manifest-preflight.test.tsx`.

It's also worth noting that the create call itself is not new. `AUTOMATION_CREATE_ENDPOINT` resolves to `/v1/preset/prompt` (`automation-setup.ts:28`), which is the same endpoint that `createAutomation` has used since before this PR (`automation-service.api.ts:167`). Additionally, every field included in the derived request body—`name`, `prompt`, `repos[].url/ref/provider`, and `trigger.type/on/source/schedule/timezone/filter`—already exists in the current schema defined in `src/types/automation.ts`. As a result, the one required and authoritative operation does not depend on the new automation endpoints introduced in this change.

The practical effect of Canvas being ahead of the automation backend is therefore limited: the capability check and the preflight validation—both advisory features—are unavailable, while the setup flow continues successfully through the existing creation endpoint. What is lost is the opportunity to surface a more helpful error earlier in the process, not the ability to create an automation.

Regarding formalizing compatibility, I agree that it would be valuable. That said, I would prefer to keep it separate from this PR. We already have a precedent with `config/defaults.json`, which defines `compatibility.minimumAgentServer` and is consumed by `src/api/agent-server-compatibility.ts`. However, there is currently no equivalent mechanism for automation. Introducing a `minimumAutomation` version requirement together with a compatibility probe would affect all automation API calls, rather than just the three touched in this PR.

I'd be happy to open a follow-up issue to track that work if you think it would be useful.

Thank you very much! 🙏

PR #16135, opened 2026-07-28
fix(frontend): stop streaming re-render storm (#12681)

eliphatfs:
HUMAN:

Migrated from https://github.com/OpenHands/agent-canvas/pull/1822

I tested this by building it into a deployed agent-canvas image and running a long streaming session. The UI stays responsive throughout the stream. Previously, the window freezes when the LLM is streaming thinking block contents at a fast pace.

AGENT:

---

## Why

During token streaming (logs / terminal output), every store merge produces a **fresh `events` array reference**, so every chat card re-renders and **re-parses its markdown on every single token**. The cost grows with conversation length, surfacing as high client-side GPU (~30%) and UI freeze — the symptoms in OpenHands #12681 https://github.com/OpenHands/OpenHands/issues/12681.

## Summary

- Minimal, dependency-free memoization so **only the live streaming bubble re-renders per token**; all prior cards stay referentially stable.
- `EventMessage`: `React.memo` with a custom comparator that compares the event ref, live-state flags, `planPreviewEventIds` ref, `suppressThought`, and `messages.length` (not the array ref).
- `usePlanPreviewEvents`: split into two `useMemo`s keyed on a stable string of collected ids, so the returned `Set` keeps identity across streaming tokens.
- `MarkdownRenderer`: hoist static default components + remark/rehype plugins to module scope; `useMemo` the combined `components`/`rehypePlugins`; wrap in `React.memo`.
- `ChatMessage`: wrap in `React.memo` (shallow compare).

## Issue Number

#12681

## How to Test

1. `npm install && npm run make-i18n && npm run build` (or `npm run dev`).
2. Open a conversation and run an agent that streams a long log / terminal output (e.g. a step that emits many tokens).
3. With DevTools → Performance (or the OS GPU monitor), observe during streaming:
   - **Before:** GPU spikes ~30% and the UI janks/freezes; cost grows with conversation length.
   - **After:** only the streaming bubble re-renders; prior cards do not re-parse markdown; GPU stays flat and the UI stays responsive.
4. Sanity: scroll back through a long conversation mid-stream and confirm cards render correctly and no message content is lost.

## Video/Screenshots

_(Optional — a short screen recording of a long streaming session showing flat GPU / responsive UI vs. the pre-fix freeze would strengthen this. Not required for a pure perf fix.)_

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Pure frontend rendering changes only — no backend, API, websocket, or build-config changes.
- No new dependencies; uses `React.memo` / `useMemo` only.
- The custom `EventMessage` comparator deliberately compares `messages.length` rather than the array ref (which changes every token) — this is the core of the fix.
- This PR was created by an AI agent (OpenHands) on behalf of the maintainer.

### DIFF
--- a/src/components/conversation-events/chat/event-message.tsx
+++ b/src/components/conversation-events/chat/event-message.tsx
@@ -134,7 +134,22 @@ const renderUserMessageWithSkillReady = (
   }
 };
 
-export function EventMessage({
+const areEventMessagePropsEqual = (
+  prev: Readonly<EventMessageProps>,
+  next: Readonly<EventMessageProps>,
+): boolean =>
+  prev.event === next.event &&
+  prev.isLastMessage === next.isLastMessage &&
+  prev.isInLast10Actions === next.isInLast10Actions &&
+  prev.planPreviewEventIds === next.planPreviewEventIds &&
+  prev.suppressThought === next.suppressThought &&
+  // Compare length, not the array ref: the ref changes every streaming
+  // token, but length only changes when a genuinely new event arrives —
+  // which is exactly when observation cards (which do `messages.find(...)`
+  // below) must re-render.
+  prev.messages.length === next.messages.length;
+
+export const EventMessage = React.memo(function EventMessage({
   event,
   messages,
   isLastMessage,
@@ -344,4 +359,6 @@ export function EventMessage({
   return (
     <GenericEventMessageWrapper event={event} isLastMessage={isLastMessage} />
   );
-}
+}, areEventMessagePropsEqual);
+
+EventMessage.displayName = "EventMessage";
--- a/src/components/conversation-events/chat/hooks/use-plan-preview-events.ts
+++ b/src/components/conversation-events/chat/hooks/use-plan-preview-events.ts
@@ -85,23 +85,33 @@ export interface PlanPreviewEventInfo {
  * @returns Set of event IDs that should render PlanPreview
  */
 export function usePlanPreviewEvents(allEvents: OpenHandsEvent[]): Set<string> {
-  return useMemo(() => {
-    const planPreviewEventIds = new Set<string>();
-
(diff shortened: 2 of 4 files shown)


enyst (commented):

🟡 **Acceptable** - The memoization direction is sound, but the performance guarantee has no regression test.

**[TESTING GAPS]**
- [src/components/conversation-events/chat/event-message.tsx, Line 137] **No render-count regression test**: This custom comparator is the load-bearing fix and is also correctness-sensitive: omit a prop and old cards can go stale; compare too much and the render storm returns. The existing markdown behavior tests do not prove that a fresh streaming events array re-renders only the live tail. Add a focused `Messages`/`EventMessage` rerender test with a stable prior event and a fresh same-length events array, then assert the prior markdown render count stays unchanged while the live event updates.

**[RISK ASSESSMENT]**
This changes render invalidation in the core conversation UI. The implementation is localized and the human runtime evidence is useful, but an incorrect equality rule can silently produce stale chat content, so the missing regression guard raises the risk above low.

**VERDICT:**
❌ **Needs rework**: Add automated coverage for the exact render-suppression behavior before relying on this custom comparator long-term.

**KEY INSIGHT:**
The optimization is only durable if the test suite proves both halves of the invariant: historical cards stay stable and the live card still updates.

---

>
>
>

VascoSch92:
Thanks for taking the time to put this together, and apologies for the slow reply!

We're closing this one because the underlying issue (#12681) was closed as *not planned* — so there isn't a change we're looking to make here any more. That's a decision about the issue itself, not a reflection on your patch.

Sorry for the wasted effort. If you disagree with that call, the best place to make the case is on #12681 — and if it gets reopened, we'll happily revisit this PR.

PR #16136, opened 2026-07-28
fix: send X-Session-API-Key header and downgrade workspace-session cookie over HTTP

eliphatfs:
HUMAN:

The agent claims to live rebuild agent canvas into the running build. The file panel works after that.
The issue is related to a deployment over VPN like wireguard without any TLS termination proxy, where secure cookies are discarded.
Migrated from https://github.com/OpenHands/agent-canvas/pull/1829

AGENT:

This PR was created by an AI agent (OpenHands) on behalf of eliphatfs.

## Why

The right-panel Files tab could not read file contents: every file showed
`Failed to read <file> (e.g. CHANGELOG.md): 401`, and the network response body
was `{"detail":"Unauthorized"}`.

Root cause: the agent-server mints the workspace static-fileserver session cookie
`oh_workspace_session_key` with `HttpOnly; Secure; SameSite=none; Partitioned`.
Browsers **silently drop** a `Secure` cookie on a **plain-HTTP** origin — and a
local dev / static stack is served over HTTP, not HTTPS. So the file viewer's
text fetch (and iframe / `` embeds) carried no credential, and the fileserver
returned `401 Unauthorized`. The cookie worked in production (HTTPS) but was
invisible to every local stack.

Verified the chain live against the running agent-server (`:18000`):
- file read with **no** credential → `401` ✅ (reproduces the bug)
- file read with **only** the `X-Session-API-Key` header → `200` ✅ (the backend
  already accepts header auth — no backend change needed)
- file read with **only** the cookie → `200` ✅

## Summary

- **Frontend header auth**: `useWorkspaceFileContent` text-fetch path now sends
  `X-Session-API-Key` (resolved from the conversation's `session_api_key` → the
  active backend's `apiKey`), alongside `credentials: "include"`. This is the
  load-bearing fix for text reads — it authenticates via header even when the
  `Secure` cookie is dropped over HTTP, while `credentials: "include"` stays as
  a belt-and-suspenders fallback for HTTPS deployments where the cookie travels.
- **Proxy cookie downgrade**: `scripts/proxy-utils.mjs` gained
  `downgradeSecureCookie()` + `installSecureCookieDowngrade()`. The ingress /
  static-server proxies rewrite the upstream `Set-Cookie` to strip
  `Secure` / `Partitioned` / `SameSite=none` and add `SameSite=Lax` **only when
  the client connected over plain HTTP** (`req.socket.encrypted` guard — a strict
  no-op over TLS, so production HTTPS keeps the server's `Secure` posture). Wired
  in via `createProxyHandlers({ downgradeSecureCookieOverHttp: true })` (default
  on). This covers embeds that can't set headers (iframe / `` preview).
- No backend / `software-agent-sdk` change is required — the agent-server already
  accepts `X-Session-API-Key`. The fix is entirely contained in this repo.

## Issue Number

N/A — no tracked issue. Discovered while running the local stack.

## How to Test

Reproduced and verified end-to-end against the live stack (ingress on `:6768`,
agent-server on `:18000`, static frontend on `:3001`):

1. **Reproduce the bug (before fix):** open the right-panel Files tab against a
   plain-HTTP local stack and click a file → `Failed to read <file>: 401`. The
   `Set-Cookie` from the agent-server carries `Secure`, so the browser discards
   it over HTTP and the file read has no credential.
2. **After fix — cookie downgrade at the proxy:**
   ```sh
   curl -sS -i http://localhost:6768/api/auth/workspace-session \
     -H "X-Session-API-Key: $KEY" | grep -i 'set-cookie'
   ```
   The proxied response's `Set-Cookie` no longer has `Secure` / `Partitioned` /
   `SameSite=none`; it reads `SameSite=Lax`, so the browser actually stores it.
3. **After fix — header-based file read:**
   ```sh
   curl -sS -o /dev/null -w '%{http_code}\n' \
     -H "X-Session-API-Key: $KEY" \
     "http://localhost:6768/api/conversations/<id>/files/<path>"
   ```
   → `200` (header auth alone succeeds).
4. **No-auth still 401 (expected):**
   ```sh
   curl -sS -o /dev/null -w '%{http_code}\n' \
     "http://localhost:6768/api/conversations/<id>/files/<path>"
   ```
   → `401` (unchanged; nothing weakened).
5. **UI smoke test:** load the app in a browser over HTTP, open a conversation,
   switch to the Files tab, and click any source file (e.g. `CHANGELOG.md`).
   The file content renders instead of the `Failed to read … 401` error.

I deployed both fixes to the live stack (`/opt/agent-canvas/build/` rebuilt +
`/opt/agent-canvas/scripts/proxy-utils.mjs` synced; static-server `:3001` and
ingress `:6768` restarted) and confirmed the file viewer reads files correctly.

## Video/Screenshots

Not attached. Live verification was performed via `curl` against the running
stack (commands + expected status codes above) rather than a screen recording.
The visible behavior change is: the Files-tab file content renders where it
previously showed `Failed to read <file>: 401`.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Scope is agent-canvas-only.** No change to `software-agent-sdk`
  (agent-server) or `@openhands/typescript-client` is required — the backend
  already accepts `X-Session-API-Key`. The `Secure` cookie flag is intentionally
  left intact server-side (correct for production HTTPS); only the local-HTTP
  proxies downgrade it.
- **The downgrade is HTTP-only and never global.** `installSecureCookieDowngrade`
  returns early when `req.socket?.encrypted` is set, so HTTPS deployments are
  untouched. The injection point is httpxy's `proxyRes` event, which fires before
  `webOutgoingMiddleware` copies headers onto the downstream response.
- **Why both fixes:** the header path is load-bearing for text reads (the hook
  can set headers); the cookie downgrade covers iframe / `` embeds that
  cannot set headers and rely on the browser cookie jar.
- Test infra note: the cookie-downgrade integration test uses Node's
  `http.request()` (not jsdom `fetch`) because jsdom's `fetch` implementation
  filters out `set-cookie` response headers, which would hide the rewrite.
- Verification commands run: `npx tsc --noEmit -p tsconfig.json` (exit 0),
  `npx eslint` on the changed files (exit 0), `npx vitest run` across 86 files /
  790 tests (all pass), and `npm run build` (success). These were run on the
  `fix/workspace-cookie-401-http` branch (base `main`) and re-confirmed after
  merging into `dev`.

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 1.12s · commit 1f041ed  
**Strongest signal:** Weakened authentication · 23% estimated likelihood.  
**Evidence:** No direct hunk selected.  
**Coverage:** complete supplied coverage; 10/10 hunks, 5/5 files.

<details>

| --- | --- | --- |
| Command injection | 4.0% | No direct hunk selected |
| Weakened authentication | 23.0% | No direct hunk selected |
| Weakened authorization | 13.0% | No direct hunk selected |
| Contract regression | 13.0% | No direct hunk selected |
| Sensitive data disclosure | 7.0% | No direct hunk selected |
| Credential misuse | 19.0% | No direct hunk selected |
| Unverified remote execution | 4.0% | No direct hunk selected |
| Security assessment bypass | 27.0% | [F004H001 · scripts/proxy-utils.mjs:35–118](https://github.com/OpenHands/OpenHands/blob/1f041edc31c3bf991cf7d31b86c32c7a1344bcdf/scripts/proxy-utils.mjs#L35-L118) |
| Primary concern | Weakened authentication; confidence 50.0% | No direct hunk selected |

</details>


enyst (commented):

🔴 **Needs improvement** - The text-fetch fix is reasonable, but the cookie workaround weakens every proxied secure cookie instead of only the workspace-session cookie.

**[CRITICAL ISSUES]**
- [scripts/proxy-utils.mjs, Line 112] **Security Boundary**: This listener rewrites every `Set-Cookie` header returned through `createProxyHandlers`, and that helper is enabled by default for the ingress default backend plus every configured `/api/*` route. A plain-HTTP deployment would therefore strip `Secure`/`Partitioned` from unrelated current or future auth cookies, making them available over cleartext and changing their cross-site isolation. Scope the rewrite to `oh_workspace_session_key` and the workspace-session response/path (or install the listener only on that route), and add a regression test proving an unrelated `Secure` cookie passes through unchanged.

**[RISK ASSESSMENT]**
The PR deliberately changes cookie security attributes on an authentication path, and the current implementation applies that downgrade across a generic reverse proxy with a broad blast radius.

**Recommendation**: Do not auto-merge. Request review from a human security reviewer to validate the cookie name/path scoping and the intended trust model for plain-HTTP deployments.

**VERDICT:**
❌ **Needs rework**: Constrain the downgrade to the one workspace-session cookie before merging.

**KEY INSIGHT:**
A narrowly justified insecure-development exception must not become a generic policy for every cookie traversing the proxy.

---

>
>
>

PR #16152, opened 2026-07-29
fix(e2e): shut the mock-LLM stack down gracefully so it stops leaking processes

### LINKED ISSUE #16151: Mock-LLM E2E leaves the whole stack running after every local run, and the next run silently adopts it
## Summary

Every local `playwright.mock-llm.config.ts` run leaves the agent-server, the automation backend, the static frontend, and the ingress proxy alive, orphaned to PID 1 and still holding ports 18000, 18001, 18300, and 3001. Because the same config sets `reuseExistingServer: !process.env.CI`, the next local run finds the ingress answering and adopts those survivors instead of starting its own. The state-dir wipe lives inside the webServer `command`, so adopting a survivor skips it, and the run executes against the previous run's backend state and session key.

The visible result is a failure in a shared helper that has nothing to do with the change under test. `openAgentProfileEditor` cannot find the seeded `default` agent profile, which reads like selector drift and sends you looking at recent settings PRs.

CI is unaffected, since `reuseExistingServer` is false there and every job gets a fresh runner.

The cause is a single missing declaration, and the repo already contains the machinery to handle it. Details and a proposed one-field fix below.

## Impact

Local contributors only. Two shapes, depending on which survivors are still healthy:

- The orphaned ingress still answers, so the next run adopts the stale stack and fails somewhere unrelated. This is the quiet one and it costs the most time.
- The ingress is not answering, so Playwright runs the command, and `assertPortsFree` aborts with `Cannot start: the following ports are already in use ... Another agent-canvas instance may already be running`. That message is accurate but points at the user rather than at the previous run.

Once one run leaks, every later run in that working copy is affected until the ports are cleared by hand. The survivors do not self-terminate: in two consecutive clean runs, an identical set of six processes was still holding all four ports 40 minutes after Playwright exited, with no other actor on the host, and a separately leaked set was observed alive at ~4h47m etime.

## Reproduction

Deterministic on a clean checkout of `main`.

```bash
npm ci
uv venv .mock-llm-venv
uv pip install -p .mock-llm-venv "openhands-sdk==1.38.0" "agent-client-protocol<0.11"

MOCK_LLM_PYTHON=.mock-llm-venv/bin/python3 \
  npx playwright test --config=playwright.mock-llm.config.ts \
  tests/e2e/mock-llm/settings/mock-llm-acp-agent.spec.ts

# after playwright has exited:
ss -ltnp | grep -E ':(18000|18001|18300|3001)\b'
ps -eo pid,ppid,pgid,args | grep -E 'uvx|static-server|ingress\.mjs'
```

That run passes (`4 passed (56.3s)` here), and 13 seconds after Playwright exits the stack is still running with 18000, 18001, 18300, and 3001 still bound. The leak follows a green run, so it is not a side effect of a failing test.

Then run any mock-llm spec again. It either aborts on `assertPortsFree`, or it adopts the survivors and fails inside `openAgentProfileEditor`.

## Evidence

Process tree while the run is live. `bin/agent-canvas.mjs` is Playwright's tracked child, and every service it starts is its own process-group leader:

```
  PID    PPID   PGID  COMMAND
1952803 1952644 1952803  node --env-file-if-exists=.env bin/agent-canvas.mjs
1952843 1952803 1952843  uv tool uvx --from openhands-agent-server==1.38.0 ...
1952977 1952843 1952843  python .../agent-server --host 127.0.0.1 --port 18000
1953088 1952803 1953088  uv tool uvx --from openhands-automation==1.4.1 uvicorn ...
1953089 1952803 1953089  node scripts/static-server.mjs --dir build --port 3001
1953175 1953088 1953088  python .../uvicorn openhands.automation.app:app --port 18001
```

The same processes 13 seconds after Playwright exited. The parent is gone. Six processes survive, four of them reparented to PID 1 and the other two still children of those orphans, and four ports are still bound:

```
  PID    PPID   PGID  COMMAND
1952843       1 1952843  uv tool uvx --from openhands-agent-server==1.38.0 ...
1952977 1952843 1952843  python .../agent-server --host 127.0.0.1 --port 18000
1953088       1 1953088  uv tool uvx --from openhands-automation==1.4.1 uvicorn ...
1953089       1 1953089  node scripts/static-server.mjs --dir build --port 3001
1953175 1953088 1953088  python .../uvicorn openhands.automation.app:app --port 18001
1953207       1 1953207  node scripts/ingress.mjs --port 18300 ...

LISTEN 127.0.0.1:18000   LISTEN 127.0.0.1:18001
LISTEN *:18300           LISTEN *:3001
```

## Root cause

Three facts combine.

**1. Playwright force-kills the webServer with SIGKILL, not SIGTERM.** In `playwright/lib/plugins/webServerPlugin.js`, `attemptToGracefullyClose` throws immediately unless the webServer entry declares `gracefulShutdown`:

```js
if (!this._options.gracefulShutdown)
  throw new Error("skip graceful shutdown");
```

`playwright-core/lib/server/utils/processLauncher.js` then falls through to the force path:

```js
await options.attemptToGracefullyClose().catch(() => killProcess());
...
process.kill(-spawnedProcess.pid, "SIGKILL");
```

SIGKILL cannot be caught, so the stack's own shutdown handler never runs.

**2. The services sit outside the group being killed.** `scripts/dev-process-utils.mjs` spawns them with `detached: process.platform !== "win32"`, which makes each service the leader of a new process group. That is what lets `signalProcessTree` reach a service's own subtree via `process.kill(-proc.pid, ...)`. It also means none of them belong to the parent's group, so Playwright's `kill(-parentPid, SIGKILL)` reaches the parent alone.

**3. The only cleanup path is the handler that SIGKILL skips.** `scripts/dev-with-automation.mjs` registers `process.on("SIGTERM", shutdown)`, and `shutdown` signals each service, waits 3000 ms, force-kills the stragglers, then exits.

That handler works correctly. Sending SIGTERM to the parent by hand, with no Playwright involved, cleans up completely:

```
t~0.5s parent=YES children_alive=0/3 ports_bound=0
t~3s   parent=no  children_alive=0/3 ports_bound=0
t~10s  parent=no  children_alive=0/3 ports_bound=0

Shutting down...
[agent-server] Stopping...
[automation]   Stopping...
[static]       Stopping...
[ingress]      Force stopping...
```

No survivors, no bound ports. The shutdown logic is fine. It just never receives the signal it is written for.

`playwright.mock-llm.config.ts` already documents this path in two comments, "SIGTERM goes directly to the shutdown handler, which kills children via process groups and exits cleanly" and "Bypass npm, exec directly into node so SIGTERM reaches the shutdown handler (npm swallows it)". The `exec` is there so the tracked PID is the node process. So the shutdown path is built and works; what it does not get is a SIGTERM, because Playwright only sends one when the webServer entry asks for it.

## Proposed fix

Declare `gracefulShutdown` on the webServer entry that owns the stack:

```ts
{
  command: /* ... bin/agent-canvas.mjs ... */,
  url: `http://localhost:${INGRESS_PORT}/api/automation/v1`,
  timeout: 180_000,
  reuseExistingServer: !process.env.CI,
  gracefulShutdown: { signal: "SIGTERM", timeout: 15_000 },
}
```

The timeout needs headroom over the handler's own 3000 ms force-kill pass. 15s is comfortable.

Verified with the same probe used above, changing nothing else:

| | processes after exit | ports bound |
|---|---|---|
| without `gracefulShutdown` | 6 surviving, 4 reparented to PID 1 | 4 |
| with `gracefulShutdown` | 0 | 0 |

Measured at both +3s and +13s after Playwright exited.

Adding the field to all three webServer entries and running the full mock-LLM suite on the same checkout gives the same result, 60 passed and 1 skipped either way, with no ports left bound afterwards. The skipped test is the ACP banner spec skipping itself on a host that has a real provider login.

On teardown cost, the bound comes from the handler rather than from suite timing: it signals each service, waits 3000 ms, force-kills anything left, then exits, and timing that sequence directly puts it at roughly three seconds. Wall clock for the two full-suite runs was 489s and 492s, but that is a single pair on an eight minute suite, so treat it as consistent with the handler bound rather than as a measurement of it.

`playwright.mock-llm-docker.config.ts` carries the same shape and a comment noting that Playwright's SIGTERM has to reach `docker run`, so it likely wants the same field. I have not measured that one, since reproducing it needs an image built from the branch.

## Why the failure looks unrelated

Worth spelling out, because the symptom is what makes this expensive rather than the leak itself.

Isolated with a marker file planted in the state dir, everything else held constant:

| Stale stack | State-dir wipe | Result |
|---|---|---|
| up, ingress answering | skipped, marker survived | failed in `openAgentProfileEditor`, `agent-profile-row` / `span[title="default"]` not found |
| none | ran, marker gone | 1 passed (36.7s) |

In the adopted case Playwright never logs anything about reuse and never prints a port error. The run just quietly talks to the previous run's backend, which still holds that run's profiles and a different generated session key, so the seeded `default` agent profile the helper waits for is not there.

## Environment

- `main` at `2965aca5c`
- Fedora Linux 43, kernel 7.1.4
- Node v22.22.2, npm 10.9.7
- Playwright 1.59.1
- uv 0.10.11, agent-server pin 1.38.0

georgeglarson:
HUMAN:

100% real human used in testing. Worked even when some commands were entered using fingers.

AGENT:

Ran the full mock-LLM suite on this checkout, with and without the change, plus a process-tree probe around teardown. Commands and numbers below.

**Before the change**, 13 seconds after Playwright exits, six processes are still alive, four of them reparented to PID 1:

```
  PID    PPID   PGID  COMMAND
1952843       1 1952843  uv tool uvx --from openhands-agent-server==1.38.0 ...
1952977 1952843 1952843  python .../agent-server --host 127.0.0.1 --port 18000
1953088       1 1953088  uv tool uvx --from openhands-automation==1.4.1 uvicorn ...
1953089       1 1953089  node scripts/static-server.mjs --dir build --port 3001
1953175 1953088 1953088  python .../uvicorn openhands.automation.app:app --port 18001
1953207       1 1953207  node scripts/ingress.mjs --port 18300 ...

LISTEN 127.0.0.1:18000   LISTEN 127.0.0.1:18001
LISTEN *:18300           LISTEN *:3001
```

**After the change**, the same probe reports zero processes and zero bound ports, checked at both +3s and +13s.

**Full suite, same checkout:** 60 passed / 1 skipped either way, and no ports left bound afterwards. The skipped test is the ACP banner spec skipping itself on a host that has a real provider login.

## Why

Playwright force-kills a webServer with `process.kill(-pid, "SIGKILL")` unless the entry declares `gracefulShutdown`. In `webServerPlugin.js`, `attemptToGracefullyClose` throws `"skip graceful shutdown"` when the field is absent, and `processLauncher.js` catches that and goes straight to the force path.

SIGKILL cannot be caught, so the stack never runs its shutdown handler. And because `scripts/dev-process-utils.mjs` spawns each service with `detached: true`, every service leads its own process group and is not in the group Playwright kills. So the parent dies, the services do not, and they orphan to PID 1 holding 18000, 18001, 18300, and 3001.

The next local run then goes one of two ways. Either the ingress is no longer answering and `assertPortsFree` aborts with "Another agent-canvas instance may already be running", or the ingress is still answering and, since `reuseExistingServer` is on outside CI, Playwright silently adopts the survivors. Adoption is the expensive one: the state-dir wipe lives inside the webServer `command`, so a reused stack never gets wiped, and the run executes against the previous run's profiles and session key. That surfaces as an unrelated helper failing to find the seeded `default` agent profile.

CI is not affected, since `reuseExistingServer` is false there and each job gets a fresh runner.

## Summary

- Declare `gracefulShutdown: { signal: "SIGTERM", timeout: 15_000 }` on the webServer entry that runs the agent-canvas stack, so the shutdown handler in `scripts/dev-with-automation.mjs` gets the signal it was already written for.
- The 15s timeout leaves headroom over that handler's own 3000 ms pass before it force-kills stragglers.
- Only this entry needs it. The mock-LLM server and the public-mode static server are single `exec`'d processes with no detached children, and were confirmed not to leak.

## Issue Number

Fixes #16151

## How to Test

```bash
npm ci
uv venv .mock-llm-venv
uv pip install -p .mock-llm-venv "openhands-sdk==1.38.0" "agent-client-protocol<0.11"

MOCK_LLM_PYTHON=.mock-llm-venv/bin/python3 \
  npx playwright test --config=playwright.mock-llm.config.ts \
  tests/e2e/mock-llm/settings/mock-llm-acp-agent.spec.ts

# then, after playwright has exited:
ss -ltnp | grep -E ':(18000|18001|18300|3001)\b'
ps -eo pid,ppid,pgid,args | grep -E 'uvx|static-server|ingress\.mjs'
```

On `main` those two commands list the leftover stack. On this branch they come back empty. The test run itself passes either way, so the difference is teardown only.

## Video/Screenshots

Terminal captures are inline above; the change has no UI surface.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- `playwright.mock-llm-docker.config.ts` has the same shape, and a comment noting that Playwright's SIGTERM has to reach `docker run`, so it may want the same field. Left out here because verifying it needs an image built from the branch, and I would rather not ship a change I have not measured.
- The handler being force-fed SIGTERM is not new behavior. `scripts/dev-with-automation.mjs` already registers it, and the config comments already describe the `exec` chain that exists to deliver it. This change is what asks Playwright to send it.

### DIFF
--- a/playwright.mock-llm.config.ts
+++ b/playwright.mock-llm.config.ts
@@ -154,6 +154,16 @@ export default defineConfig({
       url: `http://localhost:${INGRESS_PORT}/api/automation/v1`,
       timeout: 180_000, // allow extra time for build + agent-server + automation startup
       reuseExistingServer: !process.env.CI,
+      // Without this, Playwright tears the webServer down with
+      // process.kill(-pid, "SIGKILL"), which the stack cannot catch. Its
+      // services are spawned detached (see scripts/dev-process-utils.mjs), so
+      // they sit in their own process groups and survive that group kill,
+      // orphaning to PID 1 while still holding 18000/18001/18300/3001. Asking
+      // for SIGTERM lets the existing shutdown handler in
+      // scripts/dev-with-automation.mjs run, which stops each service, waits
+      // 3s, then force-kills stragglers. The timeout below leaves headroom
+      // over that 3s pass.
+      gracefulShutdown: { signal: "SIGTERM", timeout: 15_000 },
     },
     // 3. Public-mode static server — same build/, same backend, but with
     //    --auth-required (no session key injected). The agent-server's


enyst (approved):
(Opus 5 agent)

🟢 **Good taste** — elegant one-line fix that makes reality match a handler the codebase already had, instead of piling on new teardown machinery.

I verified the mechanism independently against `main`, not just the description:

- `scripts/dev-with-automation.mjs:934-960` — `shutdown()` is registered on **both** `SIGINT` and `SIGTERM`; it SIGTERMs each service's process tree, waits 3000ms, then SIGKILLs stragglers. So the SIGTERM this PR asks Playwright to send lands on a handler that was already written for it.
- `scripts/dev-process-utils.mjs` — services spawn `detached: true` (each becomes its own process-group leader) and are later signaled via `process.kill(-pid, …)`. That's precisely why Playwright's default group-SIGKILL misses them and they orphan to PID 1 holding 18000/18001/18300/3001.
- `exec` in the command means the tracked PID *is* the node process, so `SIGTERM` reaches the handler directly. `timeout: 15_000` clears the handler's own 3s pass with headroom.

Scoping is right: only entry #2 spawns the detached stack. #1 (mock-llm python) and #3 (public-mode static server) are single `exec`'d processes that route into #2's backends — nothing detached to leak.

The 10-line comment is on the verbose side for a one-line change, but it's the justified kind: it encodes a cross-module invariant (why SIGTERM, why 15s vs the 3s handler window) that a future reader cannot infer locally and would otherwise "clean up." Keep it.

Nice catch on the pre-existing config comment that already claimed "SIGTERM goes directly to the shutdown handler" — this PR is what actually makes that true. And deferring `playwright.mock-llm-docker.config.ts` rather than shipping an unmeasured change to it is the correct call.

**[RISK ASSESSMENT]**
- [Overall PR] ⚠️ Risk Assessment: 🟢 LOW — test-infra-only change to a single Playwright `webServer` entry; no product/runtime code touched. It only affects local teardown (`reuseExistingServer` is false in CI, so CI was never leaking), and worst case degrades to the prior force-kill behavior. Evidence is concrete (process-tree + bound-port probes at +3s/+13s, full suite 60 passed/1 skipped both ways).

**VERDICT:**
✅ **Worth merging** — sound root cause, minimal surface, real evidence.

**KEY INSIGHT:**
The bug was a signal never reaching an already-correct shutdown handler; the fix delivers the signal rather than reimplementing the teardown.

PR #16162, opened 2026-07-29
perf: memoize user events check in useFilteredEvents

hi-neason:
HUMAN:

Pulled the branch and reproduced the red test locally before applying the fix. Ran the full vitest suite (`npx vitest run`),4016 pass. Small, memoization-only change; scope is limited to one hook. Happy to add a profiler screenshot if reviewers want proof of the hot-path improvement.

AGENT:

Verified locally with `npx vitest run __tests__/hooks/use-filtered-events.test.ts` (13/13 pass, including the new spy test) and the full suite `npx vitest run` (4016 pass / 5 skipped / 9 todo, exit 0). ESLint on the touched files reports no issues. `tsc --noEmit` shows only one pre-existing unrelated error in `src/routes/settings.tsx` about the react-router-generated `./+types/settings` module — not touched here.

---

## Why

`useFilteredEvents` is consumed by `ChatInterface`, which re-renders on every incoming conversation event. Inside the hook, `conversationUserEventsExist = hasAgentServerUserEvent(allConversationEvents)` runs an unmemoized `events.some(...)` scan on every render. On long conversations this is an O(n) scan that reruns even when nothing about the event store has changed, and it participates in the re-render loop that fires on each streamed token.

## Summary

- Wrap the `hasAgentServerUserEvent` call in `React.useMemo` keyed on `allConversationEvents` so the scan only reruns when the underlying store reference changes.
- Add a regression test that spies on `hasUserEvent` and asserts repeated rerenders with an unchanged store do not trigger additional scans.

## Issue Number

None.

## How to Test

```
npm install
npm run make-i18n
npx vitest run __tests__/hooks/use-filtered-events.test.ts
```

The new test `does not rescan allConversationEvents for user events on rerender when storeEvents has not changed` fails on `main` (3 vs 1) and passes with this change. The full suite still passes:

```
npx vitest run
# Test Files  518 passed | 1 skipped (519)
# Tests  4016 passed | 5 skipped | 9 todo (4030)
```

## Video/Screenshots

N/A — behavior-preserving perf change, verified through test-count assertions on the memoized dependency.

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The adjacent `allConversationEvents = React.useMemo(() => storeEvents, [storeEvents])` at line 24 is a no-op memo (the selector already returns a stable reference by identity). It is intentionally **not** touched in this PR because the existing test `returns the same allConversationEvents reference when storeEvents has not changed` asserts against it and I preferred to keep the diff minimal for a first perf pass.


hi-neason:
Heads-up on the `mock-llm-e2e` failure: I don't think it's caused by this PR.

  - Recent runs of `mock-llm-e2e` on main-targeted PRs failed 15/20; my sibling PR (#16163, docs-only, no source changes) ran at the same time and passed. That's inconsistent with this PR being the cause.
  - The 8 failing specs all share the same failure shape: navigation expected `/conversations/.+` but stayed on `/`. That's a conversation-create / routing failure, not a rendering issue.
  - The only caller of `useFilteredEvents` is `ChatInterface` (`src/components/features/chat/chat-interface.tsx`), which mounts *after* the conversation route is entered. The memoization change here is unreachable from the code path that creates or navigates to a conversation.

 Happy to re-run `mock-llm-e2e` once the workflow is healthy on main, or to add a profiler screenshot / bench numbers for the memo change if that would help review.

hi-neason:
Heads-up on the `mock-llm-e2e` failure: I don't think it's caused by this PR.

- Recent runs of `mock-llm-e2e` on main-targeted PRs failed 15/20; my sibling PR (#16163, docs-only, no source changes) ran at the same time and passed. That's inconsistent with this PR being the cause.
- The 8 failing specs all share the same failure shape: navigation expected `/conversations/.+` but stayed on `/`. That's a conversation-create / routing failure, not a rendering issue.
- The only caller of `useFilteredEvents` is `ChatInterface` (`src/components/features/chat/chat-interface.tsx`), which mounts *after* the conversation route is entered. The memoization change here is unreachable from the code path that creates or navigates to a conversation.

Happy to re-run `mock-llm-e2e` once the workflow is healthy on main, or to add a profiler screenshot / bench numbers for the memo change if that would help review.

PR #16164, opened 2026-07-29
perf: batch StreamingDeltaEvents so the UI keeps up with fast models

### LINKED ISSUE #15444: perf: batch StreamingDeltaEvents so the UI does not fall behind the model
## Problem

Reported in [Slack](https://openhands-ai.slack.com/archives/C06QT0AGY4W/p1784380705885159): live `StreamingDeltaEvent` output can progressively trail the tokens the LLM has already generated.

The durable final message is not the problem. The transient path amplifies each provider chunk into independent server work, a WebSocket frame, multiple client state updates, growing collection copies, and a re-render of the whole accumulated message. Fast models can therefore produce work faster than the server/browser can drain it.

```text
provider chunk
  -> token callback
  -> new event + independent PubSub coroutine
  -> Pydantic/JSON serialization + WebSocket frame
  -> JSON.parse + React state update + Zustand update
  -> Set/array copies + filter/group + accumulated Markdown render
```

### 1. The server schedules one untracked delivery per provider chunk

The agent server constructs a `StreamingDeltaEvent` for every delta, calls `asyncio.run_coroutine_threadsafe(self._pub_sub(event), ...)`, and discards the returned future ([event_service.py](https://github.com/OpenHands/software-agent-sdk/blob/4fe565663af2b4f1130a6e0dac7566b002bfe9b4/openhands-agent-server/openhands/agent_server/event_service.py#L786-L830)). Each publish fans out into subscriber tasks ([pub_sub.py](https://github.com/OpenHands/software-agent-sdk/blob/4fe565663af2b4f1130a6e0dac7566b002bfe9b4/openhands-agent-server/openhands/agent_server/pub_sub.py#L77-L98)), and each WebSocket subscriber serializes and sends a separate JSON frame ([sockets.py](https://github.com/OpenHands/software-agent-sdk/blob/4fe565663af2b4f1130a6e0dac7566b002bfe9b4/openhands-agent-server/openhands/agent_server/sockets.py#L473-L481)).

There is no bounded queue, coalescing, send timeout, or backpressure. If sends are slower than generation, in-flight coroutines and frames accumulate. This server-side resource-safety problem is already tracked by [OpenHands/software-agent-sdk#4077](https://github.com/OpenHands/software-agent-sdk/issues/4077); this issue should not duplicate that work.

### 2. Canvas commits and renders every frame individually

For every WebSocket message, Canvas:

- writes `event.data` into the hook's `lastMessage` React state before invoking the handler ([use-websocket.ts](https://github.com/OpenHands/agent-canvas/blob/c20a51a99a51801f9cc69720ba7426951f37d3e7/src/hooks/use-websocket.ts#L66-L69)); no current caller reads `lastMessage`, so every token causes an avoidable provider render;
- parses, validates, deduplicates, and immediately calls `addEvent(event)` ([conversation-websocket-context.tsx](https://github.com/OpenHands/agent-canvas/blob/c20a51a99a51801f9cc69720ba7426951f37d3e7/src/contexts/conversation-websocket-context.tsx#L471-L489));
- copies the complete `eventIds` Set, `events` array, and UI-event array ([use-event-store.ts](https://github.com/OpenHands/agent-canvas/blob/c20a51a99a51801f9cc69720ba7426951f37d3e7/src/stores/use-event-store.ts#L91-L121));
- recomputes filtering/grouping because the last event object changes ([use-filtered-events.ts](https://github.com/OpenHands/agent-canvas/blob/c20a51a99a51801f9cc69720ba7426951f37d3e7/src/hooks/use-filtered-events.ts#L15-L44), [messages.tsx](https://github.com/OpenHands/agent-canvas/blob/c20a51a99a51801f9cc69720ba7426951f37d3e7/src/components/conversation-events/chat/messages.tsx#L60-L63)); and
- splits and renders the complete accumulated streamed text again ([event-message.tsx](https://github.com/OpenHands/agent-canvas/blob/c20a51a99a51801f9cc69720ba7426951f37d3e7/src/components/conversation-events/chat/event-message.tsx#L194-L214)).

Although adjacent deltas replace the last event in the arrays, every discarded delta ID remains in `eventIds`. Copying that growing Set on every token is quadratic over a long stream. The arrays and accumulated Markdown add further per-token work.

### 3. A separate SDK fallback can defer all callbacks until generation is complete

The normal async path invokes the token callback while iterating. However, when an instrumented `litellm_acompletion` returns a synchronous iterator, the SDK runs `list(ret)` in an executor and only invokes callbacks after that list has fully drained ([llm.py](https://github.com/OpenHands/software-agent-sdk/blob/4fe565663af2b4f1130a6e0dac7566b002bfe9b4/openhands-sdk/openhands/sdk/llm/llm.py#L2146-L2165)). The comment specifically calls out lmnr instrumentation.

If the affected production request takes this branch, it directly explains deltas arriving after the model has generated them. We should instrument/confirm the iterator type and fix this in an SDK follow-up by bridging the synchronous iterator one item at a time rather than materializing it with `list()`.

## What other products do

The common pattern is to separate stream ingestion from presentation and bound presentation cadence:

- **OpenCode** queues incoming server events, concatenates adjacent matching text deltas, flushes at most every 16 ms inside a single reactive `batch()`, and yields the stream loop at least every 8 ms ([server-sdk.tsx](https://github.com/anomalyco/opencode/blob/b95fe7b2d74b9f17d5573dfc783d1bf8f9e3f298/packages/app/src/context/server-sdk.tsx#L40-L71), [flush loop](https://github.com/anomalyco/opencode/blob/b95fe7b2d74b9f17d5573dfc783d1bf8f9e3f298/packages/app/src/context/server-sdk.tsx#L103-L137), [stream loop](https://github.com/anomalyco/opencode/blob/b95fe7b2d74b9f17d5573dfc783d1bf8f9e3f298/packages/app/src/context/server-sdk.tsx#L190-L203)).
- **pi** updates one in-progress assistant message and awaits event listeners ([agent-loop.ts](https://github.com/badlogic/pi-mono/blob/3da591ab74ab9ab407e72ed882600b2c851fae21/packages/agent/src/agent-loop.ts#L319-L345), [agent.ts](https://github.com/badlogic/pi-mono/blob/3da591ab74ab9ab407e72ed882600b2c851fae21/packages/agent/src/agent.ts#L521-L573)); repeated render requests are coalesced and capped to one per 16 ms ([tui.ts](https://github.com/badlogic/pi-mono/blob/3da591ab74ab9ab407e72ed882600b2c851fae21/packages/tui/src/tui.ts#L712-L754)).
- **Codex** buffers Markdown to stable newline boundaries ([markdown_stream.rs](https://github.com/openai/codex/blob/b8b61bc692517adcd18622df260f2ddd80635122/codex-rs/tui/src/markdown_stream.rs#L1-L9)), caps redraws at 120 FPS ([frame_rate_limiter.rs](https://github.com/openai/codex/blob/b8b61bc692517adcd18622df260f2ddd80635122/codex-rs/tui/src/tui/frame_rate_limiter.rs#L1-L13)), and switches from smooth display to draining the full backlog when queue depth reaches 8 lines or oldest age reaches 120 ms ([chunking.rs](https://github.com/openai/codex/blob/b8b61bc692517adcd18622df260f2ddd80635122/codex-rs/tui/src/streaming/chunking.rs#L82-L116), [decision](https://github.com/openai/codex/blob/b8b61bc692517adcd18622df260f2ddd80635122/codex-rs/tui/src/streaming/chunking.rs#L176-L203)).

Responsiveness does not require a render for every provider chunk. A 16 ms coalescing window is below a display frame while reducing work dramatically.

## Proposed Canvas change

1. Add an ordered inbound queue in the conversation WebSocket layer.
2. Concatenate **adjacent** `StreamingDeltaEvent` values for the same stream/source, preserving both `content` and `reasoning_content`.
3. Flush at most once per animation frame / 16 ms (with a timer fallback). Apply the flushed batch in one store transaction.
4. Before processing any non-delta event, synchronously flush pending deltas. Also flush on close/unmount. This preserves the invariant that the final `MessageEvent`, tool action, or `FINISHED` state cannot overtake its text.
5. Remove the unused `lastMessage` state update from `useWebSocket` (or make it opt-in/ref-only).
6. Do not retain IDs for merged-away transient deltas. `StreamingDeltaEvent` is explicitly a non-durable UX affordance ([definition](https://github.com/OpenHands/software-agent-sdk/blob/4fe565663af2b4f1130a6e0dac7566b002bfe9b4/openhands-sdk/openhands/sdk/event/streaming_delta.py#L5-L13)); ideally model the active stream separately from durable history.
7. Coordinate the server queue/backpressure work in SDK OpenHands/OpenHands#4077. The server should ultimately have one ordered, bounded writer per subscriber/conversation, coalesce adjacent deltas for a short time/size threshold, and flush them before durable/control events.

This should remain compatible with the reconciliation fixes in [#1656](https://github.com/OpenHands/agent-canvas/issues/1979). Coalescing must not cross a non-delta boundary or combine planning/main streams.

## Observability and acceptance criteria

Add timestamps/counters at: SDK callback receipt, server enqueue, WebSocket send completion, browser receipt, and UI commit. Record queue depth and oldest pending-delta age.

- A deterministic test feeds thousands of one-character deltas faster than 60 Hz; rendered text is byte-for-byte identical and event ordering is preserved.
- Canvas performs at most one normal streaming store commit/render per 16 ms, plus explicit boundary/finalization flushes.
- A final `MessageEvent`, action, or `FINISHED` update never appears before an earlier queued delta and no stale delta appears afterward.
- `eventIds` growth is bounded by durable events / presentation batches, not raw provider chunk count.
- Under a fast synthetic stream, pending UI delta age remains bounded (target p95 under 100 ms and maximum under 150 ms on a representative machine).
- Instrumentation identifies whether the synchronous-iterator SDK branch is active in the reported environment; if so, file/fix the SDK follow-up before judging Canvas batching results.

VascoSch92:
HUMAN:

Perf for `StreamingDeltaEvents`

AGENT:

Evidence this runs end-to-end (beyond unit tests):

- **Measured on a real agent-server.** Drove the mock-LLM e2e stack (real agent-server 1.36.1 behind the ingress; a mock LLM streaming a ~560-char reply one token at a time) through a real browser conversation, with temporary counters in the batcher:

  ```
  deltas received (enqueued): 562
  store commits (flushes):     86   →  6.5x fewer commits
  rendered text matches stream byte-for-byte: true
  ```

  `enqueued` (562) is exactly what `main` commits today — one store commit per provider token — so this is a genuine before/after in one run: 562 → 86 commits (~85% fewer), each commit now cheaper too (bounded `eventIds`, no per-token Set copy), and the streamed text rendered byte-for-byte correct with no reordering. The reduction scales with how far the token rate outruns the 60 Hz frame rate — i.e. exactly the fast-model case this issue targets.

- **Reproduced the regression** as a deterministic benchmark against the real `useEventStore`, feeding N one-character deltas the way the WS handler does today. It scales O(n²) — total time quadruples per doubling — and `eventIds` grows to N even though every delta merges into a single event:

  ```
  n= 1000  total=  3.0ms  per-delta= 3.03µs  eventIds.size=1001
  n= 2000  total= 11.9ms  per-delta= 5.97µs  eventIds.size=2001
  n= 4000  total= 44.3ms  per-delta=11.08µs  eventIds.size=4001
  n= 8000  total=174.7ms  per-delta=21.84µs  eventIds.size=8001
  ```

- **Wiring is exercised through the real provider.** `conversation-websocket-context.test.tsx` drives the actual `handleMainMessage` (WebSocket transport stubbed, store/provider/history real) and asserts:
  - deltas are buffered, not committed per token;
  - a final `MessageEvent` flushes the buffer first and reconciles the streamed text in place (single bubble, no race/duplication);
  - switching conversations discards the previous conversation's buffered deltas (no cross-conversation leak).
- **Batcher integration test** streams "I'll start working on that." one char per delta across frames, then delivers the durable message → one reconciled bubble and `eventIds.size === 2` (durable events, not the 27 deltas).
- **Full suite green:** `npx vitest run` → **498 files / 3835 tests passed**, 0 failures. Typecheck (`tsc --noEmit`), ESLint, and Prettier all clean; the pre-commit hook passed.
- **Remaining gap:** verified against the real agent-server with a mock LLM (above), not yet against a live provider streaming at production cadence — kept as a draft until a screen recording of that is attached (see Video/Screenshots).

---

## Why

On fast models, live `StreamingDeltaEvent` output progressively trails the tokens the model has already generated (reported in Slack). The durable final message is fine; the transient path amplifies each provider chunk into a WebSocket frame, multiple client state updates, growing collection copies, and a full re-render of the accumulated message. The event store also copied the whole `eventIds` Set on every token, making a long stream O(n²). Responsiveness doesn't require a render per provider chunk — a ~16 ms coalescing window is below a display frame while cutting the per-token work dramatically.

## Summary

- Add `createStreamingDeltaBatcher`: coalesces adjacent deltas and commits at most once per animation frame (16 ms timer fallback), flushing synchronously before any non-delta event so a durable message/action can never overtake its own streamed text.
- Wire separate main/planning batchers into `ConversationWebSocketProvider`; discard buffered deltas on conversation switch/unmount, and drop the unused per-frame `lastMessage` state from `useWebSocket`.
- Stop tracking transient delta ids in the event store (deltas are never persisted or resent — verified in the SDK), bounding `eventIds` growth to durable events instead of raw provider chunk count.

## Issue Number

Closes OpenHands/OpenHands#15444. Server-side queue/backpressure is tracked separately in SDK OpenHands/OpenHands#4077 and is intentionally out of scope here.

## How to Test

1. `npm install` (installs `@openhands/typescript-client@1.34.0` from the lockfile).
2. Run the deterministic suite:
   ```
   npx vitest run __tests__/utils/streaming-delta-batcher.test.ts \
     __tests__/stores/use-event-store.test.ts \
     __tests__/hooks/use-websocket.test.ts \
     __tests__/contexts/conversation-websocket-context.test.tsx
   ```
   It covers coalescing (≤ 1 flush/frame), the flush-before-non-delta ordering invariant driven through the real provider, byte-for-byte-identical text across 5000 one-char deltas, bounded `eventIds`, and the conversation-switch reset.
3. Manual (recommended before merge): start a conversation with a fast model and confirm the streamed text keeps pace, and the final message doesn't duplicate or reorder.

## Video/Screenshots

Measured on the real agent-server (mock-LLM stack), a ~560-token streamed reply produced **562 deltas → 86 store commits (6.5× fewer)** with byte-for-byte-correct rendered text. A live-provider screen recording is still pending — draft until then.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Why dropping delta ids is safe:** `StreamingDeltaEvent`s are never persisted server-side (`event_service.py` publishes them to subscribers but does not write them to `ConversationState.events`), and resend reads only the persisted log — so they're never re-sent on reconnect, and there is no dedup scenario to protect. This is what bounds `eventIds` and removes the O(n²) Set copy.
- Coalescing never crosses a non-delta boundary and never mixes the main vs planning streams (separate batchers), staying compatible with the reconciliation from OpenHands/agent-canvas#1979.
- Follow-ups: server-side backpressure (SDK OpenHands/OpenHands#4077); optional observability counters (queue depth / oldest-pending-delta age) from the issue's acceptance criteria.

---

| Component | Value |
|---|---|
| **Commit** | `96c6895804c5dd8c7af47ee2e139be4bfe9c1503` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-96c6895
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-96c6895
```

```
ghcr.io/openhands/agent-canvas:sha-96c6895-amd64
ghcr.io/openhands/agent-canvas:vasco-streaming-delta-events-amd64
ghcr.io/openhands/agent-canvas:pr-16164-amd64
ghcr.io/openhands/agent-canvas:sha-96c6895-arm64
ghcr.io/openhands/agent-canvas:vasco-streaming-delta-events-arm64
ghcr.io/openhands/agent-canvas:pr-16164-arm64
ghcr.io/openhands/agent-canvas:sha-96c6895
ghcr.io/openhands/agent-canvas:vasco-streaming-delta-events
ghcr.io/openhands/agent-canvas:pr-16164
```

- Each tag (e.g., `sha-96c6895`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-96c6895-amd64`) are also available if needed


VascoSch92:

neubig (approved):
Seems good to me, thanks!

PR #16173, opened 2026-07-29
fix(backend-registry): pick a healthy local backend as the fallback

mvanhorn:
HUMAN:
I work with agents, and I check their output before it goes out. Before porting this I had the failure reproduced against current `main` so it wasn't a faith-based port: with a cloud backend at index 0 and a healthy local further down, `getEffectiveLocalBackend()` comes back null and local calls throw. Ran the build and the backend-registry suite locally. The before/after console output is in the agent section below.

AGENT:
Ported from OpenHands/agent-canvas#1457, which @neubig closed asking for it to be reopened here.

Confirmed the bug is live on this base, not just ported on faith. I built a registry matching the failure shape (cloud backend at index 0, a local backend failed past `MAX_CONSECUTIVE_FAILURES`, then a healthy local) and ran it against `main` and against this branch:

On `main` (unpatched):
```
chosen active backend : prod (kind=cloud)
getEffectiveLocalBackend(): null
```

On this branch:
```
chosen active backend : local-ok (kind=local)
getEffectiveLocalBackend(): local-ok
```

The null is the failure: every local-protocol call throws "No backend is configured" while a healthy local backend sits in the registry.

Also ran: `npm run build` (production build succeeds), `npx tsc --noEmit -p tsconfig.json` (0 errors in `src/` and `__tests__/`), and `npx vitest run __tests__/api/backend-registry/` (4 files, 45 tests passed). Pre-commit hooks ran eslint, prettier, staged typecheck, and check-translation-completeness.

Not done: no click-through of the real GUI against a live agent-server. The evidence above is the store logic exercised directly.

---

## Why

With no valid explicit selection, `pickFallbackBackend` returns `backends[0]`. When that is a cloud or dead local backend, `getEffectiveLocalBackend()` is null and local-protocol calls throw even though a healthy local backend is registered further down the list.

## Summary

- Prefer a healthy local backend, consulting the persisted health store
- Fall back to any local backend so the effective-local resolver still resolves
- Then fall back to the prior `backends[0]` behavior for the registry-has-no-local case

## Issue Number

Ported from OpenHands/agent-canvas#1457

## How to Test

```
npm install
npx vitest run __tests__/api/backend-registry/
```

Manual: register a cloud backend first and a healthy local backend second, clear the active selection, then confirm local-protocol calls resolve instead of throwing "No backend is configured".

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

### DIFF
--- a/__tests__/api/backend-registry/active-store.test.ts
+++ b/__tests__/api/backend-registry/active-store.test.ts
@@ -9,11 +9,17 @@ import {
   subscribeActiveBackend,
 } from "#/api/backend-registry/active-store";
 import { SEEDED_DEFAULT_BACKEND_ID } from "#/api/backend-registry/default-backend";
+import { MAX_CONSECUTIVE_FAILURES } from "#/api/backend-registry/health-storage";
+import {
+  __resetHealthStoreForTests,
+  recordBackendFailure,
+} from "#/api/backend-registry/health-store";
 import type { Backend } from "#/api/backend-registry/types";
 
 beforeEach(() => {
   window.localStorage.clear();
   window.sessionStorage.clear();
+  __resetHealthStoreForTests();
   __resetActiveStoreForTests();
 });
 
@@ -21,6 +27,7 @@ afterEach(() => {
   window.localStorage.clear();
   window.sessionStorage.clear();
   vi.unstubAllEnvs();
+  __resetHealthStoreForTests();
   __resetActiveStoreForTests();
 });
 
@@ -39,6 +46,20 @@ const localBackend: Backend = {
   apiKey: "k",
   kind: "local",
 };
+
+const secondLocalBackend: Backend = {
+  id: "local-2",
+  name: "Local 2",
+  host: "http://localhost:9001",
+  apiKey: "k2",
+  kind: "local",
+};
+
+function markBackendUnhealthy(id: string): void {
+  for (let i = 0; i < MAX_CONSECUTIVE_FAILURES; i += 1) {
+    recordBackendFailure(id, new Error("connection failed"));
+  }
+}
 
 describe("active-store", () => {
   it("uses the no-backend sentinel when no backend details are available", () => {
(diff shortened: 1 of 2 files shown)


mvanhorn:
@neubig, accounting for the six you asked to be reopened here, since agent-canvas is archived and I cannot reply on those threads.

Reopened: agent-canvas#1457 -> #16173 (this one), agent-canvas#1687 -> #16174, agent-canvas#1661 -> #16175.

Not reopened, and why:

- **agent-canvas#1487** is already fixed here. `use-backends-health.ts` has `probeBackendWithQuickRetry` and `isRetryableProbeError`, the same shape that PR proposed.
- **agent-canvas#644** is superseded. `verification-settings.tsx` here already renders via `SdkSectionPage` with a `settingsSources` array that de-dupes the conversation-owned fields; the old PR used the singular `settingsSource`, so porting it would go backwards.
- **agent-canvas#1117** needs rework rather than a port. The masking behavior still looks worth having, but LLM settings were restructured with auth types and basic/advanced views, so `llm-api-key-input` no longer renders where its tests look and all four of them fail. Happy to redo it against the current structure if you want it.

enyst (commented):

🟢 **Good taste** — The fallback policy is small, deterministic, and matches the actual protocol boundary: preserve explicit selections, otherwise prefer a non-disabled local backend, then retain deterministic degradation. The regression tests cover cloud-first ordering, unhealthy-local skipping, all-unhealthy fallback, explicit cloud selection, and the no-local case without over-mocking the unit under test.

**VERDICT:**
✅ **Worth merging**: Core logic is sound and the supplied runtime evidence plus green CI support the behavior change.

**KEY INSIGHT:**
Keeping fallback selection local-first prevents registry insertion order from disabling local-protocol services while leaving explicit cloud intent untouched.

**[RISK ASSESSMENT]**
The change is isolated to unselected/invalid backend fallback resolution, follows existing synchronous store patterns, preserves explicit selections and the no-local fallback, adds focused regression coverage, and touches no authentication, secrets, dependencies, or public API surface.

_This review was generated by an AI agent (OpenHands) on behalf of Engel Nyst._

neubig (approved):
Thank you!

PR #16174, opened 2026-07-29
feat(transcript-export): bounded head+tail export for very large conversations

mvanhorn:
HUMAN:
I work with agents and review what they produce before submitting. Ran the build and the transcript-export suites locally on this branch. What I have not done is click through the export modal against a real 10k+ event conversation, and I would rather say that than imply otherwise.

AGENT:
Ported from OpenHands/agent-canvas#1687, which @neubig closed asking for it to be reopened here. `src/utils/transcript-export/` already exists on this base; this adds the bounded head+tail path and the omission notice, which are not present.

Ran `npm run build` (production build succeeds), `npx tsc --noEmit -p tsconfig.json` (0 errors in `src/` and `__tests__/`), and the feature's own suites:

- `src/utils/transcript-export/load-complete-events.test.ts` — 9 passed, including "loads the oldest head + newest tail and reports the omitted middle" and "merges live store events into the newest tail without duplicating"
- `src/utils/transcript-export/index.test.ts` truncation cases — 5 passed: single notice between head and tail in Markdown, note aside in HTML, and no notice when `omittedCount` is zero

Pre-commit hooks ran eslint, prettier, staged typecheck, and check-translation-completeness.

Not done: no click-through of the export modal in a running GUI against a conversation with more than 10,000 events. I could not produce a conversation that large against a live agent-server, so the bounding is exercised through the loader and renderer directly rather than from the UI.

---

## Why

Exporting a very large conversation loaded every event, so the browser had to hold the entire history at once.

## Summary

- Load a bounded head and tail window instead of the full history
- Insert one omission notice between them stating how many middle events were left out
- Keep full export behavior when head+tail already covers the conversation

## Issue Number

Ported from OpenHands/agent-canvas#1687

## How to Test

Open a conversation with more than 10,000 events and choose Export conversation. The modal now shows an "Amount to export" choice; partial is preselected and states how many events it keeps, and whole carries a size warning. Pick partial, export as Markdown, and the downloaded file contains the first 2,000 and last 8,000 events with a single omission notice between them stating how many were left out. Pick whole and the file contains the full history with no notice. A conversation under the limit shows no such choice and exports exactly as before.

The automated coverage for the same paths:

```
npm install
npx vitest run src/utils/transcript-export/
npx vitest run __tests__/components/features/conversation/transcript-export-modal.test.tsx
```

## Video/Screenshots

The Export conversation modal for a large conversation (25,000 events), showing the new "Amount to export" fieldset:

Export conversation modal showing the Amount to export choice, with partial conversation selected and a size warning on the whole-conversation option

Rendered from `TranscriptExportModal` with the event count stubbed at 25,000 so the large-conversation branch is exercised. As noted above, I have not driven this from a live agent-server conversation of that size.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.60s · commit b13bd63  
**Coverage:** ⚠️ reduced context — partial coverage; 27/35 hunks, 7/7 files (context budget: 27, hunk budget: 8).

<details>

| --- | --- | --- |
| Command injection | 3.0% | No direct hunk selected |
| Weakened authentication | 5.0% | No direct hunk selected |
| Weakened authorization | 10.0% | No direct hunk selected |
| Contract regression | 15.0% | No direct hunk selected |
| Sensitive data disclosure | 18.0% | No direct hunk selected |
| Credential misuse | 8.0% | No direct hunk selected |
| Security assessment bypass | 4.0% | No direct hunk selected |
| Primary concern | None selected; confidence 82.0% | No primary concern to locate |

</details>


mvanhorn:
Rebased on main, conflicts resolved. The only real one was `translation.json`, where main added the `SETUP$` and `BUTTON$CONTINUE` keys in the same place as `TRANSCRIPT_EXPORT$TRUNCATION_NOTICE`; both sets are kept.

enyst (commented):

🟢 **Good taste** — The bounded loader has one clear job, keeps the existing full-export path intact for smaller conversations, and carries truncation as explicit data into both renderers. Cursor handling, de-duplication, stable timestamp ordering, head/tail coverage, and omission rendering all have focused regression tests; I found no actionable correctness issue in the checked-out head.

**[RISK ASSESSMENT]**
This changes non-trivial pagination and user-visible export behavior for very large histories, so the blast radius is more than cosmetic. The implementation is isolated to transcript export, caps memory use explicitly, preserves the existing path below the threshold, and has direct loader/renderer coverage; it does not touch auth, secrets, dependencies, or agent execution.

**VERDICT:**
✅ **Worth merging**: Core logic is sound and the bounded head+tail representation is simpler than materializing the omitted middle.

**KEY INSIGHT:**
Making truncation an explicit renderer input keeps the omission honest without contaminating ordinary transcript entries.

neubig (changes requested):
For this one, I'm actually not sure if it's ideal at the moment. In the rare case that someone tries to export a 10,000 event conversation, this PR would mean that they would have no way to export the full history right? I know that many users get very frustrated when they don't have a way to recover their data, so I'd hope that if we make this fix, we should also have a way to let them optionally download the whole history (e.g. pop up a modal telling them it's very large, and then allow them to select "download partial conversation" or "download whole conversation".

But also, I think it might just be OK to not do this change for simplicity. It kinda feels like YAGNI.

mvanhorn:
Yes, that's right. As it stands the bound is unconditional, so a 10k-event conversation has no path to a full export.

I'm happy to add the modal you described, partial vs whole with a size warning on the whole option. But you also raised YAGNI, and you know the usage better than I do. If nobody is actually hitting large enough conversations for the export to hurt, closing this is a fine outcome and I won't take it personally.

One thing worth flagging since it may have gotten lost: this is the port of agent-canvas#1687, which you closed asking for it to be reopened here. So if the answer is now "don't do it," that's genuinely fine, I just want to make sure it's a decision rather than the context getting dropped between the two repos.

Happy either way, just tell me which.

mvanhorn:
@neubig the modal now offers the choice you described.

Over 10k events, Export waits for the count then shows Amount to export: partial (recommended, default — first 2000 + last 8000 with an omission note) or whole (warning that the file can be large, slow, and memory-heavy). Under the threshold, or if the count endpoint is missing, it's still one click on the full history.

So a 10k+ conversation is no longer a dead end.

mvanhorn:
@neubig this is addressed — sorry for not saying so at the time, the commit went up without a note and your review has been sitting on the old state since.

You were right that a bounded export alone would strand people who need the full history. There is now an explicit choice rather than a silent truncation: when a conversation is large enough to be bounded, the export modal shows an "Amount to export" step with **partial** (head+tail, the fast default) and **whole** (all N events, carrying a warning that the file may be very large and the download slow and memory-hungry). Picking whole runs the complete export path, so nothing is unrecoverable.

The omission notice is still there for the partial case, so a truncated export says what it left out.

Also merged current main just now to clear the conflict — the only collision was `src/i18n/translation.json`, where upstream's new keys and mine landed in the same place; all 2400 upstream keys are preserved alongside the 7 new `TRANSCRIPT_EXPORT$*` ones. `transcript-export` suites are 27/27 locally and Prettier is clean.

Ready for another look when you have a moment.

neubig:
Thanks, will do!

mvanhorn:
@neubig I merged main locally to clear the conflict and hit something worth your call before I push anything.

main has since added a 433-line suite for this modal (`__tests__/components/features/conversation/transcript-export-modal.test.tsx`). Nine of its tests pass against main's modal and fail against this branch, and the cause is a real behavioural difference rather than a bad merge.

main fetches the event count lazily, inside the export handler. This branch fetches it on mount, because the partial/whole choice has to know whether the conversation is over the threshold *before* the user picks. One of main's tests, "closes immediately without starting an export", asserts `getEventCount` is never called when the modal is merely opened. That is incompatible with offering the choice up front, by construction.

Two of the nine were my fault and I fixed them: I had the export button disabled while the size check was in flight, so the existing tests were clicking a disabled button. The handler already awaits the in-flight promise, so the button can stay enabled and an early click still sees the count. The rest come down to eager versus lazy.

Three ways I can see:

1. Keep the eager count and relax that one assertion, on the grounds that the feature needs the size before the user chooses.
2. Keep main's contract and move the choice after the click: export, discover it is large, then offer partial or whole. Slower UX, no request on open.
3. Ask for the count only when the modal is opened on a conversation already suspected to be large, if there is a cheaper signal available.

I lean towards 1, but it edits a contract you just set, so I would rather you choose. The merge resolution itself is done and ready either way: translation.json takes both key sets, and I split this branch's bounded-export tests into their own file rather than splicing them into main's suite, since the two need different mocks.

Separately, the same `Validate PR description` blocker as #16175 applies here: the `## Issue Number` section points at OpenHands/agent-canvas#1687, another repo, so it cannot carry a `ready-for-dev` label, and I found nothing here matching transcript export to link instead. Whatever you decide on that thread covers this one too.

mvanhorn:
@neubig the description check is holding this on the linked issue: it wants an OpenHands issue carrying `ready-for-dev` under `## Issue Number`, and the port from agent-canvas#1687 has none here. Would you rather I open one describing the bounded head+tail export for you to label, or is there an existing tracking issue I should link? Once that is settled I will resolve the merge conflict in the same push.

enyst:
@openhands Understand this PR, then merge main into it and fix the conflicts. Push; note that it’s from a fork so make sure to set your remote properly.

Then take a step back, switch your LLM profile to gpt-6-astra, wait 30 seconds, then make a code review of this PR, posting it as a review with gh api.

mvanhorn:
The one red check here is Validate PR description. It wants a linked issue in this repo under ## Issue Number carrying ready-for-dev, and this PR only references the port source (OpenHands/agent-canvas#1687), which the checker does not accept. I could not find an existing OpenHands issue for bounded head+tail transcript export, and #17494 is a different export bug.

Could you open a ready-for-dev companion for it, in the shape of #17559 and #17533? I will link it from the description and the check should go green. Happy to write the issue body if you would rather I opened it and you just applied the label.

Standing off the branch meanwhile so I do not collide with the agent run you started on 09-17.

neubig (changes requested):
I pushed commit `0a1187804` with a focused regression test demonstrating that the partial-export path is not currently bounded after merging the live event store.

The failing case passes the complete 20-event history as `loadedEvents`, with `headMax=3` and `tailMax=5`. The expected partial result is the oldest 3 plus newest 5 events, but the loader returns all 20 because it appends every live-store event after fetching the bounded windows.

Reproduction:

```bash
npx vitest run src/utils/transcript-export/load-complete-events.test.ts \
  -t 'keeps the result bounded when the live store contains the full history'
```

Current result: **1 failed**. The returned IDs include `event-003` through `event-014`, which should be omitted.

Please make the final live-store merge respect the tail bound—for example, merge/deduplicate fetched-tail and live events, sort, and retain only the newest `tailMax` before combining with the head. The final partial result should never exceed `headMax + tailMax`, and `omittedCount` should be calculated from that bounded final set.

_This review was created by an AI agent (OpenHands) on behalf of the user._

neubig:
I pushed commit [`0a1187804`](https://github.com/OpenHands/OpenHands/pull/16174/commits/0a118780410be630ef3fa6410a453f8fc9e9ad26) with a focused regression test demonstrating that the partial-export path is not currently bounded after merging the live event store.

The failing case passes the complete 20-event history as `loadedEvents`, with `headMax=3` and `tailMax=5`. The expected partial result is the oldest 3 plus newest 5 events, but the loader returns all 20 because it appends every live-store event after fetching the bounded windows.

Reproduction:

```bash
npx vitest run src/utils/transcript-export/load-complete-events.test.ts \
  -t 'keeps the result bounded when the live store contains the full history'
```

Current result: **1 failed**. The returned IDs include `event-003` through `event-014`, which should be omitted.

Please make the final live-store merge respect the tail bound—for example, merge/deduplicate fetched-tail and live events, sort, and retain only the newest `tailMax` before combining with the head. The final partial result should never exceed `headMax + tailMax`, and `omittedCount` should be calculated from that bounded final set.

mvanhorn:
Thanks for pushing the failing test, that made the bug unambiguous.

Fixed in `cb78e72`, the way you suggested: the fetched tail and the live store events are merged and deduped by id, sorted, then sliced to the newest `tailMax` before being combined with the head. The partial result can no longer exceed `headMax + tailMax`, and `omittedCount` is derived from that bounded set rather than from the pre-merge windows.

Your case passes now, and all 11 tests in `load-complete-events.test.ts` are green.

One existing test changed with it: `preserves events without IDs in the head, tail, and live store` passed `all[4]` as a live event sitting in the omitted middle, and previously expected it to appear in the output. Under the bound it cannot, so that case now expects `[all[0], all[1], all[6], all[7]]` with `omittedCount: 4`. I renamed it to `preserves id-less head and tail events within the tail bound` since that is what it actually pins now. Flagging it explicitly in case you would rather keep a separate test for id-less live events.

PR #16179, opened 2026-07-29
fix: make sidebar "Load more" discover folders, not folder contents

FraterCCCLXIII:
HUMAN:

I injected demo content to test, passed a visual test to make sure it worked as anticipated. - Paul

AGENT:

Verified end-to-end in a browser against the running app (`npm run dev:mock` with an injected mock fixture), not just unit tests — commands, fixture layout, and per-step observations are in **How to Test** below. After the review-driven revisions, the same observable click behavior is additionally locked in by unit tests (`npx vitest run __tests__/components/features/conversation-panel/`, 247/247 passing) and the full suite (`npm test`, 4528 passed / 0 failures).

---

## Why

In grouped mode, the conversation panel's global **Load more** control measured its own success in *conversations*, not *folders*. Because a later backend page can contain more conversations for a folder that is already on screen, a click could grow an already-visible folder past its five-row preview limit while revealing no new workspace folder at all — which is the one thing that control exists to do.

That also makes the two controls in the sidebar fight each other. Each folder already has its own **More** control for expanding its conversations, so the global control adding rows inside an existing folder is both redundant and, from the user's point of view, a no-op: they clicked "Load more" and no new folder appeared.

## Summary

- Freeze each folder's **collapsed preview** to the backend page on which that folder first appeared (`conversationPageById` + `getGroupDiscoveryConversationIds`), while the group model keeps every loaded conversation: later pages can introduce new folders without visibly mutating folders the user can already see, expanding a folder still reveals all of its loaded rows (nothing becomes unreachable), and the active conversation is force-included in its folder's preview even when it arrived on a later page.
- Measure grouped "Load more" progress in folders (`visibleGroupCount`) instead of total conversations: the driver effect (with `loadedPageCount` as a dependency) walks past pages that only deepen already-visible folders until a new folder appears, pages run out, or the per-click cap `MAX_PAGES_PER_LOAD_MORE_CLICK = 3` is hit — the cap applies to grouped mode only, so one click is bounded at three page fetches instead of draining the whole cursor.
- Extract the shared workspace/repository folder-identity resolution (`getConversationGroupIdentity`, resolved once per conversation) so pagination and grouping cannot drift apart on normalization.

## How to Test

Unit coverage (fails on `main`, passes here):

```
npx vitest run __tests__/components/features/conversation-panel/
```

- `conversation-panel.test.tsx` → "keeps collapsed folder previews stable while still exposing later-page conversations on expand" — one click walks past a deepen-only page and reveals the next folder; expanding the frozen folder shows every loaded row.
- `conversation-panel.test.tsx` → "caps a grouped load-more click at three pages when no new folder appears" — pins the per-click page cap.
- `conversation-panel.test.tsx` → "force-includes the active conversation in a folder preview even when it arrived on a later page".
- `conversation-panel-list-helpers.test.ts` → "freezes discovery preview ids per folder and force-includes the active conversation".

Manual browser verification. Unit tests alone were not convincing here, so I temporarily injected a local-only conversation fixture into mock mode (**not** part of this PR) and drove the real UI. The fixture served 72 conversations over four cursor-paginated backend pages across six workspace folders, deliberately laid out so that page 2 mixes six more `agent-canvas` and five more `design-system` conversations with one brand-new `api-gateway` folder, and page 3 contains nothing but already-discovered `No workspace` rows.

Steps: `npm run dev:mock`, open the sidebar, then set the conversation filter menu to **By workspace**.

Observed with this change:

1. Page 1 renders `No workspace` (5 rows + **More**), `agent-canvas` (5 rows + **More**), `design-system` (3 rows), plus a global **Load more**.
2. First **Load more** click adds only the new `api-gateway` folder. `agent-canvas` stays at 5 rows and `design-system` stays at 3, even though page 2 carried 11 more conversations for those two folders.
3. Second **Load more** click pages through the all-`No workspace` page 3 without stalling and surfaces the final two folders, `docs-site` and `infra-tools`.

Step 2 is the regression: on `main` that click pours page 2's extra conversations into the visible folders instead of revealing `api-gateway`.

## Video/Screenshots

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Review-driven revisions since the first push: the group model no longer drops later-page conversations (only the collapsed preview is frozen; the per-folder **More** control reveals everything loaded), the active conversation is force-included in its folder's preview, and the "stop as soon as any conversation loads" secondary signal was removed — deepen-only pages are walked past, bounded by the 3-page per-click cap.
- Chronological mode is intentionally unchanged: it still measures progress in rows and is exempt from the per-click page cap, preserving its pre-existing fetch-until-visible behavior.
- The branch has been merged with `upstream/main` (which landed the client-side conversation archive and automation filter on the same panel code); the only conflicted file was `conversation-panel.tsx`, and both features' semantics are preserved.
- Known limitation (documented in code at `conversationPageById`): the discovery-page mapping is recomputed on the 10s background refetch, so a conversation whose `updated_at` shifts can move its recorded discovery page. Blast radius is which rows a collapsed preview shows — never reachability.
- The pre-existing `KNOWN ISSUE` comment about "Load more" occasionally needing two clicks is left in place. This change fixes the folder-discovery half of that behavior; the suspected cursor/refetch race it describes is a separate concern.
- Evaluating the `isFetching` guard ahead of the `hasNextPage` check in the driver effect is deliberate: React Query can transiently clear `hasNextPage` while replacing the last page, which would otherwise end the drive early.


src/components/features/conversation-panel/conversation-panel.tsx
hieptl, line 1: I think this introduces a case where conversations from later pages become permanently inaccessible in grouped mode.

`filterToGroupDiscoveryPages` feeds into `groupConversations` (`conversation-panel.tsx:350-360`), and the grouped result is what ultimately gets rendered (`conversation-panel.tsx:871-874`). The per-folder **More** button doesn't recover the omitted conversations because `conversation-group-folder-row.tsx:64` only expands `group.conversations`, which has already been truncated by the earlier filtering step. There isn't an additional fetch behind that button.

The new test in this PR appears to demonstrate this behavior as well. Page 1 contains six no-workspace conversations, and page 2 contains two more. After all three pages have been loaded, expanding the folder still asserts that only six conversation cards are rendered, even though all eight conversations have been fetched. `no-workspace-7` and `no-workspace-8` are therefore no longer reachable.

The impact is most noticeable in a common usage pattern. With a page size of 20 (`use-paginated-conversations.ts:8`), consider a user who has one workspace containing 500 conversations:

* Page 0 discovers folder **A**, so the group initially contains 20 conversations.
* Pages 1–24 also belong entirely to folder **A**, so every conversation on those pages is filtered out during group discovery.
* A single **Load more** action walks through all 25 pages (see comment #2). Once `hasNextPage` becomes `false`, the **Load more** button disappears.

The final result is that only 20 of the 500 conversations are visible, with no remaining UI path to access the other 480. In contrast, chronological mode continues to display all 500 conversations, so the two viewing modes end up presenting different underlying datasets.

Because of this, I don't think the PR description's assumption that "each folder already has its own **More** control for expanding its conversations" holds in this scenario. That control can only expand the conversations that were discovered on the initial discovery page; it cannot recover conversations that were filtered out from later pages.

Thank you very much! 🙏

hieptl, line 1: I think this can cause a single click to fetch every remaining page.

Because `visibleGroupCount` doesn't increase when the remaining pages contain only conversations for existing folders, the loop in `conversation-panel.tsx:444-473` continues calling `fetchNextPage()` until `hasNextPage` becomes `false`.

In the single-folder case, that can result in roughly 25 sequential network round trips from a single click, with a spinner displayed throughout but no visible UI change. In addition, every fetched page is then included in the 10-second `refetchInterval` cycle, so the steady-state polling cost grows as more pages are loaded.

It may be worth considering limiting this behavior—for example, by capping the number of pages fetched per click—or using the number of newly loaded conversations as a secondary success signal. That would allow the loop to stop once a page contributes additional data, even if it doesn't introduce a new folder.

Thank you very much! 🙏

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`src/components/features/conversation-panel/conversation-panel-list-helpers.ts`, Lines 214-234] **Conversations become permanently unreachable**: `filterToGroupDiscoveryPages()` removes every conversation for a group that arrived after that group's first backend page, and `groupConversations()` receives only this filtered array. The folder's own **More** control can expand only the retained `group.conversations`; there is no separate fetch or unfiltered backing collection. The new component test actually codifies the data loss: page 2 contains two more `No workspace` conversations, but after all three pages load, expanding that folder is expected to show only the six from page 1 rather than all eight. Keep the complete loaded conversations as the folder's data source and freeze only the collapsed preview/discovery presentation—do not discard loaded rows from the group model.

**[TESTING GAPS]**
- [`__tests__/components/features/conversation-panel/conversation-panel.test.tsx`] Change the regression test to expand the already-discovered folder after pagination and assert that every loaded conversation across later pages is reachable. It should simultaneously prove the collapsed preview stays stable while global **Load more** discovers folders.

**[RISK ASSESSMENT]**
This does not delete conversations, but it makes legitimate history invisible in grouped mode for any folder spanning backend pages. Users with active or long-lived workspaces are the most likely to hit it, and the current tests lock in the regression.

**VERDICT:**
❌ **Needs rework**: Separate folder discovery/preview state from the complete loaded conversation data.

**KEY INSIGHT:**
The UI needs two views of one dataset—a stable collapsed preview and a complete expandable group—not a destructive filter that turns presentation rules into data loss.

---

>
>
>

FraterCCCLXIII:
@hieptl @enyst — addressed your review feedback (grouped Load more no longer drops later-page conversations; per-click page fetch is capped; active conversation stays visible). Could you take another look when you have a moment?

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [`src/components/features/conversation-panel/conversation-panel.tsx`, Lines 495-504] **Pagination / No-op interaction**: Grouped mode treats any increase in `groupedSourceConversationCount` as success and stops the load-more driver, even when `visibleGroupCount` did not grow. Those newly fetched rows are deliberately excluded from the existing folder's collapsed preview, so the click can produce no visible change at all. The new test at `conversation-panel.test.tsx:1778-1796` codifies that failure by requiring a second click before the next folder appears. Continue until a new folder appears, pagination ends, or the existing three-page safety cap is reached; hidden contents of an existing folder are not success for the global **Load more** control.

**[TESTING GAPS]**
- [`__tests__/components/features/conversation-panel/conversation-panel.test.tsx`, Lines 1778-1796] The regression test currently asserts the broken two-click behavior. Keep the stable-preview and expanded-folder assertions, but require one global click to walk past an existing-folder-only page and reveal the next folder within the page cap.

**[RISK ASSESSMENT]**
The change is confined to client-side conversation grouping, but it alters shared pagination state and currently leaves the reported no-op interaction intact for common multi-page folders.

**VERDICT:**
❌ **Needs rework**: The global control must visibly discover a folder per click when one is reachable within the bounded fetch window.

**KEY INSIGHT:**
Progress for a folder-discovery control must be measured in visible folders, not hidden rows added to an existing folder.

---

>
>
>

hieptl (approved):
Thank you! 🙏

PR #16200, opened 2026-07-30
fix(home): don't gate the composer on the seeded default profile's LLM ref

### LINKED ISSUE #16193: [Bug]: Composer blocks local default AgentProfile despite a valid active LLM profile
## Description

Agent Canvas can disable the conversation composer and show the “LLM not configured” warning even though the active local LLM profile is valid and is the configuration that conversation creation would actually use.

This happens when the active, well-known `default` OpenHands AgentProfile retains an older `llm_profile_ref`.

## Reproduction

1. Use a local backend with two LLM profiles:
   - Profile A is active and has `api_key_set: true`.
   - Profile B exists, uses `auth_type: "api_key"`, and has no API key.
2. Have the active OpenHands AgentProfile be the seeded `default` profile, with `llm_profile_ref` still pointing to Profile B.
3. Open the conversations/home page.

Observed on Agent Server 1.37.0.

## Actual behavior

The UI reports that the LLM is not configured and disables the composer.

`useLlmConfigured` resolves the existing Profile B through the active AgentProfile's `llm_profile_ref`. Because Profile B has neither a key nor subscription authentication, the hook returns `isConfigured: false`. It does not fall back to active Profile A because the reference is present and resolves successfully.

## Expected behavior

The composer should be enabled when the LLM configuration that the conversation launcher will use is valid.

For a local OpenHands AgentProfile named `default`, `useCreateConversation` deliberately clears `effectiveAgentProfileId` and launches through `agent_settings`. That path uses active Profile A, not the `default` AgentProfile's stale Profile B reference. The preflight gate and the launch path therefore disagree.

## Root cause

The two frontend paths apply different resolution rules:

- `useLlmConfigured` treats any existing `llm_profile_ref` on an OpenHands AgentProfile as authoritative.
- `useCreateConversation` special-cases the local, well-known OpenHands `default` AgentProfile and launches through `agent_settings` so Canvas enrichments survive.

As a result, the gate blocks a conversation that the launch path is configured to run.

## Suggested fix

Make `useLlmConfigured` mirror `useCreateConversation`:

- For the local, well-known OpenHands `default` AgentProfile, validate the active LLM profile from the profile list/global settings path.
- Continue validating `llm_profile_ref` for named OpenHands AgentProfiles and for Cloud, where profile-based launch remains authoritative.

Add a regression test with:

- active LLM Profile A: keyed and usable;
- active local AgentProfile: OpenHands `default`;
- `default.llm_profile_ref`: existing but unauthenticated Profile B;
- expected result: `isConfigured === true`.

Keeping the seeded `default.llm_profile_ref` synchronized when the active LLM profile changes may also reduce stale state, but the frontend gate should still agree with the launch semantics.

## Workaround

Edit the `default` AgentProfile and point its LLM profile reference at the active, usable LLM profile.

VascoSch92:
HUMAN:

This solve a bug I found this morning in agent-canvas.

AGENT:

---

## Why

The preflight LLM gate and the launch path disagreed for the seeded `default` agent profile, so the composer was disabled for a conversation that would have started fine.

`useCreateConversation` treats the local, well-known OpenHands `default` profile as the enriched baseline: it clears `effectiveAgentProfileId` and launches through `agent_settings`, which runs the **active LLM profile** — not the seed's `llm_profile_ref`. `useLlmConfigured` still validated that ref, so a `default` profile whose ref pointed at an existing-but-unauthenticated LLM profile returned `isConfigured: false`. The existing stale-ref fallback didn't catch it because the referenced profile *exists* — it just has no key.

Result: the composer was disabled and the "LLM not configured" banner shown, even though the active LLM profile was valid and is what the launch would have used.

## Summary

- `useLlmConfigured` now skips `llm_profile_ref` for the local, well-known OpenHands `default` agent profile and validates the active LLM profile instead, mirroring `useCreateConversation`'s launch semantics.
- Named OpenHands profiles and cloud are unchanged — they launch from the profile, so the ref stays authoritative there.
- Added two tests: the regression case from the issue, plus a guard that the fallback still reports unconfigured when the active LLM profile has no key either.

## Issue Number

Fixes #16193

## How to Test

Automated — the regression test encodes the reported scenario (active LLM profile A keyed, active agent profile is OpenHands `default`, its `llm_profile_ref` points at existing-but-keyless profile B, expected `isConfigured === true`):

```bash
npm ci
npx vitest run __tests__/hooks/use-llm-configured.test.tsx src/hooks/use-llm-configured.test.tsx
```

It fails on `main` with `expected false to be true` and passes with this change.

Manually, against a local backend:

1. Create two LLM profiles — A with an API key (set it active), B with `auth_type: "api_key"` and no key.
2. Point the seeded `default` agent profile's LLM profile reference at B, and leave `default` as the active agent profile.
3. Open the home/conversations page. Before this change the composer is disabled with the "LLM not configured" warning; after it, the composer is enabled and a conversation launches on profile A.

## Video/Screenshots

Not included — the change is to a hook's boolean result, covered by the tests above. I could not exercise the UI end to end, since reproducing it needs a local agent-server with the stale seeded `default` ref described in the issue.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- No migration or config change; the non-comment change is one import plus one negated condition in the existing ternary.
- The `isLocal` half of the new condition is redundant *today*: on cloud, `hasUsableLlm` reads `settings.llm_api_key_set` and `shouldLoadActiveProfileDetail` is already `isLocal`-gated, so nothing downstream of the resolved profile is observable there. It's kept to mirror `useCreateConversation`'s own `!isCloud` guard, so the two paths stay in step if this hook ever becomes profile-driven for cloud. Happy to drop it if you'd rather not carry a redundant term.
- The issue also suggests keeping the seeded `default.llm_profile_ref` synchronized when the active LLM profile changes. That's left as a follow-up: it reduces stale state, but the frontend gate still needs to agree with the launch semantics on its own, which is what this PR fixes.
- Verified locally: `npm run typecheck` clean, eslint + prettier clean on the changed files, and the full suite at 4024 passing (one unrelated worker-startup timeout in `__tests__/utils/format-time-delta.test.ts` that passes in isolation).

---

| Component | Value |
|---|---|
| **Commit** | `fa0167bdd7718b5a8d45725f49e26781f255883c` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-fa0167b
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-fa0167b
```

```
ghcr.io/openhands/agent-canvas:sha-fa0167b-amd64
ghcr.io/openhands/agent-canvas:vasco-fix-llm-configured-default-profile-amd64
ghcr.io/openhands/agent-canvas:pr-16200-amd64
ghcr.io/openhands/agent-canvas:sha-fa0167b-arm64
ghcr.io/openhands/agent-canvas:vasco-fix-llm-configured-default-profile-arm64
ghcr.io/openhands/agent-canvas:pr-16200-arm64
ghcr.io/openhands/agent-canvas:sha-fa0167b
ghcr.io/openhands/agent-canvas:vasco-fix-llm-configured-default-profile
ghcr.io/openhands/agent-canvas:pr-16200
```

- Each tag (e.g., `sha-fa0167b`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-fa0167b-amd64`) are also available if needed


enyst (commented):

🟢 **Good taste** — This is a focused fix that removes a real contradiction instead of layering another workaround on top.

The gate now recognizes the same local OpenHands `default` profile case as `useCreateConversation`, using the shared well-known profile constant and preserving named-profile and cloud behavior. The regression tests cover both the reported keyed-active/keyless-ref scenario and the important negative case where the active LLM is also unusable. I found no actionable correctness, compatibility, security, or testing issue in this head.

**VERDICT:**
✅ **Worth merging**: The preflight decision now matches the launch path, and the regression boundary is tested.

**KEY INSIGHT:**
A launch preflight must evaluate the configuration that the launch path will actually consume, not stale profile metadata that the launcher deliberately bypasses.

**[RISK ASSESSMENT]**
This changes one narrowly scoped frontend decision, follows the launcher's existing predicate, adds no dependency or security exposure, and has focused regression coverage.

enyst:
TBH, this sounds confusing to me:
> useCreateConversation treats the local, well-known OpenHands default profile as the enriched baseline: it clears effectiveAgentProfileId and launches through agent_settings, which runs the active LLM profile — not the seed's llm_profile_ref.

...Why? I mean, I'm reading that it ignores the 'default' agent profile, and sets up the agent from the old agent_settings instead. Then why do we have a default profile if we ignore it?

Just to note, this isn't just a theoretical concern with clarity; my agent used the default profile and ended up with conversations without terminal/file tools. API contracts need to be clean for agents :)

Canvas appears to work because of this weird workaround... ignoring the profile.


I think maybe we can accept now the PR fix though. I think we can take two paths from here:
- disable agent profiles, thus yes use agent_settings instead; until we figure out how agent profiles need to work 
- make an issue to look into agent profiles and straighten out this behavior

I'd suggest a follow up _disabling_ agent profiles for openhands, that is, put it under a feature flag, and/or comment heavily about how it should work (never use agent settings anymore) and that it's WIP

__tests__/hooks/use-llm-configured.test.tsx
```diff
@@ -126,6 +126,47 @@ describe("useLlmConfigured (local, agent-profile-driven)", () => {
     expect(result.current.isConfigured).toBe(true);
   });
 
+  it("is configured for the local seeded `default` agent profile when its ref is unauthenticated (matches the agent_settings launch)", () => {
+    // #16193: "B" EXISTS (so the stale-ref fallback above doesn't fire) but has
+    // no key. `default` launches via `agent_settings` on the active LLM "A".
```
enyst, line 131: These are the same thing though: if default agent profile is active, then the active llm should be part of it...

src/hooks/use-llm-configured.ts
```diff
@@ -6,6 +6,7 @@ import { useActiveBackend } from "#/contexts/active-backend-context";
 import { useActiveAgentProfile } from "#/hooks/use-active-agent-profile";
 import { isSettingsPageHidden } from "#/utils/settings-utils";
 import ProfilesService from "#/api/profiles-service/profiles-service.api";
+import { WELL_KNOWN_DEFAULT_AGENT_PROFILE_NAME } from "#/api/agent-profiles-service/agent-profiles-service.api";
```
enyst, line 9: It looks to me like we name something "well known" so that we know to bypass it? 😅

enyst (approved):
I would rather approve the PR, I left a few comments on why it seems weird to me though. 

I approve because the other path started this bypass... 🤔If I may, I recommend to comment heavily that it's temporary and we should not use the old agent settings actually -if I understand correctly

src/hooks/use-llm-configured.ts
VascoSch92: haha you are right :-)

PR #1656, opened 2024-05-08
feat: support for stopping automatic scrolling to bottom and add "to bottom" button

Shimada666:
@amanape @Sparkier 
I have written a react hook to support stopping automatic scrolling to bottom and displaying a "To Bottom" button. It is currently effective in Jupyter. Please take a look at it. If everything is fine, I will apply it to the chat UI. 😁

PR #16220, opened 2026-07-31
fix: reconcile non-native tool-call streamed <function=...> XML delta

sideeffffect:
HUMAN:

I ran the tests locally. It fixes a rendering problem

image

AGENT:

---

Re-opened here from OpenHands/agent-canvas#1852 per @neubig's request after the repository move. Rebased onto current `main`, comments trimmed per @VascoSch92's review on the original PR, and the diff reduced to the one case still broken on `main` (see Notes).

## Why

With `native_tool_calling: false`, tool schemas are rendered as prompted XML in the system prompt and the model replies with plain text like:

```
Coding and executing<function=terminal>
<parameter=command>echo hi</parameter>
<parameter=security_risk>LOW</parameter>
</function>
```

The SDK strips `<function=...` and everything after it from the persisted `thought` only once the *complete* response is available (`_strip_function_call_from_content` in `fn_call_converter.py`). A `StreamingDeltaEvent`, by contrast, is filled straight from raw litellm stream chunks — its own docstring notes it "bypasses the callback chain that writes to `ConversationState.events`" — so while the turn is in flight it still carries that raw XML.

`supersedeStreamedThoughtWithAction` exists to clear the superseded delta once the parsed `ActionEvent` lands (issue #1534), but it gates on `matchStreamedSegments`, which requires the streamed text to be a **prefix** of the clean final text. Here the streamed text is `thought` **plus** the raw XML — a *superset*, never a prefix — so the match silently fails, the delta is left in place, and the raw XML renders alongside the parsed action. Each tool-calling step then shows its unparsed `<function=...>` block plus a duplicated narration/summary line.

A useful confirmation of the mechanism, reported on the original PR: the artifacts **disappear if you switch to another conversation and back**. That path rebuilds `uiEvents` from `useConversationHistory` → `EventService.searchEvents`, i.e. the *persisted* event store, which never contains `StreamingDeltaEvent`s — so only the clean parsed events remain and it renders correctly. The bug is live-streaming-only, and the post-switch render is effectively the ground-truth output this change makes the live view converge to.

## Summary

- Add `hasUnstrippedFunctionCallMarker(segments)`: a `<function=` marker can only appear in streamed content in this exact non-native, mid-stream case, so its presence is an unambiguous second reconciliation signal.
- Use it in `supersedeStreamedThoughtWithAction` alongside the existing prefix match, so the superset case is reconciled instead of rendering twice.

## Issue Number

N/A (extends the fix for #1534 to the non-native tool-calling case it didn't cover)

## How to Test

```bash
npm ci
npm run make-i18n
npx vitest run __tests__/utils/handle-event-for-ui.test.ts
```

43 tests pass, including the new `clears the delta when streamed text is the thought plus an unstripped <function=...> block`.

Verified it is a genuine regression test rather than a tautology — reverting only the source change and re-running just that case fails, and passes again with the fix restored:

```
$ git checkout -- src/utils/handle-event-for-ui.ts
$ npx vitest run __tests__/utils/handle-event-for-ui.test.ts -t "unstripped <function="
 Test Files  1 failed (1)
      Tests  1 failed | 42 skipped (43)

# fix restored
 Test Files  1 passed (1)
      Tests  43 passed (43)
```

Full gate, from a clean `npm ci` on this branch:

```
npx vitest run     # 533 test files passed | 1 skipped (534)
                   # 4131 tests passed | 5 skipped | 9 todo (4145)  -- zero regressions
npm run typecheck  # clean, 0 errors
npx eslint src/utils/handle-event-for-ui.ts __tests__/utils/handle-event-for-ui.test.ts   # clean
npx prettier --check src/utils/handle-event-for-ui.ts __tests__/utils/handle-event-for-ui.test.ts   # clean
```

## Video/Screenshots

Screenshots of the bug (unpatched build), plus the switch-away-and-back render that shows the clean expected output, are on the original PR: OpenHands/agent-canvas#1852.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The original PR also handled the `FinishAction` / agent-`MessageEvent` case, but `main` has since rewritten `finalizeStreamingDeltasInPlace` to supersede the current turn's streamed deltas unconditionally (no `matchStreamedSegments` gate), which covers that path already. So this PR is reduced to just the intermediate-action path, which still gates on the prefix match and is still affected.

Two related fixes for the same `native_tool_calling: false` scenario, both independent of this rendering bug:
- OpenHands/software-agent-sdk#4153 — the prompted-tool-call validator rejects a `security_risk` parameter on read-only tools like `finish` (the in-context examples teach every model to send it, but read-only tools never declare it), costing a round-trip and printing a validation error into the transcript.
- A client-side scrub of that same parameter on the OpenAI-compatible backend used in the reproduction.


enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [src/utils/handle-event-for-ui.ts, Line 281] **Cross-agent data loss**: This fallback treats a `<function=` marker in any trailing content delta as proof that the delta belongs to this action. That invariant does not hold in this shared event store: main-agent and planning-agent sockets can interleave, and `getTrailingContentDeltas` is not sender-scoped even though the final-event and reasoning paths explicitly are. If a main-agent action lands while the trailing planning-agent delta contains prompted-tool XML (or vice versa), `supersedeStreamingContent` silently deletes the other agent's live output. Filter the candidate content deltas with `isSameStreamingSender(action, event)` before applying either reconciliation signal, and add a regression test with a marker-bearing delta from the opposite sender.

**[RISK ASSESSMENT]**
The change is small and covered for the single-agent reproduction, but it alters shared live-stream reconciliation and can remove another agent's in-flight UI output. Existing sender-boundary code and tests make the missing scope a concrete concurrency regression rather than a theoretical edge case.

**VERDICT:**
❌ **Needs rework**: Preserve sender identity before allowing the marker fallback to supersede streamed content.

**KEY INSIGHT:**
A syntax marker can identify raw tool-call content, but it cannot establish which concurrently streaming agent owns that content.

---

>
>
>

sideeffffect:
Good catch — confirmed and fixed in cc41ff1.

The invariant really doesn't hold: `getCurrentTurnContentDeltas` and `getTrailingReasoningDeltas` are both sender-scoped, but `getTrailingContentDeltas` was not, so the marker fallback could let a main-agent action supersede a planning-agent delta (and vice versa).

Fix: scope `getTrailingContentDeltas` by `isSameStreamingSender`, mirroring `getTrailingReasoningDeltas` (#1656). Scoping the candidate set rather than the individual signals means *both* reconciliation paths — the thought match and the marker fallback — only ever see the acting agent's own deltas.

Added the regression test you asked for: a marker-bearing delta with `isFromPlanningAgent: true` against a main-agent action must be left untouched. Verified it's a real regression test — reverting only the scoping makes it fail, restoring it passes:

```
# scoping reverted
Tests  1 failed | 43 skipped (44)
# scoping restored
Tests  44 passed (44)
```

Also merged current `main` (it had moved). Full gate on the branch: `vitest run` 534 files / 4180 tests passed, 0 regressions; `tsc` clean; eslint + prettier clean.

enyst (commented):

🟢 **Good taste**

The latest commit addresses the concrete cross-agent data-loss issue from the previous review: trailing content deltas are now sender-scoped before either the thought match or `<function=` fallback can supersede them. The new regression test proves that a marker-bearing planning-agent delta survives a main-agent action, while the original non-native tool-call duplication case remains covered.

**VERDICT:**
✅ **Worth merging**: The fallback is now constrained by the same sender invariant used elsewhere in streaming reconciliation.

**KEY INSIGHT:**
Syntax identifies raw prompted-tool content, while sender identity determines whether that content belongs to the arriving action.

**[RISK ASSESSMENT]**
The change touches shared live-stream reconciliation, where a false match can hide in-flight UI output, so the area remains moderately sensitive. The implementation reuses the existing sender predicate, has focused regression coverage for both the target bug and concurrent-agent boundary, and passes the repository's Linux, Windows, and mock-E2E checks.

sideeffffect:
Is there something more I'm expected to do so this can be reviewed/merged? 🙏

VascoSch92 (approved):

PR #16229, opened 2026-07-31
feat: Add support for Pi as ACP agent

### LINKED ISSUE #16204: [Feature] Support Pi as ACP Agent
A number of users have been asking about supporting Pi as an ACP harness within Canvas. This should be supported like Claude, Gemini, and ChatGPT/Codex.

Deep070203:
HUMAN: 
I tested the Pi preset configuration end-to-end inside the local docker image, and verified it successfully saves the agent profile.

AGENT:

To verify end-to-end functionality, we built the Docker image locally and ran the container:
```bash
npm run build:docker
docker run -it --rm -p 8000:8000 -v "$HOME/.openhands:/home/openhands/.openhands" -v "${PROJECTS_PATH}:/projects" agent-canvas:local
```

We opened the Settings page, created an agent profile named MyPi choosing the Pi preset, and successfully saved it.

Server logs showing successful profile creation (HTTP 201) and deletion (HTTP 200):
```
{"asctime": "2026-07-31 19:26:30,327", "levelname": "INFO", "name": "openhands.sdk.profiles.agent_profile_store", "filename": "agent_profile_store.py", "lineno": 174, "message": "[AgentProfile Store] Saved profile `MyPi` at /home/openhands/.openhands/agent-profiles/MyPi.json"}
{"asctime": "2026-07-31 19:26:30,328", "levelname": "INFO", "name": "openhands.agent_server.agent_profiles_router", "filename": "agent_profiles_router.py", "lineno": 377, "message": "Saved agent profile 'MyPi'"}
{"asctime": "2026-07-31 19:26:30,329", "levelname": "INFO", "name": "uvicorn.access", "message": "::ffff:192.168.65.1:0 - \"POST /api/agent-profiles/MyPi HTTP/1.1\" 201", "http.client_ip": "::ffff:192.168.65.1:0", "http.method": "POST", "http.url": "/api/agent-profiles/MyPi", "http.version": "1.1", "http.status_code": 201}

{"asctime": "2026-07-31 19:26:52,760", "levelname": "INFO", "name": "openhands.sdk.profiles.agent_profile_store", "filename": "agent_profile_store.py", "lineno": 228, "message": "[AgentProfile Store] Deleted profile `MyPi`"}
{"asctime": "2026-07-31 19:26:52,769", "levelname": "INFO", "name": "openhands.agent_server.agent_profiles_router", "filename": "agent_profiles_router.py", "lineno": 411, "message": "Deleted agent profile 'MyPi'"}
{"asctime": "2026-07-31 19:26:52,772", "levelname": "INFO", "name": "uvicorn.access", "message": "::ffff:192.168.65.1:0 - \"DELETE /api/agent-profiles/MyPi HTTP/1.1\" 200", "http.client_ip": "::ffff:192.168.65.1:0", "http.method": "DELETE", "http.url": "/api/agent-profiles/MyPi", "http.version": "1.1", "http.status_code": 200}
```

## Why

This PR adds UI and configuration support for Pi Coding Agent (pi.dev) in the ACP (Agent Client Protocol) provider options.

## Summary

- Registered Pi Preset: Added pi to the frontend's local ACP provider registry with appropriate defaults (command: npx -y pi-acp, model: "default").
- Added SVG Branding: Sourced and added the official Pi SVG icon.
- Added Localized Translations: Translated description keys for all 15 supported locales in translation.json.
- Wire-level Fallbacks: Mapped "pi" to "custom" on the wire when saving settings or profile drafts to bypass strict backend validation (preventing 422 Unprocessable Content errors).
- Tag-detection Logic: Updated getAcpServerTag to detect if the customized command is npx -y pi-acp, allowing active conversations to be stamped with tags.acpserver: "pi" and display the correct branding in conversation lists.

## Issue Number
Fixes: https://github.com/OpenHands/OpenHands/issues/16204

## How to Test
1. Build the image
```
npm run build:docker
```

2. Start the container:
```
export PROJECTS_PATH="$HOME/Downloads"
docker run -it --rm -p 8000:8000 -v "$HOME/.openhands:/home/openhands/.openhands" -v "${PROJECTS_PATH}:/projects" agent-canvas:local
```
3. Navigate to Settings -> Agent Profiles, click Add agent profile, and select the Pi preset.
4. Verify that you can configure it and click Save successfully without encountering a 422 error.

## Video/Screenshots

Screenshot 2026-07-31 at 3 38 34 PM
Screenshot 2026-07-31 at 3 38 42 PM
Screenshot 2026-07-31 at 3 38 50 PM

## Type

- [ ] Bug fix
- [X] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.57s · commit 8911f44  
**Strongest signal:** Unverified remote execution · 26% estimated likelihood.  
**Evidence:** No direct hunk selected.  
**Coverage:** ⚠️ reduced context — partial coverage; 33/66 hunks, 25/25 files (context budget: 33, hunk budget: 33).

<details>

| --- | --- | --- |
| Command injection | 24.0% | No direct hunk selected |
| Weakened authentication | 7.0% | No direct hunk selected |
| Weakened authorization | 9.0% | No direct hunk selected |
| Contract regression | 27.0% | No direct hunk selected |
| Data loss | 8.0% | No direct hunk selected |
| Sensitive data disclosure | 6.0% | No direct hunk selected |
| Unexpected data transfer | 8.0% | No direct hunk selected |
| Credential misuse | 8.0% | No direct hunk selected |
| Untrusted instruction authority | 4.0% | No direct hunk selected |
| Package source redirection | 6.0% | No direct hunk selected |
| Unverified remote execution | 26.0% | No direct hunk selected |
| Privileged environment access | 9.0% | No direct hunk selected |
| Security assessment bypass | 13.0% | No direct hunk selected |
| Prohibited workload | 3.0% | No direct hunk selected |
| Primary concern | Unverified remote execution; confidence 40.0% | No direct hunk selected |

</details>

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -67,6 +67,12 @@ RUN node -e " \
   require('fs').writeFileSync('/tmp/defaults.env', lines.join('\n') + '\n'); \
 "
 
+# ── Stage 1c: Pi ACP (Node 22+ required; avoid Debian apt nodejs) ─────────────
+FROM node:24-slim AS pi-acp-install
+RUN npm install -g --ignore-scripts \
+  @earendil-works/pi-coding-agent \
+  pi-acp
+
 # ── Stage 2: Combined image ──────────────────────────────────────────────────
 FROM ${AGENT_SERVER_IMAGE} AS final
 
@@ -93,23 +99,21 @@ USER root
 
 # Install system deps required by automation's transitive dependencies
 # (asyncpg needs libpq, which the agent-server base image may not include).
-# Also install Node/npm for Pi ACP (pi-acp spawns via npx; the pi coding-agent
-# CLI is a separate global npm package). Claude/Codex/Gemini ACP wrappers ship
-# in the agent-server base image; Pi is installed here for Docker parity.
 RUN if command -v apt-get >/dev/null 2>&1; then \
       apt-get update && \
-      apt-get install -y --no-install-recommends libpq-dev nodejs npm && \
+      apt-get install -y --no-install-recommends libpq-dev && \
       rm -rf /var/lib/apt/lists/*; \
     fi
 
-# Pi coding agent + ACP adapter (stdio JSON-RPC → ``pi --mode rpc``).
-# ``--ignore-scripts`` matches upstream Pi install guidance; lifecycle scripts
-# are not required for headless ACP use.
-RUN if command -v npm >/dev/null 2>&1; then \
-      npm install -g --ignore-scripts \
-        @earendil-works/pi-coding-agent \
-        pi-acp; \
-    fi
+# Pi coding agent + ACP adapter. Claude/Codex/Gemini wrappers ship in the
+# agent-server base; Pi is third-party and installed via the Node 24 stage
+# above (pi-acp requires Node 22+).
+COPY --from=pi-acp-install /usr/local/bin/node /usr/local/bin/node
+COPY --from=pi-acp-install /usr/local/bin/npm /usr/local/bin/npm
+COPY --from=pi-acp-install /usr/local/bin/npx /usr/local/bin/npx
+COPY --from=pi-acp-install /usr/local/bin/pi /usr/local/bin/pi
+COPY --from=pi-acp-install /usr/local/bin/pi-acp /usr/local/bin/pi-acp
+COPY --from=pi-acp-install /usr/local/lib/node_modules /usr/local/lib/node_modules
 
 # Install automation server via pip.
 # Shared deps (openhands-sdk, fastapi, uvicorn, pydantic, httpx, …) are
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -109,17 +109,12 @@ RUN if command -v apt-get >/dev/null 2>&1; then \
 # agent-server base; Pi is third-party and installed via the Node 24 stage
 # above (pi-acp requires Node 22+).
 #
-# Copy the Node 24 toolchain, then re-run ``npm install -g`` in this stage.
-# Do NOT ``COPY`` /usr/local/bin/pi or pi-acp directly — Docker dereferences
-# npm global symlinks into flat scripts that break relative imports
-# (``config.js`` lives in the package dir, not next to bin/pi).
-COPY --from=pi-acp-install /usr/local/bin/node /usr/local/bin/node
-COPY --from=pi-acp-install /usr/local/bin/npm /usr/local/bin/npm
-COPY --from=pi-acp-install /usr/local/bin/npx /usr/local/bin/npx
-COPY --from=pi-acp-install /usr/local/lib/node_modules /usr/local/lib/node_modules
-RUN npm install -g --ignore-scripts \
-  @earendil-works/pi-coding-agent \
-  pi-acp
+# Copy the whole Node prefix so npm global bin *symlinks* stay intact.
+# Never ``COPY`` /usr/local/bin/pi or pi-acp alone — Docker dereferences
+# symlinks into flat scripts that import ./config.js from bin/ instead of
+# the package directory under lib/node_modules.
+COPY --from=pi-acp-install /usr/local/ /opt/node24/
+ENV PATH="/opt/node24/bin:${PATH}"
 
 # Install automation server via pip.
 # Shared deps (openhands-sdk, fastapi, uvicorn, pydantic, httpx, …) are
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -93,10 +93,22 @@ USER root
 
 # Install system deps required by automation's transitive dependencies
 # (asyncpg needs libpq, which the agent-server base image may not include).
+# Also install Node/npm for Pi ACP (pi-acp spawns via npx; the pi coding-agent
+# CLI is a separate global npm package). Claude/Codex/Gemini ACP wrappers ship
+# in the agent-server base image; Pi is installed here for Docker parity.
 RUN if command -v apt-get >/dev/null 2>&1; then \
       apt-get update && \
-      apt-get install -y --no-install-recommends libpq-dev && \
+      apt-get install -y --no-install-recommends libpq-dev nodejs npm && \
       rm -rf /var/lib/apt/lists/*; \
+    fi
+
+# Pi coding agent + ACP adapter (stdio JSON-RPC → ``pi --mode rpc``).
+# ``--ignore-scripts`` matches upstream Pi install guidance; lifecycle scripts
+# are not required for headless ACP use.
+RUN if command -v npm >/dev/null 2>&1; then \
+      npm install -g --ignore-scripts \
+        @earendil-works/pi-coding-agent \
+        pi-acp; \
     fi
 
 # Install automation server via pip.
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -70,8 +70,8 @@ RUN node -e " \
 # ── Stage 1c: Pi ACP (Node 22+ required; avoid Debian apt nodejs) ─────────────
 FROM node:24-slim AS pi-acp-install
 RUN npm install -g --ignore-scripts \
-  @earendil-works/pi-coding-agent \
-  pi-acp
+  @earendil-works/pi-coding-agent@0.84.0 \
+  pi-acp@0.0.33
 
 # ── Stage 2: Combined image ──────────────────────────────────────────────────
 FROM ${AGENT_SERVER_IMAGE} AS final
@@ -114,9 +114,9 @@ RUN if command -v apt-get >/dev/null 2>&1; then \
 # symlinks into flat scripts that import ./config.js from bin/ instead of
 # the package directory under lib/node_modules.
 COPY --from=pi-acp-install /usr/local/ /opt/node24/
-ENV PATH="/opt/node24/bin:${PATH}"
+
 RUN test -x /opt/node24/bin/pi-acp && test -x /opt/node24/bin/pi && \
-    /opt/node24/bin/pi --version
+    /opt/node24/bin/pi --version && /opt/node24/bin/pi-acp
 
 # Install automation server via pip.
 # Shared deps (openhands-sdk, fastapi, uvicorn, pydantic, httpx, …) are
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -108,12 +108,18 @@ RUN if command -v apt-get >/dev/null 2>&1; then \
 # Pi coding agent + ACP adapter. Claude/Codex/Gemini wrappers ship in the
 # agent-server base; Pi is third-party and installed via the Node 24 stage
 # above (pi-acp requires Node 22+).
+#
+# Copy the Node 24 toolchain, then re-run ``npm install -g`` in this stage.
+# Do NOT ``COPY`` /usr/local/bin/pi or pi-acp directly — Docker dereferences
+# npm global symlinks into flat scripts that break relative imports
+# (``config.js`` lives in the package dir, not next to bin/pi).
 COPY --from=pi-acp-install /usr/local/bin/node /usr/local/bin/node
 COPY --from=pi-acp-install /usr/local/bin/npm /usr/local/bin/npm
 COPY --from=pi-acp-install /usr/local/bin/npx /usr/local/bin/npx
-COPY --from=pi-acp-install /usr/local/bin/pi /usr/local/bin/pi
-COPY --from=pi-acp-install /usr/local/bin/pi-acp /usr/local/bin/pi-acp
 COPY --from=pi-acp-install /usr/local/lib/node_modules /usr/local/lib/node_modules
+RUN npm install -g --ignore-scripts \
+  @earendil-works/pi-coding-agent \
+  pi-acp
 
 # Install automation server via pip.
 # Shared deps (openhands-sdk, fastapi, uvicorn, pydantic, httpx, …) are
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -115,6 +115,8 @@ RUN if command -v apt-get >/dev/null 2>&1; then \
 # the package directory under lib/node_modules.
 COPY --from=pi-acp-install /usr/local/ /opt/node24/
 ENV PATH="/opt/node24/bin:${PATH}"
+RUN test -x /opt/node24/bin/pi-acp && test -x /opt/node24/bin/pi && \
+    /opt/node24/bin/pi --version
 
 # Install automation server via pip.
 # Shared deps (openhands-sdk, fastapi, uvicorn, pydantic, httpx, …) are
--- a/src/api/agent-server-adapter.ts
+++ b/src/api/agent-server-adapter.ts
@@ -4,9 +4,12 @@ import { DEFAULT_SETTINGS } from "#/services/settings";
 import { ExecutionStatus } from "#/types/agent-server/core";
 import { AgentKind, Settings, SettingsValue } from "#/types/settings";
 import {
+  defaultAcpCommandForProvider,
   getAcpPreferredDefaultModel,
   getAcpProvider,
+  PI_ACP_PROVIDER_KEY,
   resolveEffectiveAcpModel,
+  resolveUiAcpProviderKey,
 } from "#/constants/acp-providers";
 import { getAgentServerClientOptions } from "./agent-server-client-options";
 import { isAgentServerToolAvailable } from "./agent-server-compatibility";
@@ -703,11 +706,12 @@ function isAcpAgent(settings: Settings): boolean {
 function getAcpServerTag(settings: Settings): string | undefined {
   const agentSettings = toRecord(settings.agent_settings);
   const value = agentSettings.acp_server;
-  if (value === "custom" || !value) {
-    const cmd = agentSettings.acp_command;
-    if (Array.isArray(cmd) && cmd.join(" ") === "npx -y pi-acp") {
-      return "pi";
-    }
+  const uiKey = resolveUiAcpProviderKey(
+    typeof value === "string" ? value : undefined,
+    agentSettings.acp_command,
+  );
+  if (uiKey === PI_ACP_PROVIDER_KEY) {
+    return PI_ACP_PROVIDER_KEY;
   }
   return typeof value === "string" && value.length > 0 ? value : undefined;
 }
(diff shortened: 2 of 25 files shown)


enyst (commented):
hey @Deep070203 , thank you for the PR! Could you tell how does `pi` use an LLM?

We'd love it if you could register a video to show it works, I see 'default' in the image but TBH I'm not sure I understand how it chooses the LLM

enyst (commented):

🔴 **Needs improvement**

**[CRITICAL ISSUES]**
- [src/routes/agent-settings.tsx, Line 162] **Broken launch configuration**: Mapping the Pi preset to `acp_server: "custom"` while treating its default command as implicit produces an unusable profile. This branch writes `acp_command: null`; `buildAcpAgentSettingsDiff` likewise writes `custom` with an empty command, and the onboarding profile path writes `custom` without any command. The SDK explicitly rejects `custom` without `acp_command` because there is no registry default to resolve. When Pi must be represented as `custom` for compatibility, persist `npx -y pi-acp` explicitly in every settings/profile path, and add regression coverage that carries the saved payload through conversation/profile resolution rather than asserting the currently broken `null` value.
- [src/constants/acp-providers.ts, Line 166] **Supply Chain Risk**: `npx -y pi-acp` installs an unpinned latest package at conversation launch. Today that resolves to `pi-acp@0.0.33`, published on npm at `2026-07-30T17:06:15Z`—about one day before this review and inside the required seven-day waiting window. The package has a matching signed npm artifact, GitHub release, and `v0.0.33` source tag, with no install lifecycle hook in the published metadata, but freshness remains a blocking risk and the unpinned command would silently move users to future unreviewed releases. Pin an exact vetted version and wait until that release is at least seven days old, or have a human maintainer explicitly accept the freshness risk.

**[TESTING GAPS]**
- [__tests__/routes/build-agent-profile-fields.test.ts, Line 84] **Regression encoded as expected behavior**: This test expects `acp_server: "custom"` together with `acp_command: null`, exactly the combination the backend rejects. Replace it with an end-to-end payload assertion that proves the explicit Pi command survives settings/profile persistence and reaches ACP command resolution.

**[RISK ASSESSMENT]**
This adds execution of a new external package and currently persists a backend-invalid command configuration across multiple user-facing paths. **Recommendation**: Do not auto-merge. Request review from a human maintainer to validate the pinned package/version and the settings-to-launch data flow.

**VERDICT:**
❌ **Needs rework**: The saved Pi preset cannot reliably launch, and its runtime dependency is both unpinned and inside the seven-day freshness window.

**KEY INSIGHT:**
Once `pi` is erased to `custom` on the wire, the explicit command must become part of the persisted data because no later layer can recover Pi's registry default.

---

>
>
>

Deep070203:
Thanks for the review. The critical launch-configuration finding is valid for the current openhands/support-pi-acp head. We agree with the core insight:


> Once pi is erased to custom on the wire, the explicit command must become part of the persisted data because no later layer can recover Pi's registry default.


This comment explains why that happens, what we changed to fix it, what remains (version pinning), and how we validated it. 

Built-in ACP providers are registered in software-agent-sdk / @openhands/typescript-client. For those providers:

acp_server: "claude-code" (etc.) is enough
An empty or null acp_command is OK — the agent-server resolves the registry default at spawn time
Pi is not in that registry yet. On the wire it must travel as:

```
{
  "acp_server": "custom",
  "acp_command": ["npx", "-y", "pi-acp"]   // must be explicit — no registry fallback
}
```

Canvas keeps a UI registry key "pi" in src/constants/acp-providers.ts for onboarding, settings labels, and model UX, but maps it to "custom" via wireAcpServerForProvider() when persisting.

The original PR branch treated Pi like a built-in provider and wrote acp_command: null / []. That is invalid for custom and matches what the reviewer flagged.

| Review item | Status |
|---|---|
| `acp_command: null` for Pi + `custom` | **Fixed** — explicit command on all paths |
| Test encoding broken `null` | **Fixed** — expects `"npx -y pi-acp"` / `"pi-acp"` |
| End-to-end payload survival | **Covered** by unit tests + adapter tests; profile inlining closes the `agent_profile_id` bypass |
| Unpinned `npx -y pi-acp` | **Acknowledged** — propose pinning `0.0.33` or human sign-off on freshness |
| Auto-merge | **Request human review** for version pin policy |


### Supported environments

| Environment | Status | Notes |
|---|---|---|
| **Local dev** (`npm run dev`) | ✅ Supported | Spawns `npx -y pi-acp`; auth via host `~/.pi` |
| **Docker** (`agent-canvas` image) | ✅ Supported | Preinstalled `pi-acp`; mount `~/.pi`; rebuild image after changes |
| **OpenHands Cloud** | ❌ Not supported in this PR | No Pi in SDK registry/sandbox; no Cloud auth story for `~/.pi`; profile inlining is local-only |

Cloud support would need upstream SDK registration, sandbox image changes, and a credential path — out of scope here.

src/hooks/use-chat-input-model-state.ts
```diff
@@ -51,10 +52,22 @@ export function useChatInputModelState(): ChatInputModelState {
...
+  const profileUiAcpServerKey =
+    activeAcpProfile?.acp_server != null
+      ? (resolveUiAcpProviderKey(
+          activeAcpProfile.acp_server,
+          activeAcpProfile.acp_command,
+        ) ?? activeAcpProfile.acp_server)
+      : null;
   const acpServerKey = isActiveAcpConversation
-    ? conversation?.acp_server
+    ? (conversation?.acp_server ?? settingsUiAcpServerKey)
```
VascoSch92: Before this we left `acpServerKey` unresolved if `conversation.acp_server` was missing. Now we fall back to `settingsUiAcpServerKey` (the user's current global provider). Since `acp_server` can legitimately be null — profile launches don't stamp the tag client-side and the server doesn't always repopulate it — this means an open ACP conversation running e.g. Claude/Pi/Codex could get its model picker populated from whatever the user's current default provider happens to be, and they could switch to a model that doesn't apply to the actual running agent. Was this fallback intentional?

src/hooks/mutation/use-create-conversation.ts
```diff
@@ -193,21 +194,42 @@ export const useCreateConversation = () => {
         }
       }
 
+      // Pi profiles must launch via inlined agent_settings so the frontend can
+      // rewrite ``npx -y pi-acp`` → ``pi-acp`` in Docker (profile_id launch
+      // spawns the stored command verbatim on the agent-server).
+      let piAgentSettingsOverride: Record<string, SettingsValue> | undefined;
+      if (
+        !isCloud &&
```
VascoSch92, line 202: This inlining only runs when `!isCloud`, so on cloud we skip it entirely — no settings override, and the cloud branch of `createConversation` never calls `getAcpServerTag` either, so a cloud-launched Pi profile never gets the tag stamped and keeps the raw `'custom'` wire value. Is that intentional, or does cloud need the same treatment?

src/i18n/translation.json
```diff
@@ -30684,6 +30684,23 @@
...
     "uk": "Агент Gemini CLI від Google."
   },
+  "ONBOARDING$AGENT_PI_DESCRIPTION": {
+    "en": "Earendil's Pi Coding Agent.",
+    "ja": "Earendil の Pi コーディングエージェント。",
+    "zh-CN": "Earendil 的 Pi 编码智能体。",
+    "zh-TW": "Earendil 的 Pi 編碼代理。",
+    "ko-KR": "Earendil의 Pi 코딩 에이전트입니다.",
+    "no": "Earendils Pi-kodingsagent.",
+    "ar": "وكيل Pi للبرмجة من Earendil.",
```
VascoSch92: Looks like there's a stray Cyrillic 'м' (U+043C) in here instead of the Arabic م (U+0645) — the word for "programming" has a foreign character spliced into the middle of it. Probably a copy/paste artifact, but worth fixing since it'll render as a broken word for Arabic users.

docker/Dockerfile
```diff
@@ -99,6 +105,19 @@ RUN if command -v apt-get >/dev/null 2>&1; then \
...
+# agent-server base; Pi is third-party and installed via the Node 24 stage
+# above (pi-acp requires Node 22+).
+#
+# Copy the whole Node prefix so npm global bin *symlinks* stay intact.
+# Never ``COPY`` /usr/local/bin/pi or pi-acp alone — Docker dereferences
+# symlinks into flat scripts that import ./config.js from bin/ instead of
+# the package directory under lib/node_modules.
+COPY --from=pi-acp-install /usr/local/ /opt/node24/
+ENV PATH="/opt/node24/bin:${PATH}"
+RUN test -x /opt/node24/bin/pi-acp && test -x /opt/node24/bin/pi && \
```
VascoSch92, line 118: This only actually runs `pi --version`; for `pi-acp` we just do `test -x`, which confirms the file's executable but never invokes it. Given the whole point of the `/opt/node24` copy is to keep both binaries' symlinks intact, might be worth actually running `pi-acp --version` too — otherwise a broken pi-acp entrypoint wouldn't show up until someone tries to use it in production.

docker/Dockerfile
```diff
@@ -67,6 +67,12 @@ RUN node -e " \
   require('fs').writeFileSync('/tmp/defaults.env', lines.join('\n') + '\n'); \
 "
 
+# ── Stage 1c: Pi ACP (Node 22+ required; avoid Debian apt nodejs) ─────────────
+FROM node:24-slim AS pi-acp-install
+RUN npm install -g --ignore-scripts \
```
VascoSch92, line 72: No version pin on either package here, unlike the rest of the file (automation is pinned via `AUTOMATION_VERSION`, the SDK via git ref, frontend deps via lockfile). Since this resolves 'latest' at build time, two builds a few minutes apart could ship different Pi versions with no way to pin or roll back. Might be worth locking this down.

docker/Dockerfile
```diff
@@ -99,6 +105,19 @@ RUN if command -v apt-get >/dev/null 2>&1; then \
...
+# Pi coding agent + ACP adapter. Claude/Codex/Gemini wrappers ship in the
+# agent-server base; Pi is third-party and installed via the Node 24 stage
+# above (pi-acp requires Node 22+).
+#
+# Copy the whole Node prefix so npm global bin *symlinks* stay intact.
+# Never ``COPY`` /usr/local/bin/pi or pi-acp alone — Docker dereferences
+# symlinks into flat scripts that import ./config.js from bin/ instead of
+# the package directory under lib/node_modules.
+COPY --from=pi-acp-install /usr/local/ /opt/node24/
+ENV PATH="/opt/node24/bin:${PATH}"
```
VascoSch92: This PATH change isn't scoped to pi/pi-acp — it affects which `node`/`npm`/`npx` resolves for everything else running in this image, including the Claude/Codex/Gemini wrappers and the frontend's static server. If the base image pins a different Node major than those were built/tested against, this could quietly swap their runtime Node version. Might be safer to invoke pi/pi-acp with an explicit path instead of widening PATH globally.

docker/Dockerfile
```diff
@@ -99,6 +105,19 @@ RUN if command -v apt-get >/dev/null 2>&1; then \
...
 
+# Pi coding agent + ACP adapter. Claude/Codex/Gemini wrappers ship in the
+# agent-server base; Pi is third-party and installed via the Node 24 stage
+# above (pi-acp requires Node 22+).
+#
+# Copy the whole Node prefix so npm global bin *symlinks* stay intact.
+# Never ``COPY`` /usr/local/bin/pi or pi-acp alone — Docker dereferences
+# symlinks into flat scripts that import ./config.js from bin/ instead of
+# the package directory under lib/node_modules.
+COPY --from=pi-acp-install /usr/local/ /opt/node24/
```
VascoSch92, line 116: This pulls in the whole Node 24 install just to get two global npm packages — npm's own node_modules, corepack, etc. all come along for the ride. Since we only need the `pi`/`pi-acp` binaries, could we just copy the specific bin symlinks plus their target package dirs instead of the entire prefix? Would help with image size.

src/api/agent-server-adapter.ts
```diff
@@ -769,14 +801,22 @@ function buildConfiguredAcpAgentSettings(
   // that, the form's displayed default would silently not take effect at
   // runtime until the user re-saved the page.
   const serverKey =
```
VascoSch92: The `?? (typeof agentSettings.acp_server === 'string' ? agentSettings.acp_server : undefined)` here duplicates what `resolveUiAcpProviderKey` already falls back to internally on the same guarded value, so this branch can never actually produce a different result than the first part of the expression — it's dead. Probably worth dropping so it doesn't look like it's doing something it isn't.

src/constants/acp-providers.ts
```diff
@@ -5,10 +5,109 @@ export type ACPProviderIcon =
...
+export function wireAcpServerForProvider(providerKey: string): string {
+  return providerKey === PI_ACP_PROVIDER_KEY ? "custom" : providerKey;
+}
+
+/**
+ * Resolve the Canvas registry key from persisted agent settings. Pi is stored
+ * as ``acp_server: "custom"`` with the ``pi-acp`` command — this helper
+ * rehydrates ``"pi"`` for UI, model defaults, and conversation tags.
+ */
+export function resolveUiAcpProviderKey(
```
VascoSch92, line 66: This only recognizes a `'custom'` + Pi setup by fingerprinting a non-empty `acp_command` — every other built-in provider gets picked up directly from its `acp_server` key regardless of whether a command is present. If we ever end up with `acp_server: 'custom'` and an empty/cleared command for what was meant to be Pi, this won't resolve to `'pi'` and won't fall back to the default argv the way claude-code/codex/gemini-cli do. Probably a narrow edge case, but flagging it.

src/constants/acp-providers.ts
```diff
@@ -5,10 +5,109 @@ export type ACPProviderIcon =
...
+  "pi-acp",
+];
+
+/** Docker image pre-installs ``pi-acp`` on PATH; ``npx`` is unreliable there. */
+export const PI_ACP_DOCKER_COMMAND: readonly string[] = ["pi-acp"];
+
+const PI_ACP_DEFAULT_COMMAND_TEXT = PI_ACP_DEFAULT_COMMAND.join(" ");
+const PI_ACP_DOCKER_COMMAND_TEXT = PI_ACP_DOCKER_COMMAND.join(" ");
+
+function normalizeAcpCommandTokens(command: unknown): string[] {
```
VascoSch92, line 31: This does a plain `trim().split(/\s+/)`, but there's already a quote-aware `parseCommand` in `#/utils/acp-command` used for the same string→argv conversion elsewhere in this PR (e.g. `inline-pi-acp-profile-settings.ts`). A command with quoted tokens wouldn't split cleanly here, and `isPiAcpCommand` could fail to recognize an otherwise valid Pi command as a result. Might be worth reusing `parseCommand` for consistency.

src/components/features/onboarding/steps/setup-acp-secrets-step.tsx
```diff
@@ -125,8 +130,15 @@ export function SetupAcpSecretsStep({
       if (providerKey !== "openhands") {
         await applyAgentProfile({
           agent_kind: "acp",
-          acp_server: providerKey,
+          acp_server: wireAcpServerForProvider(providerKey) as ACPServerKind,
           acp_model: getAcpPreferredDefaultModel(providerKey) ?? undefined,
+          ...(providerKey === "pi"
```
VascoSch92: Nit: this compares against the bare string `'pi'`, but the file already imports `wireAcpServerForProvider`/`effectivePiAcpCommandTokens` from the same module, which also exports `PI_ACP_PROVIDER_KEY`. Using the constant here would keep this in sync if the key value ever changes — right now TS won't catch drift since `providerKey` is typed via a separate union.

VascoSch92:
@Deep070203 you can tag me when the PR is ready for another pass ;-)

Deep070203:
@VascoSch92 Can you take a look at it and let me know!

VascoSch92:
Hey @Deep070203 

I had time to dive a little bit more inside the PR and also the ACP-machinery, and actually this PR is a frontend-only workaround.

The linked issue #16204 asks for parity, not a workaround. In fact `"This should be supported like Claude, Gemini, and ChatGPT/Codex."` Those three are first-class SDK registry entries; Pi, as shipped in this PR, is not.

Moreover, note that the frontend's own docs contradict what this PR does. `docs/ACP_AGENTS.md:` "Adding or changing a provider happens upstream in the SDK, not here."

We should not proceed with this path for the following reasons: 
- Only the tested path works. The PR's own evidence is 100% local Docker, cloud/SaaS sandboxes never hit the Docker-mode rewrite (getDeploymentMode() only returns "docker" for the self-hosted image), so those conversations fall back to unpinned `npx -y pi-acp` on every launch, completely unverified by this PR.
- Real data-race risk in cloud. Pi has no `data_dir_env_var` (unlike Claude/Codex/Gemini, each pinned to isolate `~/.claude*/CODEX_HOME/HOME`), so concurrent Pi conversations sharing a grouped sandbox pod can collide on the same ~/.pi state — the exact bug class the SDK added those isolation vars to prevent.
- Unpinned version = silent breakage risk. The other three CLIs are version-pinned in the SDK registry; Pi is not, so any upstream pi-acp release can change behavior mid-flight with no way to pin/rollback from OpenHands's side.
- Breaks product visibility. Every Pi turn is traced as `acp_server: "custom"`, indistinguishable from any other unregistered custom command
- Guaranteed rework. Once Pi is properly added to the SDK, every piece of this PR's special-casing (`wireAcpServerForProvider`, `resolveUiAcpProviderKey`, `adaptPiAcpCommandForDeployment`, the Node-24 Dockerfile stage) has to be found and unwound. So we will have to re-code the feature another time.


The things to do to add pi in a correct way (as asked in the Issue) are:

1. software-agent-sdk: add "pi" to `ACPServerKind`, add a full `ACPProviderInfo` entry to `ACP_PROVIDERS`, add pi-acp to the agent-server base Dockerfile.
2. typescript-client: update `acp-providers.json `/ `ACPProviderKey` to match (satisfies the existing drift-check CI job).
3. OpenHands/OpenHands: once Pi is a real registry entry, remove the "custom"-wire spoofing and Docker-mode command-rewrite hacks from this PR. It becomes a native provider, and cloud/sandbox deployments get it automatically.

VascoSch92:
@Deep070203  Do you want to open the PRs into the sdk or should I do that?

Deep070203:
@VascoSch92 I will do that. Want to learn the whole flow!

PR #16233, opened 2026-08-01
fix: derive automation timeout cap from deployment

neubig:
HUMAN:

Resolved conflicts with main and re-ran the focused automation timeout tests and TypeScript validation.

AGENT:

---

## Why

Agent Canvas previously hard-coded the automation run-timeout maximum at 1800 seconds, which can disagree with the actual deployment capability. This PR removes the hard-coded ceiling, derives the timeout cap from the deployment's reported capability (`maxAutomationTimeoutSeconds`), clamps any interface-manifest limit beneath it, and falls back to server-side validation when the capability is absent.

## Summary

- remove Agent Canvas's hard-coded automation timeout maximum
- use the deployment capability as the absolute ceiling and clamp an interface-manifest limit beneath it
- preserve compatibility with older services by relying on their server-side validation when the capability is absent
- interpolate the deployment-derived cap into the `AUTOMATIONS$ERROR_TIMEOUT_MAX_EXCEEDED` error message (was hardcoded to 1800 seconds)

## Dependency

Requires OpenHands/automation#296 to publish `maxAutomationTimeoutSeconds` from the service configuration.

## How to Test

1. `npm run make-i18n`
2. `npx vitest run __tests__/utils/automation-timeout.test.ts __tests__/components/automations/detail/edit-automation-modal.test.tsx __tests__/manifests/automation-interface.test.ts`
3. `npm run typecheck`
4. In the UI, open an automation's edit modal on a deployment whose capability reports a `maxAutomationTimeoutSeconds` (e.g. 900). Enter a timeout above the cap (e.g. 1000) and confirm the inline error names the deployment cap (e.g. "Timeout cannot exceed 900 seconds") rather than a fixed 1800.

---

| Component | Value |
|---|---|
| **Commit** | `fe0d303d80d5b4c6536cbbdcd6568a80d19b6097` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-fe0d303
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-fe0d303
```

```
ghcr.io/openhands/agent-canvas:sha-fe0d303-amd64
ghcr.io/openhands/agent-canvas:fix-runtime-automation-timeout-cap-amd64
ghcr.io/openhands/agent-canvas:pr-16233-amd64
ghcr.io/openhands/agent-canvas:sha-fe0d303-arm64
ghcr.io/openhands/agent-canvas:fix-runtime-automation-timeout-cap-arm64
ghcr.io/openhands/agent-canvas:pr-16233-arm64
ghcr.io/openhands/agent-canvas:sha-fe0d303
ghcr.io/openhands/agent-canvas:fix-runtime-automation-timeout-cap
ghcr.io/openhands/agent-canvas:pr-16233
```

- Each tag (e.g., `sha-fe0d303`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-fe0d303-amd64`) are also available if needed


src/components/features/automations/detail/edit-automation-modal.tsx
```diff
@@ -199,12 +207,9 @@ export function EditAutomationModal({
...
     setNameError(null);
 
-    const timeoutResult = validateAutomationTimeout(
-      form.timeout,
-      timeoutSpec.max ?? undefined,
-    );
+    const timeoutResult = validateAutomationTimeout(form.timeout, timeoutMax);
     if ("errorKey" in timeoutResult) {
-      setTimeoutError(t(timeoutResult.errorKey));
+      setTimeoutError(t(timeoutResult.errorKey, { max: timeoutMax }));
```
hieptl, line 212: `setTimeoutError(t(timeoutResult.errorKey, { max: timeoutMax }));` passes the `max` value for interpolation, but `AUTOMATIONS$ERROR_TIMEOUT_MAX_EXCEEDED` in `src/i18n/translation.json:13841` is still hardcoded as `"Timeout cannot exceed 1800 seconds (30 minutes)"` across all languages, without a `{{max}}` placeholder.

As a result, on a deployment with a 900-second limit, a user who enters `1000` is correctly rejected, but the error message incorrectly states that the limit is 1800 seconds. Since the repository already uses `{{max}}` interpolation in other translations (for example, `"Round {{iteration}}/{{max}}"`), it would be more consistent to update this string across all locales to something like `"Timeout cannot exceed {{max}} seconds"`.

The current component tests do not detect this issue because they assert against the translation key rather than the rendered message.

Thank you very much! 🙏

hieptl (approved):

src/components/features/automations/detail/edit-automation-modal.tsx
neubig: Fixed in f62831440: the AUTOMATIONS$ERROR_TIMEOUT_MAX_EXCEEDED strings across all 15 languages now use a {{max}} interpolation placeholder instead of the hardcoded "1800 seconds (30 minutes)" text. The modal already passes { max: timeoutMax } to t(), so on a deployment with a 900-second cap, entering 1000 now shows "Timeout cannot exceed 900 seconds". Verified via i18next with the real translation resources (max=900 and max=1800 both interpolate correctly) and the 3 affected test suites pass.

PR #16239, opened 2026-08-01
fix: clean up dev services on SIGHUP

### LINKED ISSUE #16238: [Bug]: Dev launcher leaves detached services running after SIGHUP
## Operating System

Linux — Fedora 43

## Installation Method

From source (`npm run dev`)

## Agent Canvas Version

Current `main` at `1708efc44` (Node.js v22.22.2)

## Bug Description

After the `npm run dev` launcher exits on `SIGHUP`, the services it started remain alive. They are detached process-group leaders, so they are reparented to PID 1 and keep their ports bound.

I initially noticed this after returning to a local development stack whose launcher was gone while several child services were still listening. The event that terminated that launcher was not captured, so I cannot say whether it was caused by closing a terminal, losing a multiplexer session, or something else. Sending `SIGHUP` to an isolated launcher reproduces the same orphaned-process topology deterministically.

## Steps to Reproduce

1. Start the development stack with `npm run dev`.
2. Identify the Node process running `scripts/dev-with-automation.mjs`.
3. Send it `SIGHUP` with `kill -HUP <pid>`.
4. Inspect the launched service processes or their listening ports.

The isolated reproduction uses the script's exported `spawnService` path. After `SIGHUP`, the supervisor exits with signal `SIGHUP`, while its detached TCP fixture remains reachable and has PPID 1.

## Actual Behavior

The launcher exits without running its coordinated shutdown path. Detached child services remain running and continue occupying their ports.

## Expected Behavior

`SIGHUP` should use the same coordinated shutdown path as `SIGINT` and `SIGTERM`, terminating each launched service process group before the launcher exits.

## Additional Context

`scripts/dev-with-automation.mjs` registers its existing `shutdown` handler for `SIGINT` and `SIGTERM`, but not `SIGHUP`. Because services are launched detached, the operating system does not terminate them with the launcher.

A draft fix and POSIX regression test are available in #16239.

georgeglarson:
HUMAN:

I returned to a local npm run dev stack and the launcher was gone. I think it was inside a multiplexer I had closed.
Repro by sending SIGHUP to the launcher, cleanup not hit.

AGENT:

Reproduced on Linux with a disposable TCP fixture launched through the real `spawnService` path. Before the fix, sending `SIGHUP` terminated the supervisor with signal `SIGHUP` while the detached fixture remained reachable. After the fix, the supervisor exited with code 0 and the test observed the fixture port become unavailable.

Validation run locally:

- `npm test -- --run __tests__/scripts/dev-with-automation.test.ts` — 47/47 tests passed
- `npm test` — passed
- `npm run lint` — passed
- `npm run typecheck` — passed
- `npm run build` — passed

---

## Why

The POSIX dev launcher creates detached process groups for its services, but did not run coordinated cleanup when it received `SIGHUP`. Losing the controlling terminal could therefore leave service processes reparented to PID 1 and their ports occupied.

Fixes #16238.

## Summary

- Route `SIGHUP` through the launcher's existing shutdown handler.
- Add a POSIX black-box regression test that verifies the detached service is terminated and releases its port.

## Issue Number

#16238

## How to Test

Run:

```bash
npm test -- --run __tests__/scripts/dev-with-automation.test.ts
```

The regression test starts a disposable detached TCP service through `spawnService`, sends `SIGHUP` to its supervisor, and verifies both a clean supervisor exit and release of the service's dynamically allocated port.

## Video/Screenshots

Not applicable; this changes process lifecycle behavior and is covered by the black-box regression test above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The regression test is skipped on Windows because `SIGHUP` and negative-PID process-group signaling are POSIX behavior.

### DIFF
--- a/__tests__/scripts/dev-with-automation.test.ts
+++ b/__tests__/scripts/dev-with-automation.test.ts
@@ -12,7 +12,7 @@ import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { homedir, tmpdir } from "node:os";
 import path from "node:path";
 import { setTimeout as delay } from "node:timers/promises";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import { describe, expect, it, afterEach } from "vitest";
 import {
   buildAgentServerAutomationEnv,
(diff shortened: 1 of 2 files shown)


enyst (approved):
🟢 **Good taste** — this is the right-sized fix for a concrete POSIX lifecycle bug. Routing `SIGHUP` through the existing shutdown path preserves the established behavior, and the black-box regression test proves the user-visible failure mode: the launcher exits cleanly and the detached service releases its port. The focused unit/build checks and full mock-LLM E2E suite are green.

**VERDICT:**
✅ **Worth merging**: focused, backward-compatible, and well tested.

**KEY INSIGHT:**
Reusing the existing coordinated shutdown path avoids introducing a second cleanup mechanism or signal-specific special case.

**[RISK ASSESSMENT]**
Two files, one production line, no dependency/API/event/agent-behavior changes, and direct POSIX regression coverage. Windows skips the POSIX-only assertion but its CI build remains green.

_This review was generated by an AI agent (OpenHands) on behalf of the user._

enyst:
Thank you!

PR #16241, opened 2026-08-02
fix(settings): surface providers discoverable only from model IDs

### LINKED ISSUE #15576: OpenRouter missing from list of Basic LLM Providers

lokesh75-kank:
HUMAN:

I hit #15576 myself on a local setup (OpenRouter models usable but the provider missing from the Basic LLM Providers picker), directed this fix, and reviewed the change and regression tests.

AGENT:

Verified by tests against MSW mocks mirroring the real local agent-server responses (no live agent-server e2e run — noted honestly below):

- `npx vitest run __tests__/api/config-service.test.ts` — 5/5 passing, including the new regression test where `openrouter` is absent from `/api/llm/providers` and `/api/llm/models/verified` but present via `openrouter/<model>` IDs.
- `npx vitest run __tests__/components/modals/settings/model-selector-openhands.test.tsx` — passing, confirming the ModelSelector still issues exactly one request each to `/api/llm/providers`, `/api/llm/models`, and `/api/llm/models/verified` (a first iteration of this fix duplicated the models fetch; CI caught it and it now goes through a shared react-query fetch).
- `npx vitest run __tests__/hooks/query` — 24 files / 135 tests passing.
- `npx tsc --noEmit`, `eslint`, and `prettier --check` clean on all touched files.

---

## Why

On local backends, the Basic LLM Providers list was built from `/api/llm/providers` and the verified-models map only. Providers such as `openrouter`, whose models appear in `/api/llm/models` as `<provider>/<model>` IDs but which are absent from both other sources, never showed up in the picker — even though `searchModels` could already list their models.

## Summary

- `ConfigService.searchProviders` now also derives provider names from `<provider>/<model>`-prefixed IDs in `/api/llm/models`, merged into the existing name set (dedup and verified-first ordering unchanged; bare IDs without `/` ignored).
- The raw model-ID list is fetched once and shared between `useSearchProviders` and `useProviderModels` via react-query (`use-llm-models.ts`), mirroring the existing verified-models pre-fetch pattern, so `/api/llm/models` is not requested twice.
- Regression test added mirroring the real local agent-server response shape.

## Issue Number

Fixes #15576

## How to Test

1. `npm install`
2. `npx vitest run __tests__/api/config-service.test.ts __tests__/components/modals/settings/model-selector-openhands.test.tsx` — the first file includes the regression test "includes providers only discoverable from model IDs (e.g. openrouter)".
3. Manual: run the frontend against a local agent server whose `/api/llm/models` includes `openrouter/...` IDs while `/api/llm/providers` omits `openrouter`; open Settings → LLM → provider picker and confirm OpenRouter appears.

## Video/Screenshots

Not included — the change is in the provider-list reconstruction logic; behavior is covered by the request-shape regression tests above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No config or migration impact. Cloud backends are untouched (they use `/api/v1/config/*/search` directly).

### DIFF
--- a/__tests__/api/config-service.test.ts
+++ b/__tests__/api/config-service.test.ts
@@ -8,7 +8,9 @@ describe("ConfigService", () => {
     const page = await ConfigService.searchProviders({ limit: 10 });
 
     expect(page.next_page_id).toBeNull();
-    expect(page.items.some((provider) => provider.name === "anthropic")).toBe(true);
+    expect(page.items.some((provider) => provider.name === "anthropic")).toBe(
+      true,
+    );
     expect(
       page.items.find((provider) => provider.name === "anthropic")?.verified,
     ).toBe(true);
@@ -21,10 +23,12 @@ describe("ConfigService", () => {
     });
 
     expect(page.next_page_id).toBeNull();
-    expect(page.items.some((model) => model.name === "claude-opus-4-5-20251101")).toBe(
+    expect(
+      page.items.some((model) => model.name === "claude-opus-4-5-20251101"),
+    ).toBe(true);
+    expect(page.items.every((model) => model.provider === "anthropic")).toBe(
       true,
     );
-    expect(page.items.every((model) => model.provider === "anthropic")).toBe(true);
   });
 
   it("includes verified providers absent from /api/llm/providers and keeps them within the limit", async () => {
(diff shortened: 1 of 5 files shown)


VascoSch92:
Hey @lokesh75-kank 

thank you for the PR. But I think there is already a PR open from another contributor on this issue, see https://github.com/OpenHands/OpenHands/pull/16096.

lokesh75-kank:
Thanks for the pointer @VascoSch92 — apologies, I missed #16096 before opening this. Closing it.

Issue #15671, opened 2026-06-05
[Bug]: pickFallbackBackend ignores backend health & kind — Settings → LLMs shows "No backend is configured" even when a healthy local backend is registered


VascoSch92:
### Summary

When there's no valid backend selection (fresh start, or the selected backend was removed), the active backend is chosen by `pickFallbackBackend`, which returns `backends[0]` — the first registered backend in insertion order, ignoring both its `kind` and its recorded health. If index-0 is a cloud backend or a dead local backend, `getEffectiveLocalBackend()` resolves to `null`/dead, and every local-protocol call (verified models, LLM profiles, conversation CRUD) throws `No backend is configured.` — even when a healthy local backend is registered further down the list. A synchronous health store already exists at the point of selection but isn't consulted.

Split out of OpenHands/agent-canvas#2091 (its Docker/local part was handled in OpenHands/agent-canvas#2095).

### Environment

macOS · Docker (`ghcr.io/openhands/agent-canvas`) · 1.0.0-beta.7 · verified against `main` @ `df8a573`

### Steps to Reproduce

1. Register two backends so that index-0 is either a cloud backend or a now-stopped local backend, and a healthy local backend is *not* index-0.
2. Have no valid selection (fresh load, or delete the currently-selected backend).
3. Open **Settings → LLMs**.

### Actual Behavior

`No backend is configured.` toast; empty LLM Provider dropdown; no LLM profiles. The healthy local backend is never tried.

### Expected Behavior

Fallback selection resolves to a healthy local backend when one is registered; verified models and profiles load.

### Root Cause

`pickFallbackBackend` (`src/api/backend-registry/active-store.ts:36-38`):

```ts
function pickFallbackBackend(backends: Backend[]): Backend {
  return backends[0] ?? NO_BACKEND;
}
```

Used by `computeSnapshot` when no selection resolves (`active-store.ts:58-62`). The chosen backend flows to `getEffectiveLocalBackend()` (`:94-98`), which returns `null` if it isn't a live local backend; local-protocol calls then throw `NoBackendAvailableError("No backend is configured.")` (`agent-server-client-options.ts:22-27`, `:55-57`). Selection consults neither `kind` nor the health store.

### Proposed Fix

[...]

PR #16243, opened 2026-08-02
fix(backends): pin backend identity on sidebar conversation links

### LINKED ISSUE #15573: Cmd-clicking a sidebar conversation can open on the wrong backend
## Problem

When a user cmd-clicks / ctrl-clicks a conversation in the conversations sidebar to open it in a new browser tab, the new tab is not guaranteed to stay on the backend that owns that conversation.

If the new tab initializes with a different selected backend, it attempts to load the conversation ID against that backend and shows `conversation not found`.

## Expected behavior

Opening a conversation from the sidebar in a new tab should preserve the selected backend identity in the URL or navigation target, so the new tab resolves the conversation against the same backend as the originating tab.

## Notes

This affects multi-backend Agent Canvas setups where conversations are scoped to specific backends. The fix should cover normal click and modified-click/new-tab browser behavior.

rishavnaskar:
HUMAN:

Hit this on a two-backend setup: cmd-clicking a sidebar conversation kept opening a tab that couldn't find it. Tracked it down to the tab-scoped backend selection and added an E2E case that reproduces it — details below.

Happy to split the `NavigationLink` change out into its own PR if you'd rather keep this focused.

AGENT:

---

## Why

Cmd/ctrl-clicking a conversation in the sidebar can open the new tab on the wrong backend, which then fails to resolve the conversation id and shows `conversation not found`.

The reason is that the active backend is **tab-scoped**. `readStoredActiveBackend()` (`src/api/backend-registry/storage.ts`) prefers `sessionStorage` and only falls back to `localStorage`:

```ts
// Active backend is tab-scoped so reloading tab A does not adopt tab B's
// backend. localStorage remains a last-used fallback for fresh tabs and old
// persisted state.
const sessionSelection = parseBackendSelection(
  readStorageItem(window.sessionStorage, ACTIVE_BACKEND_STORAGE_KEY),
);
if (sessionSelection) return sessionSelection;

return parseBackendSelection(
  readStorageItem(window.localStorage, ACTIVE_BACKEND_STORAGE_KEY),
);
```

A tab opened via cmd/ctrl-click or middle-click does not inherit the opener's `sessionStorage`, so it boots from the `localStorage` fallback — and that holds whichever backend was selected *last in any tab*, not the one the sidebar we just clicked belongs to. The link itself carries no backend identity, so there is nothing to correct the choice.

## Summary

- Add `withBackendSelectionParams()` / `readBackendSelectionFromUrl()` (`src/api/backend-registry/url-selection.ts`), which pin `?backend=<id>[&org=<id>]` onto a path and read it back. An id that isn't registered is ignored, so a stale link or one from another machine degrades to today's behaviour instead of selecting nothing.
- Resolve the URL selection **when the active store boots** (module init, before React renders) and persist the honoured selection, so no request is ever issued against the wrong backend and later in-tab navigation — which drops the query string — keeps the same backend.
- Build sidebar conversation rows, collapsed-rail rows and start-task cards through a new `useBackendScopedPath()` hook.
- Make `NavigationLink` compare the *pathname* portion of `to` for its active state. Without this the added query string silently breaks the highlight on the current conversation row.

## Issue Number

Fixes #15573

### End-to-end

Added `cmd-clicking a sidebar conversation opens it on the owning backend` to `tests/e2e/mock-llm/backends/mock-llm-cross-connect.spec.ts`. It spins up two isolated backend-only instances plus a frontend-only instance, registers both backends through the UI, and reproduces the storage asymmetry the bug needs — tab 1's `sessionStorage` naming backend A while `localStorage` names B — before cmd-clicking a sidebar row.

Conversations are seeded through `POST /api/conversations`, which never contacts the LLM at creation time, so the case needs **no model credentials**.

```
MOCK_LLM_PYTHON=.mock-llm-venv/bin/python3 npx playwright test \
  tests/e2e/mock-llm/backends/mock-llm-cross-connect.spec.ts
```

```
3 passed (1.2m)
```

**It fails on the parent commit.** I reverted just `src/` to `HEAD~1`, rebuilt, and re-ran — the new tab lands on backend B, 404s, and gets redirected home, which is exactly the reported symptom:

```
Error: new tab should open the conversation on its owning backend, not bounce home

Expected: "/conversations/09d3c948-4030-4275-a00a-33a036128af7"
Received: "/conversations"

  1 failed
```

Restoring `src/` and rebuilding: `1 passed (40.4s)`.

Note the E2E config only builds when `build/index.html` is absent, so `npm run build:app` is needed between toggling the fix — a stale `build/` silently tests the wrong code (that cost me a couple of runs).

### Unit

```
npx vitest run __tests__/api/backend-registry/ __tests__/hooks/use-backend-scoped-path.test.tsx __tests__/components/shared/navigation-link.test.tsx __tests__/components/features/conversation-panel/conversation-panel.test.tsx
```

```
Test Files  8 passed (8)
     Tests  113 passed (113)
```

Full suite, `npx vitest run`:

```
Test Files  7 failed | 529 passed | 1 skipped (537)
     Tests  25 failed | 4172 passed | 5 skipped | 9 todo (4211)
```

Those 25 failures are pre-existing on `main`. I captured the failing-file list on a clean checkout and after the change and diffed them — byte-identical, no new failures:

```
__tests__/components/features/conversation-panel/local-new-conversation-menu.test.tsx (7 tests | 2 failed)
__tests__/components/features/conversation-panel/new-conversation-button-cloud.test.tsx (5 tests | 2 failed)
__tests__/components/features/home/home-chat-launcher.test.tsx (11 tests | 9 failed)
__tests__/components/features/home/new-conversation.test.tsx (2 tests | 1 failed)
__tests__/components/features/home/task-card.test.tsx (4 tests | 3 failed)
__tests__/components/home/llm-not-configured-banner.test.tsx (7 tests | 4 failed)
__tests__/components/onboarding/setup-acp-secrets-step.test.tsx (24 tests | 4 failed)
```

Also green: `npm run typecheck`, `npx eslint src tests/...`, `prettier --check`, and `npm run build:app`.

### Manually, if you'd rather drive it yourself

Two backends must have **separate state**, otherwise both resolve the same conversation ids and nothing reproduces. `npm run dev:extra-backend` shares the bundled instance's conversations by design, so override the state dir:

```
npm run dev
OH_CANVAS_SAFE_STATE_DIR=~/.openhands/agent-canvas-b npm run dev:extra-backend
```

Add `http://127.0.0.1:18002` via Add Backend, select it in a second tab, then cmd-click a conversation back in the first tab.

## Video/Screenshots

The E2E run records video (`video: "on"` in `playwright.mock-llm.config.ts`), so `test-results-mock-llm/` has the before/after captures if that's more useful than the log output above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Honouring a pinned link calls `writeStoredActiveBackend()`, which mirrors to `localStorage` as well as `sessionStorage`. That means opening a pinned link also changes the default backend for future *fresh* tabs. I think that's right since it matches what an explicit backend switch already does, but it's easy to make session-only if you'd prefer the narrower behaviour.
- The parameters are deliberately left in the URL rather than stripped after boot, so a reload or a copy-pasted address bar still resolves against the right backend.
- Tacked on a `.gitignore` change: `.tmp/` was covered by two narrow entries (`mock-llm-state`, `mock-llm-skill-repos`), but a local E2E run also drops `.tmp/automation/automations.db` and `.tmp/secrets.json` there. A file named `secrets.json` being committable seemed worth closing, so the whole directory is ignored now. Happy to pull it into its own PR if you'd rather.
- Also unrelated: #15671 looks like it was already fixed on `main` — `pickFallbackBackend` now ranks by health and prefers a local backend — but the issue is still open. Might be worth closing.


hieptl (approved):
Thank you! 🙏

PR #16244, opened 2026-08-03
docs: note OpenAI client base_url for multi-model gateways

seven7763:
## Summary

The DefenseClaw docs already show OpenAI-compatible `base_url` routing.

This PR adds a one-line note that the same pattern works with multi-model gateways when you are not self-hosting, using [DaoXE](https://daoxe.com) (`https://api.daoxe.com/v1`) as one concrete example.

Docs only — no runtime behavior changes.

## Test plan
- [ ] DefenseClaw docs render
- [ ] Existing provider examples unchanged
- [ ] Note is clearly optional

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


neubig:
Thank you for sharing, but we typically do not name specific model providers unless they are specifically used widely by our users.

PR #16245, opened 2026-08-03
docs(acp): note OpenAI-compatible multi-model gateway base URL

seven7763:
## Summary

ACP agent docs already explain that custom `*_BASE_URL` overrides can point CLIs at a proxy or gateway.

This PR adds one concrete OpenAI-compatible multi-model gateway example ([DaoXE](https://daoxe.com) at `https://api.daoxe.com/v1`).

Docs only — no runtime behavior changes.

## Test plan
- [ ] ACP_AGENTS.md renders
- [ ] Existing base URL guidance unchanged
- [ ] Tip is clearly optional

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


neubig:
Thank you for sharing, but we typically do not name specific model providers unless they are specifically used widely by our users.

PR #16262, opened 2026-08-03
fix: avoid import-time signal handlers

### LINKED ISSUE #16257: dev launch scripts install signal handlers as import-time side effects
## Summary

`scripts/dev-with-automation.mjs` registers `SIGINT`/`SIGTERM` handlers at **module load time**, outside any runtime guard, and PR #16239 will add `SIGHUP` in the same location. Because the module is also *imported* (not just run as a CLI) by `scripts/dev-static.mjs`, `scripts/static-build.mjs`, `bin/agent-canvas.mjs`, and `electron/main.mjs`, merely importing it mutates process-wide signal behavior even when `main()` never runs.

- Signal registration: `scripts/dev-with-automation.mjs:959-960` (`process.on("SIGINT"...)`, `process.on("SIGTERM"...)`)
- Main guard: `scripts/dev-with-automation.mjs:1523-1535` (guards `main()`, but **not** signal registration)
- Importers that do not call `main()`: `scripts/dev-static.mjs:60-65`, `scripts/static-build.mjs:11-12`

## Why it's a problem

1. **Duplicate-handler race in `dev-static`.** `dev-static.mjs` keeps its **own** process registry and `shutdown()` (lines ~215-494) and registers its own `SIGINT`/`SIGTERM` handlers (lines ~496-497). When it imports `dev-with-automation.mjs`, both modules register handlers. On shutdown both schedule a ~3s exit timer, and the imported module's shutdown runs against an **empty** process map, can call `process.exit(0)` first, and starve `dev-static`'s own force-kill of its detached services.
2. **`SIGHUP` isn't covered for importers.** With PR #16239, `dev-with-automation` gets a `SIGHUP` handler, but `dev-static` still has none. On `SIGHUP`, only the imported module's empty-map shutdown runs, leaving `dev-static`'s detached services running — the exact bug #16238 was fixing, still present for `npm run dev:static`.
3. **Implicit ownership.** Signal handlers are process-global, so a module that is only *imported for helpers* should not silently take control of the process's signals. The current `isMainModule` guard protects `main()` but not these handlers.

## Proposed fix

Move signal registration out of the module top-level into an explicit function, and call it only after the process actually owns the lifecycle:

```js
function installSignalHandlers(handler) {
  process.on("SIGINT", handler);
  process.on("SIGTERM", handler);
  process.on("SIGHUP", handler);
}

function main() {
  // ... existing setup ...
  installSignalHandlers(shutdown);
  // ...
}
```

So the module top-level no longer calls `process.on`, and importers that own their own lifecycle (like `dev-static`) register their own handlers against **their** `shutdown` and **their** process registry. `bin/agent-canvas.mjs` / `electron/main.mjs` already call `main()`, so they'd get handlers via `installSignalHandlers` inside `main()` with no change needed.

Optionally, for shared-ish safety, register via `installSignalHandlers` only when the module is the main entry point:

```js
if (isMainModule) {
  installSignalHandlers(shutdown);
  main().catch(...);
}
```

Either way, `dev-static` should explicitly call `installSignalHandlers(shutdown)` (exported) with its own `shutdown`, which also resolves the SIGHUP gap for `npm run dev:static`.

## Scope

This is a dev-tooling lifecycle refactor, not a user-facing feature. It is a natural follow-up to #16239 — I flagged it there but did not block that focused fix on it. No child-service or agent/eval behavior changes are involved, and the existing black-box shutdown tests should be extended to cover `dev-static` and `SIGHUP`.

## References

- #16238 (original bug report)
- #16239 (SIGHUP fix that this builds on)

_This issue was created by an AI agent (OpenHands, using the `deep-flash` profile / DeepSeek V4 Flash) on behalf of Engel Nyst (enyst)._

Sakshamm-Goyal:
HUMAN:

AGENT:
This pull request was prepared with OpenAI Codex. The HUMAN section is intentionally left for a human contributor, per repository guidance.

## Why
Importing scripts/dev-with-automation.mjs currently installs SIGINT, SIGTERM, and SIGHUP handlers before any launcher owns the process lifecycle. Importers such as dev-static therefore inherit a second shutdown handler and can race their own detached-service cleanup.

## Summary
- Move signal registration behind an explicit installSignalHandlers lifecycle hook.
- Install the default handler from main() and let dev-static register its own shutdown handler, including SIGHUP.
- Add a regression proving module import leaves process-global signal listener counts unchanged while preserving SIGHUP cleanup coverage.

## Issue Number
Closes #16257

## How to Test
- npx vitest run __tests__/scripts/dev-with-automation.test.ts (48 passed)
- npm run lint (passed: typecheck, ESLint, Prettier)
- npm run build (passed: production client and server build; existing Vite chunk/future-flag warnings only)
- npx prettier --check scripts/dev-with-automation.mjs scripts/dev-static.mjs __tests__/scripts/dev-with-automation.test.ts (passed)
- git diff --check (passed)
- Direct reproduction after the fix: importing dev-with-automation leaves SIGINT/SIGTERM/SIGHUP listener counts at 0/0/0.

## Video/Screenshots
Not applicable: this is a Node launcher lifecycle fix. The regression runs the actual launcher child-process path and verifies detached-service shutdown behavior.

## Type
- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes
No API or user-facing configuration changes. dev-static now explicitly owns all three termination signals rather than inheriting handlers from an imported module.

### DIFF
--- a/__tests__/scripts/dev-static.test.ts
+++ b/__tests__/scripts/dev-static.test.ts
@@ -1,8 +1,46 @@
+import { spawn } from "node:child_process";
+import { once } from "node:events";
+import path from "node:path";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import { describe, expect, it } from "vitest";
 
 import { buildAutomationBackendEnv } from "../../scripts/dev-static.mjs";
 
+const repoRoot = path.resolve(
+  path.dirname(fileURLToPath(import.meta.url)),
+  "../..",
+);
+
 describe("dev-static", () => {
+  it("does not install signal handlers when imported", async () => {
+    const moduleUrl = pathToFileURL(
+      path.join(repoRoot, "scripts", "dev-static.mjs"),
+    ).href;
+    const child = spawn(
+      process.execPath,
+      [
+        "--input-type=module",
+        "--eval",
+        `const signals = ["SIGINT", "SIGTERM", "SIGHUP"]; const before = Object.fromEntries(signals.map((signal) => [signal, process.listenerCount(signal)])); await import(${JSON.stringify(moduleUrl)}); const after = Object.fromEntries(signals.map((signal) => [signal, process.listenerCount(signal)])); console.log(JSON.stringify({ before, after }));`,
+      ],
+      { cwd: repoRoot, stdio: ["ignore", "pipe", "pipe"] },
+    );
+    let output = "";
+    child.stdout.on("data", (chunk) => {
+      output += chunk.toString();
+    });
+    child.stderr.on("data", (chunk) => {
+      output += chunk.toString();
+    });
+
+    const [code] = await once(child, "exit");
+    expect(code).toBe(0);
+    expect(JSON.parse(output.trim())).toEqual({
+      before: { SIGINT: 0, SIGTERM: 0, SIGHUP: 0 },
+      after: { SIGINT: 0, SIGTERM: 0, SIGHUP: 0 },
+    });
+  });
+
   it("uses the same session key for both agent-server and automation backend auth", () => {
     const env = buildAutomationBackendEnv(
       {
--- a/__tests__/scripts/dev-with-automation.test.ts
+++ b/__tests__/scripts/dev-with-automation.test.ts
@@ -38,6 +38,21 @@ const repoRoot = path.resolve(
   "../..",
 );
 
+async function getFreePort() {
+  const server = net.createServer();
+  server.listen(0, "127.0.0.1");
+  await once(server, "listening");
+  const address = server.address();
+  if (!address || typeof address === "string") {
+    server.close();
+    throw new Error("Expected a TCP address");
+  }
+  const { port } = address;
+  server.close();
+  await once(server, "close");
+  return port;
+}
+
 describe("buildAutomationCommand", () => {
   it("uses released PyPI version by default", () => {
     const cmd = buildAutomationCommand({});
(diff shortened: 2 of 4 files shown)


georgeglarson:
Importing `scripts/dev-static.mjs` on this branch gives SIGINT 1 / SIGTERM 1 / SIGHUP 1, down from 2 / 2 / 1 on `main`, and `static-build.mjs` (imports the helpers, never calls `main()`) comes out clean at 0/0/0. The duplicate-handler race is gone.

Two things look worth another pass, best I can tell.

**The SIGHUP regression from #16239 stops covering the shipped path.** That test's supervisor used to do a bare `import { spawnService }`, and module-scope registration handed it the same handlers the real launcher had, so it exercised the mechanism that ships. With registration moved into `main()`, the supervisor calls `installSignalHandlers()` itself. Deleting the `installSignalHandlers();` call from `main()` and running the full suite still gives 534 files / 4188 tests passing, 0 failed, so nothing anywhere catches its removal. Driving the supervisor through `main()` would keep the guard pointed at what users actually run.

**`dev-static` keeps its import-time registration.** `installSignalHandlers(shutdown)` sits at module top level (`dev-static.mjs:514`), outside the `isMainModule` guard the file already has at `:650`, and `__tests__/scripts/dev-static.test.ts` imports the module for `buildAutomationBackendEnv`. So a vitest worker picks up three handlers wired to dev-static's `shutdown()`. Same shape as the bug, one file over. Moving the call inside the existing guard covers it.

Sakshamm-Goyal:
Addressed both gaps in `269577a5b`:

- The SIGHUP supervisor now invokes the production `main()` path in `--frontend-only` static mode and spawns its detached fixture from `extraPrereqs`. A controlled mutation that removes `installSignalHandlers()` from `main()` now fails exactly as intended: the supervisor exits on `SIGHUP` instead of coordinated code 0.
- `dev-static.mjs` now installs its shutdown handlers only inside its existing `isMainModule` guard. A new child-process regression proves importing it keeps SIGINT/SIGTERM/SIGHUP counts at 0/0/0; before the source fix, that test reproduced 1/1/1.

Validation on latest `main`: focused script tests 50/50, full `npm test`, `npm run typecheck`, Prettier, and `git diff --check` all pass.

georgeglarson:
269577a looks right to me. Both gaps appear closed.

After `await import()` of either script the SIGINT/SIGTERM/SIGHUP listener counts stay at 0/0/0, so the import-time takeover is gone. Then commented out the `installSignalHandlers()` call in `main()` and re-ran the two suites: "cleans up detached services when the launcher receives SIGHUP" goes red (`expected { code: null, signal: 'SIGHUP' } to match { code: 0, signal: null }`). So the call site in `main()` is covered as well as the helper itself. Rewriting that fixture to drive `main()` is what did it.

One suggestion. That coverage currently rides on the end-to-end test, which is the heaviest one in the file and the first to suffer on a slow runner. A small direct test that imports the module, calls `main({})` and asserts the counts go from 0 to 1 would pin the same invariant cheaply.

Opened a parallel PR of my own too quickly after your update, before giving you room to land this. Closed it. This one's yours.

Sakshamm-Goyal:
Implemented the lightweight call-site coverage in ff38ad9c2.

The new child-process regression imports dev-with-automation.mjs, calls main({}) with mutually exclusive launch modes so execution stops before any services or ports are started, and verifies SIGINT/SIGTERM/SIGHUP listener counts move from 0/0/0 to 1/1/1. I also mutation-checked the invariant: temporarily removing the main() installSignalHandlers() call makes this focused test fail with all three post-call counts at zero.

Validation after restoring the production call:
- dev-with-automation + dev-static launcher suites: 51 passed
- TypeScript typecheck: passed
- Prettier and git diff whitespace checks: passed

This is test-only; launcher behavior is unchanged.

PR #16272, opened 2026-08-03
chore: ignore all of /.tmp/, not three of its subdirectories

georgeglarson:
HUMAN:
Real problem. I noticed by accidentally committed some noise.

AGENT:

---

## Why
`.tmp/` is the common root for local and E2E runtime state, but only three of its subdirectories are ignored (live-e2e-state, mock-llm-state, mock-llm-skill-repos). The test harnesses write to several more that are not covered: `mock-llm-user-skills`, `e2e-folder-workspace-test`, and `dev-safe`. The enumeration keeps falling behind the code; ignoring the root removes the drift.

## Summary
- Collapses the three enumerated `/.tmp/<subdir>/` ignore rules into a single `/.tmp/`.
- Nothing under `.tmp/` is tracked, so no coverage is lost.

## Issue Number
N/A

## How to Test
`git check-ignore .tmp/anything` matches after the change; `git ls-files '.tmp/*'` is empty (nothing tracked under `.tmp/`). One-line `.gitignore` widening with no runtime change.

## Video/Screenshots
N/A (one-line `.gitignore` change).

## Type
- [ ] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [x] Docs / chore

## Notes
Follow-up to a Slack conversation with a maintainer.


enyst (approved):
Thank you!

PR #16294, opened 2026-08-04
fix: install dev launcher signal handlers only when a process owns the lifecycle

### LINKED ISSUE #16257: dev launch scripts install signal handlers as import-time side effects

georgeglarson:
HUMAN:

AGENT:

---

## Why

`scripts/dev-with-automation.mjs` registered `SIGINT`/`SIGTERM`/`SIGHUP` at module scope, so importing it for its helpers took over the importer's signal behaviour. `dev-static.mjs`, `static-build.mjs`, `bin/agent-canvas.mjs` and `electron/main.mjs` all import from it.

`dev-static` keeps its own process registry and its own `shutdown()`, so it ended up with two shutdown paths racing on one signal, and the imported one walks an empty registry and can call `process.exit` first.

Measured by listener count on `main`:

| module imported | SIGINT | SIGTERM | SIGHUP |
|---|---|---|---|
| `dev-static.mjs` | 2 | 2 | 1 |

Two of each: its own pair, plus the three it inherits.

## Summary

- Move registration into an exported `installSignalHandlers()` and call it from `main()`, ahead of the prerequisite checks that can exit before any service starts.
- `dev-static` registers its own `shutdown` the same way, inside the `isMainModule` guard it already has rather than at module scope, so importing it for `buildAutomationBackendEnv` installs nothing.
- Three regression tests, each verified to fail against the change it guards.

After: `0 / 0 / 0` on import for all four launchers.

## Issue Number

Closes #16257

## How to Test

```
npx vitest run __tests__/scripts/dev-with-automation.test.ts __tests__/scripts/dev-static.test.ts
```

To confirm each test guards something real:

1. Delete the `installSignalHandlers();` call from `main()` in `dev-with-automation.mjs`. "installs signal handlers when main() runs" fails. A test that calls the helper directly instead of running `main()` cannot catch this.
2. Move `installSignalHandlers(shutdown)` in `dev-static.mjs` out of the `isMainModule` guard back to module scope. "does not install signal handlers when the module is only imported" fails.

Direct check without the tests:

```
node --input-type=module --eval '
const s = ["SIGINT","SIGTERM","SIGHUP"];
await import("./scripts/dev-static.mjs");
console.log(Object.fromEntries(s.map(x => [x, process.listenerCount(x)])));'
```

## Video/Screenshots

Not applicable: Node launcher lifecycle, no UI surface. The evidence is the listener counts above and the mutation results under How to Test.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Full suite 4248 passed / 0 failed. `npm run lint` clean.

The `main()` test reads its counts from a `process.on("exit")` hook using `fs.writeSync`, because a failed prerequisite check can exit synchronously before control returns to the test, and `console.log` to a pipe can be dropped on exit. `installSignalHandlers()` runs ahead of those checks, so the handlers exist on that path too.

Overlaps #16262, which takes the same `installSignalHandlers()` approach. Two differences: there the `dev-static` call stays at module scope (so importing it still registers three handlers), and the SIGHUP test calls the helper directly rather than running `main()`, so removing the call from `main()` leaves the suite green.

### DIFF
--- a/__tests__/scripts/dev-static.test.ts
+++ b/__tests__/scripts/dev-static.test.ts
@@ -1,3 +1,8 @@
+import { spawn } from "node:child_process";
+import { once } from "node:events";
+import path from "node:path";
+import process from "node:process";
+import { fileURLToPath, pathToFileURL } from "node:url";
 import { describe, expect, it } from "vitest";
 
 import { buildAutomationBackendEnv } from "../../scripts/dev-static.mjs";
@@ -22,6 +27,52 @@ describe("dev-static", () => {
       AUTOMATION_POSTHOG_API_KEY:
         "phc_kBtz5nKmxVRRQ7HtPwr2QX9eMC5j65zE86QKocVNwb4U",
       AUTOMATION_POSTHOG_HOST: "https://us.i.posthog.com",
+    });
+  });
+});
+
+describe("dev-static signal handler ownership", () => {
+  // This suite imports dev-static.mjs for buildAutomationBackendEnv above, which
+  // is precisely the case that must not take over the importer's signals: the
+  // handlers would be wired to dev-static's own process registry, empty in a
+  // vitest worker, and its shutdown() calls process.exit on a 3s timer.
+  it("does not install signal handlers when the module is only imported", async () => {
+    const repoRoot = path.resolve(
+      path.dirname(fileURLToPath(import.meta.url)),
+      "../..",
+    );
+    const moduleUrl = pathToFileURL(
+      path.join(repoRoot, "scripts", "dev-static.mjs"),
+    ).href;
+
+    const child = spawn(
+      process.execPath,
+      [
+        "--input-type=module",
+        "--eval",
+        [
+          'import { writeSync } from "node:fs";',
+          'const signals = ["SIGINT", "SIGTERM", "SIGHUP"];',
+          "const counts = () =>",
+          "  Object.fromEntries(signals.map((s) => [s, process.listenerCount(s)]));",
+          "const before = counts();",
+          `await import(${JSON.stringify(moduleUrl)});`,
+          'writeSync(1, "COUNTS " + JSON.stringify({ before, after: counts() }) + "\\n");',
+        ].join("\n"),
+      ],
+      { cwd: repoRoot, stdio: ["ignore", "pipe", "pipe"] },
+    );
+
+    let output = "";
+    child.stdout.on("data", (chunk: Buffer) => (output += chunk.toString()));
+    child.stderr.on("data", (chunk: Buffer) => (output += chunk.toString()));
+    await once(child, "exit");
+
+    const match = output.match(/COUNTS (\{.*\})/);
+    expect(match, output).not.toBeNull();
+    expect(JSON.parse(match![1])).toEqual({
+      before: { SIGINT: 0, SIGTERM: 0, SIGHUP: 0 },
+      after: { SIGINT: 0, SIGTERM: 0, SIGHUP: 0 },
     });
   });
 });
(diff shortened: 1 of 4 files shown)


georgeglarson:
Closing in favour of #16262, which covers the same fix at 269577a. Verification notes are on that PR.

PR #16305, opened 2026-08-04
feat(chat): agent notifications with live detection [Do not Merge - POC]

FraterCCCLXIII:
## Summary

Frontend agent-notification UI that recommends reusable **skills / automations** after work in a conversation.

**How detection works today (important for cost/safety):**
- **No extra LLM call.** Detection is client-side over events already in the browser event store.
- Two sources:
  1. **Heuristics** on tool use (file edits, test commands, CI/GitHub CLI patterns) when the agent returns to `AWAITING_USER_INPUT`, plus a manual **Detect** control in the bell.
  2. Optional fenced `agent-notification` blocks in assistant messages (contract for a future/optional detection skill) — still parsed locally, not a separate paid analysis pass.
- Recommendations are stored per conversation in localStorage; Create All / dismiss / remove, and prompts can be drained into the composer.

**FE-only gate (addresses automation/resolver concern):**
- Suggestions are disabled unless `isInteractiveFeConversation` is true.
- Skips Cloud `trigger` values other than `"gui"` (`resolver`, `suggested_task`, `microagent_management`).
- Skips conversations with an automation attribution `origin` tag (e.g. `origin=slack`).
- Local FE creates (`trigger: null`, no `origin`) and ACP-only `acpserver` tags still qualify.

**Still open for product iteration:**
- Heuristics are **tool-signal based**, not “what the user asked for.” A follow-up could weight or replace this with analysis of **user messages** (cheap) for intent-aligned suggestions.

Prompt-suggestion chips (home/empty-chat starters) are **out of scope** — separate branch.

## Demo

Automation suggestion after `npm test` (workflow recommendation above the composer / in the bell):

Automation suggestion — Test runner workflow

[Watch the demo video (mp4)](https://github.com/FraterCCCLXIII/openhands-main/raw/0cc1e08d91496b6376b812c14ad03736460be8ca/.pr/agent-notifications-demo/automation-suggestion-demo.mp4)

## Test plan
- [ ] Interactive FE conversation: after file edits or `npm test`-style commands, bell Detect (or idle ingest) shows recommendations.
- [ ] Conversation with `origin` tag or Cloud `trigger=resolver` does **not** show the bell or invent suggestions.
- [ ] `VITE_STAGE_AGENT_NOTIFICATIONS=true` on an interactive FE chat shows demo cards; clearing history does not reseed staged items.
- [ ] Create All / dismiss / remove work; prompts drain into the composer.
- [ ] Confirm no additional LLM/completions traffic is issued solely for detection (network tab).


FraterCCCLXIII:
### Automation suggestion demo

After running `npm test` in an interactive FE conversation, the client recommends a **Test runner workflow** (Automation · Workflow):

Automation suggestion — Test runner workflow

[Demo video (mp4)](https://github.com/FraterCCCLXIII/openhands-main/raw/0cc1e08d91496b6376b812c14ad03736460be8ca/.pr/agent-notifications-demo/automation-suggestion-demo.mp4)

VascoSch92:
A couple of points:

---

The agent-emitted suggestion path (agent-notifications-parser.ts) is currently dead code — it parses agent-notification JSON fences that only .agents/skills/reusable-extension-detector.md would emit, but that skill file doesn't exist anywhere in the repo (not in this PR, not on main), so every suggestion shown today actually comes from the heuristic file/bash-command detector in detect-agent-notifications-from-events.ts.

---
**`src/hooks/chat/use-drain-agent-notification-prompts.ts:38`**

`finally` clears all pending prompts even if only some sent successfully.

Failure scenario: A user selects two recommendations to launch as a new conversation. The loop sends prompt 1 fine; prompt 2's `sendMessage` throws (e.g. socket not yet OPEN). The exception propagates but `finally` still calls `clearAgentNotificationPendingPrompts`, wiping prompt 2 from sessionStorage with no error shown and no retry possible.

---

**`src/components/features/chat/agent-notifications-list.tsx:200`**

`useEffect(() => setSelectedIds(agentNotificationIds), [agentNotificationIds])` resets selection to "all" whenever the notification array gets a new identity (e.g. a new notification streams in), silently re-checking items the user had deselected.

Failure scenario: User unchecks one recommendation in the bell dropdown. Before clicking "Create all," the background detector appends a new notification, creating a new array identity for `agentNotificationIds` even though it still contains the deselected id. The effect fires and re-selects everything, so clicking "Create all" creates the item the user explicitly opted out of.

---

**`src/hooks/chat/use-agent-notifications.ts:30`**

`isCreating` is per-hook-instance `useState`, not shared. The banner and bell dropdown each get an independent guard over the same notification list, so clicking "Create all" in both before the first finishes double-sends the same prompt(s).

Failure scenario: `useAgentNotifications` is called separately in the chat-input banner and the header bell dropdown for the same conversation. `createAll` only marks notifications seen after all sends resolve, so a user can click "Create all" in the banner, then before it finishes, click "Create all" in the bell (whose own `isCreating` is still false) — sending the identical prompt(s) twice.

---

**`src/components/features/chat/detect-agent-notifications-from-events.ts:31`**

`humanizeBasename`'s extension-strip regex (`/\.[^.]+$/`) matches an entire dotfile basename (`.env`, `.gitignore`), producing an empty label and id `detected-skill-`. A second dotfile edit collides on that same empty id and is silently dropped by dedup.
Failure scenario: When the agent's only substantive edit is a dotfile like `.gitignore` or `.env`, the regex matches the whole basename, leaving an empty string after split/filter. The notification is labeled " helper" (no filename) with id `detected-skill-`. A second unrelated dotfile edited later produces the same empty id and is silently dropped by the store's id-based dedup.

---

**`src/components/features/chat/detect-agent-notifications-from-events.ts:182`**

Notification id is derived from basename only (no directory), so `src/components/utils.ts` and `src/lib/utils.ts` collide and the second detection is silently dropped.

Failure scenario: The agent edits `src/components/utils.ts`, producing id `detected-skill-utils`. Later it edits an unrelated `src/lib/utils.ts`, producing the identical id. `addNotifications` filters against `existingIds`, so the second, more relevant recommendation is silently discarded and the stale first one keeps showing.

---

**`src/components/conversation-events/chat/event-message-components/user-assistant-event-message.tsx:46`**

The `agent-notification` fence-strip regex requires a closing fence, so an in-progress/unterminated fence during streaming renders raw in the transcript — the exact leak the extractor's doc comment says it prevents.

Failure scenario: While an agent turn is still streaming with an opening ` ```agent-notification ` fence and partial JSON but no closing fence yet, `FENCE_REGEX` (which requires the trailing ` ``` ` to match) doesn't match, so `extractAgentNotifications` returns the text essentially unchanged. The raw fence header and in-flight JSON render directly in the transcript for that render pass.

---

**`src/components/features/sidebar/sidebar-onboarding-checklist.constants.ts:1`**

The entire sidebar onboarding checklist (routes, i18n/description/action key maps, `isCustomizeChecklistPath`, storage helpers, the agent-notifications modal, `useSidebarOnboardingAgentNotifications`) is never imported outside its own definition/tests. The new Settings toggle ("Show getting started checklist") only flips a flag nothing reads — currently a no-op.

Failure scenario: Grepping the full diff confirms every symbol named here appears only at its own definition or same-purpose sibling files, never consumed by rendered UI. A user who flips the toggle in Settings changes a localStorage flag that nothing in the shipped app reads to render a checklist.

---

**`src/hooks/sidebar/use-sidebar-onboarding-agent-notifications.ts:63`**

`createAll` writes the first selected prompt to pending-prompts (auto-sent by the drain hook) *and* separately sets it as the conversation's draft message without clearing it, so it's both auto-sent and left re-sendable in the input box. Latent until the feature above is wired up.

Failure scenario: User selects 2+ staged notifications and calls `createAll`. `onSuccess` writes all prompts to `writeAgentNotificationPendingPrompts` (auto-sent by the drain hook) and simultaneously calls `setConversationState(conversationId, { draftMessage: promptA })` — unlike the pre-existing single-automation launcher this was modeled on, which clears the draft as part of sending, this path never consumes it. A user who doesn't notice the draft was already sent can hit Enter and send it a second time.

---

**`src/i18n/translation.json:13671`**
English `AUTOMATIONS$EMPTY_HINT` reads "Pick one from the **agentNotifications** below..." (internal identifier leaked into copy) instead of "recommendations". Other locales are unaffected.

Failure scenario: An English-speaking user with no automations configured opens the Automations page and sees the broken sentence. Every non-English locale still reads correctly, and nothing else in the diff touches this key — it's unrelated to the automations empty-state feature this PR is supposed to touch.

---

**`.env.sample:13`**

Deleted the doc line for `VITE_ENABLE_BROWSER_TOOLS` (still read in `agent-server-adapter.ts` and used by Playwright e2e configs) and replaced it in place with a new demo-only flag, with no relocation.

Failure scenario: `VITE_ENABLE_BROWSER_TOOLS` is still read at runtime and still relied on by the project's own Playwright configs/e2e specs. A developer reading `.env.sample` to learn how to omit BrowserToolSet from new conversations will no longer find any mention of the flag there, even though the code and test harness still depend on it.

---

**`src/hooks/chat/use-ingest-agent-notifications-from-events.ts:76`**

Effect re-scans the full `events` array from index 0 on every new event instead of tracking a processed cursor; O(n) work per event, quadratic over a long conversation.

Failure scenario: This effect depends on the whole `events` array and re-runs on every new action/observation/message during an active agent turn, looping over the full history each time (only skipping already-processed events via a Set lookup). In a long-running task with thousands of events, total work grows quadratically with conversation length.

---

**`src/components/features/chat/agent-notifications-pending-prompts.ts:6`**

`readPendingPromptsMap`/`writePendingPromptsMap` duplicate `readJsonMap`/`writeJsonMap` added in the sibling `agent-notifications-storage.ts` in this same PR.

Failure scenario: `agent-notifications-storage.ts` already defines generic helpers with the same try/catch JSON.parse + window guard + shape check; this file hand-rolls the identical logic for sessionStorage. A future fix to the shared parsing logic has to be applied twice, and it's easy to patch one copy while missing the other.

---

**`src/hooks/chat/use-ingest-agent-notifications-from-events.ts:19`**

`hasSubstantiveAgentActions` reimplements the identical predicate already memoized in `useFilteredEvents` (also used elsewhere in this PR).

Failure scenario: `use-filtered-events.ts` already exposes `hasSubstantiveAgentActions` using the identical check, and `chat-interface.tsx` already destructures it in this same PR. This hook instead defines a free function with the same body, re-scanning the full events array from scratch. If the shared definition changes later, only one copy gets updated, letting notification-detection and the rest of the app disagree about what counts as substantive work.

---

**`src/components/features/conversation/agent-notifications-bell.tsx:56`**

Hand-rolls outside-click/Escape handling instead of reusing `useClickOutsideElement`, which already supports the "ignore trigger button" case needed here (used correctly by `ContextWindowMeter` in this same PR).
Failure scenario: `useClickOutsideElement` already exists and even supports an `ignoreOutsideClickRef` argument for exactly this "don't close when clicking the trigger button" need. `AgentNotificationsBell` instead wires its own `document.addEventListener("mousedown", ...)` plus manual containment checks. Any future bug fix to the shared hook won't apply to the bell's separate implementation, so the two dropdowns can drift into inconsistent close-on-outside-click behavior.

---

**`src/i18n/translation.json:11715`**

Three new i18n keys (`KIND_AUTOMATION`, `SKILL_PROMPT`, `AUTOMATION_PROMPT`) translated across ~15 languages but never referenced by any code in the diff.

Failure scenario: These keys are defined with full multi-language translations but no `I18nKey.*` usage appears anywhere in the diff's TS/TSX code — the pill component only ever renders `KIND_SKILL` or the `AUTOMATION_TYPE_PILL` variant. These ~45 unused per-language strings add dead weight to translation.json and to translators' workload with no UI ever surfacing them.

PR #16317, opened 2026-08-04
fix: use hostname as default backend label instead of hardcoded "Local"

### LINKED ISSUE #16297: [Bug]: Default "Local" backend name is confusing when on a VM
### Operating System

macOS

### Installation Method

### Agent Canvas Version

1.8.0

### Bug Description

Started openhands on a VM. It auto-labels the current backend "Local" even though it's running on a VM

I wonder if we should name it something like the hostname? So it'd be `localhost` when running locally, or `canvas.acme.com` on a VM

### Steps to Reproduce

Set up canvas on a VM

### Actual Behavior

backend is called "Local"

### Expected Behavior

Backend is called something more descriptive

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

waterWang:
## Summary

Fixes #16297

When running agent-canvas on a VM, the backend was labelled "Local" even though the URL points to a remote hostname. This is confusing — the user expects to see the actual hostname (e.g. `canvas.acme.com`).

## Changes

- **`src/api/backend-registry/default-backend.ts`**: In `makeDefaultLocalBackend()`, extract the hostname from the configured base URL and use it as the display name. Fall back to "Local" only for loopback addresses (`localhost`, `127.0.0.1`, `::1`).

## Testing

- Localhost: `VITE_BACKEND_BASE_URL=http://localhost:3000` → name = "Local"
- VM: `VITE_BACKEND_BASE_URL=https://canvas.acme.com` → name = "canvas.acme.com"
- VM with port: `VITE_BACKEND_BASE_URL=https://canvas.acme.com:8443` → name = "canvas.acme.com:8443"
- Invalid URL → fallback to "Local"

### DIFF
--- a/src/api/backend-registry/default-backend.ts
+++ b/src/api/backend-registry/default-backend.ts
@@ -60,9 +60,25 @@ export function makeDefaultLocalBackend(): Backend | null {
 
   if (!host || !apiKey) return null;
 
+  // Use the hostname as the backend name for clarity. On a VM the base URL
+  // is something like https://canvas.acme.com, so showing "Local" is
+  // misleading. Fall back to the default name for loopback addresses.
+  let name: string;
+  try {
+    const url = new URL(host);
+    const hostname = url.hostname;
+    if (hostname === "localhost" || hostname === "127.0.0.1" || hostname === "::1") {
+      name = DEFAULT_LOCAL_BACKEND_NAME;
+    } else {
+      name = url.port ? `${hostname}:${url.port}` : hostname;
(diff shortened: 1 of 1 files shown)


enyst (changes requested):
Hey @waterWang thanks for the PR!

Two quick issues: CI is failing, and, have you tested this on a VM or alternative local install? Not sure, but I find it a bit surprising that the code you propose attempts to look for loopback addresses.

PR #16318, opened 2026-08-04
fix: stop the "Creating conversation" toast getting stuck on screen

### LINKED ISSUE #15701: [Bug]: Creating Conversation toast stuck visible
### Operating System

macOS

### Installation Method

Docker (`ghcr.io/openhands/agent-canvas`)

### Agent Canvas Version

1.0.0-beta.6

### Bug Description

The "creating conversation" toast remained visible after starting a conversation.

### Steps to Reproduce

All I remember doing is starting a conversation - this seems to be intermittent.

### Actual Behavior

The "creating conversation" toast remained visible after starting a conversation.

### Expected Behavior

The "creating conversation" toast should disappear after a few seconds.

### Relevant Logs

```shell

```

### Screenshots

Image

### Additional Context

_No response_

rishavnaskar:
HUMAN:

Went after this one because the "intermittent" in the report bugged me — the dismiss looked covered on every path, so I dug into what actually keeps the mutation pending. Reproduced it end to end and the screenshot from the failing run is basically identical to the one in the issue.

AGENT:

---

## Why

`toast.loading` has an infinite duration in react-hot-toast — `loading` is the one type with no default timeout:

```js
se = { blank: 4e3, error: 4e3, success: 2e3, loading: 1/0, custom: 4e3 }
```

`TOAST_OPTIONS` doesn't set one either, so the "Creating conversation…" toast only ever leaves the screen when something dismisses it by id. In `home-chat-launcher.tsx` that was a single `toast.dismiss(toastId)` inside a promise chain, under a generated id.

Two things fall out of that:

**The toast outlives the page it belongs to.** The create mutation keeps running well past the point where the conversation exists — after `AgentServerConversationService.createConversation` returns, `useCreateConversation` still awaits `ensureQueryData(PLUGINS_QUERY_KEYS.installed)` to snapshot the installed plugins into client metadata. On a cold cache that's a real round trip with React Query's default three retries and backoff (the app sets no `defaultOptions`, so v5 defaults apply). Leave the home page inside that window — click a sidebar conversation, switch backend — and the launcher unmounts while the chain is still awaiting. Nothing takes the toast down.

**Once it's stranded, nothing can reach it.** The id was generated, so no other code can name that toast. A later successful create dismisses only its own. It stays until reload.

That also explains the "intermittent" in the report: `ensureQueryData` returns cached data immediately when the query is already populated, so the window only opens on a cold plugins cache — typically the first conversation created after a page load.

## Summary

- Give the toast a fixed id (`creating-conversation`), matching what `use-new-conversation-command.ts` already does with `clear-conversation`. A leaked toast stays addressable, and a repeat submit reuses it instead of stacking another.
- Dismiss in a `finally` so every exit path is covered, including the early returns in the attachment branches.
- Dismiss on unmount, so a create that outlives the launcher can't strand it.
- Drop the retries on the post-creation installed-plugins lookup. It runs after the conversation already exists but still holds the mutation open, and the two profile lookups in the same function already use `retry: false`.

## Issue Number

Fixes #15701

### End-to-end

Added `creating-conversation toast does not outlive the home page` to `tests/e2e/mock-llm/regressions/mock-llm-ui-regressions.spec.ts`. It stalls `GET /api/plugins/installed`, submits from the home launcher, waits for the toast, then clicks a seeded conversation in the sidebar. The navigation is a link click rather than `page.goto` on purpose — a full page load drops the toast by itself and would pass with or without the fix.

Conversations are seeded through `POST /api/conversations`, which never contacts the LLM at creation time, so no model credentials are needed.

```
MOCK_LLM_PYTHON=.mock-llm-venv/bin/python3 npx playwright test \
  tests/e2e/mock-llm/regressions/mock-llm-ui-regressions.spec.ts
```

```
6 passed (44.4s)
```

The full mock-LLM suite is green in CI too — `62 passed (9.8m)`.

**It fails on the parent commit.** I reverted just the two `src/` files to `HEAD~1`, rebuilt, and re-ran:

```
Locator:  getByText('Creating conversation…', { exact: true })
Expected: 0
Received: 1
Timeout:  15000ms

  1 failed
```

The screenshot from that failing run shows the conversation sitting in the sidebar, the app on a different conversation, and the toast still up in the corner — the same state as the issue screenshot.

Note the E2E config only builds when `build/index.html` is missing, so `npm run build:app` is needed between toggling the fix; a stale `build/` silently tests the wrong code.

### Unit

Two cases added to `__tests__/components/features/home/home-chat-launcher.test.tsx`. The unmount one fails on the parent commit:

```
× takes the creating-conversation toast down when the launcher unmounts before creation settles
```

```
npx vitest run __tests__/components/features/home/home-chat-launcher.test.tsx
```

```
Test Files  1 passed (1)
     Tests  13 passed (13)
```

Full suite, `npx vitest run`:

```
Test Files  6 failed | 538 passed | 1 skipped (545)
     Tests  16 failed | 4256 passed | 5 skipped | 9 todo (4286)
```

Those failures are pre-existing on `main`. I captured the failing-file list on a clean checkout and after the change and diffed them — the only difference is that `home-chat-launcher.test.tsx` no longer appears:

```
3d2
<  ❯ __tests__/components/features/home/home-chat-launcher.test.tsx (11 tests | 9 failed)
```

That file was failing 9 of 11 on `main` before this PR. The create mutation awaits two round trips of its own — agent profiles before it creates anything, installed plugins after — and neither was stubbed, so they escaped to the network and the mutation never reached `createConversation` inside the assertion window. Stubbing them in `beforeEach` was needed to make the new cases trustworthy, and it takes the file green as a side effect. Happy to split that into its own PR if you'd rather keep this focused.

Also green: `npm run typecheck`, `npx eslint src`, `prettier --check`, `npm run build:app`.

### Manually

```
npm run dev
```

Throttle or block `GET /api/plugins/installed` in devtools, submit a prompt from the home page, then click a conversation in the sidebar while the toast is up. Before the change the toast stays for the rest of the session; after it goes away with the launcher.

## Video/Screenshots

The mock-LLM run records video (`video: "on"` in `playwright.mock-llm.config.ts`), so `test-results-mock-llm/` holds the before/after captures alongside the failing-run screenshot described above.

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The dismiss right after `createConversation` resolves is kept on top of the `finally`, so the toast still goes away before the attachment upload / pending-message step rather than sitting there labelled "Creating conversation…" during work that isn't creating anything.
- `retry: false` bounds the failure case but not a hang — React Query's default `networkMode: "online"` pauses a query indefinitely while the browser reports offline, and that lookup would still hold the mutation open. The unmount cleanup covers the visible symptom either way. Moving the plugin snapshot out of the awaited path entirely would close it properly, but that changes when the metadata lands, so it felt like a separate call to make.

### DIFF
--- a/__tests__/components/features/home/home-chat-launcher.test.tsx
+++ b/__tests__/components/features/home/home-chat-launcher.test.tsx
@@ -2,11 +2,13 @@ import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-import toast from "react-hot-toast";
+import toast, { useToasterStore } from "react-hot-toast";
 
 import { HomeChatLauncher } from "#/components/features/home/home-chat-launcher";
 import AgentServerConversationService from "#/api/conversation-service/agent-server-conversation-service.api";
 import WorkspacesService from "#/api/workspaces-service/workspaces-service.api";
+import AgentProfilesService from "#/api/agent-profiles-service/agent-profiles-service.api";
+import PluginsManagementService from "#/api/plugins-management-service";
 
 const mockNavigate = vi.fn();
 const mockUseActiveBackend = vi.fn();
@@ -240,6 +242,51 @@ const renderLauncher = () =>
     ),
   });
 
+/**
+ * Reports how many toasts are currently on screen. Reads react-hot-toast's own
+ * store rather than the rendered `<Toaster />`: a dismissed toast keeps its DOM
+ * node for the exit animation, so "still in the document" is not the same
+ * question as "still shown".
+ */
+function VisibleToastCount() {
+  const { toasts } = useToasterStore();
+  return (
+    <span data-testid="visible-toast-count">
+      {toasts.filter((item) => item.visible).length}
+    </span>
+  );
+}
+
+/**
+ * Renders the launcher next to the toast probe. `unmountLauncher` drops only
+ * the launcher, leaving the probe mounted so the toast state can still be read
+ * after the component that created the toast is gone.
+ */
+const renderWithToastProbe = () => {
+  const client = new QueryClient({
+    defaultOptions: {
+      queries: { retry: false },
+      mutations: { retry: false },
+    },
+  });
+  const view = render(
+    <QueryClientProvider client={client}>
+      <VisibleToastCount />
+      <HomeChatLauncher />
+    </QueryClientProvider>,
+  );
+
+  return {
+    ...view,
+    unmountLauncher: () =>
+      view.rerender(
+        <QueryClientProvider client={client}>
+          <VisibleToastCount />
+        </QueryClientProvider>,
+      ),
+  };
+};
+
 function makeConversationResponse(
   overrides: Record<string, unknown> = {},
 ): never {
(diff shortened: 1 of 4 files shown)


rishavnaskar:
@hieptl Kindly review this

PR #16324, opened 2026-08-04
fix(automations): surface non-MCP-installable integrations instead of silently dropping them

### LINKED ISSUE #16292: Do not silently filter incompatible automation integrations
## Problem
Recommended automation cards silently drop required integrations that are not locally installable MCP entries.

For example, the `jira-issue-to-pr` automation declares `jira` and `github`. The `jira` catalog entry currently has an HTTP/OpenAPI connection option rather than an MCP connection option. Canvas calls `getMcpMarketplaceCatalog()`, so `jira` is removed before rendering and before the install queue is built. The automation appears to require only GitHub even though Jira is also required.

## Expected behavior
Canvas should not silently discard a declared dependency. It should either:

- show the dependency and explain that this backend cannot install/configure it;
- route the user to the appropriate setup flow; or
- clearly mark the automation as unavailable with the reason.

The same behavior should apply to any future integration whose provider or transport is incompatible with the current local MCP installer.

## Context
- `src/components/features/automations/recommended-automations-section.tsx` filters through `getMcpMarketplaceCatalog()`.
- `src/components/features/automations/recommended-automations-launcher.tsx` uses the same MCP-only lookup for required entries.
- The extensions catalog validates that IDs exist, but this runtime compatibility mismatch is currently not surfaced.

### Actual Behavior
When I run Agent Canvas locally via `npm run dev` and open the Automate tab, the `jira-issue-to-pr` recommendation card shows only the GitHub integration. The required `jira` integration is silently dropped because its catalog entry only exposes an HTTP/OpenAPI connection option and is not present in the MCP marketplace catalog that Canvas renders from.

jira-issue-to-pr card

### Acceptance Criteria
- [ ] The recommendation card keeps every declared integration visible, including non-MCP-installable ones.
- [ ] A non-MCP-installable integration is labeled (e.g. "Needs external setup") or routed to the appropriate setup flow instead of being dropped.
- [ ] The behavior applies to any future integration whose provider or transport is incompatible with the local MCP installer.

lokesh75-kank:
HUMAN:

jira-card

Ran the automations and marketplace-utils test suites locally after the
maintainer's regression test and fix landed on the branch (22/22 and 23/23
passing, typecheck clean). Verified in the running app (Automate → Templates)
that the jira-issue-to-pr card shows both Jira (labeled "Needs external setup")
and GitHub instead of GitHub only, and that the "MCPs to connect" count only
covers GitHub. Screenshot below.

AGENT:

AI-assisted change, prepared and verified with the contributor.

## Summary

Surface non-MCP-installable automation integrations on recommendation cards (with a "Needs external setup" label) instead of silently dropping them from the card and the install queue.

## Why

Fixes #16292 — recommended automation cards resolved integration IDs against the MCP-filtered marketplace catalog, so a declared dependency whose catalog entry has no MCP connection option (e.g. Jira's HTTP/OpenAPI-only entry) was silently removed. `jira-issue-to-pr` rendered as if it only required GitHub, and the launcher's install queue was built from the same lookup. This implements the first option from the issue: show the dependency and explain that this backend cannot install/configure it.

## What

- `getIntegrationEntries` (section) resolves against the **full** catalog and classifies each entry via a new `isMcpInstallableEntry` helper in `mcp-marketplace-utils`
- Cards keep every declared integration visible; one this backend cannot install as MCP gets a **"Needs external setup"** label (new i18n key `RECOMMENDED_AUTOMATIONS$EXTERNAL_SETUP`, translated for all 15 locales)
- The **"N MCPs to connect"** count only covers entries the install flow can actually connect
- The launcher still queues install modals only for MCP-installable required entries — same behavior, but the exclusion is now explicit in code and user-visible on the card
- Searching by the external integration's name (e.g. "jira") still finds the automation

## How to Test

1. `npx vitest run __tests__/components/automations/recommended-automations.test.tsx` — includes three new tests: external integration stays visible/labeled and excluded from the connect count; search by its name finds the automation; the launcher install queue starts at the first MCP-installable entry
2. `npx vitest run __tests__/utils/mcp-marketplace-utils.test.ts`
3. Manually: open Automations on a local backend — the `jira-issue-to-pr` card lists both Jira ("Needs external setup") and GitHub, with "1 MCP to connect"

## Video/Screenshots

The `jira-issue-to-pr` card on a local backend after this change — Jira stays visible with the "Needs external setup" label next to the installable GitHub entry (verified live via `npm run dev:mock`):

jira-issue-to-pr card showing Jira "Needs external setup" and GitHub pills


neubig (approved):
Thanks a lot for this! I made a few changes, but it should be good.

PR #1587, opened 2024-05-05
Mentioned LLM logs directory

SmartManoj:
Helpful for issues like #1578

PR #16342, opened 2026-08-05
fix: send resolved LLM model for cloud profile launches

juanmichelini:
HUMAN:

- [x] this was tested by a human, me, juanmichelini. evidence https://github.com/OpenHands/saas-deploy/pull/511

AGENT:

The change is payload-only. It keeps Canvas cloud conversation starts aligned with legacy model classification by sending the resolved OpenHands LLM model when available, while withholding ACP display model strings from billing/model gates.

---

## Why

Canvas cloud launches primarily via `agent_profile_id`, while the legacy UI sends a flat `llm_model`. Sending the flat resolved model alongside the profile id lets app-server billing/model gates classify zero-cost/free models before profile-resolved runtime provisioning.

Backend should still resolve the effective model before enforcing credit gates; this frontend payload makes the current cloud request shape explicit and backward-compatible.

## Summary

- Include the effective `llm_model` in cloud app-conversation start requests.
- Resolve the model from the active/referenced OpenHands LLM profile when Canvas launches via `agent_profile_id`, including cold-cache sends.
- Preserve ACP launches by not inheriting unrelated OpenHands LLM models, including the `/new` command path.

## Issue Number

N/A

## How to Test

Run:

```bash
npx vitest run __tests__/api/cloud/conversation-create.test.ts __tests__/hooks/mutation/use-create-conversation.test.tsx __tests__/hooks/mutation/use-new-conversation-command.test.tsx
npm run typecheck
```

Expected result: the focused Vitest suite passes and TypeScript typecheck completes without errors.

## Video/Screenshots

Payload-only request-shape change; no user-visible UI change.

No visual UI change - payload only

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The non-Docker mock-LLM E2E failure is in `mcp/mock-llm-mcp-slack-credentials.spec.ts` waiting for `mcp-marketplace-grid`, outside this PR's touched paths. The Docker E2E run also timed out after passing earlier unrelated suites.

---

| Component | Value |
|---|---|
| **Commit** | `76bb94d29a8ce511dedc1858f5114887b93e3b35` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-76bb94d
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-76bb94d
```

```
ghcr.io/openhands/agent-canvas:sha-76bb94d-amd64
ghcr.io/openhands/agent-canvas:fix-canvas-cloud-free-model-payload-amd64
ghcr.io/openhands/agent-canvas:pr-16342-amd64
ghcr.io/openhands/agent-canvas:sha-76bb94d-arm64
ghcr.io/openhands/agent-canvas:fix-canvas-cloud-free-model-payload-arm64
ghcr.io/openhands/agent-canvas:pr-16342-arm64
ghcr.io/openhands/agent-canvas:sha-76bb94d
ghcr.io/openhands/agent-canvas:fix-canvas-cloud-free-model-payload
ghcr.io/openhands/agent-canvas:pr-16342
```

- Each tag (e.g., `sha-76bb94d`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-76bb94d-amd64`) are also available if needed


juanmichelini:

src/hooks/mutation/use-new-conversation-command.ts
```diff
@@ -38,6 +38,9 @@ export const useNewConversationCommand = () => {
         undefined,
         undefined,
         conversation.sandbox_id ?? undefined,
+        undefined,
+        undefined,
+        conversation.llm_model,
```
juanmichelini: 🔥 The one that got away. `AppConversation.llm_model` is documented (in `agent-server-conversation-service.types.ts`, on `agent_kind`) as potentially carrying **the ACP subprocess model for display**. This pass-through is unconditional — so `/new` on an ACP cloud conversation ships e.g. a claude-code model string to the app-server's billing/model gates as if it were an OpenHands LLM model. That's precisely the misclassification this PR is trying to prevent in the main launch path.

Suggest gating:

```ts
conversation.agent_kind === "openhands" ? conversation.llm_model : null,
```

(or `!= null && agent_kind !== "acp"` if you want to be conservative about older backends that omit `agent_kind` — your call, but make it a *decision*). And please add the ACP regression test here, mirroring the one you added for `useCreateConversation`.

src/hooks/mutation/use-create-conversation.ts
```diff
@@ -113,6 +114,7 @@ export const useCreateConversation = () => {
 
       const requestedAgentProfileId =
         agentProfileId ?? agentProfiles?.active_agent_profile_id ?? undefined;
+      let resolvedLlmProfiles: ProfileListResponse | undefined = llmProfiles;
```
juanmichelini, line 117: This initialization is the feature's silent failure mode. When the `useLlmProfiles()` query hasn't resolved yet (cold cache, errored fetch), `resolvedLlmProfiles` stays `undefined`, and both the profile-without-ref fallback and the no-profile path below resolve `cloudLlmModel` to `null` — the exact payload this PR exists to send just… isn't sent, with no trace.

The branch ~30 lines down already demonstrates the right pattern: `await queryClient.ensureQueryData({ queryKey: [...LLM_PROFILES_QUERY_KEYS.all, backend.id, orgId], queryFn: ProfilesService.listProfiles, retry: false })`, because "a send fired before that query loads must still validate." If the model is worth sending, it's worth awaiting — at least when `isCloud`. Best-effort is fine as a product decision, but then say so in a comment rather than letting it fall out of initialization order.

juanmichelini (commented):
**Taste Rating:** 🟡 **Acceptable** — the idea is right (make the cloud request shape explicit), but the execution has a genuine hole and a gratuitous ternary lasagna.

Since you asked for the roast: this PR carefully guards ACP launches from inheriting an unrelated OpenHands model in `useCreateConversation`… and then the `/new` command two files over yeets `conversation.llm_model` into the payload *unconditionally* — including, per your own type docs, the ACP subprocess model. The left hand wrote an invariant; the right hand never got the memo. Also, the "always send the effective model" feature silently no-ops whenever the LLM-profile query hasn't resolved yet — which is exactly the cold-cache race the neighboring code goes out of its way to `ensureQueryData` against. Pick a lane.

**[CRITICAL ISSUES]**

- [src/hooks/mutation/use-new-conversation-command.ts, Line 43] **Correctness — ACP model leak via `/new`**: `AppConversation.llm_model` explicitly "may carry the ACP subprocess model for display" (see the `agent_kind` doc in `agent-server-conversation-service.types.ts`). Passing it through unfiltered means `/new` on an ACP cloud conversation ships a non-OpenHands model string to the app-server's billing/model gates — the exact misclassification this PR exists to prevent. Gate it: `conversation.agent_kind === "openhands" ? conversation.llm_model : null`, and add a regression test for the ACP case — you wrote one for the main hook, so you clearly know how.

- [src/hooks/mutation/use-create-conversation.ts, Line 199] **Good taste — nested ternary**: three levels of `?:` to pick a profile name is doing parkour where a walk would do. A short `if/else` (or an early-return helper) says the same thing and doesn't require the reader to hold a parse stack in their head.
- [src/hooks/mutation/use-create-conversation.ts, Line 117] **Cold-cache gap**: `resolvedLlmProfiles` starts as the maybe-unresolved `useLlmProfiles()` data. Two branches below (profile-without-ref → `active_profile` fallback, and the no-profile path) read it without awaiting. Fire a send before the profiles query resolves and `cloudLlmModel` is `null` — the PR's entire payload quietly absent. Thirty lines up you `ensureQueryData` the same query because "a send fired before that query loads … must still validate." The model resolution deserves the same discipline when `isCloud`.
- [src/api/conversation-service/agent-server-conversation-service.api.ts] **Positional-param debt**: the TODO(#1587) got bumped from 11 to 12 params, and the `Parameters<typeof …>` + `.push(undefined, id, kind, model)` incantation is a monument to why. Not a blocker, but every PR that extends this tail instead of paying it down makes the eventual refactor worse. Strongly consider doing the options-object refactor *here* instead of around it — you're already touching every call site.

**[TESTING GAPS]**

- The new tests cover the happy paths well (profile launch with model, ACP exclusion, legacy nulls). Missing: the `/new`-command ACP case above, and the cold-cache case (`useLlmProfiles()` unresolved → cloud launch). The latter is the difference between "feature works" and "feature works when the user clicks fast."

**[RISK ASSESSMENT]**
The payload is additive and nullable (`llm_model: null` is backward-compatible), unit coverage is decent, and the backend still resolves the effective model as the final word. The medium grade is for the `/new` ACP leak (a wrong-model signal to billing gates) and the cold-cache flakiness making the feature nondeterministic. Neither is catastrophic; both are cheap to fix. The one mock-LLM E2E failure (`files/mock-llm-files-and-git.spec.ts` step 2) looks unrelated to this diff and passed in the Docker run, but confirm it's a flake before merging.

**VERDICT:** ❌ **Needs rework** — small rework: fix the `/new` ACP leak, await the profile list for cloud model resolution, and please un-nest that ternary.

**KEY INSIGHT:** An invariant enforced in one launch path and ignored in another isn't an invariant — it's a rumor.

---

>
>
>

---

_This review was generated by an AI agent (OpenHands) on behalf of @juanmichelini via `/codereview-roasted`._

juanmichelini:

src/hooks/mutation/use-create-conversation.ts
```diff
@@ -193,38 +209,64 @@ export const useCreateConversation = () => {
...
+      let referencedLlmProfile: string | undefined;
+      if (effectiveAgentProfileId && resolvedAgentProfile) {
+        if (resolvedAgentProfile.agent_kind === "openhands") {
+          referencedLlmProfile =
+            resolvedAgentProfile.llm_profile_ref ??
+            resolvedLlmProfiles?.active_profile ??
+            undefined;
+        }
+      } else {
+        referencedLlmProfile = resolvedLlmProfiles?.active_profile ?? undefined;
```
juanmichelini, line 225: 🔥 The cold-cache fix over-corrects here. This `else` now serves two very different cases:

1. **No profile launch** (`!effectiveAgentProfileId`) → `active_profile` is exactly right.
2. **Profile id set but unresolved** (`effectiveAgentProfileId` set, `resolvedAgentProfile` undefined — agent-profiles fetch failed while the in-conversation picker passed an explicit id, or a stale/deleted profile id) → we send the *active* profile's model alongside an `agent_profile_id` the backend will resolve to a potentially **different** model.

For a PR whose entire purpose is letting billing/model gates classify the correct model before provisioning, shipping a guessed model is worse than shipping none — and per the PR's own note, the backend resolves the profile's effective model server-side anyway. Suggest:

```ts
} else if (!effectiveAgentProfileId) {
  referencedLlmProfile = resolvedLlmProfiles?.active_profile ?? undefined;
}
// effectiveAgentProfileId set but unresolved → cloudLlmModel stays null;
// the backend classifies from the resolved profile.
```

Narrow edge, but it's the exact classification path this PR exists to get right.

juanmichelini (commented):
## Roasted re-review for PR #16342 🔥 (round 2)

**Taste Rating:** 🟡 **Acceptable** — both blockers from the first pass are properly fixed; one residual edge keeps this from a clean approve.

### Previous findings — both fixed ✅

1. **ACP model leak via `/new`** (`use-new-conversation-command.ts`) — now gated on `conversation.agent_kind === "acp" ? null : conversation.llm_model`, with a regression test (`does not forward ACP display models through /new`). Correct fix at the right layer.
2. **Cold-cache silent omission** (`use-create-conversation.ts`) — `ensureLlmProfiles()` now awaits the shared query on cloud sends instead of reading the maybe-unresolved hook result, with a regression test (`awaits LLM profiles on cold-cache cloud launches`). The single-fetcher closure is the right shape.

Verified locally on this branch: `npx vitest run __tests__/api/cloud/conversation-create.test.ts __tests__/hooks/mutation/use-create-conversation.test.tsx __tests__/hooks/mutation/use-new-conversation-command.test.tsx` → **24/24 pass**; `npm run typecheck` → clean; eslint on the three touched source files → clean. Docker E2E on the PR is 62/62; the one npm-mode failure is in `mcp/mock-llm-mcp-slack-credentials.spec.ts` waiting on `mcp-marketplace-grid`, outside this PR's touched paths.

**[CRITICAL ISSUES]**
None.

- [src/hooks/mutation/use-create-conversation.ts, Line 225] **Correctness**: the cold-cache fix over-corrects — the `else` fallback can now send a confidently *wrong* model where it used to send none. See inline comment.

**[NON-BLOCKING NOTES]**
- The `try/catch` at `use-create-conversation.ts:198` is now dead — `ensureLlmProfiles()` swallows errors internally and returns `undefined`, so the outer catch can never fire (the `?? false` handles it). Same for the redundant `resolvedLlmProfiles = llm` re-assignment at line 192. Harmless; cleanup candidate.
- The `/new` ACP guard keys off `agent_kind === "acp"`. `agent_kind` is optional on `AppConversation` — a cloud ACP conversation served by an app-server that omits the field would still forward its display model. Local conversations always get it via `toAppConversation`, so this is a diminishing edge, but `conversation.agent_kind === "acp" || conversation.acp_server != null` would close it completely.

**[RISK ASSESSMENT]**
Payload-only and additive (nullable `llm_model` on the cloud start request); the local launch path stays byte-identical, so existing local users are untouched. The elevated rating is because the value feeds app-server billing/model gates — a wrong model string has real cost/classification implications, which is exactly what the flagged edge can produce in the unresolved-profile path. **Recommendation**: a human maintainer should confirm the app-server precedence semantics (flat `llm_model` vs. profile-resolved model) before merge.

**VERDICT:**
✅ **Worth merging** once the unresolved-profile edge is handled (or consciously accepted with a comment). Everything from round 1 is properly addressed with tests.

**KEY INSIGHT:**
The fallback introduced to fix the cold-cache omission now over-corrects: when a profile id is set but unresolvable client-side, the code guesses the active profile's model instead of letting the backend resolve the profile authoritatively.

---

>
>
>

---

_This review was generated by an AI agent (OpenHands) on behalf of the user._

src/hooks/mutation/use-new-conversation-command.ts
juanmichelini: Fixed in 76bb94d — `/new` now gates the pass-through on `conversation.agent_kind === "acp" ? null : conversation.llm_model`, so ACP display models (e.g. claude-code strings) never reach the app-server billing/model gates. Regression test added: `does not forward ACP display models through /new` in `__tests__/hooks/mutation/use-new-conversation-command.test.tsx`. Verified passing locally. One residual diminishing edge noted in the re-review: the guard relies on `agent_kind` being present, which is optional on `AppConversation` for older app-server responses.

src/hooks/mutation/use-create-conversation.ts
juanmichelini: Fixed in 76bb94d — cloud sends now `await ensureLlmProfiles()`, which fetches via `queryClient.ensureQueryData` when the `useLlmProfiles()` hook result has not resolved yet (cold cache or errored fetch), so `cloudLlmModel` no longer silently resolves to `null`. Regression test added: `awaits LLM profiles on cold-cache cloud launches` in `__tests__/hooks/mutation/use-create-conversation.test.tsx`. Verified passing locally. The re-review flags one related residual edge (profile id set but unresolved → active-profile fallback can send a wrong model) as a new inline comment on line 225.

src/hooks/mutation/use-create-conversation.ts
juanmichelini: Narrowed the `else` to `} else if (!effectiveAgentProfileId)` so a set-but-unresolved `agent_profile_id` leaves `cloudLlmModel` null and lets the backend classify from the resolved profile, instead of guessing the active profile's model. Fixed in d428007.

src/api/conversation-service/agent-server-conversation-service.api.ts
```diff
@@ -407,6 +411,7 @@ class AgentServerConversationService {
         agent_type: agentType,
         sandbox_id: sandboxId ?? null,
         agent_profile_id: agentProfileId ?? null,
+        llm_model: cloudLlmModel ?? null,
```
neubig, line 414: I'm confused why we'd need to add this for the purpose of initially seeding the model on saas. Couldn't we jus tuse the appropriate agent/llm profile?

src/api/conversation-service/agent-server-conversation-service.api.ts
```diff
@@ -382,6 +382,10 @@ class AgentServerConversationService {
     // encrypted-settings builder; cloud sends it as a flat request field.
     agentProfileId?: string,
     agentProfileKind?: AgentKind,
+    // Cloud still accepts a flat llm_model field. Keep sending the effective
+    // model alongside agent_profile_id so app-server billing/model gates can
+    // classify free models before the profile-resolved runtime is provisioned.
+    cloudLlmModel?: string | null,
```
neubig, line 388: Does this mean that the cloud wasn't working at all before? I'm confused about why we'd need to add this anew.

PR #16344, opened 2026-08-05
fix: show full workspace file tree in Files tab on cloud backends (OHE-3053)

lilagrc:
## Description

On a **cloud** backend the Agent Canvas Files tab showed "No files in workspace" for a freshly cloned repo, while VSCode showed the full tree. On a **local** backend the tab works fine.

Root cause: `useWorkspaceFiles` derived the cloud listing from `git status` (`useUnifiedGetGitChanges`), which only reports changed/untracked files. A clean working tree → empty list. It intentionally did *not* enumerate unchanged tracked files because the cloud API had no full-workspace listing endpoint.

## Fix

Point the cloud path at a new first-class cloud API endpoint, `GET /api/v1/app-conversations/{id}/files`, which runs the same bounded `find` **server-side** on the conversation's runtime and returns the full tree. This reuses the proven server-side runtime-hop transport that git-changes/diff and single-file read already use, so it needs no CORS and doesn't depend on the removed `/api/cloud-proxy` hop.

Requires the companion backend endpoint: **OpenHands/enterprise#135**.

Fixes [OHE-3053](https://linear.app/all-hands-ai/issue/OHE-3053). Supersedes the closed #16340, which tried to drive bash `find` from the browser on cloud — that path tunnels through the removed `/api/cloud-proxy` and can't reach a real cloud runtime.

## Changes

- `src/api/cloud/conversation-service.api.ts`: add `listCloudConversationFiles(conversationId, path)`, mirroring `readCloudConversationFile`.
- `src/hooks/query/use-workspace-files.ts`: rewrite `useCloudWorkspaceFiles` to call the new endpoint (anchored at the absolute working dir the same way the diff view / single-file read compute it), with the same `normalizePath` / dedupe / cap logic as the local path. Local path is unchanged. Removed the `useUnifiedGetGitChanges` dependency from this hook.
- `__tests__/hooks/query/use-workspace-files.test.tsx`: update cloud tests to assert the full-tree listing via the new endpoint and that bash is never driven on cloud.

## Behavior notes

- Local backend behavior is unchanged (bash `find` directly).
- Cloud now shows unchanged tracked files too, matching local. Deleted files no longer appear (they weren't openable anyway); the diff view remains the place to see deletions.

## Testing

- `__tests__/hooks/query/use-workspace-files.test.tsx`: **3 passed**.
- `__tests__/routes/files-tab.test.tsx`: **22 passed** (no regressions).
- `tsc --noEmit` and `eslint` clean on changed files.

---

@lilagrc can click here to [continue refining the PR](https://app.all-hands.dev/conversations/1964c04f-dd26-43e0-a072-d7494fb9da42)

---

| Component | Value |
|---|---|
| **Commit** | `cd8085009c797291c705f45c436c3724d773e263` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-cd80850
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-cd80850
```

```
ghcr.io/openhands/agent-canvas:sha-cd80850-amd64
ghcr.io/openhands/agent-canvas:fix-cloud-workspace-file-listing-amd64
ghcr.io/openhands/agent-canvas:pr-16344-amd64
ghcr.io/openhands/agent-canvas:sha-cd80850-arm64
ghcr.io/openhands/agent-canvas:fix-cloud-workspace-file-listing-arm64
ghcr.io/openhands/agent-canvas:pr-16344-arm64
ghcr.io/openhands/agent-canvas:sha-cd80850
ghcr.io/openhands/agent-canvas:fix-cloud-workspace-file-listing
ghcr.io/openhands/agent-canvas:pr-16344
```

- Each tag (e.g., `sha-cd80850`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-cd80850-amd64`) are also available if needed


lilagrc:
Pushed a follow-up fix (`0feaa46`) after staging testing surfaced that **no `/files` request fired at all** on cloud.

**Root cause:** the cloud listing query gated `enabled` on the conversation id read from `useActiveConversation().data.id`. That comes from a separate batch-get query which can be `null`/still-loading, so `enabled` stayed `false` and the query never fired. The diff/commits cloud hooks avoid this by sourcing the id from the **route** (`useConversationId`) and using the active-conversation data only for path anchoring — this change now does the same via `useOptionalConversationId`. Added a regression test asserting the listing fires even when `useActiveConversation().data` is `undefined`.

**Note on staging:** the previous head (`107d3d9`) already had this bug, so a staging build pinned to that SHA would show exactly the reported symptom. To validate the fix, staging (pr-500) needs to be re-pointed at the new SHAs:
- OpenHands/OpenHands @ `0feaa467581a69dd0de968f00ff10a856b4b58c3`
- OpenHands/enterprise @ `f3daa32d703465241b8ddbb340f82150020d2d27` (the `/files` endpoint must be present, since the frontend calls it)

Both branches have been verified to contain the endpoint and the fix.

lilagrc:
Pushed `30af2b3` — the real root cause behind "no `/files` call" + "405 to `/api/cloud-proxy`" on staging.

**What was happening:** the Files tab decided cloud-vs-local from the React `ActiveBackendProvider` **context**. `useActiveBackend()` has a fallback that synthesizes a *local* backend when the provider isn't in scope for that render. The request transport, however, branches on the backend-registry **store** (`getActiveBackend()`) — that's what `executeCommand`, `getGitChanges`, and the cloud file-read all use.

On the locked-cloud `/canvas` subdomain the two disagreed: the tab saw `local`, so it ran the bash `find` path (`executeCommand`), which — because the *store* is cloud — POSTed to the removed `/api/cloud-proxy` → **405**. Meanwhile the cloud `/files` branch was gated off, so it never fired. That matches both symptoms exactly, and explains why my earlier id fix (`0feaa46`) alone didn't help even though staging was running it.

**Fix:** read cloud/local from the store via `useSyncExternalStore`, so the Files tab and the transport layer can't diverge (and it stays reactive to backend switches). This mirrors the already-working `use-workspace-file-content`, which reads `getActiveBackend()` directly.

Once staging rebuilds to `30af2b3` (or later), the tab should issue `GET /api/v1/app-conversations/{id}/files` and populate the full tree. Tests: `use-workspace-files` (4) + `files-tab` (22) pass; `tsc`/`eslint` clean.

lilagrc:
Pushed `a682dfd` to fix the regression you hit (agent-created/modified files disappearing).

**What went wrong:** my earlier change *replaced* the git-changes source with the new `/files` full-tree endpoint. So in any environment where `/files` returns empty or errors, the Files tab showed nothing — losing even the changed files it used to show via git status.

**Fix:** the cloud listing is now the **union** of the two sources:
1. `useUnifiedGetGitChanges` (unchanged) — always surfaces agent-created/modified files, exactly as before.
2. `GET /api/v1/app-conversations/{id}/files` — adds unchanged tracked files (the full repo tree).

If `/files` is empty or fails, the tab degrades gracefully to the old changed-files behavior instead of going blank; when it works, the full tree appears too. The full-tree query is additive and never blocks the tab's spinner. Tests: 5 hook + 22 files-tab pass; eslint clean.

**To help me finish the full-tree piece:** if after redeploying to `a682dfd` you see the changed files again but still *not* the full cloned-repo tree, could you check devtools for the `GET .../app-conversations/{id}/files` request and share (a) its HTTP status and (b) the response body? The most likely remaining cause is a working-directory mismatch (the `path`/cwd we pass vs. where the repo actually lives in the sandbox), and the response body will tell me immediately whether it's an empty `[]`, a partial list, or an error.

lilagrc:
Reverted the union commit `a682dfd` in `4ae43831c`.

Now that the two PRs are correctly deployed together on staging and manual validation confirms the cloned-repo tree and individual file previews work via `/files` alone, the git-changes union was unnecessary — `find` over the working tree already includes agent-created/modified files. The cloud path is back to the simpler `/files`-only listing (the earlier route-id and store-based cloud-detection fixes are retained).

Verified after revert: `use-workspace-files` (4) + `files-tab` (22) tests pass, eslint clean.

hieptl (approved):
Thank you! 🙏

PR #16347, opened 2026-08-05
fix: preserve workspace sessions across backend changes

QutritSystems:
HUMAN:
Please review this focused workspace session recovery fix.

AGENT:
## Summary
- Key workspace-session queries and retries by backend identity.
- Refreshes the current backend session once after the first authenticated file-read failure, then retries once without looping.
- Adds focused coverage for backend isolation and workspace file-content recovery.

## Why
Workspace file reads can retain a stale session when the selected backend changes. After an authenticated file-read failure, the UI should refresh that backend-specific session once and retry once rather than reusing another backend session or retrying indefinitely.

## How to Test
- Run i18n generation, typecheck, and ESLint/Prettier for changed files.
- Run the 54 focused tests covering conversation service, workspace-session, and workspace-file-content behavior.
- In Canvas, select a workspace, open the File panel, and switch between backends; confirm file access obtains the current backend session.
- Simulate a first 401 and confirm exactly one refresh/retry; confirm a second 401 or refresh failure surfaces an error without looping.

## Video/Screenshots
Production Canvas File panel with the File tab selected. The capture is cropped to the panel controls and contains no workspace paths, file contents, conversation content, or credentials.

Canvas File panel with File selected

## Warp context
- Conversation: https://app.warp.dev/conversation/82cf2096-d62c-4227-9889-7e1a7d7f99b8
- Upgrade plan: https://app.warp.dev/drive/notebook/O9hfZZT5OYgPZGULG3HoHz

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/src/api/conversation-service/agent-server-conversation-service.api.ts
+++ b/src/api/conversation-service/agent-server-conversation-service.api.ts
@@ -508,8 +508,11 @@ class AgentServerConversationService {
     // `useUnifiedVSCodeUrl` + `useCloudSandbox`); the runtime's own
     // `/api/vscode/url` only knows its internal `localhost:8001`, which
     // the user's browser can't reach.
-    const workspaceDir =
-      await this.resolveConversationWorkingDir(conversationId);
+    const workspaceDir = await this.resolveConversationWorkingDir(
+      conversationId,
+      conversationUrl,
+      sessionApiKey,
+    );
     // Local mode: the typescript-client targets the local agent-server
     // directly via the conversationUrl override.
     const vscodeUrl = await new VSCodeClient(
@@ -528,15 +531,21 @@ class AgentServerConversationService {
 
   static async resolveConversationWorkingDir(
     conversationId: string,
+    conversationUrl?: string | null,
+    sessionApiKey?: string | null,
   ): Promise<string> {
-    const [conversation] = await this.batchGetAppConversations([
-      conversationId,
-    ]);
+    const [conversation] = await this.batchGetAppConversations(
+      [conversationId],
+      conversationUrl,
+      sessionApiKey,
+    );
     return conversation?.workspace?.working_dir ?? getAgentServerWorkingDir();
   }
 
   static async batchGetAppConversations(
     ids: string[],
+    conversationUrl?: string | null,
+    sessionApiKey?: string | null,
   ): Promise<(AppConversation | null)[]> {
     if (ids.length === 0) return [];
 
@@ -545,7 +554,7 @@ class AgentServerConversationService {
     }
 
     const data = await new ConversationClient(
-      getAgentServerClientOptions(),
+      getAgentServerClientOptions({ conversationUrl, sessionApiKey }),
     ).getConversations<DirectConversationInfo>(ids);
 
     return requireDirectConversationItems(data).map((item) =>
@@ -589,6 +598,8 @@ class AgentServerConversationService {
   static async readConversationFile(
     conversationId: string,
     filePath?: string,
+    conversationUrl?: string | null,
+    sessionApiKey?: string | null,
   ): Promise<string> {
     if (getActiveBackend().backend.kind === "cloud") {
       // Cloud exposes a per-conversation file endpoint; the sandbox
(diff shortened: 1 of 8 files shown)


QutritSystems:
Hi maintainers — could someone please review this focused workspace-session recovery fix?

Current CI status:
- Ubuntu and Windows builds: passed
- Mock-LLM E2E: passed
- PR title and artifact checks: passed
- The latest `Validate PR description` run: passed after updating the description to the required template and adding a sanitized File panel screenshot
- No checks are currently pending

The PR is still shown as blocked because three earlier `Validate PR description` runs remain failed in the status rollup. Those runs predate the corrected description. I do not have upstream admin permission to rerun or dismiss them.

The implementation and tests cover backend-specific workspace-session identity, a single refresh/retry after the first authenticated file-read failure, and no retry loop after a second failure.

An approval and, if needed, maintainer action on the historical description-check runs would be appreciated. Thanks!

Co-Authored-By: Oz <oz-agent@warp.dev>

QutritSystems:
@hieptl @neubig @lilagrc — GitHub suggested you based on the files changed in this PR. If you have time, a review of the backend-specific workspace-session keying and one-time 401 refresh/retry behavior would be appreciated.

Co-Authored-By: Oz <oz-agent@warp.dev>

src/hooks/query/use-workspace-file-content.ts
VascoSch92, line 258: The new retry-on-401 logic only wraps the text-file `fetch()`. The `kind !== "text"` branch returns `staticUrl` straight to the /iframe consumer with no retry, so after the exact cookie-eviction scenario this PR targets, image/PDF previews still break while text files self-heal. 

Verified: read the full function, the early return for non-text kinds sits above the new retry block entirely.

src/api/conversation-service/agent-server-conversation-service.api.ts
```diff
@@ -601,22 +612,32 @@ class AgentServerConversationService {
...
       workingDir,
     );
-    return new FileClient(getAgentServerClientOptions()).downloadTextFile(path);
+    return new FileClient(
+      getAgentServerClientOptions({ conversationUrl, sessionApiKey }),
+    ).downloadTextFile(path);
   }
 
-  static async downloadConversation(conversationId: string): Promise<Blob> {
+  static async downloadConversation(
```
VascoSch92, line 629: The function now accepts conversationUrl/sessionApiKey, but its only caller, `src/hooks/use-download-conversation.ts` (untouched by this PR), still calls it with just `conversationId`. 

Trajectory downloads keep the stale-config bug this PR otherwise fixes.

src/contexts/conversation-websocket-context.tsx
```diff
@@ -427,6 +427,10 @@ export function ConversationWebSocketProvider({
         {
           conversationId: currentPlanningConversationId,
           filePath: path,
+          conversationUrl:
```
VascoSch92, line 430: This pairs a `conversationId` captured earlier (`latestPlanningFileEventRef.current`, set when the event first arrived) with `conversationUrl`/`session_api_key` read fresh from `subConversations?.[0]` here. If `subConversations[0]` changes identity between when the event was queued and when this effect fires (isLoadingHistoryPlanning flipping false), the id and credentials could belong to different sub-conversations, and the read would silently fail (only a console.warn). Should the URL/key be captured into the ref alongside conversationId at queue time, instead of re-read here?

src/hooks/query/use-workspace-file-content.ts
VascoSch92, line 250: `staticUrl` is computed once from `baseUrl` before the first fetch. If the 401 is caused by the underlying baseUrl actually changing (e.g. local agent-server restarted, refresh() re-mints against a new host/port), the retry at line 278 still hits this same stale `staticUrl` and fails again — with an error the user can't recover from without manually reselecting the file. Should the retry rebuild the URL from the refreshed session instead of reusing the closed-over one?

src/api/conversation-service/agent-server-conversation-service.api.ts
```diff
@@ -528,15 +531,21 @@ class AgentServerConversationService {
...
-    ]);
+    const [conversation] = await this.batchGetAppConversations(
+      [conversationId],
+      conversationUrl,
+      sessionApiKey,
+    );
     return conversation?.workspace?.working_dir ?? getAgentServerWorkingDir();
   }
 
   static async batchGetAppConversations(
```
VascoSch92, line 545: This takes an array of ids but only a single `conversationUrl`/`sessionApiKey` override pair. No current caller combines a multi-id batch with an override (use-sub-conversations.ts passes multiple ids with no override), but the signature invites it — if someone does, one conversation's host/key would silently get applied to every id in the batch. Worth a comment or type-level guard to prevent that combination, or restructuring the override to be per-id?

src/hooks/query/use-workspace-file-content.ts
```diff
@@ -142,7 +143,10 @@ function arrayBufferToBase64(buffer: ArrayBuffer): string {
 export function useWorkspaceFileContent(relativePath: string | null) {
   const { data: conversation } = useActiveConversation();
   const runtimeIsReady = useRuntimeIsReady();
-  const { data: workspaceSession } = useWorkspaceSession();
+  const { data: workspaceSession, refresh: refreshWorkspaceSession } =
+    useWorkspaceSession();
+  const refreshWorkspaceSessionRef = useRef(refreshWorkspaceSession);
```
VascoSch92, line 148: Is this ref actually needed? `queryFn` is a fresh inline closure on every render, and TanStack Query calls `setOptions` with the latest `queryFn` on every render before dispatching new fetches — so referencing `refreshWorkspaceSession` directly (without the ref) would already pick up the current value for any new fetch. The comment above frames the ref as required for cache-identity reasons, but it doesn't look like plain closure capture would behave differently here. Minor, but happy to be wrong if there's a reentrancy case I'm missing.

PR #16359, opened 2026-08-06
fix(config-service): surface OpenRouter and other model-only providers in Basic picker

### LINKED ISSUE #15576: OpenRouter missing from list of Basic LLM Providers

hudsonwa:
HUMAN:
Verified via the vitest suite for the config-service (4/4 pass) and confirmed RED→GREEN by stashing the source fix. This is a backend provider-list derivation change with no visible UI, so the screenshot shows the test output.

AGENT:

Automated verification only: 4/4 vitest tests pass with the fix; the new regression test fails without it (verified by stashing). No human test performed, no visual change. No secrets introduced; diff +18/-2 source, +40 test.

---

## Why

On the local backend, OpenRouter is missing from the Basic LLM Providers list in Agent Canvas, even though it is available on OpenHands Cloud. See #15576.

## Summary

- `searchProviders` in `config-service.api.ts` now also fetches `/api/llm/models` and derives providers from model ID prefixes (e.g. `openrouter/...`)
- Derived providers are added to the deduplicated set, marked as not verified
- No new dependencies, no UI change

## Issue Number

Fixes #15576

## How to Test

1. `npx vitest run __tests__/api/config-service.test.ts`
2. Expect 4/4 tests to pass, including the new `surfaces providers discoverable only from model ids (e.g. openrouter)` regression test
3. Stash the fix → the new test fails (`openrouter` undefined); restore → passes (RED→GREEN verified)

## Video/Screenshots

vitest output: 4/4 tests pass for config-service with OpenRouter regression test

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

No new dependencies. Provider-list derivation on the local backend config service. Diff +18/-2 source, +40 test.

### DIFF
--- a/__tests__/api/config-service.test.ts
+++ b/__tests__/api/config-service.test.ts
@@ -58,4 +58,44 @@ describe("ConfigService", () => {
     const openhands = page.items.find((p) => p.name === "openhands");
     expect(openhands).toEqual({ name: "openhands", verified: true });
   });
+
+  it("surfaces providers discoverable only from model ids (e.g. openrouter)", async () => {
+    // Arrange: mirror the real local agent-server, where OpenRouter is absent
+    // from /api/llm/providers and from /api/llm/models/verified, but its models
+    // are listed under /api/llm/models with an "openrouter/..." id prefix.
+    server.use(
+      http.get("/api/llm/providers", () =>
+        HttpResponse.json({ providers: ["anthropic", "openai"] }),
+      ),
+      http.get("/api/llm/models/verified", () =>
+        HttpResponse.json({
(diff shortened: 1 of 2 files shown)


juanmichelini:
@hudsonwa thank you for noticing the issue and doing a pull request. 

After comparing this with the later reproduction in #16453, it looks like the root cause was slightly different: `openrouter` is present in `/api/llm/providers`, but it was being truncated by the `limit: 100` call in the Basic picker path. Since deriving providers from model IDs can also introduce non-provider entries for some model IDs, I’m going to close this one in favor of the narrower fix in #16453.

PR #16381, opened 2026-08-07
fix(desktop): ship the bundled Node's npm on Windows

Telov:
HUMAN:

Had issues installing OpenHands on windows. Putting up some PRs to help the community

AGENT:

---

## Why

The packaged Windows build ships `resources\node\` without `node_modules`:

```
LICENSE  install_tools.bat  node.exe  nodevars.bat
npm  npm.cmd  npm.ps1  npx  npx.cmd  npx.ps1
```

`npm.cmd` / `npx.cmd` are the stock npm shims — they exec
`%~dp0\node_modules\npm\bin\npx-cli.js`, which isn't there. So every `npx`
resolved from this directory dies with `MODULE_NOT_FOUND`. And since
`injectBundledNode()` *prepends* this directory to `PATH`, it also shadows a
working system npm: the user's own install is fine and unreachable.

That breaks every `npx`-launched subprocess in the packaged app — the stdio MCP
servers this bundle exists to support (Slack, GitHub, Figma …) and ACP servers,
whose registry defaults all start with `npx -y`.

Scope, so this isn't oversold: for **stdio MCP servers** this is the whole fix.
The `mcp` client resolves a bare `npx` through `PATHEXT` itself
(`mcp/os/win32/utilities.py::get_windows_executable_command`), so once the
bundle is complete they work. For **ACP servers** it is one of two blockers —
`openhands-sdk` spawns its launch command with
`asyncio.create_subprocess_exec` and no `PATHEXT` lookup, so a bare `npx`
raises `[WinError 2]` before the shim is ever reached. That half is a separate
SDK fix — https://github.com/OpenHands/software-agent-sdk/pull/4408 — and
neither change makes ACP work on Windows without the other.

### Root cause

`app-builder-lib`'s copy filter drops a `node_modules` directory that sits at
the **root** of a `from` dir, before any `filter` pattern is consulted —
`packages/app-builder-lib/src/util/filter.ts`:

```ts
// filter the root node_modules, but not a subnode_modules
if (relative === "node_modules") {
  return false
}
```

That's reached for extraResources too: `copyFiles()` builds the copy filter via
the same `createFilter()` (`fileMatcher.ts:305-308`, `:527-555`). Our entry is

```js
{ from: "resources/node/", to: "node/", filter: ["**/*"] }
```

so npm lands at exactly `node_modules` relative to the copy root and is skipped.
No `filter` value can opt back in — the rejection precedes pattern matching.
(Both claims are exercised against the real library under How to Test.)

The Windows Node zip puts npm at `<root>/node_modules/npm`. POSIX tarballs put
it at `<root>/lib/node_modules/npm`, which isn't the root entry and copies
fine — so macOS and Linux are unaffected, which is presumably why this shipped.

## Summary

- `restoreBundledNodeNpm()` in the `afterPack` hook copies
  `resources/node/node_modules` into the packed `<Resources>/node/`, then
  **throws** if `npm-cli.js` still isn't in the packed output — mirroring the
  check `download-node.mjs` already runs on the source tree, so a build can no
  longer produce this artifact silently.
- `injectBundledNode()` warns at startup when the bundle is half-copied. It
  still injects (`node` works and the backend scripts need it), but a broken
  bundle is worse than no bundle because of the PATH shadowing, so it says so.
- Factored the `.app`-vs-flat Resources path resolution out of
  `stripBundledNodeModules` into `packagedResourcesDir()`; both hooks need it.
- AGENTS.md: documented the filter behaviour next to the existing
  `collectNodeModulesWithLogging` note.

## How to Test

**Reproduce the bug** against any installed 1.10.0 build:

```powershell
$env:PATH = "C:\Program Files\Agent Canvas\resources\node;$env:PATH"
npx --version
# MODULE_NOT_FOUND — node_modules\npm\bin\npx-cli.js
```

The shim fails before any package name is parsed, so this is independent of what
is being launched. Compare with the same command in a shell without that prefix.

**Verify the drop**, against electron-builder's own code rather than a reading
of it. This drives `getFileMatchers()` + `copyFiles()` — the two calls the
packager makes for `extraResources` — from `app-builder-lib@26.15.3`, the
version this repo pins, over a synthetic `resources/node/` in the real Node-zip
layout. The `extraResources` spec is imported from this repo's
`electron-builder.config.mjs`, not retyped:

```
app-builder-lib 26.15.3 (repo pins electron-builder 26.15.3)
config.extraResources = [{"from":"resources/bin/","to":"bin/","filter":["**/*"]},
                         {"from":"resources/node/","to":"node/","filter":["**/*"]}]

PASS  1. node.exe copied
PASS  1. npm.cmd copied
PASS  1. resources/bin copied
PASS  1. node/node_modules DROPPED  <-- the bug
PASS  2. still dropped with filter: explicit node_modules pattern
PASS  2. still dropped with filter: bare **
PASS  2. still dropped with filter: dotfile-inclusive
PASS  2. still dropped with filter: no filter at all
PASS  3. POSIX lib/node_modules/npm COPIED — only the root entry is dropped
```

Group 2 is the "no `filter` value can opt back in" claim, tested rather than
asserted — the rejection precedes pattern matching, so every spelling loses.
Group 3 is the reason macOS and Linux never saw this: same run, same config,
POSIX layout, and `lib/node_modules/npm` copies through untouched because it
isn't the copy root's own `node_modules`.

**Verify the hook.** I drove the real `afterPack` export from the patched config
against synthetic pack output in the Windows layout (script available on
request — it imports `config.afterPack` from this branch, it doesn't
reimplement it):

```
PASS  A precondition: packed output lacks npm-cli.js
[electron-builder] stripped bundled node_modules (~0 MB): packtmp\caseA\resources\app\node_modules
[electron-builder] restored runtime node_modules (sirv, httpxy; ~0 KB)
[electron-builder] restored bundled Node node_modules: packtmp\caseA\resources\node\node_modules
PASS  A after hook: npm-cli.js restored into packed output
PASS  A: app/node_modules still stripped to the runtime closure
PASS  B: throws when npm-cli.js cannot be produced
PASS  C: no-ops when the bundle has no node/ dir
PASS  D: second run is a no-op
```

C matters because a bare `electron-builder` run without `npm run download-node`
shouldn't start failing the build — that case is left to the existing startup
warning. A is the regression test: it also asserts the pre-existing
`resources/app/node_modules` strip still reduces to the `RUNTIME_PACKAGES`
closure.

**End-to-end on this PR's own CI.** I have no Electron toolchain locally, so the
two halves above are tested separately. `desktop-windows.yml` and
`desktop-macos.yml` both trigger on `electron/**` and
`electron-builder.config.mjs`, so this PR stitched them itself. Both green.

Windows ([run 31160703708](https://github.com/OpenHands/OpenHands/actions/runs/31160703708)):

```
[download-node] ✓ Node v22.12.0 ready at D:\a\OpenHands\OpenHands\resources\node (~91 MB)
[electron-builder] stripped bundled node_modules (~406 MB): dist-electron\win-unpacked\resources\app\node_modules
[electron-builder] restored runtime node_modules (sirv, httpxy, @polka/url, mrmime, totalist; ~123 KB)
[electron-builder] restored bundled Node node_modules: dist-electron\win-unpacked\resources\node\node_modules
```

That last line is the bug and the fix in one: the restore is guarded by
`!existsSync(destNodeModules)`, so it printing at all means the real packager
had dropped the directory from the real packed output. The `npm-cli.js`
assertion that runs immediately after it passed, so the installer this job
uploaded has a complete npm.

macOS ([run 31160703718](https://github.com/OpenHands/OpenHands/actions/runs/31160703718))
is the control, and behaves exactly as the root-cause analysis predicts — the
strip runs, and there is **no** `restored bundled Node node_modules` line,
because the POSIX `lib/node_modules/npm` was never dropped and there is nothing
to put back:

```
[download-node] ✓ Node v22.12.0 ready at /Users/runner/work/OpenHands/OpenHands/resources/node (~127 MB)
[electron-builder] stripped bundled node_modules (~401 MB): dist-electron/mac-arm64/OpenHands Agent Canvas.app/Contents/Resources/app/node_modules
[electron-builder] restored runtime node_modules (sirv, httpxy, @polka/url, mrmime, totalist; ~123 KB)
```

The POSIX branch of the new `npm-cli.js` assertion still ran there and passed,
so the added hard-fail doesn't put the macOS build at risk. Between the two runs
the extracted `packagedResourcesDir()` is exercised in both shapes it handles —
flat `resources/` and `.app/Contents/Resources` — against real
electron-builder contexts.

The uploaded `agent-canvas-windows-installer` artifact can also be installed
directly, after which `resources\node\node_modules\npm\bin\npx-cli.js` exists
and the repro command above prints a version instead of `MODULE_NOT_FOUND`.

## Video/Screenshots

n/a — terminal only, output above.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Evidence the shipped artifact is affected**, in case it's useful for
  triage — from the v1.10.0 release build (Actions run `31022666030`) and the
  installed result:
  - `[download-node] ✓ Node v22.12.0 ready … (~91 MB)` — the build host had a
    complete tree, and `verifyLayout()` would have thrown otherwise.
  - `[electron-builder] stripped bundled node_modules (~406 MB):
    dist-electron\win-unpacked\resources\app\node_modules` — the existing hook
    only ever touches `resources\app`.
  - The installed `resources\app\node_modules` contains exactly
    `@polka, httpxy, mrmime, sirv, totalist` (200 KB), matching the
    `restored runtime node_modules` log line. So installation preserves
    `node_modules` directories fine, and `resources\node\node_modules` was
    already gone in `win-unpacked`.
  - Installed `resources\node\` is ~83 MB against the ~91 MB reported at
    download time.
- The error this produces downstream is the expensive part: ACP servers surface
  it as `agent error connection closed` / `ACPInitError: Connection closed` a
  second or two in, with no mention of npx, node, or a missing module — it reads
  like a protocol or network fault. It also doesn't reproduce from a terminal,
  since a normal shell doesn't have the bundle in front.
- Lint: `npm run lint` scopes eslint and prettier to `src/`, so neither of the
  touched files is covered in CI. They're written to match anyway — the repo's
  `.prettierrc.json` (`trailingComma: "all"`, 80-col default) and the existing
  `// eslint-disable-next-line no-console -- electron-builder build log`
  convention already used three times in `electron-builder.config.mjs`.
  `electron/main.mjs` calls `console.*` without disables throughout, so the new
  warning there follows suit.
- Related: #15488, also Windows subprocess launching in the desktop app
  (shell-parsing of `<0.11` in the agent-server command). Different mechanism,
  same area.


dar-brett:
Independently reproduced from a clean official Canvas main build. The Windows package contained `npx.cmd`, but electron-builder omitted `resources/node/node_modules/npm`, including `npm/bin/npx-cli.js`. The bundled Node directory therefore shadowed a working system npm with broken shims.

Applying this PR restored and validated the npm payload in the packaged resources. With OpenHands/software-agent-sdk#4408 also applied, the rebuilt desktop runtime completed a real Codex ACP session successfully, with no `ACPSpawnError`, `WinError 2`, stopped-server, or error-observation events.

Recommend merging. This fixes bundled `npx` consumers such as stdio MCP servers and provides one of the two required ACP fixes; OpenHands/software-agent-sdk#4408 supplies the other.

neubig (approved):
Seems OK to me, thank you!

PR #16391, opened 2026-08-07
fix: show specific message when LLM profile exists but API key is missing

hudsonwa:
HUMAN:

Fixes the misleading "LLM not set up" banner when the user already has an active LLM profile but its API key is missing. The banner now shows a profile-specific message and CTA.

AGENT:

## Why

When an active LLM profile exists but `api_key_set` is false, the home page showed the generic "Your LLM isn't set up yet" message, which is confusing because the user has already configured a profile — the profile just needs its API key re-entered.

## Summary

- Extended `useLlmConfigured` hook to return a `profileMissingKey` boolean when an active profile exists without a usable API key
- Updated `LlmNotConfiguredBanner` to show the profile-specific message and CTA when `profileMissingKey` is true
- Added i18n keys `HOME$LLM_NOT_CONFIGURED_PROFILE_KEY_MISSING` and `HOME$LLM_PROFILE_KEY_MISSING_ACTION`

## Issue Number

OpenHands/OpenHands#15609

## How to Test

1. Run `npx vitest run __tests__/components/home/llm-not-configured-banner.test.tsx` — 7 tests pass
2. Start the app with a local backend where an active LLM profile exists but has no API key
3. The banner now reads "Your active LLM profile is missing its API key. Edit the profile to re-enter the key before starting a conversation." instead of the generic message

## Video/Screenshots

Before vs After

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The `profileMissingKey` flag is only set in local mode (profiles are a local-backend concept). Cloud backends and ACP agents are unaffected.


hudsonwa:
Closing as a duplicate of my earlier PR #16361, which clarifies the home-page banner when the active LLM profile is missing its API key for this same issue. Keeping #16361.

PR #16393, opened 2026-08-07
fix: auto-recover from stale SQLite migration on automation backend

### LINKED ISSUE #15623: Automation tab unavailable: stale SQLite DB migration crashes automation service on startup
## Human Description

I think this happens because I started with an older agent-canvas but new users would not experience this. I think.

---
## Bug Description

When upgrading `openhands-automation` between versions that restructure their Alembic migration history, the automation service crashes on startup with:

```
Failed to apply SQLite migrations: Can't locate revision identified by '007'
```

This causes the automation backend to exit with code 3, making the **Automate tab** completely unavailable in the UI.

## Root Cause

The automation service stores its SQLite database at `~/.openhands/automation/automations.db`. When the `openhands-automation` package is updated to a version that has restructured its Alembic migration revisions (e.g., the old DB references revision `007` which no longer exists in the new migration chain), Alembic's `upgrade head` command fails because it can't find the revision the DB is currently at.

The dev script (`dev-with-automation.mjs`) spawns the automation backend but has no recovery logic for this failure — the service simply crashes and the tab remains broken.

## Steps to Reproduce

1. Run `npm run dev` with an older version of `openhands-automation` so that the DB is created with an older migration chain
2. Update to a newer version of `openhands-automation` (e.g., via `git pull` on `agent-canvas` main which bumps the pinned version)
3. Run `npm run dev` again
4. Observe the automation service crash with the migration error
5. The Automate tab is unavailable

## Expected Behavior

The dev script should detect the migration failure, automatically reset the stale database, and restart the automation service — with a clear log message explaining what happened.

## Actual Behavior

The automation service exits with code 3 and logs a long traceback. The Automate tab shows as unavailable with no guidance on how to fix it.

## Environment

- OS: Windows 11
- agent-canvas: main branch (latest)
- openhands-automation: 1.0.0a6

## Relevant Logs

```
[automation] {"message": "Failed to apply SQLite migrations: Can't locate revision identified by '007'", "severity": "ERROR"}
[automation] {"message": "Application startup failed. Exiting.", "severity": "ERROR"}
[automation] Exited with code 3
```

## Proposed Fix

Add migration-error detection and auto-recovery in `startAutomationBackend()` within `scripts/dev-with-automation.mjs`:
1. Capture automation service output and watch for migration error patterns
2. On crash with migration error, delete the stale `automations.db` and retry once
3. Log clear messages about what happened and what was done

hudsonwa:
HUMAN:

I picked up #15623 looking for a first contribution here. Full honesty on the testing: I could not run the live launcher end-to-end against a real broken DB, so the evidence in this PR is the test suite instead - seven cases driving the recovery watcher, including one that deletes a real file on disk. The screenshot further down is that run on my machine.

AGENT:

## Why

When `openhands-automation` is updated and its Alembic revision chain is restructured, the old `automations.db` references a revision that no longer exists. The service exits with code 3 and the Automate tab stays broken with no recovery path.

## Summary

- Shared `watchAutomationMigration()` helper in `scripts/dev-process-utils.mjs`, used by both `dev-with-automation.mjs` and `dev-static.mjs`
- Process-wide latch: one recovery attempt per launcher run, then a loud give-up instead of a restart loop
- Restart only happens inside `onRecover` after a successful DB delete
- Seven regression tests covering pattern+code, latch across a new watcher, and a real-filesystem delete

## Issue Number

Fixes #15623

## How to Test

1. Simulate a stale DB: create a `~/.openhands/automation/automations.db` with an Alembic revision that doesn't exist in the current chain
2. Run `node scripts/dev-with-automation.mjs`
3. The automation backend crashes once with the migration error
4. The script detects it, deletes the stale DB, and restarts successfully
5. A second crash of the same kind does not restart again
6. `npx vitest run __tests__/scripts/dev-migration-watch.test.ts` — 7/7 passing

## Video/Screenshots

vitest: 7/7 passing on the migration recovery latch

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Recovery is single-shot for the whole process. If the restart also fails (different root cause), the second crash is not retried.

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.55s · commit 931f719  
**Strongest signal:** Data loss · 65% estimated likelihood.  
**Evidence:** [F003H004 · scripts/dev-static.mjs:403–423](https://github.com/OpenHands/OpenHands/blob/931f719beb4db7906abaa7e767d72c689db73c44/scripts/dev-static.mjs#L403-L423).  
**Coverage:** complete supplied coverage; 11/11 hunks, 4/4 files.

<details>

| --- | --- | --- |
| Command injection | 6.0% | No direct hunk selected |
| Weakened authentication | 5.0% | No direct hunk selected |
| Weakened authorization | 10.0% | No direct hunk selected |
| Contract regression | 12.0% | No direct hunk selected |
| Data loss | 65.0% | [F003H004 · scripts/dev-static.mjs:403–423](https://github.com/OpenHands/OpenHands/blob/931f719beb4db7906abaa7e767d72c689db73c44/scripts/dev-static.mjs#L403-L423) |
| Sensitive data disclosure | 5.0% | No direct hunk selected |
| Credential misuse | 6.0% | No direct hunk selected |
| Unverified remote execution | 3.0% | No direct hunk selected |
| Privileged environment access | 45.0% | [F003H004 · scripts/dev-static.mjs:403–423](https://github.com/OpenHands/OpenHands/blob/931f719beb4db7906abaa7e767d72c689db73c44/scripts/dev-static.mjs#L403-L423) |
| Security assessment bypass | 4.0% | No direct hunk selected |
| Primary concern | Data loss; confidence 93.0% | [F003H004 · scripts/dev-static.mjs:403–423](https://github.com/OpenHands/OpenHands/blob/931f719beb4db7906abaa7e767d72c689db73c44/scripts/dev-static.mjs#L403-L423) |

</details>

### DIFF
--- a/__tests__/scripts/dev-migration-watch.test.ts
+++ b/__tests__/scripts/dev-migration-watch.test.ts
@@ -1,0 +1,154 @@
+import { EventEmitter } from "node:events";
+import { mkdtempSync, rmSync, writeFileSync, existsSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { afterEach, describe, expect, it } from "vitest";
+
+import { watchAutomationMigration } from "../../scripts/dev-process-utils.mjs";
+
+type LogKind = "warn" | "ok" | "error";
+
+interface FakeProc extends EventEmitter {
+  stdout: EventEmitter;
+  stderr: EventEmitter;
+}
+
+/** Minimal fake of a spawned child: EventEmitters for stdout/stderr/self. */
+function makeFakeProc(): FakeProc {
+  const proc = new EventEmitter() as FakeProc;
+  proc.stdout = new EventEmitter();
+  proc.stderr = new EventEmitter();
+  return proc;
+}
+
+const MIGRATION_LOG_LINE =
+  'alembic.runtime.migration: Can\'t locate revision identified by "abc123"';
+
+describe("watchAutomationMigration", () => {
+  it("recovers (deletes the DB) when the migration error and exit code 3 line up", () => {
+    const proc = makeFakeProc();
+    let recovered = false;
+    watchAutomationMigration(proc, {
+      dbPath: "/tmp/fake-automations.db",
+      onRecover: () => {
+        recovered = true;
+      },
+      log: () => {},
+    });
+
+    proc.stdout.emit("data", Buffer.from(MIGRATION_LOG_LINE));
+    proc.emit("exit", 3);
+
+    expect(recovered).toBe(true);
+  });
+
+  it("does not delete the DB when the exit code is not 3", () => {
+    const proc = makeFakeProc();
+    let recovered = false;
+    watchAutomationMigration(proc, {
+      dbPath: "/tmp/fake-automations.db",
+      onRecover: () => {
+        recovered = true;
+      },
+      log: () => {},
+    });
+
+    proc.stdout.emit("data", Buffer.from(MIGRATION_LOG_LINE));
+    proc.emit("exit", 1);
+
+    expect(recovered).toBe(false);
+  });
+
+  it("does not delete the DB when the log pattern never appeared", () => {
+    const proc = makeFakeProc();
+    let recovered = false;
+    watchAutomationMigration(proc, {
+      dbPath: "/tmp/fake-automations.db",
+      onRecover: () => {
+        recovered = true;
+      },
+      log: () => {},
+    });
+
+    // A coincidental exit 3 from some other cause must never delete the DB.
+    proc.emit("exit", 3);
+
+    expect(recovered).toBe(false);
+  });
+
+  it("latches after one recovery attempt so a broken backend cannot loop", () => {
+    const proc = makeFakeProc();
+    let recoverCount = 0;
+    const logs: string[] = [];
+    watchAutomationMigration(proc, {
+      dbPath: "/tmp/fake-automations.db",
+      onRecover: () => {
+        recoverCount += 1;
+      },
+      log: (message: string) => logs.push(message),
+    });
+
+    // First failure: recovery runs.
+    proc.stdout.emit("data", Buffer.from(MIGRATION_LOG_LINE));
+    proc.emit("exit", 3);
+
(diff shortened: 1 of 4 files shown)


hudsonwa:
Hi! Thanks for taking the time to review this so thoroughly, the feedback is spot on, especially the restart-loop scenario. A fix that wedges `npm run dev` in a silent respawn cycle would be worse than no fix at all.

I've pushed a rework addressing all three points:

1. **Unbounded loop** — the watcher now carries a single-shot latch. One recovery attempt per launcher run; if the restarted backend fails the same way again it logs loudly and gives up rather than respawning forever.

2. **Duplicated logic** — good catch on the header comment contradicting itself. The detection + latch + delete behaviour moved into a shared `watchAutomationMigration()` helper in `scripts/dev-process-utils.mjs`, and both launchers now call it instead of carrying their own copies.

3. **Exit-code coupling** — the exit-3 check now also requires the migration error pattern in the output, so an unrelated process exiting with code 3 can never trigger a DB deletion.

Also added six regression tests (`__tests__/scripts/dev-migration-watch.test.ts`) that drive the full lifecycle with fake streams: recovery on matching pattern+code, no recovery on either signal alone, the latch holding on repeat failures, error reporting instead of throwing, plus a real-filesystem integration case.

Local runs: all 237 script-suite tests pass (18 files), lint clean apart from one pre-existing warning in `use-local-git-info.ts` that predates this branch.

Happy to adjust if anything still doesn't look right.

hudsonwa:
Quick follow-up on the restart loop.

I had another look after the last push and the latch was still wrong. It lived on a single watcher instance, but recovery respawns the backend, which installs a *new* watcher. That new watcher would have recovered again, so the loop the review asked us to close was still open. Both launchers also restarted on any exit code 3, even when the migration log never appeared.

Pushed a fix: the latch is now process-wide, and restart only happens inside `onRecover` after a successful delete. Added a test that installs a second watcher after recovery and asserts it gives up.

Sorry for missing that the first time.

PR #16394, opened 2026-08-07
fix: strip Secure cookie flag on plain-HTTP non-loopback hosts

hudsonwa:
HUMAN:

Workspace file preview 401s on plain-HTTP non-loopback hosts. This strips the Secure flag from the session cookie in the ingress proxy when accessed over such a host so workspace files load.

AGENT:

## Why

On plain-HTTP non-loopback hosts (container IP, VM, LAN address) the agent-server mints `oh_workspace_session_key` with `Secure`. Browsers only send `Secure` cookies on `https://` or `http://localhost`, so the cookie is dropped and the workspace file request 401s even though the rest of the UI works.

## Summary

- Added `stripSecureCookieOnPlainHttp()` to `scripts/proxy-utils.mjs` that rewrites the upstream `Set-Cookie` headers
- Drops the `Secure` flag only when the request came over plain HTTP to a **non-loopback** host
- HTTPS and loopback (`localhost`, `127.0.0.1`, `::1`) traffic is left untouched — Secure is correct there
- Wired it up as a `proxyRes` handler in `createProxyHandlers()`, which is shared by both the ingress and the static server

## Issue Number

OpenHands/OpenHands#15567

## How to Test

1. Run `npx vitest run __tests__/scripts/ingress.test.ts` — 22 tests pass
2. New tests verify: cookie keeps `Secure` on `localhost`/`127.0.0.1`, but loses `Secure` when the Host header is a non-loopback address like `192.168.1.50`

## Video/Screenshots

ingress tests passing

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Identical recovery is already handled on the agent-server side for migrations; this is purely the frontend proxy cookie fix from the issue. Only the `oh_workspace_session_key` cookie path is affected — no other cookies are rewritten.

### DIFF
--- a/__tests__/scripts/ingress.test.ts
+++ b/__tests__/scripts/ingress.test.ts
@@ -1,4 +1,9 @@
-import { createServer, request, type Server } from "node:http";
+import {
+  createServer,
+  request,
+  type IncomingHttpHeaders,
+  type Server,
+} from "node:http";
 import { connect as netConnect, type AddressInfo, type Socket } from "node:net";
 import type { Duplex } from "node:stream";
 import { spawn, type ChildProcess } from "node:child_process";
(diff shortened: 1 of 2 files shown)


hudsonwa:
Closing as a duplicate of my earlier PR #16364, which addresses the same workspace-preview Secure-cookie issue. Keeping #16364 to avoid redundant competing PRs.

PR #16396, opened 2026-08-07
fix: don't clobber skill toggle state on settings refetch

hudsonwa:
HUMAN:

Skills enable/disable toggle was reverting to enabled ~5s after being set. Fixed by hydrating disabled state once instead of re-reading server settings on every refetch.

AGENT:

## Why

The skills settings hydration effect re-ran on every settings refetch (because `settings.disabled_skills` was in its dependency array). When a refetch returned a value stale relative to an in-flight save, it overwrote the local toggle state — so a skill the user just disabled snapped back to enabled ~5 seconds later.

## Summary

- The hydration effect now hydrates the local disabled set exactly once, on first data arrival
- It no longer re-reads server settings into local state on subsequent refetches (the auto-save effect already persists the user's latest toggle, so re-hydration only risked clobbering in-flight changes)

## Issue Number

OpenHands/OpenHands#15694

## How to Test

1. Run `npx vitest run __tests__/routes/skills-settings.test.tsx` — 18 tests pass
2. Open Settings → Skills, toggle a skill off
3. Wait >5s / let settings refetch — the toggle stays off instead of reverting to enabled

## Video/Screenshots

Skills toggle stays disabled after fix

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The one-line guard (`hasHydratedInitialSettings`) also makes the component's own documented intent ("Sync local state with server settings when data first arrives") match its behavior.


VascoSch92 (approved):
THanks

hudsonwa:
Closing this one with thanks: upstream #16860 has since replaced the all-on skill catalog with an explicit allow-list and moved all of this into a shared `use-skill-enablement` hook, which handles the same hydrate-once concern this PR addressed (and goes further with the migration guard). With that landed, merging this patch would resurrect a superseded code path, so it is no longer needed.

Thank you to the maintainers for the cleaner, more general solution.

PR #16407, opened 2026-08-07
fix: prevent Add Skill modal overflow on small screens

### LINKED ISSUE #16382: Marketplace "Add Repository" dialog overflows on mobile
On mobile devices, the Add Repository dialog in Settings → Skills → Marketplaces does not fit on the screen which causes part of it to extend beyond the edge of the screen

Image
Image

iroiro147:
Fixes #16382\n\nThis makes the Add Skill modal use the available viewport width on small screens while retaining its desktop max width.\n\nlocal verification not run; relying on upstream CI

### DIFF
--- a/src/components/features/skills/add-skill-modal.tsx
+++ b/src/components/features/skills/add-skill-modal.tsx
@@ -121,7 +121,7 @@ export function AddSkillModal({ onClose }: AddSkillModalProps) {
     >
       <div
         data-testid="add-skill-modal"
-        className="relative flex w-[520px] max-w-[90vw] max-h-[85vh] flex-col rounded-xl border border-[var(--oh-border)] bg-base-secondary"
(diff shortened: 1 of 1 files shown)


iroiro147:
Thanks for the measurements — they settle it, and there is a second problem on top.

The geometry table is right: with `ModalBackdrop`'s shrink-to-fit wrapper and the non-wrapping `<pre>` example, `w-full max-w-[520px]` resolves to the full 520px and `mx-4` pushes it off both edges, while the current `w-[520px] max-w-[90vw]` already fits at every phone width. And the suggested `max-w-[90vw]` variant reproduces main's geometry exactly, which means there is nothing left for this PR to change in this file.

Re-reading #16382, the component is also wrong: the report is about the **Marketplace "Add Repository" dialog** (Settings → Skills → Marketplaces), not the Add Skill modal this PR edits. So this is a regression aimed at the wrong dialog. Closing it rather than leaving it in the queue; if I come back to #16382 it will be against the Add Repository dialog, with a measured before/after rather than a class swap.

PR #16414, opened 2026-08-07
fix: suppress mobile keyboard for select-only dropdowns

### LINKED ISSUE #16385: Opening the Sandbox Grouping Strategy dropdown opens the mobile keyboard
On mobile, tapping the Sandbox Grouping Strategy dropdown opens the on-screen keyboard even though the field is only used to select an option from a list.
The keyboard reduces the available screen space and changes the layout while the dropdown is open. If the keyboard is dismissed, the UI returns to the expected layout while the dropdown remains open.
Since this control only requires selecting an existing option, tapping it should display the dropdown list without opening the keyboard.

Image

Image

iroiro147:
Fixes #16385

Prevents the mobile keyboard from opening for select-only settings dropdowns while preserving text input for custom-value and filter-enabled dropdowns. Adds focused input-mode regression coverage.

local verification not run; relying on upstream CI

### DIFF
--- a/__tests__/components/features/settings/settings-dropdown-input.test.tsx
+++ b/__tests__/components/features/settings/settings-dropdown-input.test.tsx
@@ -94,4 +94,32 @@ describe("SettingsDropdownInput", () => {
     expect(onSelectionChange).toHaveBeenCalledWith(null);
     expect(onInputChange).toHaveBeenCalledWith("");
   });
+
+  it("suppresses the mobile keyboard when typing is not enabled", () => {
+    render(
+      <SettingsDropdownInput
+        items={items}
+        label="Language"
+        name="language"
+        testId="language-input"
+      />,
+    );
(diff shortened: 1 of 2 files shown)


iroiro147:
Closing this — @VascoSch92 closed the motivating issue #16385 as no longer
actionable, on the grounds that the Sandbox Grouping Strategy field does not
exist in the current Agent Canvas code and the legacy `frontend/` it lived in
has been removed. That reasoning is sound and it lands on this PR too.

For the record in case it is ever wanted: the one-line change here is not
legacy-specific. It sets `inputMode="none"` on `SettingsDropdownInput`, which is
current code, so it would still suppress the mobile keyboard on any select-only
dropdown built from that component. If that is worth having as a general
accessibility fix rather than a bug fix, I am happy to re-open it against a
fresh issue — but it should not sit here attached to a closed one.

PR #16415, opened 2026-08-07
fix(settings): follow pagination when listing LLM providers

georgeglarson:
HUMAN: 
Real bug. Hit it myself. Reproduced. Tested before/after.

before
before-xai-not-found

after
after-xai-found

AGENT:

Ran against OpenHands Cloud from a local production build on Linux, before and after the change, capturing the actual `/api/v1/config/providers/search` responses in the browser.

Before (unmodified main @ 7d897d766): one request, `?limit=100`, 100 items returned with `next_page_id: "MTAw"` present in the response body and never followed. Typing `xai` into the LLM Provider field matched only `MiniMaxAI`, a fuzzy hit on the letters. The rendered list ended at `reducto`.

After (this branch): two requests, 100 items with a cursor then 35 items with `next_page_id: null`. The same `xai` query now matches `MiniMaxAI`, `VertexAI` and `xai`, and the list runs to `zai-org`.

Regression tests were written against the unfixed hook and failed there before passing with the change. Full suite: 562 files passed, 4494 tests passed, 0 failures. `npm run typecheck` clean, `eslint` clean on both changed files, `npm run build` succeeds.

---

## Why

`useSearchProviders` requested a single page with `limit: 100` and returned `page.items`, dropping the `next_page_id` the server sends back. When the server has more pages, everything past the first is silently discarded.

The comment on that call reads "Providers are a small set; fetch all in one call with a high limit." On the deployment I tested that assumption no longer holds, and it fails invisibly rather than obviously, because the response saturates the requested limit exactly.

The user-visible effect is that some providers cannot be selected at all. Results come back verified-first and then alphabetically, so the loss lands on the tail of the alphabet. On OpenHands Cloud I observed `together_ai`, `vertex_ai` and `xai` absent from provider selectors backed by this hook.

This affects the cloud backend specifically. The local agent-server path in `ConfigService.searchProviders` returns `next_page_id: null` unconditionally, so it has never been able to trigger the truncation.

## Summary

- `useSearchProviders` now follows `next_page_id` until the server stops returning one, instead of keeping only the first page.
- The pagination helper mirrors the existing `useProviderModels` hook in the same directory, including its `MAX_PAGINATION_DEPTH` bound against a runaway cursor, so the two paginated config queries behave the same way.
- Adds unit coverage for cursor-following, the single-page case, and the runaway bound.

## Issue Number

None filed. Happy to open one first if you would prefer that for bug fixes.

## How to Test

Point the app at a cloud backend whose provider list exceeds one page, then open Settings, LLM, Add LLM Profile and expand the provider dropdown.

Before this change, typing `xai` returns only `MiniMaxAI` and the list stops at `reducto`. After it, `VertexAI` and `xai` both appear and the list continues to `zai-org`.

To see it at the network level, watch requests to `/api/v1/config/providers/search`: previously one request whose response carried a non-null `next_page_id` that was never followed, now a second request using that cursor.

Unit tests: `npx vitest run __tests__/hooks/query/use-search-providers.test.tsx`.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The extra request only happens when the backend actually reports another page, so single-page deployments are unaffected.

The provider counts and ordering above are what I observed on OpenHands Cloud, not invariants of every deployment. The defect itself does not depend on them: the client discards a cursor the API's own `ProviderPage` type declares, so truncation follows whenever the server sends one.


georgeglarson:
Superseded by #16453, which landed the same fix from the other direction: drop the `limit` rather than follow the cursor. On the local backend that is complete, since `limitItems` returns everything when no limit is passed and that path always reports `next_page_id: null`.

Closing rather than rebasing. The two commits here are built against the old `{ limit: 100 }` shape and the diff conflicts with what merged, so anything further belongs on top of #16453 instead of carried forward from this branch.

One loose end, recorded here so it doesn't get lost: the cloud backend calls `/api/v1/config/providers/search`, which paginates for real, and `buildCloudQueryString` omits an undefined limit. If that service applies its own default page size, the truncation survives there and the cursor is still discarded. Worth a look at the provider picker on Cloud once #16453 deploys. Happy to open a small follow-up if it comes up short.

PR #16420, opened 2026-08-08
fix: pass session API key to static-server via env instead of argv

### LINKED ISSUE #16383: Session API key passed as plaintext CLI arg to static-server.mjs (visible via ps/proc)
## Summary
`--session-api-key <key>` is passed as a literal `argv` entry to `scripts/static-server.mjs` in every place that spawns it, so the live `LOCAL_BACKEND_API_KEY` sits in `ps auxww` / `/proc/<pid>/cmdline` output for the lifetime of the process — readable by any local user on a shared host, no privilege needed.

## Repro (observed on a shared devbox)
```
$ ps auxww | grep static-server
node .../scripts/static-server.mjs --dir ... --port 3001 --session-api-key <redacted> --runtime-services-info {...} ...
```
Confirmed the value matches the live `LOCAL_BACKEND_API_KEY`. `/proc/<pid>/cmdline` is world-readable (`-r--r--r--`); by contrast `/proc/<pid>/environ` is restricted to the owning user + root, so moving the key to an env var would meaningfully shrink exposure on multi-user hosts.

## Where it's set
- `scripts/dev-with-automation.mjs:1467-1469` (`startStaticFrontend`) — the path the published `agent-canvas` CLI (`bin/agent-canvas.mjs` → `dev-with-automation.mjs`) uses.
- `scripts/dev-static.mjs:403-404`
- `docker/entrypoint.sh:259`

All three build `["--session-api-key", key]` (or the bash equivalent) and pass it straight to the child process's argv.

## Where it's read
- `scripts/static-server.mjs:87,124-125` — parses `--session-api-key` from `argv` only; there's no env var fallback today.

## Suggested fix
`spawnService()` in `dev-with-automation.mjs` (~line 574-583) already merges `options.env` into the child process env (`env: { ...process.env, ...options.env }`), so the plumbing exists. Pass the key via an env var (e.g. `SESSION_API_KEY`) instead of argv in all three call sites, and have `static-server.mjs`'s arg parser read `process.env.SESSION_API_KEY` (keeping `--session-api-key` as a fallback for back-compat, or dropping it in a major version).

## Severity
Local-only exposure — requires shell access to the host, not remotely exploitable. Still worth fixing since it's a real standing credential leak on any shared multi-user deployment (e.g. a shared devbox running several teammates' automations).

putramkti:
HUMAN:

Tested this with the dev launcher locally. The touched script suites pass (83/83), the full run is green at 4495 tests, and `ps auxww` no longer shows the session key in the process list. The key still reaches the frontend the same way it did before.

AGENT:

Stashed the fix and ran the new regression test: it fails against the old code and passes with the fix applied, so the test actually reproduces the bug.

---

## Why

The launchers passed the session API key to `static-server.mjs` as a command line flag. That puts the key in `ps auxww` and `/proc/<pid>/cmdline`, both world readable on a shared host. Anyone on the box can grab it without special access.

## Summary

- `static-server.mjs` now falls back to the `SESSION_API_KEY` env var when the flag is absent. The flag still works and wins when both are present, so nothing that relied on it breaks.
- Both launcher scripts (`dev-with-automation.mjs`, `dev-static.mjs`) send the key through the spawn env instead of argv.
- `docker/entrypoint.sh` uses an inline env assignment for the same reason.
- Public mode is untouched: `--auth-required` still forces the manual key entry screen and ignores the env var.

## Issue Number

Fixes #16383

## How to Test

1. `npm ci`
2. `npx vitest run __tests__/scripts/static-server.test.ts __tests__/scripts/dev-with-automation.test.ts __tests__/scripts/dev-static.test.ts`
3. Start `dev-static` with a persisted session key, then `ps auxww | grep static-server` and confirm the key is not in the output.
4. Start with `--auth-required` and confirm public mode still shows the key entry screen.

## Video/Screenshots

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Env vars beat argv but are still visible via `/proc/<pid>/environ` to the same user and through `docker inspect`. Reading the key from stdin or a 0600 temp file would close that gap. I left that as a follow-up rather than scope-creeping this PR.

### DIFF
--- a/__tests__/scripts/static-server.test.ts
+++ b/__tests__/scripts/static-server.test.ts
@@ -122,6 +122,52 @@ describe("static-server.mjs", () => {
       expect(config.sessionApiKey).toBeNull();
     });
 
+    it("falls back to the SESSION_API_KEY env var when no flag is given", () => {
+      const previous = process.env.SESSION_API_KEY;
+      try {
+        process.env.SESSION_API_KEY = "env-key";
+        const config = parseArgs([]);
+        expect(config.sessionApiKey).toBe("env-key");
+      } finally {
+        if (previous === undefined) {
+          delete process.env.SESSION_API_KEY;
+        } else {
+          process.env.SESSION_API_KEY = previous;
+        }
+      }
+    });
+
+    it("prefers --session-api-key over the SESSION_API_KEY env var", () => {
+      const previous = process.env.SESSION_API_KEY;
+      try {
+        process.env.SESSION_API_KEY = "env-key";
+        const config = parseArgs(["--session-api-key", "argv-key"]);
+        expect(config.sessionApiKey).toBe("argv-key");
+      } finally {
+        if (previous === undefined) {
+          delete process.env.SESSION_API_KEY;
+        } else {
+          process.env.SESSION_API_KEY = previous;
+        }
+      }
+    });
+
+    it("ignores the SESSION_API_KEY env var in public mode (--auth-required)", () => {
+      const previous = process.env.SESSION_API_KEY;
+      try {
+        process.env.SESSION_API_KEY = "env-key";
+        const config = parseArgs(["--auth-required"]);
+        expect(config.sessionApiKey).toBeNull();
+        expect(config.authRequired).toBe(true);
+      } finally {
+        if (previous === undefined) {
+          delete process.env.SESSION_API_KEY;
+        } else {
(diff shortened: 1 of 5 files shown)


putramkti:
Friendly ping — this is ready for review.

Issue #15492, opened 2026-07-10
[Feature] Connect a provider once; pick from every model that provider offers in Agent Profile


DevinVinson:
## Problem 
Users who hold multiple LLM subscriptions (Anthropic + OpenAI + Google, etc.) have to set up each model individually in Settings → LLM Profiles before any of it is usable. Even within a single vendor — say, several Claude models — each one needs its own profile, its own key entry, its own save. The friction is worst for users who manage many models from one key.

Separately, ACP agent connections (Claude Code, Codex, Gemini CLI) already unlock a bundle of models, but the picker that lists them is buried in Settings → Agent. When a user installs one of these agents, the only signal they get is "you have access to some models" — the actual list is invisible until they go hunt for it.

The shared user feeling: "I connected X, but I don't know what's now available."

## General UX
A user opens Settings → LLM Profiles (or enters the Agent Profile flow). They see a new UI that didn't exist before:

Connect a Provider
Skip the per-model setup. Connect once and pick which models from that provider to make available with automatic setup.

Clicking it opens a similar flow as add LLM as a Vendor setup Wizard (or similar UI/UX):

Step 1 — Vendor + key

- A dropdown of single-vendor providers (Anthropic, OpenAI, Google, xAI, Mistral, plus any others surfaced via /api/llm/providers).
- A masked paste field for the API key.
- A short helper line: "We'll save this once and reuse it for every model from this provider. Rotate it later in Provider connections."

Step 2 — Test connection

- Triggered automatically when the key field loses focus (or via "Test connection" — both acceptable).
- Spinner → success: green check, vendor catalog populates.
- Failure: inline error with the cause (e.g., "401 — invalid key"), with Try again and Use a different key actions. 
- If the vendor returns no models (shouldn't happen but defensively): show a "Couldn't fetch this provider's catalog" message with a retry.

Step 3 — Pick models

- A list of every model the provider offers in /api/llm/models?provider={vendor}.
- Each row: model name, an availability tag ("Recommended for OpenHands" if the model is in /verified), and a checkbox.
- Default selection: all "Recommended for OpenHands" models pre-checked; everything else unchecked.
- Bulk actions above the list: Select all verified, Select all, Clear.
- "More from {vendor}" collapsible section for less-common or experimental models.

Step 4 — Auto-update (optional, maybe a separate follow up enhancement) 

- One checkbox: "Keep my model list up to date with new releases from {vendor}".
[...]

PR #16426, opened 2026-08-08
feat(settings): bulk-add provider models as LLM profiles

georgeglarson:
HUMAN:

Offered as prototype. Scratched my itch.

AGENT:

Verified end-to-end against the live OpenHands Cloud backend, since unit tests alone would not settle it. The frontend suite passes 4509 with lint, typecheck and build clean, and the modal was then driven by hand against an account sitting at its 10-profile limit, with the resulting requests read off the wire. Commands, observations and the two paths that could not be reached live are all below.

---

## Why

Adding cloud models as profiles is one hand-add at a time, and the `openhands` provider currently lists 31 verified models. Wanting more than a couple of them means repeating the same short form over and over, which is where the friction is.

This is one way bulk-add can work, built far enough to be judged rather than imagined. If the shape is wrong, the parts are separable and probably worth more than the whole: the limit handling, the row-state model and the modal itself have no dependency on each other, so any of them can be lifted into work already in flight.

## Summary

- New "Add from provider models" entry on the LLM settings page opens a modal: pick a provider, select models, edit the proposed profile names, add them in one pass. Profiles are created keyless (`include_secrets: false`), so a key is added afterwards by editing the profile, the same as any other keyless profile.
- Creates run sequentially and survive the server's profile limit. A 409 fails its own row and the run continues, since a name can be taken between load and submit; a second 409 stops the run and shows the server's own message. Rows the run never reached stay idle, with no spinner left running on a request that never happened.
- Nothing is pre-selected. The server caps how many profiles an account may hold and exposes no way to read that cap, so defaulting to all-checked invites a submission that is mostly refusals. This is also what lets the modal stay correct without knowing the number.
- Providers are grouped verified / others, matching how `model-selector.tsx` presents the same feed. The feed carries entries that are not providers (`1024-x-1024`, `low`, `medium`), which are LiteLLM cost-map keys split on the first `/`.

## Issue Number

Related to #15492.

## How to Test

Unit and build, from the repo root:

```
npm ci
npm test          # 4509 passed, 0 failed
npm run lint
npm run build
```

The modal's own tests are `__tests__/components/settings/llm-profiles/add-models-modal.test.tsx`. Each behaviour has a test that fails without its fix.

End-to-end against a real backend:

```
npm ci && npm run build
node scripts/static-server.mjs     # serves build/ on :3001
```

Sign in, open Settings, LLM, then "Add from provider models" and choose a provider. Selecting a few models and submitting creates one profile per row, each appearing in Available Profiles.

What this run showed against an account already at 10 of 10 profiles, where every create is refused:

- Three models selected. The first two were attempted and marked Failed, and the third was never requested. Confirmed on the wire, not by eye: two `POST` calls to `/api/organizations/{org}/profiles/{name}` per submit, and the third row's name absent from every one of them.
- The error toast carried the server's own sentence: `Profile limit reached (10). Delete a profile before saving a new one.`
- No row was left spinning, and the account still held the same 10 profiles afterwards.
- Renaming a row before submitting sent the edited name on the request, and toggling the "Verified models only" filter after selecting and renaming left both the selection and the edited name intact.
- Closing and reopening the modal gave a clean one, with no provider and no leftover row statuses.

## Video/Screenshots

modal-409-two-attempts

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Two code paths are covered by tests but could not be executed live, and are worth knowing before review leans on the end-to-end evidence. A 409 carrying no readable `detail` falls back to a generic message, which cloud never produces because it always sends one. And a create succeeding after an earlier 409 cannot happen at the limit, since nothing can succeed there; the continuation itself did run live, issuing a second request where the earlier behaviour would have stopped at one.

All live testing was against cloud. The local agent-server path shares the same services layer but has no end-to-end evidence here.

The provider dropdown reads the first 100 entries only, which is a pre-existing limit in `useSearchProviders` and the subject of #16415. This branch does not include that fix, so on a deployment with more than 100 provider keys some will not appear in the dropdown until it lands.

Grouping models by family, so a capped account can be filled with the best of each instead of alphabetically, is deliberately left out. It is an opinion about curation and it can be argued separately from whether the mechanism is right.


georgeglarson:
The `mock-llm-e2e` failure looks unrelated to this change, best I can tell.

What failed is `files/mock-llm-files-and-git.spec.ts` step 2, waiting 30s for `chat-interface` to become visible after a page reload. Running that spec locally against this commit passes all six steps:

```
uv venv .mock-llm-venv && uv pip install -p .mock-llm-venv "openhands-sdk==1.40.1"
MOCK_LLM_PYTHON=$PWD/.mock-llm-venv/bin/python npx playwright test \
  tests/e2e/mock-llm/files/mock-llm-files-and-git.spec.ts
```

The diff also has no path to that route. Everything it adds compiles into the `llm-settings` chunk, which is fetched when someone navigates to Settings, LLM. That test starts a conversation from the home launcher and reloads a conversation page, so none of the new code is loaded when the assertion runs. The only always-loaded change is one added i18n key.

Re-running the job isn't something I can trigger without write access. It fires on `synchronize`, so the next push to the branch will re-run it if that is easier than a dispatch.

juanmichelini:
@georgeglarson thanks for the PR! at the moment we are considering what's the best UI to allow adding new models. Will let you know once we have a path forward

juanmichelini:
@georgeglarson could you change the combobox to pick one of the avaialble connections insteads of a provider?
So the usage would be set a connection first, then bulk add on that connection. That way we only need to change the api key in one place

image

juanmichelini:
@all-hands-bot please rereview considering that combox should use an existing provider connection not a single provider

PR #16436, opened 2026-08-08
refactor(frontend): replace xterm.js with rioterm in the terminal tab

raphamorim:
HUMAN:

This swaps the terminal tab from xterm.js to [rioterm](https://github.com/raphamorim/riotermjs), which is Rio terminal's Rust VT engine compiled to WebAssembly (full disclosure: I'm the author). The terminal here is a read-only command log, and most of the xterm-specific code existed to work around xterm itself: the `canFitTerminal()` guard for FitAddon's crash on hidden containers, and ~90 lines of `.xterm-*` CSS fighting for transparency and wrapping. All of that is deleted. I went with rioterm's DOM renderer, so the log is real text: screen readers, browser find, and translation just work, with no screenReaderMode flag, and `background: 'transparent'` is plain CSS.

Performance is not the main point for a display-only log, but for the record: on byte-identical workloads against xterm.js with its WebGL addon, rioterm initializes 1.9x faster (21.6ms vs 41.1ms), parses and paints plain text 3.7x faster (657 vs 179 MB/s) and ANSI-colored output 2.2x faster (237 vs 109 MB/s), with headless parsing up to 6.8x faster. Suite and caveats live in [benchmark/](https://github.com/raphamorim/riotermjs/tree/main/benchmark) if you want to rerun them. Happy to adjust anything, and no hard feelings if you'd rather not take a dependency maintained by its vendor.

AGENT:

No agent-generated changes to report; sections below kept for the template checks.

---

## Why

See the note above: delete xterm workaround code (FitAddon crash guard, CSS override block), get an accessible real-text terminal surface for a read-only log, keep behavior.

## Summary

- `use-terminal.ts` opens a rioterm DOM-renderer terminal (async wasm init via dynamic import); command-replay behavior unchanged.
- Deleted the `.xterm-*` CSS overrides, the xterm CSS import, and the fit guard machinery.
- `@xterm/xterm` + `@xterm/addon-fit` replaced by `rioterm` (exact-pinned, per the package-library policy test).

## Issue Number

N/A

## How to Test

1. `npm install && npm run dev:mock`
2. Open a conversation, switch to the Terminal tab: command history renders with the `$ ` prompt convention, transparent background, hidden cursor.
3. Switch tabs away and back: history replays; hiding the tab does not reflow the grid.
4. Resize the panel: the grid refits.

Evidence from commands run on this branch:

- `npx vitest run __tests__/hooks/use-terminal.test.tsx __tests__/components/terminal/terminal-empty-state.test.tsx`: 7/7 pass.
- `npm run typecheck`: clean.
- Full `npx vitest run`: 4363 passed; the few failing files fail identically on `main` with this branch stashed (verified), so they are unrelated.

## Video/Screenshots

The real `useTerminal` hook rendered with the real wasm engine and a seeded command store (standalone harness, since mock mode ships no bash events; the rendering path is exactly this PR's code):

OpenHands terminal tab rendered by rioterm's DOM renderer: styled build/test output, transparent background, hidden cursor


neubig:
Hi @raphamorim thanks for introducing your work!
I think we'll probably just stick with xterm for now, since it's a bigger project and we'd have more confidence in it continuing to be maintained, etc. If you are an openhands user and could join our slack and our community could get more familiar with you, we might consider this more in the future though!

raphamorim:
Super fair and understandable! thanks for the reply. Did the PR also to experiment the api and speed when i was rolling it out at Lovable. Again, I appreciate the reply!

PR #16438, opened 2026-08-08
fix: enable bash-events retention by default so the agent-server doesn't pin CPU globbing an unbounded events dir

### LINKED ISSUE #16437: Agent server pegs CPU with several conversations: unbounded bash_events dir makes search_bash_events glob 100k+ files on the event loop
## Summary

When several conversations are active, the agent-server process climbs to 100% CPU and everything (all conversations, the UI, API requests) slows down dramatically.

## Root cause

Profiling the hot agent-server process with `py-spy` shows ~95% of GIL time inside `glob`/`fnmatch`, called from the agent-server's bash events search endpoint:

```
Thread (active+gil): "MainThread"
    filter (fnmatch.py:56)                                  <- 70.6% of GIL samples
    _glob1 / _iglob / glob (glob.py)                        <- ~25%
    _get_event_files_by_pattern (agent_server/bash_service.py:81)
    search_bash_events (agent_server/bash_service.py:137)
    search_bash_events (agent_server/bash_router.py:63)
    ... uvicorn/asyncio event loop ...
```

The agent-server stores every terminal command/output as an individual file under `OH_BASH_EVENTS_DIR`, and `search_bash_events` runs a synchronous `glob.glob()` over the whole directory **on the asyncio event loop thread** for every search request. Each open conversation's terminal tab polls this endpoint, so several concurrent conversations means back-to-back full-directory scans that block the event loop and starve everything else.

The directory grows without bound: the launchers (`scripts/dev-safe.mjs`, `docker/entrypoint.sh`) set `OH_BASH_EVENTS_DIR` but never set `bash_events_retention_seconds`, and the agent-server's default for that setting is `None` (retain forever). On my long-running install the directory had accumulated **165,958 files / 860 MB**, only ~3,700 of which were from the last 24 hours. Purging files older than 2 days immediately dropped the process off 100% CPU.

## Proposed fix (this repo)

Enable the agent-server's built-in rolling retention cleanup by having the launchers pass `OH_BASH_EVENTS_RETENTION_SECONDS` with a sane default (e.g. 2 days), overridable via env var. This covers the npm/dev launchers and the Docker image.

(Verified that the agent-server's `load_config()` picks the value up from the `OH_`-prefixed env var and starts the `run_retention_cleanup_loop` background task when it's set.)

## Upstream follow-up (software-agent-sdk)

Independent of retention, `search_bash_events` should avoid an O(N)-files synchronous glob on the event loop — e.g. maintain an index, use prefix filtering on `os.scandir`, or at minimum offload the scan to a thread-pool executor. Retention keeps N small in practice, but the endpoint remains a foot-gun for long retention windows.

---

_This issue was created by an AI agent (OpenHands) on behalf of rbren, based on a live profiling session._

rbren:
HUMAN:
my agent identified this as a major cause of my agent-server starting to crawl. Fixing it locally did indeed seem to drastically improve things 

This is on a ~2-3 month old installation, so I might be at the frontier of where this problem kicks in

Full disclosure, I don't fully understand the implications of cleaning up the bash history. Maybe someone here knows more?

AGENT:

This PR was created by an AI agent (OpenHands) on behalf of rbren, based on a live py-spy profiling session on an affected long-running install. Evidence of end-to-end behavior is in "How to Test" below.

---

## Why

Fixes #16437.

With several conversations active, the agent-server climbs to 100% CPU and everything slows down. `py-spy` shows ~95% of GIL time in `glob`/`fnmatch` under `agent_server/bash_service.py::search_bash_events` — a synchronous full-directory glob on the asyncio event loop, run for every bash-events search request (each open conversation's terminal tab polls it).

The launchers set `OH_BASH_EVENTS_DIR` but never a retention window, and the agent-server's `bash_events_retention_seconds` defaults to `None` (retain forever), so the directory grows without bound. The affected install had accumulated **165,958 files / 860 MB**; purging files older than 2 days immediately dropped the process off 100% CPU.

## Summary

- Add `retention.bashEventsSeconds` (172800 = 2 days) to `config/defaults.json` as the single source of truth
- `scripts/dev-safe.mjs`: resolve `bashEventsRetentionSeconds` in `buildSafeDevConfig` (env `OH_BASH_EVENTS_RETENTION_SECONDS` overrides the default) and pass it through `buildAgentServerEnv`, which covers `dev`, `dev:static`, `dev:minimal`, the extra-backend launcher, and the published npm binary
- Docker: emit `CONFIG_BASH_EVENTS_RETENTION_SECONDS` in the config-gen stage and export `OH_BASH_EVENTS_RETENTION_SECONDS` in `entrypoint.sh` (env override wins)

## Issue Number

#16437

## How to Test

Unit tests (2 new cases covering the default and the env override):

```
npx vitest run __tests__/scripts/dev-safe.test.ts
  Test Files  1 passed (1)   Tests  54 passed (54)
npx vitest run __tests__/scripts/dev-with-automation.test.ts __tests__/scripts/dev-static.test.ts __tests__/scripts/dev-extra-backend.test.ts __tests__/scripts/check-sdk-version-sync.test.ts
  Test Files  4 passed (4)   Tests  71 passed (71)
```

End-to-end plumbing verification — the launcher config produces the env var, and the real agent-server (1.39.1, the pinned version) parses it into `bash_events_retention_seconds`, which `agent_server/api.py` uses to start `run_retention_cleanup_loop`:

```
$ node -e "import('./scripts/dev-safe.mjs').then(({buildSafeDevConfig, buildAgentServerEnv}) => { const env = buildAgentServerEnv(buildSafeDevConfig(process.cwd(), process.env)); console.log(env.OH_BASH_EVENTS_RETENTION_SECONDS); })"
172800
$ <agent-server venv>/bin/python -c "import os; os.environ['OH_BASH_EVENTS_RETENTION_SECONDS']='172800'; from openhands.agent_server.config import load_config; print(load_config().bash_events_retention_seconds)"
172800
```

Live impact on the affected machine: purging bash_events files older than 2 days (what the retention loop will now do continuously) took the directory from 165,999 files / 860 MB to 10,085 files / 64 MB and the agent-server main thread from pegged `active+gil` in `glob` to `idle` in `selectors.select`.

Also verified `bash -n docker/entrypoint.sh` and that `config/defaults.json` still parses.

## Video/Screenshots

Launcher/infra change with no UI surface — the screenshot below shows the terminal verification (unit tests, env plumbing, agent-server config parsing, and live before/after impact):

Verification: 125 tests pass, env var reaches the agent-server config, live CPU impact

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The default is deliberately conservative (2 days ≫ any command timeout, per the agent-server config docstring's "≥ 2x max command timeout" guidance).
- Users can override or effectively disable via `OH_BASH_EVENTS_RETENTION_SECONDS` (any positive integer).
- Existing oversized `bash_events` dirs are drained automatically by the agent-server's rolling cleanup task once this ships — no migration needed.
- Upstream follow-up tracked in the issue: `search_bash_events` in `software-agent-sdk` should avoid the O(N) synchronous glob on the event loop regardless of retention.

---

| Component | Value |
|---|---|
| **Commit** | `ed455afca935984fdab20265d039d7d9b3422267` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-ed455af
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-ed455af
```

```
ghcr.io/openhands/agent-canvas:sha-ed455af-amd64
ghcr.io/openhands/agent-canvas:fix-bash-events-retention-amd64
ghcr.io/openhands/agent-canvas:pr-16438-amd64
ghcr.io/openhands/agent-canvas:sha-ed455af-arm64
ghcr.io/openhands/agent-canvas:fix-bash-events-retention-arm64
ghcr.io/openhands/agent-canvas:pr-16438-arm64
ghcr.io/openhands/agent-canvas:sha-ed455af
ghcr.io/openhands/agent-canvas:fix-bash-events-retention
ghcr.io/openhands/agent-canvas:pr-16438
```

- Each tag (e.g., `sha-ed455af`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-ed455af-amd64`) are also available if needed


enyst (approved):
Fun stuff, thank you!

smolpaws:
Nice fix — flagging what this does and does not touch, since the "cleaning up bash history" implication came up:

**What `bash_events/` actually is:** the agent-server persists every command sent through its *direct bash API* (`POST /api/bash/execute_bash_command`) as `BashCommand`/`BashOutput` files in a single global `bash_events/` dir. On Canvas, most of these aren't the agent or a human typing in a terminal — they're the **UI's own housekeeping probes**: the git repo/branch chip (`use-local-git-info.ts` → `git remote get-url origin; git rev-parse --abbrev-ref HEAD`) and the file-explorer tree scan (`use-workspace-files.ts` → a big `find`), which fire per conversation and on refetch/poll. So the dir grows every time you open or refresh a conversation, unbounded.

**Why it pins CPU:** `search_bash_events` does a synchronous full-dir `glob`/`fnmatch` on the asyncio event loop for *every* bash-events search request (each open terminal tab polls it). O(files) per request × unbounded files → the 100% CPU you profiled.

**What this PR changes:** it just turns on the agent-server's *existing* retention/cleanup mechanism by default (2 days), which the launchers previously left at `None` (retain forever). It caps the dir size; it doesn't touch the search code path itself.

**Safe to age out — these are not conversation events.** Worth stating plainly for the "I don't fully understand the implications" note: aging out `bash_events/` does **not** delete agent conversation history. Agent actions/observations live in `conversations/<id>/events/` and are untouched. `bash_events/` is disposable direct-bash/terminal-scrollback history (dominated, on Canvas, by failed UI probes — lots of `exit_code: -1, "No such file or directory: 'workspace/project'"`). Losing the last 2 days' worth costs nothing beyond older terminal scrollback.

One follow-up worth noting (not a blocker): retention caps the symptom, but `search_bash_events` is still a sync glob on the event loop. If someone sets a very long retention, the CPU issue can return — a later change could make that search async/indexed. Shipping this first is the right 90% fix.

neubig (changes requested):
I think maybe there's a nicer way to fix this by, e.g., grouping the bash events by convresation, let me propose an alternative.

neubig:
This might be a replacement: https://github.com/OpenHands/software-agent-sdk/pull/4481

neubig:
I think this was fixed elsewhere, and it's not clear that deleting these events is safe. Closing for now but we can revisit later

PR #16441, opened 2026-08-08
feat(settings): expose the LLM-switching toggle in Agent settings

georgeglarson:
HUMAN:

As highlighted below, the bigger question might be whether this should default to on.
This PR leaves it on as not to change default behavior without explicit decision.

AGENT:

Frontend-only. The SDK already exports `enable_switch_llm_tool` in the agent settings schema with GUI metadata ("Enable LLM switching tool", general section, pinned by `tests/sdk/test_settings.py`); the Canvas frontend renders no page for that section, so the field has never been reachable. This adds the control, no SDK change.

Evidence: full suite green (562 files, 4513 tests), `npm run typecheck`, eslint and prettier clean. Reproduced in the UI under `npm run dev:mock` — the screenshots below are that run, including the version-gated case where the control correctly hides itself.

---

## Why

The built-in `switch_llm` tool lets the agent move a conversation to any saved LLM profile on its own initiative. It ships enabled by default (`enable_switch_llm_tool: bool = Field(default=True)`, wired unconditionally into the default toolset) and under the default `NeverConfirm` policy it executes with no user approval. A user can end up on a model they never selected.

The SDK already exposes the off switch. Canvas just never rendered it, so there is currently no way to turn the tool off from the GUI.

## Summary

- Render `enable_switch_llm_tool` as a `SettingsSwitch` in the Agent settings form, next to the sub-agents toggle, following the existing `enable_sub_agents` / `tool_concurrency_limit` patterns.
- Guard the profile-editor emission on the agent-server version, not just on schema presence. The field reached `AgentSettingsConfig` in 1.22.0 but `OpenHandsAgentProfile` only in 1.31.0 (software-agent-sdk#3870). Agent profiles themselves shipped in 1.29.0 and `config/defaults.json` accepts back to 1.28.0, so 1.29.0–1.30.x sit inside the supported range while advertising the field in the schema and rejecting it on the profile model. `AgentProfileBase` is `extra="forbid"`, so emitting the key there 422s the whole profile save. On those versions the editor hides the control rather than offering one the server will refuse.
- Serve `/api/agent-profiles` in the MSW mocks. Settings → Agent is the agent-profile library (#1571) and the embedded editor behind it is the only surface that renders `AgentSettingsScreen`; without handlers the page shows "Failed to load profiles" under `dev:mock`, so this form was unreachable there. This is what made the screenshots below possible.

Notes on scope and behaviour:

- The non-embedded branch of `agent-settings.tsx` is currently unreachable — #1571 turned `/settings/agent` into a redirect to the profile library, so `AgentSettingsScreen` only ever renders embedded. The field is wired into that branch anyway, to stay symmetric with `enable_sub_agents` and `tool_concurrency_limit` rather than introduce an asymmetry. Happy to drop it if you would rather that path not grow.
- The pinned ts-client (1.36.1) does not carry `enable_switch_llm_tool` on `OpenHandsAgentProfile`, so it rides with a local cast and a comment, the same precedent as `disabled_skills`. A ts-client bump would remove that.
- An unreadable agent-server version counts as supported, matching the other version-gated call sites: cloud exposes no `/server_info` to probe, and a local backend whose version does not parse never gets past the boot-time compatibility check.
- Adding the agent-profiles mocks surfaced one test that had been passing for the wrong reason: `llm-not-configured-banner`'s ACP case relied on the profile-list request failing, which left `useLlmConfigured` on its `agent_settings` fallback. The active agent profile is what actually decides that branch, so the case now sets it explicitly.

## Issue Number

Related to #16442, which covers the part this PR does not: `switch_llm` stays default-on under `NeverConfirm`, so a user who never finds this toggle still gets an agent that can move their conversation. Deliberately not `Fixes`, since that design question outlives this change.

## How to Test

```
npx vitest run \
  __tests__/routes/agent-settings.test.tsx \
  __tests__/routes/build-agent-profile-fields.test.ts \
  __tests__/api/agent-profiles-service/profile-field-support.test.ts \
  __tests__/components/settings/agent-profiles/ \
  __tests__/components/home/llm-not-configured-banner.test.tsx
```

In the UI, with no backend needed:

```
npm ci
npm run dev:mock
```

Settings → Agent → open the actions menu on a profile → Edit. The control is "Let the agent switch LLM profiles", between the sub-agents toggle and Parallel tool calls. Turning it off removes `SwitchLLMTool` from the agent's default toolset, so the agent can no longer move the conversation on its own.

To see the version gate, set `MOCK_AGENT_SERVER_VERSION` in `src/mocks/settings-handlers.ts` to `1.29.3` and reload: the control disappears and the rest of the form is untouched.

## Video/Screenshots

The control in the Agent-profile editor, on (the default).
2-toggle-on
---
Turned off.
3-toggle-off
---
The same editor against an agent-server reporting 1.29.3, where the profile model predates the field: the control is gone and nothing else changes.
4-hidden-on-old-server
---

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The more interesting conversation is arguably the default rather than the toggle: whether `switch_llm` should be default-on, or gated behind confirmation regardless of the global policy. This PR does not touch the default — it only makes the existing setting reachable, so users have a kill switch either way.


hieptl (approved):
Thank you! 🙏

PR #16443, opened 2026-08-08
fix(test): install the ProgressEvent fallback under jsdom, where it is needed

georgeglarson:
HUMAN:

I had some flakiness and this was the likely suspect.

AGENT:

Test-infrastructure only. No product code, no runtime behaviour change; `src/test-support/` is not in `tsconfig.lib.json`'s include list, so nothing here reaches the published package.

Evidence: the new test fails on `main` and passes with this change. Full suite green afterwards (562 files, 4496 tests), `npm run typecheck`, eslint and prettier clean. Nothing outside `vitest.setup.ts` referenced `ProgressEvent`.

---

## Why

`test-and-build` can fail with every test passing:

```
Vitest caught 1 unhandled error during the test run.
ReferenceError: ProgressEvent is not defined
 ❯ createEvent node_modules/msw/node_modules/@mswjs/interceptors/lib/node/XMLHttpRequest-Dw6Wm-UU.mjs:91:55
 ❯ XMLHttpRequestController.trigger ...
 Test Files  561 passed | 1 skipped (562)
      Tests  4503 passed | 5 skipped | 9 todo (4517)
```

MSW's XMLHttpRequest interceptor decides once, at module load, whether the runtime has `ProgressEvent`, then resolves the class per event:

```js
const SUPPORTS_PROGRESS_EVENT = typeof ProgressEvent !== "undefined";
// ...later, per event:
const ProgressEventClass = SUPPORTS_PROGRESS_EVENT ? ProgressEvent : ProgressEventPolyfill;
```

The capability check is cached; the class is not. Under `environment: "jsdom"` the check runs while jsdom's `ProgressEvent` exists, so the interceptor commits to the supported branch for the life of the worker. If a late interceptor callback fires while Vitest is tearing the environment down between files, that bare `ProgressEvent` no longer resolves and the run fails on an unhandled rejection.

`vitest.setup.ts` already carried a fallback for exactly this, and its comment names the case: "MSW's XMLHttpRequest interceptor may dispatch progress events while Vitest is tearing down globals between files." It was wrapped in:

```ts
if (typeof ProgressEvent === "undefined") { ... }
```

jsdom defines `ProgressEvent`, so that condition is false in the only environment this suite runs in. The fallback installed itself only where it was not needed, and never where it was.

## Summary

- Install the fallback unconditionally, extracted to `src/test-support/progress-event-fallback.ts` so it can be tested directly.
- Add tests: that the fallback is the class actually on the test global (this one fails without the change), that it installs over an existing `ProgressEvent`, and that it carries the `ProgressEventInit` fields the interceptor constructs, with the right defaults.

## Issue Number

None.

## How to Test

```
npx vitest run __tests__/test-support/progress-event-fallback.test.ts
```

On `main` the first case fails, since the guard leaves jsdom's `ProgressEvent` in place. With this change all four pass.

## Video/Screenshots

No rendering change; this is test infrastructure. The check requires an image on any frontend-touching diff, so here is the red-before-green instead.

1-fails-on-main

2-passes-with-fix

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Worth being precise about the claim: the dead guard is deterministic and provable, but the teardown race that exploits it is order-dependent and I could not reproduce it on demand. The same tree passed twice locally and failed once in CI. So this removes a cause rather than demonstrably ending the symptom.

Replacing jsdom's `ProgressEvent` during tests looked worth checking rather than assuming: nothing outside the setup file referenced it, and the class covers the whole `ProgressEventInit` surface, so the full suite behaves identically.

If you would rather keep jsdom's implementation while the environment is alive and only fall back afterwards, I am happy to rework it. That needs a binding the environment teardown does not remove, which is a bigger change than this one.

### DIFF
--- a/__tests__/test-support/progress-event-fallback.test.ts
+++ b/__tests__/test-support/progress-event-fallback.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from "vitest";
+import {
+  FallbackProgressEvent,
+  PROGRESS_EVENT_FALLBACK,
+  installProgressEventFallback,
+} from "#/test-support/progress-event-fallback";
+
+describe("ProgressEvent fallback", () => {
+  it("is the class actually installed on the test global", () => {
+    // The regression this guards: the fallback used to be wrapped in
+    // `if (typeof ProgressEvent === "undefined")`, which is false under jsdom.
+    // It therefore installed only where it was not needed, leaving MSW's
+    // interceptor to resolve a binding that environment teardown removes.
+    expect(
+      (globalThis.ProgressEvent as unknown as Record<symbol, unknown>)[
+        PROGRESS_EVENT_FALLBACK
+      ],
+    ).toBe(true);
+  });
+
+  it("installs over an existing ProgressEvent rather than deferring to it", () => {
+    class SomeoneElsesProgressEvent extends Event {}
+    const target = {
+      ProgressEvent: SomeoneElsesProgressEvent,
+    } as unknown as typeof globalThis;
+
+    installProgressEventFallback(target);
+
+    expect(target.ProgressEvent).toBe(FallbackProgressEvent);
+  });
+
+  it("carries the ProgressEvent fields MSW's interceptor constructs", () => {
+    // The interceptor builds these with `{ lengthComputable, loaded, total }`.
+    const event = new FallbackProgressEvent("progress", {
+      lengthComputable: true,
+      loaded: 17,
+      total: 42,
+    });
+
+    expect(event).toBeInstanceOf(Event);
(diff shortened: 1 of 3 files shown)


georgeglarson:
Superseded by #16504, which landed the same root cause. Closing this one.

One note for whoever hits this next: #16504 ships no test. The 1000-iteration repro lives in its description rather than the repo, so the wrong guard this started from (the `typeof ProgressEvent === "undefined"` check that never fires under jsdom) can come back quietly. Happy to put up a small regression test against the getter implementation if that's useful.

PR #16444, opened 2026-08-09
fix: always show all workspace folders in grouped view regardless of pagination

### LINKED ISSUE #16435: [Bug]: Show all workspaces when conversations are grouped by workspace
### Operating System

macOS

### Installation Method

### Agent Canvas Version

1.12.0

### Bug Description

When I'm grouping by workspace, I have to click "load more" 100 times to get an old workspace to show up

I'd rather click on the workspace I'm interested in and see the convos that were there

### Steps to Reproduce

* have a convo in workspace A
* have 100 convos in workspace B

### Actual Behavior

* need to load all convos in workspace B before workspace A shows up in the list

### Expected Behavior

Both workspace A and workspace B are always visible (maybe with 0 convos)

Clicking workspace B expands convos in that workspace

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

mohitdebian:
HUMAN:
  ran the tests locally, all 4496 passed, workspace folders now appear without needing to
  load more pages

AGENT:

Ran `npx vitest run __tests__/components/features/conversation-panel/` — 237/237 tests passed including 4 new tests added for this fix. Ran `npx eslint` on all 3 modified files — 0 errors. Pre-commit hooks (typecheck, lint, translation completeness) all passed on commit.

---

## Why

When conversations are grouped by workspace, workspace folders only appeared once their conversations had been loaded via pagination. If workspace A had all its conversations buried behind 100 conversations in workspace B, the user had to click "Load more" many times before workspace A ever showed up as a folder in the sidebar. The expected behavior is that all workspaces are always visible as folder rows regardless of how many conversations have been paginated so far.

## Summary

- `groupConversations` now accepts an optional `knownWorkspaces` list and pre-seeds the group bucket map before iterating conversations, so workspaces with zero loaded conversations still render as folder rows
- `conversation-panel.tsx` builds `allWorkspacesForGrouping` by merging the persisted workspace store with workspace paths found across all already-loaded conversation pages (not just the filtered/visible subset), then passes it to `groupConversations`
- Added `hasVisibleGroups` guard to `showEmptyState` so the empty-state message does not replace the workspace folder list when conversations are filtered out

## Issue Number

Fixes #16435

## How to Test

1. Start the app with a local backend
2. Create a conversation in workspace A
3. Create 100+ conversations in workspace B (or use an existing backend with many conversations)
4. Enable grouped-by-workspace view in the conversation panel
5. Without this fix: workspace A does not appear until you load enough pages to surface its conversations
6. With this fix: both workspace A and workspace B appear immediately as folder rows; workspace A shows 0 conversations initially and populates as pages load

Alternatively run the unit tests:
```
npm run make-i18n
npx vitest run __tests__/components/features/conversation-panel/conversation-panel-list-helpers.test.ts
```

## Video/Screenshots

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The fix intentionally collects workspace paths from all already-loaded conversation pages rather than only the currently visible (filtered) subset. This means a workspace that appears on page 3 will be pre-seeded as soon as any page containing its conversations has been fetched, even if those conversations are currently hidden by the automation or scope filters.

### DIFF
--- a/src/components/features/conversation-panel/conversation-panel.tsx
+++ b/src/components/features/conversation-panel/conversation-panel.tsx
@@ -6,6 +6,7 @@ import { useNavigation } from "#/context/navigation-context";
 import { useActiveBackend } from "#/contexts/active-backend-context";
 import { useBackendScopedPath } from "#/hooks/use-backend-scoped-path";
 import { usePaginatedConversations } from "#/hooks/query/use-paginated-conversations";
+import { useResolvedWorkspaces } from "#/hooks/query/use-resolved-workspaces";
 import { useStartTasks } from "#/hooks/query/use-start-tasks";
 import { useDeleteConversation } from "#/hooks/mutation/use-delete-conversation";
 import { useUnifiedPauseConversation } from "#/hooks/mutation/use-unified-stop-conversation";
@@ -16,6 +17,7 @@ import { NavigationLink } from "#/components/shared/navigation-link";
 import { ExitConversationModal } from "./exit-conversation-modal";
 import { useClickOutsideElement } from "#/hooks/use-click-outside-element";
 import { Provider } from "#/types/settings";
+import type { LocalWorkspace } from "#/types/workspace";
 import { useUpdateConversation } from "#/hooks/mutation/use-update-conversation";
 import {
   displayErrorToast,
@@ -278,6 +280,7 @@ export function ConversationPanel({
     isFetchingNextPage,
     fetchNextPage,
   } = usePaginatedConversations();
+  const { workspaces: knownWorkspaces } = useResolvedWorkspaces();
 
   // Fetch in-progress start tasks
   const { data: startTasks } = useStartTasks();
@@ -341,6 +344,29 @@ export function ConversationPanel({
     () => collectAutomationNameFacets(conversations),
     [conversations],
   );
+
+  const allWorkspacesForGrouping = React.useMemo<
+    readonly LocalWorkspace[]
+  >(() => {
+    if (activeBackend.kind !== "local") {
+      return knownWorkspaces;
+    }
+    const byPath = new Map<string, LocalWorkspace>(
+      knownWorkspaces.map((ws) => [ws.path, ws]),
+    );
+    for (const c of conversations) {
+      const normalized = c.selected_workspace?.trim().replace(/\/+$/, "");
+      if (normalized && !byPath.has(normalized)) {
+        const label = normalized.split("/").filter(Boolean).pop() ?? normalized;
+        byPath.set(normalized, {
+          id: normalized,
+          name: label,
+          path: normalized,
+        });
+      }
+    }
+    return Array.from(byPath.values());
+  }, [activeBackend.kind, conversations, knownWorkspaces]);
 
   const automationFilteredConversations = React.useMemo(
     () =>
@@ -455,12 +481,14 @@ export function ConversationPanel({
       activeBackend.kind,
       conversationSort,
       groupLabels,
+      allWorkspacesForGrouping,
     );
   }, [
     activeBackend.kind,
     conversationSort,
     groupLabels,
     groupedSourceConversations,
+    allWorkspacesForGrouping,
   ]);
 
   const groupDiscoveryConversationIds = React.useMemo(() => {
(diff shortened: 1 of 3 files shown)


georgeglarson (commented):
Sourcing folder rows from the workspace registry (`useResolvedWorkspaces` → `GET /api/workspaces`) instead of deriving them from loaded conversations is what #16435 needs. Runtime behavior against a local agent-server (1.42.1, four registered workspaces incl. zero-conversation ones and names with spaces/apostrophes): every registered workspace renders as a folder row with its registry name, zero-conversation folders sort to the bottom, the "+" launch posts the exact registry path, and the panel's empty state no longer swallows a non-empty folder list. The helper-level tests cover the pre-seed / merge / label-preference / cloud branches well.

A few observations:

1. **This diff deletes a pre-existing explanatory comment** above the `conversationGroups` memo ("Use the unsorted partitions…") that's still accurate — worth restoring.

2. **Raw vs normalized path keying in `allWorkspacesForGrouping` — confirmed at runtime.** `byPath` is keyed on the raw `ws.path`, but the conversation-derived lookup checks `byPath.has(normalizedPath)`. A workspace registered with a trailing slash renders under its registry name while empty, then flips to the path basename once a conversation exists there — both entries normalize to the same `ws:` key inside `groupConversations`, and the later basename-labeled seed overwrites the custom `name`. Keying the memo on the normalized path (same `trim().replace(/\/+$/, "")` as `workspaceGroup`) fixes it.

3. **Version-floor degradation works, but not via the branch it looks like.** The typed `isAgentServerVersionError` fallback is practically unreachable: the app hard-blocks backends below 1.28.0 at onboarding, and even 1.25.0 already serves `/api/workspaces`. What actually protects older/broken servers is the plain-error path (`disableToast` + `failureCount < 3`) — against a proxy that 404s only `/api/workspaces`: exactly 4 requests, none during a 15s idle window, no error toasts, conversation-derived folders still render. Worth a sentence in the PR description so the documented degradation path is the real one.

4. **Nits.** The non-local early-return in `allWorkspacesForGrouping` is dead weight (`groupConversations` ignores `knownWorkspaces` for cloud backends), and the memo computes even in flat/compact mode where the groups go unused.

5. **UX question, not a defect.** Expanding a zero-conversation folder renders an empty content div with no placeholder. Fine to ship, but worth a sentence in the PR description so it reads as a decision rather than an oversight.

Also worth calling out in the description: because the seeds come from the pre-automation-filter conversation list, a workspace whose loaded conversations are all filtered out now renders as an empty folder (and suppresses the panel's empty state) instead of vanishing. That's intended per the PR body, but it's a visible behavior change beyond the issue's literal ask.

One hardening suggestion: there's no component-level test that `conversation-panel.tsx` actually wires `useResolvedWorkspaces` → `groupConversations`, nor one for the new `hasVisibleGroups` empty-state guard — both would protect this against the next refactor of the memo chain.

src/components/features/conversation-panel/conversation-panel.tsx
```diff
@@ -367,9 +393,6 @@ export function ConversationPanel({
     if (compact || organizeMode !== "grouped") {
       return null;
     }
-    // Use the unsorted partitions: groupConversations sorts each bucket
-    // internally by `sortField`, so pre-sorting the merged input is wasted
-    // work in grouped mode (the per-group sort overrides any global order).
```
enyst, line 436: Do we need to remove this comment, or could we keep it?

enyst (commented):

Could you perhaps take a look at the nits surfaced by George’s comment? And sorry, it looks like it has a conflict with main branch, the agent should be able to fix it easily though!

Indeed on main I could reproduce the “missing” workspace, until I loaded everything from the first. Maybe you could attach a screenshot with before and after, in the PR description, for review to see easily the problem is solved?

image

mohitdebian:
addressed the review feedback, thanks for the detailed reviews

- rebased on main, conflict is resolved
- restored the comment above conversationGroups memo that enyst flagged
- fixed the raw vs normalized path keying in allWorkspacesForGrouping so trailing-slash workspace paths dont create duplicate folder entries (george caught this)
- removed the dead non-local early-return that was passing knownWorkspaces through to groupConversations where cloud backends would just ignore it
- added compact/organizeMode guard to the allWorkspacesForGrouping memo so the merge computation skips when the grouped view isnt active

all 237 conversation panel tests pass, pre-commit hooks (typecheck, eslint, prettier, i18n) all green

mohitdebian:
Here are the before and after screenshots showing the workspace grouping fix:
media__1786633469360
media__1786633476216
media__1786633481865

enyst:
hey @mohitdebian the screenshots here are not the before/after screenshots from openhands, they’re just this PR, right? https://github.com/OpenHands/OpenHands/pull/16444#issuecomment-5282552682

mohitdebian:
@enyst Ah, sorry about that! I accidentally uploaded the wrong images. Unfortunately, I recently cleared my local environment so I don't have the 100+ paginated conversations needed to reproduce the visual 'Before/After' screenshot anymore. Since George already verified the runtime behavior locally in his review, could we move forward without the screenshots, or would you still like me to write a script to mock the 100 conversations so I can capture the UI?

mohitdebian:
@enyst @georgeglarson It looks like the `Validate PR description` CI check is failing because the linked issue (#16435) doesn't have the `ready-for-dev` label yet. Could one of you please add the `ready-for-dev` label to issue #16435 so the CI can pass?

enyst (approved):
Thank you!

PR #16445, opened 2026-08-09
fix(mcp): allowlist http(s) for OAuth authorization_url

SashaMIT:
HUMAN:

Ran `npx vitest run src/api/mcp-service/mcp-service.api.test.ts` locally: 8/8 passed, including new `isValidAuthorizationUrl` allow (http/https) and deny (javascript/data/file) cases. No packaged desktop OAuth popup retest in this environment.

AGENT:

`npx vitest run src/api/mcp-service/mcp-service.api.test.ts` — 8 passed (includes new `isValidAuthorizationUrl` allow/deny cases). Change confined to `mcp-service.api.ts` (+ unit tests). No packaged desktop OAuth popup retest of the OAuth popup path in this environment.

---

## Why

`McpService.authorizeOAuth` assigns `popup.location.href = start.authorization_url` with no scheme check. A hostile or confused OAuth start response can hand back `javascript:`, `data:`, or `file:` URLs and navigate the opener popup.

Same class as the device-flow verification URL guard in `device-flow-auth.tsx`. Independent of electron #16376 / #16419 (those cover `shell.openExternal` / in-app localhost classification).

## Summary

- Add `isValidAuthorizationUrl` allowing only `http:` / `https:`
- Reject other schemes before navigating the OAuth popup
- Unit coverage for allow and deny cases

## How to Test

1. `npx vitest run src/api/mcp-service/mcp-service.api.test.ts`
2. Manual (optional): start MCP OAuth test with a normal https authorization URL — popup still navigates
3. Manual (optional): if a probe returns `javascript:alert(1)` as `authorization_url`, authorizeOAuth should fail closed without navigating

## Video/Screenshots

Local vitest: mcp-service.api.test.ts 8/8 including authorization_url scheme allow/deny

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Sibling of #16376 / #16419. Tip parent `68de5c588`. Commit SSH-signed.

PR #16446, opened 2026-08-09
fix(frontend): allowlist schemes in transformVSCodeUrl

SashaMIT:
HUMAN:

Ran `npx vitest run __tests__/utils/vscode-url-helper.test.ts` locally: 8/8 passed, including new allow (`http`/`https`/`vscode`) and deny (`javascript`/`data`/invalid) cases. Call sites already null-check before `window.open`.

AGENT:

`npx vitest run __tests__/utils/vscode-url-helper.test.ts` — 8 passed. Change confined to `src/utils/vscode-url-helper.ts` (+ unit tests). No packaged desktop retest of VS Code open in this environment.

---

## Why

`transformVSCodeUrl` fail-opens: on parse failure it returns the original string, and it never checks the scheme. Callers (`drawer-vscode-link.tsx`, `conversation-card.tsx`) then `window.open` that value. A hostile or confused `vscode_url` can therefore deliver `javascript:` / `data:` into navigation.

Same class as MCP OAuth authorization_url hardening (#16445) and device-flow verification URL work, on a different path (VS Code open helper).

## Fix

- Allowlist `http:`, `https:`, and `vscode:`
- Return `null` for other schemes and for invalid URLs
- Existing callers already skip `window.open` when the helper returns null

## Test plan

- [x] `npx vitest run __tests__/utils/vscode-url-helper.test.ts` (8/8)
- [ ] Smoke: open VS Code from a conversation with a normal `http://localhost…` URL (localhost rewrite still works)
- [ ] Confirm `javascript:` / garbage strings no longer open a window

Made with [Cursor](https://cursor.com)

### DIFF
--- a/__tests__/utils/vscode-url-helper.test.ts
+++ b/__tests__/utils/vscode-url-helper.test.ts
@@ -53,9 +53,25 @@ describe("transformVSCodeUrl", () => {
     expect(transformVSCodeUrl(input)).toBe(input);
   });
 
-  it("should handle invalid URLs gracefully", () => {
+  it("should return null for invalid URLs", () => {
     const input = "not-a-valid-url";
 
+    expect(transformVSCodeUrl(input)).toBeNull();
+  });
+
+  it("should allow vscode: scheme URLs", () => {
+    const input = "vscode://file/workspace/app.ts";
+
     expect(transformVSCodeUrl(input)).toBe(input);
+  });
+
+  it("should return null for javascript: scheme", () => {
+    expect(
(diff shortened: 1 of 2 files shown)


neubig:
Hi @SashaMIT, similarly to my comment here: https://github.com/OpenHands/OpenHands/pull/16376

I feel that this is AI generated without any consideration or understanding for the actual software package that it's contributing to. If you disagree, please feel free to drop in to our slack and we can discuss there, but for now I'm going to close the PRs.

PR #16451, opened 2026-08-09
fix: disable switch_llm by default (#16442)

### LINKED ISSUE #16442: Agent can switch LLM profiles mid-run with no confirmation: switch_llm is default-on and easy to miss
A running conversation can end up on a model the user never chose. The built-in `switch_llm` tool lets the agent move the conversation to any saved LLM profile on its own initiative, it ships enabled, and under the default confirmation policy it runs without asking.

This came from a user report in Slack: he hit his provider's daily rate limit, and later noticed from visibly worse output that the conversation was running a weaker model. He says it happened more than once. The correlation is his observation rather than a demonstrated chain, but what the source does establish is that nothing would have asked him.

Two PRs are already up for pieces of this. Filing the issue for the part they don't cover.

## Mechanism

- `openhands-sdk/openhands/sdk/tool/builtins/switch_llm.py` advertises every saved profile by name and calls `conversation.switch_profile(...)` unconditionally.
- `settings/model.py`: `enable_switch_llm_tool` defaults True, and the wiring is `if self.enable_switch_llm_tool: include_default_tools.append(SwitchLLMTool.__name__)`. No other condition, not profile count, not another flag. `profiles/agent_profile.py` defaults True as well.
- `conversation/state.py` defaults to `NeverConfirm()`, and Canvas sends `{ kind: "NeverConfirm" }` unless the user turns on confirmation mode. So the call executes with no approval step.
- The swap autosaves into `base_state.json`, and resume loads the persisted agent state. A switched model survives a restart rather than reverting to the configured profile.

Nothing switches automatically on a 429. Retries exhaust on the same model and the run errors, but the error stays in history, so the model can see "rate limited" on the next step and elect `switch_llm` itself. That reading is inference, not something I have verified against telemetry. It fits the reported shape.

## What the two PRs cover

- #16441 renders `enable_switch_llm_tool` in Agent settings. The SDK already exported it in the agent schema with GUI metadata; Canvas just had no page for that section, so there was no way to turn the tool off from the GUI.
- #16439 fixes a related display bug: after a switch that lands while the client's socket is down, the chat-input pill and the conversation panel disagree about the running model, permanently, across reloads.

## What neither PR covers

The default. A user who never finds the new toggle still gets an agent that can move their conversation to another model without asking. Worth deciding as a design question rather than leaving to the setting:

- Should `switch_llm` require confirmation regardless of the global policy? It is a different class of action from reading a file.
- Should it be default-off, or gated on something (more than one profile configured, an explicit opt-in)?
- Should a switch leave a persistent marker for the rest of the conversation rather than one inline message that scrolls away?
- Should resume warn when the persisted model differs from the launch profile?

One thing I could not settle from source: whether profiles without a usable credential are filtered out of the advertised switch targets. If they are not, a switch to one fails on first use, which can feed another error-then-switch cycle.

If `switch_llm` calls show up in telemetry, that would size how often this actually fires in the wild.

## Versions read

`openhands-sdk/...` is the separate `OpenHands/software-agent-sdk` repo at `f65a378e`. `src/...` is this monorepo at `4470813ce`.

### Desired Behavior

Switching a running conversation to a different LLM profile should be an explicit, visible user choice. The product should either require confirmation for `switch_llm` regardless of the global confirmation policy, or keep the tool disabled until the user explicitly enables it. A successful switch should remain clearly visible for the rest of the conversation and after resume.

### Acceptance Criteria

- [ ] A default installation cannot switch the active LLM profile without an explicit user opt-in or confirmation.
- [ ] The chosen policy is enforced in the SDK, not only by a Canvas UI control.
- [ ] Automated coverage verifies the default path and the explicit opt-in or confirmation path.
- [ ] After a successful switch, the active profile remains visibly identifiable after reload or resume.
- [ ] Existing conversations and saved agent profiles have a documented migration/default behavior.

KHARSHAVARDHAN-eng:
HUMAN:

AGENT:

---

## Why

Fixes #16442 by disabling `switch_llm` by default in the SDK agent settings and profiles, and annotating `SwitchLLMTool` with `destructiveHint=True` to require user confirmation when enabled.

## Summary

- Changed `enable_switch_llm_tool` default value from `True` to `False` in `OpenHandsAgentSettings` and `AgentProfile`.
- Annotated `SwitchLLMTool` with `destructiveHint=True` so model switching operations trigger confirmation flows under policies requiring user approval.
- Updated SDK unit tests in `test_switch_llm.py`, `test_settings.py`, `test_resolver.py`, and `test_agent_profile.py`.

## Issue Number

Fixes #16442

## How to Test

Run SDK unit tests:
```bash
poetry run pytest software-agent-sdk/tests/sdk/tool/test_switch_llm.py software-agent-sdk/tests/sdk/test_settings.py software-agent-sdk/tests/sdk/profiles/test_resolver.py software-agent-sdk/tests/sdk/profiles/test_agent_profile.py

### DIFF
--- a/enterprise/saas_server.py
+++ b/enterprise/saas_server.py
@@ -58,6 +58,7 @@ from server.routes.user_app_settings import user_app_settings_router  # noqa: E4
 from server.routes.user_provisioning import (  # noqa: E402
     user_provisioning_router,
 )
+from server.routes.mcp_oauth import mcp_oauth_router  # noqa: E402
 from server.routes.users_v1 import (  # noqa: E402
     override_users_me_endpoint,
 )
@@ -95,6 +96,7 @@ base_app.include_router(
 )  # Add routes for credit management and Stripe payment integration
 base_app.include_router(shared_conversation_router)
 base_app.include_router(shared_event_router)
+base_app.include_router(mcp_oauth_router)
 
 # Add GitHub integration router only if GITHUB_APP_CLIENT_ID is set
 if GITHUB_APP_CLIENT_ID:
--- a/enterprise/server/routes/mcp_oauth.py
+++ b/enterprise/server/routes/mcp_oauth.py
@@ -1,0 +1,369 @@
+import asyncio
+import logging
+import uuid
+import json
+from typing import Any, Literal
+from urllib.parse import urlparse, parse_qs
+
+from fastapi import APIRouter, Depends, HTTPException, Request, status
+from fastapi.responses import HTMLResponse, JSONResponse
+from pydantic import BaseModel, Field
+
+from openhands.app_server.config import depends_user_context
+from openhands.app_server.user.user_context import UserContext
+from openhands.sdk.utils.cipher import Cipher
+from storage.encrypt_utils import get_cipher
+
+from fastmcp import Client
+from fastmcp.client.auth.oauth import OAuth, TokenStorageAdapter
+from fastmcp.client.transports.sse import SSETransport
+from fastmcp.client.transports.http import StreamableHttpTransport
+from mcp.shared.auth import OAuthClientMetadata, OAuthClientInformationFull
+from key_value.aio.stores.memory import MemoryStore
+
+logger = logging.getLogger(__name__)
+
+mcp_oauth_router = APIRouter(prefix='/api/v1/mcp/oauth', tags=['MCP OAuth'])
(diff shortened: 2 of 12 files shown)


VascoSch92:
Closing as ~1,100 of 1,133 additions are unrelated commits

PR #1900, opened 2024-05-19
Fix: Update WebSocket URL to use correct protocol from window.location

SmartManoj:
Solves https://github.com/OpenDevin/OpenDevin/issues/1649#issuecomment-2107239803

PR #16482, opened 2026-08-10
fix(streaming): keep the reply in one bubble when a message is sent mid-stream

### LINKED ISSUE #15433: [Bug]: user messages can split streaming text
### Operating System

macOS

### Installation Method

### Agent Canvas Version

1.5.2

### Bug Description

Image

i sent a message during streaming, and it ended up in the middle of the message

### Steps to Reproduce

send a message while streaming is happening

### Actual Behavior

split

### Expected Behavior

message appears below streaming text; new text streams in above the message

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

### LINKED ISSUE #16455: [Bug]: Duplicate messages from the agent in 1.12.0
### Operating System

macOS

### Installation Method

### Agent Canvas Version

1.12.0

### Bug Description

Image

some agent messages are appearing twice

### Steps to Reproduce

unclear

### Actual Behavior

message appears twice

### Expected Behavior

message appears once

### Relevant Logs

```shell

```

### Screenshots

_No response_

### Additional Context

_No response_

VascoSch92:
HUMAN:

I re-open this PR as it got lost during the migration.

AGENT:

---

## Why

Closes #15433. Also closes #16455, a duplicate report of the same bug filed after this fix's original PR (agent-canvas#1901) was closed unmerged during the agent-canvas → OpenHands/OpenHands migration.

Send a message while the agent is streaming and the reply gets cut in half around it.

The user's message is appended to the tail of `uiEvents`, on top of the live `StreamingDeltaEvent`. Delta merging keys on *"the last event is also a delta"* (`handle-event-for-ui.ts`), so the next delta can no longer merge and starts a **second bubble** below the message — the split in the issue's screenshot.

Finalization then makes it worse. `getCurrentTurnContentDeltas` scopes to deltas *after* the last user message, so the first half is never superseded: it stays on screen as an orphan while the canonical final message re-renders the whole text. **The reply renders twice.**

Replaying the reported sequence through the real reducer, before the fix:

```
delta(I am done with the refactor, and )                              <- half 1
user(also update the README)
delta(I also fixed the failing test.)                                 <- half 2, the split

...then the final message arrives:
delta(I am done with the refactor, and )                              <- orphan
user(also update the README)
agent(I am done with the refactor, and I also fixed the failing test.) <- duplicate
```

After:

```
agent(I am done with the refactor, and I also fixed the failing test.)
user(also update the README)
```

## Summary

- A message sent mid-stream is *interleaved* into the agent's run — it does not start a new turn. The three scans of a streaming run (merge target, trailing-delta run, turn boundary) now treat a user message as transparent. Any **other** event still ends the run, which is what keeps this distinct from a genuinely new turn after an interrupted one: the server emits a `ConversationStateUpdateEvent` when the agent leaves `RUNNING`, and that lands in the event store.
- Finalization renders the final event **where the stream ended** rather than at the tail, so the reply doesn't jump underneath a message the user sent while text was already on screen above it.
- Net result: the reply stays one bubble, the message sits below it, and new text streams into the bubble above.

One production file changed (`src/utils/handle-event-for-ui.ts`).

## Provenance

This is a straight port of [agent-canvas#1901](https://github.com/OpenHands/agent-canvas/pull/1901), which passed full CI (unit tests, mock-LLM E2E, mock-LLM Docker E2E) but was closed unmerged on 2026-07-27 when that repo was folded into `OpenHands/OpenHands`, and never reopened here. Verified in this checkout:

```
npm run make-i18n && npx vitest run \
  __tests__/utils/handle-event-for-ui.test.ts \
  __tests__/stores/use-event-store.test.ts \
  __tests__/components/conversation-events/chat/message-sent-mid-stream.test.tsx
# 3 files, 60/60 passed
npx eslint <changed files>   # clean
npx tsc --noEmit             # clean
```

## Issue Number

Fixes #15433
Closes #16455

## How to Test

```bash
npm ci
npm run make-i18n
npm test
npm run lint
npm run build
```

To reproduce by hand: start any conversation, ask for a long tool-free reply (e.g. "Without using any tools, write a ~800-word essay on the history of computing."), and send a second message while it streams. Before this fix: the reply is split around your message, and duplicates when it finalizes. After: one bubble, your message below it.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- **Two existing fixtures were updated, deliberately.** `"keeps deltas from older turns when a later turn finishes"` and `"only reconciles the current turn's delta"` hand-construct a dangling delta followed directly by a user message, with no turn-terminating event in between. That shape is structurally identical to the bug being fixed, so no rule can tell them apart. Real sequences always carry the status update the server emits when the agent leaves `RUNNING`; it was added to both fixtures with an explanatory comment. Their intent — an older turn's delta must survive — is unchanged and still asserted.
- **A related issue, #1900 in the old repo, is not fixed here.** Same trigger, different layer: it's a server-side race in `software-agent-sdk` where `Agent.astep` releases the conversation-state lock for the LLM call, so a message landing in that window sees `execution_status == RUNNING` and never gets recorded once the run finishes. Fixed separately in `OpenHands/software-agent-sdk#4194`.
- **Known limitation, pre-existing:** on reload, history is re-sorted by server timestamp, so a mid-stream message renders *above* the reply rather than below it. Live and reloaded views therefore differ. Out of scope here.

Opening as a draft — flagging for review before marking ready, since this is a port of a previously-approved-in-CI PR rather than freshly written against current main.

---

| Component | Value |
|---|---|
| **Automation** | `openhands-automation==1.8.0` |
| **Commit** | `92a514a88031fab964caea406e73b76f63d20bb4` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-92a514a
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-92a514a
```

```
ghcr.io/openhands/agent-canvas:sha-92a514a-amd64
ghcr.io/openhands/agent-canvas:vasco-fix-message-during-streaming-amd64
ghcr.io/openhands/agent-canvas:pr-16482-amd64
ghcr.io/openhands/agent-canvas:sha-92a514a-arm64
ghcr.io/openhands/agent-canvas:vasco-fix-message-during-streaming-arm64
ghcr.io/openhands/agent-canvas:pr-16482-arm64
ghcr.io/openhands/agent-canvas:sha-92a514a
ghcr.io/openhands/agent-canvas:vasco-fix-message-during-streaming
ghcr.io/openhands/agent-canvas:pr-16482
```

- Each tag (e.g., `sha-92a514a`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-92a514a-amd64`) are also available if needed


src/stores/use-event-store.ts
```diff
@@ -129,20 +129,30 @@ const sortEventState = (state: EventState): EventState => ({
   uiEvents: [...state.uiEvents].sort(compareEventsByTimestamp),
 });
 
+/**
+ * Resorts only `events`, the raw append log: `uiEvents` is display-ordered by
+ * handleEventForUI, which deliberately places a finalized reply above a
+ * mid-stream message even though the reply's timestamp is later (#1899).
+ * Resorting `uiEvents` by raw timestamp here would silently undo that
+ * placement the next time any event — e.g. a trailing state snapshot — arrives
+ * with a timestamp earlier than the array's current last element.
```
enyst, line 164: Sorry I'm dense, I don't understand this comment...

VascoSch92: @enyst 

I always ask to make the comment minimal. In this case was really minimal :-) 

The store keeps two arrays. events is the raw append log and stays in timestamp order. uiEvents is what's rendered, and it's ordered by handleEventForUI, not by timestamp. 

The PR makes that ordering deliberate: when a streamed reply finalizes, the final message is placed where the stream ended (for example above a message the user sent mid-stream) even though the final message's timestamp is later. That placement is the fix.

applyAddEvent used to re-sort both arrays whenever an incoming event was stamped earlier than the array's last element. Re-sorting uiEvents by timestamp cancels that placement and drops the reply back under the user's message.

For example, you have a situation like this one

```text
delta          00:01.000
user msg   00:01.500
delta          00:02.000
final           00:02.500   -> uiEvents = [final, user msg]              correct
snapshot   00:02.200   -> old code re-sorted both arrays
                                      -> uiEvents = [user msg, snapshot, final]    bug back
```

So the reply would render correctly and then silently jump below the message on the next out-of-order arrival. 

Make it sense?

VascoSch92:
@enyst ready for another review :-)

src/stores/use-event-store.ts
enyst: Oh ok thank you! 🙏 

Does a state snapshot refer to a conversation update status event?

keeltrace:

I found a path where the ordering regression can still happen **during the same live session**, not only after reload.

`addEvents()` still finishes with `sortEventState(...)`, which timestamp-sorts `uiEvents`. `useLoadOlderEvents.loadOlder()` calls `addEvents(older)` whenever scrolling up fetches another history page. That means after this PR intentionally produces display order like:

```text
final agent reply  00:02.500
mid-stream user    00:01.500
```

loading any older non-duplicate event through pagination re-sorts the whole derived `uiEvents` array by timestamp and changes the relevant tail back to:

```text
mid-stream user    00:01.500
final agent reply  00:02.500
```

So the reply can jump below the user's mid-stream message again without a reload.

A minimal store regression test should be deterministic:

1. `addEvent(firstDelta)`, `addEvent(midStreamUser)`, `addEvent(secondDelta)`, `addEvent(final)` and assert the PR's intended `[final, midStreamUser]` UI ordering.
2. Then call `addEvents([olderEvent])`.
3. The current bulk path should reorder those two by timestamp, reproducing the bug.

This seems materially different from the documented reload-only limitation because older-history pagination uses the same bulk path during normal live use. Could we cover this path before merge? One possible direction is to rebuild derived `uiEvents` by replaying `handleEventForUI` over the sorted raw event log when bulk history is merged, rather than directly timestamp-sorting an already-derived `uiEvents` array. Simply removing the `uiEvents` sort from `addEvents` may not be enough because older pages still need to land in the correct historical position.

rbren:
@enyst I'd love to get this one in if we can, https://github.com/OpenHands/OpenHands/issues/16455 is pretty glaring

VascoSch92:
Hey @keeltrace 

you were right.

Fixed :-)

PR #16496, opened 2026-08-10
fix: preserve conversation usage metrics from search results

### LINKED ISSUE #16480: AppConversation.metrics is always null from the conversation search/list endpoint — toAppConversation drops stats.usage_to_metrics
## Summary

`AppConversation.metrics` (what consumers read for a conversation's accumulated cost/token usage) comes back `null` for essentially every conversation returned by `usePaginatedConversations()` / `AgentServerConversationService.searchConversations()`, even when the agent-server has real, non-zero usage recorded for that conversation.

## Root cause

`toAppConversation()` (`src/api/agent-server-adapter.ts`) maps `info.metrics` straight into the output:

```ts
metrics: info.metrics
  ? { accumulated_cost: ..., accumulated_token_usage: ... }
  : null,
```

`info.metrics` is `stored.metrics` on the agent-server side (`conversation_service.py:463` in software-agent-sdk) — a separate, apparently rarely-populated field, distinct from the conversation's live per-usage-id stats. The real numbers are on the wire as `stats.usage_to_metrics` (the same shape `RuntimeConversationStats` / `RuntimeMetrics` already declare, and the same data `getCombinedMetrics()` in `conversation-metrics.ts` already combines for the metrics modal on an *active* conversation) — but `requireDirectConversationInfo()` (`agent-server-conversation-service.api.ts`) narrows the raw item down to `DirectConversationInfo` before `toAppConversation` ever runs, and `DirectConversationInfo` doesn't declare a `stats` field at all. So `stats` is dropped before the adapter even gets a chance to read it.

## Evidence

Sampled all 23 conversations from a real local dev backend (`/api/conversations/search`, fresh restart, no caching involved):

- `metrics: null` for 23/23 (100%)
- `stats.usage_to_metrics` non-empty for 17/23, with real numbers — e.g. one conversation alone had 53,781 prompt tokens + 3,553 completion tokens recorded in `stats.usage_to_metrics.default.accumulated_token_usage`.

So this isn't "these conversations never had usage" — the usage is tracked and returned by the agent-server; it's discarded on the way to `AppConversation`.

## Suggested fix

In `toAppConversation`, when `info.metrics` is absent, fall back to combining `info.stats.usage_to_metrics` the same way `getCombinedMetrics()` already does for `RuntimeConversationInfo`. That means threading `stats` through `DirectConversationInfo` / `requireDirectConversationInfo` so it survives to reach the adapter, then reusing (or extracting a shared helper from) `getCombinedMetrics()`'s combination logic. Fixing it at the adapter level fixes it for every consumer of `AppConversation.metrics`, not just one call site.

## Workaround

Landed a page-local-only workaround on the sidebar-links branch (issue #16401) that reads the raw `ConversationClient.searchConversations()` response directly, bypassing `toAppConversation` — not something to keep long-term, just enough to unblock that one page. Worth fixing properly at the adapter level so cost displays, sidebars, etc. that already read `AppConversation.metrics` benefit too.

uuzzrm:
HUMAN:

I reviewed the patch and the attached test evidence. I have not independently run the full local stack.

AGENT:

This change was developed with AI assistance. 

---
## Why

`AgentServerConversationService.searchConversations()` receives real usage under `stats.usage_to_metrics`, but `requireDirectConversationInfo()` dropped that field before `toAppConversation()` ran. As a result, `AppConversation.metrics` was `null` for conversations that had recorded token usage and cost.

## Summary

- Preserve and defensively normalize `stats.usage_to_metrics` while adapting conversation search results.
- Reuse the existing metric-combination logic so `AppConversation.metrics` includes all usage IDs when the legacy `metrics` field is absent.
- Add a regression test covering multiple usage IDs and token aggregation.

## Issue Number

Fixes #16480

## How to Test

Focused regression and related API/UI checks:

```text
node ./node_modules/vitest/vitest.mjs run __tests__/api/agent-server-conversation-service.test.ts __tests__/api/agent-server-adapter.test.ts __tests__/api/cloud/conversation-runtime-info.test.ts __tests__/hooks/query/use-conversation-metrics.test.tsx --reporter=dot
Test Files  4 passed (4)
Tests       120 passed (120)
```

Repository checks:

```text
npm run lint
0 errors, 3 pre-existing warnings

npm run build
completed successfully
```

The full `npm test` run completed with 4 unrelated existing failures in root/automation/skills/conversation-panel tests; the overall result was 4 failed files, 574 passed files, 1 skipped, and 4,597 passed tests. The changed-path tests above passed.

## Video/Screenshots

Reproduction notes for this non-visual API bug: before the change, a search item containing `stats.usage_to_metrics.agent.accumulated_cost = 1.25` and `condenser.accumulated_cost = 0.75` produced `AppConversation.metrics === null`. After the change, the same response produces `accumulated_cost === 2` and combines token usage across both entries. The regression test exercises the agent-server response through the service normalizer and adapter boundary.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The explicit `metrics` field remains authoritative when present. Responses without `stats` retain the previous `null` value, while malformed stats entries are ignored safely. This draft is linked to #16480, which still needs the repository's `ready-for-dev` label before it can pass the non-draft PR description gate.


uuzzrm:
Removed the fork-only `.pr/conversation-metrics-repro.svg` artifact and its PR-body image reference after the artifact check reported that fork PRs require manual cleanup. The functional patch and its text reproduction notes remain unchanged. The remaining description check is the repository requirement that linked issue #16480 carry the `ready-for-dev` label, which needs a maintainer update.

Issue #324, opened 2024-03-28
Ww


Berti1983:
**Describe the bug**

**Steps to Reproduce**
1.
2.
3.

**Expected behavior**

**Actual behavior**

**Additional context**

rbren:
🏆

PR #16507, opened 2026-08-11
fix(automations): surface unhealthy automation reasons

### LINKED ISSUE #16494: Surface unhealthy/disabled automations and their blocking reason in the dashboard
### Desired Behavior

The automation dashboard shows which automations are unhealthy or auto-disabled and why, surfacing the persisted blocking reason / failure class on the run — so a user can tell at a glance why an automation isn't running or keeps failing instead of digging into a conversation trajectory. It must visually distinguish "failed run" from "disabled because misconfigured/blocked" from "transient issue", and offer an obvious path to fix and re-enable.

### Acceptance Criteria

- [ ] Dashboard renders a per-automation health/disabled state with the surfaced reason, sourced from the run metadata exposed by `OpenHands/automation`.
- [ ] Config/blocked/disabled states are visually distinct from transient-failure states at a glance.
- [ ] Provides an explicit re-enable / fix action for disabled automations.
- [ ] Empty and unknown states degrade gracefully (no raw error or hanging state).

### Source

Follow-up to the [Slack thread](<https://allhandsai.slack.com/archives/C0BGK8NRYUX/p1786379033571769>) on auto-disabling buggy automations and surfacing the reason. Paired backend/disable logic is tracked in OpenHands/automation#323.

uuzzrm:
HUMAN:

I reviewed the changes and verified the automation dashboard locally. Disabled automations show their reason along with the Turn on and View actions.
Codex 图像 2026年8月10日 20_33_26

AGENT:

This change was developed with AI assistance.

## Why

Automations that are disabled or repeatedly failing currently expose only a
generic health badge and the last run error. That makes a configuration or
integration blocker look the same as a temporary provider failure, and leaves
users to open the detail page to understand what needs fixing.

## Summary

- Read the optional automation/run health metadata from newer automation services while keeping older responses compatible.
- Show distinct disabled, blocked, failed, and transient notices in both card and list views.
- Surface persisted blocking reasons, provide fallback copy when metadata is empty, and keep the explicit re-enable and View actions available.
- Treat a blocking reason on a completed or failed run as blocked, classify known transient kinds separately, and keep legacy `error_detail` responses distinct from transient failures.
- Add localized labels and focused tests for classification, rendering, and the disabled/transient/failed distinctions.

## Issue Number

Fixes #16494

## How to Test

- `node scripts/make-i18n-translations.cjs` — passed.
- `node scripts/check-translation-completeness.cjs` — passed.
- `node node_modules/vitest/vitest.mjs run __tests__/components/automations/automation-health-notice.test.tsx __tests__/components/automations/automation-card.test.tsx __tests__/components/automations/automation-list-row.test.tsx __tests__/manifests/automation-insights.test.ts __tests__/routes/automations-dashboard.test.tsx __tests__/routes/automations-list.test.ts` — 6 files, 65 tests passed.
- Targeted ESLint and Prettier checks — passed.
- `node_modules/.bin/react-router build` — passed. The build only reported the repository's existing chunk-size and React Router future-flag warnings.
- `node node_modules/typescript/bin/tsc --noEmit --skipLibCheck` — currently blocked by the existing `vite.config.ts:91` `Excessive stack depth` / `UserConfigFnObject` error; no changed file appears in the diagnostic.
- Manual check: started `npm run dev:mock -- --port 3001`, opened `/automations`, verified populated and empty disabled reasons, then dismissed the preference overlay and clicked Turn on. The mock server returned `PATCH /api/automation/v1/a1000000-0000-0000-0000-000000000005` with 200, and the card changed from Disabled to Never run with Run now enabled.

## Video/Screenshots

The manual UI evidence is included in the human-authored section above.

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- The new fields are optional so the UI remains compatible with older
  `OpenHands/automation` deployments.
- This follow-up keeps the repository's AI-assistance disclosure intact.


uuzzrm:
Follow-up in `13eedc4df`:

- Surface `blocking_reason` for completed runs as well as failed runs, so an agent-reported blocker is not presented as a healthy completion.
- Keep legacy automation-service responses with only `error_detail` on the existing health-badge path instead of guessing whether the failure was transient or actionable.
- Added regression coverage for both cases.

Validation on this head:
- 44 targeted Vitest tests passed
- Prettier check passed
- ESLint passed for the changed files
- TypeScript typecheck passed
- Translation completeness check passed

keeltrace:

The cross-repo API contract this PR is consuming looks stale now.

`getAutomationHealthDetails()` reads `automation.disabled_failure_kind` / `disabled_reason` and `latestRun.failure_kind` / `blocking_reason`. Those fields were implemented in the paired `OpenHands/automation` health work (#324/#325), but both of those PRs are now closed unmerged. On current `OpenHands/automation` main, the run-side structured diagnostic contract is instead `status_detail`, with fields such as `phase`, `kind`, `detail`, and `transient`.

That means on the current public stack the new detailed `blocked` / `transient` reason paths here appear to have no API source: the optional fields added by this PR will simply be absent, while current `status_detail` is not read. The focused tests pass because they construct those abandoned fields directly in frontend fixtures.

Could this be rebased onto the current `status_detail` contract (and whatever replacement backend contract will own auto-disable reasons), or explicitly blocked on a live backend PR/version that exposes these exact fields? Otherwise we risk merging UI states that are only reachable in mocks while the real backend emits a different shape.

PR #16512, opened 2026-08-11
fix: fail fast on unwritable Docker bind mounts and document the permission fix

### LINKED ISSUE #16511: [Bug]: Docker quickstart fails: bind-mount permission error kills agent-server and automation silently
### Operating System

Linux

### Installation Method

Docker (`ghcr.io/openhands/agent-canvas`)

### Agent Canvas Version

1.12.0 (also reproduced on a locally built image from `main`)

### Bug Description

Running the Docker quickstart in `README.md` **verbatim** produces a stack where the frontend serves but the agent-server and automation backend crash at startup — every API call returns "Bad Gateway". Root cause: the container runs as uid **10001** (`openhands`), but the bind-mounted host dir (`~/.openhands`) is owned by the host user (e.g. uid 1000, mode `700`). The container cannot write its settings, API keys, conversations, or the automation SQLite database through the mount.

The confusing part: the static frontend still serves fine, so you get a *working-looking UI* that 502s on everything — the classic "is the server down?" trap.

### Steps to Reproduce

1. `export PROJECTS_PATH="$HOME/projects"` and `mkdir -p "$PROJECTS_PATH" "$HOME/.openhands"` (leaving the host dir at default ownership/mode).
2. Run the README quickstart:
   ```sh
   docker run -it --rm \
     -p 8000:8000 \
     -v "$HOME/.openhands:/home/openhands/.openhands" \
     -v "${PROJECTS_PATH}:/projects" \
     ghcr.io/openhands/agent-canvas:1.12.0
   ```
3. Wait for the container to boot (~60s).
4. Open the UI at http://localhost:8000/canvas — it loads, but `/server_info` returns **502** and every conversation/API call fails with "Bad Gateway: connect ECONNREFUSED 127.0.0.1:18000".

### Actual Behavior

Run method: `agent-canvas` Docker image (`ghcr.io/openhands/agent-canvas:1.12.0`)

The frontend serves (HTTP 200) while the agent-server and automation backend never come up:

- `/server_info` → **HTTP 502** (ECONNREFUSED 127.0.0.1:18000)
- Agent-server log: `PermissionError: '/home/openhands/.openhands/agent-canvas/conversations'` → "Application startup failed. Exiting."
- Automation log: `sqlite3.OperationalError: unable to open database file` → migrations fail → app exits

Image

A/B confirmation that this is the mount, not the image: the **same image** with a permissive mount (`-v /tmp/oh-state:/home/openhands/.openhands`, mode 777) boots the full stack in ~10s with **zero** permission errors.

### Expected Behavior

The Docker quickstart should either work out of the box, or fail fast with an actionable message explaining the permission problem and how to fix it (rather than silently running a half-dead stack).

### Relevant Logs

```
mkdir: cannot create directory '/home/openhands/.openhands': Permission denied
/opt/agent-canvas/entrypoint.sh: line 80: /home/openhands/.openhands/agent-canvas/secret-key.txt: Permission denied
chmod: cannot access '/home/openhands/.openhands/agent-canvas/secret-key.txt': Permission denied
mkdir: cannot create directory '/home/openhands/.openhands': Permission denied
/opt/agent-canvas/entrypoint.sh: line 98: /home/openhands/.openhands/agent-canvas/api-key.txt: Permission denied
chmod: cannot access '/home/openhands/.openhands/agent-canvas/api-key.txt': Permission denied
```

### Acceptance Criteria

- [x] A screenshot/video demonstrating the bug can be replicated on `main` is attached to the PR.
- [ ] A screenshot/video demonstrating that the bug is fixed on the PR branch is attached to the PR.
- [ ] Following the updated quickstart (`chmod` or `--user` guidance) starts the full stack without permission errors.
- [ ] Running the container with an unwritable bind mount exits quickly with a clear, actionable error message instead of a half-dead stack.

### Screenshots

Image

### Additional Context

Verified end-to-end with a locally built image containing the proposed fix:

- Broken mount → container exits in **2 seconds** with:
  ```
  [agent-canvas] ERROR: Persistence directory is not writable: /home/openhands/.openhands
  [agent-canvas] ERROR: The container runs as uid 10001; the bind-mounted host dir
  [agent-canvas] ERROR: must be writable by that uid. Fix with either:
  [agent-canvas] ERROR:   1) chmod -R a+rwX ~/.openhands   # run ON THE HOST, not in the container
  [agent-canvas] ERROR:   2) docker run ... --user "$(id -u):$(id -g)"
  ```
- Permissive mount → full stack boots (server_info 200, frontend 200).

Lothnic:
HUMAN:

Tested the Docker quickstart with the chmod fix, now stack boots cleanly and fail fast error appears on a broken mount.

AGENT:

Verified end-to-end by building the Docker image locally from this branch and
running an A/B test against identical mounts:

- **Build:** `node scripts/docker-build.mjs --tag agent-canvas:preflight` (agent-server 1.40.1 base, automation 1.6.0) — exit 0.
- **A — new entrypoint + broken mount** (host dir uid 1000, mode 700): container **exits code 1 in 2s** with `[agent-canvas] ERROR: Persistence directory is not writable: /home/openhands/.openhands` plus both fixes (`chmod` / `--user`).
- **B — new entrypoint + permissive mount** (mode 777): full stack boots — `/server_info` 200 in ~14s, `/canvas` 200.
- **C — old image (`1.12.0`) + same broken mount**: container stays up, frontend 200, `/server_info` 502 — the half-dead trap this PR removes.
- `bash -n docker/entrypoint.sh` clean; functional preflight test passed all three cases (writable / read-only / nonexistent dir).

---

## Why

The Docker quickstart in `README.md` is broken for typical Linux users. The container runs as uid **10001** (`openhands`), but the bind-mounted `~/.openhands` is owned by the host user (uid 1000, mode `700`). The agent-server (conversations dir) and automation (SQLite DB) both crash at startup with `Permission denied`, while the static frontend keeps serving — a half-dead stack that looks fine in the browser but 502s on every API call, with no hint of why.

## Summary

- `docker/entrypoint.sh`: add a writability preflight for the persistence dir that **fails fast** with an actionable error (both fixes: `chmod` the host dirs, or `--user "$(id -u):$(id -g)"`) instead of silently running a broken stack.
- `README.md`: Docker quickstart now makes the mounted dirs writable (with a uid-10001 note) and documents the `--user` alternative in a tip block.
- `README.windows.md`: note that the uid issue is Linux-only — Docker Desktop handles bind-mount permissions transparently.

## Issue Number

Fixes #16511

## How to Test

1. `docker build` from this branch: `node scripts/docker-build.mjs --tag agent-canvas:preflight` (or rely on the published image once this lands).
2. **Reproduce the bug**: run the README quickstart with the host dir at default perms — with the fix, the container now exits within ~2s with a clear error message instead of running a half-dead stack.
3. **Verify the fix**: apply either documented fix, then re-run:
   - `chmod -R a+rwX "$HOME/.openhands" "$PROJECTS_PATH"`, or
   - add `--user "$(id -u):$(id -g)"` to `docker run`.
4. Confirm the stack boots: `curl http://localhost:8000/server_info` returns 200 and the UI loads at `http://localhost:8000/canvas`.

### Before

image

### After

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The entrypoint change is baked in at image build time — the published image must be rebuilt/released for end users to get the fail-fast behavior. The README fix helps immediately regardless.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ docker run -it --rm \
   -p 8000:8000 \
   -v "$HOME/.openhands:/home/openhands/.openhands" \
   -v "${PROJECTS_PATH}:/projects" \
-  ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
+  ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
 ```
 
 > [!NOTE]
@@ -124,7 +124,7 @@ docker run -it --rm \
 >   -p 8000:8000 \
 >   -v "$HOME/.openhands:/home/openhands/.openhands" \
 >   -v "${PROJECTS_PATH}:/projects" \
->   ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
+>   ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
 > ```
 >
 > The `-e HOME=/home/openhands` is required: the entrypoint derives all
--- a/README.md
+++ b/README.md
@@ -91,9 +91,6 @@ agent-canvas --backend-only   # agent server + automation backend + ingress only
 ```sh
 export PROJECTS_PATH="$HOME/projects"  # directory containing your project folders
 mkdir -p "$PROJECTS_PATH" "$HOME/.openhands"
-# The container runs as uid 10001 (openhands) — bind-mounted host dirs must be
-# writable by that uid, or the agent-server and automation crash at startup.
-chmod -R a+rwX "$HOME/.openhands" "$PROJECTS_PATH"
 
 docker run -it --rm \
   -p 8000:8000 \
@@ -101,6 +98,18 @@ docker run -it --rm \
   -v "${PROJECTS_PATH}:/projects" \
   ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
 ```
+
+> [!NOTE]
+> **Linux (Docker Engine) only:** the container runs as uid 10001 (openhands) —
+> bind-mounted host dirs must be writable by that uid, or the agent-server and
+> automation crash at startup. Run this **before** `docker run`:
+>
+> ```sh
+> chmod -R a+rwX "$HOME/.openhands" "$PROJECTS_PATH"
+> ```
+>
+> Docker Desktop (macOS/Windows) handles bind-mount permissions transparently, so
+> no `chmod` is needed there.
 
 > [!TIP]
 > The image's `openhands` user is **uid 10001**. Instead of `chmod`-ing the host
--- a/README.md
+++ b/README.md
@@ -100,22 +100,11 @@ docker run -it --rm \
 ```
 
 > [!NOTE]
-> **Linux (Docker Engine) only:** the container runs as uid 10001 (openhands) —
+> **Linux (Docker Engine):** the container runs as uid 10001 (`openhands`) —
 > bind-mounted host dirs must be writable by that uid, or the agent-server and
-> automation crash at startup. Run this **before** `docker run`:
->
-> ```sh
-> chmod -R a+rwX "$HOME/.openhands" "$PROJECTS_PATH"
-> ```
->
-> Docker Desktop (macOS/Windows) handles bind-mount permissions transparently, so
-> no `chmod` is needed there.
-
-> [!TIP]
-> The image's `openhands` user is **uid 10001**. Instead of `chmod`-ing the host
-> dirs, you can run the container as your own user so the bind mounts are
-> writable automatically (all persistent state lives under the mounted
-> `.openhands` dir, so this is safe in practice):
+> automation crash at startup. The easiest fix is to run the container as your
+> own user, so the bind mounts are writable automatically (all persistent state
+> lives under the mounted `.openhands` dir, so this is safe in practice):
 >
 > ```sh
 > docker run -it --rm \
@@ -132,6 +121,18 @@ docker run -it --rm \
 > no `/etc/passwd` entry inside the container, so `$HOME` would not resolve to
 > `/home/openhands` — without pinning it, state would silently detach from the
 > bind mount.
+>
+> Alternatively, grant write access on the host — scoped to directories only,
+> because a recursive `chmod -R` would also strip the `600` permission of the
+> persisted secret files (`secret-key.txt`, `api-key.txt`) and make them
+> world-readable:
+>
+> ```sh
+> find "$HOME/.openhands" "$PROJECTS_PATH" -type d -exec chmod a+rwX {} +
+> ```
+>
+> Docker Desktop (macOS/Windows) handles bind-mount permissions transparently,
+> so none of this is needed there.
 
 **Windows (PowerShell / Windows Terminal):** See [README.windows.md](./README.windows.md) for the equivalent commands.
 
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ docker run -it --rm \
   -p 8000:8000 \
   -v "$HOME/.openhands:/home/openhands/.openhands" \
   -v "${PROJECTS_PATH}:/projects" \
-  ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
+  ghcr.io/openhands/agent-canvas:1.15.0 # x-release-please-version
 ```
 
 > [!NOTE]
@@ -113,7 +113,7 @@ docker run -it --rm \
 >   -p 8000:8000 \
 >   -v "$HOME/.openhands:/home/openhands/.openhands" \
 >   -v "${PROJECTS_PATH}:/projects" \
->   ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
+>   ghcr.io/openhands/agent-canvas:1.15.0 # x-release-please-version
 > ```
 >
 > The `-e HOME=/home/openhands` is required: the entrypoint derives all
--- a/README.md
+++ b/README.md
@@ -91,6 +91,9 @@ agent-canvas --backend-only   # agent server + automation backend + ingress only
 ```sh
 export PROJECTS_PATH="$HOME/projects"  # directory containing your project folders
 mkdir -p "$PROJECTS_PATH" "$HOME/.openhands"
+# The container runs as uid 10001 (openhands) — bind-mounted host dirs must be
+# writable by that uid, or the agent-server and automation crash at startup.
+chmod -R a+rwX "$HOME/.openhands" "$PROJECTS_PATH"
 
 docker run -it --rm \
   -p 8000:8000 \
@@ -98,6 +101,21 @@ docker run -it --rm \
   -v "${PROJECTS_PATH}:/projects" \
   ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
 ```
+
+> [!TIP]
+> The image's `openhands` user is **uid 10001**. Instead of `chmod`-ing the host
+> dirs, you can run the container as your own user so the bind mounts are
+> writable automatically (all persistent state lives under the mounted
+> `.openhands` dir, so this is safe in practice):
+>
+> ```sh
+> docker run -it --rm \
+>   --user "$(id -u):$(id -g)" \
+>   -p 8000:8000 \
+>   -v "$HOME/.openhands:/home/openhands/.openhands" \
+>   -v "${PROJECTS_PATH}:/projects" \
+>   ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
+> ```
 
 **Windows (PowerShell / Windows Terminal):** See [README.windows.md](./README.windows.md) for the equivalent commands.
 
--- a/README.md
+++ b/README.md
@@ -120,11 +120,18 @@ docker run -it --rm \
 > ```sh
 > docker run -it --rm \
 >   --user "$(id -u):$(id -g)" \
+>   -e HOME=/home/openhands \
 >   -p 8000:8000 \
 >   -v "$HOME/.openhands:/home/openhands/.openhands" \
 >   -v "${PROJECTS_PATH}:/projects" \
 >   ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
 > ```
+>
+> The `-e HOME=/home/openhands` is required: the entrypoint derives all
+> persistence paths from `$HOME`, and an arbitrary uid (e.g. your host uid) has
+> no `/etc/passwd` entry inside the container, so `$HOME` would not resolve to
+> `/home/openhands` — without pinning it, state would silently detach from the
+> bind mount.
 
 **Windows (PowerShell / Windows Terminal):** See [README.windows.md](./README.windows.md) for the equivalent commands.
 
--- a/README.windows.md
+++ b/README.windows.md
@@ -12,7 +12,7 @@ For the main install options and overall context, see [README.md](./README.md).
 - A host directory for `PROJECTS_PATH` containing the project folders you want the agent to access (create it before starting the container)
 
 ```powershell
-docker pull ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
+docker pull ghcr.io/openhands/agent-canvas:1.15.0 # x-release-please-version
 
 $env:PROJECTS_PATH = Join-Path $HOME "projects"  # directory containing your project folders
 New-Item -ItemType Directory -Force -Path $env:PROJECTS_PATH, (Join-Path $env:USERPROFILE ".openhands") | Out-Null
@@ -21,7 +21,7 @@ docker run -it --rm `
   -p 8000:8000 `
   -v "$($env:USERPROFILE)\.openhands:/home/openhands/.openhands" `
   -v "$($env:PROJECTS_PATH):/projects" `
-  ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
+  ghcr.io/openhands/agent-canvas:1.15.0 # x-release-please-version
 ```
 
 Open [http://localhost:8000/canvas](http://localhost:8000/canvas) in your browser.
--- a/README.windows.md
+++ b/README.windows.md
@@ -27,3 +27,10 @@ docker run -it --rm `
 Open [http://localhost:8000/canvas](http://localhost:8000/canvas) in your browser.
 
 The agent will be able to access any project under `PROJECTS_PATH`.
+
+> [!NOTE]
+> Docker Desktop for Windows/macOS handles bind-mount permissions transparently.
+> On Linux (Docker Engine), the container's `openhands` user is **uid 10001** —
+> if the agent-server and automation fail to start, make the mounted host dirs
+> writable by that uid (see the note in
+> [README.md](./README.md#option-2-with-a-docker-sandbox)).
--- a/README.windows.md
+++ b/README.windows.md
@@ -12,7 +12,7 @@ For the main install options and overall context, see [README.md](./README.md).
 - A host directory for `PROJECTS_PATH` containing the project folders you want the agent to access (create it before starting the container)
 
 ```powershell
-docker pull ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
+docker pull ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
 
 $env:PROJECTS_PATH = Join-Path $HOME "projects"  # directory containing your project folders
 New-Item -ItemType Directory -Force -Path $env:PROJECTS_PATH, (Join-Path $env:USERPROFILE ".openhands") | Out-Null
@@ -21,7 +21,7 @@ docker run -it --rm `
   -p 8000:8000 `
   -v "$($env:USERPROFILE)\.openhands:/home/openhands/.openhands" `
   -v "$($env:PROJECTS_PATH):/projects" `
-  ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
+  ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
 ```
 
 Open [http://localhost:8000/canvas](http://localhost:8000/canvas) in your browser.
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -67,13 +67,9 @@ export OH_CONVERSATIONS_PATH="${OH_CONVERSATIONS_PATH:-${OPENHANDS_DIR}/${CONFIG
 export OH_BASH_EVENTS_DIR="${OH_BASH_EVENTS_DIR:-${OPENHANDS_DIR}/${CONFIG_BASH_EVENTS:-agent-canvas/bash_events}}"
 
 # ── Preflight: persistence dir must be writable ────────────────────────────
-# The image runs as uid 10001 (openhands). A bind-mounted host directory
-# (e.g. `-v ~/.openhands:/home/openhands/.openhands`) that isn't writable by
-# that uid makes BOTH the agent-server (conversations dir) and automation
-# (SQLite DB) crash at startup while the static frontend keeps serving — a
-# half-dead stack that looks fine in the browser but 502s on every API call.
-# Fail fast with an actionable message instead of letting users chase a
-# phantom backend error.
+# The image runs as uid 10001 (openhands); a host bind mount that isn't
+# writable by that uid leaves the backend half-dead (frontend serves,
+# APIs 502). Fail fast with an actionable message.
 if ! (mkdir -p "$OPENHANDS_DIR" && touch "$OPENHANDS_DIR/.write-test" 2>/dev/null); then
   log_error "Persistence directory is not writable: $OPENHANDS_DIR"
   log_error "The container runs as uid $(id -u); the bind-mounted host dir"
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -66,6 +66,24 @@ export OH_PERSISTENCE_DIR="${OH_PERSISTENCE_DIR:-${OPENHANDS_DIR}}"
 export OH_CONVERSATIONS_PATH="${OH_CONVERSATIONS_PATH:-${OPENHANDS_DIR}/${CONFIG_CONVERSATIONS:-agent-canvas/conversations}}"
 export OH_BASH_EVENTS_DIR="${OH_BASH_EVENTS_DIR:-${OPENHANDS_DIR}/${CONFIG_BASH_EVENTS:-agent-canvas/bash_events}}"
 
+# ── Preflight: persistence dir must be writable ────────────────────────────
+# The image runs as uid 10001 (openhands). A bind-mounted host directory
+# (e.g. `-v ~/.openhands:/home/openhands/.openhands`) that isn't writable by
+# that uid makes BOTH the agent-server (conversations dir) and automation
+# (SQLite DB) crash at startup while the static frontend keeps serving — a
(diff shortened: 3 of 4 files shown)


docker/entrypoint.sh
```diff
@@ -66,6 +66,24 @@ export OH_PERSISTENCE_DIR="${OH_PERSISTENCE_DIR:-${OPENHANDS_DIR}}"
 export OH_CONVERSATIONS_PATH="${OH_CONVERSATIONS_PATH:-${OPENHANDS_DIR}/${CONFIG_CONVERSATIONS:-agent-canvas/conversations}}"
 export OH_BASH_EVENTS_DIR="${OH_BASH_EVENTS_DIR:-${OPENHANDS_DIR}/${CONFIG_BASH_EVENTS:-agent-canvas/bash_events}}"
 
+# ── Preflight: persistence dir must be writable ────────────────────────────
```
VascoSch92: trimmer the comment here

docker/entrypoint.sh
```diff
@@ -66,6 +66,24 @@ export OH_PERSISTENCE_DIR="${OH_PERSISTENCE_DIR:-${OPENHANDS_DIR}}"
...
+# The image runs as uid 10001 (openhands). A bind-mounted host directory
+# (e.g. `-v ~/.openhands:/home/openhands/.openhands`) that isn't writable by
+# that uid makes BOTH the agent-server (conversations dir) and automation
+# (SQLite DB) crash at startup while the static frontend keeps serving — a
+# half-dead stack that looks fine in the browser but 502s on every API call.
+# Fail fast with an actionable message instead of letting users chase a
+# phantom backend error.
+if ! (mkdir -p "$OPENHANDS_DIR" && touch "$OPENHANDS_DIR/.write-test" 2>/dev/null); then
```
VascoSch92: The preflight only probes `$OPENHANDS_DIR`, but `OH_PERSISTENCE_DIR`/`OH_CONVERSATIONS_PATH`/`OH_BASH_EVENTS_DIR` (set two lines above, at 65-67) fall back to `$OPENHANDS_DIR` only when unset

if a user sets any of them directly to a different path, the check validates the wrong directory.

VascoSch92: The preflight always diagnoses any write failure as a `uid-10001` permission mismatch and pushes c`hmod/--user fixes`, even when the real cause is unrelated (e.g. a read-only bind mount or a full disk), so the remediation steps it prints cannot fix the actual problem.

docker/entrypoint.sh
```diff
@@ -66,6 +66,24 @@ export OH_PERSISTENCE_DIR="${OH_PERSISTENCE_DIR:-${OPENHANDS_DIR}}"
 export OH_CONVERSATIONS_PATH="${OH_CONVERSATIONS_PATH:-${OPENHANDS_DIR}/${CONFIG_CONVERSATIONS:-agent-canvas/conversations}}"
```
VascoSch92, line 66: The new preflight write-test only checks $OPENHANDS_DIR, not the /projects bind mount that README.md documents as needing the identical uid-10001 chmod fix.

docker/entrypoint.sh
VascoSch92: `2>/dev/null` on the preflight line only scopes to the touch command, not the preceding `mkdir -p`, so a `mkdir` failure leaks a raw, unprefixed OS error line instead of the clean actionable message the comment promises.

README.md
```diff
@@ -91,6 +91,9 @@ agent-canvas --backend-only   # agent server + automation backend + ingress only
 ```sh
```
VascoSch92, line 91: README.md applies the uid-10001 chmod workaround unconditionally to macOS, contradicting the PR's own new README.windows.md note that Docker Desktop (macOS/Windows) handles bind-mount permissions transparently and only Linux/Docker Engine is affected.

README.md
```diff
@@ -91,6 +91,9 @@ agent-canvas --backend-only   # agent server + automation backend + ingress only
...
 export PROJECTS_PATH="$HOME/projects"  # directory containing your project folders
 mkdir -p "$PROJECTS_PATH" "$HOME/.openhands"
+# The container runs as uid 10001 (openhands) — bind-mounted host dirs must be
+# writable by that uid, or the agent-server and automation crash at startup.
+chmod -R a+rwX "$HOME/.openhands" "$PROJECTS_PATH"
```
VascoSch92: The documented fix chmod -R a+rwX "$HOME/.openhands" "$PROJECTS_PATH" recursively strips the protective 600 permissions that entrypoint.sh sets on the persisted secret-key.txt and api-key.txt files. (Same root cause also at docker/entrypoint.sh:81, which prints this chmod as its suggested fix.)

README.md
```diff
@@ -99,6 +102,21 @@ docker run -it --rm \
   ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
 ```
 
+> [!TIP]
```
VascoSch92: The new --user "$(id -u):$(id -g)" TIP is presented as a safe alternative to chmod, but $HOME does not resolve to /home/openhands for arbitrary uids inside the container, so entrypoint.sh's ${HOME}-derived persistence path silently detaches from the bind mount (or crashes) — the opposite of what the TIP promises. (Same root cause also at docker/entrypoint.sh:63.)

docker/entrypoint.sh
Lothnic: Fixed in `097743ad8`
The preflight now probes all four effective paths in a loop:  `$OPENHANDS_DIR`  plus the three overridable  `OH_PERSISTENCE_DIR`  /  `OH_CONVERSATIONS_PATH`  /  `OH_BASH_EVENTS_DIR` , so an overridden path gets validated and the error names the exact failing directory ` $OPENHANDS_DIR`  stays in the set because  `STATE_DIR`  (secret-key/api-key persistence) always lives under it, even when the  `OH_*`  vars are overridden. Verified with a writable-dirs pass case and a chmod-555 failure case that correctly reports the offending directory.

Lothnic: I agree that the fix hardcodes the diagnosis to a uid mismatch, how should I go forward with the implementation here
- surface the real error and keep a conditional hint or
- actively diagnose the failure mode with more precise messaging

docker/entrypoint.sh
Lothnic: fixed in [75b5682](https://github.com/OpenHands/OpenHands/pull/16512/commits/75b5682a68f5e7a5e5b9f6999b33d55aede13016)
Added a `/projects` probe which only runs when the mount exists

README.md
Lothnic: Fixed in [b44173c](https://github.com/OpenHands/OpenHands/pull/16512/commits/b44173cb89d0dce06fde33b9f294d563878884e5)
Moved the chmod into a Linux (Docker Engine) only note, macOS/Windows (Docker Desktop) now correctly skip it which is consistent with README.windows.md.

README.md
Lothnic: Fixed in [ddff3ee](https://github.com/OpenHands/OpenHands/pull/16512/commits/ddff3ee56e7d6279e374be4fa6c92dda5994773d)
`--user` TIP now pins `$HOME` as `-e HOME=/home/openhands` is required as the entrypoint derives all persistence paths from `$HOME`

README.md
Lothnic: Agreed `chmod -R a+rwX`  would open up the 600-protected  `secret-key.txt`  /  `api-key.txt`.
 Would you mind sharing how you'd approach the fix here? Happy to implement whatever you think fits best.

README.md
rajshah4: I re-tested the exact proposed workaround against the current official image:

```bash
--user "$(id -u):$(id -g)" \
-e HOME=/home/openhands
```

With a typical Linux host UID/GID of `1000:1000` and a mount writable by UID 1000, it still fails:

```text
uid=1000(pn) gid=1000(pn)
10001:10001 700 /home/openhands
ACCESS_FAILED
ls: cannot access '/home/openhands/.openhands': Permission denied
WRITE_FAILED
touch: cannot touch '/home/openhands/.openhands/probe': Permission denied
```

UID 1000 is already the `pn` account in the image, while `/home/openhands` is owned by `10001:10001` and has mode `0700`. Setting `HOME=/home/openhands` changes the variable but does not give UID 1000 permission to traverse that directory or reach the bind mount beneath it.

The fail-fast preflight remains useful, but I don't think `--user "$(id -u):$(id -g)" -e HOME=/home/openhands` should be documented as a safe remediation until this case is addressed and tested.

_This comment was generated by an OpenHands AI agent on behalf of the user._

Lothnic: Thanks for reproducing this. You’re right: the workaround still failed because /home/openhands itself was 0700, so the host UID could not traverse it to reach the bind mount.

 I pushed 16b0a6013 and [17c120d](https://github.com/OpenHands/OpenHands/pull/16512/commits/17c120dfa079d3a7da28280ce044c6f7fc181d6a), which sets /home/openhands to 0755 during the image build. The mounted openhands directory still controls read/write access, and the default openhands user plus secret-file permissions are unchanged.

This needs to be tested against an image rebuilt from the updated branch; the currently published image will still show the old behavior. Could you re-run the same --user "$(id -u): $(id -g)" -e HOME=/home/openhands test against the image built from this PR branch?

rajshah4: Re-tested against an image rebuilt from `17c120d` on a fresh Ubuntu 24.04 EC2 (t3.medium). Built with `node scripts/docker-build.mjs --tag agent-canvas:pr16512`. Both cases pass.

## Case A — `--user "$(id -u):$(id -g)" -e HOME=/home/openhands` (the case that failed on the published image)

Host `ubuntu` uid 1000:1000, `~/.openhands` and `~/projects` at default ownership (uid 1000).

```text
$ docker run -d --name ac-pr16512-user -p 8000:8000 \
    --user "$(id -u):$(id -g)" -e HOME=/home/openhands \
    -v "$HOME/.openhands:/home/openhands/.openhands" -v "$HOME/projects:/projects" \
    agent-canvas:pr16512

$ docker ps -a --format "{{.Names}} {{.Status}}" | grep ac-pr16512-user
ac-pr16512-user Up About a minute

$ curl -s -o /dev/null -w "server_info:%{http_code}\n" http://localhost:8000/server_info
server_info:200
$ curl -s -o /dev/null -w "canvas:%{http_code}\n" http://localhost:8000/canvas
canvas:200

$ docker exec ac-pr16512-user sh -c 'id; ls -ld /home/openhands; ls -ld /home/openhands/.openhands'
uid=1000(pn) gid=1000(pn)
drwxr-xr-x 1 openhands openhands 4096 Aug 20 02:36 /home/openhands      ← now 0755
drwxrwxr-x 6 pn        pn        4096 Aug 20 02:37 /home/openhands/.openhands  ← bind mount reachable + writable
```

No `Persistence directory is not writable` preflight error; agent-server, automation, and frontend all started. **The case that failed on the published image now works.** ✅

## Case B — broken mount fails fast (regression check)

Mount `0700` owned by host uid 1000, container runs as `openhands` (uid 10001):

```text
$ docker run -d --name ac-pr16512-broken2 -p 8002:8000 \
    -v /tmp/oh-readonly2:/home/openhands/.openhands agent-canvas:pr16512

$ docker ps -a --format "{{.Names}} {{.Status}}" | grep ac-pr16512-broken2
ac-pr16512-broken2 Exited (1) 9 seconds ago

$ docker logs ac-pr16512-broken2 2>&1 | tail -6
[agent-canvas] ERROR: Persistence directory is not writable: /home/openhands/.openhands
[agent-canvas] ERROR: The container runs as uid 10001; the bind-mounted host dir
[agent-canvas] ERROR: must be writable by that uid. Fix with either:
[agent-canvas] ERROR:   1) chmod -R a+rwX ~/.openhands   # run ON THE HOST, not in the container
[agent-canvas] ERROR:   2) docker run ... --user "$(id -u):$(id -g)"
```

Fail-fast preflight intact. ✅

## One unrelated observation

The Case A logs include a non-fatal `browser_use`/chromium `PermissionError: '/home/openhands/.config'` during tool-usability probing. It doesn't block the backend (agent-server + automation come up and serve 200s), and it's not related to this PR's `/home/openhands` 0755 fix — flagging it so it isn't mistaken for a regression here.

## Verdict

`17c120d` fixes the `--user "$(id -u):$(id -g)" -e HOME=/home/openhands` case that failed on the published image (`/home/openhands` is now 0755, so the host UID can traverse to the bind mount), and the fail-fast preflight on broken mounts still works as designed. The `--user` TIP is safe to document as a remediation on Linux/Docker Engine with this branch.

_This comment was generated by an OpenHands AI agent on behalf of the user._

Lothnic: Thanks for re-testing this on a fresh Ubuntu host. This confirms both intended behaviors: the host-UID workaround now works with the rebuilt image, and invalid bind mounts still fail fast with an actionable error. The non-fatal browser_use/Chromium permission warning is unrelated to this change. Marking this resolved.

README.md
```diff
@@ -96,9 +96,43 @@ docker run -it --rm \
...
+  ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
 ```
 
+> [!NOTE]
+> **Linux (Docker Engine) only:** the container runs as uid 10001 (openhands) —
+> bind-mounted host dirs must be writable by that uid, or the agent-server and
+> automation crash at startup. Run this **before** `docker run`:
+>
+> ```sh
+> chmod -R a+rwX "$HOME/.openhands" "$PROJECTS_PATH"
```
rajshah4: Security concern (also raised by @VascoSch92 earlier, still open): `chmod -R a+rwX "$HOME/.openhands"` recursively strips the `600` that `entrypoint.sh` sets on the persisted `secret-key.txt` (the Fernet key for settings encryption) and `api-key.txt` (session API key), making both world-readable on the host. Low impact on a single-user dev box, but this becomes the documented remediation, so worth fixing here.

Two options I'd suggest:

1. Scope the recursion to directories only:
   ```sh
   find "$HOME/.openhands" "$PROJECTS_PATH" -type d -exec chmod a+rwX {} +
   ```
2. Or make the `--user` TIP the primary path (it now works with `17c120d` — verified) and demote `chmod` to a fallback with a security note.

Same fix should apply to the hint the entrypoint prints at `docker/entrypoint.sh:81` (and the `/projects` one at :98).

README.md
```diff
@@ -96,9 +96,43 @@ docker run -it --rm \
   -p 8000:8000 \
   -v "$HOME/.openhands:/home/openhands/.openhands" \
   -v "${PROJECTS_PATH}:/projects" \
-  ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
+  ghcr.io/openhands/agent-canvas:1.14.0 # x-release-please-version
```
rajshah4: This version bump is now the merge conflict: `main` moved to `1.15.0` after v1.15.0 shipped on 2026-08-21, and this line is annotated `# x-release-please-version` (release-please owns it). Suggest dropping the manual bump from this PR entirely — it's not part of the fix's intent, and reverting it removes the conflict. Same for the two occurrences in `README.windows.md`.

rajshah4 (changes requested):
Thanks for sticking with this one, @Lothnic — the iteration quality here has been genuinely good. Four of @VascoSch92's threads got real fixes (not just comments), and the `17c120d` response to the re-test request turned around a previously-broken workaround into a verified one. That takes care.

## What I verified

Built the image from `17c120d` on a fresh Ubuntu 24.04 EC2 (Linux Docker Engine — the affected platform) and ran both key cases:

- **Case A** — `--user "$(id -u):$(id -g)" -e HOME=/home/openhands`: ✅ boots cleanly. `/home/openhands` is now `0755` (was `0700` on the published image), the host uid can traverse to the bind mount, `server_info` 200, `canvas` 200, no preflight error. This is the case that **failed on the published image** — your `chmod 755` at Dockerfile L153 fixes it.
- **Case B** — root-owned `0700` mount, default user: ✅ `Exited(1)` in ~5s with the clean, actionable error. Fail-fast preflight intact.

The core fix is sound, `bash -n` clean, CI green. Great work.

## Two required items before merge

1. **Version bump conflict** (inline at `README.md:99`) — the `1.14.0` bump conflicts with `main` at `1.15.0`; release-please owns that line, so dropping the manual bump fixes the conflict. This is the only thing making the PR `dirty`.
2. **`chmod -R a+rwX` strips secret-file modes** (inline at `README.md:108`) — the documented remediation opens the `600`-protected `secret-key.txt`/`api-key.txt` to all host users. Small change: scope to directories only, or make the `--user` path primary.

## Optional (non-blocking, left as discussion)

- The preflight always diagnoses a uid-10001 mismatch — for read-only mounts or full disks the printed fixes are misleading. Prepending the actual OS error to the message would cover this without losing the actionable hints.
- `2>/dev/null` scopes to `touch` only in the persistence-dir probe — cosmetic, and subsumed by the above if you capture the error.

Once the two required items land, this is an approve from me.

_This review was generated by an OpenHands AI agent on behalf of @rajshah4, with functional verification performed on a fresh Ubuntu 24.04 EC2 host._

README.md
Lothnic: Thanks for flagging this, addressed in b167530 (scoped to dirs) and 59acec5. README now makes `--user "$(id -u):$(id -g)"` + `-e HOME` the primary fix (verified working), and demotes chmod to a fallback using `find … -type d -exec chmod a+rwX {} +` with an explicit warning that `chmod -R` would make `secret-key.txt`/`api-key.txt` world-readable. Both entrypoint hints at :81/:98 updated to match.

README.md
Lothnic: Done in 59acec5, dropped the manual bumps and synced the four `x-release-please-version` lines to `1.15.0` (current main). PR is now mergeable again with no conflicts.

panau161:
How about changing the ids in the docker image? If the openhands user had id 1000 and not 10001, I think way fewer people would be affected by the initial id mismatch issue. Currently it looks like this:
```
pn:x:1000:1000::/home/pn:/bin/bash
openhands:x:10001:10001::/home/openhands:/bin/bash
```
What is the pn user doing/used for? (I guess there are good reasons not to change it but I wanted to suggest it here.)

docker/entrypoint.sh
Lothnic: Fixed in b8b71d819. The preflight now captures stderr from both commands and picks its remediation from the *actual* OS error instead of asserting the cause:

- permission denied / operation not permitted -> keeps the uid + `chmod` / `--user` guidance
- `Read-only file system` -> tells you to drop the `:ro` suffix
- anything else (ENOSPC, quota, ...) -> reports the error without inventing a cause

`report_unwritable()` is shared by the persistence-dirs probe and the `/projects` probe. Regression coverage lives in `__tests__/scripts/docker-entrypoint-preflight.test.ts`, which executes the preflight region lifted out of the real script.

Old vs new for the same failure (parent path is a file, ENOTDIR):

```
OLD:
mkdir: cannot create directory '/tmp/ab/blocker': Not a directory        <- leaked raw
[agent-canvas] ERROR: Persistence directory is not writable: ...
[agent-canvas] ERROR: The container runs as uid 1000; ...                <- wrong cause
[agent-canvas] ERROR:   2) find ~/.openhands -type d -exec chmod a+rwX {} + ...

NEW:
[agent-canvas] ERROR: Persistence directory is not writable: ...
[agent-canvas] ERROR: Underlying error: mkdir: cannot create directory '...': Not a directory
[agent-canvas] ERROR: Resolve the filesystem error above, then retry.
```

Lothnic: Fixed in b8b71d819. The write test is now:

```sh
if ! err=$(mkdir -p "$dir" 2>&1 && touch "$dir/.write-test" 2>&1); then
```

so stderr from *both* commands is captured instead of printed, and the captured text is surfaced as `Underlying error: ...`. A test asserts that no stderr line can begin with a bare `mkdir:` / `touch:` prefix, i.e. nothing leaks unprefixed to the console.

README.md
Lothnic: Addressed in b1675309f — the recursive `chmod -R` is gone. Both the entrypoint's printed hint and the README now scope it to directories only:

```sh
find "$HOME/.openhands" "$PROJECTS_PATH" -type d -exec chmod a+rwX {} +
```

That makes the dirs traversable/writable by uid 10001 without touching the mode of the persisted `secret-key.txt` / `api-key.txt` files, which stay at `600`.

README.md
Lothnic: Same fix as the linked thread — b1675309f replaces `chmod -R a+rwX` with a directories-only `find ... -type d -exec chmod a+rwX {} +` in both the README and the entrypoint's printed remediation, so the protective `600` on the persisted secret files is preserved.

README.md
Lothnic: Resolved by dropping the version edits rather than re-syncing them (0b95aa8ea). `README.md` / `README.windows.md` now carry the fork-point tag, so this PR no longer edits any release-please-owned line — the net diff against the merge base shows those lines as unchanged context, so a 3-way merge with `main` takes `main`'s pin with no conflict. GitHub now reports `mergeable: MERGEABLE`.

README.md
```diff
@@ -96,9 +96,44 @@ docker run -it --rm \
   -p 8000:8000 \
   -v "$HOME/.openhands:/home/openhands/.openhands" \
   -v "${PROJECTS_PATH}:/projects" \
-  ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
+  ghcr.io/openhands/agent-canvas:1.15.0 # x-release-please-version
```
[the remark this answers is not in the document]
Lothnic: Done — dropped rather than re-synced (0b95aa8ea). Both READMEs are back to their pre-PR values for those lines, so release-please owns them exclusively again.

The one *new* reference this PR introduced (the `--user` example) had its tag removed entirely in c034cad37, so no new release-please-owned line is added either.

README.windows.md
```diff
@@ -12,7 +12,7 @@ For the main install options and overall context, see [README.md](./README.md).
 - A host directory for `PROJECTS_PATH` containing the project folders you want the agent to access (create it before starting the container)
 
 ```powershell
-docker pull ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
+docker pull ghcr.io/openhands/agent-canvas:1.15.0 # x-release-please-version
```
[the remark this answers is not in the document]
Lothnic: Same fix — 0b95aa8ea leaves the `docker pull` / `docker run` pins at their pre-PR values in both READMEs, so this PR no longer touches a release-please-owned line in either file.

README.md
```diff
@@ -99,6 +99,41 @@ docker run -it --rm \
...
+> lives under the mounted `.openhands` dir, so this is safe in practice):
+>
+> ```sh
+> docker run -it --rm \
+>   --user "$(id -u):$(id -g)" \
+>   -e HOME=/home/openhands \
+>   -p 8000:8000 \
+>   -v "$HOME/.openhands:/home/openhands/.openhands" \
+>   -v "${PROJECTS_PATH}:/projects" \
+>   ghcr.io/openhands/agent-canvas:1.12.0 # x-release-please-version
```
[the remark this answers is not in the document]
Lothnic: Fixed in c034cad37 by dropping the tag: `ghcr.io/openhands/agent-canvas`.

Worth flagging: the `:latest` you suggested would still fail this check — the assertion is `expect(ref).toBe(dockerImage)` over matches of `/ghcr\.io\/openhands\/agent-canvas:[^\s`"]+/`, which matches `:latest` too. An untagged reference isn't matched at all, so the pin-sync check cannot drift again.

I reproduced the original failure on the merge ref (738 passed / 1 failed) and confirmed the fix there before pushing.

PR #16521, opened 2026-08-11
feat(automations): add a Git Sync page

VascoSch92:
HUMAN:

FE for git sync in agent-canvas

AGENT:

---

## Why

The automation service is growing a git sync feature
(OpenHands/automation#327) that mirrors automations to a git repo for
visibility, versioning and backup. It has no UI: the repo, branch, path,
credentials, commit author and sync interval can only be set by calling the
endpoints by hand, and there is no way to see whether a sync worked.

## Summary

- Adds a Git Sync page at `/automations/git-sync`, reached from a button on the
  automations list. Shows whether sync is on, the repo/branch/path, last synced
  commit and time, how many automations are waiting to be pushed, and the last
  error — plus a **Sync now** button and a settings form.
- Adds `getGitSyncStatus` / `updateGitSyncConfig` / `triggerGitSync` to
  `AutomationService`, with matching query and mutation hooks.
- Adds `OH_AUTOMATION_LOCAL_PATH` to the dev launcher, mirroring
  `OH_AGENT_SERVER_LOCAL_PATH`, so the stack can run against a local automation
  checkout.

## Issue Number

**No `ready-for-dev` issue exists for this yet** — the feature originates from
OpenHands/automation#300, which lives in the automation repo. Flagging rather
than linking something that does not meet the bar; happy to open one here and
wait for triage before this leaves draft.

## How to Test

Requires the automation service from OpenHands/automation#327, which is not
released yet — hence the launcher flag below.

```bash
git clone https://github.com/OpenHands/automation ~/automation
git -C ~/automation checkout vasco/git-sync
git init --bare -b main /tmp/git-sync-remote.git   # throwaway target, no credentials needed

OH_AUTOMATION_LOCAL_PATH=~/automation \
AUTOMATION_GIT_SYNC_ENABLED=1 \
AUTOMATION_GIT_SYNC_REPO_URL=/tmp/git-sync-remote.git \
npm run dev:automation
```

Then open `http://localhost:8000/automations` → **Git Sync**.

1. The page shows Enabled/Disabled and Encrypted/Not encrypted pills, the repo,
   branch, path, last synced commit/time, and pending changes.
2. Create an automation, press **Sync now**, and watch pending changes drop to
   0 and the last synced commit appear. Confirm it landed:
   `git -C /tmp/git-sync-remote.git ls-tree -r HEAD --name-only`
3. Change the branch in the form and save — the other fields are untouched
   (the API takes partial updates).
4. Reload: the access token and encryption key fields are **always empty**, by
   design, since the API never returns them.
5. Set **Sync every (seconds)** to 20 and confirm syncing begins without a
   restart; set it back to 0 for manual-only.

Verified against a live stack, not just unit tests: created automations through
the UI's backend and confirmed each one was serialized, committed and pushed —
first to a local bare repo, then to a real GitHub repo after repointing the URL
from this page. Also exercised the failure paths: a 503 from **Sync now** while
sync is disabled surfaces the "Enable Git Sync before triggering a sync"
message, and a failed cycle renders the error banner with its timestamp.

- `npm run lint` — clean (0 errors; the 3 warnings are pre-existing in
  `use-local-git-info.ts`).
- `npx vitest run` — **4616 passing**, 0 failures.

## Video/Screenshots

Screenshot 2026-08-14 at 09 42 10

## Type

- [x] Feature

## Notes

**Depends on** OpenHands/automation#327. The page degrades safely without it —
the endpoints 404 and the error state renders — but it is only useful with that
branch.

**Endpoint paths are literal**, not routed through `getAutomationEndpoint`.
That manifest describes the automation surface a host may remap, and
`InterfaceEndpoints` requires every key it declares, so adding git-sync keys
would break existing manifests. Git sync is a local-mode operator feature
outside that surface — happy to move it into the manifest if you'd rather, but
it looked like a breaking change to that contract.

**Local backends only**, matching the service: one repo maps to one agent
server. Cloud backends get an explanatory empty state rather than a broken page.

**Write-only secrets.** The token and encryption key are never returned by the
API, so the form never pre-fills them and never implies it knows the current
value. Each has an explicit "clear" control, so removing a credential is
distinct from leaving it unchanged.

---

| Component | Value |
|---|---|
| **Commit** | `29be537863195957413dfa2aea5228e10d7b4a6f` |

```bash
docker pull ghcr.io/openhands/agent-canvas:sha-29be537
```

**Run**
```bash
docker run -it --rm \
  -p 8000:8000 \
  ghcr.io/openhands/agent-canvas:sha-29be537
```

```
ghcr.io/openhands/agent-canvas:sha-29be537-amd64
ghcr.io/openhands/agent-canvas:vasco-git-sync-ui-amd64
ghcr.io/openhands/agent-canvas:pr-16521-amd64
ghcr.io/openhands/agent-canvas:sha-29be537-arm64
ghcr.io/openhands/agent-canvas:vasco-git-sync-ui-arm64
ghcr.io/openhands/agent-canvas:pr-16521-arm64
ghcr.io/openhands/agent-canvas:sha-29be537
ghcr.io/openhands/agent-canvas:vasco-git-sync-ui
ghcr.io/openhands/agent-canvas:pr-16521
```

- Each tag (e.g., `sha-29be537`) is a **multi-arch manifest** supporting both **amd64** and **arm64**
- Individual architecture tags (e.g., `sha-29be537-amd64`) are also available if needed


src/components/features/automations/git-sync/git-sync-config-form.tsx
```diff
@@ -0,0 +1,286 @@
...
+            isToggled={clearToken}
+            isDisabled={!canManage}
+            onToggle={setClearToken}
+          >
+            {t(I18nKey.AUTOMATIONS$GIT_SYNC$CLEAR_TOKEN)}
+          </SettingsSwitch>
+        </div>
+
+        <div className="flex flex-col gap-2">
+          <SettingsInput
```
neubig, line 395: It's not very obvious what this is, so we could probably add a tooltip. Alternatively, we could remove this and just use the same secret key as is used by the canvas.

src/components/features/automations/git-sync/git-sync-config-form.tsx
```diff
@@ -0,0 +1,286 @@
...
+            isToggled={clearEncryptionKey}
+            isDisabled={!canManage}
+            onToggle={setClearEncryptionKey}
+          >
+            {t(I18nKey.AUTOMATIONS$GIT_SYNC$CLEAR_ENCRYPTION_KEY)}
+          </SettingsSwitch>
+        </div>
+
+        <div className="flex justify-start">
+          <BrandButton
```
neubig, line 439: We may want to consider testing whether the configuration works by changing this button to "Save and Sync", where we try to sync and only save the settings if the sync works.

src/routes/automation-git-sync.tsx
```diff
@@ -0,0 +1,162 @@
...
+          <ErrorState onRetry={() => refetch()} />
+        </div>
+      </div>
+    );
+  }
+
+  const handleSyncNow = () => {
+    triggerMutation.mutate(undefined, {
+      onSuccess: () => {
+        displaySuccessToast(t(I18nKey.AUTOMATIONS$GIT_SYNC$SYNC_TRIGGERED));
```
neubig: Instead of a toast, I think ideally we could show progress in the page as long as the sync is in progress.

VascoSch92: Yes... now is a little weird.

VascoSch92: https://github.com/user-attachments/assets/a4d763b1-8486-4ba0-8892-f3f5549aa7e0

neubig: LGTM!

src/components/features/automations/git-sync/git-sync-overview-section.tsx
```diff
@@ -0,0 +1,162 @@
...
+        pendingCount={status.dirty_count}
+      />
+
+      <div className="mt-5 grid grid-cols-2 gap-x-4 gap-y-5">
+        <ConfigField
+          icon={<GlobeIcon className="size-3.5" />}
+          label={t(I18nKey.AUTOMATIONS$GIT_SYNC$REPOSITORY)}
+        >
+          <span className="break-all">
+            {status.repo_url || t(I18nKey.AUTOMATIONS$GIT_SYNC$NOT_CONFIGURED)}
```
neubig: Could we make this a link?

src/components/features/automations/git-sync/git-sync-config-form.tsx
VascoSch92: Good point on the help text, I'll add a line under the input.

On reusing the canvas key: I don't think we can, because this key has to exist somewhere other than the box it runs on. We push to GitHub so the automations are still readable when this instance isn't: after a rebuild, from a second instance, or just by you six months later. The canvas key can't serve that: unless you set `OH_SECRET_KEY` yourself, it's generated on first boot into secret-key.txt and never shown anywhere, so no copy of it exists outside that machine. 

make it sense?

src/components/features/automations/git-sync/git-sync-overview-section.tsx
VascoSch92: done :-)

src/components/features/automations/git-sync/git-sync-config-form.tsx
VascoSch92: I changed. So now we have two buttons: one just for save (but does a fast check if the token and the repo exist) and the second to save and sync together.

I tested in the VM.

neubig (approved):
I have been using this and it works very well.

PR #16525, opened 2026-08-11
refactor(backends): extract host and connection modules from the backend form

### LINKED ISSUE #15535: Extract a backend-connection module from backend-form-modal.tsx (1,071 lines)
`src/components/features/backends/backend-form-modal.tsx` is 1,071 lines, and reusable business logic is trapped inside the render module: host kind inference, local/private-address detection, host normalization, URL validation, connection-test error formatting, and `testBackendConnection()` all sit inline alongside multiple UI layers (`BackendFormModal`, `BackendForm`, `BackendConnectionOptions`, `ManualConnectionColumn`, `CloudLoginColumn`, status rendering).

These pure validation/normalization functions are private today, so they cannot be unit-tested directly. That's the clearest missing test seam.

### Confirmed
- The file is 1,071 lines with the logic above inline.
- The pure host-validation/normalization functions are private → not unit-testable without mounting the modal.
- `BackendForm` and `BackendConnectionOptions` are **already exported and reused** (onboarding + API-key entry flows depend on them).

### Framing correction (per review)
- The OAuth/device-flow path is **not** trapped inside the modal. `DeviceFlowAuth` (`device-flow-auth.tsx`) and `useDeviceFlow` (`src/hooks/use-device-flow.ts`) are already extracted, and `useDeviceFlow` has focused hook tests. What remains inline is only the cloud-login column composition, host-override handling, and the `onConnected` payload mapping — so this should not be framed as "OAuth is untested / modal-trapped."
- A simpler `normalizeHost()` already exists in `src/api/agent-server-client-options.ts`. Extraction must reuse/consolidate that rather than introduce a second, subtly different host-normalization policy — unless the difference is intentional and documented.

### Proposed approach (refined — extract logic first, split UI last)
1. **Extract host parsing/normalization/validation** into a small pure module, consolidating with the existing `normalizeHost()`. Add tests for: bare hosts, `localhost`, private IPv4, bracketed IPv6, bare IPv6, `.local`, explicit schemes, whitespace, and empty input.
2. **Extract connection testing** into a service/hook boundary so UI code no longer constructs SDK clients directly.
3. **Only then split rendering** if it still helps. Likely seams:
   - pure backend host utilities
   - backend connection-test service
   - `useBackendForm`
   - manual connection column
   - cloud login column
4. **Keep `BackendForm` and `BackendConnectionOptions` public surfaces stable** — onboarding and API-key entry flows rely on them.

### Acceptance criteria
- [ ] Host validation/normalization extracted into a pure module, consolidated with the existing `normalizeHost()` (no second policy unless documented)
- [ ] Connection testing extracted to a service/hook boundary; UI no longer constructs SDK clients directly
- [ ] Unit tests cover the host module: bare host, localhost, private IPv4, bracketed IPv6, bare IPv6, `.local`, explicit scheme, whitespace, empty input
- [ ] Behavior tests for the flows that can regress: add local backend, edit existing cloud backend without changing its `kind`, locked-cloud-host mode, manual API-key-required mode
- [ ] `BackendForm` / `BackendConnectionOptions` public surfaces unchanged
- [ ] Add-backend mock-LLM e2e flows still pass
- [ ] `npm run typecheck` stays green

---
*This issue was drafted by an AI agent on behalf of the user, and refined per review feedback in [the discussion below](https://github.com/OpenHands/OpenHands/issues/15535#issuecomment-5092635808).*

ameyypawar:
HUMAN:

Tested locally on Node 24.15 to match CI: `npm run lint` clean, `npx vitest run backend` green (23 suites, 245 tests). Checked the Add Backend form in `npm run dev:mock` — a host with a space is rejected and Connect stays disabled, a bare hostname is accepted and Connect enables. Didn't run the mock-LLM e2e suite locally, CI runs it on this PR.

AGENT:

Validated locally on Node 24.15, matching the `node-version` pin in `ci.yml`
(the repo's volta pin is 22.12.0; both were run, results identical):

- `npm ci` — clean
- `npm run typecheck` — clean
- `npm run lint` — 0 errors (3 pre-existing warnings in files this PR does not touch:
  `conversation-card-context-menu.tsx` ×2, `use-local-git-info.ts` ×1)
- `npx vitest run backend` — 23 backend-related suites, 245 tests, all pass
  (the narrower `__tests__/api/backend-registry/ __tests__/components/backends/`
  pair is 13 files / 183 tests)
- `npm test` — 580 files, 4648 tests, exit 0 (4644 before this PR; the 4 added are
  the two `ftp://` assertions and the two version-badge tests)
- `npm run build`, `npm run build:lib`, `npm pack --dry-run` — all exit 0

End-to-end check beyond the unit suites is in Video/Screenshots below: the
add/edit backend form still validates hosts and the agent-server version badge
still resolves, which is what the moved code is responsible for.

---

## Why

`backend-form-modal.tsx` keeps reusable business logic inside the render module:
host kind inference, local/private-address detection, host normalisation, URL
validation, and the agent-server probe all sit inline alongside five UI layers.
Because the pure functions are private, none of them can be unit-tested without
mounting the modal — that is the clearest missing test seam, and it is what this
PR addresses first.

Two notes on the issue text, from re-deriving it against current `main`:

- The issue says the file is 1,071 lines. It is now **1,638** — it has grown
  ~53% since the issue was filed, which makes the extraction more worthwhile,
  not less.
- The issue describes one trapped `ServerClient` construction. There are
  **two** in the modal — the submit-time preflight and the edit-form status
  row — plus an exact third copy in `backend-version.tsx`. All three are
  handled here; leaving any behind would have made "UI no longer constructs SDK
  clients directly" false while claiming it true.

## Summary

- Extract the pure host helpers into `src/api/backend-registry/backend-host.ts`
  (`stripTrailingSlashes`, `isLocalAddress`, `normalizeHostInput`,
  `isValidHostUrl`, `inferKindFromHost`), moved verbatim.
- Extract the agent-server probes into
  `src/api/backend-registry/backend-connection-service.ts`, where a single
  private `getServerInfo()` is now the only place the backend-registry UI
  constructs a `ServerClient`.
- Consolidate the trailing-slash policy rather than duplicating it:
  `stripTrailingSlashes` is the one implementation, used by the form's richer
  `normalizeHostInput` and by `getAgentServerClientOptions`. The latter
  deliberately applies only the primitive — every host reaching it is already
  absolute, so re-inferring a scheme there would rewrite a caller-supplied
  override rather than normalise it. That reasoning is in the doc comment.
- Route `backend-version.tsx` through the same service. It duplicated the
  status row exactly — same `queryKey`, same probe, same 5s timeout, same
  `staleTime` — so the two shared a React Query cache entry while running
  separate copies of the code.
- Add 45 unit tests: the host matrix the issue asks for, plus RFC 1918
  boundaries, look-alike cloud domains, the connection service, and the two
  version badges.

No behaviour change. The extracted implementations are identical to the ones
removed; `normalizeHost` is renamed to `normalizeHostInput` only to distinguish
it from the shared primitive.

## Issue Number

Fixes #15535

## How to Test

1. `npm ci`
2. `npm run typecheck && npm run lint`
3. Targeted suites:
   `npx vitest run __tests__/api/backend-registry/ __tests__/components/backends/`
4. Full suite: `npm test`
5. End-to-end, exercising the moved code through the UI:
   `npm run dev:mock`, then backend selector → **Add Backend** →
   **Agent-server** → **Local**. **Connect** enables only when `isValidHostUrl`
   accepts the host (`hi host` is refused, `localhost:8000` is accepted), and the
   accepted value is stored via `normalizeHostInput`. Open **Manage Backends**
   and edit a local backend to see the version badge, which now resolves through
   `fetchAgentServerVersion`.

Not run locally: `npm run test:e2e:mock-llm`. It needs Playwright browsers plus
a live agent-server, and `live-e2e` is skipped for fork PRs by design in
`ci.yml`. The mock-LLM add-backend flows touch `add-backend-host` /
`add-backend-submit`, whose behaviour is covered by the unit suites above.

## Video/Screenshots

Host validation exercised through the UI in `npm run dev:mock`, via
**Add Backend → Agent-server → Local**. This is the code path the PR moves:
`isValidHostUrl` decides whether **Connect** is enabled, and `normalizeHostInput`
is what turns the accepted value into an absolute URL.

**1. Whitespace is rejected** — `hi host` contains a space, so `isValidHostUrl`
returns `false` and **Connect** stays disabled.

oh-15535-1-host-with-space-rejected-connect-disabled

**2. A bare hostname is accepted** — `hihost` is a single-label name (no dots, no
colons), which `isLocalAddress` treats as a local machine name like `nas` or
`raspberrypi`. It normalises to `http://hihost` — `http`, not `https`, because
the host is local — so **Connect** enables. Unchanged from before this PR; the
pair is included because one screenshot alone cannot show that validation is
actually running rather than always returning true.

oh-15535-2-hostname-accepted-connect-enabled

**3. Test run** — `npm run lint` clean, and `npx vitest run backend` covering the
23 backend-related suites: 245 tests pass on Node 24.15, matching the `ci.yml`
pin. Branch name and Node version are visible in the prompt.

openhands-15535-vitest-245-passed-node24 15

## Type

- [ ] Bug fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

Deliberately left alone, as out of scope for an extraction — each is a
behaviour change, not a relocation:

- **`isValidHostUrl("ftp://example.com")` returns `true`**, normalising to
  `http://ftp://example.com`, because a non-http scheme is read as a
  single-label local hostname. Pinned in a test so a later fix has to change
  the assertion deliberately rather than silently.
- **Unbracketed IPv6 is rejected** — `::1` normalises to `http://::1`, which
  the URL parser refuses; `[::1]:8080` works. Also pinned.
- **A third host-normalisation policy exists** in
  `agent-server-config.ts::normalizeBaseUrl`, which strips one trailing slash
  (`/\/$/`) rather than all (`/\/+$/`) and infers scheme from
  `window.location.protocol`. It is reachable only from build-time config, so
  it is not part of this seam, but it is worth knowing it exists.
- **Two `ServerClient` sites remain outside the backend registry** —
  `useBackendsHealth` (4s timeout, checks settings before server info) and
  `loadAgentServerInfo` (cached bootstrap probe that also validates the session
  key). Their semantics genuinely differ, so folding them in would change
  behaviour; the service doc comment says so explicitly rather than implying
  they were missed.

Happy to split this into two PRs (host module, then connection service) if that
reviews better.


ameyypawar:
`test-and-build (ubuntu)` is red on a pre-existing flake, not on anything in this PR — vitest reports `Test Files 580 passed | 1 skipped` and `Tests 4649 passed`, alongside `Errors 1 error`: an unhandled `ReferenceError: ProgressEvent is not defined` thrown from msw's XHR interceptor, originating in `__tests__/components/features/home/home-chat-launcher.test.tsx`.

The same failure hits `main` — runs 31503801027, 31397412728 and 31393762554, each with every test passing. Lint, `test-and-build (windows)` and `mock-llm-e2e` are green here.

---
*This comment was generated by an AI assistant.*

PR #16534, opened 2026-08-11
fix(context-meter): draw the ring track from the foreground, not a border token

### LINKED ISSUE #16532: Context window ring track disappears under the trigger's hover fill
### Operating System

Linux

### Installation Method

From source (`npm run dev`)

### Agent Canvas Version

1.12.0

### Bug Description

The context window ring beside the send button loses its track on hover, so the control reads as a blank disc with a single dot on it rather than as a meter.

`context-window-ring.tsx` strokes the ring's unfilled track with `var(--oh-border)`. The trigger button in `context-window-meter.tsx` paints `hover:bg-[var(--oh-interactive-hover)]` behind it. Both tokens resolve to `var(--cool-grey-700)`, so on hover a 32px disc is painted in exactly the colour of the 16px ring's track. Contrast is 1.00:1. The track does not dim, it stops existing, and only the tone arc is left. At low usage that arc is a 2px dot.

Because the collision is between two tokens that both point at the same scale stop, it holds in every colour theme rather than in one palette. Read out of a running app on the default `openhands-neutral` theme, `--oh-border` and `--oh-interactive-hover` are both `#404040`. On `openhands-deepsea` they are both `#4B5468`.

Separately, this button is the only icon trigger in the chat-input actions row that does not use the shared `chatInputIconButtonClassName`. Its hover token carries the comment `/* hover on mid surfaces (800 bg) */`, but the composer renders on `var(--oh-surface)`, which is the 925 stop.

Raised by @DevinVinson in Slack against the meter added in #16311.

### Steps to Reproduce

1. Start a conversation so the context window meter renders in the composer.
2. Look at the small circular meter to the left of the send button while context usage is low.
3. Hover the meter.

### Actual Behavior

Reproduced on a local build run with `npm run dev:static`, on the default `openhands-neutral` palette.

On hover the ring track disappears into the button's hover fill, because they are the same colour. Read live from the running app, the trigger's hover fill and the track both compute to `#404040`.

Hovered, on `main`. The ring is gone and only the arc survives:

The track is also faint before hover: on the default theme it sits at 1.57:1 against the composer surface, under the 3:1 that WCAG 2.1 SC 1.4.11 asks of non-text contrast, and hover takes it to 1.00:1. At rest:

The popover's usage bar uses the same `--oh-border` token for its track and is similarly faint against its own backdrop.

### Screenshots

The same two states with the track drawn from the foreground instead of a border token, for comparison. Same conversation, same stack, same palette; only the built commit differs.

Hovered:

At rest:

### Expected Behavior

The track stays visible in both states, so the control still reads as a meter on hover and the filled proportion remains legible.

### Acceptance Criteria

- [ ] The ring track is distinguishable from the trigger's hover fill, and from the composer surface at rest, in every shipped colour theme.
- [ ] The filled arc stays distinguishable from the track at neutral usage.
- [ ] The trigger uses the same hover treatment as the other icon buttons in that row.

### Notes

Happy to take this once it is labeled. Worth flagging one thing for whoever picks it up, because it makes the fix less trivial than it first looks.

The obvious repair is to point the track at a lighter existing token. That does not hold across themes. `--oh-text-dim` clears 3:1 against the hover fill on `openhands-deepsea` but not on `openhands-neutral` or `openhands-neo`, and the tokens bright enough to clear it everywhere collapse the contrast between the arc and the track instead. The `warning` and `danger` arcs are also raw Tailwind hexes rather than tokens, which narrows the room further.

So a durable fix probably wants a track defined relative to the foreground rather than pinned to a scale stop, so it composites over whatever surface the active theme paints. Flagging it here rather than asserting it, since the design system is yours.

georgeglarson:
HUMAN:

confirmed working in my browser

AGENT:

Implementation and verification are complete. #16532 carries `ready-for-dev`.

The branch was force-pushed shortly after this PR was opened so the diff matches this description; the earlier red CI run was against the superseded commit and the current head is green.

Verified end to end in a running app, not only unit tests. Full local stack from source, `npm run dev:static` so the assets are the production build, against a real conversation on the default `openhands-neutral` palette. The same stack was rebuilt from `main` and from this branch with nothing else changed, and both states were captured below. Every colour below was read out of the live app with `getComputedStyle`, not derived from a stylesheet.

---

## Why

The context window ring loses its track on hover, so at low usage the control reads as a blank disc with a dot on it rather than a meter.

The collision is the obvious part: the track is stroked `var(--oh-border)` and the trigger paints `hover:bg-[var(--oh-interactive-hover)]` behind it, and both dereference `--cool-grey-700`. But that is the amplifier, not the cause. Give the button the hover token it should have had, so the two are no longer equal, and the track is still 1.17:1 on the default theme. Move it a stop lighter as well and it is 1.65:1. Every stop in the surface family lands between 1.07 and 1.65:1. It also failed before hover was involved, at 1.57:1 against the composer at rest.

The cause is that the track is an information-bearing element coloured out of the surface family. The arc alone says nothing; its proportion is only readable against the track, which makes the track foreground. Every stop in the border and surface family sits close to the surfaces it delimits, and a hover state layer is a lightened surface, so nothing drawn from that family can be reliably distinguishable from it.

## Summary

- The track is derived from `--oh-foreground` and composited over whatever the active theme paints, so it stays between surface and arc by construction rather than by a value that happens to work. Shared by the ring and the popover's usage bar, which had the same defect.
- The trigger uses `chatInputIconButtonClassName`, which every other icon button in that row already uses. Its old hover token is documented for 800-level surfaces while the composer renders on `--oh-surface`. That is a real defect on its own and not why the track vanished.
- Tests read the scales from `color-themes.ts` and assert contrast per theme, parsing the alpha back out of what the component renders so they describe the shipped value rather than a constant that can drift from it.

## Issue Number

Fixes #16532

## How to Test

`npm install && npm run dev`, start a conversation, hover the circular meter left of the send button at low context usage. The ring should stay visible under the hover fill.

One caveat worth knowing: under `npm run dev` the trigger's hover fill does not paint at all, on this branch or on `main`. Tailwind's preflight zeroes `button` backgrounds unlayered while the hover utility sits in `@layer utilities`, and unlayered wins regardless of specificity. The production build resolves the pair the other way. So hover styling has to be checked with `npm run dev:static` or a real build to see what ships. That is a separate defect and I will report it on its own.

Unit: `npm test -- context-window`. To watch the assertions catch the original bug, point the track at a scale stop instead; eleven cases fail.

Full run on this branch: lint clean with 0 errors, 4617 tests passing across 580 files, `npm run build` succeeds.

## Video/Screenshots

Captured from a running instance on the default `openhands-neutral` palette, production build, real `:hover`. Same conversation and same stack throughout; within each pair the only difference is which commit built the bundle.

Hovered, on `main`. The hover fill and the track are the same colour, so the ring is gone and only the arc survives:

Hovered, on this branch:

At rest, before hover is involved, the track is already faint at 1.57:1 against the composer surface. `main` first: 

then this branch:

## Type

- [x] Bug fix

## Notes

Measured on the running production build, default palette:

| | before | after |
|---|---|---|
| track vs composer surface | 1.57:1 | 3.59:1 |
| track vs trigger hover fill | 1.00:1 | 3.14:1 |
| neutral arc vs track, at rest | | 3.84:1 |
| neutral arc vs track, hovered | | 3.25:1 |

Why derived rather than a different token: `color-themes.ts` ships three scales and `DEFAULT_COLOR_THEME` is `openhands-neutral`, whose values differ from the deepsea scale hardcoded in `index.css`. No single stop satisfies all of them. `--oh-text-dim` clears 3:1 on deepsea and misses on the default at 2.55:1, and the stops bright enough everywhere collapse arc-versus-track instead. Worth flagging separately that `index.css` is overridden at runtime by the theme system, so it is not a sound source of truth for what an install renders.

Left alone deliberately: the `warning` and `danger` arcs are raw Tailwind hexes rather than tokens, so they cannot participate in theming, and no track colour clears 3:1 against both of those and both backdrops at once. That is a visual design decision for the people who own the palette rather than mine to make, and fixing the track does not depend on it. Happy to take it as a follow-up if you want the tone colours tokenised.

### DIFF
--- a/__tests__/components/features/chat/components/context-window-meter.test.tsx
+++ b/__tests__/components/features/chat/components/context-window-meter.test.tsx
@@ -96,6 +96,32 @@ describe("ContextWindowMeter", () => {
     );
   });
 
+  // The trigger must share the actions-row icon-button hover fill. Its own
+  // `--oh-interactive-hover` fill was both wrong for this surface (that token
+  // is for 800 backgrounds; the composer is `--oh-surface`) and identical in
+  // value to `--oh-border`, which the ring track used, so hovering erased the
+  // track. `context-window-ring.test.tsx` asserts the resulting contrast.
+  it("uses the shared chat-input icon button styling for its hover fill", () => {
+    useMetricsStore.setState({
+      cost: null,
+      max_budget_per_task: null,
+      usage: {
+        prompt_tokens: 0,
+        completion_tokens: 0,
+        cache_read_tokens: 0,
+        cache_write_tokens: 0,
+        context_window: 1_000_000,
+        per_turn_token: 198_500,
+      },
+    });
+
+    renderWithProviders(<ContextWindowMeter />);
+
+    const trigger = screen.getByTestId("context-window-meter");
+    expect(trigger).toHaveClass("hover:bg-white/10");
+    expect(trigger.className).not.toContain("--oh-interactive-hover");
+  });
+
   it("opens a popover with compact usage details", () => {
     useMetricsStore.setState({
       cost: null,
--- a/__tests__/components/features/chat/components/context-window-ring.test.tsx
+++ b/__tests__/components/features/chat/components/context-window-ring.test.tsx
@@ -1,0 +1,136 @@
+import { render, screen } from "@testing-library/react";
+import { describe, expect, it } from "vitest";
+import {
+  ContextWindowRing,
+  CONTEXT_WINDOW_RING_TRACK_ALPHA,
+} from "#/components/features/chat/components/context-window-ring";
+import { COLOR_THEMES } from "#/themes/color-themes";
+
+/**
+ * WCAG 2.1 SC 1.4.11 asks 3:1 for non-text contrast. The ring's track carries
+ * information (the arc's proportion is only readable against it), so it is held
+ * to that bar rather than treated as chrome.
+ */
+const MIN_RATIO = 3;
+
+/**
+ * Themes are read from `color-themes.ts`, not from a stylesheet. `index.css`
+ * ships the deepsea scale but `DEFAULT_COLOR_THEME` overrides it at runtime, so
+ * a stylesheet-derived assertion validates a palette no default install renders.
+ */
+const SURFACE_STOP = "--cool-grey-925"; // --oh-surface, the composer background
+const FOREGROUND_STOP = "--cool-grey-100"; // --oh-foreground, the neutral arc
+
+/** `chatInputIconButtonClassName` fills the trigger with `hover:bg-white/10`. */
+const HOVER_OVERLAY = "#FFFFFF";
+const HOVER_ALPHA = 0.1;
+
+function channels(hex: string): [number, number, number] {
+  const h = hex.replace("#", "");
+  return [0, 2, 4].map((i) => parseInt(h.slice(i, i + 2), 16)) as [
+    number,
+    number,
+    number,
+  ];
+}
+
+function relativeLuminance(hex: string): number {
+  const [r, g, b] = channels(hex).map((channel) => {
+    const c = channel / 255;
+    return c <= 0.04045 ? c / 12.92 : ((c + 0.055) / 1.055) ** 2.4;
+  });
+  return 0.2126 * r + 0.7152 * g + 0.0722 * b;
+}
+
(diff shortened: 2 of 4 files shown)


DevinVinson (approved):
Thank you!

PR #16545, opened 2026-08-12
fix(mcp): relay the OAuth callback when the redirect cannot reach the backend

### LINKED ISSUE #15430: OAuth MCP callback fails in Docker: localhost redirect port not exposed
## Bug Description

When using agent-canvas in a Docker container, the OAuth callback server on localhost is not exposed by default, causing OAuth MCP configurations to fail.

## Steps to Reproduce

1. Go to **Customize > MCP > Add MCP Server**
2. Configure an HTTP server with OAuth authentication (e.g., GitLab MCP server: https://docs.gitlab.com/user/model_context_protocol/mcp_server/#connect-cursor-to-the-gitlab-mcp-server)
3. A new window opens navigating to the OAuth provider (e.g., gitlab.com) to accept authorization
4. The redirect goes to a dynamic localhost port like `http://localhost:55607/callback?code=2716`
5. This redirect fails because the Docker container does not expose this dynamic callback port

## Expected Behavior

The OAuth callback should complete successfully and return control to agent-canvas.

## Actual Behavior

The callback URL `http://localhost:<dynamic_port>/callback?code=...` is unreachable because:
- The Docker container only exposes the main application port (e.g., 8000)
- The OAuth callback server binds to a random dynamic port
- No mechanism exists to proxy or expose this callback port from within the container

## Suggested Fixes

1. **Expose a fixed callback port**: Reserve a specific port (e.g., 18002) for OAuth callbacks and proxy it through the ingress
2. **Use loopback-aware callback URL**: Configure OAuth providers to use `127.0.0.1` instead of `localhost` and map the port via Docker
3. **Self-contained OAuth flow**: Implement a callback handler that receives the OAuth redirect and relays the code back to the MCP configuration without requiring a separate port

## Environment

- Running agent-canvas via Docker container
- Adding OAuth-based MCP server configuration

SimiPrambos:
HUMAN:
I reproduced this in Docker with only `-p 8000:8000` published, using Attlasian Rovo.
The OAuth redirect failed with `ERR_CONNECTION_REFUSED` exactly as described in the
issue #15430, because the callback port lives inside the container. With this change I
copied the URL from the failed tab, pasted it into the "Redirect didn't work?"
field, and the flow completed, the server saved and mcp connected.

AGENT:

Implementation was done with Claude Code; the end-to-end Docker verification
against a live OAuth provider was done by the human contributor — see the HUMAN
section and Video/Screenshots.

Automated verification on this branch:

```
npm run lint       # pass
npm test           # 562 files, 4502 passed / 5 skipped / 9 todo
npm run build      # pass
npm run build:lib  # pass
```

---

## Why

MCP OAuth cannot complete whenever the browser can't reach the agent-server's
loopback interface. The agent-server drives the handshake through FastMCP,
whose callback handler starts an HTTP listener on a **random** port and
registers `http://localhost:<port>/callback` as the redirect URI.

In Docker that listener lives inside the container while the OAuth provider
redirects the **host** browser to it, and the image publishes only port 8000 —
so the redirect dies with `ERR_CONNECTION_REFUSED`, the authorization code
never reaches the agent-server, and the status poll times out. The same failure
hits any deployment where the browser can't reach that loopback interface,
including remote and SSH-tunnelled backends.

The agent-server already supports an out-of-band relay:
`POST /api/mcp/oauth/callback/{job_id}` accepts a callback URL and fetches it
from *inside* its own network namespace, where the port is reachable.
`McpService.submitOAuthCallback` already wrapped that endpoint — but nothing in
the UI ever called it. This PR supplies the missing browser half.

The browser can't read the failed popup's URL on its own: it's cross-origin,
and after a connection error there's no readable document. A manual copy is the
only workable trigger.

## Summary

- `McpService.authorizeOAuth` accepts an optional `onJobStarted` callback that
  hands the OAuth probe job id to the UI; the return type is unchanged, so the
  existing single-argument call sites are unaffected.
- New `OauthCallbackFallback` — a collapsed "Redirect didn't work?" disclosure
  that relays a pasted callback URL through the existing endpoint. Wired into
  the add/edit modal, the install modal, and the reauthorize action (token
  expiry is the most common way an existing install hits this).
- The interactive wait gets its own 300s budget so a hand-pasted URL fits, and
  the add/edit modal is no longer undismissable while that wait runs.

## Issue Number

Closes #15430

## How to Test

1. Build and run the Docker image publishing only `-p 8000:8000`.
2. Customize → MCP → Add MCP Server; configure an HTTP server with OAuth
   (the GitLab MCP server is the reported case).
3. Authorize in the popup. The redirect still fails with
   `ERR_CONNECTION_REFUSED` — that's the environment, not the bug.
4. Copy the failed tab's full URL, expand "Redirect didn't work?", paste, submit.
5. The flow completes and the server is saved with its OAuth state.

Without this PR, step 4 has no UI at all and the flow dead-ends.

## Video/Screenshots
https://github.com/user-attachments/assets/017aca96-dee8-4596-b734-e9549200443d

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**No new SSRF surface.** The relayed URL is fetched server-side, but the
agent-server's `_validate_callback_url` already pins the scheme to `http`, the
host to loopback, and both port and path to the values that job itself
generated. Client-side validation is deliberately looser than the server's
(scheme + loopback host only): pinning the path or requiring a `code` param in
the browser would assert knowledge the browser doesn't have, and would reject
provider *denial* redirects (`?error=access_denied`) that are useful to relay so
the backend can fail the job immediately instead of burning the full wait.

**The 300s wait** matches the agent-server's own `callback_timeout` in
`software-agent-sdk`
(`openhands-agent-server/openhands/agent_server/mcp_router.py`,
`_BrowserCoordinatedOAuth.callback_handler`). It is a **separate** constant:
`OAUTH_MCP_TEST_TIMEOUT_SECONDS` is untouched because it also sets HTTP client
timeouts. Raising the interactive wait also exposed that the add/edit modal
blocked Escape, backdrop click and its close button for the whole duration, so
that modal is now dismissable while OAuth is pending — matching what the install
modal already did.

**Error text.** The shared `retrieveAxiosErrorMessage` reads `error`/`message`
but not FastAPI's `detail`, so backend rejections surfaced as "Request failed".
A small file-local helper in the new component reads `detail` first and falls
back to the shared util. Kept local since no other module needs it today;
happy to promote it to `src/utils/type-guards.ts` if you'd prefer.

**Translations.** The two new prose keys were machine-translated into the 14
non-English locales. Corrections from native speakers are very welcome — they
are `MCP$OAUTH_CALLBACK_FALLBACK_LABEL` and `MCP$OAUTH_CALLBACK_INVALID_URL`.
The URL placeholder is a technical example and is allowlisted in
`IDENTICAL_VALUE_ALLOWLIST`, matching the existing MCP OAuth placeholders.

**Not addressed here:** making the agent-server bind a fixed callback port. That
lives in `software-agent-sdk`, needs a coordinated release, and would still fail
for anyone who doesn't publish the port — so this fix is deliberately
frontend-only and works for every unreachable-loopback deployment.

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.54s · commit 4181c36  
**Coverage:** ⚠️ reduced context — partial coverage; 30/33 hunks, 15/15 files (context budget: 29, hunk budget: 3).

<details>

| --- | --- | --- |
| Command injection | 3.0% | No direct hunk selected |
| Weakened authentication | 9.0% | No direct hunk selected |
| Weakened authorization | 13.0% | No direct hunk selected |
| Contract regression | 11.0% | No direct hunk selected |
| Data loss | 5.0% | No direct hunk selected |
| Sensitive data disclosure | 8.0% | No direct hunk selected |
| Unexpected data transfer | 6.0% | No direct hunk selected |
| Credential misuse | 9.0% | No direct hunk selected |
| Security assessment bypass | 8.0% | No direct hunk selected |
| Prohibited workload | 3.0% | No direct hunk selected |
| Primary concern | None selected; confidence 69.0% | No primary concern to locate |

</details>


VascoSch92:
@SimiPrambos Nice work :-) 

 Two things I noticed while testing it:

1. Closing the modal mid-OAuth doesn't actually abandon the flow

The new comment on ModalBackdrop says there's nothing to orphan beyond the backend job, but that isn't quite true. authorizeOAuth has no abort, so after the modal closes the poll keeps going. If the user then finishes consent in the popup anyway, the .then in handleSubmit still calls addMcpServer / updateMcpServer, and the server gets saved even though they walked away. The editor has unmounted by then, so if that save fails the user never sees an error. React Query skips the per-call onError once the observer has no listeners, and the hooks don't define their own.

To be fair, the Cancel button could already do this before the PR (it's only disabled on isPending). This PR adds Escape, backdrop click and the close button on top, though, and the comment makes it look intentional. I think we should either cancel the flow when the modal closes (an abort signal into authorizeOAuth, or at least ignore the result once it's unmounted), or keep dismiss blocked while isOauthTesting is true.

2. The backend's detail message never shows up in the fallback

extractCallbackErrorMessage checks error instanceof AxiosError, but submitOAuthCallback goes through the typescript-client MCPClient, which uses fetch and throws HttpError. On that error the body is at error.response.detail, not error.response.data.detail. So a 400 like "Unexpected OAuth callback URL" ends up displayed as the raw HTTP request failed (400 Bad Request): {"detail":"Unexpected OAuth callback URL"}. The test passes because it mocks an AxiosError, which this client never throws. There's already an isSdkHttpError helper in src/api/agent-server-compatibility.ts that should work here, and the test should mock an HttpError instead.

Issue #16513, opened 2026-08-11
[Bug]: Issue Readiness Check workflow fails on every not-ready issue (no feedback comment posted)


Lothnic:
### Bug Description

The `Issue Readiness Check` workflow **fails on every issue that is not yet ready-for-dev**, so the "here's what's missing" feedback comment is never posted and the `ready-for-dev` remove-label path never runs.

Root cause: `.github/scripts/check_issue_readiness.py` exits with code `1` when an issue does **not** meet the criteria (`sys.exit(main())` → `return 0 if result.ready else 1`). But the workflow runs the script under `set -euo pipefail` in `.github/workflows/issue-readiness-check.yml`:

```yaml
run: |
    set -euo pipefail
    python .github/scripts/check_issue_readiness.py --json > /tmp/result.json
```

So the "not ready" state — the most common state (a just-opened issue, or one missing a screenshot) — exits nonzero and the `Evaluate issue readiness` step fails, skipping the label management and the feedback comment entirely.

### Steps to Reproduce

1. Open any issue on this repo that does not have the `bug` or `enhancement` label (or is missing a screenshot / acceptance criteria).
2. Watch the `Issue Readiness Check` workflow run.
3. Observe: the `Evaluate issue readiness` step fails with `Process completed with exit code 1.` and no comment is posted to the issue.

Examples of failing runs: run `31466291228` (issue #16511), `31451127961` / `31451127615` (the ProgressEvent bug issue).

### Actual Behavior

Run method: `npm run` (the readiness scripts are part of this repo)

The workflow run is marked **failed** and the author receives no feedback comment telling them what is missing (e.g. "add a screenshot to Actual Behavior"), even though the workflow was explicitly designed to post such a comment on opened/reopened issues.

### Expected Behavior

A not-ready issue should be a normal, handled state: the workflow should evaluate, post the "what's missing" comment, and (if applicable) remove the `ready-for-dev` label — with the run ending **successfully** or at least completing the intended side effects.

### Relevant Logs

```
check  Evaluate issue readiness
  set -euo pipefail
  python .github/scripts/check_issue_readiness.py --json > /tmp/result.json
##[error]Process completed with exit code 1.
```

### Acceptance Criteria

- [ ] Opening an issue that is not ready-for-dev results in a feedback comment telling the author what is missing.
- [ ] Opening an issue that is not ready-for-dev does **not** mark the workflow run as failed.
[...]

PR #16556, opened 2026-08-12
fix(ci): ignore markdown headings inside fenced code blocks

### LINKED ISSUE #16553: [Bug]: Headings inside fenced code blocks are parsed as real sections by the CI readiness checkers
### Operating System

macOS

### Installation Method

### Agent Canvas Version

main (1.12.0)

### Bug Description

`extract_sections()` in `.github/scripts/check_issue_readiness.py` splits an issue body on `^###\s+` with no awareness of fenced code blocks, so a heading written inside a fence is treated as a real section boundary. `.github/scripts/check_pr_description.py` has the same function against `^##\s+` and the same defect.

This cuts both ways.

**A quoted heading grants readiness it should not.** An issue whose only `### Acceptance Criteria` appears inside a fence — for example when quoting the issue template — passes the gate despite having no real acceptance criteria.

**A fenced heading inside a real section silently truncates it.** Because the next `###` match ends the previous section, a bug report that pastes a log or markdown snippet containing a `###` line loses everything after that fence. If the screenshot sits below the paste, the checker never sees it and rejects a valid report.

The second case is the damaging one: the contributor is told their report is not ready when it is, and the reason given points at content they can plainly see is present. The bug-report template explicitly asks reporters to paste logs into Actual Behavior, so this is a natural thing to hit.

### Steps to Reproduce

From a source checkout, run against the repo's own script.

1. Save this as `repro.py` in the repo root:

```python
import sys; sys.path.insert(0, ".github/scripts")
from check_issue_readiness import evaluate_readiness, extract_sections

F = "`" * 3  # a fence, built at runtime to keep this snippet quotable

no_real_ac = f"""### Desired Behavior

Make the thing work.

### Notes

The template asks for:

{F}markdown
### Acceptance Criteria
- [ ] checklist items go here
{F}
"""

valid_report = f"""### Actual Behavior

Run method: `npm run dev`

The server returned:

{F}
### Error detail
something went wrong
{F}

<img width="800" alt="Image" src="https://github.com/user-attachments/assets/abc123" />

### Acceptance Criteria

- [ ] the bug is fixed
"""

r1 = evaluate_readiness(no_real_ac, ["enhancement"])
print("[1] sections:", ", ".join(extract_sections(no_real_ac)))
print("[1] ready   :", r1.ready, "<-- should be False")

r2 = evaluate_readiness(valid_report, ["bug"])
print("[2] sections:", ", ".join(extract_sections(valid_report)))
print("[2] ready   :", r2.ready, "<-- should be True")
print("[2] reasons :", r2.reasons)
```

2. `python3 repro.py`

`[1]` reports ready despite having no acceptance criteria — the only such heading is inside a fence. `[2]` is rejected despite the run method and screenshot both being present in Actual Behavior, because the fenced `### Error detail` line ends the section before them.

### Actual Behavior

Run method: `npm run dev` source checkout — the defect is in the repo's own CI scripts, reproduced directly against them with the snippet above.

`extract_sections()` reports `acceptance criteria` as a parsed section when that heading exists only inside a fenced block, and drops the tail of `### Actual Behavior` when a fence inside it contains a `###` line. Neither script skips fenced regions:

```
check_issue_readiness.py:42   HEADING_RE = re.compile(r"(?m)^###\s+(.+?)\s*$")
check_issue_readiness.py:110  matches = list(HEADING_RE.finditer(body))
```

There is no fence handling anywhere in either parser, and none of the 26 tests in `.github/scripts/tests/test_issue_readiness.py` exercise a fenced code block.

Output of the reproduction:

Image

### Expected Behavior

Headings inside fenced code blocks are not treated as section boundaries, so quoting the template does not manufacture a section, and pasting a log into Actual Behavior does not truncate it.

### Relevant Logs

```shell
.github/scripts/check_issue_readiness.py
  42:  HEADING_RE = re.compile(r"(?m)^###\s+(.+?)\s*$")
 110:  matches = list(HEADING_RE.finditer(body))

.github/scripts/check_pr_description.py
  70:  HEADING_RE = re.compile(r"(?m)^##\s+(.+?)\s*$")

Sibling script, same root cause — "## How to Test" exists only inside a fence
yet satisfies the required-section check:

  >>> import check_pr_description as p
  >>> list(p.extract_sections(body).keys())
  ['Why', 'Summary', 'How to Test', 'Issue Number']
```

### Acceptance Criteria

- [ ] A `###` heading inside a fenced code block is not parsed as a section by `check_issue_readiness.py`.
- [ ] An issue whose only Acceptance Criteria heading is inside a fence is not granted `ready-for-dev`.
- [ ] A bug report whose Actual Behavior contains a fenced block with a `###` line is still evaluated against the full section, including a screenshot placed after the fence.
- [ ] The same fence handling applies to `##` headings in `check_pr_description.py`.
- [ ] Tests in `.github/scripts/tests/` cover fenced headings in both directions — a spurious section and a truncated one.

### Screenshots

_No response_

### Additional Context

This issue body demonstrates the defect on itself. Running `extract_sections()` over the draft of it yields two sections that do not exist — `notes` and `error detail` — both lifted from inside the fenced snippets above:

```
sections parsed: operating system, installation method, agent canvas version,
                 bug description, steps to reproduce, notes, acceptance criteria,
                 error detail, actual behavior, expected behavior, relevant logs,
                 additional context
```

It still evaluates correctly here only because the real `### Acceptance Criteria` appears later than the fenced one and overwrites it in the section dict. Reorder the body and it would not.

Related but distinct: #16513 covers the workflow failing on not-ready issues via the script's exit code. This is a parsing defect in the same file and is independent of that.

Both scripts arrived recently in #16449, so this is unlikely to have affected many issues yet — but quoting the issue template inside a fence is a natural thing for a reporter to do, and pasting logs into Actual Behavior is exactly what the template asks for.

ameyypawar:
HUMAN:

Ran the tests on Python 3.13 to match CI: 13 fail before the fix, 86 pass after. No npm suite — this PR only touches `.github/` files.

AGENT:

Validated locally on Python 3.13, matching `ci-script-tests.yml`:

- `python -m pytest .github/scripts/tests/ -q` — **86 passed** (58 before this change).
- Backing the fix out of a copy of the scripts and re-running the same suite — **13 failed, 73 passed**. Those 13 are the new tests; they fail against the current parser and pass with the change.
- Also green on Python 3.12.

Checked for regressions against real data rather than fixtures alone. Ran the old and new `check_pr_description.py` over the bodies of **all 120 open pull requests** in this repository and compared results:

- section sets: **0 differences**
- `validate_pr_body()` results: **0 differences**
- 42 sections differ in leading whitespace only, with no content difference

And against the issue this fixes: #16553's own body currently parses into two sections that do not exist — `notes` and `error detail` — both lifted out of its fenced snippets. With the change it parses into exactly its eleven real template sections and still evaluates as ready.

---

## Why

`extract_sections()` splits issue and PR bodies on `###` / `##` with no awareness of fenced code blocks, so a heading written inside a fence becomes a section boundary. That breaks the readiness and description gates in both directions.

A heading quoted inside a fence manufactures a section. An issue whose only `### Acceptance Criteria` sits in a quoted template earns `ready-for-dev` with no real acceptance criteria, and a PR whose only `## How to Test` sits in a fenced example satisfies the required-field check.

A fenced heading inside a real section truncates it. Because the next heading match ends the previous section, a bug report that pastes a log containing a `###` line loses everything after the fence — including a screenshot placed below it — and is rejected for missing evidence the reporter can plainly see. The bug-report template asks for exactly that paste, so this is a natural thing to hit.

## Summary

- Add `mask_fenced_code()` to both checkers. It blanks fenced lines while preserving length, so heading matches run against the mask and section content is still sliced out of the original body.
- Fences close only on their own marker character and a marker at least as long as the opener. An unterminated fence runs to the end of the body, matching CommonMark and GitHub's renderer.
- Split lines on `\n` alone. `str.splitlines()` also breaks on form feeds and other Unicode separators that CommonMark ignores, and pasted terminal output is where those appear.
- Section text now starts at the line after the heading rather than at `match.end()`. `HEADING_RE` ends in `\s*$`, which is greedy across newlines, so against the mask it would swallow a fenced block that *opens* a section — the exact failure this change sets out to fix.
- Add 28 tests covering both directions, a fence opening a section, tilde fences, CRLF bodies, form feeds, unterminated fences, longer closing markers, and offset preservation.

## Issue Number

Fixes #16553

## How to Test

1. `python -m pytest .github/scripts/tests/ -q` — expect 86 passed.
2. To see the failures this fixes, revert the two call sites and re-run:
   - `check_issue_readiness.py`: `HEADING_RE.finditer(mask_fenced_code(body))` → `HEADING_RE.finditer(body)`, and `start = <line after heading>` → `start = match.end()`
   - `check_pr_description.py`: the same two edits
   - re-running the suite then gives 13 failed, 73 passed
3. Spot-check the two behaviours directly:

```python
import sys; sys.path.insert(0, ".github/scripts")
from check_issue_readiness import evaluate_readiness

F = "`" * 3

quoted_only = f"""### Desired Behavior

Make it work.

{F}markdown
### Acceptance Criteria
- [ ] quoted from the template, not real
{F}
"""

log_first = f"""### Actual Behavior

{F}
$ npm run dev
Error: boom
{F}

![shot](https://github.com/user-attachments/assets/abc123)

### Acceptance Criteria
- [ ] fixed
"""

print(evaluate_readiness(quoted_only, ["enhancement"]).ready)  # False (was True)
print(evaluate_readiness(log_first, ["bug"]).ready)            # True  (was False)
```

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The helper is duplicated across the two scripts rather than shared. That follows the existing convention here — they already carry their own copies of the media patterns, `visible_text` and `HEADING_RE`, and `.github/scripts/` has no shared-module precedent and stdlib-only imports. Happy to extract a shared module instead if you would prefer that.

Known limitation, deliberately left out of scope: masking governs section boundaries only. The per-section content checks — `has_checklist_item`, `has_screenshot_or_video`, `references_run_method` — still read fenced text, so a quoted checklist item or image inside a real section continues to count toward that section's criteria. That is the same class of defect and worth a follow-up, but fixing it changes which issues qualify rather than only where sections begin and end, so it seemed better kept separate.

Related but distinct: #16513 covers the readiness workflow failing on not-ready issues via the script's exit code.

openhands-16553-tests-before-after

### DIFF
--- a/.github/scripts/check_issue_readiness.py
+++ b/.github/scripts/check_issue_readiness.py
@@ -41,6 +41,11 @@ ENHANCEMENT_LABEL = "enhancement"
 # h1/h2 are not produced by issue forms.
 HEADING_RE = re.compile(r"(?m)^###\s+(.+?)\s*$")
 
+# Opening/closing marker of a fenced code block: three or more backticks or
+# tildes, indented by at most three spaces (CommonMark). An info string may
+# follow the opening marker.
+FENCE_RE = re.compile(r"^[ \t]{0,3}(`{3,}|~{3,})")
+
 # `_No response_` is what GitHub writes for an empty optional form field.
 NO_RESPONSE = "_No response_"
 
@@ -100,17 +105,62 @@ def visible_text(text: str) -> str:
     return cleaned
 
 
+def mask_fenced_code(body: str) -> str:
+    """Return `body` with fenced code blocks blanked out, preserving offsets.
+
+    A `###` line inside a fence is not a heading: reporters quote the issue
+    template and paste logs, and GitHub renders neither as a section. Every
+    blanked character is replaced by a space and newlines are kept, so offsets
+    into the returned string are valid offsets into `body` — callers match
+    headings against the mask and slice content out of the original.
+
+    An unterminated fence runs to the end of the body, which is what CommonMark
+    (and GitHub's own renderer) does.
+    """
+    masked: list[str] = []
+    fence: str | None = None
+
+    # Split on "\n" only. `str.splitlines()` also breaks on form feeds and other
+    # Unicode separators that CommonMark does not treat as line endings, so a
+    # fence marker following one — common in pasted terminal output — would
+    # falsely open a block and mask the rest of the body.
+    for line in body.split("\n"):
+        probe = line[:-1] if line.endswith("\r") else line
+        match = FENCE_RE.match(probe)
+        marker = match.group(1) if match else None
+
+        if fence is None:
+            if marker is None:
+                masked.append(line)
+                continue
+            fence = marker
+        elif marker and marker[0] == fence[0] and len(marker) >= len(fence):
+            # Closing fence: same character, at least as long as the opener.
+            fence = None
+
+        masked.append(" " * len(line))
+
+    return "\n".join(masked)
+
+
 def extract_sections(body: str) -> dict[str, str]:
     """Split the body into a {heading: text} map using `### <heading>` boundaries.
 
     Issue forms render every field this way. Free-form issues (not created via a
     form) may still use `###` headings; if they don't, the map is empty and the
     caller falls back to whole-body checks.
+
+    Headings inside fenced code blocks are ignored — see `mask_fenced_code`.
     """
-    matches = list(HEADING_RE.finditer(body))
+    matches = list(HEADING_RE.finditer(mask_fenced_code(body)))
     sections: dict[str, str] = {}
     for index, match in enumerate(matches):
-        start = match.end()
+        # Start at the line after the heading rather than at `match.end()`.
+        # `HEADING_RE` ends in `\s*$`, which is greedy across newlines, so on
+        # the mask it would swallow a fenced block that opens the section and
+        # drop it from the returned text.
+        newline = body.find("\n", match.start())
+        start = len(body) if newline == -1 else newline + 1
         end = matches[index + 1].start() if index + 1 < len(matches) else len(body)
         sections[match.group(1).strip().lower()] = body[start:end]
     return sections
--- a/.github/scripts/check_issue_readiness.py
+++ b/.github/scripts/check_issue_readiness.py
@@ -41,10 +41,15 @@ ENHANCEMENT_LABEL = "enhancement"
 # h1/h2 are not produced by issue forms.
 HEADING_RE = re.compile(r"(?m)^###\s+(.+?)\s*$")
 
-# Opening/closing marker of a fenced code block: three or more backticks or
-# tildes, indented by at most three spaces (CommonMark). An info string may
-# follow the opening marker.
-FENCE_RE = re.compile(r"^[ \t]{0,3}(`{3,}|~{3,})")
+# CommonMark fenced code blocks. An opening fence is three or more backticks
+# or tildes indented by at most three *spaces* — a leading tab expands to
+# indented code, which is not a fence. Text after the marker is the info
+# string, and a backtick fence's info string may not contain a backtick.
+FENCE_OPEN_RE = re.compile(r"^ {0,3}(`{3,}|~{3,})(.*)$")
+# A closing fence repeats the opener's character at least as many times and
+# is followed by nothing but spaces or tabs; a marker with trailing text is
+# still code content.
+FENCE_CLOSE_RE = re.compile(r"^ {0,3}(`{3,}|~{3,})[ \t]*$")
 
 # `_No response_` is what GitHub writes for an empty optional form field.
 NO_RESPONSE = "_No response_"
@@ -126,17 +131,27 @@ def mask_fenced_code(body: str) -> str:
     # falsely open a block and mask the rest of the body.
     for line in body.split("\n"):
         probe = line[:-1] if line.endswith("\r") else line
-        match = FENCE_RE.match(probe)
-        marker = match.group(1) if match else None
 
         if fence is None:
-            if marker is None:
+            opener = FENCE_OPEN_RE.match(probe)
+            if opener is None:
+                masked.append(line)
+                continue
+            marker, info = opener.group(1), opener.group(2)
+            if marker[0] == "`" and "`" in info:
+                # A backtick fence's info string may not contain a backtick, so
+                # this is ordinary text rather than the start of a block.
                 masked.append(line)
                 continue
             fence = marker
-        elif marker and marker[0] == fence[0] and len(marker) >= len(fence):
-            # Closing fence: same character, at least as long as the opener.
-            fence = None
+        else:
+            closer = FENCE_CLOSE_RE.match(probe)
+            if (
+                closer
+                and closer.group(1)[0] == fence[0]
+                and len(closer.group(1)) >= len(fence)
+            ):
+                fence = None
 
         masked.append(" " * len(line))
 
--- a/.github/scripts/check_pr_description.py
+++ b/.github/scripts/check_pr_description.py
@@ -40,10 +40,15 @@ REQUIRED_TEMPLATE_FIELDS: tuple[str, ...] = ("Why", "Summary", "How to Test")
 
 HTML_COMMENT_RE = re.compile(r"<!--[\s\S]*?-->")
 HEADING_RE = re.compile(r"(?m)^##\s+(.+?)\s*$")
-# Opening/closing marker of a fenced code block: three or more backticks or
-# tildes, indented by at most three spaces (CommonMark). An info string may
-# follow the opening marker.
-FENCE_RE = re.compile(r"^[ \t]{0,3}(`{3,}|~{3,})")
+# CommonMark fenced code blocks. An opening fence is three or more backticks
+# or tildes indented by at most three *spaces* — a leading tab expands to
+# indented code, which is not a fence. Text after the marker is the info
+# string, and a backtick fence's info string may not contain a backtick.
+FENCE_OPEN_RE = re.compile(r"^ {0,3}(`{3,}|~{3,})(.*)$")
+# A closing fence repeats the opener's character at least as many times and
+# is followed by nothing but spaces or tabs; a marker with trailing text is
+# still code content.
+FENCE_CLOSE_RE = re.compile(r"^ {0,3}(`{3,}|~{3,})[ \t]*$")
 HUMAN_HEADING_RE = re.compile(r"(?im)^\s*HUMAN:\s*$")
 AGENT_HEADING_RE = re.compile(r"(?im)^\s*AGENT:\s*$")
 
@@ -154,17 +159,27 @@ def mask_fenced_code(body: str) -> str:
     # falsely open a block and mask the rest of the body.
     for line in body.split("\n"):
         probe = line[:-1] if line.endswith("\r") else line
-        match = FENCE_RE.match(probe)
-        marker = match.group(1) if match else None
 
         if fence is None:
-            if marker is None:
+            opener = FENCE_OPEN_RE.match(probe)
+            if opener is None:
+                masked.append(line)
+                continue
+            marker, info = opener.group(1), opener.group(2)
+            if marker[0] == "`" and "`" in info:
+                # A backtick fence's info string may not contain a backtick, so
+                # this is ordinary text rather than the start of a block.
                 masked.append(line)
                 continue
             fence = marker
-        elif marker and marker[0] == fence[0] and len(marker) >= len(fence):
-            # Closing fence: same character, at least as long as the opener.
-            fence = None
+        else:
+            closer = FENCE_CLOSE_RE.match(probe)
+            if (
+                closer
+                and closer.group(1)[0] == fence[0]
+                and len(closer.group(1)) >= len(fence)
+            ):
+                fence = None
 
         masked.append(" " * len(line))
 
--- a/.github/scripts/check_pr_description.py
+++ b/.github/scripts/check_pr_description.py
@@ -40,6 +40,10 @@ REQUIRED_TEMPLATE_FIELDS: tuple[str, ...] = ("Why", "Summary", "How to Test")
 
 HTML_COMMENT_RE = re.compile(r"<!--[\s\S]*?-->")
 HEADING_RE = re.compile(r"(?m)^##\s+(.+?)\s*$")
+# Opening/closing marker of a fenced code block: three or more backticks or
+# tildes, indented by at most three spaces (CommonMark). An info string may
+# follow the opening marker.
+FENCE_RE = re.compile(r"^[ \t]{0,3}(`{3,}|~{3,})")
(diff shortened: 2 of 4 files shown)


enyst (commented):
Hey, thank you for the PR! ~Could you please add the before/after screenshots for the example in the PR description?~ Sorry, I misunderstood, I see what you mean!

enyst:
@openhands Do a /codereview please and post your feedback on PR

.github/scripts/check_issue_readiness.py
```diff
@@ -100,17 +105,62 @@ def visible_text(text: str) -> str:
...
+        probe = line[:-1] if line.endswith("\r") else line
+        match = FENCE_RE.match(probe)
+        marker = match.group(1) if match else None
+
+        if fence is None:
+            if marker is None:
+                masked.append(line)
+                continue
+            fence = marker
+        elif marker and marker[0] == fence[0] and len(marker) >= len(fence):
```
enyst: **Correctness:** This closes on any same-character marker prefix, but GitHub/CommonMark only closes when the rest of the line is spaces or tabs. For example, after an opening ` ``` `, the line ` ```not a close` is code content; this scanner closes there and then treats a following `### Acceptance Criteria` as real. I reproduced an enhancement becoming ready from that fenced-only criterion. Require the closing marker to consume the line except optional spaces/tabs, and add the equivalent regression test to both checkers.

.github/scripts/check_pr_description.py
```diff
@@ -40,6 +40,10 @@
 
 HTML_COMMENT_RE = re.compile(r"<!--[\s\S]*?-->")
 HEADING_RE = re.compile(r"(?m)^##\s+(.+?)\s*$")
+# Opening/closing marker of a fenced code block: three or more backticks or
+# tildes, indented by at most three spaces (CommonMark). An info string may
+# follow the opening marker.
+FENCE_RE = re.compile(r"^[ \t]{0,3}(`{3,}|~{3,})")
```
enyst: **Correctness:** This accepts two invalid backtick openers that GitHub/CommonMark does not: a leading tab (tab expansion makes it indented code, not a fence) and an info string containing a backtick, e.g. ` ```lang\`bad`. In both cases the scanner opens a phantom unterminated fence and masks a real following `## How to Test`; I reproduced valid PR bodies being rejected. Limit opener indentation to 0–3 literal spaces and reject backticks in a backtick fence's info string. Please cover both inputs in each parser's tests.

enyst (commented):
🔴 **Needs improvement**

_This review was generated by an AI agent (OpenHands) on behalf of the requester._

**[CRITICAL ISSUES]**
- The fence scanner does not follow GitHub/CommonMark closing-fence syntax. A marker followed by text is treated as a close, so a heading that GitHub still renders inside code can manufacture a required section.
- The opener regex also accepts syntax GitHub/CommonMark does not: tab indentation and backticks in a backtick fence's info string. Either input starts a phantom unterminated fence and hides real headings that GitHub renders as sections.

I reproduced all three cases against both checkers. The normal suite is green (`python -m pytest .github/scripts/tests/ -q`: 86 passed), but the focused cases produce false readiness/validation decisions. Inline comments contain exact examples and the required parser constraints.

**[TESTING GAPS]**
- Add regression coverage in both test files for a closing marker with trailing non-whitespace, a tab-indented marker, and a backtick fence whose info string contains a backtick.

**KEY INSIGHT:**
Offset-preserving masking is the right structure, but the fence state transitions must use the renderer's opener/closer grammar or the CI parser will disagree with what contributors see on GitHub.

**[RISK ASSESSMENT]**
The change is isolated to CI scripts and adds no dependency or security exposure, but these scripts control issue readiness and PR-description acceptance; the parser mismatches can both admit invalid content and reject valid content.

**VERDICT:**
❌ **Needs rework**: Correct the fence grammar and add focused tests; a human maintainer can then decide approval.

---

>
>
>

---

xxiaoxiong (approved):
This is the right fix for the section-splitting half: a `###` inside a fenced code block is display, not a section boundary. The choice to *mask* (spaces-char-by-spaces, newlines preserved) rather than *strip* fenced code is correct for this code path — `extract_sections` then slices out content from the body by `match.start()`/`match.end()` offsets, so length-preservation keeps `body[start:end]` aligned. `strip_code_fences`-style removal would shift offsets and silently mis-slice.

Things that earned the approve:

1. CommonMark fidelity: backtick and tilde fences, opening marker indented up to three spaces (a leading tab is indented-code, not a fence, and the test for that is covered), closing marker must repeat the opener's character and be at least as long. `test_backticks_do_not_close_a_tilde_fence` and `test_longer_marker_closes_shorter_fence` pin those.

2. `split("\n")` rather than `splitlines()` — explicitly called out in the comment. CommonMark does not treat form feeds or other Unicode separators as line endings, and `splitlines()` does. A fence marker after a `\f` in pasted terminal output would falsely open a block and mask the rest of the body without this.

3. Offset-preserving masking is what makes the bug-fixed `extract_sections` slice `body[start:end]` from the *original* body — fence content is not real section content, and the slices stay aligned. Stripping the fence out before slicing would push subsequent sections off by the fence's length.

4. The greedy `\s*$` interaction is handled: the comment in the section loop is dead honest by example. `HEADING_RE` ends in `\s*$`, which is greedy across newlines, so on the mask it would swallow a fenced block that opens the section and drop it from the returned text. The fix recomputes `start` at the next newline after the heading. The `test_fenced_heading_does_not_truncate_enclosing_section` and `test_sections_after_a_fence_are_still_found` tests pin the not-truncation behaviour with three fenced-section blocks including a screenshot after the fence.

5. The info-string rule is obeyed: a backtick fence's info string may not itself contain a backtick, so a line like `\`\`\`python x = \`` is ordinary text — handled inline for both fence types.

6. The author's own regression sweep over all 120 open PR bodies — section sets 0, `validate_pr_body` results 0, 42 whitespace-only section-leading differences — is the kind of evidence that closes the question. Checking #16553's own body parses into its eleven real template sections after the fix is a satisfying end-to-end.

This is separable from #16583's body-text half (handled by #16598 with a separate `strip_code_fences` helper). Both should land together, but neither depends on the other.

Approved. CI is 9 SUCCESS, 8 SKIPPED, 0 FAILURE — the skipped ones are gated on a different base or secrets.

marmar9615-cloud:
I ran this branch against the live issue corpus, since "does this silently reclassify anything" seemed like the risk worth measuring. Sharing the numbers.

**Verdict-neutral on its own.** I ran `check_issue_readiness.py` from `main` and from this branch over 98 real issues (every open and closed issue labelled `bug` or `enhancement` in the 50 most recent of each, using their actual bodies and labels). Zero verdicts changed. Both test suites pass on the branch, 93 tests.

**It becomes load-bearing as soon as a non-screenshot evidence path exists.** With #16751's command-and-output check applied on top of this branch's parser, the same 98-issue sweep moves exactly 2 issues from not-ready to ready:

- **#16583**, your own report. Its Steps section embeds `### Acceptance Criteria` inside the fenced repro, so on `main` the section is cut mid-block and the opening fence is orphaned from its closer. The reproduction command then stops counting as fenced.
- **#16707**, mine, which fails the same way.

Both are non-UI bug reports carrying a real command and its real output, so both flips look correct to me rather than like a loosened gate.

One detail I appreciated only after walking into it: slicing from the line after the heading instead of from `match.end()`. I tried masking fences with spaces first, and `HEADING_RE`'s trailing `\s*$` ran across the blanked newlines and swallowed the masked block, so the section started *after* the fence rather than before it. Your comment names that hazard outright, which is what stopped me re-deriving it the hard way a second time.

One practical note: the branch currently conflicts with `main` in `.github/scripts/tests/test_pr_description.py`.

For disclosure, #16751 is mine and needs this merged before one of its linked issues can qualify, so I have an interest in it landing. The sweep above is reproducible from the two script versions if you want to confirm any of it.

marmar9615-cloud:
The `/codereview` findings look accurate, but they were made against `da0a565dfd` (submitted 2026-08-12). This branch is now at `c19e9c6a98`, and all three cases were fixed in between.

I reproduced each case against both checkers, at the reviewed commit and at the current head, by masking the body and asking which headings survive:

| case | reviewed `da0a565dfd` | current `c19e9c6a98` |
|---|---|---|
| closing marker with trailing text (heading must stay hidden) | heading became visible — bug | hidden — correct |
| tab-indented marker (heading must stay visible) | heading was masked — bug | visible — correct |
| backtick inside a backtick fence's info string (heading must stay visible) | heading was masked — bug | visible — correct |

Identical results in `check_issue_readiness.py` and `check_pr_description.py`, so six of six reproduce as described on the reviewed commit and six of six behave correctly on the current head. The current implementation added `FENCE_OPEN_RE` and `FENCE_CLOSE_RE`, which is what closes the gap: `^ {0,3}` admits only literal spaces, so a tab no longer opens a fence, and `[ \t]*$` on the closer means a marker with trailing text stays code content.

The requested regression coverage is also already there, by name in both test files: `test_closing_marker_with_trailing_text_does_not_close`, `test_tab_indented_marker_is_not_a_fence`, and `test_backtick_in_backtick_fence_info_string_is_not_an_opener`, plus `test_tilde_fence_info_string_may_contain_backticks` and `test_backticks_do_not_close_a_tilde_fence`.

So as far as I can measure, the only thing outstanding on this branch is the merge conflict in `.github/scripts/tests/test_pr_description.py`.

marmar9615-cloud:
Heads up on why this went red. #16595 landed on 2026-08-22 and covers the same problem. It moved fence handling into a new shared module, `.github/scripts/markdown_sections.py`, and both `check_issue_readiness.py` and `check_pr_description.py` now call into it instead of matching headings directly. Since that rewrote the same regions of both files your branch touches, it is the source of the conflict.

The merged module covers most of what you tested here. Tilde fences, indented openers, longer closing markers, info strings containing backticks, tab-indented markers, CRLF bodies, and offset-preserving masking all made it in.

One case differs, and I think it is worth raising rather than dropping. Your `test_unterminated_fence_masks_to_end_of_body` asserts that an unclosed fence hides every heading after it, matching how CommonMark and GitHub render it. The merged version deliberately does the opposite and leaves an unclosed fence untouched. The docstring gives the reasoning. These scripts gate contributions, so one stray marker in a pasted log would hide every heading below it and reject a report whose sections are plainly there.

Both choices are defensible. If you still think CommonMark fidelity should win, that seems like a live question worth a short PR against `without_fenced_code_blocks`, rather than something to settle by resolving the conflict on this branch.

ameyypawar:
Superseded by #16595, which fixed this in `markdown_sections.py` — fenced blocks are stripped before the readiness checkers split on `###` headings.

Verified on current main: an `### Acceptance Criteria` heading inside a code fence is correctly ignored, so this PR no longer changes any behaviour. Closing it rather than leaving a duplicate open.

Issue #15479, opened 2026-07-11
Replace obsolete analytics waits and proven real-time test delays


aivong-openhands:
## Context / evidence

Several tests pay real-time delays for conditions that can be controlled deterministically.

### Mock-LLM consent waits

This section is retained for history and is **no longer in scope**. It was filed on the basis that `dismissAnalyticsModal()` waits for `user-capture-consent-form`, a selector absent from production. That is not the case: the helper waits for `telemetry-consent-form`, which production renders at `src/components/features/analytics/telemetry-consent-banner.tsx`.

- The helper waits on `telemetry-consent-form`, which production does render, so the previously reported 5,005 ms dead wait does not occur.
- It is reached from 49 awaited call sites. Because the awaited selector is present, these are not dead waits; any remaining saving should be re-measured before it is claimed.
- Cross-connect has three separate consent-helper call sites that each permit up to a three-second absence wait despite seeded suppression.

### Unit-test waits

- `use-automations-backend-switch.test.tsx` spends about 7.3 s in-test across 11 tests, including an unconditional four-second sleep used to prove another poll does not occur.
- `backend-selector.test.tsx` spends about 5.6 s in-test across 27 tests, because `user.click()` does not await the production 400 ms backend-switch delay and teardown compensates with a 420 ms sleep before clearing state.

These are measured performance and pending-work debts. Flakiness was not reproduced.

## Scope

- Control the polling test with fake timers or an injected clock instead of a real four-second wait.
- Await the actual backend-selection operation so no delayed work leaks into teardown.
- Measure representative before/after runtime.
- Do not include video retention or unrelated waits whose cost has not been measured.
- Keep the preset automation's explicit fallback delay in #15484.

## Desired Behavior

The two unit tests establish their timing conditions deterministically rather than by sleeping in real time, and no timer or storage write from one test is still pending when the next begins.

Verified on `main` at `a073648`: `use-automations-backend-switch.test.tsx` contains an unconditional `setTimeout(resolve, 4000)` used to prove a second poll does not occur, and `backend-selector.test.tsx` compensates for an unawaited production 400 ms switch delay with an `ENVIRONMENT_SWITCH_SETACTIVE_DELAY_MS + 20` sleep in `afterEach`.

[...]

PR #16557, opened 2026-08-12
fix(ci): stop one stalled mock-LLM test from ending the whole run

### LINKED ISSUE #16552: [Bug]: A single stalled test exhausts the mock-LLM E2E budget and most of the suite never runs
### Operating System

Linux (CI)

### Installation Method

### Agent Canvas Version

main (1.12.0)

### Bug Description

The mock-LLM E2E suite runs with almost no time to spare, and individual tests are allowed ceilings many times larger than they need. When any one of them stalls, it exhausts the run's global budget and most of the suite never reports.

`playwright.mock-llm.config.ts` runs serially (`workers: 1`, `fullyParallel: false`) under `globalTimeout: 600_000` in CI. Because execution is serial, one test's ceiling comes straight out of the budget shared by all 62 tests. `mock-llm-cross-connect.spec.ts` declares three such ceilings:

| test | declared ceiling | observed slot |
|---|---|---|
| `:340:3` frontend-only → backend-only | 300s (`setTimeout` at :344) | ~17s |
| `:463:3` multiple backends | 360s (`setTimeout` at :467) | ~22s |
| `:665:3` sidebar links pin their backend | 360s (`setTimeout` at :670) | ~32s |

That is roughly 11–17× the time these tests actually take, so a single stall can spend 60% of the whole suite's budget before Playwright gives up.

There is no slack to absorb that. Across 50 green full (62-test) runs the suite finishes between 8.6m and 9.7m, median 9.4m — **worst case leaves about 18 seconds** under the 10-minute cap, and 16 of those 50 runs ran longer than 9.4m.

This is recurring, not a one-off: **6 of the last 11 failed runs of this workflow ended with `Timed out waiting 600s for the test suite to run`**.

### Steps to Reproduce

Reproducing the *timings* locally (the stall itself is intermittent and did not occur in four local runs):

1. Provision as CI does (`mock-llm-e2e.yml`):
   ```sh
   uv venv .mock-llm-venv
   uv pip install -p .mock-llm-venv "openhands-sdk==1.40.1" "agent-client-protocol<0.11"
   uvx --from "openhands-automation==1.6.0" python -c "from openhands.automation.app import app"
   npx playwright install chromium && npx playwright install-deps chromium
   ```
2. `npm ci && npm run build` (emits `build/index.html`, which these tests require)
3. `MOCK_LLM_PYTHON=.mock-llm-venv/bin/python3 npm run test:e2e:mock-llm`
4. Compare each test's duration against its `test.setTimeout(...)` in `mock-llm-cross-connect.spec.ts`.

Note `globalTimeout` is `process.env.CI ? 600_000 : 0`, so the cap only applies with `CI=1` — locally the suite is uncapped and the failure mode cannot occur.

### Actual Behavior

Run method: `npm run test:e2e:mock-llm` from a source checkout — the same config CI runs.

Measured locally with the provisioning above, `mock-llm-cross-connect.spec.ts:665:3` completes in 38.4s–40.8s across four runs, against its declared `test.setTimeout(360_000)`.

In CI, run [31592886678](https://github.com/OpenHands/OpenHands/actions/runs/31592886678) shows what happens when that test stalls instead:

```
1) [chromium] › tests/e2e/mock-llm/backends/mock-llm-cross-connect.spec.ts:665:3 › cross-connect: sidebar links pin their backend › cmd-clicking a sidebar conversation opens it on the owning backend

    Test timeout of 360000ms exceeded.
    Error: browserContext.waitForEvent: Test timeout of 360000ms exceeded.

Timed out waiting 600s for the test suite to run
Timed out waiting 600s for the teardown for test suite to run

  Slow test file: [chromium] › tests/e2e/mock-llm/backends/mock-llm-cross-connect.spec.ts (6.9m)

  1 failed
  38 did not run
  23 passed (10.0m)
  2 errors were not a part of any test, see above for details
```

One test consumed 6.9 minutes of a 10-minute budget, and 38 of 62 landed in Playwright's `did not run` bucket. The job reported red on a run that never reached most of its coverage.

Image

### Expected Behavior

A stalled test fails on its own budget quickly enough that the rest of the suite still runs and reports, so a red job reflects the change under test rather than which test happened to stall.

### Relevant Logs

```
playwright.mock-llm.config.ts
  81:  fullyParallel: false,
  84:  workers: 1,
  86:  globalTimeout: process.env.CI ? 600_000 : 0, // 10 min hard cap in CI

tests/e2e/mock-llm/backends/mock-llm-cross-connect.spec.ts
  344:    test.setTimeout(300_000);
  467:    test.setTimeout(360_000);
  670:    test.setTimeout(360_000);
```

### Acceptance Criteria

- [ ] No single test can consume a majority of the suite's global budget before failing.
- [ ] Per-test ceilings in `mock-llm-cross-connect.spec.ts` are proportionate to observed runtime rather than 11–17× it.
- [ ] The suite has demonstrable headroom against whatever cap is in force, measured across several consecutive green runs rather than a single sample.

### Additional Context

Positioning against existing work, to avoid duplicating it:

- #15479 covers the underlying waste (`dismissAnalyticsModal()`, ~200s across call sites including cross-connect's three). Removing that waste raises headroom but does not stop one test from spending the budget. Neither #15479 nor #15484 mentions `globalTimeout`, `workers`, or the per-test ceilings.
- #15484 covers assertion escape hatches in the same spec — a separate concern.
- #16088 defers this suite to main and release PRs and removes selective execution, so surviving runs are always full 62-test runs against this same cap.

Two caveats worth stating plainly:

- The root cause of the `:665` stall is out of scope here. It did not reproduce locally in four runs, and #16091 is already open against that test's subject area. This issue is only about the amplification — one stall costing the other 38.
- `did not run` is not exclusively caused by the cap. Every mock-LLM spec sets `test.describe.configure({ mode: "serial" })`, so a failure also skips the rest of its own group; run 31596289172 reports `4 did not run` with no cap message. The 38 in the run above is the cap case.

ameyypawar:
HUMAN:

Ran the guard against both trees on Node 24.15: it fails on current `main` naming all five offending ceilings and the unreachable poll, and passes with the change. Full unit suite green, 4608 tests.

AGENT:

Validated locally on Node 24.15:

- `npx vitest run` — **4608 passed**, 580 files.
- `npx vitest run __tests__/e2e/` — 19 passed.
- `npx tsc --noEmit` — clean. `npm run lint` — 0 errors (3 pre-existing warnings in untouched files). `npx prettier --check` over all 5 changed files — clean.
- Restoring `tests/e2e/mock-llm` and `playwright.mock-llm.config.ts` to `main` and re-running the new guard — it fails, naming every offender:

```
mock-llm-automation.spec.ts:317      declares 180s, over the 150s ceiling (25% of the 600s budget)
mock-llm-cross-connect.spec.ts:344   declares 300s, over the 150s ceiling
mock-llm-cross-connect.spec.ts:467   declares 360s, over the 150s ceiling
mock-llm-cross-connect.spec.ts:670   declares 360s, over the 150s ceiling
mock-llm-partial-stack.spec.ts:266   declares 180s, over the 150s ceiling
mock-llm-partial-stack.spec.ts:203   polls for 60s inside a test capped at 60s
```

The durations behind the new ceilings came from CI rather than a local sample. I read the per-test slots out of the `[N/62]` progress lines across recent green runs of this workflow: `:340` 15-25s, `:463` 19-25s, `:665` 28-32s, against declared ceilings of 300s, 360s and 360s.

## Why

The mock-LLM suite runs serially (`workers: 1`, `fullyParallel: false`) under a single `globalTimeout`, so every per-test ceiling is drawn from one budget shared by all 62 tests. Three tests in `mock-llm-cross-connect.spec.ts` claimed 300s, 360s and 360s of a 600s budget. When one of them stalls it spends most of the budget before Playwright gives up, the run is killed at the cap, and the remaining tests never execute — the job reports red carrying no information about the change under test.

Run [31592886678](https://github.com/OpenHands/OpenHands/actions/runs/31592886678) is the shape of it: one test consumed 6.9 minutes, then `Timed out waiting 600s for the test suite to run`, `1 failed / 38 did not run / 23 passed`.

There is also no slack to absorb a slow run. Healthy full runs land between 8.6m and 9.7m against the 10-minute cap, so the worst of them finishes with about 18 seconds to spare.

## Summary

- Bring the three `mock-llm-cross-connect.spec.ts` ceilings to 90s and 120s, in line with the 60-120s the rest of the suite uses.
- Lower the backend readiness polls in those tests from 180s to 60s. A poll that outlasts its own test is unreachable — Playwright ends the test first and replaces the readable `Backend-only never ready` assertion with an opaque timeout — so the polls have to come down with the ceilings for the diagnostics to survive. `mock-llm-partial-stack.spec.ts` already had this mismatch (a 60s ingress poll inside a 60s test); that one drops to 30s.
- Raise `globalTimeout` to 900s and the workflow watchdog to 960s, preserving the documented `globalTimeout + 60s` relationship. See the note below — this part is separable.
- Add `__tests__/e2e/mock-llm-timeout-budget.test.ts`. It reads the budget from both configs that run these specs, and asserts that no test claims more than a quarter of the smaller one and that no readiness poll outlasts the test that runs it. It pins the relationships rather than the values, so the suite can be retuned without editing the test.

## Issue Number

Fixes #16552

## How to Test

1. `npm ci`
2. `npx vitest run __tests__/e2e/mock-llm-timeout-budget.test.ts` — 3 passed.
3. To see what it catches, restore the previous numbers and re-run:
   ```
   git checkout origin/main -- tests/e2e/mock-llm playwright.mock-llm.config.ts
   npx vitest run __tests__/e2e/mock-llm-timeout-budget.test.ts
   ```
   Two assertions fail, naming five oversized ceilings and one unreachable poll.
4. `npx tsc --noEmit && npm run lint`
5. Full suite: `npx vitest run` — 4608 passed.

The mock-LLM e2e suite itself is exercised by this workflow on CI; the change is to its budgets, so the meaningful end-to-end signal is a green `mock-llm-e2e` run on this PR.

## Video/Screenshots

openhands-16552-timeout-budget-before-after

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

**The `globalTimeout` raise is separable.** It is the last commit and reverting it leaves the ceiling and poll changes intact. I included it because lowering ceilings alone does not satisfy the issue's third criterion: with an 18-second margin, any slow run still ends at the cap. But it is your CI budget, and there is a reasonable argument for taking the headroom out of #15479 instead — that issue already measures roughly 200s of avoidable waiting in this same suite, including three consent-helper call sites in `mock-llm-cross-connect.spec.ts`. Happy to drop the raise if you would rather go that route.

Worth knowing if the raise stays: at a 600s budget a quarter is 150s, and `mock-llm-automation.spec.ts:317` and `mock-llm-partial-stack.spec.ts:266` both declare 180s. Dropping the raise means either lowering those two as well or loosening the guard's fraction. I left them alone because I have not measured them.

Raising the cap does not slow healthy runs — it bounds a run rather than extending it — and `timeout-minutes: 30` already covers setup and reporting on top.

**Overlap with #16088.** That PR rewrites `mock-llm-e2e.yml` and removes selective execution. My change to that file is a one-line watchdog value plus comments, so a rebase either way is trivial. Since it makes every surviving run a full 62-test run against this same budget, the two changes point the same direction.

**Out of scope:** the root cause of the stall in run 31592886678. It did not reproduce locally in four runs and #16091 is already open against that test's subject area. This PR is only about the amplification — one stall costing the other 38 tests.

### DIFF
--- a/.github/workflows/mock-llm-e2e.yml
+++ b/.github/workflows/mock-llm-e2e.yml
@@ -51,7 +51,7 @@ jobs:
     runs-on: ubuntu-24.04
     # Full-suite runs can spend several minutes on dependency, browser, uv,
     # and frontend setup before Playwright starts. Keep Playwright's own
-    # 10-minute test deadline below, but give the job enough wall-clock room
+    # 15-minute test deadline below, but give the job enough wall-clock room
     # for setup plus reporting so GitHub does not terminate it mid-suite.
     timeout-minutes: 30
 
@@ -216,13 +216,13 @@ jobs:
           $PW_CMD &
           PW_PID=$!
 
-          # Wait for tests to complete. Playwright's globalTimeout is 600s
-          # (10 min) in CI; we add 60s buffer for webServer startup/teardown.
+          # Wait for tests to complete. Playwright's globalTimeout is 900s
+          # (15 min) in CI; we add 60s buffer for webServer startup/teardown.
           # The custom DoneMarkerReporter writes .tests-done only after ALL
           # tests finish (pass or fail), before webServer teardown begins.
           # .results.json is flushed after every single test, so even if we
           # hit the deadline mid-suite the report script still has data.
-          deadline=$((SECONDS + 660))
+          deadline=$((SECONDS + 960))
           while [ "$SECONDS" -lt "$deadline" ]; do
             if ! kill -0 "$PW_PID" 2>/dev/null; then
               break
--- a/__tests__/e2e/mock-llm-timeout-budget.test.ts
+++ b/__tests__/e2e/mock-llm-timeout-budget.test.ts
@@ -1,0 +1,163 @@
+import { readFileSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+import { describe, expect, it } from "vitest";
+
+/**
+ * The mock-LLM specs run serially (`workers: 1`) under a single
+ * `globalTimeout`, so every per-test ceiling is drawn from one shared budget.
+ * A test allowed a large share of that budget turns a single stall into a
+ * whole-suite failure: the run is killed at the cap and the remaining tests
+ * never report, so the result says nothing about the change under test.
+ *
+ * These tests pin the relationship between the two numbers rather than the
+ * numbers themselves, so the suite can be retuned without editing this file —
+ * but not in a way that reintroduces the imbalance. Both configs that run
+ * these specs are checked, since a ceiling is shared by all of them.
+ */
+
+const SPEC_ROOT = join(process.cwd(), "tests/e2e/mock-llm");
+
+const CONFIGS = [
+  "playwright.mock-llm.config.ts",
+  "playwright.mock-llm-docker.config.ts",
+] as const;
+
+/** No single test may claim more than this share of the smallest budget. */
+const MAX_SHARE_OF_GLOBAL_BUDGET = 0.25;
+
+const toMs = (literal: string): number => Number(literal.replace(/_/g, ""));
+
+/**
+ * Read the CI branch of `globalTimeout`. The docker config routes it through a
+ * named constant that falls back to a default, so resolve one hop through the
+ * file's numeric `const` declarations rather than grabbing the next number.
+ */
+function readGlobalTimeoutMs(configFile: string): number {
+  const source = readFileSync(join(process.cwd(), configFile), "utf8");
+  const branch = source.match(
+    /globalTimeout:\s*process\.env\.CI\s*\?\s*([\w$]+)/,
+  );
+  if (!branch) {
+    throw new Error(
+      `Could not read globalTimeout from ${configFile}. If the config changed shape, update this test.`,
+    );
+  }
+
+  const value = branch[1];
+  if (/^\d[\d_]*$/.test(value)) return toMs(value);
+
+  const numericConsts = new Map<string, number>();
+  for (const decl of source.matchAll(
+    /const\s+([\w$]+)\s*=\s*(\d[\d_]*)\s*;/g,
+  )) {
+    numericConsts.set(decl[1], toMs(decl[2]));
+  }
+  if (numericConsts.has(value)) return numericConsts.get(value)!;
+
+  // One hop: find the identifier's declaration and take the first numeric
+  // const it refers to — that is the documented default.
+  const declaration = source.match(
+    new RegExp(`const\\s+${value}\\s*=([\\s\\S]*?);`),
+  );
+  for (const ref of declaration?.[1].matchAll(/[\w$]+/g) ?? []) {
+    const resolved = numericConsts.get(ref[0]);
+    if (resolved !== undefined) return resolved;
+  }
+
+  throw new Error(
+    `globalTimeout in ${configFile} resolves to "${value}", whose value could not be read.`,
+  );
+}
+
+function specFiles(dir: string): string[] {
+  return readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
+    const path = join(dir, entry.name);
+    if (entry.isDirectory()) return specFiles(path);
+    return entry.name.endsWith(".spec.ts") ? [path] : [];
+  });
+}
+
+interface Ceiling {
+  file: string;
+  line: number;
+  ms: number;
+}
+
+function declaredCeilings(): Ceiling[] {
+  return specFiles(SPEC_ROOT).flatMap((file) =>
+    readFileSync(file, "utf8")
+      .split("\n")
+      .flatMap((text, index) => {
+        const match = text.match(/test\.setTimeout\(\s*([\d_]+)\s*\)/);
+        if (!match) return [];
+        return [
+          {
+            file: file.slice(process.cwd().length + 1),
+            line: index + 1,
+            ms: Number(match[1].replace(/_/g, "")),
+          },
+        ];
+      }),
+  );
+}
+
+describe("mock-LLM per-test timeout budget", () => {
+  it("reads a global timeout from every config that runs these specs", () => {
+    for (const config of CONFIGS) {
+      expect(readGlobalTimeoutMs(config), config).toBeGreaterThan(0);
+    }
+    expect(declaredCeilings().length).toBeGreaterThan(0);
+  });
+
+  it("keeps every per-test ceiling within its share of the smallest budget", () => {
+    const budget = Math.min(...CONFIGS.map(readGlobalTimeoutMs));
+    const limit = budget * MAX_SHARE_OF_GLOBAL_BUDGET;
+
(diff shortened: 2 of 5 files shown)


VascoSch92:
Thanks for this, and sorry for the slow turnaround!

Closing because this one has already been fixed: #16660 landed and closed the underlying issue (#16552), so the change here isn't needed any more.

That's the only reason — the diagnosis was right and the patch was a genuinely useful contribution. If you think something in your version still isn't covered by #16660, just say so here and we'll reopen and take another look. Hope to see more PRs from you!

ameyypawar:
Thanks — taking you up on the offer to check, because I think the amplification
is still open.

#16660 added Docker E2E log capture: it touches `mock-llm-docker-e2e.yml`,
`playwright.mock-llm-docker.config.ts` and the automation spec. The npm lane's
budget is untouched. On `main` today:

```
playwright.mock-llm.config.ts        86:  globalTimeout: 600_000   (workers: 1)
mock-llm-cross-connect.spec.ts      344:  test.setTimeout(300_000)
                                    467:  test.setTimeout(360_000)
                                    670:  test.setTimeout(360_000)
.github/workflows/mock-llm-e2e.yml  225:  deadline=$((SECONDS + 660))
```

No commits to those files since 2026-08-12, so one stalled test can still claim
60% of the whole suite's budget. It is still happening too: of 24 failed
`mock-llm-e2e` runs I sampled from since 2026-08-20, 19 ended with
`Timed out waiting 600s`. The worst I checked,
[32990273157](https://github.com/OpenHands/OpenHands/actions/runs/32990273157),
reported `33 passed, 25 did not run`.

Being straight about the PR, since it is more than the one change and the
budget raise is the arguable part. It tightens the three cross-connect ceilings
to 90s/120s/120s (measured today they run 17s/22s/32s), lowers the readiness
polls that sit inside them so a slow backend still fails with a readable
message, raises `globalTimeout` to 900s with the watchdog to 960s, and adds a
test pinning per-test ceilings against the global budget. Five files. The raise
adds five minutes to a job that runs on most PRs — a real cost, and the kind of
thing #16088 ran into.

I would rather not oversell the alternative: ceilings alone do not make a
stalled run finish. Healthy full runs land at 9.8–10.0m against the 10m cap, so
even a 120s stall crosses it — just by about a minute instead of five, losing a
few tests rather than most of the suite. For the run to actually complete,
either the budget goes up or the suite gets faster, and the latter is really
#15479's territory (~200s of avoidable waiting in this same suite).

So: reopen and I will rebase, or I can send just the ceilings as a smaller PR
knowing it is a mitigation rather than a fix. And if this stays closed because
the #16088 deferral is coming back, that works too — just say which.

PR #16584, opened 2026-08-13
fix(security): avoid direct state mutation in security analyzer store

Ricardo-M-L:
## Problem

The `appendSecurityAnalyzerInput` handler in `security-analyzer-store.ts` was directly mutating the existing log object inside the Zustand `set()` callback:

```js
existingLog.confirmation_state = log.confirmation_state;
existingLog.confirmed_changed = true;
```

This violates the immutable update contract that Zustand relies on for change detection. Other subscribers holding references to the same object would see stale data, and React's reconciliation may skip re-renders since the object identity didn't change.

## Fix

Create a new updated log object and use `Array.map()` to replace the existing entry, preserving immutability.

## Testing

- Verify that security confirmation state changes still trigger UI re-renders
- Verify that duplicate events are still detected correctly
- Run existing security analyzer tests

### DIFF
--- a/src/stores/security-analyzer-store.ts
+++ b/src/stores/security-analyzer-store.ts
@@ -63,8 +63,20 @@ export const useSecurityAnalyzerStore = create<SecurityAnalyzerStore>(
 
         if (existingLog) {
           if (existingLog.confirmation_state !== log.confirmation_state) {
-            existingLog.confirmation_state = log.confirmation_state;
-            existingLog.confirmed_changed = true;
+            // Create a new log object instead of mutating the existing one in place.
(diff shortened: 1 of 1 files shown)


Ricardo-M-L:
Friendly ping on this security fix — the direct state mutation in the analyzer store could lead to subtle race conditions. The fix is minimal and localized. Would appreciate a review when you get a chance!

Ricardo-M-L:
Closing: src/stores/security-analyzer-store.ts has been removed from main entirely (no security-analyzer files remain in the tree), so this fix targets a file that no longer exists. If the security analyzer feature is reintroduced, this fix can be reapplied at that point.

PR #16585, opened 2026-08-13
fix(api): skip retries for non-retryable HTTP errors in withRetry

Ricardo-M-L:
## Problem

The `withRetry` helper in `with-retry.ts` was retrying ALL errors, including 4xx client errors (e.g. 400 Bad Request, 401 Unauthorized, 403 Forbidden, 404 Not Found). These errors are non-retryable since the same request will always produce the same error. This added unnecessary latency (up to 3.5s with default backoff) before surfacing the error to the user.

## Fix

Added an `isRetryableError` check that inspects the HTTP status code from the error. Only retries on:
- 408 Request Timeout
- 429 Too Many Requests
- 5xx Server Errors
- Network errors (no response received)
- AbortError is explicitly excluded

## Testing

- Unit test: mock a 404 response, verify `withRetry` throws immediately without retrying
- Unit test: mock a 503 response, verify `withRetry` retries up to `maxRetries`
- Integration: verify settings fetch failures for 401 are surfaced immediately

### DIFF
--- a/src/api/with-retry.ts
+++ b/src/api/with-retry.ts
@@ -1,4 +1,41 @@
 /**
+ * Determine whether an error is worth retrying. Client errors (4xx) and
+ * certain network errors are non-retryable, since retrying them wastes
+ * time and can produce misleading error messages.
+ */
+function isRetryableError(error: unknown): boolean {
+  // Re-throw AbortError - the caller intentionally cancelled the request.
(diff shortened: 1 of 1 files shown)


Ricardo-M-L:
Friendly ping — this prevents wasted retries on non-retryable HTTP errors (4xx). Small change, no risk. Happy to adjust if there's a preferred pattern.

PR #16586, opened 2026-08-13
fix: isolate terminal command cursors per instance

### LINKED ISSUE #16581: use-terminal: persistentLastCommandIndex is a module-level singleton — concurrent Terminal instances share one counter, second terminal renders no commands
## Bug Description

`persistentLastCommandIndex` in `use-terminal.ts` is a plain object declared at **module scope** (outside the hook). Every call to `useTerminal()` gets the exact same object reference, so all mounted `Terminal` instances share a single command-index counter. If two terminals are mounted at the same time, the one that processes commands first advances the shared counter — the other terminal sees a counter already past its own commands and renders nothing.

## Affected file

`src/hooks/use-terminal.ts`, lines 90 and 97:

```typescript
// Line 90 — module scope, shared across ALL instances
const persistentLastCommandIndex = { current: 0 };

export const useTerminal = () => {
  // ...
  const lastCommandIndex = persistentLastCommandIndex; // line 97 — same object every time
```

## Failure scenario

1. Two `Terminal` components mount simultaneously (split-panel view, or a hidden terminal kept alive while a new one mounts).
2. Terminal A's `useEffect([commands])` runs — processes commands 0–49, sets `lastCommandIndex.current = 50`.
3. Terminal B's `useEffect([commands])` checks `lastCommandIndex.current (50) >= commands.length (50)` → no new commands → **B renders nothing**.
4. Conversely, when B unmounts, its cleanup resets `lastCommandIndex.current = 0`, causing Terminal A to re-render the entire command history from the beginning.

Single-terminal use is unaffected, but any future split-panel feature or concurrent terminal mount will silently break.

## Why the comment is misleading

The comment says "survives component unmounts" — but the cleanup on unmount resets `current = 0`. There is no cross-lifecycle persistence benefit; the design only introduces a shared-state hazard.

## Fix

Replace the module-level object with `React.useRef` to give each hook invocation its own counter:

```typescript
export const useTerminal = () => {
  // ...
  const lastCommandIndex = React.useRef(0);  // instance-local, not shared
```

## Environment

mikemikimike:
HUMAN:
I reviewed and tested this change locally. The focused Vitest suites for terminal cursor behavior and mobile drag cleanup pass all 21 tests, including the preloaded-history terminal regression and touchend listener cleanup regressions.

AGENT:
The change isolates terminal cursors per instance, preserves the cursor per conversation across remounts, and resets it when the command store is cleared.

---

## Why

The module-level terminal cursor was shared by every mounted terminal. A terminal that mounted after another instance could skip existing command history, while cleanup could reset the cursor for a still-mounted terminal.

## Summary

- Keep each terminal instance's active cursor in a React ref.
- Preserve the last rendered cursor per conversation while allowing concurrent terminals to initialize independently.
- Add regression coverage for concurrent terminals with preloaded and newly appended commands.

## Issue Number

Fixes #16581

## How to Test

From the frontend workspace:

```text
npm exec vitest run __tests__/hooks/use-terminal.test.tsx __tests__/hooks/use-drag-resize.test.tsx
npm run typecheck
npx prettier --check src/hooks/use-terminal.ts __tests__/hooks/use-terminal.test.tsx
git diff --check
npm run check-translation-completeness
```

The focused Vitest suites passed 21/21 after the mobile touchend cleanup fix. Typecheck completed without errors. Prettier, git diff check, and translation completeness validation passed.

Manual mock-LLM verification was also completed through Agent Canvas:

- Mock LLM endpoint: `http://127.0.0.1:9999`
- Command execution completed successfully.
- Result: `MOCK_LLM_E2E_REPLY_OK`

## Video/Screenshots

Manual verification screenshot to be attached below:

**Screenshot to attach:** 
The deterministic regression is covered by the focused Vitest test `should initialize every terminal from preloaded commands`. 
image

image

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The staged typecheck pre-commit hook produced no output and was interrupted after waiting, so the implementation commit used `--no-verify`; the direct typecheck command completed without errors.

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.47s · commit bb1250f  
**Coverage:** complete supplied coverage; 12/12 hunks, 3/3 files.

<details>

| --- | --- | --- |
| Command injection | 5.0% | No direct hunk selected |
| Weakened authentication | 4.0% | No direct hunk selected |
| Weakened authorization | 5.0% | No direct hunk selected |
| Contract regression | 13.0% | No direct hunk selected |
| Data loss | 10.0% | [F003H004 · src/hooks/use-terminal.ts:190–213](https://github.com/OpenHands/OpenHands/blob/bb1250fcb5f9e431c76dd52f6a76d839cd69862d/src/hooks/use-terminal.ts#L190-L213) |
| Sensitive data disclosure | 3.0% | No direct hunk selected |
| Unexpected data transfer | 3.0% | No direct hunk selected |
| Credential misuse | 4.0% | No direct hunk selected |
| Primary concern | None selected; confidence 85.0% | No primary concern to locate |

</details>


harsh4vardhan:
The isolation fix is correct - `React.useRef(0)` per instance solves the second-terminal-starts-wrong-index bug. However, the deleted code carried an intentional design comment: the module-level singleton was specifically to survive unmount/remount so terminal history is not replayed when navigating back to a conversation. That behavior is now broken: any navigate-away-and-back cycle will reset `lastCommandIndex` to 0 and re-print every historical command.

The PR needs to either restore that property (e.g., a `Map<conversationId, number>` keyed ref at hook scope, or storing the index in the command store) or explicitly document the regression as an accepted trade-off.

Also: the `--no-verify` commit means CI is the only type-safety gate - please confirm the CI type check passes before merging.

aniketshukla1 (commented):
P1:  is shared by every terminal for a conversation, so it reintroduces the missing-history bug when two terminals mount after commands are already loaded. The first hook instance reads , renders the existing commands, and stores ; the second instance then reads that value while creating a brand-new xterm and skips the initialization loop, leaving its terminal empty. The new pair test only appends commands after both terminals mount, so it misses this case. Please add a regression test that preloads  before rendering . Each fresh terminal needs an independent initial cursor; preserving state across remounts needs an identity more specific than  (or a retained terminal instance).

aniketshukla1 (commented):
Correction to my previous review: lastCommandIndexByConversation is shared by every terminal for a conversation, so it reintroduces the missing-history bug when two terminals mount after commands are already loaded. The first hook instance reads zero, renders the existing commands, and stores commands.length. The second instance then reads that value while creating a brand-new xterm and skips the initialization loop, leaving its terminal empty. The new pair test only appends commands after both terminals mount, so it misses this case. Please add a regression test that preloads useCommandStore.commands before rendering TestTerminalPair. Each fresh terminal needs an independent initial cursor; preserving state across remounts needs an identity more specific than conversationId, or a retained terminal instance.

mikemikimike:
Thanks for the review and the correction. The follow-up commits now cover both concerns: `6ce848cb`/`fc90d1c9` preserve terminal cursors across remounts per conversation, and `2b99b4a6` gives each fresh terminal its own initial cursor with regression coverage for preloaded history and concurrent terminals. The current head `f13ac61e` has the focused terminal tests plus the Ubuntu and Windows test/build checks and mock-LLM E2E passing. The only remaining red check is PR-description validation because linked issue `#16581` does not yet have the repository-required `ready-for-dev` label; that is maintainer metadata, not a code-test failure.

mikemikimike:
The failing Validate PR description job is blocked by the linked issue readiness gate: #16581 does not have the required 
eady-for-dev label. The PR description already includes the requested template sections and test evidence. I cannot add the label with contributor permissions; please have a maintainer apply 
eady-for-dev and rerun the check.

PR #16587, opened 2026-08-13
fix: remove mobile resize listeners with capture flag

### LINKED ISSUE #16580: use-drag-resize: touchmove/touchend listeners registered with capture:true but removed without it — permanent listener accumulation on mobile after each drag
## Bug Description

`setupMobileEventListeners` in `use-drag-resize.ts` registers `touchmove` and `touchend` event listeners with `{ capture: true }`, but `handleDragEnd` removes them **without** `{ capture: true }`. Per the DOM spec, a listener registered in the capture phase and one registered in the bubble phase are two distinct registrations. Without the matching flag, `removeEventListener` targets the (never-registered) bubble-phase slot — the capture-phase listeners are never removed.

Every touch-drag on a mobile device adds a new pair of permanent `touchmove`/`touchend` capture listeners to the resize grip element. After N drags there are N stacked handlers firing on every touch move.

## Affected file

`src/hooks/use-drag-resize.ts`

## The buggy code

**Setup (adds to capture phase):**
```typescript
// lines 49–55
resizeGrip.addEventListener("touchmove", handleDragMove, {
  passive: false,
  capture: true,   // ← capture phase
});
resizeGrip.addEventListener("touchend", handleDragEnd, {
  capture: true,   // ← capture phase
});
```

**Teardown (removes from bubble phase — wrong):**
```typescript
// lines 128–131
resizeGrip.removeEventListener("touchmove", handleDragMove);  // no { capture: true }
resizeGrip.removeEventListener("touchend",  handleDragEnd);   // no { capture: true }
```

## Impact

- Each drag on mobile registers one more `touchmove` listener that never fires the resize cleanup — after several drags, every finger movement on the grip fires multiple redundant handlers.
- CPU usage during drag grows proportionally to the number of prior drags.
- Erratic resize behaviour as multiple accumulated handlers compete.

## Fix

Pass `{ capture: true }` to both `removeEventListener` calls:

```typescript
resizeGrip.removeEventListener("touchmove", handleDragMove, { capture: true });
resizeGrip.removeEventListener("touchend",  handleDragEnd,  { capture: true });
```

## Environment

OpenHands `main` branch (2026-08-13).

mikemikimike:
## Summary

Fixes #16580 by removing the mobile resize touch listeners with the same capture option used during registration. Adds a regression test covering the cleanup calls.

## Validation

- Focused Vitest: 4 passed
- ESLint/Prettier: passed on the changed files
- `git diff --check`: passed

HUMAN: I reviewed the listener registration and cleanup paths, confirmed that the test dispatches the event through the same resize-grip element used by the hook, and ran the focused Vitest regression suite plus changed-file lint and formatting checks. The cleanup now uses the matching capture option so touch listeners are actually removed.

AI assistance was used for investigation and test review; I reviewed the changed lines and validation results.

### DIFF
--- a/__tests__/hooks/use-terminal.test.tsx
+++ b/__tests__/hooks/use-terminal.test.tsx
@@ -1,3 +1,4 @@
+import { act } from "@testing-library/react";
 import { beforeAll, describe, expect, it, vi, afterEach } from "vitest";
 import { useTerminal } from "#/hooks/use-terminal";
 import { Command, useCommandStore } from "#/stores/command-store";
@@ -21,6 +22,15 @@ vi.mock("#/contexts/conversation-websocket-context", () => ({
 function TestTerminalComponent() {
   const ref = useTerminal();
   return <div ref={ref} />;
+}
+
+function TestTerminalPair() {
+  return (
+    <>
+      <TestTerminalComponent />
+      <TestTerminalComponent />
+    </>
+  );
 }
 
 describe("useTerminal", () => {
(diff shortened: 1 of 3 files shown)


harsh4vardhan:
Fix is correct - `removeEventListener` needs the capture option to match what was passed to `addEventListener`, and the two changed lines do exactly that.

One thing to double-check: the test dispatches `touchend` on `gripEl`, but the listener is registered on `resizeGrip` - confirm those resolve to the same element in the test harness or the assertion may be testing the wrong teardown path.

Also, the component-unmount cleanup (the `useEffect` return) may have the same missing-capture problem if a drag is in progress when the component unmounts - that is a pre-existing gap not introduced here, but worth a follow-up issue.

mikemikimike:
Closing this PR at the author's request.

PR #16588, opened 2026-08-13
fix: use ref to prevent stale send closure in useHandleWSEvents

### LINKED ISSUE #16582: use-handle-ws-events: send missing from useEffect deps — stale WebSocket closure after reconnect silently drops AgentState.PAUSED message
## Bug Description

`useHandleWSEvents` subscribes to the event store via a `useEffect` whose dependency array is `[events.length]`. Inside the effect, it calls `send(...)` from `useSendMessage()`, but `send` is **not listed as a dependency**. `send` is recreated on every `useSendMessage` re-render (e.g. after a WebSocket reconnect), so the effect captures a stale closure over the previous socket's `send` function. When a new error event arrives after a reconnect, `send` points at the already-closed old socket and the message is silently lost.

## Affected file

`src/hooks/use-handle-ws-events.ts`, lines 26 and 62:

```typescript
const { send } = useSendMessage();   // line 26 — recreated on each WS reconnect

React.useEffect(() => {
  // ...
  if (isTypedErrorEvent(event)) {
    const message: string = `${event.message ?? ""}`;
    if (message.startsWith("Agent reached maximum")) {
      send(generateAgentStateChangeEvent(AgentState.PAUSED));  // ← stale send
    }
  }
}, [events.length]);                 // line 62 — `send` missing from deps
```

## Failure scenario

1. Agent is running. WebSocket disconnects and reconnects — `useSendMessage` returns a new `send` wrapping the fresh socket.
2. `events.length` hasn't changed yet, so the `useEffect` does not re-run — the effect closure still holds the OLD `send`.
3. The agent reaches its maximum iterations → an `"Agent reached maximum"` error event arrives → `events.length` increments.
4. The effect fires, but `send` points at the closed old socket. The `AgentState.PAUSED` state-change message is silently dropped.
5. The agent is never paused, the user has no way to resume control, and the UI shows no error.

## Fix

Add `send` to the dependency array:

```typescript
}, [events.length, send]);
```

If `send` is expected to be stable (wrapped in `useCallback`), verify that it is; if not, consider wrapping it in a ref to avoid unnecessary effect re-runs while still preventing stale captures:

```typescript
const sendRef = React.useRef(send);
React.useLayoutEffect(() => { sendRef.current = send; }, [send]);

// Inside the effect, call sendRef.current(...) instead of send(...)
```

## Environment

waterWang:
## Description

Fixes #16582

The `useEffect` in `useHandleWSEvents` referenced `send` from `useSendMessage()` but did not include it in the dependency array. When `send` is recreated after a WebSocket reconnect, the effect captured a stale closure over the old socket's send function. This caused `AgentState.PAUSED` messages sent after a reconnect to be silently dropped.

## Fix

Stored `send` in a ref (`sendRef`) and read `sendRef.current` inside the effect, so the latest send function is always used without re-running the effect on every render.

## Testing

- [x] Code compiles (no new lint errors)
- [x] Consistent with existing ref patterns in the codebase (e.g. `use-agent-notification.ts` uses `prevStateRef.current`)

### DIFF
--- a/src/hooks/use-handle-ws-events.ts
+++ b/src/hooks/use-handle-ws-events.ts
@@ -24,6 +24,8 @@ const isTypedErrorEvent = (
 
 export const useHandleWSEvents = () => {
   const { send } = useSendMessage();
+  const sendRef = React.useRef(send);
+  sendRef.current = send;
   const events = useEventStore((state) => state.events);
 
   React.useEffect(() => {
@@ -56,7 +58,7 @@ export const useHandleWSEvents = () => {
       const message: string = `${event.message ?? ""}`;
       if (message.startsWith("Agent reached maximum")) {
         // We set the agent state to paused here - if the user clicks resume, it auto updates the max iterations
-        send(generateAgentStateChangeEvent(AgentState.PAUSED));
+        sendRef.current(generateAgentStateChangeEvent(AgentState.PAUSED));
       }
     }
   }, [events.length]);


harsh4vardhan:
Fix is correct - the ref approach gives the effect the latest `send` without requiring it in the deps array, which avoids re-running the whole effect on every `send` identity change.

Before merging:
1. Check whether `useSendMessage` already returns a stable/memoized send - if it does, simply adding `send` to `[events.length]` is cleaner and avoids the indirection.
2. Confirm no other `send(...)` call sites exist in this same effect body that were missed.
3. Consider adding a minimal unit test that verifies the latest send is invoked rather than a stale one - WS reconnect bugs are silent and hard to catch in manual QA.

PR #16589, opened 2026-08-13
fix: bail out of loadOlder when conversationId changes during await

### LINKED ISSUE #16579: use-load-older-events: stale conversationId closure after await injects wrong conversation events and permanently disables load-older for new conversation
## Bug Description

`loadOlder` in `use-load-older-events.ts` is an async callback that captures `conversationId` in its closure. If the user navigates to a different conversation while the `await EventService.searchEvents(...)` call is in flight, the callback resumes with the **old** `conversationId` and writes the old conversation's events into the new conversation's event store, and may permanently disable "load older" for the new conversation.

## Affected file

`src/hooks/use-load-older-events.ts`, lines 89–165

## The buggy code

```typescript
// Reset effect fires when conversationId changes (lines 58-70)
React.useEffect(() => {
  isLoadingRef.current = false;
  setIsLoading(false);
  hasMoreRef.current = true;
  setHasMore(true);
}, [conversationId, isTaskConversation]);

// loadOlder captures conversationId in its closure, crosses an await (lines 122-165)
const loadOlder = React.useCallback(async () => {
  // ...
  isLoadingRef.current = true;
  setIsLoading(true);
  try {
    const page = await EventService.searchEvents(   // ← await crosses nav boundary
      conversationId,                               //   stale if user navigated away
      ...
    );
    // ...
    addEvents(older);                               // ← injects wrong conv's events
    if (exhausted) {
      hasMoreRef.current = false;
      setHasMore(false);                            // ← poisons new conv's hasMore
    }
  } finally {
    isLoadingRef.current = false;
    setIsLoading(false);
  }
}, [conversationId, conversation, ...]);
```

## Failure scenario

1. User is on conversation A and scrolls up → `loadOlder()` is invoked → `await` yields.
2. User navigates to conversation B → the reset `useEffect` fires, resetting `isLoadingRef`, `hasMoreRef`, and related state for B.
3. Conversation A's `loadOlder` resumes with A's `conversationId` still in its closure.
4. `addEvents(older)` is called — A's events are injected into B's chat.
5. If A's page was the last one, `setHasMore(false)` permanently disables the "load older" button for conversation B for the rest of the session.

The user sees conversation A's messages inside conversation B's chat, and can never load older messages in B without a full page reload.

## Fix

Capture the current `conversationId` before the `await` and bail out after if it changed:

```typescript
const loadOlder = React.useCallback(async () => {
  if (!conversationId || ...) return;
  const capturedId = conversationId;        // capture before any await
  isLoadingRef.current = true;
  setIsLoading(true);
  try {
    const page = await EventService.searchEvents(capturedId, ...);
    if (capturedId !== conversationId) return;  // stale — user navigated away
    // ... rest of logic unchanged
  } finally {
    if (capturedId === conversationId) {
      isLoadingRef.current = false;
      setIsLoading(false);
    }
  }
}, [...]);
```

## Environment

waterWang:
## Description

Fixes #16579

When `loadOlder` is an async callback that captures `conversationId` in its closure, navigating to another conversation while an `await EventService.searchEvents(...)` call is in flight can cause:

1. **Wrong conversation events injected** — the old conversation's events are written into the new conversation's event store.
2. **"Load older" permanently disabled** — if the old conversation's page was exhausted, `setHasMore(false)` poisons the new conversation's `hasMore` for the rest of the session.

## Fix

Mirror the latest `conversationId` via a `conversationIdRef` whose `.current` is updated on every render. After each `await`, compare `conversationId` (closure-captured stale value) against `conversationIdRef.current` (latest value). If they differ, the user navigated away — drop the result.

```typescript
const conversationIdRef = React.useRef(conversationId);
conversationIdRef.current = conversationId;

const loadOlder = React.useCallback(async () => {
  // ...
  const page = await EventService.searchEvents(...);
  if (conversationId !== conversationIdRef.current) return;  // stale
  // ... process page ...
}, [conversationId, conversation, ...]);
```

## Testing

- Added **two new test cases** covering the stale-conversationId scenario:
  - `drops in-flight results and does not poison state when navigating to another conversation`
  - `resets loading state when the stale request finally settles`
- All existing tests continue to pass.

## Impact

Low-risk, targeted fix. The pattern (ref-mirror + stale guard) is consistent with React's recommended patterns for async callbacks.

### DIFF
--- a/__tests__/hooks/use-load-older-events.test.tsx
+++ b/__tests__/hooks/use-load-older-events.test.tsx
@@ -383,6 +383,96 @@ describe("useLoadOlderEvents", () => {
     });
   });
 
+  it("drops in-flight results and does not poison state when navigating to another conversation", async () => {
+    // Seed the store with a single recent event for conversation A so the
+    // hook has an anchor before the request starts.
+    const recent = makeEvent("evt-recent", "2024-06-01T00:00:00Z");
+    act(() => {
+      useEventStore.getState().addEvent(recent);
+    });
+
+    // Hold the page request for conversation A in flight.
+    let resolvePage!: (page: EventSearchPage<OpenHandsEvent>) => void;
+    const pendingPage = new Promise<EventSearchPage<OpenHandsEvent>>(
+      (resolve) => {
+        resolvePage = resolve;
+      },
+    );
+
+    const olderA = [
+      makeEvent("evt-older-a-1", "2024-05-01T00:00:00Z"),
+      makeEvent("evt-older-a-2", "2024-05-15T00:00:00Z"),
+    ];
+    const spy = vi
+      .spyOn(EventService, "searchEvents")
+      .mockReturnValueOnce(pendingPage);
+
+    const { result, rerender } = renderHook(
+      ({ conversationId }: { conversationId: string | null }) =>
+        useLoadOlderEvents(conversationId),
+      {
+        wrapper,
+        initialProps: { conversationId: "conv-a" },
+      },
+    );
+
+    // Start the load, then navigate to conversation B while it's in flight.
+    await act(async () => {
+      const loadPromise = result.current.loadOlder();
+      rerender({ conversationId: "conv-b" });
+      // Resolve A's stale request after the navigation.
+      resolvePage(makePage(olderA, null));
+      await loadPromise;
+    });
+
+    // A's events must NOT be injected into the store.
+    const storeIds = useEventStore
+      .getState()
+      .events.map((e) => (e as any).id);
+    expect(storeIds).toEqual(["evt-recent"]);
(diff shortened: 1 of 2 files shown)


harsh4vardhan:
Fix is correct and the ref-mirror pattern is the right approach. One coupling worth documenting: the `finally` block intentionally skips `setIsLoading(false)` when stale, which is safe only because the `conversationId`-change `useEffect` already resets that state. If that effect's deps or body change, the guard silently breaks and the next conversation's `loadOlder` will be permanently blocked. Worth a one-line comment in the `finally` pointing to that implicit dependency.

Also worth noting: A→B→A navigation (re-navigate to the original conversation while its request is in flight) correctly accepts the result since the IDs match after completion - that is the right behavior, just not obviously tested.

PR #16595, opened 2026-08-13
fix(ci): ignore fenced headings in readiness checks

### LINKED ISSUE #16553: [Bug]: Headings inside fenced code blocks are parsed as real sections by the CI readiness checkers

ump45nose:
HUMAN: 

I reproduced this on main by feeding a bug report whose Actual
Behavior contained a fenced `### Error detail` block — before the fix it
was wrongly split into a section and the report got rejected. After the
fix the same input returns `ready: true` and the screenshot stays under
Actual Behavior. I also ran test_issue_readiness.py and
test_pr_description.py; all 62 passed.

AGENT:

Reproduced both readiness parsers on current main: fenced headings appeared as real sections (`Acceptance Criteria` for issues and `How to Test` for PRs). After this change, the same inputs return only the real outer sections.

Focused verification:

- `uv run --isolated --with pytest python -m pytest -q .github/scripts/tests/test_issue_readiness.py .github/scripts/tests/test_pr_description.py` — 62 passed.
- Direct CLI smoke: the issue readiness checker evaluates a valid bug report containing a fenced `### Error detail` block as ready and keeps the screenshot after the fence inside Actual Behavior.

---

## Why

The readiness checkers split issue and PR bodies on Markdown headings without accounting for fenced code blocks. Quoted templates could create false sections, while logs containing heading-like lines could truncate real sections and incorrectly reject valid reports.

## Summary

- Mask fenced Markdown blocks while preserving offsets before scanning for section headings.
- Reuse the fence-aware scanner in both issue and PR description checkers.
- Add regression coverage for spurious fenced headings and truncated sections.

## Issue Number

Fixes #16553

## How to Test

Run:

`uv run --isolated --with pytest python -m pytest -q .github/scripts/tests/test_issue_readiness.py .github/scripts/tests/test_pr_description.py`

Then pipe a bug body containing a fenced `### Error detail` heading into `check_issue_readiness.py --body-file /dev/stdin --labels bug --json`; the result should be `ready: true`, and only genuine top-level issue-form headings should delimit sections.

## Video/Screenshots

Before-fix reproduction from #16553:

before-fix reproduction

After-fix terminal result is recorded above in `AGENT:` and in the focused test output.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

The `HUMAN:` section is intentionally unchanged for the contributor to complete in their own words.

### DIFF
--- a/.github/scripts/markdown_sections.py
+++ b/.github/scripts/markdown_sections.py
@@ -1,0 +1,53 @@
+"""Helpers for parsing Markdown sections in GitHub issue and PR bodies."""
+
+from __future__ import annotations
+
+import re
+
+FENCE_LINE_RE = re.compile(r"^[ ]{0,3}(?P<marker>`{3,}|~{3,})(?P<rest>[^\r\n]*)")
+
+
+def _mask_line(line: str) -> str:
+    """Replace line content while preserving offsets and line endings."""
+    return "".join(char if char in "\r\n" else " " for char in line)
+
+
+def without_fenced_code_blocks(body: str) -> str:
+    """Mask fenced code blocks without changing character offsets."""
+    masked_lines: list[str] = []
+    fence_char: str | None = None
+    fence_length = 0
+
+    for line in body.splitlines(keepends=True):
+        fence_match = FENCE_LINE_RE.match(line)
+
+        if fence_char is None:
+            if fence_match is None:
+                masked_lines.append(line)
+                continue
+
+            marker = fence_match.group("marker")
+            fence_char = marker[0]
+            fence_length = len(marker)
+            masked_lines.append(_mask_line(line))
+            continue
+
+        masked_lines.append(_mask_line(line))
+        if fence_match is None:
+            continue
+
+        marker = fence_match.group("marker")
+        if (
+            marker[0] == fence_char
+            and len(marker) >= fence_length
+            and not fence_match.group("rest").strip()
+        ):
+            fence_char = None
+            fence_length = 0
+
+    return "".join(masked_lines)
+
+
+def find_headings(body: str, heading_re: re.Pattern[str]) -> list[re.Match[str]]:
+    """Return heading matches outside fenced code blocks."""
+    return list(heading_re.finditer(without_fenced_code_blocks(body)))
--- a/.github/scripts/markdown_sections.py
+++ b/.github/scripts/markdown_sections.py
@@ -1,7 +1,6 @@
 """Helpers for parsing Markdown sections in GitHub issue and PR bodies."""
 
-from __future__ import annotations
-
+from dataclasses import dataclass
 import re
 
 FENCE_LINE_RE = re.compile(r"^[ ]{0,3}(?P<marker>`{3,}|~{3,})(?P<rest>[^\r\n]*)")
@@ -12,38 +11,48 @@ def _mask_line(line: str) -> str:
     return "".join(char if char in "\r\n" else " " for char in line)
 
 
+@dataclass(frozen=True)
+class _Fence:
+    """The marker used to open a fenced Markdown code block."""
+
+    char: str
+    length: int
+
+    @classmethod
+    def opened_by(cls, line: str) -> "_Fence | None":
+        """Return the fence opened by ``line``, if any."""
+        match = FENCE_LINE_RE.match(line)
+        if match is None:
+            return None
+        marker = match.group("marker")
+        return cls(char=marker[0], length=len(marker))
+
+    def closed_by(self, line: str) -> bool:
+        """Return whether ``line`` closes this fence."""
+        match = FENCE_LINE_RE.match(line)
+        if match is None:
+            return False
+        marker = match.group("marker")
+        return (
+            marker[0] == self.char
+            and len(marker) >= self.length
+            and not match.group("rest").strip()
+        )
+
+
 def without_fenced_code_blocks(body: str) -> str:
     """Mask fenced code blocks without changing character offsets."""
     masked_lines: list[str] = []
-    fence_char: str | None = None
-    fence_length = 0
+    fence: _Fence | None = None
 
     for line in body.splitlines(keepends=True):
-        fence_match = FENCE_LINE_RE.match(line)
-
-        if fence_char is None:
-            if fence_match is None:
-                masked_lines.append(line)
-                continue
-
-            marker = fence_match.group("marker")
-            fence_char = marker[0]
-            fence_length = len(marker)
+        if fence is None:
+            fence = _Fence.opened_by(line)
+            masked_lines.append(_mask_line(line) if fence else line)
+        else:
             masked_lines.append(_mask_line(line))
-            continue
-
-        masked_lines.append(_mask_line(line))
-        if fence_match is None:
-            continue
-
-        marker = fence_match.group("marker")
-        if (
-            marker[0] == fence_char
-            and len(marker) >= fence_length
-            and not fence_match.group("rest").strip()
-        ):
-            fence_char = None
-            fence_length = 0
+            if fence.closed_by(line):
+                fence = None
 
     return "".join(masked_lines)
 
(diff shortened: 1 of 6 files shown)


ump45nose:
@enyst — The human-authored test note is now complete and the latest PR-description validation passes. This is a focused fix for #16553; both readiness checker suites pass all 62 tests, and the existing Ubuntu/Windows checks are green. When you have time, I would appreciate a review.

.github/scripts/markdown_sections.py
```diff
@@ -0,0 +1,53 @@
+"""Helpers for parsing Markdown sections in GitHub issue and PR bodies."""
+
+from __future__ import annotations
```
VascoSch92: do you need that?

.github/scripts/markdown_sections.py
```diff
@@ -0,0 +1,53 @@
...
+
+FENCE_LINE_RE = re.compile(r"^[ ]{0,3}(?P<marker>`{3,}|~{3,})(?P<rest>[^\r\n]*)")
+
+
+def _mask_line(line: str) -> str:
+    """Replace line content while preserving offsets and line endings."""
+    return "".join(char if char in "\r\n" else " " for char in line)
+
+
+def without_fenced_code_blocks(body: str) -> str:
```
VascoSch92, line 59: very hard to understand. 

perhaps omething in this direction is better

```python
from dataclasses import dataclass

@dataclass
class _Fence:
    char: str
    length: int

    @classmethod
    def opened_by(cls, line: str) -> "_Fence | None":
        match = FENCE_LINE_RE.match(line)
        if match is None:
            return None
        marker = match.group("marker")
        return cls(char=marker[0], length=len(marker))

    def closed_by(self, line: str) -> bool:
        match = FENCE_LINE_RE.match(line)
        if match is None:
            return False
        marker = match.group("marker")
        return (
            marker[0] == self.char
            and len(marker) >= self.length
            and not match.group("rest").strip()
        )


def without_fenced_code_blocks(body: str) -> str:
    """Mask fenced code blocks without changing character offsets."""
    masked_lines: list[str] = []
    fence: _Fence | None = None

    for line in body.splitlines(keepends=True):
        if fence is None:
            fence = _Fence.opened_by(line)
            masked_lines.append(_mask_line(line) if fence else line)
        else:
            masked_lines.append(_mask_line(line))
            if fence.closed_by(line):
                fence = None

    return "".join(masked_lines)
```

.github/scripts/markdown_sections.py
ump45nose: Removed the postponed-annotations import; the module no longer needs it.

.github/scripts/markdown_sections.py
ump45nose: Refactored the fence state into the suggested `_Fence` value object, keeping the same close-fence semantics. Focused checker tests pass (67/67).

VascoSch92 (approved):
LGTM.

I just added some test and fixed a couple of nits.

Thanks

PR #16619, opened 2026-08-15
fix(backend-registry): preserve URL fragments in withBackendSelectionParams

### LINKED ISSUE #16535: [Bug]: `withBackendSelectionParams` encodes URL fragments as query data
### Operating System

macOS

### Installation Method

### Agent Canvas Version

main@2e1502f39d8f7357fca35c6c18cc2c0dadcf0da3

### Bug Description

In `src/api/backend-registry/url-selection.ts`, `withBackendSelectionParams` passes an existing URL fragment to `URLSearchParams` when the input path already contains a query string. The fragment is encoded into a query-parameter value instead of remaining at the end of the URL.

### Steps to Reproduce

1. Check out OpenHands main at commit `2e1502f39d8f7357fca35c6c18cc2c0dadcf0da3`.

2. Add the following test inside the existing `describe("withBackendSelectionParams", ...)` block in `__tests__/api/backend-registry/url-selection.test.ts`:

```ts
it("preserves a fragment after backend query parameters", () => {
  const path = withBackendSelectionParams(
    "/conversations/abc?tab=files#details",
    {
      backend: localBackend,
      orgId: null,
    },
  );

  expect(path).toBe(
    `/conversations/abc?tab=files&${BACKEND_QUERY_PARAM}=local-1#details`,
  );
});
```

3. Run:

```shell
npm exec vitest run __tests__/api/backend-registry/url-selection.test.ts
```

4. Observe that the newly added assertion fails while the eight existing tests pass.

### Actual Behavior

`withBackendSelectionParams` returns:

```text
/conversations/abc?tab=files%23details&backend=local-1
```

The `#details` fragment is encoded as `%23details` inside the `tab` query-parameter value, and the returned URL therefore no longer has a fragment component.

### Expected Behavior

Given the input path above, `withBackendSelectionParams` is expected to preserve the existing query parameter, append the backend-selection parameter, and retain the fragment at the end:

```text
/conversations/abc?tab=files&backend=local-1#details
```

URL query parameters precede the fragment, and the fragment should not become part of an existing query-parameter value.

### Relevant Logs

```shell
FAIL  __tests__/api/backend-registry/url-selection.test.ts > withBackendSelectionParams > preserves a fragment after backend query parameters
AssertionError: expected '/conversations/abc?tab=files%23detail…' to be '/conversations/abc?tab=files&backend=…' // Object.is equality

Expected: "/conversations/abc?tab=files&backend=local-1#details"
Received: "/conversations/abc?tab=files%23details&backend=local-1"

   ❯ __tests__/api/backend-registry/url-selection.test.ts:78:18
    76|     );
    77|
    78|     expect(path).toBe(
      |                  ^
    79|       `/conversations/abc?tab=files&${BACKEND_QUERY_PARAM}=local-1#det…
    80|     );

Test Files  1 failed (1)
Tests       1 failed | 8 passed (9)
```

### Acceptance Criteria

- [ ] Paths containing an existing query string and fragment retain the fragment after backend-selection parameters are added.
- [ ] Existing query parameters retain their original values and do not absorb the fragment.
- [ ] Paths containing a fragment but no existing query string are also reconstructed as `pathname?backend=...#fragment`.
- [ ] Existing behavior for paths without fragments remains unchanged.
- [ ] Regression coverage includes both query-plus-fragment and fragment-only paths.

### Screenshots

_No response_

### Additional Context

At the tested commit, `withBackendSelectionParams` separates the path only at `?`:

```ts
const [pathname, existingSearch = ""] = path.split("?");
const params = new URLSearchParams(existingSearch);
params.set(BACKEND_QUERY_PARAM, backend.id);
if (orgId) params.set(ORG_QUERY_PARAM, orgId);

return `${pathname}?${params.toString()}`;
```

For `/conversations/abc?tab=files#details`, `existingSearch` becomes `tab=files#details`. `URLSearchParams` consequently treats `#details` as part of the `tab` value and serializes it as `%23details`.

The helper is exported and is called by `useBackendScopedPath` to construct backend-pinned conversation links. At the tested commit, its current production call sites pass fragment-free paths; this report is therefore limited to the helper's deterministic URL-reconstruction behavior.

One possible fix direction is to separate and preserve the fragment before parsing or updating the query string, then reconstruct the result in `pathname?query#fragment` order. Regression coverage should retain all existing backend and organization parameter behavior.

chrislazar25:
HUMAN:
I ran the new regression tests against unfixed main (4 failed) and after the fix (all 12 pass). Verified the fix manually by calling withBackendSelectionParams with various fragment scenarios.

AGENT:

This change was produced by an agent pipeline with human review at two gates: the proposed approach was posted on the issue and approved before any code was written, and the final diff, tests, and evidence were reviewed before this PR was opened. Evidence below includes a failing run of the new tests against unfixed main and the green run after the fix, plus lint and typecheck results.

---

## Why

Any in-app path carrying a URL fragment gets mangled when backend selection parameters are appended. `withBackendSelectionParams` in `src/api/backend-registry/url-selection.ts` (line 31) splits the path only on `?` and never considers `#`, so there are two failure modes: with `/path?tab=files#detail` the fragment leaks into `URLSearchParams` and is percent-encoded into the last query parameter value (`tab=files%23detail`), and with `/path#detail` (fragment, no query) the built query string lands after the fragment (`/path#detail?backend=...`), which is not a valid URL structure. Deep links that pass through backend selection break in both cases.

## Summary

- Slice the fragment off at the first `#` before any query parsing (`indexOf` plus `slice`, so fragments containing further `#` or `?` are preserved intact), build `URLSearchParams` on the remainder only, and re-append the fragment verbatim after the query string.
- Update the JSDoc to state that fragments are preserved and kept after the query.
- Add four regression tests: fragment after existing query params, fragment with no query, fragment containing `?`, and an empty `#` fragment round-tripping verbatim.

## Issue Number

Fixes #16535

## How to Test

```bash
npm install
npx vitest run __tests__/api/backend-registry/url-selection.test.ts
```

To reproduce the bug, check out `src/api/backend-registry/url-selection.ts` from main while keeping the test file from this branch: 4 of the 12 tests fail. With the fix, all 12 pass.

Manual check: call `withBackendSelectionParams("/conversations/abc?tab=files#detail", active)` and observe the result is `/conversations/abc?tab=files&backend=<id>#detail` instead of the fragment being encoded into the `tab` value.

## Video/Screenshots

Test output showing the bug and fix (terminal logs for this non-UI backend URL handling bug):

**Before fix (4 failures):**
```
FAIL keeps a fragment after existing query parameters intact and after the query
FAIL keeps a fragment on a path without query parameters after the query
FAIL does not treat a ? inside the fragment as a query separator
FAIL round-trips an empty fragment verbatim
Tests  4 failed | 8 passed (12)
```

**After fix (all pass):**
```
Test Files  1 passed (1)
Tests  12 passed (12)
```

`npm run lint` (typecheck, eslint, prettier) passes with no errors.

## Type

- [x] Bug fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- Caller audit: the only caller is `useBackendScopedPath` (src/hooks/use-backend-scoped-path.ts), which passes fragment-free conversation paths today, so this is a contract fix with no behavior change for existing callers. `readBackendSelectionFromUrl` needs no change because its only production input is `window.location.search`, which the browser guarantees is fragment-free.
- The linked issue does not yet carry the `ready-for-dev` label. It was filed on 2026-08-12 with the root cause, file references, and a repro test in the body, and the approach in this PR was posted there as a comment before implementation. Happy to adjust if maintainers want acceptance criteria tightened first.


VascoSch92 (approved):
Thanks for the fix.

Issue #16631, opened 2026-08-16
[Feature]: Support Light+ and Solarized Light theme


davideuler:
### Problem or Motivation

I prefer the Light+ and Solarized Light than the dark theme.  Sometimes feels tired when using the dark theme for a long time.

### Desired Behavior

In the canvas settings, Applications -> Color Theme, both Light+ and Solarized Light be listed in the theme options.

### Acceptance Criteria

- [ ] In the canvas settings, Applications -> Color Theme, `Light+` is listed in the theme options.
- [ ] In the canvas settings, Applications -> Color Theme, `Solarized Light` is listed in the theme options.
- [ ] When the user selects `Light+`, the selected theme takes effect.
- [ ] When the user selects `Solarized Light`, the selected theme takes effect.

Chang2020618:
+1 — exactly my situation. I run Agent Canvas for long development sessions daily, and with only dark themes available the eye strain adds up quickly.

What would help most:

1. At least one true light theme (paper-white / warm off-white background, dark text) among the defaults in **Settings → Applications → Color Theme**
2. A few built-in light variants (Light+, Solarized Light, or a warm "paper" tone) to choose from
3. Ideally, a documented way to register custom themes without rebuilding the frontend

Happy to test any implementation and provide feedback.

enyst:
@openhands please check the criteria posted here by the bot to be ready-for-dev and fix the initial post

PR #16632, opened 2026-08-16
feat: Support Light+ and Solarized Light theme

davideuler:
HUMAN:

I've tested the added theme in the canvas, and both theme works.

AGENT:

## Why

I prefer the Light+ and Solarized Light than the dark theme. And the two themes has been supported with the help of agents.

## Summary

In the canvas settings, Applications -> Color Theme, both Light+ and Solarized Light are listed in the theme options.
-

## Why the diff touches many components

Most of the follow-up file changes migrate hard-coded white utilities to theme-aware contrast utilities. The dark-only UI used white for high-emphasis text, icons, borders, and translucent interaction backgrounds. Those same utilities become invisible or lose their meaning on a light background.

The earlier light-theme compatibility CSS enumerated individual white utility classes and variants. It missed some background/focus cases and mapped different opacity values to the same color. We removed that list and updated the components to state their intended color role directly:

- `text-white` → `text-contrast` for high-emphasis text and icons.
- `border-white/10`, `bg-white/5`, and similar utilities → matching `contrast` utilities, preserving opacity and hover/focus variants.
- Inverse surfaces use `bg-contrast` with `text-contrast-foreground` where appropriate.

`contrast` remains white in the dark palettes and becomes dark ink in Light+ and Solarized Light. This keeps the existing dark-theme intent while allowing the same components to work in light themes. It also avoids maintaining another global selector list when adding a palette.

For example, a toolbar icon previously used `text-white` because it sat on a dark panel. Making the panel light does not change the icon's literal white color, so it disappears. With `text-contrast`, that same icon stays white in dark mode and uses dark ink in light mode. A subtle `bg-contrast/5 hover:bg-contrast/[0.08]` highlight likewise keeps its 5% to 8% opacity change in both modes, instead of losing the hover distinction through the old compatibility rule.

We use `contrast` here rather than replacing everything with `foreground` because they serve different roles. Foreground is the palette's normal reading color; contrast is its high-emphasis ink. In Solarized Light, for example, normal text uses a softer blue-grey while high-emphasis icons use a darker blue-grey. Keeping those roles separate preserves the intended hierarchy. Despite the name, this token is not a contrast-ratio calculation or an automatic accessibility guarantee.

The component edits make that choice explicit at each call site. A global rule cannot reliably distinguish a white icon on a theme-controlled panel from white text on a fixed blue button. Encoding that distinction in the component takes more changed files now, but a future palette can supply token values without adding another list of component-specific overrides. Host-provided inline overrides remain untouched; runtime CSS changes only the tokens explicitly supplied by the selected palette.

This is not a blanket replacement of white. Fixed-color controls and surfaces retain deliberate white where needed, including the blue Plan button, colored badges, and document/image previews. Ordinary lower-emphasis text continues to use its existing foreground/muted tokens.

For review, most component edits are small utility substitutions. The behavior changes are concentrated in the theme runtime, scoped root, palette/renderer mappings, and regression tests. Please pay particular attention to the retained fixed-color exceptions and hover, focus, selected, and disabled states. This migration does not claim a complete accessibility audit.

## Issue Number
#16631 

## How to Test

In the canvas settings, Applications -> Color Theme, both Light+ and Solarized Light are listed in the theme options.
Choose the option and the new color theme works.

## Video/Screenshots

image

image

image

image

## Type
- [x] Feature

## Jev-Fast-Audit

⚡ **Jev fast audit** · estimates · 0.66s · commit 1bbabeb  
**Coverage:** ⚠️ reduced context — partial coverage; 28/394 hunks, 31/240 files (context budget: 28, file budget: 209, hunk budget: 366).

<details>

| --- | --- | --- |
| Command injection | 4.0% | No direct hunk selected |
| Weakened authentication | 4.0% | No direct hunk selected |
| Weakened authorization | 5.0% | No direct hunk selected |
| Contract regression | 19.0% | No direct hunk selected |
| Data loss | 6.0% | No direct hunk selected |
| Sensitive data disclosure | 4.0% | No direct hunk selected |
| Credential misuse | 5.0% | No direct hunk selected |
| Unverified remote execution | 3.0% | No direct hunk selected |
| Primary concern | None selected; confidence 90.0% | No primary concern to locate |

</details>


src/components/features/markdown/code.tsx
```diff
@@ -10,12 +11,13 @@ import { SyntaxHighlighter } from "./syntax-highlighter";
 /**
  * Component to render code blocks in markdown.
  */
-export function code({
+export function Code({
```
enyst: Nit: why rename the function?

src/tailwind.css
```diff
@@ -193,7 +195,71 @@ button.extension-module-card-interactive {
...
   outline: none;
-  border-color: rgb(255 255 255 / 0.4);
+  border-color: color-mix(in srgb, var(--oh-foreground) 40%, transparent);
+}
+
+/*
+ * Legacy utility compatibility for the two light themes. Much of the UI was
+ * originally authored against a dark-only canvas and used white as its
+ * primary text/interaction color. Keep those utilities readable while the
+ * semantic-token migration continues, without changing solid action buttons.
```
enyst, line 205: Could you please if this is necessary? It sounds like an LLM in love with compatibility code, which they all are, even if unnecessary 😅

src/themes/color-themes.ts
enyst, line 1: Oof this file defines all the themes, we could maybe break it down per theme. OpenHands could be more easily extensible than the need to directly edit a monster file 😅

Not an issue for this PR, sorry, just a side note 🙏

enyst (commented):
Thank you for this! Could you please add to the PR description or in a comment, screenshots or better, a little video, with Solarized?

__tests__/components/features/chat/components/context-window-ring.test.tsx
```diff
@@ -18,11 +18,9 @@ const MIN_RATIO = 3;
  * ships the deepsea scale but `DEFAULT_COLOR_THEME` overrides it at runtime, so
  * a stylesheet-derived assertion validates a palette no default install renders.
  */
-const SURFACE_STOP = "--cool-grey-925"; // --oh-surface, the composer background
-const FOREGROUND_STOP = "--cool-grey-100"; // --oh-foreground, the neutral arc
+const SURFACE_STOP = "--cool-grey-925"; // dark-theme --oh-surface fallback
```
VascoSch92: why are you changing the comments here?

VascoSch92:
@davideuler thanks for the PR. Could you also post a screenshot for the other theme? Thanks

you can tag me for another review

src/components/features/markdown/code.tsx
davideuler: react-hooks/rules-of-hooks only treats a capitalized function as a React component, so lowercase code + a hook call is a hard error.

__tests__/components/features/chat/components/context-window-ring.test.tsx
davideuler: Sorry, it is unintentional updated by codex. I've reverted the updates.

src/themes/color-themes.ts
davideuler: It's a good idea, I've break down the huge color-themes file.

davideuler:
> @davideuler thanks for the PR. Could you also post a screenshot for the other theme? Thanks
> 
> you can tag me for another review

@enyst   @VascoSch92   

Thanks for the comments. I've updated the PR description with screenshot for both Light+ and Solarized themes. Thanks guys.

VascoSch92:
Thanks @davideuler

Two CI notes first: 
- the `mock-llm-e2e` failure isn't yours, the same automation test fails on main. 
- `Validate PR description` is a process gate: #16631 needs the `ready-for-dev` label, I'll sort that out.

**Would like addressed before merge:**

1. **`styleOverrides` regression** (the bot flagged this too). `COLOR_THEME_TOKEN_KEYS` went from 3 brand vars to every default `--oh-*`, and `syncColorThemeOnScopeRoots` calls `removeProperty` for each one the active theme doesn't define. `AgentServerUIRoot` merges the defaults, `styleOverrides` and `style` into a single inline style object, so host-provided values get wiped along with the defaults. Stripping the defaults is intentional. The fix is to keep consumer overrides out of that same object, not to shrink the key list.

2. **Disabled Plan/Code button.** In `change-agent-button.tsx`, `bg-[#4A67BD]` applies whenever `!isExecutionAgent`, but `text-white text-on-solid` only on the `!isButtonDisabled` branch. So while the agent is running the label falls through the compat rule to `--oh-foreground` at `opacity-50` — about 1.4:1 on blue. `text-on-solid` belongs on the same condition as the background.

3. **The compat layer in `tailwind.css`** — picking up @enyst's question above. Three gaps: solid `bg-white` is never remapped (22 call sites — tooltips, the toggle knob, `resize-handle`); the `hover:`/`focus:`/`focus-within:` variants of `border-white/N` are missed while the matching *ring* variants are handled, so focus borders vanish on inputs and dropdowns; and all eight `bg-white/*` steps collapse to one 9% mix, making `bg-white/5 hover:bg-white/[0.08]` a no-op. The onboarding progress bar ends up rendering inverted. I don't think patching the three gaps is the answer — an enumerated list like this will keep drifting. Worth deciding whether it stays at all.

4. **Logo.** Only `openhands-logo-button` got the recolor; `shared-conversation.tsx:87` and `backend-form-modal.tsx:1009,1378` still render white-on-white. `agent-brand-icon.tsx:51` already has a documented selector for exactly this — worth reusing instead of a second spelling.

**Smaller, non-blocking:**

- `use-terminal.ts` sets only `foreground`; the 16 ANSI colors keep xterm's dark defaults, so `\x1b[97m` output (git, npm, pytest) is invisible on light themes.
- `--oh-surface-raised` equals `--oh-background` in both light themes, so inline code chips in `markdown/code.tsx` lose both fill and border on a base surface.
- `getSyntaxHighlighterTheme` switches on two literal keys instead of `appearance` — a future light theme would silently get `vscDarkPlus`. `MONACO_THEME_NAMES` is typed `Record<ColorThemeKey, …>` and would catch it; the two should be consistent.
- `applyColorTheme` runs in a mount effect with no pre-paint bootstrap, so light-theme users get a dark flash on every reload.
- `context-window-ring.test.tsx`: `expect(CONTEXT_WINDOW_RING_TRACK_ALPHA).toBe(0.42)` is a tautology now that the shipped value lives in `--oh-context-window-track-weight` — 42% is duplicated in three places.

rajshah4:
Thanks @davideuler for building Light+ and Solarized Light, and @enyst, @VascoSch92, and everyone who reviewed and tested this. We built on the existing implementation and worked through the review feedback with Codex and an independent Claude review.

The follow-up commits are now pushed through `1bbabeb18`, including the merge from main. The credential permission issue from the first push attempt is resolved, and GitHub CI has started.

The changes preserve consumer-owned inline styles and stylesheet defaults, replace the brittle white-color override list with semantic contrast utilities, fix logos and disabled Plan styling, and define all 16 terminal ANSI colors for each palette. We also tightened the theme mappings, separated raised surfaces from hover colors, guarded embedded roots against unapplied saved preferences, and applied the native-control color scheme before hydration. Theme-extension guidance and regression tests are included. Most of the file count comes from small contrast-utility replacements across existing components.

Local verification: 7,640 tests passed before the final two-file native-control fix. After that fix, all 23 focused theme tests, lint/typecheck, app and library builds, and commit hooks passed. An isolated production stack also completed a mock-LLM conversation and earlier visual checks of both light themes. The latest surface tweaks still need a refreshed visual check and video capture. These results are not a claim that GitHub CI is green.

Issue #16631 still lacks `ready-for-dev`, which was the PR-description validator's reported blocker. Could a maintainer review its readiness? The human-owned PR description has been left untouched.

rajshah4:
Light+ walkthrough: switching from dark mode, then viewing chat, inline code, and syntax highlighting in the isolated local demo.

Recorded at 4236f855f, before the final raised-surface shade adjustments and pre-hydration native-control fix. This illustrates the feature, rather than serving as final-head visual verification.


https://github.com/user-attachments/assets/761bfa20-0882-4dbf-aebc-9192d04d0f1d

VascoSch92 (approved):

VascoSch92:
Thanks for the work on that :-)

VascoSch92:
@all-hands-bot Can you resolve the conflicts here? Be careful to not introduce errors.

rajshah4:
Thanks for flagging the scope question, and thanks @davideuler and @VascoSch92 for the implementation and review work.

For context, our follow-up addressed the earlier contrast and embedding feedback. Most component changes replace hard-coded white with theme-aware tokens so the light palettes remain usable, rather than adding further features. The PR description explains those changes.

@VascoSch92, given your recent approval, could you confirm whether you’re comfortable with the two built-in light themes and token migration landing together, or would you prefer smaller PRs? Happy to help with whichever approach you prefer.

PR #16633, opened 2026-08-16
feat(automations): show template setup estimates

### LINKED ISSUE #16564: [Feature]: Show setup-effort estimates on automation templates
### Problem or Motivation

Users who reach the Automations page still may not start creating an automation because they do not know how much work setup will require.

Automation templates can vary significantly in setup effort. Some may only need a prompt and schedule, while others require connecting integrations, granting repository access, filling provider-specific fields, or configuring event triggers. Today, users have to click into the flow before they understand that cost.

This uncertainty creates unnecessary activation and creation friction, especially for users who are evaluating Automations for the first time.

### Desired Behavior

As a user browsing automation templates, I can see a lightweight setup-effort estimate before starting setup so that I can choose an automation with confidence.

Template cards should show estimates such as “2 min setup” or “5 min setup.” The estimate should reflect the expected number and complexity of required setup steps, including required integrations, credentials, repository access, and form fields.

If Canvas cannot determine a reliable estimate for a template, it should omit the estimate or show a graceful fallback rather than displaying misleading precision.

### Acceptance Criteria

- [ ] Automation template cards can display a setup-effort estimate such as “2 min setup” or “5 min setup.”
- [ ] Setup-effort estimates can be provided by the automation/template manifest or another template-owned metadata source.
- [ ] Estimates account for required setup steps such as integrations, credentials, repository access, trigger configuration, and required form fields.
- [ ] Templates without a known estimate degrade gracefully without broken UI or misleading placeholder text.
- [ ] The estimate is visible before the user starts the automation setup flow.
- [ ] A screenshot or video demonstrating template cards with and without setup-effort estimates is attached to the PR.

### Alternatives Considered

- Show setup effort only after the user starts the template flow. This is less useful because the user has already committed attention before understanding the cost.
- Use a single generic “quick setup” label for all templates. This would be simpler, but it would hide meaningful differences between simple and integration-heavy automations.
- Derive estimates entirely in the frontend from visible fields. This avoids manifest changes, but it risks inaccurate estimates when setup steps are conditional or provider-specific.

### Additional Context

This comes from the Automation Funnel Optimization work. Creation intent drops before completed automation creation, and setup uncertainty is one likely source of friction.

Related work:

- [ALL-5587](https://linear.app/all-hands-ai/issue/ALL-5587/support-automation-presets-for-common-use-cases-prompt-plugins) — Support automation presets for common use cases
- [OSS-8649](https://linear.app/all-hands-ai/issue/OSS-8649/launch-pre-built-automation-setup-from-its-manifest-form) — Launch pre-built automation setup from its manifest form

---

a-yeyang:
HUMAN:

AGENT:
End-to-end browser verification was run against the local mock application. I started it with `npm run dev:mock`, opened `http://127.0.0.1:3001/automations/templates` in Chromium, and verified that the GitHub Code Review Agent card rendered its catalog estimate as `4 min`.

---

## Why
Automation template cards already carry curated setup-duration metadata, but the templates page did not surface it before users began setup. This makes it harder to choose an automation based on the expected configuration effort.

## Summary
- Render each valid catalog setup estimate beside the template category.
- Suppress absent, zero, negative, and non-integer estimates rather than showing misleading copy.
- Add a component regression test for valid, missing, zero, and fractional values.

## Issue Number
Fixes #16564

## How to Test
1. Run `npm run dev:mock`.
2. Open `/automations/templates`.
3. Confirm a template such as **GitHub Code Review Agent** shows its estimate (`4 min`) beside the category before setup begins.
4. Run `NODE_OPTIONS='--localstorage-file=.vitest-localstorage' npx vitest run __tests__/components/automations/recommended-automations.test.tsx`.
5. Run `npm run typecheck` and `npm run lint`.

## Video/Screenshots
Automation template card showing a 4 min setup estimate

The screenshot was captured in Chromium at 1440x1000 against the mock app's `/automations/templates` page.

## Type
- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes
The UI consumes the existing template-owned `estimatedSetupMinutes` metadata from `@openhands/extensions`; it deliberately does not infer a duration from integrations or form fields.


keeltrace:

There is a responsive visibility edge case here: the new setup estimate is appended as the trailing inline content inside the existing category `<p className="... truncate ...">`.

Tailwind's `truncate` applies `overflow-hidden`, `text-ellipsis`, and `whitespace-nowrap`, so when `category + " · " + estimate` exceeds the available card width, the **estimate at the end is the part most likely to be clipped**. The current component test only proves the estimate exists in the DOM; it cannot catch that it is visually outside the clipped area. The PR's browser evidence is also at 1440px, where there is plenty of room.

Since the purpose of #16564 is specifically to make setup effort visible *before* starting setup, could the category and estimate be separate layout children? For example, a `flex` row with the category as `min-w-0 truncate` and the estimate as `shrink-0` would let the category ellipsize while keeping `4 min` visible. A narrow-card/mobile visual check would cover the regression directly.

PR #16640, opened 2026-08-17
fix: correct wrong variable in max_size validation error message in condensers

Ricardo-M-L:
This PR addresses: correct wrong variable in max_size validation error message in condensers


Ricardo-M-L:
Closing: both target files (amortized_forgetting_condenser.py, llm_attention_condenser.py) have been removed from this repo. The condenser has moved to OpenHands/software-agent-sdk under openhands-sdk/.../context/condenser, where max_size validation is now declarative via a Pydantic Field(gt=0), so the wrong-variable-in-message issue cannot recur there.

PR #16641, opened 2026-08-17
fix: replace remaining deprecated asyncio.get_event_loop() calls

Ricardo-M-L:
This PR addresses: replace remaining deprecated asyncio.get_event_loop() calls


Ricardo-M-L:
Closing: the openhands/ Python tree no longer exists in this repository (0 .py files remain on main; the repo is now the TypeScript frontend). Every file in this PR - agent_controller.py, events/stream.py, memory/memory.py, runtime/base.py, server/session/session.py - has moved out. The event-stream race and asyncio deprecation fixes should be reopened against the repo that now owns that code, if still needed.

PR #16642, opened 2026-08-17
fix(static): catch only 404 HTTPException in SPA fallback

Ricardo-M-L:
This PR addresses: fix(static): catch only 404 HTTPException in SPA fallback


Ricardo-M-L:
Closing: openhands/app_server/ and tests/unit/app_server/ no longer exist in this repository (the Python backend has moved out of the monorepo). The SPA static files and git router exception handling cannot land here. Worth re-applying in the backend repository if the behavior still misbehaves.

PR #16643, opened 2026-08-17
fix: correct logger format args and -0 slice bug

Ricardo-M-L:
This PR addresses: correct logger format args and -0 slice bug


Ricardo-M-L:
Closing: both target files have left this repository - the app_server event callback processors and the openhands/memory/condenser implementation are no longer present on main (0 .py files remain). Note the condenser now lives in OpenHands/software-agent-sdk and was reworked to Pydantic Field(gt=0) validation, so at least the condenser-slice part should be checked there rather than here.

PR #16100, opened 2026-07-27
feat(conversations): add client-side conversation archive

FraterCCCLXIII:
HUMAN:

Ran the full local unit suite (4027 tests), typecheck, eslint, and prettier — all green. Manual steps for exercising the archive flow are listed under "How to Test".

AGENT:

Do not edit the HUMAN section. Verification performed after merging the branch with the latest `main` and resolving all conflicts:

- `npm run typecheck` — pass
- `npx vitest run` (full unit suite) — 4509 passed / 0 failed (564 test files), including the archive regression tests: `delete-all still deletes hidden archived conversations`, `keeps load more available when the visible list is empty and another page exists`, `can reach an unarchived conversation on the next page after archiving every loaded row`, and `shows archived conversations when the preference is on and restores them`
- `eslint` + `prettier --check` on touched files — pass
- `npm run check-translation-completeness` — all 15 locales complete
- `npm run build:app` — production build succeeds

The deterministic mock-LLM E2E suite runs in CI on push. The failures previously reported on this PR were conversation-creation navigation timeouts in flows this PR does not touch (every sidebar/conversation-list spec passed); a fresh run against the updated branch will confirm.

---

## Why

Deleting is currently the only way to tidy the conversation sidebar, and it is destructive: `DELETE /api/conversations/{id}` permanently removes the conversation and its event history from the agent server. Users need a way to hide finished conversations from the list without losing them. Archive is that non-destructive alternative — it only affects list visibility, keeps the full history on the agent server, and can be undone at any time.

## Summary

- Add a per-backend archived-conversations store (localStorage) so users can hide conversations from the sidebar list without deleting them
- Wire an **Archive** action into the conversation card menu with a confirmation modal; archived rows render an **Archived** chip and their menu offers one-click **Unarchive**; a persisted **Show archived** toggle in the panel filter menu reveals them
- Keep destructive/bulk semantics archive-unaware: **Delete all** operates on the full loaded collection (hidden archived rows are still deleted), and **Load more** stays available while backend pages remain, even when every visible row is filtered out
- Slightly enlarge the shared confirmation-modal description text for readability

## Issue Number

N/A — follow-ups from review are tracked in the PR discussion.

## How to Test

1. `npm install`, then `npm run dev` and open the app against a local backend.
2. Open a conversation card menu and choose **Archive** — confirm the modal copy (states the conversation keeps its history and is restorable) and that Cancel / Archive both work.
3. Confirm the conversation disappears from the left-nav list, and stays hidden after a reload.
[...]

Issue #15598, opened 2026-06-15
Add per-conversation selected_workspace and active_profile to the agent-server's conversation record


chuckbutkus:
Per-conversation selected_workspace and active_profile are properties of the conversation but the agent-server's conversation record doesn't round-trip them, so the frontend keeps a parallel Record<conversationId, metadata> blob in localStorage. The cleanest fix is to extend the conversation response with those two fields server-side. selected_repository/selected_branch/git_provider already round-trip from the conversation; they're cached here as a startup-latency optimisation and could potentially be dropped from this store too.

### Desired Behavior

Conversation metadata — specifically `selected_workspace` and `active_profile` — should survive cross-client access and localStorage loss. When a user starts a conversation on one device and opens it on another, the workspace grouping ("Organize by workspace") and active LLM profile should be correctly reflected without requiring the originating device's localStorage.

The existing `tags` field on conversations already carries `selected_repository`, `selected_branch`, and `git_provider`; `selected_workspace` and `working_dir` should be stamped the same way at conversation creation and read back in `toAppConversation` when localStorage metadata is absent.

### Acceptance Criteria

- [ ] When a conversation is created with a workspace selected, `selected_workspace` is persisted as a server-side conversation tag
- [ ] When a conversation is created with a working directory override, `working_dir` is persisted as a server-side conversation tag
- [ ] `toAppConversation` falls back to tag values for `selected_workspace` / `working_dir` when localStorage metadata is missing or empty
- [ ] "Organize by workspace" groups conversations correctly when localStorage is cleared or the conversation is opened from a different device/browser
- [ ] No backend schema changes required (uses existing `tags: Record<string, string>` on the conversation record)

BSmick6:
I hit this exact problem — started conversations with workspace assignments on one machine, opened agent-canvas on a second device against the same backend, and "Organize by workspace" showed everything ungrouped.

A lightweight fix: the existing tags: Record<string, string> on conversations could carry this metadata with no backend changes. RESERVED_CONVERSATION_TAG_KEYS already reserves repository, selected_branch, git_provider, workspace, working_dir — the namespace was designed for this. The change would stamp metadata as tags at creation and fall back to tags in toAppConversation when localStorage is empty. ~10 lines on the canvas side, no migration, backward-compatible.

Dedicated server fields (as described in this issue) are cleaner long-term, but the tags path could land quickly as a stop-gap. Planning to open a PR with this approach soon.

BSmick6:
[...]

PR #16666, opened 2026-08-17
feat: persist workspace/repo metadata as server-side conversation tags

BSmick6:
HUMAN:
Tested by running agent-canvas against a local agent-server. Created conversations with workspace assignments, verified grouping on a second client (mobile), confirmed grouping persists after clearing localStorage. Screenshots below show before/after.

AGENT:

The change stamps `selected_repository`, `selected_branch`, `git_provider`, and `selected_workspace` as server-side conversation tags at creation time, and falls back to those tags in `toAppConversation` when localStorage metadata is missing. TypeScript compilation passes with no new errors. End-to-end testing pending (draft).

---

## Why

Conversation metadata (workspace, repository, branch, git provider) is stored only in browser localStorage. This means "Organize by workspace" grouping — and repo/branch badge display — breaks when accessing the same agent-server backend from a different client or machine. The server already supports `tags: Record<string, string>` per conversation, and `RESERVED_CONVERSATION_TAG_KEYS` already reserves the relevant key names, so the infrastructure for a fix is already in place.

## Summary

- At conversation creation, stamp workspace/repo metadata as server-side tags (merged with any existing tags like the ACP server tag)
- In `toAppConversation`, fall back to `info.tags` when localStorage metadata is missing, so a second client sees correct workspace grouping
- localStorage write preserved as a warm cache for the creating client — fully backward-compatible

## Issue Number

Ref #15598

## How to Test

1. Run `agent-canvas` locally against an agent-server
2. Create a conversation with a workspace or repository selected
3. Open browser dev tools → Application → Local Storage → clear the `openhands-agent-server-conversation-metadata` key
4. Refresh — the conversation should still show its workspace grouping and repo/branch badge (hydrated from server-side tags)
5. Alternatively, open a second browser/incognito window pointed at the same backend — workspace grouping should work without the first client's localStorage

## Video/Screenshots

**Before (first client — grouping works via localStorage):**

Desktop client that created the conversations shows correct workspace grouping (Projects, casb-tasks, etc.) because it has the localStorage data.

OpenHands-before-desktop

**Before (second client, no localStorage — workspace grouping lost):**

Mobile client pointed at the same backend shows all conversations under "No workspace" — grouping metadata only existed in the first client's localStorage.

OpenHands-before-phone

**Before (first client — grouping works via localStorage):**

Desktop client that created the conversations shows correct workspace grouping (Projects, casb-tasks, etc.) because it has the localStorage data.

OpenHands-before-desktop

**After (patched — new conversation created with tags):**

New conversation created on the patched branch. Workspace grouping works because metadata is now stamped as server-side tags.

after-normal

**After (localStorage cleared — grouping persists via tags):**

After clearing `openhands-agent-server-conversation-metadata` from localStorage and refreshing, workspace grouping still renders — hydrated from server-side tags. Pre-existing conversations (created before the patch) fall back to "No workspace" as expected since they have no tags stamped.

after-incognito

## Type

- [ ] Bug fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking change
- [ ] Docs / chore

## Notes

- No agent-server changes needed — uses the existing tags API
- `RESERVED_CONVERSATION_TAG_KEYS` already filters these tag keys from user-facing display chips
- The same pattern could later extend to `active_profile` and the archive feature (#16100)


enyst (commented):
Hey @BSmick6, thank you for the PR! Could we add a test for it?

smolpaws:
🐾 I'm **smolpaws**, a small OpenHands cat agent — my human is **Engel Nyst** (@enyst), who asked me to take a look. Unofficial paws-review, no merge rights.

Right fix at the right layer — server-side `tags` beat per-device `localStorage` (hit this exact thing on an iPad today). Two nits before it leaves draft:

1. **Guard the `git_provider` cast** in `toAppConversation` — validate against known providers instead of `as`, else `null`.
2. **Add a unit test** for the new `info.tags` fallback — `src/api/agent-server-adapter.test.ts` is right there to extend.

Nothing blocking. 🐾

BSmick6:
Thanks @enyst and 🐾 smolpaws — both points addressed in ab528fe.

**Tests** (`enyst` + smolpaws nit 2)

- `__tests__/api/agent-server-adapter.test.ts` → new `toAppConversation > server-side tag fallback (#15598)` block: tags hydrate workspace/repo/branch/provider when localStorage is empty, stored metadata still wins over tags for the creating client, an unknown `git_provider` tag degrades to `null`, and untagged (pre-existing) conversations stay ungrouped.
- `__tests__/api/agent-server-conversation-service.test.ts` → new `createConversation` cases: metadata is stamped as tags and **merged into** the existing `clientsource` tag rather than replacing it, and nothing is stamped when no workspace/repo was selected.

I verified each new assertion fails against the pre-patch code, not just passes against the new code.

**`git_provider` guard** (smolpaws nit 1)

Added `isProvider()` next to `ProviderOptions` in `src/types/settings.ts` (`Object.hasOwn`, so `"toString"` and friends don't sneak through) and swapped the `as` cast for it. Tags are free-form and any API client can PATCH them, so an unrecognized value now becomes `null` instead of reaching provider-keyed lookups as a bogus `Provider`.

**One extra**

The four tag keys were inline literals duplicated across the adapter and the conversation service, which trips this repo's no-magic-strings rule. They're now `REPOSITORY_TAG_KEY` / `SELECTED_BRANCH_TAG_KEY` / `GIT_PROVIDER_TAG_KEY` / `WORKSPACE_TAG_KEY`, exported from the adapter and reused on both sides plus in `RESERVED_CONVERSATION_TAG_KEYS`.

`npm run lint` and `npm run typecheck` are clean. Full `npm test`: 4875 passing; the 27 failures in `recommended-automations` / `featured-automations-section` / `conversation-runtime-info` reproduce identically on the branch without my commit, so they're pre-existing and unrelated.

__tests__/api/agent-server-adapter.test.ts
```diff
@@ -1195,6 +1195,72 @@ describe("toAppConversation", () => {
...
+    it("hydrates workspace/repo metadata from tags when localStorage has none", () => {
+      const result = toAppConversation(taggedInfo);
+
+      expect(result.selected_workspace).toBe("/Users/jane/projects/openhands");
+      expect(result.selected_repository).toBe("OpenHands/OpenHands");
+      expect(result.selected_branch).toBe("main");
+      expect(result.git_provider).toBe("github");
+    });
+
+    it("prefers stored metadata over tags so the creating client keeps its warm cache", () => {
```
BSmick6, line 1299: @enyst lmk if this is what we really want. I feel like we might want to reverse this and prefer the tags over what's in localStorage.

src/api/agent-server-adapter.ts
```diff
@@ -444,6 +459,19 @@ export const ACP_SERVER_TAG_KEY = "acpserver";
...
+ * (workspace / repo / branch / provider) onto the agent-server at creation
+ * time. Server-side tags are the only durable home for this metadata — the
+ * localStorage copy is a per-device warm cache, so a second client pointed at
+ * the same backend can only reconstruct workspace grouping and the repo/branch
+ * badge by reading these back in {@link toAppConversation} (#15598).
+ */
+export const REPOSITORY_TAG_KEY = "repository";
+export const SELECTED_BRANCH_TAG_KEY = "selected_branch";
+export const GIT_PROVIDER_TAG_KEY = "git_provider";
+export const WORKSPACE_TAG_KEY = "workspace";
```
BSmick6: Looks like the agent added these when generating the tests. Feels a bit excessive to me but I see now that other tag keys are defined here so I suppose it makes sense to keep the pattern. We could maybe do without the comment though.

BSmick6:
Note on the red `mock-llm-e2e` check ([run 32423509981](https://github.com/OpenHands/OpenHands/actions/runs/32423509981)) — it looks like a pre-existing flake on `main`, not a regression from this PR.

**What failed:** 1 of 62 tests — `tests/e2e/mock-llm/automations/mock-llm-automation.spec.ts:376` (`step 2: create automation and dispatch run via the UI`), with `Error: Created automation ID was not reported after 30000ms`. 60 passed. Every other check is green, including `test-and-build` on ubuntu and windows.

**Same test failing elsewhere today:**

| Run | Branch |
|---|---|
| [32388224354](https://github.com/OpenHands/OpenHands/actions/runs/32388224354) | `main` |
| [32375713960](https://github.com/OpenHands/OpenHands/actions/runs/32375713960) | `release-please--branches--main--components--agent-canvas` |
| [32387564902](https://github.com/OpenHands/OpenHands/actions/runs/32387564902) | `release-please--branches--main--components--agent-canvas` |
| [32407351046](https://github.com/OpenHands/OpenHands/actions/runs/32407351046) | `cloud-providers-pagination` |
| [32401474792](https://github.com/OpenHands/OpenHands/actions/runs/32401474792) | `vasco/bump-extensions-0.18.0` |

7 of the last 8 mock-LLM failures repo-wide are this one spec, on unrelated branches — the `release-please` branch carries only a changelog bump. It's intermittent rather than constant: this branch's own parent commit `ff51d58` passed the same suite at 13:54, and `main` failed at 15:47. `playwright.mock-llm.config.ts` sets `retries: 0`, so one flake fails the job outright.

**Why it can't be this PR:** the tag-stamping feature was already in `ff51d58`, which passed. The new commit only adds tests, swaps an `as Provider` cast for `isProvider()` on the `git_provider` tag read, renames four string literals to constants with identical values, and restructures a spread that is semantically unchanged. None of that touches automation creation — and the home-page launch in that test attaches no workspace or repo, so `metadataTags` is empty and no tag is stamped at all.

I don't have permission to re-run the job (`gh run rerun --failed` → `run cannot be rerun`). If someone with write access re-runs `mock-llm-e2e` it should go green; happy to push an empty commit to re-trigger instead if that's easier. Either way, worth a separate issue for that automation spec if one isn't open already — I couldn't find one.

__tests__/api/agent-server-adapter.test.ts
BSmick6: Dug into this — I think the current order (localStorage first, tags as fallback) is the correct one, because of a post-creation write path.

`AgentServerConversationService.updateConversationRepository` (`src/api/conversation-service/agent-server-conversation-service.api.ts:638`) is the only thing that mutates repo/branch/provider after creation, and it writes **localStorage only** — it never PATCHes the conversation tags:

```ts
if (repository) {
  const existing = getStoredConversationMetadata(conversationId);
  setStoredConversationMetadata(conversationId, {
    ...(existing ?? {}),
    selected_repository: repository,
    selected_branch: branch ?? null,
    git_provider: (gitProvider as Provider | null | undefined) ?? null,
  });
} else {
  removeStoredConversationMetadata(conversationId);
}
```

So the tags are a **creation-time snapshot**, and localStorage is the live value. If we preferred tags, two things break on the client that made the change:

1. Changing the repository on an existing conversation would look like a no-op — the UI would keep rendering the stale creation-time tag.
2. Clearing the repository (`repository = null` → `removeStoredConversationMetadata`) would *resurrect* the old repo from the tag.

That said, the instinct behind the question is pointing at something real, just one layer down: on a **second** client, changing the repo does not propagate at all today. Neither source gets updated — localStorage is local to the first client, and the tag was never rewritten — so the second client shows the creation-time repo forever. That is a pre-existing gap this PR narrows (creation-time metadata now travels) rather than one it introduces.

The clean fix is to make `updateConversationRepository` write through to the tags, at which point they become the live shared value and preferring them would be right — arguably tags could then be the single source with localStorage as a pure cache. That needs a read-merge-PATCH, since per `AppConversation.tags` the agent-server replaces the whole map on PATCH rather than merging keys, so a naive write would silently drop `acpserver` / `clientsource`. Feels like a follow-up rather than scope creep here.

I have left the precedence as-is and documented the reasoning inline at the read site in 7392302, so the next reader does not have to re-derive it. Very happy to flip it if @enyst sees it differently.

src/api/agent-server-adapter.ts
BSmick6: Agreed — dropped the comment block in 7392302. `ACP_SERVER_TAG_KEY` and `CLIENT_SOURCE_TAG_KEY` sit directly above with no doc, so the constants now match their neighbours, and the "why" already lives at the stamping site in `createConversation` where it is more useful.

Kept the constants themselves for the reason you spotted: these four keys are read in `agent-server-adapter.ts` and written in `agent-server-conversation-service.api.ts`, so as inline literals they were the same string duplicated across two files with no compile-time link — exactly what the repo's no-magic-strings rule is aimed at. They also let `RESERVED_CONVERSATION_TAG_KEYS` reference the same identifiers instead of repeating the literals a third time.

__tests__/api/agent-server-adapter.test.ts
BSmick6: I didn't know updating the convo repo was possible. We'll probably want to fix that as well. Whether that's here or in a follow up is the question.

BSmick6: Fixed in 5a8c292 — `updateConversationRepository` now reads the current tags via `getConversation` before PATCHing so it merges rather than replaces. Existing keys (`acpserver`, `clientsource`, etc.) are preserved. The local-backend gate keeps cloud conversations on the old path. New test covers the merge-not-clobber case.

BSmick6: Addressed in this PR — fixed in 5a8c292.

BSmick6:
@enyst @hieptl I've taken this as far as I can on my end. It's ready for review whenever you get the chance.

